PR #480, opened 2018-10-26
[Combobox] Add tests for arrow keys and clean up test suite

solonaarmstrong:
Codecov for [this PR](https://github.com/Shopify/polaris-react/pull/477) is failing, so I added some more tests.

As per https://github.com/Shopify/web-foundation/blob/master/Best%20practices/React/Testing.md#split-test-files I've split this into multiple test files because it was getting unruly.


src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
```diff
@@ -50,16 +52,16 @@ describe('<ComboBox/>', () => {
...
     it('passes options to OptionList', () => {
-      const comboBox = mountWithAppProvider(
-        <ComboBox
-          options={options}
-          selected={[]}
-          textField={renderTextField()}
-          onSelect={noop}
-        />,
-      );
+      const comboBox = mountWithAppProvider(<ComboBox {...mockProps} />);
```
kvendrik: Make sure to specify `options`.

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
```diff
@@ -50,16 +52,16 @@ describe('<ComboBox/>', () => {
...
 
       comboBox.simulate('click');
```
kvendrik: I know this is not a change you made in this PR but we probably wanna stick with manually triggering the callback instead of simulating an actual click event on the component:`trigger(comboBox.find('div').first(), 'onClick')`.

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
```diff
@@ -50,16 +52,16 @@ describe('<ComboBox/>', () => {
...
```
kvendrik: Instead of this
```ts
expect(optionListOptions[0].value).toBe('cheese_pizza');
expect(optionListOptions[0].label).toBe('Cheese Pizza');
expect(optionListOptions[1].value).toBe('macaroni_pizza');
expect(optionListOptions[1].label).toBe('Macaroni Pizza');
```

can't you do a simple assertion on the prop?

```ts
expect(comboBox.find(OptionList).prop('options')).toEqual(expectedOptions)
```

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
```diff
@@ -288,40 +238,22 @@ describe('<ComboBox/>', () => {
...
-        <ComboBox
-          options={options}
-          selected={[]}
-          textField={<input type="text" />}
-          onSelect={noop}
-        />,
+        <ComboBox {...mockProps} textField={<input type="text" />} />,
       );
       expect(comboBox.find('input')).toHaveLength(1);
       expect(comboBox.find(TextField)).toHaveLength(0);
```
kvendrik: This test will also pass if you render a `TextField` so you might wanna test this differently e.g.

```ts
const input = <input type="text" />;
const comboBox = mountWithAppProvider(<ComboBox {...mockProps} textField={input} />);
expect(comboBox.contains(input)).toBeTruthy()
```

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
```diff
@@ -418,82 +340,204 @@ describe('<ComboBox/>', () => {
...
-        await expect(comboBox.state('popoverActive')).toBe(false);
-      };
+    describe('down arrow', () => {
+      it('selects nothing if there are no options', () => {
+        const options: OptionDescriptor[] = [];
+        const comboBox = mountWithAppProvider(
+          <ComboBox {...mockProps} options={options} />,
+        );
+
+        async () => {
```
kvendrik: If this is `async` you probably wanna just make the whole test an `async` test.

```ts
it('selects nothing if there are no options', async () => {});
```

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
```diff
@@ -418,82 +340,204 @@ describe('<ComboBox/>', () => {
...
+          comboBox.find(TextField).simulate('click');
```
kvendrik: `trigger(comboBox.find(TextField), 'onFocus')`?

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
```diff
@@ -418,82 +340,204 @@ describe('<ComboBox/>', () => {
...
+          listenerMap.keyup({keyCode: Key.DownArrow});
```
kvendrik: `trigger(comboBox.find(TextField), 'onKeyPress', { keyCode: Key.DownArrow })`?

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
kvendrik: [All that seems to happen with `TextField` is that it gets passed into `Popover`](https://github.com/Shopify/polaris-react/blob/c0be5a316fc4f818554d3885650f0f0e6e8409ec/src/components/Autocomplete/components/ComboBox/ComboBox.tsx#L270) which makes me feel like these test might not belong in this suite?

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
solonaarmstrong: Typescript doesn't like comboBox being declared within the async function.

kvendrik: What happens when you do?

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
solonaarmstrong: I changed them to the tab keypress.

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
```diff
@@ -287,41 +234,23 @@ describe('<ComboBox/>', () => {
...
-          options={options}
-          selected={[]}
-          textField={<input type="text" />}
-          onSelect={noop}
-        />,
+        <ComboBox {...mockProps} textField={input} />,
       );
-      expect(comboBox.find('input')).toHaveLength(1);
-      expect(comboBox.find(TextField)).toHaveLength(0);
+      expect(comboBox.contains(input)).toBeTruthy;
```
tmlayton:
```suggestion
      expect(comboBox.contains(input)).toBeTruthy();
```

tmlayton: Tests are failing on this

solonaarmstrong:
Apologies, for some reason github wasn't syncing any of my repos, so local tests and CI were passing locally, but failing remotely. Should be fixed now.

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
AndrewMusgrave: Why do we need `async/await` in these tests? I don't think the code in the anonymous function is being executed, you probably want a `IIFE`. But as  Koen suggested adding `async` to the `it` block is preferred.

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
kvendrik: Simulate won't work in all cases. In this specific case though its mostly that I would argue that using `simulate` is pretty vague as it triggers a click event but its not clear on what.

solonaarmstrong: Finding the first div feels fragile to me and others have discouraged it in other PRs. But it makes sense to directly trigger `onClick`.

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
solonaarmstrong: There is no `onKeyPress` on the text field.

chloerice:
Is this a WIP or on hold at the moment?

PR #484, opened 2018-10-26
[Navigation] Add new indicators to nav items 

dfmcphee:

Currently we don't have a way to indicate to merchants that an item in the navigation has a sub item that is new.

This adds a `new` prop to items and subitems that will callout a new item to the merchant.

#### Example

screen shot 2018-10-26 at 5 27 53 pm

## 

### How to 🎩

<details>

```jsx
import {Page, AppProvider, Navigation} from '@shopify/polaris';

interface State {}

  render() {
    return (
      <AppProvider>
        <Page title="Playground">
          <Navigation
            location="/"
            userMenu={
              <Navigation.UserMenu
                name="Ellen Ochoa"
                detail="Ochoa Crafts"
                actions={[
                  {
                    id: '123',
                    items: [{content: 'Perform action'}],
                  },
                ]}
                message={{
                  title: 'Message title',
                  description: 'Message description',
                  action: {content: 'Perform action'},
                  link: {to: 'path/to/place', content: 'Link content'},
                }}
                avatarInitials="EO"
              />
            }
          >
            <Navigation.Section
              items={[
                {
                  url: '/',
                  label: 'Products',
                  icon: 'products',
                  subNavigationItems: [
                    {
                      url: '/',
                      label: 'All Products',
                    },
                    {
                      url: '/inventory',
                      label: 'Inventory',
                      new: true,
                    },
                    {
                      url: '/collections',
                      label: 'Collections',
                    },
                  ],
                },
              ]}
            />
          </Navigation>
        </Page>
      </AppProvider>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Navigation/README.md
+++ b/src/components/Navigation/README.md
@@ -127,6 +127,7 @@ The content of the navigation component consists of navigation items. Each item 
 | badge              | string \| null      | A string property allowing content to be displayed in a badge next to the navigation item                         |
 | label              | string              | A string property allowing content to be displayed as link text in the navigation item                            |
 | disabled           | boolean             | A boolean property indicating whether the navigation item is disabled                                             |
+| new                | boolean             | A boolean property indicating whether the navigation item is new                                                  |
 | accessibilityLabel | string              | A visually hidden label for screen readers to understand the content of a navigation item                         |
 | selected           | boolean             | A boolean property indicating whether the navigation item is the currently-selected item                          |
 | subNavigationItems | SubNavigationItem[] | A collection of navigation items rendered as nested secondary navigation items                                    |
--- a/src/components/Navigation/README.md
+++ b/src/components/Navigation/README.md
@@ -115,24 +115,24 @@ The content of the navigation component consists of navigation items. Each item 
 
 #### Item properties
 
-| Prop               | Type                | Description                                                                                                       |
-| ------------------ | ------------------- | ----------------------------------------------------------------------------------------------------------------- |
-| url                | string              | A location for the navigation item to navigate to when clicked                                                    |
-| matches            | boolean             | A boolean property indicating whether the navigation item should respond to a closely matching location property  |
-| exactMatch         | boolean             | A boolean property indicating whether the navigation item should respond to an exactly matching location property |
-| matchPaths         | string[]            | A string property providing a collection of additional paths for the navigation item to respond to                |
-| excludePaths       | string[]            | A string property providing an explicit collection of paths the navigation item should not respond to             |
-| icon               | IconProps['source'] | An icon to be displayed next to the navigation item                                                               |
-| iconBody           | string              | An `SVG` element passed as a string from a server and displayed next to the navigation item                       |
-| badge              | string \| null      | A string property allowing content to be displayed in a badge next to the navigation item                         |
-| label              | string              | A string property allowing content to be displayed as link text in the navigation item                            |
-| disabled           | boolean             | A boolean property indicating whether the navigation item is disabled                                             |
-| new                | boolean             | A boolean property indicating whether the navigation item is new                                                  |
-| accessibilityLabel | string              | A visually hidden label for screen readers to understand the content of a navigation item                         |
-| selected           | boolean             | A boolean property indicating whether the navigation item is the currently-selected item                          |
-| subNavigationItems | SubNavigationItem[] | A collection of navigation items rendered as nested secondary navigation items                                    |
-| secondaryAction    | SecondaryAction     | Renders an icon-only action as a supplementary action next to a navigation item                                   |
-| onClick()          | function            | A callback function to handle clicking on a navigation item                                                       |
+| Prop               | Type                | Description                                                                                                                                |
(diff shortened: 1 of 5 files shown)


src/components/Navigation/README.md
```diff
@@ -127,6 +127,7 @@ The content of the navigation component consists of navigation items. Each item
 | badge              | string \| null      | A string property allowing content to be displayed in a badge next to the navigation item                         |
 | label              | string              | A string property allowing content to be displayed as link text in the navigation item                            |
 | disabled           | boolean             | A boolean property indicating whether the navigation item is disabled                                             |
+| new                | boolean             | A boolean property indicating whether the navigation item is new                                                  |
```
ry5n: Should we mention that this adds the indicator dot to the parent? And that it will be overridden by the `badge` prop?

ry5n: (I don’t think the descriptions here need to give the datatype either; there’s a separate column for that)

src/components/Navigation/components/Item/Item.tsx
```diff
@@ -88,20 +99,34 @@ export default class Item extends React.Component<Props, State> {
...
+        .length > 0;
+
+    const indicatorMarkup = hasNewChild ? (
+      <span className={styles.Indicator}>
+        <Indicator pulse />
+      </span>
+    ) : null;
+
+    const badgeMarkup =
+      badge ||
```
ry5n: Is this a regression? Looks like `badgeMarkup` now gets the bare value of `badge`. It should still get wrapped in markup.

src/components/Navigation/components/Item/Item.tsx
```diff
@@ -134,6 +159,7 @@ export default class Item extends React.Component<Props, State> {
           >
             {iconMarkup}
             <span className={styles.Text}>{label}</span>
+            {indicatorMarkup}
```
ry5n: Is there anything we can do about the indicator wrapping?

src/components/Navigation/components/Item/Item.tsx
```diff
@@ -226,7 +252,10 @@ export default class Item extends React.Component<Props, State> {
           onClick={this.getClickHandler(onClick)}
         >
           {iconMarkup}
-          <span className={styles.Text}>{label}</span>
+          <span className={styles.Text}>
+            {label}
+            {indicatorMarkup}
```
ry5n: For the UnstyledLink code path here, the indicator markup is a child of the `.Text` span. For the button path it’s the next sibling. Can they not be done the same way?

src/components/Navigation/components/Item/Item.tsx
dfmcphee: Oh this is just bad formatting, the bare badge value was just to use as a conditional for the rendering. I will change it to a ternary to match the other conditional markup.

src/components/Navigation/components/Item/Item.tsx
dfmcphee: Meaning can we prevent it from wrapping while allowing the text to still wrap if needed?

src/components/Navigation/components/Item/Item.tsx
dfmcphee: Good catch. They should be the same. I moved this content up to a variable defined above so they always stay in sync.

dfmcphee:
Addressed you comments @ry5n. The only outstanding question is how we want to handle the wrapping for the indicator on long labels. I am also ok with deferring this if we aren't sure yet.

src/components/Navigation/components/Item/Item.tsx
ry5n: Yeah like avoiding the indicator from wrapping below the text, before the text itself wraps. We could try to mitigate this by turning on hyphenation for the text, and having a non-breaking space separate the word and the indicator, but that would need to be tested. I’m also happy to defer this, but it’s worth testing with the menu items we know are going to get this soon, combined with our current set of locales. In other words, it’s OK to defer unless it’s going to immediately look bad to some merchants :)

src/components/Navigation/components/Item/Item.tsx
```diff
@@ -57,7 +68,7 @@ enum MatchState {
   NoMatch,
 }
 
-export default class Item extends React.Component<Props, State> {
+export class BaseItem extends React.Component<CombinedProps, State> {
```
AndrewMusgrave: I'm curious, why did you change the class to `BaseItem` and export like 
```tsx
export const Item = withAppProvider<Props>()(BaseItem);
export default Item
```
rather than 
```tsx
export default withAppProvider<Props>(Item)
````

dfmcphee: I needed to do this because we use `Item` to [create the sub-navigation](https://github.com/Shopify/polaris-react/blob/49294ee23209ec3a844e916f1a7af4cc524fc933/src/components/Navigation/components/Item/Item.tsx#L226) items and this also needs to be wrapped in `withAppProvider`.

tmlayton:
Want to ship this with v3.0.0?

PR #493, opened 2018-10-29
[Layout Readme] Fix link text to FormLayout component from Layout readme

andrewlo:

Link to `FormLayout` from `Layout`'s readme's related component section uses the term `layout` instead of `form layout`.

Change it to `form layout` to be consistent with the component's name & other references to `FormLayout` component.


BPScott:
Thanks for the fix @andrewlo!

PR #503, opened 2018-10-30
[Autocomplete] Adds emptyState prop

darrenhebner:
This PR allows developers to render an optional empty state if there are no results for their search query.

This PR adds the `emptyState` prop. This prop takes a function that renders some JSX. `emptyState` is called when `options` is empty and there are no before or after actions.

empty state

### How to 🎩

Include the following snippet in the Playground file. Try searching for something that does not return any results. You should see the empty state.

<details>

```jsx
/* eslint-disable */

import {Page, AppProvider} from '@shopify/polaris';
import {Autocomplete} from '../src/components';

interface State {}

  state = {
    query: '',
  };

  options = [
    {label: 'One', value: '1'},
    {label: 'Two', value: '2'},
    {label: 'Three', value: '3'},
  ];

  render() {
    const {query} = this.state;
    return (
      <AppProvider>
        <Page title="Playground">
          <Autocomplete
            selected={[]}
            options={this.getOptions()}
            onSelect={() => {
              return;
            }}
            allowMultiple
            emptyState={<span>No results</span>}
            textField={
              <Autocomplete.TextField
                label="Search"
                value={query}
                onChange={(value) => this.setState({query: value})}
              />
            }
          />
        </Page>
      </AppProvider>
    );
  }

  getOptions() {
    const {query} = this.state;

    if (query.length === 0) {
      return this.options;
    }

    return this.options.filter((option) =>
      option.label.toLowerCase().includes(query.toLowerCase()),
    );
  }
}

/* eslint-enable */
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Autocomplete/Autocomplete.tsx
+++ b/src/components/Autocomplete/Autocomplete.tsx
@@ -30,6 +30,8 @@ export interface Props {
   loading?: boolean;
   /** Indicates if more results will load dynamically */
   willLoadMoreResults?: boolean;
+  /** Is rendered when there are no options */
+  emptyState?: React.ReactNode;
   /** Callback when the selection of options is changed */
   onSelect(selected: string[]): void;
   /** Callback when the end of the list is reached */
@@ -58,6 +60,7 @@ export class Autocomplete extends React.PureComponent<CombinedProps, never> {
       loading,
       actionBefore,
       willLoadMoreResults,
+      emptyState,
       onSelect,
       onLoadMoreResults,
       polaris: {intl},
@@ -91,6 +94,7 @@ export class Autocomplete extends React.PureComponent<CombinedProps, never> {
         actionsBefore={conditionalAction}
         onSelect={onSelect}
         onEndReached={onLoadMoreResults}
+        emptyState={emptyState}
       />
     );
   }
--- a/src/components/Autocomplete/README.md
+++ b/src/components/Autocomplete/README.md
@@ -302,6 +302,98 @@ class AutocompleteExample extends React.Component {
 
 ---
 
+### Autocomplete with empty state
+
+Use to indicate there are no search results.
+
+```jsx
+class AutocompleteExample extends React.Component {
+  options = [
+    {value: 'rustic', label: 'Rustic'},
+    {value: 'antique', label: 'Antique'},
+    {value: 'vinyl', label: 'Vinyl'},
+    {value: 'vintage', label: 'Vintage'},
+    {value: 'refurbished', label: 'Refurbished'},
+  ];
+
+  state = {
+    selected: [],
+    inputText: '',
+    options: this.options,
+    loading: false,
+  };
+
+  render() {
+    return (
+      <div style={{height: '225px'}}>
+        <Autocomplete
+          options={this.state.options}
+          selected={this.state.selected}
+          onSelect={this.updateSelection}
+          emptyState={
+            <React.Fragment>
+              <Icon source="search" />
+              <div style={{textAlign: 'center'}}>
+                <TextContainer>Could not find any results</TextContainer>
+              </div>
+            </React.Fragment>
+          }
+          loading={this.state.loading}
+          textField={
+            <Autocomplete.TextField
+              onChange={this.updateText}
+              label=""
+              value={this.state.inputText}
+              prefix={<Icon source="search" color="inkLighter" />}
+              placeholder="Search"
+            />
+          }
+        />
+      </div>
+    );
+  }
+
+  updateText = (newValue) => {
+    this.setState({inputText: newValue});
+    this.filterAndUpdateOptions(newValue);
+  };
+
+  filterAndUpdateOptions = (inputString) => {
+    if (!this.state.loading) {
+      this.setState({loading: true});
+    }
+
+    setTimeout(() => {
+      if (inputString === '') {
+        this.setState({
+          options: this.options,
+          loading: false,
+        });
+        return;
+      }
+      const filterRegex = new RegExp(inputString, 'i');
+      const resultOptions = this.options.filter((option) =>
+        option.label.match(filterRegex),
+      );
+      this.setState({
+        options: resultOptions,
+        loading: false,
+      });
+    }, 300);
+  };
+
+  updateSelection = (updatedSelection) => {
+    const selectedText = updatedSelection.map((selectedItem) => {
+      const matchedOption = this.options.filter((option) => {
+        return option.value.match(selectedItem);
+      });
+      return matchedOption[0] && matchedOption[0].label;
+    });
+    this.setState({selected: selectedText, inputText: selectedText});
+  };
+}
+```
+
 ## Related components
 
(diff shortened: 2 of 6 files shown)


src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -30,6 +30,8 @@ export interface Props {
   loading?: boolean;
   /** Indicates if more results will load dynamically */
   willLoadMoreResults?: boolean;
+  /** Is rendered when there are no options */
+  renderEmptyState?(): React.ReactElement<any>;
```
solonaarmstrong: I'm curious why it's a function?

solonaarmstrong:
We would also want to write a test in Autocomplete.test.tsx to test that the prop is passed to Combobox.

src/components/Autocomplete/Autocomplete.tsx
darrenhebner: No real reason. It's pretty common in `web` to use a function here. But since we aren't passing it any params, it doesn't really matter.

The old autocomplete component also used a function here, so it would make it easier for consumers of the old autocomplete to migrate over to the new one.

darrenhebner: @AndrewMusgrave - You mean `renderEmptyState?(): React.ReactNode`? I'm cool with that.

ry5n:
@AndrewMusgrave Seems reasonable, although I wonder in what cases this should be used. I can imagine that sometimes if there are no results we’d want to show nothing (hide the popover). Without the scenarios mapped out it’s hard to tell.

darrenhebner:
It's also optional so there is still the option to hide the popover if there are no options.

Here's our use case. We are fetching resources from an external API. We need to let the merchant know that no results were returned for their query.

src/components/Autocomplete/Autocomplete.tsx
solonaarmstrong: I agree, I think a React.ReactNode is sufficient and will accomplish what's needed without a function. Looking at empty states in lists on index pages in production, a consumer could pass an icon or a text container with a message.

darrenhebner:
@AndrewMusgrave @solonaarmstrong - All changes should be good to go. 🚢 ?

darrenhebner:
@tmlayton - Great, I'll add an example to the docs quickly.

solonaarmstrong:
A few things:
- Yes, I think we should provide an example. (I can do that)
- This prop also might want to be called simply emptyState because it's no longer a function
- It doesn't work properly with loading

screen shot 2018-11-01 at 1 33 14 pm

ComboBox is not aware of loading, so to keep this within autocomplete, it needs some logic:

```
let emptyStateMarkup = renderEmptyState;
    if (loading) {
      emptyStateMarkup = null;
    }
```

or something resembling that...

src/components/Autocomplete/README.md
```diff
@@ -61,6 +61,14 @@ class AutocompleteExample extends React.Component {
...
           selected={this.state.selected}
           onSelect={this.updateSelection}
+          renderEmptyState={
+            <React.Fragment>
+              <Icon source="search" />
+              <div style={{textAlign: 'center'}}>
+                <TextContainer>Could not find any results</TextContainer>
+              </div>
+            </React.Fragment>
+          }
```
solonaarmstrong: I can't test this in the style guide because the build is failing.

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
```diff
@@ -231,6 +237,13 @@ export default class ComboBox extends React.PureComponent<Props, State> {
       <div ref={this.popoverScrollContainer} />
     );
 
+    const emptyStateMarkup = !actionsBefore &&
+      !actionsAfter &&
+      options.length === 0 &&
```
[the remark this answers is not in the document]
AndrewMusgrave: What are the pitfalls to this approach? We use it all the times for string

src/components/Autocomplete/README.md
AndrewMusgrave: Should we assign the markup to a const first?
```tsx
render() {
    const emptyStateMarkup = (
        ...
     );

     return (
        ...
          renderEmptyState={emptyStateMarkup}
        ...
    )
}
```

AndrewMusgrave: Same comment for the two examples below

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
```diff
@@ -499,6 +499,89 @@ describe('<ComboBox/>', () => {
...
+          options={[]}
+          selected={[]}
+          textField={renderTextField()}
+          onSelect={noop}
+          emptyState={<EmptyState />}
+        />,
+      );
+
+      comboBox.simulate('click');
+      expect(comboBox.find(EmptyState).exists()).toBe(true);
```
AndrewMusgrave: @solonaarmstrong probably has more opinions on this. Is our general pattern
```tsx
expect(comboBox.find(EmptyState).exists()).toBe(true);
```

or

```tsx
expect(comboBox.find(EmptyState)).toHaveLength(1);
```

solonaarmstrong: We favour the latter. https://github.com/Shopify/web-foundation/blob/master/Best%20practices/Enzyme.md#expectations

src/components/Autocomplete/README.md
solonaarmstrong: TextField, in this example, is not assigned to a variable, so I'm going to leave it as is and clean up the examples in another PR.

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
darrenhebner: We are doing `expect(comboBox).toContainComponent(EmptyState)` in `web`. I'll update to use `expect(comboBox.find(EmptyState)).toHaveLength(1);` though if we don't have the new matchers in this project.

darrenhebner: Also, the entire `ComboBox.test.tsx` is using 

`expect(comboBox.find(EmptyState).exists()).toBe(true);`

darrenhebner:
@solonaarmstrong @AndrewMusgrave @tmlayton - All changes should be addressed. Really hoping to ship 🤞

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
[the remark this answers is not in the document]
tmlayton: On second thought, that statement was too strong. I think we should avoid type coercion because this can result in unexpected cases, but in this case it is technically safe. Although, I think the main benefit of avoiding type coercion in a condition is for readability. We have this rule outlined [here](https://github.com/shopify/javascript#11.5).

PR #515, opened 2018-10-31
Add eslint shopify/jest plugin

BPScott:

To ensure our tests follow best practices. This caught a few places where we didn't remember to make assertions in our tests.

Adds the shopify/jest eslint plugin (along with some reordering of plugins to make them not stamp over each other).
Runs `yarn format` to autofix several items
Makes manual fixes where required

I suggest reviewing this per-commit, as the first commit contains all the autofixes which are pretty brainless, while the second commit is code I had to modify myself

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -26,7 +26,6 @@
         "allowBlockStart": false
       }
     ],
-    "jest/no-disabled-tests": "off",
     "react/button-has-type": "off",
     "react/no-array-index-key": "off",
     "shopify/jest/no-vague-titles": "off",
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,10 +1,9 @@
 {
   "extends": [
-    "plugin:shopify/esnext",
+    "plugin:shopify/typescript-react",
+    "plugin:shopify/jest",
     "plugin:shopify/node",
     "plugin:shopify/polaris",
-    "plugin:shopify/typescript",
-    "plugin:shopify/typescript-react",
     "plugin:shopify/typescript-prettier"
   ],
   "settings": {
@@ -27,8 +26,10 @@
         "allowBlockStart": false
       }
     ],
+    "jest/no-disabled-tests": "off",
     "react/button-has-type": "off",
     "react/no-array-index-key": "off",
+    "shopify/jest/no-vague-titles": "off",
     "jsx-a11y/label-has-for": [
       2,
       {
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -111,7 +111,7 @@ The autocomplete component is an input field that provides selectable suggestion
 - Move to use sewing-kit for test running, updating to Jest 23 in the process. This gives us working sourcemaps for code coverage
 - Improved accessibility testing checklist
 - Updated development node environment to 8.12.0
-- Added shopify/jest plugin to eslint config.
+- Added shopify/jest plugin to eslint config
 
 #### Open development
 
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -111,6 +111,7 @@ The autocomplete component is an input field that provides selectable suggestion
 - Move to use sewing-kit for test running, updating to Jest 23 in the process. This gives us working sourcemaps for code coverage
 - Improved accessibility testing checklist
 - Updated development node environment to 8.12.0
+- Added shopify/jest plugin to eslint config.
 
 #### Open development
 
--- a/examples/create-react-app/src/tests/App.test.js
+++ b/examples/create-react-app/src/tests/App.test.js
@@ -6,7 +6,7 @@ import App from '../App';
 import {mountWithAppProvider} from './utils/enzyme';
 
 // Sample test to demonstrate how you should test with AppProvider context
-it('renders page', () => {
+test('renders page', () => {
   const wrapper = mountWithAppProvider(<App />);
   expect(wrapper.find(Page).exists()).toBe(true);
 });
--- a/src/components/ActionList/components/Section/tests/Section.test.tsx
+++ b/src/components/ActionList/components/Section/tests/Section.test.tsx
@@ -77,7 +77,7 @@ describe('<Section />', () => {
   });
 
   it('passes the onActionAnyItem callback to Item', () => {
-    const callback = () => {};
+    const mockOnActionAnyItem = jest.fn();
     const section = mountWithAppProvider(
       <Section
         hasMultipleSections
@@ -87,15 +87,15 @@ describe('<Section />', () => {
             {content: 'Export file', onAction: noop},
           ],
         }}
-        onActionAnyItem={callback}
+        onActionAnyItem={mockOnActionAnyItem}
       />,
     );
 
-    expect(
-      section
-        .find(Item)
-        .first()
-        .prop('onAction'),
-    );
+    section
+      .find('Item button')
+      .first()
+      .simulate('click');
+
+    expect(mockOnActionAnyItem).toHaveBeenCalledTimes(1);
   });
 });
--- a/src/components/ActionList/components/Section/tests/Section.test.tsx
+++ b/src/components/ActionList/components/Section/tests/Section.test.tsx
@@ -18,7 +18,7 @@ describe('<Section />', () => {
       />,
     );
 
-    expect(section.find(Item).length).toBe(2);
+    expect(section.find(Item)).toHaveLength(2);
   });
 
   it('renders items as li when hasMultipleSections is false', () => {
@@ -34,7 +34,7 @@ describe('<Section />', () => {
       />,
     );
 
-    expect(section.find('li').length).toBe(2);
+    expect(section.find('li')).toHaveLength(2);
   });
 
   it('wraps items in an li when hasMultipleSections is true', () => {
@@ -50,7 +50,7 @@ describe('<Section />', () => {
       />,
     );
 
-    expect(section.find('li').length).toBe(3);
+    expect(section.find('li')).toHaveLength(3);
   });
 
   it('passes content to Item', () => {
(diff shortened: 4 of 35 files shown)


src/components/Modal/tests/Modal.test.tsx
```diff
@@ -483,15 +483,6 @@ describe('<Modal>', () => {
       modal.unmount();
       expect(appBridgeModalMock.unsubscribe).toHaveBeenCalledTimes(1);
     });
-
-    it('unsubscribes on unmount', () => {
```
BPScott: Deleting this as it is identical to the above test

src/components/ResourceList/components/FilterControl/components/DateSelector/tests/DateSelector.test.tsx
```diff
@@ -134,9 +134,7 @@ describe('<DateSelector />', () => {
 
       expect(wrapper.find(DatePicker).exists()).toBe(false);
     });
-  });
 
-  describe('filterValue', () => {
```
BPScott: The above describe block is also called "filterValue", so put all the tests together

examples/create-react-app/src/tests/App.test.js
```diff
@@ -6,7 +6,7 @@ import App from '../App';
 import {mountWithAppProvider} from './utils/enzyme';
 
 // Sample test to demonstrate how you should test with AppProvider context
-it('renders page', () => {
+test('renders page', () => {
```
solonaarmstrong: I don't understand this change.

src/components/ActionList/components/Section/tests/Section.test.tsx
```diff
@@ -77,7 +77,7 @@ describe('<Section />', () => {
   });
 
   it('passes the onActionAnyItem callback to Item', () => {
-    const callback = () => {};
+    const mockOnActionAnyItem = jest.fn();
```
solonaarmstrong: Should we use `spy` here?

src/components/CalloutCard/tests/CalloutCard.test.tsx
```diff
@@ -19,7 +19,12 @@ describe('<CalloutCard />', () => {
   );
 
   it('renders its children', () => {
-    expect(calloutCard.contains('<p>Content</p>'));
+    expect(
+      calloutCard
+        .find('p')
+        .first()
+        .contains('Content'),
+    ).toBe(true);
```
solonaarmstrong: I guess we can't assert `toContain` here?

solonaarmstrong: https://github.com/Shopify/web-foundation/blob/master/Best%20practices/Enzyme.md#expectations

src/components/DatePicker/components/Day/tests/Day.test.tsx
```diff
@@ -43,7 +43,7 @@ describe('<Day />', () => {
         />,
       );
       day.simulate('click');
-      expect(spy).toHaveBeenCalled;
+      expect(spy).toHaveBeenCalled();
```
solonaarmstrong: We should probably do `toHaveBeenCalledTimes(1)`

src/components/ResourceList/components/FilterControl/components/FilterValueSelector/tests/FilterValueSelector.test.tsx
```diff
@@ -407,7 +407,7 @@ describe('<FilterValueSelector />', () => {
         expect(operatorsSelect.exists()).toBe(false);
       });
 
-      it('calls onFilterKeyChange when the filter key was changed', () => {
+      it('calls onChange when the filter key was changed', () => {
```
solonaarmstrong: We should decide on a convention. Do we use `spy` or the name of the function being called?

src/utilities/tests/compose.test.ts
```diff
@@ -20,7 +20,7 @@ const toScreamingCaseSnake = compose(
 
 describe('compose', () => {
   it('will return a function', () => {
-    expect(typeof toScreamingCaseSnake === 'function');
+    expect(typeof toScreamingCaseSnake === 'function').toBe(true);
```
solonaarmstrong: `expect(typeof toScreamingCaseSnake).toBe('function');` ?

solonaarmstrong:
This is fantastic, Ben! I've been wanting to clean some of these up and you made it so easy.

examples/create-react-app/src/tests/App.test.js
BPScott: https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/consistent-test-it.md are the docs for this rule.

It seems we're using the default configuration of top level tests use `test` and all tests nested within `describe` use `it`.

src/components/CalloutCard/tests/CalloutCard.test.tsx
BPScott: We're not currently using jest-enzyme so `toContain` doesn't exist.

@kvendrik raised #529 that says we should add it and I agree with you both. toContain() is a much nicer approach but lets add jest-enzyme and updating tests to use its new functionality in a separate PR.

src/components/ResourceList/components/FilterControl/components/FilterValueSelector/tests/FilterValueSelector.test.tsx
BPScott: The name of the function being called IMO. `spy` is a detail of how we do the testing rather than the functionality we want to check.

Also if we call this `spy` then we would violate the [no-identical-title](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-identical-title.md) rule as this test and the test below would both be `calls spy when the filter key was changed` instead of:

`calls onChange when the filter key was changed` and `calls onFilterKeyChange when the filter key was changed`

src/components/CalloutCard/tests/CalloutCard.test.tsx
kvendrik: @BPScott [`toContain`](https://jestjs.io/docs/en/expect#tocontainitem) is part of Jest so you should be able to use it here 😄.

src/components/ResourceList/components/FilterControl/components/FilterValueSelector/tests/FilterValueSelector.test.tsx
kvendrik: The title should always use the name of the method I would say. In the method you then create a spy named `[name of the method]Spy` to assert the call.

src/components/ActionList/components/Section/tests/Section.test.tsx
BPScott: Per conversation in slack: My personal preference is for a more verbose name, but `spy` is the existing dominant convention. I'll use spy here to maintain existing convention, then we can update everything to more verbose names in a separate PR.

.eslintrc
```diff
@@ -27,8 +26,10 @@
         "allowBlockStart": false
       }
     ],
+    "jest/no-disabled-tests": "off",
     "react/button-has-type": "off",
     "react/no-array-index-key": "off",
+    "shopify/jest/no-vague-titles": "off",
```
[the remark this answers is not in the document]
BPScott: enabling [no-vague-titles](https://github.com/Shopify/eslint-plugin-shopify/blob/master/docs/rules/jest/no-vague-titles.md) results in 97 errors and I can't be bothered with going through and manually rewording that many test names at this moment.

Following our chat in slack I've enabled `no-disabled-tests` globally and ignored the specific skipped tests we want to keep pending jsdom allowing us to test what we need.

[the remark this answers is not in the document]
kvendrik: For the `no-vague-titles` thing maybe create an issue so we can come back to it later on?

src/components/Frame/tests/Frame.test.tsx
```diff
@@ -191,6 +193,7 @@ describe('<Frame />', () => {
   // JSDOM 11.12.0 does not support setting/reading custom properties so we are
   // unable to assert that we set a custom property
   // See https://github.com/jsdom/jsdom/issues/1895
+  // eslint-disable-next-line jest/no-disabled-tests
```
kvendrik: 🎉

src/components/CalloutCard/tests/CalloutCard.test.tsx
BPScott: oh hang on yes, I'm talking garbage sorry.

I did try doing this with toContain and it was failing which was rather odd :/

kvendrik: It does work with strings though so this should work:
```tsx
expect(calloutCard.find('p').first()).toContain('Content');
```

.eslintrc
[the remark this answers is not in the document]
BPScott: Done: #530

UNRELEASED.md
```diff
@@ -104,6 +104,7 @@ The autocomplete component is an input field that provides selectable suggestion
 - Move to use sewing-kit for test running, updating to Jest 23 in the process. This gives us working sourcemaps for code coverage
 - Improved accessibility testing checklist
 - Updated development node environment to 8.12.0
+- Added shopify/jest plugin to eslint config.
```
```suggestion
- Added shopify/jest plugin to eslint config
```

src/components/CalloutCard/tests/CalloutCard.test.tsx
tmlayton: No need for `toContain`

```ts
expect(calloutCard.find('p').first().text()).toBe('Content');
```

src/components/ResourceList/components/FilterControl/components/FilterValueSelector/tests/FilterValueSelector.test.tsx
tmlayton: I’m good with this too. Agreed that the title should use the name of the method. 

> In the method you then create a spy named [name of the method]Spy (e.g. onChangeSpy) to assert the call.

If you are using `spyOn`, yes. Otherwise, it should be `onChangeMock` if using `jest.fn()`. Although, I find myself preferring to drop the mock or spy suffix for simple tests.

PR #517, opened 2018-10-31
Fix focus invariant error

### LINKED ISSUE #507: Focus attempts to use findDOMNode() when it is possibly unmounted
# Issue summary

Focus attempts to run [`findDOMNode()`](https://github.com/Shopify/polaris-react/blob/a3c2898ac4d210b2eed6b4e91fa5d7c2a8021d3e/src/components/Focus/Focus.tsx#L17-L24) in callback during `didMount`. Unfortunately, by the point this happens, the component can have become unmounted. This causes an error (for those at Shopify, here's the error in Web: https://app.bugsnag.com/shopify/shopify-web-client/errors/5ae71d22fa94dc00178c5a21?event_id=5bd8c91700309c5bd43b0000&i=sk&m=fq)

## Expected behavior

No error.

## Actual behavior

An invariant violation (http://reactjs.org/docs/error-decoder.html?invariant=188). IMO we can remove the `write()` callback stuff altogether, it's an optimization we probably never needed.

I Haven't checked but presumably rendering a `Focus`, then setting state and unmounting it immediately would lead to this issue.

AndrewMusgrave:
Resolves #507

Previously `Focus` had it self focusing logic in a called back passed to `write` which uses `requestAnimationFrame` under the hood. It's an unneeded optimization which was causing the calledback to be invoked after the component has unmounted causing an invariant error.

### How to 🎩

<details>

Check the console for errors

```jsx
/* eslint-disable */

import {Page, AppProvider, Focus, TextStyle} from '@shopify/polaris';

interface State {
  toggle: boolean;
}

  state = {
    toggle: true,
  };

  componentDidMount() {
    this.setState({toggle: false});
  }

  render() {
    const {toggle} = this.state;

    const focusMarkup = toggle && (
      <Focus>
        <TextStyle>Polaris</TextStyle>
      </Focus>
    );

    return (
      <AppProvider>
        <Page title="Playground">{focusMarkup}</Page>
      </AppProvider>
    );
  }
}

/* eslint-enable */
```

</details>

#### Notes
* Focus is not exposed so a changelog entry is not needed

### DIFF
--- a/src/components/Focus/Focus.tsx
+++ b/src/components/Focus/Focus.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import * as ReactDOM from 'react-dom';
 import {focusFirstFocusableNode} from '@shopify/javascript-utilities/focus';
-import {write} from '@shopify/javascript-utilities/fastdom';
 
 export interface Props {
   children?: React.ReactNode;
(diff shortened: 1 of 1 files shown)


src/components/Focus/Focus.tsx
```diff
@@ -10,24 +10,35 @@ export interface Props {
...
+    if (root) {
+      if (!root.querySelector('[autofocus]')) {
+        focusFirstFocusableNode(root, false);
       }
-    });
+    }
   }
 
   render() {
-    const Fragment = (React as any).Fragment;
```
AndrewMusgrave: Types for react are correct now so this can be removed

PR #525, opened 2018-11-01
fixes absolute paths

kvendrik:
fixes https://github.com/Shopify/polaris-react-deprecated/issues/2410
fixes https://github.com/Shopify/polaris-react-deprecated/issues/2469
fixes part of https://github.com/Shopify/polaris-react-deprecated/issues/2413

#### Description
This PR implements the missing pieces we need to make absolute paths work. It implements a fix for the issues we experienced with the TS type definitions, implements a test to make sure the fix keeps working and reverts all imports back to favouring absolute paths over relative ones.

Apart from this the only changes that have been necessary so far were small config changes. This PR implements all thats left to make everything work which gives us the opportunity to look at the changes necessary and decide how we wanna move forward. Do we want to support absolute paths or do we wanna stick with relatives ones? If we decide to stick with relative paths I'll revert back the changes previously made to support absolute imports and if not we can merge this in and move forward with the new import standard.

#### The case for absolute paths
Preferring absolute paths is mainly about readability. It prevents paths from becoming confusing when nested deeply into the folder structure. E.g.

```tsx
import {handleMouseUpByBlurring} from '../../../../../../utilities/focus';
import Indicator from '../../../../../Indicator';
import {Icon, UnstyledLink} from '../../../../../../components';
import {IconableAction, DisableableAction} from '../../../../../../types';
```

versus

```tsx
import {handleMouseUpByBlurring} from 'utilities/focus';
import Indicator from 'components/Indicator';
import {Icon, UnstyledLink} from 'components';
import {IconableAction, DisableableAction} from 'types';
```

#### Trade-offs
- Absolute to relative imports transformation is [not something TypeScript currently supports](https://github.com/Microsoft/TypeScript/issues/15479) for type definitions.
- There are a few [workarounds](https://medium.com/@grrowl/fixing-absolute-imports-in-typescript-797f405176eb) none of which are ideal.

This PR goes with one of the less ideal options and [adds a small script](https://github.com/Shopify/polaris-react/blob/b52d1093adb6da26bec1f3b545d77143e5d05049/scripts/ts-paths-transform.js) that transforms absolute module paths in the type definitions to relative ones.

### What this PR doesn't do
No matter if we decide to go with absolute paths or not we [should still work on a full test of the consumer build](https://github.com/Shopify/polaris-react/issues/2413) for obvious reasons.

### Open questions
- How do we feel about the trade-offs described above?
- How do we feel about the pros and cons of sticking with relative imports?
- What would be the best way to test the types (`yarn check-types`). Should it be part of `build.test.js` or should it run separately?

### How to 🎩 

**Types**
1. In `web` switch to branch `v3-rc.1`.
2. In `polaris-react` switch to this branch and run `yarn build-consumer web`.
3. In `web` run `yarn type-check` and make sure it passes without errors.

**Full prod build**
1. In `web` switch to branch `v3-rc.1`.
2. In `polaris-react` switch to this branch and run `yarn build-consumer web`.
3. In `web` run `yarn prod` and open the admin on `https://shop1.myshopify.io/admin`.

**Note: on hold until v3 has been released.**


danrosenthal:
@kvendrik is this still a change you'd like to see merged? I'd recommend rebasing and re-pinging our closing.

kvendrik:
@danrosenthal yep this change was on hold until the v3 release. I've pinged @BPScott yesterday about finding someone to finish this up if he would like to see it merged soon as I'm currently busy with other projects.

dpersing:
Friendly neighborhood staleness checker 🍞! Got a bit of time to post an update @BPScott @kvendrik?

kvendrik:
This PR is waiting for review by @BPScott 😄

BPScott:
@kvendrikv Thanks for finding the time to update this.

I'm trying to compare the build output of this vs master but it is a little out of date. I have tried rebasing this on top of master but have ran into large amounts of merge conflicts during the rebase process, probably as a result of this branch starting off before v3.

Can you squash this down into a single commit (or possibly two - one for the mapping build step and one for updating all the import statements) based off the tip of master so that I can easily test this?

kvendrik:
@BPScott just looked into squashing the commits. It seems like you say to come up with a bunch of conflicts. It might be easier to simply run a build on this branch, one on master and then compare the output. I just did, the diff LGTM, here are some steps to reproduce:

1. In `polaris-react` go up one folder and create a folder named `build-test`. In this folder create folders  `node_modules/@shopify`.
2. In `polaris-react` switch to this branch and run `yarn run build-consumer build-test`.
3. When thats done rename the newly created folder in `build-test/node_modules/@shopify` to `ts-imports-build`.
4. In `polaris-react` switch to master and run `yarn run build-consumer build-test`.
5. In `build-test/node_modules/@shopify` now run `git diff --no-index polaris ts-imports`.

BPScott:
I've worked out how to squash without pain - using `git merge --squash` and creating a second temporary branch.

```
git checkout master
git checkout -b fix/ts-imports-alt
git merge --squash fix/ts-imports
git commit -m "Allow using absolute paths in imports"
```

At this point the `fix/ts-imports-alt` branch has a single commit containing all the changes in this PR, based off master.

From there you can `git diff fix/ts-imports-alt origin/fix/ts-imports` to compare the state of your new local branch and this PR - they should be identical.

Then you can set the upstream branch, and force push into this PR.

src/components/Modal/components/Dialog/index.ts
```diff
@@ -1,4 +1 @@
-import Dialog from './Dialog';
```
BPScott: Can you please revert these stylistic changes for the moment?

I agree that the `export {default, Props} from './Dialog';` style is better but  these changes adds a some noise when comparing the build.

Ideally I'd like to be able to compare the build output and see that they are identical, and this adds changes.

kvendrik:
> I would expect that there would be no differences between the builds.

The esnext build will contain the absolute imports. Afaik we haven't had any issues with that which is why this PR only converts absolute imports in TS types to relative ones. There is a good chance this does cause issues and that we just haven't noticed them before so we should probably investigate.

> I'm also seeing several changes in the types

These seem to be all from stylistic changes and seem to be legit but I agree probably best to remove those stylistic changes from the PR for the moment.

kvendrik:
@AndrewMusgrave working on a solution for the esnext build today, will make sure to update you once I have something up 🙂

kvendrik:
@AndrewMusgrave this is ready for a (very thorough) review 😄

AndrewMusgrave:
Sorry it's taking so long to review, I'll get to it on monday 🙏

dpersing:
@AndrewMusgrave Do you have any more thoughts on who should take a gander at this?

BPScott:
RIGHT! Yes. Sorry.

Bad news: despite my initial interest I've now gone kinda cold on this. Having done some research into our build tooling as part of my recent work to update to use Babel 7, I'm leaning towards the opinion that using a baseUrl and custom resolution paths is a [remnant of AMD-style loaders](https://www.typescriptlang.org/docs/handbook/module-resolution.html#additional-module-resolution-flags) and introducing them is an extra step of complexity in our build process that I don't think is worth the gain in developer experience.

As part of the Babel 7 upgrade work I've became interested in the idea that we could simplify our current two-step build process (where we build the contents of src into build-intermediate with `tsc`, then build the contents of build-intermediate with `babel`) into a single-step process by having Babel read our source typescript files directly [now that Babel can read typescript files](https://github.com/Microsoft/TypeScript-Babel-Starter) - We'd still use `tsc` for type checking but it could be removed from "transforming our files" part of our build process. This has the potential to speed up transpilation which would be great for test running and work requiring build-consumer. It also has the potential to result in smaller output files as we end up with just babel polyfills provided by babel-preset-env instead of babel's polyfills and `tsc`'s polyfills. I'd also like to see if we can move towards a build purely using rollup instead of a series of scripts as part of this bid for simplification.

The work to add these file mapping transforms would complicate that potential future work which I think has greater value that the value provided by this PR.

I still like this idea of simpler import paths but I'd like to attempt to solve it after that investigation into simplifying the build. As it currently stands this PR increases the friction for any future build improvement work, and I'd like to avoid that for the moment.

Sorry :)

PR #526, opened 2018-11-01
adds missing tests to Page and its sub-components

kvendrik:
fixes https://github.com/Shopify/polaris-react-deprecated/issues/2438

- Adds missing tests to Page and its sub-components.
- Removes indicator logic (more details here https://github.com/Shopify/polaris-react/pull/526/commits/ca283f3b8dad9090b2308831a3a89542345c3373).


danrosenthal:
@kvendrik is this still a WIP, or is it ready to be reviewed? Looks like the `WIP` bot is picking up a work in progress from a linked PR.

kvendrik:
@danrosenthal this is ready to be review, tagged @solonaarmstrong last week but looks like she hasn't had time yet. 😄

solonaarmstrong:
> You make writing tests look easy

That he does!

src/components/Page/components/Header/components/Action/tests/Action.test.tsx
```diff
@@ -0,0 +1,99 @@
+import * as React from 'react';
+import {Icon, UnstyledLink} from 'components';
```
solonaarmstrong: I'm not sure we can import this way?

src/components/Page/components/Header/components/Action/tests/Action.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  describe('disabled', () => {
+    it('gets passed into the button', () => {
+      const action = mountWithAppProvider(<Action disabled />);
+      expect(action.find('button').prop('disabled')).toBeTruthy();
+    });
+  });
+
+  describe('children', () => {
+    it('get rendered when an icon is present', () => {
```
solonaarmstrong: gets?

src/components/Page/components/Header/components/Action/tests/Action.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  describe('onAction()', () => {
+    it('triggers when the button gets clicked', () => {
+      const onActionSpy = jest.fn();
+      const action = mountWithAppProvider(<Action onAction={onActionSpy} />);
+      trigger(action.find('button'), 'onClick');
+      expect(onActionSpy).toHaveBeenCalledTimes(1);
+    });
+  });
+});
```
solonaarmstrong: Should this have a test for `disclosure`?

src/components/Page/components/Header/components/ActionGroup/ActionGroup.tsx
```diff
@@ -20,11 +17,9 @@ class ActionGroup extends React.Component<Props, never> {
   render() {
     const {actions, details, title, icon, active} = this.props;
 
-    const detailsMarkup = details ? (
+    const detailsMarkup = details && (
```
solonaarmstrong: Why this change?

src/components/Page/components/Header/components/Action/tests/Action.test.tsx
kvendrik: There is a test that makes sense the `children` get rendered when disclosure if defined (because of the if statement that controls how they get rendered based on if `disclosure` is defined). Apart from that the only related behaviour is the addition of an icon which is purely presentational and we therefor want to leave to the VRT.

src/components/Page/components/Header/components/ActionGroup/ActionGroup.tsx
kvendrik: Wasn't necessary just came past is and decided to clean it up. 😄

solonaarmstrong: I'm not clear on when we use `&&`. Wouldn't this need to be null if it doesn't pass the condition, since it's markup?

kvendrik: [Its explained more in-depth in React's docs](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator) but the short answer is that `undefined` is part of type `ReactNode`. 😄

PR #532, opened 2018-11-01
Update sewing-kit

BPScott:

* Keeping up to date
* Removes the need for some overrides in our jest configuration
* Improves our linting, by using an updated version of eslint-plugin-shopify
  * fixes some cases where linting rules were mistakenly disabled thanks to the typescript and react configs having precedence fights. We thought these rules were always enabled but they weren't. Now they are.
  * import linting rules now work within typescript files. This enforces the order of imports (node builtins, then node module, then parent imports, then siblings), and ensures [paths are a simple as possible](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-useless-path-segments.md).

Updates to the latest version of sewing-kit
Updates files to comply with linting rules.

Review this PR one commit at a time as about half the size of this pr comes from 7f835d3d which is the result of running `yarn format` rather than any manual changes.

### How to 🎩

`yarn lint` and `yarn test` should both pass. 

Change around the ordering of imports so sibling imports come before parent imports and see `yarn lint` complain

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,6 +2,7 @@
   "extends": [
     "plugin:shopify/typescript",
     "plugin:shopify/react",
+    "plugin:shopify/eslint-comments",
     "plugin:shopify/jest",
     "plugin:shopify/node",
     "plugin:shopify/polaris",
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,7 @@
 {
   "extends": [
-    "plugin:shopify/typescript-react",
+    "plugin:shopify/typescript",
+    "plugin:shopify/react",
     "plugin:shopify/jest",
     "plugin:shopify/node",
     "plugin:shopify/polaris",
@@ -51,15 +52,17 @@
   },
   "overrides": [
     {
-      "files": ["*.js"],
+      "files": ["examples/**/*.js"],
       "rules": {
-        "typescript/no-var-requires": "off"
+        "import/no-extraneous-dependencies": "off",
+        "import/no-unresolved": "off"
       }
     },
     {
       "files": ["playground/Playground.tsx"],
       "rules": {
-        "react/prefer-stateless-function": "off"
+        "react/prefer-stateless-function": "off",
+        "shopify/react-initialize-state": "off"
       }
     }
   ]
--- a/.eslintrc
+++ b/.eslintrc
@@ -4,7 +4,7 @@
     "plugin:shopify/jest",
     "plugin:shopify/node",
     "plugin:shopify/polaris",
-    "plugin:shopify/typescript-prettier"
+    "plugin:shopify/prettier"
   ],
   "settings": {
     "react": {
@@ -26,9 +26,12 @@
         "allowBlockStart": false
       }
     ],
+    "import/no-cycle": "off",
+    "import/no-named-as-default": "off",
     "react/button-has-type": "off",
     "react/no-array-index-key": "off",
     "shopify/jest/no-vague-titles": "off",
+    "shopify/jsx-no-complex-expressions": "off",
     "jsx-a11y/label-has-for": [
       2,
       {
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -1,15 +1,15 @@
 /* eslint-disable no-console */
 
 import {execSync} from 'child_process';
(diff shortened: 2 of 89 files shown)


.eslintrc
```diff
@@ -26,9 +28,12 @@
         "allowBlockStart": false
       }
     ],
+    "import/no-cycle": "off",
```
BPScott: Initially disabled rules!

* `import/no-cycle` shall be enabled in a follow-up PR as this one is already noisy enough as-is, and fixing it will require a lot of changes and there's potential for weirdness as I unpick the cycles.
* `import/no-named-as-default` I'm not really sure what to do with. This currently has lots of errors and fixing it may require rethinking how we we write imports/exports. It doesn't help that I think the rule itself might be a bit buggy.
* `shopify/jsx-no-complex-expressions` we probably want to keep disabled, This has 10 violations currently and in those cases we can't avoid it without some small rewrites that are more complex than I am comfortable putting in . PR with a bunch of brainless changes.

scripts/build.js
```diff
@@ -38,9 +38,6 @@ rm('-rf', resolvePath(root, 'types/src'));
 
 mv(resolvePath(intermediateBuild, 'src/*'), intermediateBuild);
 
-const srcReadme = resolvePath(root, './src/components/README.md');
```
BPScott: These variables were never used

scripts/pa11y.js
```diff
@@ -62,7 +61,7 @@ async function runPa11y() {
     ),
   );
 
-  urls.map((path) => {
+  urls.forEach((path) => {
```
BPScott: This stops eslint complaining that there is no return value as in this case we don't care about the return value

src/components/AppProvider/utils/index.tsx
```diff
@@ -139,6 +138,7 @@ export function withSticky() {
       | React.ComponentClass<OwnProps & WithAppProviderProps> & C
       | React.SFC<OwnProps & WithAppProviderProps> & C,
   ): any & C {
+    // eslint-disable-next-line shopify/react-initialize-state
```
BPScott: I'm not really sure what the correct fix is here so I hand waved it away as this is deep dark typescript magic that I didn't want to mess with

src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -42,7 +42,6 @@ export default class ColorPicker extends React.PureComponent<Props, State> {
       return;
     }
 
-    // eslint-disable-next-line react/no-did-mount-set-state
```
BPScott: calling setState in componentDidMount shouldn't happen that often but it is a legitimate use according to the react docs. See https://github.com/yannickcr/eslint-plugin-react/issues/1754 for context.

src/components/Page/tests/Page.test.tsx
```diff
@@ -5,7 +5,9 @@ import {Page, DisplayText, Card} from 'components';
 import {noop} from '../../../utilities/other';
 import {LinkAction} from '../../../types';
 import {Header} from '../components';
+// eslint-disable-next-line shopify/strict-component-boundaries
```
BPScott: I'm not 100% sure what the right fix for this is, I think it might be pulling these types up into a higher level, as they're referenced in some other places too. But that's a fix more complex than what I want to do in this PR

.eslintrc
kvendrik: All sounds good 👍 I think we might wanna create follow up issues for all of these so we don't forget about them.

src/components/AppProvider/utils/index.tsx
kvendrik: I think there is an issue with the order of the generics here. `<{}, OwnProps & WithAppProviderProps>` would mean `{}` is the props and `OwnProps & WithAppProviderProps` is the state which is probably why you're getting this error. Try changing it to `<OwnProps & WithAppProviderProps, never>`.

BPScott: I'm scared, and @AndrewMusgrave was poking around this area in https://github.com/Shopify/polaris-react-deprecated/pull/2274.

I think that PR might solve this, and then because we've got the eslint-comments plugin it'll remind us to delete the ignore line if it is not needed

PR #557, opened 2018-11-06
Update Avatar and Image component to handle onError with images.

dleroux:

Resolves https://github.com/Shopify/shopify/issues/174824

The default option we want for our Avatar component is to show the  logged in user initials if an avatar image is not found. Gravatar gives us the possibility of providing a default image when an image is not found or to return a 404. 

In this [pr](https://github.com/Shopify/web/pull/5377) we attempted to return a transparent image. This was done with the following option in the query:

```
avatar(fallback: TRANSPARENT) {
      src
    }
```

The idea being that if the fallback was a transparent image it would simply lay over the Initials and wouldn't be visible.

However the Initials are in circle coloured background. As a result when an image was present it would show through [pr](https://github.com/Shopify/polaris-react-deprecated/pull/1157)

This PR is taking the is relying on gravatar returning a 404.

```
avatar(fallback: NOT_FOUND) {
      src
    }
```

Both the Image component and Avatar component have an onError prop now. If an image requested returns a 404 the onError attribute of the image tag will trigger the onError and no image will be outputted.

This is one approach. Another approach would be to have web check for the image, or to have the graphQl return a boolean. identifying if the fall back was used. I see some benefits of making the change in Polaris as well as web. 

I'm putting up this PR for discussion on the approach. I will polish and update tests if we chose to go this route.

## 

### How to 🎩

<details>

```jsx
import {Page, Avatar, Image, Stack} from '@shopify/polaris';

interface State {}

  render() {
    return (
      <Page title="Playground">
        <Stack>
          <Avatar size="large" initials="DL" customer />
          <Avatar
            size="large"
            initials="DL"
            source="https://avatars2.githubusercontent.com/u/1229901?s=80&v=4"
          />
          <Avatar size="large" initials="DL" source="https://bla" />
        </Stack>
        <br />
        <Image
          alt="Just an image"
          source="https://avatars2.githubusercontent.com/u/1229901?s=80&v=4"
        />
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -6,6 +6,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- `Avatar` falls back to `initials` when the image fails to load ([#557](https://github.com/Shopify/polaris-react/pull/557))
+
 ### Bug fixes
 
 ### Documentation
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import {classNames, variationName} from '@shopify/react-utilities/styles';
+import {autobind} from '@shopify/javascript-utilities/decorators';
 
 import {withAppProvider, WithAppProviderProps} from '../AppProvider';
 import Image from '../Image';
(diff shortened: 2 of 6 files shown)


src/components/Avatar/Avatar.tsx
```diff
@@ -32,85 +34,116 @@ export interface Props {
...
+      finalSource = source;
+    } else if (customer) {
+      finalSource = customerPlaceholder(nameString);
+    }
+
+    const className = classNames(
+      styles.Avatar,
+      styles[variationName('style', styleClass(nameString))],
+      size && styles[variationName('size', size)],
+      finalSource && !hasError && styles.hasImage,
```
dleroux: will only apply the styles if the image loading is successful

src/components/Avatar/Avatar.tsx
```diff
@@ -32,85 +34,116 @@ export interface Props {
...
-  } else if (name) {
-    label = name;
-  } else if (initials) {
-    const splitInitials = initials.split('').join(' ');
-    label = intl.translate('Polaris.Avatar.labelWithInitials', {
-      initials: splitInitials,
-    });
-  } else {
-    label = intl.translate('Polaris.Avatar.label');
+export class Avatar extends React.PureComponent<CombinedProps, State> {
```
dleroux: Most of this has not changed except for handling of the new state.

src/components/Image/Image.scss
```diff
@@ -0,0 +1,7 @@
+.hidden {
```
dleroux: I added the CSS because of the re-rendering the image would show a broken image icon in the browser.

src/components/Image/Image.tsx
```diff
@@ -7,34 +14,104 @@ export interface SourceSet {
 
 export type CrossOrigin = 'anonymous' | 'use-credentials' | '' | undefined;
 
+export interface State {
+  hasError: boolean;
+  hasLoaded: boolean;
```
dleroux: has Loaded is only set to true if the image loads properly. This is used to change the style on the image.

src/components/Image/Image.tsx
```diff
@@ -7,34 +14,104 @@ export interface SourceSet {
...
+}
+
 export interface Props extends React.HTMLProps<HTMLImageElement> {
   alt: string;
+  className: string;
```
dleroux: className was being passed down, now we need it in here as well for the visibility.

tmlayton:
> I see some benefits of making the change in Polaris as well as web

Can you elaborate on what you think the pros and cons are for fixing here or in web?

I’m of two minds here. My initial thought is this problem should be solved by the implementor and it is their responsibility to provide a valid source. Doing heavy lifting and magic in a component can be great if you get it right, but confusing and may seem buggy if done wrong. In this PR thus far, you are opinionated that an `<Image />` should not render at all once `hasError` is true. Are we sure this is what everyone wants? This could potentially swallow 404 errors and prevent someone from understanding what is happening.

That being said, I do like `<Avatar />` being resilient considering the API is already designed to have a fallback. Also, adding the `onError` handler to `<Image />` is nice.

So after some more thought I’m in favor of this, but I think the approach needs more consideration. Will make a few comments.

src/components/Image/Image.tsx
```diff
@@ -7,34 +14,104 @@ export interface SourceSet {
...
+      source,
+      crossOrigin,
+      onError,
+      className,
+      ...rest
+    }: Props = this.props;
+    const {hasError = false, hasLoaded = false} = this.state;
+
+    if (hasError) {
+      return null;
```
tmlayton: I don’t think we should do this or any of the load and error state tracking. I would pass the `onError` handler directly to `img` and leave it at that. It feels like we are trying to do too much and smells weird to me.

dleroux: I don't see how we'd get around this.  The nice this thing is that someone could provide a custom image for images that are not found.

dleroux:
Essentially if we do this in web they would need to do the same we do. Check if the image is valid (or returns a 404) then re-render to omit the source.

Like you I felt that not displaying a broken image icon could lead people wondering. However the onError prop could be useful to provide a fallback.

That said, maybe I can simply control the visibility of the Image in Avatar as oppose to Image. That way if an Image is used on it's own it will show the a broken image image icon,  but Avatar won't which is ok because of the fallback initials. What do you think?

tmlayton:
> That said, maybe I can simply control the visibility of the Image in Avatar as oppose to Image. That way if an Image is used on it's own it will show the a broken image image icon, but Avatar won't which is ok because of the fallback initials. What do you think?

I like this approach!

src/components/Avatar/tests/Avatar.test.tsx
```diff
@@ -0,0 +1,49 @@
+import * as React from 'react';
+import {mountWithAppProvider, trigger} from 'test-utilities';
+import {Image} from 'components';
+import Avatar from '../Avatar';
+
+describe('<Avatar />', () => {
+  describe('intials', () => {
+    it('renders intials if not image is provided ', () => {
+      const avatar = mountWithAppProvider(<Avatar initials="DL" />);
+      expect(avatar.find('span[role="img"] span svg').exists()).toBeTruthy();
```
AndrewMusgrave:
```suggestion
      expect(avatar.find('span[role="img"] span svg')).toHaveLength(1);
```

src/components/Avatar/tests/Avatar.test.tsx
```diff
@@ -0,0 +1,49 @@
+import * as React from 'react';
+import {mountWithAppProvider, trigger} from 'test-utilities';
+import {Image} from 'components';
+import Avatar from '../Avatar';
+
+describe('<Avatar />', () => {
+  describe('intials', () => {
+    it('renders intials if not image is provided ', () => {
```
AndrewMusgrave:
```suggestion
    it('renders intials if the image is not provided ', () => {
```

src/components/Avatar/tests/Avatar.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+    });
+  });
+
+  describe('on Error with Intials', () => {
+    it('renders initials if the Image onError prop is triggered and the Intials are provided.', () => {
+      const src = 'image/path/';
+      const avatar = mountWithAppProvider(
+        <Avatar size="large" initials="DL" source={src} />,
+      );
+      expect(avatar.find('span[role="img"] span svg').exists()).toBeFalsy();
```
AndrewMusgrave:
```suggestion
      expect(avatar.find('span[role="img"] span svg')).toHaveLength(0);
```

src/components/Avatar/tests/Avatar.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+      trigger(avatar.find(Image), 'onError');
+      expect(avatar.find('span[role="img"] span svg').exists()).toBeTruthy();
```
AndrewMusgrave:
```suggestion
      expect(avatar.find('span[role="img"] span svg')).toHaveLength(1);
```

src/components/Avatar/Avatar.tsx
```diff
@@ -34,83 +35,122 @@ export interface Props {
...
+      finalSource = source;
+    } else if (customer) {
+      finalSource = customerPlaceholder(nameString);
+    }
+
+    const className = classNames(
+      styles.Avatar,
+      styles[variationName('style', styleClass(nameString))],
+      size && styles[variationName('size', size)],
+      hasImage && !hasLoaded && styles.hidden,
```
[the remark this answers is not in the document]
dleroux: It work work in this case because we won't want the hasImage class if there is no image.

UNRELEASED.md
```diff
@@ -50,6 +50,7 @@ The autocomplete component is an input field that provides selectable suggestion
 - Added `onPortalCreated` prop to `Portal`
 - Improved consistency of `Badge` styling
 - Explicitly specifying `list-style` on `List`
+- `Avatar` now fallsback to `initials` when the image fails to load ([#557](https://github.com/Shopify/polaris-react/pull/557))
```
```suggestion
- `Avatar` falls back to `initials` when the image fails to load ([#557](https://github.com/Shopify/polaris-react/pull/557))
```

danrosenthal:
@dleroux what's the status on this PR? Do you need another round of review?

dleroux:
@danrosenthal should be good to go but @tmlayton had commented on it so just waiting on his approval of the changes. I've pinged him on it.

src/components/Avatar/Avatar.tsx
```diff
@@ -34,83 +35,124 @@ export interface Props {
   accessibilityLabel?: string;
 }
 
+export interface State {
```
tmlayton: Do we always export our `State` interfaces?

src/components/Avatar/Avatar.tsx
```diff
@@ -34,83 +35,124 @@ export interface Props {
...
-  } else if (initials) {
-    const splitInitials = initials.split('').join(' ');
-    label = intl.translate('Polaris.Avatar.labelWithInitials', {
-      initials: splitInitials,
-    });
-  } else {
-    label = intl.translate('Polaris.Avatar.label');
+export class Avatar extends React.PureComponent<CombinedProps, State> {
+  state: State = {
+    hasError: false,
```
tmlayton: image

tmlayton: Sorry, I could not resist 😄

src/components/Avatar/Avatar.tsx
```diff
@@ -34,83 +35,124 @@ export interface Props {
...
+    hasLoaded: false,
+  };
+
+  ComponentDidUpdate({source: previousSource}: Props) {
```
tmlayton: Just noticed this is misspelled, and also why the linter is not complaining about setting state in the update lifecycle

tmlayton:
```suggestion
  componentDidUpdate({source: previousSource}: Props) {
```

src/components/Avatar/Avatar.tsx
dleroux: yes we do

dleroux:
Awesome thanks. :shipit: ?

tmlayton:
:shipit:

PR #573, opened 2018-11-09
[TextField: Spinner] Support press and hold to increment/decrement value

### LINKED ISSUE #420: Spinner on numeric TextField doesn't support press-and-hold
# Feature request summary
Make spinner buttons on numeric text fields respond to being pressed and held down, instead of only firing a single click event when the button is released.

## Rationale
Currently pressing and holding the spinner on a `TextField` with `type="number"` doesn't continually increment/decrement the value displayed:

numeric-text-field-spinner

However, we feel that most users would expect to be able to press and hold the spinner buttons to change the value more quickly than having to click multiple times. For example, in one of our apps, we have numeric text fields which control the positioning of a component on screen, and in that case, being able to hold the spinner down to "slide" the component around the screen is a useful ability to have.

tanhengyeow:
Resolves #420

Support press and hold to increment/decrement value in `Spinner` of the numeric `TextField`.

## 

<details>

```jsx

interface State {}

  state = {
    value: '1',
  };

  handleChange = (value) => {
    this.setState({value});
  };

  render() {
    return (
      <Page title="Playground">
        <TextField
          label="Quantity"
          type="number"
          value={this.state.value}
          onChange={this.handleChange}
        />
      </Page>
    );
  }
}
```

</details>

### DIFF
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -120,6 +120,7 @@ export default class TextField extends React.PureComponent<Props, State> {
   }
 
   private input: HTMLElement;
+  private buttonPressTimer: number;
 
   constructor(props: Props) {
     super(props);
@@ -205,7 +206,11 @@ export default class TextField extends React.PureComponent<Props, State> {
 
     const spinnerMarkup =
       type === 'number' && !disabled ? (
-        <Spinner onChange={this.handleNumberChange} />
+        <Spinner
+          onChange={this.handleNumberChange}
+          onMouseDown={this.handleButtonPress}
+          onMouseUp={this.handleButtonRelease}
+        />
       ) : null;
 
     const style = multiline && height ? {height} : null;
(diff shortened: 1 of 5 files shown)


src/components/TextField/TextField.tsx
```diff
@@ -383,6 +388,16 @@ export default class TextField extends React.PureComponent<Props, State> {
   private handleClick() {
     this.input.focus();
   }
+
+  @autobind
+  private handleButtonPress(onChange: Function) {
+    this.buttonPressTimer = setInterval(onChange, 200);
```
tmlayton: `200` might be a bit long and I think we should consider adding a `setTimeout` first. That way when clicking and holding, there is an initial delay of maybe `500` then the number quickly increments maybe at an interval of `100`. Thoughts?

sarahill: ☝️ I like this suggestion of gradually increasing the speed. Overall, this looks great!

dpersing:
Hi @tanhengyeow! Are you able to provide any updates around tests? Thanks!

tanhengyeow:
Hey everyone, thanks for the review :smile:

The suggestion sounds great and I'll look into adding tests :) I'm having exams now so I can only get back to this during December, hope that's fine :D

danrosenthal:
> I'm having exams now so I can only get back to this during December, hope that's fine

Totally fine. I'll add `[WIP]` to the title to indicate that this pull request is a work in progress.

danrosenthal:
I might be missing something here, but when I tophat, it still just seems to increment and decrement by one digit.

tmlayton:
When you `git log` do you see the latest commits? Depending on how you checked this out an upstream might not be set.

PR #582, opened 2018-11-09
Format breakpoints.scss

### LINKED ISSUE #571: File 'breakpoints.scss' is flagged by Prettier
# Issue summary
File `breakpoints.scss` is flagged by Prettier after running `yarn format`.

## Expected behavior
Running `yarn format` should not result in format changes to files one hasn't made yet.

## Actual behavior
Running `yarn format` resulted in `src/styles/shared/breakpoints.scss` to reflect suggested format changes.

1. Run `yarn format` on `master`

## Specifications

- Polaris version number: 2.x
- Browser:
- Device:
- Operating System:

BPScott:

Something weird is happening with stylelint where it wants to reformat
theses line when doing --fix, despite there being no issues reported
when running a regular lint. This is weird and confusing but the change
is straightforward enough, so lets allow it for the sake of `yarn
format` doing nothing when `yarn lint` reports no errors.

Fixes #571

Run `yarn format`


BPScott:
Long-term fix is to add that `at-rule-name-space-after` as a rule to be disabled because prettier handles it in stylelint-config-prettier

PR #587, opened 2018-11-09
[WIP] Sample component-level accessibility content

dpersing:

Provides sample WIP component-level accessibility recommendations for review and discussion.

Edits `README.md` files for example components to include samples of language around accessibility as part of exploration for https://unicorn.shopify.io/projects/6177.

See https://github.com/Shopify/polaris-styleguide/pull/2377 for higher level documentation examples.

### DIFF
--- a/src/components/Button/README.md
+++ b/src/components/Button/README.md
@@ -256,6 +256,8 @@ Use when a button has been pressed and the associated action is in progress.
 
 <!-- content-for: web -->
 
+---
+
 ### Accessibility
 
 #### Labeling
--- a/src/components/Button/README.md
+++ b/src/components/Button/README.md
@@ -30,7 +30,7 @@ keywords:
 
 # Button
 
-Buttons are used to make common actions immediately visible and easy to perform with one click or tap. Merchants can use it to navigate, or take action.
+Buttons are used to make common actions immediately visible and easy to perform with one click, tap or keypress. Merchants can use it to navigate, or take action.
 
 ---
 
@@ -253,6 +253,29 @@ Use when a button has been pressed and the associated action is in progress.
 ```jsx
 <Button loading>Save product</Button>
 ```
+
+<!-- content-for: web -->
+
+### Accessibility
+
+#### Labeling
+
+The `accessibilityLabel` prop adds an `aria-label` attribute to the Button, which can be accessed by assistive technologies like screen readers. Typically, this text will be read in place of the visible text on the button for assistive technology users.
+
+For a Button, `accessibilityLabel` should be used if:
+
+- The visible text of the Button is not sufficient for non-visual users to understand the purpose of the Button in the context of the page.
+- The Button has no text and relies on an icon instead to convey its purpose.
+
+To help support people who use speech activation software and sighted screen reader users, make sure that the `aria-label` text includes any button text that is visible. Mismatches in visible and programmatic labeling can cause confusion, and can cause voice recognition commands to not work.
+
+When possible, give the button visible text that clearly conveys its purpose without the use of `accessibilityLabel`. Duplicating the Button text with `accessibilityLabel` when no additional content is needed is not necessary.
+
+#### Keyboard support
+
+Keyboard users will expect to be able to give buttons keyboard focus with the `tab` key, and to activate them with the `enter`/`return` and `space` keys.
+
+<!-- /content-for -->
 
 ---
 
--- a/src/components/DatePicker/README.md
+++ b/src/components/DatePicker/README.md
@@ -102,3 +102,11 @@ class DatePickerExample extends React.Component {
 ![Date picker on iOS](components/DatePicker/ios/default.png)
 
 <!-- /content-for -->
+
+<!-- content-for: web -->
+
+#### Accessibility
+
+Date picker interactions can be challenging for some users and very easy for others. To give users a choice, consider allowing users to enter the date using a [Text field](/components/forms/text-field), with the datepicker provided as optional.
+
+<!-- /content-for -->
--- a/src/components/Heading/README.md
+++ b/src/components/Heading/README.md
@@ -103,6 +103,16 @@ Use for the title of each top-level page section.
 
 <!-- /content-for -->
 
+<!-- content-for: web -->
+
+---
+
+## Accessibility
+
+Headings default to an `<h2>` in HTML. Typically, major sections of a page should have a level 2 heading. A clear and consistent heading structure helps people who have difficulty with reading or language, and also helps screen reader users navigate the page using keystrokes custom to their screen reader.
+
+<!-- /content-for -->
+
 ---
 
 ## Related components
--- a/src/components/InlineError/README.md
+++ b/src/components/InlineError/README.md
@@ -82,6 +82,24 @@ Use when the merchant has entered invalid information into multiple fields insid
 
 <!-- /content-for -->
 
+<!-- content-for: web -->
+
+---
+
+## Accessibility
+
+### Structure
+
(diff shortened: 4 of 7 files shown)


src/components/Button/README.md
```diff
@@ -30,7 +30,7 @@ keywords:
 
 # Button
 
-Buttons are used to make common actions immediately visible and easy to perform with one click or tap. Merchants can use it to navigate, or take action.
+Buttons are used to make common actions immediately visible and easy to perform with one click, tap or keypress. Merchants can use it to navigate, or take action.
```
dpersing: Example keyboard-inclusive language.

src/components/Button/README.md
```diff
@@ -254,6 +254,31 @@ Use when a button has been pressed and the associated action is in progress.
 <Button loading>Save product</Button>
 ```
 
+<!-- content-for: web -->
```
dpersing: Example section summarizing accessibility-related information for the component, including the recommended use of props and expected interactions.

src/components/DatePicker/README.md
```diff
@@ -102,3 +102,11 @@ class DatePickerExample extends React.Component {
 ![Date picker on iOS](components/DatePicker/ios/default.png)
 
 <!-- /content-for -->
+
+<!-- content-for: web -->
+
+#### Accessibility
```
dpersing: The Date picker could use the same keyboard interaction treatment, but I also want to be conscientious of how we call out known issues with a component. How detailed should we get about behavior or functionality that should be present but isn't, and has issues logged?

For this example I chose to focus on larger context around fitting the component into the functionality of the page.

src/components/Heading/README.md
```diff
@@ -103,6 +103,16 @@ Use for the title of each top-level page section.
 
 <!-- /content-for -->
 
+<!-- content-for: web -->
+
+---
+
+## Accessibility
```
dpersing: Some general recs for headings. I used the same content, tweaked slightly, for both the Heading and Subheading components.

src/components/InlineError/README.md
```diff
@@ -82,6 +82,24 @@ Use when the merchant has entered invalid information into multiple fields insid
 
 <!-- /content-for -->
 
+<!-- content-for: web -->
+
+---
+
+## Accessibility
+
+### Structure
```
dpersing: This example gives context for HTML output that users may not be familiar with, and provides rationale for using Inline errors instead of some other text type. I also called out a visual design choice that aids accessibility to help dissuade consumers from hiding the icon.

src/components/Tabs/README.md
```diff
@@ -195,3 +195,23 @@ Also known as [Segmented controls](https://developer.apple.com/design/human-inte
 ![Fixed tabs on iOS](components/Tabs/ios/fixed.png)
 
 <!-- /content-for -->
+
+<!-- content-for: web -->
+
+---
+
+## Accessibility
```
dpersing: Here's an example where I go more into depth about the structure of the feature. I could go either way with relying on the linked spec or including specific detail with the component. If we go in the broader direction of linking to known issues or PRs for a specific component, the structure and keyboard detail might help, I think.

src/components/Button/README.md
```diff
@@ -254,6 +254,31 @@ Use when a button has been pressed and the associated action is in progress.
...
+
+---
+
+### Accessibility
```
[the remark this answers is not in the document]
dpersing: Yup, typo on my part!

src/components/Subheading/README.md
```diff
@@ -88,6 +88,16 @@ Use to structure content in a card.
 
 ---
 
+<!-- content-for: web -->
+
+---
+
+## Accessibility
+
+Subheadings default to an `<h3>` in HTML. Typically, subsections of a page should have a level 3 heading. A clear and consistent heading structure helps people who have difficulty with reading or language, and also helps screen reader users navigate the page using keystrokes custom to their screen reader.
```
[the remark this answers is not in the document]
dpersing: Super good points. Folks probably go with the defaults normally.

src/components/DatePicker/README.md
danrosenthal: I'll give this some thought and get back to you

AndrewMusgrave:
@dpersing do you need re-reviews or anything to help push this along?

dpersing:
> @dpersing do you need re-reviews or anything to help push this along?

Emerging from my 🦃 coma to wrap up this phase based on feedback from folks. Will likely close this PR this week and open individual ones for each component to get published as part of the next phase.

src/components/Heading/README.md
```diff
@@ -103,6 +103,16 @@ Use for the title of each top-level page section.
...
+
+Headings default to an `<h2>` in HTML. Typically, major sections of a page should have a level 2 heading. A clear and consistent heading structure helps people who have difficulty with reading or language, and also helps screen reader users navigate the page using keystrokes custom to their screen reader.
+
```
danrosenthal: This content feels like it could be more actionable. Dom's feedback below about referencing the `element` prop makes sense to me.

src/components/Tabs/README.md
danrosenthal: IMO, linking to the spec is the right approach here. This says that we didn't just imagine the implementation, but that we are actually following industry best practices. Also the spec is pretty safe to rely on always being there, and not changing or becoming stale. I'm still iffy about linking to a specific issue or PR in the styleguide, as their lifecycle is far shorter, and more at risk of becoming stale.

In general I would think about these things in terms of lifecycle, where a resource with a long lifecycle should not link back to a resource with a shorter lifecycle. In terms of lifecycles, I see it kind of like this, ranging from shortest lifespan to longest:

`pull request > issue > component readme/styleguide page`

Other elements I see with a similar lifespan to READMEs would be a master component issue (hub with spokes to other short lifespan issues), and component metadata.

Happy to expand on this 1:1 if it makes sense to do so.

src/components/DatePicker/README.md
dpersing: > Maybe the answer is some sort of boilerplate caveats section that links to a master issue, and is just part of every accessibility section.

I like the idea of having a boilerplate caveat in general, since the accessibility status of a given implementation might change for an individual user based on a software update, and there will always need to be some unknowns.

Linking to the ARIA spec as the default, and going into further detail if we deviate from it, also helps avoid the case where we need to tweak documentation more frequently.

src/components/Button/README.md
```diff
@@ -254,6 +254,31 @@ Use when a button has been pressed and the associated action is in progress.
...
+The `accessibilityLabel` prop adds an `aria-label` attribute to the Button, which can be accessed by assistive technologies like screen readers. Typically, this text will be read in place of the visible text on the button for assistive technology users.
+
+For a Button, `accessibilityLabel` should be used if:
+
+- The visible text of the Button is not sufficient for non-visual users to understand the purpose of the Button in the context of the page.
+- The Button has no text and relies on an icon instead to convey its purpose.
+
+To help support people who use speech activation software and sighted screen reader users, make sure that the `aria-label` text includes any button text that is visible. Mismatches in visible and programmatic labeling can cause confusion, and can cause voice recognition commands to not work.
+
+When possible, give the button visible text that clearly conveys its purpose without the use of `accessibilityLabel`. Duplicating the Button text with `accessibilityLabel` when no additional content is needed is not necessary.
```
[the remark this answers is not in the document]
dpersing: I like the idea of using that format for more binary or "if this than that" decision making.

dpersing:
Closing this WIP branch to replace it with individual branches for individual component updates.

PR #600, opened 2018-11-12
[Toggle] Add toggle component

solonaarmstrong:

Resolves https://github.com/Shopify/polaris-react/issues/594

Adds a new toggle component, based off [`Switch` in web](https://github.com/Shopify/web/blob/master/app/components/Switch/Switch.tsx).

I changed a few things. The disabled color uses `sky` instead of `ink` `lightest` because the contrast between enabled and disabled seemed difficult to distinguish for those with visual impairments. I added a border to compensate for this change.

I also used our transition speeds and easing.

I added some props:
`accessibilityLabel`
`label`
`labelHidden`
`prefix`
`suffix`

And I removed `name`.

toggle

## 

### How to 🎩

<details>

```jsx
import {Page, Toggle} from '@shopify/polaris';

interface State {
  checked: boolean;
}

  state: State = {
    checked: true,
  };

  toggleToggle = (value: boolean) => {
    this.setState({checked: value});
  };

  render() {
    return (
      <Page title="Playground">
        <Toggle
          label="Label"
          labelHidden
          prefix="Off"
          suffix="On"
          checked={this.state.checked}
          onChange={this.toggleToggle}
        />
      </Page>
    );
  }
}

```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -4,6 +4,12 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ---
 
+### New components
+
+#### [Toggle](http://polaris.shopify.com/components/forms/toggle)
+
+- Use the toggle component to present a single choice to enable or disable a setting. ([#600](https://github.com/Shopify/polaris-react/pull/600))
+
 ### Enhancements
 
 ### Bug fixes
--- a/src/components/Toggle/README.md
+++ b/src/components/Toggle/README.md
@@ -19,7 +19,7 @@ keywords:
 
 # Toggle
 
-Use to present a single choice to enable or disable a setting.
+Use to present a single choice to enable or disable a setting. Toggle is best used when multiple settings can be enabled or disabled and then saved all at once.
 
 ---
 
@@ -27,8 +27,9 @@ Use to present a single choice to enable or disable a setting.
 
 Toggle should:
 
-- Always be used with an associated label component.
+- Be used when there are only two options available (enabled or disabled) and not a range of selections (zero, one, or multiple).
 - Be used to switch a single setting on and off.
+- Always be used with an associated label component.
 
 ---
 
@@ -77,22 +78,50 @@ Use toggle to let merchants enable or disable a setting.
 ```jsx
 class ToggleExample extends React.Component {
   state = {
-    checked: false,
+    visaChecked: true,
+    mcChecked: false,
   };
 
-  toggleToggle = (value) => {
-    this.setState({checked: value});
+  toggleVisaToggle = (value) => {
+    this.setState({visaChecked: value});
+  };
+
+  toggleMCToggle = (value) => {
+    this.setState({mcChecked: value});
   };
 
   render() {
     const {value} = this.state;
     return (
       <Stack vertical>
-        <Toggle
-          label="Accept email marketing"
-          checked={this.state.checked}
-          onChange={this.toggleToggle}
-        />
+        <Card title="Payment methods">
+          <Card.Section>
+            <Stack distribution="fillEvenly">
+              <Stack.Item>Visa</Stack.Item>
+              <Stack.Item>Credit card</Stack.Item>
+              <Stack.Item>2.99% + $0.30</Stack.Item>
+              <Toggle
+                label=""
+                labelHidden
+                checked={this.state.visaChecked}
+                onChange={this.toggleVisaToggle}
+              />
+            </Stack>
+          </Card.Section>
+          <Card.Section>
+            <Stack distribution="fillEvenly">
+              <Stack.Item>MasterCard</Stack.Item>
+              <Stack.Item>Credit card</Stack.Item>
+              <Stack.Item>2.99% + $0.30</Stack.Item>
+              <Toggle
+                label=""
+                labelHidden
+                checked={this.state.mcChecked}
+                onChange={this.toggleMCToggle}
+              />
+            </Stack>
+          </Card.Section>
+        </Card>
       </Stack>
     );
   }
@@ -103,29 +132,119 @@ class ToggleExample extends React.Component {
 
 <!-- example-for: web -->
 
-Use to provide more clarity to merchants about which setting they are selecting.
+Use to provide more clarity to merchants about which setting they are choosing.
 
 ```jsx
 class ToggleExample extends React.Component {
   state = {
-    checked: false,
+    visaChecked: true,
+    mcChecked: false,
   };
 
-  toggleToggle = (value) => {
-    this.setState({checked: value});
+  toggleVisaToggle = (value) => {
+    this.setState({visaChecked: value});
+  };
+
+  toggleMCToggle = (value) => {
+    this.setState({mcChecked: value});
   };
 
   render() {
     const {value} = this.state;
     return (
       <Stack vertical>
-        <Toggle
-          label="Accept email marketing"
-          prefix="Off"
-          suffix="On"
-          checked={this.state.checked}
-          onChange={this.toggleToggle}
-        />
+        <Card title="Payment methods">
+          <Card.Section>
+            <Stack distribution="fillEvenly">
+              <Stack.Item>Visa</Stack.Item>
+              <Stack.Item>Credit card</Stack.Item>
+              <Stack.Item>2.99% + $0.30</Stack.Item>
+              <Toggle
+                label=""
+                labelHidden
+                prefix="Disable"
+                suffix="Enable"
+                checked={this.state.visaChecked}
+                onChange={this.toggleVisaToggle}
+              />
+            </Stack>
+          </Card.Section>
+          <Card.Section>
+            <Stack distribution="fillEvenly">
+              <Stack.Item>MasterCard</Stack.Item>
+              <Stack.Item>Credit card</Stack.Item>
+              <Stack.Item>2.99% + $0.30</Stack.Item>
+              <Toggle
+                label=""
+                labelHidden
+                prefix="Disable"
+                suffix="Enable"
+                checked={this.state.mcChecked}
+                onChange={this.toggleMCToggle}
+              />
+            </Stack>
+          </Card.Section>
+        </Card>
+      </Stack>
+    );
+  }
+}
+```
+
+### Disabled toggle
+
+<!-- example-for: web -->
+
+Use for settings that aren’t currently able to change.
+
+```jsx
+class ToggleExample extends React.Component {
+  state = {
+    visaChecked: true,
+    mcChecked: false,
+  };
+
+  toggleVisaToggle = (value) => {
+    this.setState({visaChecked: value});
+  };
+
+  toggleMCToggle = (value) => {
+    this.setState({mcChecked: value});
+  };
+
+  render() {
+    const {value} = this.state;
+    return (
+      <Stack vertical>
+        <Card title="Payment methods">
+          <Card.Section>
+            <Stack distribution="fillEvenly">
+              <Stack.Item>Visa</Stack.Item>
+              <Stack.Item>Credit card</Stack.Item>
+              <Stack.Item>2.99% + $0.30</Stack.Item>
+              <Toggle
+                label=""
+                labelHidden
+                checked={this.state.visaChecked}
+                onChange={this.toggleVisaToggle}
+              />
+            </Stack>
+          </Card.Section>
+          <Card.Section>
+            <Stack distribution="fillEvenly">
+              <Stack.Item>MasterCard</Stack.Item>
+              <Stack.Item>Credit card</Stack.Item>
+              <Stack.Item>2.99% + $0.30</Stack.Item>
+              <Toggle
+                label=""
+                labelHidden
+                disable
+                checked={this.state.mcChecked}
+                onChange={this.toggleMCToggle}
+              />
+            </Stack>
+          </Card.Section>
+        </Card>
       </Stack>
     );
   }
@@ -136,6 +255,7 @@ class ToggleExample extends React.Component {
 
 ## Related components
 
+- To give merchants control over a feature or option that can be turned on or off, [use the setting toggle component](/components/actions/setting-toggle). Setting toggle is best used when the setting is saved immediately on change.
 - To present a list of options where merchants must make a single selection, [use the radio button component](/components/forms/radio-button)
 - To give merchants a way to make a range of selections (zero, one, or multiple) [use the checkbox component](/components/forms/checkbox)
 - To present merchants with a list of checkboxes, [use the choice list component](/components/forms/choice-list) with the “allow multiple” option
--- a/src/components/Toggle/README.md
+++ b/src/components/Toggle/README.md
@@ -79,50 +79,63 @@ Use toggle to let merchants enable or disable a setting.
 class ToggleExample extends React.Component {
   state = {
     visaChecked: true,
-    masterCardChecked: false,
+    americanExpressChecked: false,
   };
 
   toggleVisaToggle = (value) => {
     this.setState({visaChecked: value});
   };
 
-  toggleMasterCardToggle = (value) => {
-    this.setState({masterCardChecked: value});
+  toggleAmericanExpressToggle = (value) => {
+    this.setState({americanExpressChecked: value});
   };
 
   render() {
     const {value} = this.state;
     return (
-      <Stack vertical>
-        <Card title="Payment methods">
-          <Card.Section>
-            <Stack distribution="fillEvenly">
-              <Stack.Item>Visa</Stack.Item>
-              <Stack.Item>Credit card</Stack.Item>
-              <Stack.Item>2.99% + $0.30</Stack.Item>
+      <Card title="Payment methods">
+        <Card.Section>
+          <Stack distribution="fillEvenly">
+            <Stack.Item>
+              <Stack spacing="loose">
+                <Image source="/assets/toggle/visa.svg" alt="Visa" />
+                <Stack.Item>Visa</Stack.Item>
+              </Stack>
+            </Stack.Item>
+            <Stack.Item>Credit card</Stack.Item>
+            <Stack.Item>2.99% + $0.30</Stack.Item>
+            <Stack.Item>
               <Toggle
                 label=""
                 labelHidden
                 checked={this.state.visaChecked}
                 onChange={this.toggleVisaToggle}
               />
-            </Stack>
-          </Card.Section>
-          <Card.Section>
-            <Stack distribution="fillEvenly">
-              <Stack.Item>MasterCard</Stack.Item>
-              <Stack.Item>Credit card</Stack.Item>
-              <Stack.Item>2.99% + $0.30</Stack.Item>
-              <Toggle
-                label=""
-                labelHidden
-                checked={this.state.masterCardChecked}
-                onChange={this.toggleMasterCardToggle}
-              />
-            </Stack>
-          </Card.Section>
-        </Card>
-      </Stack>
+            </Stack.Item>
+          </Stack>
+        </Card.Section>
+        <Card.Section>
+          <Stack distribution="fillEvenly">
+            <Stack.Item>
+              <Stack spacing="loose">
+                <Image
+                  source="/assets/toggle/amex.svg"
+                  alt="American Express"
+                />
+                <Stack.Item>AMEX</Stack.Item>
+              </Stack>
+            </Stack.Item>
+            <Stack.Item>Credit card</Stack.Item>
+            <Stack.Item>2.99% + $0.30</Stack.Item>
+            <Toggle
+              label=""
+              labelHidden
+              checked={this.state.americanExpressChecked}
+              onChange={this.toggleAmericanExpressToggle}
+            />
+          </Stack>
+        </Card.Section>
+      </Card>
     );
   }
 }
@@ -138,54 +151,67 @@ Use to provide more clarity to merchants about which setting they are choosing.
 class ToggleExample extends React.Component {
   state = {
     visaChecked: true,
-    masterCardChecked: false,
+    americanExpressChecked: false,
   };
 
   toggleVisaToggle = (value) => {
     this.setState({visaChecked: value});
   };
 
-  toggleMasterCardToggle = (value) => {
-    this.setState({masterCardChecked: value});
+  toggleAmericanExpressToggle = (value) => {
+    this.setState({americanExpressChecked: value});
   };
 
   render() {
     const {value} = this.state;
     return (
-      <Stack vertical>
-        <Card title="Payment methods">
-          <Card.Section>
-            <Stack distribution="fillEvenly">
-              <Stack.Item>Visa</Stack.Item>
-              <Stack.Item>Credit card</Stack.Item>
-              <Stack.Item>2.99% + $0.30</Stack.Item>
+      <Card title="Payment methods">
+        <Card.Section>
+          <Stack distribution="fillEvenly">
+            <Stack.Item>
+              <Stack spacing="loose">
+                <Image source="/assets/toggle/visa.svg" alt="Visa" />
+                <Stack.Item>Visa</Stack.Item>
+              </Stack>
+            </Stack.Item>
+            <Stack.Item>Credit card</Stack.Item>
+            <Stack.Item>2.99% + $0.30</Stack.Item>
+            <Stack.Item>
               <Toggle
                 label=""
                 labelHidden
-                prefix="Disable"
-                suffix="Enable"
+                prefix="Disabled"
+                suffix="Enabled"
                 checked={this.state.visaChecked}
                 onChange={this.toggleVisaToggle}
               />
-            </Stack>
-          </Card.Section>
-          <Card.Section>
-            <Stack distribution="fillEvenly">
-              <Stack.Item>MasterCard</Stack.Item>
-              <Stack.Item>Credit card</Stack.Item>
-              <Stack.Item>2.99% + $0.30</Stack.Item>
-              <Toggle
-                label=""
-                labelHidden
-                prefix="Disable"
-                suffix="Enable"
-                checked={this.state.masterCardChecked}
-                onChange={this.toggleMasterCardToggle}
-              />
-            </Stack>
-          </Card.Section>
-        </Card>
-      </Stack>
+            </Stack.Item>
+          </Stack>
+        </Card.Section>
+        <Card.Section>
+          <Stack distribution="fillEvenly">
+            <Stack.Item>
+              <Stack spacing="loose">
+                <Image
+                  source="/assets/toggle/amex.svg"
+                  alt="American Express"
+                />
+                <Stack.Item>AMEX</Stack.Item>
+              </Stack>
+            </Stack.Item>
+            <Stack.Item>Credit card</Stack.Item>
+            <Stack.Item>2.99% + $0.30</Stack.Item>
+            <Toggle
+              label=""
+              labelHidden
+              prefix="Disabled"
+              suffix="Enabled"
+              checked={this.state.americanExpressChecked}
+              onChange={this.toggleAmericanExpressToggle}
+            />
+          </Stack>
+        </Card.Section>
+      </Card>
     );
   }
 }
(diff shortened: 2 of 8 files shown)


src/components/Toggle/tests/Toggle.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+import TextContainer from '../../TextContainer';
+import Toggle from '../Toggle';
+
+describe('<Toggle />', () => {
+  const mockProps = {
+    label: 'label',
+    onChange: noop,
+  };
+
+  describe('accessibilityLabel', () => {
```
mehdi-loup: You might want to test all accessibility props in this `describe` (ie. aria-checked and  aria-disabled)

solonaarmstrong: @mehdi-loup Did you have any thoughts about the change in background color?

mehdi-loup: You mean the transition for the bg color? I think it is great you added this.

For the colours themselves, if that is your question, I remember that Fred Audet validated them.

src/components/Toggle/README.md
```diff
@@ -0,0 +1,142 @@
...
+Use to present a single choice to enable or disable a setting.
+
+---
+
+## Best practices
+
+Toggle should:
+
+- Always be used with an associated label component.
+- Be used to switch a single setting on and off.
```
mehdi-loup: It would be good to help developers figure how to choose between a toggle and a checkbox. What should drive that decision? Is it the surrounding context, the space available, the size of the label,..?

src/components/Toggle/README.md
```diff
@@ -0,0 +1,142 @@
...
+---
+
+## Examples
+
+### Default toggle
+
+Use toggle to let merchants enable or disable a setting.
+
+```jsx
+class RadioButtonExample extends React.Component {
```
danrosenthal:
```suggestion
class ToggleExample extends React.Component {
```

src/components/Toggle/README.md
```diff
@@ -0,0 +1,142 @@
...
+  render() {
+    const {value} = this.state;
+    return (
+      <Stack vertical>
+        <Toggle
+          label="Accept email marketing"
+          labelHidden
+          checked={this.state.checked}
+          onChange={this.toggleToggle}
+        />
```
danrosenthal: This example feels a bit too close to the example we're using for a Checkbox. What is the differentiator here? Shouldn't this be more like a settings toggle in the admin?

src/components/Toggle/README.md
```diff
@@ -0,0 +1,142 @@
...
+      <Stack vertical>
+        <Toggle
+          label="Accept email marketing"
+          labelHidden
+          prefix="Off"
+          suffix="On"
+          checked={this.state.checked}
+          onChange={this.toggleToggle}
+        />
+      </Stack>
```
danrosenthal: This feels like it should be the default state for this component. It's far more accessible and creates a better UX when we provide these labels.

src/components/Toggle/README.md
```diff
@@ -0,0 +1,142 @@
...
+  }
+}
+```
+
+---
+
+## Related components
+
+- To present a list of options where merchants must make a single selection, [use the radio button component](/components/forms/radio-button)
+- To present merchants with a list of checkboxes, [use the choice list component](/components/forms/choice-list) with the “allow multiple” option
```
danrosenthal: Would the `checkbox` not be a related component?

src/components/Toggle/Toggle.scss
```diff
@@ -0,0 +1,84 @@
+$toggle-track-length: rem(46px);
+$toggle-track-height: rem(24px);
+$toggle-thumb-size: rem(20px);
+
+.Toggle {
+  display: flex;
+  align-items: center;
+
+  &.disabled {
+    opacity: 0.8;
```
danrosenthal: This number feels a bit magical. Should we put it in a variable with a descriptive name?

src/components/Toggle/Toggle.scss
```diff
@@ -0,0 +1,84 @@
...
+  border-radius: border-radius(extra-large);
+  border: border-width() solid color('sky', 'dark');
+  background-color: color('sky');
+  transition: background-color duration(slow), border duration(slow);
+  -webkit-tap-highlight-color: transparent;
+}
+
+.ToggleThumb {
+  position: absolute;
+  top: rem(2px);
```
danrosenthal: Let's throw these guys in a descriptive variable as well.

src/components/Toggle/Toggle.scss
```diff
@@ -0,0 +1,84 @@
...
+  border-radius: 50%;
+  transition: left duration(slow) easing();
+  background-color: color('white');
+}
+
+.ToggleChecked {
+  .ToggleTrack {
+    background-color: color('indigo');
+    border: border-width() solid color('indigo', 'dark');
+    outline: none;
```
danrosenthal: Why are we removing the outline on this?

src/components/Toggle/Toggle.scss
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  .ToggleThumb {
+    left: $toggle-track-height;
+  }
+}
+
+.ToggleDisabled {
+  cursor: not-allowed;
```
danrosenthal: Hmm, why `not-allowed` here as opposed to another cursor property? What does this give us?

src/components/Toggle/Toggle.scss
```diff
@@ -0,0 +1,84 @@
...
+  opacity: 0.5;
```
danrosenthal: I'd like to see a bit more rationale on these opacity values. At the very least, they should be in a descriptive variable.

src/components/Toggle/Toggle.scss
```diff
@@ -0,0 +1,84 @@
...
+.Input {
+  position: relative;
+  display: inline-block;
+  padding: 0;
+  border: 0;
+  border-radius: border-radius(extra-large);
+  appearance: none;
+  cursor: pointer;
+  background-color: transparent;
+  user-select: none;
```
danrosenthal: What does this do?

src/components/Toggle/Toggle.scss
```diff
@@ -0,0 +1,84 @@
...
+  -webkit-tap-highlight-color: transparent;
```
danrosenthal: What's the purpose of this style?

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,90 @@
+import * as React from 'react';
+import {classNames} from '@shopify/react-utilities/styles';
+import {createUniqueIDFactory} from '@shopify/javascript-utilities/other';
+import Labelled from '../Labelled';
+
+import * as styles from './Toggle.scss';
+
+export interface Props {
+  /** Visually hidden text for screen readers */
+  accessibilityLabel?: string;
```
danrosenthal: Do we typically provide both an `accessibilityLabel` and a `label` prop? Wouldn't it be simpler to just use the `label` prop and hide it if `labelHidden`?

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,90 @@
...
+  /** Element to display after the input */
+  suffix?: React.ReactNode;
+  /** Callback when the toggle value is changed */
+  onChange(newChecked: boolean): void;
+}
+
+const getUniqueID = createUniqueIDFactory('Toggle');
+
+export default function Toggle({
+  id = getUniqueID(),
```
danrosenthal: 👍

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,90 @@
...
+  checked = false,
+  disabled = false,
```
danrosenthal: Shouldn't the `false` on each of these props be implicit?

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,90 @@
...
+  onChange,
+  accessibilityLabel,
+  label,
+  labelHidden = false,
```
danrosenthal: Ditto here per this being implicit.

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,90 @@
...
+  prefix,
+  suffix,
+}: Props) {
+  const className = classNames(
```
danrosenthal: Any reason not to make this `toggleClassName`?

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,90 @@
...
+  );
+
+  const inputClassName = classNames(
+    styles.Input,
+    checked && styles.ToggleChecked,
+    disabled && styles.ToggleDisabled,
+  );
+
+  function handleChange(event: React.MouseEvent<HTMLButtonElement>) {
+    if (disabled || !onChange) {
```
danrosenthal: `onChange` is a required prop, so the second condition here would never be reached.

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,90 @@
...
+      return;
+    }
+
+    const isButtonChecked =
+      event.currentTarget.getAttribute('aria-checked') === 'true';
+
+    onChange(!isButtonChecked);
```
danrosenthal: The negated condition here feels a bit strange, especially considering this is only used here. Why not make it `buttonNotChecked` or something like that and avoid it altogether?

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  const prefixMarkup = prefix && <div className={styles.Prefix}>{prefix}</div>;
```
danrosenthal: Though we do this with a short-circuit (`&&`) in a couple places in our codebase, I prefer the more explicit ternary returning markup or null pattern when it comes these.

```suggestion
  const prefixMarkup = prefix ? <div className={styles.Prefix}>{prefix}</div> : null;
```

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  const suffixMarkup = suffix && <div className={styles.Suffix}>{suffix}</div>;
```
danrosenthal: Ditto here.

```suggestion
  const suffixMarkup = suffix ? <div className={styles.Suffix}>{suffix}</div> : null;
```

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,90 @@
...
+      <div className={className}>
+        {prefixMarkup}
+        <button
+          id={id}
+          type="button"
+          className={inputClassName}
+          onClick={handleChange}
+          aria-checked={checked}
+          aria-disabled={disabled}
+          aria-label={accessibilityLabel}
```
danrosenthal: What would be different about the `accessibilityLabel` vs the `label` here?

src/components/Toggle/tests/Toggle.test.tsx
```diff
@@ -0,0 +1,156 @@
...
+        <Toggle {...mockProps} onChange={onChangeSpy} />,
+      );
+
+      const button = toggle.find('button');
+      button.simulate('click');
+
+      expect(onChangeSpy).toHaveBeenCalledTimes(1);
+    });
+  });
+});
```
danrosenthal: Code coverage here is an absolute beaut. Nice work!

src/components/index.ts
```diff
@@ -263,6 +263,8 @@ export {
   DEFAULT_TOAST_DURATION,
 } from './Toast';
 
+export {default as Toggle} from './Toggle';
```
danrosenthal: We should be exporting props here as well.

```suggestion
export {default as Toggle, Props as ToggleProps} from './Toggle';
```

src/styles/foundation/border-radius.scss
```diff
@@ -1,6 +1,7 @@
 $border-radius-data: (
   base: 3px,
   large: 6px,
+  extra-large: 30px,
```
danrosenthal: This seems like  fairly radical jump from `3px` to `6px`, all the way up to `30px`. I doubt this value needs to be shared throughout the system, and should probably be a variable local to the `Toggle` component.

mbaumbach:
I'm not sure if this is the same use cases as Shopify, but we use a `Toggle` that is modeled after the one available in Shopify Flow (which this one looks like too -- so we can't wait to toss our code and use this one :) ).

Our general use case for a `Toggle` vs a `Checkbox` is when we want a `Checkbox` that is a standalone setting and also invokes an immediate save, similar to the Shopify Flow editor causing the flow to be enabled. Perhaps there are other use cases, but that one felt like the biggest differentiator for us. We still use `Checkbox` when it's simply a boolean-like setting that will be saved along with other form elements/settings.

src/components/Toggle/README.md
solonaarmstrong: @danrosenthal Do you mean that the state should not be called `checked`? Perhaps something like `enabled`?

danrosenthal: It's less to do with the state, and more to do with the content of the example. What I'm looking for is an example that shows me why I would want to use a `Toggle` vs. a `Checkbox`. Does that make sense?

src/components/Toggle/Toggle.scss
solonaarmstrong: The only other one that makes sense to me here is `none`

src/components/Toggle/Toggle.tsx
solonaarmstrong: For form components, it looks like don't provide the `accessibilityLabel`. I fixed it.

src/components/Toggle/Toggle.tsx
solonaarmstrong: It seems to be our convention to name this `className`

solonaarmstrong:
Same! I also love the icons. And the inset shadow looks great.

solonaarmstrong:
@mbaumbach Thank you for such good wording between checkbox and toggle.

solonaarmstrong:
I'm loving these icons! They deviate slightly from your illustrations, @sarahill because of the size of our icons in relation to the size of the toggle.

screen shot 2018-11-13 at 9 20 29 am
screen shot 2018-11-13 at 9 20 34 am

src/components/Toggle/README.md
solonaarmstrong: Yes!

dpersing:
Love the icons! For the "Off" `X` icon, what do folks think about using `#637381` (`Lighter Ink` from the palette)? I think the current contrast would be good for a `disabled` state if one is provided.

A few other notes on AT support:

**Keyboard focus indicator:** The switch can receive keyboard focus, but it's not possible to see it because the browser default is being overridden (`outline: 0;`) and there is no replacement. @sarahill, what do you think about applying the thicker purple border we use for text inputs and buttons on focus?

**The shape of the switch isn't visible in Windows High Contrast.** We could add either a transparent border to the switch container by default, or add a border using the [`-ms-high-contrast` query](http://www.gwhitworth.com/blog/2017/04/how-to-use-ms-high-contrast) so that the shape is visible. Right now you can see the icons and the text, but the missing shape makes the control unclear until you interact with it.

Switch shape not visible in Windows High Contrast

**The state and/or role aren't announced clearly for VoiceOver on iOS, VoiceOver on macOS, and NVDA on Windows.** [Scott O'Hara does a breakdown of quirks here](https://github.com/scottaohara/aria-switch-button#screen-reader-quirks), which I'm able to confirm as still being the case, and to which I'd add JAWS/Edge not recognizing the role and reporting the control as a button. JAWS with Firefox is the only combo that appears to read the control correctly.

An almost identical ARIA pattern with better support is the the `checkbox` role, which I think would be worth considering, @solonaarmstrong. Here's the [`checkbox` spec](https://www.w3.org/TR/wai-aria-1.1/#checkbox) and [an official example](https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20160317/examples/checkbox/checkbox-1.html).

danrosenthal:
Great feedback @dpersing!

solonaarmstrong:
I increased the disabled opacity. Here are 4 states. As per Devon's suggestion, the `X` icon uses ink lighter.

screen shot 2018-11-13 at 3 02 06 pm

src/components/Toggle/Toggle.scss
solonaarmstrong: It adds 🚫 to the cursor.

screen shot 2018-11-13 at 3 29 43 pm

solonaarmstrong:
Latest changes:
- `X` is now ink lighter (darker)
- opacity for disabled is slightly higher (more visible)
- there is a transparent border on both the thumb and the entire component
- removed outline: none
- improved docs

---

Marki reminded me about the SettingToggle component. https://polaris.shopify.com/components/actions/setting-toggle

I've listed it as a related component. In addition to distinguishing toggle from checkbox, I feel like something should also be said to distinguish it from setting toggle. From what I can tell, toggle is replacing setting toggle in certain parts of the admin.

I'm unclear on whether or not these save differently. From what I understand, this toggle component  creates a dirty state that requires a further save, whereas the setting toggle saves on change?

I've included these changes in my latest commit.

src/components/Toggle/README.md
solonaarmstrong: I stumbled here because toggle is generally associated with a longer, inline description, rather than a label that's positioned above.

src/components/Toggle/Toggle.scss
```diff
@@ -0,0 +1,87 @@
+$toggle-track-length: rem(46px);
+$toggle-track-height: rem(24px);
+$toggle-thumb-size: rem(20px);
+$disabled-opacity: 0.6;
+$toggle-thumb-displacement: rem(2px);
+
+.Toggle {
+  display: flex;
+  align-items: center;
+  border: 1px solid transparent;
```
danrosenthal: This `1px` can be moved into a `toggle-border-size` variable and used throughout the component styles.

src/components/Toggle/Toggle.scss
```diff
@@ -0,0 +1,87 @@
+$toggle-track-length: rem(46px);
+$toggle-track-height: rem(24px);
+$toggle-thumb-size: rem(20px);
+$disabled-opacity: 0.6;
```
danrosenthal: This should be consistent with your naming convention, so `toggle-disabled-opacity`.

src/components/Toggle/Toggle.scss
```diff
@@ -0,0 +1,87 @@
...
+
+  &.disabled {
+    opacity: $disabled-opacity;
+  }
+}
+
+.ToggleTrack {
+  width: $toggle-track-length;
+  height: $toggle-track-height;
+  border-radius: rem(30px);
```
danrosenthal: This number should be placed in a variable.

src/components/Toggle/README.md
```diff
@@ -0,0 +1,261 @@
...
+
+### Default toggle
+
+Use toggle to let merchants enable or disable a setting.
+
+```jsx
+class ToggleExample extends React.Component {
+  state = {
+    visaChecked: true,
+    mcChecked: false,
```
danrosenthal: I'd like to see these, and the handlers use a more consistent naming convention. Something like `masterCardChecked` and `toggleMasterCardToggle`, since you use `visa` in the other ones.

sarahill:
@solonaarmstrong this is looking great! 

For the size of the `x` icon we do have a `cancel small` that exists in the icons kit. Can we use that one here instead? I think it'll look more consistent with the size of the `tick`.

screen shot 2018-11-14 at 9 41 31 am

sarahill:
Should the `Off` and `On` labels be clickable? In the case where there are labels, I tend to want to click them. I would expect them to still select the corresponding option.

Thoughts?

solonaarmstrong:
> For the size of the x icon we do have a cancel small that exists in the icons kit. Can we use that one here instead?

@sarahill It is using the `cancelSmall` icon.

sarahill:
Oh got it. It was the differences in size of the toggle I was using. Sorry, missed that point

danrosenthal:
> Should the Off and On labels be clickable? In the case where there are labels, I tend to want to click them. I would expect them to still select the corresponding option.

Definitely they should be clickable

sarahill:
The size of the `tick` seems off to me. In the polaris-icons sketch file, the minor tick is slightly bigger. Hm just noticed the icons in general also look a bit more rounded. Wonder if they got updated in the file and not in code? cc @jessebc 

screen shot 2018-11-15 at 10 00 27 am

solonaarmstrong:
Good eye, @sarahill ! It looks like the explicit size of the container is affecting the icon. I'm working on it.

dpersing:
This might be in the middle of getting worked on @solonaarmstrong, but I'm not able to activate the control with the space bar anymore. The checkbox role is being clearly read, though!

solonaarmstrong:
@dpersing I was going to ask you if I should add explicit key controls, since it's no longer an input?

src/components/Toggle/Toggle.scss
```diff
@@ -0,0 +1,99 @@
...
+
+.ToggleTrack {
+  width: $toggle-track-length;
+  height: $toggle-track-height;
+  border-radius: $toggle-border-radius;
+  box-shadow: 1px 1px 1px color('ink', 'lightest') inset;
+  background-color: color('sky');
+  transition: background-color duration(slow), border duration(slow),
+    box-shadow duration(slow);
+  -webkit-tap-highlight-color: transparent;
```
solonaarmstrong: @danrosenthal https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-tap-highlight-color

I tested it on my phone and it doesn't make a difference, so I will remove it.

solonaarmstrong: Incidentally, it looks pretty sexy on my phone. 😍

src/components/Toggle/Toggle.scss
solonaarmstrong: @danrosenthal https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-tap-highlight-color

I tested it on my phone and it doesn't make a difference, so I will remove it.

Incidentally, it looks pretty sexy on my phone 😍

dpersing:
> @dpersing I was going to ask you if I should add explicit key controls, since it's no longer an input?

Yup, since it's a `<div>` now you'll need to explicitly add support for the `space` key to toggle the `aria-checked` value and the toggle the appearance along with a mouse click.

tmlayton:
This is looking great! I just pulled the latest to do some 🎩’n and noticed a few things:

|Alignment is off|Odd focus area|Keyboard doesn’t toggle the switch|
|---|---|---|
|image|image|toggle|

UNRELEASED.md
```diff
@@ -4,6 +4,12 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ---
 
+### New components
+
+#### [Toggle](http://polaris.shopify.com/components/forms/toggle)
+
+- Use the toggle component to present a single choice to enable or disable a setting. ([#600](https://github.com/Shopify/polaris-react/pull/600))
```
```suggestion
- Use the toggle component to present a single choice to enable or disable a setting with an immediate save. ([#600](https://github.com/Shopify/polaris-react/pull/600))
```

src/components/Toggle/README.md
```diff
@@ -0,0 +1,481 @@
...
+Use to present a single choice to enable or disable a setting. Toggle is best used when multiple settings can be enabled or disabled and then saved all at once.
+
+---
+
+## Best practices
+
+Toggle should:
+
+- Be used when there are only two options available (enabled or disabled) and not a range of selections (zero, one, or multiple).
+- Be used to switch a single setting on and off.
```
tmlayton:
```suggestion
- Be used to switch a single setting on and off with an immediate save.
```

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,113 @@
...
+    disabled && styles.ToggleDisabled,
+  );
+
+  function handleChange(event: React.MouseEvent<HTMLElement>) {
+    if (disabled) {
+      return;
+    }
+
+    const buttonNotChecked =
+      event.currentTarget.getAttribute('aria-checked') === 'true';
```
tmlayton: If I understand this correctly it is to call the callback with what the new value will be. We shouldn’t have an aria attribute as the source of truth for this and would expect it to be the checked prop. Something like:

```js
function handleChange(event: React.MouseEvent<HTMLElement>) {
  if (disabled) {
    return;
  }
  
  const newValue = !checked;
  onChange(newValue);
}

File: src/components/Toggle/Toggle.tsx [hunk 35]
```diff
@@ -0,0 +1,113 @@
...
+  prefix?: React.ReactNode;
+  /** Element to display after the input */
+  suffix?: React.ReactNode;
+  /** Callback when the toggle value is changed */
+  onChange(newChecked: boolean): void;
+}
+
+const getUniqueID = createUniqueIDFactory('Toggle');
+
+export default function Toggle({
```
tmlayton: Why a function and not a class extending a React component?

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,113 @@
...
+          <div className={styles.ToggleThumb} />
+          {iconMarkup}
+        </div>
+        {suffixMarkup}
+      </div>
+    </Labelled>
+  );
+}
+
+export function IconWrapper({children, className}: any) {
```
tmlayton: Should we be exporting this?

src/components/Toggle/Toggle.tsx
tmlayton: And I think we will want to wrap this with the app provider:

```js
function Toggle (...) {...}

export default withAppProvider<Props>()(Toggle);
```

src/components/Toggle/tests/Toggle.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+      const onChangeSpy = jest.fn();
+
+      const toggle = mountWithAppProvider(
+        <Toggle {...mockProps} onChange={onChangeSpy} />,
+      );
+
+      const input = findByTestID(toggle, 'ToggleInput');
+      input.simulate('click');
+
+      expect(onChangeSpy).toHaveBeenCalledTimes(1);
```
tmlayton:
```suggestion
      expect(onChangeSpy).toHaveBeenCalledTimes(1);
      expect(onChangeSpy).toHaveBeenCalledWith(true);
```

src/components/Toggle/Toggle.tsx
solonaarmstrong: I am probably going to delete this.

src/components/Toggle/tests/Toggle.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+      expect(input.prop('role')).toBe('checkbox');
+    });
+  });
+
+  describe('checked', () => {
+    it('sets aria-checked if true', () => {
+      const toggle = mountWithAppProvider(<Toggle {...mockProps} checked />);
+      const input = findByTestID(toggle, 'ToggleInput');
+
+      expect(input.prop('aria-checked')).toBeTruthy();
```
tmlayton: We should avoid `toBeTruthy` and instead be literal in our expectations. In this case, the test can give a false positive because `expect('false').toBeTruthy()` will pass.

tmlayton: Although, after looking at the Enzyme API it looks like the `prop` method does some magic to turn these into booleans.

tmlayton:
```suggestion
      expect(input.prop('aria-checked')).toBe(true);
```

src/components/Toggle/tests/Toggle.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+    });
+
+    it('does not set aria-checked if false', () => {
+      const toggle = mountWithAppProvider(<Toggle {...mockProps} />);
+      const input = findByTestID(toggle, 'ToggleInput');
+
+      expect(input.prop('aria-checked')).toBeFalsy();
```
tmlayton:
```suggestion
      expect(input.prop('aria-checked')).toBeUndefined();
```

src/components/Toggle/tests/Toggle.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+    });
+
+    it('is set to false by default', () => {
+      const toggle = mountWithAppProvider(<Toggle {...mockProps} />);
+      const input = findByTestID(toggle, 'ToggleInput');
+
+      expect(input.prop('checked')).toBeFalsy();
```
tmlayton:
```suggestion
      expect(input.prop('checked')).toBe(false);
```

Which will actually fail because the value defaults to `undefined`.

src/components/Toggle/tests/Toggle.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+    });
+  });
+
+  describe('disabled', () => {
+    it('sets aria-disabled if true', () => {
+      const toggle = mountWithAppProvider(<Toggle {...mockProps} disabled />);
+      const input = findByTestID(toggle, 'ToggleInput');
+
+      expect(input.prop('aria-disabled')).toBeTruthy();
```
tmlayton: Update all of these truthy and falsy checks, I will spare the noise from making a bunch of comments 😄

src/components/Toggle/Toggle.tsx
solonaarmstrong: We export it here: https://github.com/Shopify/polaris-react/blob/master/src/components/Button/Button.tsx#L198

solonaarmstrong:
@jessebc 

> we should only use these when the changes take place immediately

Do you mean that this toggle should be used when saved as soon as it changes? My understanding was that this is how the settings toggle component works and that what distinguishes the two is when they save. What would the deciding factor be when choosing between the two?

> The tricky thing with labels or icons inside a toggle is that it's hard to discern if the icon is showing you a status or if it's labeling an action you can take. 

I can't think of any examples at the moment, but I've experienced this confusion with other UI and agree that it should be clear. For merchants who are able to see colour and contrast, there are other visual attributes that make the enabled/disabled state clear. The confusion would be most relevant for those with visual impairments.

How do we make this decision? Should I remove the icons?

sarahill:
@jessebc makes a good point about the potential for added confusion. Also, in the payment methods example above, the icons actually feel like a bit more cognitive load.

I like personally like the icons but if they cause confusion and don't always visually work I lean towards removing them.

I agree with the comments about the toggles triggering an immediate change. Our current solution is the SettingsToggle which, to @solonaarmstrong's point, raises the question of when would you use one over the other.

Also, in the case of filters, it may be more appropriate to use a checkbox depending on which direction we go. One of the ideas we have is the expanded section so the filters would sit above the list of customers, products, etc. In that case, the list would update (ideally) as the user is applying changes so the toggle might work.

solonaarmstrong:
> Sound like we need to clarify how we intend these to be used

If we decide to use them with filters, then it's clear they can't save right away. The payment methods toggle does not save on change. It creates a dirty state and multiple toggles can be changed before saving.

Also, I'm concerned that there are different designs for toggles across Shopify. I'm surprised teams don't align more to create consistency. If we do merge this component, would we reach out to the other teams to align? Should we align now?

image

dpersing:
> Should we align now?

I think it would be great to standardize the toggle, but I wonder if that might be a project in and of itself. In the short term, I wonder if we could use a checkbox for filters specifically, and work on a standard toggle in parallel? After?

sarahill:
Seems like a good idea to table this for now and use a checkbox. I'm also
unsure at this stage if the toggle is really appropriate in this context or
not. As we iterate on the prototype/design we can decide what makes sense.

> Also, I'm concerned that there are different designs for toggles across
Shopify. I'm surprised teams don't align more to create consistency. If we
do merge this component, would we reach out to the other teams to align?
Should we align now?

We should reach out to teams and align as much as possible. There may be
some differences between web and native mobile but aligning the style makes
sense.

UNRELEASED.md
solonaarmstrong: It would be the opposite.

solonaarmstrong:
Closing this in favour of further discussion https://github.com/Shopify/polaris-ux/issues/216

ziaulrehman40:
Come-on, we need this component, link in last comment is broken.

tmlayton:
> Come-on, we need this component, link in last comment is broken.

Hey @ziaulrehman40 thanks for the feedback, it’s helpful to know you would find it useful. And the link is not broken, it’s a private repo. Although the summary of that issue is there is a lot of contention, for very good reason, around this pattern so I don’t anticipate it being added anytime soon.

PR #601, opened 2018-11-12
[DatePicker] Fix date comparision when different references

nimzco:

When DatePicker allow ranges but `start` and `end` dates are different references `start === end` will return false even if they are the same date.

Compare date instead of references.

### How to 🎩

Best way to tophat is to try this PR with and without the changes:
https://github.com/Shopify/web/pull/8192

Before: 

before

After:

after

### 🎩 checklist


nimzco:
Sure thing 👍

nimzco:
I've added a test that fails without the change and passes with it! 🚀

PR #602, opened 2018-11-12
[Autocomplete] Add disabled prop

jfangrad:
Previously the `Autocomplete` component could not be disabled to prevent user interaction. Like a `TextField` the `Autocomplete` component should have the ability to be disabled and stop user interaction.

Adds in the `disabled` prop to the `Autocomplete` and `ComboBox` component.

This ensures that if `Autocomplete` is disabled, the `Popover` does not get displayed.

### How to 🎩

<details>

```jsx
import {Autocomplete, Icon, Page, TextField} from '@shopify/polaris';

interface State {
  options: {value: string; label: string}[];
  inputText: string;
  selected: string[];
}

  options = [
    {value: 'rustic', label: 'Rustic'},
    {value: 'antique', label: 'Antique'},
    {value: 'vinyl', label: 'Vinyl'},
    {value: 'vintage', label: 'Vintage'},
    {value: 'refurbished', label: 'Refurbished'},
  ];

  state = {
    selected: [],
    inputText: '',
    options: this.options,
  };

  updateText = (newValue: string) => {
  };

  filterAndUpdateOptions = (inputString: string) => {
    if (inputString === '') {
      this.setState({
        options: this.options,
      });
      return;
    }

    );
    this.setState({
      options: resultOptions,
    });
  };

  updateSelection = (updatedSelection: string[]) => {
    const selectedText = updatedSelection.map((selectedItem) => {
      const matchedOption = this.options.filter((option) => {
        return option.value.match(selectedItem);
      });
      return matchedOption[0] && matchedOption[0].label;
    });
    this.setState({selected: selectedText, inputText: ''});
  };

  render() {
    return (
      <Page title="playground">
        <div style={{height: '225px'}}>
          <Autocomplete
            options={this.state.options}
            textField={
              <Autocomplete.TextField
                onChange={this.updateText}
                label=""
                value={this.state.inputText}
                prefix={<Icon source="search" color="inkLighter" />}
                placeholder="Search"
                disabled
              />
            }
          />
        </div>
        <TextField label="" value="hey" disabled />
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist


src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -32,6 +32,8 @@ export interface Props {
   willLoadMoreResults?: boolean;
   /** Is rendered when there are no options */
   emptyState?: React.ReactNode;
+  /** Whether the attached text field should be disabled */
```
darrenhebner: This doesn't actually disable the text field, does it? It just prevents the popover form showing.

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
```diff
@@ -253,11 +256,13 @@ export default class ComboBox extends React.PureComponent<Props, State> {
...
+
     return (
       <div
         onClick={this.handleClick}
         role="combobox"
-        aria-expanded={this.state.popoverActive}
+        aria-expanded={active}
         aria-owns={this.state.comboBoxId}
         aria-controls={this.state.comboBoxId}
         aria-haspopup
```
darrenhebner: I'm wondering if someone from Polaris might know if there are some aria attributes that should be added to inform the user that the combobox is disabled?

src/components/Autocomplete/Autocomplete.tsx
jfangrad: Correct, since the text field is passed in as a prop you would have to disable it separately.

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
jfangrad: Yeah I was wondering as well... perhaps an `aria-disabled`?

dleroux: I would think that yes we would need an aria-disabled on the div since it's the one handling the interaction. cc @dpersing

dleroux:
I worry that having to disable both and Autocomplete and the Textfield will lead to inconsistencies. Another approach would be to disable the Textfield which would disable the Autocomplete. Or to disable the Autocomplete which would disable the Textfield. Both would mostly likely have to be done on context. Just not sure which would make more sense. What do you think? Disabling the Textfield would sort of be in line with our we handle popovers.

jfangrad:
@dleroux That makes sense and would give a better experience. Should we just go based off the `TextField` disabled prop? Since the `Autocomplete` gets passed the TextField as a prop we could just use `textfield.props.diabled` to enable / disable everything?

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
jfangrad: @dpersing Should we be doing both though? Disabling the text input and putting the `aria-disabled` on the div since the div is what the user clicks to open/close the combobox.

dpersing: @jfangrad It technically wouldn't hurt, but since the use case for assistive tech that could process the `aria-disabled` attribute on the `<div>` is pretty much limited to screen readers at the moment, the majority of users who would get that info would be users accessing the widget with the keyboard instead of the mouse. The `disabled` state on the input would prevent tab focus to start with, and would visually signal to sighted users that the control isn't available. I agree with @dleroux that confining disabling to just the control that triggers the interaction, and using the HTML native option, should lead to less inconsistency.

dleroux:
Yes let’s go with the disabled on Textfield. Thanks!

UNRELEASED.md
```diff
@@ -6,6 +6,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Added `disabled` prop to `Autocomplete` component.
```
dleroux: This is not the case anymore, I might even call it a bug fix. maybe?

- Added the ability to disable the `Autocomplete` by setting the  `Autocomplete.TextField` `disabled` prop true.

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
```diff
@@ -378,6 +378,22 @@ describe('<ComboBox/>', () => {
     });
   });
 
+  describe('disabled', () => {
+    it('does not set Popover to active if TextField is disabled', () => {
```
dleroux: I might also consider adding a similar to AutoComplete. Just for the sake of documenting how it's done. Also to check it's not focusable maybe.

dleroux:
One thing I noticed is that if a textfield is disabled it does not get a focus rectangle when clicked. However even when disabled the Autocomplete still does:

autocomplete

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
```diff
@@ -253,17 +253,20 @@ export default class ComboBox extends React.PureComponent<Props, State> {
...
         onFocus={this.handleFocus}
         onBlur={this.handleBlur}
-        tabIndex={0}
+        tabIndex={disabled ? -1 : 0}
```
dleroux: I just noticed that the Focus Rectangle is still visible with this tabIndex when navigating with keyboard.

I just tried quickly removing the tabIndex all-together and it looked like it worked because of the events bubbling from the text field. That was a quick test so we need to ensure it does work properly.

Also you might not need: 

```
.ComboBox > div:focus {
  outline: none;
}
```

If we do this. (This doesn't show up in the PR's changed files for some reason?) If we do need it however we should probably change to a classname so we don't rely on the div. Although like I said it might not be needed with the removal of the tab index.

dleroux:
@jfangrad I think you've uncovered bigger issues here. Maybe you want to hold off and @danrosenthal will keep us updated as he will be looking at our best options for this.

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
dpersing: @dleroux If the attributes are moved, we'd have to fall back to the [ARIA 1.0 combobox pattern](https://www.w3.org/TR/wai-aria-practices/#combobox) instead of the current 1.1 model, since how the attributes are nested and related as parents or siblings changes the semantic structure of the whole widget. Here's some more detail on the current accessibility state: https://github.com/Shopify/polaris-react-deprecated/issues/1969

danrosenthal:
Closing this PR in favor of re-architecting the `Autocomplete` component. See https://github.com/Shopify/polaris-react/issues/689 for more details.

PR #605, opened 2018-11-13
[Popover] Fix inconsistent border-radius

beefchimi:
A quick ~one~ two-liner.

I noticed that the `border-radius` for `<Popover />` was funky, and tracked it down to inconsistent `border-radius` values.

This PR resolves the visual issue.

### Before

Here you can see some of the `Popover` background poking through in the top left and right corner:

before

### After

Top corners now resolved:

after

### DIFF
--- a/src/components/Popover/Popover.scss
+++ b/src/components/Popover/Popover.scss
@@ -49,14 +49,14 @@ $content-max-width: rem(400px);
   position: relative;
   overflow: hidden;
   background-color: color('white');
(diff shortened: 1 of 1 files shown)


danrosenthal:
@HYPD or @sarahill what is the intended border radius on this component? Should it be `large` or just the `base`?

HYPD:
`base` would be the correct value to use here.

beefchimi:
@HYPD updated to use the `base` variant

PR #606, opened 2018-11-13
[TextStyle] Change <b> to <span>

beefchimi:
~`<b>` tags have gone the way of powdered wigs. Out with the old, in with the new. `TextStyle > strong` now uses a `<strong>` tag!~

We do not want to use `b` for "strong" variants... so using `span` tag instead.


AndrewMusgrave:
I [originally](https://github.com/Shopify/polaris-react-deprecated/pull/1889) had it as `strong` and Ryan had opinions about this 

> The style guide says to use strong “to mark text representing user input, or to emphasize the totals row in a price table.” We probably don’t always want it to be `strong`.  We could try to differentiate the two cases, but it’s probably not worth it. For now, let’s leave this as either `<span>` or `<b>`.

What are your thoughts?

beefchimi:
Looking into this - it looks like the `<b>` tag _(formerly known as the "bold" tag)_ has been repurposed as a "bring attention to" tag. Sounds like Ryan does not want to always add a special emphasis to the element, so using `b` or `strong` would not be the right choice.

I will update to use `span`.

PR #612, opened 2018-11-13
Update @shopify/polaris-tokens to the latest version

kaelig:

Polaris-tokens were updated to 2.0.0. It's a major version bump with no visual changes (only naming & infra).

Release notes: https://github.com/Shopify/polaris-tokens/blob/51ae550fe952678d5824c34fb3094462309a4a40/CHANGELOG.md#200---2018-10-23

Bumps up polaris-tokens from `^1.2.0` to `^2.0.0`.

No visual changes should occur.


chloerice:
Will the naming updates be a breaking change to consumers of `polaris-react` if they are consuming the tokens at a lower version?

kaelig:
> Will the naming updates be a breaking change to consumers of polaris-react if they are consuming the tokens at a lower version?

Ideally, those folks should be consuming polaris-tokens explicitly, not via an implicit dependency of polaris-react. Meaning that if they correctly have polaris-tokens in their dependencies, this breaking change should not impact them until they update the polaris-tokens dependency itself.

AndrewMusgrave:
Is this good to 🚢 ? @kaelig

PR #615, opened 2018-11-14
[DataTable] Update class targeting to data attribute targeting

keyfer:
### Problem

Currently, our DataTable looks for `class*=header` when calculating column visibility. This doesn't work in a production environment where class names are obfuscated, hence this working in a development environment, but not a production environment like an embedded app.

### Solution
This PR changes the way we target the `headerCells`, from the class-based targeting mentioned above to a data attribute system using `data-polaris-headercell`.

#### Alternative solution
The other option here would be to use `refs` instead of data attributes, but with `Cell` being a subcomponent inside of `DataTable`, we would need to do a ref callback, and that seemed like more trouble than what it was worth.

### How to 🔝 👒 

 - Load up the `DataTable` component with some dummy data
 - Make sure classes are obfuscated in the DOM
 - Confirm that the fixed column works, and there is no `offsetWidth` error in the console

### 🎩 checklist

### DIFF
--- a/src/components/DataTable/DataTable.tsx
+++ b/src/components/DataTable/DataTable.tsx
@@ -10,6 +10,7 @@ import {measureColumn, getPrevAndCurrentColumns} from './utilities';
 import {Cell, CellProps, Navigation} from './components';
 
 import * as styles from './DataTable.scss';
+import {headerCell} from 'components/shared';
 
 export type CombinedProps = Props & WithAppProviderProps;
 export type TableRow = Props['headings'] | Props['rows'] | Props['totals'];
@@ -328,7 +329,7 @@ export class DataTable extends React.PureComponent<CombinedProps, State> {
     } = this;
     if (collapsed && table && scrollContainer && dataTable) {
       const headerCells = table.querySelectorAll(
-        '[data-polaris-headercell]',
+        headerCell.selector,
       ) as NodeListOf<HTMLElement>;
       const collapsedHeaderCells = Array.from(headerCells).slice(1);
       const fixedColumnWidth = headerCells[0].offsetWidth;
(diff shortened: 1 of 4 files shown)


src/components/DataTable/DataTable.tsx
```diff
@@ -328,7 +328,7 @@ export class DataTable extends React.PureComponent<CombinedProps, State> {
     } = this;
     if (collapsed && table && scrollContainer && dataTable) {
       const headerCells = table.querySelectorAll(
-        '[class*=header]',
+        '[data-polaris-headercell]',
```
tmlayton:
```suggestion
        '[data-polaris-header-cell]',
```

src/components/DataTable/components/Cell/Cell.tsx
```diff
@@ -99,6 +99,7 @@ function Cell({
 
   const headingMarkup = header ? (
     <th
+      data-polaris-headercell
```
tmlayton:
```suggestion
      data-polaris-header-cell
```

solonaarmstrong:
LGTM. Echo Dan and Tim's comments and then it's 👍

src/components/DataTable/DataTable.tsx
AndrewMusgrave: Could you add the attribute to this file https://github.com/Shopify/polaris-react/blob/master/src/components/shared.ts 🙏

danrosenthal:
Changes look good but the failure in CI is legit. You'll need to change the import order to pass. After that, this will be good to go.

src/components/DataTable/components/Cell/Cell.tsx
```diff
@@ -99,6 +99,7 @@ function Cell({
 
   const headingMarkup = header ? (
     <th
+      data-polaris-header-cell
```
danrosenthal: You'll want to add an import in this component bringing in `headerCell`:

```
import {headerCell} from 'components/shared'
```

And then update this line to:

```suggestion
      {...headerCell.props}
```

PR #638, opened 2018-11-16
[DataTable] Use relative path

AndrewMusgrave:
Causes build to fail

Uses a relative import

### How to 🎩
TBD: Not totally sure. Possibly run web in production?

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -8,7 +8,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
-- Fix selector import in `DataTable` and `Cell` ([#638](https://github.com/Shopify/polaris-react/pull/638))
+- Fixed selector import in `DataTable` and `Cell` ([#638](https://github.com/Shopify/polaris-react/pull/638))
 
 ### Documentation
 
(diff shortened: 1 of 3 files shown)


danrosenthal:
@GoodForOneFare, do you have any advice on tophatting this to make sure it won't cause the same problem that led to https://github.com/Shopify/web/pull/8518?

UNRELEASED.md
```diff
@@ -8,6 +8,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- Fix selector import in `DataTable` and `Cell` ([#638](https://github.com/Shopify/polaris-react/pull/638))
```
solonaarmstrong: Fixed

danrosenthal: This feels like a great opportunity to use the code suggestion feature.

```suggestion
- Fixed selector import in `DataTable` and `Cell` ([#638](https://github.com/Shopify/polaris-react/pull/638))
```

solonaarmstrong: Oh right, I have to rewire my brain to remember to use it.

danrosenthal:
@GoodForOneFare, also curious if you have any insight as to why this didn't come up earlier in the web CI process? Why didn't this fail until it was in Shipit?

GoodForOneFare:
shrink-ray is temporarily disabled.  It usually catches these problems.

PR #639, opened 2018-11-17
Fix `yarn build` on Windows

dustinsoftware:

Building styles did not work at all on Windows...
image

image

...but it does now! I have tested these changes on both Mac and Windows.

Fixes a few assumptions around string quotes and paths.

cc @BPScott


dustinsoftware:
CLA signed

PR #643, opened 2018-11-19
[Pagination] Fixed event from being called when has previous or next is false

AndrewMusgrave:
Key events can't be disabled like buttons and will currently fire when `hasNext` or `hasPrevious` are true

* Add default of true for `hasNext` and `hasPrevious`
* Don't add key events if `hasNext` or `hasPrevious` are true
* Add tests
* Add changelog

### How to 🎩
Use the playground code to toggle `hasNext` and `hasPrevious`

<details>

```jsx
import {Page, Pagination, Button} from '@shopify/polaris';

interface State {
  next: boolean;
  prev: boolean;
  [key: string]: any;
}

  state = {
    next: true,
    prev: true,
  };

  toggleState = (key: string) => () => this.setState({[key]: !this.state[key]});

  render() {
    const {next, prev} = this.state;

    return (
      <Page title="Playground">
        <Button onClick={this.toggleState('prev')}>Toggle Previous</Button>

        <Button onClick={this.toggleState('next')}>Toggle Next</Button>

        <Pagination
          hasPrevious={prev}
          previousKeys={[74]}
          previousTooltip="j"
          onPrevious={() => {
            console.log('Previous');
          }}
          hasNext={next}
          nextKeys={[75]}
          nextTooltip="k"
          onNext={() => {
            console.log('Next');
          }}
        />
      </Page>
    );
  }
}
```

</details>

#### Notes
* base to be changed when we create a feature branch to merge to during the code freeze

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -24,15 +24,9 @@ export interface PaginationDescriptor {
   nextURL?: string;
   /** The URL of the previous page */
   previousURL?: string;
-  /**
-   * Whether there is a next page to show
-   * @default true
-   */
+  /** Whether there is a next page to show */
   hasNext?: boolean;
-  /**
-   * Whether there is a previous page to show
-   * @default true
-   */
+  /** Whether there is a previous page to show */
   hasPrevious?: boolean;
   /** Accessible label for the pagination */
   accessibilityLabel?: string;
@@ -50,8 +44,8 @@ export interface Props extends PaginationDescriptor {
 export type CombinedProps = Props & WithAppProviderProps;
 
 function Pagination({
-  hasNext = true,
-  hasPrevious = true,
+  hasNext,
+  hasPrevious,
   nextURL,
   previousURL,
   onNext,
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -24,9 +24,15 @@ export interface PaginationDescriptor {
   nextURL?: string;
   /** The URL of the previous page */
   previousURL?: string;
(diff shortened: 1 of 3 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -24,9 +24,15 @@ export interface PaginationDescriptor {
...
-  /** Whether there is a next page to show */
+  /**
+   * Whether there is a next page to show
+   * @default true
+   */
   hasNext?: boolean;
-  /** Whether there is a previous page to show */
+  /**
+   * Whether there is a previous page to show
+   * @default true
```
solonaarmstrong: Why does this default to true?

src/components/Pagination/tests/Pagination.test.tsx
```diff
@@ -139,6 +139,24 @@ describe('<Pagination />', () => {
 
       expect(spy).toHaveBeenCalledTimes(1);
     });
+
+    it('does not navigate the browser when hasNext or hasPrevious is false', () => {
+      const spy = jest.fn();
```
solonaarmstrong: We've started naming the spy according to what it's spying on, so in this case it would be called `anchorClickSpy`.

solonaarmstrong:
This looks 👍 Just a small comment on the variable name.

solonaarmstrong:
Also, I think the Percy error for Page might be legit.

src/components/Pagination/Pagination.tsx
AndrewMusgrave: I was to quick on adding this, percy should be fixed.

AndrewMusgrave:
I'll resolve the conflict and 🚢 tomorrow

PR #647, opened 2018-11-19
Prevent polaris-react 3.0.x hotfixes from getting deployed under the 'latest' tag

kaelig:

- Versions published from the `web` branch should not overwrite the version available on CDN via the `/latest/` release tag.
- `npm publish`, by default, publishes packages under the `latest` tag.

- Prevents `web` branch deploys from getting uploaded to S3 in the `/latest/` directory.
- Prevents `npm publish` from overriding the `latest` dist-tag when deploying from the `web` branch.

(Not sure how to test this without creating a whole new package, stack, s3 bucket, etc… probably not worth the hassle)

### DIFF
--- a/shipit.yml
+++ b/shipit.yml
@@ -16,3 +16,5 @@ merge:
 
 deploy:
   max_commits: false
+  override:
+    - npm publish
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
     "url": "https://github.com/Shopify/polaris-react/issues"
   },
   "publishConfig": {
-    "access": "public"
+    "access": "public",
+    "tag": "web"
   },
   "sideEffects": [
     "**/*.css",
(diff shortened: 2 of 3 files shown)


shipit.yml
```diff
@@ -16,3 +16,5 @@ merge:
 
 deploy:
   max_commits: false
+  override:
+    - npm publish
```
kaelig: This ensures `publishConfig.tag` is respected - because Shipit wouldn't have done so ([See issue in shipit-engine](https://github.com/Shopify/shipit-engine/issues/843)).

BPScott:
We need to take care not to merge this into master after the change freeze.

kaelig:
> We need to take care not to merge this into master after the change freeze.

Yes, because if we do… 💥 😨

kaelig:
> I'd be tempted to change latest: isStableVersion, to latest: false && isStableVersion, and that way you don't need to do the other line deletions to keep the diff as small as possible but that's just me golfing.

Thanks for the great suggestion - much better.

PR #648, opened 2018-11-20
Provide better error messaging when README examples have syntax errors

BPScott:

If you fatfingered an example in a README then `yarn tophat` can give a cryptic error message. Make it less cryptic by adding more information.

Throw an error message that contains the file and example name that the problem occurs in.

What was:

```
Unexpected token, expected "{" (6:24)
```

Is now:

```
🚨 [Text field] Example "Default text field" contains a syntax error in src/components/TextField/README.md: Unexpected token, expected "{" (6:24)
```

### How to 🎩

* Go make a syntax error in an example. For instance change `handleChange = (value) =>` to `handleChange (value) =>` in TextField/README.md's "Default text field" example.
* Run `yarn tophat`
* See that the console reports the file and example name within which the error occurs. It'll look something like:

```
   🚨 [Text field] Example "Default text field" contains a syntax error in src/components/TextField/README.md: Unexpected token, expected "{" (6:24)
```

### DIFF
--- a/documentation/Component READMEs.md
+++ b/documentation/Component READMEs.md
@@ -132,16 +132,27 @@ Example description.
 
 ## Troubleshooting
 
-When running `yarn tophat` or CI tests, when you see:
+When running `yarn tophat` or CI tests you may run into errors.
+Here are some common fixes.
+
+### No examples found
 
 ```console
 🚨 [Top bar] No examples found in src/components/TopBar/README.md
 ```
 
-A `---` may be missing before the `## Examples` heading:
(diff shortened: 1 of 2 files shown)


solonaarmstrong:
Thank you! I should probably get back into the habit of using `yarn tophat`. At some point, I switched to `dev server`.

BPScott:
I'm working on some plans to unify the two so we can have the empty playground for quick fiddling and the rendered README examples (eventually with live editing) all in one server.

PR #650, opened 2018-11-20
[FormLayout] min-width added to Item only when in Group. 

dleroux:

Resolves https://github.com/Shopify/web/issues/4902

There's a min-width set on FormLayout Items t

hat is used for wrapping of items when inside a group. In the side column however that min-width is too wide.

Adding the min-width to items when in groups only.

Before | After
------------ | -------------
Content from cell 1 | Content from cell 2
formgroupmaxwidth-b4 | formgroupmaxwidth
screen shot 2018-11-20 at 10 12 07 am | screen shot 2018-11-20 at 9 46 00 am

## 

### How to 🎩

1. Below is the Polaris-React Playground
2. `yarn build-consumer web` to 🎩 in web

<details>

```jsx
import {Page, FormLayout, TextField} from '@shopify/polaris';

interface State {}

  render() {
    return (
      <Page title="Playground">
        <FormLayout>
          <FormLayout.Group>
            <TextField label="Length" onChange={() => {}} />
            <TextField label="Width" onChange={() => {}} />
            <TextField label="Height" onChange={() => {}} />
            <TextField label="Unit" onChange={() => {}} />
          </FormLayout.Group>
          <FormLayout.Group condensed>
            <TextField label="Length" onChange={() => {}} />
            <TextField label="Width" onChange={() => {}} />
            <TextField label="Height" onChange={() => {}} />
            <TextField label="Unit" onChange={() => {}} />
          </FormLayout.Group>
        </FormLayout>
      </Page>
    );
  }
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/FormLayout/FormLayout.scss
+++ b/src/components/FormLayout/FormLayout.scss
@@ -5,11 +5,6 @@ $item-min-size: rem(220px);
   margin-left: -1 * spacing(loose);
 }
 
-.condensed .Item {
-  flex-basis: (0.5 * $item-min-size);
-  min-width: (0.5 * $item-min-size);
-}
-
 .Title {
   margin-bottom: -1 * spacing(tight);
   padding: spacing() spacing(loose) 0;
@@ -29,6 +24,14 @@ $item-min-size: rem(220px);
   flex: 1 1 $item-min-size;
   margin-top: spacing();
   margin-left: spacing(loose);
-  min-width: $item-min-size;
   max-width: calc(100% - #{spacing(loose)});
+
+  .grouped & {
+    min-width: $item-min-size;
+  }
+
+  .condensed & {
+    flex-basis: (0.5 * $item-min-size);
+    min-width: (0.5 * $item-min-size);
+  }
 }
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -8,6 +8,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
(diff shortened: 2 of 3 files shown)


src/components/FormLayout/FormLayout.scss
```diff
@@ -29,6 +24,14 @@ $item-min-size: rem(220px);
   flex: 1 1 $item-min-size;
```
dleroux: Opted to keep the flex-basis as part of the Item to minimize the change.

src/components/FormLayout/FormLayout.scss
```diff
@@ -29,6 +24,14 @@ $item-min-size: rem(220px);
...
   margin-left: spacing(loose);
-  min-width: $item-min-size;
   max-width: calc(100% - #{spacing(loose)});
+
+  .grouped & {
+    min-width: $item-min-size;
+  }
+
+  .condensed & {
+    flex-basis: (0.5 * $item-min-size);
```
AndrewMusgrave: Why `.5`? Should we save this to a variable?

dleroux: This was like this so I opted not to touch it. It's only used once so I don't see an issue with it. Do you?

danrosenthal: I think this is fine as is. I think this pretty clearly communicates that it is half the item min size.

danrosenthal: We do the same thing with `* -1` to make a number negative. We don't add comments for that.

PR #653, opened 2018-11-20
[WIP] [App Bridge] [Frame] improve documentation for Frame component and those components which delegate to App Bridge

### LINKED ISSUE #599: [Frame] Update example to follow stand-alone app guidelines
Per our [stand-alone app guidelines ](https://polaris.shopify.com/guides/designing-apps#section-make-your-stand-alone-app-visually-distinct), update the `<Frame />` example to include the app name in the navigation as well as a "Back to Shopify" link.

danrosenthal:
Closes #599

#### Modal
- key issue with Modal is that for embedded context, the `src` prop must be used, and React children are invalid
- this component also a "this component delegates to app bridge" boilerplate message

#### Page
- key issue with Page is that, when in an embedded context, the UI looks completely different
	- it needs some text and a screenshot to explain the difference
- this component needs a "this component delegates to app bridge" boilerplate message

#### ContextualSaveBar
- this one is a bit weird, as it does not work in an embedded context at all. Not sure how, or if, this fits within the app bridge delegation boilerplate

#### Toast
- this component needs a "this component delegates to app bridge" boilerplate message
- should also state that, at this time, only one toast message is supported at a time

#### Loading
- this component needs a "this component delegates to app bridge" boilerplate message

#### Frame
- the issue here is that our examples do not follow our guidelines, so we need to make fixes to align with our guidelines
- the secondary issue is that we do not provide a `Shopify admin` example that could be easily opened in a code sandbox


danrosenthal:
## Screenshots of work in progress improvements to Frame examples

I've added a second one to help with rapidly prototyping the admin.

|Stand-alone|Shopify admin|
|-|-|
|screen shot 2018-11-21 at 1 58 14 pm|screen shot 2018-11-21 at 1 58 52 pm|

davekilljoy:
The new frame example looks solid. Could we update the `Orders` nav item to say `Jaded Pixel Orders` to make sure we're there is no confusion between Shopify `Orders` section and `[Stand-alone app] Orders` sections.

I don't know if we should follow the same though pattern with the `Billing details` for example. Should we say `Jaded Pixel billing details` to be extremely clear? What do y'all think?

adamwillking:
Yeah I think it's good to default to using the app name in front of resources already exist in Shopify. So the perception is more that they are children of the main orders/products/whatever. 

In this case Jaded Pixel billing details is a little bit long, so it may be a good example to point out that if the title is too long to include that context, make sure it's clear in the section description. For instance. In this case the section description isn't clear that it's specific to the Jaded Pixel app. I think that would be a nice compromise when the title gets crazy long. WDYT @davekilljoy?

markiiikram:
Hey @danrosenthal, Thanks for these changes. I think these updates are a great step in the right direction in clarifying the differences when building for embedded and stand-alone apps. As we spoke about, please add the embedded page screenshot https://github.com/Shopify/polaris-react/pull/653/files#diff-63ec484b9233a6121ad12bc626225d26R45 once it's been added to the  polaris-styleguide

src/components/Page/components/Header/Header.tsx
```diff
@@ -35,23 +35,21 @@ export interface PrimaryActionProps
...
   /** Page title, in large type */
   title: string;
-  /** Important and non-interactive status information shown immediately after the title. */
+  /** Important and non-interactive status information shown immediately after the title. (stand-alone app use only) */
   titleMetadata?: React.ReactNode;
-  /** Visually hide the title */
+  /** Visually hide the title (stand-alone app use only) */
   titleHidden?: boolean;
-  /** App icon, for pages that are part of Shopify apps */
-  icon?: string;
```
danrosenthal: This was 🔥 in `3.0.0` but we never removed the prop from this interface.

tmlayton: Good catch

src/components/Toast/README.md
```diff
@@ -33,7 +33,15 @@ The toast component is a non-disruptive message that appears at the bottom of th
 
 ## Required components
 
-The toast component must be wrapped in the [frame](/components/structure/frame) component.
+The toast component must be wrapped in the [frame](/components/structure/frame) component unless used in an embedded application.
```
tmlayton:
```suggestion
The toast component must be wrapped in the [frame](/components/structure/frame) component or used in an embedded application.
```

src/components/Page/components/Header/Header.tsx
tmlayton: As discussed, we should keep this prop and instead add a console warning that the prop no longer does anything and to use the Partners dashboard to upload your icon.

src/components/Frame/README.md
```diff
@@ -373,6 +368,303 @@ class FrameExample extends React.Component {
...
+        </Layout>
+      </Page>
+    );
+
+    const loadingMarkup = isLoading ? <Loading /> : null;
+
+    const pageMarkup = isLoading ? loadingPageMarkup : actualPageMarkup;
+
+    return (
+      <AppProvider
```
tmlayton: This needs a fixed height container

src/components/Loading/README.md
```diff
@@ -32,7 +32,13 @@ Use to indicate that the page is loading.
...
 ## Required components
 
-The loading component must be wrapped in the [frame](/components/structure/frame) component.
+The loading component must be wrapped in the [frame](/components/structure/frame) component unless used in an embedded application.
+
+---
+
+## Use in an embedded application
+
+Passing an API key to the [app provider component](https://polaris.shopify.com/components/structure/app-provider#section-initializing-the-shopify-app-bridge) causes the loading component to delegate to the [Shopify App Bridge](https://help.shopify.com/en/api/embedded-apps/app-bridge) instead of rendering as it would in a stand-alone application.
```
chloerice: Can you elaborate on what this actually means? I think a code example of how it should be used in an embedded application would be helpful.

src/components/ContextualSaveBar/README.md
```diff
@@ -17,6 +17,18 @@ The contextual save bar tells merchants their options once they have made change
...
 
+## Required components
+
+The contextual save bar component must be wrapped in the [frame](/components/structure/frame) component.
+
+---
+
+## Use in an embedded application
+
+Use of the contextual save bar component in an embedded application is [not currently supported](link/to/github/issue).
```
tmlayton:
```suggestion
Use of the contextual save bar component in an embedded application is not currently supported. If this is a feature you would like to see supported by Shopify App Bridge, let us know in the [forums](https://ecommerce.shopify.com/c/shopify-apis-and-technology).
```

src/components/Loading/README.md
chloerice: For instance, the App Bridge docs instruct us to use the package's `Loading` action, and following the example in the `AppProvider` docs, it follows that you need to do the following: 

```jsx
import {Loading} from '@shopify/app-bridge/actions';
...

startLoading = () => {
  const loading = Loading.create(this.context.polaris.appBridge);
  loading.dispatch(Loading.Action.START);
}
```

But this will cause their Polaris app to not compile.

src/components/Modal/Modal.tsx
```diff
@@ -33,27 +33,27 @@ const IFRAME_LOADING_HEIGHT = 200;
 export type Size = 'Small' | 'Medium' | 'Large' | 'Full';
 
 export interface Props extends FooterProps {
-  /** Whether the modal is open or not */
+  /** Whether the modal is open or not (stand-alone app use only) */
```
chloerice: 👌

src/components/Modal/README.md
```diff
@@ -38,6 +38,14 @@ Modals are overlays that prevent merchants from interacting with the rest of the
 
 ---
 
+## Use in an embedded application
+
+Passing an API key to the [app provider component](https://polaris.shopify.com/components/structure/app-provider#section-initializing-the-shopify-app-bridge) causes the modal component to delegate to the [Shopify App Bridge](https://help.shopify.com/en/api/embedded-apps/app-bridge) instead of rendering as it would in a stand-alone application.
+
+Note in the props table that a number of properties are only available in stand-alone applications, and won't work in an embedded context.
+
```
chloerice: An embedded app use code example would be really helpful here.

danrosenthal:
Thanks for the great feedback @tmlayton and @chloerice.

> I saw in a comment you made earlier that you mentioned a CodeSandbox, is that something you're planning to include? Or is that something we want to figure out in a separate exploration?

Not sure when I mentioned a code sandbox, unless it was in reference to opening the new frame examples in a code sandbox. No plans to make anything else that isn't represented in this PR.

Code examples on the components you mentioned sounds like a reasonable thing to add. I'll throw some simple, code-only examples in below the sections that need it.

danrosenthal:
@chloerice and @tmlayton, I've addressed both of your feedback. Mind having another look?

tmlayton:
Added a test for the icon warning to sort out the coverage check

src/icons/index.ts
```diff
@@ -51,3 +51,8 @@ export {default as risk} from './risk.svg';
 export {default as save} from './save.svg';
 export {default as search} from './search.svg';
 export {default as view} from './view.svg';
+export {default as customers} from './customers.svg';
```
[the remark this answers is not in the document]
danrosenthal: Gonna burn these and the admin example in favor of an internal code sandbox example to come once the polaris icons project is ready.

PR #654, opened 2018-11-20
[ContextualSaveBar]: Add prop to extend contents flush with left edge.

oros:

Related to: https://github.com/Shopify/business-platform/issues/684

We are currently presenting a separate TopBar in specific scenarios. When they make changes within `web`, the context bar shows the old logo. We are looking to simply hide the logo in this dirty state.

image

This PR adds a `hideLogo` prop to the ContextualSaveBar component which essentially eliminates the markup for the Logo section in the ContextualSaveBar. Existing CSS for the `message` allows that section to fill in the remaining space out of the box.

## 

### How to 🎩

Essentially, loading the ContextualSaveBar and toggling the `alignContentFlush` prop between `true | false` should be enough to see the results. Stretching the viewport to smaller dimensions and observing the behaviour should remain consistent.

<details>

```jsx
import {AppProvider, ContextualSaveBar, Frame} from '@shopify/polaris';

interface State {}

  render() {
    return (
        <AppProvider
          theme={{
            logo: {
              width: 124,
              contextualSaveBarSource:
            },
          }}
        >
          <Frame>
            <ContextualSaveBar
              alignContentFlush
              message="Unsaved changes"
              saveAction={{
                onAction: () => console.log('add form submit logic'),
                loading: false,
                disabled: false,
              }}
              discardAction={{
                onAction: () => console.log('add clear form logic'),
              }}
            />
          </Frame>
        </AppProvider>
      </div>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -9,6 +9,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - `TextField` no longer uses `componentWillReceiveProps`([#628](https://github.com/Shopify/polaris-react/pull/628))
 - EventListener`no longer uses`componentWillUpdate` ([#628](https://github.com/Shopify/polaris-react/pull/628))
 - Allowed `Icon` to accept a React Node as a source ([#635](https://github.com/Shopify/polaris-react/pull/635)) (thanks to [@mbriggs](https://github.com/mbriggs) for the [original issue](https://github.com/Shopify/polaris-react/issues/449))
+- Added `alignContentFlush` prop to ContextualSaveBar ([#654](https://github.com/Shopify/polaris-react/pull/654))
 
 ### Bug fixes
 
--- a/src/components/Frame/components/ContextualSaveBar/ContextualSaveBar.tsx
+++ b/src/components/Frame/components/ContextualSaveBar/ContextualSaveBar.tsx
@@ -29,6 +29,7 @@ class ContextualSaveBar extends React.PureComponent<CombinedProps, State> {
     const {discardConfirmationModalVisible} = this.state;
 
     const {
+      alignContentFlush,
       message,
       discardAction,
       saveAction,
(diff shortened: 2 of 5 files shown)


oros:
Sounds good to me! I'll go ahead and make the changes.

oros:
@danrosenthal Went with `alignContentFlush`, happy for any additional feedback you might have.

danrosenthal:
This is looking great! Can you add an example in the `README.md` file with a sentence of rationale for why you would use this prop in your application?

oros:
Percy / visual regression testing is a bit trickier to validate as the "All Examples" renders three fixed ContextualSaveBar components on top of eachother; but can confirm that the individual example behaves as expected.

image

oros:
@danrosenthal Added a README example and I *think* the Percy regression testing is actually behaving as we would expect. It's a bit muddy due to three bars with fixed positioning all overlapping.

src/components/Frame/components/ContextualSaveBar/ContextualSaveBar.tsx
```diff
@@ -100,12 +101,16 @@ class ContextualSaveBar extends React.PureComponent<CombinedProps, State> {
       />
     );
 
+    const logoMarkup = alignContentFlush ? null : (
```
AndrewMusgrave: `alignContentFlush` doesn't default to true so null will be shown by default making this a breaking change unless i'm missing something?

oros: I'm not sure now if I'm misreading it. But I think...

- If the flag isn't specified, it will be `falsey`, showing the logo markup as it has before.

- If the flag is specified, it will be `truthy`, resulting in no logo section thus the content will grow.

- If the flag is specified and specifically set to `false`, the logo markup will show.

UNRELEASED.md
```diff
@@ -9,6 +9,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - `TextField` no longer uses `componentWillReceiveProps`([#628](https://github.com/Shopify/polaris-react/pull/628))
 - EventListener`no longer uses`componentWillUpdate` ([#628](https://github.com/Shopify/polaris-react/pull/628))
 - Allowed `Icon` to accept a React Node as a source ([#635](https://github.com/Shopify/polaris-react/pull/635)) (thanks to [@mbriggs](https://github.com/mbriggs) for the [original issue](https://github.com/Shopify/polaris-react/issues/449))
+- Adds `alignContentFlush` prop to ContextualSaveBar ([#654](https://github.com/Shopify/polaris-react/pull/654))
```
```suggestion
- Added `alignContentFlush` prop to ContextualSaveBar ([#654](https://github.com/Shopify/polaris-react/pull/654))
```

PR #672, opened 2018-11-24
Tophat: Updates to mimic webgen created projects and fix TS validations

BPScott:

Making the playground folder structure look more similar to a webgen project layout.

This helps with familiarity if people are used to other webgen projects, and makes it easier to build the the playground using sewing-kit in the future.

- Switch to ts-loader as that's what sk uses under the hood in dev mode
- Ensure typescript compiles (most of the tsconfig changes are alphabetising the file to make - comparing to webgen's tsconfig easier)
- Move client rendering into its own folder

### How to 🎩

* Run `yarn dev`
* Ensure playground changes still hot-reloads in a timely fashion
* Ensure component modifications (e.g. changing src/components/DisplayText/DisplayText.tsx) still hot-reloads in a timely fashion
* Ensure README modifications (e.g. changing src/components/DisplayText/README.md) still hot-reloads in a timely fashion


playground/webpack.config.js
```diff
@@ -134,25 +134,23 @@ module.exports = (env = {production: false}) => ({
         test: /\.tsx?$/,
         use: [
           {
-            loader: 'awesome-typescript-loader',
+            loader: 'babel-loader',
```
amrocha: Is this because we were applying babel through the ts loader before, so now we need to do this manually?

amrocha: As a follow-up, have you looked into whether we can just use babel to compile TS?

playground/webpack.config.js
```diff
@@ -134,25 +134,23 @@ module.exports = (env = {production: false}) => ({
...
+                ['shopify/web', {modules: false}],
+                ['shopify/react', {hot: true}],
+              ],
+              cacheDirectory: path.resolve(__dirname, 'build/cache/typescript'),
+            },
+          },
+          {
+            loader: 'ts-loader',
             options: {
               silent: true,
```
amrocha: why silent?

package.json
```diff
@@ -98,8 +98,8 @@
     "@types/react-helmet": "^5.0.6",
     "@types/react-hot-loader": "^3.0.6",
     "@types/react-router-dom": "^4.2.7",
+    "@types/rollup": "0.51.4",
     "archiver": "^2.1.0",
-    "awesome-typescript-loader": "5.2.1",
```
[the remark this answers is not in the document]
BPScott: Now there's a story. sewing-kit uses awesome-typescript-loader v3.1.3 in production, and ts-loader in development due to heap allocation issues. I using using the ts-loader version sk uses is preferable to using a version of atl different to what sk uses.

playground/webpack.config.js
BPScott: Bingo. I havent looked into using Babel for compiling TS, as I figured sticking close to what SK does under the hood shall result in the least confusion.

playground/webpack.config.js
BPScott: We used silent before, and it's what SK does. It doesn't look like it has any effect on output as most errors are output via webpack rather than `console.log` per [their docs](https://github.com/TypeStrong/ts-loader#silent-boolean-defaultfalse)

package.json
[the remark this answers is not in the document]
amrocha: Huh, I thought it was all the same loader. Is it for performance reasons?

playground/webpack.config.js
amrocha: I saw TS recently announced that they fully support babel, would be cool if we could use that to simplify build configs

package.json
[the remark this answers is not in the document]
BPScott: ts-loader seems to blow up and fail to compile in development for large projects from what I can tell. That was however an old version of ts-loader so it might have been fixed, foundations might reevaluate their usage and see if they can consolidate on a single thing at some point when they have time.

PR #683, opened 2018-11-27
Fix straight quotes into apostrophes

kaelig:
What do you do when you have a few spare minutes while waiting for folks to join a meeting?

```
ag "[a-zA-Z]'[a-zA-Z]"

-> fix occurrences of ' into ’
```

[What is `ag`?](https://geoff.greer.fm/ag/)

### Why?

https://polaris.shopify.com/content/grammar-and-mechanics#section-punctuation

screenshot 2018-11-27 10 46 07

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -52,7 +52,7 @@
 
     "DiscardConfirmationModal": {
       "title": "Discard all unsaved changes",
-      "message": "If you discard changes, you'll delete any edits you made since you last saved.",
+      "message": "If you discard changes, you’ll delete any edits you made since you last saved.",
       "primaryAction": "Discard changes",
       "secondaryAction": "Continue editing"
     },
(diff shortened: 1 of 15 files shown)


src/locales/en.json
```diff
@@ -52,7 +52,7 @@
 
     "DiscardConfirmationModal": {
       "title": "Discard all unsaved changes",
-      "message": "If you discard changes, you'll delete any edits you made since you last saved.",
+      "message": "If you discard changes, you’ll delete any edits you made since you last saved.",
```
kaelig: Will this break anything, or is it something we can change?

kaelig:
Does this require a changelog entry?

perrupa:
> Does this require a changelog entry?

I can't imagine any case in which this would break something or anyone caring about this. I'd error on the side of no changelog entry required. 👍

kaelig:
Thank you @perrupa and @danrosenthal!

PR #685, opened 2018-11-27
Tophat: Compile all examples in a single pass

BPScott:

Tophat's generation of examples from markdown is kinda clunky. It runs babel over every example, which results in excess work during the build, and very fat example strings - as every example contains all of babel's helper functions in addition to the example code.

It also means we have to call babel ourselves, instead of delegating to our webpack configuration.

Run all examples through babel in a single pass using webpack.
Removes all babel dependencies related to manually running it instead of letting webpack handle it.

This drastically simplifies generated example code, leading to the main asset bundle (which contains app code, excluding vendor and Polaris code) shrinking from 868kb to 198kb when doing a production build.

### How to 🎩

- Run `yarn dev`
- Ensure playground changes still hot-reloads in a timely fashion
- Ensure component modifications (e.g. changing src/components/DisplayText/DisplayText.tsx) still hot-reloads in a timely fashion
- Ensure README modifications (e.g. changing src/components/DisplayText/README.md) still hot-reloads in a timely fashion
- Note the size reduction in the main bundle in shrink-ray


playground/webpack.config.js
```diff
@@ -97,7 +97,23 @@ module.exports = (env = {production: false}) => ({
     rules: [
       {
         test: /\.md$/,
-        use: [{loader: `${__dirname}/webpack/parseMarkdown.js`}],
+        use: [
+          {
```
amrocha: How does this work? You parse MD files with babel first?

playground/webpack/parseMarkdown.js
```diff
@@ -16,9 +27,65 @@ module.exports = function loader() {
   });
   this.cacheable();
 
-  const stringyData = JSON.stringify(parseMarkdown(files), null, 2);
+  const data = parseMarkdown(files);
+
+  // Work around JSON.stringify() not supporting functions.
```
amrocha: I'm gonna have to ask you to explain this one in person later today

playground/webpack.config.js
BPScott: Other way around.  [Loaders are ran bottom to top](https://webpack.js.org/concepts/loaders/#configuration).

Order of operations is:
- Use the loader defined in parseMarkdown to turn the md files into  a single es6 module
- run that module through babel using babel-loader.

Same sort of thing as below - we run ts files through ts-loader then babel-loader.

playground/webpack/parseMarkdown.js
BPScott: It's a solid brainfuck. Let's chat and I can improve the comments

BPScott:
> I was confused by the second magnifier emoji:

Gonna go with the tick. There's a bit of lag time between when the readmes are done parsing and when webpack starts outputing stuff and I want people to know it's not the readme parsing being slow :p

playground/webpack/parseMarkdown.js
```diff
@@ -16,9 +27,65 @@ module.exports = function loader() {
...
+    const scopeValues = scopeKeys.map((key) => scope[key]);
+
+    // Replace the empty parameter list with a list based upon the scope.
+    // We can't use a placeholder in the parmeter list and search/replace that
+    // because the placeholder's name may be mangled when the code is minified.
+    const fnString = fn
+      .toString()
+      .replace(/^function(\s*)\(\)/, `function$1(${scopeKeys.join(', ')})`);
+
+    // Use apply here as it compiles to something shorter than the spread operator
```
amrocha: Is the difference big enough to be worth using apply? Like, I'm familiar with apply but I would say the spread operator is a lot more readable, and this code is already complex so I would optimize for readability

playground/webpack/parseMarkdown.js
```diff
@@ -16,9 +27,65 @@ module.exports = function loader() {
...
+  // It will transform:
+  // function() { return (<Button>Hi</Button>) }
+  // into:
+  // function(React, Button) { return (<Button>Hi</Button>) }
+  // and then call that function with React and Button as the arguments
+  const codeInvoker = function(fn, scope) {
+    const scopeKeys = Object.keys(scope);
+    const scopeValues = scopeKeys.map((key) => scope[key]);
+
+    // Replace the empty parameter list with a list based upon the scope.
```
amrocha: This is just my opinion so if you disagree feel free to leave it as is:

I really don't like long explanatory comments for every line like this. I usually feel like they make the code less readable by breaking up the code block. I'd prefer to have this included in the explanation above.

playground/webpack/parseMarkdown.js
amrocha: Our linting rules seem to agree as well

playground/webpack/parseMarkdown.js
```diff
@@ -16,9 +27,65 @@ module.exports = function loader() {
...
+    // We can't use a placeholder in the parmeter list and search/replace that
```
kaelig:
```suggestion
    // We can't use a placeholder in the parameter list and search/replace that
```

playground/webpack/parseMarkdown.js
BPScott: It was a difference in one of my first passes at this where I inlined the codeInvoker function in every example code block. It's a negligble difference now I only define it once +1

Will go back to using spread

PR #690, opened 2018-11-28
[ResourceList.Item] Add support for cmd / ctrl click

kyledurand:

Resolves an issue outlined in a project I'm working now where users cannot command / control click to open up list items in new tabs.

This is useful for opening more than one item at a time.

Fixes the above ☝️ 

### 🎩 checklist


lemonmade:
This isn't specifically relevant to this implementation, but why are we not just relying on native link behaviours here?

danrosenthal:
> This isn't specifically relevant to this implementation, but why are we not just relying on native link behaviours here?

Resource list items often contain links and buttons, so we have not implemented the item as a link as links can't contain other links.

dfmcphee:
> Resource list items often contain links and buttons, so we have not implemented the item as a link as links can't contain other links.

Didn't we used to have a version that overlayed a visually hidden `a` tag so that you got this behaviour without the nesting?

danrosenthal:
> Didn't we used to have a version that overlayed a visually hidden a tag so that you got this behaviour without the nesting?

We do something like this in the [polaris rails implementation of resource list](https://polaris-rails.shopify.com/components/resource_list), but that's not without its problems (issues like highlighting and selecting text.)

src/components/ResourceList/components/Item/Item.tsx
```diff
@@ -334,6 +335,10 @@ export class Item extends React.PureComponent<CombinedProps, State> {
       onClick(id);
     }
 
+    if (ctrlKey || metaKey) {
+      return window.open(url, '_blank');
```
dleroux: Should do a check for `url` because it's not required if there's an onClick.

Also I wonder, if there is an onClick and a URL should the this command click happen first and the onClick be ignored?

dleroux:
> Didn't we used to have a version that overlayed a visually hidden a tag so that you got this behaviour without the nesting?

We render a link if a URL is provided, put it behind the item content and call `anchor.click();` after everything we care about in the clicking has occurred.

src/components/ResourceList/components/Item/Item.tsx
kyledurand: I figured we should err on the side of caution here and allow the onClick to fire regardless.

kyledurand:
Thanks for the tophat and comments @dleroux. I've updated this PR

PR #695, opened 2018-11-29
[Page][Prototype] ✨ add `forceRender` to force render of the title, actions & breadcrumbs to be in page vs in the top bar.

michelleyschen:

Resolves https://github.com/Shopify/polaris-react/issues/669#issuecomment-442485366

Added `forceRender` to title, breadcrumbs, and actions render in the app when app-bridge is use

Default using app bridge 
ie. `<AppProvider apiKey={YOUR_API_KEY}>`
by default

Default without app bridge
ie. `<AppProvider apiKey={undefined}>`
with both

Using app bridge & forceRenderTitle
ie. `<AppProvider apiKey={YOUR_API_KEY}><Page title="test title" forceRenderTitle /></AppProvider>`
with forcerendertitle

Using app bridge & forceRenderActions
ie. `<AppProvider apiKey={YOUR_API_KEY}><Page title="" primaryAction={{content: 'Save'}} forceRenderActions /></AppProvider>`
with forcerenderactions

Using app bridge, forceRenderTitle, and forceRenderActions
ie. `<AppProvider apiKey={YOUR_API_KEY}><Page title="test title" primaryAction={{content: 'Save'}} forceRenderTitle forceRenderActions /></AppProvider>`
with both

Using app bridge, forceRender
ie. `<AppProvider apiKey={YOUR_API_KEY}><Page title="test title" primaryAction={{content: 'Save'}} forceRender /></AppProvider>`
with both

## 

### How to 🎩

<details>

```jsx

interface State {}

  render() {
    return (
      <AppProvider apiKey={YOUR_API_KEY}>
          <Page title="Playground" breadcrumbs={[{content: 'previous page'}]} primaryAction={{content: 'Save'}}>
          </Page>
      </AppProvider>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -17,6 +17,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Added `onKeyPress`, `onKeyDown`, and `onKeyUp` to `Button` ([#860](https://github.com/Shopify/polaris-react/pull/860))
 - Added `monochrome` prop to `Button` and `Link` component ([#821](https://github.com/Shopify/polaris-react/pull/821))
 - Updated `Frame` layout and made `TopBar.UserMenu` visible on mobile ([#852](https://github.com/Shopify/polaris-react/pull/852))
+- Added a `forceRender` prop to `Page` to not delegate to the app bridge TitleBar action ([#695](https://github.com/Shopify/polaris-react/pull/695))
 - Changed `Tabs` example to contain children so the `Panel` renders for accessibility ([#893](https://github.com/Shopify/polaris-react/pull/893))
 - Fixed timezone not being accounted for in `ResourceList` date filter control ([#710](https://github.com/Shopify/polaris-react/pull/710))
 
--- a/src/components/Page/Page.tsx
+++ b/src/components/Page/Page.tsx
@@ -23,6 +23,13 @@ export interface Props extends HeaderProps {
   fullWidth?: boolean;
   /** Decreases the maximum layout width. Intended for single-column layouts */
   singleColumn?: boolean;
+  /**
+   * Force render in page and do not delegate to the app bridge TitleBar action
+   * @default false
+   * @embeddedAppOnly
+   * @see {@link https://polaris.shopify.com/components/structure/page#section-use-in-an-embedded-application|Shopify Page Component docs}
+   * */
+  forceRender?: boolean;
 }
 
 export type ComposedProps = Props & WithAppProviderProps;
@@ -39,7 +46,7 @@ export class Page extends React.PureComponent<ComposedProps, never> {
   private titlebar: AppBridgeTitleBar.TitleBar | undefined;
 
   componentDidMount() {
-    if (this.props.polaris.appBridge == null) {
+    if (this.delegateToAppbridge === false) {
       return;
     }
 
@@ -50,7 +57,7 @@ export class Page extends React.PureComponent<ComposedProps, never> {
   }
 
   componentDidUpdate(prevProps: ComposedProps) {
-    if (this.props.polaris.appBridge == null || this.titlebar == null) {
+    if (this.titlebar == null || this.delegateToAppbridge === false) {
       return;
     }
 
@@ -64,7 +71,7 @@ export class Page extends React.PureComponent<ComposedProps, never> {
   }
 
   componentWillUnmount() {
-    if (this.props.polaris.appBridge == null || this.titlebar == null) {
+    if (this.titlebar == null || this.delegateToAppbridge === false) {
       return;
     }
 
@@ -81,8 +88,7 @@ export class Page extends React.PureComponent<ComposedProps, never> {
     );
 
     const headerMarkup =
-      this.props.polaris.appBridge ||
-      this.hasHeaderContent() === false ? null : (
+      this.delegateToAppbridge || this.hasHeaderContent() === false ? null : (
         <Header {...rest} />
       );
 
@@ -94,7 +100,16 @@ export class Page extends React.PureComponent<ComposedProps, never> {
     );
   }
 
-  private hasHeaderContent() {
+  private get delegateToAppbridge(): boolean {
+    const {
+      polaris: {appBridge},
+      forceRender = false,
+    } = this.props;
+
+    return appBridge != null && forceRender === false;
+  }
+
+  private hasHeaderContent(): boolean {
     const {
       title,
       primaryAction,
--- a/src/components/Page/tests/Page.test.tsx
+++ b/src/components/Page/tests/Page.test.tsx
@@ -42,6 +42,20 @@ describe('<Page />', () => {
     };
   }
 
+  describe('forceRender renders children in page', () => {
+    const {
+      createSpy: titleBarCreateSpy,
+      restore: restoreTitleBarCreateMock,
+    } = mockTitleBarCreate();
+    const {page} = mountWithAppBridge(
+      <Page {...mockProps} title="new title" forceRender />,
+    );
+
+    expect(page.find(Header).exists()).toBeTruthy();
+    expect(titleBarCreateSpy).not.toHaveBeenCalled();
+    restoreTitleBarCreateMock();
+  });
+
   describe('children', () => {
     it('renders its children', () => {
       const card = <Card />;


resistorsoftware:
I am using latest Polaris and my Page component is not respecting this idea of forceRenderingActions seriously.. my setup is.. 

    <Page
        title="Dashboard"
        icon={this.props.inventory.icon}
        primaryAction={{content: 'Recommendations', target: 'APP', url: '/recommendations'}}
        forceRenderActions
      >

And yet my primaryAction button is rendering in that ugly separate top heading bar... what? I am now super confused.

michelleyschen:
@resistorsoftware the idea of `forceRenderActions` is only in this PR and and is meant to be a possible way of addressing exactly the confusion you mention above. 

This is unfortunately not part of Polaris 3.1.1 yet so the only way to see `forceRenderActions` in action is pulling this branch locally.

There are a few work around mentioned in https://github.com/Shopify/polaris-react/issues/669 to have this working today. (I personally like https://github.com/Shopify/polaris-react/issues/669#issuecomment-442972860 the most) 

I hope this help!

resistorsoftware:
Thanks @michelleyschen 

Personally, I hate wasting my time, and using workarounds is just that. I am not into re-visiting work a few months or longer down the road to simply remove workarounds that fix what should be fixed for real, now. 

I am seeing countless complaints and errors and bugs with App Bridge et al... including what happened last night. The button I see (rendered in the "wrong" place belongs to a completely different App running in a different tab, in the same shop! 

I am uncomfortable being a guinea pig as the Shopify team attempts to bring the App Bridge to life, and also, I am not pleased the documentation and directions do not follow when implementing Polaris. It should be WYSIWYG but instead, as this shows, it is more like chaos.

src/components/Page/Page.tsx
```diff
@@ -94,36 +95,71 @@ export class Page extends React.PureComponent<ComposedProps, never> {
...
 
-    return (
-      (title != null && title !== '') ||
-      primaryAction != null ||
-      (secondaryActions != null && secondaryActions.length > 0) ||
-      (actionGroups != null && actionGroups.length > 0) ||
-      (breadcrumbs != null && breadcrumbs.length > 0)
-    );
+    const hasHeaderTitle =
+      title != null && title !== '' && (appBridge ? forceRenderTitle : true);
```
a-mooz: Quick note: whenever you're comparing the string `''`, always consider if you need to `trim()` the string first to avoid bugs using extra spaces at the end.

michelleyschen: Because this is the existing logic in Polaris, I am un-sure about changing the check @tmlayton double checking if this is the desired behaviour or can I introduce the trim check?

resistorsoftware:
Do we have to wait till 2019 for Polaris to work with an Embedded App and Page component? Why is it taking so long to fix and make it consistent with the dreams and documentation?

tmlayton:
@michelleyschen I’d prefer to keep this approach a bit simpler where `forceRender` skips delegation to App Bridge entirely. Then if an app still wants to use App Bridge’s `TitleBar`, they can pull the `appBridge` instance off of context and interact with app bridge directly.

Personally, I’m leaning heavily towards removing app bridge from Polaris entirely. It seems to have caused more issues and confusion than it has been helpful. I’m thinking we would still provide a set of React components which wrap app bridge (maybe even keep the same names), as well as an `AppBridgeProvider` for instantiation. Wherever these components live, we can move the `ShopifyRoutePropagator` there as well.

@resistorsoftware the criticisms are appreciated. Given I did most of the work integrating App Bridge with Polaris, it’s disappointing it hasn’t gone smoothly. It’s a fair point about being forced into early adoption (guinea pig) with App Bridge. While App Bridge is ultimately the path forward at Shopify, we were too optimistic including it in v3.0.0 as App Bridge was being launched.

That being said, this is a pull request. Please keep comments specific to @michelleyschen’s approach to force-rendering `<Page />` when in an embedded context and direct issues and criticisms to either [GitHub issues](https://github.com/Shopify/polaris-react/issues) or, for App Bridge specifically, the [Shopify APIs & SDKs forum](https://ecommerce.shopify.com/c/shopify-apis-and-technology).

src/components/Page/Page.tsx
```diff
@@ -23,6 +23,10 @@ export interface Props extends HeaderProps {
   fullWidth?: boolean;
   /** Decreases the maximum layout width. Intended for single-column layouts */
   singleColumn?: boolean;
+  /** Force render of title and breadcrumbs in page on embedded app, default false */
+  forceRenderTitle?: boolean;
```
tmlayton: I'm thinking only a `forceRender` prop which simply skips app bridge delegation.

vividviolet:
> I am seeing countless complaints and errors and bugs with App Bridge et al... including what happened last night.

@resistorsoftware can you clarify who is complaining? Is is the merchant or app developers? 

> The button I see (rendered in the "wrong" place belongs to a completely different App running in a different tab, in the same shop!

Can you please log an issue in app-bridge so we can dig into this?

> Personally, I’m leaning heavily towards removing app bridge from Polaris entirely. It seems to have caused more issues and confusion than it has been helpful

@tmlayton I think a lot of work has already been done to integrate App Bridge with Polaris. The documentation does outline what's expected when using Polaris components inside an embedded app. There is definitely room to improve the docs. The nice thing about having the App Bridge integrated with Polaris is that developers can use the components as usual without thinking about App Bridge aside from the initial AppProvider set up. I think we need to work toward clarifying the docs and resolving the remaining bugs instead of removing App Bridge from Polaris entirely.

I think separating App Bridge out of Polaris would have the following drawbacks:

- inconsistent user experience between an embedded app using normal Polaris components vs. apps using the Polaris wrapped App Bridge components. For example, a lot of work has been put into standardizing the app title bar so that it is consistent and accounts for new features such as secondary nav. @katbuilt might have more thoughts on this
- inconsistent user experience when we allow an app to use a mix of both normal and App Bridge components
- developers being confused as to when to use the Polaris wrapped App Bridge components vs. the regular ones. For example, they might not know to use the App Bridge modal which results in the app displaying a modal that does not cover the full Admin and other UI differences

michelleyschen:
@katbuilt 

> The screenshots here only show the placement of primary action in relation to breadcrumb and title (purple button) Could you post some screenshots of what happens when there are primary actions and the maximum number of allowed secondary action buttons? What about apps using pagination?

The `forceRender` meant the app will render like any app without app bridge connection. So the actions buttons & pagination will work exactly like what the `Page` component listed on polaris guide. https://polaris.shopify.com/components/structure/page#navigation

> What would be the backwards-facing implications of this PR to apps who don't add the forceRender to their app code?

Any app or `Page` that does not have `forceRender` will continue to work like today.

michelleyschen:
@tmlayton update the code to just have `forceRender` prop. The `build-consumer` seems to be broken, how have you been testing appBridge related things recently?

keyfer:
I got the `build-consumer` to work fine, so I will try to pick this up and carry it towards the finish line.

src/components/Page/Page.tsx
```diff
@@ -23,6 +23,8 @@ export interface Props extends HeaderProps {
   fullWidth?: boolean;
   /** Decreases the maximum layout width. Intended for single-column layouts */
   singleColumn?: boolean;
+  /** Force render of title, breadcrumbs, and actions in page on embedded app, default false */
```
tmlayton:
```suggestion
  /**
   * Force render in page and do not delegate to the app bridge TitleBar action
   * @embeddedAppOnly
   */
```

src/components/Page/Page.tsx
```diff
@@ -94,7 +95,20 @@ export class Page extends React.PureComponent<ComposedProps, never> {
...
   }
 
-  private hasHeaderContent() {
+  private get renderWithAppBridge(): boolean {
+    const {
+      polaris: {appBridge},
+      forceRender = false,
+    } = this.props;
+
+    if (appBridge && !forceRender) {
```
tmlayton:
```suggestion
    if (appBridge != null && forceRender === false) {
```

src/components/Page/Page.tsx
```diff
@@ -94,7 +95,20 @@ export class Page extends React.PureComponent<ComposedProps, never> {
     );
   }
 
-  private hasHeaderContent() {
+  private get renderWithAppBridge(): boolean {
```
tmlayton: actually just update this whole method, also I prefer `delegateToAppBridge` because we have no concerns with how app bridge handles the action call.

```suggestion
  private get delegateToAppBridge(): boolean {
    const {
      polaris: {appBridge},
      forceRender = false,
    } = this.props;
    
    return appBridge != null && forceRender === false;
  }
```

src/components/Page/Page.tsx
```diff
@@ -39,7 +41,7 @@ export class Page extends React.PureComponent<ComposedProps, never> {
   private titlebar: AppBridgeTitleBar.TitleBar | undefined;
 
   componentDidMount() {
-    if (this.props.polaris.appBridge == null) {
+    if (!this.renderWithAppBridge) {
```
tmlayton:
```suggestion
    if (this. delegateToAppBridge === false) {
```

src/components/Page/Page.tsx
```diff
@@ -50,7 +52,7 @@ export class Page extends React.PureComponent<ComposedProps, never> {
   }
 
   componentDidUpdate(prevProps: ComposedProps) {
-    if (this.props.polaris.appBridge == null || this.titlebar == null) {
+    if (!this.titlebar || !this.renderWithAppBridge) {
```
tmlayton:
```suggestion
    if (this.titlebar == null || this. delegateToAppBridge === false) {
```

src/components/Page/Page.tsx
```diff
@@ -64,7 +66,7 @@ export class Page extends React.PureComponent<ComposedProps, never> {
   }
 
   componentWillUnmount() {
-    if (this.props.polaris.appBridge == null || this.titlebar == null) {
+    if (!this.titlebar || !this.renderWithAppBridge) {
```
tmlayton:
```suggestion
    if (this.titlebar == null || this. delegateToAppBridge === false) {
```

src/components/Page/Page.tsx
```diff
@@ -81,8 +83,7 @@ export class Page extends React.PureComponent<ComposedProps, never> {
     );
 
     const headerMarkup =
-      this.props.polaris.appBridge ||
-      this.hasHeaderContent() === false ? null : (
+      this.renderWithAppBridge || this.hasHeaderContent() === false ? null : (
```
tmlayton:
```suggestion
      this. delegateToAppBridge || this.hasHeaderContent() === false ? null : (
```

src/components/Page/Page.tsx
```diff
@@ -94,7 +95,20 @@ export class Page extends React.PureComponent<ComposedProps, never> {
     );
   }
 
-  private hasHeaderContent() {
+  private get renderWithAppBridge(): boolean {
+    const {
+      polaris: {appBridge},
```
a-mooz: `polaris` looks unused?

src/components/Page/Page.tsx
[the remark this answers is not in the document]
keyfer: Eslint isn't super happy about this. I could add an eslint ignore here, but I don't think that's needed.

src/components/Page/Page.tsx
keyfer: Here we're just pulling out `appBridge` from `polaris.`

src/components/Page/Page.tsx
```diff
@@ -23,6 +23,12 @@ export interface Props extends HeaderProps {
   fullWidth?: boolean;
   /** Decreases the maximum layout width. Intended for single-column layouts */
   singleColumn?: boolean;
+  /**
+   * Force render in page and do not delegate to the app bridge TitleBar action
```
a-mooz: I'm thinking from the perspective of someone without any context on the issue, would this description be helpful for that person?

src/components/Page/Page.tsx
```diff
@@ -50,7 +56,7 @@ export class Page extends React.PureComponent<ComposedProps, never> {
   }
 
   componentDidUpdate(prevProps: ComposedProps) {
-    if (this.props.polaris.appBridge == null || this.titlebar == null) {
+    if (this.titlebar == null || this.delegateToAppbridge === false) {
```
a-mooz: Would `!this.titlebar` work better here, since it also accounts for `this.titlebar` being `undefined`?  Is there a reason `undefined` should be treated differently from `null`, that I've glossed over?

src/components/Page/Page.tsx
```diff
@@ -81,8 +87,7 @@ export class Page extends React.PureComponent<ComposedProps, never> {
     );
 
     const headerMarkup =
-      this.props.polaris.appBridge ||
-      this.hasHeaderContent() === false ? null : (
+      this.delegateToAppbridge || this.hasHeaderContent() === false ? null : (
```
a-mooz: I think I had some feedback on this previously, can we add brackets to make the logic here explicit at a glance? 
```
(this.delegateToAppBridge ||  this.delegateToAppbridge || this.hasHeaderContent() === false) ? null : ...
```
vs
```this. delegateToAppbridge || (this.hasHeaderContent() === false ? null : ...)`

File: src/components/Page/Page.tsx (hunk 12 above)
Comment (keyfer) [reply src/components/Page/Page.tsx]: So, you're exactly right that `undefined` and `null` should be treated the same here. `==` actually does that for us.

File: src/components/Page/Page.tsx (hunk 13 above)
Comment (keyfer) [reply src/components/Page/Page.tsx]: We could do this, but it will throw some linting errors for "uneeded" brackets, so if we wanted to go down this route we'd have to add a comment ignoring the linting rule first, which I don't think is a great option.

File: src/components/Page/Page.tsx (hunk 12 above)
Comment (keyfer) [reply src/components/Page/Page.tsx]: You're exactly right that `undefined` and `null` should be treated the same here, and in fact the reason to use `==` is so that we can check whether a reference is either `null` _or_ `undefined`.

File: src/components/Page/Page.tsx (hunk 11 above)
Comment (tmlayton) [reply src/components/Page/Page.tsx]: I’m biased – what do you think? If someone, without context, uses this prop and the TitleBar no longer renders, I’d like them to know how they can still use TitleBar. Maybe we can add a link to [TitleBar](https://help.shopify.com/en/api/embedded-apps/app-bridge/actions/titlebar)?

File: src/components/Page/Page.tsx (hunk 13 above)
Comment (tmlayton) [reply src/components/Page/Page.tsx]: For styling and lint rules, we largely use the Shopify config. You can open an issue here https://github.com/Shopify/eslint-plugin-shopify or if it is a project-specific setting, open an issue here.

File: src/components/Page/Page.tsx (hunk 12 above)
Comment (tmlayton) [reply src/components/Page/Page.tsx]: What you have is good.

File: src/components/Page/Page.tsx [hunk 14]
```diff
@@ -94,7 +99,20 @@ export class Page extends React.PureComponent<ComposedProps, never> {
...
+    const {
+      polaris: {appBridge},
+      forceRender = false,
+    } = this.props;
+
+    if (appBridge != null && forceRender === false) {
+      return true;
+    }
+
+    return false;
```
tmlayton: You can just return the boolean evaluation above

```tsx
  private get delegateToAppBridge(): boolean {
    const {
      polaris: {appBridge},
      forceRender = false,
    } = this.props;
    
    return appBridge != null && forceRender === false;
  }
```

src/components/Page/Page.tsx
keyfer: We could link to TitleBar, but that does feel a bit weird too. Perhaps we should add something to here: https://polaris.shopify.com/components/structure/page#section-use-in-an-embedded-application, and link to that?

tmlayton: Follow the doc format like so and we can support these tags later in the style guide: https://github.com/Shopify/polaris-react/blob/ad4b8bdf06b1159cbccfba45125ef5bdf0f2f84b/src/components/AppProvider/types.ts#L24

tmlayton:
Changelog is still including entries that have since been released

keyfer:
Don't know how I missed that twice, sorry @tmlayton.

Should be good now.

UNRELEASED.md
```diff
@@ -12,7 +12,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - Added `onKeyPress`, `onKeyDown`, and `onKeyUp` to `Button` ([#860](https://github.com/Shopify/polaris-react/pull/860))
 - Added `monochrome` prop to `Button` and `Link` component ([#821](https://github.com/Shopify/polaris-react/pull/821))
-- Updated `Frame` layout and made `TopBar.UserMenu` visible on mobile ([#852](https://github.com/Shopify/polaris-react/pull/852))
```
michelleyschen: was this meant to be an edit or just addition?

src/components/Page/Page.tsx
```diff
@@ -23,6 +23,13 @@ export interface Props extends HeaderProps {
   fullWidth?: boolean;
   /** Decreases the maximum layout width. Intended for single-column layouts */
   singleColumn?: boolean;
+  /**
+   * Force render in page and do not delegate to the app bridge TitleBar action
+   * @default false
+   * @embeddedAppOnly
+   * @see {@link https://polaris.shopify.com/components/structure/page#section-use-in-an-embedded-application|Shopify Page Component docs}
```
michelleyschen: are we...updating the doc in this link?

src/components/Page/tests/Page.test.tsx
```diff
@@ -42,6 +42,18 @@ describe('<Page />', () => {
     };
   }
 
+  describe('forceRender renders children in page', () => {
+    const {
+      createSpy: titleBarCreateSpy,
+      restore: restoreTitleBarCreateMock,
+    } = mockTitleBarCreate();
+    const {page} = mountWithAppBridge(<Page {...mockProps} forceRender />);
```
michelleyschen: instead of using just `{...mockProps}`, I would do `{...mockProps} title="new title"`

this ensure even if the default props change, this test won't fail.

PR #696, opened 2018-11-29
[AppProvider] Fix SSR

### LINKED ISSUE #372: window is not defined: AppProvider component seems not SSR-ready for embedded apps
# Issue summary

I am building a Shopify app using Polaris 2.3.1, Typescript and Next.js, which server-side renders (SSR). Everything is working smooth while not running embedded in the admin.

Running server-side rendered in embedded mode results in a server error: `window is not defined`

## Expected behavior

Polaris should be able to run embedded in admin even if server rendered.

## Actual behavior

Crashes server with `500: window is not defined` when running embedded in admin.

Starting a SSR app in the admin with `apiKey` and `shopOrigin` passed to `AppProvider` results in:

```
{ ReferenceError: window is not defined
  at checkFrameRedirect (.../node_modules/@shopify/polaris/index.js:1098:5)
```
Yikes. There is no `window` object on the server.

```
  ...
  
  static async getInitialProps(ctx) {
    return { ...ctx.req.shopify };
  }

  constructor(props) {
    super(props);
  }

  render() {
    return (
      <AppProvider
        apiKey={this.props.apiKey}
        shopOrigin={`https://${this.props.shop}.myshopify.com`}
      >
        <Page>
         ....
    }
```

## Specifications

- Polaris version number: 2.3.1
- Browser: Chrome Canary
- Device: MacBook
- Operating System: MacOS 10.13.5

## Thoughts
A potential fix could be to move the inter-frame-messaging and redirection setup to a later react live-cycle which runs in browser, not server.

tmlayton:
To prevent server side rendering from failing, which fixes #372.

Not instantiating app bridge if server side. The specific culprit is `getShopOrigin()` which ultimately reads `window.location`

## 

### How to 🎩

The setup here is fairly involved. I worked through this with @katiedavis and confirmed this fixes the issue with AppProvider. There still may be other components where we are missing a server side check. If you would like to 🎩 please reach out to one of us.

Because this is hard to 🎩 I at least committed a test first which fails then committed the changes which satisfies the test.

### Checklist

* [x] 🎩

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -17,6 +17,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Removed added space in `ChoiceList` when choice has children on selection but is not selected ([#665](https://github.com/Shopify/polaris-react/issues/665))
 - Fixed `errorOverlayText` on `Dropzone` ([#671](https://github.com/Shopify/polaris-react/pull/671))
 - Updated the `InlineError` text color, the error border-color on form fields and the error `Icon` color to be the same red. ([#676](https://github.com/Shopify/polaris-react/pull/676))
+- Fixed `AppProvider` server side rendering support ([#696](https://github.com/Shopify/polaris-react/pull/696)) (thanks [@sbstnmsch](https://github.com/sbstnmsch) for the [original issue](https://github.com/Shopify/polaris-react/issues/372))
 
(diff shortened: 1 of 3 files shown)


danrosenthal:
Make sure it has a CHANGELOG entry and we are good to go. Folks will be very happy with this fix in place.

PR #698, opened 2018-11-29
Final remove cyclical types

BPScott:

Fixes the final few cyclical depdendency complains and enables the no-cycle eslint rule

Moves types that are related to frame into the Frame component's types file. This includes moving some files that were in src/component/types into src/component/Frame/types too.

I've also moved DEFAULT_TOAST_DURATION into Frame/components/Toast too so that the Frame component is the source of these sorts of values, and the top-level Toast imports what it needs from Frame. (@danrosenthal and I talked about moving this into Frame/types but then I realized it's a value not a type so it doesn't really fit there.)

### How to 🎩

* yarn test
* yarn lint
* yarn run sk type-check


src/components/Frame/components/Toast/Toast.tsx
```diff
@@ -5,13 +5,13 @@ import {Key} from '../../../../types';
 
 import Icon from '../../../Icon';
 import KeypressListener from '../../../KeypressListener';
-import {Props as ToastProps, DEFAULT_TOAST_DURATION} from '../../../Toast';
+import {ToastProps} from '../../types';
 
 import * as styles from './Toast.scss';
 
-export type Props = ToastProps;
+export const DEFAULT_TOAST_DURATION = 5000;
```
danrosenthal: 👍

PR #700, opened 2018-11-30
Upgrade to use TypeScript 3.1.6

BPScott:

Keeping up to date.

Notably Typescript 3 offers [big improvements in error UX](https://blogs.msdn.microsoft.com/typescript/2018/07/30/announcing-typescript-3-0/#improved-errors-and-ux) - error messages due to incorrect typing shall be a lot more human readable now.

Update to typescript 3.1.6. Fix new warnings raised by running `yarn run sk type-check`.

### How to 🎩

Run tests and linting.

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
   },
   "devDependencies": {
     "@percy/puppeteer": "^0.4.0",
-    "@shopify/jest-dom-mocks": "^2.0.5",
+    "@shopify/jest-dom-mocks": "^2.1.1",
     "@shopify/js-uploader": "github:shopify/js-uploader",
     "@shopify/react-serialize": "^1.0.6",
     "@shopify/sewing-kit": "0.64.2",
@@ -187,6 +187,6 @@
     "lodash-decorators": "^4.3.5",
     "prop-types": "^15.6.1",
     "react-transition-group": "^2.4.0",
-    "tslib": "^1.8.0"
+    "tslib": "^1.9.3"
   }
 }
--- a/package.json
+++ b/package.json
@@ -146,7 +146,7 @@
     "shelljs": "^0.7.7",
     "shx": "^0.2.2",
     "svgo": "^0.7.2",
-    "typescript": "^2.9.1",
+    "typescript": "~3.1.6",
     "webpack": "~4.25.1",
     "webpack-cli": "3.1.0",
     "yargs": "^12.0.1"
@@ -156,7 +156,7 @@
     "react-dom": "^16.3.1"
   },
   "resolutions": {
-    "typescript-eslint-parser": "17.0.1"
+    "typescript-eslint-parser": "20.0.0"
   },
   "files": [
     "esnext",
--- a/src/components/DataTable/DataTable.tsx
+++ b/src/components/DataTable/DataTable.tsx
@@ -272,9 +272,7 @@ export class DataTable extends React.PureComponent<CombinedProps, State> {
     } = this;
     let {heights} = this.state;
     if (table) {
-      const rows = Array.from(table.getElementsByTagName('tr') as NodeListOf<
-        HTMLElement
-      >);
+      const rows = Array.from(table.getElementsByTagName('tr'));
 
       if (!truncate) {
         return (heights = rows.map((row) => {
--- a/src/components/DropZone/utils/index.ts
+++ b/src/components/DropZone/utils/index.ts
@@ -58,5 +58,5 @@ function isDragEvent(event: DropZoneEvent): event is DragEvent {
 function isChangeEvent(
   event: DropZoneEvent,
 ): event is React.ChangeEvent<HTMLInputElement> {
-  return 'target' in event;
+  return event.hasOwnProperty('target');
 }
--- a/src/components/DropZone/utils/index.ts
+++ b/src/components/DropZone/utils/index.ts
@@ -10,7 +10,7 @@ export function fileAccepted(file: File, accept: string | undefined) {
 }
 
 export function getDataTransferFiles(event: DropZoneEvent) {
-  if (isDragEvent(event)) {
+  if (isDragEvent(event) && event.dataTransfer) {
     const dt = event.dataTransfer;
 
     if (dt.files && dt.files.length) {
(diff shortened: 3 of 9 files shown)


src/utilities/breakpoints.ts
```diff
@@ -10,15 +10,19 @@ const noWindowMatches = {
   addListener: noop,
   removeListener: noop,
   matches: false,
-};
+  onchange: noop,
+  addEventListener: noop,
+  removeEventListener: noop,
+  dispatchEvent: (_: Event) => true,
+} as MediaQueryList;
```
BPScott: By casting this object as an instance of MediaQueryList then the two function below can infer their return type correctly, so we don't need to explictly state it

src/components/DropZone/utils/index.ts
```diff
@@ -10,7 +10,7 @@ export function fileAccepted(file: File, accept: string | undefined) {
 }
 
 export function getDataTransferFiles(event: DropZoneEvent) {
-  if (isDragEvent(event)) {
+  if (isDragEvent(event) && event.dataTransfer) {
```
BPScott: TS was concerned that event.dataTransfer may be null

src/components/DropZone/utils/index.ts
```diff
@@ -20,7 +20,7 @@ export function getDataTransferFiles(event: DropZoneEvent) {
       // events and uses `items` instead of `files` in this case.
       return Array.from(dt.items);
     }
-  } else if (event.target && event.target.files) {
+  } else if (isChangeEvent(event) && event.target.files) {
```
BPScott: TS was concerned that event.target may not exist because it didn't realize that at this point we must have a ChangeEvent for an HTMLInputElement - which does have a target

src/components/ResourceList/ResourceList.tsx
```diff
@@ -555,7 +555,7 @@ export class ResourceList extends React.Component<CombinedProps, State> {
 
       const overlay = listNode.getBoundingClientRect();
       const viewportHeight = Math.max(
-        document.documentElement.clientHeight,
+        document.documentElement ? document.documentElement.clientHeight : 0,
```
BPScott: TS was concerned that document.documentElement may be null

danrosenthal:
I've approved this, because I don't see any issue with it. But I'd love to also see @AndrewMusgrave and @tmlayton weigh in.

src/components/DataTable/DataTable.tsx
```diff
@@ -251,9 +251,7 @@ export class DataTable extends React.PureComponent<
     } = this;
     let {heights} = this.state;
     if (table) {
-      const rows = Array.from(table.getElementsByTagName('tr') as NodeListOf<
```
AndrewMusgrave: 🎉

src/components/DropZone/utils/index.ts
```diff
@@ -54,3 +54,9 @@ function accepts(file: File, acceptedFiles: string | string[] | undefined) {
 function isDragEvent(event: DropZoneEvent): event is DragEvent {
   return dragEvents.indexOf(event.type) > 0;
 }
+
+function isChangeEvent(
+  event: DropZoneEvent,
+): event is React.ChangeEvent<HTMLInputElement> {
+  return 'target' in event;
```
AndrewMusgrave: Should we use `hasOwnProperty` since `in` will return true if `target` is anywhere in the prototype chain?

src/utilities/breakpoints.ts
AndrewMusgrave: Rather than casting we should type the variable

```tsx
const noWindowMatches: MediaQueryList  = {
```

BPScott:
I've opened https://github.com/Shopify/quilt/pull/426 which shall let us remove that MediaQueryListListener interface definition

src/components/Sticky/Sticky.tsx
```diff
@@ -106,3 +107,7 @@ export default class Sticky extends React.Component<Props, State> {
     }
   }
 }
+
+function isFunction(arg: any): arg is Function {
```
kaelig: Should this be available as a utility somewhere? Or… should we use Lodash's isFunction instead?

package.json
```diff
@@ -158,7 +158,7 @@
     "react-dom": "^16.3.1"
   },
   "resolutions": {
-    "typescript-eslint-parser": "17.0.1"
+    "typescript-eslint-parser": "20.0.0"
```
kaelig: Should we (can we) get rid of this resolution?

src/components/Sticky/Sticky.tsx
BPScott: It's so small I don't think it's worth it. It's a bit of a hack to work around some of typescript's type shortening in this one case rather than something generally useful.

BPScott: We need this because of the 'arg is function' typescript hint. using lodash's isFunction wouldn't give us that.

package.json
BPScott: We could as this is what sewing-kit uses under the hood, however the next time sewing-kit changes it version of this we'll have to add it back to keep us in sync with the version of typescript that we use. I think it's worth keeping as a reminder.

src/components/Sticky/Sticky.tsx
```diff
@@ -53,8 +53,9 @@ export default class Sticky extends React.Component<Props, State> {
     const {style, isSticky} = this.state;
     const {children} = this.props;
 
-    const childrenContent =
-      typeof children === 'function' ? children(isSticky) : children;
```
AndrewMusgrave: Any idea why this needed to change? Normally the type check is enough of a guard.

BPScott: I believe this is due to this [breaking change](https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#narrowing-functions-now-intersects--object-and-unconstrained-generic-type-parameters) in how narrowing works for when children can be either a function a ReactNode. We now have to explicitly narrow it ourself with as type guard

BPScott:
yo @AndrewMusgrave, I've rebased this onto latest master, and have removed the hack for MediaQueryList typings. Can you give this one last check and I'll merge tomorrow.

BPScott:
I've ran `yarn build-consumer web` and ran `yarn run sk type-check` and `yarn run test` and nothing seems to be related to this update. (There's a few warnings around IconProps due to changes in v3.2.0 as web isn't using that yet)

PR #704, opened 2018-11-30
Fix percy false positives

amrocha:

Because percy has false positives

Adds a 2s delay before taking a screenshot

## 

### How to 🎩

Run percy a few times locally and make sure the problems that came up because we were taking a scerenshot too fast go away


amrocha:
It fixes some issues but the card thing is still broken :(
Gonna try to figure that one out before merging

amrocha:
Haven't had the time to dig into this. Since this change does fix a few examples I'm gonna go ahead and merge it, and address the card popup thing later

PR #708, opened 2018-11-30
[TextField] Use autocomplete="nope" instead of autocomplete="off"

jineshshah36:

Some browsers (I'm looking at you Chrome) ignore autocomplete="off" because they think they know better. As a result, setting autocomplete="off" does not disable autocomplete. A better way to disable autocomplete is to use any string that does not match a valid autocomplete value such as "nope". See https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#Disabling_autocompletion for more details.

This will not make any UI changes

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -7,7 +7,6 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Enhancements
 
 - `TextField` no longer uses `componentWillReceiveProps`([#628](https://github.com/Shopify/polaris-react/pull/628))
-- `TextField` uses autocomplete="nope" when `autoComplete` prop is `false` ([#708](https://github.com/Shopify/polaris-react/pull/708))
 - Fixed `Pagination` from calling `onNext` and `onPrevious` while `hasNext` and `hasPrevious` are false for key press events ([#643](https://github.com/Shopify/polaris-react/pull/643))
 - `EventListener` no longer uses `componentWillUpdate` ([#628](https://github.com/Shopify/polaris-react/pull/628))
 - Allowed `Icon` to accept a React Node as a source ([#635](https://github.com/Shopify/polaris-react/pull/635)) (thanks to [@mbriggs](https://github.com/mbriggs) for the [original issue](https://github.com/Shopify/polaris-react/issues/449))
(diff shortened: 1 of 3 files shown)


UNRELEASED.md
```diff
@@ -7,6 +7,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Enhancements
 
 - `TextField` no longer uses `componentWillReceiveProps`([#628](https://github.com/Shopify/polaris-react/pull/628))
+- `TextField` uses autocomplete="nope" when `autoComplete` prop is `false` ([#708](https://github.com/Shopify/polaris-react/pull/708))
```
danrosenthal: Can you change this to use past tense instead of present tense (`Fixed`) and move it into the `Bug fixes` section, as this feels more like a bug fix than an enhancement?

danrosenthal:
🎩 this PR on branch `tophat-708` with this playground:

```tsx

interface State {
  value: string | undefined;
}

  state: State = {
    value: undefined,
  };

  render() {
    return (
      <Page
        title="Playground"
      >
        <TextField
          label="Store name"
          value={this.state.value}
          onChange={this.handleChange}
          autoComplete
          type="email"
        />
      </Page>
    );
  }

  handleChange = (value: string) => {
    this.setState({value});
  };
}
```

UNRELEASED.md
jineshshah36: Done

mbaumbach:
@danrosenthal @jineshshah36 I'm actually seeing the reverse situation in Chrome 71. When `autocomplete="nope"` is used, I'm getting autocomplete results, but when it's set to `autocomplete="off"` I don't. It looks like Kent C. Dodd and the PayPal team had to pull a 180° on this as well (see: https://github.com/paypal/downshift/issues/527#issuecomment-407413031)

I'm specifically testing this with the `Autocomplete` component and setting the `Autocomplete.TextField` to have `autoComplete={false}`.

jineshshah36:
I'm also seeing this now. I'm thinking we should also revert this

StupidSexyJake:
If you change it to `autocomplete='__nope' ` then it works.

PR #709, opened 2018-12-03
[TextField / Labelled] Add character count

### LINKED ISSUE #491: TextField remaining characters count
# Feature request summary

It would be nice if TextFields had the option to show the number of characters remaining when `maxLength` prop is set (see example below), or just the number of characters in the input if no `maxLength` is set.

screen shot 2018-10-29 at 1 48 42 pm

Perhaps a simple `showCharacterCount` prop can be used to enable/disable the character count being displayed.

## Rationale

It is quite common when building forms to have text input fields that are character limited. Simply setting a max length on the field means that the user has no visual feedback that they have run out of characters except for that when they continue to type characters are no longer added to the input.

This also means that users can't see how many characters they have remaining while trying to decide what to type.

---

## Possible Implementation
Because the `TextField` component uses the `Labelled` component to show the label and `action` (if it exists), my initial thought was to extend the `Labelled` component to optionally display the number of characters remaining in the `TextField`. But this does mean that `TextField` has to pass optional information about the character count (as well as whether or not to show the count) to `Labelled`.

Another implementation could be simply creating another component that wraps the existing `TextField` and displays the character count. I feel like this might need some interesting css though to get the count to line up properly and adjust its position for when a `label action` is also used (like below).

screen shot 2018-10-29 at 3 46 39 pm

jfangrad:

Resolves #491

TextField lacks the ability to display the remaining characters available when the TextField has a max character length. This PR adds that ability.

Adds in the prop `showCharacterCount` to TextField to enable the displaying of the remaining character count.

The `Labelled` component will then display the number of characters out of the total allowed characters that have been inputted into the TextField.

screen shot 2018-12-03 at 4 21 40 pm

One item of note is that things can get a little busy as can be seen below:

screen shot 2018-11-29 at 4 20 26 pm

@dpersing was giving insight into this on the issue.

## 

### How to 🎩

<details>

```jsx

interface State {
  value: string;
}

  state = {
    value: '',
  };

  onChange = (value: string) => {
    this.setState({value});
  };

  render() {
    const {value} = this.state;
    return (
      <Page
        title="Playground"
      >
        <TextField
          label="This is where the label lives"
          value={value}
          onChange={this.onChange}
          maxLength={10}
          showCharacterCount
        />
        <TextField
          label="This is where the label lives"
          value={value}
          onChange={this.onChange}
          maxLength={10}
          multiline         
          showCharacterCount
        />
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/TextField/README.md
+++ b/src/components/TextField/README.md
@@ -497,7 +497,7 @@ class HelpTextExample extends React.Component {
 
 Use as a special form of help text that works best inline.
 
-- Use a prefix for things like currency symbols (e.g. “$”, “¥”, “£”).
+- Use a prefix for things like currency symbols (e.g. “\$”, “¥”, “£”).
 - Use suffix for things like units of measure (e.g. “in”, “cm”).
 
 ```jsx
@@ -735,6 +735,36 @@ Use to show that a textfield is not available for interaction. Most often used i
 
 ```jsx
 <TextField label="Store name" disabled />
+```
+
+### Text field with character count
+
+<!-- example-for: web -->
+
+Use to display the current number of characters in a text field. Use in conjunction with max length to display the current remaining number of characters in the text field.
+
+```jsx
+class TextFieldExample extends React.Component {
+  state = {
+    value: 'Jaded Pixel',
+  };
+
+  handleChange = (value) => {
+    this.setState({value});
+  };
+
+  render() {
+    return (
+      <TextField
+        label="Store name"
+        value={this.state.value}
+        onChange={this.handleChange}
+        maxLength={20}
+        showCharacterCount
+      />
+    );
+  }
+}
 ```
 
 ---
--- a/src/components/TextField/TextField.scss
+++ b/src/components/TextField/TextField.scss
@@ -170,6 +170,18 @@ $stacking-order: (
   pointer-events: none;
 }
 
+.CharacterCount {
+  @include text-emphasis-subdued;
+  z-index: z-index(contents, $stacking-order);
+  margin-right: $backdrop-horizontal-spacing;
+  line-height: control-height();
+  pointer-events: none;
+}
+
+.AlignFieldBottom {
+  align-self: flex-end;
+}
+
 .Spinner {
   z-index: z-index(contents, $stacking-order);
   display: flex;
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -7,6 +7,7 @@ import Labelled, {Action, helpTextID, labelID} from '../Labelled';
 import Connected from '../Connected';
 
 import {Error, Key} from '../../types';
+import {withAppProvider, WithAppProviderProps} from '../AppProvider';
 import {Resizer, Spinner} from './components';
 import * as styles from './TextField.scss';
 
@@ -96,6 +97,8 @@ export interface BaseProps {
   ariaActiveDescendant?: string;
   /** Indicates what kind of user input completion suggestions are provided */
   ariaAutocomplete?: string;
+  /** Indicates whether or not the character count should be displayed */
+  showCharacterCount?: boolean;
   /** Callback when value is changed */
   onChange?(value: string, id: string): void;
   /** Callback when input is focused */
@@ -112,17 +115,19 @@ export type Props = NonMutuallyExclusiveProps &
     | {disabled: true}
     | {onChange(value: string, id: string): void});
 
+export type CombinedProps = Props & WithAppProviderProps;
+
 const getUniqueID = createUniqueIDFactory('TextField');
 
-export default class TextField extends React.PureComponent<Props, State> {
-  static getDerivedStateFromProps(nextProps: Props, prevState: State) {
+class TextField extends React.PureComponent<CombinedProps, State> {
+  static getDerivedStateFromProps(nextProps: CombinedProps, prevState: State) {
     return {id: nextProps.id || prevState.id};
   }
 
   private input: HTMLElement;
   private buttonPressTimer: number;
 
-  constructor(props: Props) {
+  constructor(props: CombinedProps) {
     super(props);
 
     this.state = {
@@ -132,7 +137,7 @@ export default class TextField extends React.PureComponent<Props, State> {
     };
   }
 
-  componentDidUpdate({focused}: Props) {
+  componentDidUpdate({focused}: CombinedProps) {
     if (
       this.input &&
       focused !== this.props.focused &&
@@ -177,6 +182,8 @@ export default class TextField extends React.PureComponent<Props, State> {
       ariaActiveDescendant,
       ariaAutocomplete,
       ariaControls,
+      showCharacterCount,
+      polaris: {intl},
     } = this.props;
 
     const {height} = this.state;
@@ -202,6 +209,35 @@ export default class TextField extends React.PureComponent<Props, State> {
     const suffixMarkup = suffix ? (
       <div className={styles.Suffix} id={`${id}Suffix`}>
         {suffix}
+      </div>
+    ) : null;
+
+    const characterCount = value.length;
+    const characterCountLabel = intl.translate(
+      maxLength
+        ? 'Polaris.TextField.characterCountWithMaxLength'
+        : 'Polaris.TextField.characterCount',
+      {count: characterCount, limit: maxLength},
+    );
+
+    const characterCountClassName = classNames(
+      styles.CharacterCount,
+      multiline && styles.AlignFieldBottom,
+    );
+
+    const characterCountText = !maxLength
+      ? characterCount
+      : `${characterCount}/${maxLength}`;
+
+    const characterCountMarkup = showCharacterCount ? (
+      <div
+        id={`${id}CharacterCounter`}
+        className={characterCountClassName}
+        aria-label={characterCountLabel}
+        aria-live="polite"
+        aria-atomic="true"
+      >
+        {characterCountText}
       </div>
     ) : null;
 
@@ -231,6 +267,9 @@ export default class TextField extends React.PureComponent<Props, State> {
     }
     if (helpText) {
       describedBy.push(helpTextID(id));
+    }
+    if (showCharacterCount) {
+      describedBy.push(`${id}CharacterCounter`);
     }
 
     const labelledBy = [labelID(id)];
@@ -303,6 +342,7 @@ export default class TextField extends React.PureComponent<Props, State> {
             {prefixMarkup}
             {input}
             {suffixMarkup}
+            {characterCountMarkup}
             {spinnerMarkup}
             <div className={styles.Backdrop} />
             {resizer}
@@ -414,3 +454,5 @@ function normalizeAutoComplete(autoComplete?: boolean) {
   }
   return autoComplete ? 'on' : 'off';
 }
+
+export default withAppProvider<Props>()(TextField);
--- a/src/components/TextField/tests/TextField.test.tsx
+++ b/src/components/TextField/tests/TextField.test.tsx
@@ -305,6 +305,41 @@ describe('<TextField />', () => {
     });
   });
 
+  describe('characterCount', () => {
+    it('displays number of characters entered in input field', () => {
+      const textField = mountWithAppProvider(
+        <TextField
+          value="test"
+          showCharacterCount
+          label="TextField"
+          id="MyField"
+          onChange={noop}
+        />,
+      );
+
+      const characterCount = textField.find('#MyFieldCharacterCounter');
+
+      expect(characterCount.text()).toBe('4');
+    });
+
+    it('displays remaining characters as fraction in input field with maxLength', () => {
+      const textField = mountWithAppProvider(
+        <TextField
+          value="test"
+          maxLength={10}
+          showCharacterCount
+          label="TextField"
+          id="MyField"
+          onChange={noop}
+        />,
+      );
+
+      const characterCount = textField.find('#MyFieldCharacterCounter');
+
+      expect(characterCount.text()).toBe('4/10');
+    });
+  });
+
   describe('type', () => {
     it('sets the type on the input', () => {
       const type = shallowWithAppProvider(
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -192,6 +192,11 @@
 
     "Tag": {
       "ariaLabel": "Remove {children}"
+    },
+
+    "TextField": {
+      "characterCount": "{count} characters",
+      "characterCountWithMaxLength": "{count} of {limit} characters used"
     }
   }
 }
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -10,6 +10,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- `TextField` accepts a `showCharacterCount` prop to enable the display of character count ([#709](https://github.com/Shopify/polaris-react/pull/709))
+
 ### Bug fixes
 
 - Fixed vertical misalignment in `Banner.Header`([#870](https://github.com/Shopify/polaris-react/pull/870))


src/components/Labelled/Labelled.tsx
```diff
@@ -48,6 +57,17 @@ export default function Labelled({
     </div>
   ) : null;
 
+  const characterCountMarkup = showCharacterCount &&
+    maxLength && (
+      <span
+        id={`${id}-labeled-character-counter`}
+        className={styles.CharacterCount}
+        aria-live="polite"
```
dpersing: This looks great, @jfangrad!

I tested with a few common screen reader/browser combos and the character count help text is read when the field gets focus, and the updates get read. One small recommendation I'd make is to add `aria-atomic="true"` to the `aria-live` container here. Right now, NVDA with Firefox is only reading the updated character count, not the whole string, so the context is not quite as clear. Adding that attribute will cause the whole `<span>` to be read when it changes.

src/components/Labelled/Labelled.tsx
```diff
@@ -36,6 +42,9 @@ export default function Labelled({
   helpText,
   children,
   labelHidden,
+  characterCount,
+  maxLength,
+  showCharacterCount,
```
danrosenthal: The `showCharacterCount` prop here feels unnecessary. As an implementor I would expect that if I pass a character count, it should show a character count. And if I don't pass a character count, it should not show a character count.

src/components/Labelled/Labelled.tsx
```diff
@@ -48,6 +57,17 @@ export default function Labelled({
     </div>
   ) : null;
 
+  const characterCountMarkup = showCharacterCount &&
+    maxLength && (
```
danrosenthal: Would there ever be a case where this might be used without a maxLength? Does it make sense to only render it if the `maxLength` is provided?

src/components/Labelled/Labelled.tsx
```diff
@@ -48,6 +57,17 @@ export default function Labelled({
...
+      <span
+        id={`${id}-labeled-character-counter`}
+        className={styles.CharacterCount}
+        aria-live="polite"
+      >
+        {`${characterCount} of ${maxLength} characters used`}
```
danrosenthal: This string will need to be externalized and internationalized. Some languages will change the position of words here, and we avoid hard-coding strings in our components. This is done using the `AppProvider` component and the `intl` context it provides. Strings live in the `en.json` file at the root of this project.

Grepping around the project should show you some examples of how that is done.

src/components/Labelled/Labelled.tsx
danrosenthal: Bumping @dpersing's suggestion to add `aria-atomic="true"` to the `aria-live` container.

src/components/TextField/tests/TextField.test.tsx
```diff
@@ -595,13 +595,18 @@ describe('<TextField />', () => {
...
         />,
       );
 
       expect(textField.find(Labelled)).toHaveLength(1);
       expect(textField.find(Labelled).prop('label')).toBe('TextField');
       expect(textField.find(Labelled).prop('id')).toBe('MyField');
       expect(textField.find(Labelled).prop('helpText')).toBe('Help text');
+      expect(textField.find(Labelled).prop('showCharacterCount')).toBe(true);
+      expect(textField.find(Labelled).prop('characterCount')).toBe(0);
+      expect(textField.find(Labelled).prop('maxLength')).toBe(5);
```
danrosenthal: This could also use a test for when the field's value is of length greater than zero, asserting that the character count does in fact count the characters.

solonaarmstrong:
I wonder if it would make more sense to position the text count here: 
screen shot 2018-12-14 at 7 03 08 am

This is how it is already implemented across the admin. It looks like it's using text style as a sibling to the label.

danrosenthal:
I'm also in favor of that positioning, @solonaarmstrong. cc: @sarahill to confirm

jfangrad:
@solonaarmstrong @danrosenthal I also agree and think that it looks better in that position. The one reason, that came up in the issue, for putting it below was that if the textfield had a lot of text in it the user might not be able to see the count as they type if it was at the top of the field.

Its probably a rare case but just thought I would bring that up again to be included in the consideration.

danrosenthal:
> The one reason, that came up in the issue, for putting it below was that if the textfield had a lot of text in it the user might not be able to see the count as they type if it was at the top of the field.

I think that's an acceptable limitation.

sarahill:
I like @solonaarmstrong suggestion for placement.

Couple questions/thoughts:
- How would this look on smaller screens.
    - We could either not show the count on or reduce it to just the count `4 / 70`.
- Do we need to worry about a long label that could potentially run into the character count?
    - For this we could set a max-width and allow the label to wrap if necessary and base align the label + character count.

solonaarmstrong:
@sarahill This is how it currently behaves on small screens in the admin (for SEO, as an example). It simply needs to stack.

screen shot 2018-12-14 at 9 07 37 am

src/components/Labelled/Labelled.tsx
solonaarmstrong: I had the same feeling. If there is no `maxLength`, I would change the copy to `n characters used` or simply `n characters`.

src/components/Labelled/Labelled.tsx
```diff
@@ -26,6 +26,12 @@ export interface Props {
   children?: React.ReactNode;
   /** Visually hide the label */
   labelHidden?: boolean;
+  /** Character count in the labled input */
+  characterCount?: number;
+  /** Maximum character length for an input */
+  maxLength?: number;
```
solonaarmstrong: It seems a bit unclear what has a `maxLength`: the input or the label. This might want to be named `inputMaxLength`.

src/components/TextField/TextField.tsx
```diff
@@ -96,6 +96,8 @@ export interface BaseProps {
   ariaActiveDescendant?: string;
   /** Indicates what kind of user input completion suggestions are provided */
   ariaAutocomplete?: string;
+  /** Indicates whether the character count should be displayed or not */
```
solonaarmstrong:
```suggestion
  /** Indicates whether or not the character count should be displayed */
```

src/components/Labelled/Labelled.tsx
```diff
@@ -26,6 +26,12 @@ export interface Props {
   children?: React.ReactNode;
   /** Visually hide the label */
   labelHidden?: boolean;
+  /** Character count in the labled input */
```
solonaarmstrong:
```suggestion
  /** Character count in the labeled input */
```

solonaarmstrong: Or better yet
```suggestion
  /** Displays the number of characters entered in the input */
```

jfangrad:
@danrosenthal @solonaarmstrong I made the changes to the placement etc.

One thing that I ran into is that the placement of the label action gets a little weird when also showing the character count. Currently this is what I have it doing when both are present:
screen shot 2018-12-17 at 1 34 08 pm

Its probably a super rare case to have all three...
To have the label action always be on the far right I would have to move the character count code down another level into the `Label` component which wouldn't be super nice.

label action will still be on the far right if the `showCharacterCount` is false

jfangrad:
@danrosenthal I definitely agree with you. It felt weird putting all the logic in Labelled.

I like the idea of passing it through the action prop, my only concern is that you will then not be able to have both at the same time. I know for my project that isn't a concern. I guess if it becomes a problem in the future that can be addressed then?

My main question then is if someone passes a `labelAction` prop and the `showCharacterCount` prop to the `TextField` which should we pass to the `Labelled` components `action` prop? Which takes precedence? My initial thought is action should probably take precedence since it has been around longer.

I am also worried about people trying to do an action and a character count and wondering why one of them doesn't show up (But maybe just saying why in the docs is enough?)

danrosenthal:
It feels to me like there should only be one element ever rendered in that slot, similar to how we handle it in [Polaris rails](https://polaris-rails.shopify.com/components/text_field#side-controls) with side controls.

This feels like a good time to bring in another opinion, so I've asked @tmlayton to weigh in.

tmlayton:
A couple things off the bat:

- There are visual regressions here where labels are being rendered when they shouldn’t (check the Percy link)
- Let’s add an example for this

--- 

As to the approach, yes this is a concern of the text field, at least the logic and string generation. The label should be naive here and only care about layout and styling of the string, not its contents. It seems to me we are adding two features, 1) text field generating a character count / remaining characters string and 2) labelled supporting upper right aligned tertiary text.

Similar to how we pass help text to labelled, I would expect a prop like `tertiaryText` or `infoText` on labelled and we simply pass the generated count string from text field to the new labelled prop. I would also expect `labelHidden` to not visually render this `infoText`, but still set semantic/aria attributes.

tmlayton:
A new prop in addition to label action – I think using the action prop distorts the API.

> Also since the labelHidden prop doesn't hide the labelAction why should it hide the character count (or infoText).

I can see the reasoning here, but actions need to interacted with thus have a dependency on being rendered to serve any purpose. Labels, however, don't need to be rendered and can still serve a purpose as aria labels, etc. So my idealism-based position would be if it is purely information (nothing to interact with) `labelHidden` should hide it. But...

> I can think of a couple cases in a project I am working on where we want a hidden label but would still want the character count to show.

...I think pragmatism will be better here 😄 (aka `labelHidden` hides the label, not the character count)

sarahill:
Some quick suggestions for displaying the character count. I think this would work with the `labelHidden`. I don't know if we want to change it that much but just some thoughts.

image

cc: @dpersing

dpersing:
I like including the count as part of the field! The only case where things might be weird is if the field is very long and magnification cuts off the number.

For non-visual users, I think we'd want to provide a bit more context for the number. Something like "X characters remaining" would be good (and still with the `aria-atomic="true"` for consistency).

jfangrad:
@danrosenthal how do you feel about moving to the design Sara has suggested? It does eliminate a lot of the issues we were running into and make the implementation potentially a lot simpler.

danrosenthal:
@jfangrad I feel good about it. Would you mind taking that on?

src/components/TextField/TextField.scss
```diff
@@ -170,6 +170,22 @@ $stacking-order: (
...
+  z-index: z-index(contents, $stacking-order);
+  margin-right: $backdrop-horizontal-spacing;
+  pointer-events: none;
+}
+
+.MaxCharacters {
+  color: color('red');
+}
+
+.FlexEnd {
```
danrosenthal: I think this could use a more semantic name that describes its purpose.

src/components/TextField/TextField.scss
```diff
@@ -170,6 +170,22 @@ $stacking-order: (
...
+  align-self: flex-end;
+  margin-bottom: rem(4px);
```
danrosenthal: What's the purpose of setting a `margin-bottom` here? Also, this could use `spacing(extra-tight)`.

src/components/TextField/TextField.scss
```diff
@@ -170,6 +170,22 @@ $stacking-order: (
   pointer-events: none;
 }
 
+.CharacterCount {
+  @include text-emphasis-subdued;
+  z-index: z-index(contents, $stacking-order);
```
danrosenthal: Is the `z-index` necessary to specify here?

src/components/TextField/TextField.tsx
```diff
@@ -204,6 +211,32 @@ export default class TextField extends React.PureComponent<Props, State> {
...
+      maxLength
+        ? 'Polaris.TextField.characterCountWithMaxLength'
+        : 'Polaris.TextField.characterCount',
+      {count: characterCount},
+    );
+
+    const characterCountClassName = classNames(
+      styles.CharacterCount,
+      multiline && styles.FlexEnd,
+      maxLength && characterCount === 0 && styles.MaxCharacters,
```
danrosenthal: If possible, it would be great to add code coverage for these conditions as well. (224 and 225)

danrosenthal:
@dpersing you can use the deployed example to tophat here: https://polaris-react-pr-709.herokuapp.com/text-field/text-field-with-character-count

dpersing:
Tested the updated version with VoiceOver/Safari on macOS and iOS and NVDA with Firefox on Windows 10.

On macOS, VoiceOver is totally fine with the `aria-label` for the live region update. On iOS, and on Windows, the `aria-label` is skipped, but the region content is read out as expected. Since a user will hear the initial "X characters remaining" information when the field gets focus (or after a short delay), I think this is good to ship with.

solonaarmstrong:
Personally, I would do this a bit differently. I would either offer the consumer the option to count up or count down. Or I would be opinionated about counting up.

In production, the character count reads as `10 out of 30 characters used` (it counts up). I think counting up is more common in the wild. I initially was confused why 11 characters were displaying `9`. There is also no way to append `characters used` or `characters left`.

solonaarmstrong:
As well, @jfangrad if you want me to write some more tests, I like writing tests. Just don't want to step on your toes.

src/components/TextField/TextField.tsx
```diff
@@ -204,6 +211,32 @@ export default class TextField extends React.PureComponent<Props, State> {
...
+    );
+
+    const characterCountMarkup = showCharacterCount ? (
+      <div
+        id={`${id}-character-counter`}
```
solonaarmstrong:
```suggestion
         id={`${id}CharacterCounter`}
```

src/components/TextField/tests/TextField.test.tsx
```diff
@@ -305,6 +305,41 @@ describe('<TextField />', () => {
...
+        <TextField
+          value="test"
+          showCharacterCount
+          label="TextField"
+          id="MyField"
+          onChange={noop}
+        />,
+      );
+
+      const characterCount = textField.find('#MyField-character-counter');
```
solonaarmstrong: This is probably a good case for using `findByTestID`. Throw a `testID` on the div element and
```suggestion
       const characterCount = findByTestID(textField, 'CharacterCounter');
```

src/components/TextField/tests/TextField.test.tsx
```diff
@@ -305,6 +305,41 @@ describe('<TextField />', () => {
...
+
+      expect(characterCount.text()).toBe('4');
+    });
+
+    it('counts characters in input field with max length', () => {
```
solonaarmstrong:
```suggestion
     it('counts remaining characters in input field with maxLength', () => {
```

src/components/TextField/tests/TextField.test.tsx
```diff
@@ -305,6 +305,41 @@ describe('<TextField />', () => {
     });
   });
 
+  describe('characterCount', () => {
+    it('counts characters in input field', () => {
```
solonaarmstrong:
```suggestion
     it('displays number of characters entered in input field', () => {
```

src/components/TextField/README.md
```diff
@@ -737,6 +737,36 @@ Use to show that a textfield is not available for interaction. Most often used i
 <TextField label="Store name" disabled />
 ```
 
+### Text field with character count
+
+<!-- example-for: web -->
+
+Use to show the number of characters remaining in a text field with a max length. If max length is not provided it will display the current number of characters in the text field.
```
solonaarmstrong: Same comment as I've already left, but I find this a bit unclear.

src/components/TextField/TextField.scss
TheEnzoMar: This is needed because of https://github.com/Shopify/polaris-react/blob/master/src/components/TextField/TextField.scss#L140 which sets a `z-index`

TheEnzoMar:
Hi @solonaarmstrong and @danrosenthal 
Justin's won't be able to finish the PR as his internship ended, so I'm going to taking over it. I'm currently reviewing the code and your comments/feedback. Cheers!

TheEnzoMar:
Hi @ry5n, I tested what the text field w. character count looks like when other attributes are enabled. Here are the results.

image

Let me know what you think and/or any comments or concerns.

Cheers!

TheEnzoMar:
@solonaarmstrong, regarding
> Personally, I would do this a bit differently. I would either offer the consumer the option to count up or count down - with clear indications of what the number represents. Or I would be opinionated about counting up...

I see value in your comment and agree that a more consistent way of counting would be better. I think for simplicity we should count up, and when there is a max character limit the label should denote `x of y`.

What do you think of something like,

image

solonaarmstrong:
@TheEnzoMar Yes, that makes much more sense to me.

src/components/TextField/README.md
tmlayton:
```suggestion
Use to show the number of characters remaining in a text field with a max length. If max length is not provided it will display the current number of characters in the text field.
```

dpersing:
Hey @TheEnzoMar! Sounds like there's agreement on the counting method. Let us know if we can provide any additional help on this PR.

emarchak:
It ended up making more sense to provide a fixed line-height to the CharacterCount instead of shifting around the margins if it's a multiline text field. I believe we still need the `.AlignFieldBottom` since the layout had inconsistent behaviour when I applied `align-self: flex-end;` to a single line input field. 

image

I believe this should be good to go pending the two ✅ and someone approving percy. @solonaarmstrong and @dpersing could you please take another look?

dpersing:
Hi @emarchak! It looks like the `aria-label` for the "X of Y" content is reading out "X characters remaining", which is the opposite of the visual. For example, when 0 characters have been typed, it says "0 characters remaining", whereas really it should be "10 characters remaining" since none have been typed yet:

```
<div id="TextField1CharacterCounter" class="TextField-CharacterCount_7V_ks" aria-label="0 characters remaining" aria-live="polite" aria-atomic="true">0/10</div>
```

I'm thinking that "X of Y characters used" would make more sense with the visual?

emarchak:
Fixed!

src/locales/en.json
```diff
@@ -192,6 +192,11 @@
 
     "Tag": {
       "ariaLabel": "Remove {children}"
+    },
+
+    "TextField": {
+      "characterCount": "{count} characters",
+      "characterCountWithMaxLength": "{count} of {limit} characters used"
```
dpersing: 🎉

fchienvuhoang:
I'm using bellow code example but not work, my Polaris v3.8.0, any update?
```
 <TextField
        label="Store name"
        value={this.state.value}
        onChange={this.handleChange}
        maxLength={20}
        showCharacterCount
        multiline
      />
```

solonaarmstrong:
@fchienvuhoang I just tried and I see it in there. 

screen shot 2019-02-22 at 9 31 15 am

PR #712, opened 2018-12-04
[Avatar] Makes Avatar Image load only if Image load without error 

dleroux:

Resolves https://github.com/Shopify/shopify/issues/174824

The default option we want for our Avatar component is to show the  logged in user initials if an avatar image is not found. Gravatar gives us the possibility of providing a default image when an image is not found or to return a 404. 

In this [pr](https://github.com/Shopify/web/pull/5377) we attempted to return a transparent image. This was done with the following option in the query:

```
avatar(fallback: TRANSPARENT) {
      src
    }
```

The idea being that if the fallback was a transparent image it would simply lay over the Initials and wouldn't be visible.

However the Initials are in circle coloured background. As a result when an image was present it would show through [pr](https://github.com/Shopify/polaris-react-deprecated/pull/1157)

This PR is taking the is relying on gravatar returning a 404.

```
avatar(fallback: NOT_FOUND) {
      src
    }
```

If an image requested returns a 404 the onError attribute of the image tag will trigger and no image will be outputted.

In previous attempts server side rendering was causing the onLoad and onError of the image to not trigger. This PR adds the image only on the componentDidMount Lifecycle.

## 

### How to 🎩

<details>

```jsx
import {Page, Avatar, Image, Stack} from '@shopify/polaris';

interface State {}

  render() {
    return (
      <Page title="Playground">
        <Stack>
          <Avatar size="large" initials="DL" customer />
          <Avatar
            size="large"
            initials="DL"
            source="https://avatars2.githubusercontent.com/u/1229901?s=80&v=4"
          />
          <Avatar size="large" initials="DL" source="https://bla" />
        </Stack>
        <br />
        <Image
          alt="Just an image"
          source="https://avatars2.githubusercontent.com/u/1229901?s=80&v=4"
        />
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ The format is based on [these versioning and changelog guidelines](https://git.i
 
 ### Enhancements
 
+- Improved `Avatar` so it falls back to `initials` when the image fails to load ([#557](https://github.com/Shopify/polaris-react/pull/557))
 - Added `onScrolledToBottom` prop to `Scrollable` ([#568](https://github.com/Shopify/polaris-react/pull/568))
 
 ### Bug fixes
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -9,6 +9,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - `ResourceList.Item` opens url in new tab if command or control keys are pressed during click ([#690](https://github.com/Shopify/polaris-react/pull/690))
 - Added `primaryAction` prop to `SkeletonPage` ([#488](https://github.com/Shopify/polaris-react/pull/488))
 - Added support for press-and-hold to increment and decrement value in a type="number" `TextField` ([#573](https://github.com/Shopify/polaris-react/pull/573)) (thanks to [@andrewpye](https://github.com/andrewpye) for the [original issue](https://github.com/Shopify/polaris-react/issues/420))
+- Forced `Avatar` to fall back to `initials` when the image fails to load ([#712](https://github.com/Shopify/polaris-react/pull/712))
 
 ### Bug fixes
 
--- a/src/components/Avatar/Avatar.scss
+++ b/src/components/Avatar/Avatar.scss
@@ -19,6 +19,10 @@ $large-size: rem(60px);
   }
 }
 
+.hidden {
+  visibility: hidden;
+}
+
 .sizeSmall {
   width: $small-size;
 }
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -1,5 +1,7 @@
 import * as React from 'react';
+import {isServer} from '@shopify/react-utilities/target';
 import {classNames, variationName} from '@shopify/react-utilities/styles';
+import {autobind} from '@shopify/javascript-utilities/decorators';
 
 import {withAppProvider, WithAppProviderProps} from '../AppProvider';
 import Image from '../Image';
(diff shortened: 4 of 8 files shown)


dleroux:
Leaving this as a WIP until we can verify that this approach will work in production.

src/components/Avatar/Avatar.tsx
```diff
@@ -28,89 +29,138 @@ export interface Props {
...
+  }
+
+  componentDidUpdate({source: previousSource}: Props) {
+    const {source} = this.props;
+    const {hasError, hasLoaded} = this.state;
+    if (previousSource === source || (!hasError && !hasLoaded)) {
+      return;
+    }
+    // eslint-disable-next-line react/no-did-update-set-state
+    this.setState({hasError: false, hasLoaded: false});
```
danrosenthal: Couldn't this cause needless re-renders?

src/components/Image/tests/Image.test.tsx
```diff
@@ -20,5 +22,33 @@ describe('<Image />', () => {
...
+        <Image alt="alt text" source="404" onError={spy} />,
+      );
+
+      trigger(image.find('img'), 'onError');
+      expect(spy).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  describe('onLoad', () => {
+    it('calls the onLoad callback when the image on onLoad is triggered', () => {
```
danrosenthal:
```suggestion
    it('calls the onLoad callback when the image onLoad is triggered', () => {
```

src/components/Avatar/Avatar.tsx
danrosenthal: Also appears to be untested

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ The format is based on [these versioning and changelog guidelines](https://git.i
 
 ### Enhancements
 
+- Improved `Avatar` so it falls back to `initials` when the image fails to load ([#557](https://github.com/Shopify/polaris-react/pull/557))
```
danrosenthal: I don't think we want to add this to the `CHANGELOG` in this PR.

src/components/Avatar/Avatar.scss
```diff
@@ -19,6 +19,10 @@ $large-size: rem(60px);
   }
 }
 
+.hidden {
+  visibility: hidden;
```
danrosenthal: Don't we have a mixin for visually hiding an element?

src/components/Avatar/Avatar.tsx
dleroux: I believe if state doesn't change it doesn't re-render.

We don't test the state per se but we do test it renders what it should.

src/components/Avatar/Avatar.scss
dleroux: We do it does more than this and  I'm not sure it has value here because we're not wanting if to be accessible if the image is broken.

tmlayton:
A couple questions: 

- Is this still a WIP?  
- Can you isolate the commit which address why #557 had to be reverted?

src/components/Avatar/Avatar.tsx
```diff
@@ -28,89 +29,138 @@ export interface Props {
...
+    const className = classNames(
+      styles.Avatar,
+      styles[variationName('style', styleClass(nameString))],
+      size && styles[variationName('size', size)],
+      hasImage && !hasLoaded && styles.hidden,
+      hasImage && styles.hasImage,
+    );
+
+    const imageMarkUp =
+      finalSource && hasMounted && !hasError ? (
```
tmlayton: How about we remove this extra `hasMounted` state, use `!isServer` instead, and add a test that expects the image markup not render when `isServer` is true?

dleroux:
I’m okay either way. I prefer taking advantage of the lifecycle hooks, it feels more like a  React approach but I’ll give this a shot.

The other PR was reverted because it broke in prod because of an issue with how Avatar was being exported.

tmlayton:
> I’m okay either way. I prefer taking advantage of the lifecycle hooks, it feels more like a React approach but I’ll give this a shot.

I would disagree – just because we use a lifecycle hook doesn’t necessarily mean it is a better React approach. In this case, it adds complexity (extra state and lifecycle hook) and is harder to test. It also adds an extra render we can avoid:

`constructor` -> `render` -> `componentDidMount` -> `setState` -> `render` -> `componentDidUpdate`

image

CHANGELOG.md
danrosenthal: Bumping this

src/components/Avatar/Avatar.tsx
```diff
@@ -28,89 +30,136 @@ export interface Props {
...
+    if (props.source !== state.prevSource) {
+      return {
+        prevSource: props.source,
+        hasError: false,
+        hasLoaded: false,
+      };
+    }
+
+    return null;
   }
```
danrosenthal: This feels like the right approach over doing this in lifecycle hooks.

UNRELEASED.md
```diff
@@ -6,7 +6,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
-### Bug fixes
+- `Avatar` falls back to `initials` when the image fails to load ([#712](https://github.com/Shopify/polaris-react/pull/712))### Bug fixes
```
```suggestion
- Forced `Avatar` to fall back to `initials` when the image fails to load ([#712](https://github.com/Shopify/polaris-react/pull/712))

### Bug fixes
```

src/components/Avatar/Avatar.tsx
```diff
@@ -28,89 +30,136 @@ export interface Props {
...
 
-  if (source) {
-    finalSource = source;
-  } else if (customer) {
-    finalSource = customerPlaceholder(nameString);
+  state: State = {
+    hasError: false,
+    hasLoaded: false,
```
tmlayton:
```suggestion
    hasLoaded: false,
    prevSource: this.props.source,
```

I think we’ll want to initialize `prevSource` in state

src/components/Avatar/Avatar.tsx
```diff
@@ -28,89 +30,136 @@ export interface Props {
...
-  } else if (initials) {
-    const splitInitials = initials.split('').join(' ');
-    label = intl.translate('Polaris.Avatar.labelWithInitials', {
-      initials: splitInitials,
-    });
-  } else {
-    label = intl.translate('Polaris.Avatar.label');
+export class Avatar extends React.PureComponent<CombinedProps, State> {
+  static getDerivedStateFromProps(props: Props, state: State) {
+    // store prevSource in state so we can compare when props change
```
tmlayton:
```suggestion
```

I’m okay if you want to remove this comment, I was just adding it to give more context to the suggestion.

src/components/Avatar/tests/Avatar.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+  describe('on Error with changed props', () => {
+    it('re-renders the image if a the source prop is changed after an error', () => {
+      const src = 'image/path/';
+      const workingSrc = 'image/goodPath/';
+      const avatar = mountWithAppProvider(
+        <Avatar size="large" initials="DL" source={src} />,
+      );
+      trigger(avatar.find(Image), 'onError');
+      avatar.setProps({source: workingSrc});
+      avatar.update();
```
tmlayton:
```suggestion
```

`setProps` will trigger a re-render and the new state should be derived before, so this may not be needed.

src/components/Avatar/tests/Avatar.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+  });
+
+  describe('on Error with changed props', () => {
+    it('re-renders the image if a the source prop is changed after an error', () => {
+      const src = 'image/path/';
+      const workingSrc = 'image/goodPath/';
+      const avatar = mountWithAppProvider(
+        <Avatar size="large" initials="DL" source={src} />,
+      );
+      trigger(avatar.find(Image), 'onError');
```
tmlayton:
```suggestion
      trigger(avatar.find(Image), 'onError');
      expect(avatar.find(Image)).toHaveLength(0);
```

I feel like we should add this check that the image indeed does not render after an error then re-renders

tmlayton: Also is there a preference for `trigger` vs `simulate`?

src/components/Avatar/Avatar.tsx
danrosenthal: This feels unnecessary, since it's never used after this initialization, and getDerivedStateFromProps is always used before render.

CHANGELOG.md
danrosenthal: Never mind, I see that this was the result of un-reverting a previous commit.

src/components/Avatar/index.ts
```diff
@@ -1,4 +1,4 @@
 import Avatar from './Avatar';
 
-export * from './Avatar';
```
dleroux: @danrosenthal this should ensure we don't get the issue we experienced in web. There's no need to export * now that Avatar is a class component.

PR #724, opened 2018-12-05
[Layout] Adds oneHalf and oneThird props

darrenhebner:
These changes are introduced to allow us to properly achieve an even 2 or 3 column layout. This replaces the old recommendation of using two `secondary` sections. 

This PR adds `oneHalf` and `oneThird` props to `Layout.Section` that adds styles for even width columns. 

### What does it look like?
**Wide screen**
screen shot 2018-12-14 at 10 56 33 am

**Narrow screen**
screen shot 2018-12-14 at 11 05 27 am

### How to 🎩

Try viewing this layout at multiple viewport widths. Make sure that the columns do not become too narrow before wrapping.

<details>

```jsx
import {Page, Layout, Banner, Card, FooterHelp} from '@shopify/polaris';

interface State {}

  render() {
    return (
      <Page
        title="Playground"
      >
        <Layout>
          <Layout.Section fullWidth>
            <Card>
              <Banner title="Successfully created" status="success" />
            </Card>
          </Layout.Section>
          <Layout.Section oneHalf>
            <Card title="Half" sectioned>
              <p>Some content</p>
            </Card>
          </Layout.Section>
          <Layout.Section oneHalf>
            <Card title="Half" sectioned>
              <p>Some content</p>
            </Card>
          </Layout.Section>
          <Layout.Section oneThird>
            <Card title="Third" sectioned>
              <p>Some content</p>
            </Card>
          </Layout.Section>
          <Layout.Section oneThird>
            <Card title="Third" sectioned>
              <p>Some content</p>
            </Card>
          </Layout.Section>
          <Layout.Section oneThird>
            <Card title="Third" sectioned>
              <p>Some content</p>
            </Card>
          </Layout.Section>
          <Layout.Section fullWidth>
            <FooterHelp>Learn more</FooterHelp>
          </Layout.Section>
        </Layout>
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -7,6 +7,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Enhancements
 
 - Moved icons to a separate npm package ([#686](https://github.com/Shopify/polaris-react/pull/686))
+- Added `oneHalf` and `oneThird` props to `Layout` component ([#724](https://github.com/Shopify/polaris-react/pull/724))
 
 ### Bug fixes
 
--- a/src/components/Layout/Layout.scss
+++ b/src/components/Layout/Layout.scss
@@ -1,5 +1,7 @@
 $secondary-basis: layout-width(secondary, min);
 $primary-basis: layout-width(primary, min);
+$one-half-basis: layout-width(one-half-width, base);
+$one-third-basis: layout-width(one-third-width, base);
 $relative-size: $primary-basis / $secondary-basis;
 
 .Layout {
@@ -23,6 +25,16 @@ $relative-size: $primary-basis / $secondary-basis;
 
 .Section-fullWidth {
   flex: 1 1 100%;
+}
+
+.Section-oneHalf {
+  flex: 1 1 $one-half-basis;
+  min-width: 0;
+}
+
+.Section-oneThird {
+  flex: 1 1 $one-third-basis;
+  min-width: 0;
 }
 
 .AnnotatedSection {
--- a/src/styles/foundation/layout.scss
+++ b/src/styles/foundation/layout.scss
@@ -14,6 +14,12 @@ $layout-width-data: (
     min: rem(240px),
     max: rem(320px),
   ),
+  one-half-width: (
+    base: rem(450px),
+  ),
+  one-third-width: (
+    base: rem(240px),
+  ),
   nav: (
     base: rem($navigation-width),
   ),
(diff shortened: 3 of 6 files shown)


UNRELEASED.md
```diff
@@ -6,6 +6,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Adds `evenWidth` prop to `Layout`
```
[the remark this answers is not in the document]
darrenhebner: Oh great, I wasn't totally sure what to put here.

darrenhebner:
I've updated the README examples to use the new `evenWidth` prop. I think the preview window in the Polaris docs might be too narrow to show the two columns side by side. The code example still demonstrates how to use it though.

dfmcphee:
> * It is only possible to do this 50/50 layout if you add `fullWidth` to the `Page` component since there is not enough room otherwise.

Do you think we should be using something in-between the primary and secondary flex-basis? Would that two evenWidth columns to show side by side in a default page?

darrenhebner:
@dfmcphee I think it could be possible to reduce the flex basis slightly so they fit side by side in a regular width page. With this current implementation you can not have side by side columns in with a regular width page. 

I could use a designers eye for figuring out how narrow we would want to go. For our use case we wouldn't want to go too narrow.

I'll figure out how narrow we would need to go in order for it to work with a regular width page.

darrenhebner:
I've updated the branch to use a slightly smaller flex-basis. This would allow two columns to be side by side on a regular width page.

I don't think it looks too small. Here's what it would look just before the columns wrap.

screen shot 2018-12-12 at 1 58 15 pm

src/components/Layout/Layout.scss
```diff
@@ -1,5 +1,6 @@
 $secondary-basis: layout-width(secondary, min);
 $primary-basis: layout-width(primary, min);
+$even-width-basis: layout-width(page-content, partially-condensed);
```
darrenhebner: This is the exact width that I want, am I using this function correctly though? I'm not sure what `page-content` and `partially-condensed` represent.

dfmcphee: Ah, yeah. That is the width for the breakpoint where we start to condense the padding on the page. I would add a new value to the map for `equal-width` here (even if it is the same value): https://github.com/Shopify/polaris-react/blob/master/src/styles/foundation/layout.scss

src/styles/foundation/layout.scss
```diff
@@ -14,6 +14,9 @@ $layout-width-data: (
     min: rem(240px),
     max: rem(320px),
   ),
+  equal-width: (
```
dfmcphee: Small nitpick, but this should probably also be `even-width` to match the prop name.

PR #733, opened 2018-12-10
[Playground] Add Google Analytics tracking back

kaelig:
Tracking used to be present in the `yarn tophat` developer tool, but was recently removed (probably when /tophat and /playground got merged).

- Adds tracking back to the playground (and /examples)
- Puts `utf-8` in lowercase, not sure it has an impact but it seems to be the norm

### DIFF
--- a/playground/index.html
+++ b/playground/index.html
@@ -1,12 +1,21 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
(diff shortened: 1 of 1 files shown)


kaelig:
The script is loaded correctly:

screenshot 2018-12-10 15 39 00

PR #734, opened 2018-12-11
Export ContextualSaveBarProps from src/components/index

BPScott:
266f1297 accidentally removed this import as part of refactoring

---

#698 Stopped exporting ToastProps and ContextualSaveBarProps. This was my bad as I didn't understand enough about how and where these props are used and didn't realize they are part of our public API. 

#722 Brought back ToastProps. This finishes the job.


BPScott:
Pinging @kvendrik and @tmlayton so they're aware of this.

Issue #684, opened 2018-11-27
Host component documentation images in this repository


kaelig:
# Issue summary

Images referenced in `src/components/**/README.md` files are hosted in a separate repository.

This leads to broken images on GitHub and confusing workflows:

![screen shot 2018-11-27 at 11 42 16 am](https://user-images.githubusercontent.com/85783/49107033-8b5d3800-f239-11e8-9299-53927239729f.png)

---

Some context: I've historically been against this practice because I thought it'd lead to a large increase in `@shopify/polaris-react` package size and have a performance impact on dependency installs, build times, etc. Turns out, it would amount to ~3.3MB. This shouldn't have any measurable adverse impact. Apologies for being too conservative in the past… 😔

PR #753, opened 2018-12-13
Change paths to images to match the style guide's folder structure

kaelig:
Related to https://github.com/Shopify/polaris-styleguide/pull/2446

This is an interim solution, allowing us to simplify image path parsing code on polaris-styleguide until we figure out a path forward for #684.


BPScott:
I haven't tested this in conjunction with the styleguide changes. In light of the our recent version churn due little bugs minor bumps, I think it might be worth doing a beta version to prove it works before releasing a stable version

PR #755, opened 2018-12-13
[Exploration] making Polaris fully theme-able

kvendrik:
## Why are these changes being introduced?
Exploration for https://github.com/Shopify/polaris-react/issues/744.

## What is this PR doing?
It adds the necessary infrastructure to scale our theming efforts. The goal is to provide third parties with a way to theme Polaris to fit their branding. 

### Options
During the exploration I explored two different approaches:

1. https://github.com/Shopify/polaris-react/commit/2ceb348254bdda2369c7eb32a6ac6b6d6046648d allows for theming based on components. Its very flexible and allows for basically anything to be themed (as long as we enable it).
2. The latest version of this PR contains a different approach that implements a way to theme Polaris by theming certain global variables. Its less flexible but faster to use.

If we want to go with option 1 or option 2 really depends on our goal.

#### Option 1: Component specific styling
In a scenario where we expect third parties to use Polaris for their own products with their own carefully thought through branding flexibility is a must (yeey option 1).

#### Option 2: Global styling
If we simply want to provide a quick and easy way to theme Polaris with the goal being to enable devs to legally use it in their commercial products (our policy states that you can't create a product that looks like the admin so theme-ability is a must) option 2 does a better job. Especially because with the lesser amount of theme-able keys we could even auto generate missing keys based on certain given base values ([like we currently do for the `TopBar`](https://github.com/Shopify/polaris-react/blob/master/src/components/ThemeProvider/utils/index.ts#L79)).

What is noteworthy is that also with this option devs will still have some flexibility thanks to the support for overriding theme keys. Have a look at the example below where the `AppProvider` sets a `yellow` background colour for all layered components (e.g. `Card`s). The actual `Card` is then wrapped in a second `ThemeProvider` which sets the background colour to `indigo`. Because we support theme overrides the background colour of the card will be `indigo`.

```tsx
<AppProvider
  theme={{
    styles: {layeredBackground: 'yellow'},
  }}
>
  <ThemeProvider
    theme={{
      styles: {layeredBackground: 'indigo'},
    }}
  >
    <Card />
  </ThemeProvider>
</AppProvider>
```

## Playing around with the changes
This PR contains a playground with everything you need to play around with option 2. To play around with option 1 simply check out commit https://github.com/Shopify/polaris-react/commit/2ceb348254bdda2369c7eb32a6ac6b6d6046648d and start the playground from there.

screen shot 2018-12-13 at 3 56 21 pm

## Challenges
I have noticed that one of the main challenges with making Polaris theme-able is that it challenges our current infrastructure. The main issue is that we `darken` and `lighten` colours throughout our codebase with SCSS functionality that is only available when you use static colour variables. When you introduce theme-ability to that it breaks as it would have to count on variables that are unknown until runtime.

**⚠️This PR is an exploration, contains breaking changes and should not be merged in its current state.**

cc @dfmcphee @danrosenthal


danrosenthal:
Hey bud, gonna go ahead and close these as they're an exploration (this and #763), but feel free to reopen if you want to merge or want to discuss.

PR #759, opened 2018-12-13
Fix cyclical imports redux (part2)

BPScott:
Continuing the cyclical type import dance (almost there)

#698 Did some refactoring to remove some cyclical type imports. It turns out those changes broke the Props Explorer in the styleguide, as the script in the styleguide that generates props data assumes that each component has an interface called `Props` in it and #698 moved some props declarations outside of the component to remove circularity - as the Props were shared by two components.

#727 reverted the problematic parts of that PR and made the styleguide work again, at the cost of bringing back those cyclical type imports.

This PR fixes the cyclical types in a way that does not break the props explorer.

This PR consists of two commits. The first reverts #727. The second applies a change that allows the props explorer to be generated correctly while retaining no circular imports. It does this by adding  a `Props` interface to AppProvider, Toast and ContextualSaveBar that extends from the actual Props interface we want to use.

Essentially changing the following code that causes problems with the props explorer:

```ts
import { ContextualSaveBarProps } from '../Frame';

class ContextualSaveBar extends React.PureComponent<ContextualSaveBarProps, never>  {}
```

to the following, which the props explorer works with:

```ts
import { ContextualSaveBarProps } from '../Frame';

interface Props extends ContextualSaveBarProps {}

class ContextualSaveBar extends React.PureComponent<Props, never>  {}
```

I tried doing this with a type alias `type Props = ContextualSaveBarProps;` but it didn't work so we have to create a new interface.

### How to 🎩

* Run tests
* Run `yarn run sk type-check` to prove typescript is happy
* Run `yarn build-consumer web` then run `yarn run sk type-check` in web to prove there are no missing exports (You may have to install polaris-icons in web as a workaround because build-consumer isn't smart enough to handle updating transitive dependencies)
* Run `yarn build-consumer polaris-styleguide` then run `yarn run dev` in polaris-styleguide and check the props explorer for the AppProvider, Toast and ContextualSaveBar components still render the same (You may have to install polaris-icons in web as a workaround because build-consumer isn't smart enough to handle updating transitive dependencies)

### DIFF
--- a/src/components/AppProvider/AppProvider.tsx
+++ b/src/components/AppProvider/AppProvider.tsx
@@ -12,14 +12,20 @@ import {
   polarisAppProviderContextTypes,
 } from './types';
 
-export default class AppProvider extends React.Component<AppProviderProps> {
+// The script in the styleguide that generates the Props Explorer data expects
+// a component's props to be found in the Props interface. This silly workaround
+// ensures that the Props Explorer table is generated correctly, instead of
+// crashing if we write `AppProvider extends React.Component<AppProviderProps>`
+interface Props extends AppProviderProps {}
+
+export default class AppProvider extends React.Component<Props> {
   static childContextTypes = polarisAppProviderContextTypes;
   public polarisContext: Context;
   private stickyManager: StickyManager;
   private scrollLockManager: ScrollLockManager;
   private subscriptions: {(): void}[] = [];
 
-  constructor(props: AppProviderProps) {
+  constructor(props: Props) {
     super(props);
     this.stickyManager = new StickyManager();
     this.scrollLockManager = new ScrollLockManager();
@@ -46,7 +52,7 @@ export default class AppProvider extends React.Component<AppProviderProps> {
     apiKey,
     shopOrigin,
     forceRedirect,
-  }: AppProviderProps) {
+  }: Props) {
     if (
       i18n !== this.props.i18n ||
       linkComponent !== this.props.linkComponent ||
--- a/src/components/AppProvider/AppProvider.tsx
+++ b/src/components/AppProvider/AppProvider.tsx
@@ -1,42 +1,25 @@
 import * as React from 'react';
 import {autobind} from '@shopify/javascript-utilities/decorators';
-import ThemeProvider, {Theme} from '../ThemeProvider';
-import {LinkLikeComponent} from '../UnstyledLink';
+import ThemeProvider from '../ThemeProvider';
 import {
   StickyManager,
   ScrollLockManager,
-  TranslationDictionary,
   createAppProviderContext,
 } from './utilities';
-import {Context, polarisAppProviderContextTypes} from './types';
+import {
+  AppProviderProps,
+  Context,
+  polarisAppProviderContextTypes,
+} from './types';
 
-export interface Props {
-  /** A locale object or array of locale objects that overrides default translations */
-  i18n?: TranslationDictionary | TranslationDictionary[];
-  /** A custom component to use for all links used by Polaris components */
-  linkComponent?: LinkLikeComponent;
-  /** The API key for your application from the Partner dashboard */
-  apiKey?: string;
-  /**
-   * The current shop’s origin, provided in the session from the Shopify API (to be provided without the https://)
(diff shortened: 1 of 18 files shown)


src/components/AppProvider/AppProvider.tsx
```diff
@@ -12,7 +12,13 @@ import {
   polarisAppProviderContextTypes,
 } from './types';
 
-export default class AppProvider extends React.Component<AppProviderProps> {
+// The script in the styleguide that generates the Props Explorer data expects
+// a component's props to be found in the Props interface. This silly workaround
```
AndrewMusgrave: I could be wrong but I recall it working on types as well e.g. `type Props = AppProviderProps`

BPScott: I tried that first and it didn't work :(

AndrewMusgrave: Too bad :/

src/components/AppProvider/AppProvider.tsx
```diff
@@ -36,7 +25,7 @@ export default class AppProvider extends React.Component<Props> {
   private scrollLockManager: ScrollLockManager;
   private subscriptions: {(): void}[] = [];
 
-  constructor(props: Props) {
+  constructor(props: AppProviderProps) {
```
AndrewMusgrave: Should we keep this `Props` to stay consistent with the generic on `L21`
```suggestion
  constructor(props: Props) {
```

src/components/AppProvider/AppProvider.tsx
```diff
@@ -63,7 +52,7 @@ export default class AppProvider extends React.Component<Props> {
     apiKey,
     shopOrigin,
     forceRedirect,
-  }: Props) {
+  }: AppProviderProps) {
```
AndrewMusgrave: ^
```suggestion
  }: Props) {
```

src/components/AppProvider/types.ts
```diff
@@ -1,8 +1,34 @@
...
+export interface AppProviderProps {
+  /** A locale object or array of locale objects that overrides default translations */
+  i18n?: TranslationDictionary | TranslationDictionary[];
+  /** A custom component to use for all links used by Polaris components */
+  linkComponent?: LinkLikeComponent;
+  /** The API key for your application from the Partner dashboard */
+  apiKey?: string;
+  /**
+   * The current shop’s origin, provided in the session from the Shopify API (to be provided without the https://)
+   * @default getShopOrigin()
```
AndrewMusgrave: Unless I'm mistaken I don't think our parser can handle this?

src/components/index.ts
```diff
@@ -138,6 +135,8 @@ export {default as FormLayout, Props as FormLayoutProps} from './FormLayout';
 export {
   default as Frame,
   Props as FrameProps,
+  ContextualSaveBarProps,
+  ToastProps,
```
AndrewMusgrave: 🎉

solonaarmstrong:
🎉

src/components/Toast/Toast.tsx
```diff
@@ -6,12 +6,18 @@ import {
...
   FrameContext,
   frameContextTypes,
-  ToastProps as Props,
+  ToastProps,
 } from '../Frame';
 import {withAppProvider, WithAppProviderProps} from '../AppProvider';
 
 const createId = createUniqueIDFactory('Toast');
 
+// The script in the styleguide that generates the Props Explorer data expects
```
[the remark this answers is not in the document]
BPScott: My big motivation is make future-Ben think twice when he sees this in 6 months time, because I know he'll think "that looks pointless, I should tidy that up to be "ComposedProps = ToastProps & WithAppProviderProps" and then everything will break

src/components/AppProvider/types.ts
BPScott: I believe you're right, the default stuff can only handle strings ATM, but this is how it was before. I've only moved the definitions around, I haven't modified them so I'm going to claim fixing this up is out of scope for this PR :)

BPScott:
Have rebased to add Andrew's points about consistently using Props within AppProvider.

PR #763, opened 2018-12-14
[Exploration] Styled components

kvendrik:
## What is this?
While working on https://github.com/Shopify/polaris-react/pull/755 it became clear that making Polaris fully theme-able poses some challenges. One of the biggest ones being that we `lighten` and `darken` a lot of our colours in our SCSS which is complicated to do with values that are being passed into the `ThemeProvider` as those values are not known until runtime.

A potential workaround would be to generate the problematic CSS rules in JS and pass them into the component at runtime but if a lot of components are going to need that you might start to wonder if using SCSS is even still in our advantage.

Hence this exploration which recreates one of our components using JS styles, which would give us a lot more flexibility. The cons of course being that no other codebase at Shopify currently does this and that porting everything over would be a lot of work. We could also partially use JS styling to work around some of the issues we're having.

No matter what way we think this might be useful in I wanted to share the exploration for future reference and to hear your thoughts on if and how this might be useful in the future.

## Playing around with the changes
This exploration contains a playground with the JS styled button.

**⚠️This PR is an exploration, contains breaking changes and should not be merged in its current state.**

cc @dfmcphee @kaelig (I know you explored this at some point 🙂)

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,6 +95,7 @@
     "@types/react-hot-loader": "^3.0.6",
     "@types/react-router-dom": "^4.2.7",
     "@types/rollup": "0.51.4",
+    "@types/styled-components": "^4.1.4",
     "archiver": "^2.1.0",
     "aws-sdk": "^2.58.0",
     "babel-cli": "^6.26.0",
@@ -188,6 +189,7 @@
     "lodash-decorators": "^4.3.5",
     "prop-types": "^15.6.1",
     "react-transition-group": "^2.4.0",
+    "styled-components": "^4.1.2",
     "tslib": "^1.9.3"
   }
 }
--- a/playground/Playground.tsx
+++ b/playground/Playground.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import {Page} from '@shopify/polaris';
+import {Page, Button} from '@shopify/polaris';
 
 interface State {}
 
@@ -10,7 +10,7 @@ export default class Playground extends React.Component<{}, State> {
         title="Playground"
         primaryAction={{content: 'View Examples', url: '/examples'}}
       >
-        {/* Add the code you want to test here */}
+        <Button>Click me!</Button>
       </Page>
     );
   }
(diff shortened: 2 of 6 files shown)


AndrewMusgrave:
My two-cents

> While working on #755 it became clear that making Polaris fully theme-able poses some challenges. One of the biggest ones being that we lighten and darken a lot of our colours in our SCSS which is complicated to do with values that are being passed into the ThemeProvider as those values are not known until runtime.

We won't have to worry about this much longer. I built a library that'll make manipulating easy. We'll also have more options than what we currently have in sass and not have to worry about the color type. It won't be as complicated as it is now. I haven't had time recently to finish publishing it and haven't pushed for it recently as I thought theming was on hold till further exploration was done.

 > A potential workaround would be to generate the problematic CSS rules in JS and pass them into the component at runtime but if a lot of components are going to need that you might start to wonder if using SCSS is even still in our advantage.Hence this exploration which recreates one of our components using JS styles, which would give us a lot more flexibility. The cons of course being that no other codebase at Shopify currently does this and that porting everything over would be a lot of work. We could also partially use JS styling to work around some of the issues we're having.

I'm not sure if this transition will be worth the effort. It'll create a lot of work migrating from scss for not much benefit other than an alternative way to theme. I'm not sure if `styled-components` have improved there performance recently but AFAIK inline-styles still beat them.

We would be adding an extra dependency not managed by shopify. Granted `styled-components` are widely used, we would still be putting our trust in an outside dependency and owning all the bugs and vulnerabilities that come along with it.

Styled-components don't align currently align with the common tech stack at shopify. That doesn't mean it can't change, or we can't use them. But all our tooling, build's, etc have been built with sass in mind. We also "export" our sass functions so that consumers are able to use them, for example web. If we moved away from sass what would this mean for them? Would they lose our utilities, would we have to port them over creating more debt? If we created a utility for styled-components would we need to create it in web as well?

kaelig:
Thanks for sharing this exploration! Very interesting to see how design tokens can be leveraged in JS.

danrosenthal:
Hey bud, gonna go ahead and close these as they're an exploration (this and #755), but feel free to reopen if you want to merge or want to discuss.

PR #764, opened 2018-12-14
[Documentation] Add syntax highlighting to console logging guidelines

kaelig:
@tmlayton as mentioned earlier, this is where we have _some_ docs around deprecation.

This PR just makes it look nicer.


BPScott:
Oh, because now you've added the language hint prettier knows it is the blocks are javascript and can format them to its expectations

PR #721, opened 2018-12-04
Revert "Fix several cyclical type imports"

kvendrik:
### Description
Reverts Shopify/polaris-react#691 as it breaks color exports.

### How to 🎩 
1. On this branch:`yarn build-consumer web`.
2. In Web: `yarn run type-check`.
3. Make sure you get no more errors related to missing color types and helpers.

Issue #718, opened 2018-12-04
Color transformation functions (hsbToHex, etc.) not exported anymore


mbaumbach:
# Issue summary

Importing `hsbToHex` and other color transformation functions is no longer available at the top-level `@shopify/polaris` package. These are often useful when working with the HSBA-based `ColorPicker` component. It appears to have been removed in commit f0e56546bc37d75f3a70b2c7c63d5e6fb695abe7.

## Expected behavior

The following import used to work:

```
import { hsbToHex } from '@shopify/polaris`;
```

## Actual behavior

As of Polaris 3.2.0, this throws an error saying that `hsbToHex` is not exported from `@shopify/polaris`.

## Steps to reproduce the problem

1. Attempt to import any of the color transformation functions.

## Specifications

- Are you using the React components? (Y/N): Y
- Polaris version number: 3.2.0
- Browser: Google Chrome 70.0.3538.110
- Device: MacBook Pro
- Operating System: macOS Mojave 10.14.1

mbaumbach:
I'm not sure if this is considered a breaking change or was never meant to be public API and we should be providing/acquiring our own transformation functions. If it's not meant to be public API, feel free to close. :)

danrosenthal:
Thanks @mbaumbach, and sorry for any trouble. This is indeed fixed in `3.2.1`

PR #767, opened 2018-12-15
Fix cyclical type imports relating to color-picker helpers

BPScott:

The final cyclical import is dead (again). We can now enable the import/no-cycle eslint rule (again).

#691 removed some cyclical imports. It also stopped exporting a few color helper functions in the final build (see #718) as I incorrectly assumed that they should be internal. It turns out people are using them, so that PR got reverted in #721.

I reapplied the not-the-color-helper parts of #691 in #754. This PR is the follow up that removes the cyclical dependency in color helpers, without touching any exports.

It moves the color types into the utility folder next to the color utility functions

### How to 🎩

* Run tests
* Run `yarn run sk type-check` to prove typescript is happy
* Note that while an import path is changed the contents of that file remain identical.
* Run `yarn build-consumer web` then run `yarn run sk type-check` in web to prove there are no missing exports (You may have to install polaris-icons in web as a workaround because build-consumer isn't smart enough to handle updating transitive dependencies) - web consumes the HSBColor type in some places so this will prove that the types are still exported.


UNRELEASED.md
```diff
@@ -19,6 +19,6 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Development workflow
 
 - Added a slight delay to the Percy screenshot script to give time for components to render fully ([#704](https://github.com/Shopify/polaris-react/pull/704))
-- Refactors to remove cyclical type imports ([#759](https://github.com/Shopify/polaris-react/pull/759) and [#754](https://github.com/Shopify/polaris-react/pull/754))
+- Refactors to remove cyclical type imports ([#759](https://github.com/Shopify/polaris-react/pull/759), [#754](https://github.com/Shopify/polaris-react/pull/754) and [#767](https://github.com/Shopify/polaris-react/pull/767))
```
```suggestion
- Refactors to remove cyclical type imports ([#759](https://github.com/Shopify/polaris-react/pull/759), [#754](https://github.com/Shopify/polaris-react/pull/754), and [#767](https://github.com/Shopify/polaris-react/pull/767))
```

Oxford comma

BPScott: The strippers, JFK, and Stalin appreciate your attention to detail.

(I've rebased to include this)

PR #773, opened 2018-12-17
Ensure Buttons with a url prop output valid HTML

### LINKED ISSUE #757: [Button] `disabled` attribute is not allowed on element `a`
# Issue summary

A `disabled` button renders invalid markup when passed a URL prop:

```jsx
<Button disabled={true} url="https://foo.com">Add product</Button>
```

## Expected behavior

HTML is valid, without a `disabled` attribute on the anchor element:

```html
<a class="Polaris-Button Polaris-Button--disabled" href="https://foo.com" data-polaris-unstyled="true"><span class="Polaris-Button__Content"><span>Add product</span></span></a>
```

## Actual behavior

HTML is invalid (`Error: Attribute disabled not allowed on element a at this point`):

```html
<a class="Polaris-Button Polaris-Button--disabled" disabled="" href="https://foo.com" data-polaris-unstyled="true"><span class="Polaris-Button__Content"><span>Add product</span></span></a>
```

1. Go to https://codesandbox.io/s/nrz4zwlnqj?module=App.js
1. Look at the rendered markup
1. Validate markup against https://validator.w3.org/nu/#textarea

## Specifications

- Polaris version number: 3.3.0

BPScott:

Ensure Buttons with a url prop output valid HTML

Fixes #757.

When a button has a url it is rendered as an `<a>` tag, which does not
have a `disabled` attribute.

Stop adding the `disabled` attribute in that case

### How to 🎩

<details>

```jsx
import {Button, Page} from '@shopify/polaris';

interface State {}

  render() {
    return (
      <Page
        title="Playground"
      >
        <h2>Normal</h2>
        <Button>ONE</Button>
        <Button url="/foo">TWO</Button>
        <h2>Disabled</h2>
        <Button disabled>ONE</Button>
        <Button disabled url="/foo">
          TWO
        </Button>
      </Page>
    );
  }
}
```

</details>

Ensure that the `<Button disabled url="/foo">TWO</Button>` element does not render a `disabled` attribute in the generated HTML markup.


BPScott:
> Thinking about this more, I think it would make sense to convert the "disabled" link to a non-interactive link by removing the `href`. That is, if someone produces a "disabled" link, it shouldn't have a valid `href` so it can't be accessed. This renders the `<a>` but makes it non-interactive.

I think this is the preferable solution too. This will require a wee refactor as (sensibly) you can't omit the url property when calling `<UnstyledLink>` so we'd have to output a `<span>` or an `<a>` without a href within `Button` rather than deferring to `UnstyledLink` but I think that's fine.

kaelig:
Should we still merge this?

BPScott:
Dragging it up to date now

BPScott:
Daft realisation: why bother rendering span or a's without hrefs - just render `<button disabled>` instead!

src/components/Button/Button.tsx
```diff
@@ -157,23 +157,36 @@ function Button({
...
-      {content}
-    </UnstyledLink>
-  ) : (
+  if (url) {
+    return isDisabled ? (
+      // <a disabled> is not valid HTML so render an <a> without a href attribute.
+      // This results in a minimal DOM change when the Button moves between
+      // enabled and disabled states.
+      // eslint-disable-next-line jsx-a11y/anchor-is-valid
+      <a id={id} className={className} aria-label={accessibilityLabel}>
```
kaelig: Would this still appear as a navigable anchor in the accessibility tree (just like a good old `<a name="foo"></a>`)?

src/components/Button/Button.tsx
```diff
@@ -157,23 +157,36 @@ function Button({
...
-      disabled={isDisabled}
-      aria-label={accessibilityLabel}
-    >
-      {indicatorMarkup}
-      {content}
-    </UnstyledLink>
-  ) : (
+  if (url) {
+    return isDisabled ? (
+      // <a disabled> is not valid HTML so render an <a> without a href attribute.
```
dleroux: I don't think we need to add these comments.

src/components/Button/Button.tsx
```diff
@@ -157,23 +157,36 @@ function Button({
...
+        className={className}
+        aria-label={accessibilityLabel}
+      >
+        {indicatorMarkup}
+        {content}
+      </UnstyledLink>
+    );
+  }
+
+  // UnstyledLink cant be disabled so always render a button when disabled
```
dleroux: Same with this one. The code is clear enough imo.

BPScott:
Whoops, I thought I merged this last week!

PR #774, opened 2018-12-17
[Tooltip] Updated example to initially be active

chloerice:

Along with https://github.com/Shopify/polaris-styleguide/pull/2463, this fixes the tooltip's markup not being captured and displayed with the component example markup.

Sets the tooltip to start out as `active`. This initial render of the example with the tooltip showing allows for the markup to be correctly captured and displayed with the rest of the example's markup.

### How to 🎩

- `git checkout tooltip-example-always-active && yarn dev
- Open `http://localhost:8080/tooltip/default-tooltip`
    - The example should render with the tooltip initially active (rendered)
    - The hover responses should still work as expected (unmounts on the mouse leaving it or the activator link, mounts on mouse entering the activator link)

### 🎩 checklist

* Attempted to test using `yarn run build-consumer polaris-styleguide`, but it results in an odd, unrelated server error:

```sh
[server] Error in server execution, check the console for more info.
/Users/chloe/Projects/polaris-styleguide/build/server/main.js:8315
		var e = new Error("Cannot find module '" + req + "'");
          ^
Error: Cannot find module './/public_images/components/Avatar/android/default@2x@2x.png'
```

### DIFF
--- a/src/components/Tooltip/README.md
+++ b/src/components/Tooltip/README.md
@@ -68,7 +68,7 @@ Use only when necessary to provide an explanation for an interface element.
 
 ```jsx
 <div style={{padding: '75px 0'}}>
-  <Tooltip content="This order has shipping labels.">
+  <Tooltip active content="This order has shipping labels.">
     <Link>Order #1001</Link>
   </Tooltip>
 </div>


chloerice:
> I wonder if we should also keep the old one to have one that still works with the hover?

It still works with hover, it just starts out as active (so if you hover over it and then stop hovering over it it still has the same behavior). But yeah, if we want to avoid inevitable confusion maybe it's best to have an example that's explicitly for an initially active tooltip?

dfmcphee:
> It still works with hover, it just starts out as active (so if you hover over it and then stop hovering over it it still has the same behavior). But yeah, if we want to avoid inevitable confusion maybe it's best to have an example that's explicitly for an initially active tooltip?

Ah, ok. I didn't realize that. I would say this is fine as is then. 👍

PR #775, opened 2018-12-17
Revert "Move from local icons to polaris-icons (#686)"

AndrewMusgrave:
This reverts commit d847fc68bba7f54a3560b450c5d5a237fd5d609a.

Polaris icon's cannot be resolved

screen shot 2018-12-17 at 4 57 31 pm

Reverting #686

### How to 🎩
 Build consumer to web or polaris styleguide


AndrewMusgrave:
cc/ @amrocha @kaelig

kaelig:
Please don't merge. The error is due to a limitation in `build-consumer`, which doesn't install transitive dependencies.

BPScott:
[Transitive dependencies](https://en.wikipedia.org/wiki/Transitive_dependency) are dependencies of dependencies. In this case polaris-styleguide does not directly depend upon polaris-icons, but on polaris-react, which in turn depends upon polaris-icons.

polaris-icons is a new dependency of polaris-react that only exists on master.

Running `yarn build consumer polaris-styleguide` does not install any of its dependencies of polaris-react into polaris-styleguide. Most of the time this works by fluke as we don't change dependencies that often (and styleguide will use dependencies of what it has in its package.json). However in this case polaris-styleguide can't find the polaris-icons because it has not been installed.

This will fix itself once we release a new version of polaris-react and update polaris-styleguide to use that version.

AndrewMusgrave:
> Running yarn build consumer polaris-styleguide does not install any of polaris-react's dependencies into polaris-styleguide. Most of the time this works by fluke as we don't change dependencies that often (and styleguide will use dependencies of what it has in its package.json). However in this case polaris-styleguide can't find the polaris-icons package because it has not been installed - it isn't there in polaris 3.3.0. This will fix itself once we release a new version of polaris-react and update polaris-styleguide to use that version.

Thanks, TIL. That's unfortunate, `build-consumer` may need to be improved since it's heavily used for testing on web, where we should be checking our changes and builds before release. If this is fixed with a release, is there a reason why we haven't done one already?

dleroux:
But technically we should't release a new version of Polaris unless we are able to build-consumer and properly tophat our release?

Issue #595, opened 2018-11-09
[RangeSlider] Add dual input


solonaarmstrong:
# Feature request summary

Add a second input (button) to the existing range slider component to create a range of values instead of only one value.

![screen shot 2018-11-09 at 3 20 57 pm](https://user-images.githubusercontent.com/6239480/48293281-2b8e2100-e433-11e8-90be-5f28d297b456.jpg)

To make this accessible, the two input values will each need to be accompanied by text (number) inputs that can control the values.

## Rationale

This feature is needed for some list filters. For example:
- To select customers with a range of money spent
- To select customers with a range of number of orders placed

AndrewMusgrave:
Rather than an addition prop, have we considered value accepting number or tuple? For example
```tsx
interface Props {
  ...
  value: number | [number, number]
  ...
}
```

solonaarmstrong:
@AndrewMusgrave When I was exploring this, I created a `Range` interface, similar to `DatePicker`, but I think your suggestion is better. 👍

solonaarmstrong:
@tmlayton This is a good starting point for a technical design.

dpersing:
A few notes for a11y that we've discussed elsewhere but thought I'd include here for technical design discussion:

- There is an [ARIA 1.1 pattern for a multi-thumb range slider](https://www.w3.org/TR/wai-aria-practices/#slidertwothumb), but there's no equivalent in HTML, so it relies entirely on ARIA and JavaScript for semantics and functionality. This should be fine for keyboard users, but might get hairy for desktop screen reader users.
- ARIA-only range sliders, including single-thumb ones, aren't supported well by mobile screen readers. VoiceOver and TalkBack really only seem to like sliders that build off the native `input="range"` and choke on spec-perfect single-thumb ARIA-only sliders.
- There's no native touch gesture for adjusting a multi-thumb slider, since the screen reader cursor focus goes to the whole slider and the user swipes up and down to move the handle/thumb. The ARIA multi-thumb pattern doesn't currently include any touch/mobile info.

A good progressive enhancement solution for setting any range value is a labeled text input or set of inputs for a multi-thumb slider. Letting users enter values in the text inputs or via the slider and having the values update either way for both sets of controls would be ideal.

solonaarmstrong:
CC @tmlayton

PR #784, opened 2018-12-19
[Range Slider] Add second thumb

solonaarmstrong:

Resolves [#595](https://github.com/Shopify/polaris-react/issues/595). The need for a dual thumb range slider arose with the redesign of list filters. Some filters are ranges and we thought this type of input might be a good solution.

This is an example of how it might be used:

moneyspent

Adds a second thumb to range slider, using manually constructed divs, rather than the native range element.

Technical spec: https://docs.google.com/document/d/1IIssOR8bwoARk4XtXNPmUK32008c19xKcW89a1-_8oo/edit#

### To do

* [x] Button size, border, hover, focus, and active states still aren't consistent with thumbs between dual and single - SA: https://github.com/Shopify/polaris-react/pull/1034
* [x] Keep the treatment the same for the thumbs. So keep the box shadow instead of the solid border. There is some transparency overlaying that happens but it still gives a little more depth than the solid border. - SA: https://github.com/Shopify/polaris-react/pull/1034
* [x] Fix bug: disabled prop doesn't prevent the thumbs from moving. https://github.com/Shopify/polaris-react/pull/1036
* [ ] Test mouse event handlers
* [ ] Test resize event handler

### How to 🎩

### Notes
- The position of the thumbs is controlled by exact pixel values and not by percentages. This was necessary to properly adjust the upper thumb position so its right edge doesn't surpass the track. See the screenshots (I moved the thumbs above the track so you can see what I mean).
#### Before adjustment
screen shot 2019-01-03 at 7 46 44 am

#### After adjustment
screen shot 2019-01-03 at 7 47 01 am

<details>

```jsx
import {Page, RangeSlider, Heading, Stack, TextField} from '../src';
import {Key} from '../src/types';

interface State {
  intermediateTextFieldValue: [number, number];
  dualValue: [number, number];
  singleValue: number;
}

const initialValue: [number, number] = [500, 1000];

  state: State = {
    intermediateTextFieldValue: initialValue,
    dualValue: initialValue,
    singleValue: 50,
  };

  handleRangeSliderChange = (value) => {
    this.setState({dualValue: value, intermediateTextFieldValue: value});
  };

  handleLowerTextFieldChange = (value) => {
    const upperValue = this.state.dualValue[1];
    this.setState({intermediateTextFieldValue: [parseInt(value), upperValue]});
  };

  handleUpperTextFieldChange = (value) => {
    const lowerValue = this.state.dualValue[0];
    this.setState({intermediateTextFieldValue: [lowerValue, parseInt(value)]});
  };

  handleLowerTextFieldBlur = () => {
    const upperValue = this.state.dualValue[1];
    const value = this.state.intermediateTextFieldValue[0];
    this.setState({dualValue: [Number(value), upperValue]});
  };

  handleUpperTextFieldBlur = () => {
    const lowerValue = this.state.dualValue[0];
    const value = this.state.intermediateTextFieldValue[1];
    this.setState({dualValue: [lowerValue, Number(value)]});
  };

  handleEnterKeyPress = (event) => {
    const newValue = this.state.intermediateTextFieldValue;
    const oldValue = this.state.dualValue;
    if (event.keyCode === Key.Enter && newValue !== oldValue) {
      this.setState({dualValue: newValue});
    }
  };

  handleSingleChange = (value: number) => {
  };

  render() {
    const {dualValue, intermediateTextFieldValue} = this.state;
    const prefix = '$';
    const min = 0;
    const max = 2000;
    const step = 10;
    const disabled = false;
    const lowerTextFieldValue =
      intermediateTextFieldValue[0] === dualValue[0]
        ? dualValue[0]
        : intermediateTextFieldValue[0];
    const upperTextFieldValue =
      intermediateTextFieldValue[1] === dualValue[1]
        ? dualValue[1]
        : intermediateTextFieldValue[1];

    return (
      <Page title="Playground">
        <div style={{marginTop: '20px'}} onKeyDown={this.handleEnterKeyPress}>
          <RangeSlider
            label="Money spent is between"
            value={dualValue}
            prefix={prefix}
            output
            min={min}
            max={max}
            step={step}
            disabled={disabled}
            onChange={this.handleRangeSliderChange}
          />
          <Stack distribution="equalSpacing" spacing="extraLoose">
            <TextField
              label="Min money spent"
              type="number"
              value={`${lowerTextFieldValue}`}
              prefix={prefix}
              min={min}
              max={max}
              step={step}
              disabled={disabled}
              onChange={this.handleLowerTextFieldChange}
              onBlur={this.handleLowerTextFieldBlur}
            />
            <TextField
              label="Max money spent"
              type="number"
              value={`${upperTextFieldValue}`}
              prefix={prefix}
              min={min}
              max={max}
              step={step}
              disabled={disabled}
              onChange={this.handleUpperTextFieldChange}
              onBlur={this.handleUpperTextFieldBlur}
            />
          </Stack>
        </div>

        <div style={{width: '300px', height: '100px', marginTop: '100px'}}>
          <Heading>Native range slider</Heading>
          <br />
          <RangeSlider
            label=""
            labelHidden
            value={this.state.singleValue}
            output
            min={5}
            max={30}
            prefix="low"
            suffix="high"
          />
        </div>
      </Page>
    );
  }
}

```
</details

 
### 🎩 checklist

* [ ] Dual thumb looks and behaves as the single thumb
* [ ] Thumbs don't surpass lower and upper boundaries 
* [ ] Thumbs don't surpass each other
* [ ] If initial values provided are not within bounds, they adjust
* [ ] Up and right arrows increment
* [ ] Down and left arrows decrement
* [ ] Clicking on the track moves the nearest thumb to the click position
* [ ] Clicking on the track and moving the cursor drags the nearest thumb
* [ ] Values increment and decrement by the provided step (or 1, as default)
* [ ] Values are properly reflected in output (tooltips) and accessibility inputs
* [ ] Manually typing a value in an accessibility input changes the thumb value on enter keypress or blur
* [ ] Changing the value in an accessibility input with the spinners changes the thumb value
* [ ] If values entered are not within bounds or not a multiple of step, they adjust
* [ ] Accessibility inputs fit at different widths and with varying numbers of characters
* [ ] prefix and suffix show up
* [ ] When there is an error, it returns to indigo on active and focus (like the native single thumb)
* [ ] Disables properly (and cursor is 🚫)
* [ ] Indigo indicates the selected range on the track
* [ ] Tests pass
* [ ] Tests are complete
* [ ] Style guide example is clear and works well

### DIFF
--- a/src/components/RangeSlider/README.md
+++ b/src/components/RangeSlider/README.md
@@ -29,7 +29,7 @@ Range sliders should:
 - When a label is visible, it should clearly communicate the purpose of the range input and its values (min, max, step, value)
 - Be labeled as “Optional” when you need to request input that’s not required
 - Validate input as soon as merchants have finished interacting with a field (but not before)
-- Always be used with `accessibilityInputs` when range slider has dual thumbs, to provide accessible alternatives to sliding the thumbs
+- Always be used with two text field components when range slider has dual thumbs, to provide accessible alternatives to both the lower and upper thumbs
 
 ---
 
@@ -231,21 +231,109 @@ class RangeSliderExample extends React.Component {
 Use a dual thumb range slider when merchants need to select a range of values.
 
 ```jsx
+const initialValue = [900, 1000];
+
 class RangeSliderExample extends React.Component {
-  handleChange = (value) => {
-    console.log({value});
+  state = {
+    intermediateTextFieldValue: initialValue,
+    value: initialValue,
+  };
+
+  handleRangeSliderChange = (value) => {
+    this.setState({value, intermediateTextFieldValue: value});
+  };
+
+  handleLowerTextFieldChange = (value) => {
+    const upperValue = this.state.value[1];
+    this.setState({intermediateTextFieldValue: [parseInt(value), upperValue]});
+  };
+
+  handleUpperTextFieldChange = (value) => {
+    const lowerValue = this.state.value[0];
+    this.setState({intermediateTextFieldValue: [lowerValue, parseInt(value)]});
+  };
+
+  handleLowerTextFieldBlur = () => {
+    const upperValue = this.state.value[1];
+    const value = this.state.intermediateTextFieldValue[0];
+
+    this.setState({value: [parseInt(value), upperValue]});
+  };
+
+  handleUpperTextFieldBlur = () => {
+    const lowerValue = this.state.value[0];
+    const value = this.state.intermediateTextFieldValue[1];
+
+    this.setState({value: [lowerValue, parseInt(value)]});
+  };
+
+  handleEnterKeyPress = (event) => {
+    const newValue = this.state.intermediateTextFieldValue;
+    const oldValue = this.state.value;
+
+    if (event.keyCode === Key.Enter && newValue !== oldValue) {
+      this.setState({value: newValue});
+    }
   };
 
   render() {
+    const {value, intermediateTextFieldValue} = this.state;
+    const prefix = '$';
+    const min = 0;
+    const max = 2000;
+    const step = 10;
+    const disabled = false;
+
+    const lowerTextFieldValue =
+      intermediateTextFieldValue[0] === value[0]
+        ? value[0]
+        : intermediateTextFieldValue[0];
+    const upperTextFieldValue =
+      intermediateTextFieldValue[1] === value[1]
+        ? value[1]
+        : intermediateTextFieldValue[1];
+
     return (
       <Card sectioned>
-        <RangeSlider
-          label=""
-          value={[35, 60]}
-          onChange={this.handleChange}
-          accessibilityInputs
-          step={5}
-        />
+        <div style={{marginTop: '20px'}} onKeyDown={this.handleEnterKeyPress}>
+          <RangeSlider
+            label="Money spent is between"
+            value={value}
+            prefix={prefix}
+            output
+            min={min}
+            max={max}
+            step={step}
+            disabled={disabled}
+            onChange={this.handleRangeSliderChange}
+          />
+          <Stack distribution="equalSpacing" spacing="extraLoose">
+            <TextField
+              label="Min money spent"
+              type="number"
+              value={`${lowerTextFieldValue}`}
+              prefix={prefix}
+              min={min}
+              max={max}
+              step={step}
+              disabled={disabled}
+              onChange={this.handleLowerTextFieldChange}
+              onBlur={this.handleLowerTextFieldBlur}
+            />
+            <TextField
+              label="Max money spent"
+              type="number"
+              value={`${upperTextFieldValue}`}
+              prefix={prefix}
+              min={min}
+              max={max}
+              step={step}
+              disabled={disabled}
+              onChange={this.handleUpperTextFieldChange}
+              onBlur={this.handleUpperTextFieldBlur}
+            />
+          </Stack>
+        </div>
       </Card>
     );
   }
--- a/src/components/RangeSlider/README.md
+++ b/src/components/RangeSlider/README.md
@@ -29,6 +29,7 @@ Range sliders should:
 - When a label is visible, it should clearly communicate the purpose of the range input and its values (min, max, step, value)
 - Be labeled as “Optional” when you need to request input that’s not required
 - Validate input as soon as merchants have finished interacting with a field (but not before)
+- Always be used with `accessibilityInputs` when range slider has dual thumbs, to provide accessible alternatives to sliding the thumbs
 
 ---
 
@@ -218,6 +219,32 @@ class RangeSliderExample extends React.Component {
           onChange={this.handleChange}
           prefix={<p>Hue</p>}
           suffix={<p style={suffixStyles}>{this.state.value}</p>}
+        />
+      </Card>
+    );
+  }
+}
+```
+
+### Dual thumb range slider
+
+Use a dual thumb range slider when merchants need to select a range of values.
+
+```jsx
+class RangeSliderExample extends React.Component {
+  handleChange = (value) => {
+    console.log({value});
+  };
+
+  render() {
+    return (
+      <Card sectioned>
+        <RangeSlider
+          label=""
+          value={[35, 60]}
+          onChange={this.handleChange}
+          accessibilityInputs
+          step={5}
         />
       </Card>
     );
--- a/src/components/RangeSlider/RangeSlider.tsx
+++ b/src/components/RangeSlider/RangeSlider.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import {createUniqueIDFactory} from '@shopify/javascript-utilities/other';
 import {withAppProvider, WithAppProviderProps} from '../AppProvider';
 import {Props, RangeSliderValue} from './types';
+import {DEFAULT_RANGE_SLIDER_PROPS} from './utilities';
 
 import {SingleThumb, DualThumb} from './components';
 
@@ -12,9 +13,9 @@ const getUniqueID = createUniqueIDFactory('RangeSlider');
 export function RangeSlider(props: CombinedProps) {
   const {
     id = getUniqueID(),
-    min = 0,
-    max = 100,
-    step = 1,
+    min = DEFAULT_RANGE_SLIDER_PROPS.min,
+    max = DEFAULT_RANGE_SLIDER_PROPS.max,
+    step = DEFAULT_RANGE_SLIDER_PROPS.step,
     value,
     ...rest
   } = props;
--- a/src/components/RangeSlider/RangeSlider.tsx
+++ b/src/components/RangeSlider/RangeSlider.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import {createUniqueIDFactory} from '@shopify/javascript-utilities/other';
 import {withAppProvider, WithAppProviderProps} from '../AppProvider';
-import {Props, RangeSliderValue} from './types';
+import {Props, RangeSliderValue, DualValue} from './types';
 import {RangeSliderDefault} from './utilities';
 
 import {SingleThumb, DualThumb} from './components';
@@ -35,7 +35,7 @@ export function RangeSlider(props: CombinedProps) {
   );
 }
 
-function isDualThumb(value: RangeSliderValue): value is [number, number] {
+function isDualThumb(value: RangeSliderValue): value is DualValue {
   return Array.isArray(value);
 }
 
--- a/src/components/RangeSlider/RangeSlider.tsx
+++ b/src/components/RangeSlider/RangeSlider.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import {createUniqueIDFactory} from '@shopify/javascript-utilities/other';
 import {withAppProvider, WithAppProviderProps} from '../AppProvider';
 import {Props, RangeSliderValue} from './types';
-import {DEFAULT_RANGE_SLIDER_PROPS} from './utilities';
+import {RangeSliderDefault} from './utilities';
 
 import {SingleThumb, DualThumb} from './components';
 
@@ -13,9 +13,9 @@ const getUniqueID = createUniqueIDFactory('RangeSlider');
 export function RangeSlider(props: CombinedProps) {
   const {
     id = getUniqueID(),
-    min = DEFAULT_RANGE_SLIDER_PROPS.min,
-    max = DEFAULT_RANGE_SLIDER_PROPS.max,
-    step = DEFAULT_RANGE_SLIDER_PROPS.step,
+    min = RangeSliderDefault.Min,
+    max = RangeSliderDefault.Max,
+    step = RangeSliderDefault.Step,
     value,
     ...rest
   } = props;
--- a/src/components/RangeSlider/RangeSlider.tsx
+++ b/src/components/RangeSlider/RangeSlider.tsx
@@ -1,205 +1,41 @@
 import * as React from 'react';
-import {autobind} from '@shopify/javascript-utilities/decorators';
 import {createUniqueIDFactory} from '@shopify/javascript-utilities/other';
-import {classNames} from '@shopify/react-utilities/styles';
-
-import {Error} from '../../types';
 import {withAppProvider, WithAppProviderProps} from '../AppProvider';
-import Labelled, {Action, helpTextID} from '../Labelled';
+import {Props, RangeSliderValue} from './types';
 
-import styles from './RangeSlider.scss';
+import {SingleThumb, DualThumb} from './components';
 
-export interface State {
-  id: string;
-}
-
-export interface BaseProps {
-  /** Label for the range input */
-  label: string;
-  /** Adds an action to the label */
-  labelAction?: Action;
-  /** Visually hide the label */
-  labelHidden?: boolean;
-  /** ID for range input */
-  id?: string;
-  /** Initial value for range input */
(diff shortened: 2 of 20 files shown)


src/components/RangeSlider/RangeSlider.tsx
```diff
@@ -106,7 +113,7 @@ export class RangeSlider extends React.PureComponent<CombinedProps, State> {
       ? describedBy.join(' ')
       : undefined;
 
-    const sliderProgress = ((value - min) * 100) / (max - min);
+    const sliderProgress = ((value as number - min) * 100) / (max - min);
```
AndrewMusgrave:
```suggestion
    const sliderProgress = (((value as number) - min) * 100) / (max - min);
```

This may fix it

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -19,9 +21,7 @@ export interface State {
 }
 
 export interface Props extends RangeSliderProps {
-  cssVarPrefix: string;
   value: [number, number];
-  accessibilityInputs?: boolean;
```
danrosenthal: @tmlayton why did you get rid of this?

danrosenthal: Never mind I see now

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -0,0 +1,594 @@
...
+  @autobind
+  private handleKeypressLower(event: KeyboardEvent) {
+    const {valueLower, valueUpper} = this.state;
+    const {step, min} = this.props;
+    event.preventDefault();
+    event.stopPropagation();
+
+    let newValue = valueLower;
+
+    if (event.keyCode === Key.DownArrow || event.keyCode === Key.LeftArrow) {
```
tmlayton: We will want to make this method no-op for key presses we don't care about it, otherwise we are unnecessarily setting state.

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -0,0 +1,594 @@
...
+        valueUpper: valueWithinBoundsUpper,
+      },
+      () => {
+        this.handleChange();
+      },
+    );
+  }
+
+  @autobind
+  private handleKeypressLower(event: KeyboardEvent) {
```
tmlayton: We will want to make this method no-op for key presses we don't care about it, otherwise we are unnecessarily setting state.

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -0,0 +1,594 @@
...
+    const {valueUpper, valueLower} = this.state;
+    const steppedValue = roundToNearestStepValue(valueUpper, step);
+
+    const valueWithinBoundsUpper = keepValueWithinBoundsUpper(
+      steppedValue,
+      valueLower,
+      max,
+      step,
+    );
+    this.setState(
```
tmlayton: We are only holding the track width and the value in state, yet are setting state in 10 different places. We should be able to pair that down to just a couple places where state is being set.

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -0,0 +1,594 @@
...
+    );
+
+    this.setState({
+      valueLower: valueWithinBoundsLower,
+      valueUpper: valueWithinBoundsUpper,
+    });
+
+    if (this.track.current) {
+      this.setState({
+        trackWidth: this.track.current.getBoundingClientRect().width,
```
tmlayton: track width also needs to be updated when its container resizes, with either a resize observer or a window resize event

solonaarmstrong: For traceability: https://github.com/Shopify/polaris-react/issues/840

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
solonaarmstrong: I'll create an issue.

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -0,0 +1,594 @@
...
+        valueLower: valueWithinBoundsLower,
+      },
+      () => {
+        this.handleChange();
+      },
+    );
+  }
+
+  @autobind
+  private handleKeypressUpper(event: KeyboardEvent) {
```
tmlayton: There is a decent amount of duplicated logic throughout this component and handlers are doing too much imo. Handlers should be naive delegators that simply map to the underlying private API and should have minimal code in them.

A good way to think about this is pretend you have access to the private API and you can only control the component with a command line, so you would type commands like `incrementLowerThumb` and `decrementLowerThumb` which would end up being your private methods. Then once the "command line interface" is in place, layer in event handlers which interact with the component through that interface.

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -0,0 +1,594 @@
...
+    );
+
+    const valueWithinBoundsUpper = keepValueWithinBoundsUpper(
+      roundToNearestStepValue(valueUpper, step),
+      valueLower,
+      max,
+      step,
+    );
+
+    this.setState({
```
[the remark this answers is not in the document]
solonaarmstrong: So, similar to this? https://github.com/Shopify/polaris-react/pull/834/files#diff-cfa660a9862c8aeb7af82de8921b1d45R38

dleroux:
Just starting my review of this starting with a 🎩 .  I noticed that because the `grab` cursor is set on the `Thumb` it causes a bit of a weird UX when your mouse moves faster than the `Thumb`. I'm thinking maybe the `grab` cursor should remain as long as the mouse is pressed. You can see it a little bit in the animation but it's flickers more than that.

dualrange

src/components/RangeSlider/components/DualThumb/DualThumb.scss
```diff
@@ -0,0 +1,210 @@
...
+    opacity: 0.8;
+    cursor: not-allowed;
+  }
+
+  &.error {
+    &:active,
+    &:focus {
+      // stylelint-disable selector-max-specificity
+      // stylelint-disable selector-max-class
+      .Track {
```
dleroux: Is there maybe a way to add the style to track maybe? Not sure if it's doable but it would be nice to have all the .Track styles under that 1 selector.

src/components/RangeSlider/components/DualThumb/DualThumb.scss
```diff
@@ -0,0 +1,210 @@
...
+        var(--Polaris-RangeSlider-progress-lower),
+      var(--selected-range, transparent)
+        var(--Polaris-RangeSlider-progress-upper),
+      var(--unselected-range, transparent)
+        var(--Polaris-RangeSlider-progress-upper),
+      var(--unselected-range, transparent) 100%;
+    background-image: linear-gradient(to right, var(--gradient-colors));
+  }
+
+  --gradient-colors: var(--unselected-range, transparent) 0%,
```
dleroux: nit: maybe moving this up before the .error & block, so the styles on .Track are together followed by the variant.

src/components/RangeSlider/components/DualThumb/DualThumb.scss
```diff
@@ -0,0 +1,210 @@
...
+    cursor: not-allowed;
+  }
+
+  &.error {
+    &:active,
+    &:focus {
+      // stylelint-disable selector-max-specificity
+      // stylelint-disable selector-max-class
+      .Track {
+        --unselected-range: #{color('sky', 'dark')};
```
dleroux: Do we need the actual colors as css vars? 

$unselected-range: color('sky', 'dark');
$selected-range: color('indigo');
$error-selected-range: color('red');

Might make the css simpler?

src/components/RangeSlider/components/DualThumb/DualThumb.scss
```diff
@@ -0,0 +1,210 @@
+$stacking-order: (
+  output: 10,
+  input: 20,
+);
+
+$range-track-height: rem(4px);
+$range-thumb-size: rem(24px);
```
dleroux: for accessibility it's recommended that the area be 48px., have we experimented with that?

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -0,0 +1,477 @@
...
+
+    if (error) {
+      describedBy.push(`${id}Error`);
+    }
+
+    const ariaDescribedBy = describedBy.length
+      ? describedBy.join(' ')
+      : undefined;
+
+    const classNameTrackWrapper = classNames(
```
dleroux: nit: we normally do `trackWrapperClassName`

src/components/RangeSlider/types.ts
```diff
@@ -0,0 +1,43 @@
...
+  /** Initial value for range input */
+  value: RangeSliderValue;
+  /** Minimum possible value for range input */
+  min?: number;
+  /** Maximum possible value for range input */
+  max?: number;
+  /** Increment value for range input changes */
+  step?: number;
+  /** Provide a tooltip while sliding, indicating the current value */
+  output?: boolean;
```
dleroux: I had to look this up because of the <output> html tag.  Would this be more clear as `tooltip?: boolean`

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -0,0 +1,477 @@
...
+  [key: string]: () => void;
+}
+
+enum Control {
+  Lower,
+  Upper,
+}
+
+const THUMB_SIZE = 24;
+const OUTPUT_TIP_SIZE = 8;
```
dleroux: We removed the tip from tooltips and popover, do we want it here?

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -0,0 +1,477 @@
...
+export default class DualThumb extends React.Component<Props, State> {
+  static getDerivedStateFromProps(props: Props, state: State) {
+    const {min, step, max, value, onChange, id} = props;
+    const {prevValue} = state;
+
+    if (isEqual(prevValue, value)) {
+      return null;
+    }
+
+    const sanitizedValue: DualValue = sanitizeValue(value, min, max, step);
```
dleroux: `: DualValue` you shouldn't need this?

AndrewMusgrave:
Hoping to get to this today. But if not then Monday

dleroux:
The tooltip is not centered on the right Thumb for some reason.
screen shot 2019-01-25 at 12 47 11 pm

src/components/RangeSlider/types.ts
solonaarmstrong: That would be a breaking change, unfortunately, because this prop already exists in the single range slider.

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
solonaarmstrong: That's a good point. I didn't realize we'd also removed it from tooltips, I thought it was just the popover. If we're not using them anymore, then I agree it makes sense to remove it here too.

src/components/RangeSlider/components/DualThumb/DualThumb.scss
solonaarmstrong: Yes, you're right. The size is based on the native range slider. Turns out there is a browser-compatible way to style the native thumb, so maybe we should increase the size on both.

solonaarmstrong:
> I'm thinking maybe the grab cursor should remain as long as the mouse is pressed.

Maybe I can throw a class onto the thumb while the mouse is down.

solonaarmstrong:
> The tooltip is not centered on the right Thumb for some reason.

Yes, that's because of the number of characters widening the output and because the positioning is controlled by an exact pixel value. If we remove the tip, this should resolve on its own.

src/components/RangeSlider/components/DualThumb/DualThumb.scss
solonaarmstrong: I don't have a strong opinion one way or another. I found it a bit unnecessary to use a variable, as well, but this is the way it was done originally in the single range slider, so it's consistent with that file. It might be worth opening a separate issue to simplify the CSS in both files. I'll skip it here so keep the scope clear.

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
solonaarmstrong: I may actually do this in a followup PR.

src/components/RangeSlider/components/DualThumb/DualThumb.scss
solonaarmstrong: Chatted in Slack and I'm changing it in this PR. 😄

solonaarmstrong: I just changed these and it's far less readable. What I do like about keeping them in the variables is that it makes the gradients clearer in terms of which are the selected and unselected areas. What do you think?

src/components/RangeSlider/components/DualThumb/DualThumb.scss
dpersing: [WCAG 2.1 has added target size guidance at the AAA level](https://www.w3.org/TR/WCAG21/#target-size) (44px), which is technically out of scope for what we're targeting, from a """compliance""" perspective, since we (and most orgs) are focusing on A and AA level. That said, it might be helpful to release the slider as is and get some usability feedback! We could always adjust the size and other factors later.

solonaarmstrong:
The thumbs are 2px smaller than the native one. And the question about opacity on the border was never resolved.

screen shot 2019-02-08 at 12 48 11 pm

src/components/RangeSlider/README.md
```diff
@@ -225,6 +226,120 @@ class RangeSliderExample extends React.Component {
...
+    intermediateTextFieldValue: initialValue,
+    value: initialValue,
+  };
+
+  handleRangeSliderChange = (value) => {
+    this.setState({value, intermediateTextFieldValue: value});
+  };
+
+  handleLowerTextFieldChange = (value) => {
+    const upperValue = this.state.value[1];
```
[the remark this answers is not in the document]
solonaarmstrong: Do we use deconstruction in the style guide? I've seen a few examples that don't.

src/components/RangeSlider/README.md
```diff
@@ -225,6 +226,120 @@ class RangeSliderExample extends React.Component {
...
 ```
 
+### Dual thumb range slider
+
+Use a dual thumb range slider when merchants need to select a range of values.
+
+```jsx
+const initialValue = [900, 1000];
+
+class RangeSliderExample extends React.Component {
```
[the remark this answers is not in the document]
solonaarmstrong: React.PureComponent for an example?

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
solonaarmstrong: https://github.com/Shopify/polaris-react/pull/988

tmlayton:
Closing in lieu of #1436

Issue #785, opened 2018-12-19
Make Polaris StrictMode compliant


jineshshah36:
# Issue summary

Currently, there are many components that are not StrictMode compliant within Polaris because they using the legacy context API or use deprecated lifecycle methods. Given that React 16.7 will support a concurrent mode which should hopefully improve application performance for most situations, it would be great if Polaris because StrictMode compliant so that we could use it in the future without issue.

## Expected behavior

Polaris should be StrictMode compliant

## Actual behavior

Polaris is not StrictMode compliant

## Steps to reproduce the problem

1. Wrap Polaris components with the <React.StrictMode></React.StrictMode> component
2. The console will begin spitting out errors.

## Reduced test case

N/A

## Specifications

N/A

mbaumbach:
Related issue: #457. Also, React 16.7 was just released and doesn't include concurrent mode (it's now in an unannounced future version), but agreed, it would be great to see Polaris compliant with `StrictMode`.

AndrewMusgrave:
Hi @jineshshah36 , as Ben mentioned we're currently in the process of removing deprecated lifecycle hooks as tracked in #519. We don't have an issue tracking the switch to the new context api, but we've already begun using the new context api in a few components and will be switching more in the future.

jineshshah36:
I'm getting a 500 error when I try to visit https://github.com/Shopify/polaris-react/issues/519, but that is good to hear

PR #803, opened 2018-12-21
[WIP] Refactor legacy frame context

danrosenthal:
Addresses #785.

### DIFF
--- a/src/components/Frame/components/Context/Context.tsx
+++ b/src/components/Frame/components/Context/Context.tsx
@@ -1,0 +1,19 @@
+import * as React from 'react';
+import {FrameContext} from '../../types';
+
+const defaultContext: FrameContext = {
+  frame: {
+    showToast: noop,
+    hideToast: noop,
+    setContextualSaveBar: noop,
+    removeContextualSaveBar: noop,
+    startLoading: noop,
+    stopLoading: noop,
+  },
+};
+
+function noop() {}
+
+const {Provider, Consumer} = React.createContext<FrameContext>(defaultContext);
+
+export {Provider, Consumer};
--- a/src/components/Loading/Loading.tsx
+++ b/src/components/Loading/Loading.tsx
@@ -1,21 +1,25 @@
 import * as React from 'react';
+import compose from '@shopify/react-compose';
 import {Loading as AppBridgeLoading} from '@shopify/app-bridge/actions';
-import {FrameContext, frameContextTypes} from '../Frame';
+import {FrameContext, Consumer} from '../Frame';
+import withContext from '../WithContext';
+import {WithContextTypes} from '../../types';
 import {withAppProvider, WithAppProviderProps} from '../AppProvider';
 
 export interface Props {}
-export type ComposedProps = Props & WithAppProviderProps;
+export type ComposedProps = Props &
+  WithAppProviderProps &
+  WithContextTypes<FrameContext>;
 
 export class Loading extends React.PureComponent<ComposedProps, never> {
-  static contextTypes = frameContextTypes;
-  context: FrameContext;
   private appBridgeLoading: AppBridgeLoading.Loading | undefined;
 
   componentDidMount() {
-    const {appBridge} = this.props.polaris;
+    const {polaris, context} = this.props;
+    const {appBridge} = polaris;
 
     if (appBridge == null) {
-      this.context.frame.startLoading();
+      context.frame.startLoading();
     } else {
       this.appBridgeLoading = AppBridgeLoading.create(appBridge);
       this.appBridgeLoading.dispatch(AppBridgeLoading.Action.START);
@@ -23,10 +27,11 @@ export class Loading extends React.PureComponent<ComposedProps, never> {
   }
 
   componentWillUnmount() {
-    const {appBridge} = this.props.polaris;
+    const {polaris, context} = this.props;
+    const {appBridge} = polaris;
 
     if (appBridge == null) {
-      this.context.frame.stopLoading();
+      context.frame.stopLoading();
     } else if (this.appBridgeLoading != null) {
       this.appBridgeLoading.dispatch(AppBridgeLoading.Action.STOP);
     }
@@ -37,4 +42,7 @@ export class Loading extends React.PureComponent<ComposedProps, never> {
   }
 }
 
-export default withAppProvider<Props>()(Loading);
+export default compose<Props>(
+  withContext<Props, WithAppProviderProps, FrameContext>(Consumer),
+  withAppProvider<Props>(),
+)(Loading);
--- a/src/components/Frame/Frame.tsx
+++ b/src/components/Frame/Frame.tsx
@@ -11,12 +11,8 @@ import Backdrop from '../Backdrop';
 import TrapFocus from '../TrapFocus';
 import {dataPolarisTopBar, layer, Duration} from '../shared';
 import {setRootProperty} from '../../utilities/setRootProperty';
-import {
-  ContextualSaveBarProps,
-  FrameContext,
-  frameContextTypes,
-  ToastProps,
-} from './types';
+import {ContextualSaveBarProps, FrameContext, ToastProps} from './types';
+import {Provider} from './components/Context';
 import {ToastManager, Loading, ContextualSaveBar} from './components';
 
 import * as styles from './Frame.scss';
@@ -54,8 +50,6 @@ const APP_FRAME_LOADING_BAR = 'AppFrameLoadingBar';
 export type CombinedProps = Props & WithAppProviderProps;
 
 export class Frame extends React.PureComponent<CombinedProps, State> {
-  static childContextTypes = frameContextTypes;
-
   state: State = {
     skipFocused: false,
     globalRibbonHeight: 0,
@@ -68,19 +62,6 @@ export class Frame extends React.PureComponent<CombinedProps, State> {
   private contextualSaveBar: ContextualSaveBarProps | null;
 
   private globalRibbonContainer: HTMLDivElement | null = null;
-
-  getChildContext(): FrameContext {
-    return {
-      frame: {
-        showToast: this.showToast,
-        hideToast: this.hideToast,
-        startLoading: this.startLoading,
-        stopLoading: this.stopLoading,
-        setContextualSaveBar: this.setContextualSaveBar,
-        removeContextualSaveBar: this.removeContextualSaveBar,
-      },
-    };
-  }
 
   componentDidMount() {
     this.handleResize();
@@ -240,28 +221,30 @@ export class Frame extends React.PureComponent<CombinedProps, State> {
       ) : null;
 
     return (
-      <div
-        className={frameClassName}
-        {...layer.props}
-        {...navigationAttributes}
-      >
-        {skipMarkup}
-        {topBarMarkup}
-        {contextualSaveBarMarkup}
-        {loadingMarkup}
-        {navigationOverlayMarkup}
-        {navigationMarkup}
-        <main
-          className={styles.Main}
-          id={APP_FRAME_MAIN}
-          data-has-global-ribbon={Boolean(globalRibbon)}
+      <Provider value={this.getContext}>
+        <div
+          className={frameClassName}
+          {...layer.props}
+          {...navigationAttributes}
         >
-          <div className={styles.Content}>{children}</div>
-        </main>
-        <ToastManager toastMessages={toastMessages} />
-        {globalRibbonMarkup}
-        <EventListener event="resize" handler={this.handleResize} />
-      </div>
+          {skipMarkup}
+          {topBarMarkup}
+          {contextualSaveBarMarkup}
+          {loadingMarkup}
+          {navigationOverlayMarkup}
+          {navigationMarkup}
+          <main
+            className={styles.Main}
+            id={APP_FRAME_MAIN}
+            data-has-global-ribbon={Boolean(globalRibbon)}
+          >
+            <div className={styles.Content}>{children}</div>
+          </main>
+          <ToastManager toastMessages={toastMessages} />
+          {globalRibbonMarkup}
+          <EventListener event="resize" handler={this.handleResize} />
+        </div>
+      </Provider>
     );
   }
 
@@ -389,6 +372,20 @@ export class Frame extends React.PureComponent<CombinedProps, State> {
     if (key === 'Escape') {
       this.handleNavigationDismiss();
     }
+  }
+
+  @autobind
+  get getContext(): FrameContext {
+    return {
+      frame: {
+        showToast: this.showToast,
+        hideToast: this.hideToast,
+        startLoading: this.startLoading,
+        stopLoading: this.stopLoading,
+        setContextualSaveBar: this.setContextualSaveBar,
+        removeContextualSaveBar: this.removeContextualSaveBar,
+      },
+    };
   }
 }
 
--- a/src/components/Frame/components/Context/index.ts
+++ b/src/components/Frame/components/Context/index.ts
@@ -1,0 +1,1 @@
+export {Provider, Consumer} from './Context';
--- a/src/components/Frame/components/index.ts
+++ b/src/components/Frame/components/index.ts
@@ -5,3 +5,4 @@ export {
 } from './ToastManager';
 export {default as Loading, Props as LoadingProps} from './Loading';
 export {default as ContextualSaveBar} from './ContextualSaveBar';
+export {Consumer} from './Context';
--- a/src/components/Frame/index.ts
+++ b/src/components/Frame/index.ts
@@ -2,13 +2,8 @@ import Frame from './Frame';
 
 export {Props} from './Frame';
 
-export {DEFAULT_TOAST_DURATION} from './components';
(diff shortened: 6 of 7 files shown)


src/components/ContextualSaveBar/ContextualSaveBar.tsx
```diff
@@ -1,32 +1,36 @@
...
-    this.context.frame.removeContextualSaveBar();
+    this.props.context.frame.removeContextualSaveBar();
   }
 
-  componentDidUpdate(oldProps: Props) {
-    if (contextualSaveBarHasChanged(this.props, oldProps)) {
-      this.context.frame.setContextualSaveBar(this.props);
+  componentDidUpdate(oldProps: ComposedProps) {
+    const {props} = this;
+    const {polaris, context, ...rest} = props;
```
danrosenthal: This bit feels wrong to me. I don't think I should need to pull `polaris` and `context` off `props` in order for this to work.

src/components/ContextualSaveBar/tests/ContextualSaveBar.test.tsx
```diff
@@ -12,58 +12,111 @@ describe('<ContextualSaveBar />', () => {
...
+    const mockFrameContext = {
+      frame: {
+        setContextualSaveBar: jest.fn(),
+        removeContextualSaveBar: jest.fn(),
+        showToast: noop,
+        hideToast: noop,
+        startLoading: noop,
+        stopLoading: noop,
+      },
+    };
```
danrosenthal: This should be moved into a function that generates a mock frame context and accepts spies for methods available there.

src/components/ContextualSaveBar/tests/ContextualSaveBar.test.tsx
```diff
@@ -12,58 +12,111 @@ describe('<ContextualSaveBar />', () => {
...
+
+    mountWithAppProvider(
+      <Provider value={mockFrameContext}>
+        <ContextualSaveBar {...props} />
```
danrosenthal: I currently get a TypeScript error on `ContextualSaveBar` saying:

```
Type '{ saveAction: { content: string; onAction: () => void; }; discardAction: { content: string; onAction: () => void; }; message: string; }' is missing the following properties from type 'Readonly<Props & WithAppProviderProps & WithContextTypes<FrameContext>>': polaris, context
```

The code still compiles properly, and tests pass, but this feels wrong to me, and maybe related to the issue in `ContextualSaveBar` where I have to pull `polaris`, and `context` off `this.props`.

src/components/ContextualSaveBar/tests/ContextualSaveBar.test.tsx
```diff
@@ -12,58 +12,111 @@ describe('<ContextualSaveBar />', () => {
...
+    const mockFrameContext = {
+      frame: {
+        setContextualSaveBar: jest.fn(),
+        removeContextualSaveBar: jest.fn(),
+        showToast: noop,
+        hideToast: noop,
+        startLoading: noop,
+        stopLoading: noop,
+      },
+    };
```
danrosenthal: Ditto here per making a function that mocks these.

src/components/ContextualSaveBar/tests/ContextualSaveBar.test.tsx
```diff
@@ -12,58 +12,111 @@ describe('<ContextualSaveBar />', () => {
...
+
+    const frame = mountWithAppProvider(
+      <Provider value={mockFrameContext}>
+        <ContextualSaveBar {...props} />
```
danrosenthal: Ditto here per the typing issue.

src/components/ContextualSaveBar/tests/ContextualSaveBar.test.tsx
```diff
@@ -12,58 +12,111 @@ describe('<ContextualSaveBar />', () => {
...
+    const mockFrameContext = {
+      frame: {
+        setContextualSaveBar: jest.fn(),
+        removeContextualSaveBar: jest.fn(),
+        showToast: noop,
+        hideToast: noop,
+        startLoading: noop,
+        stopLoading: noop,
+      },
+    };
```
danrosenthal: Need a function for mocking this.

src/components/ContextualSaveBar/tests/ContextualSaveBar.test.tsx
```diff
@@ -12,58 +12,111 @@ describe('<ContextualSaveBar />', () => {
...
+
+    const frame = mountWithAppProvider(
+      <Provider value={mockFrameContext}>
+        <ContextualSaveBar {...props} />
```
danrosenthal: Typing issue here too.

src/components/ContextualSaveBar/tests/ContextualSaveBar.test.tsx
```diff
@@ -12,58 +12,111 @@ describe('<ContextualSaveBar />', () => {
...
+    const mockFrameContext = {
+      frame: {
+        setContextualSaveBar: jest.fn(),
+        removeContextualSaveBar: jest.fn(),
+        showToast: noop,
+        hideToast: noop,
+        startLoading: noop,
+        stopLoading: noop,
+      },
+    };
```
danrosenthal: Ditto mocking function.

src/components/ContextualSaveBar/tests/ContextualSaveBar.test.tsx
```diff
@@ -12,58 +12,111 @@ describe('<ContextualSaveBar />', () => {
...
+
+    const frame = mountWithAppProvider(
+      <Provider value={mockFrameContext}>
+        <ContextualSaveBar {...props} />
```
danrosenthal: Ditto typing issue.

src/components/ContextualSaveBar/ContextualSaveBar.tsx
```diff
@@ -1,32 +1,36 @@
...
-class ContextualSaveBar extends React.PureComponent<Props, never> {
-  static contextTypes = frameContextTypes;
-  context: FrameContext;
-
+class ContextualSaveBar extends React.PureComponent<ComposedProps, never> {
   componentDidMount() {
-    this.context.frame.setContextualSaveBar(this.props);
+    const {props} = this;
+    const {polaris, context, ...rest} = props;
+    props.context.frame.setContextualSaveBar(rest);
```
danrosenthal: It feels a bit gross having all of this available on a `context` object inside of `props`. Do we think there might be a cleaner way to organize this?

https://github.com/Shopify/polaris-react/blob/master/src/types.ts#L243

dleroux: you wouldn't need the `props.` since you are pulling off the context.

dleroux:
I believe the `Toast` tests will need to be updated with the `Provider` just like you did for the `ContextualSavebar`.

src/components/ContextualSaveBar/tests/ContextualSaveBar.test.tsx
danrosenthal: Looks like this was a result of using the wrong version of typescript (`3.2.2`, instead of `3.1.6`).

src/components/ContextualSaveBar/ContextualSaveBar.tsx
dleroux: I think this is just the reality because `setContextualSaveBar` accepts `ContextualSaveBarProps` which never included `polaris`.

danrosenthal:
Looks like the typing issue was to do with using the wrong version of TS in VSCode. So that problem is resolved. @dleroux this is still a WIP, so I haven't updated the other tests yet. Those tests failing was expected.

src/components/ContextualSaveBar/ContextualSaveBar.tsx
danrosenthal: In this case I was more referring to the HOC that adds all of our context to a `context` object, as shown in the linked code in the original comment. Wondering if this is the best pattern going forward.

cc: @AndrewMusgrave

AndrewMusgrave:
> Looks like the typing issue was to do with using the wrong version of TS in VSCode. So that problem is resolved. @dleroux this is still a WIP, so I haven't updated the other tests yet. Those tests failing was expected.

Seems there is an issue with `compose` returns the type with context in TS v.3.2.2. 🔥 [C](https://github.com/Shopify/quilt/blob/c1540e4b18ef840b55fcf8a00aada110ae0a277c/packages/react-compose/src/index.ts#L26) should fix it.

> Also feeling a little weird about putting everything onto a context object in props. @AndrewMusgrave can you explain the thinking behind this a bit? Does this feel right to you to keep as a pattern?

I think this will be ok to keep as a pattern. Rather than directly using a render prop this'll allow us to keep our render logic cleaner and force similar implementations across context usage. If we decide to change something in the future we'll only have to change it in the HoC. And this'll allow us to use the same injector pattern we're familiar with (e.g. withAppProvider, withSticky). 

I decided to call the prop `context` to keep it consistent with old react context (`this.context`). Although this can definitely be changed.

It's a very common pattern to pass context down as props, it's just not normally called context. You'll see it in most major libraries, for example `redux`, `react-router`, `apollo`, etc.

src/components/ContextualSaveBar/ContextualSaveBar.tsx
AndrewMusgrave: The other option is it spread context onto the components props rather than placing context on props.

E.g.
```tsx
this.props.context.a
// or
this.props.a
```

However if we don't namespace the context we could encounter naming collisions.

src/components/ContextualSaveBar/ContextualSaveBar.tsx
AndrewMusgrave: Technically it'll work without deconstruction since we don't normally spread props. I'm surprised we didn't have a TS error before, since `polaris` always existed on props. I wonder if we'll still have one now? 

This is just a by product of the injector pattern. Since our smart components will always receive additional props through `HoC`s that won't be passed to the component upon usage.

danrosenthal:
That rationale makes sense to me @AndrewMusgrave. Would you see us moving `polaris` context under the `context` object once we refactor AppProvider to use the new context API?

src/components/ContextualSaveBar/ContextualSaveBar.tsx
danrosenthal: I think that keeping it namespaced to `context` makes sense here. I'm good with this as a pattern.

src/components/ContextualSaveBar/ContextualSaveBar.tsx
```diff
@@ -1,32 +1,36 @@
...
+  WithContextTypes<FrameContext>;
 
-class ContextualSaveBar extends React.PureComponent<Props, never> {
-  static contextTypes = frameContextTypes;
-  context: FrameContext;
-
+class ContextualSaveBar extends React.PureComponent<ComposedProps, never> {
   componentDidMount() {
-    this.context.frame.setContextualSaveBar(this.props);
+    const {props} = this;
```
AndrewMusgrave: Should we do the deconstruction on the same line?
```tsx
const {
  props: {
    polaris,
    context,
    ...rest,
  }
}
```

src/components/ContextualSaveBar/ContextualSaveBar.tsx
AndrewMusgrave:
```suggestion
    context.frame.setContextualSaveBar(rest);
```

src/components/ContextualSaveBar/ContextualSaveBar.tsx
```diff
@@ -1,32 +1,36 @@
...
   componentWillUnmount() {
-    this.context.frame.removeContextualSaveBar();
+    this.props.context.frame.removeContextualSaveBar();
   }
 
-  componentDidUpdate(oldProps: Props) {
-    if (contextualSaveBarHasChanged(this.props, oldProps)) {
-      this.context.frame.setContextualSaveBar(this.props);
+  componentDidUpdate(oldProps: ComposedProps) {
+    const {props} = this;
```
AndrewMusgrave: Same deconstruction comment as above

src/components/ContextualSaveBar/ContextualSaveBar.tsx
```diff
@@ -1,32 +1,36 @@
...
+    const {polaris, context, ...rest} = props;
+    if (contextualSaveBarHasChanged(rest, oldProps)) {
+      props.context.frame.setContextualSaveBar(rest);
```
AndrewMusgrave:
```suggestion
      context.frame.setContextualSaveBar(rest);
```

src/components/ContextualSaveBar/tests/ContextualSaveBar.test.tsx
```diff
@@ -12,58 +12,87 @@ describe('<ContextualSaveBar />', () => {
...
     );
-    expect(frame.setContextualSaveBar).toHaveBeenCalledTimes(1);
     const newProps = {
       saveAction: {content: 'Save', onAction: noop, loading: true},
       discardAction: {content: 'Discard', onAction: noop},
       message: 'Unsaved changes',
     };
-    contextualSaveBar.setProps({...newProps});
-    expect(frame.setContextualSaveBar).toHaveBeenCalledWith({
+    frame.setProps({
```
AndrewMusgrave: Just curious, why do we need to `setProps` on frame now?

src/components/Loading/Loading.tsx
```diff
@@ -1,32 +1,37 @@
...
+  WithContextTypes<FrameContext>;
 
 export class Loading extends React.PureComponent<ComposedProps, never> {
-  static contextTypes = frameContextTypes;
-  context: FrameContext;
   private appBridgeLoading: AppBridgeLoading.Loading | undefined;
 
   componentDidMount() {
-    const {appBridge} = this.props.polaris;
+    const {polaris, context} = this.props;
```
AndrewMusgrave: Maybe a single line deconstruction ?

```tsx
    const {polaris: {appBridge}, context} = this.props;
```

src/components/Loading/Loading.tsx
```diff
@@ -1,32 +1,37 @@
...
+      context.frame.startLoading();
     } else {
       this.appBridgeLoading = AppBridgeLoading.create(appBridge);
       this.appBridgeLoading.dispatch(AppBridgeLoading.Action.START);
     }
   }
 
   componentWillUnmount() {
-    const {appBridge} = this.props.polaris;
+    const {polaris, context} = this.props;
```
AndrewMusgrave: ^

src/components/Frame/components/Context/Context.tsx
```diff
@@ -0,0 +1,19 @@
...
+    showToast: noop,
+    hideToast: noop,
+    setContextualSaveBar: noop,
+    removeContextualSaveBar: noop,
+    startLoading: noop,
+    stopLoading: noop,
+  },
+};
+
+function noop() {}
```
AndrewMusgrave: How come we're using a custom `noop`?

src/components/Frame/utilities/createFrameContext/createFrameContext.ts
```diff
@@ -0,0 +1,31 @@
+import {noop} from '@shopify/javascript-utilities/other';
+import {FrameContext, ToastProps, ContextualSaveBarProps} from '../../types';
+
+export interface CreateFrameContext {
+  showToast?(toast: {id: string} & ToastProps): void;
```
AndrewMusgrave: Since we reuse `{id: string}` should we create a new type for it?

```ts
interface A {id: string;}
type B = A & ToastProps;
```

src/components/ContextualSaveBar/tests/ContextualSaveBar.test.tsx
danrosenthal: This is an alternative to mounting and unmounting the child component, since we no longer have direct access to the child component. Calling `setProps` on Frame with children that have new props triggers the `componentDidUpdate` lifecycle method for those children.

lemonmade:
I took a little stab at improving the typing of decorators, I think having the following types should allow you to grab just the static types on the decorated component:

```ts
type StaticFields<Object> = {
  [Key in keyof Object]: Key extends 'prototype' ? never : Key
}[keyof Object];
type Statics<Object> = Pick<Object, StaticFields<Object>>;

// Example
class MyComponent extends React.Component {
  static foo = 'bar';
  static Baz(qux: string) {
    return <div>{qux}</div>;
  }

  render() {
    return null;
  }
}

type StaticMembers = Statics<typeof MyComponent>; // {foo: string; Baz: () => JSX.Element}
```

danrosenthal:
@dleroux, @AndrewMusgrave, @tmlayton this is ready for another look

danrosenthal:
@lemonmade is the Apollo issue preventing `Provider`/`Consumer` components at the Frame level no longer blocking this PR merging?

lemonmade:
Yes, we've shipped and reverted the Apollo upgrade several times so we are getting close, but it is currently blocking.

tmlayton:
Want to create a 4.0 branch and target it with this PR?

PR #830, opened 2019-01-08
Rename Option and Group types in Select to workaround typedoc oddness

BPScott:

Typedoc (which we use for generating prop tables on the styleguide) has
a bug in it where it doesn't correctly resolve types that are imported
from a different file (or we're using it wrong).
Instead of looking at imports it seems to look at a global bucket of
names. In this case there was a clash between Option (as living in Select)
and Option (as living in OptionList/components). 

This resulted it it thinking that the type for ResourceList -> sortOptions was `src/components/OptionList/components/Option/Option.tsx`'s default export instead of the type in `src/components/Select/Select.tsx`

Work around this by renaming Select's Option to "SelectOption".

This allows our type documentation generation in the styleguide to work
correctly.

Strictly speaking the change to SelectGroup isn't needed but i think
"SelectOption and SelectGroup" sounds nicer than "SelectOption and
Group".

### How to 🎩

- In polaris-styleguide checkout this PR: https://github.com/Shopify/polaris-styleguide/pull/2488
- In polaris-styleguide run `yarn run typedoc` and swear loudly because it fails.
- In polaris-react run `yarn build-consumer polaris-styleguide`
- In polaris-styleguide run `yarn run typedoc` again and see that it runs without errors. Cheer / run your temples /swear loudly again.
- In polaris-styleguide run `yarn dev` and see that the prop table on the ResourceList page correctly states that `sortOptions` has the typing `string | StrictOption[]` [just like in production](https://polaris.shopify.com/components/lists-and-tables/resource-list/sortOptions)

### DIFF
--- a/src/components/ResourceList/ResourceList.tsx
+++ b/src/components/ResourceList/ResourceList.tsx
@@ -8,7 +8,7 @@ import EventListener from '../EventListener';
 import Sticky from '../Sticky';
 import Spinner from '../Spinner';
 import {withAppProvider, WithAppProviderProps} from '../AppProvider';
-import Select, {Option} from '../Select';
+import Select, {SelectOption} from '../Select';
 import EmptySearchResult from '../EmptySearchResult';
 import selectIcon from './icons/enable-selection.svg';
 
@@ -60,7 +60,7 @@ export interface Props {
   /** Current value of the sort control */
   sortValue?: string;
   /** Collection of sort options to choose from */
-  sortOptions?: Option[];
+  sortOptions?: SelectOption[];
   /** ReactNode to display instead of the sort control */
   alternateTool?: React.ReactNode;
   /** Callback when sort option is changed */
--- a/src/components/index.ts
+++ b/src/components/index.ts
@@ -222,8 +222,8 @@ export {default as ScrollLock} from './ScrollLock';
 export {
   default as Select,
   Props as SelectProps,
-  Option as SelectOption,
-  Group as SelectGroup,
+  SelectOption,
+  SelectGroup,
 } from './Select';
 
 export {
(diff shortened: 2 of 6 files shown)


BPScott:
From @AndrewMusgrave on Slack "Don’t ask me why but this seems to be the culprit (https://github.com/Shopify/polaris-react/commit/719ca6694bbc65f617e646d3fa38f708befab467). I can’t for the life of me see why though :thinkhard:"

I think that file changed the order in which typedoc discovered files which resulted in a different `Option` being set as the "global Option" (for want of a better phrase).

src/components/index.ts
```diff
@@ -222,8 +222,8 @@ export {default as ScrollLock} from './ScrollLock';
 export {
   default as Select,
   Props as SelectProps,
-  Option as SelectOption,
-  Group as SelectGroup,
+  SelectOption,
```
BPScott: As we're not changing the export names in `src/components/index.ts` this is **not** a breaking change to our public API

src/components/ResourceList/ResourceList.tsx
```diff
@@ -8,7 +8,7 @@ import EventListener from '../EventListener';
 import Sticky from '../Sticky';
 import Spinner from '../Spinner';
 import {withAppProvider, WithAppProviderProps} from '../AppProvider';
-import Select, {Option} from '../Select';
+import Select, {SelectOption} from '../Select';
```
BPScott: I tried changing just the import statement to `import Select, {Option as SelectOption} from '../Select';` but that did not work.

`yarn run typedoc` compiled correctly but when viewing the types it just said `SelectOption[]` instead of exposing what that intersection type referred to  (i.e. `string | StrictOption[]`) like on production.

PR #834, opened 2019-01-08
[RangeSlider] Clamp single thumb

### LINKED ISSUE #815: [Range Slider] Fix values out of bounds
# Issue summary

While [adding a dual thumb](https://github.com/Shopify/polaris-react/pull/784), I discovered the native input doesn't correct a value out of bounds. This can happen:

screen shot 2019-01-02 at 8 45 40 am

## Expected behavior

The value should be adjusted, based on min and max.

<details>
<summary>Playground</summary>

```jsx

import {Page, RangeSlider} from '@shopify/polaris';

interface State {
  singleValue: number;
}

  state = {
    singleValue: 50 as number,
  };

  handleSingleChange = (value: number) => {
  };

  render() {
    return (
      <Page title="Playground">
        <RangeSlider
          label="Native ranger slider"
          value={this.state.singleValue}
          output
          min={5}
          max={30}
          prefix="low"
          suffix="high"
        />
      </Page>
    );
  }
}

```
</details>

Set the value higher than the max.

AndrewMusgrave:

Resolves #815

Creates a clamp function that adjusts a given number based on given min/max bounds.

### How to 🎩

<details>

```jsx
import {Page, RangeSlider} from '@shopify/polaris';

interface State {
  singleValue: number;
}

  state = {
    singleValue: 50 as number,
  };

  handleSingleChange = (value: number) => {
  };

  render() {
    return (
      <Page title="Playground">
        <div style={{width: '200px'}}>
          <RangeSlider
            label="Native ranger slider"
            value={this.state.singleValue}
            output
            min={5}
            max={30}
            prefix="low"
            suffix="high"
          />
        </div>
      </Page>
    );
  }
}

```

</details>

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -18,6 +18,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - Ensure disabled `Button` components with a `url` prop output valid HTML ([#773](https://github.com/Shopify/polaris-react/pull/773))
 - Fixed `DropZone` which was unable to add a duplicate file back to back or add a file again once removed [#782](https://github.com/Shopify/polaris-react/pull/782). Thank you [@jzsplk](https://github.com/jzsplk) for the contribution [#425](https://github.com/Shopify/polaris-react/issues/425) and [@vladucu](https://github.com/vladucu) for the clear example.
+- Added clamp to `RangeSlider` so its initial value doesn't fall outside of min or max bounds ([#834](https://github.com/Shopify/polaris-react/pull/834))
 
 ### Documentation
 
--- a/src/components/RangeSlider/components/SingleThumb/SingleThumb.tsx
+++ b/src/components/RangeSlider/components/SingleThumb/SingleThumb.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import {classNames} from '@shopify/react-utilities/styles';
+import clamp from '../../../../utilities/clamp';
 import Labelled, {helpTextID} from '../../../Labelled';
 
 import {invertNumber, CSS_VAR_PREFIX} from '../../utilities';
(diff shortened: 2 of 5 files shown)


solonaarmstrong:
🎉

UNRELEASED.md
```diff
@@ -12,6 +12,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Updated `Page` header layout so actions take up less room on small screens ([#707](https://github.com/Shopify/polaris-react/pull/707))
 - Added `alternateTool` prop to `ResourceList` component ([#812](https://github.com/Shopify/polaris-react/pull/812))
 - Added second thumb to `RangeSlider` ([#784](https://github.com/Shopify/polaris-react/pull/784))
+- Added clamp to `RangeSlider` so its initial value doesn't exceed min or max ([#834](https://github.com/Shopify/polaris-react/pull/834))
```
tmlayton: I don’t think we need a changelog item for this

tmlayton:
Let’s add some tests for this

UNRELEASED.md
tmlayton: On second thought, is it better to target master with this PR and add this changelog item as a bug fix?

solonaarmstrong: It can't target master just yet because it's relying on the changes in the dual thumb. But yes, I think this should be a bug fix. I am waiting to rebase with `range-slider-dual-input` when it has rebased with master.

tmlayton: Yep, I’m suggesting refactoring this PR to fix the production bug

solonaarmstrong: We chatted about this in Slack and I'm going to stick with pointing it to the feature branch. The reasoning being that this bug is entirely dependent on a consumer error, so I don't consider it an urgent fix. The merge conflict would be a headache and outweighs the slight possibility that anyone would encounter this bug.

solonaarmstrong: Also, the reason I think we need a changelog entry is that it's fixing an existing component (the single thumb slider), whereas everything else targets the dual thumb, which is new.

solonaarmstrong:
I changed `lower` and `upper` to `min` and `max` because the naming makes more sense to me.

PR #836, opened 2019-01-08
Remove all usages of babel-node

BPScott:

That extra transpile step means an extra dependency but doesn't add
any value

- Replaces all calls to `babel-node` with `node`.
- Removes the dependency on babel-cli
- Updates all scripts so they work with plain node. This boiled down to replacing `import` lines with `require()` and replacing `export default` with `module.exports = `

#### build and build-consumer

- Check out master branch
- Run `mkdir -p old-build/node_modules/@shopify/polaris-react && yarn run build-consumer polaris-react/old-build` This will create a baseline build to compare against
- Checkout this branch and `yarn install`
- Run `yarn clean && mkdir -p new-build/node_modules/@shopify/polaris-react && yarn run build-consumer polaris-react/new-build` This creates a new build
- Run `diff -r -u old-build new-build` to compare the build outputs. The only difference should be within the package.json

#### readme-update-version

- Modify the version in package.json (e.g change it to `3.4.0-rc.1`)
- Run `yarn run readme-update-version` and see that README file contents have been updated

#### open-styleguide-pr

Kinda hard without making a release to test.


BPScott:
Rebased to include a changelog entry 👍

PR #837, opened 2019-01-09
Update build tools (Babel, Postcss, Rollup)

BPScott:

Keeping up to date

- Update babel to use version 7
- Update postCSS to use version 7
- Update rollup to version 1
- Update babel, postcss and rollup plugins to versions referenced in sewing-kit

babel-core "7.0.0-bridge.0" is needed for the moment as Jest 23 does not
support Babel 7 out the box (support will be in Jest 24)

### How to 🎩

- Ensure tests pass
- Click around storybook and check percy passes
- https://github.com/Shopify/polaris-styleguide/pull/2498 is a styleguide PR that uses the beta version from this branch. Ensure Percy passes and click around https://polaris-styleguide-pr-2498.herokuapp.com to ensure functionality still works.
- https://github.com/Shopify/web/pull/9681 is a web PR in web that uses the beta version from this branch. The shrink ray build suggests no meaningful changes to the build output (the PR has false-positives because shrink-ray compares your PR against master, not commit your PR is based off). Checkout that branch, run `dev server` and click around to ensure functionality still works.
- Create a create-react-app app, and ensure polaris still renders:
  ```
  npx create-react-app cra-test-polaris && cd cra-test-polaris
  yarn add @shopify/polaris^3.5.0-beta.1
  ```  
  then replace the contents of App.js with: 

  ```
  import React, { Component } from 'react';
  import {AppProvider, Page, Button} from '@shopify/polaris';
  import '@shopify/polaris/styles.css';

  class App extends Component {
    render() {
      return (
        <AppProvider><Page><Button>HI THERE!</Button></Page></AppProvider>
      );
    }
  }

  export default App;
  ```

then run `npm  start` to see the page still renders with polaris content


BPScott:
@lemonmade this seemed quite straightforward. I can't work out if I should be suspicious or relieved.

BPScott:
Total bundle size has increased (by ~10k) but believe that is due to babel-plugin-shopify taking a more conservative approach to compilation targets by default.

@lemonmade do you have any opinions on how we should handle this? Should we target a more aggressive baseline by setting a browserlist?

I **think** if we choose a reasonably aggressive baseline, then any consumers of polaris would be able to transpile our code into something with wider support if they use babel?

BPScott:
Adding a browserlist config seems to have sorted that. We're using the [shopify standard one](https://github.com/Shopify/browserslist-config/blob/master/index.js) but we can't extend from it as apps using older versions of sewing-kit don't like the `@extends` syntax. It has shrunk down our bundle size nicely, as out compile targets are now more restrictive (previously babel used [this list, hardcoded in babel-preset-shopify](https://github.com/Shopify/babel-preset-shopify/blob/v16.2.0/browsers.js), and the postcss used the the browserlist default from years ago). Updating our versions of the caniuse library also means we get up to date browser references - e.g. "last chrome browser" is v71, instead of being the last chrome browser at whatever point in time we last updated our caniuse dependency.

On master:
```
$ yarn install && yarn run build && du {index.*,styles.*}
$ // A bunch of build output, and then:
684K    index.es.js
696K    index.js
224K    styles.css
4.0K    styles.scss
```

Meanwhile on this branch:

```
$ yarn install && yarn run build && du {index.*,styles.*}
$ // A bunch of build output, and then:
624K    index.es.js
640K    index.js
196K    styles.css
4.0K    styles.scss
```

kaelig:
Very nice for performance! I think this deserves the @addyosmani stamp of approval ⚡️ 

Addy

addyosmani:
Nice! Based on the comments, I approve of the bytes shaved on this branch 👍

BPScott:
Deployed this to web and styleguide. This is now ready for review.

kaelig:
Oops, misclick!

I was just going to say I just ran a [SpeedCurve test on polaris.shopify.com](https://speedcurve.com/shopify/core-foundations/deploy-status/?b=chrome&dl=327322&dp=317254&r=us-east-1&s=39950&u=75317&share=xqwmih55678bzck9g0ue73yngezz1l) so we have a baseline to see what impact this new version has on the style guide's performance.

BPScott:
> I was just going to say I just ran a SpeedCurve test on polaris.shopify.com so we have a baseline to see what impact this new version has on the style guide's performance.

The change is nil as apps that use sewing-kit push polaris through webpack anyway so they're already using a tree-shaken minimal build. Check the shrink-ray report on polaris-styleguide and the build is effectively unchanged.

tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "baseUrl": "./src",
     "rootDir": "./",
-    "target": "es2016",
+    "target": "es2017",
```
kaelig: Is this a breaking change?

.storybook/webpack.config.js
```diff
@@ -42,16 +42,13 @@ module.exports = (baseConfig, env, config) => {
           loader: 'babel-loader',
           options: {
             babelrc: false,
+            configFile: `${__dirname}/.babelrc`,
```
kaelig: I must be missing something… about why this line is here, as it seems to contradict the above line (`babelrc: false`)?

.storybook/webpack.config.js
```diff
@@ -42,16 +42,13 @@ module.exports = (baseConfig, env, config) => {
...
             // Don't use the production environment as it contains optimisations
             // that break compilation. The shopify/react preset enables the
             // babel-plugin-transform-react-constant-elements plugin which
             // somehow hoists things up into an undesirable location.
-            forceEnv: isProduction ? 'not-production' : undefined,
+            envName: isProduction ? 'not-production' : undefined,
```
kaelig: Does compilation still break on Heroku if this workaround is removed?

package.json
```diff
@@ -123,24 +130,24 @@
...
     "react-dom": "^16.4.0",
     "react-test-renderer": "^16.3.1",
     "rimraf": "^2.6.1",
-    "rollup": "^0.50.0",
-    "rollup-plugin-babel": "^3.0.2",
-    "rollup-plugin-commonjs": "^8.0.2",
-    "rollup-plugin-json": "^2.1.1",
-    "rollup-plugin-node-resolve": "^3.0.0",
-    "rollup-pluginutils": "^2.0.1",
+    "rollup": "^1.0.2",
```
kaelig: 1.1.0 is the latest

package.json
```diff
@@ -123,24 +130,24 @@
...
+    "rollup-plugin-babel": "^4.2.0",
```
kaelig: 4.3.0 is the latest

kaelig:
Should `"babel-runtime": "^6.23.0",` (in `dependencies`) also be bumped?

package.json
```diff
@@ -98,21 +107,19 @@
...
     "@types/lodash-decorators": "^4.0.0",
     "@types/node": "^8.10.17",
-    "@types/rollup": "0.51.4",
     "archiver": "^2.1.0",
     "aws-sdk": "^2.58.0",
-    "babel-core": "^6.26.3",
-    "babel-loader": "^7.1.5",
-    "babel-preset-shopify": "^16.2.0",
+    "babel-core": "7.0.0-bridge.0",
+    "babel-loader": "^8.0.4",
```
kaelig: Doesn't matter too much, but `version "8.0.5"` appears in yarn.lock, so we may as well update it here.

tsconfig.json
BPScott: Not a breaking change. Per the commit message:

Prior to this change TS would always output polyfills for es2017 features.
By targeting es2017 TS will leave es2017 features untouched and then
babel shall use babel-preset-env to polyfill them based upon our browser
support (as defined in browserslist in package.json)

This means that eventually if our browerslist gets to the point where everything we support doesn't need a es2017 polyfill (e.g. Object.values()) then we won't output a polyfill for it at all.

.storybook/webpack.config.js
BPScott: Per https://babeljs.io/docs/en/config-files, .babelrc is a file relative config, while the value passed into configFile is project-wide. "configFile" here does the equivalent of copying the contents of .storybook/.babelrc directly into this config object (which is what it looked like prior to this PR) so this is a chance for avoiding repition.

Originally we needed to disable .babelrc files because the root of the repo had one one that wasn't suitable for web compilation, but I was able to remove that as part of #836. So I **think** I can remove that `babelrc: false` line now.

.storybook/webpack.config.js
BPScott: Heroku doesn't break as that runs the storybook dev server (we talked about switching that to a production build yesterday).

It sill breaks the production build for the reason in the comment.

package.json
BPScott: I'm trying to match the dependeny versions that sewing-kit uses to avoid duplication of deps. Same goes for why the rollup versions are as they are.

BPScott:
Good spot babel-runtime should also be bumped

.storybook/webpack.config.js
kaelig: Ok! I was wondering if the upgrades in this PR would have magically fixed it.

PR #851, opened 2019-01-10
[RangeSlider] handle resize

### LINKED ISSUE #840: [RangeSlider] add a resize handler for track width and consolidate width lookups into method
As implemented in #784, the `DualThumb` version of range slider does not respond to the viewport resizing. This results in alignment of the thumbs and computation of the value becoming broken when the browser resizes.

We need to implement a resize listener that sets the `trackWidth` state to the value returned by a track width getter that will need to be implemented as well. Once that's in place, any other lookups of track width can be refactored to point to the tack width getter.

|before resize|after resize|
|-|-|
|screen shot 2019-01-09 at 1 23 27 pm|screen shot 2019-01-09 at 1 23 33 pm|

danrosenthal:
Closes #840 and depends on #850.


tmlayton:
So the sanitize logic can be broken by passing something like:

`<RangeSlider value={[105, 110]} />`

You would end up with 105 and 100 in state respectively.

danrosenthal:
@tmlayton I am going to work on sanitization in a separate PR. I think this contains enough good improvements we want to merge as is.

src/components/RangeSlider/utilities/index.ts
```diff
@@ -1,7 +1,7 @@
 export const CSS_VAR_PREFIX = '--Polaris-RangeSlider-';
-export const DEFAULT_RANGE_SLIDER_PROPS = {
-  min: 0,
-  max: 100,
-  step: 1,
-};
+export enum RangeSliderDefault {
```
solonaarmstrong: Just curious about the rationale behind choosing screaming snake case vs not?

danrosenthal: Took a quick look at our [naming guidelines](https://github.com/Shopify/javascript#2.2), and PascalCase is what we use for enums.

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -27,26 +30,53 @@ export interface Props extends RangeSliderProps {
...
+
 const THUMB_SIZE = 24;
 const OUTPUT_TIP_SIZE = 8;
 
 export default class DualThumb extends React.Component<Props, State> {
+  static getDerivedStateFromProps(props: Props, state: State) {
+    const {min, step, max, value} = props;
+    const {prevValue} = state;
+
+    if (isEqual(prevValue, value) === false) {
```
solonaarmstrong: I think we prefer early returns.

```suggestion
    if (isEqual(prevValue, value)) {
      return
    }
    
    return {
        prevValue: value,
        valueLower: sanitizeValueLower(value[0], min, step, value[1]),
        valueUpper: sanitizeValueUpper(value[1], max, step, value[0]),
      };
```

solonaarmstrong:
I don't think we need to fuss with this, but posting this video here so we're aware. The thumb wobbles when resizing, especially when the track is becoming wider.

range-slider-resize-wobble

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -253,7 +326,8 @@ export default class DualThumb extends React.Component<Props, State> {
   }
 
   @autobind
-  private handleMouseDownThumbUpper() {
+  private handleMouseDownThumbUpper(event: MouseEvent) {
+    if (event.button !== 0) return;
```
solonaarmstrong: `if (event.button > 0)` ? 

I'm not clear on what this is doing.

solonaarmstrong:
🎩 looks great with the example. The interaction between the thumbs and text fields work as expected. Resizing works well.

> This is okay, IMO, as most folks won't be resizing their UI all the time
This is exactly how I feel about it.

Code looks good!

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
danrosenthal: This is looking to see that the primary mouse button is the source of the click. The primary button should always return 0, so we want this specific check.

tmlayton: Yep the JS API for the mouse button event just sucks

PR #853, opened 2019-01-10
Add deprecation guidelines

tmlayton:


documentation/Deprecation guidelines.md
```diff
@@ -0,0 +1,11 @@
+# Deprecation guidelines
+
+- Ship backwards compatible changes which include supporting both the old and new API
+- Support backwards compatibility for at least half of a major release cycle, but never more that 2 major release cycles
+  - For example, before or part of 3.5, okay to remove in 4.0. After 3.5, remove in 5.0
+  - Large changes consider a full major release cycle. For example, a large change in 3.1 would be removed in 5.0. But a large change in 2.9 would never wait until 5.0 to remove
+- Use the `@deprecated` doc tag for props and components ([for example, add above where the component is defined](https://github.com/Shopify/polaris-react/blob/8e49e4c65fbbf25d40617ba2d0ff0b3747320f17/src/components/Navigation/components/UserMenu/UserMenu.tsx#L27)) and state the upgrade path
```
tmlayton:
```suggestion
- Use the `@deprecated` doc tag for props and components ([for example, add above where the component is defined](https://github.com/Shopify/polaris-react/blob/8e49e4c65fbbf25d40617ba2d0ff0b3747320f17/src/components/Navigation/components/UserMenu/UserMenu.tsx#L27)) state the reason, and upgrade path
```

documentation/Deprecation guidelines.md
```diff
@@ -0,0 +1,11 @@
...
+- Add a console warning for any respective `@deprecated` doc tag and state the upgrade path
```
tmlayton:
```suggestion
- Add a console warning for any respective `@deprecated` doc tag, state the reason, and state the upgrade path.
```

documentation/Deprecation guidelines.md
```diff
@@ -0,0 +1,11 @@
+# Deprecation guidelines
+
+- Ship backwards compatible changes which include supporting both the old and new API
+- Support backwards compatibility for at least half of a major release cycle, but never more that 2 major release cycles
```
BPScott: How do you quantify a release cycle, and thus how long is half a release cycle? AFAIK we don't have a fixed period for when we release major versions.

I think it is reasonable to expect people to upgrade to the final release of a major version, fix any deprecations and then move onwards to the next major version.

tmlayton: > How do you quantify a release cycle, and thus how long is half a release cycle

It’s not meant to be a quantified number, otherwise I would provide a formula 😄That’s also why these are guidelines and not rules – use best judgement in determining what is an appropriate amount of time for a deprecation.

PR #855, opened 2019-01-10
[Range Slider] Add example with accessibility inputs for dual thumb

solonaarmstrong:

Resolves https://github.com/Shopify/polaris-react/issues/854

The example uses a stack component, which does not fully meet the requirements of the design, but it's not possible to do in the example without CSS media queries.

I'm questioning if these really should be handled by the component. A debounce or timeout might solve the issue we were having?

screen shot 2019-01-10 at 1 12 18 pm

image

## Error state
<details>
The need to handle errors seems like an edge case, so we left it out.

image

image

**Devon Persing**
Hmm. If someone isn’t interacting directly with the slider it might weird. (Since it’s not going to work with all AT.)

What are the scenarios where an error would get triggered?

**Solona Armstrong**
I had a similar question, Devon, in terms of scenarios. I don't really know. The `type="number"` prevents characters from being typed and the `min` `max` prevent the number from exceeding its bounds. If no value is entered, the thumb value can fill it in, so I can't imagine what an error would be.

I just figured we should consider error state. 🤷‍♀️ (edited)

**Devon Persing**
Yeah, it’s a good question, but I’m having the same thoughts you were. :smile:

**Devon Persing**
If a person leaves the slider alone altogether, they would just get all results.

**Solona Armstrong**
exactly

**Solona Armstrong**
The slider can't have no values, so _something_ will be set. I can't think of how something would be entered incorrectly (without being adjusted by the component), so maybe we don't handle errors? I figured there might be an edge case I hadn't thought of that would come back as a question later, but perhaps I'm overcomplicating. (edited)

**Devon Persing**
Yeah, I think we’re okay to ship with no error states. I’m sure we’ll find a scenario in the wild, but for now…

</details>

### DIFF
--- a/src/components/RangeSlider/README.md
+++ b/src/components/RangeSlider/README.md
@@ -29,7 +29,7 @@ Range sliders should:
 - When a label is visible, it should clearly communicate the purpose of the range input and its values (min, max, step, value)
 - Be labeled as “Optional” when you need to request input that’s not required
 - Validate input as soon as merchants have finished interacting with a field (but not before)
-- Always be used with `accessibilityInputs` when range slider has dual thumbs, to provide accessible alternatives to sliding the thumbs
+- Always be used with two text field components when range slider has dual thumbs, to provide accessible alternatives to both the lower and upper thumbs
 
 ---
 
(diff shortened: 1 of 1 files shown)


solonaarmstrong:
I'm not a big fan of the amount of CSS this needs. Ideally, it would make use of media queries in order to match the designs, but whichever format I tried breaks. I'm questioning if these really should be handled by the component. A debounce or timeout might solve the issue we were having?

solonaarmstrong:
@dpersing The issues you pointed out with resizing should be solved here: https://github.com/Shopify/polaris-react/pull/851

And yes, there is a bug with the range slider updating values, which I'm also working on.

Thanks for testing it out and I'm glad it works otherwise.

danrosenthal:
#851 should also fix the values updating

solonaarmstrong:
This is rebased off of https://github.com/Shopify/polaris-react/pull/866. The extra commits will disappear once that is merged.

PR #856, opened 2019-01-11
Use the latest version of yarn to upgrade Polaris on the style guide

kaelig:

ShipIt comes with an old version of Yarn, that doesn't support integrity hashes. This means each polaris-styleguide PR coming from polaris-react deploys is deleting thousands of lines (hashes) from yarn.lock.

By leveraging a more recent version of Yarn, we know those hashes will be kept in.

---

To tophat, let's create a new Shipit stack, create a beta version, and see if the resulting PR has a yarn.lock that still contains integrity hashes.

See this PR https://github.com/Shopify/polaris-styleguide/pull/2501/files where the resulting yarn.lock still contains integrity hashes.

### DIFF
--- a/scripts/open-styleguide-pr.js
+++ b/scripts/open-styleguide-pr.js
@@ -65,7 +65,7 @@ execSync(
 );
 
 execSync(
-  `yarn upgrade @shopify/polaris@${releaseVersion.replace(
+  `npx yarn@latest upgrade @shopify/polaris@${releaseVersion.replace(
     'v',
     '',
   )} --no-progress --ignore-engines`,
--- a/scripts/open-styleguide-pr.js
+++ b/scripts/open-styleguide-pr.js
@@ -11,6 +11,7 @@ const polarisBotName = 'Shopify Polaris Bot';
 const polarisBotEmail = 'shopify-polaris-bot@users.noreply.github.com';
 const polarisBotToken = require('../secrets.json').github['shopify-polaris'];
 
+const YARN_VERSION = '10.13.0';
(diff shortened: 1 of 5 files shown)


kaelig:
Here's what this PR is attempting to fix:

screenshot_2019-01-10_16_13_16

BPScott:
`@latest` is potentially a big aggressive given that we haveto opt into new dev versions of it via dev, but I think it's better than the alternative of having up update every so often.

kaelig:
I agree it may be a bit optimistic / aggressive. Changed it to `10.13.0` instead!

scripts/open-styleguide-pr.js
```diff
@@ -11,6 +11,7 @@ const polarisBotName = 'Shopify Polaris Bot';
 const polarisBotEmail = 'shopify-polaris-bot@users.noreply.github.com';
 const polarisBotToken = require('../secrets.json').github['shopify-polaris'];
 
+const YARN_VERSION = '10.13.0';
```
tmlayton: Can we pull the version number from the yml file?

tmlayton: https://www.npmjs.com/package/js-yaml or something like this?

kaelig: Oh wait... That's the node version, not Yarn. Silly me. I'll revert that last change.

kaelig: Done in de7dbfe. Can you have a look?

dpersing:
@kaelig @BPScott @tmlayton Looks like this is ready to merge?

PR #857, opened 2019-01-11
Remove the last vestiges of Indicator from Button

BPScott:

This was dead code that would never be called

🔥 🔥 🔥


tmlayton:
For context, although it is not terribly helpful, the indicator was added in https://github.com/Shopify/polaris-react-deprecated/pull/1295 and hidden in https://github.com/Shopify/polaris-react-deprecated/pull/1544.

PR #859, opened 2019-01-11
[TopBar] Remove `title` attribute from UserMenu

### LINKED ISSUE #842: Mouseover on a UserMenu component shows 'altText'
# Issue summary

The title inside of the UserMenu is defined as `altText` https://github.com/Shopify/polaris-react/blob/master/src/components/TopBar/components/UserMenu/UserMenu.tsx#L52

When a user does a mouseover on the component, it shows `altText`

There is no props to override the value.

## Expected behavior

To not show `altText` by default

## Actual behavior

screen shot 2019-01-09 at 2 05 24 pm

1. Open a page with a TopBar that has a UserMenu (Admin)
2. Mouseover the name for a few seconds.

## Specifications

- Are you using the React components? Y
- Browser: Chrome

ry5n:
Fixes #842 by removing a `title` attribute we don’t need.

- The `title` attribute is [ineffective for people using assistive technology](https://www.24a11y.com/2017/the-trials-and-tribulations-of-the-title-attribute/)
- It seems like the content here should already be available to screen readers


dpersing:
@ry5n I think this is ready to go except for an `UNRELEASED.md` entry.

PR #860, opened 2019-01-11
[Button] keypress, keydown, and keyup events

tmlayton:

Porting over from https://github.com/Shopify/polaris-react-deprecated/pull/1820

Add events for the aforementioned.

## 

### How to 🎩

<details>

```jsx

interface State {}

  render() {
    return (
      <Page title="Playground">
        <Button onKeyPress={log} onKeyDown={log} onKeyUp={log}>
          Test
        </Button>
      </Page>
    );
  }
}

function log(event: KeyboardEvent) {
  console.log(event.keyCode);
}
```

</details>

### 🎩 checklist

* [x] 🎩 
* [ ] ~~Tested on [mobile](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting.md#cross-browser-testing)~~
* [ ] ~~Tested on [multiple browsers](https://help.shopify.com/en/manual/intro-to-shopify/shopify-admin/supported-browsers)~~

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -10,6 +10,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Added `onKeyPress`, `onKeyDown`, and `onKeyUp` to `Button` ([#860](https://github.com/Shopify/polaris-react/pull/860))
+
 ### Design updates
 
 ### Bug fixes
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -59,11 +59,11 @@ export interface Props {
   /** Callback when focus leaves button */
   onBlur?(): void;
   /** Callback when a keypress event is registered on the button */
-  onKeyPress?(): void;
+  onKeyPress?(event: React.KeyboardEvent<HTMLButtonElement>): void;
   /** Callback when a keyup event is registered on the button */
-  onKeyUp?(): void;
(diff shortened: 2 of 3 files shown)


dpersing:
I think we will want to provide some documentation around giving merchants instructions if non-standard button events are used. I can draft something as part of the accessibility guidelines for buttons ([current draft here](https://docs.google.com/document/d/1ONoa4fUsqG19i5h0h2Kz2w9CvmFSN926YmoNVMsGPgw/edit#heading=h.x10j6gpyvlsr)), but we I think it might warrant expanding that into an section on non-standard interactions on the [Interaction states page](https://polaris.shopify.com/design/interaction-states) in the style guide.

@danrosenthal @francinen What would you think about that type of guidance, and/or perhaps working on a pattern guide around events? We could use the drag-and-drop interaction as an example.

I think it would also be worthwhile to consider making a drag-and-drop component.

cc @tmlayton

francinen:
@dpersing I think that would be worth investing in. Is there a specific approach to providing that context to merchants that the documentation should endorse, e.g., using `aria-describedby` to point to a node containing text instructions for interacting with the component?

dpersing:
> using aria-describedby to point to a node containing text instructions for interacting with the component

@francinen I think having visible instructions (for sighted keyboard users) tied to the feature with `aria-describedby` or similar would be a great example. To conserve space, maybe the instructions are in a tooltip that comes before the drag-and-drop controls?

tmlayton:
> I think we will want to provide some README.md/style guide documentation around giving merchants instructions if non-standard button events are used

@dpersing do you think we should make it part of this PR?

PR #874, opened 2019-01-15
[WIP] [DatePicker] Add locale prop for Date Picker i18n

joshuajay:
** WIP - Need to add some tests **

Resolves https://github.com/Shopify/polaris-react/issues/456

Currently the `DatePicker` component is only available in English. 

Anywhere a specific date or range of dates is rendered visible (or screen readable) to the user, it is formatted in the specified locale as per the `Intl.DateTimeFormat` standard  (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat).
This includes the dates, months, years, and weekdays.

_Something to note: The weekday labels will be slightly different in English after this change, instead of the 2-letter abbreviations that we have currently, Intl defines 'short' weekday names in the 'en' locale to be 3-letters._

## 

### How to 🎩

Tophatting is as simple as adding a valid locale to the component ('ja', 'it', 'pt-BR', 'de', 'zh').

Check that:
 - everything makes sense
 - the next / previous month buttons work as expected
 - the longer weekday abbreviations still fit on small screens

<details>

```jsx
import {Page, DatePicker} from '@shopify/polaris';

interface State {
  month: number;
  year: number;
  selected: {
    start: Date,
    end: Date,
  };
}

  state = {
    month: 1,
    year: 2018,
    selected: {
      start: new Date('Wed Feb 07 2018 00:00:00 GMT-0500 (EST)'),
      end: new Date('Wed Feb 07 2018 00:00:00 GMT-0500 (EST)'),
    },
  };

  render() {
    const {month, year, selected} = this.state;

    return (
      <Page title="Playground">
        <DatePicker
          locale="fr"
          month={month}
          year={year}
          onChange={this.handleChange}
          onMonthChange={this.handleMonthChange}
          selected={selected}
        />
      </Page>
    );
  }

  handleChange = (value) => {
    this.setState({selected: value});
  };

  handleMonthChange = (month, year) => {
    this.setState({
      month,
      year,
    });
  };
}
```

</details>

### 🎩 checklist


dpersing:
Hi @joshuajay! Let us know if we can help answer any questions about feedback.

joshuajay:
Thanks @chloerice and @dpersing !

I've included the suggested CSS changes, and also figured out what was causing the re-render.
Since I changed the the way the visible month / year props are passed down to be a Date object instead, and it was re-creating a new Date object each time, the props check to determine if it should re-render was counting each date object as different, even if they were the same date.

Adding a `shouldComponentUpdate()` lifecycle method that checks the value of the date objects seems to have fixed it :)

chloerice:
> Update: scratch that last attempt, it ended up breaking more than it fixed, so instead I've opted to change the Month props back to what they were, which actually simplifies a bunch of things so it's definitely better this way 👍

Cool, will take another 👀

dpersing:
@chloerice Have you had a chance to take another look?

src/components/DatePicker/components/Day/Day.tsx
```diff
@@ -83,7 +87,7 @@ export class Day extends React.PureComponent<CombinedProps, never> {
         aria-disabled={disabled}
         role="gridcell"
       >
-        {date}
+        {Intl.DateTimeFormat(locale, {day: 'numeric'}).format(day)}
```
nimzco: Based on this Slack convo:
image

We should keep `date` for now.

src/components/DatePicker/DatePicker.tsx
```diff
@@ -84,32 +82,22 @@ export class DatePicker extends React.PureComponent<CombinedProps, State> {
       disableDatesBefore,
       disableDatesAfter,
       weekStartsOn = Weekdays.Sunday,
-      polaris: {intl},
+      locale = 'en',
```
nimzco: Can't we get the locale at the provider level or something like that? Or maybe not?

nimzco: Also, this is more for the Polaris team to answer, but is this a pattern they want to have in their library for their components?

chloerice:
> @chloerice Have you had a chance to take another look?

@dpersing Sorry I forgot my recent convos with @joshuajay were in Slack! I took another look and the popover issue mentioned in my review is still there (double render flash on initial render). I thought maybe it was a browser cache issue and quit Chrome + rebooted and I still have it.

PR #878, opened 2019-01-16
Adds example to deprecation guidelines

kvendrik:
We give a lot of instructions on how to deprecate a feature and it might be useful to provide some examples.

It adds [the deprecation of `Navigation.UserMenu`](https://github.com/Shopify/polaris-react/pull/849) as an example.

### DIFF
--- a/documentation/Deprecation guidelines.md
+++ b/documentation/Deprecation guidelines.md
@@ -11,4 +11,5 @@
 - Call out deprecations in our changelog
 
 ## Examples
+
 - [Deprecation of `Navigation.UserMenu`](https://github.com/Shopify/polaris-react/pull/849).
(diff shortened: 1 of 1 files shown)


tmlayton:
Should we use this as an example? From my understanding the upgrade path was not yet available so not sure it is the best example.

tmlayton:
Or maybe we can clarify in the first point that the full upgrade path should be available before shipping any deprecations:

```diff
- Ship backwards compatible changes which include supporting both the old and new API
+ Ship backwards compatible changes which include supporting both the old and new API (make sure the full upgrade path is available)
```

kvendrik:
> Should we use this as an example? From my understanding the upgrade path was not yet available so not sure it is the best example.

Yeah I agree it might not be the best example considering the upgrade path is provided in a separate PR. I think if we specify like you suggested that a full upgrade path should be provided before shipping this is okay to add so that devs have an example for now but can be removed later.

PR #881, opened 2019-01-16
fix safe area for mixin with fallback

danrosenthal:


danrosenthal:
> Will we be able to get this in a patch that is easy for web to consume?

This will need to be a minor release because of https://github.com/Shopify/polaris-react/pull/881/files#diff-1f9a24a2e063326eff13af43500f8b74R11 @lemonmade.

PR #901, opened 2019-01-18
[TextField] [Resizer] Stop replacing newline with line break tag 

devshahani:
The `TextField` component sizes itself incorrectly due to the way it handles some whitespace characters. 

For example, here’s some text that a user submitted in a Rails app. Notice how the text includes carriage returns (`\r`):
image

This is how the content gets rendered in a multiline `TextField`:
image

If we take out all the carriage returns, the exact same piece of text renders like this:
image

The extra height visible seems to stem from two factors: 
1. [This CSS property](https://github.com/Shopify/polaris-react/blob/master/src/components/TextField/TextField.scss#L207):
```
white-space: pre-wrap;
```
2. [This code](https://github.com/Shopify/polaris-react/blob/master/src/components/TextField/components/Resizer/Resizer.tsx#L93) in the `Resizer` component:
``` js
const ENTITIES_TO_REPLACE = {
  '&': '&amp;',
  '<': '&lt;',
  '>': '&gt;',
  '\n': '<br>',
};

const REPLACE_REGEX = /[\n&<>]/g;

function replaceEntity(entity: keyof typeof ENTITIES_TO_REPLACE) {
  return ENTITIES_TO_REPLACE[entity] || entity;
}
```
The expectation here is that each newline is converted to a `<br>` so the `DummyInput` div can correctly size itself, but somewhere in the process the remaining carriage returns are also being considered as line breaks, so the `DummyInput` div now contains one extra line for each carriage return character, resulting in the `TextField` component having additional space at the bottom unrelated to it's contents actually height.

Another solution that worked was replacing all the carriage returns with nothing, along the lines of:
```
  '\n': '<br>',
  '\r': '',
```

## 

### How to 🎩

<details>

```jsx

interface State {}

  render() {
    return (
      <Page title="Playground">
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/TextField/components/Resizer/Resizer.tsx
+++ b/src/components/TextField/components/Resizer/Resizer.tsx
@@ -90,9 +90,10 @@ const ENTITIES_TO_REPLACE = {
   '&': '&amp;',
   '<': '&lt;',
   '>': '&gt;',
+  '\n': '<br>',
 };
 
-const REPLACE_REGEX = /[&<>]/g;
+const REPLACE_REGEX = /[\n&<>]/g;
 
 function replaceEntity(entity: keyof typeof ENTITIES_TO_REPLACE) {
   return ENTITIES_TO_REPLACE[entity] || entity;
--- a/src/components/TextField/components/Resizer/Resizer.tsx
+++ b/src/components/TextField/components/Resizer/Resizer.tsx
@@ -90,10 +90,9 @@ const ENTITIES_TO_REPLACE = {
   '&': '&amp;',
   '<': '&lt;',
   '>': '&gt;',
-  '\n': '<br>',
 };
 
-const REPLACE_REGEX = /[\n&<>]/g;
+const REPLACE_REGEX = /[&<>]/g;
 
 function replaceEntity(entity: keyof typeof ENTITIES_TO_REPLACE) {
   return ENTITIES_TO_REPLACE[entity] || entity;
--- a/src/components/TextField/components/Resizer/Resizer.tsx
+++ b/src/components/TextField/components/Resizer/Resizer.tsx
@@ -94,12 +94,13 @@ const ENTITIES_TO_REPLACE = {
   '\r': '',
 };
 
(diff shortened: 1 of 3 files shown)


devshahani:
hey @tmlayton, I just updated the PR with a bunch of details. Let me know if it seems like something that should have an issue created first.

tmlayton:
@devshahani thanks for updating the pull request and sorry for the delay – I’m partial to removing the carriage returns (`\r`). Based on how the different operating systems use carriage returns and newlines, it looks like we can reliably remove the carriage returns and replace newlines with the `<br />` tag.

Because the core of this problem has nothing to do with the newline character and is a symptom of the carriage returns, I’d rather isolate the change to the carriage returns only.

For further context, it looks like we need `<br />` replacements or the measuring does not work: https://github.com/Shopify/web/pull/509/files#r109464096

src/components/TextField/components/Resizer/Resizer.tsx
```diff
@@ -90,10 +90,9 @@ const ENTITIES_TO_REPLACE = {
   '&': '&amp;',
   '<': '&lt;',
   '>': '&gt;',
-  '\n': '<br>',
```
tmlayton: I prefer your suggestion of removing `\r` carriage returns:

```
'\n': '<br>',
'\r': '',

File: src/components/TextField/components/Resizer/Resizer.tsx [hunk 2]
```diff
@@ -90,10 +90,9 @@ const ENTITIES_TO_REPLACE = {
...
 };
 
-const REPLACE_REGEX = /[\n&<>]/g;
+const REPLACE_REGEX = /[&<>]/g;
```
tmlayton:
```suggestion
const REPLACE_REGEX = /[\r\n&<>]/g;
```

tmlayton:
Thanks for this fix @devshahani! I pushed a change to get the test coverage to 100% by removing the undefined condition which would never execute. Just to be safe, now generating the regex from the keys in the entity object.

PR #902, opened 2019-01-18
Version 4.0.0

danrosenthal:


BPScott:
Things I'd like to see (a non-exhaustive list):

* Migrate legacy context API usage to the modern version (this PR starts on this already)
* Remove the dependency on prop-types (it looks like moving to the new context API removes most of our existing needs for this).
* Migrate usage of deprecated lifecycle hooks (this can be done in 3.x releases) - https://github.com/Shopify/polaris-react/issues/519
* Strict mode compliance and some way to assert this in tests/storybook (I think the removal of legacy context and deprecated lifecycle hooks are prerequisites for this) - https://github.com/Shopify/polaris-react/issues/785
* Update minimum version of react/react-dom to at least 16.8 so that we can use hooks.
* Write a withAppProvider hook and use it in place of the withAppProvider() higher-order-component. This ensures means that every component can be a function/class instead of a potentially being wrapped in a HoC. This also means we don't have to do annoying stuff with ComposedProps when building a component
* Latest typescript version (this can be done in 3.x releases, assuming https://github.com/Shopify/quilt/issues/489 is fixed) (hopefully they'll have merged a fix for https://github.com/Microsoft/TypeScript/issues/20463 by then)


* Builds solely using Babel for compilation, instead of the TS then Babel two step dance

danrosenthal:
@AndrewMusgrave, looking at the Percy report, and it appears `Banner` lost `withinContentContainer` context inside of `Navigation`, and is rendering the ordinary `Banner`. Must have lost that change during the rebase

AndrewMusgrave:
Thanks for pointing that out. I'll take a look. Just checking back at the percy report from the merge pr, it looks like it didn't fail because it was a new example and didn't have anything to compare it to. I'll keep a closer 👁 next time.

AndrewMusgrave:
The failing check is from a "flakey" test we'll need to change in the future

danrosenthal:
> AndrewMusgrave changed the title [WIP] Version 4.0.0 Version 4.0.0 23 hours ago

🎉

tmlayton:
Would love to see us get over the 90% coverage target

AndrewMusgrave:
@tmlayton 😎

danrosenthal:
I can't approve my own PR, but consider this my ✅. Nice work, @AndrewMusgrave and @BPScott!

PR #908, opened 2019-01-21
[DropZone] Constrain drop zone height based on inherited wrapper height

### LINKED ISSUE #742: DropZone height is not respected
# Issue summary

The [`<DropZone />`](https://polaris.shopify.com/components/actions/drop-zone) determines its size (height and width) based on width only. This makes certain valid use cases, where the default `DropZone` height isn't appropriate, impossible. For instance, this one:

`DropZone` is configurable to work at various sizes, however. For example:

**Default**

**50 height, 50 width**
screen shot 2018-12-12 at 2 06 50 pm

So far, so good. Until...

**50 height, unspecified width**
screen shot 2018-12-12 at 2 08 19 pm

## Expected behavior

I would expect something more like:

**50 height, unspecified width**
screen shot 2018-12-12 at 2 12 07 pm

## Actual behavior

**50 height, unspecified width**
screen shot 2018-12-12 at 2 08 19 pm

Notice that the DropZone is cut off at the bottom. That is because it actually returns to its default size (shown above).

1. Go to https://polaris.shopify.com/components/actions/drop-zone
2. Paste this code: `<div style={{height: 50}}><DropZone /></div>`

## Specifications

- Are you using the React components? (Y):
- Polaris version number: latest
- Browser: All
- Device: All
- Operating System: All

elizabethletourneau:

Resolves #742  

The `DropZone` determines its size (height and width) based on width only. This makes certain valid use cases, where the default `DropZone` height isn't appropriate, impossible. 

This pull request allows the consumer to set a fixed height on a wrapping parent element so that the `DropZone` takes its parent's height if one is provided.

Current Behaviour:  (wrapped with a div given a height of 50px)
image

New Behaviour: (wrapped with a div given a height of 50px)
image

Fixed small Height small Width:
image
Fixed Medium Height Medium Width:
image
Fixed Large Height Large Width: 
image
Fixed ExtraLarge Height ExtraLarge Width:
image
Fixed Small Width ExtraLarge Height
image
Fixed ExtraLarge Width Small Height
image

(Note: there are in between size variations (ie. small width, medium height etc.) The `FileUpload` will always render the smaller measurement of width or height.

Default Width and Heights:
image

Fixed Small Height No fixed width:

Fixed Medium Height No fixed Width: 
image

Fixed Large Height no Fixed width:
image

Fixed Extra Large Height no Fixed Width:
image

### How to 🎩

<details>
<summary>Code Example with small width, small height </summary>

```jsx

import {Page, DropZone, Stack, Thumbnail, Caption} from '../src';

interface State {
  files: object;
}

  state = {
    files: [],
  };

  render() {
    const {files} = this.state;
    const validImageTypes = ['image/gif', 'image/jpeg', 'image/png'];

      <Stack vertical>
        {files.map((file, index) => (
            <Thumbnail
              size="small"
              alt={file.name}
              source={
                validImageTypes.indexOf(file.type) > 0
                  ? window.URL.createObjectURL(file)
                  : 'https://cdn.shopify.com/s/files/1/0757/9955/files/New_Post.png?12678548500147524304'
              }
            />
            <div>
            </div>
          </Stack>
        ))}
      </Stack>
    );

    return (
      <Page title="Playground">
        <div style={{width: 50, height: 50}}>
          <DropZone
              this.setState({
              });
            }}
          >
            {uploadedFiles}
            {fileUpload}
          </DropZone>
        </div>
      </Page>
    );
  }
}



```

</details>
<details>
<summary>medium width medium height </code>:</summary>

```jsx

import {Page, DropZone, Stack, Thumbnail, Caption} from '../src';

interface State {
  files: object;
}

  state = {
    files: [],
  };

  render() {
    const {files} = this.state;
    const validImageTypes = ['image/gif', 'image/jpeg', 'image/png'];

      <Stack vertical>
        {files.map((file, index) => (
            <Thumbnail
              size="small"
              alt={file.name}
              source={
                validImageTypes.indexOf(file.type) > 0
                  ? window.URL.createObjectURL(file)
                  : 'https://cdn.shopify.com/s/files/1/0757/9955/files/New_Post.png?12678548500147524304'
              }
            />
            <div>
            </div>
          </Stack>
        ))}
      </Stack>
    );

    return (
      <Page title="Playground">
        <div style={{width: 150, height: 150}}>
          <DropZone
              this.setState({
              });
            }}
          >
            {uploadedFiles}
            {fileUpload}
          </DropZone>
        </div>
      </Page>
    );
  }
}


```

</details>
<details>
<summary>Large width, large height</summary>

```jsx

import {Page, DropZone, Stack, Thumbnail, Caption} from '../src';

interface State {
  files: object;
}

  state = {
    files: [],
  };

  render() {
    const {files} = this.state;
    const validImageTypes = ['image/gif', 'image/jpeg', 'image/png'];

      <Stack vertical>
        {files.map((file, index) => (
            <Thumbnail
              size="small"
              alt={file.name}
              source={
                validImageTypes.indexOf(file.type) > 0
                  ? window.URL.createObjectURL(file)
                  : 'https://cdn.shopify.com/s/files/1/0757/9955/files/New_Post.png?12678548500147524304'
              }
            />
            <div>
            </div>
          </Stack>
        ))}
      </Stack>
    );

    return (
      <Page title="Playground">
        <div style={{width: 250, height: 250}}>
          <DropZone
              this.setState({
              });
            }}
          >
            {uploadedFiles}
            {fileUpload}
          </DropZone>
        </div>
      </Page>
    );
  }
}


```

</details>

<details>
<summary>ExtraLarge width ExtraLarge height</summary>

```jsx

import {Page, DropZone, Stack, Thumbnail, Caption} from '../src';

interface State {
  files: object;
}

  state = {
    files: [],
  };

  render() {
    const {files} = this.state;
    const validImageTypes = ['image/gif', 'image/jpeg', 'image/png'];

      <Stack vertical>
        {files.map((file, index) => (
            <Thumbnail
              size="small"
              alt={file.name}
              source={
                validImageTypes.indexOf(file.type) > 0
                  ? window.URL.createObjectURL(file)
                  : 'https://cdn.shopify.com/s/files/1/0757/9955/files/New_Post.png?12678548500147524304'
              }
            />
            <div>
            </div>
          </Stack>
        ))}
      </Stack>
    );

    return (
      <Page title="Playground">
        <div style={{width: 500, height: 500}}>
          <DropZone
              this.setState({
              });
            }}
          >
            {uploadedFiles}
            {fileUpload}
          </DropZone>
        </div>
      </Page>
    );
  }
}


```

</details>
<details>
<summary>Default width and height</summary>

```jsx

import {Page, DropZone, Stack, Thumbnail, Caption} from '../src';

interface State {
  files: object;
}

  state = {
    files: [],
  };

  render() {
    const {files} = this.state;
    const validImageTypes = ['image/gif', 'image/jpeg', 'image/png'];

      <Stack vertical>
        {files.map((file, index) => (
            <Thumbnail
              size="small"
              alt={file.name}
              source={
                validImageTypes.indexOf(file.type) > 0
                  ? window.URL.createObjectURL(file)
                  : 'https://cdn.shopify.com/s/files/1/0757/9955/files/New_Post.png?12678548500147524304'
              }
            />
            <div>
            </div>
          </Stack>
        ))}
      </Stack>
    );

    return (
      <Page title="Playground">
        <DropZone
            this.setState({
            });
          }}
        >
          {uploadedFiles}
          {fileUpload}
        </DropZone>
      </Page>
    );
  }
}


```

</details>
<details>
<summary>Small width, Large Height</summary>

```jsx

import {Page, DropZone, Stack, Thumbnail, Caption} from '../src';

interface State {
  files: object;
}

  state = {
    files: [],
  };

  render() {
    const {files} = this.state;
    const validImageTypes = ['image/gif', 'image/jpeg', 'image/png'];

      <Stack vertical>
        {files.map((file, index) => (
            <Thumbnail
              size="small"
              alt={file.name}
              source={
                validImageTypes.indexOf(file.type) > 0
                  ? window.URL.createObjectURL(file)
                  : 'https://cdn.shopify.com/s/files/1/0757/9955/files/New_Post.png?12678548500147524304'
              }
            />
            <div>
            </div>
          </Stack>
        ))}
      </Stack>
    );

    return (
      <Page title="Playground">
        <div style={{width: 50, height: 500}}>
          <DropZone
              this.setState({
              });
            }}
          >
            {uploadedFiles}
            {fileUpload}
          </DropZone>
        </div>
      </Page>
    );
  }
}


```

</details>

<details>
<summary>Large Width Small Height</summary>

```jsx

import {Page, DropZone, Stack, Thumbnail, Caption} from '../src';

interface State {
  files: object;
}

  state = {
    files: [],
  };

  render() {
    const {files} = this.state;
    const validImageTypes = ['image/gif', 'image/jpeg', 'image/png'];

      <Stack vertical>
        {files.map((file, index) => (
            <Thumbnail
              size="small"
              alt={file.name}
              source={
                validImageTypes.indexOf(file.type) > 0
                  ? window.URL.createObjectURL(file)
                  : 'https://cdn.shopify.com/s/files/1/0757/9955/files/New_Post.png?12678548500147524304'
              }
            />
            <div>
            </div>
          </Stack>
        ))}
      </Stack>
    );

    return (
      <Page title="Playground">
        <div style={{width: 500, height: 50}}>
          <DropZone
              this.setState({
              });
            }}
          >
            {uploadedFiles}
            {fileUpload}
          </DropZone>
        </div>
      </Page>
    );
  }
}


```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/DropZone/DropZone.scss
+++ b/src/components/DropZone/DropZone.scss
@@ -15,12 +15,17 @@ $dropzone-overlay-border-color-error: color('red');
 $dropzone-overlay-background: color('indigo', 'lighter');
 $dropzone-overlay-background-error: color('red', 'lighter');
 
+.DropZoneWrapper {
+  height: inherit;
+}
+
 .DropZone {
   position: relative;
   display: flex;
   justify-content: center;
   background-color: $dropzone-background;
   border-radius: $dropzone-border-radius;
+  height: inherit;
 }
 
 .hasOutline {
@@ -34,6 +39,10 @@ $dropzone-overlay-background-error: color('red', 'lighter');
 
 .isDisabled {
   cursor: not-allowed;
+}
+
+.isMeasuring {
+  visibility: hidden;
 }
 
 .sizeExtraLarge {
@@ -57,6 +66,9 @@ $dropzone-overlay-background-error: color('red', 'lighter');
 
 .Container {
   flex: 1;
+  display: flex;
+  justify-content: center;
+  align-items: center;
 }
 
 .Overlay {
--- a/src/components/DropZone/DropZone.tsx
+++ b/src/components/DropZone/DropZone.tsx
@@ -21,7 +21,7 @@ import {FileUpload, Provider} from './components';
 import {dragDrop, alertCircle} from './icons';
 
 import {fileAccepted, getDataTransferFiles} from './utils';
-import {DropZoneContext} from './types';
+import {DropZoneContext, Size} from './types';
 
 import styles from './DropZone.scss';
 
@@ -29,13 +29,15 @@ export type Type = 'file' | 'image';
 
 export interface State {
   id: string;
-  size: string;
+  height: Size;
+  width: Size;
   type?: string;
   error?: boolean;
   dragging: boolean;
   overlayText?: string;
   errorOverlayText?: string;
   numFiles: number;
+  measuring: boolean;
 }
 
 export interface Props {
@@ -108,7 +110,6 @@ export interface Props {
 export type CombinedProps = Props & WithAppProviderProps;
 
 const getUniqueID = createUniqueIDFactory('DropZone');
-
 export class DropZone extends React.Component<CombinedProps, State> {
   public static FileUpload: typeof FileUpload = FileUpload;
   public static defaultProps: Partial<CombinedProps> = {
@@ -170,19 +171,23 @@ export class DropZone extends React.Component<CombinedProps, State> {
     this.state = {
       type,
       id: props.id || getUniqueID(),
-      size: 'extraLarge',
+      height: Size.ExtraLarge,
       dragging: false,
       error: false,
       overlayText: translate(`Polaris.DropZone.overlayText${suffix}`),
       errorOverlayText: translate(`Polaris.DropZone.errorOverlayText${suffix}`),
       numFiles: 0,
+      width: Size.ExtraLarge,
+      measuring: true,
     };
   }
 
   get getContext(): DropZoneContext {
+    const {width, height, type = 'file'} = this.state;
     return {
-      size: this.state.size,
-      type: this.state.type || 'file',
+      width,
+      height,
+      type,
     };
   }
 
@@ -191,9 +196,11 @@ export class DropZone extends React.Component<CombinedProps, State> {
       id,
       dragging,
       error,
-      size,
+      height,
+      width,
       overlayText,
       errorOverlayText,
+      measuring,
     } = this.state;
     const {
       label,
@@ -223,11 +230,28 @@ export class DropZone extends React.Component<CombinedProps, State> {
       styles.DropZone,
       outline && styles.hasOutline,
       (active || dragging) && styles.isDragging,
+      measuring && styles.isMeasuring,
       error && styles.hasError,
-      size && size === 'extraLarge' && styles.sizeExtraLarge,
-      size && size === 'large' && styles.sizeLarge,
-      size && size === 'medium' && styles.sizeMedium,
-      size && size === 'small' && styles.sizeSmall,
+      height === Size.ExtraLarge && styles.sizeExtraLarge,
+      height === Size.Large && styles.sizeLarge,
+      height === Size.Medium && styles.sizeMedium,
+      height === Size.Small && styles.sizeSmall,
+    );
+
+    const extraLargeDropZoneSize =
+      width === Size.ExtraLarge && height !== Size.Small;
+
+    const mediumLargeDropZoneSize =
+      (width === Size.Medium || width === Size.Large) && height !== Size.Small;
+
+    const dragOverlayDisplayText = extraLargeDropZoneSize && (
+      <DisplayText size="small" element="p">
+        {overlayText}
+      </DisplayText>
+    );
+
+    const dragOverlayCaption = mediumLargeDropZoneSize && (
+      <Caption>{overlayText}</Caption>
     );
 
     const dragOverlay =
@@ -235,49 +259,48 @@ export class DropZone extends React.Component<CombinedProps, State> {
         <div className={styles.Overlay}>
           <Stack vertical spacing="tight">
             <Icon source={dragDrop} color="indigo" />
-            {size === 'extraLarge' && (
-              <DisplayText size="small" element="p">
-                {overlayText}
-              </DisplayText>
-            )}
-            {(size === 'medium' || size === 'large') && (
-              <Caption>{overlayText}</Caption>
-            )}
+            {dragOverlayDisplayText}
+            {dragOverlayCaption}
           </Stack>
         </div>
       ) : null;
+
+    const dragErrorOverlayDisplayText = extraLargeDropZoneSize && (
+      <DisplayText size="small" element="p">
+        {errorOverlayText}
+      </DisplayText>
+    );
+
+    const dragErrorOverlayCaption = mediumLargeDropZoneSize && (
+      <Caption>{errorOverlayText}</Caption>
+    );
 
     const dragErrorOverlay =
       dragging && error ? (
         <div className={styles.Overlay}>
           <Stack vertical spacing="tight">
             <Icon source={alertCircle} color="red" />
-            {size === 'extraLarge' && (
-              <DisplayText size="small" element="p">
-                {errorOverlayText}
-              </DisplayText>
-            )}
-            {(size === 'medium' || size === 'large') && (
-              <Caption>{errorOverlayText}</Caption>
-            )}
+            {dragErrorOverlayDisplayText}
+            {dragErrorOverlayCaption}
           </Stack>
         </div>
       ) : null;
 
     const dropZoneMarkup = (
-      <div
-        ref={this.setNode}
-        className={classes}
-        aria-disabled={disabled}
-        onClick={this.handleClick}
-        onDragStart={handleDragStart}
-      >
-        {dragOverlay}
-        {dragErrorOverlay}
-        <div className={styles.Container}>{children}</div>
-        <VisuallyHidden>
-          <input {...inputAttributes} />
-        </VisuallyHidden>
+      <div ref={this.setNode} className={styles.DropZoneWrapper}>
+        <div
+          className={classes}
+          aria-disabled={disabled}
+          onClick={this.handleClick}
+          onDragStart={handleDragStart}
+        >
+          {dragOverlay}
+          {dragErrorOverlay}
+          <div className={styles.Container}>{children}</div>
+          <VisuallyHidden>
+            <input {...inputAttributes} />
+          </VisuallyHidden>
+        </div>
       </div>
     );
 
@@ -356,24 +379,34 @@ export class DropZone extends React.Component<CombinedProps, State> {
   }
 
   @autobind
+  private setWrapperSize(size: string, node: HTMLElement) {
+    let wrapperSize;
+    const getSize = size === 'height' ? 'height' : 'width';
+    const wrapper = node.getBoundingClientRect()[getSize];
+
+    if (wrapper < Size.Small) {
+      wrapperSize = Size.Small;
+    } else if (wrapper < Size.Medium) {
+      wrapperSize = Size.Medium;
+    } else if (wrapper < Size.Large) {
+      wrapperSize = Size.Large;
+    } else {
+      wrapperSize = Size.ExtraLarge;
+    }
+    return wrapperSize;
+  }
+
+  @autobind
   @debounce(50, {trailing: true})
   private adjustSize() {
     if (!this.node) {
       return;
     }
 
-    let size = 'extraLarge';
-    const width = this.node.getBoundingClientRect().width;
+    const height = this.setWrapperSize('height', this.node);
+    const width = this.setWrapperSize('width', this.node);
 
-    if (width < 100) {
-      size = 'small';
-    } else if (width < 160) {
-      size = 'medium';
-    } else if (width < 300) {
-      size = 'large';
-    }
-
-    this.setState({size});
+    this.setState({height, width, measuring: false});
   }
 
   @autobind
--- a/src/components/DropZone/components/FileUpload/FileUpload.tsx
+++ b/src/components/DropZone/components/FileUpload/FileUpload.tsx
@@ -15,7 +15,7 @@ import TextStyle from '../../../TextStyle';
 import withContext from '../../../WithContext';
 import withRef from '../../../WithRef';
 
-import {DropZoneContext} from '../../types';
+import {DropZoneContext, Size} from '../../types';
 import {dragDrop} from '../../icons';
 import {fileUpload, imageUpload} from '../../images';
 
@@ -79,71 +79,72 @@ export class FileUpload extends React.Component<CombinedProps, State> {
   }
 
   render() {
+    const fileUploadMarkup = this.renderFileUploadMarkup();
+
+    return <div className={styles.FileUpload}>{fileUploadMarkup}</div>;
+  }
+
+  private renderFileUploadMarkup() {
     const {
-      context: {size, type},
+      context: {width, height, type},
     } = this.props;
+
     const {actionTitle, actionHint} = this.state;
     const imageClasses = classNames(
       styles.Image,
-      size && size === 'extraLarge' && styles.sizeExtraLarge,
-      size && size === 'large' && styles.sizeLarge,
+      width === Size.ExtraLarge && styles.sizeExtraLarge,
+      width === Size.Large && styles.sizeLarge,
     );
 
-    const extraLargeView =
-      size === 'extraLarge' ? (
-        <Stack vertical>
-          {type === 'file' && (
-            <img className={imageClasses} src={fileUpload} alt="" />
-          )}
-          {type === 'image' && (
-            <img className={imageClasses} src={imageUpload} alt="" />
-          )}
-          <Button>{actionTitle}</Button>
-          <TextStyle variation="subdued">{actionHint}</TextStyle>
-        </Stack>
-      ) : null;
+    const fileType = type === 'file';
+    const imageType = type === 'image';
 
-    const largeView =
-      size === 'large' ? (
-        <Stack vertical spacing="tight">
-          {type === 'file' && (
-            <img className={imageClasses} src={fileUpload} alt="" />
-          )}
-          {type === 'image' && (
-            <img className={imageClasses} src={imageUpload} alt="" />
-          )}
-          <Button size="slim">{actionTitle}</Button>
-          <Caption>
+    const size = Math.min(width, height);
+    switch (size) {
+      case Size.ExtraLarge:
+        return (
+          <Stack vertical>
+            {fileType && (
+              <img className={imageClasses} src={fileUpload} alt="" />
+            )}
+            {imageType && (
+              <img className={imageClasses} src={imageUpload} alt="" />
+            )}
+            <Button>{actionTitle}</Button>
             <TextStyle variation="subdued">{actionHint}</TextStyle>
-          </Caption>
-        </Stack>
-      ) : null;
-
-    const mediumView =
-      size === 'medium' ? (
-        <Stack vertical spacing="tight">
-          <Link>{actionTitle}</Link>
-          <Caption>
-            <TextStyle variation="subdued">{actionHint}</TextStyle>
-          </Caption>
-        </Stack>
-      ) : null;
-
-    const smallView =
-      size === 'small' ? (
-        <Stack vertical spacing="tight">
-          <Icon source={dragDrop} color="inkLightest" />
-        </Stack>
-      ) : null;
-
-    return (
-      <div className={styles.FileUpload}>
-        {smallView}
-        {mediumView}
-        {largeView}
-        {extraLargeView}
-      </div>
-    );
+          </Stack>
+        );
+      case Size.Large:
+        return (
+          <Stack vertical spacing="tight">
+            {fileType && (
+              <img className={imageClasses} src={fileUpload} alt="" />
+            )}
+            {imageType && (
+              <img className={imageClasses} src={imageUpload} alt="" />
+            )}
+            <Button size="slim">{actionTitle}</Button>
+            <Caption>
+              <TextStyle variation="subdued">{actionHint}</TextStyle>
+            </Caption>
+          </Stack>
+        );
+      case Size.Medium:
+        return (
+          <Stack vertical spacing="tight">
+            <Link>{actionTitle}</Link>
+            <Caption>
+              <TextStyle variation="subdued">{actionHint}</TextStyle>
+            </Caption>
+          </Stack>
+        );
+      case Size.Small:
+        return (
+          <Stack vertical spacing="tight">
+            <Icon source={dragDrop} color="inkLightest" />
+          </Stack>
+        );
+    }
   }
 }
 
--- a/src/components/DropZone/components/FileUpload/tests/FileUpload.test.tsx
+++ b/src/components/DropZone/components/FileUpload/tests/FileUpload.test.tsx
@@ -3,11 +3,12 @@ import {Link, Icon, Button, Caption, TextStyle} from 'components';
 import {mountWithAppProvider} from 'test-utilities';
 import {Provider} from '../../Context';
 import FileUpload from '../FileUpload';
+import {Size} from '../../../types';
 
 describe('<FileUpload />', () => {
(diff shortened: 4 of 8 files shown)


src/components/DropZone/DropZone.scss
```diff
@@ -15,12 +15,17 @@ $dropzone-overlay-border-color-error: color('red');
...
+  height: inherit;
+}
+
 .DropZone {
   position: relative;
   display: flex;
   justify-content: center;
   background-color: $dropzone-background;
   border-radius: $dropzone-border-radius;
+  height: inherit;
```
elizabethletourneau: This will allow for a fixed height above 300px on the `DropZone` container to be set by the consumer. I'm open for discussion if we should put a constraint on the maximum height the `DropZone` container should be.

dleroux: I don't there needs to be a maxheight

src/components/DropZone/DropZone.tsx
```diff
@@ -363,18 +367,25 @@ export class DropZone extends React.Component<CombinedProps, State> {
...
-    } else if (width < 300) {
-      size = 'large';
+    let height = Size.ExtraLarge;
+    const wrapperHeight = this.node.getBoundingClientRect().height;
+    if (wrapperHeight < Size.Small) {
+      height = Size.Small;
+    } else if (wrapperHeight < Size.Medium) {
+      height = Size.Medium;
+    } else if (wrapperHeight < Size.Large) {
+      height = Size.Large;
```
dleroux: This is fine but you could: 

`let height` and then add the Size.ExtraLarge to Size.ExtraLarge; as oppose to mutating height. Same with width below.

src/components/DropZone/DropZone.tsx
```diff
@@ -22,15 +22,16 @@ import IconDragDrop from './icons/drag-drop.svg';
...
+import {DropZoneContext, Size} from './types';
 
 import * as styles from './DropZone.scss';
 
 export type Type = 'file' | 'image';
 
 export interface State {
   id: string;
-  size: string;
+  height: number;
```
dleroux:
```suggestion
  height: Size;
```

src/components/DropZone/DropZone.tsx
```diff
@@ -22,15 +22,16 @@ import IconDragDrop from './icons/drag-drop.svg';
...
+  width: number;
```
dleroux:
```suggestion
  width: Size;
```

src/components/DropZone/components/FileUpload/FileUpload.tsx
```diff
@@ -80,71 +80,73 @@ export class FileUpload extends React.Component<CombinedProps, State> {
   }
 
   render() {
+    const fileUploadMarkup = this.renderFileUploadMarkup();
```
dleroux: 👍

src/components/DropZone/types.ts
```diff
@@ -1,6 +1,13 @@
...
 
+export enum Size {
+  ExtraLarge = 500,
+  Large = 300,
+  Medium = 160,
+  Small = 100,
+}
 export interface DropZoneContext {
-  size: string;
+  width: number;
```
dleroux:
```suggestion
  width: Size;
```

src/components/DropZone/types.ts
```diff
@@ -1,6 +1,13 @@
...
+  height: number;
```
dleroux:
```suggestion
  height: Size;
```

dleroux:
Code looks good! I left a few comments. There's one thing I noticed while tophatting. The default loads and then get replaced by the correct size, which I find was distracting. I wonder if it would be better to not render anything until we know what to render. What do you think? Maybe a designer can weigh in here @sarahill ?

dropzone

src/components/DropZone/DropZone.tsx
```diff
@@ -224,26 +230,27 @@ export class DropZone extends React.Component<CombinedProps, State> {
       styles.DropZone,
       outline && styles.hasOutline,
       (active || dragging) && styles.isDragging,
+      loading && styles.isLoading,
       error && styles.hasError,
-      size && size === 'extraLarge' && styles.sizeExtraLarge,
-      size && size === 'large' && styles.sizeLarge,
-      size && size === 'medium' && styles.sizeMedium,
-      size && size === 'small' && styles.sizeSmall,
+      height && height === Size.ExtraLarge && styles.sizeExtraLarge,
```
dleroux: 👍

src/components/DropZone/DropZone.tsx
```diff
@@ -224,26 +230,27 @@ export class DropZone extends React.Component<CombinedProps, State> {
...
             <Icon source={IconDragDrop} color="indigo" />
-            {size === 'extraLarge' && (
-              <DisplayText size="small" element="p">
-                {overlayText}
-              </DisplayText>
-            )}
-            {(size === 'medium' || size === 'large') && (
-              <Caption>{overlayText}</Caption>
-            )}
+            {width === Size.ExtraLarge &&
```
dleroux: not necessary but I would have probably pulled this logic out into it's own variable.

src/components/DropZone/DropZone.tsx
```diff
@@ -253,32 +260,33 @@ export class DropZone extends React.Component<CombinedProps, State> {
...
             <Icon source={IconAlertCircle} color="red" />
-            {size === 'extraLarge' && (
-              <DisplayText size="small" element="p">
-                {errorOverlayText}
-              </DisplayText>
-            )}
-            {(size === 'medium' || size === 'large') && (
-              <Caption>{errorOverlayText}</Caption>
-            )}
+            {width === Size.ExtraLarge &&
```
dleroux: same here

dleroux:
It's looking good! what will be the impact for people already using the dropzone you think?

src/components/DropZone/DropZone.tsx
```diff
@@ -22,21 +22,23 @@ import IconDragDrop from './icons/drag-drop.svg';
...
-  size: string;
+  height: Size;
+  width: Size;
   type?: string;
   error?: boolean;
   dragging: boolean;
   overlayText?: string;
   errorOverlayText?: string;
   numFiles: number;
+  loading: boolean;
```
solonaarmstrong: Is it possible to solve the flashing problem without adding state to the component?

src/components/DropZone/components/FileUpload/FileUpload.tsx
```diff
@@ -80,71 +80,73 @@ export class FileUpload extends React.Component<CombinedProps, State> {
...
-          {type === 'file' && (
-            <img className={imageClasses} src={AssetFileUpload} alt="" />
-          )}
-          {type === 'image' && (
-            <img className={imageClasses} src={AssetImageUpload} alt="" />
-          )}
-          <Button size="slim">{actionTitle}</Button>
-          <Caption>
+    const size = Math.min(width, height);
+    switch (size) {
```
solonaarmstrong: Just curious why you used a switch case here?

src/components/DropZone/components/FileUpload/tests/FileUpload.test.tsx
```diff
@@ -30,11 +31,20 @@ describe('<FileUpload />', () => {
...
-      <Provider value={{size: 'small', type: 'file'}}>
+      <Provider value={{width: Size.Small, height: Size.Small, type: 'file'}}>
         <FileUpload />
       </Provider>,
     );
 
     expect(fileUpload.find(Icon)).toHaveLength(1);
   });
+
+  it('renders the smaller view when width is small but height is big', () => {
```
solonaarmstrong:
```suggestion
  it('renders the smaller view when width is small but height is large', () => {
```

dleroux:
Is it me of the text not centered on smaller screens? Could be something with Stack?

screen shot 2019-02-08 at 10 27 37 am

src/components/DropZone/components/FileUpload/FileUpload.tsx
dleroux: I think it makes sense, the 4 layouts are different enough. I'd probably leave the Size.Small as default just in case.

src/components/DropZone/DropZone.tsx
AndrewMusgrave: @solona what is the flashing problem?

src/components/DropZone/DropZone.tsx
```diff
@@ -171,18 +172,21 @@ export class DropZone extends React.Component<CombinedProps, State> {
...
       numFiles: 0,
+      width: Size.ExtraLarge,
+      loading: true,
     };
   }
 
   get getContext(): DropZoneContext {
     return {
-      size: this.state.size,
+      width: this.state.width,
```
AndrewMusgrave: Since we're using state three times in this function we could deconstruct and use that sweet sweet es6 object syntax
```tsx
const {width, height, type = 'file'} = this.state;

return {width, height, type};
```

src/components/DropZone/DropZone.tsx
AndrewMusgrave: I don't think we need `height &&`, height will always be truthy (I think?) but it was undefined we still should be able to compare it the enum

src/components/DropZone/DropZone.tsx
AndrewMusgrave: Maybe a switch statement to determine the markup rendered?

src/components/DropZone/DropZone.tsx
```diff
@@ -363,18 +371,29 @@ export class DropZone extends React.Component<CombinedProps, State> {
...
+    } else if (wrapperHeight < Size.Medium) {
+      height = Size.Medium;
+    } else if (wrapperHeight < Size.Large) {
+      height = Size.Large;
+    } else {
+      height = Size.ExtraLarge;
     }
-
-    this.setState({size});
+    const wrapperWidth = this.node.getBoundingClientRect().width;
```
AndrewMusgrave: If we only need the width, should we use `this.node.offsetWidth`?

src/components/DropZone/components/FileUpload/FileUpload.tsx
```diff
@@ -80,71 +80,73 @@ export class FileUpload extends React.Component<CombinedProps, State> {
...
-      context: {size, type},
+      context: {width, height, type},
     } = this.props;
+
     const {actionTitle, actionHint} = this.state;
     const imageClasses = classNames(
       styles.Image,
-      size && size === 'extraLarge' && styles.sizeExtraLarge,
-      size && size === 'large' && styles.sizeLarge,
+      width && width === Size.ExtraLarge && styles.sizeExtraLarge,
```
AndrewMusgrave: Same as with `height &&` above

src/components/DropZone/components/FileUpload/FileUpload.tsx
AndrewMusgrave: We don't need to encase the switch statement in a block 😄. To Dan we should be ok the way width and height are typed, if we wanted to be safe we could always create a min function that returns a type safe variable. Or return null by default? 🤷‍♂️

src/components/DropZone/DropZone.tsx
elizabethletourneau: I removed the variable sections out. I think a switch statement would repeat to much logic. I made the caption its own conditional variable and the Display Text its own conditional variable since the Icon itself will always render.

src/components/DropZone/DropZone.tsx
elizabethletourneau: It's Dan's gif  [above](https://github.com/Shopify/polaris-react/pull/908#issuecomment-458548206).  Is there another example @solona I could take a look at?

elizabethletourneau:
> Is it me of the text not centered on smaller screens? Could be something with Stack?
> 
> screen shot 2019-02-08 at 10 27 37 am

@daniedleroux  I measured it and they seem to be identical on both sides. Tried a bunch of difference sizes. Do you remember what size you tried it on?

elizabethletourneau:
> It's looking good! what will be the impact for people already using the dropzone you think?

The functionality of Dropzone is essentially the same. It now handles the case where when a fixed height is given without a fixed width, dropzone is not getting re-rendered because of its width. (ie. small height large width is not readjusting to large height large width). I can't think of a case where this would be a breaking change.

elizabethletourneau:
If anyone is reviewing please hold off. ;)

src/components/DropZone/DropZone.tsx
elizabethletourneau: The tests, she didn't like this suggestion.

dleroux:
this size:

image

the 2 `o's` on the second line should have the same distance to the edge. I think it's 8px difference. -1.6rem as oppose to -0.8rem (in that screen it's 24 on one side and 32px on the other)

elizabethletourneau:
> this size:
> 
> image
> 
> the 2 `o's` on the second line should have the same distance to the edge. I think it's 8px difference. -1.6rem as oppose to -0.8rem (in that screen it's 24 on one side and 32px on the other)

Dan and I are not seeing the same thing so we are going to look at this together on Monday.

src/components/DropZone/DropZone.tsx
dleroux: This measuring happens only initial load. So we could use componentDidMount and forceUpdate() on a private variable but since I think since this change only occurs once it's probably more performant as state.

I'd probably stick to `measuring` as oppose to `loading` for the state key, a little more clear. What do you think?

dleroux:
It looks like the 8px off is a bug in Chrome 72. So maybe just open an issue for Stack and steps to reproduce.

solonaarmstrong:
> @solona what is the flashing problem?

@AndrewMusgrave I'm pretty sure it was Dan that pointed it out, but the issue was that the size defaulted to ExtraLarge and then readjusted, based on the DOM, so it initially flashed ExtraLarge and then snapped to its new size.

src/components/DropZone/tests/DropZone.test.tsx
```diff
@@ -244,7 +244,7 @@ describe('<DropZone />', () => {
       expect(caption).toHaveLength(0);
     });
 
-    it('renders a Caption containing the overlayText on medium screens', () => {
+    it('renders a Caption containing the overlayText when the bounding container is a small size', () => {
```
solonaarmstrong:
```suggestion
    it('renders a Caption containing the overlayText when the bounding container is a medium size', () => {
```

src/components/DropZone/tests/DropZone.test.tsx
```diff
@@ -264,7 +264,7 @@ describe('<DropZone />', () => {
       expect(captionText.contains(overlayText)).toBe(true);
     });
 
-    it('renders a DisplayText containing the overlayText on extra-large screens', () => {
+    it('renders a DisplayText containing the overlayText when the bounding container is an extra large size', () => {
       setBoundingClientRect('extraLarge');
```
solonaarmstrong: Should these all be using the enum here?

AndrewMusgrave: +1

solonaarmstrong:
```
Element.prototype.getBoundingClientRect;
const widths = {	const widths = {
  small: 99,	  small: 99,
  medium: 159,	  medium: 159,
  large: 299,	  large: 299,
  extraLarge: 1024,	  extraLarge: 1024,
};	};
```

These don't quite correspond.

elizabethletourneau: I'm not sure an enum would work here unless I created a new one and not import the `Size` from types. These sizes (99, 159, 299) are all less than the bounding size which renders which overlay gets shown however the Size enum used in DropZone is to determine the breaking point when they change (100, 160, 300) by comparing values less then them. (The extralarge doesn't matter really what number it is because its the default). I might be misinterpreting the goal here though. Can you clarify if I should be creating a new enum or using the existing one?   For example replace `small: 99` with `small: 100` and you see the tests fail.

solonaarmstrong: I would create a new enum.

src/components/DropZone/DropZone.tsx
```diff
@@ -224,61 +231,73 @@ export class DropZone extends React.Component<CombinedProps, State> {
...
-      size && size === 'large' && styles.sizeLarge,
-      size && size === 'medium' && styles.sizeMedium,
-      size && size === 'small' && styles.sizeSmall,
+      height === Size.ExtraLarge && styles.sizeExtraLarge,
+      height === Size.Large && styles.sizeLarge,
+      height === Size.Medium && styles.sizeMedium,
+      height === Size.Small && styles.sizeSmall,
     );
 
+    const dragOverlayDisplayText = width === Size.ExtraLarge &&
```
AndrewMusgrave: Should we save the condition into a `const`? it seems we reuse it.

src/components/DropZone/DropZone.tsx
```diff
@@ -224,61 +231,73 @@ export class DropZone extends React.Component<CombinedProps, State> {
...
+      height !== Size.Small && (
+        <DisplayText size="small" element="p">
+          {overlayText}
+        </DisplayText>
+      );
+
+    const dragOverlayCaption = (width === Size.Medium ||
```
AndrewMusgrave: Maybe we should save them all to const so they can have specific names, since we don't add comments it maybe easier for people touching the code later to quickly see what's being done.

src/components/DropZone/DropZone.tsx
```diff
@@ -363,18 +382,29 @@ export class DropZone extends React.Component<CombinedProps, State> {
...
-    if (width < 100) {
-      size = 'small';
-    } else if (width < 160) {
-      size = 'medium';
-    } else if (width < 300) {
-      size = 'large';
+    let height;
+    let width;
+    const wrapperHeight = this.node.getBoundingClientRect().height;
+    if (wrapperHeight < Size.Small) {
```
AndrewMusgrave: Since this if statement and the below if statement are the same we could turn it into a function to reduce code, but I don't have strong feelings about this since. So whatever you feel is better is 👌

src/components/DropZone/components/FileUpload/FileUpload.tsx
```diff
@@ -80,71 +80,69 @@ export class FileUpload extends React.Component<CombinedProps, State> {
...
-            <img className={imageClasses} src={AssetImageUpload} alt="" />
-          )}
-          <Button size="slim">{actionTitle}</Button>
-          <Caption>
+    const size = Math.min(width, height);
+    switch (size) {
+      case Size.ExtraLarge:
+        return (
+          <Stack vertical>
+            {type === 'file' && (
```
AndrewMusgrave: Should we bring these two img elements into a `const` for reuse here and below?

src/components/DropZone/tests/DropZone.test.tsx
```diff
@@ -330,11 +331,11 @@ function createEvent(name: string, files: any) {
...
 }
 
-function setBoundingClientRect(size: keyof typeof widths) {
+function setBoundingClientRect(size: Width) {
   Element.prototype.getBoundingClientRect = jest.fn(() => {
     return {
-      width: widths[size],
-      height: 100,
+      width: size,
+      height: size,
```
solonaarmstrong: It seems weird to use a Width enum as the type of size, which is then used for height.

src/components/DropZone/types.ts
```diff
@@ -1,6 +1,13 @@
 export type DropZoneEvent = DragEvent | React.ChangeEvent<HTMLInputElement>;
 
+export enum Size {
+  ExtraLarge = 500,
+  Large = 300,
+  Medium = 160,
+  Small = 100,
+}
```
solonaarmstrong: Nit: the other enum goes from small to extra large, so for someone reading through this, it could be a bit more difficult to compare them.

src/components/DropZone/DropZone.tsx
elizabethletourneau: I like that idea. I tried out a function.  Let me know what you think.

src/components/DropZone/tests/DropZone.test.tsx
```diff
@@ -28,12 +28,13 @@ const duplicateFiles = [
...
 const rejectedFiles = [files[1]];
 const origGetBoundingClientRect = Element.prototype.getBoundingClientRect;
-const widths = {
-  small: 99,
-  medium: 159,
-  large: 299,
-  extraLarge: 1024,
-};
+
+enum Size {
```
solonaarmstrong:
```suggestion
enum TestSize {
```

?

PR #912, opened 2019-01-21
Run rollup once, to generate both cjs and es-modules bundles

BPScott:

Gotta go fast.

Previously we ran rollup for twice, but as the only thing that differs
is the output format that's not needed.

This reduces the build time by ~5.5 seconds (~15%) which is particularly nice if you spend time waiting for `build-consumer` to run or waiting for `build.test.js` to complete when running the test suite.

Runs rollup() once, then generate both the cjs and the es-modules files from that single run.

As we're only running rollup() once we can also remove a bunch of conditionals from the config generation leading to simpler configuration

### How to 🎩

- On master, create a build output: `git checkout master && mkdir master-build/node_modules/@shopify && yarn run build-consumer polaris-react/master-build`
- On this branch create a build output: `git checkout run-rollup-once && mkdir new-build/node_modules/@shopify && yarn run build-consumer polaris-react/new-build`
- Compare the outputs and ensure they are the same: `diff -r -u master-build new-build`

### DIFF
--- a/config/rollup/index.js
+++ b/config/rollup/index.js
@@ -1,5 +1,4 @@
 const {resolve} = require('path');
-const {readJSONSync} = require('fs-extra');
 const nodeResolve = require('rollup-plugin-node-resolve');
 const babel = require('rollup-plugin-babel');
 const json = require('rollup-plugin-json');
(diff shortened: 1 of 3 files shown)


BPScott:
> Can you run a diff of the master build and this branch's build?

That's what the tophat instructions do. No changes occur.

tmlayton:
> Gotta go fast.

Also, 
image

PR #919, opened 2019-01-22
[Toast] Add `action` prop

### LINKED ISSUE #553: Action prop for `Toast`
# Feature request summary

Add an `Action` prop for `Toast` component.

## Rationale

In the `online-store` rebuild, we have two use cases for that, one is a `Toast` for no internet connection, it would be useful to have an action `Retry` on this `Toast`.

Another use case is still on the rebuild, but in `theme-editor` where you can `Undo` actions on the success `Toast`.

---

MateusJabour:

Resolves #553

As highlight on this comment https://github.com/Shopify/polaris-react/issues/553#issuecomment-436714929, there are some use cases for the action inside `Toast` in `online-store-web` project. 

Adds the action prop to `Toast`, shows a `<Button />` when `action` is defined. Follows designs from #553. Adds clarification in the docs about how to use it. As far as app-bridge, for now, there will be no action.

### How to 🎩

<details>

```jsx
import {Page, Frame, Toast, Button, Pagination, Key} from '../src';

interface State {
  showToast: boolean;
}

  state = {
    showToast: false,
  };

  toggleToast = () => {
    this.setState(({showToast}) => ({showToast: !showToast}));
  };

  render() {
    const {showToast} = this.state;
    const toastMarkup = showToast ? (
      <Toast
        content="Image deleted"
        // duration={10000}
        onDismiss={this.toggleToast}
        action={{content: 'Retry'}}
      />
    ) : null;
    return (
      <Page title="Playground">
        <Frame>
          <Button onClick={this.toggleToast}>Show Toast</Button>
          {toastMarkup}
        </Frame>
      </Page>
    );
  }
}

```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -14,6 +14,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - Updated `Navigation` badge prop to accept a react node ([#1142](https://github.com/Shopify/polaris-react/pull/1142))
 - Changed max width on `Search` to 694px so that it is perfectly centered in the top bar ([#1107](https://github.com/Shopify/polaris-react/issues/1107))
+- Added `action` prop to `Toast` ([#919](https://github.com/Shopify/polaris-react/pull/919))
 
 ### Bug fixes
 
--- a/src/components/Frame/components/Toast/Toast.scss
+++ b/src/components/Frame/components/Toast/Toast.scss
@@ -4,7 +4,7 @@ $Backdrop-opacity: 0.88;
 .Toast {
   @include text-style-display-small;
   display: inline-flex;
-  max-width: rem(400px);
+  max-width: rem(500px);
   padding: spacing(tight) spacing();
   border-radius: border-radius();
   background: rgba(color('black'), $Backdrop-opacity);
@@ -15,6 +15,11 @@ $Backdrop-opacity: 0.88;
   @include breakpoint-after($breakpoint) {
     padding: spacing();
   }
+}
+
+.Action {
+  margin-left: spacing(extra-loose);
+  margin-right: spacing();
 }
 
 .error {
--- a/src/components/Frame/components/Toast/Toast.tsx
+++ b/src/components/Frame/components/Toast/Toast.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import {classNames} from '@shopify/react-utilities';
 
 import {Key} from '../../../../types';
+import Button from '../../../Button';
 
 import Icon from '../../../Icon';
 import KeypressListener from '../../../KeypressListener';
@@ -10,6 +11,7 @@ import {ToastProps as Props} from '../../types';
 import styles from './Toast.scss';
 
 export const DEFAULT_TOAST_DURATION = 5000;
+export const DEFAULT_TOAST_DURATION_WITH_ACTION = 10000;
 
 export default class Toast extends React.Component<Props, never> {
   private timer?: number;
@@ -27,13 +29,26 @@ export default class Toast extends React.Component<Props, never> {
   }
 
   render() {
-    const {content, onDismiss, error} = this.props;
+    const {content, onDismiss, error, action} = this.props;
 
     const dismissMarkup = (
-      <button type="button" className={styles.CloseButton} onClick={onDismiss}>
+      <button
+        type="button"
+        className={styles.CloseButton}
+        onClick={onDismiss}
+        testID="closeButton"
+      >
         <Icon source="cancel" />
       </button>
     );
+
+    const actionMarkup = action ? (
+      <div className={styles.Action}>
+        <Button plain monochrome onClick={action.onAction}>
+          {action.content}
+        </Button>
+      </div>
+    ) : null;
 
     const className = classNames(styles.Toast, error && styles.error);
 
@@ -41,6 +56,7 @@ export default class Toast extends React.Component<Props, never> {
       <div className={className}>
         <KeypressListener keyCode={Key.Escape} handler={onDismiss} />
         {content}
+        {actionMarkup}
         {dismissMarkup}
       </div>
     );
@@ -53,11 +69,26 @@ export default class Toast extends React.Component<Props, never> {
   }
 
   private triggerDismissalTimeout() {
-    const {onDismiss, duration = DEFAULT_TOAST_DURATION} = this.props;
+    const {onDismiss, duration, action} = this.props;
+
+    let timeoutDuration = duration || DEFAULT_TOAST_DURATION;
+
+    if (action && !duration) {
+      timeoutDuration = DEFAULT_TOAST_DURATION_WITH_ACTION;
+    } else if (
+      action &&
+      duration &&
+      duration < DEFAULT_TOAST_DURATION_WITH_ACTION
+    ) {
+      // eslint-disable-next-line-no-console
+      console.log(
+        'Toast with action should persist for at least 10,000 milliseconds to give the merchant enough time to act on it.',
+      );
+    }
 
     this.clearDismissalTimeout();
-    if (onDismiss != null && duration != null) {
-      this.timer = window.setTimeout(onDismiss, duration);
+    if (onDismiss != null) {
+      this.timer = window.setTimeout(onDismiss, timeoutDuration);
     }
   }
 }
--- a/src/components/Frame/components/Toast/Toast.tsx
+++ b/src/components/Frame/components/Toast/Toast.tsx
@@ -80,7 +80,7 @@ export default class Toast extends React.Component<Props, never> {
       duration &&
       duration < DEFAULT_TOAST_DURATION_WITH_ACTION
     ) {
-      // eslint-disable-next-line-no-console
+      // eslint-disable-next-line no-console
       console.log(
         'Toast with action should persist for at least 10,000 milliseconds to give the merchant enough time to act on it.',
       );
--- a/src/components/Frame/components/Toast/tests/Toast.test.tsx
+++ b/src/components/Frame/components/Toast/tests/Toast.test.tsx
@@ -1,7 +1,8 @@
 import * as React from 'react';
 import {timer} from '@shopify/jest-dom-mocks';
-import {mountWithAppProvider} from 'test-utilities';
+import {mountWithAppProvider, trigger, findByTestID} from 'test-utilities';
 import {noop} from 'utilities/other';
+import Button from '../../../../Button';
 import {ToastProps as Props} from '../../../types';
 import Toast from '../Toast';
 import {Key} from '../../../../../types';
@@ -37,6 +38,68 @@ describe('<Toast />', () => {
     });
   });
 
+  describe('action', () => {
+    const mockAction = {
+      content: 'Do something',
+      onAction: noop,
+    };
+
+    it('does not render when not defined', () => {
+      const message = mountWithAppProvider(<Toast {...mockProps} />);
+      expect(message.find(Button)).toHaveLength(0);
+    });
+
+    it('passes content as button text', () => {
+      const message = mountWithAppProvider(
+        <Toast {...mockProps} action={mockAction} />,
+      );
+
+      expect(message.find(Button).text()).toContain(mockAction.content);
+    });
+
+    it('triggers onAction when button is clicked', () => {
+      const spy = jest.fn();
+      const mockActionWithSpy = {...mockAction, onAction: spy};
+      const message = mountWithAppProvider(
+        <Toast {...mockProps} action={mockActionWithSpy} />,
+      );
+
+      trigger(message.find(Button), 'onClick');
+
+      expect(spy).toHaveBeenCalledTimes(1);
+    });
+
+    it('defaults to a duration of 10000ms if no duration is provided and action is provided', () => {
+      const spy = jest.fn();
+      const mockActionWithSpy = {...mockAction, onAction: spy};
+
+      mountWithAppProvider(
+        <Toast
+          content="Image uploaded"
+          onDismiss={spy}
+          action={mockActionWithSpy}
+        />,
+      );
+      expect(spy).not.toHaveBeenCalled();
+
+      const defaultDuration = 10000;
+
+      timer.runTimersToTime(defaultDuration);
+      expect(spy).toHaveBeenCalledTimes(1);
+    });
+
+    it('warns that a duration of 10000ms is recommended with action if duration is lower than 10000', () => {
+      const warnSpy = jest.spyOn(console, 'log');
+      mountWithAppProvider(
+        <Toast {...mockProps} action={mockAction} duration={9000} />,
+      );
+
+      expect(warnSpy).toHaveBeenCalledWith(
+        'Toast with action should persist for at least 10,000 milliseconds to give the merchant enough time to act on it.',
+      );
+    });
+  });
+
   describe('onDismiss()', () => {
     it('is called when the dismiss button is pressed', () => {
       const spy = jest.fn();
@@ -44,7 +107,7 @@ describe('<Toast />', () => {
         <Toast content="Image uploaded" onDismiss={spy} />,
       );
 
-      message.find('button').simulate('click');
+      findByTestID(message, 'closeButton').simulate('click');
       expect(spy).toHaveBeenCalled();
     });
 
@@ -76,6 +139,18 @@ describe('<Toast />', () => {
       expect(spy).not.toHaveBeenCalled();
 
       timer.runTimersToTime(duration);
+      expect(spy).toHaveBeenCalledTimes(1);
+    });
+
+    it('is called after the default duration is reached and no duration was provided', () => {
+      const spy = jest.fn();
+
+      mountWithAppProvider(<Toast content="Image uploaded" onDismiss={spy} />);
+      expect(spy).not.toHaveBeenCalled();
+
+      const defaultDuration = 5000;
+
+      timer.runTimersToTime(defaultDuration);
       expect(spy).toHaveBeenCalledTimes(1);
     });
 
--- a/src/components/Frame/types.ts
+++ b/src/components/Frame/types.ts
@@ -1,4 +1,5 @@
 import * as PropTypes from 'prop-types';
+import {Action} from '../../types';
 
 export interface FrameManager {
   showToast(toast: {id: string} & ToastProps): void;
@@ -63,4 +64,6 @@ export interface ToastProps {
   error?: boolean;
   /** Callback when the dismiss icon is clicked */
   onDismiss(): void;
+  /** Adds an action next to the message (stand-alone app use only) */
+  action?: Action;
 }
--- a/src/components/Toast/README.md
+++ b/src/components/Toast/README.md
@@ -68,10 +68,18 @@ class EmbeddedAppToastExample extends React.Component {
 
 Toast should:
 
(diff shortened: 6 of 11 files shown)


solonaarmstrong:
@BPScott Do you know what the CLA failure is about?

solonaarmstrong:
This looks great!

We should probably add an example in the style guide with the action prop. https://github.com/Shopify/polaris-react/blob/master/src/components/Toast/README.md#examples
Especially if the action is `undo`, have we agreed on what the duration should be (for an example)?

I'm not sure if app bridge has an equivalent action? Should we investigate whether it needs something here: https://github.com/Shopify/polaris-react/blob/master/src/components/Toast/Toast.tsx#L53

🎩 doesn't animate the toast and I think this is because of the Storybook iframe. @BPScott do you have any thoughts on that? But, it does render and it looks good.

I did notice that the spacing between the action and the close button is smaller than Sara's design, but you used `extra-loose`, which is the largest. Either that's enough or it would need a custom spacing variable of something like `rem($unit * 16)`.

screen shot 2019-01-22 at 6 38 25 pm
screen shot 2019-01-22 at 6 38 45 pm

src/components/Frame/types.ts
```diff
@@ -63,4 +64,5 @@ export interface ToastProps {
   error?: boolean;
   /** Callback when the dismiss icon is clicked */
   onDismiss(): void;
+  action?: Action;
```
solonaarmstrong: The Polaris style guide renders the comment above a prop as the prop description. Something like...

```suggestion
/** An action to render beside the dismiss button */
  action?: Action;
```

src/components/Frame/components/Toast/Toast.scss
```diff
@@ -17,6 +17,10 @@ $Backdrop-opacity: 0.88;
   }
 }
 
+.ActionLink {
+  margin-left: (spacing(extra-loose));
```
solonaarmstrong:
```suggestion
  margin-left: spacing(extra-loose);
  margin-right: spacing();
```

I did notice that the spacing between the action and the close button is smaller than Sara's design. Throwing a right margin on there fits the design more closely. Also, I don't think those extra parentheses are needed.

screen shot 2019-01-22 at 6 38 25 pm
screen shot 2019-01-22 at 6 38 45 pm

src/components/Frame/components/Toast/tests/Toast.test.tsx
```diff
@@ -37,6 +38,45 @@ describe('<Toast />', () => {
     });
   });
 
+  describe('action <Link />', () => {
+    const mockAction = {
+      content: 'Do something',
+      onAction: noop,
+    };
+
+    it('does not render when not defined', () => {
```
solonaarmstrong: Should this be preceded by a mount without the `action` prop?

src/components/Frame/components/Toast/tests/Toast.test.tsx
```diff
@@ -37,6 +38,45 @@ describe('<Toast />', () => {
...
+    });
+
+    it('triggers onAction when button is clicked', () => {
+      const spy = jest.fn();
+      const mockActionWithSpy = {...mockAction, onAction: spy};
+      const message = mountWithAppProvider(
+        <Toast {...mockProps} action={mockActionWithSpy} />,
+      );
+
+      message.find(Link).simulate('click');
```
solonaarmstrong: Would it be better to use `trigger` here?

MateusJabour:
@solonaarmstrong I think we can wait for her, she has all the context.

src/components/Frame/components/Toast/Toast.tsx
```diff
@@ -2,6 +2,7 @@ import * as React from 'react';
 import {classNames} from '@shopify/react-utilities';
 
 import {Key} from '../../../../types';
+import Link from '../../../Link';
```
gf3: this should probably be a `Button`, no?

MateusJabour: Could you develop more on that, please?

gf3: generally a Link is for something that takes you somewhere whereas a Button is for something that performs an action

MateusJabour: Cool, I'll make the change, thank you for the insight 👍

BPScott:
> @BPScott Do you know what the CLA failure is about?

Sorry for missing this first time around. It looks like @MateusJabour doesn't have his shopify email account associated with his github account - note [his commits from git aren't linked to his useraccount](https://github.com/Shopify/polaris-react/pull/919/commits) (the ones done through the GH UI are though). @MateusJabour can you please [add your shopify email to your GH account](https://help.github.com/articles/adding-an-email-address-to-your-github-account/) so that the bot  can recognize you.

dpersing:
If users can retry their action in the toast, we’ll need to make sure that functionality is available elsewhere. Otherwise, the toast can time out before users can access it. This would have a big impact on anyone relying on the keyboard alone, since they would need to tab all the way to the bottom to access the toast. It would also have an impact on low vision users relying on magnification, since they may miss the toast entirely.

A few options that come to mind:

- Allow merchants to retry the action with the original control they used. This is where keyboard focus would be.
- Create a keyboard shortcut to allow users to manually move focus to the toast. This would need to be visible for sighted keyboard users and conveyed programmatically to screen reader users. The shortcut would also need to use a modifier key and big conflict with screen reader default shortcuts.
- Time retrying the action and let the user know the timing, or cue the user to retry (this would still require a separate method or a shortcut). This is how Android handles their toast equivalent, snack bars.

solonaarmstrong:
Content for toast: https://docs.google.com/document/d/1IbTKFxhaWyo2ebiE56qgr_ysZZ5JX_ml_HRaa9rbzNs/edit#heading=h.lpzdeyxbvb1d

solonaarmstrong:
I spoke to Mateus and I'm going to wrap this up.

src/components/Frame/components/Toast/Toast.scss
```diff
@@ -4,7 +4,7 @@ $Backdrop-opacity: 0.88;
 .Toast {
   @include text-style-display-small;
   display: inline-flex;
-  max-width: rem(400px);
+  max-width: rem(720px);
```
solonaarmstrong: I'm not sure the width needs to be increased to 720px. If the content guidelines suggest a message no longer than 3 words, does it really need to exceed 500px?

screen shot 2019-02-12 at 5 32 21 pm

src/components/Frame/components/Toast/Toast.scss
```diff
@@ -4,7 +4,7 @@ $Backdrop-opacity: 0.88;
 .Toast {
   @include text-style-display-small;
   display: inline-flex;
-  max-width: rem(400px);
+  max-width: rem(500px);
```
solonaarmstrong: This was changed to 720px, but I'm not sure the width needs to be increased that much. If the content guidelines suggest a message no longer than 3 words, does it really need to exceed 500px?

screen shot 2019-02-12 at 5 32 21 pm

src/components/Toast/README.md
```diff
@@ -91,23 +99,30 @@ Messages should be:
...
 
 <!-- end -->
 
-### Action
+### Toasts with actions
+
+Only include actions in toasts if the same action is available elsewhere on the page. For example:
+
+- If merchants need to reload a section, offer the call to action [Reload] in the toast message. If they miss the toast, they can also refresh the entire page.
+- If merchants delete an image, offer the option to [Undo] the deletion. If they miss it in the toast, they can still retrieve it from {x}.
```
solonaarmstrong: Still waiting for clarity on what `{x}` might be.

solonaarmstrong:
@dpersing 

> Allow merchants to retry the action with the original control they used. This is where keyboard focus would be.

I think this would have to be in guidelines because the implementation would be up to the consumer. Toast itself only knows to be active or not active, it doesn't know its activator (original control) in order to focus it. I also don't know how this would work with something like a `retry` action because we can't place focus on the browser's refresh/reload button.

> Create a keyboard shortcut to allow users to manually move focus to the toast. This would need to be visible for sighted keyboard users and conveyed programmatically to screen reader users. The shortcut would also need to use a modifier key and big conflict with screen reader default shortcuts.

- Where to make it visible (tooltip)?
- How to convey it programmatically
- Which modifier
- How to detect modifier (combo)?
- role alert and aria-live ?

> Time retrying the action and let the user know the timing, or cue the user to retry (this would still require a separate method or a shortcut). This is how Android handles their toast equivalent, snack bars.

Do you mean a countdown?

src/components/Toast/README.md
solonaarmstrong: @clauderic Do you have any input on a scenario where there would be an undo action and where that action could be taken elsewhere?

src/components/Toast/README.md
```diff
@@ -91,23 +99,31 @@ Messages should be:
...
 
 Action should:
 
-- Keep the action label short. Preferably 1 verb
-- Not have actions for dismissing toast
+- Keep the action label short, preferably 1 verb.
+- Not have actions, like [Cancel], for dismissing toast. The [X] to dismiss is already included in the component.
+- Be used with a duration of 10000 milliseconds for accessibility.
```
solonaarmstrong: @dpersing Is this where this belongs? I was thinking this was good enough for now until there is an accessibility section.

solonaarmstrong:
@sarahill Did you intend for the action to be uppercase or title case?

image

image

src/components/Toast/README.md
```diff
@@ -302,7 +362,7 @@ On iOS, icons are available for cases where you want to re-inforce the message.
 
 <!-- example-for: android, ios, web -->
 
-Use error toast to indicate that an operation has failed. For example, your phone is offline and need to reconnect to the internet. For all other error message types, follow the [error message guidelines](/patterns/error-messages).
+Although error toasts are still available and used in the system, we discourage their use. Reserve them for errors not caused by merchants, like a connection issue. Error toasts should convey what went wrong in plain language and should not go over 3 words. For all other error message types, follow the [error message guidelines](/patterns/error-messages).
```
solonaarmstrong: Changes to error were included in this PR because adding an action to an error toast modified the guidelines.

src/components/Toast/README.md
```diff
@@ -68,18 +68,26 @@ class EmbeddedAppToastExample extends React.Component {
 
 Toast should:
 
-- Be used for short messages to confirm an action. Maximum of 2 lines of text.
-- Not be used for actionable links or messages.
-- Not be used for error messages.
-- Be displayed once at the time. If you need multiple toasts, queue them.
+- Be used for short messages to confirm an action.
```
selenehinkley:
```suggestion
- Be used for short messages to confirm an action
```

src/components/Toast/README.md
```diff
@@ -68,18 +68,26 @@ class EmbeddedAppToastExample extends React.Component {
...
+- Not go over 3 words.
```
selenehinkley:
```suggestion
- Not go over 3 words
```

src/components/Toast/README.md
```diff
@@ -68,18 +68,26 @@ class EmbeddedAppToastExample extends React.Component {
...
+- Rarely be used for error messages.
```
selenehinkley:
```suggestion
- Rarely be used for error messages
```

selenehinkley: We don't punctuate lists unless one of the items in the list is two sentences (then alllll the list is punctuated)

src/components/Toast/README.md
```diff
@@ -68,18 +68,26 @@ class EmbeddedAppToastExample extends React.Component {
...
+
+When to use:
+
+- For success messages
+- Only for non-critical errors that are relevant in the moment and can be explained in 3 words. For example, if there’s an internet connection issue, the toast would say, [Internet disconnected].
```
selenehinkley:
```suggestion
- Only for non-critical errors that are relevant in the moment and can be explained in 3 words. For example, if there’s an internet connection issue, the toast would say, Internet disconnected.
```

src/components/Toast/README.md
```diff
@@ -91,23 +99,31 @@ Messages should be:
 - Product updated
 - Collection added
 - Customer updated
-- No internet connection
+- Internet disconnected
+- Connection timed out
 
 #### Don’t
 
+- No internet connection
```
selenehinkley:
```suggestion
- No internet connection
- Can’t charge negative tax rates
- Your online store has a maximum of 20 themes. Delete unused themes to add more.
```

selenehinkley: Added a couple more error msg examples

src/components/Toast/README.md
solonaarmstrong: Ah ok, thank you. I wasn't sure how to treat that.

dpersing:
@solonaarmstrong and I chatted more about this and we decided to lean on documentation for the best practices for this one. As we see how people use the action, we can make some clearer recs or build more into the component. At present we're going to work with our content geniuses to add an accessibility section to the component-specific docs to say that consumers should be aware that it may be difficult for people to access the action in the toast, and reiterate the recommended timeout. We'll encourage that they device an additional way to do the same action in the context of the page or the browser.

sarahill:
I originally designed the action to be uppercase but after we decided to
use the monochrome links we switched to title case.

All that to say they should be title case.

src/components/Toast/README.md
clauderic: Having a history section with a list of changes where each individual change can be undone could be an alternative, @brian-lyn recently explored what that might look like for the Theme editor

solonaarmstrong: 🙏 Thank you!

sadiesaurus:
@solonaarmstrong 

I'm not seeing the Accessibility heading, and it looks like the accessibility content might not be rendering in the right area? 🤔 

alt

src/components/Toast/README.md
```diff
@@ -377,3 +438,25 @@ Keep the action label short, preferably 1 verb action.
 
 - To present a small amount of content or a menu of actions in a non-blocking overlay, [use the popover component](/components/popover)
 - To communicate a change or condition that needs the merchant’s attention within the context of a page, [use the banner component](/components/feedback-indicators/banner)
+
+  ***
```
solonaarmstrong: @sadiesaurus This was the problem. My keyboard has gone awry and this happened. Thanks for pointing it out. 🤦‍♀️

dpersing: This looks great to me!

solonaarmstrong: 🎉

src/components/Frame/components/Toast/Toast.scss
sarahill: This seems fine to me. `720px` seems pretty wide and probably unnecessary for most cases.

dleroux: What happens on small screens, would this be more useful at ‘100vw’ ?

solonaarmstrong: I recently learned that vw and vh don't work well in Safari. On small screens, a long toast message wraps.

src/components/Frame/types.ts
```diff
@@ -63,4 +64,6 @@ export interface ToastProps {
   error?: boolean;
   /** Callback when the dismiss icon is clicked */
   onDismiss(): void;
+  /** Adds an action next to the message */
```
tmlayton:
```suggestion
  /** Adds an action next to the message (stand-alone app use only) */
```

src/components/Toast/README.md
```diff
@@ -91,23 +99,33 @@ Messages should be:
...
+- If merchants need to reload a section, offer the call to action [Reload] in the toast. If they miss the toast message, they can also refresh the entire page.
+- If merchants delete an image, offer the option to [Undo] the deletion. If they miss it in the toast message, they can still retrieve it from somewhere else.
 
 Action should:
 
-- Keep the action label short. Preferably 1 verb
-- Not have actions for dismissing toast
+- Keep the action label short, preferably 1 verb.
+- Not have actions, like [Cancel], for dismissing toast. The [X] to dismiss is already included in the component.
+- Be used with a duration of at least 10,000 milliseconds for accessibility.
```
tmlayton: This seems reasonable to enforce in code that toast with actions have a default of 10000ms. If an action is present and the duration is less than 10000ms we should at least log a warning and force the duration to 10000.

solonaarmstrong: Thank you! I had suggested this, but it seemed we shouldn't have one prop have a side effects on another prop.

UNRELEASED.md
```diff
@@ -12,6 +12,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Added `action` prop to `Toast` (https://github.com/Shopify/polaris-react/pull/919)
```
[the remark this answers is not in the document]
solonaarmstrong: Oops!

src/components/Toast/README.md
tmlayton: If no duration is provided, but an action is, then the default could be 10000ms instead of 5000ms so we don't override. Then only log a warning if the provided duration is less than 10k when an action is present. Not terribly opinionated either way, but I don't really see any downside to overriding a provided duration prop other than the principle that it is gross.

solonaarmstrong: Overriding a provided duration feels a bit heavy-handed, but a log message sounds like a great idea to me. Providing a default of 10,000ms if no duration is provided makes a lot of sense to me.

solonaarmstrong:
action-font-size

The font sizes are different. The message is 20px and the plain button is 14px, so I made it a large size, which bumps it up to 16px. I would need to make a custom button to increase it more.

action-font-size-lg

src/components/Frame/components/Toast/Toast.tsx
```diff
@@ -53,11 +69,26 @@ export default class Toast extends React.Component<Props, never> {
...
+    if (action && !duration) {
+      timeoutDuration = DEFAULT_TOAST_DURATION_WITH_ACTION;
+    } else if (
+      action &&
+      duration &&
+      duration < DEFAULT_TOAST_DURATION_WITH_ACTION
+    ) {
+      // eslint-disable-next-line no-console
+      console.log(
+        'Toast with action should persist for at least 10,000 milliseconds to give the merchant enough time to act on it.',
```
solonaarmstrong: This should be translated, but I'm not entirely sure how to import intl between this file and the public-facing Toast.tsx.

dpersing:
@solonaarmstrong It looks like there are a couple of issues with Toast that aren't specific to this PR, so let me know if you'd prefer that I open separate issues. (Toast wasn't out when I did my initial review of the system, so it probably didn't get tested except with VoiceOver.)

- NVDA doesn't announce the `aria-live` container when it appears. This is related to how NVDA parses `aria-live` regions. To get the content to be read, the container with `aria-live="..."`, `role="alert"`, or `role="status"` needs to be in the DOM already when its contents are changed. This is something that doesn't seem to be likely to change. (Here's an issue about it for some more context: https://github.com/nvaccess/nvda/issues/8873)
- The container will be a bit difficult to see against the background if it's viewed in Windows High Contrast, since the background color is stripped out. Adding a border (either a transparent one using `border: 1px solid transparent`, or one that's only displayed using the `ms-high-contrast` query. @AndrewMusgrave has done some work with high contrast styles, so you can likely ping either of us about this!

Toast in High Contrast

PR #929, opened 2019-01-23
Use `import styles from` notation when importing scss

BPScott:

The css-modules spec and [examples](https://github.com/css-modules/webpack-demo/blob/master/src/components/0-Logo/Logo.js#L1) says to use this format, and we transform our
imports as part of the build to use this format so we should use this
style initially to save that indirection

Removes a custom babel plugin (yay simpler builds)

Speaking to @lemonmade and @TheMallen [on slack](https://shopify.slack.com/archives/C43QLSB42/p1548193916122100) the legacy reason for doing this was some weirdness in some old typescript version, however that issue was now seems to be fixed and we should move to use the default import style.

- Changes all instances of `import * as styles from './blah.scss'` to `import styles from './blah.scss'`
- Remove a custom babel transform that changed the `import * as styles from` to `import styles from` when building polaris.js and polaris.es.js - It is unneeded if we're always writing `import styles from` in the first place
- Update the typescript typings for *.scss files to reflect that using the default import is the correct way. This means that if we ever write `import * as styles` in the future it will be reported as a type error (so no chance for an accidental regression).

The build output for `polaris.js` `polaris.es.js` and `styles.css` are unchanged. This change is only visible in the `esnext` folder - which is only consumed by shopify projects using sewing-kit

### How to 🎩

- On master, create a build output: `git checkout master && mkdir master-build/node_modules/@shopify && yarn run build-consumer polaris-react/master-build`
- On this branch create a build output: `git checkout use-default-style-imports && mkdir new-build/node_modules/@shopify && yarn run build-consumer polaris-react/new-build`
- Compare the outputs and ensure the only differences are in the esnext folder: `diff -r -u master-build new-build`
- Run `yarn run build-consumer polaris-styleguide` and ensure the styleguide still renders and passes running `yarn run sk type-check` and `yarn run test`
- Run `yarn run build-consumer web` and ensure web still renders and passes running `yarn run sk type-check` and `yarn run test`


UNRELEASED.md
```diff
@@ -36,6 +36,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Code quality
 
+- Use `import styles from './foo.scss` instead of non-standard `import * as styles from './foo.scss` when importing scss files ([#929](https://github.com/Shopify/polaris-react/pull/929))
```
```suggestion
- Use `import styles from './foo.scss';` instead of non-standard `import * as styles from './foo.scss';` when importing scss files ([#929](https://github.com/Shopify/polaris-react/pull/929))
```

UNRELEASED.md
```diff
@@ -36,6 +36,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Code quality
 
+- Use `import styles from './foo.scss'` instead of non-standard `import * as styles from './foo.scss'` when importing scss files ([#929](https://github.com/Shopify/polaris-react/pull/929))
```
tmlayton: Why don’t you like including the semicolon?

```suggestion
- Use `import styles from './foo.scss';` instead of non-standard `import * as styles from './foo.scss';` when importing scss files ([#929](https://github.com/Shopify/polaris-react/pull/929))
```

BPScott: Double resolved!

dleroux:
@BPScott 

I think this PR is fine, but then while tophatting in web I got this error:

BPScott:
https://github.com/Shopify/web/blob/master/app/sections/Apps/AppList/components/AppListItem/AppListItem.scss#L80 seems to be passing a number as the 3rd argument to color() instead of a color value. Web is doing something strange and wrong.

BPScott:
Rebased atop latest master

PR #931, opened 2019-01-24
Setup shopify build and the build consumer test

amrocha:

We've had issues in the past where we've released bad versions of Polaris that somehow broke consumers. This PR adds a test that builds a local copy of polaris, adds it to a consumer and builds it.

I've configured it to build two consumers, `polaris-styleguide` and `web`.

Shopify build + bash scripts

### How to 🎩

You can try it out locally or run some tests on buildkite to make sure it's working


amrocha:
@BPScott This is the PR for the consumer build test, what do you think of the approach? any red flags?

Right now a build takes ~17-22 minutes. Do we want to enable this only for master like we talked about? I think that makes the most sense. Right now I have buildkite set up so that it only builds this branch and master

build-consumer.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
BPScott: Add a  `set -e` at the top of these scripts. That'll make the script fail and stop immediately instead of trying to keep running the commands if a prior one failed

build-consumer.sh
```diff
@@ -0,0 +1,10 @@
...
+
+ls -l
+ls -l ../
+git clone ssh://git@github.com/Shopify/$1
+cd $1
+ls -l
+yarn
```
BPScott: yarn supports a `--cwd=` option, and git supports a `-C` option to specify the current working directory, so you can get rid of the cd above, and I believe doing a `yarn add` implies installing anything else that is missing as well, so I think this can be slightly shortened:

```
git clone -C=..ssh://git@github.com/Shopify/$1
yarn --cwd=../$1 add file:../polaris-react
yarn --cwd=../$1 build
```

.shopify-build/polaris-react.yml
```diff
@@ -0,0 +1,31 @@
+containers:
+  default:
+    docker: gcr.io/shopify-docker-images/apps/production/web-platform-ci-node:10.11.0-yarn-1.12.3
+  styleguide:
+    docker: circleci/node:10.13.0
+
+shared:
```
BPScott: You're not sharing anything between steps here so it seems odd to use the shared key. It might be simpler without faffing with yaml's referencing stuff, which means you can remove the shared key:


```yaml
steps:
  - label: 'Build web'
    container: default
    # etc
  - label: 'Build styleguide'
    container: styleguide
    # etc
```

kvendrik:
For reference: https://github.com/Shopify/polaris-react/issues/678

.shopify-build/polaris-react.yml
amrocha: I prefer it this way since it clearly delineates what a step is doing, and where in the pipeline that step is executed. That, and the rest of our projects using Shopify Build follow the pattern I used, and I believe there's value in being consistent here.

amrocha:
. If so then I don't think we need to bother with the contents of move-directory-in-ci.

. Perhaps instead of a build we can run a type-check and tests instead (using a parallelise script like what web does)

>I think only triggering on the master branch is fine for the moment.

I'll go ahead and proceed with that then

>I'd be tempted to merge the build-consumer and move-directory-in-ci scripts into one

They had to be split in an earlier iteration but they can be merged now, I've gone ahead and done that

>and place it in the .shopify-build folder

I don't know about this, I feel like the `.shopify-build` folder should only contain config pertaining to Shopify Build itself. Do you know of any examples of other projects doing this? If it's not gonna be a one-off then I'm ok with it

>Do you need to have polaris-react and the consumer as siblings? Does it work if you clone web inside the polaris-react directory? 

Yep we do, they can't be nested. When you `yarn add` a folder all Yarn does is copy that folder to node_modules, and install dependencies. I believe this causes issues if the repositories are nested. I was running into errors locally until I made them siblings like this.

>Is running a build the ideal test? I note that web's PR builds just runs [linting and unit tests (https://github.com/Shopify/web/blob/master/.shopify-build/pipelines.yml#L70-L75)] - it never runs a build

I thought about this some more and I think we need to run a build. If you'll remember, one of the incidents we caused was only caught at deploy time when web tried building and failed because of an issue with Polaris. To prevent that from happening I believe the only way is to run a build.

amrocha:
@BPScott Can you take a look at this again? I want to merge today or tomorrow if possible

amrocha:
https://buildkite.com/shopify/polaris-react/builds/90 This is the build associated with this commit. I'll turn it on for master after this PR is merged.

amrocha:
Talked to Ben, we're going to try running web's tests as well to see if we can catch a few more errors

There's also some room to improve running time with caching

.shopify-build/setup-consumer-test.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+shopt -s extglob
+set -e
+
+ls -l
+mkdir ../tmp/polaris-react
+mv !(node_modules) ../tmp/polaris-react
+mv ../tmp/polaris-react polaris-react
+git clone ssh://git@github.com/Shopify/$1 --depth 1
```
kaelig: Curious if this has to be `ssh://git@github.com/...` or if `https://github.com/...` would also work?

amrocha: It doesn't, CI can't authenticate with HTTP the only thing it has is the SSH key

.shopify-build/polaris-react.yml
```diff
@@ -0,0 +1,45 @@
+containers:
```
kaelig: This file looks really good. Very nicely organized and easy to read.

amrocha:
Here's the build for the latest commit that fails correctly: https://buildkite.com/shopify/polaris-react/builds/148#_

I'm going to increase the timeout, and rebase on master, will post the new build here once it passes

amrocha:
Here's the rebased build passing https://buildkite.com/shopify/polaris-react/builds/242#_

amrocha:
Talked to the team and everyone is in favor of merging

@BPScott can I have a +1 when you have time?

.shopify-build/polaris-react.yml
```diff
@@ -0,0 +1,45 @@
...
+    container: styleguide
+    timeout: 20m
+    dependencies:
+      - yarn
+    run:
+      - yarn run build
+      - ./.shopify-build/setup-consumer-test.sh polaris-styleguide
+      - yarn --cwd ./polaris-styleguide run build
+  unit-tests: &unit-tests
+    label: '🤓 Unit tests'
```
BPScott: "Web Unit tests" perhaps, as it's not clear if immediatly clear if they're for web or styleguide.

Related: Should we run the styleguide tests too?

amrocha: We could, but the styleguide tests are very spotty and I doubt they'd help much, they rarely fail. What do you think?

BPScott: Cool, happy to leave them out if you don't think they're worth it.

PR #932, opened 2019-01-25
convert extends React.Component to React.PureComponent to avoid needless rerenders

danrosenthal:


danrosenthal:
From Andrew: `Furthermore, React.PureComponent's shouldComponentUpdate() skips prop updates for the whole component subtree.`

https://reactjs.org/docs/react-api.html#reactpurecomponent

😢

PR #943, opened 2019-01-29
Improve Stack vertical alignment example

BPScott:
Per title

### DIFF
--- a/src/components/Stack/README.md
+++ b/src/components/Stack/README.md
@@ -84,11 +84,17 @@ Use to control spacing of items in a stack in standard increments. Use tight for
 
 ### Vertical centering with a stack
 
-Use to vertically center a set of items.
+Use to vertically center a set of items that have different heights.
 
 ```jsx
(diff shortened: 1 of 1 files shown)


src/components/Stack/README.md
```diff
@@ -87,7 +87,7 @@ Use to control spacing of items in a stack in standard increments. Use tight for
 Use to vertically center a set of items.
 
 ```jsx
-<Stack alignment="center">
+<Stack vertical alignment="center">
```
solonaarmstrong: I don't think that's the intention here. This uses the `align-items` flex attribute.

image

solonaarmstrong: So, it's not an example of vertically stacking items, rather _centering_ them along a vertical axis.

solonaarmstrong:
I didn't 🎩 because I'm mid-process firing up core and don't want to interrupt it. But code looks good! 👍

PR #957, opened 2019-02-01
Removes logo background

kvendrik:
We're polishing the new top bar layout by removing the background of the logo in the `TopBar`.

- Removes the `background-color` rule on the `LogoContainer` class.
- Removes `backgroundDarker` theming option as it isn't used anywhere else.

### How to 🎩

<details>
<summary>1. Copy-paste this code in <code>playground/Playground.tsx</code>:</summary>

```jsx
import {AppProvider, TopBar} from '../src';

interface State {}

  render() {
    return (
      <AppProvider
        theme={{
          colors: {
            topBar: {
              background: '#357997',
            },
          },
          logo: {
            width: 124,
            topBarSource:
            contextualSaveBarSource:
            url: 'http://jadedpixel.com',
            accessibilityLabel: 'Jaded Pixel',
          },
        }}
      >
        <TopBar />
      </AppProvider>
    );
  }
}
```
</details>
2. Run `yarn dev` and make sure everything looks as expected.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -11,6 +11,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Enhancements
 
 - Set `Collapsible` to use `overflow: visible;` once fully open ([#951](https://github.com/Shopify/polaris-react/pull/951))
+- Removes `TopBar` logo background ([#957](https://github.com/Shopify/polaris-react/pull/957))
 
 ### Bug fixes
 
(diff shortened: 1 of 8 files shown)


kvendrik:
@danrosenthal I feel like that's possible although I'm not sure changing the design would be a breaking changing. I was slightly concerned about removing the `backgroundDarker` property but considering its a dynamic type it won't cause a type issue. If changing the design is a breaking issue I feel like more of the recently shipped PRs would've had a to trigger a major version release. What do you think?

PR #960, opened 2019-02-02
[TextField] Remove on/off restriction for autocomplete

schonert:
## The issue 
Browsers are constantly changing the way they handle autocomplete. Chrome is ignoring `off`, `nope`, etc. and is aiming to autocomplete all inputs. Restricting developers to `on` and `off` seems unnecessary and prevents fully supporting the browsers native functionality - here I'm referring to [W3C](https://www.w3.org/TR/html52/sec-forms.html#sec-autofill)

## The solution
Enabling custom `autoComplete` values.
```
<TextField autoComplete="new-password" ... />
```

This PR has no breaking changes. Passing a boolean will fallback to `on`/`off`

### DIFF
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -451,7 +451,7 @@ class TextField extends React.PureComponent<CombinedProps, State> {
 function normalizeAutoComplete(autoComplete?: string | boolean) {
   if (autoComplete == null) {
     return autoComplete;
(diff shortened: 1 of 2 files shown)


danrosenthal:
Hey @schonert, looks like this ended up getting done in [another PR](https://github.com/Shopify/polaris-react/blob/master/src/components/TextField/TextField.tsx#L79). Sorry about that! Going to go ahead and close yours

PR #964, opened 2019-02-04
[TextField] Fix error thrown when value is null or undefined

### LINKED ISSUE #959: [TextField] null value throws cannot read property length of null
# Issue summary

When a `TextField` component is given a value of `null`, an error is thrown saying cannot read property length of null.

## Expected behavior

Providing a `null` value should render a `TextField` with no contents.

## Actual behavior

The following error is thrown:

```
Uncaught TypeError: Cannot read property 'length' of null
    at TextField.render (index.es.js:5177)
```

This appears to be due to the character count trying to check the length of the value. Ideally `null` shouldn't be passed to a `TextField`, but I would expect just the typical warning from React about an uncontrolled component and not a crash.

This is the piece of code in [TextField.tsx](https://github.com/Shopify/polaris-react/blob/2c7d6bc5caadb6bd4ea0923bb7c44ab895a18154/src/components/TextField/TextField.tsx#L215) that's crashing:

`const characterCount = value.length;`

I think a simple fix would be:

`const characterCount = value ? value.length : 0;`

1. Create the following component: `<TextField value={null} />`

## Reduced test case

The following code sandbox reproduces the issue: https://codesandbox.io/s/1omnl2q7vj

## Specifications

- Polaris version number: 3.6.0
- Browser: Chrome 72.0.3626.81 (Official Build) (64-bit)
- Device: MacBook Pro
- Operating System: macOS High Sierra 10.14.3

mbaumbach:

Resolves #959

When the value passed in is `null` or `undefined` the `TextField` crashes due to the character count not being able to get the length of the string.

The PR is checking to see if value is truthy and if it is it will calculate the character count like it was doing, otherwise it sets the character count to `0`.

### How to 🎩

<details>

```jsx

interface State {}

  render() {
    return (
      <Page title="Playground">
        <TextField value={null} showCharacterCount />
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist


src/components/TextField/TextField.tsx
```diff
@@ -212,7 +212,7 @@ class TextField extends React.PureComponent<CombinedProps, State> {
       </div>
     ) : null;
 
-    const characterCount = value.length;
+    const characterCount = value ? value.length : 0;
```
tmlayton:
```suggestion
    const characterCount = typeof value === 'string' ? value.length : 0;
```

Let be explicit in checking that value is indeed a string.

mbaumbach:
Thanks @tmlayton for the feedback! I modified the check on value to be more explicit, added a release note, and added a test. Since the tests are in TypeScript, it'll throw a warning since `null` isn't a valid assignment to `string?`. I'm not familiar enough with TypeScript to know if there's a way to suppress that or if that's even desired.

AndrewMusgrave:
Typescript is throwing an error when you'r using `null` because we use the [strick null check compiler option](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#--strictnullchecks), so for our environment `interface Foo {value?: string}` can only ever be a string or undefined. Meaning this can't be normally tested as you've encountered because typescript won't accept null as a value. To avoid this in the test you'll need to use [type assertion](https://www.typescriptlang.org/docs/handbook/basic-types.html#type-assertions). You can achieve this by either asserting the type as any, or a string. e.g. `value={value as any}` or `value={(value as unknown) as string}`. Unknown is needed because the null and string don't overlap.

mbaumbach:
Thanks @AndrewMusgrave, that did the trick for fixing the failure, though there's still a warning. I'm not positive what the percy failure is, as it looks like something I may not have access to see and I don't know why a visual regression test would have failed without any visual changes. Is that test a bit finicky like most visual regression test systems I've seen? :)

mbaumbach:
@AndrewMusgrave Not sure if that question was posed for @tmlayton or myself, but I think that's a perfectly fine solution to me too. There is precedence of that within the file already in the `handleNumberChange` function which attempts to parse the value into a number similarly.

We'd need to modify the `BaseProps` such that `value` has a type of `any` or some composite type that allows `null` too, correct?

tmlayton:
I like that suggestion – remove the deconstruction default assignment, normalize the value, and remove the test.

mbaumbach:
Thanks for the suggestions @AndrewMusgrave and @tmlayton, I removed the test and adjusted the value to be normalized outside of the prop destructuring.

src/components/TextField/TextField.tsx
```diff
@@ -186,11 +186,13 @@ class TextField extends React.PureComponent<CombinedProps, State> {
       polaris: {intl},
     } = this.props;
 
+    const normalizedValue = value ? value : '';
```
tmlayton: Nitpick here, but our [preference for null or undefined checks](https://github.com/Shopify/javascript#11.6) is `value == null` 

```suggestion
    const normalizedValue = value == null ? value : '';
```

mbaumbach: Wouldn't this mean that `undefined` no longer works to defaulting to the empty string now, which was the behavior before with the default value in the destructuring? Or are you saying we should do a check like `value === null || value === undefined`?

tmlayton: Sorry, typo! `value != null`

tmlayton: > EDIT: Just noticed the double equals, not triple. Makes sense. :)

Yep, exactly. The "Abstract Equality Comparison" (I think that’s the official ES2015 name) does a strict check for `null` or `undefined`

mbaumbach: Thanks @tmlayton 👍, I committed that change.

tmlayton:
I wish GitHub had a :highfive: emoji! The raised hands will have to do 🙌

mbaumbach:
Thanks for the help in getting it in!  🎉

PR #966, opened 2019-02-04
Adds contextControl prop to <TopBar /> and <Navigation />

kvendrik:

As part of the context switching updates ([details](https://github.com/Shopify/web/issues/9665)) this PR proposes a `contextControl` prop thats is added to the `<TopBar />` and `<Navigation />` components and accepts a component that is ideally used to provide a way for users to switch between admin contexts.

Adds the `contextControl` prop to `<TopBar />` and `<Navigation />`.

### How to 🎩
<details>
<summary>1. Copy-paste this code in <code>playground/Playground.tsx</code>:</summary>

```jsx
import {TopBar, AppProvider, Frame, Navigation, Button} from '../src';

interface State {}

  render() {
    const topBar = (
      <TopBar
        contextControl={
          <TopBar.Menu
            action={[]}
            activatorContent="Click here to switch from one context to another"
          />
        }
      />
    );

    const nav = (
      <Navigation
        location="/"
        contextControl={
          <Button>Click here to switch from one context to another</Button>
        }
      />
    );

    return (
      <AppProvider
        theme={{
          logo: {
            width: 124,
            topBarSource:
          },
        }}
      >
        <Frame topBar={topBar} navigation={nav} showMobileNavigation />
      </AppProvider>
    );
  }
}
```

</details>

2. Run `yarn dev` and navigate to `http://localhost:6006/iframe.html?selectedKind=Playground&selectedStory=Playground`.
3. Make sure the `<TopBar.Menu />` shows up in the far left corner of the top bar.
4. Remove the `contextControl` prop from the top bar.
5. Make sure the logo shows up and looks/works as you would expect.
6. Make sure the `Button` that was passed into `Navigation` doesn't show up on the page.
6. Resize the screen to a mobile size.
7. Make sure the `Button` that was passed into `Navigation` shows up in the side nav.


src/components/TopBar/tests/TopBar.test.tsx
```diff
@@ -226,6 +226,39 @@ describe('<TopBar />', () => {
       expect(topBar.find(UnstyledLink).prop('style')).toEqual({width: '104px'});
     });
   });
+
+  describe('contextControl', () => {
+    const mockContextControl = (
+      <TopBar.Menu
+        actions={[]}
+        activatorContent="Switch stores"
```
tmlayton:
```suggestion
        activatorContent="Switch contexts"
```
🕵️‍♀️

src/components/TopBar/TopBar.tsx
```diff
@@ -83,20 +86,25 @@ export class TopBar extends React.PureComponent<ComposedProps, State> {
...
-        <Image
-          source={logo.topBarSource || ''}
-          alt={logo.accessibilityLabel || ''}
-          className={styles.Logo}
-          style={{width}}
-        />
-      </UnstyledLink>
-    ) : null;
+    const contextMarkup =
+      !contextControl && logo ? (
```
tmlayton: We will want to use if-else flow control here:

if context control -> render context control
else if logo -> render logo
else -> null

kvendrik: Makes sense that makes it a bit clearer and resolves the no `contextControl` nor `logo` case.

kvendrik:
> We are missing a test case for top bar without context control and without logo. Currently that code path with still render and empty context control div.

Yes good catch! Will fix that!

danrosenthal:
During my :tophat: I noticed this was able to be gotten into a weird state if the initial render was on desktop and you resize it to mobile, where the `Backdrop` was visible but not the `Navigation`. When initial render is on mobile, the `Navigation` renders as expected. This was pretty weird, but I don't see how it could be related to these changes.

PR #974, opened 2019-02-05
Small icon improvements

BPScott:

Prepping for polaris-icons v2

- Don't use ellipsis as it is deprecated in favor of horizontalDots. The
  svgs are identical so this is a drop-in replacement
- Don't use an icon from src/icons if you can pass a string to `<Icon>`
  instead. These do the same thing but it avoids an extra import.
- Adds "menu" as a bundled icon - The exports of `src/icons/index.ts` and BUNDLED_ICONS in `src/components.Icon/Icon.tsx` are now identical.

### How to 🎩

- Ensure tests pass
- Ensure Navigation section Ellipsis look identical - see Navigation -> Navigation with section rollup example.
- Ensure TopBar menus icon looks identical - see any Topbar example


src/icons/index.ts
```diff
@@ -30,7 +30,6 @@ export const deleteIcon = POLARIS_ICONS.deleteIcon;
 export const disable = POLARIS_ICONS.disable;
 export const dispute = POLARIS_ICONS.dispute;
 export const duplicate = POLARIS_ICONS.duplicate;
-export const ellipsis = POLARIS_ICONS.ellipsis;
```
kaelig: For backwards compat, would this make sense?

```js
export const ellipsis = POLARIS_ICONS.horizontalDots;
```

Also, show a deprecation message when somebody is using ellipsis instead of horizontal dots?

And then in v4 we'd be able to completely remove it.

(unless I've missed something and this icon was already deprecated in polaris-react?)

PR #975, opened 2019-02-05
[DataTable] Remove top border

mehdi-loup:
The top border of the datable will often overlap the border of its container, ie. popover.

This issue can be seen in the documentation itself:

Resolves https://github.com/Shopify/polaris-ux/issues/238

Simply set the datatable border top to 0

## 

### 🎩 checklist

* [x] Tested in the playground on mobile and different browsers (chrome, firefox & safari)


UNRELEASED.md
```diff
@@ -14,6 +14,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- Remove datatable top border so it does not overlaps with its container borders. ([#975](https://github.com/Shopify/polaris-react/pull/975))
```
```suggestion
- Fixed top border of `DataTable` overlapping its container borders ([#975](https://github.com/Shopify/polaris-react/pull/975))
```

tmlayton:
> Looks good. @tmlayton I don’t think we consider a visual change like this breaking, but it could require remediation in a consuming app that relied on it for visual separation. Any idea how we might deal with that?

In this case it is a bug fix so not breaking. And theoretically, components should never have visual breaking changes because separation of visual concerns, but never say never 😄

I suspect we would only consider a visual change a breaking change if a consumer could end up in a state where we would classify the app as visually broken. That’s a subjective line, but I think it is something we would know when we see it.

PR #976, opened 2019-02-05
[AppProvider]: Fix server side rendering for embedded apps

hannachen:

Resolves https://github.com/Shopify/polaris-react/issues/824 https://github.com/Shopify/app-bridge/issues/722

**Polaris** `Page` component renders content differently on the server vs client side because **App Bridge** is set to `undefined` when `window` is not available.

Always call `createApp` and allow **App Bridge** to return a noop version of App Bridge when `window` is unavailable.

#### Install and test sample app
1. Check out `shopify` and run test branch `app-bridge/test/ss-app`: `dev up; dev server`
    *This will make the sample server side app available to your local shop*
1. Check out and run Hanna's **Test Polaris App Bridge Embedded App**: https://github.com/hannachen/nextjs-app-test
1. Once the app is installed to your shop, it will initially render the correct UI (server side), but mangled once **App Bridge** client kicks in:
    screen shot 2019-02-05 at 2 19 27 pm

#### Build App Bridge for sample app
1. Check out `app-bridge` from this PR: https://github.com/Shopify/app-bridge/pull/724
1. Build `app-bridge` by running `dev nuke; dev up; dev build`
1. Pack and apply to `nextjs-app-test` by running the following from `app-bridge` project root: 
    ```sh
   dev build-consumer ../hannachen/nextjs-app-test app-bridge
    ```
    *Please change the relative path to `nextjs-app-test` according to your setup*

#### Build Polaris for sample app
1. Build `polaris-react` at the project root by running:
    ```sh
    dev build-consumer ../hannachen/nextjs-app-test
    ```
    *This will build and copy `polaris` into the test app*

#### Test Changes
1. Restart the sample app and  view the test app again
    This time there should be no difference between client and server side render:
    screen shot 2019-02-05 at 3 48 10 pm

### 🎩 checklist

* [N/A] Tested on [multiple browsers](https://help.shopify.com/en/manual/intro-to-shopify/shopify-admin/supported-browsers)
* [N/A] Tested for [accessibility](https://github.com/Shopify/polaris-react/blob/master/documentation/Accessibility%20testing.md)

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -19,7 +19,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Bug fixes
 
 - Fixed unnecessary height on `TextField` due to unhandled carriage returns ([#901](https://github.com/Shopify/polaris-react/pull/901))
-- Ensured server side rendering matches client side rendering for embedded app components (ex. `<Page>`)
+- Ensured server side rendering matches client side rendering for [embedded app components](https://github.com/Shopify/polaris-react/blob/master/documentation/Embedded%20apps.md#components-which-wrap-shopify-app-bridge) ([#976](https://github.com/Shopify/polaris-react/pull/976))
 
 ### Documentation
 
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -19,6 +19,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Bug fixes
 
 - Fixed unnecessary height on `TextField` due to unhandled carriage returns ([#901](https://github.com/Shopify/polaris-react/pull/901))
+- Ensured server side rendering matches client side rendering for embedded app components (ex. `<Page>`)
 
 ### Documentation
 
--- a/src/components/AppProvider/utilities/createAppProviderContext/createAppProviderContext.ts
+++ b/src/components/AppProvider/utilities/createAppProviderContext/createAppProviderContext.ts
@@ -19,6 +19,25 @@ export interface CreateAppProviderContext extends AppProviderProps {
   unsubscribe?(callback: () => void): void;
 }
 
+const serverAppBridge = {
+  dispatch<A>() {
+    return {} as A;
+  },
+  error() {
+    return noop;
+  },
+  featuresAvailable() {
+    return new Promise(noop);
+  },
+  getState() {
+    return new Promise(noop);
+  },
+  localOrigin: '',
+  subscribe() {
+    return noop;
+  },
+};
+
 export default function createAppProviderContext({
   i18n,
   linkComponent,
@@ -32,14 +51,18 @@ export default function createAppProviderContext({
 }: CreateAppProviderContext = {}): Context {
   const intl = new Intl(i18n);
   const link = new Link(linkComponent);
-  const appBridge =
-    apiKey && !isServer
-      ? createApp({
+
+  let appBridge;
+
+  if (apiKey) {
+    appBridge = isServer
+      ? serverAppBridge
+      : createApp({
           apiKey,
           shopOrigin: shopOrigin || getShopOrigin(),
           forceRedirect,
-        })
-      : undefined;
+        });
+  }
 
   if (appBridge && appBridge.hooks) {
     appBridge.hooks.set(LifecycleHook.DispatchAction, setClientInterfaceHook);
--- a/src/components/AppProvider/utilities/createAppProviderContext/createAppProviderContext.ts
+++ b/src/components/AppProvider/utilities/createAppProviderContext/createAppProviderContext.ts
@@ -4,7 +4,6 @@ import createApp, {
   LifecycleHook,
   DispatchActionHook,
 } from '@shopify/app-bridge';
-import {isServer} from '@shopify/react-utilities/target';
 import {AppProviderProps, Context} from '../../types';
 import {StickyManager} from '../withSticky';
 import ScrollLockManager from '../ScrollLockManager';
@@ -19,25 +18,6 @@ export interface CreateAppProviderContext extends AppProviderProps {
   unsubscribe?(callback: () => void): void;
 }
 
-const serverAppBridge = {
-  dispatch<A>() {
-    return {} as A;
-  },
-  error() {
-    return noop;
-  },
-  featuresAvailable() {
-    return new Promise(noop);
-  },
-  getState() {
-    return new Promise(noop);
-  },
-  localOrigin: '',
-  subscribe() {
-    return noop;
-  },
-};
-
 export default function createAppProviderContext({
   i18n,
   linkComponent,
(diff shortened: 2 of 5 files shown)


src/components/AppProvider/utilities/createAppProviderContext/createAppProviderContext.ts
```diff
@@ -27,14 +26,13 @@ export default function createAppProviderContext({
...
-      ? createApp({
-          apiKey,
-          shopOrigin: shopOrigin || getShopOrigin(),
-          forceRedirect,
-        })
-      : undefined;
+  const appBridge = apiKey
+    ? createApp({
+        apiKey,
+        shopOrigin: shopOrigin || getShopOrigin(),
```
gf3: is this change intended for the current version of polaris or the next major release?

hannachen: This is intended to be a bug fix for the current version of Polaris.

gf3: 👍

hannachen:
> Looks good - will this PR be updated to consume a new package of app bridge?

Yep! When we have the changes in App Bridge we'll cut a new release and bump to the new version here. Would that be the correct approach?

vividviolet:
That sounds great @hannachen! I'll do a final top hat once we get the new App Bridge package released

UNRELEASED.md
```diff
@@ -16,6 +16,12 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- Fixed the top border of `DataTable` overlapping its container’s border ([#975](https://github.com/Shopify/polaris-react/pull/975))
+- Fixed the `DataTable` sort direction not reversing on second sort of the initially sorted column ([#918](https://github.com/Shopify/polaris-react/pull/918)) (thanks [@tabrez96](https://github.com/tabrez96) for the [issue report](https://github.com/Shopify/polaris-react/issues/873))
+- Changed the offset from 5px to 4px in `Tooltip` between activator and message to be consistent with `Popover` ([#1019](https://github.com/Shopify/polaris-react/pull/1019))
+- Fixed `Card` header not showing when `title` empty or not set ([#1031](https://github.com/Shopify/polaris-react/pull/1032))
```
tmlayton: Looks like a rebase picked up these changes

UNRELEASED.md
```diff
@@ -16,6 +16,12 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
...
+- Ensured server side rendering matches client side rendering for embedded app components (ex. `<Page>`)
```
tmlayton: Can you add the PR number and link similar to the other entries?

UNRELEASED.md
hannachen: Oops, yes! That got picked up by mistake.

UNRELEASED.md
hannachen: Thanks for taking the time to go through this Tim! I'm getting a few changes for App Bridge in so we can cut a new release to finish up this PR.

UNRELEASED.md
```diff
@@ -23,6 +23,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Removed reference to `window.Polaris`, which in some cases could be undefined ([#1104](https://github.com/Shopify/polaris-react/issues/1104))
 - Added padding and margin to `subdued` sections for proper spacing between the header and footer ([#1082](https://github.com/Shopify/polaris-react/pull/1082))
 - Removed left margin from vertical `Stack` to prevent overflow ([#1024](https://github.com/Shopify/polaris-react/pull/1024))
+- Ensured server side rendering matches client side rendering for embedded app components (ex. `<Page>`) ([#976](https://github.com/Shopify/polaris-react/pull/976))
```
```suggestion
- Ensured server side rendering matches client side rendering for [embedded app components](https://github.com/Shopify/polaris-react/blob/master/documentation/Embedded%20apps.md#components-which-wrap-shopify-app-bridge) ([#976](https://github.com/Shopify/polaris-react/pull/976))
```

src/components/AppProvider/utilities/createAppProviderContext/tests/createAppProviderContext.test.tsx
```diff
@@ -10,18 +11,19 @@ import Link from '../../Link';
...
-jest.mock('@shopify/app-bridge');
-(createApp as jest.Mock<{}>).mockImplementation((args) => args);
-
 const actualIsServer = targets.isServer;
 
 function mockIsServer(value: boolean) {
   (targets as any).isServer = value;
 }
 
+let createAppSpy = jest.spyOn(appBridge, 'default');
```
vividviolet: I think we should avoid calling `jest.spyOn` outside a test. We should do `jest.mock('@shopify/app-bridge')` outside of the tests and within the tests call `jest.spyOn`

hannachen: The reason behind the `jest.spyOn` outside of tests is that I wanted to keep using the same mock for most tests, but unmock the implementation for one of the tests using [`mockRestore()`](https://jestjs.io/docs/en/mock-function-api.html#mockfnmockrestore). Should I just repeat the mock for each test in this case? I've updated the test to do that instead, does this look better now?

vividviolet: If you want the same mock for each test the preferred way is to use a factory function that is called in your test to set up the mocks rather than setting up a global. You can also just set up the mock in `beforeEach`. We also should avoid calling `jest.restoreAllMocks` since it's too general and explicitly call to restore individual mocks.

See the guide here: https://github.com/Shopify/web-foundation/blob/master/Best%20practices/Testing.md#setup-perform-assert

src/components/AppProvider/utilities/createAppProviderContext/tests/createAppProviderContext.test.tsx
```diff
@@ -85,36 +85,33 @@ describe('createAppProviderContext()', () => {
     expect(context).toEqual(mockContext);
   });
 
-  it('does not instantiate app bridge if server side rendering', () => {
-    mockIsServer(true);
+  it('initializes a noop app bridge if server side rendering', () => {
+    jest.spyOn(redirect, 'getWindow').mockReturnValueOnce(undefined);
```
vividviolet: We should treat app-bridge as a blackbox and avoid reaching into its internals. I think we can just extend the `returns the right context with properties` with assertions for `dispatch`, `getState` et... App-bridge is responsible to handle the case when window is undefined - Polaris does not need to care about this anymore

hannachen: Ah, that makes things much easier, it also makes way more sense. Thanks!

PR #982, opened 2019-02-07
Update to use polaris-icons v2

BPScott:

This adds treeshaking support, esm builds and has renamed the exports.

- Update to `@shopify/polaris-icons` v2.0.0
- Update imports to reflect new export names

Note that the names you put into Icon source remain unaffected - you'd still do `<Icon source="add">`, but that now add maps to the named export `addMinor`.

### How to 🎩

- Assert percy passes (and thus icons still render)
- Assert build passes (and thus there's no incorrect imports)


src/icons/index.ts
```diff
@@ -1,54 +1,54 @@
-import POLARIS_ICONS from '@shopify/polaris-icons';
```
BPScott: This whole file could all be replaced with just `export * from '@shopify/polaris-icons'` but IIRC @amrocha and I decided that it's best to be explicit.

I'm still kinda tempted to 🔥 this file entirely and just import icons direct within the Icon component but we can decide if we want to do that later

amrocha: Like we discussed, I believe having an index file creates good habits. Prune a list of icons down to the ones you need and use them, instead of potentially adding a new icon every time

On the `export *` side though, I talked to Kaelig and it sounds like there's only a few icons that are problematic, so I'm gonna put something together soon, run it by Jesse and see if we can get it merged. If we do then what I said above doesn't apply anymore and we can get rid of this file

PR #984, opened 2019-02-07
[DO-NOT-MERGE][Button] Add pressed state

solonaarmstrong:

Resolves https://github.com/Shopify/polaris-react/issues/877

Sometimes buttons are used like a radio button, usually in a group but potentially on their own (as a toggle).

Adds a `pressed` prop with associated styling.

## 

### How to 🎩

<details>

```jsx
import {Page, Button, Stack} from '../src';

interface State {}

  render() {
    return (
        <Stack vertical>
          <Stack>
            <Button pressed>Pressed</Button>
            <Button>Button</Button>
          </Stack>
          <Stack>
            <Button primary pressed>
              Primary pressed
            </Button>
            <Button primary>Primary</Button>
          </Stack>
          <Stack>
            <Button outline pressed>
              Outline pressed
            </Button>
            <Button outline>Outline</Button>
          </Stack>
        </Stack>
      </Page>
    );
  }
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -54,12 +54,55 @@ $spinner-size: rem(20px);
   margin-left: -($spinner-size / 2);
 }
 
+.pressed {
+  background: linear-gradient(
+    to bottom,
+    color('sky', 'light'),
+    color('sky', 'light')
+  );
+  border-color: color('sky', 'dark');
+  box-shadow: inset 0 1px 1px 0 rgba(color('ink', 'lighter'), 0.1),
+    inset 0 1px 4px 0 rgba(color('ink', 'lighter'), 0.2);
+
+  &:focus {
+    border-color: color('indigo');
+    box-shadow: inset 0 1px 1px 0 rgba(color('ink', 'lighter'), 0.1),
+      inset 0 1px 4px 0 rgba(color('ink', 'lighter'), 0.2),
+      0 0 0 1px color('indigo');
+  }
+
+  @media (-ms-high-contrast: active) {
+    &.pressed {
+      color: ms-high-contrast-color('button-text');
+      background: ms-high-contrast-color('button-text-background');
+    }
+  }
+}
+
 .primary {
   @include button-filled(color('indigo'), color('indigo', 'dark'));
   @include recolor-icon(color('white'));
 
   &.disabled {
     @include button-filled-disabled(color('indigo'));
+  }
+
+  &.pressed {
+    background: linear-gradient(
+      to bottom,
+      darken(color('indigo'), 10%),
+      darken(color('indigo'), 10%)
+    );
+    border-color: darken(color('indigo', 'dark'), 15%);
+    box-shadow: inset 0 0 0 0 transparent, shadow(faint),
+      0 0 1px 0 darken(color('indigo', 'dark'), 15%);
+  }
+
+  @media (-ms-high-contrast: active) {
+    &.pressed {
+      color: ms-high-contrast-color('button-text');
+      background: ms-high-contrast-color('button-text-background');
+    }
   }
 }
 
@@ -77,6 +120,19 @@ $spinner-size: rem(20px);
 
   &.disabled {
     @include button-outline-disabled(color('ink', 'lighter'));
+  }
+
+  &.pressed {
+    background: rgba(color('ink', 'lighter'), 0.1);
+    border-color: color('ink', 'lighter');
+    box-shadow: none;
+  }
+
+  @media (-ms-high-contrast: active) {
+    &.pressed {
+      color: ms-high-contrast-color('button-text');
+      background: ms-high-contrast-color('button-text-background');
+    }
   }
 }
 
@@ -124,7 +180,8 @@ $spinner-size: rem(20px);
 
   &:hover,
   &:focus,
-  &:active {
+  &:active,
+  &.pressed:hover {
     @include recolor-icon(color('blue', 'dark'));
     background: transparent;
     border: 0;
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -55,27 +55,46 @@ $spinner-size: rem(20px);
 }
 
 .pressed {
-  background: linear-gradient(
-    to bottom,
-    color('sky', 'light'),
-    color('sky', 'light')
-  );
-  border-color: color('sky', 'dark');
-  box-shadow: inset 0 1px 1px 0 rgba(color('ink', 'lighter'), 0.1),
-    inset 0 1px 4px 0 rgba(color('ink', 'lighter'), 0.2);
+  background: color(sky);
+  border-color: darken(color(sky, dark), 4%);
+  box-shadow: 0 0 0 0 transparent,
+    inset 0 1px 1px 0 rgba(color('ink', 'lighter'), 0.1),
+    inset 0 1px 4px 0 rgba(color('ink', 'lighter'), 0.1);
+
+  &:hover {
+    transition-duration: duration(fast);
+    background: darken(color(sky), 2%);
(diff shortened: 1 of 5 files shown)


solonaarmstrong:
There's not much to test, since this is only adding a CSS class.

dpersing:
@solonaarmstrong @daniedleroux Could this be tied to the existing `ariaPressed` prop? My concern with having a style-only prop is that it doesn't have that programmatic state, and consumers would need to add both.

solonaarmstrong:
@dpersing That would be my instinct, too, but from previous experience, it seems we don't allow one prop to have side-effects on another prop and leave it up to the consumer to implement the `ariaPressed`. I was thinking this would be something we could include in the accessibility best practices.

dpersing:
> I was thinking this would be something we could include in the accessibility best practices.

Sounds good! I can add that.

dleroux:
I agree, pressed should be sufficient. I should have looked at rails before implementing the ariaPressed. Great opportunity to put in practice our deprecation guidelines :)

solonaarmstrong:
Deprecation it is. 🎉

solonaarmstrong:
@ry5n Between you and Ben, I feel as though I'm levelling up on my vernacular.

dpersing:
Taking a closer look at this since it came up in discussion around the state of the "Saved" filter button. @solonaarmstrong @ry5n 

We should consider a way to visually distinguish between pressed and default states for Windows High Contrast users, as well. The change is currently relying on a color shift for `background`, `border-color`, and `box-shadow`, so the styles are the same in High Contrast. 🤔 This will primarily be of concern for grouped buttons, if one or more needs to look selected compared to others.

We can use `-ms-high-contrast` to create a specific style, but if we could do something that doesn't rely on it, that would be ideal.

src/components/Button/Button.scss
```diff
@@ -54,13 +54,56 @@ $spinner-size: rem(20px);
...
+    box-shadow: inset 0 1px 1px 0 rgba(color('ink', 'lighter'), 0.1),
+      inset 0 1px 4px 0 rgba(color('ink', 'lighter'), 0.2),
+      0 0 0 1px color('indigo');
+  }
+
+  @media (-ms-high-contrast: active) {
+    &.pressed {
+      color: ms-high-contrast-color('button-text');
+      background: ms-high-contrast-color('button-text-background');
+    }
```
solonaarmstrong: @dpersing The other 3 similar components: radio, checkbox, and select use `-ms-high-contrast` so I added `button-text` and `button-text-background`.

dpersing: @solonaarmstrong That looks awesome! Thank you.

solonaarmstrong:
> Do we have designs/references for how this should look in different states?

No. If you have opinions about this @ry5n that would be great.

Also, the issue shows only the basic button, but I added primary and outline, in case.

ry5n:
Cool. I’ll work out some styles for this (probably will closely match what exists in Rails) and post back here

ry5n:
Sorry these styles took so long. It ended up being best to design them in code, so I could get a feel for the transitions and so on. I pushed up a branch with what I came up with: 

[Design styles for toggle button (#1060)](https://github.com/Shopify/polaris-react/pull/1060)

### Preview

screen shot 2019-02-19 at 6 26 27 pm

solonaarmstrong:
I cherry-picked your commit, @ry5n. Seemed the most straightforward.

src/components/Button/Button.scss
```diff
@@ -54,13 +54,75 @@ $spinner-size: rem(20px);
...
+  &.disabled {
+    background: lighten(color(sky), 4%);
+    color: color('ink', 'lightest');
+    border-color: color(sky, dark);
+    box-shadow: inset 0 1px 1px 0 rgba(color('ink', 'lighter'), 0.1);
+  }
+
+  @media (-ms-high-contrast: active) {
+    color: ms-high-contrast-color('button-text');
+    background: ms-high-contrast-color('button-text-background');
```
ry5n: I haven’t checked if these look good in Windows

yourpalsonja:
Hey everyone, change of plans! I've talked to @ry5n @solonaarmstrong and @danrosenthal, and we're going to take the changes that have been made here, and use them to build a new `ToggleButton` component in `polaris-next` for the "Saved" button feature in BetterListFilters.

Some reasoning:
- Not every type of `Button` needs to support a `pressed` state
- The difference between `pressed` and `:active` can be somewhat confusing
- We need this change quickly, and creating a "test" component in `polaris-next` will let us try it out with low overhead.

Once I've created the `ToggleButton`, I'll link to the commit here! Leaving this PR open with a *DO NOT MERGE* for now, we can close it if we decide to keep `ToggleButton`.

ry5n:
@yourpalsonja @solonaarmstrong I happened to take a look at the new guidelines for deprecation; would be worth looking over if we go ahead with this PR: https://github.com/Shopify/polaris-react/blob/master/documentation/Deprecation%20guidelines.md

solonaarmstrong:
@ry5n Was there something in the deprecation guideline that I missed?

*[ ] Ship backwards compatible changes which include supporting both the old and new API (make sure the full upgrade path is available)
*[ ] Support backwards compatibility for at least half of a major release cycle, but never more that 2 major release cycles
For example, before or part of 3.5, okay to remove in 4.0. After 3.5, remove in 5.0
Large changes consider a full major release cycle. For example, a large change in 3.1 would be removed in 5.0. But a large change in 2.9 would never wait until 5.0 to remove
*[ ] Use the @deprecated doc tag for props and components (for example, add above where the component is defined), state the reason, and upgrade path
Add a console warning for any respective @deprecated doc tag, state the reason, and state the upgrade path
*[ ] Add the console warning in the constructor for classes and in the render method for functional components Guidelines for writing good console messages
*[ ] For significant deprecations, add a section to the component documentation with rationale. State the upgrade path and include a link to a new component if applicable
*[ ] Call out deprecations in our changelog

ry5n:
The guidelines are new to me, and we would have a deprecation if we went forward with this PR, so I just thought I’d drop a reminder.

yourpalsonja:
@ry5n @solonaarmstrong Speaking of deprecation, I didn't include the backwards-compatible prop `ariaPressed` on `ToggleButton`. Even though it resembles `Button`, and accepts most of the same props, it's a different component, and we're deprecating `ariaPressed` anyhow. Let me know if you have a different opinion!

solonaarmstrong:
It's a different component altogether, so that makes sense to me.

yourpalsonja:
@ry5n @dpersing @danrosenthal @dleroux How does everyone feel about closing this for now, since `ToggleButton` has been merged in https://github.com/Shopify/web/pull/11029? We can always re-open if needed, but closing and adding something about deprecating `aria-pressed` for `Button` to the backlog seems like a better move.

yourpalsonja:
See notes above.

BPScott:
Heya folks. I need this for some work in polaris-icons. What will be the best way to bring it back to life?

dleroux:
This PR specifically, or the the ToggleButton in polaris-next?

beefchimi:
I also need this `ToggleButton` and will be copying it from `web` into our project for now.

Any ETA and it joining the Polaris family?

yourpalsonja:
@beefchimi We're planning on putting it in, but are waiting on feedback from a few folks that have grabbed it from `web` like you, so we make sure we build something that's useful and reusable. Let us know how you implemented it and it if needed any changes!

yourpalsonja:
@beefchimi We're planning on putting it in, but are waiting on feedback from a few folks that have grabbed it from `web` like you, so we make sure we build something that's useful and reusable. Let us know how you implemented it and it if needed any changes!

PR #987, opened 2019-02-07
[Range slider] Widen track click target area

solonaarmstrong:

Resolves https://github.com/Shopify/polaris-react/issues/798. Depends on https://github.com/Shopify/polaris-react/pull/784.

The clickable area for the track was too small.

Widens it. It also adds track click functionality to the manual dual thumb.

## 

### How to 🎩

http://localhost:6006/?selectedKind=All%20Components%7CRange%20slider&selectedStory=Dual%20thumb%20range%20slider&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

http://localhost:6006/?selectedKind=All%20Components%7CRange%20slider&selectedStory=Default%20range%20slider&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

<details>

```jsx
import {Page, RangeSlider} from '../src';

interface State {
  singleValue: number;
  dualValue: [number, number];
}

  state: State = {
    singleValue: 50,
    dualValue: [50, 90],
  };

  handleSingleChange = (value) => {
  };

  handleDualChange = (value) => {
    this.setState({dualValue: value});
  };

  render() {
    const {singleValue, dualValue} = this.state;

    return (
      <Page title="Playground">
        <RangeSlider
          label="LLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLL"
          value={singleValue}
          prefix="Hue"
        />
        <div style={{marginTop: '50px'}} />
        <RangeSlider
          label="LLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLLL"
          value={dualValue}
          onChange={this.handleDualChange}
          prefix="Hue"
        />
      </Page>
    );
  }
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/RangeSlider/components/DualThumb/DualThumb.scss
+++ b/src/components/RangeSlider/components/DualThumb/DualThumb.scss
@@ -3,8 +3,9 @@ $stacking-order: (
   input: 20,
 );
 
+$range-wrapper: rem(28px);
 $range-track-height: rem(4px);
-$range-thumb-size: rem(26px);
+$range-thumb-size: rem(24px);
 $range-thumb-border-size: rem(2px);
 $range-thumbs-border: rem(1px) solid color('sky', 'dark');
 $range-thumbs-border-error: rem(2px) solid color('red');
@@ -15,6 +16,7 @@ $range-thumbs-border-active: rem(2px) solid color('indigo');
   position: relative;
   width: 100%;
   display: flex;
+  align-items: center;
 }
 
 .TrackWrapper {
@@ -22,7 +24,8 @@ $range-thumbs-border-active: rem(2px) solid color('indigo');
   display: flex;
   align-items: center;
   width: 100%;
-  min-height: $range-thumb-size;
+  min-height: $range-wrapper;
+  cursor: pointer;
 
   &.disabled {
     opacity: 0.8;
--- a/src/components/RangeSlider/components/DualThumb/DualThumb.tsx
+++ b/src/components/RangeSlider/components/DualThumb/DualThumb.tsx
@@ -197,7 +197,11 @@ export default class DualThumb extends React.Component<Props, State> {
         >
           <div className={styles.Wrapper} id={id}>
             {prefixMarkup}
-            <div className={trackWrapperClassName}>
+            <div
+              className={trackWrapperClassName}
+              onMouseDown={this.handleMouseDownTrack}
+              testID="trackWrapper"
+            >
               <div
                 className={styles.Track}
                 style={cssVars}
@@ -272,6 +276,7 @@ export default class DualThumb extends React.Component<Props, State> {
   ) {
     if (event.button !== 0) return;
     registerMouseMoveHandler(this.handleMouseMoveThumbLower);
+    event.stopPropagation();
   }
 
   @autobind
@@ -289,6 +294,7 @@ export default class DualThumb extends React.Component<Props, State> {
   ) {
     if (event.button !== 0) return;
     registerMouseMoveHandler(this.handleMouseMoveThumbUpper);
+    event.stopPropagation();
   }
 
   @autobind
@@ -396,6 +402,23 @@ export default class DualThumb extends React.Component<Props, State> {
         },
         this.dispatchValue,
       );
+    }
+  }
+
+  @autobind
+  private handleMouseDownTrack(event: React.MouseEvent) {
+    if (event.button !== 0) return;
+    const clickXPosition = this.actualXPosition(event.clientX);
+    const {value} = this.state;
+    const distanceFromLowerThumb = Math.abs(value[0] - clickXPosition);
+    const distanceFromUpperThumb = Math.abs(value[1] - clickXPosition);
+
+    if (distanceFromLowerThumb <= distanceFromUpperThumb) {
+      this.setValue([clickXPosition, value[1]], Control.Upper);
+      registerMouseMoveHandler(this.handleMouseMoveThumbLower);
+    } else {
+      this.setValue([value[0], clickXPosition], Control.Lower);
+      registerMouseMoveHandler(this.handleMouseMoveThumbUpper);
     }
   }
 
--- a/src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
+++ b/src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
@@ -14,10 +14,6 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
   display: flex;
   align-items: center;
 
-  &:not(:first-child) {
-    margin-top: spacing(tight);
-  }
-
   &.disabled {
     opacity: 0.8;
   }
@@ -29,6 +25,12 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
   align-items: center;
   flex: 1 1 auto;
   height: $range-thumb-size;
+
+  input {
+    padding: rem(12px) 0;
+    background-color: transparent;
+    cursor: pointer;
+  }
 }
 
 .Prefix {
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -11,6 +11,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Enhancements
 
 - Added support for dual values to `RangeSlider` component ([#784](https://github.com/Shopify/polaris-react/pull/784))
+- Widened click target area of track for `RangeSlider` ([#987](https://github.com/Shopify/polaris-react/pull/987))
 
 ### Bug fixes
 
--- a/src/components/RangeSlider/components/DualThumb/tests/DualThumb.test.tsx
+++ b/src/components/RangeSlider/components/DualThumb/tests/DualThumb.test.tsx
@@ -520,6 +520,65 @@ describe('<DualThumb />', () => {
       expect(onChangeSpy).not.toHaveBeenCalled();
     });
 
+    it('moves the lower thumb when the track is clicked closer to it than the upper thumb', () => {
+      const onChangeSpy = jest.fn();
+      const dualThumb = mountWithAppProvider(
+        <DualThumb {...mockProps} value={[5, 40]} onChange={onChangeSpy} />,
+      );
+
+      clickTrack(dualThumb, 0.2);
+
+      expect(onChangeSpy).toHaveBeenCalledWith([10, 40], mockProps.id);
+    });
+
+    it('moves the upper thumb when the track is clicked closer to it than the lower thumb', () => {
+      const onChangeSpy = jest.fn();
+      const dualThumb = mountWithAppProvider(
+        <DualThumb {...mockProps} value={[5, 40]} onChange={onChangeSpy} />,
+      );
+
+      clickTrack(dualThumb, 0.6);
+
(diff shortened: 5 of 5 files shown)


src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -399,6 +402,32 @@ export default class DualThumb extends React.Component<Props, State> {
     }
   }
 
+  @autobind
```
beefchimi: All you want is for the "hitbox" to be larger than the "visual" line of the track, correct?

Is there no way to pull that off with CSS? I worry when I see `getBoundingClientRect` being used and the potential workload involved in calculating this stuff each time the user clicks.

I don't have much time to dig into this today unfortunately 😢 but would love to hear more about other solutions you have explored.

solonaarmstrong: The single thumb can be achieved with CSS because the native input includes the functionality of being able to click somewhere on the track and move the thumb.

The dual thumb is built manually and would need to recreate this behaviour before being able to use CSS to widen the hit area.

beefchimi: sorry - is it the `thumb` or the `track` that is "too thin to click"? Which are you trying to increase the `hitbox` for?

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -197,7 +197,10 @@ export default class DualThumb extends React.Component<Props, State> {
         >
           <div className={styles.Wrapper} id={id}>
             {prefixMarkup}
-            <div className={trackWrapperClassName}>
+            <div
+              className={trackWrapperClassName}
+              onClick={this.handleTrackClick}
```
tmlayton: We’ll want this to be a `onMouseDown` to follow the [UI Events spec](https://www.w3.org/TR/uievents/) as it will be the start of a drag event. A mouse down and hold should permit dragging the respective thumb.

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -399,6 +402,32 @@ export default class DualThumb extends React.Component<Props, State> {
     }
   }
 
+  @autobind
+  private handleTrackClick(event: React.MouseEvent) {
+    const clickXPosition = this.actualXPosition(event.clientX);
```
tmlayton: Early return if it is a secondary button click

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -399,6 +402,32 @@ export default class DualThumb extends React.Component<Props, State> {
...
+    let xPositionLower = this.props.min;
+    let xPositionUpper = this.props.max;
```
tmlayton: No need for props or looking at the DOM (`getBoundingClientRect`), instead calculate the midpoint based on the current upper and lower values and determine if the relative X position is above or below and update the respective value.

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -399,6 +402,32 @@ export default class DualThumb extends React.Component<Props, State> {
...
+      xPositionLower = this.actualXPosition(
+        this.thumbLower.current.getBoundingClientRect().left,
+      );
+      xPositionUpper = this.actualXPosition(
+        this.thumbUpper.current.getBoundingClientRect().left,
+      );
+    }
+
+    if (clickXPosition > xPositionLower && clickXPosition < xPositionUpper) {
+      return;
```
tmlayton: We won’t need an early return for out-of-bounds X positions because the event is tied to the track wrapper, which constrains the value so it is in-bounds.

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -29,6 +29,11 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
   align-items: center;
   flex: 1 1 auto;
   height: $range-thumb-size;
+
+  input {
+    padding: spacing() 0;
+    background-color: transparent;
+  }
```
tmlayton: The default browser cursor, in this case `pointer`, will not respect padding for the input so this will need an explicit `cursor: pointer;`

tmlayton:
We’ll also want to add styles to increase the hit area on the dual thumb and set the cursor to pointer there as well.

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
solonaarmstrong: The track.

solonaarmstrong:
> We’ll also want to add styles to increase the hit area on the dual thumb and set the cursor to pointer there as well.

The wrapper is already the full height.

screen shot 2019-02-08 at 11 01 57 am

👍 to add the cursor pointer.

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
solonaarmstrong: Even if this isn't a dragging motion? The intention is to be able to click and have the thumb jump to the clicked position.

click-track

tmlayton:
> The wrapper is already the full height.

The change to the padding on single thumb makes the hit area larger than the container height, even if that wasn’t the intent I think it is a nicer experience. My vote is for larger hit areas on both, but regardless I think they should at least be consistent:

|Single|Dual|
|---|---|
|screen shot 2019-02-08 at 11 52 14 am|screen shot 2019-02-08 at 11 52 04 am|

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
solonaarmstrong: If the click happens in the selected range, you think whichever thumb is closest should be affected?

clicks-on-track

tmlayton: Yep, exactly. The UX is a bit odd if clicking near the midpoint, but we are really trying to, in the simplest way, solve for moving the thumb nearest to click. We could potentially add a kind of dead zone in the middle that does nothing when clicked, but that's us being too opinionated (in my opinion 😄)  about user intent and adds complexity which I don't think adds value.

solonaarmstrong: Great, thanks!

solonaarmstrong:
The native single input calculates from the center of the thumb, but the dual calculates from the left edge of the thumb, so there's inconsistency between their respective behaviours.

middle-vs-left

solonaarmstrong:
Making these consistent is a little tricky because they're different beasts.

There is an 8px distance between text field and its label. I think the single thumb should have the same distance, so I removed the top margin.

screen shot 2019-02-08 at 6 41 41 pm
screen shot 2019-02-08 at 6 41 57 pm

In order to make the heights consistent, I need to use base * 3 for padding, which doesn't exist in the Sass spacing map.

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -29,6 +25,12 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
   align-items: center;
   flex: 1 1 auto;
   height: $range-thumb-size;
+
+  input {
+    padding: rem(12px) 0;
```
solonaarmstrong: In order to make the heights consistent, I need to use base * 3 for padding, which doesn't exist in the Sass spacing map.

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -14,10 +14,6 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
   display: flex;
   align-items: center;
 
-  &:not(:first-child) {
-    margin-top: spacing(tight);
-  }
-
```
solonaarmstrong: Making these consistent is a little tricky because they're different beasts.

There is an 8px distance between text field and its label. I think the single thumb should have the same distance, so I removed the top margin.

screen shot 2019-02-08 at 6 41 41 pm
screen shot 2019-02-08 at 6 41 57 pm

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -399,6 +402,28 @@ export default class DualThumb extends React.Component<Props, State> {
...
+    if (event.button !== 0) return;
+
+    const {value} = this.state;
+
+    const distanceFromLowerThumb = Math.abs(value[0] - clickXPosition);
+    const distanceFromUpperThumb = Math.abs(value[1] - clickXPosition);
+
+    if (
+      distanceFromLowerThumb < distanceFromUpperThumb ||
+      distanceFromLowerThumb === distanceFromUpperThumb
```
solonaarmstrong: I chose to favour the lower thumb if the click is equidistant (for no particular reason).

src/components/RangeSlider/components/DualThumb/DualThumb.scss
```diff
@@ -15,14 +16,16 @@ $range-thumbs-border-active: rem(2px) solid color('indigo');
   position: relative;
   width: 100%;
   display: flex;
+  align-items: center;
```
solonaarmstrong: To fix prefix and suffix.

screen shot 2019-02-08 at 7 10 48 pm

tmlayton:
Looks like a bad rebase

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -399,6 +402,28 @@ export default class DualThumb extends React.Component<Props, State> {
...
+    const clickXPosition = this.actualXPosition(event.clientX);
+    if (event.button !== 0) return;
+
+    const {value} = this.state;
+
+    const distanceFromLowerThumb = Math.abs(value[0] - clickXPosition);
+    const distanceFromUpperThumb = Math.abs(value[1] - clickXPosition);
+
+    if (
+      distanceFromLowerThumb < distanceFromUpperThumb ||
```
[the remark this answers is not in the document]
solonaarmstrong: 🤦‍♀️

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -399,6 +402,28 @@ export default class DualThumb extends React.Component<Props, State> {
...
+      distanceFromLowerThumb === distanceFromUpperThumb
+    ) {
+      this.setValue([clickXPosition, value[1]], Control.Lower);
+      registerMouseMoveHandler(this.handleMouseMoveThumbLower);
+    } else if (distanceFromUpperThumb < distanceFromLowerThumb) {
```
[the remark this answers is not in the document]
solonaarmstrong: 👍  I debated this one. I wasn't sure if it should be an else or explicitly name the condition.

solonaarmstrong:
🎉

kaelig:
Fixed in https://github.com/Shopify/polaris-react/pull/987

PR #990, opened 2019-02-08
[TextField] Fix `focused` prop so it triggers focus state

### LINKED ISSUE #985: [TextField] Focused prop is not setting focus on the element
# Issue summary

The `focused` prop on the `<TextField />` component is not setting focus on the input.

## Expected behavior

When `focused` is set to `true`, the input (or textarea if `multiline` is set) should be on its focus state.

## Actual behavior

The input/textarea is on its default state.

1. Create a `<TextField />` with the `focused` prop set to `true`
2. Verify the input is not on its focus state

## Reduced test case

https://codesandbox.io/s/n41rrjovl0?module=App.js

## Specifications

- Polaris version number: 3.6.0
- Browser: Google Chrome 71.0.3578.98
- Device: MacBook Pro
- Operating System: macOS Mojave 10.14.2

amireynoso:

Resolves #985 

The <TextField /> component accepts a `focused` prop which should force the field to be focused. However, it is not currently working (take a look at [the issue](https://github.com/Shopify/polaris-react/issues/985) for a reduced test case).

I based my changes in the [`<SearchField />`](https://github.com/Shopify/polaris-react/tree/master/src/components/TopBar/components/SearchField) implementation (component and tests), which also accepts a `focused` prop and it works as expected. This is a breakdown of the changes:

- It now takes the `focused` prop into consideration when setting the value for `state.focus` (used to be `false` by default)
- Forces the input's focus state on `componentDidMount()` if the prop is `true`.
- Forces  the input's focus state (or blur state) on `componentDidUpdate()` in case the prop is toggled
- Scoped the tests related to this prop into their own `describe` block
- The current test claimed to be testing something it was not testing (nor doing): "focuses input **and calls onFocus()** when focused prop has been updated to true"
   - I refactored this test a little bit to test that the element is in focus when the prop is set to `true`
- Added two new tests to check the prop can be toggled

### How to 🎩

<details>

```jsx

interface State {
  value: string;
}

  state = {
    value: 'Lorem ipsum',
  };

  handleChange = (value: string) => {
    this.setState({value});
  };

  render() {
    return (
      <Page title="Playground">
        <TextField
          label="Focused text field"
          value={this.state.value}
          onChange={this.handleChange}
          multiline
          focused
        />
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist


src/components/TextField/TextField.tsx
```diff
@@ -132,21 +132,29 @@ class TextField extends React.PureComponent<CombinedProps, State> {
 
     this.state = {
       height: null,
-      focus: false,
+      focus: props.focused || false,
```
AndrewMusgrave: We can probably leave this as
```suggestion
      focus: props.focused,
```

src/components/TextField/TextField.tsx
```diff
@@ -132,21 +132,29 @@ class TextField extends React.PureComponent<CombinedProps, State> {
...
   }
 
-  componentDidUpdate({focused}: CombinedProps) {
-    if (
-      this.input &&
-      focused !== this.props.focused &&
-      this.props.focused === true
-    ) {
+  componentDidMount() {
+    if (this.input && this.props.focused) {
```
AndrewMusgrave: The ref should always be available in `componentDidMount`? We should be able to omit that check. The JS styleguide also prefers early returns so we can probably change this to
```tsx
if (!this.props.focused) return;
this.input.focus();
```

src/components/TextField/TextField.tsx
```diff
@@ -132,21 +132,29 @@ class TextField extends React.PureComponent<CombinedProps, State> {
...
       this.input.focus();
     }
   }
 
+  componentDidUpdate({focused: wasFocused}: CombinedProps) {
+    const {focused} = this.props;
+
+    if (this.input) {
```
AndrewMusgrave: Same comment about the early return here as well

src/components/TextField/TextField.tsx
```diff
@@ -132,21 +132,29 @@ class TextField extends React.PureComponent<CombinedProps, State> {
...
+      if (!wasFocused && focused) {
+        this.input.focus();
+      } else if (wasFocused && !focused) {
+        this.input.blur();
```
AndrewMusgrave: Should we blur when the props change? Using the `focused` prop forces focus on the textfield but omitting the prop doesn't necessarily mean they want the field to be blurred. What are your thoughts on this?

src/components/TextField/TextField.tsx
amireynoso: True! I just did this for type "sanity" but I'll change it back.

src/components/TextField/TextField.tsx
amireynoso: Sounds good!

src/components/TextField/TextField.tsx
amireynoso: In this case, the `blur` even would only trigger if the `focused` prop changed values and it wouldn't be triggered on mount. 

One use case is if `focused` is being set programatically based on an interaction with another component. If the focus is removed (programatically), I'd expect it to trigger the blur event.

I saw this on the [`<SearchField />`](https://github.com/Shopify/polaris-react/tree/master/src/components/TopBar/components/SearchField) component and thought it was a nice addition.

src/components/TextField/TextField.tsx
amireynoso: Oh, nevermind. Since `focused` is possibly `undefined`, we need to set a `boolean` default value.

amireynoso:
Hey @AndrewMusgrave, thanks for your review! I addressed the comments, would you mind taking another look?

amireynoso:
@solonaarmstrong Hey Solona! Would you help me with a review of this PR?  I wasn't sure who else to tag, feel free to tag someone else. Thank you!

AndrewMusgrave:
Sorry I already gave feedback so this doesn't show up on my review list anymore, I'll get to it later today 🙏

src/components/TextField/TextField.tsx
AndrewMusgrave: Sorry I thought it optional 😅

UNRELEASED.md
```diff
@@ -12,6 +12,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- Fixed the `focused` prop on `TextField` so it sets the focus state ([#985](https://github.com/Shopify/polaris-react/pull/985))
```
```suggestion
- Fixed the `focused` prop on `TextField` so it sets the focus state ([#990](https://github.com/Shopify/polaris-react/pull/990))
```

src/components/TextField/TextField.tsx
AndrewMusgrave: Works for me, how about you @solonaarmstrong ?

solonaarmstrong:
What's the difference between `focused` and `autoFocus`. It looks as though `autoFocus` is the one that actually places the cursor in the field?

AndrewMusgrave:
`Autofocus` will focus the textfield when it renders, focused maintains focus on the textfield

UNRELEASED.md
amireynoso: 🤦‍♀️Thanks for catching that!

PR #781, opened 2018-12-18
Update Card.Section to accept ReactNode as title

drose-shopify:
### WHY are these changes introduced?

According to the [Polaris Design Doc](https://polaris.shopify.com/components/structure/card#navigation) it allows titles to be `React.ReactNode`. This is already implemented within `Card.Header`, however for sections this is not the case and currently enforces the title to be a string. 
### WHAT is this pull request doing?
Updating the `Card.Section` to allow for `React.ReactNode`.

**Note** This breaks breaks the design slightly for existing items that use a `React.ReactNode` in the title. Previously it would render within a `Subheading` element, this change removes that.
I am happy to provide a form of backwards compatibility and am open to suggestions.

|Before|After|
|------|-----|
|screen shot 2018-12-18 at 4 46 53 pm|screen shot 2018-12-18 at 4 46 57 pm|

## 

### How to 🎩

🖥 [Local development instructions](https://github.com/Shopify/polaris-react/blob/master/README.md#development)
🗒 [General tophatting guidelines](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting.md)
📄 [Changelog guidelines](https://github.com/Shopify/polaris-react/blob/master/documentation/Versioning%20and%20changelog.md)

<details>
<summary>Copy-paste this code in <code>playground/Playground.tsx</code>:</summary>

```jsx
import * as React from 'react';
import {Page, Card, Stack, TextStyle, Subheading} from '@shopify/polaris';

interface State {}

export default class Playground extends React.Component<{}, State> {
  render() {
    return (
      <Page
        title="Playground"
        primaryAction={{content: 'View Examples', url: '/examples'}}
      >
        <Card title="Hello World Before">
          <Card.Section
            title={<Subheading>{this.renderTitleMarkup()}</Subheading>}
          />
        </Card>
        <Card title="Hello World After">
          <Card.Section title={this.renderTitleMarkup()} />
        </Card>
      </Page>
    );
  }

  private renderTitleMarkup(): JSX.Element | null {
    return (
      <Stack>
        <Stack.Item fill>
          <Stack alignment="center" spacing="tight">
            <h4>
              <TextStyle variation="strong">Sub Title</TextStyle>
            </h4>
          </Stack>
        </Stack.Item>
      </Stack>
    );
  }
}
```

</details>

### 🎩 checklist

* [ ] Tested on [mobile](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting.md#cross-browser-testing)
[...]

PR #994, opened 2019-02-08
Revert "Revert "Update Card.Section to accept ReactNode as title""

tmlayton:
Reverts Shopify/polaris-react#993


tmlayton:
Inheriting approvals from #781, merging on green

PR #1005, opened 2019-02-11
[DatePicker] Extract months and week names into en.json

nimzco:

Resolves https://github.com/Shopify/intl-domestic-country-fit/issues/340

Currently, months and week names are hard coded in English, therefore, it breaks in other languages. See:
image

- Extract strings in `en.json` and use `intl.translate` to load them based on the month or weekday

### How to 🎩

- Replace the new translations in `en.json` and see them rendered on the storybook.

### 🎩 checklist


nimzco:
Oh, I didn't realize @joshuajay was taking a stab at this here: https://github.com/Shopify/polaris-react/pull/874 🙈 

I guess I can close this then @joshuajay?

joshuajay:
They both handle the localisation in a different way, not sure which is better.

Mine will take date formats into account when displaying the month/year together and does not need to be maintained by Shopify's translation efforts since it leverages an already implemented international standard for date formatting.

This PR gives us more control over the translations, so there might be value in this over the PR I opened.

nimzco:
Awesome, thank you Chloe! 🙏

Issue #991, opened 2019-02-08
Remove lodash / lodash-decorators


AndrewMusgrave:
### Lodash decorators
- [x] Throttle #1009
- [x] Remove from package #1009

### Lodash
- [x] Merge #1018
- [x] Get #1013
- [x] Replace #1012
- [ ] isEqual #1017
- [x] capitalize #1015
- [x] pick #1020
- [x] isObject #1011
- [ ] remove from package #1016

### Lodash in `examples/create-react-app`
- [x] Merge #1010

Related - #980

sijad:
`denounce` is missed in above list:

https://github.com/Shopify/polaris-react/blob/b16c134fe6e450daf04a94c60d76e121ee118c8d/src/components/MediaQueryProvider/MediaQueryProvider.tsx#L2
https://github.com/Shopify/polaris-react/blob/44060b1ef10a346408675af954d48e03ac1286f6/src/components/Frame/components/Loading/Loading.tsx#L2
...

`isEqual` can replaced with https://www.npmjs.com/package/fast-deep-equal or https://www.npmjs.com/package/react-fast-compare

PR #1010, opened 2019-02-11
Add lodash to CRA example

AndrewMusgrave:
 
We shouldn't import packages that aren't dependencies. Lodash isn't a dep. https://github.com/Shopify/polaris-react/blob/master/examples/create-react-app/package.json

Part of #991

Adding lodash

### TODO:
* add changelog

### How to 🎩
Run the example 👍 

### 🎩 checklist

### DIFF
--- a/examples/create-react-app/package.json
+++ b/examples/create-react-app/package.json
@@ -4,6 +4,7 @@
   "private": true,
   "dependencies": {
     "@shopify/polaris": "^2.2.0",
+    "lodash": "^4.17.11",
     "react": "^16.3.1",
     "react-dom": "^16.3.1",
     "react-scripts": "1.1.4"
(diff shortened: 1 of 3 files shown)


BPScott:
I'm not too familar with how mergeAppProviderOptions()  and legacy context stuff does its thing, but do we need a deep merge there or could we get away with a shallow one using Object.assign()? If we can do that we wouldn't need this dependency.

examples/create-react-app/package.json
```diff
@@ -4,6 +4,7 @@
   "private": true,
   "dependencies": {
     "@shopify/polaris": "^2.2.0",
+    "lodash": "^4.17.11",
```
elizabethletourneau: 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

solonaarmstrong:
shrink ray needs a rerun here too

PR #1011, opened 2019-02-11
Remove lodash isObject

AndrewMusgrave:

Part of removing lodash #991

Creating a `isObject` function and using that instead of `_.isObject`

### How to 🎩
Test pass? 💯 

### TODO:
* add changelog

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -242,3 +242,13 @@ export const contentContextTypes: ValidationMap<any> = {
 export interface WithContextTypes<IJ> {
   context: IJ;
 }
+
+export enum TypeOf {
+  Undefined = 'undefined',
+  Object = 'object',
+  Boolean = 'boolean',
+  Number = 'number',
+  String = 'string',
+  Symbol = 'symbol',
+  Function = 'function',
+}
--- a/src/utilities/isObject.ts
+++ b/src/utilities/isObject.ts
@@ -1,8 +1,6 @@
 import {TypeOf} from '../types';
 
-function isObject(value: any) {
+export function isObject(value: any) {
   const type = typeof value;
   return value != null && (type === TypeOf.Object || type === TypeOf.Function);
 }
-
-export default isObject;
--- a/src/utilities/isObject.ts
+++ b/src/utilities/isObject.ts
@@ -1,0 +1,8 @@
+import {TypeOf} from '../types';
+
+function isObject(value: any) {
+  const type = typeof value;
+  return value != null && (type === TypeOf.Object || type === TypeOf.Function);
+}
+
+export default isObject;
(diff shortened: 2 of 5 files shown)


src/utilities/isObject.ts
```diff
@@ -0,0 +1,8 @@
+import {TypeOf} from '../types';
+
+function isObject(value: any) {
+  const type = typeof value;
+  return value != null && (type === TypeOf.Object || type === TypeOf.Function);
+}
+
+export default isObject;
```
BPScott: It looks like all of our other files in src/utilities export named functions, and don't export any default. Should we keep consistent with that behaviour here too?

src/types.ts
```diff
@@ -242,3 +242,13 @@ export const contentContextTypes: ValidationMap<any> = {
 export interface WithContextTypes<IJ> {
   context: IJ;
 }
+
+export enum TypeOf {
```
BPScott: If we're only using this in a single file would it be easier to keep this in isObject.ts too?

Do you see us remembering to use this enum in the future or will we end up sticking to plain JS and doing things like `typeof x === 'boolean'`. My gut is thinking we'll forget that this exists and we might as well do `(type === 'object' || type === 'function');`.

src/utilities/isObject.ts
AndrewMusgrave: That's fair, keeping it consistent would be nice. Do you think for utilities that will only have one export they should export a default, or should we continue with the way things are now?

src/types.ts
AndrewMusgrave: I used it in a few different utility methods I burned from lodash. That's a good question, we'll probably wants others opinion on that. cc/ @daniedleroux @danrosenthal

src/utilities/isObject.ts
BPScott: I don't think there's much difference between a single named export and a default. As theres no inherently superior solution I reckon stick with named export so we are consistent - that way the next time someone needs to add a utility they can see we have a convention and can follow it instead of having to ask "hey we do this in two ways, which one should I follow".

Then if we want to revisit this in the future we can change all utilities at once.

src/types.ts
BPScott: Cool, I wasn't aware of that.

danrosenthal: I don't have strong feelings. If others find it useful, I don't have any problem with it

danrosenthal:
The failed shrinkray report slightly concerns me. Get that passing and add an UNRELEASED entry and this should be good to go

AndrewMusgrave:
> The failed shrinkray report slightly concerns me.

Shrinkray is a little finicky sometimes. It was just acting up when I made the PR 😄

PR #1013, opened 2019-02-11
Remove lodash get

AndrewMusgrave:

Part of removing lodash #991 

Creating a `get` function and using that instead of `_.get`

### How to 🎩
Test pass? 💯 

### TODO:
* add changelog

### DIFF
--- a/src/utilities/get.ts
+++ b/src/utilities/get.ts
@@ -1,0 +1,32 @@
+import {GeneralObject} from '../types';
+
+const OBJECT_NOTATION_MATCHER = /\[(.*?)\]|(\w+)/g;
+
+export function get<T>(
+  obj: GeneralObject | undefined,
+  keypath: string | string[],
+  defaultValue?: T,
+): T | any {
+  if (obj == null) return undefined;
+
+  const keys = Array.isArray(keypath) ? keypath : getKeypath(keypath);
+  let acc = obj;
+  for (let i = 0; i < keys.length; i++) {
+    const val = acc[keys[i]];
+    if (val === undefined) return defaultValue;
+    acc = val;
+  }
+
+  return acc;
+}
+
+function getKeypath(str: string) {
+  const path = [];
+  let result: RegExpExecArray | null;
+  while ((result = OBJECT_NOTATION_MATCHER.exec(str))) {
+    const [, first, second] = result;
+    path.push(first || second);
+  }
+
+  return path;
+}
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -48,5 +48,6 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Update `create-react-app` example dependencies ([#1010](https://github.com/Shopify/polaris-react/pull/1010))
 - Replace all occurrences of `_.capitalize` with a custom `capitalize` function ([#1015](https://github.com/Shopify/polaris-react/pull/1015))
 - Replace all occurrences of `_.isObject` with a custom `isObject` function ([#1011](https://github.com/Shopify/polaris-react/pull/1011))
+- Replace all occurrences of `_.get` with a custom `get` function ([#1013](https://github.com/Shopify/polaris-react/pull/1013))
 
 ### Deprecations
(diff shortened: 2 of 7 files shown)


solonaarmstrong:
In case you didn't know, if this doesn't need a changelog entry, you can add the `Skip Changelog` label to the PR and get that sweet green ✅

src/utilities/get.ts
```diff
@@ -0,0 +1,17 @@
+import {GeneralObject} from '../types';
+
+function get(obj: GeneralObject | undefined, keypath: string) {
```
solonaarmstrong: I guess we never use a default with the lodash function, so you didn't a default in your custom function?

src/utilities/get.ts
```diff
@@ -0,0 +1,17 @@
...
+  if (obj == null) return undefined;
+
+  const keys = keypath.split('.');
+  let acc = obj;
```
solonaarmstrong: Nit: I find it challenging to decipher abbreviated variable names, but that's a personal preference.

src/utilities/get.ts
```diff
@@ -0,0 +1,17 @@
+import {GeneralObject} from '../types';
+
+function get(obj: GeneralObject | undefined, keypath: string) {
+  if (obj == null) return undefined;
+
+  const keys = keypath.split('.');
```
solonaarmstrong: I'm also assuming that this is only checking for `a.b.c` format and not `['a', '0', 'b', 'c']` format?

src/utilities/get.ts
elizabethletourneau: Are we attempting to mimic lodash _.get method or use our own? The _.get function seems to take an array as a `keypath` as well as a string. Here's my hacky attempt at playing around with it:

 `function get(obj: GeneralObject | undefined, keypath: Array<any> | string) {
  if (obj == null) return undefined;
  let keys;
  if (typeof keypath === 'string') {
    keys = keypath.split('.');
  } else {
    keys = keypath;
  }`

src/utilities/get.ts
elizabethletourneau: I noticed this also that it doesn't seem to support this use case but maybe that's because of what Solona said, it's not meant to be as robust. 
 `const object = { 'a': [{ 'b': { 'c': 3 } }] };
 get(object, 'a[0].b.c');
// => 3` 

// => actually returns undefined

src/utilities/get.ts
elizabethletourneau: and a fun little use case from lodash:
` const object = {a: [{b: {c: 3}}]};`
   ` get(object, ['a', '0', 'b', 'c']);`
   ` // => 3`

src/utilities/get.ts
BPScott: `accumulator`, which is the name used in the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce).

(more tedious-than-fun-fact: because I got familiar with functional programming using ruby my brain defaults to calling this parameter "[memo](https://ruby-doc.org/core-2.1.0/Enumerable.html#method-i-reduce)")

solonaarmstrong: ...which then creates confusion with `memoize`. Naming is hard.

solonaarmstrong: I would prefer `accumulator`, but that's just me.

src/utilities/get.ts
AndrewMusgrave: Yup but it's easy enough to add so I might as well

AndrewMusgrave: Create our own minimal function although I could add a form of this. I'll put something tomorrow

src/utilities/get.ts
AndrewMusgrave: Comment above ^

src/utilities/get.ts
```diff
@@ -0,0 +1,32 @@
+import {GeneralObject} from '../types';
+
+const OBJECT_NOTATION_MATCHER = /\[(.*?)\]|(\w+)/g;
```
AndrewMusgrave: I'm by no means a regex wizard so all suggestions are greatly welcomed 🙏

PR #1017, opened 2019-02-11
Remove lodash isEquals

AndrewMusgrave:

Part of removing lodash #991 

Creating a `isEquals` function and using that instead of `_.isObjectsEqual`

### How to 🎩
Test pass? 💯 

### TODO:
* add changelog


AndrewMusgrave:
WIP - trying to fix a circular import issue

src/components/Icon/types.ts
```diff
@@ -0,0 +1,156 @@
+import {SVGSource} from '@shopify/images';
```
chloerice: 👌

src/components/Icon/types.ts
```diff
@@ -0,0 +1,156 @@
...
+  | 'yellowLighter'
+  | 'yellow'
+  | 'yellowDark'
+  | 'orange'
+  | 'redLighter'
+  | 'red'
+  | 'redDark'
+  | 'purple';
+
+export const BUNDLED_ICONS = {
```
BPScott: This isn't a type. Having it in type.ts doesn't feel right

dleroux:
@AndrewMusgrave the original issue was getting sandboxed so I rebased this is this branch `rm-ld-isequals-rebased`. It looks like this does fix this issue: https://github.com/Shopify/polaris-react/issues/450

I didn't want to overwrite your branch because if I'm not mistaken the icon stuff for the circular import is not required anymore.

danrosenthal:
Hey @AndrewMusgrave, is this still something you're looking to merge? Anything else you need here?

dleroux:
We should merge it asap, it’s affecting a few partners.

AndrewMusgrave:
As I've let this PR go a little stale. I'm going to open a new pr that'll include, addressing function comparison overhead and cyclic values

PR #1018, opened 2019-02-11
Remove lodash merge

AndrewMusgrave:

Part of removing lodash #991 

Creating a less featureful `merge` function and using that instead of `_.merge`

### How to 🎩
Test pass? 💯 

### TODO:
* add changelog

### DIFF
--- a/src/utilities/merge.ts
+++ b/src/utilities/merge.ts
@@ -1,0 +1,29 @@
+import {GeneralObject} from '../types';
+
+export default function merge(...objs: GeneralObject[]) {
+  const final = {};
+
+  for (const obj of objs) {
+    mergeRecursively(final, obj);
+  }
+
+  return final;
+}
+
+function mergeRecursively(objA: GeneralObject, objB: GeneralObject) {
+  for (const key in objB) {
+    if (!objB.hasOwnProperty(key)) {
+      continue;
+    } else if (isMergeableValue(objB[key]) && isMergeableValue(objA[key])) {
+      objA[key] = mergeRecursively(objA[key], objB[key]);
+    } else {
+      objA[key] = objB[key];
+    }
+  }
+
+  return objA;
+}
+
+function isMergeableValue(value: any) {
+  return value !== null && typeof value === 'object';
+}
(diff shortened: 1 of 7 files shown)


BPScott:
Repeating my query from https://github.com/Shopify/polaris-react/pull/1010 here too:

I'm not too familiar with how mergeAppProviderOptions() and legacy context stuff does its thing, but do we need a deep merge there  and in the withAppProvider / withSticky HoCs or could we get away with a shallow one using Object.assign()?

AndrewMusgrave:
Unfortunately merging context requires a deep merge since the user could define partial context and have it collide with `polaris`.

src/utilities/merge.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  return final;
+}
+
+function mergeRecursively(objA: GeneralObject, objB: GeneralObject) {
+  for (const key in objB) {
+    if (!objB.hasOwnProperty(key)) {
+      continue;
+    } else if (isMergableValue(objB[key]) && isMergableValue(objA[key])) {
```
solonaarmstrong: `isMergeableValue` ? Sorry to pick on spelling, I just find it easier to understand.

src/utilities/merge.ts
```diff
@@ -0,0 +1,29 @@
...
+  for (const obj of objs) {
+    mergeRecursively(final, obj);
+  }
+
+  return final;
+}
+
+function mergeRecursively(objA: GeneralObject, objB: GeneralObject) {
+  for (const key in objB) {
+    if (!objB.hasOwnProperty(key)) {
```
solonaarmstrong: Would it be worth testing this?

```
   it('returns the first argument if it receives only one argument', () => {
    const objA = {keyA: 1};
    expect(merge(objA)).toEqual(objA);
  });
```

solonaarmstrong:
Elegant! 👌

AndrewMusgrave:
Ready for 👀

dleroux:
I rebased it because conflicts with Michelle's autobind.

dleroux:
I assume we skip change logs for these?

dleroux:
Although I wonder why percy changed.

AndrewMusgrave:
> I assume we skip change logs for these?

Iv'e been adding them to `Code Quality` 😄

PR #1020, opened 2019-02-11
Remove lodash pick

AndrewMusgrave:

Part of removing lodash #991 

Creating a `pick` function and using that instead of `_.pick`

### How to 🎩
Test pass? 💯 

### TODO:
* add changelog

### DIFF
--- a/src/utilities/pick.ts
+++ b/src/utilities/pick.ts
@@ -1,0 +1,41 @@
+import {GeneralObject, TypeOf} from '../types';
+
+function pickValueAndLength(obj: GeneralObject, key: string) {
+  const keyPaths = key.split('.');
+  let value = obj;
+  for (const key of keyPaths) {
+    if (!value.hasOwnProperty(key)) {
+      return null;
+    }
+
+    value = value[key];
+  }
+
+  return {keyPaths, value};
+}
+
+function pick(obj: GeneralObject | null, ...keyPaths: (string | string[])[]) {
+  const flattenedKeypaths = ([] as string[]).concat(...keyPaths);
+  if (obj == null || flattenedKeypaths.length === 0) return {};
+  return flattenedKeypaths.reduce((acc, key) => {
+    if (typeof key !== TypeOf.String || obj.hasOwnProperty(key)) {
+      return {...acc, [key]: obj[key]};
+    }
+
+    const pickedValues = pickValueAndLength(obj, key);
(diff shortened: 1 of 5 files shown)


src/utilities/pick.ts
```diff
@@ -0,0 +1,8 @@
+import {GeneralObject} from '../types';
+
+function pick(obj: GeneralObject | null, keyPaths: string[]) {
+  if (obj == null || keyPaths.length === 0) return {};
+  return keyPaths.reduce((acc, key) => ({...acc, [key]: obj[key]}), {});
```
tmlayton: If a key path is passed that doesn’t exist in the object, it will return the key with value `undefined`. Is that your intent? Either way, I would add a test case for the expectation if a key path is not present in the object.

src/utilities/tests/pick.test.ts
```diff
@@ -0,0 +1,23 @@
...
+    expect(pick({propA: 1}, [''])).toEqual({});
+  });
+
+  it('does not mutate the original object', () => {
+    const obj = {propA: 1};
+    expect(pick(obj, ['propA'])).not.toBe(obj);
+  });
+
+  it('returns a object with values picked from the provided argument', () => {
+    expect(pick({propA: 1, propB: 2, propC: 3}, ['propA', 'propB'])).toEqual({
```
tmlayton:
```suggestion
    expect(pick({propA: 1, propB: 2, propC: 3}, ['propA', 'propC'])).toEqual({
```

src/utilities/tests/pick.test.ts
```diff
@@ -0,0 +1,23 @@
...
+      propA: 1,
+      propB: 2,
```
tmlayton:
```suggestion
      propC: 3,
```

tmlayton:
I’m in favor of this being a simpler, but less feature-rich implementation of lodash pick because we don’t use most of what the lodash pick version supports. But for comparison, here is the lodash test suite for pick running with this version https://codepen.io/timlayton/pen/MLdNrv.

The only thing I could see worth adding is deeply nested picking, but I’m good with shallow picking for now.

AndrewMusgrave:
Sorry addressing feedback took so long, still finishing / catching up from my tour 🙇 I added nested picking and don't include unmatched keypaths. Let me know what you think or if you have any suggestions!

PR #1027, opened 2019-02-13
Add download attribute to Button and UnstyledLink

BPScott:

I've got a use case where I want a Button that triggers the download of a file instead of opening it.

The web provides support for this using the download attribute (see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-download) but there's no way to expose that currently

Adds a `download` prop to the Button and UnstyledLink components. It may either be `true` to render `<a href="foo" download>` or a string to render `<a href="foo" download="downloadPropValue">` which acts as a hint for the filename to use when saving the file.

This only has an effect when you pass a `url` to a button but I think that's ok, as `external` acts in a similar way.

### How to 🎩

On the playground click the various buttons and see the behaviour.

<details>

```jsx
import {Button, Page, UnstyledLink} from '../src';

interface State {}

  render() {
    return (
      <Page title="Playground">
        <Button url="/">Link (no download attribute)</Button>
        <br />
        <Button url="/" download={false}>
          Link (download attribute is false)
        </Button>
        <br />
        <Button url="/" download>
          Download without hint (same as download attribute being true)
        </Button>
        <br />
        <Button url="/" download="page.html">
          Download with hint
        </Button>

        <hr />

        <UnstyledLink url="/">Link (no download attribute)</UnstyledLink>
        <br />
        <UnstyledLink url="/" download={false}>
          Link (download attribute is false)
        </UnstyledLink>
        <br />
        <UnstyledLink url="/" download>
          Download without hint (same as download attribute being true)
        </UnstyledLink>
        <br />
        <UnstyledLink url="/" download="page.html">
          Download with hint
        </UnstyledLink>
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -11,6 +11,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Enhancements
 
 - Updates `TopBar.UserMenu` interaction states styling ([#1006](https://github.com/Shopify/polaris-react/pull/1006))
+- Added `download` prop to `Button` and `UnstyledLink` components that enables setting the download attribute ([#1027](https://github.com/Shopify/polaris-react/pull/1027))
 
 ### Bug fixes
 
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -46,6 +46,8 @@ export interface Props {
   monochrome?: boolean;
   /** Forces url to open in a new tab */
   external?: boolean;
+  /** Tells the browser to download the url instead of opening it. Provides a hint for the downloaded filename if it is a string value. */
+  download?: string | boolean;
   /** Icon to display to the left of the button content */
   icon?: React.ReactNode | IconSource;
   /** Visually hidden text for screen readers */
@@ -91,6 +93,7 @@ function Button({
   onKeyPress,
   onKeyUp,
   external,
+  download,
   icon,
   primary,
   outline,
@@ -183,6 +186,7 @@ function Button({
         id={id}
         url={url}
         external={external}
+        download={download}
         onClick={onClick}
         onFocus={onFocus}
         onBlur={onBlur}
--- a/src/components/UnstyledLink/tests/UnstyledLink.test.tsx
+++ b/src/components/UnstyledLink/tests/UnstyledLink.test.tsx
@@ -39,4 +39,27 @@ describe('<UnstyledLink />', () => {
       expect(anchorElement.prop('rel')).toBe('noopener noreferrer');
     });
   });
+
+  describe('download', () => {
+    it('adds the correct boolean attributes', () => {
+      const anchorElement = mountWithAppProvider(
+        <UnstyledLink download url="https://shopify.com" />,
+      ).find('a');
+      expect(anchorElement.prop('download')).toBe(true);
+    });
+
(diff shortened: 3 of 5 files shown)


UNRELEASED.md
```diff
@@ -10,6 +10,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Added `download` prop to `Button` and `UnstyledLink` components to let you set the download attribute ([#1027](https://github.com/Shopify/polaris-react/pull/1027))
```
solonaarmstrong:
```suggestion
- Added `download` prop to `Button` and `UnstyledLink` components that enables the download attribute ([#1027](https://github.com/Shopify/polaris-react/pull/1027))
```

src/components/Button/Button.tsx
```diff
@@ -46,6 +46,8 @@ export interface Props {
   monochrome?: boolean;
   /** Forces url to open in a new tab */
   external?: boolean;
+  /** Tells the browser to download the url instead of opening it. Value is the hint for the saved filename.. */
```
solonaarmstrong:
```suggestion
  /** Tells the browser to download the url instead of opening it. Value is the hint for the saved filename. */
```

solonaarmstrong: I was expecting `Value` to be a property. Maybe it would be helpful to clarify that the string value is used as the hint?

src/components/UnstyledLink/tests/UnstyledLink.test.tsx
```diff
@@ -39,4 +39,20 @@ describe('<UnstyledLink />', () => {
...
+      expect(anchorElement.prop('download')).toBe(true);
+    });
+
+    it('adds the correct string attributes', () => {
+      const anchorElement = mountWithAppProvider(
+        <UnstyledLink download="file.txt" url="https://shopify.com" />,
+      ).find('a');
+      expect(anchorElement.prop('download')).toBe('file.txt');
+    });
+  });
```
solonaarmstrong: Was it not worth adding a test to check that download is false when not set?

UNRELEASED.md
BPScott: "that enables" feels like it's missing a word. "... components that enables setting the download attribute" feels nicer to me

solonaarmstrong: I was mostly just trying to avoid the use of `you`, but I don't feel strongly about it.

src/components/Button/Button.tsx
BPScott: [We talked on slack, and we have a plan!]

src/components/UnstyledLink/tests/UnstyledLink.test.tsx
BPScott: Added

UNRELEASED.md
BPScott: "that enables setting" instead of "that enables" still removes the usage of "you" :)

BPScott:
Comments addressed! I don't think it is worth adding an example - there's no visual change and finding a non-contrived example might be tricky as `download` only works on same origin requests and there's not really any useful files that both storybook and the styleguide would have in common.

PR #1029, opened 2019-02-13
[TextField] Prevent runaway intervals and non-primary clicks

tmlayton:

Discovered by @AndrewMusgrave [in this comment](https://github.com/Shopify/polaris-react/pull/784#pullrequestreview-198355115) while 🎩’n #784.

- A `mouseup` event outside of the spinner button would not stop an active interval
- A non-primary click would not be filtered out, which could also cause a runaway interval

Ignores non-primary clicks and binds a one-time `mouseup` handler to `document` on `mousedown`. Added tests first which you can see fail, then the code to satisfy the tests.

#### MouseUp outside of spinner
|Before|After|
|---|---|
|infinite|all good|

#### Non-primary click on spinner
|Before|After|
|---|---|
|right click|all good|

## 

### How to 🎩

1. Run the local server and [view the number example](http://localhost:6006/?selectedKind=All%20Components%7CText%20field&selectedStory=Number%20field&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel) 
1. Make sure it doesn’t get stuck in an infinite internal

### 🎩 checklist

* [ ] ~~Tested on [mobile](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting.md#cross-browser-testing)~~
* [ ] ~~Tested on [multiple browsers](https://help.shopify.com/en/manual/intro-to-shopify/shopify-admin/supported-browsers)~~


UNRELEASED.md
```diff
@@ -12,6 +12,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- Fixed a bug with `type="number"` `TextField` where press-and-hold could increment or decrement infinitely ([#1029](https://github.com/Shopify/polaris-react/pull/1029))
```
solonaarmstrong:
```suggestion
- Fixed `type="number"` `TextField` so press-and-hold doesn't increment or decrement infinitely ([#1029](https://github.com/Shopify/polaris-react/pull/1029))
```

https://github.com/Shopify/polaris-react/blob/master/documentation/Versioning%20and%20changelog.md#how-to-write-a-changelog-entry

src/components/TextField/TextField.tsx
```diff
@@ -442,6 +443,10 @@ class TextField extends React.PureComponent<CombinedProps, State> {
       this.buttonPressTimer = window.setTimeout(onChangeInterval, interval);
     };
     this.buttonPressTimer = window.setTimeout(onChangeInterval, interval);
+
+    addEventListener(document, 'mouseup', this.handleButtonRelease, {
+      once: true,
+    });
```
solonaarmstrong: Should this be torn down in `handleButtonRelease` if it will start listening again on subsequent button clicks?

solonaarmstrong:
Works and tests pass.

UNRELEASED.md
tmlayton:
```suggestion
- Fixed `type="number"` `TextField` to prevent conditions where press-and-hold could increment or decrement infinitely ([#1029](https://github.com/Shopify/polaris-react/pull/1029))
```

solonaarmstrong: Looks great!

src/components/TextField/TextField.tsx
tmlayton: The `once` option will tear down the event after it runs one time.

Issue #885, opened 2019-01-16
[Navigation Item] Render inline svg in an image tag


theodoretan:
## Problem
In the `Item` component svg strings are rendered inline with `dangerouslySetInnerHTML`. This can lead to an XSS vulnerability where if an svg with scripts is rendered on a page, they will be executed.

## Solution
We can convert the svg string to a data URI and display it in an `img` tag where it doesn’t execute scripts when rendered.

I propose that the logic that implements this lives within the `Icon` component. This way other users that render svg strings can do so safely. The `Icon` component would have a `color` prop to recolour the image when the item’s state changes (hover/focus/selected). 

Doing this also decouples Navigation Item from Icon as Item relies on the implementation Icon returning an svg tag to re-style. By passing in the fill colour instead of having these states in `Navigation.scss`, the icon that Item is rendering doesn’t break if it’s not an svg tag.

theodoretan:
cc @ragalie

nwtn:
> The Icon component would have a color prop to recolour the image when the item’s state changes (hover/focus/selected).

Just a heads-up that this has the potential to send you down a rabbit-hole. Luckily, it’s one I’ve already explored and prepared you for. When icons are rendered inline in an `img` tag, CSS can’t be applied to elements within the SVG. This means that things like color styling for :hover, :focus, and :selected get complicated.

The best solution we’ve come up with so far is to apply CSS filters on the entire `img` the change the icon’s color when selected. [polaris-tokens](https://github.com/Shopify/polaris-tokens) include `filter` values for all our supported colors to achieve this. More info in the [README](https://github.com/Shopify/polaris-tokens#css-filters).

kaelig:
>  if an svg with scripts is rendered on a page, they will be executed.

Is that something that can happen / has happened?

>We can convert the svg string to a data URI and display it in an img tag where it doesn’t execute scripts when rendered.

Data URIs don't compress well. Just mentioning it so we can take this performance cost into account during the decision-making process.

theodoretan:
> Is that something that can happen / has happened?

Yes, the scrubber that sanitizes our sales channel icons has failed twice already and if we rely on this, we'll always be playing catch-up to new ways of XSS. More context: https://github.com/Shopify/partners/issues/7098

kaelig:
Thanks - I wish this be solved with a CSP directive that specifically declares scripts in inline SVGs are unsafe, but it seems like such fine-grained control isn't possible.

[...]

PR #1042, opened 2019-02-14
Switch to svgr

amrocha:

As part of the solution to #885 we're changing our internal icon loading strategy to use SVGR instead of our own custom loader.

This will allow us to treat any icons that we receive as functions as safe, and any icons that we receive as strings as dangerous, and apply different rendering techniques. 

Adds SVGR, and fixes an issue with our current SVGR implementation.

Our current SVGR implementation accepts a `React.Element`, and renders it. The usage is as follows:
```js
import { ReactComponent as Logo }  from './logo.svg';
import {Icon} from '@shopify/polaris';
[...]
<Icon source={<Logo />} />
```
The problem with this is that `Logo` is not modifiable from the outside. That means we can't set any props on it. Currently, our internal icons rely on a class being set on the `<svg>` element in order to work correctly. By switching to SVGR, we lose the ability to do that with our current implementation.

My proposed change is to accept a `React.Component`-like (because SVGR produces functions that behave alike to `React.Component`s, but doesn't use it as a prototype) instead of a `React.Element`. That allows us to set props before rendering the icon, as such:

```js
import { ReactComponent as Logo }  from './logo.svg';
import {Icon} from '@shopify/polaris';
[...]
<Icon source={Logo} />
```

`Icon.tsx`
```js
if (typeof source === 'function') {
  const SourceComponent = source;
  contentMarkup = <SourceComponent {...defaultIconProps} />;
}
```

 This also solves another issue we had before: our icons are designed for a 20x20 rendering box, but we couldn't set that property for users using SVGR, so we just had to tell them to set it themselves. Now we're able to set it within `polaris-react`

### How to 🎩

Try out all the modified icon rendering paths and make sure they work

### 🎩 checklist


amrocha:
@dleroux this is the PR we talked about on slack, lmk if you have any concerns

amrocha:
Something to keep in mind is that this is a breaking change, but it's also a bug fix. It was a bug that we did not apply the props needed to render certain icons.

If this approach is approved then I'll look into deprecating the old SVGR implementation

src/components/Banner/icons/confetti.svg
```diff
@@ -1 +0,0 @@
-<svg viewBox="0 0 20 20" xmlns="http://www.w3.org/2000/svg"><g fill="none" fill-rule="evenodd"><path d="M1.65 18.329a2.22 2.22 0 0 0 2.33.517L13 14 6 7l-4.867 9a2.22 2.22 0 0 0 .517 2.329" fill="#FFF"/><path d="M6.707 6.293a1.003 1.003 0 0 0-1.587.232l-4.866 9a.986.986 0 0 0-.06.133 3.23 3.23 0 0 0 .75 3.378 3.23 3.23 0 0 0 3.377.749c.045-.016.09-.036.132-.058l9.021-4.846a.998.998 0 0 0 .233-1.588l-7-7zM3.58 17.926a1.222 1.222 0 0 1-1.525-1.524l4.187-7.746 5.101 5.1-7.763 4.17zM15 4a1 1 0 0 1 1 1 1 1 0 1 0 2 0 1 1 0 0 1 1-1 1 1 0 1 0 0-2 1 1 0 0 1-1-1 1 1 0 1 0-2 0 1 1 0 0 1-1 1 1 1 0 1 0 0 2M1 4a1 1 0 0 1 1 1 1 1 0 1 0 2 0 1 1 0 0 1 1-1 1 1 0 1 0 0-2 1 1 0 0 1-1-1 1 1 0 1 0-2 0 1 1 0 0 1-1 1 1 1 0 1 0 0 2m18 12a1 1 0 0 1-1-1 1 1 0 1 0-2 0 1 1 0 0 1-1 1 1 1 0 1 0 0 2 1 1 0 0 1 1 1 1 1 0 1 0 2 0 1 1 0 0 1 1-1 1 1 0 1 0 0-2m-7-7a.997.997 0 0 0 .707-.293l1-1a.999.999 0 1 0-1.414-1.414l-1 1A.999.999 0 0 0 12 9m6.684.052l-3 1a1 1 0 1 0 .633 1.896l3-1a1 1 0 1 0-.633-1.896m-10-4.104a1.001 1.001 0 0 0 1.265-.632l1-3A1 1 0 0 0 9.052.683l-1 3a1 1 0 0 0 .632 1.265" fill="#919EAB"/></g></svg>
```
amrocha: This file wasn't used anywhere

src/components/Icon/Icon.tsx
```diff
@@ -207,25 +207,39 @@ function Icon({
...
   if (source === 'placeholder') {
     contentMarkup = <div className={styles.Placeholder} />;
-  } else if (React.isValidElement(source)) {
-    contentMarkup = source;
+  } else if (typeof source === 'function') {
+    const SourceComponent = source;
+    contentMarkup = <SourceComponent {...defaultIconProps} />;
+  } else if (typeof source === 'string' && isBundledIcon(source)) {
+    // const SourceComponent = BUNDLED_ICONS[source];
+    // contentMarkup = <SourceComponent {...defaultIconProps} />;
```
amrocha: This is the SVGR implementation, below is the old one. I need to hold on the new implementation until we've shipped SVGR in `polaris-icons`, so for now I commented this out and left the old implementation below for ease of testing

dleroux:
I wonder since this is a breaking change if we should we find a more suitable type for the prop other than React.ReactNode?  Maybe: `React.SFC<React.SVGProps<SVGSVGElement>>` would work.

src/components/Icon/Icon.tsx
```diff
@@ -207,25 +207,39 @@ function Icon({
     backdrop && styles.hasBackdrop,
   );
 
+  const defaultIconProps = {
+    className: styles.Svg,
+    viewBox: '0 0 20 20',
```
BPScott: svgr icons coming from polaris-icons already have a viewbox set. I think an svg without a viewbox won't really work, so should we be overwriting it here?

amrocha:
>I wonder since this is a breaking change if we should we find a more suitable type for the prop other than React.ReactNode? Maybe: `React.SFC<React.SVGProps<SVGSVGElement>>` would work.

I'm a bit out of my element when it comes to TS so I'll defer to you, but it seems like SVGR doesn't export a "React Component" per se, it exports a function that behaves like a React Component. This means we can use it like we can use any other React Component, but checks like `isReactComponent` fail. Would we still want to use that type in this case?

src/components/Icon/Icon.tsx
amrocha: We are currently telling people using SVGR to set a 20x20 viewbox, this means we don't have to anymore. I also tested this and it wasn't overwriting anything, if there's already a viewbox defined then it preserves that

amrocha:
@tmlayton @danrosenthal Hey guys, I added two deprecation warnings in this PR. Since I've never done this before can one of you take a look let me know if there's any issues?

src/components/Icon/Icon.tsx
```diff
@@ -172,9 +172,11 @@ export type IconSource =
   | SVGSource
   | 'placeholder'
   | BundledIcon
-  | UntrustedSVG;
+  | UntrustedSVG
+  | React.SFC<React.SVGProps<SVGSVGElement>>;
 export interface Props {
   /** The SVG contents to display in the icon. Icons should be in a 20 X 20 pixel viewbox */
+  /** @deprecated Passing a React Element as source is deprecated. Pass a React Component instead */
```
danrosenthal: It's my understanding that this prop itself is not actually deprecated, only its ability to receive a React Element. Can you clarify? If that's the case, I don't think we want the `deprecated` tag here, only the warning on line 239.

src/components/Icon/Icon.tsx
```diff
@@ -213,15 +216,35 @@ function Icon({
...
+  } else if (
+    typeof source === 'function' &&
+    React.isValidElement(<SourceComponent />)
+  ) {
+    contentMarkup = <SourceComponent {...defaultIconProps} />;
   } else if (React.isValidElement(source)) {
+    // eslint-disable-next-line no-console
+    console.warn(
+      'Deprecation: passing a React Element to the Icon component is deprecated and will be removed in the next major version. Pass a React Component instead.',
+    );
```
danrosenthal: I think we'll want tests for these warnings.

src/components/Icon/Icon.tsx
```diff
@@ -213,15 +216,35 @@ function Icon({
...
     contentMarkup = source;
-  } else if (isBundledIcon(source)) {
-    const iconSource = BUNDLED_ICONS[source] as SVGSource;
-    contentMarkup = renderSVG(iconSource);
   } else if (untrusted && isUntrustedSVG(source)) {
+    // eslint-disable-next-line no-console
+    console.warn(
+      'Deprecation: the untrusted prop is deprecated and will be removed. All raw strings passed into the Icon component will be assumed to be untrusted.',
+    );
```
danrosenthal: Ditto this one

src/components/Icon/Icon.tsx
amrocha: You're right, the prop itself isn't deprecated, just part of its interface 👍

danrosenthal: In that case the `deprecated` tag should be removed here

src/components/Icon/tests/Icon.test.tsx
```diff
@@ -1,3 +1,4 @@
+import {debug} from 'util';
```
danrosenthal: 🔥

config/rollup/index.js
```diff
@@ -41,13 +42,25 @@ module.exports = function createRollupConfig({entry, cssPath}) {
           moduleDirectory: ['../build-intermediate', 'node_modules'],
         },
       }),
+      svgr({
+        exclude: 'node_modules/**',
+        svgoConfig: svgOptions(),
+        replaceAttrValues: {
+          '#FFF': 'currentColor',
```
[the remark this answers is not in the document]
kaelig: I was toying around with CSS Variables and thought we could switch to a more semantic and meaningful technique down the way:

```js
      replaceAttrValues: {
        '#FFF': 'var(--color-icon-secondary, #fff)',
        '#fff': 'var(--color-icon-secondary, #fff)',
        '#212B36': 'var(--color-icon-primary, #212B36)',
        '#212b36': 'var(--color-icon-primary, #212B36)',
      },
```

This way, the code would be self explanatory.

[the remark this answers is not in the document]
amrocha: >Do we need to replace longhand #ffffff and #FFFFFF, too?

Yes, but the optimization pass we do converts those to the shorthand #FFF, so it isn't really necessary

>Are we sure we always want to replace these values, there is never an exception?

I'm personally not a fan of this transform either, but:
- This is a common enough use case that enough people have asked about it over at SVGO (https://github.com/svg/svgo/pull/977). The idea is that you want the SVG file itself to display something, but you also want to use currentColor for styling
- This is the way our icons are currently set up, and not replacing the values could introduce inconsistencies

src/components/Icon/tests/Icon.test.tsx
```diff
@@ -33,7 +34,24 @@ describe('<Icon />', () => {
...
+      expect(iconElement.find(AddMinor)).toHaveLength(1);
+    });
+
+    it('spits out a console warning when rendering a React Element', () => {
+      const spy = jest.spyOn(global.console, 'warn');
+      shallowWithAppProvider(<Icon source={<AddMinor />} />);
+      expect(spy).toHaveBeenCalled();
+    });
+
+    it('renders a React Element when source is given a React Component', () => {
```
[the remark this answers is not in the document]
amrocha: Oh I didn't mean to shorthand, that was just an oversight on my end, but I like your more descriptive line better anyway :D

UNRELEASED.md
```diff
@@ -29,4 +30,9 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Code quality
 
+- Banner, DropZone, and ResourceList’s icons were moved to polaris-icons ([#1042](https://github.com/Shopify/polaris-react/pull/1042))
+
 ### Deprecations
+
+- Passing a React Element into the Icon component is now deprecated ([#1042](https://github.com/Shopify/polaris-react/pull/1042))
```
kaelig: Should we mention that passing a react component instead would work?

amrocha: There's a link to the PR and the warning explains what to do too, I think it's fine to leave as is but if you want to make a suggestion like Tim did then sure I'll commit it

```suggestion
- Passing a React Element into the Icon component is now deprecated in favor of passing a React Component ([#1042](https://github.com/Shopify/polaris-react/pull/1042))
```

kaelig:
```suggestion
- Passing a React Element into the Icon component is now deprecated in favor of passing a React Component ([#1042](https://github.com/Shopify/polaris-react/pull/1042))
```

config/rollup/index.js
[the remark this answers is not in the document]
tmlayton: Something more semantic would be nice because I’m lazy, but if this is good for now I’m fine with leaving it

src/components/Icon/tests/Icon.test.tsx
```diff
@@ -46,6 +64,14 @@ describe('<Icon />', () => {
       expect(element.find('img')).toHaveLength(1);
     });
 
+    it('spits out a console warning when rendering an untrusted SVG', () => {
+      const spy = jest.spyOn(global.console, 'warn');
+      const svg =
+        "<svg><path d="M10 10"/></svg>";
```
tmlayton:
```suggestion
        "<svg><path d='M10 10'/></svg>";
```

Sorry, I suggested the wrong quotes!

tmlayton:
Looks like the `EnableSelectionMinor` icon still isn’t showing up in the resource list select button. Any thoughts there?

amrocha:
Oh hmm, I thought it was, let me double check

amrocha:
@tmlayton I had to update the version of polaris-icons, it should be good now

tmlayton:
Looks like some missing icons

src/components/Icon/Icon.tsx
```diff
@@ -213,15 +215,35 @@ function Icon({
...
+  let SourceComponent = source as React.SFC<React.SVGProps<SVGSVGElement>>;
   let contentMarkup: React.ReactNode;
   if (source === 'placeholder') {
     contentMarkup = <div className={styles.Placeholder} />;
+  } else if (isBundledIcon(source)) {
+    SourceComponent = BUNDLED_ICONS[source];
+    contentMarkup = <SourceComponent {...defaultIconProps} />;
+  } else if (
+    typeof source === 'function' &&
+    React.isValidElement(<SourceComponent />)
```
BPScott: By repeating the <SourceComponent> here in the if clause and again in the if body I think we end up invoking the component twice - can get that down to one by tweaking the ordering of this somehow, or assuming that if `typeof source === function` then it's going to be a React SFC?

src/components/Banner/icons/index.ts
```diff
@@ -1,5 +1,7 @@
-export {default as circleCheckMark} from './circle-check-mark.svg';
-export {default as flag} from './flag.svg';
-export {default as circleAlert} from './circle-alert.svg';
-export {default as circleBarred} from './circle-barred.svg';
-export {default as circleInformation} from './circle-information.svg';
+export {
```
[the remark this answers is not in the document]
amrocha: Our recommended approach to using polaris-icons is an index file so I don't want to reference the library directly like that. I can move these imports to the root level `src/icons/index.ts` though, and reference that file from Banner, while removing any other indexes.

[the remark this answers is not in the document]
BPScott: Yep, that sounds great.

src/components/Icon/Icon.tsx
amrocha: I thought about this too. I do want to keep the `React.isValidElement` check in there though.

Right now, the Icon component just fails silently if you pass in something that it doesn't understand. I want to change that in a follow-up PR and make it throw an exception, but without the `isValidElement` check in there this would accept any function and not be as useful.

What do you think of something like this:
```js
} else if (typeof source ==='function') {
  const functionOutput = (<SourceComponent {...defaultIconProps} />);
  if (React.isValidElement(functionOutput)) {
    contentMarkup = functionOutput;
  }
}
```

amrocha:
shrink-ray is saying the bundle increased because we added a lot more icons to `polaris-icons`, so the package itself is a lot bigger now. The good news is that:
- The unused icons are not included in the bundle of a consumer if they're using tree shaking
- This gzips extremely well, so adding ~50 new icons only resulted in an increase in size of 6KB

amrocha:
This got delayed because an issue came up in web. `sewing-kit` doesn't use our bundle, it recompiles polaris. This recompilation step was causing issues because icons were being imported using different loaders. To deal with this we explored a few solutions and decided to modify Spinner to be consistent with the rest of the SVGs in the repository, which are known to work with `sewing-kit`'s compilation.

Tested in web and it works as expected. 

@BPScott Can you take another look?

kaelig:
In Icon/README.md, should the example `User provided icon` stop using the `untrusted` prop, since it's deprecated?

src/components/Spinner/Spinner.tsx
```diff
@@ -58,12 +59,13 @@ function Spinner({
   const spinnerSVG = size === 'large' ? spinnerLarge : spinnerSmall;
 
   return (
-    <svg
-      viewBox={spinnerSVG.viewBox}
-      dangerouslySetInnerHTML={{__html: spinnerSVG.body}}
+    <Image
+      alt="Spinner"
```
BPScott: I don't think we want this image to be announced to screenreaders. Set alt to an empty string: `alt=""`

src/components/Icon/Icon.tsx
```diff
@@ -213,19 +215,42 @@ function Icon({
...
+  let SourceComponent = source as React.SFC<React.SVGProps<SVGSVGElement>>;
   let contentMarkup: React.ReactNode;
   if (source === 'placeholder') {
     contentMarkup = <div className={styles.Placeholder} />;
+  } else if (isBundledIcon(source)) {
+    SourceComponent = BUNDLED_ICONS[source];
+    contentMarkup = <SourceComponent {...defaultIconProps} />;
+  } else if (typeof source === 'function') {
+    const sourceElement = <SourceComponent {...defaultIconProps} />;
+    if (React.isValidElement(sourceElement)) {
```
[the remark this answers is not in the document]
amrocha: I do want to keep the React.isValidElement check in there.

Right now, the Icon component just fails silently if you pass in something that it doesn't understand. I want to change that in a follow-up PR and make it throw an exception, but without the isValidElement check in there this would accept any function and not be as useful.

PR #1045, opened 2019-02-15
Use index imports for all svg files

BPScott:

To ensure consistency we have introduced [a new eslint rule](https://github.com/Shopify/eslint-plugin-shopify/pull/219) that disables referencing svg files directly and instead ensures you reference them via an index file.

This PR enables that rule (well it's enabled by default in eslint-plugin-shopify 26.2.0) and fixes any current violations

Update to eslint-plugin-shopify 26.2.0 which contains the rule that identified all cases of refering to an image directly instead of via an index file

Fix any violations of that rule.

### How to 🎩

Ensure tests, type-check and percy pass.


src/components/Banner/tests/Banner.test.tsx
```diff
@@ -27,31 +29,31 @@ describe('<Banner />', () => {
 
   it('uses a greenDark successIcon if status is success', () => {
```
solonaarmstrong: It might be clearer to change the test description, as well.

solonaarmstrong:
Other than the test descriptions for drop zone 👍

PR #1047, opened 2019-02-15
[TextField] Fix the weird Backdrop issue on Chrome

alexandcote:
Fixed: #1048

We were facing an issue only on Chrome when we used the `TextField` inside a `Collapsible` and which the wrapper can scroll. The `Backdrop` popped over the `input`when you focus on an input that was fully hidden on the first rendering (you need to scroll) which hide value inside the input (See screenshot below). 

You can check this small [CodePen](https://codepen.io/anon/pen/QYZWrZ) if you want to find which attribute causing the issue 😛 

We change the way we do the animation when we focus a `Textfield`. We add a pseudo element on the `Backdrop` and change the `opacity` instead of animate the `box-shadow`.

**We could solve the problem by adding the `backface-visibility: hidden;` to the input element but it seems more a hack fix.**

### How to 🎩

- Get the focus on the last element of the scrollable area, the content should not disappear. 
- The focus style should be the same for all variations of the `TextField`.

<details>

```jsx
import {
  Page,
  Card,
  Button,
  Scrollable,
  Collapsible,
  FormLayout,
  TextField,
} from '../src';

interface State {
  value: string;
}

  state = {
    value: 'Jaded Pixel',
  };

  handleChange = (value) => {
    this.setState({value});
  };

  render() {
    return (
      <Page title="Playground">
        <Card sectioned>
          <Scrollable style={{height: '420px'}}>
            <Collapsible open id="collapsible">
              <FormLayout>
                <TextField
                  label="Field #1"
                  value={this.state.value}
                  onChange={this.handleChange}
                />
                <TextField label="Field #2" type="number" value="1" />
                <TextField
                  label="Field #3"
                  type="email"
                  value={this.state.value}
                  onChange={this.handleChange}
                />
                <TextField
                  label="Field #4"
                  value={this.state.value}
                  onChange={this.handleChange}
                  multiline
                />
                <TextField
                  label="Field #5"
                  value={this.state.value}
                  onChange={this.handleChange}
                  prefix="123"
                />
                <TextField label="Field #6" value="Partially hidden" />
                <TextField label="Field #7" value="Fully hidden from sight!" />
              </FormLayout>
            </Collapsible>
          </Scrollable>
        </Card>
        <Card sectioned>
          <FormLayout>
            <TextField
              label="Normal"
              value={this.state.value}
              onChange={this.handleChange}
            />
            <TextField
              type="number"
              label="Number"
              value={this.state.value}
              onChange={this.handleChange}
            />
            <TextField
              label="Multiline"
              value={this.state.value}
              onChange={this.handleChange}
              multiline
            />
            <TextField
              label="Error"
              value={this.state.value}
              onChange={this.handleChange}
              error
            />
            <TextField
              label="Disabled"
              value={this.state.value}
              onChange={this.handleChange}
              disabled
            />
            <TextField
              label="Connected Left"
              value={this.state.value}
              onChange={this.handleChange}
              connectedLeft={<Button>Left</Button>}
            />
            <TextField
              label="Connected Right"
              value={this.state.value}
              onChange={this.handleChange}
              connectedRight={<Button>Right</Button>}
            />
            <TextField
              label="Connected"
              value={this.state.value}
              onChange={this.handleChange}
              connectedLeft={<Button>Left</Button>}
              connectedRight={<Button>Right</Button>}
            />
          </FormLayout>
        </Card>
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -20,6 +20,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Fixed the `DataTable` sort direction not reversing on second sort of the initially sorted column ([#918](https://github.com/Shopify/polaris-react/pull/918)) (thanks [@tabrez96](https://github.com/tabrez96) for the [issue report](https://github.com/Shopify/polaris-react/issues/873))
 - Changed the offset from 5px to 4px in `Tooltip` between activator and message to be consistent with `Popover` ([#1019](https://github.com/Shopify/polaris-react/pull/1019))
 - Fixed `Card` header not showing when `title` empty or not set ([#1031](https://github.com/Shopify/polaris-react/pull/1032))
+- Fixed an issue on Chrome when you use a `TextField` inside `Collapsible` which is inside a scrollable element, the text disappeared if you focused a fully hidden `TextField` ([#1047](https://github.com/Shopify/polaris-react/pull/1047))
 
 ### Documentation
 
--- a/src/components/TextField/TextField.scss
+++ b/src/components/TextField/TextField.scss
@@ -1,3 +1,4 @@
+$textfield-backdrop-offset: -(border-width(thick));
 $backdrop-horizontal-spacing: spacing(tight) * 1.5;
 $addon-horizontal-spacing: 0.25em;
 $spinner-icon-size: rem(12px);
@@ -40,8 +41,6 @@ $stacking-order: (
   // stylelint-disable-next-line selector-max-class, selector-max-combinators, selector-max-specificity
   ~ .Backdrop::after {
     opacity: 1;
-    border-color: color('indigo');
-    box-shadow: inset shadow(transparent), 0 0 0 1px color('indigo');
   }
 }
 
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ The format is based on [these versioning and changelog guidelines](https://git.i
 
 - Moved character counter to bottom of multiline text input ([#992](https://github.com/Shopify/polaris-react/pull/992))
 - Aligned `TopBar` search input and results with page content ([#1008](https://github.com/Shopify/polaris-react/issues/1008))
+- Fixed an issue on Chrome when you use a `TextField` inside `Collapsible` which is inside a scrollable element, the text disappeared if you focused a fully hidden `TextField` ([#1047](https://github.com/Shopify/polaris-react/pull/1047))
```
solonaarmstrong: We use the `UNRELEASED.md` file for entries (and then move them over during a release).

solonaarmstrong:
Unrelated, but it looks like a text field in a scrollable slightly cuts off the focus border.

**In scrollable**
screen shot 2019-02-15 at 9 57 35 am

**Normal**
screen shot 2019-02-15 at 9 57 41 am

solonaarmstrong:
It appears as though this bug was introduced here: https://github.com/Shopify/polaris-react-deprecated/pull/636 while trying to solve a stutter with the focus. I believe `will-change` is affecting the stack order of the backdrop.

solonaarmstrong:
❤️ @beefchimi

PR #1059, opened 2019-02-20
[Toast] Add ToastDuration enum

solonaarmstrong:

Resolves https://github.com/Shopify/polaris-react/issues/1055

Adds an enum for Toast duration values. These are the values I chose, but I'm open to doing it differently:

```
Fast = 3000 // Used in web and the old default
Base = 5000 // The default
Slow = 10000 // The minimum duration required to accompany an action
Slower = 15000 // An arbitrary number that's even slower. I added this because we say that toast with action should be at least 10,000, so I thought this might encourage longer persistence. 🤷‍♀️ 
```

## 

### How to 🎩

<details>

```jsx
import {Page, Frame, Toast, Button, ToastDuration} from '../src';

interface State {
  showToast: boolean;
}

  state = {
    showToast: false,
  };

  toggleToast = () => {
    this.setState(({showToast}) => ({showToast: !showToast}));
  };

  render() {
    const {showToast} = this.state;
    const toastMarkup = showToast ? (
      <Toast
        content="Image deleted"
        duration={ToastDuration.Slower}
        onDismiss={this.toggleToast}
      />
    ) : null;
    return (
      <Page title="Playground">
        <Frame>
          <Button onClick={this.toggleToast}>Show Toast</Button>
          {toastMarkup}
        </Frame>
      </Page>
    );
  }
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Frame/types.ts
+++ b/src/components/Frame/types.ts
@@ -1,4 +1,5 @@
 import * as PropTypes from 'prop-types';
+import {Action} from '../../types';
 
 export interface FrameManager {
   showToast(toast: {id: string} & ToastProps): void;
@@ -63,4 +64,6 @@ export interface ToastProps {
   error?: boolean;
   /** Callback when the dismiss icon is clicked */
   onDismiss(): void;
+  /** Adds an action next to the message (stand-alone app use only) */
+  action?: Action;
 }
--- a/src/components/Frame/types.ts
+++ b/src/components/Frame/types.ts
@@ -52,6 +52,13 @@ export interface ContextualSaveBarProps {
 
 // Toast
 
+export enum ToastDuration {
+  Fast = 3000,
+  Base = 5000,
+  Slow = 10000,
+  Slower = 15000,
+}
+
 export interface ToastProps {
   /** The content that should appear in the toast message */
   content: string;
--- a/src/components/Toast/README.md
+++ b/src/components/Toast/README.md
@@ -68,10 +68,18 @@ class EmbeddedAppToastExample extends React.Component {
 
 Toast should:
 
-- Be used for short messages to confirm an action. Maximum of 2 lines of text.
-- Not be used for actionable links or messages.
-- Not be used for error messages.
-- Be displayed once at the time. If you need multiple toasts, queue them.
+- Be used for short messages to confirm an action
+- Not go over 3 words
+- Rarely be used for error messages
+
+When to use:
+
+- For success messages
+- Only for non-critical errors that are relevant in the moment and can be explained in 3 words. For example, if there’s an internet connection issue, the toast would say, Internet disconnected.
+
+When not to use:
+
+- Avoid using toast for error messages. Always try to use a banner to prominently inform merchants about persistent errors.
 
 ---
 
@@ -79,7 +87,7 @@ Toast should:
 
 ### Message
 
-Messages should be:
+Toast messages should be:
 
 - Short and affirmative
 - Written in the pattern of: noun + verb
@@ -91,10 +99,14 @@ Messages should be:
 - Product updated
 - Collection added
 - Customer updated
-- No internet connection
+- Internet disconnected
+- Connection timed out
 
 #### Don’t
 
+- No internet connection
+- Can’t charge negative tax rates
+- Your online store has a maximum of 20 themes. Delete unused themes to add more.
 - Your product has been successfully updated
 - We were unable to save the customer
 - Your Order was Archived Today
@@ -102,12 +114,18 @@ Messages should be:
 
 <!-- end -->
 
-### Action
+### Toast with action
+
+Only include an action in toast if the same action is available elsewhere on the page. For example:
+
+- If merchants need to reload a section, offer the call to action [Reload] in the toast. If they miss the toast message, they can also refresh the entire page.
+- If merchants delete an image, offer the option to [Undo] the deletion. If they miss it in the toast message, they can still retrieve it from somewhere else.
 
 Action should:
 
-- Keep the action label short. Preferably 1 verb
-- Not have actions for dismissing toast
+- Keep the action label short, preferably 1 verb.
+- Not have actions, like [Cancel], for dismissing toast. The [X] to dismiss is already included in the component.
+- Be used with a duration of at least 10,000 milliseconds for accessibility.
 
 <!-- usagelist -->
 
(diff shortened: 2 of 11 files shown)


src/components/Toast/README.md
```diff
@@ -236,7 +236,7 @@ class ToastExample extends React.Component {
       <Toast
         content="Message sent"
         onDismiss={this.toggleToast}
-        duration={4500}
+        duration={Duration.Fast}
```
BPScott: I think this needs to be `ToastDuration` - to match the name of the export in Polaris otherwise our examples might complain about an unknown variable

solonaarmstrong: Oh good grief. 🤦‍♀️ Thank you

mateus:
You might also wanna update the `DEFAULT_TOAST_DURATION` on the `Toast.tsx` to use `ToastDuration`

https://github.com/Shopify/polaris-react/blob/8f7392d33dbce11e09455ae96f24b7c5d023f546/src/components/Frame/components/Toast/Toast.tsx#L12

src/components/Toast/README.md
```diff
@@ -236,7 +236,7 @@ class ToastExample extends React.Component {
       <Toast
         content="Message sent"
         onDismiss={this.toggleToast}
-        duration={4500}
+        duration={ToastDuration.Fast}
```
ry5n: I could be wrong, but it strikes me as an odd API to import the duration enum and use it this way, as props to the component it came from. Since these delay values are intended only for Toast, I would expect instead for the `duration` prop to accept the keywords directly.

For example, I would expect to write `<Toast duration="fast" />` like `<Button size="slim" />`, while still being able to write `<Toast duration={6000} />`.

ry5n: We can still export the enum—we seem to export `type Size` from `Button`, but I would expect it to be rarely imported.

AndrewMusgrave: 👍

solonaarmstrong: @AndrewMusgrave Is your 👍 in reference to reverting back to a number?

AndrewMusgrave: Sorry I was agreeing with Ryan on the api. I would expect the usage to be `duration="fast"` or `duration={5000}` rather than forcing consumers to import out enum

solonaarmstrong: Thank you :)

src/components/Frame/types.ts
```diff
@@ -51,6 +51,13 @@ export interface ContextualSaveBarProps {
 
 // Toast
 
+export enum ToastDuration {
+  Fast = 3000,
+  Base = 5000,
```
AndrewMusgrave: How do we feel about using `base` as a value? We use it a lot for our sass functions, but it's a default. So we would use `spacing()` rather than `spacing(base)`

solonaarmstrong: I'm not clear on how a px value relates to a ms value?

AndrewMusgrave: It doesn't I was just using it for comparison. Since we would be using `duration="base"`, which I don't find very descriptive.

solonaarmstrong: Ohhh! When you asked how we feel about base, I thought you were suggesting to use it and was quite confused. Now I understand what you're saying.

Would it make more sense to call this `Default`? Or?

solonaarmstrong:
What about:

```
Fast = 3000
Default = 5000
Slow = 7000
Slower = 10000
Slowest = 15000
```

src/components/Frame/types.ts
solonaarmstrong: What about:

```
Fast = 3000
Default = 5000
Slow = 7000
Slower = 10000
Slowest = 15000
```

AndrewMusgrave:
```tsx
enum Duration {
  Fastest = 1000
  Fast = 3000
  Default = 5000
  Slow = 7000
  Slower = 9000
}
```

Feels pretty natural to me, would love to hear others thoughts on it. Since if we ever implement another component with a duration prop we'll *probably* want to reuse this enum

solonaarmstrong: We increased the default duration for accessibility reasons from 3000 to 5000. Some toasts exist in web with 3000 as a duration because it used to be the default. But I don't know why we would go any less than 3000. 1000ms is too fast for many people to actually see it and would have accessibility issues.

We need 10000 for the action prop: https://github.com/Shopify/polaris-react/pull/919/files#diff-f6f39e05b67f0498645df16bf2a7c4eaR460

src/components/Toast/README.md
BPScott: This is a good call. I don't think we use enums in other places, and instead passing in strings to the prop is how we do this in other places (e.g. `<Button size="">`)

ry5n:
@mateus I know you’ve reviewed this, but do you want to pick it up?

mateus:
After playing with it for a little bit here's what I found.

One way to use an enum but still have something similar to what we have for `<Button size="" />` would be by making a list based on the keys of the enum. This would create a list with Capital elements that I'm not sure you guys wanna use since it's not the pattern on the other components.

```ts
(property) ToastProps.duration?: number | "Fast" | "Base" | "Slow" | "Slower" | undefined
```

```ts
export enum ToastDurationEnum {
  Fast = 3000,
  Base = 5000,
  Slow = 10000,
  Slower = 15000,
}

export type ToastDuration = keyof typeof ToastDurationEnum;

export interface ToastProps {
  /** The content that should appear in the toast message */
  content: string;
  /**
   * The length of time in milliseconds the toast message should persist
   * @default 5000
   */
  duration?: ToastDuration | number;
  /** Display an error toast. */
  error?: boolean;
  /** Callback when the dismiss icon is clicked */
  onDismiss(): void;
}
```

If you wanna see what is necessary to have that working [check this brach](https://github.com/Shopify/polaris-react/compare/toast-duration). Thoughts?

mateus:
Now if we are ok not using enums we can simply write an object and get its keys like this:

```ts
(property) ToastProps.duration?: number | "fast" | "base" | "slow" | "slower" | undefined
```

```ts
export const ToastDurationMap = {
  fast: 3000,
  base: 5000,
  slow: 10000,
  slower: 15000,
};

export type ToastDuration = keyof typeof ToastDurationMap;

export interface ToastProps {
  /** The content that should appear in the toast message */
  content: string;
  /**
   * The length of time in milliseconds the toast message should persist
   * @default 5000
   */
  duration?: ToastDuration | number;
  /** Display an error toast. */
  error?: boolean;
  /** Callback when the dismiss icon is clicked */
  onDismiss(): void;
}
```

The rest of the implementation remains basically the same.

ry5n:
Hmmm, I’m not familiar enough with the codebase or with TypeScript to have an opinion here. Can you ping some Polaris team members?

mateus:
@danrosenthal mind having a look? You can basically start from https://github.com/Shopify/polaris-react/pull/1059#issuecomment-468076388

danrosenthal:
@mateus sorry for the delay on getting back to you on this. Would you mind rebasing this PR so that we get a fresh deploy to heroku?

Code is looking good, just want to be able to :tophat: there and then I will 👍 .

PR #1063, opened 2019-02-20
[DropZone] Use createRef() API 

whitetig3r:
These changes capitalize on the enhancements made for refs in >React 16.3.
Resolves https://github.com/Shopify/polaris-react/issues/910

The refs to the DOM nodes are now initialized using createRef(). These refs are linked to the respective DOM elements in the render method, through the ref prop. The refs are now accessed using _this.refname.current_. There are no visible UI changes.

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -47,6 +47,7 @@ Upgraded Storybook to v5 ([#1140](https://github.com/Shopify/polaris-react/pull/
 
 ### Code quality
 
+- Migrated the refs in `DropZone` to use the new createRef API ([#1063](https://github.com/Shopify/polaris-react/pull/1063))
 - Updated `ResourceList` to no longer use `componentWillReceiveProps`([#1235](https://github.com/Shopify/polaris-react/pull/1235))
 - Updated `Tabs` to no longer use `componentWillReceiveProps`([#1221](https://github.com/Shopify/polaris-react/pull/1221))
 - Removed an unneeded media query from Modal's `Header` component ([#1272](https://github.com/Shopify/polaris-react/pull/1272))
--- a/src/components/DropZone/DropZone.tsx
+++ b/src/components/DropZone/DropZone.tsx
@@ -153,19 +153,18 @@ export class DropZone extends React.Component<CombinedProps, State> {
     return Object.keys(newState).length ? newState : null;
   }
 
-  private node: HTMLElement | null = null;
-  private dropNode: HTMLElement | HTMLDocument | null = null;
+  private node = React.createRef<HTMLDivElement>();
   private dragTargets: EventTarget[] = [];
-  private fileInputNode: HTMLInputElement;
+  private fileInputNode = React.createRef<HTMLInputElement>();
 
   private adjustSize = debounce(
     () => {
-      if (!this.node) {
+      if (!this.node.current) {
         return;
       }
 
       let size = 'extraLarge';
-      const width = this.node.getBoundingClientRect().width;
+      const width = this.node.current.getBoundingClientRect().width;
 
       if (width < 100) {
         size = 'small';
@@ -211,6 +210,10 @@ export class DropZone extends React.Component<CombinedProps, State> {
     };
   }
 
+  get dropNode() {
+    return this.props.dropOnPage ? document : this.node.current;
+  }
+
   render() {
     const {
       id,
@@ -239,7 +242,7 @@ export class DropZone extends React.Component<CombinedProps, State> {
       disabled,
       type: 'file',
       multiple: allowMultiple,
-      ref: this.setInputNode,
+      ref: this.fileInputNode,
       onChange: this.handleDrop,
       autoComplete: 'off',
     };
@@ -291,7 +294,7 @@ export class DropZone extends React.Component<CombinedProps, State> {
 
     const dropZoneMarkup = (
       <div
-        ref={this.setNode}
+        ref={this.node}
         className={classes}
         aria-disabled={disabled}
         onClick={this.handleClick}
@@ -339,6 +342,8 @@ export class DropZone extends React.Component<CombinedProps, State> {
     addEventListener(this.dropNode, 'dragleave', this.handleDragLeave);
     addEventListener(window, 'resize', this.adjustSize);
 
+    this.adjustSize();
+
     if (this.props.openFileDialog) {
       this.triggerFileDialog();
     }
@@ -371,11 +376,11 @@ export class DropZone extends React.Component<CombinedProps, State> {
   };
 
   private open = () => {
-    if (!this.fileInputNode) {
+    if (!this.fileInputNode.current) {
       return;
     }
 
-    this.fileInputNode.click();
+    this.fileInputNode.current.click();
   };
 
   private getValidatedFiles = (files: File[] | DataTransferItem[]) => {
@@ -405,19 +410,6 @@ export class DropZone extends React.Component<CombinedProps, State> {
       acceptedFiles,
       rejectedFiles,
     };
-  };
-
-  private setNode = (node: HTMLElement | null) => {
-    const {dropOnPage} = this.props;
-
-    this.node = node;
-    this.dropNode = dropOnPage ? document : node;
-
-    this.adjustSize();
-  };
-
-  private setInputNode = (node: HTMLInputElement) => {
-    this.fileInputNode = node;
   };
 
   private handleClick = (event: React.MouseEvent<HTMLElement>) => {


src/components/DropZone/DropZone.tsx
```diff
@@ -151,10 +151,10 @@ export class DropZone extends React.Component<CombinedProps, State> {
     return Object.keys(newState).length ? newState : null;
   }
 
-  private node: HTMLElement | null = null;
-  private dropNode: HTMLElement | HTMLDocument | null = null;
+  private node: React.RefObject<HTMLDivElement>;
+  private dropNode: React.RefObject<HTMLDivElement> | Document;
```
AndrewMusgrave: Why did you change the type from `HTMLDocument` to `Document`?

src/components/DropZone/DropZone.tsx
```diff
@@ -177,6 +177,9 @@ export class DropZone extends React.Component<CombinedProps, State> {
       errorOverlayText: translate(`Polaris.DropZone.errorOverlayText${suffix}`),
       numFiles: 0,
     };
+    this.node = React.createRef();
```
AndrewMusgrave: Rather than creating the refs in the constructor we can do it above

src/components/DropZone/DropZone.tsx
```diff
@@ -151,10 +151,10 @@ export class DropZone extends React.Component<CombinedProps, State> {
     return Object.keys(newState).length ? newState : null;
   }
 
-  private node: HTMLElement | null = null;
-  private dropNode: HTMLElement | HTMLDocument | null = null;
+  private node: React.RefObject<HTMLDivElement>;
```
AndrewMusgrave:
```suggestion
  private node: React.createRef<HTMLDivElement>();
```

src/components/DropZone/DropZone.tsx
```diff
@@ -304,30 +307,58 @@ export class DropZone extends React.Component<CombinedProps, State> {
 
     this.setState({error: this.props.error});
 
-    if (!this.dropNode) {
+    const dropNode: HTMLDivElement | Document | null =
```
AndrewMusgrave: Typescript should infer the return type so we can omit it

src/components/DropZone/DropZone.tsx
```diff
@@ -304,30 +307,58 @@ export class DropZone extends React.Component<CombinedProps, State> {
...
+      'current' in this.dropNode ? this.dropNode.current : document;
```
AndrewMusgrave: Could we change this to
```tsx
const dropNode = this.dropNode.current || document;
```

src/components/DropZone/DropZone.tsx
```diff
@@ -304,30 +307,58 @@ export class DropZone extends React.Component<CombinedProps, State> {
...
+
+    if (!dropNode) {
       return;
     }
 
-    addEventListener(this.dropNode, 'drop', this.handleDrop);
-    addEventListener(this.dropNode, 'dragover', this.handleDragOver);
-    addEventListener(this.dropNode, 'dragenter', this.handleDragEnter);
-    addEventListener(this.dropNode, 'dragleave', this.handleDragLeave);
+    addEventListener(dropNode as EventTarget, 'drop', this.handleDrop);
```
AndrewMusgrave: Any idea why we have to assert the type? Ideally we shouldn't have to force it.

src/components/DropZone/DropZone.tsx
```diff
@@ -304,30 +307,58 @@ export class DropZone extends React.Component<CombinedProps, State> {
...
+
     if (this.props.openFileDialog) {
       this.triggerFileDialog();
     }
   }
 
   componentWillUnmount() {
-    if (!this.dropNode) {
+    const dropNode: HTMLDivElement | Document | null =
+      'current' in this.dropNode ? this.dropNode.current : document;
```
AndrewMusgrave: Same as above 😄

src/components/DropZone/DropZone.tsx
```diff
@@ -541,7 +557,12 @@ export class DropZone extends React.Component<CombinedProps, State> {
     }
 
     this.dragTargets = this.dragTargets.filter((el: Node) => {
-      return el !== event.target && this.dropNode && this.dropNode.contains(el);
+      return (
+        el !== event.target &&
+        'current' in this.dropNode &&
```
AndrewMusgrave: We can omit this check since we do it below

src/components/DropZone/DropZone.tsx
whitetig3r: Thanks for pointing that out. Will change all usage of `Document` back to `HTMLDocument` 😄

src/components/DropZone/DropZone.tsx
whitetig3r: Will move the refs above!

src/components/DropZone/DropZone.tsx
whitetig3r: Will add these changes! 😄 

eg: `private node = React.createRef<HTMLDivElement>()` like this?

src/components/DropZone/DropZone.tsx
whitetig3r: Yes! 👍  That's right! Thank you! Will omit that check.

src/components/DropZone/DropZone.tsx
whitetig3r: `[ts]
Property 'current' does not exist on type 'RefObject<HTMLDivElement> | Document'.
  Property 'current' does not exist on type 'Document'.`

I think that a type check seems to be imperative in this case, before accessing the current property

src/components/DropZone/DropZone.tsx
whitetig3r: `[ts]
Argument of type 'HTMLDivElement | Document | null' is not assignable to parameter of type 'EventTarget'.
  Type 'null' is not assignable to type 'EventTarget'. `

I guess this has to do with the definitions of the event listeners in `@shopify/javascript-utilities/events` which requires that the first argument be of type _EventTarget_.

Any other workarounds? 😄

src/components/DropZone/DropZone.tsx
whitetig3r: Will remove that check! 😄

src/components/DropZone/DropZone.tsx
whitetig3r: `[ts] Object is possibly 'null'.`

Is the check necessary since,`in` just checks existence while `this.dropNode.current` checks whether it is null before trying to see whether it contains `el` ?

src/components/DropZone/DropZone.tsx
```diff
@@ -152,10 +152,13 @@ export class DropZone extends React.Component<CombinedProps, State> {
     return Object.keys(newState).length ? newState : null;
   }
 
-  private node: HTMLElement | null = null;
-  private dropNode: HTMLElement | HTMLDocument | null = null;
+  private node = React.createRef<HTMLDivElement>();
+  private dropNode = this.props.dropOnPage
```
AndrewMusgrave: I don't think we need to use `createRef` here since it's never actually it's own ref. It's either document or `this.node`. This assignment will also only be ran once opposed to every render.

src/components/DropZone/DropZone.tsx
```diff
@@ -327,30 +330,36 @@ export class DropZone extends React.Component<CombinedProps, State> {
 
     this.setState({error: this.props.error});
 
-    if (!this.dropNode) {
+    const dropNode = this.dropNode.current || document;
```
AndrewMusgrave: An alternative solution that may fix the problem above is user a `getter`. Rather than a `private property` we can use a dynamic value. Something like the example below may work.

```tsx
private get dropNode() {
  return this.props.dropOnPage && this.node.current ? document : null;
}
```

src/components/DropZone/DropZone.tsx
```diff
@@ -327,30 +330,36 @@ export class DropZone extends React.Component<CombinedProps, State> {
...
-    addEventListener(this.dropNode, 'dragover', this.handleDragOver);
-    addEventListener(this.dropNode, 'dragenter', this.handleDragEnter);
-    addEventListener(this.dropNode, 'dragleave', this.handleDragLeave);
+    addEventListener(dropNode, 'drop', this.handleDrop);
+    addEventListener(dropNode, 'dragover', this.handleDragOver);
+    addEventListener(dropNode, 'dragenter', this.handleDragEnter);
+    addEventListener(dropNode, 'dragleave', this.handleDragLeave);
     addEventListener(window, 'resize', this.adjustSize);
 
+    this.adjustSize();
```
AndrewMusgrave: Previously this was being called every render now it's just being called in `componentDidMount`

src/components/DropZone/DropZone.tsx
whitetig3r: Thank you for pointing that out! Yes, it would be quite redundant.

src/components/DropZone/DropZone.tsx
whitetig3r: That's a perfect fix! I've added that change. 😄

src/components/DropZone/DropZone.tsx
whitetig3r: After some study on ref callbacks, I've understood that ref callbacks are called before the component mounts (if the component is rendered) and not on every render. So, with the new refs API, it would be best to call it in ComponentDidMount alone?

whitetig3r:
I've been unable to fresh a review, for some reason. 😅

AndrewMusgrave:
> I've been unable to fresh a review, for some reason.

Weird, it's a new tool so there's bound to be some bugs. I'll refresh myself 😄

src/components/DropZone/DropZone.tsx
```diff
@@ -152,10 +152,13 @@ export class DropZone extends React.Component<CombinedProps, State> {
     return Object.keys(newState).length ? newState : null;
   }
 
-  private node: HTMLElement | null = null;
-  private dropNode: HTMLElement | HTMLDocument | null = null;
+  private node = React.createRef<HTMLDivElement>();
+  private get dropNode() {
```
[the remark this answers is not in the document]
whitetig3r: Makes it look neat! 😄 Added that change!

whitetig3r:
> This is looking really good and almost ready to 🚢 Would you mind rebasing with master to resolve the conflicts and squash the commits down 😄

That's awesome! Done rebasing with `master` and squashed down to one commit! 😄

UNRELEASED.md
```diff
@@ -22,4 +22,6 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Code quality
 
+- Migrated the refs in `DropZone` to use the new createRef API ([#910](https://github.com/Shopify/polaris-react/issues/910))
```
```suggestion
- Migrated the refs in `DropZone` to use the new createRef API ([#1063](https://github.com/Shopify/polaris-react/pull/1063))
```

whitetig3r:
> Sorry this has taken me so long to re-review @whitetig3r 🙏 . I wasn't aware it was ready for review, I'm guessing there are still issues with refreshing review? Looks like there's a conflict again in the changelog, and suggested a small change to match the pull request. Feel free to @ me when you'r ready for this to be merged. Thanks for your contribution 🎉 🎉

That's absolutely fine 😄 . Yes, I guess it's because I'm unable to request a review! Sorry for the inconsistency with the changelog addition. I've resolved that and squashed down for a clean commit history. Thanks a ton for the code-reviews @AndrewMusgrave ! Glad to have been able to contribute! 😄

AndrewMusgrave:
I was going to merge this before more conflicts pop up but `codecov` seems to be stuck, will check back later

whitetig3r:
> I was going to merge this before more conflicts pop up but `codecov` seems to be stuck, will check back later

Oh oh! It's been stuck for quite a while now. Should we wait on it or do a `rebase` and `force push` to make it cycle through all the checks again @AndrewMusgrave ?

AndrewMusgrave:
Probably @whitetig3r , Iv'e checked `codecov` and the task finished but it seems github never received a finished status for the check

whitetig3r:
> Probably @whitetig3r , Iv'e checked `codecov` and the task finished but it seems github never received a finished status for the check

Good to merge or should anything else be done? :smiley:

AndrewMusgrave:
> Good to merge or should anything else be done?

Seemed to be stuck again 😬 so I force merged it. Thanks for your contribution 😄 🎉

whitetig3r:
> > Good to merge or should anything else be done?
> 
> Seemed to be stuck again  so I force merged it. Thanks for your contribution

Happy to have been able to contribute and hoping to contribute more! Thank you for the guidance :smiley:  @AndrewMusgrave

PR #1069, opened 2019-02-20
Add a retry on failure of update version, PR for update Polaris on multiple repos

### LINKED ISSUE #1066: Open styleguide PR fails when latest version is not ready on npm
# Issue summary

We just did a deploy for version `3.8.0`. Everything went smoothly until the creation of the branch in `polaris-styleguide`. 

This failed as the `npm publish` script had completed however the npm registry had not updated the version. This meant that when running `npx yarn upgrade @shopify/polaris@3.8.0` it failed as it could not find the version. Testing the version now locally there is no problem.
( scroll down to the bottom of this link )
https://shipit.shopify.io/shopify/polaris-react/production/deploys/723859

This would also cause an error if the registry was ever down or experiencing some minor outages.

## Expected behavior

When we update `polaris-react` on the new branch in the `polaris-styleguide` if there are any issues we should pause and try again ( maybe a few times ). After the re-attempts fail it should throw an error as the registry may be having issues.

https://github.com/Shopify/polaris-react/blob/1cf3f2a70b43584a6518898053856b4709e28bbe/scripts/open-styleguide-pr.js#L73-L79

## Actual behavior

If the registry does not have the specified version at the time of update the branch creation fails.

This is a bit tricky as sometimes the registry will be working fast and other times it might be slow or offline.

---

Also I am happy to do a small PR to catch when the update script fails and try again after a delay. Just want to confirm this is an issue first.

### LINKED ISSUE #1067: Add step that automatically opens a PR in web upon a successful deploy
We already do this for the styleguide, let's just do it for web as well.
We should look into #1066 at the same time

alex-page:

Fixes #1067 and fixes #1066

This allows the update of polaris npm packages to fail and retry when creating a branch to update the styleguide or other projects.

It also gives new functionality to allow us to specify repositories where we want to automatically create a branch and update the polaris npm package.

### Testing

I am not too sure about the best way to test these changes, maybe someone can help.


scripts/new-version-pr-generator.js
```diff
@@ -0,0 +1,193 @@
...
+`.trim();
+
+const updatePostObject = {
+  title: `Update @shopify/polaris to ${releaseVersion} 🚀`,
+  body: updateBody,
+  head: `update-polaris-${releaseVersion}`,
+  base: baseBranch,
+};
+
+ExecSyncQueue(
```
amrocha: Does this work? The function is called `execSyncQueue` without capitalization

amrocha:
I see why they're executing at the same time here
```
ids.map((id) => {
    return new Promise((resolve, reject) => {
```

This line is the culprit. Since the promises created here are an async object, they all get created at the same time and start running at the same time.

I think a better approach would be to get rid of `execSyncQueue`, and instead make a single function, let's say `makeRequest()`, that returns a promise. You can then chain these promises as such:
```
makeRequest()
  .then(() => makeRequest())
  .then(() => makeRequest())
```

etc

I also think this is more readable, and more idiomatic in JS. What do you think of this approach?

scripts/new-version-pr-generator.js
alex-page: Nope I have cleaned this up, my bad!

kaelig:
I just had a though: is there a reason why we retry running the whole flow (which could fail for other reasons than the npm registry not being up to date yet)?

### Flow in this PR

(if I understand correctly)

1. Run all steps
1. Retry x times if anything fails

### Could this be simpler?

1. Wait until `@shopify/polaris` is available on npm (using similar to `npm info @shopify/polaris@${VERSION}`)
1. Run installation steps

---

I'm thinking it would ensure minimal changes to the current scripts, which could be useful as this flow is a hard thing to test.

amrocha:
This is a WIP PR, the only reason Alex pushed it is because I asked so I could help him out. The approach we take might still change

>is there a reason why this PR would retry running the whole flow

Yes, network operations are flaky and can fail randomly

>rather than retrying lookups on npm and only then the flow once?

Because then the errors could still happen if any other request fails

>this flow is a hard thing to test

We're not changing any of the operations, just how they're run. We can also do mockup tests where we use no-ops for network responses.

We're having these errors because we didn't think of edge cases when first writing this script. Let's not repeat that mistake and make it resilient this time.

alex-page:
Markdown needs to be fixed up, the changing of the file has broken the indentation!

alex-page:
Hey friends, I amhappy with this now but I am not too sure the best way to test this is working with bots etc. any recommendations or reviews would be appreciated.

@kaelig @amrocha

scripts/new-version-pr-generator.js
```diff
@@ -0,0 +1,180 @@
...
+
+const {version: PACKAGE_VERSION} = require('../package.json');
+const secrets = require('../secrets.json');
+const Retry = require('./utilities/retry');
+
+const YARN_VERSION = yaml
+  .safeLoad(readFileSync(pathResolve(__dirname, '..', 'dev.yml'), 'utf8'))
+  .up.find((obj) => Object.keys(obj).includes('node')).node.yarn;
+
+// File paths and directories
```
amrocha: Can you all the comments in this file? We prefer to write self-evident code, and only comment when there's something really unusual going on. Comments like the ones in this file actually make the code less readable by breaking the logical flow and being visual noise

scripts/new-version-pr-generator.js
```diff
@@ -0,0 +1,180 @@
...
+const polarisBotToken = secrets.github['shopify-polaris'];
+const shopifyPolarisBotGitOverride = `GIT_COMMITTER_NAME='${polarisBotName}' GIT_COMMITTER_EMAIL='${polarisBotEmail}'`;
+
+/**
+ * BulkExecute - Run a bunch of execSync one at a time
+ *
+ * @param {array}  tasks   - List of tasks to run
+ * @param {object} options - Specific options for execSync
+ */
+function BulkExecute(tasks, options) {
```
amrocha: I don't think this function is necessary. You're running it once so it doesn't save you any code, and it's simple enough that it doesn't benefit from being encapsulated either. What do you think of removing it?

scripts/new-version-pr-generator.js
```diff
@@ -0,0 +1,180 @@
...
+
+// Iterate through the repositories
+const tasks = repositories.map((repository) => {
+  // Create a new promise for each repository
+  return new Promise((resolve, reject) => {
+    try {
+      const repositoryDirectory = pathResolve(sandbox, repository);
+
+      // Message that is logged on a failed update
+      const failedUpdateMessage = `It seems your automatic branch creation for ${repository} failed. This can be due to many reasons. To resolve this follow these steps and replace [release-version] with the version you are publishing:
```
amrocha: Can you make this a JS template and move it to the top of the file? This big chunk of text in the middle of the logic makes it hard to understand what's going on

scripts/new-version-pr-generator.js
```diff
@@ -0,0 +1,180 @@
...
+      2. Get the latest version of master "git checkout master && git pull"
+      3. Create a new branch based on the version that failed "git checkout -b update-polaris-v[release-version]"
+      4. Update "shopify/polaris" by running "npx yarn upgrade @shopify/polaris@[release-version]"
+      5. Add the changed files to git by running "git add package.json yarn.lock"
+      6. Commit the changes with "git commit -m 'Update @shopify/polaris to [release-version]'
+      7. Push the changes with "git push origin update-polaris-v[release-version]"
+      8. Create a new pull request for the branch in "${repository}"`;
+
+      // Pull request template
+      const updateBody = `
```
amrocha: Same here, make it a template and move it up

scripts/new-version-pr-generator.js
```diff
@@ -0,0 +1,180 @@
...
+    \`\`\`bash
+    git push origin update-polaris-${releaseVersion}
+    \`\`\`
+1. Repeat until the build & tests go ✅
+1. 🎩
+1. :shipit:
+</details>
+`.trim();
+
+      const updatePostObject = {
```
amrocha: I think this object can just be inlined, what do you think? You're only using it once, might as well create it where it's actually going to be used

scripts/new-version-pr-generator.js
```diff
@@ -0,0 +1,180 @@
...
+        title: `Update @shopify/polaris to ${releaseVersion} 🚀`,
+        body: updateBody,
+        head: `update-polaris-${releaseVersion}`,
+        base: baseBranch,
+      };
+
+      // Clone the repository to the repository directory
+      execSync(
```
amrocha: Why is this not defined as a value in `commands`?

scripts/new-version-pr-generator.js
```diff
@@ -0,0 +1,180 @@
...
+        ],
+        {
+          cwd: repositoryDirectory,
+          stdio: 'inherit',
+        },
+      );
+
+      // Try and run update polaris using retry default settings
+      // 3 attempts maximum with a 10 second delay between each attempt
+      Retry(commands, failedUpdateMessage);
```
amrocha: This isn't doing what we want it to.

Right now, if the last request fails you retry every single one of these requests. What we want is to retry that individual request.

scripts/utilities/retry.js
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Retry - Retry a function on failure
+ *
+ * @param {function} Run          - The function to retry on failure
+ * @param {string}   errorMessage - Additional information when the limit is reached
+ * @param {number}   attempts     - The total number of attempts
+ * @param {number}   limit        - The maximum number of attempts
+ * @param {number}   delay        - The seconds to wait between attempts
+ */
+const Retry = function(Run, errorMessage, attempts = 0, limit = 3, delay = 10) {
```
amrocha: We only use Uppercase for class names in JS, I'm not sure why this isn't being caught by the linter 🤔 

See https://github.com/Shopify/javascript

scripts/utilities/retry.js
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Retry - Retry a function on failure
+ *
+ * @param {function} Run          - The function to retry on failure
+ * @param {string}   errorMessage - Additional information when the limit is reached
+ * @param {number}   attempts     - The total number of attempts
+ * @param {number}   limit        - The maximum number of attempts
+ * @param {number}   delay        - The seconds to wait between attempts
+ */
```
amrocha: No need for this docstring imo, the names of the variables should be self evident

scripts/utilities/retry.js
```diff
@@ -0,0 +1,45 @@
+/* eslint-disable no-console */
+
+/**
+ * Retry - Retry a function on failure
+ *
+ * @param {function} Run          - The function to retry on failure
+ * @param {string}   errorMessage - Additional information when the limit is reached
+ * @param {number}   attempts     - The total number of attempts
```
amrocha: I think this would be clearer if you renamed it to `currentAttempt`, what do you think?

scripts/utilities/retry.js
```diff
@@ -0,0 +1,45 @@
...
+ * @param {number}   limit        - The maximum number of attempts
```
amrocha: Can we rename this to something like `maxAttempts`?

scripts/utilities/retry.js
amrocha: We don't use this function syntax either, we do something like
```
function retry([...]) {
}
```

scripts/utilities/retry.js
```diff
@@ -0,0 +1,45 @@
...
+ * @param {number}   delay        - The seconds to wait between attempts
+ */
+const Retry = function(Run, errorMessage, attempts = 0, limit = 3, delay = 10) {
+  // Check that there is a function to try
+  if (typeof Run !== 'function') {
```
amrocha: We don't do defensive programming, see https://github.com/Shopify/ruby-style-guide#general

That's the ruby styleguide but it applies generally to programming at Shopify

scripts/utilities/retry.js
```diff
@@ -0,0 +1,45 @@
...
+    console.error('Retry must have a function to run and retry on failure');
+    return;
+  }
+
+  let currentAttempts = attempts;
```
amrocha: This assignment doesn't really make sense to me, why can't you just use the function parameter the whole way through?

scripts/utilities/retry.js
```diff
@@ -0,0 +1,45 @@
...
+
+  // Try the function
```
amrocha: Again remove these comments, they don't add anything

scripts/utilities/retry.js
```diff
@@ -0,0 +1,45 @@
...
+  try {
+    // Check if we are at maximum attempts
+    if (attempts < limit) {
```
amrocha: This should be at the top of the function. We prefer early returning to checking if we're allowed to perform an operation, like this
```
function retry() {
if (currentAttempt > maxAttempts) {
  return null
}
```

scripts/utilities/retry.js
```diff
@@ -0,0 +1,45 @@
...
+      currentAttempts += 1;
```
amrocha: it doesn't really make sense to increase this here, this makes your recursion step more confusing

package.json
```diff
@@ -50,11 +50,11 @@
...
     "cdn:deploy": "node ./scripts/deploy.js",
     "cdn": "npm-run-all secrets cdn:deploy",
-    "preopen-styleguide-pr": "rimraf sandbox && yarn run secrets",
-    "open-styleguide-pr": "node ./scripts/open-styleguide-pr",
+    "prenew-version-pr-generator": "rimraf sandbox && yarn run secrets",
+    "new-version-pr-generator": "node ./scripts/new-version-pr-generator",
     "copy-polaris-tokens": "rimraf ./src/styles/polaris-tokens && shx cp -r ./node_modules/@shopify/polaris-tokens/dist ./src/styles/polaris-tokens",
     "prepublish": "in-publish && yarn run build || :",
-    "postpublish": "in-publish && npm-run-all cdn open-styleguide-pr || :",
+    "postpublish": "in-publish && npm-run-all cdn new-version-pr-generator || :",
```
amrocha: Right now we have that `|| :` at the end of this line so that this command never fails with a non-zero exit code, I think we should remove that and just handle the exit code ourselves in the script

scripts/utilities/retry.js
```diff
@@ -0,0 +1,45 @@
...
+
+      // Try the function
+      Run();
+    }
+    // Catch any errors
+  } catch (error) {
+    // If we are at maximum
+    if (attempts === limit) {
+      const message = errorMessage ? `${error}\n\n${errorMessage}` : error;
```
amrocha: Showing an error message is not the responsibility of the `retry` function, we should bubble this error up and let the script itself handle it

scripts/utilities/retry.js
```diff
@@ -0,0 +1,45 @@
...
+      console.error(message);
+    } else {
+      // Otherwise try again after a delay
+      setTimeout(() => {
+        Retry(Run, errorMessage, currentAttempts, limit, delay);
```
amrocha: So because you increased `currentAttempts` earlier this line doesn't make sense by itself. It looks as if you're calling the function again with the exact same arguments. The following makes more sense:
```
Retry(Run, errorMessage, currentAttempts + 1, limit, delay)
```

scripts/new-version-pr-generator.js
alex-page: Agreed!

scripts/new-version-pr-generator.js
alex-page: Yeah I will do that for both, I will need to pass the repository into the string.

scripts/new-version-pr-generator.js
alex-page: 👍

scripts/new-version-pr-generator.js
alex-page: Sounds good to me!

scripts/new-version-pr-generator.js
alex-page: This one needs to run in a different folder, sandbox. So it can't go in the other array of commands that run in the repository folder.

scripts/utilities/retry.js
alex-page: Sounds good to me!

scripts/utilities/retry.js
alex-page: Nice one thanks!

scripts/utilities/retry.js
alex-page: Yep!

scripts/utilities/retry.js
alex-page: Sounds good to me, this is just me used to writing tests. Should I be writing tests for this?

scripts/utilities/retry.js
alex-page: I had an odd linting error, i'll fix it up!

scripts/utilities/retry.js
amrocha: Hmm, I want to say that we should write at least some simple tests but I don't know if we even have testing set up for scripts right now. Let me check and get back to you, but don't worry about it for now

alex-page: Thanks mate :)

alex-page:
@amrocha Thanks for the review, I have requested another review. Is there a good way to test this?

scripts/new-version-pr-generator.js
```diff
@@ -0,0 +1,154 @@
...
+const pathResolve = require('path').resolve;
+const {mkdir} = require('shelljs');
+const yaml = require('js-yaml');
+const semver = require('semver');
+
+const {version: PACKAGE_VERSION} = require('../package.json');
+const secrets = require('../secrets.json');
+const Retry = require('./utilities/retry');
+
+const repositories = ['polaris-styleguide', 'web'];
```
kaelig: Have we checked with the web foundations team that this is something they'd welcome?

scripts/new-version-pr-generator.js
```diff
@@ -0,0 +1,154 @@
...
+const polarisBotToken = secrets.github['shopify-polaris'];
+const shopifyPolarisBotGitOverride = `GIT_COMMITTER_NAME='${polarisBotName}' GIT_COMMITTER_EMAIL='${polarisBotEmail}'`;
+
+function pullRequestTemplate(repository, version) {
+  return `
+## Version ${version} of @shopify/polaris just got published!
+
+See what’s new: https://github.com/Shopify/polaris-react/releases/tag/${version}
+
+cc @kaelig @dfmcphee @amrocha @BPScott
```
kaelig: This list of users is relevant for the style guide (although we'll want to remove Dom on Friday), but may not be as relevant for Shopify/web – maybe a few project owners should be pinged instead?

scripts/new-version-pr-generator.js
```diff
@@ -0,0 +1,154 @@
...
+    \`\`\`
+1. Repeat until the build & tests go ✅
+1. 🎩
+1. :shipit:
+</details>
+`.trim();
+}
+
+function failedUpdateMessage(repository, version) {
+  return `It seems your automatic branch creation for ${repository} failed. This can be due to many reasons. To resolve this follow these steps:
```
kaelig: I'd replace "your" with something less personal, as the person reading the logs may not have triggered the build themselves.

scripts/new-version-pr-generator.js
```diff
@@ -0,0 +1,154 @@
...
+
+  1. Checkout "${repository}"
+  2. Get the latest version of master "git checkout master && git pull"
+  3. Create a new branch based on the version that failed "git checkout -b update-polaris-v${version}"
+  4. Update "shopify/polaris" by running "npx yarn upgrade @shopify/polaris@${version}"
```
kaelig:
```suggestion
  4. Update "@shopify/polaris" by running "npx yarn upgrade @shopify/polaris@${version}"
```

scripts/new-version-pr-generator.js
```diff
@@ -0,0 +1,154 @@
...
+  5. Add the changed files to git by running "git add package.json yarn.lock"
+  6. Commit the changes with "git commit -m 'Update @shopify/polaris to ${version}'
```
kaelig:
```suggestion
  6. Commit the changes with "git commit -m 'Update @shopify/polaris to ${version}'"
```

scripts/new-version-pr-generator.js
```diff
@@ -0,0 +1,154 @@
...
+    git push origin update-polaris-${version}
+    \`\`\`
+1. Repeat until the build & tests go ✅
+1. 🎩
+1. :shipit:
+</details>
+`.trim();
+}
+
+function failedUpdateMessage(repository, version) {
```
kaelig: I like the failure message returned by this function a lot!

scripts/new-version-pr-generator.js
```diff
@@ -0,0 +1,154 @@
...
+          {
+            cwd: sandbox,
+            stdio: 'inherit',
+          },
+        ),
+      );
+
+      const commands = [
+        `git checkout -b update-polaris-${releaseVersion}`,
+        'npm config set @shopify:registry https://registry.npmjs.org',
```
kaelig: I don't know if Shopify/web has dependencies that live on our private packagecloud, but if so, this line could make the upgrade fail.

scripts/new-version-pr-generator.js
alex-page: I was thinking of maybe creating a channel, then people can leave and join as necessary, do you have any thoughts?

scripts/new-version-pr-generator.js
alex-page: Great point, I can check with them. I think that our process is creating a PR in the web repository. 

So it would eliminate work for our team but might be frustrating for their team.

scripts/new-version-pr-generator.js
kaelig: I'd look at how "Spy" does this sort of things automatically on some stacks at deploy time. Not sure how to invoke it from a script like this one, but it's where I'd start my research.

scripts/new-version-pr-generator.js
alex-page: I am going to remove it as they have to do a lot of tophatting for each PR so it would probably be more of an annoyance. It would be good to test this is working against another repo... Is there any you can think of that use `polaris/react`?

scripts/new-version-pr-generator.js
alex-page: Can we just use the npm registry? Is there any reason we specify the @shopify one?

kaelig: By default, `@shopify`-scoped are querying Shopify's private packagecloud registry. 

I believe this line is changing this setting because "yarn upgrade" would fail due to certain packages being both on the private and the public registries but not all versions would be available on the private one. 

It may not be an issue anymore, so this is worth testing to make sure.

amrocha: This was already here before, Alex just moved it from the old script, it works

scripts/new-version-pr-generator.js
```diff
@@ -0,0 +1,154 @@
...
+  });
+});
+
+Promise.all(tasks)
+  .then((repository) =>
+    console.log(
+      `Done: Pull request made for ${releaseVersion} in ${repository}`,
+    ),
+  )
+  .catch((error) => {
```
amrocha: I'm not entirely sure how `Promise.all` interacts with `reject()`, make sure you test this out

scripts/new-version-pr-generator.js
```diff
@@ -0,0 +1,154 @@
...
+    console.error(error);
+    process.exit(1);
```
amrocha: We still want to exit with a 0 code here I think. Basically, we don't want to "fail" the deploy if this script runs into issues.

scripts/new-version-pr-generator.js
alex-page: @amrocha agreed we are going to use the Shopify/polaris-reviewers team. Feel free to join.

scripts/new-version-pr-generator.js
alex-page: Yeah lets remove it and give it a go 👍

scripts/new-version-pr-generator.js
alex-page: From previous experience, reject should go into the the catch. I will test it though, thanks @amrocha 👍

scripts/new-version-pr-generator.js
alex-page: Ok no worries, I will update that. I thought we may want to fail it but happy for it to keep going.

amrocha: It's just that if a deploy fails then shipit acts as if the commits weren't deployed properly, which is misleading. This is more like the post-deploy script failed, but we don't really have a way of indicating that right now :/

scripts/new-version-pr-generator.js
kaelig: >This was already here before, Alex just moved it from the old script, it works

Yes it was already there, and it works when running `yarn upgrade` on `Shopify/polaris-styleguide`, but there's a good chance it'll make `yarn upgrade` fail on `Shopify/web` because it also depends on private packages.

kaelig:
To 🎩 this PR, I think these steps _should_ work:

1. Create a new Shipit stack for this branch
1. Temporarily set `baseBranch` to `pr-retry-and-multiple-repos` (see code below)
    https://github.com/Shopify/polaris-react/blob/9f1919a4c57a3081093388f0df20862379705747/scripts/new-version-pr-generator.js#L32
1. Create a new alpha (or beta) version in this branch, push (`yarn version`, then `git push origin pr-retry-and-multiple-repos --follow-tags`)
1. Hit the "Deploy" button in the new Shipit stack to trigger the scripts

If it fails, push a fix and repeat steps 2, 3 and 4 until it works!

scripts/new-version-pr-generator.js
alex-page: Thanks @amrocha for the explanation, happy with your recommended change for now

alex-page:
@kaelig @amrocha @AndrewMusgrave I have fixed up the issues. It successfully made a Pull Request for `web` and `styleguide`.

Have a look here:
https://github.com/Shopify/polaris-styleguide/pull/2638
https://github.com/Shopify/web/pull/11663

scripts/utilities/retry.js
```diff
@@ -0,0 +1,19 @@
+function Retry(
```
kaelig: If this is used in only one script and doesn't have tests, is it worth having this utility in its own file?

alex-page: I really don't mind where it lives, I did this as it might be useful in the future and doesn't hide the function in a file that is unrelated to it.

UNRELEASED.md
```diff
@@ -13,6 +13,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Enhancements
 
 - Updated `Navigation` badge prop to accept a react node ([#1142](https://github.com/Shopify/polaris-react/pull/1142))
+- We now can create pull requests to update `polaris-react` across multiple repos
```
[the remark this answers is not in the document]
alex-page: My bad!

scripts/utilities/retry.js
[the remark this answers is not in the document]
alex-page: I am used to functions being uppercase. I will fix that up.

scripts/utilities/retry.js
```diff
@@ -0,0 +1,19 @@
...
+  functionToTry,
+  maxAttempts = 3,
+  delay = 1000,
+  currentAttempt = 0,
```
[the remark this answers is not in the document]
alex-page: I think you are on to something. Maybe I should remove `currentAttempt` and reduce `maxAttempts` each iteration. This way there is no need for `currentAttempt`.

scripts/new-version-pr-generator.js
```diff
@@ -0,0 +1,144 @@
+/* eslint-disable no-console */
+const {execSync} = require('child_process');
+const {readFileSync} = require('fs');
+const pathResolve = require('path').resolve;
```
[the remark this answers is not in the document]
alex-page: Yep, this was done as I had promises before and `resolve` vs `resolve`.

scripts/new-version-pr-generator.js
```diff
@@ -0,0 +1,144 @@
...
+    console.log(
+      `Pull request made in shopify/${repository} for version ${releaseVersion}`,
+    );
+  } catch (error) {
+    const errorMessage = `${error}\n\n${failedUpdateMessage(
+      repository,
+      releaseVersion,
+    )}`;
+    console.error(errorMessage);
+    process.exit(0);
```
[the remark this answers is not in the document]
alex-page: Happy to discuss this more or leave it. @amrocha and I talked about this before.
Screen Shot 2019-03-14 at 9 27 17 AM

package.json
```diff
@@ -50,11 +50,11 @@
     "precdn": "yarn run build",
     "cdn:deploy": "node ./scripts/deploy.js",
     "cdn": "npm-run-all secrets cdn:deploy",
-    "preopen-styleguide-pr": "rimraf sandbox && yarn run secrets",
-    "open-styleguide-pr": "node ./scripts/open-styleguide-pr",
+    "prenew-version-pr-generator": "rimraf sandbox && yarn run secrets",
+    "new-version-pr-generator": "node ./scripts/new-version-pr-generator",
     "copy-polaris-tokens": "rimraf ./src/styles/polaris-tokens && shx cp -r ./node_modules/@shopify/polaris-tokens/dist ./src/styles/polaris-tokens",
     "prepublish": "in-publish && yarn run build || :",
-    "postpublish": "in-publish && npm-run-all cdn open-styleguide-pr || :",
```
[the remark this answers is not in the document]
alex-page: Related convo...
Screen Shot 2019-03-14 at 9 35 32 AM

scripts/new-version-pr-generator.js
```diff
@@ -0,0 +1,146 @@
...
+      `curl -d '${JSON.stringify({
+        title: `Update @shopify/polaris to ${releaseVersion} 🚀`,
+        body: pullRequestTemplate(repository, releaseVersion),
+        head: `update-polaris-${releaseVersion}`,
+        base: baseBranch,
+      })}' -X POST https://api.github.com/repos/shopify/${repository}/pulls?access_token=${polarisBotToken}`,
+    ];
+
+    // Run the commands in the cloned repository directories
+    commands.forEach((command) => {
```
amrocha: Are you sure this works? I'm pretty sure this `forEach` is queueing up all your commands at the same time. See the following console output:
```
arr = [1,2,3,4,5,6,7,8,9,10]
(10) [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
const func = async (number) => { console.log(number); const ms = Math.random()*500; await new Promise(resolve => setTimeout(resolve, ms)); console.log(ms); console.log(number) }
undefined
arr.forEach(func)
1
2
3
4
5
6
7
8
9
10

50.21044903148086
9
61.93249680065127
6
79.9692515536532
2
80.7784522197541
8
111.60789086028467
5
154.02466156309458
3
191.62890656360963
1
297.53635281697353
10
368.0370214847132
7
388.1897108923006
4
```

scripts/utilities/retry.js
```diff
@@ -0,0 +1,14 @@
+function retry(functionToTry, maxAttempts = 3, delay = 1000) {
+  if (maxAttempts <= 0) {
+    throw new Error();
```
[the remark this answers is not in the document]
alex-page: Sure thing!

scripts/new-version-pr-generator.js
alex-page: This is great thanks @amrocha

alex-page: @amrocha I have approached this slightly different to how we discussed but I think it resolves the asynchronous problems you raised. Let me know if you have any feedback.

alex-page:
This is currently working in parelell:
https://github.com/Shopify/web/pull/11950
https://github.com/Shopify/polaris-styleguide/pull/2659

scripts/new-version-pr-generator.js
```diff
@@ -90,57 +90,65 @@ The automatic branch creation for ${repository} failed. This can be due to many
...
-
-    // Run the commands in the cloned repository directories
-    commands.forEach((command) => {
-      retry(
-        execSync(command, {
-          cwd: repositoryDirectory,
-          stdio: 'inherit',
-        }),
+const jobs = repositories.map((repository) => {
+  return new Promise(async (resolve, reject) => {
```
amrocha: You can simplify this as:
```js
const jobs = repositories.map((repository) => (
  new Promise(async (resolve, reject) => {
```

scripts/new-version-pr-generator.js
```diff
@@ -90,57 +90,65 @@ The automatic branch creation for ${repository} failed. This can be due to many
...
+
+      // Run the commands in the cloned repository directories
+      commands.forEach(async (command) => {
+        await retry(() => {
+          execSync(command, {
+            cwd: repositoryDirectory,
+            stdio: 'inherit',
+          });
+        });
+      });
```
amrocha: I think this has has the same problem we talked about before, these commands are executed in parallel not sequentially. Try replacing the retry with a sleep like we we tried Friday.

For this to be executed sequentially you **can not use an array abstraction**. All array abstractions are going to be executed in parallel.

Either chain promises or use execSync sequentially, either works.
```js
execSync(`git checkout -b update-polaris-${releaseVersion}`, {
  cwd: repositoryDirectory,
  stdio: 'inherit',
});
execSync(`npx yarn@${YARN_VERSION} upgrade @shopify/polaris@${releaseVersion.replace(
  'v',
  '',
)} --no-progress --ignore-engines`, {
  cwd: repositoryDirectory,
  stdio: 'inherit',
});
```

scripts/utilities/retry.js
```diff
@@ -1,16 +1,20 @@
...
-      setTimeout(() => {
-        retry(functionToTry, remainingAttempts, delay);
-      }, delay);
-    }
-  }
-}
+const retry = (functionToTry, maxAttempts = 3, delay = 1000) => {
+  return new Promise((resolve, reject) => {
+    let remainingAttempts = maxAttempts;
+    const intervalId = setInterval(() => {
```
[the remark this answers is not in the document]
alex-page: Yeah great point, thanks @amrocha

scripts/new-version-pr-generator.js
alex-page: I tried doing this but the linter is not liking the end `);`

scripts/new-version-pr-generator.js
alex-page: I have fixed this up, let me know if you find the solution ok!

scripts/new-version-pr-generator.js
amrocha: oh I didn't realize we had a linting rule that blocked this

scripts/utilities/retry.js
```diff
@@ -1,20 +1,22 @@
...
+    try {
+      functionToTry();
+      resolve();
+    } catch (error) {
+      const remainingAttempts = maxAttempts - 1;
+      if (remainingAttempts <= 0) {
+        reject(error);
+      } else {
+        await timeout(delay);
+        retry(functionToTry, remainingAttempts, delay);
```
amrocha: Did you make this that the retry logic works here? There's a weird interaction here and I'm not sure how things interact

If the execution ends up in this `else` block then `resolve()` never gets called. Does that mean that the promise never resolves? I'm not sure

alex-page: Good point i'll fix that up!

alex-page:
Thanks for the feedback @amrocha . I fixed up that silly logic in retry and renamed the function to `sleep`. Nice idea.

I tested the retry function previously and it was infinite loop rejecting, so I have fixed that up. I tested it with this:
```
(async () => {
  await retry(() => {
    console.log('heyyy');
    throw new Error('sigh');
  });
})();
```

kaelig:
Removing myself from the reviewers - Andre's review is as good, if not better, as mine!

scripts/utilities/retry.js
```diff
@@ -0,0 +1,22 @@
...
+      functionToTry();
+      resolve();
+    } catch (error) {
+      const remainingAttempts = maxAttempts - 1;
+      if (remainingAttempts > 0) {
+        await sleep(delay);
+        await retry(functionToTry, remainingAttempts, delay);
+      }
+
+      reject(error);
```
amrocha: This catch block worries me a little bit.

The `reject` gets executed regardless of the output of the retry, and since it returns a new Promise I don't think a resolve output actually exits out of the entire call stack as you would expect.

Can you test that when the first request is errors out and is retried, but the second one succeeds, the original `retry` call still terminates with the `resolve` block instead of the `catch` block?

alex-page: Thanks @amrocha. You are very right! I wrote a quick test:

```js
let i = 0;
const c = () => {
  console.log(i);
  if (i < 1) {
    i += 1;
    throw new Error('c');
  } else {
    console.log('c');
  }
};

retry(c);
```

The output of this is:
```shell
0
1
c
(node:33745) UnhandledPromiseRejectionWarning: Error: c
```

I updated the code to fix up the output to resolve if it passed the retry function.
```shell
0
1
c
```

amrocha:
I'm testing a few things out, this still feels like it won't work as expected

scripts/utilities/retry.js
```diff
@@ -0,0 +1,23 @@
...
+  return new Promise(async (resolve, reject) => {
+    try {
+      functionToTry();
+      resolve();
+    } catch (error) {
+      const remainingAttempts = maxAttempts - 1;
+      if (remainingAttempts > 0) {
+        await sleep(delay);
+        await retry(functionToTry, remainingAttempts, delay);
+        resolve();
```
amrocha: Add a return statement after the resolve to cut off execution of the rest of the function

alex-page:
@amrocha I agree that the recursive promises is really hard to understand. I have reworked them a bit if you want we could do something like:
```
function retry(functionToTry, maxAttempts = 3, delay = 1000) {
  return new Promise(async (resolve, reject) => {
    let remainingAttempts = maxAttempts;
    while (remainingAttempts > 0) {
      remainingAttempts -= 1;
      try {
        functionToTry();
        break;
      } catch (error) {
        if (remainingAttempts === 0) {
          reject(error);
        }
        await sleep(delay);
      }
    }
    resolve();
  });
}
```

The tests I am running are:
```
let i = 0;
const c = () => {
  console.log(i);
  if (i < 1) {
    i += 1;
    throw new Error('c');
  } else {
    console.log('trying c');
  }
};

retry(c);

const b = () => {
  console.log('trying b');
  throw new Error('b');
};

retry(b);

const a = () => {
  console.log('trying a');
};

retry(a);
```

amrocha:
Yeah that works too. I don't really like using for loops and while loops in JS, but we don't have any rules against it so if you like this better then it's good!

PR #1074, opened 2019-02-20
Remove custom svg transformer for tests

BPScott:

Simplifying our test config by using the default svg transformer one provided by sewing-kit. We're about to change how icons get loaded so there's no point having a transformer for our tests that mimics the old way.

The only reason why we needed it was for some Spinner tests, which can
be expressed in a different way

- Removes the custom transformer
- Update Spinner tests


src/components/Spinner/tests/Spinner.test.tsx
```diff
@@ -15,24 +15,24 @@ describe('<Spinner />', () => {
   describe('size', () => {
     it('renders a large spinner by default', () => {
       const spinner = shallowWithAppProvider(<Spinner />);
-      expect(spinner.find('svg').prop('viewBox')).toBe('0 0 44 44');
+      expect(spinner.find('svg').hasClass('sizeLarge')).toBeTruthy();
```
solonaarmstrong: If `hasClass` is a valid thing to test, why don't we use it elsewhere, I wonder?

config/jest/transformers/svg.js
```diff
@@ -1,18 +0,0 @@
...
-      ? '0 0 44 44'
-      : '0 0 20 20';
-
-    const content = ICON_REGEX.test(filename)
-      ? {body: '<path />', viewBox: svgViewbox}
-      : basename(filename);
-
-    return `module.exports = ${JSON.stringify(content)};`;
-  },
-};
```
solonaarmstrong: I no longer remember the original reason for adding this config. Spinner renders fine in both sizes, though.

PR #1075, opened 2019-02-21
Loading a11y docs

dpersing:

Adds accessibility guidance for the loading component, to appear in `polaris-react` docs and in the style guide.

[See the draft Google doc for editing history for these changes and updates for other components and pages.](https://docs.google.com/document/d/1ONoa4fUsqG19i5h0h2Kz2w9CvmFSN926YmoNVMsGPgw/edit#heading=h.w5x5ph7nfna7)

* [X] Adds accessibility documentation for the loading component
* [X] Removes recommendations for labeling, since this component has no labeling props and accessibility-related information is provided in code
* [x] Adds an entry to `UNRELEASED.md`

## How to 🎩

1. Check out `master` from `polaris-styleguide`.
1. In another Terminal tab or window, check out this branch from `polaris-react` and [run the instructions for testing in a consuming project](https://github.com/Shopify/polaris-react/#testing-in-a-consuming-project).
1. In the `polaris-styleguide` tab, run `dev up && dev start`.
1. View changes after examples and props: https://polaris.myshopify.io/components/feedback-indicators/loading

## Screenshot

Web content:
Loading component accessibility docs

### DIFF
--- a/src/components/Loading/README.md
+++ b/src/components/Loading/README.md
@@ -67,17 +67,6 @@ The loading component should:
 
 ---
 
-## Content guidelines
-
-### Accessibility label
-
-The loading accessibility label should:
-
-- Accurately explain the state of the requested action. For example, “Loading your search results”.
-- Use as few words as possible to describe the state.
-
----
-
 ## Related components
 
 - To indicate that an action has been received, use the [Spinner](/components/feedback-indicators/spinner)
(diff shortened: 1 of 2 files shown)


src/components/Loading/README.md
```diff
@@ -67,19 +67,36 @@ The loading component should:
 
 ---
 
-## Content guidelines
+## Related components
```
sadiesaurus: I took a look through the style guide and it seems we haven't got a consistent way of documenting "Related components". 

Sometimes component names are:
- capitalized
- preceded with "the"
- followed by "component"

My preference is to use lowercase, "the", and "component": "the spinner component". BUT that might be out of scope here, especially since the other components in this section look like the content that's been added here. I think as long as the pattern is consistent, then for now that's 👌

dpersing: @sadiesaurus Would it make sense to fold that into the content audit that (I think?) is happening? I think we could do a bunch of those formatting updates as part of separate PRs.

sadiesaurus: @dpersing It _is_ happening 😄 and that's a great idea. I will add this in.

PR #1099, opened 2019-02-26
[CRA Example] Update deps

AndrewMusgrave:

Security alerts in our [CRA example](https://github.com/shopify/polaris-react/network/alerts)

Updating deps

### How to 🎩

* Run example
* Run tests

🎉

### DIFF
--- a/examples/create-react-app/package.json
+++ b/examples/create-react-app/package.json
@@ -3,11 +3,11 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@shopify/polaris": "^2.2.0",
+    "@shopify/polaris": "^3.9.0",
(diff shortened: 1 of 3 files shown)


examples/create-react-app/package.json
```diff
@@ -16,7 +16,13 @@
     "eject": "react-scripts eject"
   },
   "devDependencies": {
-    "enzyme": "^3.3.0",
+    "enzyme": "^3.9.0",
     "enzyme-adapter-react-16": "^1.1.1"
-  }
+  },
+  "browserslist": [
```
AndrewMusgrave: `browserslist ` is react from `react-scripts` (this is the default)

PR #1105, opened 2019-02-27
Provide an internal reference to Polaris version that is compile-safe

### LINKED ISSUE #1104: Provide an internal reference to Polaris version that is compile-safe
# Issue summary

https://github.com/Shopify/polaris-react/pull/1033 introduced a reference to `window.Polaris.VERSION`, which can be undefined when Polaris is build in production mode using sewing-kit.

## Fix

Export Polaris version from `src/config.ts` and use that internally instead.

iainmcampbell:

Resolves #1104

Providing an internal method to reference Polaris version.

### Tophat

Run an embedded app using Polaris, compiled using sewing kit [in production mode](https://github.com/Shopify/sewing_kit#how-can-i-test-a-production-verison-of-my-changes). Should be no errors when dispatching App Bridge actions (ie opening a modal, resource picker, etc).

Additionally, `APP::` prefixed actions in the Redux inspector should have an actual version number for `clientInterface.version`, not `'{{POLARIS_VERSION}}'`.

I'm working on getting a repo case up and running.

### 🎩 checklist

### DIFF
--- a/src/configure.ts
+++ b/src/configure.ts
@@ -13,4 +13,4 @@ if (typeof window !== 'undefined') {
   window.Polaris.VERSION = '{{POLARIS_VERSION}}';
 }
 
-export const POLARIS_VERSION = '{{POLARIS_VERSION}}';
+export const polarisVersion = '{{POLARIS_VERSION}}';
--- a/src/configure.ts
+++ b/src/configure.ts
@@ -12,3 +12,5 @@ if (typeof window !== 'undefined') {
   window.Polaris = window.Polaris || {};
   window.Polaris.VERSION = '{{POLARIS_VERSION}}';
 }
+
+export const POLARIS_VERSION = '{{POLARIS_VERSION}}';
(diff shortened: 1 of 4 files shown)


src/configure.ts
```diff
@@ -12,3 +12,5 @@ if (typeof window !== 'undefined') {
   window.Polaris = window.Polaris || {};
   window.Polaris.VERSION = '{{POLARIS_VERSION}}';
 }
```
amrocha: This is technically a change to our public interface, so maybe there's a case for a deprecation warning? @tmlayton what do you think?

tmlayton: I don’t think this is a deprecation, rather a way to access the version number from within the library. It’s fine if folks import or reference in a consuming application from window, but maybe I’m missing some part of the thinking @amrocha?

tmlayton:
You should be able to get around the test failures by changing the constant name or reworking the test

andrewlo:
Tested that this definitely fixed the original error in POS Channel

src/configure.ts
iainmcampbell: Yeah, this isn't a deprecation, it adds a way to access the version number internally.

I don’t think the window global should be deprecated; it should be safe for use by consuming applications. The only case where it _could_ be unavailable is when Polaris is compiled using aggressive tree shaking (ie, sewing-kit). I think that’s an issue for another PR though.

iainmcampbell:
ok all, I’ve fixed the test, we’re good for review!

PR #1112, opened 2019-02-28
[Checkbox] Fix focus and problem with events not firing

AndrewMusgrave:

`Checkbox` can lose focus or not receive events.

#### Shift clicks
https://cl.ly/44715b1c9430

#### Losing focus
https://cl.ly/4ca13de5131c

* Removing pointer events on adjacent elements to precent blocking and bubbling
* Switch from functional to class components. We can utilize on shallow prop comparison with `PureComponent` and removes the need for scoped functions
* Focus input after every change
* Use event delegation with onClick / onKeyUp handlers
* Use a no-op for onChange (input element)

### Question

Is this a good approach? Are there good alternatives?

### How to 🎩

Load up the playground and test checkboxes on multiple browsers and devices 😄 

* Fast clicks
* Slow clicks
* Modified clicks (e.g shift)
* Clicks with focus and without focus
* Keyboard interactions

### 🎩 checklist

cc/ @dpersing How does this look to you in terms of accessibility? Previously the focus state was a little off, now the checkbox should remained focused after every interaction with it.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -17,6 +17,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Bug fixes
 
 - Fixed disabled states while loading for `ResourceList` ([#1237](https://github.com/Shopify/polaris-react/pull/1237))
+- Fixed `Checkbox` from losing focus and not receiving some modified events([#1112](https://github.com/Shopify/polaris-react/pull/1112))
 
 ### Documentation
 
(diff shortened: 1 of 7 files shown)


dpersing:
@AndrewMusgrave This looks good to me from a behavior perspective.

One thing I noticed that isn't specific to this PR but has an impact on testing form fields is that we don't wrap our form field examples in a `<form>` container. This can lead to some weirdness when you try to access native form controls with some screen readers. I'll open a separate issue with that for consideration!

PR #1126, opened 2019-03-05
[ComboBox] Change tabindex prop to a negative number

adamcarchibald:
Resolves https://github.com/Shopify/polaris-react/issues/1089

Fixes tab navigation behaviour in the Autocomplete component by not selecting the ComboBox when tabbing through the component.

Changes the `tabIndex` prop value from `0` to `-1` for the `ComboBox` component.  ~~Adds a regression test.~~

### DIFF
--- a/src/components/Autocomplete/components/ComboBox/ComboBox.tsx
+++ b/src/components/Autocomplete/components/ComboBox/ComboBox.tsx
@@ -236,6 +236,7 @@ export default class ComboBox extends React.PureComponent<Props, State> {
       emptyState && <div className={styles.EmptyState}>{emptyState}</div>;
 
     return (
+      /*  eslint-disable-next-line jsx-a11y/interactive-supports-focus */
       <div
         onClick={this.handleClick}
         role="combobox"
@@ -245,7 +246,6 @@ export default class ComboBox extends React.PureComponent<Props, State> {
         aria-haspopup
         onFocus={this.handleFocus}
         onBlur={this.handleBlur}
-        tabIndex={0}
       >
         <KeypressListener
           keyCode={Key.DownArrow}
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -17,6 +17,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Bug fixes
 
 - Fixed `Popover` fade-in flutter on iOS by switching Transition component for CSSTransition [#1400](https://github.com/Shopify/polaris-react/pull/1400)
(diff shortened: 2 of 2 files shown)


adamcarchibald:
I have removed the regression test as I am unable to determine a pattern for testing the tab navigation behaviour.  I was following the pattern set out in similar tests for ComboBox but we discovered they are false positives as per https://github.com/Shopify/polaris-react/issues/1128

dpersing:
I'm wondering if a `tabindex` attribute is needed at all? The container won't take focus naturally, so I think it's safe to remove it altogether (unless there's something hooked into that specifically, in which case I'm wondering why that might be!).

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
```diff
@@ -263,7 +263,7 @@ export default class ComboBox extends React.PureComponent<Props, State> {
         aria-haspopup
         onFocus={this.handleFocus}
         onBlur={this.handleBlur}
-        tabIndex={0}
+        tabIndex={-1}
```
AndrewMusgrave: As far as I can tell we don't need the tab index on the `div` at all

```suggestion
```

adamcarchibald: Originally I removed the tabIndex entirely, however it failed CircleCI tests that check if an item has an onFocus method, it must have a tabIndex.

dpersing: Hmm. The container shouldn't need to get focus because the text input does naturally. I wonder why `onFocus`/`onBlur` is being handled on the `<div>` and not the field?

ry5n:
I chatted with @adamcarchibald this morning. Rather than have this stall, the simplest thing might be to add a comment to ignore the lint rule to get this unblocked. 

(In that case, The Polaris team should circle back later to fix the combobox implementation.)

danrosenthal:
What's the next step here, @adamcarchibald? Should this be closed?

adamcarchibald:
It was suggested that I simply remove the linting rule that is blocking this pr.  Will make changes now and post update.

danrosenthal:
I spent some time this morning tophatting this change, and it has negative impact on the focus and keyboard behavior when compared to its current state in master. Both are unusable when interacted with using the arrow keys to move through available items.

We've documented the extensive problems with this component here: https://github.com/Shopify/polaris-react/issues/689. Rather than spending time trying to patch something with deep issues, we should focus on rebuilding.

I think it makes sense to close this pull request and for the Polaris team to prioritize fixing this component.

PR #1127, opened 2019-03-06
DropZone - File upload positioning issue 

dleroux:

In 3.9 [this PR](https://github.com/Shopify/polaris-react/pull/908) added the ability to specify a height to the drop zone.

When a height container does have a height, we added `display: flex` to ensure that the `DropZone.FileUpload` was vertically and horizontally aligned. This however causes issue because we really have no control what the children of that container will be (In web this breaks the ProductShow implementation)

From the DropZone component we also don't know what the children are. It could be 'DropZone.FileUpload` or not. So the best solution I could come up with, is the add a Flex layout to the FileUpload if it's the only child. I really don't like the fact the FileUpload is responsible for it's positioning, but it's the safest approach I could think of.

**This PR is not ready to deploy. It's here for discussion. Also note I branched right from 3.9 in case we end up doing a patch release for this **
 
## 

### How to 🎩

<details>

```jsx
import {Page, DropZone, Stack, Thumbnail, Caption} from '../src';

interface State {
  files: object;
}

  state = {
    files: [],
  };

  render() {
    const {files} = this.state;
    const validImageTypes = ['image/gif', 'image/jpeg', 'image/png'];

      <Stack vertical>
        {files.map((file, index) => (
            <Thumbnail
              size="small"
              alt={file.name}
              source={
                validImageTypes.indexOf(file.type) > 0
                  ? window.URL.createObjectURL(file)
                  : 'https://cdn.shopify.com/s/files/1/0757/9955/files/New_Post.png?12678548500147524304'
              }
            />
            <div>
            </div>
          </Stack>
        ))}
      </Stack>
    );

    return (
      <Page title="Playground">
        <div style={{width: 500, height: 500}}>
          <DropZone
              this.setState({
              });
            }}
          >
            {uploadedFiles}
            {fileUpload}
          </DropZone>
        </div>
      </Page>
    );
  }
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/DropZone/DropZone.scss
+++ b/src/components/DropZone/DropZone.scss
@@ -66,9 +66,12 @@ $dropzone-overlay-background-error: color('red', 'lighter');
 
 .Container {
   flex: 1;
-  display: flex;
-  justify-content: center;
-  align-items: center;
+
+  &.centerAlignChildren {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+  }
 }
 
 .Overlay {
--- a/src/components/DropZone/DropZone.tsx
+++ b/src/components/DropZone/DropZone.tsx
@@ -83,6 +83,11 @@ export interface Props {
   disabled?: boolean;
   /** The child elements to render in the dropzone. */
   children?: string | React.ReactNode;
+  /**
+   * Align child elements to the center of the dropzone.
+   * @default true
+   */
+  centerAlignChildren?: boolean;
   /** Allows a file to be dropped anywhere on the page */
   dropOnPage?: boolean;
   /** Sets the default file dialog state */
@@ -213,6 +218,7 @@ export class DropZone extends React.Component<CombinedProps, State> {
       active,
       overlay,
       allowMultiple,
+      centerAlignChildren = true,
     } = this.props;
 
     const inputAttributes: object = {
@@ -286,6 +292,11 @@ export class DropZone extends React.Component<CombinedProps, State> {
         </div>
       ) : null;
 
+    const containerClassnames = classNames(
+      styles.Container,
+      centerAlignChildren && styles.centerAlignChildren,
+    );
+
     const dropZoneMarkup = (
       <div ref={this.setNode} className={styles.DropZoneWrapper}>
         <div
@@ -296,7 +307,7 @@ export class DropZone extends React.Component<CombinedProps, State> {
         >
           {dragOverlay}
           {dragErrorOverlay}
-          <div className={styles.Container}>{children}</div>
+          <div className={containerClassnames}>{children}</div>
           <VisuallyHidden>
             <input {...inputAttributes} />
           </VisuallyHidden>


src/components/DropZone/components/FileUpload/FileUpload.scss
```diff
@@ -3,6 +3,16 @@ $fileupload-padding: rem(15px);
 .FileUpload {
   padding: $fileupload-padding;
   text-align: center;
+
+  &:only-child {
+    // top: 50%;
+    // position: relative;
+    // transform: translateY(-50%);
```
dleroux: these comments are a different layout solution.

dleroux:
I see Percy spotted an issue with the "Nested Drop Zone" example.

dleroux:
Another option discussed on slack would be to add a prop to <DropZone.FileUpload centered={false} /> set to `true` by default.

AndrewMusgrave:
At a glance feels odd to me to have non-structure components have props that affect positioning. Curious what everyone else thinks about this?

dleroux:
An option would be to leave it as is, and have the consumer override the display: flex with their own container if they need to.

danrosenthal:
I've pushed a fix that adds a `centerAlignChildren` prop that defaults to `true`. This breaks our component model in ways, as its not typically desirable to have parents lay out their children. It's also not great to have props that default to true. In this case, since this will technically avoid a breaking change, I think it's okay.

I don't have much more time to focus on this, so if @elizabethletourneau with the support of @AndrewMusgrave could finish this off, that would be appreciated.

|screenshots|
|-|
|screen shot 2019-03-06 at 9 56 46 am|
|screen shot 2019-03-06 at 9 56 53 am|

```jsx
<React.Fragment>
  <div style={{width: 50}}>
    <DropZone>
      <DropZone.FileUpload />
    </DropZone>
  </div>
  <div style={{width: 114}}>
    <DropZone>
      <DropZone.FileUpload />
    </DropZone>
  </div>
  <div style={{width: 50}}>
    <DropZone centerAlignChildren={false}>
      <DropZone.FileUpload />
    </DropZone>
  </div>
  <div style={{width: 114}}>
    <DropZone centerAlignChildren={false}>
      <DropZone.FileUpload />
    </DropZone>
  </div>
  <div style={{width: 500}}>
    <DropZone centerAlignChildren={false}>
      <DropZone.FileUpload />
    </DropZone>
  </div>
  <div style={{width: 500, height: 500}}>
    <DropZone centerAlignChildren={false}>
      <DropZone.FileUpload />
    </DropZone>
  </div>
</React.Fragment>
```

danrosenthal:
Closing in favor of #1129

nairim:
So I took a look at the related issues, and @elizabethletourneau helped me to understand a bit more of the problem (thank you, Liz! <3)

Taking in consideration the fact that `Boolean props should often be false by default` (unless there’s a legitimate exception) and, paraphrasing @danrosenthal on issue #1127, “its not typically desirable to have parents lay out their children” I believe is safe to accept that it should be up to the consumer to style drop zone’s children if they want to customize the height of the component.

I thought that maybe there’s an opportunity here to add some guidelines to the style guide in how to do centralize it horizontally once you customize the height of the drop zone.

Unrelated question: I was playing with the props and tried the `Disabled` one. Specifically, when we have the option of having a “Add file” button, should the button appear on the disabled state if the disabled prop is set to `true`?

PR #1137, opened 2019-03-06
[ComboBox] Fix tests

AndrewMusgrave:

Resolve https://github.com/Shopify/polaris-react/issues/1124

Using real events instead of mocking a listener map

### DIFF
--- a/src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
+++ b/src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
@@ -6,37 +6,7 @@ import {mountWithAppProvider} from 'test-utilities';
 import {TextField} from '../components';
 import {Key} from '../../../../../types';
 
-interface HandlerMap {
-  [eventName: string]: (event: any) => void;
-}
-
-const listenerMap: HandlerMap = {};
-
 describe('<ComboBox/>', () => {
-  let addEventListener: jest.SpyInstance;
-  let removeEventListener: jest.SpyInstance;
-
-  beforeEach(() => {
-    addEventListener = jest.spyOn(document, 'addEventListener');
-    addEventListener.mockImplementation((event, callback) => {
-      listenerMap[event] = callback;
-    });
-
-    removeEventListener = jest.spyOn(document, 'removeEventListener');
-    removeEventListener.mockImplementation((event) => {
-      listenerMap[event] = noop;
-    });
-  });
-
-  afterEach(() => {
-    if (document.activeElement) {
-      (document.activeElement as HTMLElement).blur();
-    }
-
-    addEventListener.mockRestore();
-    removeEventListener.mockRestore();
-  });
-
   const options = [
     {value: 'cheese_pizza', label: 'Cheese Pizza'},
     {value: 'macaroni_pizza', label: 'Macaroni Pizza'},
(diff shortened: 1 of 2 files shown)


src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
```diff
@@ -445,20 +413,13 @@ describe('<ComboBox/>', () => {
...
-        await expect(spy).toHaveBeenCalledTimes(1);
-        expect(comboBox.prop('selected')[0]).toBe('cheese_pizza');
-      };
+      comboBox.find(TextField).simulate('click');
+      dispatchKeyup(Key.DownArrow);
+      dispatchKeyup(Key.Enter);
+      expect(spy).toHaveBeenCalledWith(['cheese_pizza']);
     });
 
-    it('activates the popover when the tab key is pressed', () => {
```
AndrewMusgrave: From what I could tell we don't listen for a `tab` keypress, we're trying to test if focus activates the popover

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
```diff
@@ -484,16 +443,11 @@ describe('<ComboBox/>', () => {
         />,
       );
 
-      async () => {
-        comboBox.find(TextField).simulate('click');
-        await listenerMap.keyup({keyCode: Key.DownArrow});
```
AndrewMusgrave: `DownArrow` press isn't required in this test

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
```diff
@@ -588,3 +542,8 @@ function renderTextField() {
 function renderNodeWithId() {
   return <div id="CustomNode" />;
 }
+
+function dispatchKeyup(key: Key) {
+  const event: KeyboardEventInit & {keyCode: Key} = {keyCode: key};
```
AndrewMusgrave: This is needed to properly type `KeyboardEvent` params

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
```diff
@@ -445,20 +413,13 @@ describe('<ComboBox/>', () => {
...
-        expect(comboBox.prop('selected')[0]).toBe('cheese_pizza');
-      };
+      comboBox.find(TextField).simulate('click');
+      dispatchKeyup(Key.DownArrow);
+      dispatchKeyup(Key.Enter);
+      expect(spy).toHaveBeenCalledWith(['cheese_pizza']);
     });
 
-    it('activates the popover when the tab key is pressed', () => {
+    it('activates the popover when the combobx is focused', () => {
```
danrosenthal:
```suggestion
    it('activates the popover when the combobox is focused', () => {
```

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
chloerice: I'm not well versed (or opinionated) on test architecture, but I don't know if we need to go through actually dispatching a real event here. Triggering the handler is another option:

```suggestion
function dispatchKeyup(wrapper: ReactWrapper, key: Key) {
  const listener = wrapper.findWhere((listener) => {
    return listener.prop('keyCode') === key;
  });

  trigger(listener, 'handler', {keyCode: key});
}
```

and then in the tests you'd just pass in the `comboBox` wrapper as the first arg, or `comboBox.find(Popover)` in the case of the `Key.Escape` test since the listener for that is in the popover component. But imo that test should be 🔥 because it's testing functionality of a parent that already tests that.

AndrewMusgrave: I maybe totally off but the issue I see with trigger is that it'll dispatch `onSelect` for example however when a key is pressed the `ComboBox` has it's own internal logic that handle the keyup and then call `onSelect` with the result of ComboBox's internal logic

danrosenthal: `trigger` will just look for a corresponding handler passed through props, and trigger that handler, so it doesn't feel useful for us in this case. Dispatching the event feels like the way to go from my perspective

PR #1142, opened 2019-03-07
[Navigation] badge accepts react element

CameronGorrie:
Minimizes impact of re-rendering navigation if the badge relies on a query to resolve. More detail here: https://github.com/Shopify/web/issues/10481

Allows the nav items badge property to take a react node prop value.

### How to 🎩

<details>

```tsx
import {Page, Navigation, Badge} from '../src';

export default class Playground extends React.Component<{}, never> {
  render() {
    return (
      <Page title="Playground">
        <Navigation location="/">
          <Navigation.Section
            items={[
              {
                url: '/path/to/place',
                label: 'Orders',
                icon: 'orders',
                badge: <Badge>Hello world</Badge>,
              },
            ]}
          />
        </Navigation>
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -12,6 +12,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Updated `Navigation` badge prop to accept a react node ([#1142](https://github.com/Shopify/polaris-react/pull/1142))
+
 ### Bug fixes
 
 - Fixed unnecessary height on `TextField` due to unhandled carriage returns ([#901](https://github.com/Shopify/polaris-react/pull/901))
--- a/src/components/Navigation/components/Item/Item.tsx
+++ b/src/components/Navigation/components/Item/Item.tsx
@@ -40,7 +40,7 @@ interface SecondaryAction {
 export interface Props extends ItemURLDetails {
   icon?: IconProps['source'];
   iconBody?: string;
-  badge?: string | null;
+  badge?: React.ReactNode;
   label: string;
   disabled?: boolean;
   accessibilityLabel?: string;
@@ -90,16 +90,16 @@ export class BaseItem extends React.Component<CombinedProps, State> {
       url,
       icon,
       label,
-      badge,
       subNavigationItems = [],
       secondaryAction,
       disabled,
       onClick,
       accessibilityLabel,
       iconBody,
+      selected: selectedOverride,
+      badge,
       new: isNew,
       polaris: {intl},
-      selected: selectedOverride,
     } = this.props;
 
     const {location, onNavigationDismiss} = this.context;
@@ -116,15 +116,6 @@ export class BaseItem extends React.Component<CombinedProps, State> {
       </span>
     ) : null;
 
-    const badgeMarkup =
-      badge || isNew ? (
-        <div className={styles.Badge}>
-          <Badge status="new" size="small">
-            {badge || intl.translate('Polaris.Badge.STATUS_LABELS.new')}
-          </Badge>
-        </div>
-      ) : null;
-
     const iconMarkup = iconBody ? (
       <div className={styles.Icon}>
         <Icon source={iconBody} />
@@ -137,6 +128,28 @@ export class BaseItem extends React.Component<CombinedProps, State> {
       )
     );
 
+    let badgeMarkup: React.ReactNode = null;
+    if (isNew) {
+      badgeMarkup = (
+        <Badge status="new" size="small">
+          {intl.translate('Polaris.Badge.STATUS_LABELS.new')}
+        </Badge>
+      );
+    } else if (typeof badge === 'string') {
+      badgeMarkup = (
+        <Badge status="new" size="small">
+          {badge}
+        </Badge>
+      );
+    } else {
+      badgeMarkup = badge;
+    }
+
+    const wrappedBadgeMarkup =
+      badgeMarkup == null ? null : (
+        <div className={styles.Badge}>{badgeMarkup}</div>
+      );
+
     const itemContentMarkup = (
       <React.Fragment>
         {iconMarkup}
@@ -144,7 +157,7 @@ export class BaseItem extends React.Component<CombinedProps, State> {
           {label}
           {indicatorMarkup}
         </span>
-        {badgeMarkup}
+        {wrappedBadgeMarkup}
       </React.Fragment>
     );
 
--- a/src/components/Navigation/components/Item/tests/Item.test.tsx
+++ b/src/components/Navigation/components/Item/tests/Item.test.tsx
@@ -100,6 +100,34 @@ describe('<Nav.Item />', () => {
       const link = item.find(UnstyledLink);
       expect(link.exists()).toBe(true);
     });
+
+    it('renders a small badge with new status if the prop is provided with a string', () => {
+      const item = mountWithAppProvider(<Item label="some label" badge="1" />);
+
+      expect(item.find(Badge).props()).toMatchObject({
+        status: 'new',
+        size: 'small',
+        children: '1',
+      });
+    });
+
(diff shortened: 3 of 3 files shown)


src/components/Navigation/components/Item/Item.tsx
```diff
@@ -40,7 +40,7 @@ interface SecondaryAction {
 export interface Props extends ItemURLDetails {
   icon?: IconProps['source'];
   iconBody?: string;
-  badge?: string | null;
+  badge?: string | React.ReactElement<BadgeProps> | null;
```
lemonmade: `BadgeProps` doesn't help here, really. The component you have in the Web PR doesn't conform to that prop shape, unless all Badge's props are optional, but that's OK. You don't depend on the props for anything, so I would just make this prop a `React.ReactNode` instead and accept anything

src/components/Navigation/components/Item/Item.tsx
```diff
@@ -263,6 +251,37 @@ export class BaseItem extends React.Component<CombinedProps, State> {
     );
   }
 
+  private getBadgeMarkup() {
```
lemonmade: I would really rather avoid these kinds of methods. I know people like them because we are taught to separate out logic, but in React, separation should either be with a component, or not at all. In this case, I would opt for none at all.

src/components/Navigation/components/Item/Item.tsx
```diff
@@ -263,6 +251,37 @@ export class BaseItem extends React.Component<CombinedProps, State> {
...
+
+    if (badge == null && !isNew) {
+      return null;
+    }
+
+    let badgeMarkup: Props['badge'] = null;
+    if (isNew) {
+      badgeMarkup = intl.translate('Polaris.Badge.STATUS_LABELS.new');
+    } else if (typeof badge === 'string') {
+      badgeMarkup = badge;
```
lemonmade: Seems weird to call this `badgeMarkup` when it is actually only the text contents?

src/components/Navigation/components/Item/Item.tsx
CameronGorrie: Right, I overlooked that those interfaces don't match. Will update.

UNRELEASED.md
```diff
@@ -12,6 +12,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Updated `Navigation` to accept children ([#1142](https://github.com/Shopify/polaris-react/pull/1142))
```
lemonmade: I don't think this quite describes what you are doing

src/components/Navigation/components/Item/Item.tsx
```diff
@@ -40,7 +40,7 @@ interface SecondaryAction {
 export interface Props extends ItemURLDetails {
   icon?: IconProps['source'];
   iconBody?: string;
-  badge?: string | null;
+  badge?: string | React.ReactNode | null;
```
lemonmade: `string | null` are included in `React.ReactNode` so I think you are OK to just use that type

src/components/Navigation/components/Item/Item.tsx
```diff
@@ -137,14 +128,38 @@ export class BaseItem extends React.Component<CombinedProps, State> {
       )
     );
 
+    let badgeMarkup: Props['badge'] = null;
```
lemonmade: I would just use `React.ReactNode` for the type, it feels weird to use `Props['badge']` since you make it consume the badge prop and transform it

src/components/Navigation/components/Item/Item.tsx
```diff
@@ -137,14 +128,38 @@ export class BaseItem extends React.Component<CombinedProps, State> {
...
+    if (isNew) {
+      badgeMarkup = (
+        <Badge status="new" size="small">
+          {intl.translate('Polaris.Badge.STATUS_LABELS.new')}
+        </Badge>
+      );
+    }
+
+    if (badge != null && !isNew) {
```
lemonmade: Can you do this in an `else` to the previous block to remove the `isNew` check?

src/components/Navigation/components/Item/Item.tsx
```diff
@@ -137,14 +128,38 @@ export class BaseItem extends React.Component<CombinedProps, State> {
...
     const itemContentMarkup = (
       <React.Fragment>
         {iconMarkup}
         <span className={styles.Text}>
           {label}
           {indicatorMarkup}
         </span>
-        {badgeMarkup}
+
+        {badgeMarkup == null ? null : (
```
lemonmade: I think I'd store this chunk in an intermediate variable, or manually wrap each case of the badge in the div when creating `badgeMarkup`

src/components/Navigation/components/Item/tests/Item.test.tsx
```diff
@@ -100,6 +100,65 @@ describe('<Nav.Item />', () => {
       const link = item.find(UnstyledLink);
       expect(link.exists()).toBe(true);
     });
+
+    it('renders a badge if prop is provided', () => {
+      const spy = jest.fn();
+      matchMedia.setMedia(() => ({addListener: spy}));
```
lemonmade: Is this needed for the test?

src/components/Navigation/components/Item/tests/Item.test.tsx
```diff
@@ -100,6 +100,65 @@ describe('<Nav.Item />', () => {
...
+      const item = mountWithAppProvider(
+        <Item
+          label="some label"
+          url="/admin/orders"
+          badge="1"
+          subNavigationItems={[
```
lemonmade: Is this or the field below needed for the test?

src/components/Navigation/components/Item/tests/Item.test.tsx
```diff
@@ -100,6 +100,65 @@ describe('<Nav.Item />', () => {
       const link = item.find(UnstyledLink);
       expect(link.exists()).toBe(true);
     });
+
+    it('renders a badge if prop is provided', () => {
```
lemonmade: I think this case is specifically testing the string case?

src/components/Navigation/components/Item/Item.tsx
CameronGorrie: Yeah the only downside is if `badge` is undefined we'll be setting `badgeMarkup = undefined` but the later check handles this, and it's easier to read than nested if statements.

src/components/Navigation/components/Item/tests/Item.test.tsx
CameronGorrie: Nope. This is just for the subnav items. Bad copy-pasta.

src/components/Navigation/components/Item/tests/Item.test.tsx
```diff
@@ -100,6 +100,28 @@ describe('<Nav.Item />', () => {
       const link = item.find(UnstyledLink);
       expect(link.exists()).toBe(true);
     });
+
+    it('renders a badge if the prop is provided with a string', () => {
+      const item = mountWithAppProvider(<Item label="some label" badge="1" />);
```
lemonmade: Should we ensure the badge has these contents + the right props?

src/components/Navigation/components/Item/tests/Item.test.tsx
```diff
@@ -100,6 +100,28 @@ describe('<Nav.Item />', () => {
...
+
+      expect(item.find(Badge).text()).toContain('Custom badge');
+    });
+
+    it('renders a new badge even if a badge prop is provided', () => {
+      const item = mountWithAppProvider(
+        <Item label="some label" badge={<Badge>Custom badge</Badge>} new />,
+      );
+
+      expect(item.find(Badge).text()).toContain('New');
```
lemonmade: Should we test to make sure the `Custom badge` text is not present/ there is only one badge?

PR #1146, opened 2019-03-08
Burning memoize

michelleyschen:

Related issue: https://github.com/Shopify/polaris-react/issues/991

The `memoize` method from `javascript-utilities` depend on `lodash`, which is a heavy library that can slow our performance. This PR is part of the process to get rid of `lodash` by removing `memoize` because it does not offer any substantial performance enhancement. 

There is no UI Changes

## 

### How to 🎩

### 🎩 checklist

### DIFF
--- a/src/components/ResourceList/components/FilterControl/tests/FilterControl.test.tsx
+++ b/src/components/ResourceList/components/FilterControl/tests/FilterControl.test.tsx
@@ -117,8 +117,11 @@ describe('<FilterControl />', () => {
         </Provider>,
       );
 
-      const tags = filterControl.find(Tag);
-      trigger(tags.at(0), 'onRemove', mockAppliedFilters[0].key);
+      const tag = filterControl.find(Tag).at(0);
+      const filterId = `${mockAppliedFilters[0].key}-${
+        mockAppliedFilters[0].value
+      }`;
+      trigger(tag, 'onRemove', filterId);
 
       expect(onFiltersChange).toBeCalledWith([
         ...mockAppliedFilters.slice(1, mockAppliedFilters.length),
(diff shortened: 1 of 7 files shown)


AndrewMusgrave:
Looks like we're getting some legit test failures

michelleyschen:
yes 😢 
it seems to fail if the entire suite run but pass when it's the only test.
I am a bit confuse right now - still digging into it thou.

michelleyschen:
These test failure are totally confusing to me.
Here are what I discover so far.

- in all 3 places, if I run the falling test with `only` it passed, if I run the test with the whole file it fail.
- if I mock `memoize` the test passes, so it is my decorator that’s conflicting it.
- so far I found one bug in one of the test, but the test is still failing (in `FilterControl`, the tag remove had the wrong filterId passed in, it was only passing in key where it the component has it as a combo of key + value)
- I also ran Playground with `FilterControl` and I was able to remove tag and see the `onFiltersChange` being trigger

src/components/ResourceList/components/FilterControl/tests/FilterControl.test.tsx
```diff
@@ -117,8 +117,11 @@ describe('<FilterControl />', () => {
         </Provider>,
       );
 
-      const tags = filterControl.find(Tag);
-      trigger(tags.at(0), 'onRemove', mockAppliedFilters[0].key);
+      const tag = filterControl.find(Tag).at(0);
+      const filterId = `${mockAppliedFilters[0].key}-${
+        mockAppliedFilters[0].value
+      }`;
+      trigger(tag, 'onRemove', filterId);
```
michelleyschen: I discover that the filterId should be a combine string of key and value, not just key.

michelleyschen:
Closing this in favor of https://github.com/Shopify/polaris-react/pull/1148

PR #1148, opened 2019-03-11
♻️ replace autobind with instance method

michelleyschen:

Related issue: #991

The `autobind` method from javascript-utilities depend on lodash, which is a heavy library that can slow our performance. This PR is part of the process to get rid of lodash by 
- replacing most of the `autobind` with instance method.
- removing `memoize` because it does not offer any substantial performance enhancement. 

Another PR will follow to handle the cases where I can't directly replace `autobind` with instance method. 

## 

### How to 🎩

<details>

```jsx

interface State {}

  render() {
    return (
      <Page title="Playground">
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Autocomplete/components/ComboBox/ComboBox.tsx
+++ b/src/components/Autocomplete/components/ComboBox/ComboBox.tsx
@@ -4,7 +4,6 @@ import {
   addEventListener,
   removeEventListener,
 } from '@shopify/javascript-utilities/events';
-import {autobind} from '@shopify/javascript-utilities/decorators';
 import OptionList, {OptionDescriptor} from '../../../OptionList';
 import ActionList from '../../../ActionList';
 import Popover from '../../../Popover';
@@ -132,7 +131,7 @@ export default class ComboBox extends React.PureComponent<Props, State> {
   getChildContext(): Context {
     return {
       comboBoxId: this.state.comboBoxId,
-      selectedOptionId: this.getSelectedOptionId(),
+      selectedOptionId: this.selectedOptionId,
       subscribe: this.subscribe,
       unsubscribe: this.unsubscribe,
     };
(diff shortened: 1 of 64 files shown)


src/components/ResourceList/ResourceList.tsx
```diff
@@ -111,7 +111,6 @@ export class ResourceList extends React.Component<CombinedProps, State> {
     );
   }
 
-  @autobind
```
michelleyschen: getter always have the proper scope, we don't need autobind

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
```diff
@@ -132,7 +131,7 @@ export default class ComboBox extends React.PureComponent<Props, State> {
   getChildContext(): Context {
     return {
       comboBoxId: this.state.comboBoxId,
-      selectedOptionId: this.getSelectedOptionId(),
+      selectedOptionId: this.getSelectedOptionId,
```
AndrewMusgrave: check

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
```diff
@@ -510,29 +491,26 @@ export default class ComboBox extends React.PureComponent<Props, State> {
...
     }
     if (newOption) {
       newOption.active = true;
     }
-  }
+  };
 
-  @autobind
-  private getSelectedOptionId(): string | undefined {
+  private get getSelectedOptionId(): string | undefined {
```
AndrewMusgrave: check

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
AndrewMusgrave:
```suggestion
      selectedOptionId: this.selectedOptionId,
```

michelleyschen:
i didn't replace all of them mostly because the ones that's left cause un-expected test failure and require more looking into. This PR is already large enough as is so my thought is to get these low risk item out first and high risk item when I have more time.

What do you think?

michelleyschen:
I removed a few more `autobind`, however, this mean I had to replace `debounce` with `lodash/debounce`. The idea will be that we can eventually replace `lodash/dedounce` with our own version as well. but it should be add to the list in #991

michelleyschen:
That's all of `@shopify/javascript-utilities/decorators`!!

PR #1183, opened 2019-03-13
Update v4 with master - 19/03/12

AndrewMusgrave:

Keep version 4 update to date with master

Merging master into `version-4.0.0`

### Tips

Rather than going through 10k+ lines of changes it might be easier to compare my branch against master and compare `version-4.0.0` against master 😄


src/components/ContextualSaveBar/tests/ContextualSaveBar.test.tsx
```diff
@@ -11,7 +11,7 @@ describe('<ContextualSaveBar />', () => {
     message: 'Unsaved changes',
   };
 
-  it('calls the contextual save bar on mount with correct values', () => {
```
AndrewMusgrave: Had to fix the test text, our linter doesn't allow `correct`... not sure why it wasn't caught before? 🤔

src/components/ContextualSaveBar/tests/ContextualSaveBar.test.tsx
```diff
@@ -45,7 +45,7 @@ describe('<ContextualSaveBar />', () => {
     expect(mockFrameContext.frame.removeContextualSaveBar).toHaveBeenCalled();
   });
 
-  it('calls the contextual save bar with correct values if its props change after it mounted', () => {
```
AndrewMusgrave: ^

AndrewMusgrave:
@BPScott Any idea why the `CLA` label was added?

danrosenthal:
If the CLA check is non-blocking, I would just skip it and not worry about making it work. Looks like it's a result of Solona not being part of the org anymore.

danrosenthal:
> Is the only change in v4 Dan's? If so we could have just branched and cherry picked too I guess.

There's a handful of commits there. I think this is the right approach

AndrewMusgrave:
> Is the only change in v4 Dan's? If so we could have just branched and cherry picked too I guess.

Even if we `cherry-picked` merge commits, we would still have to `cherry-pick` multiple times which results in more work and more conflicts. Using `merge` or `rebase (cherry-pick)` isn't a right or wrong question, they both have their own advantages and disadvantages that this [stack overflow post](https://stackoverflow.com/questions/1241720/git-cherry-pick-vs-merge-workflow) can explain a lot better than I could 😄 .

With that being said since `version-4.0.0` is an extension of master we will want to only merge into it rather than pushing our changes in.

PR #1187, opened 2019-03-13
Update deps

AndrewMusgrave:

Keep up to date 😄


kaelig:
Can you try running `npx yarn-dedupe yarn.lock` and see if the build still passes?

The idea is to prevent some large dependencies (like puppeteer) to be downloaded in multiple versions (and perhaps even help with the polaris build size itself!).

AndrewMusgrave:
@kaelig It causes the build it fail. I'm a little confused though, I thought `yarn install` also does a `de-dupe`, what's the difference?

chloerice:
@AndrewMusgrave how do you recommend we tophat these changes? What are the potential break points?

AndrewMusgrave:
@chloerice There isn't much we have to worry about. All the major updates happened to all be dev dependencies. Which includes, archiver, copyfiles, fs-extra, svgo and yargs. Most of the packages changes don't affect us. `copy` changed in `fs-extra`, and `svgo.optimize` now returns a promise. I would say the most important part of 🎩 this change is making sure the build is still 👌

kaelig:
@AndrewMusgrave yarn-install's dedupe is very conservative, which can be good for the developers but not always so great for the users or for build times.

Alternatively I sometimes edit the yarn.lock file manually, and group certain dependencies together. I checked in polaris-styleguide and `node_modules/puppeteer` is 195 MB, so it could be worth a quick manual edit here as well.

AndrewMusgrave:
Taking another stab at updating deps from scratch #1327

PR #1196, opened 2019-03-15
Pass icon components into `<Icon>` instead of the bundled icon name

BPScott:

We are looking at deprecating the concept of bundled icons in the
future so lets get our house in order first.

Replaces usage of bundled icon strings with passsing in a React Component into `<Icon>`

### How to 🎩

Click around storybook, ensuring you can still see the same icons

### DIFF
--- a/src/components/DataTable/components/Cell/Cell.tsx
+++ b/src/components/DataTable/components/Cell/Cell.tsx
@@ -1,9 +1,10 @@
 import * as React from 'react';
+import {CaretUpMinor, CaretDownMinor} from '@shopify/polaris-icons';
 import {classNames} from '@shopify/react-utilities/styles';
 
 import {headerCell} from '../../../shared';
 import {withAppProvider, WithAppProviderProps} from '../../../AppProvider';
-import Icon, {IconSource} from '../../../Icon';
+import Icon from '../../../Icon';
 import {SortDirection} from '../../types';
 
 import styles from '../../DataTable.scss';
@@ -71,7 +72,7 @@ function Cell({
   };
 
   const direction = sorted ? sortDirection : defaultSortDirection;
-  const source = `caret${direction === 'ascending' ? 'Up' : 'Down'}`;
+  const source = direction === 'ascending' ? CaretUpMinor : CaretDownMinor;
   const oppositeDirection =
     sortDirection === 'ascending' ? 'descending' : 'ascending';
 
@@ -82,10 +83,7 @@ function Cell({
 
   const iconMarkup = (
     <span className={iconClassName}>
-      <Icon
-        source={source as IconSource}
-        accessibilityLabel={sortAccessibilityLabel}
-      />
+      <Icon source={source} accessibilityLabel={sortAccessibilityLabel} />
     </span>
   );
 
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -38,5 +38,6 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - Replaced all occurrences of `_.pick` with a custom pick function ([#1020](https://github.com/Shopify/polaris-react/pull/1020))
 - Deleted the icons index file that would re-export icons, and replaced it with direct imports ([#1195](https://github.com/Shopify/polaris-react/pull/1195))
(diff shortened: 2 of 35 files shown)


BPScott:
Some better ways to review this:


### Already checked by percy diffing storybook

- src/components/DataTable/components/Cell/Cell.tsx -> Sortable DataTable (click up/down to see the toggle)
- src/components/FooterHelp/FooterHelp.tsx -> Footer help box
- src/components/InlineError/InlineError.tsx -> Basic inline error
- src/components/Modal/components/CloseButton/CloseButton.tsx -> Basic modal
- src/components/Navigation/components/Section/Section.tsx -> Navigation with section rollup
- src/components/Pagination/Pagination.tsx -> Default pagination
- src/components/Select/Select.tsx -> Default select
- src/components/Tag/Tag.tsx -> Default tag
- src/components/TopBar/TopBar.tsx ->  Frame in stand-alone component

### Visible in a story but need to be checked manually because they require using interaction to see the icon

- src/components/Frame/Frame.tsx -> Frame in stant-alone component (shrink to mobile size then expand menu)
- src/components/Frame/components/Toast/Toast.tsx -> Basic toast (click show toast)
- src/components/OptionList/components/Checkbox/Checkbox.tsx -> Option list with sections (click a checkbox)
- src/components/ResourceList/components/BulkActions/components/BulkActionButton/BulkActionButton.tsx -> Resource List with Bulk Actions (click selected and see "More actions" has caret)
- src/components/ResourceList/components/FilterControl/FilterControl.tsx ->
- src/components/Tabs/Tabs.tsx -> Default tabs (shrink to mobile size and see ... appears to the right)
- src/components/TextField/components/Spinner/Spinner.tsx -> Number field
- src/components/TopBar/components/SearchField/SearchField.tsx -> Tob bar with all of its elements (type into search box and see X appears)


### No relevant story, needs a playground demo:

-src/components/Breadcrumbs/Breadcrumbs.tsx
-src/components/Button/Button.tsx
-src/components/Checkbox/Checkbox.tsx
-src/components/Page/components/Header/components/Action/Action.tsx

Add the following to the playground:

```ts
import {Page, Checkbox, Button, Breadcrumbs} from '../src';

  return (
    <Page title="Playground">
      <Breadcrumbs
        breadcrumbs={[{content: '1', url: '?'}, {content: '2', url: '?'}]}
      />
      <hr />
      <Button disclosure>Click!</Button>
      <hr />
      <Checkbox checked="indeterminate" label="Foo" />
      <hr />
      <Page
        title="Page Title"
        secondaryActions={[{content: 'Duplicate', icon: 'add'}]}
      />
    </Page>
  );
}
```

Full disclosure: that leaves src/components/ResourceList/components/FilterControl/components/DateSelector/DateSelector.tsx but I trust reading-that-one-line-change to be happy that works.

amrocha:
What's the effect of this on bundle size? My understanding is that since we were already importing these icons it should be net neutral, can you confirm?

BPScott:
Yep net-neutral as they were already included thanks to bundled icons.

amrocha:
I made `Icon.tsx` throw an error if it tried to render a string and got a bunch of errors, are you sure you caught all instances of strings being passed in?

src/components/DataTable/components/Cell/Cell.tsx
```diff
@@ -71,7 +72,7 @@ function Cell({
   };
 
   const direction = sorted ? sortDirection : defaultSortDirection;
-  const source = `caret${direction === 'ascending' ? 'Up' : 'Down'}`;
```
amrocha: String interpolation like this should be illegal

BPScott:
> I made Icon.tsx throw an error if it tried to render a string and got a bunch of errors, are you sure you caught all instances of strings being passed in?

We chatted locally, this was due to me not fixing up cases where `<Button icon="string">` exists. I originally added a fix to cover these cases in this the PR but it's a bit more complex that I thought. To avoid it getting swallowed up in the noise of this PR I'll follow up in a new one.

src/components/DataTable/components/Cell/Cell.tsx
kaelig: 💯

Issue #1204, opened 2019-03-15
[Card] / Extra top padding in Card section


loic-d:
# Issue summary

In the `Card` component, there is extra padding at the top in the section.

<img width="588" alt="Screen Shot 2019-03-15 at 2 52 36 PM" src="https://user-images.githubusercontent.com/3925905/54455405-19e6be00-4732-11e9-8ef9-cfe03c47ee94.png">
<img width="596" alt="Screen Shot 2019-03-15 at 2 53 22 PM" src="https://user-images.githubusercontent.com/3925905/54455412-1c491800-4732-11e9-9b95-390d0e50bc3c.png">

## Expected behavior

There should not be extra spacing at the top of the card section.

## Actual behavior

There is extra spacing at the top of the card section.

Could be related to https://github.com/Shopify/polaris-react/pull/1082

## Steps to reproduce the problem

1. Add content to a `Card` component

## Reduced test case

The best way to get your bug fixed is to provide a reduced test case. This [CodeSandbox template](https://codesandbox.io/s/q82mlq0m26) is a great starting point.

## Specifications

- Are you using the React components? (Y/N): Y
- Polaris version number: 3.10
- Browser: Chrome
- Device: Mac OS X
- Operating System:

tmlayton:
It does look like a bit too much space, but this is the expected behavior. Here is an example using v3.5.0 https://codesandbox.io/s/m77q5yzvyy.

I’m going to close for now, but feel free to reopen if you think this should be addressed.

loic-d:
@tmlayton I see your point. I still feel that it the UI is not matching our Discounts UX requirements. Maybe the solution is to increase the spacing after the `Banner` to make it less awkward?

Try https://codesandbox.io/s/y7p9o983pz. I am using a `ChoiceList` instead of a paragraph and it looks a bit off.

loic-d:
cc/ @papoirier for visibility. You originally flagged the issue.

PR #1082, opened 2019-02-21
[Card] Fix subdued section spacing issues

solonaarmstrong:
### WHY are these changes introduced?

Resolves https://github.com/Shopify/polaris-react/issues/954

The first fix was here: https://github.com/Shopify/polaris-react/pull/962 but it created too much spacing with non-subdued sections and the header. This fix targets only the subdued sections.

### WHAT is this pull request doing?

Add spacing and a border between header and subdued section & footer and subdued section.

### Before

screen shot 2019-02-21 at 10 26 59 am
screen shot 2019-02-21 at 10 27 12 am
screen shot 2019-02-21 at 10 27 20 am

### After

screen shot 2019-02-21 at 10 27 38 am
screen shot 2019-02-21 at 10 27 45 am
screen shot 2019-02-21 at 10 27 49 am

## 

### How to 🎩

🖥 [Local development instructions](https://github.com/Shopify/polaris-react/blob/master/README.md#development)
🗒 [General tophatting guidelines](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting.md)
📄 [Changelog guidelines](https://github.com/Shopify/polaris-react/blob/master/documentation/Versioning%20and%20changelog.md)

<details>
<summary>Copy-paste this code in <code>playground/Playground.tsx</code>:</summary>

```jsx
import * as React from 'react';
import {Page, Card, List, DisplayText, TextContainer} from '../src';

interface State {}

export default class Playground extends React.Component<{}, State> {
  render() {
    return (
      <Page title="">
        <TextContainer>
          <DisplayText size="medium">Card.Section</DisplayText>
          <Card>
            <Card.Section title="Deactivated reports" subdued>
              <List>
                <List.Item>Payouts</List.Item>
                <List.Item>Total Sales By Channel</List.Item>
              </List>
            </Card.Section>
          </Card>
        </TextContainer>
        <br />
        <hr />
        <br />
        <TextContainer>
          <DisplayText size="medium">
            Footer actions + subdued Card.Section
          </DisplayText>
          <Card
            secondaryFooterAction={{content: 'Dismiss'}}
            primaryFooterAction={{content: 'Export Report'}}
          >
            <Card.Section title="Deactivated reports" subdued>
              <List>
                <List.Item>Payouts</List.Item>
                <List.Item>Total Sales By Channel</List.Item>
              </List>
            </Card.Section>
          </Card>
        </TextContainer>
        <br />
        <TextContainer>
[...]

PR #1205, opened 2019-03-15
Revert "Merge pull request #1082 from Shopify/card-subdued-spacing"

tmlayton:
#1082 is causing regressions, for example #1204


ry5n:
What led to identifying #1082 is the cause? There was a previous change that causes similar regressions, but as far as I can tell #1082 it should only affect situations with subdued sections.

tmlayton:
It’s an unchecked assumption at this point, will create a reduced test case to verify.

tmlayton:
Looks like this is the expected behavior https://codesandbox.io/s/m77q5yzvyy. I didn’t have a chance to test so was going off what other’s saw.

PR #1219, opened 2019-03-18
[Banner] Add focus method

AndrewMusgrave:

There are use cases in web where banner's need to be focused

Adding a function so Banner can focus itself

### How to 🎩

<details>

```jsx
import {react} from 'babel-types';
import {Page, Banner, Button} from '../src';

interface State {}

  node = React.createRef();

  render() {
    return (
      <Page title="Playground">
        <Banner title="Order archived" onDismiss={() => {}} ref={this.node}>
          <p>This order was archived on March 7, 2017 at 3:12pm EDT.</p>
        </Banner>
        <Button onClick={() => this.node.current && this.node.current.focus()}>
          Mr.Focus
        </Button>
      </Page>
    );
  }
}

```

</details>

### DIFF
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -43,6 +43,11 @@ export interface Props {
 
 export default class Banner extends React.PureComponent<Props, never> {
   static contextTypes = contentContextTypes;
+  private wrapper = React.createRef<HTMLDivElement>();
+
+  public focus() {
+    this.wrapper.current && this.wrapper.current.focus();
+  }
 
   render() {
     const {
@@ -154,6 +159,7 @@ export default class Banner extends React.PureComponent<Props, never> {
         className={className}
         // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
         tabIndex={0}
+        ref={this.wrapper}
         role={ariaRoleType}
         aria-live="polite"
         onMouseUp={handleMouseUp}
--- a/src/components/Banner/README.md
+++ b/src/components/Banner/README.md
@@ -500,6 +500,38 @@ class ModalExample extends React.Component {
 }
 ```
 
+### Banner with focus
+
+<!-- example-for: web -->
+
+Banner with focus receives focus programmatically when it appears. Use this component to communicate problems that merchants need to resolve before they can complete a task.
+
+```jsx
+class ModalExample extends React.Component {
+  banner = React.createRef();
+
+  componentDidMount() {
+    this.banner.current.focus();
+  }
+
+  render() {
+    return (
+      <Banner
+        title="High risk of fraud detected"
+        onDismiss={() => {}}
+        status="critical"
+        ref={this.banner}
+      >
+        <p>
+          Before fulfilling this order or capturing payment, please review the
+          fraud analysis and determine if this order is fraudulent
+        </p>
+      </Banner>
+    );
+  }
+}
+```
+
 ### Banner in a card
 
 <!-- example-for: web -->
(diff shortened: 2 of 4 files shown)


src/components/Banner/Banner.tsx
```diff
@@ -44,6 +44,11 @@ export interface Props {
 
 export default class Banner extends React.PureComponent<Props, never> {
   static contextTypes = contentContextTypes;
+  private wrapper = React.createRef<HTMLDivElement>();
+
+  public focus() {
+    this.wrapper.current && this.wrapper.current.focus();
```
AndrewMusgrave: Normally this would be an early return but that leaves untestable code so this is an alternative.

src/components/Banner/README.md
```diff
@@ -500,6 +500,38 @@ class ModalExample extends React.Component {
 }
 ```
 
+### Banner with focus
+
+<!-- example-for: web -->
+
+Banner's can be focused through the ref API
```
dleroux: Use when a banner needs to be programmatically focused to help guide the user. For example, an error banner in a form after it's been submitted.

??

AndrewMusgrave: @selenehinkley @sadiesaurus Do you both have any opinions?

sadiesaurus: `Banner with focus receives focus programmatically when it appears. Use this component to communicate problems that merchants need to resolve before they can complete a task.` 

- In the component a11y documentation, we usually say "receive focus" instead of "be focused", so I'm favour of keeping things consistent. 
- The second line is adapted from the description for critical banners. I'm not sure if the same use case applies here, though. Does this work? 
- Should we mention when to use critical banners instead?

src/components/Banner/tests/Banner.test.tsx
```diff
@@ -113,4 +114,26 @@ describe('<Banner />', () => {
     const button = bannerWithContentContext.find(Button);
     expect(button.prop('size')).toBe('slim');
   });
+
+  describe('focus', () => {
+    it('exposes a function to allow the the banner to be focused', () => {
```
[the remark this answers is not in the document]
sadiesaurus: I like this suggestion! And would also add `to receive focus programatically` instead of `to be programmatically focused` (for consistency).

src/components/Banner/README.md
AndrewMusgrave: Thanks @sadiesaurus I'll make the change 😄 I think for now we don't have to mention it's use with critical banners for now. But maybe it's something we could develop a guideline / pattern for?

AndrewMusgrave: @sadiesaurus Since this won't only be used for critical banners I adjusted it slightly. What do you think, have any suggestions?

```
Banner can programmatically receive focus. Use this functionality to draw the merchants attention to the banner.
```

sadiesaurus: Ah, cool, good point! A couple of questions for ya: 
- When you use this banner, do you get to choose whether it receives focus when it appears, or will it always receive focus when it appears? 
- If the former is true, then I think we can adjust the content a bit to make this clear
- If the latter is true, then do we need to explain when it's appropriate to redirect the merchant's attention?

AndrewMusgrave: We'll get to choose when Banner receives focus it won't always be  when it appears! @sadiesaurus

dleroux: The key is "Banner can programmatically receive focus."

So unlike a prop which is set on the component and would automatically set focus the banner, this allows us to focus the banner at any given time. For instance when a user tries to submit the form.

sadiesaurus: Cool, thank you for the context 🙏  

I think your adjustment works @AndrewMusgrave , with one wee punctuation edit: 
`the merchants attention` > `the merchant’s attention`

PR #1221, opened 2019-03-18
[Tabs] Remove componentWillReceiveProps

AndrewMusgrave:

Removes a deprecated lifecycle method

* Makes use of `getDerivedStateFromProps`
* Doesn't close the disclosure everytime props are changed
* Doesn't removed focus outline when a tab is selected from the disclosure

### How to 🎩

Checkout the tabs examples  at different screen sizes

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -39,6 +39,23 @@ export interface State {
 export default class Tabs extends React.PureComponent<Props, State> {
   static Panel = Panel;
 
+  static getDerivedStateFromProps(nextProps: Props, prevState: State) {
+    const {disclosureWidth, tabWidths, containerWidth} = prevState;
+    const {visibleTabs, hiddenTabs} = getVisibleAndHiddenTabIndices(
+      nextProps.tabs,
+      nextProps.selected,
+      disclosureWidth,
+      tabWidths,
+      containerWidth,
+    );
+
+    return {
+      visibleTabs,
+      hiddenTabs,
+      selected: nextProps.selected,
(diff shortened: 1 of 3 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -60,14 +49,23 @@ export default class Tabs extends React.PureComponent<Props, State> {
       containerWidth,
     );
 
-    this.setState({
+    return {
       visibleTabs,
       hiddenTabs,
-      tabToFocus: selected === nextProps.selected ? -1 : tabToFocus,
```
AndrewMusgrave: Discussed this with Devon and we think it'll be best if we maintain the focus outline when selecting a tab from the popover

src/components/Tabs/Tabs.tsx
```diff
@@ -60,14 +49,23 @@ export default class Tabs extends React.PureComponent<Props, State> {
...
-      showDisclosure: false,
```
AndrewMusgrave: And we'll want to keep the popover open

src/components/Tabs/tests/Tabs.test.tsx
```diff
@@ -229,17 +229,6 @@ describe('<Tabs />', () => {
       expect(tabs.find(TabMeasurer).prop('tabToFocus')).toBe(-1);
     });
 
-    it('resets focus when selected gets updated to the previously selected value', () => {
```
AndrewMusgrave: test is no longer needed

AndrewMusgrave:
cc/ @dpersing

PR #1226, opened 2019-03-19
[TextField] Add clear button to textfield

### LINKED ISSUE #1115: [TextField] Add prop for clear button
# Feature request summary

Add prop to include a clear button like this:

image
image

## Rationale

This is already being done in various places, but requires a custom component because of the limitation of `TextField`

---

andrewmcgov:

Resolves #1115 

Adds a `clearButton` prop to the `textField` component to show a button that clears the value from the input.

The clear button only shows on `type="search"` inputs, unless `clearButton={false}`.

## 

### How to 🎩

<details>

```jsx

interface State {}

  state = {
    value: 'test value',
  };

  handleChange = (value) => {
    this.setState({value});
  };

  onClearButtonClick = () => {
    this.setState({value: ''});
  };

  render() {
    return (
      <Page title="Playground">
        <TextField
          label="Store name"
          type="text"
          value={this.state.value}
          onChange={this.handleChange}
          onClearButtonClick={this.onClearButtonClick}
          clearButton
        />
      </Page>
    );
  }
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Icon/Icon.scss
+++ b/src/components/Icon/Icon.scss
@@ -3,16 +3,6 @@ $stacking-order: (
   icon: 2,
 );
 
-@mixin color-icon($value, $hue: base) {
-  svg {
-    fill: color($value, $hue);
-  }
-
-  img {
-    filter: filter($value, $hue);
-  }
-}
-
 .Icon {
   display: block;
   height: rem(20px);
--- a/src/components/TextField/TextField.scss
+++ b/src/components/TextField/TextField.scss
@@ -129,6 +129,15 @@ $stacking-order: (
   }
 }
 
+.Input-hasClearButton {
+  // Hide webkit search input clear button
+  &[type='search'] {
+    &::-webkit-search-cancel-button {
+      appearance: none;
+    }
+  }
+}
+
 .Input-suffixed {
   padding-right: 0;
 }
@@ -194,7 +203,7 @@ $stacking-order: (
 .CharacterCount {
   @include text-emphasis-subdued;
   z-index: z-index(contents, $stacking-order);
-  margin-right: $backdrop-horizontal-spacing;
+  margin: 0 $backdrop-horizontal-spacing 0 $addon-horizontal-spacing;
   pointer-events: none;
   text-align: right;
 }
@@ -203,6 +212,21 @@ $stacking-order: (
   align-self: flex-end;
   width: 100%;
   padding-bottom: spacing(tight);
+}
+
+.ClearButton {
+  @include unstyled-button;
+  z-index: z-index(contents, $stacking-order);
+  margin: 0 $backdrop-horizontal-spacing 0 $addon-horizontal-spacing;
+
+  &:hover:enabled,
+  &:focus:enabled {
+    @include color-icon('ink', 'lighter');
+  }
+
+  &:disabled {
+    cursor: default;
+  }
 }
 
 .Spinner {
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -8,6 +8,7 @@ import Connected from '../Connected';
 
 import {Error, Key} from '../../types';
 import {withAppProvider, WithAppProviderProps} from '../AppProvider';
+import Icon from '../Icon';
 import {Resizer, Spinner} from './components';
 import styles from './TextField.scss';
 
@@ -53,6 +54,8 @@ export interface BaseProps {
   labelHidden?: boolean;
   /** Disable the input */
   disabled?: boolean;
+  /** Show a clear text button in the input */
+  clearButton?: boolean;
   /** Disable editing of the input */
   readOnly?: boolean;
   /** Automatically focus the input */
@@ -103,6 +106,8 @@ export interface BaseProps {
   showCharacterCount?: boolean;
   /** Determines the alignment of the text in the input */
   align?: Alignment;
+  /** Callback when clear button is clicked */
+  onClearButtonClick?(id: string): void;
   /** Callback when value is changed */
   onChange?(value: string, id: string): void;
   /** Callback when input is focused */
@@ -169,6 +174,7 @@ class TextField extends React.PureComponent<CombinedProps, State> {
       autoComplete,
       autoFocus,
       connectedLeft,
+      clearButton,
       connectedRight,
       disabled,
       error,
@@ -256,6 +262,18 @@ class TextField extends React.PureComponent<CombinedProps, State> {
       </div>
     ) : null;
 
+    const clearButtonMarkup =
+      clearButton && normalizedValue !== '' ? (
+        <button
+          testID="clearButton"
+          className={styles.ClearButton}
+          onClick={this.handleClearButtonPress}
+          disabled={disabled}
+        >
+          <Icon source="circleCancel" color="inkLightest" />
+        </button>
+      ) : null;
+
     const spinnerMarkup =
       type === 'number' && !disabled && !readOnly ? (
         <Spinner
@@ -299,6 +317,7 @@ class TextField extends React.PureComponent<CombinedProps, State> {
       styles.Input,
       align && styles[variationName('Input-align', align)],
       suffix && styles['Input-suffixed'],
+      clearButton && styles['Input-hasClearButton'],
     );
 
     const input = React.createElement(multiline ? 'textarea' : 'input', {
@@ -359,6 +378,7 @@ class TextField extends React.PureComponent<CombinedProps, State> {
             {input}
             {suffixMarkup}
             {characterCountMarkup}
+            {clearButtonMarkup}
             {spinnerMarkup}
             <div className={styles.Backdrop} />
             {resizer}
@@ -398,6 +418,15 @@ class TextField extends React.PureComponent<CombinedProps, State> {
 
     const newValue = Math.min(max, Math.max(numericValue + steps * step, min));
     onChange(String(newValue.toFixed(decimalPlaces)), this.state.id);
+  };
+
+  private handleClearButtonPress = () => {
+    const {
+      state: {id},
+      props: {onClearButtonClick},
+    } = this;
+
+    onClearButtonClick && onClearButtonClick(id);
   };
 
   private handleExpandingResize = (height: number) => {
--- a/src/components/TextField/tests/TextField.test.tsx
+++ b/src/components/TextField/tests/TextField.test.tsx
@@ -1,6 +1,10 @@
 import * as React from 'react';
 import {noop} from '@shopify/javascript-utilities/other';
-import {shallowWithAppProvider, mountWithAppProvider} from 'test-utilities';
+import {
+  shallowWithAppProvider,
+  mountWithAppProvider,
+  findByTestID,
+} from 'test-utilities';
 import {InlineError, Labelled, Connected, Select} from 'components';
 import {Resizer, Spinner} from '../components';
 import TextField from '../TextField';
@@ -839,6 +843,65 @@ describe('<TextField />', () => {
       expect(textField.find(Connected).prop('right')).toStrictEqual(
         connectedRight,
       );
+    });
+  });
+
+  describe('clearButton', () => {
+    it('renders a clear button when true', () => {
+      const textField = mountWithAppProvider(
+        <TextField
+          id="MyTextField"
+          label="TextField"
+          onChange={noop}
+          type="text"
+          value="test value"
+          clearButton
+        />,
+      );
+      expect(findByTestID(textField, 'clearButton').exists()).toBeTruthy();
+    });
+
+    it('does not render in inputs without a value', () => {
+      const textField = mountWithAppProvider(
+        <TextField
+          id="MyTextField"
+          label="TextField"
+          type="text"
+          onChange={noop}
+          clearButton
+        />,
+      );
+      expect(findByTestID(textField, 'clearButton').exists()).toBeFalsy();
+    });
+
+    it('calls onClearButtonClicked() with an id when the clear button is clicked', () => {
+      const spy = jest.fn();
+      const textField = mountWithAppProvider(
+        <TextField
+          id="MyTextField"
+          label="TextField"
+          type="search"
+          onChange={noop}
+          onClearButtonClick={spy}
+          value="test value"
+          clearButton
+        />,
+      );
+      findByTestID(textField, 'clearButton').simulate('click');
+      expect(spy).toHaveBeenCalledWith('MyTextField');
+    });
+
+    it('does not render a clear button by default', () => {
+      const textField = mountWithAppProvider(
+        <TextField
+          id="MyTextField"
+          label="TextField"
+          onChange={noop}
+          type="text"
+          value="test value"
+        />,
+      );
+      expect(findByTestID(textField, 'clearButton').exists()).toBeFalsy();
     });
   });
 });
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -14,9 +14,9 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
-– Enhanced `NavigationItem`s colour accessibility for `active`, `focus`, `hover` and `Selected` states([1304](https://github.com/Shopify/polaris-react/pull/1304))
-
+- Enhanced `NavigationItem`'s color accessibility for `active`, `focus`, `hover` and `Selected` states ([1304](https://github.com/Shopify/polaris-react/pull/1304))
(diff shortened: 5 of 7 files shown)


andrewmcgov:
@kvendrik maybe you can take a look when you get a chance?

A couple things I am wondering about:

- Should the clear button be hidden if the input is already empty?
- Should I separate the logic at the top of the `clearButtonMarkup` somehow?

src/components/TextField/TextField.scss
```diff
@@ -193,6 +200,19 @@ $stacking-order: (
...
+.ClearButton {
+  z-index: z-index(contents, $stacking-order);
+  appearance: none;
+  border: none;
+  padding: 0;
+  margin: 0 spacing(tight);
+  cursor: pointer;
+
+  svg:hover {
+    fill: color('ink', 'lighter');
```
kaelig: Should the hover be on the SVG itself? And what about the focus state (if applicable)?

Also, should this SVG be colored using `@include color-icon('ink', 'lighter');`? 

See https://github.com/Shopify/polaris-react/blob/570d4293256f1da47e5bf311115dc1e89b7454ec/src/components/Icon/Icon.scss#L6-L14

src/components/TextField/TextField.tsx
```diff
@@ -52,6 +53,8 @@ export interface BaseProps {
   labelHidden?: boolean;
   /** Disable the input */
   disabled?: boolean;
+  /** Show a clear text button in the input */
+  clearButton?: boolean;
```
AndrewMusgrave: Rather than add an additional prop should we only show the clear button if `type="search"`?

src/components/TextField/TextField.scss
```diff
@@ -127,6 +127,13 @@ $stacking-order: (
       margin: 0;
     }
   }
+
+  // Hide webkit search clear button
+  &[type='search'] {
+    &::-webkit-search-cancel-button {
```
kvendrik: What happens in FF and IE?

src/components/TextField/TextField.tsx
```diff
@@ -252,6 +256,17 @@ class TextField extends React.PureComponent<CombinedProps, State> {
       </div>
     ) : null;
 
+    const clearButtonMarkup =
```
kvendrik: I feel like it might make sense to not show the button when the value is empty?

src/components/TextField/tests/TextField.test.tsx
```diff
@@ -812,4 +812,69 @@ describe('<TextField />', () => {
...
+      const spy = jest.fn();
+      const textField = mountWithAppProvider(
+        <TextField 
+          id="MyTextField"
+          label="TextField"
+          onChange={spy}
+          clearButton
+        />
+      )
+      textField.find('.ClearButton').simulate('click');
```
kvendrik: Officially we don't use test IDs anymore either but I feel like here it might still make more sense than using the class in this case.

src/components/TextField/tests/TextField.test.tsx
```diff
@@ -812,4 +812,69 @@ describe('<TextField />', () => {
       expect(textField.find(Connected).prop('right')).toEqual(connectedRight);
     });
   });
+
+  describe('clearButton', () => {
+    it('adds a clear button that clears the input value', () => {
```
kvendrik: Considering what you're actually testing here is that `onChange()` is called with an empty value maybe separating this into two tests would make more sense?
1. `renders a clear button when given`
1. `calls onChange() with an empty value when the clear button clicked`

src/components/TextField/tests/TextField.test.tsx
```diff
@@ -812,4 +812,69 @@ describe('<TextField />', () => {
...
+          onChange={spy}
+          clearButton
+        />
+      )
+      textField.find('.ClearButton').simulate('click');
+      expect(spy).toHaveBeenCalledWith('', 'MyTextField');
+    })
+  });
+
+  it('does not render when the clearButton prop is not passed', () => {
```
kvendrik:
```suggestion
  it('doesn’t render by default', () => {
```

src/components/TextField/tests/TextField.test.tsx
```diff
@@ -812,4 +812,69 @@ describe('<TextField />', () => {
...
+      <TextField 
+        id="MyTextField"
+        label="TextField"
+        onChange={noop}
+      />
+    )
+    expect(textField.find('.ClearButton').exists()).toBeFalsy();
+  })
+
+  it('does not render in number inputs, even if clearButton prop is passed', () => {
```
kvendrik:
```suggestion
  it('does not render a clear button when type="number"', () => {
```

src/components/TextField/tests/TextField.test.tsx
```diff
@@ -812,4 +812,69 @@ describe('<TextField />', () => {
...
+        id="MyTextField"
+        label="TextField"
+        onChange={noop}
+        type="search"
+      />
+    )
+    expect(textField.find('.ClearButton').exists()).toBeTruthy();
+  })
+
+  it('does not render in search inputs when clearButton is false', () => {
```
kvendrik:
```suggestion
  it('does not render in search inputs when false', () => {
```

src/components/TextField/TextField.tsx
```diff
@@ -397,6 +413,15 @@ class TextField extends React.PureComponent<CombinedProps, State> {
     onChange(String(newValue.toFixed(decimalPlaces)), this.state.id);
   }
 
+  @autobind
+  private handleClearButtonPress() {
+    const {onChange} = this.props;
+    if (onChange == null) {
+      return;
+    }
+    onChange('', this.state.id);
```
kvendrik: Maybe destructure this above?

src/components/TextField/TextField.tsx
andrewmcgov: Good question @AndrewMusgrave . The [issue](https://github.com/Shopify/polaris-react/issues/1115) suggested adding a prop, but I'm not sure if there is a use case for a clear button on non search inputs.

I wonder if we should only have this on search inputs, but still have the prop so that it could be set to `false` in the event that a clear button is not wanted on a search input?

andrewmcgov:
@AndrewMusgrave @kaelig @kvendrik 

I think I covered all the feedback! If there's anything I missed let me know. 

Im also not sure where to add a note in `UNRELEASED.md` - would this be a breaking change? If existing search inputs have had a similar clear button added by making a custom component, I imagine switching to this text field may cause issues.

yourpalsonja:
Pinging @kaelig @kvendrik @AndrewMusgrave – anyone able to +1 this?

yourpalsonja:
@andrewmcgov looks like you might need to merge in `master`, the branch is out of date!

andrewmcgov:
Hey @yourpalsonja! Just merged master in. Let me know if there is anything else i can adjust!

AndrewMusgrave:
I'm going to dismiss my self as a review as I won't have time to +1 this, however, would you mind adding a test for lines 419 / 420, it should bring `codecov` above 90%, and add a changelog entry to `UNRELEASED.md`. Those two things should make all your checks green 😄 
Screen Shot 2019-05-02 at 11 00 47 AM

kaelig:
Also dismissing myself as I won't have time to look into this contribution.

danrosenthal:
@andrewmcgov @yourpalsonja is this in need of another round of review?

andrewmcgov:
@danrosenthal this is almost good for another review. 

I just have a couple of questions: 
1. I need to add a note to `UNRELEASED.md`, but I am not sure if this would be an enhancement or breaking change. The reason I think it might be a breaking change is that #1115 says that this is already being done in some places, and I'm not sure if this would cause an error with the workaround that might already be in place in some inputs. 
2. Add testing for line 419 / 420 - This part would only run if no `onChange` was passed, and typescript will show a warning for a test without an `onChange`. I could test that the value of the input is unchanged on clearButton click, but I am not sure how to get around the error.

Thanks for the input!

src/components/Icon/Icon.scss
```diff
@@ -3,16 +3,6 @@ $stacking-order: (
...
-@mixin color-icon($value, $hue: base) {
-  svg {
-    fill: color($value, $hue);
-  }
-
-  img {
-    filter: filter($value, $hue);
-  }
-}
-
```
danrosenthal: I see you moved this to a shared icon style file. Makes sense! 👍

src/components/TextField/TextField.scss
danrosenthal: +1 to this question, is this related to the `x` that appears in chrome for search inputs? If so, I think this makes sense, but maybe should be scoped to cases where the clear button is present.

src/components/TextField/TextField.scss
```diff
@@ -193,6 +200,21 @@ $stacking-order: (
...
 }
 
+.ClearButton {
+  z-index: z-index(contents, $stacking-order);
+  appearance: none;
+  background: none;
+  border: none;
+  padding: 0;
+  margin: 0 spacing(tight);
+  cursor: pointer;
```
danrosenthal: We have a mixin called `unstyled-button` that would be perfect for handling most of these styles.

src/components/TextField/TextField.tsx
danrosenthal: I agree with your approach here @andrewmcgov. There may be cases where we want to use the clear button even when the type is not search.

Let's keep it the way you have it, and have it not dependent on the input type.

src/components/TextField/TextField.tsx
danrosenthal: This is a good question for design. @sarahill, do you think it makes sense for the clear button to always be present, or for it to appear once the merchant has typed into the field?

src/components/TextField/TextField.tsx
```diff
@@ -251,6 +255,20 @@ class TextField extends React.PureComponent<CombinedProps, State> {
       </div>
     ) : null;
 
+    const clearButtonMarkup =
+      type === 'search' &&
+      clearButton !== false &&
+      !disabled &&
+      normalizedValue ? (
```
danrosenthal: This logic can be simplified a lot.

```tsx
      clearButton &&
```

This gets rid of the check for the type of the field. Just checking if `clearButton` is falsy will be sufficient. We'd still want to display the button even if the field is disabled. And `nomalizedValue` should always be truthy here, since it's set to an empty string if null.

src/components/TextField/TextField.tsx
```diff
@@ -251,6 +255,20 @@ class TextField extends React.PureComponent<CombinedProps, State> {
...
+        <button
+          testID="clearButton"
+          className={styles.ClearButton}
+          onClick={this.handleClearButtonPress}
```
danrosenthal: This should probably be disabled if disabled is true.

src/components/TextField/TextField.tsx
```diff
@@ -353,6 +371,7 @@ class TextField extends React.PureComponent<CombinedProps, State> {
             {input}
             {suffixMarkup}
             {characterCountMarkup}
+            {clearButtonMarkup}
```
danrosenthal: Curious what this looks like (:tophat:) with a suffix and a character count. Do things get pretty crazy? How's the alignment?

src/components/TextField/TextField.tsx
```diff
@@ -394,6 +413,15 @@ class TextField extends React.PureComponent<CombinedProps, State> {
     onChange(String(newValue.toFixed(decimalPlaces)), this.state.id);
   };
 
+  private handleClearButtonPress = () => {
+    const {id} = this.state;
+    const {onChange} = this.props;
```
danrosenthal: This is mostly stylistic, but questionably more performant if destructured together, so you limit the number of times you reference `this`.

```tsx
const {state: {id}, props: {onChange}} = this;
```

danrosenthal:
> I need to add a note to UNRELEASED.md, but I am not sure if this would be an enhancement or breaking change. The reason I think it might be a breaking change is that #1115 says that this is already being done in some places, and I'm not sure if this would cause an error with the workaround that might already be in place in some inputs.

Since this adds a new prop, we'd consider it an enhancement. If you avoid any logic around type search, and only render when the prop is passed, you avoid any chance of it being a breaking change.

> Add testing for line 419 / 420 - This part would only run if no onChange was passed, and typescript will show a warning for a test without an onChange. I could test that the value of the input is unchanged on clearButton click, but I am not sure how to get around the error.
Thanks for the input!

Since `onChange` is a required prop, you shouldn't need the early return at all. Just remove it and you'll have solid code cov.

src/components/TextField/TextField.tsx
sarahill: I agree with @kvendrik. The clear button only appears once the merchant has typed.

src/components/TextField/TextField.tsx
danrosenthal: Based on Sarah's feedback, you'd also want to not render if `normalizedValue === ''`

src/components/TextField/TextField.scss
andrewmcgov: @danrosenthal Thats right, this is to hide the `x` that appears in chrome / firefox search inputs. I've adjusted it so that it is scoped only to inputs with a clear button. 

Say we have a `TextField` component with `type="search"` and `clearButton={false}` - would we want the browser clearButton to appear in that case? That is how I have it set up right now, but I am not sure if that is the way to go.

src/components/TextField/TextField.tsx
andrewmcgov: Its a bit cramped, but seems okay. I adjusted the margins on the clear button and character count so both match the margins on the suffix:

andrewmcgov:
@danrosenthal thanks for all the feedback! I think I covered everything. 

As for removing the early return for when `onChange` is `null`, typescript shows an error without the early return, so I'm still not sure how to hit the code cov target for those lines since typescript also shows an error when mounting a component with `onChange={null}`.

src/components/TextField/TextField.tsx
```diff
@@ -103,6 +106,8 @@ export interface BaseProps {
   showCharacterCount?: boolean;
   /** Determines the alignment of the text in the input */
   align?: Alignment;
+  /** Callback when clear button is clicked */
+  onClearButtonClick?(id: string): void;
```
danrosenthal: @andrewmcgov I made a change here that I think makes a lot of sense. Instead of using the `onChange` handler passed as a prop to handle the clear button click, we're leaving that implementation up to the consumer. This is more in line with how we think about handlers when a component needs to have its state managed externally.

This also makes these changes a lot easier to cover with testing.

What do you think?

src/components/TextField/TextField.tsx
```diff
@@ -400,6 +420,15 @@ class TextField extends React.PureComponent<CombinedProps, State> {
     onChange(String(newValue.toFixed(decimalPlaces)), this.state.id);
   };
 
+  private handleClearButtonPress = () => {
+    const {
+      state: {id},
+      props: {onClearButtonClick},
+    } = this;
+
+    onClearButtonClick && onClearButtonClick(id);
```
danrosenthal: It's used here.

We can also avoid an early return (and testing for that early return) by simply short circuiting if there is no handler provided.

src/components/TextField/TextField.tsx
andrewmcgov: @danrosenthal I like this - definitely solves the testing problem I was running into. It also lets a consumer do something like `setState({value: 0})` for number inputs, which could be really handy.

I wonder if we should add a default `clearInput` function to the textfield that clears the value to `''` if `onClearButtonClick` is `null`? 

This way adding a clearButton is as easy as passing a single prop, but it still gives the consumer the option to set up their own behaviour if they want to. 

I'm also fine leaving this the way it is, since you mentioned this is more in line with how we would normally set up a component with external state management.

danrosenthal: > I wonder if we should add a default clearInput function to the textfield that clears the value to '' if onClearButtonClick is null?
> 
> This way adding a clearButton is as easy as passing a single prop, but it still gives the consumer the option to set up their own behaviour if they want to.
> 
> I'm also fine leaving this the way it is, since you mentioned this is more in line with how we would normally set up a component with external state management.

This would be tricky given that the value state is managed externally from the component. The click on the clear button being a noop I think is fine and would signal to the consumer that they need to add it.

danrosenthal:
I went ahead and also added an example to the README so folks know how to use this.

|styleguide|
|-|
|Screen Shot 2019-05-13 at 2 01 34 PM|

danrosenthal:
I am good with merging this as-is if you are @andrewmcgov

andrewmcgov:
@danrosenthal i think it is good to go! Just need someone with access to Percy to approve the new change.

PR #1247, opened 2019-03-25
[Link] Warn users when links open in a new window

ry5n:

Fixes Shopify/polaris-ux#146 

This PR takes the position that we should automatically do the right thing for accessibility whenever possible. It automatically adds screen reader text and an icon when the `external` prop is `true`.

Unfortunately, implementation is complicated by CSS things, historical things, and international things (see comments below).

**I’m looking for comments on this approach before going any further.**

### How to 🎩

- The icon will be blank when tophatting—we’re waiting on the correct icon to be available (should be in the next release of Polaris Icons).
- Change the viewport width to make sure the icon wraps with the last word in the link, never on its own.

<details>

```jsx
import {Page, Link, TextContainer} from '../src';

interface State {}

  render() {
    return (
      <Page title="Playground">
        <TextContainer>
          <p>
            This is some sample text to make the paragraph longer, in order to
            test text wrapping. Links to external sites, like the{' '}
            <Link url="https://help.shopify.com/manual" external>
              Shopify Help Center
            </Link>
            , have a trailing icon and open in a new browser tab.
          </p>
          <p>
            これはテキストの折り返しをテストするために段落を長くするためのサンプルテキストです。{' '}
            <Link url="https://help.shopify.com/manual" external>
              Andersonヘルプセンター
            </Link>
            などの外部サイトへのリンクには末尾のアイコンがあり、新しいブラウザタブで開きます。
          </p>
        </TextContainer>
      </Page>
    );
  }
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Link/Link.scss
+++ b/src/components/Link/Link.scss
@@ -1,4 +1,5 @@
 .Link {
+  @include recolor-icon(currentColor);
   appearance: none;
   display: inline;
   text-align: inherit;
@@ -15,6 +16,23 @@
   &:active {
     outline: none;
     color: color('blue', 'dark');
+  }
+}
+
+.IconLockup {
+  display: inline;
+  white-space: nowrap;
+
+  &::before {
+    content: '\2060';
+  }
+}
+
+.IconLayout {
+  display: inline-flex;
+
+  &::before {
+    content: '\2060';
   }
 }
 
--- a/src/components/Link/Link.tsx
+++ b/src/components/Link/Link.tsx
@@ -1,8 +1,11 @@
 import * as React from 'react';
 
 import {classNames} from '@shopify/react-utilities';
+import {ExternalSmallMinor} from '@shopify/polaris-icons';
 
+import {withAppProvider, WithAppProviderProps} from '../AppProvider';
 import UnstyledLink from '../UnstyledLink';
+import Icon from '../Icon';
 
 import styles from './Link.scss';
 
@@ -22,16 +25,37 @@ export interface BaseProps {
 }
 
 export interface Props extends BaseProps {}
+export type CombinedProps = Props & WithAppProviderProps;
 
-export default function Link({
+function Link({
   url,
   children,
   onClick,
   external,
   id,
   monochrome,
-}: Props) {
+  polaris,
+}: CombinedProps) {
   const className = classNames(styles.Link, monochrome && styles.monochrome);
+  let childrenMarkup = children;
+
+  if (external && typeof children === 'string') {
+    const iconLabel = polaris.intl.translate(
+      'Polaris.Common.newWindowAccessibilityHint',
+    );
+
+    childrenMarkup = (
+      <React.Fragment>
+        {children}
+        <span className={styles.IconLockup}>
+          <span className={styles.IconLayout}>
+            <Icon accessibilityLabel={iconLabel} source={ExternalSmallMinor} />
+          </span>
+        </span>
+      </React.Fragment>
+    );
+  }
+
   return url ? (
     <UnstyledLink
       onClick={onClick}
@@ -39,12 +63,15 @@ export default function Link({
       url={url}
       external={external}
       id={id}
+      polaris={polaris}
     >
-      {children}
+      {childrenMarkup}
     </UnstyledLink>
   ) : (
     <button type="button" onClick={onClick} className={className} id={id}>
-      {children}
+      {childrenMarkup}
     </button>
   );
 }
+
+export default withAppProvider<Props>()(Link);
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -12,6 +12,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Added an icon and screen reader hint when `Link` opens a new tab ([#1142](https://github.com/Shopify/polaris-react/pull/1247))
+
 ### Bug fixes
 
 - Fixed Search overlay stretching below the viewport
--- a/src/components/Link/README.md
+++ b/src/components/Link/README.md
@@ -92,6 +92,16 @@ Use for text links that are the same color as the surrounding text.
 </Link>
 ```
 
+### External link
+
+Use for text links that point to a different website. They will open in a new browser tab or window.
+
+```jsx
+<Link url="https://help.shopify.com/manual" external>
+  Shopify Help Center
+</Link>
+```
+
 ---
 
 ## Related components
--- a/src/components/Link/tests/Link.test.tsx
+++ b/src/components/Link/tests/Link.test.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import {mountWithAppProvider} from 'test-utilities';
-import {UnstyledLink} from 'components';
+import {UnstyledLink, Icon} from 'components';
+import en from '../../../locales/en.json';
 import Link from '../Link';
 
 describe('<Link />', () => {
@@ -36,6 +37,48 @@ describe('<Link />', () => {
         <Link url="https://shopify.com" id={id} />,
       );
       expect(link.find(UnstyledLink).prop('id')).toBe(id);
+    });
+  });
+
+  describe('external link', () => {
+    it('has a trailing icon', () => {
+      const link = mountWithAppProvider(
+        <Link url="https://help.shopify.com/" external>
+          Shopify Help Center
+        </Link>,
+      );
+      expect(
+        link
+          .children()
+          .last()
+          .find(Icon)
+          .exists(),
+      ).toBe(true);
+    });
+
+    it('informs screen readers that it opens in a new window', () => {
+      const link = mountWithAppProvider(
+        <Link url="https://help.shopify.com/" external>
+          Shopify Help Center
+        </Link>,
+      );
+      const hintText = en.Polaris.Common.newWindowAccessibilityHint;
+      expect(
+        link
+          .children()
+          .last()
+          .find(Icon)
+          .prop('accessibilityLabel'),
+      ).toBe(hintText);
+    });
+
+    it('doesn’t have a trailing icon for non-string children', () => {
+      const link = mountWithAppProvider(
+        <Link url="https://help.shopify.com/" external>
+          <span>Shopify Help Center</span>
+        </Link>,
+      );
+      expect(link.find(Icon).exists()).toBe(false);
     });
   });
 });
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -31,7 +31,8 @@
     "Common": {
       "checkbox": "checkbox",
       "undo": "Undo",
-      "cancel": "Cancel"
+      "cancel": "Cancel",
+      "newWindowAccessibilityHint": "(opens a new window)"
     },
 
     "ContextualSaveBar": {


src/components/Link/Link.scss
```diff
@@ -18,6 +18,13 @@
   }
 }
 
+.IconLockup {
+  display: inline-flex;
+  align-items: center;
+
+  @include recolor-icon(currentColor);
+}
```
ry5n: In order to center the icon with the text, and to prevent the icon from wrapping by itself to the next line, the best solution I found is to wrap the last word and the icon in an inline-flex container.

src/components/Link/Link.tsx
```diff
@@ -22,29 +25,54 @@ export interface BaseProps {
...
+      <React.Fragment>
+        {rest}
+        <span className={styles.IconLockup}>
+          {lastWord}
+          <Icon accessibilityLabel={iconLabel} source={ExternalSmallMinor} />
+        </span>
+      </React.Fragment>
+    );
+  }
+
```
ry5n: `Link` in Polaris was originally intended only for text links within sentences. In practice, `Link` is used more like React Router’s `Link`: a generic navigation routing component. As such, it can be used to wrap images and block-level elements, and is typed accordingly (`children` accepts React nodes). This makes determining the last word in `children` impossible where `children` is not a string.

This PR is solving only for the case where Link is used as a text link, and is passed a single string in `children`. Based on a quick audit of real-world usage, this is the common case.

src/components/Link/Link.scss
```diff
@@ -18,6 +18,13 @@
   }
 }
 
+.IconLockup {
+  display: inline-flex;
+  align-items: center;
```
ry5n: In order to center the icon with the text, and to prevent the icon from wrapping by itself to the next line, the best solution I found is to wrap the last word and the icon in an inline-flex container.

This way of preventing unwanted wrapping won’t work with languages like Japanese because [those languages don’t use spaces between words](https://en.wikipedia.org/wiki/Japanese_writing_system#Spacing_and_punctuation). I’m not sure what to do about this.

src/components/Link/Link.tsx
```diff
@@ -22,29 +25,54 @@ export interface BaseProps {
...
   external,
   id,
   monochrome,
-}: Props) {
+  polaris,
+}: CombinedProps) {
   const className = classNames(styles.Link, monochrome && styles.monochrome);
+  let childrenMarkup = children;
+
+  if (external && typeof children === 'string') {
```
ry5n: `Link` in Polaris was originally intended only for text links within sentences. In practice, `Link` is used more like React Router’s `Link`: a generic navigation routing component. As such, it can be used to wrap images and block-level elements, and is typed accordingly (`children` accepts React nodes). This makes determining the last word in `children` impossible where `children` is not a string.

This PR is solving only for the case where Link is used as a text link, and is passed a single string in `children`. Based on a quick audit of real-world usage, this is the common case.

src/components/Link/Link.tsx
```diff
@@ -22,29 +25,54 @@ export interface BaseProps {
...
+      'Polaris.Common.newWindowAccessibilityHint',
+    );
+
+    childrenMarkup = (
+      <React.Fragment>
+        {rest}
+        <span className={styles.IconLockup}>
+          {lastWord}
+          <Icon accessibilityLabel={iconLabel} source={ExternalSmallMinor} />
+        </span>
```
ry5n: In order to center the icon with the text, and to prevent the icon from wrapping by itself to the next line, the best solution I found is to wrap the last word and the icon in an inline-flex container.

This way of preventing unwanted wrapping won’t work with languages like Japanese because [those languages don’t use spaces between words](https://en.wikipedia.org/wiki/Japanese_writing_system#Spacing_and_punctuation). I’m not sure what to do about this.

yourpalsonja:
What if we ignore the content of `Link` and just rely on the `external` prop to add a class that:


1. Adds a(n) `:before/:after` element that contains the icon (since we know it won't change) and is positioned absolutely, so that it won't wrap to the next line _but also_
1. Adds padding or a margin to the end of the element so that the entire link will wrap to the next line if things get too jammed up.
1. This would work for both RTL languages, and languages like Japanese.

Caveats: Could be weird if someone decides to use a `:pseudo` element for something else, but I don't think that can happen by default based on my reading of the code.

ry5n:
@yourpalsonja That’s an interesting idea! I tried it a couple ways, and I think something like it could work.

Polaris React uses inline SVGs, relying on them for being able to recolor icons for interaction effects like hover. I am not sure, but I don’t think inline SVGs play well with CSS generated content. BUT, if we made the link `inline-flex`, we would achieve the same overall effect (without needing absolute positioning).

The downside is that as you pointed out, the whole link wraps as a unit, which will cause some disruption to the flow of the text on narrower columns when links are long. Example:

Screen Shot 2019-03-27 at 4 15 22 PM

Both approaches have tradeoffs. Are there other approaches?

dleroux:
I tried inserting a `&nbsp;` at the end of the string and then concatenating the Icon hoping it wouldn't wrap but it did. Guessing because the Icon is html and not just text.

Have we considered something like a Tooltip? In the end visually impaired users are warned of the new window via the `rel` attribute. A simple tooltip would keep the ui clean and meet the requirement of warning sighted users. It's one of the examples provided by the w3c: https://www.w3.org/TR/WCAG20-TECHS/G201.html

1 caveat with our Tooltip implementation is that the link would not wrap.

dpersing:
> Have we considered something like a Tooltip?

@daniedleroux We currently have a few logged a11y issues around tooltips (https://github.com/Shopify/polaris-ux/issues/154, https://github.com/Shopify/polaris-react/issues/793), so I'd like to avoid adding more complexity for the link.

The `rel` attribute isn't actually conveyed to screen reader users via the accessibility API, which is why we've had to implement workarounds like this. There's also concern for sighted keyboard-only users, who will want to be able to fully control their experience, as well as low vision users relying on tools like magnification, and users with cognitive issues, who may not be expecting or notice a new tab/window.

> Adds a(n) :before/:after element that contains the icon (since we know it won't change) and is positioned absolutely, so that it won't wrap to the next line but also

@yourpalsonja If we add the icon as a pseudo-element, we'll need a way to provide the text equivalent for the icon as part of the link content. It doesn't appear that `accessibilityLabel` can be used on the `Link` component, so I recommended including the icon inline.

yourpalsonja:
@ry5n I was thinking the link wrapping in this case would be a good thing, so the text doesn't look too crowded with the link beside it? I could be wrong though.

@dpersing would we need to provide accessible text for the icon, or could we just add text that is read by the screen reader in addition to the icon? (like, "Visit our docs" [icon]), and a screenreader would read "Visit our docs, external link".

ry5n:
Here’s a side-by-side of how the 2 wrapping options behave:

 | Link itself can wrap | Link wraps as a whole |
| --- | --- |
| Screen Shot 2019-03-28 at 6 24 14 PM | Screen Shot 2019-03-28 at 6 24 39 PM |

Here’s what the the current PR does. Screen readers will see “(opens a new window)” appended to the link content. (This follows [Devon’s recommendation in the original issue](https://github.com/Shopify/polaris-ux/issues/146)):

Screen Shot 2019-03-28 at 6 15 32 PM

dpersing:
> would we need to provide accessible text for the icon, or could we just add text that is read by the screen reader in addition to the icon? (like, "Visit our docs" [icon]), and a screenreader would read "Visit our docs, external link".

@yourpalsonja I think it's 6 of 1 and half a dozen of the other!

I also want to clarify that it's less that it's an external link and more that it's a link that opens a new tab/window. The issue isn't with the content, but with the browser behavior, if that makes sense. @ry5n's "(opens a new window)" text does the trick!

ry5n:
OK, I think I found another way. This is close to what Dan was thinking. I’ve tested this with different font sizes, and in Chrome, Firefox, Safari and Edge (CodeSandbox doesn’t work in IE11, so I couldn’t test there). I’ve also tested it with our different system fonts (San Francisco, Roboto, and Segoe UI).

https://codesandbox.io/s/9wymvnyrw

Here’s how this works:

1. It uses a non-breaking space between the Link text and the icon (not sure why this works when it didn’t for Dan).
2. It wraps the Icon component in a span set to inline-flex so that the icon can sit inline with the text.
3. A zero-width space character is included in the span along with the Icon. Because the space is an actual character with font metrics (while the icon isn’t), this ensures that the span has the same height as the rest of the text. Because it has the right height, flexbox centers the icon correctly.
5. The span is pulled backwards with negative margin to negate the width of the non-breaking space. I wanted to use a zero-width non-breaking space, but browsers don’t actually enforce the no-break behavior for this character.

dpersing:
Looks great when text is resized too, @ry5n ! ❤️

ry5n:
……except it actually doesn’t work in Polaris React, trying it for real. This is probably what Dan ran into. I’m stumped.

dleroux:
Looks like the span around the last word might be it. https://snook.ca/archives/html_and_css/icon-wrap-after-text

ry5n:
> Looks like the span around the last word might be it.

Yeah. That brings us back to the code in this PR. I pushed an update that fixes cases where there are no spaces. It seems to work OK with Japanese at least.

I also reached out to the international team to see if there are any more gotchas with this approach. Once I’m confident in internationalization things, I can write some tests and will ask for final code review. Thanks all!

src/components/Link/Link.tsx
michelleyschen: For languages that does not use space between words, you can grab the last character + punctuation as a word. Since each character is consider a word, a single character on the second line is perfectly fine. But punctuation should never be wrap onto the next line by itself.

✅ 
你好
嗎。

❎ 
你好嗎
。

ry5n: Thanks @michelleyschen! Would we want an icon at the end of a text link to behave the same way as punctuation?

michelleyschen: yes, I would consider icon a punctuation

src/components/Link/Link.tsx
```diff
@@ -1,8 +1,11 @@
 import * as React from 'react';
 
 import {classNames} from '@shopify/react-utilities';
+import {ChevronLeftMinor} from '@shopify/polaris-icons';
```
ry5n: This needs to be changed to `ExternalSmallMinor` once this icon is included in Polaris Icons. I believe this is the last blocker for this PR.

src/components/Link/Link.tsx
ry5n: Thanks. I have a new implementation that doesn’t manipulate the children string. Should work for all languages now 😄

ry5n:
OK folks, I think this is almost there. @dleroux I went back to your approach again, really dug into Snook’s article and codepen, and found a way to make it work with inline SVGs. I’m confident now this will work for multiple languages.

The only real thing still blocking this IMO is the correct icon isn’t available in Polaris Icons yet.

dleroux:
Nice work! 🎩 looks good. 

Is the plan to roll this in now or wait until version 4?

src/components/Link/Link.scss
```diff
@@ -18,6 +19,26 @@
   }
 }
 
+// Prevents the icon from wrapping on its own to the next line
```
dleroux: I think these comments can go, it's pretty self explanatory. Could even call this `ExternalIconLockUp` if we want more clarity.

ry5n:
The external link icon merged this morning, so this is ready for final review 😀

Screen Shot 2019-04-08 at 12 53 10 PM

dleroux:
I guess today breaking would mean an api change. I just question the importance of a change when it could visually impact a design. Since we are checking for a ‘string’ the risk is minimal, so either way it should be ok.

PR #1249, opened 2019-03-26
Autocomplete on load more results fix

### LINKED ISSUE #1246: [Autocomplete/Combobox] onLoadMoreResults only called when bottom of list is reached via keyboard
# Issue summary

The autocomplete component's `onLoadMoreResults` prop exists to support lazy loading of options for best performance, but it's only being called when the bottom of the option list is reached with the down arrow key.

Related issue #1245 

## Expected behavior

`onLoadMoreResults` should be invoked when the bottom of the list is reached via keyboard or mouse scroll.

## Actual behavior

`onLoadMoreResults` is only invoked when the bottom of the list is reached via keyboard using the down arrow key.

1. In the sandbox linked below, open the sandbox console so you can see the list length increase on reaching the bottom of the list. 

2. Click or keypress the autocomplete text field to activate the option list popover.

3. Scroll to the bottom of the list, the list length will remain the same and no additional logs will print.

4. Now close and reopen the option list popover and use the down arrow key to navigate the to the bottom of the list. The list length will log each time you hit the bottom of the list and more and more options will be visible.

## Reduced test case

https://codesandbox.io/s/21omo7y01n

## Specifications

- Polaris version number: v3.11.0
- Browser: Chrome
- Device: Macbook Pro
- Operating System: OSX Mojave 10.14.3

chloerice:

Part two [of two](https://github.com/Shopify/polaris-react/pull/1248), fixes #1246

Replaces custom scroll listening with use of `Popover.Pane`'s `onScrollToBottom` prop.

## 

### How to 🎩

Play with the new lazy load example in the deployment below (click "Show environments" for link), or copy/paste the playground code below into `playground/Playground.tsx` if you want to play around with and try to break this change.

<details>
<summary>Click to view collapsed playground code:</summary>

```jsx
import {Page, Card, Stack, Autocomplete, Tag} from '../src';

interface State {}

  paginationInterval = 25;

  options = Array.from(Array(1000)).map((_, index) => ({
    value: `rustic ${index}`,
    label: `Rustic ${index}`,
  }));

  state = {
    selected: [],
    inputText: '',
    options: this.options,
    visibleOptionIndex: this.paginationInterval,
  };

  render() {
    const {visibleOptionIndex, selected, options} = this.state;
    const textField = (
      <Autocomplete.TextField
        onChange={this.updateText}
        label=""
        value={this.state.inputText}
        placeholder="Vintage, cotton, summer"
      />
    );

    const optionList = options.slice(0, visibleOptionIndex);
    console.log(optionList.length);
    const selectedTagMarkup =
      selected.length > 0 ? (
        <Stack spacing="extraTight">{this.renderTags()}</Stack>
      ) : null;
    return (
      <Page title="Autocomplete lazy loading">
        <Card sectioned>
          <Stack vertical>
            {selectedTagMarkup}
            <Autocomplete
              allowMultiple
              options={optionList}
              textField={textField}
              listTitle="Suggested Tags"
              onLoadMoreResults={this.handleLoadMoreResults}
            />
          </Stack>
        </Card>
      </Page>
    );
  }

  handleLoadMoreResults = () => {
    const {visibleOptionIndex} = this.state;
    const nextVisibleOptionIndex = visibleOptionIndex + this.paginationInterval;
    if (nextVisibleOptionIndex <= this.options.length - 1) {
      this.setState({visibleOptionIndex: nextVisibleOptionIndex});
    }
  };

  updateText = (newValue) => {
  };

  removeTag = (tag) => () => {
    const newSelected = this.state.selected;
    newSelected.splice(newSelected.indexOf(tag), 1);
    this.setState({selected: newSelected});
  };

  renderTags = () => {
    return this.state.selected.map((option) => {
      let tagLabel = '';
      tagLabel = option.replace('_', ' ');
      tagLabel = titleCase(tagLabel);
      return (
        <Tag key={`option${option}`} onRemove={this.removeTag(option)}>
          {tagLabel}
        </Tag>
      );
    });
  };

  filterAndUpdateOptions = (inputString) => {
    if (inputString === '') {
      this.setState({options: this.options});

      return;
    }

    );
    let endIndex = resultOptions.length - 1;
    if (resultOptions.length === 0) {
      endIndex = 0;
    }
    this.setState({
      options: resultOptions,
    });
  };

  updateSelection = (selected) => this.setState({selected});
}

function titleCase(string) {
  return string
    .toLowerCase()
    .split(' ')
    .map((word) => {
      return word.replace(word[0], word[0].toUpperCase());
    })
    .join(' ');
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Autocomplete/README.md
+++ b/src/components/Autocomplete/README.md
@@ -58,7 +58,7 @@ class AutocompleteExample extends React.Component {
     const textField = (
       <Autocomplete.TextField
         onChange={this.updateText}
-        label=""
+        label="Tags"
         value={this.state.inputText}
         prefix={<Icon source="search" color="inkLighter" />}
         placeholder="Search"
@@ -134,7 +134,7 @@ class MultiAutocompleteExample extends React.Component {
     const textField = (
       <Autocomplete.TextField
         onChange={this.updateText}
-        label=""
+        label="Tags"
         value={this.state.inputText}
         placeholder="Vintage, cotton, summer"
       />
@@ -163,7 +163,7 @@ class MultiAutocompleteExample extends React.Component {
   };
 
   removeTag = (tag) => {
-    let newSelected = this.state.selected;
+    const {selected: newSelected} = this.state;
     newSelected.splice(newSelected.indexOf(tag), 1);
     this.setState({selected: newSelected});
   };
@@ -241,7 +241,7 @@ class AutocompleteExample extends React.Component {
     const textField = (
       <Autocomplete.TextField
         onChange={this.updateText}
-        label=""
+        label="Tags"
         value={this.state.inputText}
         prefix={<Icon source="search" color="inkLighter" />}
         placeholder="Search"
@@ -282,6 +282,7 @@ class AutocompleteExample extends React.Component {
       const resultOptions = this.options.filter((option) =>
         option.label.match(filterRegex),
       );
+
       this.setState({
         options: resultOptions,
         loading: false,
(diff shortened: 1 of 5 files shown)


src/components/Autocomplete/README.md
```diff
@@ -301,6 +301,123 @@ class AutocompleteExample extends React.Component {
 }
 ```
 
+### Autocomplete with lazy loading
+
+```jsx
+class AutoCompleteLazyLoadExample extends React.Component {
+  paginationInterval = 25;
+
+  options = Array.from(Array(1000)).map((_, index) => ({
```
danrosenthal:
```suggestion
  options = Array(1000).map((_, index) => ({
```

`Array.from` shouldn't be necessary here as `Array(1000)` already returns an array data structure

src/components/Autocomplete/README.md
```diff
@@ -301,6 +301,123 @@ class AutocompleteExample extends React.Component {
...
+    options: this.options,
+    visibleOptionIndex: this.paginationInterval,
+  };
+
+  render() {
+    const {visibleOptionIndex, selected, options} = this.state;
+    const textField = (
+      <Autocomplete.TextField
+        onChange={this.updateText}
+        label=""
```
danrosenthal: It's probably good to include a label in our examples, or at least a hidden one with an accessibility label provided. That way we're illustrating a11y best practices.

src/components/Autocomplete/README.md
```diff
@@ -301,6 +301,123 @@ class AutocompleteExample extends React.Component {
...
+    const resultOptions = this.options.filter((option) =>
+      option.label.match(filterRegex),
+    );
+    let endIndex = resultOptions.length - 1;
+    if (resultOptions.length === 0) {
+      endIndex = 0;
+    }
+    this.setState({
+      options: resultOptions,
+    });
```
danrosenthal: Having `and` in the function name feels like a good sign this should be two methods. I'd suggest pulling the `setState` out into its own function `updateOptions(newOptions)` and call it here.

src/components/Autocomplete/README.md
chloerice: Array(number) returns an array with its length set to the number passed, but it remains an empty array. Passing it to Array.from() is a hack to turn that empty array into an array with that length and undefined at each index so you have something to map over and set whatever you actually want at each index.

danrosenthal: Weird. I was not aware of that hack/behavior. TIL. Thanks for the explanation!

src/components/Autocomplete/README.md
```diff
@@ -301,6 +302,128 @@ class AutocompleteExample extends React.Component {
 }
 ```
 
+### Autocomplete with lazy loading
+
+```jsx
+class AutoCompleteLazyLoadExample extends React.Component {
+  paginationInterval = 25;
+
+  options = Array.from(Array(1000)).map((_, index) => ({
```
AndrewMusgrave: Should we reduce `1000` to `500` or `300`? After loading 500 the list is extremely janky

src/components/Autocomplete/README.md
```diff
@@ -301,6 +302,128 @@ class AutocompleteExample extends React.Component {
...
+  };
+
+  updateText = (newValue) => {
+    this.setState({inputText: newValue});
+    this.filterAndUpdateOptions(newValue);
+  };
+
+  removeTag = (tag) => () => {
+    const newSelected = this.state.selected;
+    newSelected.splice(newSelected.indexOf(tag), 1);
```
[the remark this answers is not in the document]
danrosenthal: Or we could do `const {selected: newSeleced} = this.state;`, and `splice` would no longer be doing work directly on the state property, since destructuring creates something new.

src/components/Autocomplete/README.md
chloerice: Yeah, I noticed while playing around with this example in the deployment yesterday that with large lists (and maybe regardless of list length) there's a bug with selection not working onClick and instead scrolling back to the top of the list that again doesn't seem to be an issue with keyboard, but that's another problem for another PR...

Issue #1213, opened 2019-03-15
[List filters] [a11y] Sheet blends into page in High Contrast


dpersing:
When the sheet is viewed in windows High Contrast with Edge, it blends into the page header and list, making it slightly difficult to pick out. This is caused by relying on a background color alone to distinguish it from the content underneath.

<img width="1872" alt="View of the sheet in Edge" src="https://user-images.githubusercontent.com/1462085/54459407-d7b68000-4723-11e9-8de2-231de7ef6feb.png">

<img width="1872" alt="High contrast view of the sheet" src="https://user-images.githubusercontent.com/1462085/54459314-9625d500-4723-11e9-9302-8c65316e56c4.png">

Giving the sheet a transparent left border (`border: 1px solid transparent`) or any border using the `-ms-high-contrast` query will make a border appear in High Contrast so the sheet is easier to discern.

dpersing:
We looked at browser stats and decided the impact is small enough to not prioritize this for release just now, and any short-term fixes would be bandaid-y for a lot of reasons. When we do redesign work for the new brand we’ll avoid implementations that cause issues with borders by default.

dpersing:
This has been addressed in the sheet component (#1250) but still needs to be rolled out to production.

tmlayton:
This will be addressed by #1167

PR #1250, opened 2019-03-26
[Sheet] build sheet component

danrosenthal:
## Reviewers

⚠️ **This PR is now ready for review**. ~~The only outstanding work needed is improving the existing example, which will be done before merging. Reviewers should not wait on that example to provide review.~~

## Sheet

A sheet is a large container that enters from the edge of the screen when triggered by the merchant. It’s used to provide merchants with actions and information contextual to the page. It doesn’t interrupt their flow like a modal.

|implementation in web|playground desktop|playground mobile|
|-|-|-|
|sheet-web|sheet-playground|sheet-playground-mobile|

## Approach

While I previously planned on allowing this component to be used with an embedded application by delegating rendering to the App Bridge, we have decided to wait on implementing that functionality. So, for now this component is for use only within a standalone application, and will not return markup and will throw a warning if it is not rendered inside the `Frame` component.

This results in a simpler implementation, and allows us to give more thought to whether and how we want to enable use in an embedded context.

## Playground

```tsx
import {
  Page,
  Sheet,
  Button,
  Frame,
  AppProvider,
  TopBar,
  Card,
  TextField,
} from '../src';

interface State {
  sheetActive: boolean;
  value: string;
}

  state: State = {
    sheetActive: false,
    value: '',
  };

  render() {
    const theme = {
      colors: {
        topBar: {
          background: '#357997',
        },
      },
      logo: {
        width: 124,
        topBarSource:
        contextualSaveBarSource:
        url: 'http://jadedpixel.com',
        accessibilityLabel: 'Jaded Pixel',
      },
    };

    const {sheetActive, value} = this.state;
    return (
      <AppProvider theme={theme}>
        <Frame topBar={<TopBar />}>
          <Page title="Sheet">
            <Card sectioned>
              <Button onClick={this.handleToggleSheet}>
                {sheetActive ? 'Close sheet' : 'Open sheet'}
              </Button>
            </Card>
            <Sheet open={sheetActive} onClose={this.handleCloseSheet}>
              <TextField
                label="form"
                onChange={this.handleTextChange}
                value={value}
              />
            </Sheet>
          </Page>
        </Frame>
      </AppProvider>
    );
  }

  handleOpenSheet = () => {
    this.setState({sheetActive: true});
  };

  handleCloseSheet = () => {
    this.setState({sheetActive: false});
  };

  handleTextChange = (value: string) => {
    return this.setState({value});
  };

  handleToggleSheet = () => {
    const {sheetActive} = this.state;

    if (sheetActive) {
      this.handleCloseSheet();
    } else {
      this.handleOpenSheet();
    }
  };
}
```

### DIFF
--- a/src/components/Backdrop/Backdrop.scss
+++ b/src/components/Backdrop/Backdrop.scss
@@ -9,10 +9,14 @@ $backdrop-color: rgba(33, 43, 54, 0.4);
   bottom: 0;
   left: 0;
   display: block;
-  background: $backdrop-color;
+  background-color: $backdrop-color;
   animation: fade-in duration() $entry-iterations forwards;
   opacity: 1;
   will-change: opacity;
+}
+
+.transparent {
+  background-color: transparent;
 }
 
 .belowNavigation {
--- a/src/components/Sheet/README.md
+++ b/src/components/Sheet/README.md
@@ -1,0 +1,284 @@
+---
+name: Sheet
+category: Overlays
+platforms:
+  - web
+keywords:
+  - sheet
+  - modal
+  - open
+  - title
+  - overlay
+  - drawer
+  - dialog
+  - web
+fullSizeExamples: true
+---
+
+# Sheet
+
+A sheet is a large container that enters from the edge of the screen when triggered by the merchant. It’s used to provide merchants with actions and information contextual to the page. It doesn’t interrupt their flow like a modal.
+
+---
+
+## Required components
+
+The sheet component must be wrapped in the [frame](/components/structure/frame) component.
+
+---
+
+## Use in an embedded application
+
+Use of the sheet component in an embedded application is not currently supported. If this is a feature you would like to see supported by Shopify App Bridge, let us know in the [forums](https://community.shopify.com/c/Shopify-APIs-SDKs/bd-p/shopify-apis-and-technology).
+
+---
+
+## Accessibility
+
+Sheets provide an opportunity to let merchants dig into more detail on their current task, or access information for their current task in a different way. Although merchants may be able to see content in the sheet and the main page content at the same time, they should only be expected to interact with one or the other at any given time.
+
+### Keyboard support
+
+- Use the `onClose` prop so that the sheet can be closed with the <kbd>esc</kbd> key as well as with button-based controls
+- Use a button to open the sheet
+- When the sheet opens, focus moves to it so merchants who rely on the keyboard and screen readers can access it
+- Focus is kept in the sheet until it is dismissed
+- When the sheet closes, focus moves back to the button that launched it
+
+---
+
+## Responsive behavior
+
+At small screen sizes, the sheet component enters the page from the bottom of the screen. At larger screen sizes, the sheet component enters the page from the right side of the scren.
+
+---
+
+## Best practices
+
+The sheet component should:
+
+- Include a heading that summarizes the actions and information in the sheet, for example, More filters
+- Be openable through clear actions, like a link or button
+- Be close-able through clear actions, like Done, the [X] button, and the esc key
+- Include information and actions contextual to the current task
+- Not block merchants from completing their task, like a modal would
+- Not open from within another sheet (only one sheet can be open at a time)
+- Preserve its state—the settings and actions won’t reset when it’s closed
+
+The sheet component is best used in cases where the merchant needs to see elements behind it, and for that reason it uses a transparent backdrop. The backdrop is a full screen overlay which closes its parent component when pressed.
+
+---
+
+## Examples
+
+### Basic sheet
+
+<!-- example-for: web -->
+
+Use as the default option for a sheet.
+
+```jsx
+class SheetExample extends React.Component {
+  state = {
+    sheetActive: true,
+    title: 'Big yellow socks',
+    description:
+      "They're big, yellow socks. What more could you possibly want from socks? These socks will change your life.\n\nThey're made from light, hand-loomed cotton that's so soft, you'll feel like you are walking on a cloud.",
+    salesChannels: [
+      {value: 'onlineStore', label: 'Online Store'},
+      {value: 'facebook', label: 'Facebook'},
+      {value: 'googleShopping', label: 'Google shopping'},
+      {value: 'facebookMarketing', label: 'Facebook Marketing'},
+    ],
+    selected: [],
+  };
+
+  render() {
+    const {
+      state: {sheetActive, title, description, salesChannels, selected},
+      handleCloseSheet,
+      handleOpenSheet,
+      handleChange,
+      handleToggleSheet,
+      hasSelectedSalesChannels,
+      selectedSalesChannels,
+    } = this;
+
+    const theme = {
+      colors: {
+        topBar: {
+          background: '#357997',
+        },
+      },
+      logo: {
+        width: 124,
+        topBarSource:
+          'https://cdn.shopify.com/s/files/1/0446/6937/files/jaded-pixel-logo-color.svg?6215648040070010999',
+        contextualSaveBarSource:
+          'https://cdn.shopify.com/s/files/1/0446/6937/files/jaded-pixel-logo-gray.svg?6215648040070010999',
+        url: 'http://jadedpixel.com',
+        accessibilityLabel: 'Jaded Pixel',
+      },
+    };
+
+    const salesChannelsCardMarkup = hasSelectedSalesChannels ? (
+      <List>
+        {selectedSalesChannels.map((channel, index) => (
+          <List.Item key={index}>{channel}</List.Item>
+        ))}
+      </List>
+    ) : (
+      <div
+        style={{
+          alignItems: 'center',
+          display: 'flex',
+          justifyContent: 'space-between',
+          width: '100%',
+        }}
+      >
+        <p>No sales channels selected</p>
+        <Button onClick={handleToggleSheet}>Manage sales channels</Button>
+      </div>
+    );
+
+    const salesChannelAction = hasSelectedSalesChannels
+      ? [
+          {
+            onAction: handleOpenSheet,
+            content: 'Manage sales channels',
+          },
+        ]
+      : null;
+
+    return (
+      <div style={{maxHeight: '640px', overflow: 'visible'}}>
+        <AppProvider theme={theme}>
+          <Frame topBar={<TopBar />}>
+            <Page singleColumn title="Big yellow socks">
+              <Card sectioned>
+                <FormLayout>
+                  <TextField
+                    label="Title"
+                    onChange={handleChange('title')}
+                    value={title}
+                    readOnly
+                  />
+                  <TextField
+                    label="Description"
+                    onChange={handleChange('description')}
+                    value={description}
+                    multiline
+                  />
+                </FormLayout>
+              </Card>
+              <Card
+                sectioned
+                subdued
+                title="Product availability"
+                actions={salesChannelAction}
+              >
+                {salesChannelsCardMarkup}
+              </Card>
+              <Sheet open={sheetActive} onClose={handleCloseSheet}>
+                <div
+                  style={{
+                    display: 'flex',
+                    flexDirection: 'column',
+                    height: '100%',
+                  }}
+                >
+                  <div
+                    style={{
+                      alignItems: 'center',
+                      borderBottom: '1px solid #DFE3E8',
+                      display: 'flex',
+                      justifyContent: 'space-between',
+                      padding: '1.6rem',
+                      width: '100%',
+                    }}
+                  >
+                    <Heading>Manage sales channels</Heading>
+                    <Button
+                      accessibilityLabel="Cancel"
+                      icon="cancel"
+                      onClick={handleCloseSheet}
+                      plain
+                    />
+                  </div>
+                  <Scrollable style={{padding: '1.6rem', height: '100%'}}>
+                    <ChoiceList
+                      choices={salesChannels}
+                      selected={selected}
+                      allowMultiple
+                      onChange={handleChange('selected')}
+                    />
+                  </Scrollable>
+                  <div
+                    style={{
+                      alignItems: 'center',
+                      borderTop: '1px solid #DFE3E8',
+                      display: 'flex',
+                      justifyContent: 'space-between',
+                      padding: '1.6rem',
+                      width: '100%',
+                    }}
+                  >
+                    <Button onClick={handleCloseSheet}>Cancel</Button>
+                    <Button primary onClick={handleCloseSheet}>
+                      Done
+                    </Button>
+                  </div>
+                </div>
+              </Sheet>
+            </Page>
+          </Frame>
+        </AppProvider>
+      </div>
+    );
+  }
+
+  handleOpenSheet = () => {
+    this.setState({sheetActive: true});
+  };
+
+  handleCloseSheet = () => {
+    this.setState({sheetActive: false});
+  };
+
+  handleToggleSheet = () => {
+    const {
+      state: {sheetActive},
+      handleCloseSheet,
+      handleOpenSheet,
+    } = this;
+
+    sheetActive ? handleCloseSheet() : handleOpenSheet();
+  };
+
+  handleChange = (field) => {
+    return (value) => this.setState({[field]: value});
+  };
+
+  get hasSelectedSalesChannels() {
+    return this.selectedSalesChannels.length > 0;
+  }
+
+  get selectedSalesChannels() {
+    const {salesChannels, selected} = this.state;
+
+    return selected.map((key) => {
+      return salesChannels.reduce((accumulator, current) => {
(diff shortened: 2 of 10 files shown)


src/components/Sheet/Sheet.tsx
```diff
@@ -0,0 +1,114 @@
...
+  private handleResize = debounce(
+    () => {
+      const {mobile} = this.state;
+      if (mobile !== isMobile()) {
+        this.setState({mobile: !mobile});
+      }
+    },
+    40,
+    {leading: true, trailing: true, maxWait: 40},
+  );
```
danrosenthal: At this point, this method is the only code path in this component without code coverage. Does anyone have thoughts on how I might test this method?

cc: @tmlayton @AndrewMusgrave

danrosenthal:
@gf3 @Shopify/app-bridge @tmlayton

I plan on making this new component available shortly. The plan is to also make it available through the `AppBridge` in an implementation similar to `Modal`, using a `src` prop that takes an `iframe`. Reason being is this component covers the entire screen, so its rendering will need to be a concern of the `Frame` component.

I'm pinging you to let you know my intent, and to get eyes on the plan I've outlined in the original comment at the top of this PR. If you wouldn't mind taking a look at the plan and calling out anything I've missed, I would hugely appreciate it. Also, as I've never added any new features to the `AppBridge` codebase before, if there's anything I need to look out for, I'd appreciate a heads up.

Thanks in advance!

src/components/Sheet/Sheet.scss
```diff
@@ -0,0 +1,71 @@
...
+}
+.enterRightActive {
+  transform: translateX(0%);
+}
+.exitRight {
+  transform: translateX(0%);
+}
+.exitRightActive {
+  transform: translateX(100%);
+}
```
danrosenthal: I've optimized the transition code for `Bottom` and `Right` (used [here](https://github.com/Shopify/polaris-react/pull/1250/files#diff-a9e19997da37d37567719a0d15889f25R102)) for ease of extraction into some sort of `Motion` component. This is something I'm considering digging into for hackdays, but I'd like to keep that functionality localized here until that's done.

src/components/Sheet/Sheet.tsx
```diff
@@ -0,0 +1,114 @@
...
+      return (
+        <div className={containerClassName} {...layer.props} {...overlay.props}>
+          <TrapFocus trapping={open}>
+            <div role="dialog" tabIndex={-1} className={sheetClassName}>
+              {children}
+            </div>
+          </TrapFocus>
+        </div>
+      );
+    }
```
danrosenthal: For reasons I don't yet understand, in order to take advantage of `CSSTransition`'s ability to mount and unmount a child component, that component must actually be a React component (a func), or it always gets mounted. I would like to understand this limitation better.

src/components/Sheet/Sheet.tsx
```diff
@@ -0,0 +1,114 @@
...
+        <KeypressListener keyCode={Key.Escape} handler={onClose} />
+        <EventListener event="resize" handler={this.handleResize} />
+      </Portal>
+    );
+  }
+}
+
+export function isMobile(): boolean {
+  return navigationBarCollapsed().matches;
+}
```
danrosenthal: Exporting this function as a handy utility for use in components implementing `Sheet`. Consider a use-case where the implementor wants different content to render in a `desktop` or `mobile` context. This method makes implementing that, and keeping it in sync with `Sheet`'s mobile state trivial.

iainmcampbell:
hey @danrosenthal, your plan re App Bridge looks solid 👍

just one lil extra: you’ll want to add an action validator for your new Sheet action set, so people without Typescript can get informative errors. Here’s [the docs for our validation middleware](https://github.com/Shopify/app-bridge/tree/master/packages/app-bridge/src/validate), [the validator for Modal](https://github.com/Shopify/app-bridge/blob/master/packages/app-bridge/src/validate/actions/modal.ts), and [the much simpler validator for Toast](https://github.com/Shopify/app-bridge/blob/master/packages/app-bridge/src/validate/actions/toast.ts). You’ll also need to add your validator to the [middleware file](https://github.com/Shopify/app-bridge/blob/master/packages/app-bridge/src/validate/validator.ts#L39) under `updateValidator()` and `dispatchPayloadValidator()`.

dpersing:
I expanded on the [draft `Accessibility` section](https://docs.google.com/document/d/1Ttg-EtgI-SNBH0lPfw6cAUS_vJnFpmVW_GwfZ6BK8tU/edit#heading=h.ikft37ia630g) to describe expected keyboard behavior and some general concerns about accessing information in sheets. I feel like we've captured quite a bit of the accessibility concerns in the best practices and the documentation that describes what sheet does (and that's good, it's general info!). Very open to edits and thoughts on what else should be included. I'd love to include accessibility-related info for the examples we've discussed as well, once those are mocked up.

sarahill:
@danrosenthal Wondering if we should add `x` for closing the sheet by default? On mobile, it's not clear how to close the sheet.

danrosenthal:
@sarahill 

> Wondering if we should add x for closing the sheet as a default? On mobile, it's not clear how to close the sheet.

I've intentionally avoided this, as adding this button has a cascade of consequences that only add more complexity to what is now a very simple and un-opinionated container. If you add the `X` button, you then have to allow for it in layout, which generally means adding in a title, and thus a title prop. You're then forced to wrap the contents in a `Scrollable` to enable scrolling behavior while the header sticks, and the complexity only continues from there.

My decision instead was to document the need for a close button (something already present in the docs), and to show in an example how a `X` button is best used. This approach allows for the greatest flexibility, while preserving absolute simplicity. I feel strongly it's the right one.

sarahill:
☝️ Fair point. As long as this is clear to designers and developers in the documentation, I'm fine with this approach.

dpersing:
>As long as this is clear to designers and developers in the documentation, I'm fine with this approach.

This will be a good test of whether folks actually look at the docs. 🤞

danrosenthal:
@dpersing @sarahill it's also worth noting that consumers could express that they want this down the line, and we could implement it. It's a lot easier to add when we determine something is desirable, vs. taking something away when we added it prematurely.

src/components/Sheet/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+## Required components
+
+The sheet component must be wrapped in the [frame](/components/structure/frame) component.
+
+---
+
+## Use in an embedded application
+
+Use of the sheet component in an embedded application is not currently supported. If this is a feature you would like to see supported by Shopify App Bridge, let us know in the [forums](https://ecommerce.shopify.com/c/shopify-apis-and-technology).
```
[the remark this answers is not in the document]
iainmcampbell: this forum board is where app bridge related discussions are usually posted: https://community.shopify.com/c/Shopify-APIs-SDKs/bd-p/shopify-apis-and-technology

src/components/Sheet/README.md
```diff
@@ -0,0 +1,203 @@
...
+              <div
+                style={{
+                  display: 'flex',
+                  flexDirection: 'column',
+                  height: '100%',
+                }}
+              >
+                <div
+                  style={{
+                    alignItems: 'center',
```
keyfer: Doens't look like `alignItems` is being accounted for correctly:

src/components/Backdrop/Backdrop.scss
```diff
@@ -9,12 +9,16 @@ $backdrop-color: rgba(33, 43, 54, 0.4);
   bottom: 0;
   left: 0;
   display: block;
-  background: $backdrop-color;
+  background-color: $backdrop-color;
```
dpersing: The current version has the issue where the sheet is difficult to see in Windows High Contrast, since it relies on a background color and drop shadow to distinguish it from the content "underneath" (#1213). I'd like to address this before we ship so it doesn't get lost.

The border color doesn't matter, since it should translate to whatever the user's default is.

```
@media screen and (-ms-high-contrast: active) {
    border-color-left: 1px solid transparent;
  }
```

src/components/Sheet/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+The sheet component should:
+
+- Include a heading that summarizes the actions and information in the sheet, for example, More filters
+- Be openable through clear actions, like a link or button
+- Be close-able through clear actions, like Done, the [X] button, and the esc key
+- Include information and actions contextual to the current task
+- Not block merchants from completing their task, like a modal would
+- Not open from within another sheet (only one sheet can be open at a time)
+- Preserve its state—the settings and actions won’t reset when it’s closed
```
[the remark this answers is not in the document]
ry5n: This is an em dash (hard to see in a monospace font), which we use for interjections. It will look like this in a non-monospace font:

> Preserve its state—the settings and actions won’t reset when it’s closed

I believe this correctly follows Polaris content guidelines.

src/components/Backdrop/Backdrop.scss
danrosenthal: 👍 Great, thanks for the reminder on this. Will get that implemented!

src/components/Sheet/Sheet.scss
```diff
@@ -0,0 +1,71 @@
...
+  bottom: 0;
+  left: 0;
+
+  &.desktop {
+    left: auto;
+    width: $sheet-desktop-width;
+  }
+}
+
+// Bottom
```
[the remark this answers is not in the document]
danrosenthal: Nope, I'll 🔥 them.

src/components/Sheet/README.md
danrosenthal: |:tophat:|
|-|
|Screen Shot 2019-05-03 at 8 51 55 AM|

LGTM

danrosenthal:
@keyfer I've tophatted in web and confirmed that the scrolling issue is a bug with how the Frame is rendered in Storybook. My best guess is there is some extra padding somewhere in the Storybook iframe.

|:tophat:|
|-|
|web-sheet-tophat|

src/components/Sheet/Sheet.tsx
```diff
@@ -0,0 +1,163 @@
...
+      props: {
+        polaris: {intl},
+      },
+    } = this;
+
+    if (mobile !== isMobile()) {
+      this.setState({mobile: !mobile});
+    }
+
+    if (!frame) {
```
keyfer: Not sure in Polaris specifically, but I'd recommend removing the type coercion here/shorthand boolean comparison.

src/components/Sheet/Sheet.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+  render() {
+    const {
+      props: {children, open, onClose},
+      state: {mobile},
+      context: {frame},
+      handleResize,
+    } = this;
+
+    const sheetClassName = classNames(styles.Sheet, !mobile && styles.desktop);
```
keyfer: Same thing here for `!mobile`.

src/components/Sheet/Sheet.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+    const containerClassName = classNames(
+      styles.Container,
+      !mobile && styles.desktop,
```
keyfer: And here, again if this is a patten/best practice in Polaris, then more power to ya!

src/components/Sheet/Sheet.tsx
```diff
@@ -0,0 +1,163 @@
...
+    const finalTransitionProps = {
+      classNames: mobile ? BOTTOM_CLASS_NAMES : RIGHT_CLASS_NAMES,
+      ...sharedTransitionProps,
+    };
+
+    const backdropMarkup = open ? (
+      <Backdrop transparent onClick={onClose} />
+    ) : null;
+
+    return frame ? (
```
keyfer: My preference is on an early return instead of a ternary here, but don't feel too strongly about it.

```javascript

if (frame == null) {
  return null;
}

return (
  <Portal ...>
)
```

danrosenthal:
I've added an updated example that I feel a lot better about:

|Example|
|-|
|Screen Shot 2019-05-03 at 11 18 26 AM|

danrosenthal:
|Styleguide|
|-|
|screencapture-polaris-myshopify-io-components-overlays-sheet-2019-05-03-11_58_37|

danrosenthal:
|updated screenshot of the example in the styleguide with better overflow/height|
|-|
|Screen Shot 2019-05-03 at 12 54 18 PM|

src/components/Sheet/README.md
```diff
@@ -0,0 +1,280 @@
...
+
+## Accessibility
+
+Sheets provide an opportunity to let merchants dig into more detail on their current task, or access information for their current task in a different way. Although merchants may be able to see content in the sheet and the main page content at the same time, they should only be expected to interact with one or the other at any given time.
+
+### Keyboard support
+
+- Use the `onClose` prop so that the sheet can be closed with the <kbd>esc</kbd> key as well as with button-based controls
+- Use a button to open the sheet
+- When the sheet opens, move focus to it so merchants who rely on the keyboard and screen readers can access it
```
dpersing:
```suggestion
- When the sheet opens, move focus to it so merchants who rely on the keyboard and screen readers can access it
- Keep focus in the sheet until it is dismissed
```

I realize I left this out in the sheet draft, but we included it in the original list filters spec.

src/components/Sheet/README.md
```diff
@@ -0,0 +1,280 @@
...
+
+---
+
+## Use in an embedded application
+
+Use of the sheet component in an embedded application is not currently supported. If this is a feature you would like to see supported by Shopify App Bridge, let us know in the [forums](https://community.shopify.com/c/Shopify-APIs-SDKs/bd-p/shopify-apis-and-technology).
+
+---
+
+## Accessibility
```
dpersing: There's a trick to getting the accessibility content to show up in the right place in the style guide, since it's platform specific. I'll open a PR to update that.

danrosenthal:
|up to date docs screenshot|
|-|
|screencapture-polaris-myshopify-io-components-overlays-sheet-2019-05-03-15_15_57|

src/components/Sheet/Sheet.tsx
```diff
@@ -0,0 +1,180 @@
...
+    const {
+      props: {children, open, onClose},
+      state: {mobile},
+      context: {frame},
+      handleResize,
+    } = this;
+
+    const sheetClassName = classNames(
+      styles.Sheet,
+      mobile === false && styles.desktop,
```
yourpalsonja: This could be replaced with a const, since we're checking for the same thing below!

src/components/Sheet/Sheet.tsx
```diff
@@ -0,0 +1,180 @@
...
+            containerClassName={containerClassName}
+            open={open}
+            sheetClassName={sheetClassName}
+          >
+            {children}
+          </Container>
+        </CSSTransition>
+        <KeypressListener keyCode={Key.Escape} handler={onClose} />
+        <EventListener event="resize" handler={handleResize} />
+        {backdropMarkup}
```
yourpalsonja: I wonder if it is worth it to abstract this–as someone scrolling through, it might make more sense to see the code from ln 124 down here, e.g.:

```{open && <Backdrop ...>```

File: src/components/Sheet/Sheet.tsx [hunk 18]
```diff
@@ -0,0 +1,180 @@
...
+  exit: classNames(styles.Bottom, styles.exitBottom),
+  exitActive: classNames(styles.Bottom, styles.exitBottomActive),
+};
+
+export const RIGHT_CLASS_NAMES = {
+  enter: classNames(styles.Right, styles.enterRight),
+  enterActive: classNames(styles.Right, styles.enterRightActive),
+  exit: classNames(styles.Right, styles.exitRight),
+  exitActive: classNames(styles.Right, styles.exitRightActive),
+};
```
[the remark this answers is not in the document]
danrosenthal: I think we definitely could, but I feel like it would be at the expense of readability and searchability. It would involve string interpolation, which would make those classes inherently hard to search for, and hard to read. Not sure it's worth it for the sake of dryness

src/components/Sheet/Sheet.tsx
danrosenthal: With the media query, this is no longer even needed, so I'll just 🔥.

src/components/Sheet/Sheet.tsx
```diff
@@ -0,0 +1,180 @@
...
+      },
+      handleToggleMobile,
+    } = this;
+
+    if (frame == null) {
+      // eslint-disable-next-line no-console
+      console.warn(intl.translate('Polaris.Sheet.warningMessage'));
+    }
+
+    if (mobile !== isMobile()) {
```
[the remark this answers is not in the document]
danrosenthal: I don't think so, since we want to avoid calling `handleToggleMobile` each time this line of code is reached. The idea here is if the component's mobile state doesn't equate to the viewport's mobile state, we want to run the toggle

PR #1254, opened 2019-03-27
[ThemeProvider] Remove componentWillReceiveProps

AndrewMusgrave:

Remove deprecated lifecycle methods. This is in preparation for using the new context api.

Storing context on state, getChildContext will be ran every time state changes and making use of componentDidUpdate.

### How to 🎩

* try out theming examples
* https://github.com/Shopify/polaris-react-deprecated/pull/1947 has quite a few playgrounds from when the theme provider was originally added

</details>

### DIFF
--- a/src/components/ThemeProvider/tests/ThemeProvider.test.tsx
+++ b/src/components/ThemeProvider/tests/ThemeProvider.test.tsx
@@ -96,4 +96,37 @@ describe('<ThemeProvider />', () => {
       '--top-bar-color': 'rgb(255, 255, 255)',
     });
   });
+
+  it('updates themes', () => {
+    const wrapper = mountWithAppProvider(
+      <ThemeProvider
+        theme={{
+          colors: {
+            topBar: {
+              background: '#108043',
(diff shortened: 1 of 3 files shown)


src/components/ThemeProvider/tests/ThemeProvider.test.tsx
```diff
@@ -96,4 +96,37 @@ describe('<ThemeProvider />', () => {
       '--top-bar-color': 'rgb(255, 255, 255)',
     });
   });
+
+  it('updates themes', () => {
```
AndrewMusgrave: Additional coverage to bring us to 💯

AndrewMusgrave:
> I am interested why there is a slight delay in changing the search bar colour. I feel like we could look into that in the future though.

Without pulling this down to check, I'm pretty sure it's our animation that causing the delay.

AndrewMusgrave:
Thanks for putting all the examples into one @chloerice , it was a great idea ❤️

PR #1257, opened 2019-03-27
[FilterControl] Adds placeholder prop

### LINKED ISSUE #1256: ResourceList placeholder prop
# Feature request summary
Adding a `placeholder` prop to the `ResourceList` to allow users to set their own messages in the search area.

## Rationale

image

Adding a prop like this would allow users to customize the placeholder text shown in the search area. Right now it defaults to the value in the `resourceName` prop.

---

Rchristiani:

Fixes #1256 

Currently the `FilterControl` component will use the `resourceName.plural` as the placeholder to the `TextField`. This PR adds a `placeholder` prop to `FilterControl` to allow users to have a custom placeholder.

I added an optional `placeholder` prop to the `FilterControl` component. If passed the `placeholder` on the `TextField` will use the value provided, if not passed it will default to the `resourceName.plural` value.

I also removed a `,` from the `mockDefaultContext` plural string, as it seemed unused. Correct me if i am wrong!

I didn't test accessibility since it is passing a string to the `TextField` so I don't think that is required?

### How to 🎩

<details>

```jsx
import {Page, ResourceList, Avatar, TextStyle, FilterType, Card} from '../src';

interface State {}

  state = {
    searchValue: '',
  };

    this.setState({searchValue});
  };

  renderItem = (item) => {

    return (
        <h3>
        </h3>
        <div>{location}</div>
      </ResourceList.Item>
    );
  };

  render() {
    const resourceName = {
      singular: 'customer',
      plural: 'customers',
    };

    const items = [
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
      },
    ];

    const filterControl = (
      <ResourceList.FilterControl
        placeholder="by name, email or phone"
        additionalAction={{
          content: 'Save',
        }}
      />
    );

    return (
      <Card>
        <ResourceList
          resourceName={resourceName}
          items={items}
          renderItem={this.renderItem}
          filterControl={filterControl}
        />
      </Card>
    );
  }
}

  render() {
    return (
      <Page title="Playground">
        <ResourceListExample />
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/ResourceList/components/FilterControl/FilterControl.tsx
+++ b/src/components/ResourceList/components/FilterControl/FilterControl.tsx
@@ -23,6 +23,7 @@ export interface Props {
   additionalAction?: ComplexAction;
   focused?: boolean;
   filters?: Filter[];
+  placeholder?: string;
   onSearchBlur?(): void;
   onSearchChange(searchValue: string, id: string): void;
   onFiltersChange?(appliedFilters: AppliedFilter[]): void;
@@ -40,18 +41,18 @@ export class FilterControl extends React.Component<CombinedProps> {
       additionalAction,
       focused = false,
       filters = [],
+      placeholder,
       onSearchBlur,
       onSearchChange,
       polaris: {intl},
       context: {selectMode, resourceName},
     } = this.props;
 
-    const textFieldLabel = intl.translate(
-      'Polaris.ResourceList.FilterControl.textFieldLabel',
-      {
-        resourceNamePlural: resourceName.plural.toLocaleLowerCase(),
-      },
-    );
+    const textFieldLabel = placeholder
+      ? placeholder
+      : intl.translate('Polaris.ResourceList.FilterControl.textFieldLabel', {
+          resourceNamePlural: resourceName.plural.toLocaleLowerCase(),
+        });
 
     if (additionalAction) {
       additionalAction.disabled = selectMode;
--- a/src/components/ResourceList/components/FilterControl/tests/FilterControl.test.tsx
+++ b/src/components/ResourceList/components/FilterControl/tests/FilterControl.test.tsx
@@ -22,7 +22,7 @@ describe('<FilterControl />', () => {
     selectMode: false,
     resourceName: {
       singular: 'item',
-      plural: 'items,',
+      plural: 'items',
     },
     selectable: false,
   };
@@ -633,6 +633,37 @@ describe('<FilterControl />', () => {
 
       expect(filterControl.find(FilterCreator).prop('filters')).toMatchObject(
         mockFilters,
+      );
+    });
+  });
+
+  describe('placeholder', () => {
+    it('renders default text if no placeholder passed.', () => {
+      const filterControl = mountWithAppProvider(
+        <Provider value={mockDefaultContext}>
+          <FilterControl {...mockDefaultProps} filters={mockFilters} />
+        </Provider>,
+      );
+
+      expect(filterControl.find(TextField).prop('placeholder')).toBe(
+        'Search items',
+      );
+    });
+
+    it('renders the placeholder prop value if provided', () => {
+      const placeholder = 'Search by name, email or phone';
+      const filterControl = mountWithAppProvider(
+        <Provider value={mockDefaultContext}>
+          <FilterControl
+            {...mockDefaultProps}
+            filters={mockFilters}
+            placeholder={placeholder}
+          />
+        </Provider>,
+      );
+
+      expect(filterControl.find(TextField).prop('placeholder')).toBe(
+        'Search by name, email or phone',
       );
     });
   });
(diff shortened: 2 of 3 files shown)


src/components/ResourceList/components/FilterControl/tests/FilterControl.test.tsx
```diff
@@ -22,7 +22,7 @@ describe('<FilterControl />', () => {
     selectMode: false,
     resourceName: {
       singular: 'item',
-      plural: 'items,',
+      plural: 'items',
```
dleroux: why did you remove the `s`

src/components/ResourceList/components/FilterControl/FilterControl.tsx
```diff
@@ -53,6 +55,12 @@ export class FilterControl extends React.Component<CombinedProps> {
       },
     );
 
+    const placeholderLabel = placeholder
```
dleroux: I would think it's probably best that when a custom placeHolderLabel is provided, the consumer of in responsible for passing the translation.

```
const textFieldLabel = placeholder 
 ? placeholder 
 : intl.translate(
      'Polaris.ResourceList.FilterControl.textFieldLabel',
      {
        resourceNamePlural: resourceName.plural.toLocaleLowerCase(),
      },
    );
```

src/components/ResourceList/components/FilterControl/tests/FilterControl.test.tsx
```diff
@@ -637,6 +637,37 @@ describe('<FilterControl />', () => {
...
+  describe('placeholder', () => {
+    it('renders default text if no placeholder passed.', () => {
+      const filterControl = mountWithAppProvider(
+        <Provider value={mockDefaultContext}>
+          <FilterControl {...mockDefaultProps} filters={mockFilters} />
+        </Provider>,
+      );
+
+      expect(filterControl.find(TextField).prop('placeholder')).toBe(
+        'Search items',
```
dleroux: You'll need to translate that string, but if we opt to not translate it and let the consumer pass the translate text then we just need to check that it was passed as is.

src/components/ResourceList/components/FilterControl/tests/FilterControl.test.tsx
```diff
@@ -637,6 +637,37 @@ describe('<FilterControl />', () => {
...
+      );
+    });
+
+    it('renders the placeholder prop value if provided', () => {
+      const placeholder = 'by name, email or phone';
```
dleroux: See this does makes me think it's better to be 100% up to the user what they pass in. Search... was opinionated base we knew what we would render, but having to provide a placeholder that will be prepended by Search requires the consumer to know this.

src/components/ResourceList/components/FilterControl/tests/FilterControl.test.tsx
Rchristiani: I didn't remove the `s` it was a `,` that seemed to have no tests against it. So when I was writing my test and checking the default the value was `Search items,` and that made no sense to me.

src/components/ResourceList/components/FilterControl/FilterControl.tsx
Rchristiani: Are you thinking that instead of translating the `Search ` part we can just leave the transitional completely up to the consumer? That sounds like a good idea!

src/components/ResourceList/components/FilterControl/tests/FilterControl.test.tsx
Rchristiani: I think if we go with the consumer doing the translation, then I am cool just testing the value passed as is.

src/components/ResourceList/components/FilterControl/tests/FilterControl.test.tsx
Rchristiani: OK, sounds good! So just to confirm I will remove the `Search` part and the `placeholder` can be anything the user wants it to be?

src/components/ResourceList/components/FilterControl/FilterControl.tsx
dleroux: Yeah, either they use the default we have today, nothing gets changed there but if they do provide a placeholder text it gets used for both the label and the placeholder (because in thsi case they would be the same) as provided by the consumer.

src/components/ResourceList/components/FilterControl/tests/FilterControl.test.tsx
dleroux: Yeah I would think so.

src/locales/en.json
```diff
@@ -171,7 +171,8 @@
       },
 
       "FilterControl": {
-        "textFieldLabel": "Search {resourceNamePlural}"
+        "textFieldLabel": "Search {resourceNamePlural}",
+        "placeholderLabel": "Search {placeholder}"
```
dleroux: 🔥

PR #1259, opened 2019-03-27
[TextField] Allow autoComplete prop of TextField to accept string values

### LINKED ISSUE #1244: Chrome may override disabled autoComplete in TextFields
# Issue summary

Chrome is known to sometimes ignore the autoComplete=false setting on inputs: https://bugs.chromium.org/p/chromium/issues/detail?id=370363#c7

The workaround is to set it to a string field named differently from the ones autocomplete looks for:
https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill

## Expected behavior

TextField should be able to handle a string prop as well as a boolean prop for autocomplete, and then set autocomplete to the value of the string if the prop is a string.

## Actual behavior

TextField only accepts boolean or null values.

## Specifications

- Polaris version number: 3.10.0
- Browser: Chrome Version 72.0.3626.121 (Official Build) (64-bit)
- Device: mac
- Operating System: Mojave 10.14.3

marina101:

Fixes #1244  This will allow the autoComplete prop of TextField to accept a string and set the autocomplete value of the html input to a string. This is because Chrome is known to ignore autocomplete=off, but setting autocomplete=fieldname seems to disable wrongful autocompletion by chrome. 

For background on how chrome ignores autocomplete=off when it comes to autofill: https://developers.google.com/web/updates/2015/06/checkout-faster-with-autofill

Longer term if would be good to directly support autofill, but this is an urgent quick fix in the meantime.

cc: @axelerator @Senjai 

Allows autoComplete prop to accept string value and set the value of html input autocomplete to that string (while keeping origin behavior when true, false, or null are given as prop value)

## 

### How to 🎩

<details>

```jsx

interface State {}

  render() {
    return (
      <Page title="Playground">
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/TextField/tests/TextField.test.tsx
+++ b/src/components/TextField/tests/TextField.test.tsx
@@ -190,6 +190,17 @@ describe('<TextField />', () => {
       );
       expect(textField.find('input').prop('autoComplete')).toBe('on');
     });
+
+    it('sets autoComplete to string value when string is given', () => {
+      const textField = shallowWithAppProvider(
+        <TextField
+          label="TextField"
+          autoComplete="happyString"
+          onChange={noop}
+        />,
+      );
+      expect(textField.find('input').prop('autoComplete')).toBe('happyString');
+    });
   });
 
   describe('helpText', () => {
--- a/src/components/TextField/tests/TextField.test.tsx
+++ b/src/components/TextField/tests/TextField.test.tsx
@@ -195,11 +195,11 @@ describe('<TextField />', () => {
       const textField = shallowWithAppProvider(
         <TextField
           label="TextField"
-          autoComplete="happyString"
+          autoComplete="firstName"
           onChange={noop}
         />,
       );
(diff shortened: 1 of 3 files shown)


src/components/TextField/tests/TextField.test.tsx
```diff
@@ -190,6 +190,17 @@ describe('<TextField />', () => {
       );
       expect(textField.find('input').prop('autoComplete')).toBe('on');
     });
+
+    it('sets autoComplete to string value when string is given', () => {
+      const textField = shallowWithAppProvider(
+        <TextField
+          label="TextField"
+          autoComplete="happyString"
```
danrosenthal: This is a bit of a nitpick, but I wonder if the example in the test case ought to better illustrate a real-world use-case for passing a string. That way folks reading the tests can learn more about its usage. What do you think?

marina101: 👍 will change the string to "fieldName" since thats the intention for it.

PR #1266, opened 2019-03-29
export overlay and layer data attributes for use in external components

danrosenthal:


AndrewMusgrave:
Is this still a WIP?

danrosenthal:
No, if this is something we want to do, feel free to merge @AndrewMusgrave. @CameronGorrie does your approval mean this is something you want to use?

CameronGorrie:
I opted to add the data attribute where I needed it. The export would be nicer but if the sheet component lands in Polaris at some point my use case for exposing it goes away.

PR #1268, opened 2019-03-29
Upgrade to polaris-tokens 2.5.0

beefchimi:
> We have recently added all Polaris `duration` values to the `polaris-tokens` repo and just cut a new `2.5.0` release! [Read more here](https://github.com/Shopify/polaris-tokens/releases/tag/v2.5.0).

The goal of this PR is to consume the newest `polaris-tokens` release and make sure all `duration` references in `polaris-react` are updated and consistent. This PR would resolve: https://github.com/Shopify/polaris-react/issues/1258

We have upgraded from `2.3.0` to `2.5.0`. The `2.4.0` release appears to be just an enhancement - but perhaps @tmlayton can speak more about that release?

**This PR accomplishes the following:**
- [x] Upgrades our version of `polaris-tokens`:
  - [x] Upgrades to the `polaris-tokens > 2.4.0-beta.1` package for initial tophatting.
  - [x] Upgrades to the `polaris-tokens > 2.4.0-beta.2` package for additional tophatting.
  - [x] Upgrades to the stable release of `2.5.0` after positive review.
- [x] Updates `duration.scss` to:
  - [x] Remove the hard coded SASS map.
  - [x] Begin consuming the `duration` map from `polaris-tokens`.
  - [x] Makes any necessary adjustments to the `duration()` function.
  - [x] Consume correct `duration` SASS map? cc/ @kaelig _(see more context below)_
- [x] Updates any relevant JS files:
  - [x] Remove the `enum Duration`.
  - [x] Update all files consuming the `enum Duration` and replace with appropriate `duration` value from `polaris-tokens`.
- [x] Updates `Frame.scss` to use Polaris values _(this task might be up for debate)_:
  - [x] Replace custom `$transition-duration` with closest Polaris `duration` _(`duration(slow)`)_.
  - [x] Replace browser `ease` timing-function with closest Polaris `easing` _(`easing(out)`)_.

@kaelig both `spacing` and `color` export 2 different SCSS maps... and Polaris seems to prefer the `spacing.spacing-map.scss` version. Did we accidentally omit this from the `beta` release? I ended up modifying the `duration` SCSS function to retain the same API _(I just interpolate the string to include the `duration-` prefix)_.

I'm curious what everyone thinks of me simply using `parseInt(durationValue, 10)` on the imported `duration` from `polaris-tokens`? This seems to work perfectly for stripping off the `ms` suffix and converting to a proper `ms` value for JavaScript. Are we cool with this?

I tophatted in the local `dev` storybook and everything looked good, but I'd love another tophat from someone on the Polaris team.

cc/ @AndrewMusgrave + @dleroux on the `Frame.scss` task... curious why the previous values were being used? Was that intentional?

### DIFF
--- a/src/components/Frame/Frame.scss
+++ b/src/components/Frame/Frame.scss
@@ -1,11 +1,12 @@
 $button-size: rem(32px);
-$transition-duration: 260ms;
 $skip-vertical-offset: rem(10px);
+
 .Frame {
   width: 100%;
   min-height: 100vh;
   display: flex;
   background-color: color('sky', 'light');
+
   @include when-printing {
     background-color: none;
   }
@@ -46,16 +47,19 @@ $skip-vertical-offset: rem(10px);
 .Navigation-enter {
   transform: translateX(-100%);
 }
+
 .Navigation-enterActive {
   transform: translateX(0%);
-  transition: transform $transition-duration ease;
+  transition: transform duration(slow) easing(out);
 }
+
 .Navigation-exit {
   transform: translateX(0%);
 }
+
 .Navigation-exitActive {
   transform: translateX(-100%);
-  transition: transform $transition-duration ease;
+  transition: transform duration(slow) easing(out);
 }
 
 .NavigationDismiss {
@@ -72,17 +76,20 @@ $skip-vertical-offset: rem(10px);
   opacity: 0;
   pointer-events: none;
   will-change: opacity;
-  transition: opacity duration(fast) easing();
   cursor: pointer;
+  transition: opacity duration(fast) easing();
+
   .Navigation-visible & {
     pointer-events: all;
     opacity: 1;
   }
+
   &:focus {
     border-radius: border-radius();
     background-color: rgba(color('white'), 0.16);
     outline: none;
   }
+
   @include frame-when-nav-displayed {
     display: none;
   }
@@ -104,19 +111,23 @@ $skip-vertical-offset: rem(10px);
   left: 0;
   width: 100%;
 }
+
 .ContextualSaveBar-enter {
   opacity: 0;
 }
+
 .ContextualSaveBar-enterActive {
   opacity: 1;
-  transition: opacity duration() ease;
+  transition: opacity duration() easing(out);
 }
+
 .ContextualSaveBar-exit {
   opacity: 1;
 }
+
 .ContextualSaveBar-exitActive {
   opacity: 0;
-  transition: opacity duration() ease;
+  transition: opacity duration() easing(out);
 }
 
 .Main {
@@ -158,6 +169,7 @@ $skip-vertical-offset: rem(10px);
   z-index: z-index(global-ribbon, $fixed-element-stacking-order);
   bottom: 0;
   width: 100%;
+
   @include frame-when-nav-displayed {
     .hasNav & {
       left: layout-width(nav);
@@ -174,6 +186,7 @@ $skip-vertical-offset: rem(10px);
   top: 0;
   right: 0;
   left: 0;
+
   @include frame-when-nav-displayed {
     .hasTopBar & {
       top: top-bar-height();
@@ -189,6 +202,7 @@ $skip-vertical-offset: rem(10px);
   left: spacing(tight);
   opacity: 0;
   pointer-events: none;
+
   &.focused {
     pointer-events: all;
     opacity: 1;
--- a/src/components/Frame/Frame.tsx
+++ b/src/components/Frame/Frame.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import {MobileCancelMajorMonotone} from '@shopify/polaris-icons';
 import {classNames} from '@shopify/react-utilities/styles';
+import {durationSlow} from '@shopify/polaris-tokens';
 import {CSSTransition} from 'react-transition-group';
 import {navigationBarCollapsed} from '../../utilities/breakpoints';
 import Button from '../Button';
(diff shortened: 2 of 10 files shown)


src/components/Modal/components/Dialog/Dialog.tsx
```diff
@@ -54,7 +54,7 @@ export default function Dialog({
       {...props}
       mountOnEnter
       unmountOnExit
-      timeout={Duration.Base}
+      timeout={parseInt(durationBase, 10)}
```
kaelig: I thought this was already going to be an integer. Is it not?

If not - we could adapt polaris-tokens to export an int as expected.

beefchimi: Looks like that is not the case 😭 

```
  durationBase: "200ms";
  durationFast: "100ms";
  durationNone: 0;
  durationSlow: "300ms";
  durationSlower: "400ms";
  durationSlowest: "500ms";
```

Screen Shot 2019-03-29 at 10 18 14 AM

kaelig: Thanks! This is a bug, let me look into it!

beefchimi: Should we be writing tests to check for expected values?

kaelig: There is a set of snapshot tests, which have been pretty helpful so far, but more tests wouldn't hurt!

src/styles/foundation/duration.scss
```diff
@@ -1,24 +1,19 @@
...
 
 /// Returns the duration value for a given variant.
 ///
 /// @param {String} $variant - The key for the given variant.
-/// @return {Number} The spacing for the variant (in miliseconds).
+/// @return {Number} The duration for the variant (in miliseconds).
 
 @function duration($variant: base) {
-  $fetched-value: map-get($duration-data, $variant);
+  $interpolated-value: 'duration-' + $variant;
```
beefchimi: I guess the only question that remains is whether or not we want to keep these _"foundation sass functions"_ consistent and always consume a `token.token-map.scss` format vs the `token.map.scss` format I am currently using.

In the solution presented above, I have to "interpolate" the _"token prefix"_ in order to keep the same consumer API (`color: token('variant');`) – otherwise, we would need to pass in the full name of the key. Alternatively, you can forgo the custom function and just use `map-get`:
`transition-duration: map-get($polaris-duration-map, 'duration-tight');`

Using `spacing` as an example, here is the difference:

**spacing.map.scss**
```scss
$polaris-spacing-map: (
  'spacing-none': (
    0,
  ),
  'spacing-extra-tight': (
    4px,
  ),
  // ...
);
```

**spacing.spacing-map.scss**
```scss
$polaris-spacing: (
  'none': 0,
  'extra-tight': 4px,
  // ...
);
```

Does anyone have an opinion on this? If we want the `token.token-map.scss` format, we will need to do another `polaris-tokens` beta.

src/components/Modal/components/Dialog/Dialog.tsx
beefchimi: Updated to `beta2` and no longer using `parseInt`

beefchimi:
Over Slack, @kaelig brought up an interesting question:

> I wonder if those are usable by the iOS and Android teams or if they have completely different timings to respect because of the differences in platforms

Anyone know the answer to this? cc/ @alex-page

src/components/Frame/Frame.scss
```diff
@@ -1,6 +1,6 @@
 $button-size: rem(32px);
-$transition-duration: 260ms;
```
kaelig: This PR adds `40ms` to this transition, as `260ms` is going up to `300ms` (`duration-slow = 300ms`), was there agreement reached to do this? Also: should we mention this change in the release notes?

dleroux: To answer to @beefchimi I think this was an oversight, probably just ported over when we moved frame from web. I'll 🎩

src/components/Frame/Frame.tsx
```diff
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import {MobileCancelMajorMonotone} from '@shopify/polaris-icons';
 import {classNames} from '@shopify/react-utilities/styles';
+import {durationBase} from '@shopify/polaris-tokens';
```
dleroux: Since you are using `duration(slow)` in the CSS this one should use `DurationSlow`.

The timeout on the `CSSTransition` needs to match the css timing.

src/components/Popover/components/PopoverOverlay/PopoverOverlay.tsx
```diff
@@ -61,12 +62,7 @@ export default class PopoverOverlay extends React.PureComponent<Props, never> {
   render() {
     const {active} = this.props;
     return (
-      <Transition
-        in={active}
-        timeout={Duration.Fast}
-        mountOnEnter
-        unmountOnExit
-      >
+      <Transition in={active} timeout={durationFast} mountOnEnter unmountOnExit>
```
dleroux: Pretty certain this one was wrong too. Should be like the CSS which is base. Can you double check that I am correct here?

alex-page:
> Over Slack, @kaelig brought up an interesting question:
> 
> > I wonder if those are usable by the iOS and Android teams or if they have completely different timings to respect because of the differences in platforms
> 
> Anyone know the answer to this? cc/ @alex-page

I am actually not sure how ios and android do transitions 😕. Looking at these pages:
- https://www.swiftbysundell.com/basics/animations
- https://developer.android.com/reference/android/transition/Transition.html

It seems the tokens could be used on ios and android.

src/styles/foundation/duration.scss
dleroux: Not sure I follow entirely. Do you know why we provide both?

beefchimi: @kaelig and I made an executive decision to stop providing the snowflake `{token}.{token}-map.scss` style map.

Read my in-depth comment here:
https://github.com/Shopify/polaris-tokens/pull/54#issuecomment-481746556

In our opinion, we should remove this extra build step from `polaris-tokens` to produce the "snowflake" map and go all in on providing just the `{token}.map.scss`. Future token maps will require Polaris to adjust their "property function" slightly, but thats a trivial task and upgraders can just look at this PR for guidance!

PR #1273, opened 2019-04-02
[DataTable] Fix sortable header clickable area

### LINKED ISSUE #1230: [DataTable] Sortable column headings clickable area is inconsistent with styling
# Issue summary

When a DataTable column is sortable, hovering over the column header with your mouse changes the color and reveals the sorting icon, indicating that it is clickable.

However, the area where it is actually clickable is smaller than the area where the hover styles are applied, making the user think they can click to sort when they can't

In this GIF I'm clicking, it has no effect until it's right on the text label:

sortable-data-table-click

## Expected behavior

The clickable area & hover style area are the same

## Actual behavior

The clickable area is smaller than the hover style area

1. can see on any sortable DataTable

## Reduced test case

## Specifications

- Polaris version number:
- Browser:
- Device:
- Operating System:

andrewlo:

Fixes #1230

When a `DataTable` column is sortable, hovering over the column header with your mouse changes the colour and reveals the sorting icon, indicating that it is clickable.

However, the area where it is actually clickable is smaller than the area where the hover styles are applied, making the user think they can click to sort when they can't.

In this GIF I'm clicking, it has no effect until it's right on the text label:

sortable-data-table-click

Remove padding from the surrounding `Cell` and add it to the `button` when the column is sortable, to expand the clickable area to the whole cell. The button CSS seems to already have the hover styles that now apply to the whole area.

The whole cell area is now clickable:

table sort good

### How to 🎩

Can be 🎩 on the sortable `DataTable` examples on Storybook

</details>

### 🎩 checklist

### DIFF
--- a/src/components/DataTable/DataTable.scss
+++ b/src/components/DataTable/DataTable.scss
@@ -132,18 +132,7 @@ $breakpoint: 768px;
 }
 
 .Cell-sortable {
-  cursor: pointer;
-
-  &:hover {
-    .Heading {
-      @include recolor-icon(color('indigo'));
-      color: color('indigo');
-    }
-
-    .Icon {
-      opacity: 1;
-    }
-  }
+  padding: 0;
 }
 
 .Icon {
@@ -160,6 +149,8 @@ $breakpoint: 768px;
   justify-content: flex-end;
   align-items: baseline;
   transition: color duration() easing();
+  padding: spacing();
+  cursor: pointer;
 
   &:hover,
   &:focus {
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -16,6 +16,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - Fixed Search overlay stretching below the viewport
 - Stopped passing the `polaris` context into the div rendered by `Scrollable` ([#1271](https://github.com/Shopify/polaris-react/pull/1271))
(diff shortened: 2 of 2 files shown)


src/components/DataTable/DataTable.scss
```diff
@@ -160,6 +149,8 @@ $breakpoint: 768px;
   justify-content: flex-end;
   align-items: baseline;
   transition: color duration() easing();
+  padding: spacing();
```
ry5n: What is the padding value achieving?

src/components/DataTable/DataTable.scss
```diff
@@ -132,18 +132,7 @@ $breakpoint: 768px;
...
-  cursor: pointer;
-
-  &:hover {
-    .Heading {
-      @include recolor-icon(color('indigo'));
-      color: color('indigo');
-    }
-
-    .Icon {
-      opacity: 1;
```
ry5n: Any idea what the opacity here was for? Any issues with removing it?

chloerice: That's the hover style that starts at 0.

src/components/DataTable/DataTable.scss
andrewlo: This padding was on the Cell surrounding the button. In order to make the button take up the whole cell, I made the padding 0 on the Cell and moved the padding here to the button instead

PR #1285, opened 2019-04-08
Added sentence about in-card/modal/context banners

yangchen1:
Gives clear answer about when to use the in-context banner vs. full banner without having to look for specific examples

Updates best practices section in Polaris.


yangchen1:
@chloerice Done! Just need for changelog and and percy to pass :) thanks for the commands

BPScott:
Percy tests were flakey (I've fixed them in a different PR, and we don't need a changelog entry for this), so I've went ahead and used my adminness to merge

PR #1290, opened 2019-04-08
Stop false positives in percy

BPScott:

We keep getting false positives in percy because dynamic values change between renderings

Uses Percy's media query (see https://docs.percy.io/docs/percy-specific-css) to fix the rendering of elements only when rendered by percy

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -7,4 +7,22 @@
   body {
     background-color: initial !important;
   }
+
+  /**
+   * Override dynamically generated values to be fixed values instead,
+   * to eliminate false positives due to ever-changing values between Percy
+   * snapshots. See https://docs.percy.io/docs/percy-specific-css.
+   *
+   * Selectors need to be in the format `[class^='Thing']` because compiled
+   * class names have unique suffixes attached to them (for example: `Thing_20YWn`).
+   */
+  @media only percy {
+    [class^='Loading-Level'] {
+      transform: scaleX(0.1) !important;
+    }
+
+    [class^='SkeletonPage-Action'] {
+      width: 80px !important;
+    }
+  }
 </style>


.storybook/preview-head.html
```diff
@@ -7,4 +7,26 @@
   body {
     background-color: initial !important;
   }
+
+  /**
+   * Percy-specific CSS doesn't play nicely with css-in-js solutions, so we
+   * need to add overrides here instead of in the relevant css files.
+   * See See https://docs.percy.io/docs/percy-specific-css
```
kaelig:
```suggestion
   * See https://docs.percy.io/docs/percy-specific-css
```

kaelig:
Is there a way to strip those media rules in the production build?

BPScott:
> Is there a way to strip those media rules in the production build?

The moment you commented I tried a different approach that puts the overrides within percy instead of the source code.

.storybook/preview-head.html
```diff
@@ -7,4 +7,26 @@
   body {
     background-color: initial !important;
   }
+
+  /**
+   * Percy-specific CSS doesn't play nicely with css-in-js solutions, so we
+   * need to add overrides here instead of in the relevant css files.
+   * See https://docs.percy.io/docs/percy-specific-css
+   *
+   * Selectors need to be in the format `[class^='thing'] because the actual
```
kaelig:
```suggestion
   * Selectors need to be in the format `[class^='thing']` because the actual
```

.storybook/preview-head.html
```diff
@@ -7,4 +7,26 @@
...
+   * class name will have a unique suffix attached to it.
```
kaelig: For people with little knowledge around this, perhaps an example could help?

```suggestion
   * class name (for example: `thing_20YWn`) will have a unique suffix attached to it.
```

.storybook/preview-head.html
```diff
@@ -7,4 +7,26 @@
...
+   * need to add overrides here instead of in the relevant css files.
+   * See https://docs.percy.io/docs/percy-specific-css
+   *
+   * Selectors need to be in the format `[class^='Thing']` because the actual
+   * class name will have a unique suffix attached to it (e.g. `Thing_20YWn`).
+   *
+   * These overrides exist because normally these values are dynamically
+   * generated in JS and thus the differ between Percy snapshots, which lead
+   * to constant false positives unless we force them to a fixed value.
+  */
```
kaelig: Not sure the first paragraph is that useful?

How about a more minimalist approach:

```js
/**
 * Override dynamically generated values to fixed values instead,
 * eliminating false positives due to ever-changing values between Percy snapshots.
 *
 * Selectors need to be in the format `[class^='Thing']` because compiled
 * class names have unique suffixes attached to them (for example: `Thing_20YWn`).
 * 
 * @see https://docs.percy.io/docs/percy-specific-css
 */
```

PR #1291, opened 2019-04-09
Add initial configuration for translation platform

translation-platform:
**Owners:** @Shopify/polaris

### Welcome to Shopify [Translation Platform](https://development.shopify.io/engineering/developing_at_Shopify/internationalization/translation_platform)

This Pull Request contains the configuration files required to have your source (English) strings sent for translation, and receive translations back into your repository.

### Please review this PR before merging it!

- [x] Review the configuration to familiarize yourself with the [configuration options](https://development.shopify.io/engineering/developing_at_Shopify/internationalization/translation_platform#Configuration_file) available.
- [x] Make sure the specified components in `translation.yml` cover all your source files.
- [x] Make sure the name of the component represents the audience of your project:
  - `buyer` for end users of Shopify stores
  - `merchant` where the users are merchants
  - `partner` for Shopify partners

Questions about this Pull Request? Hop in the [#intl-services](https://shopify.slack.com/messages/C7TJQLVC7) Slack channel. We're happy to help!

@Shopify/translation-platform

### DIFF
--- a/translation.yml
+++ b/translation.yml
@@ -1,0 +1,8 @@
+source_language: en
+target_languages: [de, es, fr, hi, it, ja, ms, nl, pt-BR, zh-CN, zh-TW]
+components:
+  - name: 'default'
+    scheme: react
+    audience: merchant
+    paths:
+      - src/locales/{{language}}.json


nimzco:
Technically, since only the english translation is loaded here: https://github.com/Shopify/polaris-react/blob/master/src/locales/index.ts we could start to have all the translations and then figure out how to make it work.

Meaning: Polaris will still come in English by default, but we'll have all translations available for developers using Polaris in other repos.

chloerice:
> Technically, since only the english translation is loaded here: https://github.com/Shopify/polaris-react/blob/master/src/locales/index.ts we could start to have all the translations and then figure out how to make it work.
> 
> Meaning: Polaris will still come in English by default, but we'll have all translations available for developers using Polaris in other repos.

Should we document this in the `README.md`?

nimzco:
> Should we document this in the README.md?

We can, for sure 👍

danrosenthal:
@nimzco and @chloerice is this fine to be merged? If so, can one of you handle that?

BPScott:
Hi @nimzco, It looks like this is using the base config for some kind of rails app.

We've already got translations available in the "react" style in https://github.com/Shopify/polaris-react/blob/master/src/locales/en.json. Rather than create a whole new yml file (that is i no way hooked up to our build process should we point the translation platform at that existing file instead?


Having had a quick skim read of https://development.shopify.io/engineering/developing_at_Shopify/internationalization/translation_platform#Adding_your_project it suggests that as we already have translations present that we'd like to import into the platform we should contact  #intl-services before kicking this process off. Have you done that?

Rather than merge a config file that points at nonexistent translation sources I'd rather we configure this to use the translation sources we already have.

nimzco:
Hey @BPScott, thanks for your message. I totally overlooked that it was pointing to `yml` files... 🤦‍♂️ 
We can just follow what we have [in web](https://github.com/Shopify/web/blob/master/translation.yml). Replacing `yml` by `json` should just make things work, I've double checked with my team. As long as there is just the `en` translations, we're good to go 👍

nimzco:
I've updated the file 👍 
I'd merge this before adding anything into the README.

BPScott:
Nice, that looks more like it, thanks!

Comparing with https://github.com/Shopify/web/blob/master/translation.yml should we also add a `scheme: react` so placeholder interpolation is all good (per [docs](https://development.shopify.io/engineering/developing_at_Shopify/internationalization/translation_platform#Component_Scheme)

Other than that, this looks good to go.

nimzco:
Good call! 👍

nimzco:
Looking forward to seeing translations coming! 👀

PR #1306, opened 2019-04-11
[Code Quality] Consistent use of `text-breakword` mixin

beefchimi:
Was cruisin' through the codes and noticed some missed opportunities to use a `typography mixin`.

All I have done is consistently employ the `text-breakword` mixin.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -16,7 +16,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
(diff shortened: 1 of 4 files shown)


beefchimi:
`Percy` seems to be failing but I'm not spotting the problem? Anyone with laser eyes able to help me identify what Percy is taking an issue with?

chloerice:
> Percy seems to be failing but I'm not spotting the problem? Anyone with laser eyes able to help me identify what Percy is taking an issue with?

It was just a flaky test (the images in that example change depending on whether they were finished loading or not by the time the snapshot is done).

PR #1309, opened 2019-04-12
[Labelled] Contents can now wrap onto multiple lines

beefchimi:
In our app, we use forms in a `width` restricted space _(max width of around `260px`)_. When using a form that uses `Labelled` under the hood, and renders both a `label` + `action`, we can get some very ugly squished unwrapping content. Now that we are utilizing more translations in our app, this "ugly squishy" problem is more and more prevalent.

All this PR does is allow `Labelled` to use `flex-wrap: wrap;`, and then wraps the `action` in a `div` so I can apply `flex: 0 0 auto;`.

### Current (less than ideal) behaviour
less-than-ideal

### New wrapping behaviour
fixed

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -17,6 +17,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Enhancements
 
 - Added an onChange handler to `CheckableButton` ([#1326](https://github.com/Shopify/polaris-react/pull/1326))
+- `Labelled` now wraps its content, no longer causing a `label + action` to get unreasonably squished ([#1309](https://github.com/Shopify/polaris-react/pull/1309))
 - Updated `polaris-tokens` from `2.3.0` to `2.5.0` and converted all use of `duration` values ([#1268](https://github.com/Shopify/polaris-react/pull/1268))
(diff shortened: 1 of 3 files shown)


dleroux:
Just 🎩  this and I question if we do what to change this (or at least wait until the next version)? Looking at textfield the visual change happens pretty quickly. I can see it looking off if we have had a few textfield in a row. @ry5n thoughts?

chloerice:
Hey @beefchimi 👋 From a code standpoint it works as expected 👍Will tophat the affect on the other form fields tomorrow morning.

Can you give us some visual context for the restricted space you're working in? It really helps to see where the current design is failing 🙏

ry5n:
@chloerice let me know if you need anything

owain:
👋Just checking in on this since we'll be looking to start rollout next week but this could be a blocker.

dleroux:
This is more the scenario I was thinking, but you are right in the admin we are safe. If this isn't a concern then code looks good.

Screen Shot 2019-04-18 at 8 19 50 AM

beefchimi:
Thank you for the review + code snippet @chloerice 

The only scenario where I see this being an issue is when using `FormLayout` to have many side-by-side fields:

moon

But that is going to break / get wonky _anyways_ if we have long unwrapping lines:

sun

So in the end, I do not believe any use of `Labelled` has gotten worse 😃

PR #1311, opened 2019-04-12
add breakpoint mixin for frame with nav max width

patsissons:

We can currently use the `page-when-not-max-width` mixin to break pages as soon as they being to resize, but this is not applicable for layouts that use a frame with the nav section. For these cases we need to also take the nav width in account for the breaking width.

This PR adds a new breakpoint variable `$frame-with-nav-max-width` as well as a new mixin to break before that defined width `frame-with-nav-when-not-max-width()`. 

image
image
image

### 🎩 checklist

* ~[ ] Tested on [mobile](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting.md#cross-browser-testing)~
* ~[ ] Tested on [multiple browsers](https://help.shopify.com/en/manual/intro-to-shopify/shopify-admin/supported-browsers)~
* ~[ ] Tested for [accessibility](https://github.com/Shopify/polaris-react/blob/master/documentation/Accessibility%20testing.md)~

The actual breakpoint doesn't line up perfectly in storybook due to the extra margins/padding involved, but the intent is still pretty clear from the screenshots above. We want the break to occur as soon as the page content begins to resize and become responsive.

<details>
  <summary>
    Paste code into <code>playground/Playground.tsx</code> → <code>yarn dev</code> → open <a href="http://localhost:6006/?path=/story/playground-playground--playground">storybook</a> → 🎩
  </summary>

```tsx

import {Banner, Frame, Navigation, Page, TopBar} from '../src';
import * as styles from './Playground.scss';

export default class Playground extends React.PureComponent {
  render() {
    return (
      <Frame navigation={<Navigation location="/" />} topBar={<TopBar />}>
        <Page title="Playground">
          <Banner>
            <p className={styles.Content} />
          </Banner>
        </Page>
      </Frame>
    );
  }
}
```

Requires the following `Playground.scss` content
```scss
.Content::after {
  content: 'Frame Max Width Page';

  @include frame-with-nav-when-not-max-width {
    content: '*** Frame Condensed Page ***';
  }
}
```
</details>

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -24,6 +24,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Updated open styleguide pr to create multiple pull requests to update `polaris-react` across multiple repos ([#1069](https://github.com/Shopify/polaris-react/pull/1069))
 - Updated the pull request creation to retry when it fails ([#1069](https://github.com/Shopify/polaris-react/pull/1069))
 - Exported overlay and layer data attributes for use in consumer components ([#1266](https://github.com/Shopify/polaris-react/pull/1266))
+- Added new `frame-with-nav-max-width` variable and matching `frame-with-nav-when-not-max-width` mixin ([#1311](https://github.com/Shopify/polaris-react/pull/1311))
 
 - Updated `Resizer` to schedule `handleHeightCheck` to run in next animation frame ([#1301](https://github.com/Shopify/polaris-react/pull/1301))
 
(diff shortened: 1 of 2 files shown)


chloerice:
Hey Pat 👋Is this a WIP?

patsissons:
@chloerice not exactly. It could be ready to go (with a changelog update). But i wanted to get feedback from the Polaris team if this was something that belonged or not before pushing forward. IMO i think it belongs somewhere central, but i know that all these breakpoints are a bit of a pain point already. If you want me to get this prepped for merging i can do that!

chloerice:
Since it's a breakpoint that's dependent on the width of Polaris components, I don't see a reason not to add it here 👍.

Can you add some playground code that illustrates that this breaks when you expect it to?

patsissons:
can do!

patsissons:
🎩 in web confirmed 👍 I will queue this up first thing Monday.

PR #1327, opened 2019-04-16
Update developer dependencies

AndrewMusgrave:

Second attempt to update deps 😄 #1187


UNRELEASED.md
```diff
@@ -35,6 +35,7 @@ Upgraded Storybook to v5 ([#1140](https://github.com/Shopify/polaris-react/pull/
 
 ### Dependency upgrades
 
+- Updated most dependencies and dev dependencies ([#1187](https://github.com/Shopify/polaris-react/pull/1187))
```
AndrewMusgrave: I'll have to update this 🙈

BPScott:
Looks like this adds a lot of duplicates, can we remove some by running an `npx yarn-deduplicate yarn.lock`?

It looks like we'd also want to pull in the latest version of babel-plugin-shopify to shut up the coreJS complaining like what SK does in https://github.com/Shopify/sewing-kit/pull/1217

BPScott:
In my poking around with https://github.com/Shopify/polaris-react/pull/1344 I think I've learnt that we want to either: update @babel/core and babel-plugin-shopify, or leave them both where they were.

Using the latest will mean that the stand-alone build will be a bit bigger due to including corejs polyfills thanks to the babel-preset-env now using useBuiltIns: 'usage' (which changed in babel-plugin-shopify v18). This is due to our support for old ios. This doesn't affect polaris when used as part of sewing-kit because sk doesn't use the stand-alone build.

I don't mind which one you go for.

AndrewMusgrave:
Ending up keeping this scoped to `devDependencies`, and deduping causes broken builds. Build seems to be passing now with all checks ✅

PR #1337, opened 2019-04-18
remove-tag-button-transition

johanstromqvist:
Tag button hover state should be immediate.

Removes excessive hover state transition on tag button.

## 

### How to 🎩

<details>

```jsx

interface State {}

  render() {
    return (
      <Page title="Playground">
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -19,6 +19,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Made the `action` prop optional on `EmptyState` ([#1583](https://github.com/Shopify/polaris-react/pull/1583))
 - Prevented Firefox from showing an extra dotted border on focused buttons ([#1409](https://github.com/Shopify/polaris-react/pull/1409))
 - Added `resolveItemId` prop to `ResourceList` which is used in the new multiselect feature ([#1261](https://github.com/Shopify/polaris-react/pull/1261))
+- Removed transition on tag button hover state [#1337](https://github.com/Shopify/polaris-react/pull/1337)
 
 ### Bug fixes
 
--- a/src/components/Tag/Tag.scss
+++ b/src/components/Tag/Tag.scss
@@ -36,10 +36,6 @@ $icon-size: rem(16px);
   height: $height;
   width: $height;
   margin-left: spacing(extra-tight);
-  will-change: background, fill;
-  transition-property: background, fill;
-  transition-duration: duration();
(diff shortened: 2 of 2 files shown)


danrosenthal:
Hey @johanstromqvist, would you mind providing a bit of rationale for this change, and pinging a couple reviewers?

cc: @ry5n

ry5n:
I’d want to understand the rationale as well.

danrosenthal:
@johanstromqvist one more bump for context on this change

johanstromqvist:
Sorry about the delay guys – just got back from vacation today. Expect rationale within the next couple of days, or let me know if it's urgent and blocking you. I'll also provide more rationale in future PRs.

johanstromqvist:
Sorry about the delay.

The short story is that the touch target is behaving inconsistently and instead of fixing it I simply removed the behaviour. I asked around in the Polaris team but we couldn't find any rationale for the transition, and I couldn't see any myself. We could see tho that the component hadn't been updated in a long time.

**Why do I think transition is excessive**
The transition doesn't solve any problems or improve UX in my opinion, and I can see how adding transitions to hover states can add a sluggish delay which distances users from direct control.

**Why do I suggest hover to be immediate**
As a rule of thumb, I'd suggest that hover feedback should default to immediate, unless the hover triggers state change that would benefit from clarification. Before suggesting that as a general guideline, I'd like to try out on a couple of scenarios and run it by you guys. It's my current opinion tho.

**The inconsistency**
The current tag button has a 100-200ms transition on the background of the touch target and 0ms transition on the `x` icon in the centre, causing it to feel wonky and uncrafted.

UNRELEASED.md
```diff
@@ -24,6 +24,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Updated open styleguide pr to create multiple pull requests to update `polaris-react` across multiple repos ([#1069](https://github.com/Shopify/polaris-react/pull/1069))
 - Updated the pull request creation to retry when it fails ([#1069](https://github.com/Shopify/polaris-react/pull/1069))
 - Exported overlay and layer data attributes for use in consumer components ([#1266](https://github.com/Shopify/polaris-react/pull/1266))
+- Removed transition on tag button hover state [#1337](https://github.com/Shopify/polaris-react/pull/1337)
```
```suggestion
- Removed transition on tag button hover state ([#1337](https://github.com/Shopify/polaris-react/pull/1337))
```

PR #1344, opened 2019-04-24
Update to latest sewing-kit

BPScott:

Keeping up to date, includes new linting rules mostly, meaning when eslint-plugin-shopify jumps to use the new typescript parser there will be less effort to update it and get onto a newer typescript version.

Update a few dependencies that we want to keep in sync with sewing-kit

Use `sewing-kit optimize` to optimize svgs instead of maintaining our
own script which does the same thing.

---

An update to babel-plugin-shopify has changed the default value of "useBuiltIns". We now explicitly set "useBuildIns: 'entry'" in our babel.config.js to maintain the same behaviour as previously. Not doing this results in adding ~50kb extra polyfill stuff.

This doesn't affect polaris when used as part of sewing-kit because sk doesn't use the stand-alone build.

### To Review

This PR is split into multiple commits, one of which is all the automatic fixes from running `yarn run format` which contains the majority of the code changes (apart from the yarn.lock file). It might be easier to look at the commits one-by-one rather than scrolling through a load of automatic fixes.

### To Test

* Ensure tests / linting / type checks pass
* Ensure aria labels still appear as you'd expect in the components where I had to add translations


.eslintrc
```diff
@@ -56,20 +59,24 @@
     {
       "files": ["**/*.test.{ts,tsx}"],
       "rules": {
+        "jest/no-truthy-falsy": "off",
+        "shopify/jsx-no-hardcoded-content": "off",
```
BPScott: We want to check for hardcoded content to make sure we're using translated strings in our source, but we don't care about that in our tests and examples.

.eslintrc
```diff
@@ -28,13 +27,16 @@
...
+    "babel/no-unused-expressions": "off",
     "import/named": "off",
     "import/no-named-as-default": "off",
     "react/button-has-type": "off",
     "react/no-array-index-key": "off",
     "react/jsx-fragments": ["error", "element"],
     "shopify/jsx-no-complex-expressions": "off",
+    "shopify/jsx-prefer-fragment-wrappers": "off",
     "shopify/no-ancestor-directory-import": "error",
+    "shopify/react-prefer-private-members": "off",
```
BPScott: Sometimes we do actually want pubic functions

.eslintrc
```diff
@@ -28,13 +27,16 @@
...
       }
     ],
+    "babel/no-unused-expressions": "off",
     "import/named": "off",
     "import/no-named-as-default": "off",
     "react/button-has-type": "off",
     "react/no-array-index-key": "off",
     "react/jsx-fragments": ["error", "element"],
     "shopify/jsx-no-complex-expressions": "off",
+    "shopify/jsx-prefer-fragment-wrappers": "off",
```
BPScott: Sometimes we want empty wrapping divs

.eslintrc
```diff
@@ -28,13 +27,16 @@
         "allowBlockStart": false
       }
     ],
+    "babel/no-unused-expressions": "off",
```
BPScott: This doesn't seem to work in the way that I'd expect

.eslintrc
```diff
@@ -56,20 +59,24 @@
     {
       "files": ["**/*.test.{ts,tsx}"],
       "rules": {
+        "jest/no-truthy-falsy": "off",
```
BPScott: we use isTruthy() / isFalsy() A LOT and I don't want to go and refactor everything

src/components/Frame/components/Toast/tests/Toast.test.tsx
```diff
@@ -113,18 +113,22 @@ describe('<Toast />', () => {
 
     it('is called when the escape key is pressed', () => {
       const listenerMap: HandlerMap = {};
-      document.addEventListener = jest.fn((event, cb) => {
-        listenerMap[event] = cb;
-      });
+      jest
```
AndrewMusgrave: Should we clear this in an afterEach?

src/components/Frame/components/Toast/tests/Toast.test.tsx
```diff
@@ -113,18 +113,22 @@ describe('<Toast />', () => {
...
 
       const spy = jest.fn();
       mountWithAppProvider(<Toast content="Image uploaded" onDismiss={spy} />);
 
       listenerMap.keyup({keyCode: Key.Escape});
 
-      document.removeEventListener = jest.fn((event) => {
-        listenerMap[event] = noop;
-      });
+      jest
```
AndrewMusgrave: ^

src/components/KeypressListener/tests/KeypressListener.test.tsx
```diff
@@ -14,11 +14,11 @@ const listenerMap: HandlerMap = {};
 
 describe('<KeypressListener />', () => {
   beforeEach(() => {
-    document.addEventListener = jest.fn((event, cb) => {
+    jest.spyOn(document, 'addEventListener').mockImplementation((event, cb) => {
```
AndrewMusgrave: We'll want to change the afterEach to a clear rather than reassign the original value

src/components/Page/tests/Page.test.tsx
```diff
@@ -170,7 +170,7 @@ describe('<Page />', () => {
       };
 
       const {restore: restoreTitleBarCreateMock} = mockTitleBarCreate();
-      AppBridgeButton.create = jest.fn().mockReturnValue(buttonMock);
+      jest.spyOn(AppBridgeButton, 'create').mockReturnValue(buttonMock);
```
AndrewMusgrave: ^

src/components/TextField/tests/TextField.test.tsx
```diff
@@ -694,11 +694,13 @@ describe('<TextField />', () => {
         const origialAddEventListener = document.addEventListener;
 
         beforeAll(() => {
-          document.addEventListener = jest.fn(
-            (eventType: string, callback: EventCallback) => {
-              documentEvent[eventType] = callback;
-            },
-          );
+          jest
```
AndrewMusgrave: ^

src/components/TopBar/components/UserMenu/context/Modifier.tsx
```diff
@@ -22,7 +22,7 @@ class Modifier extends React.Component<ComposedProps, {}> {
     return null;
   }
 
-  state = {};
+  state: any = {};
```
AndrewMusgrave: We can probably 🔥 this and change line 14 to `never`

AndrewMusgrave:
I requested translations and since the translations have already been released feel free to mark as `async` when you want to 🚢

danrosenthal:
What's the status on this PR @BPScott?

src/components/Page/tests/Page.test.tsx
BPScott: It looks like define restoration - see line 178

src/components/TextField/tests/TextField.test.tsx
BPScott: Have changed the reassigning to use mockRestore().

src/components/TopBar/components/UserMenu/context/Modifier.tsx
BPScott: I've got no idea what this does. So I elected for not changing the code and just making typescript happy. It looks like this component still has state it's just defined in the passed setMobileUserMenuProps function.

Thinking a little deeper, matching this type with the second arg of `React.Component<ComposedProps, {}>` is a better plan. I've changed it to `state: {}`.

src/components/Frame/components/Toast/tests/Toast.test.tsx
BPScott: All these changes to use jest.spyOn are automatic fixes rather than manual ones. The end result is the same as the current behaviour. the symmetry of using jest.spyOn then mockRestore is nice but if i worked before it'll work after. I'd like to not get too bogged down in tidying up our tests as they work in the same way as before :)

I've updated the cases that you raised and found an extra one in Tabs.

BPScott:
I've rebased this atop master, fixed additional issues (autofixing jest's toEqual to toStrictEqual) and addressed PR feedback. I've also updated the babel config to retain the old useBuiltIns behaviour, which means the bundle size is unaffected (previously it grew by a substantial amount).

I've responded to your comments and this is now read for re-review @AndrewMusgrave

.eslintrc
AndrewMusgrave: I wonder if we should enable this and add an ignore whenever public is used, which is only in `Banner` i think?

src/components/Frame/components/Toast/tests/Toast.test.tsx
```diff
@@ -111,22 +111,40 @@ describe('<Toast />', () => {
...
+        jest
+          .spyOn(document, 'removeEventListener')
+          .mockImplementation((event) => {
+            listenerMap[event] = noop;
+          });
+      });
 
-      listenerMap.keyup({keyCode: Key.Escape});
+      afterEach(() => {
+        (document.addEventListener as jest.Mock).mockRestore();
```
AndrewMusgrave: I'm ok with this, however an alternative could have been
```tsx
let removeEventListener: jest.SpyInstace;
...
removeEventListener = jest.spyOn(...)
...
removeEventListener.mockRestore
```

.eslintrc
BPScott: There's 36 violations currently, 1/3 of which are false positives warning about examples written in JS when you can't even use private members (bug raised: https://github.com/Shopify/eslint-plugin-shopify/issues/259), and many of the others seem related to legacy context stuff - subscribe/unsubscribe stuff. I don't know enough about the code they touch to know what should be private or not.

I'd totally be open to reevaluating this in the future, but I don't want to touch them now :)

PR #1349, opened 2019-04-24
App Bridge: Update to v1.3.0

vividviolet:

Fixes https://github.com/Shopify/polaris-react/issues/998
Fixes https://github.com/Shopify/polaris-react/issues/1028

Allows apps to subscribe to redirect actions and handle its own redirects. This allows single page apps using its own routing (ex. React Router) to navigate without reloading the app.

**Before (iframe is reloaded)**

polaris-redirect-before

**After (iframe is not reloaded)**

polaris-redirect-after

#### Using a custom handler for app redirects

The following is an example of a component utilizing the App Bridge context from Polaris to subscribe to redirects and add its own handler.

```js
import { RouterProps } from 'react-router';

export default class Section extends React.Component {
  static contextTypes = {
    polaris: PropTypes.object,
  };

  cleanup = () => {};

  componentDidMount() {
    const {
      polaris: { appBridge },
    } = this.context;
    this.cleanup = appBridge.subscribe(Redirect.ActionType.APP, ({ path }) => {
      // Call your own method here with the relative app path
    });
  }

  componentWillUnmount() {
   // Clean up subscriptions when component unmounts
    this.cleanup();
  }
}

```

Update to use App Bridge v1.3.0

### How to 🎩

1. Clone https://github.com/Shopify/app-bridge-test-apps and do `cd app-bridge-polaris-react-router`
2. Run `dev up`
3. Navigate to the directory containing `polaris-react` and do `yarn build-consumer app-bridge-test-apps/app-bridge-polaris-react-router`.
4. Navigate back to the test app directory and follow the [instructions](https://github.com/Shopify/app-bridge/blob/master/packages/app-bridge-playground/README.md#install-playground-to-local-shop) to install the app on either local Shopify or a production store
5. Open the app and verify that clicking on the secondary navigation redirects to the correct page without reloading the iframe
6. Verify that clicking on the app name or app icon when NOT on the index page navigates back to the index page without reloading the iframe

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -27,6 +27,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - Bump react-utilites to remove a transitive dependency on core-js. ([#1343](https://github.com/Shopify/polaris-react/pull/1343))
 
+- Updated App Bridge to version 1.3.3 ([#1349](https://github.com/Shopify/polaris-react/pull/1349))
+
 ### Code quality
 
 ### Deprecations
--- a/package.json
+++ b/package.json
@@ -177,7 +177,7 @@
   ],
   "dependencies": {
     "@babel/runtime": "^7.1.6",
-    "@shopify/app-bridge": "^1.1.2",
+    "@shopify/app-bridge": "^1.3.0",
     "@shopify/images": "^1.1.0",
     "@shopify/javascript-utilities": "^2.2.1",
     "@shopify/polaris-icons": "^3.3.0",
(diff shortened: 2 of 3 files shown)


UNRELEASED.md
```diff
@@ -34,6 +34,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - Bump react-utilites to remove a transitive dependency on core-js. ([#1343](https://github.com/Shopify/polaris-react/pull/1343))
 
+- Updated App Bridge to version 1.3.3 ([#1349](https://github.com/Shopify/polaris-react/pull/1349))
```
tmlayton: Typo here, should be `1.3.0`

vividviolet: My bad! I'll put up another pr to fix it

PR #1358, opened 2019-04-25
Add translations: default

translation-platform:
This PR adds translations to the project. It will append multiple commits. Feel free to merge anytime.

### What's next?
- [ ] Check translations for malicious HTML.
- [ ] Approve the Pull Request using the GitHub Review system.
- [ ] :ship: Ship it!
- [ ] :fire: the branch.

Questions about the format of this Pull Request? Hop in the [#intl-services](https://shopify.slack.com/messages/C7TJQLVC7) Slack channel.

Questions about the translations in this Pull Request? Hop in the [#translation](https://shopify.slack.com/messages/C3XPBC62U) Slack channel.

We're happy to help!

@Shopify/polaris

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -16,6 +16,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Added translations for all supported locales ([#1358](https://github.com/Shopify/polaris-react/pull/1358))
 - Improved the performance of `ResourceList` ([#1313](https://github.com/Shopify/polaris-react/pull/1313))
 - Adds `withinContentContainer` context to `Navigation` ([#1393](https://github.com/Shopify/polaris-react/pull/1393))
 
--- a/src/locales/de.json
+++ b/src/locales/de.json
@@ -1,0 +1,200 @@
+{
+  "Polaris": {
+    "Avatar": {
+      "label": "Avatar",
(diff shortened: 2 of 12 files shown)


danrosenthal:
@lemonmade, do you have any objections to us merging this? @tmlayton mentioned you were opposed to it, as it would add bundle size. Is this tradeoff worth it so that we can enable third parties to easily localize our components?

lemonmade:
How are these locales actually pulled in to Polaris, or would they only be pulled in by consumers?

danrosenthal:
> How are these locales actually pulled in to Polaris, or would they only be pulled in by consumers?

AFAIK, they would need to be pulled in by consumers. At some point I could see us being smarter about it, and determining which translation to use based on a locale passed to the `AppProvider`. We would then import them from the [locales index](https://github.com/Shopify/polaris-react/blob/master/src/locales/index.ts#L3).

lemonmade:
I would recommend having some docs about pulling these in, it's a bit concerning if it's not clear how to bring them in. That being said, if it is not actually imported directly anywhere in Polaris, it should be "safe" to release from a bundle size perspective.

AndrewMusgrave:
As Dan mentioned above the translations will be pulled in by consumers and not directly imported anywhere in Polaris. Something we may want to consider is making i18n a required prop on AppProvider and not include our English dictionary by default.

danrosenthal:
I've documented the issue in #1418, and am going to proceed with merging this on 🍏

PR #1362, opened 2019-04-25
Add checklist and instructions updating and tophatting documentation changes

dpersing:

Currently we don't consistently check to see if documentation needs to change when we make updates to a component.

Adds PR checklist items for:

- Updating the component `README.md` (if necessary)
- Tophatting that documentation change in the style guide (if it is visible there)

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -62,7 +62,7 @@ export default class Playground extends React.Component<never, State> {
 * [ ] Tested on [multiple browsers](https://help.shopify.com/en/manual/intro-to-shopify/shopify-admin/supported-browsers)
 * [ ] Tested for [accessibility](https://github.com/Shopify/polaris-react/blob/master/documentation/Accessibility%20testing.md)
 * [ ] Updated the component's `README.md` with documentation changes
-* [ ] Tophatted documentation changes in the style guide
+* [ ] [Tophatted documentation](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting%20documentation.md) changes in the style guide
 
 <!--
   When tophatting, please check the UNRELEASED entry for consistency, as per the [guidelines](https://github.com/Shopify/polaris-react/blob/master/documentation/Versioning%20and%20changelog.md).
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -61,6 +61,8 @@ export default class Playground extends React.Component<never, State> {
 * [ ] Tested on [mobile](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting.md#cross-browser-testing)
 * [ ] Tested on [multiple browsers](https://help.shopify.com/en/manual/intro-to-shopify/shopify-admin/supported-browsers)
 * [ ] Tested for [accessibility](https://github.com/Shopify/polaris-react/blob/master/documentation/Accessibility%20testing.md)
+* [ ] Updated `README.md` with documentation changes
+* [ ] Tophatted documentation changes in the style guide
 
 <!--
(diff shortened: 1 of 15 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -61,6 +61,8 @@ export default class Playground extends React.Component<never, State> {
 * [ ] Tested on [mobile](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting.md#cross-browser-testing)
 * [ ] Tested on [multiple browsers](https://help.shopify.com/en/manual/intro-to-shopify/shopify-admin/supported-browsers)
 * [ ] Tested for [accessibility](https://github.com/Shopify/polaris-react/blob/master/documentation/Accessibility%20testing.md)
+* [ ] Updated `README.md` with documentation changes
```
amrocha: I think this might be confusing since there's multiple `README.md` files in the repo, can you use language that's bit more specific like `if you modified a component, then update that component's README file` or something like that

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -61,6 +61,8 @@ export default class Playground extends React.Component<never, State> {
...
+* [ ] Tophatted documentation changes in the style guide
```
amrocha: This would be nice to include too but there's no easy way of doing this check right now. What if we told people to check in a MD previewer, do you think that's enough? We could also write a quick doc for how to tophat changes in the styleguide and link it here too

dpersing: > We could also write a quick doc for how to tophat changes in the styleguide and link it here too

Can do. I think testing the look in the style guide is important, since folks in lots of different roles need to do it.

amrocha: Sounds good, it should just be a matter of documenting the `build-consumer` command and telling people to set up `polaris-styleguide` as well 👍

.github/PULL_REQUEST_TEMPLATE.md
dpersing: Updated to specify the component-level `README.md`.

.github/PULL_REQUEST_TEMPLATE.md
dpersing: Added a new doc, copying over some info from the `Tophatting.md` doc and adding instructions.

documentation/Tophatting documentation.md
```diff
@@ -0,0 +1,25 @@
...
+
+- When you're adding content to a component's `README.md` (such as `/src/components/Button/README.md`) to change or add best practices,accessibility documentation, or other information.
+
+## What should you 🎩?
+
+Check the location, appearance, and readability of the content you add to the style guide.
+
+## How to 🎩documentation
+
+1.Check out the `master` branch from `[polaris-styleguide](https://github.com/Shopify/polaris-styleguide)`
```
kaelig: You'll want a space between the `.` and the first word:

```suggestion
1. Check out the `master` branch from `[polaris-styleguide](https://github.com/Shopify/polaris-styleguide)`
```

or the list will render as:

Screenshot 2019-04-26 12 04 23

documentation/Tophatting documentation.md
```diff
@@ -0,0 +1,25 @@
...
+
+When components are updated in ways that impact how people use them, their `README.md` documentation should be updated to reflect this.
+
+When you update the `README.md` of a component, that content can change the documentation that appears in the [Polaris style guide](https://polaris.shopify.com/).
+
+Manually testing documentation (we call it “[tophatting](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting.md#history-of-the-phrase)”, or 🎩 for short) adds that extra degree of certainty that what you’re shipping looks like what you expect.
+
+## When should you 🎩 documentation?
+
+- When you're adding content to a component's `README.md` (such as `/src/components/Button/README.md`) to change or add best practices,accessibility documentation, or other information.
```
ry5n:
```suggestion
- When you're adding content to a component's `README.md` (such as `/src/components/Button/README.md`) to change or add best practices, accessibility documentation, or other information.
```

PR #1388, opened 2019-04-30
Updating to typescript 3.2

BPScott:

Consumers can't move past typescript 3.1.6 thanks to us erroring in those versions

Updating to typescript  3.2.4

https://github.com/Shopify/quilt/pull/602 shows how quilt and react-compose were able to compile in TS 3.2, hopefully similar fixes should do the job in our HoCs but my TS-fu isn't up to the task

Thanks to conflicting version dependencies with our linting tools we can't go all the way up to 3.4:

The last release of typescript-eslint-parser is 22.0.0 which only
supports up to TS 3.2.x. We're stuck on this because
eslint-plugin-shopify still uses typescript-eslint-parser and has not
yet moved onto its replacement package @typescript-eslint/parser - which
supports TS 3.2.x and above.

Somehow among all this we now get false positives for violations of the
import/named rule. It's fine to disable that as TS will shout at us if
we reference an unknown import anyway.

Let's stick with 3.2.x for the moment, and then when eslint-plugin-shopify updates to use  typescript-eslint-parser we can update beyond that

### How to 🎩

`yarn run sk type-check`

### DIFF
--- a/package.json
+++ b/package.json
@@ -155,7 +155,7 @@
     "shelljs": "^0.8.3",
     "shx": "^0.3.2",
     "svgo": "^1.2.2",
-    "typescript": "~3.1.6",
+    "typescript": "~3.2.4",
     "yargs": "^13.2.2"
   },
   "peerDependencies": {
@@ -163,7 +163,7 @@
     "react-dom": "^16.3.1"
   },
   "resolutions": {
-    "typescript-eslint-parser": "20.0.0"
+    "typescript-eslint-parser": "22.0.0"
   },
   "files": [
     "esnext",
@@ -182,9 +182,9 @@
     "@shopify/javascript-utilities": "^2.2.1",
     "@shopify/polaris-icons": "^3.3.0",
     "@shopify/polaris-tokens": "^2.5.0",
-    "@shopify/react-compose": "^0.1.6",
+    "@shopify/react-compose": "^1.0.0",
     "@shopify/react-utilities": "^2.0.4",
-    "@types/prop-types": "^15.5.5",
+    "@shopify/useful-types": "^1.2.4",
     "@types/react": "^16.4.7",
     "@types/react-dom": "^16.0.6",
     "@types/react-transition-group": "^2.0.7",
--- a/src/components/ActionList/ActionList.tsx
+++ b/src/components/ActionList/ActionList.tsx
@@ -30,7 +30,9 @@ export default function ActionList({
   }
 
   const hasMultipleSections = finalSections.length > 1;
-  const Element: string = hasMultipleSections ? 'ul' : 'div';
+  // Type asserting to any is required for TS3.2 but can be removed when we update to 3.3
+  // see https://github.com/Microsoft/TypeScript/issues/28768
+  const Element: any = hasMultipleSections ? 'ul' : 'div';
   const sectionMarkup = finalSections.map((section, index) => {
     return (
       <Section
--- a/src/components/ChoiceList/ChoiceList.tsx
+++ b/src/components/ChoiceList/ChoiceList.tsx
@@ -59,9 +59,9 @@ function ChoiceList({
   error,
   name = getUniqueID(),
 }: CombinedProps) {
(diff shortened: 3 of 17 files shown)


BPScott:
This is rebased atop latest master and is now ready for review

package.json
```diff
@@ -182,15 +182,15 @@
...
+    "@shopify/react-compose": "^1.0.0",
     "@shopify/react-utilities": "^2.0.4",
-    "@types/prop-types": "^15.5.5",
+    "@shopify/useful-types": "^1.2.4",
     "@types/react": "^16.4.7",
     "@types/react-dom": "^16.0.6",
     "@types/react-transition-group": "^2.0.7",
     "hoist-non-react-statics": "^2.5.0",
     "lodash": "^4.17.4",
-    "prop-types": "^15.6.1",
```
AndrewMusgrave: With the removal of `prop-types.d.ts` I don't think we'll have to update this

src/components/ActionList/ActionList.tsx
```diff
@@ -30,7 +30,7 @@ export default function ActionList({
   }
 
   const hasMultipleSections = finalSections.length > 1;
-  const Element: string = hasMultipleSections ? 'ul' : 'div';
+  const Element: any = hasMultipleSections ? 'ul' : 'div';
```
AndrewMusgrave: I'm surprised we can't type this as `keyof HTMLElementTagNameMap` or `Element` or something similar. Do you think this is because of something in our tsconfig or a version of a types package?

src/components/ChoiceList/ChoiceList.tsx
```diff
@@ -59,9 +59,7 @@ function ChoiceList({
   error,
   name = getUniqueID(),
 }: CombinedProps) {
-  const ControlComponent: typeof Checkbox | typeof RadioButton = allowMultiple
-    ? Checkbox
-    : RadioButton;
+  const ControlComponent: any = allowMultiple ? Checkbox : RadioButton;
```
AndrewMusgrave: Something like this should work

```tsx
type CheckboxComponent = React.ComponentClass<Props, State> & typeof Checkbox;
type RadioComponent = React.SFC<Props> & typeof RadioButton;
const ControlComponent: CheckboxComponent | RadioComponent = allowMultiple ? Checkbox : RadioButton;
```

src/components/ActionList/ActionList.tsx
BPScott: I tried those and had no luck.

When doing `<ul></ul>` it says the type is of `JSX.Element` so I tried doing a `keyof JSX.Element` (which resolves to an array of strings containing ul and div) but I still got an error.

It looks like the root cause of this is detailed at https://github.com/Microsoft/TypeScript/issues/28768. It seems this was fixed in typescript 3.3 (so you don't need any casting at all), but we can't move onto that till our linting tooling is ready :(

BPScott: We could change the return line to the following, but I think that fix is worse than changing the typing:

```
return React.createElement(
    Element,
    {className: styles.ActionList},
    ...sectionMarkup,
  );
```

src/components/ChoiceList/ChoiceList.tsx
BPScott: No dice, for the same reasons as above. https://github.com/Microsoft/TypeScript/issues/28768 breaks us, but it's no longer an error in newer versions of TS.

src/components/ActionList/ActionList.tsx
AndrewMusgrave: If it's fixed in a later version of typescript I'm fine with using `any` 😄  We should just leave a note something so we don't forget about it.

src/components/ChoiceList/ChoiceList.tsx
AndrewMusgrave: I tried it locally yesterday and it seemed to work but I may have had a different version of TS locally (3.4 is my guess). Same comment as above for this 👌

PR #1392, opened 2019-05-01
[v4] Update React - preparations

AndrewMusgrave:

Unfortunately we cannot continue updating our react components to prepare for react 16.8.* until we update to 16.8.* 😅 [Enzyme doesn't support the new context](https://github.com/airbnb/enzyme/issues/1908) at our current version, a bump in the version require a bump in other packages require a bump in react 

Updating react / enzyme


config/typescript/prop-types.d.ts
```diff
@@ -1,5 +0,0 @@
-// https://github.com/Microsoft/TypeScript/issues/24599
-declare module 'prop-types' {
```
AndrewMusgrave: This was causing namespace issues

src/components/AppProvider/tests/AppProvider.test.tsx
```diff
@@ -1,57 +1,9 @@
 import * as React from 'react';
-import TestUtils from 'react-dom/test-utils';
 import {mount} from 'enzyme';
-import {createThemeContext} from '../../ThemeProvider';
-import {StickyManager, createAppProviderContext} from '../utilities';
 import {polarisAppProviderContextTypes} from '../types';
 import AppProvider from '../AppProvider';
 
 describe('<AppProvider />', () => {
-  it('passes i18n and withComponent properties to context', () => {
```
AndrewMusgrave: This test will be updated and improved when we switch `AppProvider` to the new context API

src/components/Modal/tests/Modal.test.tsx
```diff
@@ -312,18 +312,20 @@ describe('<Modal>', () => {
   });
 
   describe('with app bridge', () => {
+    let AppBridgeModalCreate: jest.SpyInstance;
```
AndrewMusgrave: The next few files are using jest API to spy on functions rather than overwrite the function with a spy

AndrewMusgrave:
@danrosenthal I was planning on merging this into `v4` since it has a ✅and every feature merged into this was ✅. Let me know if I should wait for a review from yourself/Tim 😄

danrosenthal:
> @danrosenthal I was planning on merging this into v4 since it has a ✅and every feature merged into this was ✅. Let me know if I should wait for a review from yourself/Tim 😄

No need, was just adding us there so we could follow along.

PR #1394, opened 2019-05-01
[TextField] `prefix/suffix` props: Allow clickable actions but retain unselectable text

beefchimi:
2 years ago, @ry5n added the following code: https://github.com/Shopify/polaris-react/commit/cbbe3df9a95e7934786174eb183c022658e4dd04

This was specifically to:
> Fix left and right ends of `TextField` not responding to clicks

So the intention was definitely to prevent clicks on any `prefix` or `suffix` and have those clicks pass through, allowing the `focus` state to be triggered for the component.

2 years later I find myself attempting to achieve the following design:

online-store

In the screenshot above, you will see a `<TextField />` that uses:
- An `icon` for a `prefix`
- A "Visit" `<Link />` as a `suffix`
- And a  "X" `<Button />` for `connectedRight`

It turns out, my "Visit" `Link` was not interactable... and after some digging, I arrived on the `pointer-events: none;` code that @ry5n had added.

### Solution Proposed

I am proposing that we swap `pointer-events: none;` for `user-select: none;`. This allows me to have an interactable `Button` or `Link` as `prefix/suffix` while still allowing unselectable and _click-throughable_ `prefix/suffix` text.

Screen Shot 2019-05-01 at 3 16 30 PM

### Playground

<details>
<summary>Please 🎩 using the following Playground code:</summary>

```jsx
/* eslint-disable shopify/react-no-multiple-render-methods */

import {CancelSmallMinor} from '@shopify/polaris-icons';
import {Button, Link, FormLayout, TextField, Page} from '../src';

interface State {
  exampleOneValue?: string;
  exampleTwoValue?: string;
  exampleThreeValue?: string;
  exampleFourValue?: string;
}

  state: State = {
    exampleOneValue: '2.00',
    exampleTwoValue: 'hello there',
    exampleThreeValue: undefined,
    exampleFourValue: undefined,
  };

  handleExampleOneChange = (value) => {
    this.setState({exampleOneValue: value});
  };

  handleExampleTwoChange = (value) => {
    this.setState({exampleTwoValue: value});
  };

  handleExampleThreeChange = (value) => {
    this.setState({exampleThreeValue: value});
  };

  handleExampleFourChange = (value) => {
    this.setState({exampleFourValue: value});
  };

  render() {
    return (
      <Page title="Playground">
        <FormLayout>
          {this.renderExampleOne()}
          {this.renderExampleTwo()}
          {this.renderExampleThree()}
          {this.renderExampleFour()}
        </FormLayout>
      </Page>
    );
  }

  renderExampleOne() {
    const {exampleOneValue} = this.state;

    return (
      <TextField
        label="Example 1"
        type="number"
        value={exampleOneValue}
        placeholder="Placeholder text"
        prefix="$$$"
        onChange={this.handleExampleOneChange}
      />
    );
  }

  renderExampleTwo() {
    const {exampleTwoValue} = this.state;

    return (
      <TextField
        label="Example 2"
        type="text"
        value={exampleTwoValue}
        placeholder="Placeholder text"
        suffix={
          <Button onClick={this.handleOnClick} plain external>
            Log to console
          </Button>
        }
        onChange={this.handleExampleTwoChange}
      />
    );
  }

  renderExampleThree() {
    const {exampleThreeValue} = this.state;

    return (
      <TextField
        label="Example 3"
        type="text"
        value={exampleThreeValue}
        placeholder="Placeholder text"
        prefix={
          <Link url="https://help.shopify.com/manual" external>
            Clickable
          </Link>
        }
        suffix="Not selectable"
        onChange={this.handleExampleThreeChange}
      />
    );
  }

  renderExampleFour() {
    const {exampleFourValue} = this.state;
    const connectedActions = (
      <div className="MyConnectedActions">
        <Button url="https://shopify.ca" external>
          Visit
        </Button>
        <Button icon={CancelSmallMinor} disabled />
      </div>
    );

    return (
      <TextField
        label="Example 4 (using connectedRight instead)"
        type="text"
        value={exampleFourValue}
        placeholder="Placeholder text"
        prefix="Prefix"
        suffix="Suffix"
        connectedRight={connectedActions}
        onChange={this.handleExampleFourChange}
      />
    );
  }

  private handleOnClick(event: any) {
    // Try commenting this out, and you will see the `focus` state/className remain on the input
    event.stopPropagation();
    console.log('Just stopped propagation');
  }
}
```

</details>

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -10,8 +10,6 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - `Page > Header` now must pass `content` to the `actions` within `secondaryActions` and `actionGroups` ([#1653](https://github.com/Shopify/polaris-react/pull/1653))
 
-### Deprecations
-
 ### New components
 
 - `ActionMenu`: Use for display of actions and action groups within the context of a header ([#1653](https://github.com/Shopify/polaris-react/pull/1653))
@@ -19,6 +17,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Enhancements
 
 - Add `selectable` prop to `ResourceList` component ([#1614](https://github.com/Shopify/polaris-react/pull/1614))
+- Allow `Link` and `Button` interactions when rendered as `prefix/suffix` within `<TextField />` ([#1394](https://github.com/Shopify/polaris-react/pull/1394))
 - `ActionList` can now pass a unique `accessibilityLabel` to each `Item` ([#1653](https://github.com/Shopify/polaris-react/pull/1653))
 - Greatly reduced complexity of `Page > Header` ([#1653](https://github.com/Shopify/polaris-react/pull/1653))
 - Long `Page > Header` breadcrumb labels will now truncate instead of breaking layout ([#1653](https://github.com/Shopify/polaris-react/pull/1653))
--- a/src/components/TextField/tests/TextField.test.tsx
+++ b/src/components/TextField/tests/TextField.test.tsx
@@ -329,6 +329,49 @@ describe('<TextField />', () => {
       expect(textField.find(`#${labels[1]}`).text()).toBe('$');
       expect(textField.find(`#${labels[2]}`).text()).toBe('.00');
     });
+
+    it('does not set focus `onClick` for the <input /> if the `target` is the `prefix`', () => {
+      const onClickSpy = jest.fn();
+      const mockButtonId = 'MockPrefix';
+      const mockPrefixButton = (
+        <button id={mockButtonId} onClick={onClickSpy} />
+      );
+      const textField = mountWithAppProvider(
+        <TextField
+          label="TextField"
+          prefix={mockPrefixButton}
+          onChange={noop}
+        />,
+      );
+
+      textField.find(`#${mockButtonId}`).simulate('click');
+
+      expect(onClickSpy).toHaveBeenCalled();
+      expect(textField.getDOMNode().querySelector('input')).not.toBe(
+        document.activeElement,
+      );
+    });
+
+    it('does not set focus `onFocus` for the <input /> if the `target` is the `prefix`', () => {
+      const mockButtonId = 'MockPrefix';
+      const mockPrefixButton = <button id={mockButtonId} onClick={noop} />;
+      const textField = mountWithAppProvider(
+        <TextField
+          label="TextField"
+          prefix={mockPrefixButton}
+          onChange={noop}
+        />,
+      );
+
+      textField.find(`#${mockButtonId}`).simulate('focus');
+
+      const connectedInteriorWrapper = textField
+        .find(Connected)
+        .find('div')
+        .first();
+
+      expect(connectedInteriorWrapper.hasClass('focus')).toBe(false);
+    });
   });
 
   describe('suffix', () => {
@@ -343,6 +386,49 @@ describe('<TextField />', () => {
       expect(labels).toHaveLength(2);
       expect(textField.find(`#${labels[0]}`).text()).toBe('TextField');
       expect(textField.find(`#${labels[1]}`).text()).toBe('kg');
+    });
+
+    it('does not set focus `onClick` for the <input /> if the `target` is the `suffix`', () => {
+      const onClickSpy = jest.fn();
+      const mockButtonId = 'MockSuffix';
+      const mockSuffixButton = (
+        <button id={mockButtonId} onClick={onClickSpy} />
+      );
+      const textField = mountWithAppProvider(
+        <TextField
+          label="TextField"
+          suffix={mockSuffixButton}
+          onChange={noop}
+        />,
+      );
+
+      textField.find(`#${mockButtonId}`).simulate('click');
+
+      expect(onClickSpy).toHaveBeenCalled();
+      expect(textField.getDOMNode().querySelector('input')).not.toBe(
+        document.activeElement,
+      );
+    });
+
+    it('does not set focus `onFocus` for the <input /> if the `target` is the `suffix`', () => {
+      const mockButtonId = 'MockSuffix';
+      const mockSuffixButton = <button id={mockButtonId} onClick={noop} />;
+      const textField = mountWithAppProvider(
+        <TextField
+          label="TextField"
+          suffix={mockSuffixButton}
+          onChange={noop}
+        />,
+      );
+
+      textField.find(`#${mockButtonId}`).simulate('focus');
+
+      const connectedInteriorWrapper = textField
+        .find(Connected)
+        .find('div')
+        .first();
+
+      expect(connectedInteriorWrapper.hasClass('focus')).toBe(false);
     });
   });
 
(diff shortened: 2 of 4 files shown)


beefchimi:
cc/ @owain and @utkarshsaxena93 since you have both been able to see this pattern in action

utkarshsaxena93:
This is great! It is possible that we don't want this pattern Polaris wide and that the use case for the theme editor is a special case. I'll let the Polaris team decide if this is an acceptable pattern for all consumers of Polaris (which includes 3rd party apps).

owain:
I don't love this pattern, tbh. I'm not sure it should be encouraged. Especially, worried about a11y e.g. Without an icon will people that struggle with colour know it's a link?

With that said I'm sure the Polaris UX team can judge better than me.

beefchimi:
Some good talking points here.

> Without an icon...

Nothing stopping anyone from using an icon. In my screenshot above, **Example 3** uses an external icon.

> ...will people that struggle with colour know it's a link?

If we are concerned with the color of the link, then that same criticism applies to all `<Link />` styling. That being said, I'd be curious what Polaris' thoughts on a text-underline are.

Worth mentioning that Dan Rosenthal pointed me to a `polaris-rails` style that is almost visually equivalent to what I'm doing:
https://polaris-rails.shopify.com/components/password

An alternative approach to what I am attempting would be **Example 4** - as you can see, the layout is not quite correct, so some time would need to be spent to get connected button groups working _(unless I've simply done it wrong...)_. Worth considering.

ry5n:
@beefchimi I’m planning to provide review on Tuesday. If you need this sooner let me know.

dpersing:
>the component’s existing behavior causes focus to move to the field when clicking anywhere inside the textfield

Do you mean that clicking inside the field moves the keyboard focus indicator? That's standard. The prefix and suffix would be separate elements, so we shouldn't have to do any additional focus management. Unless I'm missing something? @ry5n

ry5n:
I think we’re doing something to move focus to the field (or at least draw a focus ring) when the prefix and suffix are clicked. That was fine when those were non-interactive, but if they are buttons, they should retain focus if they are clicked. That seems to not be happening.

beefchimi:
@ry5n good catch!

So whats happening is the prefix/suffix clicks are propagating up to the wrapping `div` with `onClick={this.handleClick}`, so that `handleClick` gets called, which sets `.focus()` on the input.

I've updated my Playground example to use `stopPropagation` on the first suffix example _(you can see it log the console now as well)_.

So if someone is passing a clickable element into suffix/prefix, they will need to handle the elements event behaviour.

Another thing that is happening though is a "flash of focus state" on that wrapping div... because it ALSO has a `handleFocus` that sets `focus` true. So, its basically like a `:active` style getting fired on the input when the button is clicked.

That part I'm not sure how to solve... perhaps we could check for "original target" within the `handleFocus`? Not sure if that will work or how big of a concern it is.

ry5n:
Do we actually need the focus and click handlers? What happens if we remove them?

beefchimi:
@ry5n thats a great question and I feel like there are _reasons_ for it that I might not be aware of. Not sure who to ping for more context... @AndrewMusgrave do you happen to have any insight into the application of these handlers?

ry5n:
I feel like part of it is emulating `:focus-within`, but we might not need that. Not sure if that’s all it’s doing though.

AndrewMusgrave:
You shouldn't have to stop propagation. I don't think we really get much value from `handleClick`. From an accessibility standpoint, we should let the browser handle focus rather than forcing it on the input. How would we want the spinner to act? Right now it seems `handleClick` focuses the input when our spinner is interacted with.

From what I can tell the handleFocus/handeBlur methods are for providing a visual focus state to the input when a connected field or spinner are pressed.  A simple option might be to stop propagation for blur/focus events on the pre/suffix. Do we want the input to be focused when a prefix or suffix are pressed? @ry5n

ry5n:
I think the original reason for the click and focus handlers was to “[p]revent the focus ring flashing when clicking e.g. the type="number" spinbuttons. This is accomplished by maintaining the focus ring when anything inside the field has focus”. (from https://github.com/Shopify/polaris-react-deprecated/pull/644)

However, we now need prefix/suffix elements to be able to have focus separately from the field itself, so I think this needs to be refactored. Note the spin button behavior should remain the same (native spin buttons, at least on Chrome, return focus to the field after they’re clicked).

AndrewMusgrave:
> Note the spin button behavior should remain the same (native spin buttons, at least on Chrome, return focus to the field after they’re clicked).

Chrome/safari focus the input when the spinner is used. Firefox is a little different and if the input is focused using a spinner will not affect the focus state, however, if you use the spinner and the input is not focused it does not receive focus.

ry5n:
> if you use the spinner and the input is not focused it does not receive focus

Huh, you’re right. The weird thing is that I’m not sure what element has focus in Firefox, after clicking a spin button. Hitting tab afterwards focuses the next interactive element after the input, but shift-tab focuses the input. It’s as if the spin buttons have focus, but there’s no focus indicator, and up/down keyboard keys don’t do anything.

The UI seems to be unspecified by the HTML spec. Edge doesn’t have spin buttons at all. I think the Chrome/Safari behavior makes the most sense still.

AndrewMusgrave:
> The UI seems to be unspecified by the HTML spec. Edge doesn’t have spin buttons at all. I think the Chrome/Safari behavior makes the most sense still.

Agreed 👍

ry5n:
Awesome! @AndrewMusgrave do you have time to look into how to make this work?

AndrewMusgrave:
There are two quick fixes for this. We could add wrappers around the prefix/suffix and stop propagation, or we move around our handlers which would require some style changes as well.

ry5n:
It would be nice to not add additional markup, but I guess whatever is reasonable in terms of time has my 👍

beefchimi:
@AndrewMusgrave if you let me know exactly what you want from this PR, I might be able to squeeze in the work today or tomorrow so we can ship this thing.

AndrewMusgrave:
After looking at textfield's code, we wouldn't need to add wrappers and can achieve this through delegation.

> if you let me know exactly what you want from this PR, I might be able to squeeze in the work today or tomorrow so we can ship this thing.

We're looking to keep all of the existing functionality, except for prefix and suffix where we want to block our handlers. Here's a [diff on some changes](https://github.com/Shopify/polaris-react/compare/textfield-prefix-suffix-clicks...tf-w-delegated-handlers?expand=1) that achieve this

dpersing:
Since this is non-standard, I'd like to recommend we include two things in this PR or one following closely behind:

- An example of a best practice in the example section
- Best practices documentation in the `README.md`/ style guide for when this pattern should be used.

ry5n:
> An example of a best practice in the example section

The best example in my mind is the eye icon (or text button) for showing/hiding a password.

> Best practices documentation in the README.md

👍 To me prefix/suffix actions should be used for actions that directly affect the text input itself—like showing/hiding the password value or appearance. Actions that don’t directly affect the textfield itself should prefer the connected left/right feature instead. For example, a unit selection attached to a weight field feels like it’s closely related, but it’s still a separate input field. It should almost certainly be implemented as a `Select` passed to `connectedRight`. 

Also, worth mentioning that we have a dedicated API for enabling a clear button when the field has a value. Use `clearButton` instead of prefix/suffix for that interaction.

AndrewMusgrave:
@beefchimi Let me know when this is ready for review!

beefchimi:
@AndrewMusgrave I'm not sure how to test most of this new code? Any chance I can get an assist?

UNRELEASED.md
```diff
@@ -10,14 +10,13 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - `Page > Header` now must pass `content` to the `actions` within `secondaryActions` and `actionGroups` ([#1653](https://github.com/Shopify/polaris-react/pull/1653))
 
-### Deprecations
```
AndrewMusgrave: Could we add this back?

src/components/TextField/tests/TextField.test.tsx
```diff
@@ -344,6 +387,49 @@ describe('<TextField />', () => {
...
+
+      textField.find(`#${mockButtonId}`).simulate('click');
+
+      expect(onClickSpy).toHaveBeenCalled();
+      expect(textField.getDOMNode().querySelector('input')).not.toBe(
+        document.activeElement,
+      );
+    });
+
+    it.only('does not set focus `onFocus` for the <input /> if the `target` is the `suffix`', () => {
```
AndrewMusgrave:
```suggestion
    it('does not set focus `onFocus` for the <input /> if the `target` is the `suffix`', () => {
```

As you pointed out I forgot to remove this 😄

UNRELEASED.md
beefchimi: its still there! what i removed was a duplicate :)

beefchimi:
@ry5n all issues should not be addressed. Any chance we can get a final review from you?

beefchimi:
All issues should be resolved. Just waiting on final approval from @ry5n

beefchimi:
>  The only thing I noticed is that the text field still gets its focus outline when the prefix/suffix action is focused.

@ry5n can you point to where you are seeing this? We specifically addressed this issue - and I am not able to reproduce:

shoiooop

ry5n:
Maybe I didn’t have the latest version. Gif looks good. Happy to see this go in 🎉

PR #1397, opened 2019-05-01
Applying font-family to button elements

arapehl:
Due to the [cascade order](https://www.w3.org/TR/CSS21/cascade.html#cascading-order), the browser's User Agent Stylesheet, which defines a style for `button` is overriding the `font-family` value [specified by Polaris](https://github.com/Shopify/polaris-react/blob/master/src/styles/global/elements.scss#L9) on `html, body`. This causes the style not to be inherited as intended for the `button` element. It's why we see different fonts when an `ActionList` contains both `a` and a `button` items.

Fixes https://github.com/Shopify/polaris-react/issues/1340

Adding `button` to the elements targeted by the `font-family` style rule separating out `html, body` so that only `font-family` is applied to all three.

## 

### How to 🎩

<details>

```jsx

interface State {}

  render() {
    return (
      <Page title="Playground">
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -18,6 +18,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- Applied `font-family` to `button` elements which were being overridden by User Agent Stylesheet ([#1397](https://github.com/Shopify/polaris-react/pull/1397))
 - Allowed `Tooltip` content to wrap no nonbreaking strings [#1395](https://github.com/Shopify/polaris-react/pull/1395)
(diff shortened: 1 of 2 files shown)


arapehl:
@beefchimi Added UNRELEASED entry and committed a fixup.

UNRELEASED.md
```diff
@@ -18,6 +18,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- Apply `font-family` to `button` elements which were being overridden by User Agent Stylesheet ([#1397](https://github.com/Shopify/polaris-react/pull/1397))
```
kaelig: Use past tense

PR #1400, opened 2019-05-02
Switching Transition component for CSSTransition

arapehl:
Popovers were [seen to "flutter"](https://github.com/Shopify/online-store-web/issues/2636) when animating on iOS. The `<Transition>` component was calling `renderOverlay` 5 times per `render` which had the effect of interrupting the CSS transition of the `.Popover` element and causing a flutter. By using the `<CSSTransition>` component instead, we better manage the transition and avoid the flutter.

Fixes https://github.com/Shopify/polaris-react/issues/1315

Swapping `<Transition>` component for `<CSSTransition>`.

|Before | After|
|--|--|
|flickery_popover|May-02-2019 17-59-00|

## 

### How to 🎩

<details>

```jsx

interface State {}

  render() {
    return (
      <Page title="Playground">
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -16,6 +16,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- Fixed `Popover` fade-in flutter on iOS by switching Transition component for CSSTransition [#1400](https://github.com/Shopify/polaris-react/pull/1400)
+
 ### Documentation
 
 ### Development workflow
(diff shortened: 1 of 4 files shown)


beefchimi:
Just wanted to point out that Ara and I chatted over Slack about this.

I forked his branch and experimented with resolving my concerns but was not able to get everything correct. Mainly, my issue with the current solution is that it seems to only transition in... but disappears immediately on exit.

You can see my changes here: https://github.com/Shopify/polaris-react/commit/ab27dbc87f044094d046c698ddf5036d89dde869

Included in my changes are:
- "Popover" transition state classes should be variants _(`.PopoverEnter` should be `.enter`)_
- `will-change` should be able to be moved into the main class declaration
- Not sure if `TransitionStatus` is needed anymore? Either way I think there are a few lines that can be removed.
- My solution only works on mobile when the duration length is long enough... `durationBase` seems to be too short and causes immediate transitions.

arapehl:
I had to remove the `.exiting` class name as well as supporting code as it was interfering with the `.PopoverExitActive` class name the new code uses.

Issue #1377, opened 2019-04-29
Update React Context to use `React.createContext`


AndrewMusgrave:
# Issue summary

As part of `v4` we're changing our react context to use the new API. This issue is a list of what still needs to be done. All work should be done on the `version-4.0.0` branch.

* [x] AppProvider
* [x] withSticky
* [x] ComboBox
* [x] Navigation
* [x] Scrollable
* [x] ThemeProvider
* [x] ScrollLock
* [x] Sticky

PR #1402, opened 2019-05-02
[Navigation] Use new context api

AndrewMusgrave:

Part of #1377

Nothing special, using the new context API and updating tests

### DIFF
--- a/src/components/Navigation/Navigation.tsx
+++ b/src/components/Navigation/Navigation.tsx
@@ -2,8 +2,8 @@ import * as React from 'react';
 
 import Scrollable from '../Scrollable';
 
-import {UserMenu, Section, Item} from './components';
-import {contextTypes, SectionType} from './types';
+import {UserMenu, Section, Item, Provider} from './components';
+import {SectionType} from './types';
 
 import styles from './Navigation.scss';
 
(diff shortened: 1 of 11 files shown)


src/components/Navigation/components/Item/tests/Item.test.tsx
```diff
@@ -468,8 +468,13 @@ function itemForLocation(location: string, overrides: Partial<ItemProps> = {}) {
...
       {...overrides}
     />,
-    {
-      context: {location},
-    },
+    {location},
   ).find(Item);
 }
+
+function mountWithNavigationProvider(
```
AndrewMusgrave: These mount* functions are only temporary and will be replaced once we're fully moved over to the new context API

src/components/Navigation/Navigation.tsx
```diff
@@ -34,7 +33,7 @@ export default class Navigation extends React.Component<Props, never> {
     }
   }
 
-  getChildContext() {
+  get getContext() {
```
BPScott: `getContext` makes me think this is a function, and `get getContext` feels odd. Should this just be `get context` then accessed with this.context?

Should this be private?

danrosenthal: `this.context` feels a bit too similar (the same as) the old context API, though I agree that the double `get` is a bit semantically weird

PR #1403, opened 2019-05-02
[AutoComplete] Use new context api

AndrewMusgrave:

Part of #1377

Nothing special, using the new context API and updating tests

### DIFF
--- a/src/components/Autocomplete/components/ComboBox/ComboBox.tsx
+++ b/src/components/Autocomplete/components/ComboBox/ComboBox.tsx
@@ -9,9 +9,9 @@ import ActionList from '../../../ActionList';
 import Popover from '../../../Popover';
 import {PreferredPosition} from '../../../PositionedOverlay';
 import {ActionListItemDescriptor, Key} from '../../../../types';
-import {contextTypes} from '../types';
+import {ComboBoxContext} from '../types';
 import KeypressListener from '../../../KeypressListener';
-import {TextField} from './components';
+import {TextField, Provider} from './components';
 
 import styles from './ComboBox.scss';
 
@@ -58,17 +58,9 @@ export interface Props {
   onEndReached?(): void;
 }
 
-export interface Context {
-  comboBoxId: string;
-  selectedOptionId?: string;
-  subscribe(callback: () => void): void;
-  unsubscribe(callback: () => void): void;
-}
-
 export default class ComboBox extends React.PureComponent<Props, State> {
   static TextField = TextField;
   static OptionList = OptionList;
-  static childContextTypes = contextTypes;
 
   static getDerivedStateFromProps(
     {
@@ -123,17 +115,14 @@ export default class ComboBox extends React.PureComponent<Props, State> {
     popoverWasActive: false,
   };
 
-  private subscriptions: {(): void}[] = [];
   private popoverScrollContainer: React.RefObject<
     HTMLDivElement
   > = React.createRef();
 
-  getChildContext(): Context {
+  get getContext(): ComboBoxContext {
     return {
       comboBoxId: this.state.comboBoxId,
       selectedOptionId: this.selectedOptionId,
-      subscribe: this.subscribe,
-      unsubscribe: this.unsubscribe,
     };
   }
 
@@ -161,7 +150,6 @@ export default class ComboBox extends React.PureComponent<Props, State> {
   componentDidUpdate(_: Props, prevState: State) {
     const {contentBefore, contentAfter, emptyState} = this.props;
     const {navigableOptions, popoverActive, popoverWasActive} = this.state;
-    this.subscriptions.forEach((subscriberCallback) => subscriberCallback());
 
     const optionsChanged =
       navigableOptions &&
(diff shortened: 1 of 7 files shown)


src/components/Autocomplete/components/types.ts
```diff
@@ -1,9 +1,4 @@
-import * as PropTypes from 'prop-types';
-import {ValidationMap} from 'react';
-
-export const contextTypes: ValidationMap<any> = {
-  selectedOptionId: PropTypes.string,
-  comboBoxId: PropTypes.string,
-  subscribe: PropTypes.func,
-  unsubscribe: PropTypes.func,
-};
+export interface ComboBoxContext {
```
BPScott: What do you think to moving this interface into the Context subcomponent so it can live with the Provider / Consumer?

Is there every any case where we'd need this but not the Provider or Consumer?

src/components/Autocomplete/components/ComboBox/components/TextField/TextField.tsx
```diff
@@ -1,34 +1,21 @@
 import * as React from 'react';
 
-import {contextTypes} from '../../../types';
+import {ComboBoxContext} from 'components/Autocomplete/components/types';
```
BPScott: This needs to be a relative path. (if you decide to keep it instead of moving the ComboBoxContent int the Context subcomponent)

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
```diff
@@ -123,17 +115,14 @@ export default class ComboBox extends React.PureComponent<Props, State> {
     popoverWasActive: false,
   };
 
-  private subscriptions: {(): void}[] = [];
   private popoverScrollContainer: React.RefObject<
     HTMLDivElement
   > = React.createRef();
 
-  getChildContext(): Context {
+  get getContext(): ComboBoxContext {
```
BPScott: Is there any value in having this be a function if it's only used once? Should we inline it into render():

```js
const context = {
  comboBoxId: this.state.comboBoxId
  selectedOptionId: this.selectedOptionId,
} as ComboBoxContext;

// ...

return (
  <Provider value={context}>
    {/*... */}
  </Provider>
)
```

src/components/Autocomplete/components/types.ts
AndrewMusgrave: Depends, the two cases I see using this outside of `Context.tsx` is for typing context in the component that creates it and passes it to the provider and if we ever want to create a hook for using the context (we probably won't).

src/components/Autocomplete/components/ComboBox/components/TextField/TextField.tsx
AndrewMusgrave: Auto import got me 🙈

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
AndrewMusgrave: Not really in this situation. Other components may have a more complex context, but `ComboBox` is rather simple. It's more for a common convention / easily searchable. I wouldn't be opposed to changing it if others don't think we get value from it.

src/components/Autocomplete/components/ComboBox/components/Context/Context.tsx
```diff
@@ -0,0 +1,6 @@
+import * as React from 'react';
+import {ComboBoxContext} from '../../../types';
+
+const {Provider, Consumer} = React.createContext<ComboBoxContext>({});
```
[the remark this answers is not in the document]
AndrewMusgrave: If `useContext` only accepts `createContext` I can make a quick change once all the context's are change, and agree it would go better in `{Component}/Context.tsx`

src/components/Autocomplete/components/types.ts
BPScott: In the first case you'd be able to do `import {ComboBoxContext, Provider} from 'X';` so you'd only need one import instead of 2. If we wanted a hook then we'd need to access the result of `React.createContext` so having the type in the same file would save you an import there too.

I reckon stick it in the same file as where you create the context.

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
BPScott: Looking at your other PRs it looks like their equivalent of this function is only used in their render function too. It seems the simple style of single usage is the common case so I think we should stick with that. The hooks-style of coding seems to be sticking everything into the render function anyway so making it follow that pattern feels nicer for consistency

src/components/Autocomplete/components/ComboBox/components/Context/Context.tsx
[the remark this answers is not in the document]
BPScott: (nitpick: should that be `{Component}/context.tsx` as capital letters are for files that export a react component?)

I'm happy to do everything in this style for now and then we change over to the proposed new style all in one go :)

BPScott:
> I still recommend resolving the conversations with Ben before merging.

I'm happy to say merge this as is (+ fixing that autoimport goof) as it follows the current style, then we can go back and update all context usage to be a context.ts instead of a subcomponent in a follow-up PR once the existing legacy context migration PRs are merged ([per Andrew's comment](https://github.com/Shopify/polaris-react/pull/1403#discussion_r280799506))

src/components/Autocomplete/components/types.ts
BPScott: Resolving this for now, as we can move the typing into a ComponentName/context.ts file in a follow up PR

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
BPScott: I wouldn't think to search for "getContext" to find contexts - I'd search for React.createContext and follow its usage.

Inlining these can be done as part of the follow up PR so I'm going to resolve this thread for now.

src/components/Autocomplete/components/ComboBox/components/Context/Context.tsx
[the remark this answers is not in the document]
BPScott: Resolving this as we can make this change, along with an equivalent change for all usages of context in a follow up PR

PR #1414, opened 2019-05-06
[ColorPicker] Improve touch interactions

beefchimi:
While working with `ColorPicker` and testing on mobile, I noticed I was getting a ton of console logs:

Screen Shot 2019-05-06 at 2 54 15 PM

Seemed pretty straight forward - we called `event.preventDefault()` within the `handleMove`, and the browser was defaulting to `passing: true`. In a nutshell, `passive=true` means you do not intend to call `preventDefault` within your handler, so the browser doesn't need to check for it. For those interested, you can read more here: https://developers.google.com/web/updates/2017/01/scrolling-intervention

But while I was at it, I also noticed the page was scrolling alongside touch interactions on the ColorPicker. This is an issue a few of us have already come across. The solution is to listen for `touchmove` on the `window` and `preventDefault` when we want to favour _some other_ interaction - in this case, dragging the color slider.

Please tophat this in Chrome + iOS. Run `yarn dev` and play around in the Storybook ColorPicker examples for both `master` and this branch.

cc/ @tsov

### DIFF
--- a/src/components/ColorPicker/components/Slidable/Slidable.tsx
+++ b/src/components/ColorPicker/components/Slidable/Slidable.tsx
@@ -55,11 +55,19 @@ export default class Slidable extends React.PureComponent<Props, State> {
     };
 
     const moveListener = dragging ? (
-      <EventListener event="mousemove" handler={this.handleMove} />
+      <EventListener
+        event="mousemove"
+        handler={this.handleMove}
+        passive={false}
+      />
     ) : null;
 
     const touchMoveListener = dragging ? (
-      <EventListener event="touchmove" handler={this.handleMove} />
+      <EventListener
+        event="touchmove"
+        handler={this.handleMove}
+        passive={false}
+      />
     ) : null;
 
     const endDragListener = dragging ? (
@@ -121,7 +129,10 @@ export default class Slidable extends React.PureComponent<Props, State> {
   private handleMove = (event: MouseEvent | TouchEvent) => {
     event.stopImmediatePropagation();
     event.stopPropagation();
-    event.preventDefault();
+
+    if (event.cancelable) {
+      event.preventDefault();
+    }
 
     if (event.type === 'mousemove') {
       const mouseEvent = event as MouseEvent;
--- a/src/components/ColorPicker/components/Slidable/Slidable.tsx
+++ b/src/components/ColorPicker/components/Slidable/Slidable.tsx
@@ -18,6 +18,25 @@ export interface Props {
   onDraggerHeight?(height: number): void;
 }
 
+let isDragging = false;
+
+// Required to solve a bug causing the underlying page/container to scroll
+// while trying to drag the ColorPicker controls.
+// This must be called as soon as possible to properly prevent the event.
+// `passive: false` must also be set, as it seems webkit has changed the "default" behaviour
+// https://bugs.webkit.org/show_bug.cgi?id=182521
+window.addEventListener(
+  'touchmove',
+  (event) => {
+    if (!isDragging) {
+      return;
+    }
+
+    event.preventDefault();
+  },
+  {passive: false},
+);
+
 export default class Slidable extends React.PureComponent<Props, State> {
   state: State = {
     dragging: false,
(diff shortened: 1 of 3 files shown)


src/components/ColorPicker/components/Slidable/Slidable.tsx
```diff
@@ -18,6 +18,25 @@ export interface Props {
...
+  'touchmove',
+  (event) => {
+    if (!isDragging) {
+      return;
+    }
+
+    event.preventDefault();
+  },
+  {passive: false},
+);
```
danrosenthal: Did you try adding this to the `constructor` or `componentDidMount` instead?

src/components/ColorPicker/components/Slidable/Slidable.tsx
```diff
@@ -18,6 +18,25 @@ export interface Props {
   onDraggerHeight?(height: number): void;
 }
 
+let isDragging = false;
```
danrosenthal: Would it make more sense to add this as a private member on the class? That feels more conventional

src/components/ColorPicker/components/Slidable/Slidable.tsx
tsov: This is done on purpose, since webkit seems to have issues with their `touchmove` events, [see webkit bug](https://bugs.webkit.org/show_bug.cgi?id=184250)

The problem is, the handler needs to be attached as early as possible, and not as part of another event trigger, otherwise the event becomes uncancelable on safari mac & ios. Chrome does not have the same issues when testing with their dev device tools

We had to do the same for draggable: https://github.com/Shopify/draggable/blob/master/src/Draggable/Sensors/TouchSensor/TouchSensor.js#L17

src/components/ColorPicker/components/Slidable/Slidable.tsx
```diff
@@ -111,17 +138,22 @@ export default class Slidable extends React.PureComponent<Props, State> {
...
+    isDragging = false;
     this.setState({dragging: false});
   };
 
   private handleMove = (event: MouseEvent | TouchEvent) => {
     event.stopImmediatePropagation();
     event.stopPropagation();
-    event.preventDefault();
+
+    if (event.cancelable) {
```
tsov: ✨ yes!

src/components/ColorPicker/components/Slidable/Slidable.tsx
beefchimi: We need this in scope for the `window.addEventListener`, so there is no way to make this a private member of `Slidable`

danrosenthal: Makes sense!

beefchimi:
@danrosenthal added two tests for the `EventListener` component - but I'm not sure how we would test the `window.addEventListener`. I don't see this being done elsewhere in the codebase. It might not be a very useful/reliable test.

src/components/ColorPicker/components/Slidable/Slidable.tsx
```diff
@@ -18,6 +18,25 @@ export interface Props {
...
+// Required to solve a bug causing the underlying page/container to scroll
+// while trying to drag the ColorPicker controls.
+// This must be called as soon as possible to properly prevent the event.
+// `passive: false` must also be set, as it seems webkit has changed the "default" behaviour
+// https://bugs.webkit.org/show_bug.cgi?id=182521
+window.addEventListener(
+  'touchmove',
+  (event) => {
+    if (!isDragging) {
+      return;
```
danrosenthal: 🔴

src/components/ColorPicker/components/Slidable/Slidable.tsx
danrosenthal: 💛

src/components/ColorPicker/components/Slidable/Slidable.tsx
```diff
@@ -111,17 +138,22 @@ export default class Slidable extends React.PureComponent<Props, State> {
...
     this.setState({dragging: false});
   };
 
   private handleMove = (event: MouseEvent | TouchEvent) => {
     event.stopImmediatePropagation();
     event.stopPropagation();
-    event.preventDefault();
+
+    if (event.cancelable) {
+      event.preventDefault();
```
danrosenthal: 🔴

beefchimi:
@danrosenthal yes - if you don't mind, I'd love to see how these `codecov` lines can be tested. Even if you are able to just point me at somewhere else in the codebase that accomplishes something similar, I can try to take the same approach.

PR #1417, opened 2019-05-06
[Navigation.Section] Prevent rollup from collapsing when sub nav and on mobile

### LINKED ISSUE #1004: Navigation rollup always collapses when section heading is clicked
# Issue summary
The Navigation rollup always collapses `onClick`, and there is no way to prevent that from happening due to the fact that the state ownership to decide whether the rollup should be expanded or collapsed is nested within the `Navigation.Section` component (https://github.com/Shopify/polaris-react/blob/master/src/components/Navigation/components/Section/Section.tsx#L181-L183)

This issue is blocking fixing https://github.com/Shopify/web/pull/10471 in `shopify/web`.

## Expected behavior

There should either be a way to prevent the default behaviour of collapsing the rollup onClick (for instance, by returning false in the onClick handler), or the state ownership should be lifted out of the Navigation.Section component and passed down as props instead so that the consumer can decide when to collapse / expand the rollup.

The larger issue is that the Navigation is dismissed when a rollup section header is clicked on mobile.  Instead, the navigation should remain open and the rollup disclosure should open/close. Perhaps this should be fixed directly in Polaris as opposed to conditionally passing the `onNavDismiss` handler in the consumer (`web`)

cc @chloerice

tmlayton:

Fixes #1004

Prevents the rollup from collapsing when an item clicked has sub nav and at the mobile breakpoint 

|Before|After|
|---|---|
|rollup bug|prevent rollup|

## 

### How to 🎩

<details>

```jsx
import {Page, Navigation} from '../src';

interface State {}

  render() {
    return (
      <Page title="Playground">
        <Navigation location="/">
          <Navigation.Section
            rollup={{
              after: 1,
              view: 'view',
              hide: 'hide',
              activePath: '/',
            }}
            items={[
              {
                url: '/path/to/place',
                label: 'Home',
                icon: 'home',
              },
              {
                url: '/path/to/place',
                label: 'Orders',
                icon: 'orders',
                badge: '15',
              },
              {
                url: '/path/to/place',
                label: 'Products',
                icon: 'products',
                subNavigationItems: [
                  {
                    url: '/path/to/place',
                    label: 'Foo',
                  },
                  {
                    url: '/path/to/place',
                    label: 'Bar',
                  },
                  {
                    url: '/path/to/place',
                    label: 'Baz',
                  },
                ],
              },
            ]}
          />
        </Navigation>
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

* [ ] ~~Updated the component's `README.md` with documentation changes~~
* [ ] ~~[Tophatted documentation](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting%20documentation.md) changes in the style guide~~

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -34,6 +34,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Fixed styling issue for Pagination component previous/next buttons when tooltips present ([#1277](https://github.com/Shopify/polaris-react/pull/1277))
 - Fixed a regression introduced in Polaris v3 where certain children of a `TextContainer` would have no top margin ([#1357](https://github.com/Shopify/polaris-react/pull/1357))
 - Added border to `Tooltip` in Windows high contrast mode ([#1405](https://github.com/Shopify/polaris-react/pull/1405))
+- Fixed `Navigation.Section` rollup to prevent collapsing when a sub navigation was expanding ([#1417](https://github.com/Shopify/polaris-react/pull/1417))
 
 ### Documentation
 
--- a/src/components/Navigation/components/Section/Section.tsx
+++ b/src/components/Navigation/components/Section/Section.tsx
@@ -3,6 +3,7 @@ import {HorizontalDotsMinor} from '@shopify/polaris-icons';
 import {classNames} from '@shopify/react-utilities/styles';
 import {createUniqueIDFactory} from '@shopify/javascript-utilities/other';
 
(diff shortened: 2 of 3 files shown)


chloerice:
Hey Tim, thanks for working on this!! I just tophatted in `web` and it doesn't seem to have fixed the issue. I thought it might just be a caching issue on my phone so I started a new session in an incognito window and it's still the same. Can you try this branch in `web` and see if it works for you?

PR #1426, opened 2019-05-08
[DataTable] Fix for iOS fixed column issue

dleroux:
resolves: https://github.com/Shopify/polaris-react/issues/489

On iOS, the overflow-x auto on ios doesn't allow the position absolute column to be visible.

In trying to solve this I changed the margin-left to padding-left. The result is that scrolls bar appears across the entire bottom, which IMO, is a decent compromise:

datatable

As a result though, on iOS, the first columns scrolls with the content so this still needs some work but might be easier to solve than the current issue?

Fixes: https://github.com/Shopify/polaris-react/issues/489

## 

### How to 🎩

<details>

```jsx
import {Page, DataTable, Card} from '../src';

interface State {}

  state = {
    sortedRows: null,
  };

  sortCurrency = (rows, index, direction) => {
    return [...rows].sort((rowA, rowB) => {
      const amountA = parseFloat(rowA[index].substring(1));
      const amountB = parseFloat(rowB[index].substring(1));

      return direction === 'descending' ? amountB - amountA : amountA - amountB;
    });
  };

  handleSort = (rows) => (index, direction) => {
    this.setState({sortedRows: this.sortCurrency(rows, index, direction)});
  };

  render() {
    const {sortedRows} = this.state;
    const initiallySortedRows = [
      ['Emerald Silk Gown', '$875.00', 124689, 140, '$122,500.00'],
      ['Mauve Cashmere Scarf', '$230.00', 124533, 83, '$19,090.00'],
      [
        '$445.00',
        124518,
        32,
        '$14,240.00',
      ],
    ];

    return (
        <Card>
          <DataTable
            columnContentTypes={[
              'text',
              'numeric',
              'numeric',
              'numeric',
              'numeric',
            ]}
            headings={[
              'Product',
              'Price',
              'SKU Number',
              'Net quantity',
              'Net sales',
            ]}
            rows={rows}
            sortable={[false, true, false, false, true]}
            initialSortColumnIndex={4}
            onSort={this.handleSort(rows)}
          />
        </Card>
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/DataTable/DataTable.scss
+++ b/src/components/DataTable/DataTable.scss
@@ -1,5 +1,9 @@
 $fixed-column-width: rem(145px);
 $breakpoint: 768px;
+$stacking-order: (
+  fixed-cell: 2,
+  icon: 1,
+);
 
 .DataTable {
   position: relative;
@@ -26,7 +30,7 @@ $breakpoint: 768px;
   }
 
   .ScrollContainer {
-    margin-left: rem($fixed-column-width);
+    padding-left: rem($fixed-column-width);
   }
 }
 
@@ -58,7 +62,7 @@ $breakpoint: 768px;
 .ScrollContainer {
   overflow-x: auto;
   // account for a mysterious gap in Safari when not collapsed
-  margin-left: rem(140px);
+  padding-left: rem($fixed-column-width);
   -webkit-overflow-scrolling: touch;
 }
 
@@ -95,12 +99,14 @@ $breakpoint: 768px;
 }
 
 .Cell {
+  position: relative;
+  vertical-align: top;
   padding: spacing();
   border-bottom: border-width() solid color('sky', 'light');
+  background-color: color('white');
   white-space: nowrap;
   text-align: left;
   transition: background-color 0.2s ease-in-out;
-  vertical-align: top;
 }
 
 .Cell-numeric {
@@ -111,6 +117,7 @@ $breakpoint: 768px;
   @include text-emphasis-normal;
   @include text-breakword;
   position: absolute;
+  z-index: z-index(fixed-cell, $stacking-order);
   top: auto;
   left: 0;
   width: $fixed-column-width;
@@ -136,6 +143,7 @@ $breakpoint: 768px;
 }
 
 .Icon {
+  z-index: z-index(icon, $stacking-order);
   display: flex;
   align-self: flex-end;
   opacity: 0;
--- a/src/components/DataTable/DataTable.tsx
+++ b/src/components/DataTable/DataTable.tsx
@@ -75,7 +75,12 @@ export class DataTable extends React.PureComponent<
     } = this;
     let collapsed = false;
     if (table && scrollContainer) {
-      collapsed = table.scrollWidth > scrollContainer.clientWidth;
+      const headerCells = table.querySelectorAll(
+        headerCell.selector,
+      ) as NodeListOf<HTMLElement>;
+      const fixedColumnWidth = headerCells[0].offsetWidth;
+      collapsed =
+        table.scrollWidth + fixedColumnWidth > scrollContainer.clientWidth;
       scrollContainer.scrollLeft = 0;
     }
     this.setState(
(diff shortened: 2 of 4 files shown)


markiiikram:
@chloerice and @dleroux would you be open to shipping this as in interim fix, so the UI shows something instead of nothing?

chloerice:
> @chloerice and @dleroux would you be open to shipping this as in interim fix, so the UI shows something instead of nothing?

I'd say that's the right call. Tophatting meow 🐱 👀

chloerice:
Quick update - tophatted and while this fixes the cell content not rendering in iOS, the content of sortable cells is having some z-index issues that'll need to be addressed. I'll poke around at it tomorrow.

dleroux:
Noticing the Header of the fixed column will
also need a background colour. I’m away today, feel free to take care if you have time, if not I can make some time next week.

src/components/DataTable/DataTable.scss
```diff
@@ -111,6 +112,7 @@ $breakpoint: 768px;
   @include text-emphasis-normal;
   @include text-breakword;
   position: absolute;
+  z-index: 3;
```
dleroux: our icons have a z-index of 2.

dleroux:
@chloerice I was curious about this z-index thing and sorted it out. Not sure if there's a better way than hardcoding 3 in there.

src/components/DataTable/DataTable.tsx
```diff
@@ -75,7 +75,12 @@ export class DataTable extends React.PureComponent<
     } = this;
     let collapsed = false;
     if (table && scrollContainer) {
-      collapsed = table.scrollWidth > scrollContainer.clientWidth;
+      const headerCells = table.querySelectorAll(
+        headerCell.selector,
+      ) as NodeListOf<HTMLElement>;
+      const fixedColumnWidth = headerCells[0].offsetWidth;
+      collapsed =
+        table.scrollWidth + fixedColumnWidth > scrollContainer.clientWidth;
```
dleroux: looks like we will need to add the `fixedColumnwidth` to this calculation as well.

dleroux: Since the CSS has a hardcoded value, for performance, I wonder if we aren't better using a token or hardcoded value for this. Thoughts?

chloerice: This was something I was looking to centralize when first implementing, but there wasn't a good way for us to share a single value between the SCSS and the JSX at the time. Now that we have `polaris-tokens` it's actually possible 🙌 Let's ship this quick fix (tophatting in iOS again now 👀) and I'll open separate PRs for adding and using a token for the `fixedColumnWidth`.

src/components/DataTable/DataTable.tsx
```diff
@@ -75,7 +75,12 @@ export class DataTable extends React.PureComponent<
     } = this;
     let collapsed = false;
     if (table && scrollContainer) {
-      collapsed = table.scrollWidth > scrollContainer.clientWidth;
+      const headerCells = table.querySelectorAll(
```
tmlayton: Can we just `table.querySelector` to get the first header cell since we are only using the first element below (`headerCells[0].offsetWidth`)?

dleroux:
Will need tests to get to 90% so this quick fix isn’t so quick anymore. I don’t have the bandwidth to add tests right now.

I agree that it is strange that it was built for mobile devices but none respect the fixed column. I believe this is as expected because the ux of having scrollbars while swiping can be cumbersome. (With iframes the frame just expands)

That said how do you feel we should proceed? Write tests and ship, just ship, or just put on hold. I lean towards writing ‘some’ tests and shipping until we get time do rethink this component because right now it isn’t usable on ios.

chloerice:
> Will need tests to get to 90% so this quick fix isn’t so quick anymore. I don’t have the bandwidth to add tests right now.
> 
> I agree that it is strange that it was built for mobile devices but none respect the fixed column. I believe this is as expected because the ux of having scrollbars while swiping can be cumbersome. (With iframes the frame just expands)
> 
> That said how do you feel we should proceed? Write tests and ship, just ship, or just put on hold. I lean towards writing ‘some’ tests and shipping until we get time do rethink this component because right now it isn’t usable on ios.

No worries. I will close this PR and create a new one removing the fixed column UX.

chloerice:
> Will need tests to get to 90% so this quick fix isn’t so quick anymore. I don’t have the bandwidth to add tests right now.
> 
> I agree that it is strange that it was built for mobile devices but none respect the fixed column. I believe this is as expected because the ux of having scrollbars while swiping can be cumbersome. (With iframes the frame just expands)
> 
> That said how do you feel we should proceed? Write tests and ship, just ship, or just put on hold. I lean towards writing ‘some’ tests and shipping until we get time do rethink this component because right now it isn’t usable on ios.

No worries. I will close this PR and create a new one removing the fixed column UX.

PR #1431, opened 2019-05-08
[AnnotatedSection] Update type definition for title prop to accept React.ReactNode

francinen:
We came across a use case where we wanted to render another component, such as a `Badge` alongside the heading for an `AnnotatedSection`:

image

We considered passing a React component to the `title` prop that rendered both a `Heading` and a `Badge` component:

```jsx
<Layout.AnnotatedSection
    title={
        <Stack>
            <Heading />
            <Badge />
        </Stack>
    }
>
``` 

While this renders successfuly in the browser, it raises a TypeScript error in the text editor because, at the moment, the `title` prop only supports strings.

image

After consulting @BPScott, we felt it seemed reasonable to change the type restriction for the `title` prop in this scenario: https://shopify.slack.com/archives/C8H54T86Q/p1557331031274400

This PR updates the type definition for the `title` prop of the `AnnotatedSection` component to accept `React.ReactNode` instead of `string`.

### How to 🎩

<details>

```jsx

interface State {}

  render() {
    return (
      <Page title="Playground">
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -16,6 +16,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - Added support for dual values to `RangeSlider` component ([#784](https://github.com/Shopify/polaris-react/pull/784))
 
+- Updated type restrictions for `AnnotatedSection` to allow its `title` prop to accept `React.ReactNode` instead of `string` ([#1431](https://github.com/Shopify/polaris-react/pull/1431))
+
 ### Bug fixes
 
 - Fixed an issue where the JavaScript breakpoints incorrectly set the navigation bar collapsed breakpoint ([#1475](https://github.com/Shopify/polaris-react/pull/1475))
--- a/src/components/Layout/components/AnnotatedSection/AnnotatedSection.tsx
+++ b/src/components/Layout/components/AnnotatedSection/AnnotatedSection.tsx
@@ -5,7 +5,7 @@ import styles from '../../Layout.scss';
 
 export interface Props {
   children?: React.ReactNode;
-  title?: string;
+  title?: React.ReactNode;
   description?: React.ReactNode;
 }
 


francinen:
Thanks for pointing that out, @danrosenthal! Should I something to the README that recommends including a <Heading /> or <h2> in the component being passed to the `title` prop?

danrosenthal:
@francinen looks like this needs some formatting on the UNRELEASED doc

BPScott:
Thanks for opening this up @francinen and sorry about being told one thing in slack and another in GitHub by different people.

---

> We should be conditionally wrapping it in the Heading based on the prop's type like we do in other components which have had this requirement relaxed. You'll also want to add tests for that condition

I disagree with this behaviour. I think changing the wrapping component based up on if we pass a string or a ReactNode is very unintuitive and we should not do it.

 I don't think consumers will realize that changing`<AnnotatedSection title="foo" />` to `<AnnotatedSection title={<>Foo <Badge>Hi</Badge></>} />` will delete the heading and that they'll need to add it themselves: (`<AnnotatedSection title={<Heading>Foo <Badge>Hi</Badge></Heading>} />` to get the desired behaviour.

We want to discourage lots of content within the section header and I think making it so that your title is always wrapped in a `<Heading>` is a good way to discourage that - it makes it impossible to do complex layout style things.

---

It also looks like wrapping is currently the exception rather than the rule in the vast majority of cases:

Taking a look at existing components accept a prop that can be a ReactNode (and could be either a thus either a string or an element):

- `<AccountConnection title>` does not change its wrapping element
- `<Modal title>` does not change its wrapping element
- `<Choice label helptext>` do not change their wrapping elements
- `<Connected left right>` do not change their wrapping elements
- `<EmptyState footerContent>` does not change its wrapping element
- `<FormLayout.Group helpText` does not change its wrapping element
- `<Labelled helpText>` does not change its wrapping element
- `<OptionList.Option label>`does not change its wrapping element
- `<RangeSlider helpText prefix suffix>` do not change their wrapping elements
- `<Select helpText>` does not change its wrapping element
- `<SettingAction action>` does not change its wrapping element
- `<TextField prefix suffix helpText connectedLeft connectedRight>`  do not change their wrapping elements

- `<Card title>` changes its wrapping element - removing the `<Header>`
- `<Card.Section title>` changes its wrapping element - removing the `<Subheading>`
- `<Layout.AnnotatedSection description>` changes its wrapping element - removing the `<p>`
- `<Navigation.Item badge>` changes its wrapping element - removing the `<Badge>`

- `<Page title titleMetadata>` takes a different approach - the title is always a string, but if you add a metadata then it gets injected next to the title. I dislike this and think we should let consumers deal with this layout instead of us trying to second-guess it.

In addition every instance of  the `children` prop doesn't check if it was given a string, but I've omitted those from the list as thats not quite the same use-case and there's a lot of them.

Not wrapping is clearly a more prevalent option. It's only Card/Card.Section titles, Layout.AnnotatedSection descriptions and Navigation.Item badges that pull this trick.

BPScott:
Also ping @ry5n and @yourpalsonja also as you both had opinions on this when I asked in Slack:

> Ben:
> FOLKS! Looks like we’ve found another case where people want us to widen some strictness in typing - they want to put in arbitary JSX instead of a string. Does anybody have a good reason why Layout.AnnotatedSection’s title should continue to be a string, instead of us widening the type to be React.Node so they can go put a badge next to their heading text.
>
> Ryan:
> I’m OK with this, but I’d want to expand design guidance to discourage lots of content in there. Previously in admin we had content and actions in there and they get lost visually. We’ve tried to get away from that.
> 
> Sonja: 
> If anyone is opening an issue for this, we should add a task in there about making sure we put some rationale in the styleguide docs.

danrosenthal:
That's good feedback @BPScott! Thanks for pointing out that this is the exception rather than the rule. This feels like one of those best practices we should be documenting once we've reached consensus on it.

BPScott:
@danrosenthal with all that in mind, do you have an updated opinion of what this PR should do?

Do you think we should keep the "only wrap in a Heading if it's a string" logic that @francinen added in 683c21ad or should we revert that commit?

(For the record I'm firmly in the "avoid changing behaviour based on types, lets the revert that commit" camp)

danrosenthal:
> @danrosenthal with all that in mind, do you have an updated opinion of what this PR should do?
> 
> Do you think we should keep the "only wrap in a Heading if it's a string" logic that @francinen added in 683c21a or should we revert that commit?
> 
> (For the record I'm firmly in the "avoid changing behaviour based on types, lets the revert that commit" camp)

+1, this sounds like the right approach Ben. @francinen please let me know if you have any questions

francinen:
Thanks for all the context, @BPScott and @danrosenthal! I reverted that last commit and fixed the formatting issues in the UNRELEASED.md.

PR #1427, opened 2019-05-08
[ColorPicker] Do not call `window` if "undefined"

beefchimi:
I shipped some code recently that solved a issue with ColorPicker `touchmove` behaviour: https://github.com/Shopify/polaris-react/pull/1414

I made a silly mistake - one that I had already resolved in my own app but forget to transfer over to my `polaris-react` PR.

We need to call `window.addEventListener`, but depending on how your app is configured, that code _could_ get executed on the `server`, where `window` does not exist.

```
[server] Error in server execution, check the console for more info.
window.addEventListener('touchmove', function (event) {
^
ReferenceError: window is not defined
```

All I need to do is wrap that event listener in a conditional to check if `window` is not `undefined`. I've tophatted in my app it everything is working.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -22,6 +22,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- Fix `window` call on `server` executed code [#1427](https://github.com/Shopify/polaris-react/pull/1427)
 - Fixed onClick from firing three times when using the enter key on a `ResourceList` item ([#1188](https://github.com/Shopify/polaris-react/pull/1188))
 - Resolve console `[Intervention]` errors for touch interactions on `ColorPicker`, along with preventing page scrolling while interacting with the color slider ([#1414](https://github.com/Shopify/polaris-react/pull/1414))
 - Applied `font-family` to `button` elements which were being overridden by User Agent Stylesheet ([#1397](https://github.com/Shopify/polaris-react/pull/1397))
--- a/src/components/ColorPicker/components/Slidable/Slidable.tsx
+++ b/src/components/ColorPicker/components/Slidable/Slidable.tsx
@@ -1,5 +1,7 @@
 import * as React from 'react';
+import {isServer} from '@shopify/react-utilities/target';
 import EventListener from '../../../EventListener';
+
 import styles from '../../ColorPicker.scss';
 
 export interface Position {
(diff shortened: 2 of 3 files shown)


beefchimi:
So I have introduced a conditional, which means `codecov` is complaining. Basically, we now wrap the `window` call with `if (!isServer)`.

I see this tested elsewhere in the codebase, such as `Avatar.test.tsx`, but that component's use of `isServer` is within the `render` method, and the test checks for if something is/isn't rendered.

Our use is more complicated. I can't just do `mockIsServer(true)` and expect it to work, because it doesn't actually set `window` to `undefined`, it just changes the `boolean` that is returned. And we are calling `window.dispatchEvent` within our tests...

The only way I could see this being solved is by conditionally dispatching those events... which seems to be a testing no-no?

My attempt:

```
import * as targets from '@shopify/react-utilities/target';

const actualIsServer = targets.isServer;

function mockIsServer(value: boolean) {
  (targets as any).isServer = value;
  return (targets as any).isServer;
}

describe('EventListener', () => {
  afterEach(() => {
    mockIsServer(actualIsServer);
    window.dispatchEvent(new Event('mouseup'));
  });

  it('allows default for touchmove events if dragging the slider but window is undefined', () => {
    const isServer = mockIsServer(true);
    const colorPicker = mountWithAppProvider(
      <ColorPicker color={red} onChange={noop} />,
    );

    colorPicker
      .find(Slidable)
      .first()
      .simulate('mousedown');

    const touch = {clientX: 0, clientY: 0};
    const event = new TouchEvent('touchmove', {
      touches: [touch],
      cancelable: true,
    } as TouchEventInit);
    Object.assign(event, {preventDefault: jest.fn()});

    // eslint-disable-next-line shopify/jest/no-if
    if (!isServer) {
      window.dispatchEvent(event);
    }

    expect(event.preventDefault).not.toHaveBeenCalled();
  });
});
```

Anyone have any opinions? @tmlayton or @AndrewMusgrave

danrosenthal:
I think that, because the listener gets added outside the body of the class, it is always going to add that event listener to the window at the time of import. So any mocking will be too late, as the event listener will always be attached, and so the handler will always get called.

I wonder if there's any way around this... I think not

danrosenthal:
I was able to test this here: https://github.com/Shopify/polaris-react/commit/3f1216c2454bd74a954a721936827e06687dd171.

I was right about the handler always getting attached. So if you add `isServer` as a condition on the early return, the handler added on window outside of the class body never gets called

AndrewMusgrave:
An alternative to Dan's R. approach could be

```tsx
import package from '../path';

describe('...block') {
let spy: jest.SpyInstance;

beforeEach(() => {
  spy = jest.spyOn(package, 'key');
  spy.mockReturnValue(true);
)};

afterEach(() => {
  spy.mockRestore();
});
```

AndrewMusgrave:
Sorry it looks like `isServer` is a variable and not a function so it cannot be mocked like that, however, we can mock the module
```tsx
jest.mock('@shopify/react-utilities/target', () => ({
  ...require.requireActual('@shopify/react-utilities/target'),
  isServer: true,
}));
```

danrosenthal:
> jest.mock('@shopify/react-utilities/target', () => ({
>   ...require.requireActual('@shopify/react-utilities/target'),
>   isServer: true,
> }));

This works, but unfortunately causes existing tests to fail which relied on `isServer` being false. 

Would there be a way to use the actual isServer, or scope this mock to a describe? I tried using `doMock` scoped to the test, but it doesn't seem to have the intended effect.

AndrewMusgrave:
@danrosenthal mentioned wanting to mock this variable for individual tests. There might be a cleaner way but I was able to get it working with
```tsx
jest.resetModules();
jest.mock('@shopify/react-utilities/target', () => ({
  ...require.requireActual('@shopify/react-utilities/target'),
  isServer: () => {
    console.log('Mock is server');
    return true;
  },
}));
const ColorPicker = require('../ColorPicker').default;
```

PR #1435, opened 2019-05-08
PR generator script: remove ping to non-existent team

kaelig:
The @Shopify/polaris-reviewers team doesn't exist, should we remove it from the PR that gets opened when polaris-react is deployed?


BPScott:
The team exists in GitHub but yeah it's defunct. Might as well delete it, any objections?

kaelig:
Oh, so it does exist, but somehow it doesn't transform into a link when mentioned by the bot in polaris-styleguide PRs?

BPScott:
Turns out @dfmcphee is the only maintainer of that group, so only he can delete it.

PR #1436, opened 2019-05-09
[Range slider] Dual range slider

### LINKED ISSUE #595: [RangeSlider] Add dual input

tmlayton:

Closes #595. Adds support for a dual range slider.
Screen Shot 2019-05-08 at 6 55 45 PM

Extends the `value` prop of `RangeSlider` to accept a tuple and splits the internals of `RangeSlider` into two subcomponents, `SingleThumb` (current `RangeSlider`) and `DualThumb` (new component to support a dual range slider)

### 🎩 checklist

### DIFF
--- a/src/components/RangeSlider/README.md
+++ b/src/components/RangeSlider/README.md
@@ -29,6 +29,7 @@ Range sliders should:
 - When a label is visible, it should clearly communicate the purpose of the range input and its values (min, max, step, value)
 - Be labeled as “Optional” when you need to request input that’s not required
 - Validate input as soon as merchants have finished interacting with a field (but not before)
+- Always be used with two text field components when range slider has dual thumbs, to provide accessible alternatives to both the lower and upper thumbs
 
 ---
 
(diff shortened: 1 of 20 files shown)


src/components/RangeSlider/README.md
```diff
@@ -225,6 +227,116 @@ class RangeSliderExample extends React.Component {
...
+    const upperTextFieldValue =
+      intermediateTextFieldValue[1] === value[1]
+        ? value[1]
+        : intermediateTextFieldValue[1];
+
+    return (
+      <Card sectioned>
+        <div style={{marginTop: '20px'}} onKeyDown={this.handleEnterKeyPress}>
+          <RangeSlider
+            label="Money spent is between"
```
dpersing: For the dual-thumb version, the `<label>` is technically invalid since there isn't actually an `<input>` to tie it to, and there isn't a clear way for screen reader users to read it while they're tabbing through controls or a form. I wonder if we could use `aria-describedby` on the thumbs instead to tie the visible label to the controls?

```
<label id="RangeSlider1Label" for="RangeSlider1" class="Label-Text_SY9aZ">Money spent is between</label>
<div class="DualThumb-Wrapper_q5XRy" id="RangeSlider1">...</div>
```

would become

```
<span id="RangeSlider1Label" class="Label-Text_SY9aZ">Money spent is between</span>
...
<button id="RangeSlider1Lower" class="DualThumb-Thumbs_1FGL0" role="slider" aria-valuemin="0" aria-valuemax="2000" aria-valuenow="900" aria-invalid="false" style="left: 920.728px;" aria-describedby="RangeSlider1Label"></button>
...
<button id="RangeSlider1Upper" class="DualThumb-Thumbs_1FGL0" role="slider" aria-valuemin="0" aria-valuemax="2000" aria-valuenow="1000" aria-invalid="false" style="left: 1023.03px;" aria-describedby="RangeSlider1Label"></button>
```

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -0,0 +1,593 @@
...
+      error && styles.error,
+      disabled && styles.disabled,
+    );
+
+    const thumbLowerClassName = classNames(
+      styles.Thumbs,
+      styles.ThumbLower,
+      disabled && styles.disabled,
+    );
+    const thumbUpperClassName = classNames(
```
yourpalsonja: I think _maybe_ there is a case for dryness here in the classnames!

src/components/RangeSlider/README.md
dpersing: Any chance we could get this tweak in, @tmlayton and @yourpalsonja? It will help prevent automated errors in pa11y once that is up and running again.

PR #1451, opened 2019-05-10
[Resource Picker] Fix reference to old callback function

katzenbar:

When developing an app that adds line items to a draft order, I noticed that my ResourcePicker kept resetting my line items to the initial state. Since this component requires being inside of an embedded app and I am working on a private project, I put together an example app that demonstrates the issue: https://github.com/katzenbar/shopify-resource-picker-bug

This example app takes the products selected in the ResourcePicker, and adds them to the end of the list in the card. https://github.com/katzenbar/shopify-resource-picker-bug/blob/master/app/javascript/components/ProductPage.tsx#L24-L28

But when `handleSelection` is run, the `products` array is empty. This results in the old selection being replaced.

bug-demo

The expected behavior should be that the newly selected products are added to the end of the list.

fix-demo

I believe part of reproducing this issue is using React Hooks instead of `setState`, but I have not tested to verify that is the case.

When pulling out `onSelection` and `onCancel` in `componentDidMount` through destructuring, you will keep a reference to the version of that function when the component was mounted. If data these functions reference is updated, these function references become stale and are not updated. In theory, if you were to initially have these callbacks be null and then change them to a real function they would never get called.

To fix this, I changed the code to always register a subscriber for `onSelection` and `onCancel` and get the current reference of these from `this.props` when the callback fires. This seems more efficient that trying to track subscribe/unsubscribe inside of `componentDidUpdate` in the general case. I've tested this on my sample application in Chrome, and it works as I expect now.

If anyone else is experiencing this issue, a workaround is to treat the `ResourcePicker` like the `Toast` example in the documentation and only render it when you are about to open the modal.

```
{showResourcePicker && (
        <ResourcePicker
          resourceType="Product"
          initialQuery={productSearchQuery}
          open={showResourcePicker}
          onSelection={handleResourcePickerSelection}
          onCancel={closeResourcePicker}
        />
      )}
```

### DIFF
--- a/src/components/ResourcePicker/tests/ResourcePicker.test.tsx
+++ b/src/components/ResourcePicker/tests/ResourcePicker.test.tsx
@@ -68,6 +68,28 @@ describe('<ResourcePicker />', () => {
       expect(onSelection).toHaveBeenCalledWith(selectionPayload);
     });
 
+    it('calls the new selection callback when changed', () => {
+      const onSelection = jest.fn();
+      const selectionPayload = {selection: []};
+      const {resourcePicker} = mountWithAppBridge(
+        <ResourcePicker resourceType="Product" open />,
+      );
+
+      resourcePicker.setProps({onSelection});
+
+      expect(appBridgeResourcePickerMock.unsubscribe).toHaveBeenCalledTimes(1);
+      expect(appBridgeResourcePickerMock.subscribe).toHaveBeenCalledTimes(1);
+
+      const [
+        firstArg,
+        secondArg,
+      ] = appBridgeResourcePickerMock.subscribe.mock.calls[0];
+      expect(firstArg).toBe(AppBridgeResourcePicker.Action.SELECT);
+      secondArg(selectionPayload);
+      expect(onSelection).toHaveBeenCalledTimes(1);
+      expect(onSelection).toHaveBeenCalledWith(selectionPayload);
+    });
+
     it('subscribes the cancel callback', () => {
       mountWithAppBridge(
         <ResourcePicker resourceType="Product" open onCancel={noop} />,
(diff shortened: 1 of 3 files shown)


katzenbar:
@tmlayton - Updated PR to address feedback. Sorry about not getting back to this quickly, I caught a nasty cold last week!

src/components/ResourcePicker/tests/ResourcePicker.test.tsx
```diff
@@ -68,6 +68,27 @@ describe('<ResourcePicker />', () => {
...
+      const onSelection = jest.fn();
+      const selectionPayload = {selection: []};
+      const {resourcePicker} = mountWithAppBridge(
+        <ResourcePicker resourceType="Product" open />,
+      );
+
+      resourcePicker.setProps({onSelection});
+
+      expect(appBridgeResourcePickerMock.subscribe).toHaveBeenCalledTimes(1);
+
```
tmlayton: Let’s also check that unsubscribe was called once `expect(appBridgeResourcePickerMock.unsubscribe).toHaveBeenCalledTimes(1);`

src/components/ResourcePicker/tests/ResourcePicker.test.tsx
```diff
@@ -80,6 +101,25 @@ describe('<ResourcePicker />', () => {
...
+    it('calls the new cancel callback when changed', () => {
+      const onCancel = jest.fn();
+      const {resourcePicker} = mountWithAppBridge(
+        <ResourcePicker resourceType="Product" open />,
+      );
+
+      resourcePicker.setProps({onCancel});
+
+      expect(appBridgeResourcePickerMock.subscribe).toHaveBeenCalledTimes(1);
+
```
tmlayton: Same here, checking that we call `unsubscribe`.

katzenbar:
@tmlayton Updated

tmlayton:
Looks great, thanks for the contribution!

PR #1453, opened 2019-05-10
[Checkbox] Small fix in checkbox logic

keyfer:

Saw this code while running through some component logic, thought it was worth the quick fix. 

Simple fix to checkbox logic.

### DIFF
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -53,8 +53,8 @@ class Checkbox extends React.PureComponent<CombinedProps, never> {
 
   componentDidUpdate({disabled: prevDisabled}: Props) {
     if (
+      prevDisabled === true &&
       this.props.disabled &&
-      !prevDisabled !== true &&
       this.inputNode.current &&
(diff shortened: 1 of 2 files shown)


src/components/Checkbox/Checkbox.tsx
```diff
@@ -53,8 +53,8 @@ class Checkbox extends React.PureComponent<CombinedProps, never> {
 
   componentDidUpdate({disabled: prevDisabled}: Props) {
     if (
+      prevDisabled === false &&
       this.props.disabled &&
-      !prevDisabled !== true &&
```
danrosenthal: This makes my eyes go crossed. Good fix

keyfer: Obviously could just make this `prevDisabled`, but I'd rather avoid Type Coercion.

keyfer: Wait. I actually don't even think we need this line. 

`{disabled: prevDisabled}: Props` means we can access this with `this.props.disabled`, which is already the condition underneath.

PR #1466, opened 2019-05-14
Fix issue where tsx would copy the package.json file where it shouldn't

amrocha:

We can't update `polaris-react` in `web` because of this bug.

This issue was introduced when we updated Typescript from 3.1.6 to 3.2.4.

In 3.2.4, `tsc` includes the root directory's `package.json` file in the output build by default
This is a problem because the "files" key in the root `package.json` ends up filtering which files get included in the build output. Ideally the package.json file wouldn't be copied, but I couldn't easily figure out how to configure that, so I delete the `package.json` file before copying the `tsc` output to its proper folder.

## 

### How to 🎩

Run a build, and run the following command:
```
npm pack 2>&1 | tee packoutput.txt
```

And then inspect `packoutput.txt`. Check that the files included in the `/esnext/` folder look like this: https://unpkg.com/@shopify/polaris@3.13.0/esnext/

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ yarn add @shopify/polaris
 ```html
 <link
   rel="stylesheet"
-  href="https://sdks.shopifycdn.com/polaris/3.14.1-rc.2/polaris.min.css"
+  href="https://sdks.shopifycdn.com/polaris/3.14.1-rc.3/polaris.min.css"
 />
 ```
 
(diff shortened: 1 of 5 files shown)


BPScott:
I've worked out where that package.json was coming from and have stopped it from making its way into build-intermediate in the first place.

[It was listed as an explicit include in our tsconfig.json](https://github.com/Shopify/polaris-react/commit/8fcc093fdb085927fd253cc1d191e92fb24d42a5#diff-e5e546dd2eb0351f813d63d1b39dbc48R32) and I guess TS 3.2 now explicitly adds everything listed in your includes to your built output folder, even if you never reference a particular file.

Since the PR that added that in the first place we've changed our approach to how we inject version numbers into the project. Nothing ever imports package.json so it's fine to remove it from the includes array, which in turn stops it appearing in the build-intermediate and esnext folders.

amrocha:
That would do it, good job figuring this out everyone!

Issue #1363, opened 2019-04-25
[Checkbox] Disabled does not disable control


aishad:
# Issue summary
When setting the checkbox to disabled, it’s rendered as disabled (greyed/subdued colour) yet it can still be toggled. 

## Expected behavior
When set to disabled, controls should not be interacted with / should not be able to toggle. 

## Steps to reproduce the problem
https://polaris.shopify.com/components/forms/checkbox

![giphy](https://user-images.githubusercontent.com/5339802/56762462-e7d36d80-676d-11e9-9d7f-f7a8f944d45a.gif)

cc: @AndrewMusgrave

PR #1467, opened 2019-05-14
[Checkbox] Fix regression when disabled

AndrewMusgrave:

It’s unfortunate that we missed this in the `build-consumer` test. This seems to be a regression caused by not fully understanding #1363. In #1369 I made it so Checkbox cannot be disabled and checked. however, it seems we wanted to remove the ability to fire click events while disabled and keep the ability to change `Checkbox` checked state.

* removing unnecessary code
* moving logic from `Choice` to `Checkbox`
* updating tests 

### How to 🎩

* Simple playground to toggle `checked` and `disabled` states  ⬇️&
* Build-consumer and check the failing test in web

<details>

```jsx
import {Checkbox, Button} from '@shopify/polaris';

interface State {
  checked: boolean;
  disabled: boolean;
}

export default class Playground extends React.Component {
  state = {
    checked: true,
    disabled: false,
  };

  render() {
    const {checked, disabled} = this.state;

    return (
      <div>
        <Checkbox
          checked={checked}
          disabled={disabled}
          label="Basic checkbox"
          onChange={this.handleChange}
        />
        <Button
          onClick={() => this.setState(({disabled}) => ({disabled: !disabled}))}
        >
          toggle disabled
        </Button>
        <Button
          onClick={() => this.setState(({checked}) => ({checked: !checked}))}
        >
          toggle checked
        </Button>
      </div>
    );
  }

  handleChange = (value) => {
    this.setState({checked: value});
  };
}

```

</details>

### DIFF
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -45,27 +45,10 @@ const getUniqueID = createUniqueIDFactory('Checkbox');
 class Checkbox extends React.PureComponent<CombinedProps, never> {
   private inputNode = React.createRef<HTMLInputElement>();
 
-  componentDidMount() {
-    const {checked, disabled} = this.props;
-
-    if (checked && disabled) this.handleInput();
-  }
-
-  componentDidUpdate({disabled: prevDisabled}: Props) {
-    if (
-      prevDisabled === true &&
-      this.props.disabled &&
-      this.inputNode.current &&
-      this.inputNode.current.checked
-    ) {
-      this.handleInput();
-    }
-  }
-
   handleInput = () => {
-    const {onChange, id} = this.props;
+    const {onChange, id, disabled} = this.props;
 
-    if (onChange == null || this.inputNode.current == null) {
+    if (onChange == null || this.inputNode.current == null || disabled) {
       return;
     }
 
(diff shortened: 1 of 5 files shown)


src/components/Checkbox/Checkbox.tsx
```diff
@@ -45,27 +45,10 @@ const getUniqueID = createUniqueIDFactory('Checkbox');
 class Checkbox extends React.PureComponent<CombinedProps, never> {
   private inputNode = React.createRef<HTMLInputElement>();
 
-  componentDidMount() {
```
AndrewMusgrave: This is the logic that caused the regression

src/components/Checkbox/tests/Checkbox.test.tsx
```diff
@@ -165,6 +158,18 @@ describe('<Checkbox />', () => {
       );
       expect(element.find('input').prop('disabled')).toBeFalsy();
     });
+
+    it('can change values when disabled', () => {
```
AndrewMusgrave: Added this test to prevent regression in the future

src/components/Choice/Choice.tsx
```diff
@@ -35,20 +35,14 @@ export default function Choice({
   helpText,
   onClick,
 }: Props) {
-  function handleClick() {
```
AndrewMusgrave: I feel like this logic is out of scope for the `Choice` component, so it was moved to the `handleInput ` function in `Checkbox`

AndrewMusgrave:
@danrosenthal If this looks good to you and the logic makes sense / 🎩 / review are ✅ , feel free to merge this so you can release a patch

PR #1473, opened 2019-05-14
deprecated `react-utilties`

michelleyschen:

`react-utilties` is now [deprecated](https://github.com/Shopify/react-utilities/pull/34)
This PR remove the use of this library by doing the following
 
-  Replace `className` & `variationName` from `@shopify/react-utilities/styles` with `@shopify/css-utilities`
- Replace `ReactComponent` from `@shopify/react-utilities` with `React.ComponentType`
- Move `wrapWithComponent`, `isElementOfType` and `elementChildren` &  from `@shopify/react-utilities/styles/components` to `src/utilities/components.tsx`
- Move `isServer`  from `@shopify/react-utilities/styles/target` to `src/utilities/target.ts`

## 

### How to 🎩

<details>

```jsx

interface State {}

  render() {
    return (
      <Page title="Playground">
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/AppProvider/utilities/createAppProviderContext/tests/createAppProviderContext.test.tsx
+++ b/src/components/AppProvider/utilities/createAppProviderContext/tests/createAppProviderContext.test.tsx
@@ -1,6 +1,5 @@
 import * as React from 'react';
 import * as appBridge from '@shopify/app-bridge';
-import * as targets from '@shopify/react-utilities/target';
 import createAppProviderContext, {
   setClientInterfaceHook,
 } from '../createAppProviderContext';
@@ -17,19 +16,12 @@ jest.mock('../../Link', () => ({
   __esModule: true,
 }));
 
-const actualIsServer = targets.isServer;
-
-function mockIsServer(value: boolean) {
-  (targets as any).isServer = value;
-}
-
 describe('createAppProviderContext()', () => {
   const createAppSpy: jest.SpyInstance<any> = jest.spyOn(appBridge, 'default');
   const Intl: jest.Mock<{}> = require.requireMock('../../Intl').default;
   const Link: jest.Mock<{}> = require.requireMock('../../Link').default;
 
   afterEach(() => {
-    mockIsServer(actualIsServer);
     createAppSpy.mockReset();
     Intl.mockReset();
     Link.mockReset();
--- a/src/components/Avatar/tests/Avatar.test.tsx
+++ b/src/components/Avatar/tests/Avatar.test.tsx
@@ -1,14 +1,8 @@
 import * as React from 'react';
-import * as targets from '@shopify/react-utilities/target';
 import {mountWithAppProvider, trigger} from 'test-utilities';
 import {Image} from 'components';
+
 import Avatar from '../Avatar';
-
-const actualIsServer = targets.isServer;
-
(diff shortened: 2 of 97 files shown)


src/components/AppProvider/utilities/createAppProviderContext/tests/createAppProviderContext.test.tsx
```diff
@@ -1,7 +1,8 @@
 import * as React from 'react';
 import * as appBridge from '@shopify/app-bridge';
 import {noop} from '@shopify/javascript-utilities/other';
-import * as targets from '@shopify/react-utilities/target';
+
+import * as targets from '../../../../../utilities/target';
```
BPScott: `import * as ....` is [on its way out to enable compiling TS with babel directly](https://github.com/Shopify/web/pull/13571).

I know this isn't strictly in the scope of this PR but would you be able to look at working out how we can avoid doing  an `import * as targets` for the tests that mock the result of isServer?

src/components/Avatar/tests/Avatar.test.tsx
```diff
@@ -1,7 +1,8 @@
 import * as React from 'react';
-import * as targets from '@shopify/react-utilities/target';
 import {mountWithAppProvider, trigger} from 'test-utilities';
 import {Image} from 'components';
+
+import * as targets from '../../../utilities/target';
```
BPScott: Can we dodge this `import *`?

src/components/Portal/tests/Portal.test.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from 'react';
-import * as targets from '@shopify/react-utilities/target';
 import {mountWithAppProvider} from 'test-utilities';
+import * as targets from '../../../utilities/target';
```
BPScott: Can we dodge this `import *`?

src/components/Portal/tests/Portal.test.tsx
```diff
@@ -12,38 +11,21 @@ const {
...
-  });
-
   describe('children', () => {
     it('get passed into the portal creation method', () => {
       const children = <div />;
       mountWithAppProvider(<Portal>{children}</Portal>);
       expect(createPortalSpy).toHaveBeenCalledWith(children, expect.anything());
     });
-
-    it('dont get rendered when the env is the server', () => {
```
michelleyschen: the `isServer` check is no longer part of the `Portal` component?

danrosenthal:
Looking at our consumer build tests, and it appears this might be responsible for failures there: https://buildkite.com/shopify/polaris-react/builds/729#cb6d7e9a-b2ef-40a9-864a-73867eb586ed

cc: @bpscott

BPScott:
Looks like all of that is down to a component in web trying to import from react-utilities. Web doesn’t have an explicit dependency on react-utilities, and now that it’s no longer available transitively it’s exploding. This is why projects should explicitly state their deps rather than rely on transitive dependencies.

Solution is to add an explicit dependency on react-utilities in web
Longer term solution is to fixup web so it doesn't depend on react-utilities

michelleyschen:
This is resolved now. 
https://buildkite.com/shopify/polaris-react/builds/733

PR #1474, opened 2019-05-14
Remove dependency on @shopify/images

BPScott:
This was never needed at runtime, we only used it for a trivial type
that we can easily duplicate (and we'll remove the need for that type in
v4)

We don't need it at build time as sk handles our svg optimization

To test:

- Icon unit tests
- Search for `@shopify/images` and see the only mentions are in the yarn.lock files of our examples.

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -4,9 +4,6 @@
 
 const path = require('path');
 
-const {
-  svgOptions: svgOptimizationOptions,
-} = require('@shopify/images/optimize');
 const postcssShopify = require('postcss-shopify');
 
 // Use the version of webpack-bundle-analyzer (and other plugins/loaders) from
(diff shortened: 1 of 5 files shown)


src/components/Icon/Icon.tsx
```diff
@@ -163,6 +162,11 @@ const COLORS_WITH_BACKDROPS = [
   'inkLighter',
 ];
 
+interface SVGSource {
```
BPScott: Taken from https://github.com/Shopify/images/blob/master/index.d.ts

.storybook/webpack.config.js
```diff
@@ -4,9 +4,6 @@
 
 const path = require('path');
 
-const {
```
BPScott: Turns out we missed deleting this line when removing some of our icon loading logic a while back.

BPScott: Turns out we missed deleting this line when removing some of our icon loading logic a while back.

PR #1475, opened 2019-05-14
[Breakpoints] fix breakpoint off by 1

danrosenthal:
While tophatting a change in web to replace the web implementation of `Sheet` with the Polaris implementation, I saw that the breakpoint in the JavaScript is wrong, as navigation should be displayed inclusive of 768px, but not 769px. This results in a sheet which, for one pixel at `769px` wide appears as the desktop sheet but functions as the mobile sheet (closing towards the bottom). This fixes that issue.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -16,6 +16,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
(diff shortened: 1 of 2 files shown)


danrosenthal:
cc: @keyfer @asmockler 🤦‍♂

danrosenthal:
I think this is actually wrong. Hang tight

danrosenthal:
The JavaScript is actually what needed to be changed here. This has been broken for a long time and we just weren't noticing.

PR #1476, opened 2019-05-14
[Navigation] Remove deprecated UserMenu option for styleguide

### LINKED ISSUE #1470: Update Navigation.UserMenu example to align with deprecation notice
**Example to update**
https://polaris.shopify.com/components/navigation/navigation

**Deprecation notice**
https://github.com/Shopify/polaris-react/blob/master/src/components/Navigation/components/UserMenu/UserMenu.tsx#L34

yourpalsonja:

Fixes #1470  

- Removes `UserMenu` option and shortens description now that this sub-component has been deprecated.

## 

### How to 🎩

### 🎩 checklist

### DIFF
--- a/src/components/Navigation/README.md
+++ b/src/components/Navigation/README.md
@@ -171,17 +171,6 @@ Action allows a complementary icon-only action to render next to the section tit
 
 The user menu component displays the current user’s avatar and name, and actions that are related to the current logged in user. The menu is displayed at the top of the navigation sidebar at small screen sizes. The user menu can display any messages that the current user has available to read.
 
-### User menu properties
-
-| Prop           | Type                    | Description                                                                                             |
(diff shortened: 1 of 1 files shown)


kaelig:
If people can still technically use it, should we keep it documented (until v4) and flag it clearly as deprecated?

yourpalsonja:
@kaelig, do you mean keep the props, starting here on [line 147](https://github.com/Shopify/polaris-react/pull/1476/files#diff-4466766b5c445c910b7e1abc6f1b5f13L174)? 

Removing the example, at least, seems like the right thing to do!

kaelig:
I suppose both examples and props could be kept?

I don't have enough context on the importance of this particular subcomponent, but in most cases, I'd be in favor of communicating clearly to consumers that something is deprecated (flickering the lights before the darkness) before completely removing docs and examples.

PR #1478, opened 2019-05-15
Remove all usages of importing noop functions

BPScott:
Define them inline to move us towards removing our dependency on
`@shopify/javascript-utilities`, which is going to be deprecated eventually


danrosenthal:
Is this ready for review @BPScott?

BPScott:
Yep, this is ready for review

amrocha:
Why not keep a noop function in a test utilities file?

BPScott:
> Why not keep a noop function in a test utilities file?

As Andrew says it's less code to write a noop function than to write the import statement to pull in said utility.

PR #1482, opened 2019-05-15
[v4] Add usePolaris/AppBridge 

AndrewMusgrave:

Preparing for hooks

* create a new directory for hooks `src/hooks`
* create `usePolaris` hook for internal use
* create `useAppBridge` hook for external use
* 🎉 tests 🎉 

### 🎩 

Trust in the powers of tests or try the playground - you'll need credentials for testing `useAppBridge`

<details>

```jsx
import {Page, AppProvider, Card} from '../src';
import {usePolaris, useAppBridge} from '../src/hooks';

interface State {}

const credentials = {
  apiKey: undefined,
  shopOrigin: undefined,
}

export default function() {
  const {apiKey, shopOrigin} = credentials;
  return (
    <AppProvider apiKey={apiKey} shopOrigin={shopOrigin}>
      <Page title="Playground">
        <Card title="usePolaris">
          {JSON.stringify(usePolaris(), null, 2)}
        </Card>

        <Card title="useAppBridge">
          {!apiKey && !shopOrigin ? "You'll need real credentials or this is undefined :shocked:" : '' }
          {JSON.stringify(useAppBridge())}
        </Card>
      </Page>
    </AppProvider>
  );
}
```

</details>

### DIFF
--- a/src/components/index.ts
+++ b/src/components/index.ts
@@ -9,6 +9,7 @@ export {
   default as AppProvider,
   Props as AppProviderProps,
   createAppProviderContext,
+  AppProviderContext,
   createPolarisContext,
   withAppProvider,
   WithAppProviderProps,
@@ -257,6 +258,7 @@ export {
   default as ThemeProvider,
   Props as ThemeProviderProps,
   ThemeProviderContextType,
+  ThemeProviderContext,
   createThemeContext,
 } from './ThemeProvider';
 
@@ -306,3 +308,5 @@ export {default as Indicator, Props as IndicatorProps} from './Indicator';
 export {default as withContext} from './WithContext';
 
 export {default as withRef, WithRef} from './WithRef';
+
+export {PolarisContext} from './types';
--- a/src/hooks/use-app-bridge/use-app-bridge.tsx
+++ b/src/hooks/use-app-bridge/use-app-bridge.tsx
@@ -1,0 +1,10 @@
+import * as React from 'react';
+import {AppProviderContext} from '../../components';
+
+function useAppBridge() {
+  const {appBridge} = React.useContext(AppProviderContext);
+
+  return appBridge;
+}
+
+export default useAppBridge;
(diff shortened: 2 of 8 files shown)


src/components/index.ts
```diff
@@ -9,6 +9,7 @@ export {
   default as AppProvider,
```
AndrewMusgrave: Need to use these in `src/hooks` and we can't dive into component folders from outside components

src/hooks/use-app-bridge/use-app-bridge.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
```
AndrewMusgrave: We'll want this to expose `appBridge` rather than exporting `usePolaris`

src/hooks/use-polaris/tests/use-polaris.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  it('returns context', () => {
+    let context;
+    function Component() {
+      context = usePolaris();
+      return null;
+    }
+
+    mountWithAppProvider(<Component />);
+    expect(JSON.stringify(context)).toEqual(
```
AndrewMusgrave: Feel odd about this test. But we've done it before and I can't think of a better alternative.

BPScott:
I'm a little late to the party on this one but I've got a thought:

Hooks should be pretty small and self contained - each hook would never sprawl across multiple files (unlike components that may have partner scss files and things like that). As such I don't think there's any value in creating a "folder per hook" hierarchy as there's only ever going to be one file (+ the test) for each hook within it.

With that in mind I don't think there's any value in having `src/hooks/use-polaris/use-polaris.tsx` it adds depth for no value, I think we should just put hooks directly in the hooks folder e.g.: `src/hooks/use-polaris.ts` and its test in `src/hooks/tests/use-polaris.ts`.

PR #1488, opened 2019-05-15
[Storybook] Upgrade dependencies to improve search

kaelig:
In v5.1.0-alpha.14, Storybook upgraded from using https://npmjs.com/package/fuzzy-search to https://fusejs.io to power search in the sidebar.

This improves the search experience by ~124566.9%.

This PR upgrades to the latest working alpha of Storybook (`v5.1.0-alpha.39`) that includes this change, as unfortunately, upgrading to the latest beta fails due to this issue: https://github.com/storybooks/storybook/issues/6764


alex-page:
@kaelig the percy issue isn't showing locally. Just pinging you as I think this is safe to merge.

kaelig:
I've asked Percy about this, in case they know what is causing the issue. Will get back to you as soon as I know more!

kaelig:
tl;dr: the visual regression in Percy is caused by a possible bug in Storybook where the browser automatically scrolls to the middle of the Card's component "All examples" story.

---

As stated earlier, I got in touch with Percy in case they had seen this before.

They asked me to run some debugging steps, which I did:

>I ran `yarn run build-storybook`, it generated a `storybook-static` folder. The output is as expected. We have a custom script to build storybook, so just in case… I ran it too (`yarn storybook:build`). Same: output looks good.
>Then, I loaded the story in a browser and found something different between master and my PR:
>- Master: https://polaris-react.herokuapp.com/?path=/story/all-components-card--all-examples -> Scroll position is at the top of the page
>- PR: https://polaris-react-pr-1488.herokuapp.com/?path=/story/all-components-card--all-examples -> Scrolls to "Card with separate header" on load
>
>Could it have to do with the new version of Storybook? Is this something Percy should handle differently?

Here's Percy's answer:

>Hey Kaelig,
>I cloned the repo and ran that branch on a Percy project that had both Chrome and Firefox enabled and it looks like it's just Chrome that's experincing that problem, so at least that narrows it down for us. 
>We're going to keep investigating to see what's causing this issue. 🤔

---

I'll merge this branch as it definitely improves the developer experience. Hopefully Percy or Storybook fixes this for us.

Issue #1484, opened 2019-05-15
[v4] Remove withRef / withContext / compose


AndrewMusgrave:
### withRef

- [x] DropZone.FileUpload
- [x] UnstyledLink

### withContext

- [x] Banner
- [x] ContextualSaveBar
- [x] DropZone.FileUpload
- [x] Loading
- [x] Navigation.Item
- [x] ResourceList.FilterControl
- [x] ResourceList.Item
- [x] Scrollable.ScrollTo
- [x] Toast
- [x] TopBar.UserMenu

### Delete

- [x] withRef
- [x] withContext
- [x] react-compose
- [x] WithContextTypes interface
- [x] dependency on hoist-non-react-statics

PR #1491, opened 2019-05-15
[v4][FileUpload] Remove withContext/withRef

AndrewMusgrave:

Part of #1484

I'm converting FileUpload to a functional component and removing `withRef` / `withContext`

## 

### How to 🎩

### DIFF
--- a/src/components/DropZone/components/FileUpload/FileUpload.tsx
+++ b/src/components/DropZone/components/FileUpload/FileUpload.tsx
@@ -2,150 +2,98 @@ import * as React from 'react';
 import {classNames} from '@shopify/react-utilities/styles';
 import {DragDropMajorMonotone} from '@shopify/polaris-icons';
 
-import {WithContextTypes} from '../../../../types';
-import compose from '../../../../utilities/react-compose';
 import capitalize from '../../../../utilities/capitalize';
 
-import {withAppProvider, WithAppProviderProps} from '../../../AppProvider';
 import Link from '../../../Link';
 import Icon from '../../../Icon';
 import Stack from '../../../Stack';
 import Button from '../../../Button';
 import Caption from '../../../Caption';
 import TextStyle from '../../../TextStyle';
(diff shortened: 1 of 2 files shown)


src/components/DropZone/components/FileUpload/FileUpload.tsx
```diff
@@ -2,150 +2,98 @@ import * as React from 'react';
...
-import withRef from '../../../WithRef';
+
 import {fileUpload, imageUpload} from '../../images';
-import DropZoneContext, {DropZoneContextType} from '../../context';
 
-import styles from './FileUpload.scss';
+import DropZoneContext from '../../context';
+import {usePolaris} from '../../../../hooks';
 
-export interface State {
```
AndrewMusgrave: We have no need for state. See below 👀

src/components/DropZone/components/FileUpload/FileUpload.tsx
```diff
@@ -2,150 +2,98 @@ import * as React from 'react';
...
-          )}
-          <Button>{actionTitle}</Button>
+function FileUpload(props: Props) {
+  const {
+    intl: {translate},
+  } = usePolaris();
+  const {size, type} = React.useContext(DropZoneContext);
+  const suffix = capitalize(type);
+  const {
+    actionTitle = translate(`Polaris.DropZone.FileUpload.actionTitle${suffix}`),
```
AndrewMusgrave: We use the provided actionTitle/Hint and default it to what was previously the default on state

dleroux: I wonder why we had done that?

AndrewMusgrave: Seems like it was to avoid string translation however object lookup is inexpensive so I'm not sure why, or this approach didn't occur when developing the component? 🤷‍♂

PR #1499, opened 2019-05-15
[v4][Scrollable.ScrollTo] Remove withContext

AndrewMusgrave:

Part of #1484

I'm converting Scrollable.ScrollTo to a functional component and removing `withContext`

## 

### How to 🎩


src/components/Scrollable/components/ScrollTo/tests/ScrollTo.test.tsx
```diff
@@ -18,4 +18,20 @@ describe('<Scrollable.ScrollTo />', () => {
 
     expect(spy).toHaveBeenCalled();
   });
+
+  it("does not call scrollToPosition when it's undefined", () => {
```
AndrewMusgrave: Coverage was failing without this

PR #1501, opened 2019-05-15
[v4][UnstyledLink] Remove withRef

AndrewMusgrave:

Part of #1484

I'm converting UnstyledLink to a functional component and removing `withContext`

## 

### How to 🎩

### DIFF
--- a/src/components/Navigation/components/Item/tests/Item.test.tsx
+++ b/src/components/Navigation/components/Item/tests/Item.test.tsx
@@ -279,7 +279,10 @@ describe('<Nav.Item />', () => {
         {location: 'bar'},
       );
 
-      item.find(UnstyledLink).simulate('click');
+      item
+        .find(UnstyledLink)
+        .find('a')
+        .simulate('click');
       expect(spy).toHaveBeenCalledTimes(1);
     });
 
@@ -310,7 +313,10 @@ describe('<Nav.Item />', () => {
         <Item label="some label" url="foo" disabled={false} />,
         {...context},
       );
-      item.find(UnstyledLink).simulate('click');
+      item
+        .find(UnstyledLink)
+        .find('a')
+        .simulate('click');
       expect(context.onNavigationDismiss).toHaveBeenCalledTimes(1);
     });
 
@@ -338,6 +344,7 @@ describe('<Nav.Item />', () => {
       item
         .find(UnstyledLink)
         .last()
+        .find('a')
         .simulate('click');
       expect(context.onNavigationDismiss).toHaveBeenCalledTimes(1);
     });
--- a/src/components/UnstyledLink/UnstyledLink.tsx
+++ b/src/components/UnstyledLink/UnstyledLink.tsx
@@ -1,9 +1,7 @@
 import * as React from 'react';
 import {ReactComponent} from '@shopify/react-utilities/types';
 import {unstyled} from '../shared';
-import {withAppProvider, WithAppProviderProps} from '../AppProvider/utilities';
-import compose from '../../utilities/react-compose';
-import withRef from '../WithRef';
+import {usePolaris} from '../../hooks';
 
 export interface Props extends React.HTMLProps<HTMLAnchorElement> {
   /** A destination to link to */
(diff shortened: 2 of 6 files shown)


src/hooks/use-app-bridge/use-app-bridge.tsx
```diff
@@ -1,5 +1,6 @@
 import * as React from 'react';
-import {AppProviderContext} from '../../components';
+// eslint-disable-next-line shopify/strict-component-boundaries
```
AndrewMusgrave: I was running into circular dependency issues. I couldn't find the culprit so I improved from the file directly and that worked 👌 Do you have any thoughts on this? @BPScott

src/hooks/use-polaris/use-polaris.tsx
```diff
@@ -1,9 +1,9 @@
 import * as React from 'react';
-import {
-  AppProviderContext,
-  ThemeProviderContext,
-  PolarisContext,
-} from '../../components';
+/* eslint-disable shopify/strict-component-boundaries */
```
AndrewMusgrave: ^

src/components/Navigation/components/Item/tests/Item.test.tsx
```diff
@@ -279,7 +279,10 @@ describe('<Nav.Item />', () => {
         {location: 'bar'},
       );
 
-      item.find(UnstyledLink).simulate('click');
+      item
```
AndrewMusgrave: These tests broke when `React.memo` was used instead of `PureComponent`. [Simulate will use the root node](https://airbnb.io/enzyme/docs/api/ReactWrapper/simulate.html) which I'm guessing `React.memo` interferes with

src/hooks/use-polaris/use-polaris.tsx
```diff
@@ -1,9 +1,9 @@
...
+import {AppProviderContext} from '../../components/AppProvider';
+import {ThemeProviderContext} from '../../components/ThemeProvider';
+import {PolarisContext} from '../../components/types';
+/* eslint-enable shopify/strict-component-boundaries */
 
 function usePolaris() {
```
dleroux: I know this is not part of this PR, but did we discuss calling this `usePolarisContext`?

src/components/UnstyledLink/UnstyledLink.tsx
```diff
@@ -18,28 +16,20 @@ export interface Props extends React.HTMLProps<HTMLAnchorElement> {
...
-export class UnstyledLink extends React.PureComponent<CombinedProps, never> {
-  render() {
-    const {polaris, external, url, ...rest} = this.props;
-    if (polaris && polaris.link) {
-      const LinkComponent = polaris.link.getLinkComponent();
-      if (LinkComponent) {
-        const {polaris, ...rest} = this.props;
-        return <LinkComponent {...unstyled.props} {...rest} />;
-      }
+export default React.memo(function UnstyledLink(props: Props) {
```
dleroux: Is it fair to say that `PureComponents` should make use of `React.memo`?

src/hooks/use-app-bridge/use-app-bridge.tsx
BPScott: Circular dependencies and strict-component-boundary violations are usually a sign that you've got something that is a dependency of multiple things nested deeply inside part of the components folder and then things get funky when they start to be interdepent.

In the past I've found the solution is to pull up that common dependency to a higher level to avoid things importing from siblings. That removes the cicularity of dependencies between siblings as you end up with two siblings that both depend upon a 3rd file that's higher up the hierarchy instead of on each other.

---

I saw in https://github.com/Shopify/polaris-react/pull/1482 you added AppProviderContext, ThemeProviderContext and PolarisContext to src/components/index's exports. I think this is part of the root cause of this - you have something in hooks depending on something in components (namely this context), and something in components depending on this hook.

We don't want to expose these contexts as part of our public API - which adding them to src/components/index will do.

I think the solution to both of these problems (circular deps and accidentally making that public) is to move these global contexts that are consumed by both things in in /components/ and /hook/ out of components and into src/contexts/. At which point both this file, and components/AppProvider will `import {AppProviderContext} from 'ROOTPATHHERE/src/components/context';`.

AndrewMusgrave: I gave moving `AppProvinder/ThemeProvider` context up into `src/contexts` a try however I still ran into circular dependencies (different ones). I'll create an issue to discuss re-architecting app provider and it's utilities. Since some of it's utilities will need to be moved out as well, e.g. `StickyManager`, `Link`, etc.

PR #1502, opened 2019-05-15
[v4][Navigation.Item] Remove withRef

AndrewMusgrave:

Part of #1484

I'm converting Navigation.Item to a functional component and removing `withContext`

## 

### How to 🎩

### DIFF
--- a/src/components/Navigation/components/Item/tests/Item.test.tsx
+++ b/src/components/Navigation/components/Item/tests/Item.test.tsx
@@ -17,7 +17,7 @@ describe('<Nav.Item />', () => {
     matchMedia.restore();
   });
 
-  it('sets expanded to false on resize when !navigationBarCollapsed and location does not match', () => {
+  it.skip('sets expanded to false on resize when !navigationBarCollapsed and location does not match', () => {
     const spy = jest.fn();
     matchMedia.setMedia(() => ({addListener: spy}));
     const item = mountWithNavigationProvider(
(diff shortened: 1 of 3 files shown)


src/components/Navigation/components/Item/tests/Item.test.tsx
```diff
@@ -18,7 +18,8 @@ describe('<Nav.Item />', () => {
     matchMedia.restore();
   });
 
-  it('sets expanded to false on resize when !navigationBarCollapsed and location does not match', () => {
+  // eslint-disable-next-line jest/no-disabled-tests
+  it.skip('sets expanded to false on resize when !navigationBarCollapsed and location does not match', () => {
```
AndrewMusgrave: Unfortunately, this test uses a slightly hacky way of testing a resize (stores a reference to a previous call with jest) that doesn't work anymore in a functional component. I'm not quite sure how this can be tested with a functional component. I'm skipping the test for now and creating an issue for this.

alex-page: @AndrewMusgrave do you have an issue number?

AndrewMusgrave:
> There are some warnings from deprecation, not sure if that should be happening:
 It's all good, they haven't been removed yet 😄

src/components/Navigation/components/Item/tests/Item.test.tsx
AndrewMusgrave: #1585

PR #1503, opened 2019-05-15
[v4][ResourceList.Item] Remove withContext

AndrewMusgrave:

Part of #1484

I'm removing `withContext` in favor of a wrapper component

## 

### How to 🎩

### DIFF
--- a/src/components/ResourceList/components/Item/Item.tsx
+++ b/src/components/ResourceList/components/Item/Item.tsx
@@ -2,7 +2,6 @@ import React from 'react';
 import {HorizontalDotsMinor} from '@shopify/polaris-icons';
 import {classNames} from '@shopify/css-utilities';
 import {createUniqueIDFactory} from '@shopify/javascript-utilities/other';
-import compose from '@shopify/react-compose';
 import isEqual from 'lodash/isEqual';
 import {DisableableAction, WithContextTypes} from '../../../../types';
 import ActionList from '../../../ActionList';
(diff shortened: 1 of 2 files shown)


AndrewMusgrave:
A lot of the lacking coverage has been removed in master with simpler interaction states. So when `master` is merged into `version-4.0.0` it should be 👌

src/components/ResourceList/components/Item/Item.tsx
```diff
@@ -65,273 +64,258 @@ export type CombinedProps =
...
+      onKeyUp={handleKeypress}
+      testID="Item-Wrapper"
+      data-href={url}
+    >
+      {accessibleMarkup}
+      {containerMarkup}
+    </div>
+  );
+
+  function handleAnchorFocus() {
```
BPScott: Having a bunch of functions defined beneath the return feels odd to me. Can you pull these up to be towards the start of the function.

This is a functional component and callback functions that we define need to be wrapped in [useCallback](https://reactjs.org/docs/hooks-reference.html#usecallback). Otherwise a new callback function ends up being created every render, and when you pass that down to a child component you end up rerendering it - as its props have changed becase it's onWhatever handler is a different function to the one that was used before.

Wrapping these declarations in useCallback means the same function will be passed to the child component if given the same arguments, thus avoiding unneeded rerenders.

```js
const handleFocusedBlur  = useCallback(() => {
   setFocusState({focused: true, focusedInner: true});
  }, [setFocusState]);
```

See https://github.com/BPScott/ksw-quiz/commit/30b4c3f609f6eeae8edf550e466af129e785bf87 for an example of how I dealt with this in my little toy project for playing with react

AndrewMusgrave: I find putting the functions below cleans up the code, Tim does the same as well I think. However, if we use functional expressions we'll want to pull it up. Thanks for pointing this out, I was worried about it but wasn't aware there was a hook to solve it 🙏

PR #1514, opened 2019-05-16
[v4] Pull hooks up

BPScott:
Per my slightly late comment on #1482, there's no need for sub folders as each hook will always be a single file so having a nested directory structure doesn't offer any value

- Move `src/hooks/use-polaris/use-polaris.ts` to `src/hooks/use-polaris.ts`
- Move `src/hooks/use-polaris/tests/use-polaris.ts` to `src/hooks/tests/use-polaris.ts`
- Same for `use-app-bridge`

### DIFF
--- a/src/hooks/tests/use-app-bridge.test.tsx
+++ b/src/hooks/tests/use-app-bridge.test.tsx
@@ -1,8 +1,7 @@
 import * as React from 'react';
 import isEqual from 'lodash/isEqual';
-import {createPolarisContext} from '../../../components';
-import {mountWithAppProvider} from '../../../test-utilities/enzyme';
-
+import {createPolarisContext} from '../../components';
+import {mountWithAppProvider} from '../../test-utilities/enzyme';
 import useAppBridge from '../use-app-bridge';
 
 describe('useApp', () => {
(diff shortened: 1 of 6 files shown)


AndrewMusgrave:
[Web suggests](https://github.com/Shopify/web-foundation/blob/master/Best%20practices/React/Organization.md#apphooks-and-componenthooks) something similar to what we're doing now except without an index file, what do you think?

BPScott:
> [Web suggests](https://github.com/Shopify/web-foundation/blob/master/Best%20practices/React/Organization.md#apphooks-and-componenthooks) something similar to what we're doing now except without an index file, what do you think?

Thanks for the link, I think I saw that ages ago but forgot it existed.

I still don't think there's much value in having one hook per folder.

As part of a wider rethink of how we structure this stuff (i.e. thinking about pulling contexts up into a contexts folder) I quite like the context-based library idea - thought I'm going to call it concern-based as context already means something in react land. That doc suggests having a 'utilities' folder but we already have one of those as a general dumping ground and this feels a bit more focused.


Hypothetical idea:

```
/src/concerns/app-provider/context.ts (containing the context that use-polaris and the AppProvider bits use)
/src/concerns/app-provider/hooks.ts (containing use-polaris)
/src/concerns/app-provider/index.ts (exporting both the hook and the concern, the create utility).
```

Other stuff like the create utility function can live in there too.


This is kinda half-baked as I need to jump on a plane but what do you think?

BPScott:
Spoke to Andrew and the plan for this is go with subfolders in utilities.

PR #1515, opened 2019-05-16
Update to use default import for react and prop-types

BPScott:
`import * from React` and `import React` are functionally identical and
the latter is about to become the new recommendation (see https://github.com/Shopify/eslint-plugin-shopify/pull/262 and https://github.com/Shopify/web/pull/13571).

This will also allow us to import hooks without needing prefixes in v4, e.g.:
`import React, {useState} from 'react';`

- Replace `import * as React from 'react';` with `import React from 'react';`
- Replace `import * as PropTypes from 'prop-types';` with `import PropTypes from 'prop-types';`

---

We don't need to explicitly enable `esModuleInterop` just yet (which is used to handle libraries without a default export) as these two libraries both have a default export already. Which is just as well because currently explodes due to react-utilities doing something silly.

We can safely enable esModuleInterop once https://github.com/Shopify/polaris-react/pull/1473 removes the dependency on react-utilities.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -38,7 +38,7 @@ Fixes #0000 <!-- link to issue if one exists -->
 <summary>Copy-paste this code in <code>playground/Playground.tsx</code>:</summary>
 
 ```jsx
-import * as React from 'react';
+import React from 'react';
 import {Page} from '@shopify/polaris';
 
 interface State {}
--- a/.storybook/stories-from-readme.js
+++ b/.storybook/stories-from-readme.js
@@ -1,4 +1,4 @@
-import * as React from 'react';
+import React from 'react';
 import * as Polaris from '../src';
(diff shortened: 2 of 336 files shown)


BPScott:
@AndrewMusgrave Once this is merged in, we should merge master into the v4 branch and then open up a new PR to do the same for any new files that have been added as part of v4.

AndrewMusgrave:
As discussed over slack we'll want to aim this at `v4`  to prevent apps from breaking. Super stoked about this change, we'll get a lot of use out of it now that we can use hooks 😄

BPScott:
I'm reasonably confident that this won't break things that consume the compiled artifacts (index.js/index.es.js).

However apps that use SK (and thus do the in-app partial build thing) would require either esModuleInterop or allowSyntheticImports to be enabled in their tsconfig for this to work. While that *should* be an easy thing to enable and is the recommended config in the future it seems strange to ask them to do that for a minor release of polaris-react.

Going to close this and make this change in the v4 branch for the sake of least surprise.

PR #1528, opened 2019-05-17
Check that v4 changes are listed in the correct UNRELEASED file

kaelig:
This should tell the changelog bot that v4-related changes are to be made in the v4 branch.

@tmlayton Not sure this will work, as the bot may be getting its config from the `master` branch. What do you think?

### DIFF
--- a/.github/changelog.yml
+++ b/.github/changelog.yml
@@ -1,4 +1,4 @@
-filename: UNRELEASED.md
+filename: UNRELEASED-V4.md
 include:
   - /^config\//
   - /^documentation\//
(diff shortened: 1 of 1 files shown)


tmlayton:
Looks like this won’t work: https://github.com/probot/probot/issues/264

- `config` docs: https://probot.github.io/api/latest/classes/context.html#config
- `config` source: https://github.com/probot/probot/blob/e975bbda0049863f3f1c808a31133f5f15acbb52/src/context.ts#L142

kaelig:
Thanks for looking into it, this is sad.

Can we tell the bot not to run on certain branches?

tmlayton:
> Can we tell the bot not to run on certain branches?

Not sure – if not, we own the code for the bot so could just add that feature

PR #1530, opened 2019-05-17
Make i18n required

### LINKED ISSUE #1418: Document configuration of i18n on AppProvider for pulling in translations and remove english as default
Related to this comment: https://github.com/Shopify/polaris-react/pull/1358#issuecomment-488477477, we want to clearly document the approach for pulling in non-english translations to the `AppProvider`. We may consider making `i18n` a required prop, and removing the default English translations.

AndrewMusgrave:

Fixes #1418

* making `i18n` required can reduce the bundle size of consuming apps that provide translations
* two sets of translations won't have to be loaded (polaris translations + consumer provided translations)

* Make `i18n` required
* Remove translations currently used in polaris
* Update ResourceList.FilterControl context to generate a default label in the component
* Adding locales to our npm package
* Adding `i18n` where required
* Update app provider docs and add a section for using translations

### Todo

* add changelog


kaelig:
To ease the upgrade for users of v3, should we start giving them a heads-up by telling them this prop will be mandatory in v4 (using a console.log)?

(It'd be the inverse of a deprecation)

tmlayton:
> To ease the upgrade for users of v3, should we start giving them a heads-up by telling them this prop will be mandatory in v4 (using a console.log)?
> 
> (It'd be the inverse of a deprecation)

I think we should do this

AndrewMusgrave:
> To ease the upgrade for users of v3, should we start giving them a heads-up by telling them this prop will be mandatory in v4 (using a console.log)?
> 
> (It'd be the inverse of a deprecation)

I think we should only do this if we add an upgrade path in `v3` (we don't export our translations, and I don't see why we can't in v3), otherwise, it may be confusing to consumers.

.storybook/stories-from-readme.js
```diff
@@ -71,7 +72,7 @@ export function addPlaygroundStory(playgroundModule) {
 function AppProviderDecorator(story) {
   return (
     <div style={{padding: '8px'}}>
-      <Polaris.AppProvider>{story()}</Polaris.AppProvider>
+      <Polaris.AppProvider i18n={en}>{story()}</Polaris.AppProvider>
```
AndrewMusgrave: It would be cool if we had a select to toggle between languages in dev

package.json
```diff
@@ -167,6 +167,7 @@
     "esnext",
     "styles",
     "types",
+    "locales",
```
AndrewMusgrave: This adds locales to our package

src/components/AppProvider/README.md
```diff
@@ -38,7 +38,21 @@ The app provider component is required to use Polaris. Without it, the component
 AppProvider works by default without any additional options passed to it.
```
AndrewMusgrave: Most of the changes not in this file are adding `i18n` to app providers and translations to our test provider.

src/components/AppProvider/README.md
```diff
@@ -76,9 +90,9 @@ AppProvider works by default without any additional options passed to it.
 </AppProvider>
 ```
 
-### With i18n object
```
AndrewMusgrave: I don't think we really get much default from having `object` here

src/components/ResourceList/components/FilterControl/FilterControl.tsx
```diff
@@ -44,11 +44,15 @@ export class FilterControl extends React.Component<CombinedProps> {
       polaris: {intl},
       context: {selectMode, resourceName},
     } = this.props;
+    const filterResourceName = resourceName || {
```
AndrewMusgrave: The default was originally initialized in context. These changes move it to the render method

src/components/ResourceList/components/FilterControl/components/FilterCreator/FilterCreator.tsx
```diff
@@ -18,7 +18,7 @@ export interface Props {
     singular: string;
     plural: string;
   };
-  disabled: boolean;
+  disabled?: boolean;
```
AndrewMusgrave: Part of cleaning up default context

src/components/ResourceList/context.tsx
```diff
@@ -1,27 +1,18 @@
 import * as React from 'react';
```
AndrewMusgrave: ^

src/test-utilities/enzyme.tsx
```diff
@@ -104,7 +105,7 @@ export function mountWithAppProvider<P>(
 ): PolarisContextReactWrapper<P, any> {
   const {context: ctx = {}} = options;
 
-  const polarisDefault = createPolarisContext();
+  const polarisDefault = createPolarisContext({i18n: translations});
```
AndrewMusgrave: default translations for out tests

AndrewMusgrave:
This is ready for the first round of reviews/thoughts, is this something we want to move forward with?

.storybook/stories-from-readme.js
BPScott: addon-contexts should be able to let us do that https://www.npmjs.com/package/@storybook/addon-contexts

AndrewMusgrave: Awesome 🎉

danrosenthal:
Is the codecov failure a false negative?

src/components/AppProvider/README.md
```diff
@@ -76,9 +90,9 @@ AppProvider works by default without any additional options passed to it.
...
+### With i18n
 
-With an `i18n` object, the app provider component will override default English translations.
+With an `i18n`, `AppProvider` will provide these translations to polaris components. For usage see [using translations](/components/structure/app-provider#using-translations)
```
danrosenthal:
```suggestion
With an `i18n`, `AppProvider` will provide these translations to polaris components. See [using translations](/components/structure/app-provider#using-translations)
```

✨ simplicity ✨

src/components/AppProvider/README.md
```diff
@@ -391,6 +429,19 @@ ReactDOM.render(
 
 ---
 
+## Using translations
+
+When using Polaris, you are able to import translations from over 10 different langauges and consume them through the `i18n` prop.
```
danrosenthal: Let's either list out the available languages here, or omit the number entirely in favor of something like `all languages supported by the core Shopify product`

AndrewMusgrave:
> Are we concerned with the increase in bundle size at all? 5K is non-negligible.

Shrinkray only compares against master so the 5k is from updating `react`/`react-dom`. From what I understand shrink ray uses web pack and creates our bundle and uses that to compare sizes. But it doesn't compare package sizes, so shrink ray won't catch size increases from adding translations since we don't import them into our code. Including translations in the package will increase its size, however, it's tree-shakable and will not increase bundle size for consuming app, and reduce size for apps not currently using our translations like web. Is that correct @BPScott ?

src/components/AppProvider/README.md
AndrewMusgrave: I can see us forgetting to update the docs when new languages are added, I'm in favor of the latter.

BPScott:
I think you also need to update path in the translation.yml too https://github.com/Shopify/polaris-react/blob/master/translation.yml


> Including translations in the package will increase its size, however, it's tree-shakable and will not increase bundle size for consuming app, and reduce size for apps not currently using our translations like web. Is that correct @BPScott ?

Pretty much yes.

Can you talk us through how apps that consume the compiled build (i.e. not using sewing-kit) would set a locale's information? Demoing that in the examples might be the best way to see that

AndrewMusgrave:
> Can you talk us through how apps that consume the compiled build (i.e. not using sewing-kit) would set a locale's information? Demoing that in the examples might be the best way to see that

Even apps that use sewing-kit will have to determine locale on their own. If I recall, web uses the graphql api. There are a lot of ways to determine locale, many of browser dependent. Since we tend to keep example basic and don't want an example that only works part of the time, we should probably leave this up to the consumer to implement.

BPScott:
> Even apps that use sewing-kit will have to determine locale on their own. If I recall, web uses the graphql api.

I was thinking more like "I know I need to load the english translations, how do i access the given locale's json file and inject it into the AppProvider". Sorry for not being clear :)

AndrewMusgrave:
@BPScott I've [already added](https://github.com/Shopify/polaris-react/pull/1530/files#diff-7a936f09d89747421f1ed11ca7e51421R432) an example around importing locales and passing them to `AppProvider` 😄

AndrewMusgrave:
Just rebased with `version-4.0.0` and fixed all the conflicts/errors that came along with it. Hoping to merge this later today unless anyone has any other comments

PR #1534, opened 2019-05-20
Deprecate passing a stringy name of a bundled icon into <Icon source>

BPScott:

To shrink polaris-react we should stop bundling icons, and instead make people explicitly reference the icon they need.

This PR deprecates passing a string into `<Icon source />` and instead tells people to import an icon from polaris-icons instead. e.g. instead of 

```jsx
const x = <Icon source="add" />
```

do:

```jsx
import {PlusMinor} from '@shopify/polaris-icons';
const x = <Icon source={PlusMinor} />
```

The groundwork to enable this - updating our own code to import from polaris-icons was done in #1196 and #1297.

This PR also tightens the type for `<Button icon>` from `ReactNode|IconSource` to `ReactElement|IconSource` as ReactNode is far too broad - we only want people passing in some kind of html element, we don't want to handle strings or anything else that a ReactNode could be. I consider this a bug fix change as this would have never resulted in a useable output in the first place and makes typescript point out when you're about to do something undesired.

### How to 🎩

Check that deprecation notices are triggered when you do `<Icon source="add" />`.

<details>

```jsx
import {Icon, Page, TextField} from '../src';

interface State {}

  render() {
    function handleChange() {}

    return (
      <Page title="Playground">
        HELLO
        <Icon source="add" />
        <TextField label="hi" value="FOO" onChange={handleChange} clearButton />
      </Page>
    );
  }
}
```

</details>

### DIFF
--- a/src/components/Icon/Icon.tsx
+++ b/src/components/Icon/Icon.tsx
@@ -238,6 +238,11 @@ function Icon({
     contentMarkup = <div className={styles.Placeholder} />;
   } else if (isBundledIcon(source)) {
     SourceComponent = BUNDLED_ICONS[source];
+    const componentImportName = importForStringIcon(source);
+    // eslint-disable-next-line no-console
+    console.warn(
+      `Deprecation: passing the string "${source}" into <Icon source> is deprecated and will be removed in the next major version. Pass in the "${componentImportName}" React Component from the @shopify/polaris-icons package instead.`,
+    );
     contentMarkup = <SourceComponent {...defaultIconProps} />;
   } else if (typeof source === 'function') {
     const sourceElement = <SourceComponent {...defaultIconProps} />;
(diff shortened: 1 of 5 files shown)


src/components/Icon/Icon.tsx
```diff
@@ -238,6 +238,11 @@ function Icon({
     contentMarkup = <div className={styles.Placeholder} />;
   } else if (isBundledIcon(source)) {
     SourceComponent = BUNDLED_ICONS[source];
+    const componentImportName = importForStringIcon(source);
+    // eslint-disable-next-line no-console
+    console.warn(
+      `Deprecation: passing the string "${source}" is deprecated and will be removed in the next major version. Pass in the "${componentImportName}" React Component from the @shopify/polaris-icons package instead.`,
```
danrosenthal: For the sake of consistency, I think we should continue translating these until we make the blanket decision not to

tmlayton: > until we make the blanket decision not to

I thought we made the decision to not translate these? I think it is largely useless as our Polaris docs are not translated.

danrosenthal: 10-4, I think I misinterpreted that as saying we should keep translating these, and work to translate our docs.

amrocha:
Oh, can you make sure that none of the examples we have trigger the deprecation warning? That'd be awkward...

BPScott:
> Oh, can you make sure that none of the examples we have trigger the deprecation warning? That'd be awkward...

I've rebased on latest master now that #1561 is merged and I've confirmed no deprecation warnings are raised in all examples.

PR #1535, opened 2019-05-20
[Color system] deprecate warning banner status in favor of new attention banner status

danrosenthal:
Partially resolves https://github.com/Shopify/polaris-react/issues/1430, related to https://github.com/Shopify/polaris-react/issues/1489 and https://github.com/Shopify/polaris-react/issues/1433


danrosenthal:
@AndrewMusgrave any idea why this `Banner` `focus` test would start failing after this change?

AndrewMusgrave:
Using `withAppProvider` wraps `Banner` so any `ref` that gets placed on the `Banner` component will actually be placed on `withAppProvider`. You can try using the `withRef` higher order component however I'm not sure the `forwardRef` API will work in this situation. A preferred approach would be to target `v4` and use `usePolaris` since we're removing the `withRef` util in v4. If `withRef` doesn't work and you can't target `v4` you can just use the app provider context directly rather than the HoC

BPScott:
Or don't translate the deprecation notices :p

PR #1538, opened 2019-05-20
[Color system] update the info variant of the badge component to use teal instead of blue for greater system consistency

danrosenthal:
|Blue|Teal|
|-|-|
|Screen Shot 2019-05-20 at 11 49 46 AM|Screen Shot 2019-05-20 at 11 49 27 AM|

Partially addresses https://github.com/Shopify/polaris-react/issues/1486, https://github.com/Shopify/polaris-react/issues/1421, and https://github.com/Shopify/polaris-react/issues/1430

Closes https://github.com/Shopify/polaris-ux/issues/110

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -15,8 +15,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Enhancements
 
 - Added support for dual values to `RangeSlider` component ([#784](https://github.com/Shopify/polaris-react/pull/784))
-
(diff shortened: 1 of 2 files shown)


danrosenthal:
> having a small change in colour having two different semantic meanings will cause problems in the future

Strong agreement from me on this point @ry5n. The rationale for making this blue feels flimsy, and I think moving in this direction would create a snowflake in terms of semantics tied to color. I can see a stronger argument being made for making both badges and banners teal in this case, and changing the variant naming to something which implies timeliness (teal otherwise conveys timeliness in our system, thinking loading, progress bar, spinner, `new`, etc.). Then switching all of the existing `informational` banners which don't carry a timely message to the default grey variant of banner and badge. I also take issue with the `informational` naming convention. Aren't all banner and badges informational in some way? Don't they all convey information?

PR #1544, opened 2019-05-21
[TabMeasure] Prevents reflow when measuring tabs

darrenhebner:
This PR prevents the reflow that happens when we initialize the `Tabs` component. We need to measure the width of the tabs to determine how many we can fit in to a row and how many we need to hide.

This PR uses `requestAnimationFrame` in the `handleMeasurement` method to defer the actions that cause the reflow.

Here's what the reflow looked like on a slow device. `handleMeasurement` took about ~500ms.
Screen Shot 2019-05-21 at 11 01 53 AM

## 

### How to 🎩

Check that the tab component still behaves as expected. This should only be a performance improvement.

<details>

```jsx

export default class Playground extends React.Component {
  state = {
    selected: 0,
  };

  handleTabChange = (selectedTabIndex) => {
    this.setState({selected: selectedTabIndex});
  };

  render() {
    const {selected} = this.state;
    const tabs = [
      {
        id: 'all-customers',
        content: 'All',
      },
      {
        id: 'accepts-marketing',
        content: 'Accepts marketing',
      },
      {
        id: 'repeat-customers',
        content: 'Repeat customers',
        panelID: 'repeat-customers-content',
      },
      {
        id: 'prospects',
        content: 'Prospects',
        panelID: 'prospects-content',
      },
    ];

    return (
      <Card>
        <Tabs tabs={tabs} selected={selected} onSelect={this.handleTabChange}>
          </Card.Section>
        </Tabs>
      </Card>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Tabs/components/TabMeasurer/TabMeasurer.tsx
+++ b/src/components/Tabs/components/TabMeasurer/TabMeasurer.tsx
@@ -24,6 +24,7 @@ export interface Props {
 
 export default class TabMeasurer extends React.PureComponent<Props, never> {
   private containerNode: HTMLElement | null = null;
+  private animationFrame: number | null = null;
 
   componentDidMount() {
     this.handleMeasurement();
@@ -83,25 +84,31 @@ export default class TabMeasurer extends React.PureComponent<Props, never> {
   };
 
   private handleMeasurement = () => {
-    if (this.containerNode == null) {
-      return;
+    if (this.animationFrame) {
+      cancelAnimationFrame(this.animationFrame);
     }
 
-    const {handleMeasurement} = this.props;
-    const containerWidth = this.containerNode.offsetWidth;
-    const tabMeasurerNode = findDOMNode(this);
-    const hiddenTabNodes =
-      tabMeasurerNode instanceof Element && tabMeasurerNode.children;
-    const hiddenTabNodesArray: HTMLElement[] = [].slice.call(hiddenTabNodes);
-    const hiddenTabWidths = hiddenTabNodesArray.map((node) => {
-      return node.getBoundingClientRect().width;
-    });
-    const disclosureWidth = hiddenTabWidths.pop() as number;
+    this.animationFrame = requestAnimationFrame(() => {
+      if (this.containerNode == null) {
+        return;
+      }
 
-    handleMeasurement({
-      containerWidth,
-      disclosureWidth,
-      hiddenTabWidths,
+      const {handleMeasurement} = this.props;
+      const containerWidth = this.containerNode.offsetWidth;
+      const tabMeasurerNode = findDOMNode(this);
+      const hiddenTabNodes =
+        tabMeasurerNode instanceof Element && tabMeasurerNode.children;
+      const hiddenTabNodesArray: HTMLElement[] = [].slice.call(hiddenTabNodes);
+      const hiddenTabWidths = hiddenTabNodesArray.map((node) => {
+        return node.getBoundingClientRect().width;
+      });
+      const disclosureWidth = hiddenTabWidths.pop() as number;
+
+      handleMeasurement({
+        containerWidth,
+        disclosureWidth,
+        hiddenTabWidths,
+      });
     });
   };
 }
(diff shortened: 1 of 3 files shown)


src/components/Tabs/components/TabMeasurer/TabMeasurer.tsx
```diff
@@ -83,25 +84,31 @@ export default class TabMeasurer extends React.PureComponent<Props, never> {
...
+      }
+
+      const {handleMeasurement} = this.props;
+      const containerWidth = this.containerNode.offsetWidth;
+      const tabMeasurerNode = findDOMNode(this);
+      const hiddenTabNodes =
+        tabMeasurerNode instanceof Element && tabMeasurerNode.children;
+      const hiddenTabNodesArray: HTMLElement[] = [].slice.call(hiddenTabNodes);
+      const hiddenTabWidths = hiddenTabNodesArray.map((node) => {
+        return node.getBoundingClientRect().width;
```
jim-chou-shopify: if the intent is to get width then I suggest `offsetWidth`.

src/components/Tabs/components/TabMeasurer/TabMeasurer.tsx
```diff
@@ -83,25 +84,31 @@ export default class TabMeasurer extends React.PureComponent<Props, never> {
...
-      return node.getBoundingClientRect().width;
-    });
-    const disclosureWidth = hiddenTabWidths.pop() as number;
-
-    handleMeasurement({
-      containerWidth,
-      disclosureWidth,
-      hiddenTabWidths,
+    this.animationFrame = requestAnimationFrame(() => {
+      if (this.containerNode == null) {
```
jim-chou-shopify: I think the check for `containerNode` should be outside of the frame.

src/components/Tabs/components/TabMeasurer/TabMeasurer.tsx
darrenhebner: This code already existed.

src/components/Tabs/components/TabMeasurer/TabMeasurer.tsx
darrenhebner: I tried that as well, but you'll see a type error. I think it makes sense to check the container node just before we use it. Since we are now doing the measurements async, it's possible that the container node could change in between when we schedule and when we measure.

AndrewMusgrave:
This is awesome 🎉 Would you be able to add tests/changelog entry? 🙏

darrenhebner:
@AndrewMusgrave - I've updated the changelog. I don't think any tests would need to be changed since this is just an implementation detail.

AndrewMusgrave:
However, it seems like a lot of code is [not being executed anymore](https://codecov.io/gh/Shopify/polaris-react/compare/0fc1462f9719806637f08b30c3839f45a774735c...b78bf239556950a8ba114db45ec38c1b6d6ed11f/diff), which is odd because `handleMeasurement` is called in `didMount`. Stubbing out `requestAnimationFrame` should fix up codecov and allow you to merge

```js
let requestAnimationFrameSpy: jest.SpyInstance;

beforeEach(() => {
  requestAnimationFrameSpy = jest.spyOn(window, 'requestAnimationFrame');
  requestAnimationFrameSpy.mockImplementation((cb) => cb());
});

afterEach(() => {
  requestAnimationFrameSpy.mockRestore();
});
```

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ This release fixes an issue introduced in `v3.14.0` that caused the `esnext` bui
 - Enhanced `NavigationItem`'s color accessibility for `active`, `focus`, `hover` and `Selected` states ([1304](https://github.com/Shopify/polaris-react/pull/1304))
 - Added `align` prop to `TextField` ([#1428](https://github.com/Shopify/polaris-react/pull/1428))
 - Added `clearButton` prop to `TextField` ([#1226](https://github.com/Shopify/polaris-react/pull/1226))
+- Improves performance of `TabMeasure` component ([#1544](https://github.com/Shopify/polaris-react/pull/1544))
```
AndrewMusgrave: Sorry I should have been more specific, we place all our entries [here](https://github.com/Shopify/polaris-react/blob/master/UNRELEASED.md) and move them over to the changelog when we do a release

darrenhebner:
Sorry for leaving this PR open, I just got back from vacation. I'll fix up the last couple of things and merge 👍

PR #1547, opened 2019-05-21
Upgrade storybook to v5.1.0-rc.0

kaelig:
Following up on https://github.com/Shopify/polaris-react/pull/1488

storybooks/storybook#6764 was addressed and it's now possible for us to upgrade to the latest.


tmlayton:
🎩 looks good

PR #1548, opened 2019-05-21
[v4] Merge master into v4

AndrewMusgrave:

To keep v4 update to date with master

* Resovles ~80 conflicts
* Fix misc errors after conflicts, e.g. type errors or test changes
* Convert sheet component to a functional component and make frame context default to null for compatibility
* Refactor dualthumbs tests to not use setState and fix linter errors

This will need a good set of eyes since it was such a huge merge with many conflicts. If you look at the last few commits you'll see my changes after the conflicts were resolved.

### How to 🎩

Comparing [this branch -> master](https://github.com/Shopify/polaris-react/compare/master...19-05-21-master-to-v4?expand=1) against [v4 to master](https://github.com/Shopify/polaris-react/pull/902/files) is a good start since this diff is rather large.


AndrewMusgrave:
> You've changed it to a draft, does that mean you don't want review now?
> 
> At first read-through this seems reasonable, but, as you said, it's a big delta. Maybe we could do a pair review on this so you could walk me through your decisions on some of those last few commits. My confidence in approving this otherwise is low, but I'm okay giving it a 👍 since it is merging to a feature branch and we'll continue to have eyes on this before it makes its way into master.
>
> I think it makes sense to prioritize getting this merged ASAP, so it doesn't become more dated. And in the future to avoid big rebases like this by merging master into this branch each week.

@danrosenthal Sorry it was originally a draft and it seems I forgot to turn it into a regular pull request 😰 Feel free to book some time if you want to go over some of the changes that had to be done to pass CI

src/components/AppProvider/tests/AppProvider.test.tsx
```diff
@@ -7,9 +7,7 @@ describe('<AppProvider />', () => {
   it('updates polaris context when props change', () => {
     const Child: React.SFC<{}> = (_props) => (
       <AppProviderContext.Consumer>
-        {({link: {linkComponent}}) =>
-          linkComponent ? <div id="child" /> : null
-        }
+        {({link}) => (link.getLinkComponent() ? <div id="child" /> : null)}
```
AndrewMusgrave: `linkComponent` is private so we use getters and setters for link

src/components/AppProvider/utilities/withAppProvider/withAppProvider.tsx
```diff
@@ -46,7 +46,7 @@ function withScrollable<P, T>(WrappedComponent: ReactComponent<P, T>) {
                 stickyManager: this.stickyManager,
               }}
             >
-              <WrappedComponent {...this.props} />
+              <WrappedComponent {...this.props as any} />
```
AndrewMusgrave: On `line 115` return the component as `React.ComponentClass<OwnProps> & C` so this won't have any effect on the return type

src/components/AppProvider/utilities/withAppProvider/withAppProvider.tsx
```diff
@@ -89,7 +89,7 @@ export default function withAppProvider<OwnProps>({
 
                     return (
                       <WrappedComponent
-                        {...this.props}
+                        {...this.props as any}
```
AndrewMusgrave: ^

src/components/Banner/Banner.tsx
```diff
@@ -9,23 +9,13 @@ import {
   CircleInformationMajorTwotone,
```
AndrewMusgrave: This change is only removing `withAppProvider` since it's not being used and using `WithinContentContext` consumer rather than `withContext`

src/components/DropZone/DropZone.tsx
```diff
@@ -203,13 +203,6 @@ export class DropZone extends React.Component<CombinedProps, State> {
     };
   }
 
-  get getContext(): DropZoneContext {
```
AndrewMusgrave: This got added in the merge when it shouldn't have been

src/components/Navigation/Navigation.tsx
```diff
@@ -50,6 +50,7 @@ export default class Navigation extends React.Component<Props, never> {
     const context = {
       location,
       onNavigationDismiss: onDismiss,
+      withinContentContainer: true,
```
AndrewMusgrave: This got removed during the merge and was added back

src/components/Navigation/tests/Navigation.test.tsx
```diff
@@ -39,9 +37,15 @@ describe('<Navigation />', () => {
     });
 
     it('has a child with contentContext', () => {
-      const Child: React.SFC<{}> = (_props, context) =>
-        context.withinContentContainer ? <div /> : null;
-      Child.contextTypes = childContextTypes;
+      const Child: React.SFC<{}> = (_props) => {
```
AndrewMusgrave: This text needed to be converted to the new context type

src/components/Scrollable/Scrollable.tsx
```diff
@@ -251,4 +251,4 @@ function prefersReducedMotion() {
   }
 }
 
-export default withAppProvider({withinScrollable: true})(Scrollable);
+export default withAppProvider<Props>({withinScrollable: true})(Scrollable);
```
AndrewMusgrave: There's an issue somewhere (in the old repo) with `Scrollable` being typed as any. Previous HoC changes + this should add it's types back. Without this were now running into type errors.

src/components/Frame/context.tsx
```diff
@@ -11,15 +10,6 @@ export interface FrameContextType {
...
-  showToast: noop,
-  hideToast: noop,
-  setContextualSaveBar: noop,
-  removeContextualSaveBar: noop,
-  startLoading: noop,
-  stopLoading: noop,
-};
-
-const FrameContext = React.createContext<FrameContextType>(defaultContext);
+const FrameContext = React.createContext<FrameContextType | null>(null);
```
AndrewMusgrave: `Sheet` checks `frame == null` but since we always have defaults with the new context API it'll always be true. So I'm defaulting to null and adding more explicit checks below

src/components/Sheet/tests/Sheet.test.tsx
```diff
@@ -67,11 +66,12 @@ describe('<Sheet />', () => {
 
   it('warns when not mounted inside of the Frame component', () => {
     const warnSpy = jest.spyOn(console, 'warn');
-    mountWithAppProvider(
+    mount(
```
AndrewMusgrave: Using `mount` instead of `withAppProvider` so we don't have any context

src/components/AppProvider/utilities/withAppProvider/tests/withAppProvider.test.tsx
```diff
@@ -21,6 +21,10 @@ describe('withAppProvider', () => {
       );
```
AndrewMusgrave: This commit is fixing linter errors and removing `setState` from tests

AndrewMusgrave:
> My confidence isn't 100%, but I think anything that was missed will emerge in tophatting and future changes before this makes its way into master.

I'll run @dleroux through the changes as well before merging to have a 3rd pair of eyes on it

PR #1555, opened 2019-05-22
[Range slider] Fix types and styles

tmlayton:

Props table was not rendering in the style guide AND the styles were not getting built

- Defines `Props` in the component. This is what the style guide parses.
- Adds an empty-ish top-level SCSS file so the styles get built

### 🎩 checklist


tmlayton:
> This is a pretty grim fix for the style stuff

Definitely, as mentioned this should be intentionally bad until we fix it proper in the build. We don't want to obfuscate this.

PR #1557, opened 2019-05-22
[v4][OptionList] Remove componentWillReceiveProps

AndrewMusgrave:

Part of https://github.com/Shopify/polaris-react/issues/519

* Removing `componentWillReceiveProps` from option list
* Fix tests
* Create a hook to address deep comparison
* Add changelog

### How to 🎩

Tests / Percy / Heroku deployment playground

### Notes

The reason for a deep comparison hook is that react doesn't do very much work comparing your dependencies. They only use [Object.is](https://github.com/facebook/react/blob/101901dc2d09509fa0b667196e1ac63ec17bc656/packages/react-reconciler/src/ReactFiberHooks.js#L300)(link to source code) and offer suggestions like [`JSON.stringify`](https://github.com/facebook/react/issues/14476#issuecomment-471199055). We often have complex types that we need to compare so this hook will be very useful in the future as well.

### DIFF
--- a/src/components/OptionList/OptionList.tsx
+++ b/src/components/OptionList/OptionList.tsx
@@ -2,10 +2,11 @@ import * as React from 'react';
 import {createUniqueIDFactory} from '@shopify/javascript-utilities/other';
 
 import {arraysAreEqual} from '../../utilities/arrays';
-import {withAppProvider, WithAppProviderProps} from '../AppProvider';
+import {WithAppProviderProps} from '../AppProvider';
 import {Props as IconProps} from '../Icon';
 import {Props as AvatarProps} from '../Avatar';
 import {Props as ThumbnailProps} from '../Thumbnail';
+import {useDeepCompare} from '../../hooks';
 
 import {Option} from './components';
 import styles from './OptionList.scss';
@@ -31,6 +32,8 @@ export interface SectionDescriptor {
   /** Section title */
   title?: string;
 }
+
+type Descriptor = OptionDescriptor | SectionDescriptor;
 
 const getUniqueId = createUniqueIDFactory('OptionList');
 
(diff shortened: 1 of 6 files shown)


src/components/OptionList/tests/OptionList.test.tsx
```diff
@@ -107,7 +107,7 @@ describe('<OptionList />', () => {
     ];
```
AndrewMusgrave: This just fixes the tests by adding `update` after setProps

src/hooks/use-deep-compare/use-deep-compare.tsx
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react';
```
AndrewMusgrave: See notes for an explanation 😄

src/hooks/use-deep-compare/use-deep-compare.tsx
```diff
@@ -0,0 +1,29 @@
...
+import isEqual from 'lodash/isEqual';
+
+type EffectCallback = () => void | (() => void | undefined);
+type DependencyList = ReadonlyArray<unknown>;
+type Comparator = (a: DependencyList, b: DependencyList) => boolean;
+
+function useDeepCompareRef(
+  dependencies: DependencyList,
+  comparator: Comparator = isEqual,
```
AndrewMusgrave: We won't always want to use isEqual, for example in options list we want to compare arrays and our method is cheaper.

src/components/OptionList/OptionList.tsx
```diff
@@ -61,129 +64,112 @@ export interface State {
...
+  optionRole,
+  onChange,
+  id: propId,
+}: Props) {
+  const [normalizedOptions, setNormalizedOptions] = React.useState(
+    createNormalizedOptions(options, sections, title),
+  );
+  const id = React.useRef(propId || getUniqueId());
+
+  React.useEffect(
```
AndrewMusgrave: I separated this out from the other content that was in `componentWillReceiveProps` since it's unrelated

src/components/OptionList/OptionList.tsx
```diff
@@ -212,6 +198,25 @@ function createNormalizedOptions(
   ];
 }
 
+function isSection(arr: Descriptor[]): arr is SectionDescriptor[] {
+  return typeof arr[0] === 'object' && arr[0].hasOwnProperty('options');
+}
+
+function optionArraysAreEqual(
```
AndrewMusgrave: Comparison function that is passed to `useDeepCompare`

src/components/OptionList/OptionList.tsx
```diff
@@ -212,6 +198,25 @@ function createNormalizedOptions(
...
+  firstArray: Descriptor[],
+  secondArray: Descriptor[],
+) {
+  if (isSection(firstArray) && isSection(secondArray)) {
+    return arraysAreEqual<SectionDescriptor>(
+      firstArray,
+      secondArray,
+      testSectionsPropEquality,
```
AndrewMusgrave: This argument is why we need a separate function to call `arraysAreEqual`

src/components/OptionList/OptionList.tsx
```diff
@@ -212,6 +198,25 @@ function createNormalizedOptions(
   ];
 }
 
+function isSection(arr: Descriptor[]): arr is SectionDescriptor[] {
```
AndrewMusgrave: This is a function for `arr is SectionDescriptor[]` which enables some type-safe features from typescript

src/hooks/use-deep-compare/use-deep-compare.tsx
tmlayton:
```suggestion
import React from 'react';
```

src/components/OptionList/OptionList.tsx
```diff
@@ -61,129 +64,109 @@ export interface State {
...
-      this.setState({
-        normalizedOptions: createNormalizedOptions(
-          nextOptions,
-          nextSections,
-          nextTitle,
-        ),
-      });
-    }
+export default function OptionList({
+  options,
```
tmlayton: Do you have a reason for preferring destructing in the function signature instead of in the block? I personally don’t like it because it makes signatures harder to read.

tmlayton:
We are updating all of our React imports to import React as default so we can import hooks without having the `React.` prefix all over the place. (`import React from 'react';`)

src/components/OptionList/OptionList.tsx
AndrewMusgrave: Not particularly, I just went with it because it's generally the pattern we've been using for functions / functional components.

PR #1561, opened 2019-05-23
Support polaris-icon imports in examples

BPScott:

We're about to deprecate passing a string into `<Icon>` in #1534 in favor of passing in a component imported from `polaris-icons`.

We should update our examples to reflect that new best practice.

The counterpart to this PR in the styleguide is https://github.com/Shopify/polaris-styleguide/pull/2769

- Tweaks how we inject polaris components in storybook to make it a bit simpler
- Add a subset of icons into the scope for examples
- Update readme files, replacing times where we define an icon as a string to instead refer to an icon from polaris-icons.

### How to 🎩

- Check percy to see there are no regressions in READMEs

---

The selected subset of icons is based upon  the ["bundled icons" in Icon](https://github.com/Shopify/polaris-react/blob/18f1a6fcd3a614bd5b64002495fafdafbe843141/src/components/Icon/Icon.tsx#L100). Importing EVERY icon would incur a significant bundle size increase. Working out what icons are used and only importing them is possible but would increase build times and have knock-on effects as it wouldn't work when live-editing examples on the styleguide.

Importing just a subset is a useful middle ground - the bundle size isn't a bloated but we give access to the most commonly used icons.


.storybook/polaris-readme-loader.js
```diff
@@ -49,32 +49,87 @@ module.exports = function loader(source) {
...
-  // It will transform:
-  // function() { return (<Button>Hi</Button>) }
-  // into:
-  // function(React, Button) { return (<Button>Hi</Button>) }
-  // and then call that function with React and Button as the arguments
-  const codeInvoker = function(fn, scope) {
-    const scopeKeys = Object.keys(scope);
-    const scopeValues = scopeKeys.map((key) => scope[key]);
+  // codeInvoker is responsible for injecting Polaris into the scope for a
+  // function so that it will work. It does this by creating a new fuction with
```
chloerice:
```suggestion
  // function so that it will work. It does this by creating a new function with
```

BPScott:
> Is there a reason to leave these examples as is?

The reason was I can't read :) I missed the cases where we use `IconProps['source']` in objects that we pass into jsx components (e.g. `<ActionList items={[{content: 'Import file', icon: ImportMinor}]}  />`) instead of passing it into a component (e.g. `<Button icon={ImportMinor}`).

I've hunted those down and added them, thanks for the spot.

Issue #519, opened 2018-10-04
Avoid using deprecated/UNSAFE_* React methods


BPScott:
[React has deprecated its use of a selection of lifecycle methods it has deemed to be unsafe](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html).

The `componentWillMount`, `componentWillReceiveProps` and `componentWillUpdate` lifecycle methods have the potential to cause headaches and We should use the newer, more stable lifecycle methods of `getDerivedStateFromProps` and `getSnapshotBeforeUpdate` instead.

We *could*  shuffle onto using `UNSAFE_componentWillMount` etc but that doesn't fix the root problem.

The good news is ESLint now warns about our usage of these methods so we no longer have to think about this when doing PR reviews, however there are still a few places in existing code where they are used. Any existing use of these functions should be removed

---

Search the codebase for `react/no-deprecated` and refactor the usage of the `componentWillMount`, `componentWillReceiveProps` and `componentWillUpdate` to use the new lifecycle functions.

As of 2018-10-04, searching the codebase for `react/no-deprecated` shows the following Components are affected:
* [x] AppProvider
* [x] Collapsible (WIP: @dleroux )
* [x] FileUpload (pr https://github.com/Shopify/polaris-react/pull/1233)
* [x] EventListener
* [x] OptionList
* [x] PositionedOverlay
* [x] ResourceList
* [x] Tabs
* [x] TextField
* [x] ThemeProvider

BPScott:
Following an update to eslint-plugin-react in https://github.com/yannickcr/eslint-plugin-react/pull/2069 componentWillMount, componentWillReceiveProps and componentWillUpdate will not trigger a react/no-deprecated warning as they are considered legacy componenets but they're not officially deprecated just yet.

We should still move away from them.

BPScott:
Enabling react/no-unsafe with `checkAliases: true` option shall complain if we try and use and of the unsafe lifecycle methods, including the unprefixed versions

PR #1563, opened 2019-05-23
[v4][PositionedOverlay] Add unsafe to componentWillReceiveProps

AndrewMusgrave:

Part of #519

* adds unsafe and eslint ignores
* skipping the changelog entry since this isn't a meaningful change

Removing `componentWillReceiveProps` is out of scope for the `v4` project and can be a stretch goal. It seems there is a larger issue then `componentWillReceiveProps` being used. As mentioned by [Chris, Dom, Matt and Utkarsh](https://github.com/Shopify/polaris-react-deprecated/pull/506) this was only supposed to be a temporary change that seems to have fallen through the cracks and can't really be converted to other lifecycles/hooks. We will probably need to rearchitect `PositionedOverlay` to fix this problem. After this is merged I'll open an issue outlining the problem.

### DIFF
--- a/src/components/PositionedOverlay/PositionedOverlay.tsx
+++ b/src/components/PositionedOverlay/PositionedOverlay.tsx
@@ -52,6 +52,7 @@ export interface State {
   lockPosition: boolean;
 }
 
+// eslint-disable-next-line react/no-unsafe
 export default class PositionedOverlay extends React.PureComponent<
   Props,
   State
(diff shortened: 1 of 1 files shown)


AndrewMusgrave:
I thought it had but after a quick test, it doesn't seem like it will. `UNSAFE_` allows consumers to upgrade to `v17` however not strict mode. (https://codesandbox.io/s/old-bird-5s5k4). It will probably delay `v4` at least a week (quick estimate) but this is probably something we'll want to tackle

PR #1564, opened 2019-05-23
Make the a11y test fail if there's issues

amrocha:

Last year we wrote an automated a11y test that would use `pa11y`, an a11y testing tool, to automatically check for errors in our code. We enabled it for about a week, before we realized that the test in its current shape was too fragile and was slowing the team down more than it was helping it. We decided to disable the test, and revisit it later.

Flash-forward to today. We've made several improvements in our examples and our playground (huge thanks to Ben for his work on Storybook) that allow us to re-enable the a11y test with more confidence in it.

I enabled the test for every new commit made to the repo, but made it never fail. I have been periodically checking in, and haven't noticed any issues so far. The test seems to work consistently, and identifies real issues.

Having gone through that testing period, I'm confident that we can enable the test in full.

The way the test works is by booting up storybook, and running `pa11y` through all the examples available. Any issues pa11y finds are reported. However, when we ran pa11y, we found there were lots of pre-existing issues in the repo. In order to stop the bleeding, we decided to implement a [shitlist based approach](https://sirupsen.com/shitlists/). This lets us prevent any further issues from being introduced into the codebase while we work to fix the issues in the shitlist.

This was put in place by previous PRs. What this PR does is simply make the test fail in CI if it finds any issues.

## 

### How to 🎩

Delete an item in the shitlist and then run pa11y locally to check that it actually flags the issue.

### DIFF
--- a/scripts/pa11y.js
+++ b/scripts/pa11y.js
@@ -156,7 +156,6 @@ The following issues were discovered and need to be fixed before this code can b
   }
 
   if (results.length || remainingIssues) {
-    // This is temporarily returning 0 while we ensure pa11y doesn't throw any false positives
     process.exit(1);
   }
   process.exit(0);
--- a/scripts/pa11y.js
+++ b/scripts/pa11y.js
@@ -157,7 +157,7 @@ The following issues were discovered and need to be fixed before this code can b
 
   if (results.length || remainingIssues) {
     // This is temporarily returning 0 while we ensure pa11y doesn't throw any false positives
-    process.exit(0);
+    process.exit(1);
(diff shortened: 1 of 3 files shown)


scripts/pa11y.js
```diff
@@ -157,7 +157,7 @@ The following issues were discovered and need to be fixed before this code can b
 
   if (results.length || remainingIssues) {
     // This is temporarily returning 0 while we ensure pa11y doesn't throw any false positives
```
BPScott: This comment can be removed now.

Issue #762, opened 2018-07-02
[a11y] Drop zone has focusable off-screen element


dpersing:
## Issue summary

The Drop zone component replaces the default file upload method with a custom button or a drag-and-drop-only zone, which causes a few issues for keyboard-only and screen reader users:

- All versions of the component have a visually hidden `<input>` element, which is visually hidden but can receive focus and is read by screen readers. When the component has a custom button, focus goes to both the button and the hidden `<input>`. Screen reader users will hear all instances of the upload controls, and sighted keyboard-only users will not be able tell where focus is.
- When the native control is hidden but it is not replaced visually with another control, there does not appear to be a way to upload files with the keyboard.
- If an item has been added to the dropzone, the visible button is removed. Keyboard users may have difficulty adding multiple items.
- If items are added using the drag and drop or separate `<button>`, the visually hidden `<input>` reports that no files have been uploaded.

![Drop zone with no visible button](https://user-images.githubusercontent.com/1462085/42187559-565d6d12-7e05-11e8-996f-6cf804409210.png)

![Small-size Drop zone with no visible button](https://user-images.githubusercontent.com/1462085/42187572-62243248-7e05-11e8-951d-82dde71fc46c.png)

#### Visually hidden file control (HTML/CSS)

```
<span class="Polaris-VisuallyHidden"><input type="file" multiple="" autocomplete="off"></span>
```

```
.Polaris-VisuallyHidden {
    position: absolute!important;
    top: 0;
    clip: rect(1px,1px,1px,1px)!important;
    overflow: hidden!important;
    height: 1px!important;
    width: 1px!important;
    padding: 0!important;
    border: 0!important;
}
```

### Hidden button

1. Using any browser, navigate to the [Drop zone](https://polaris.shopify.com/components/actions/drop-zone#navigation) examples and <kbd>tab</kbd> through them.
2. Repeat the process using any supported browser/screen reader combination.
3. Using the button provided, add items to the dropzone.
4. Read the updated information in the dropzone.

### Expected

- Focus will not go to off-screen controls.
- Controls will be consistent.
- State will be clear.

### Actual

- Focus goes to an off-screen control.
- Visible controls are hidden after uploads are added.
- The visually hidden but readable off-screen control reports a different upload state than is reflected by the items.

## Recommendations

- Use `display: none` to hide the off-screen `<input>` so it cannot be accessed by users, or use the standard `<input>` and not the additional button.
[...]

PR #1565, opened 2019-05-24
Fixing accessibility issues in shitlist

alex-page:

The shitlist contained a bunch of real accessibility issues that have not been fixed.

Fixes https://github.com/Shopify/polaris-react/issues/1458

Fixed accessibility issues in `DropZone`, `Form`, `Modal`, `Section`, `Page`, `Tabs`, `TextField` and `TopBar`.


src/components/Modal/README.md
```diff
@@ -346,8 +346,8 @@ class ModalExample extends React.Component {
           onClose={this.toggleModal}
           title="Get a shareable link"
           primaryAction={{
-            content: 'Close',
-            onAction: this.toggleModal,
+            content: 'Copy link',
```
ry5n: Why this change? I would expect a modal dialog to either have a confirmation action paired with cancel, or a simple close button like in the original example. I would expect any of these actions to close the modal, and I don’t feel like the copy link action fits that expectation.

src/components/Modal/components/CloseButton/CloseButton.tsx
```diff
@@ -11,15 +13,20 @@ export interface Props {
...
+
+function CloseButton({title = true, onClick, polaris: {intl}}: CombinedProps) {
   const className = classNames(
     styles.CloseButton,
     !title && styles.withoutTitle,
   );
 
   return (
     <button onClick={onClick} className={className}>
+      <VisuallyHidden>{intl.translate('Polaris.Common.close')}</VisuallyHidden>
```
ry5n: Nothing wrong with `VisuallyHidden` here, but [our recommendation in the style guide](https://polaris.shopify.com/components/images-and-icons/icon) is to use the `Icon` component’s `accessibilityLabel` prop.

src/components/Modal/README.md
ry5n: Can you say more about this change? I would expect a modal dialog to either have a confirmation action paired with cancel, or a simple close button like in the original example.

src/components/OptionList/OptionList.tsx
```diff
@@ -147,7 +147,7 @@ export class OptionList extends React.Component<CombinedProps, State> {
               {titleMarkup}
               <ul
                 className={styles.Options}
-                id={this.id}
+                id={`${this.id}-${sectionIndex}`}
```
ry5n: 👍

src/components/TextField/README.md
```diff
@@ -733,7 +733,11 @@ class SeparateValidationErrorExample extends React.Component {
             <InlineError message={errorMessage} fieldID={textFieldID} />
           </div>
         </Stack.Item>
-        <Button icon={DeleteMinor} />
+        <Button
+          icon={DeleteMinor}
+          accessibilityLabel="Remove item"
+          labelHidden
```
ry5n: Does `Button` have a `labelHidden` prop?

src/locales/en.json
```diff
@@ -118,7 +122,8 @@
 
     "Page": {
       "Header": {
-        "iconWarningMessage": "The icon prop has been removed from Page. Upload an application icon in the Shopify Partners Dashboard 'App setup' section instead."
+        "iconWarningMessage": "The icon prop has been removed from Page. Upload an application icon in the Shopify Partners Dashboard 'App setup' section instead.",
+        "rollupButton": "Toggle roll up"
```
ry5n: This opens a popover with page-level actions, and I think we can do better with the copy. The menu may contain sections, where there is a convention that one of them is “More actions” (see Orders and Discounts list pages). Maybe just “Actions”?

src/components/TextField/README.md
alex-page: Nope my bad. This line should be removed.
```suggestion
```

src/locales/en.json
alex-page:
```suggestion
        "rollupButton": "Actions"
```

src/components/Modal/components/CloseButton/CloseButton.tsx
alex-page: I have fixed this up

src/components/Form/Form.tsx
```diff
@@ -49,12 +52,15 @@ export default class Form extends React.PureComponent<Props> {
...
       target,
+      polaris: {intl},
     } = this.props;
     const autoCompleteInputs = normalizeAutoComplete(autoComplete);
 
     const submitMarkup = implicitSubmit ? (
       <VisuallyHidden>
-        <button type="submit" aria-hidden="true" />
+        <button type="submit" aria-hidden="true">
+          {intl.translate('Polaris.Common.submit')}
```
ry5n: If the button is aria hidden, it won’t be seen by screen reader users anyway, right? Is this intended, or should we be removing the `aria-hidden`?

alex-page: Thats true, however buttons always need content inside of them. This is probably a false positive but I would prefer the button to have content instead of being empty.

ry5n: Although I’d expect tests to pass with empty content and `aria-hidden="true"`, I’m cool with this fix.

src/components/DropZone/DropZone.tsx
```diff
@@ -304,6 +305,9 @@ export class DropZone extends React.Component<CombinedProps, State> {
         {dragErrorOverlay}
         <div className={styles.Container}>{children}</div>
         <VisuallyHidden>
+          <label htmlFor={id}>
```
dpersing: A couple of items for the drop zone label:

- We have the `label` prop to allow consumers to add a visible label to the drop zone. This prop adds a proper `<label>` with a `for`/`id` relationship for the visually hidden input for the component. However, when the default label is added, it creates two labels for the same component in the output. Only one will get read (typically the first one in the DOM).
- How would the visually hidden default label work if there are multiple drop zone components on the same page? I think the `label` prop is there as a way to prevent duplicate labels.

Here's what the output looks like right now for the example showing a consumer-provided label:

```
<label id="DropZone12Label" for="DropZone12" class="Label-Text_SY9aZ">Theme files</label>
...
<span class="VisuallyHidden-VisuallyHidden_36MaW">
<label for="DropZone12">Upload file</label>
<input id="DropZone12" type="file" multiple="" autocomplete="off">
</span>
```

I'm thinking we could:

- Provide a default visually hidden label unless the consumer provides their own through the prop, and document recommending providing visual labels for drop zones, especially if there are multiple ones on the page
- Or make the `label` prop required, and allow consumers to make it visible or visually hidden

This is complicated by how the control had an additional visible button, and the input is off-screen when it gets focus, which is an issue (#762).

src/components/Modal/README.md
```diff
@@ -364,17 +364,17 @@ class ModalExample extends React.Component {
               <Stack.Item fill>
                 <TextField
                   ref={this.bindNode}
-                  label=""
+                  label="Discount link"
```
dpersing: This is great, but I'm going to log a separate issue I've now noticed with how text field adds `<label>`, `aria-label`, and `aria-labelledby`, which are all duplicative.

src/components/Modal/components/CloseButton/CloseButton.tsx
```diff
@@ -11,15 +12,23 @@ export interface Props {
...
+type CombinedProps = Props & WithAppProviderProps;
+
+function CloseButton({title = true, onClick, polaris: {intl}}: CombinedProps) {
   const className = classNames(
     styles.CloseButton,
     !title && styles.withoutTitle,
   );
 
   return (
-    <button onClick={onClick} className={className}>
```
dpersing: Looks like this is being addressed in PR #1582 (which I happened to get to first today).

src/components/Sheet/README.md
```diff
@@ -207,9 +207,11 @@ class SheetExample extends React.Component {
                   </div>
                   <Scrollable style={{padding: '1.6rem', height: '100%'}}>
                     <ChoiceList
+                      title="Select a sales channel"
```
dpersing: This is good. Making a note that it also caused me to find an unnecessary and invalid `aria-describedby` attribute on the `<fieldset>` output. Will write that up.

src/components/DropZone/DropZone.tsx
alex-page: My bad this should be a quick fix!

alex-page: I have created a new issue for making it required and fixed the issue for now when there is no label provided it uses a sensible default.

src/components/Modal/README.md
dpersing: New issue: #1587

src/components/Sheet/README.md
dpersing: New issue: #1589

src/components/Modal/components/CloseButton/CloseButton.tsx
robin-drexler: Would it make sense to close https://github.com/Shopify/polaris-react/pull/1582 if the same thing is already done here? :)

UNRELEASED.md
```diff
@@ -18,6 +18,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- Fixed accessibility issues in `DropZone`, `Form`, `Modal`, `Section`, `Page`, `Tabs`, `TextField` and `TopBar` ([#1565](pulls/1565)).
```
```suggestion
- Fixed accessibility issues in `DropZone`, `Form`, `Modal`, `Section`, `Page`, `Tabs`, `TextField` and `TopBar` ([#1565](https://github.com/Shopify/polaris-react/pull/1565)).
```

src/components/Modal/components/CloseButton/CloseButton.tsx
alex-page: @robin-drexler i'll merge your PR into here :)

src/components/Tabs/Tabs.tsx
```diff
@@ -35,7 +38,7 @@ export interface State {
   tabToFocus: number;
 }
 
-export default class Tabs extends React.PureComponent<Props, State> {
+export class Tabs extends React.PureComponent<CombinedProps, State> {
```
AndrewMusgrave: We had to export classes in a previous TS version however I don't think that's the case anymore. We should be able to 🔥 it
```suggestion
class Tabs extends React.PureComponent<CombinedProps, State> {
```

src/components/Form/Form.tsx
```diff
@@ -36,7 +37,9 @@ export interface Props {
   onSubmit(event: React.FormEvent<HTMLFormElement>): void;
 }
 
-export default class Form extends React.PureComponent<Props> {
+type CombinedProps = Props & WithAppProviderProps;
+
+export class Form extends React.PureComponent<CombinedProps, never> {
```
AndrewMusgrave: ^

```suggestion
class Form extends React.PureComponent<CombinedProps, never> {
```

src/components/DropZone/tests/DropZone.test.tsx
```diff
@@ -80,47 +77,41 @@ describe('<DropZone />', () => {
...
-  it('calls the onDragEnter callback when a dragEnter event is fired', () => {
+  it('calls the onDragEnter callback when a dragenter event is fired', () => {
     const dropZone = mountWithAppProvider(<DropZone onDragEnter={spy} />);
-    const event = createEvent('dragenter', files);
-    dropZone.getDOMNode().dispatchEvent(event);
+    fireEvent({element: dropZone, eventType: 'dragenter'});
     expect(spy).toHaveBeenCalled();
   });
 
   it('calls the onDragOver callback when a dragOver event is fired', () => {
```
AndrewMusgrave: If we're fixing the typo above as might as well do it here as well 😄 

```suggestion
  it('calls the onDragOver callback when a dragover event is fired', () => {
```

src/components/DropZone/tests/DropZone.test.tsx
```diff
@@ -80,47 +77,41 @@ describe('<DropZone />', () => {
...
     const dropZone = mountWithAppProvider(<DropZone onDragOver={spy} />);
-    const event = createEvent('dragover', files);
-    dropZone.getDOMNode().dispatchEvent(event);
+    fireEvent({element: dropZone, eventType: 'dragover'});
     expect(spy).toHaveBeenCalled();
   });
 
   it('calls the onDragLeave callback when a dragLeave event is fired', () => {
```
AndrewMusgrave: ^

```suggestion
  it('calls the onDragLeave callback when a dragleave event is fired', () => {
```

src/components/Navigation/components/Section/Section.tsx
```diff
@@ -112,7 +112,6 @@ export default class Section extends React.Component<Props, State> {
             <span className={styles.Icon}>
               <Icon source={HorizontalDotsMinor} />
             </span>
-            {ariaLabel}
```
AndrewMusgrave: Just for context, why was the label removed here?

src/components/DropZone/DropZone.tsx
```diff
@@ -292,38 +293,32 @@ export class DropZone extends React.Component<CombinedProps, State> {
...
-      dropZoneMarkup
-    );
-
     return (
-      <Provider value={this.getContext}>{labelledDropzoneMarkup}</Provider>
+      <Provider value={this.getContext}>
+        <Labelled
+          id={id}
+          label={
+            label ? label : intl.translate('Polaris.DropZone.FileUpload.label')
```
AndrewMusgrave: As a general pattern, we tend to use variables rather than having the logic inline to keep the return clean

src/components/DropZone/DropZone.tsx
```diff
@@ -292,38 +293,32 @@ export class DropZone extends React.Component<CombinedProps, State> {
...
+          }
+          action={labelAction}
+          labelHidden={label ? labelHidden : true}
```
AndrewMusgrave: ^

src/components/Navigation/components/Section/Section.tsx
alex-page: Duplicate label on the button container above.

PR #1567, opened 2019-05-24
[wip] Playroom

kaelig:
Exploratory work around https://vault.shopify.io/projects/8178


danrosenthal:
Hey @kaelig, can we go ahead and support this PR, since playroom support is no longer planned?

PR #1570, opened 2019-05-24
[EmptyState] Add content context variant

chloerice:

This PR is one of two that will enable the improvement of transitions from skeleton loading states to empty states by adding support for use of the empty state component in a content context. For example, within a `Card` or `ResourceList`**. 
 
Adds content context consumer to `EmptyState`

Empty state in a card (small screen on right):

Screen Shot 2019-10-01 at 7 39 01 PM

**Note: I'm adding empty state handling to `ResourceList` in a [separate PR](https://github.com/Shopify/polaris-react/pull/2160).

### How to 🎩

<details>
<summary>Click to view collapsed tophatting instructions</summary>
<br />

View the "Empty state in content context" example in the [Chroma deployment](https://www.chromaui.com/component?appId=5d559397bae39100201eedc1&name=All%20Components%7CEmpty%20state&buildNumber=783&specName=Empty%20state%20within%20a%20content%20context). The latest deployment can be found by clicking the "Details" link in the `ci/chroma` check line item of the PR checks below.

I recommend reviewing the code with whitespace changes removed from the diff so it's easier to see what actually changed (since I converted the component from a class to a function, most of the code in `EmptyState.tsx` shifted left one tab).

Screen Shot 2019-09-17 at 7 27 02 PM

*_or_*

<details>

```jsx
import {Page, Layout, EmptyState, Card} from '../src';

  const [contentContext, setContentContext] = useState(true);
  const toggleContentContext = useCallback(
    () => setContentContext((contentContext) => !contentContext),
    [],
  );

  const emptyStateMarkup = contentContext ? (
    <Card sectioned>
      <EmptyState
        heading="Upload a file to get started"
        action={{content: 'Upload files'}}
        image="https://user-images.githubusercontent.com/18447883/59945230-fa4be980-9434-11e9-9106-a373f0efbe08.png"
      >
        <p>
          You can use the Files section to upload images, videos, and other
          documents
        </p>
      </EmptyState>
    </Card>
  ) : (
    <EmptyState
      heading="Manage your inventory transfers"
      action={{content: 'Add transfer'}}
      secondaryAction={{content: 'Learn more', url: 'https://help.shopify.com'}}
      image="https://cdn.shopify.com/s/files/1/0757/9955/files/empty-state.svg"
    >
      <p>Track and receive your incoming inventory from suppliers.</p>
    </EmptyState>
  );

  const markup = contentContext ? (
    <Card>{emptyStateMarkup}</Card>
  ) : (
    emptyStateMarkup
  );

  const secondaryLayoutSection = contentContext ? (
    <Layout.Section secondary>{markup}</Layout.Section>
  ) : null;

  const title = contentContext
    ? 'Empty state content context'
    : 'Empty state standard';

  return (
    <Page
      title={title}
      primaryAction={{
        content: 'Toggle content context',
        onAction: toggleContentContext,
      }}
    >
      <Layout>
        <Layout.Section>{markup}</Layout.Section>
        {secondaryLayoutSection}
      </Layout>
    </Page>
  );
}
```

</details>

</details>

---

### How to test your use case in a content context 👩‍🔬
 
<details>
<summary>Click to view collapsed instructions for testing your use case</summary>

Add a comment to this PR with an explanation of your use case with your empty state image attached/embedded in the comment. 

For example, the image attached below is here so the playground code collapsed above will work for anyone locally. This is because the `EmptyState` `image` prop has the url of the attachment set as its string value. 

You can copy the link to your image attachment by right clicking on the image and selecting "Copy image address" from the right click menu.

File-Upload-Spot

</details>

---

### PR checklist

~(n/a) Tested for [accessibility](https://github.com/Shopify/polaris-react/blob/master/documentation/Accessibility%20testing.md)~

### DIFF
--- a/src/components/EmptyState/EmptyState.scss
+++ b/src/components/EmptyState/EmptyState.scss
@@ -1,7 +1,6 @@
 @import '../../styles/common';
 
 $detail-width: rem(336px);
-
 $stacking-order: (
   illustration: 0,
   details: 10,
@@ -70,14 +69,70 @@ $illustration-width-cropped: calc(
   }
 }
 
-.Details {
-  position: relative;
-  z-index: z-index(details, $stacking-order);
-  padding: 0 spacing();
-  width: $detail-width;
+.withinContentContainer {
+  .Section {
+    position: unset;
+    flex-direction: column-reverse;
+    align-items: center;
+    justify-content: center;
+  }
 
-  @include page-content-when-not-fully-condensed {
-    padding: 0;
+  .DetailsContainer {
+    padding-top: spacing();
+  }
+
+  .Details {
+    display: flex;
+    text-align: center;
+    flex-direction: column;
+    align-items: center;
+
+    @include page-content-when-not-fully-condensed {
+      max-width: $detail-width;
+    }
+  }
+
+  .Image {
+    margin: 0;
+    width: initial;
+  }
+
+  .Content {
+    @include text-style-body;
+    padding-bottom: spacing(tight);
+  }
+}
+
+.withinPage {
+  .Details {
+    position: relative;
+    z-index: z-index(details, $stacking-order);
+    padding: 0 spacing();
+    width: $detail-width;
+
+    @include page-content-when-not-fully-condensed {
+      padding: 0;
+    }
+  }
+
+  .Image {
+    position: relative;
+    z-index: z-index(illustration, $stacking-order);
+    margin-top: spacing(loose) * -1;
+    margin-left: $illustration-left-margin;
+    width: $illustration-width-cropped;
+    max-width: none;
+
+    @include page-content-when-not-fully-condensed {
+      margin-left: 0;
+      width: 100%;
+    }
+
+    @include page-content-when-not-partially-condensed {
+      margin-top: 0;
+      margin-left: $illustration-left-margin;
+      width: $illustration-width;
+    }
   }
 }
 
@@ -88,26 +143,6 @@ $illustration-width-cropped: calc(
 
 .Actions {
   margin-top: spacing();
-}
-
-.Image {
-  position: relative;
-  z-index: z-index(illustration, $stacking-order);
-  margin-top: spacing(loose) * -1;
-  margin-left: $illustration-left-margin;
-  width: $illustration-width-cropped;
-  max-width: none;
-
-  @include page-content-when-not-fully-condensed {
-    margin-left: 0;
-    width: 100%;
-  }
-
-  @include page-content-when-not-partially-condensed {
-    margin-top: 0;
-    margin-left: $illustration-left-margin;
-    width: $illustration-width;
-  }
 }
 
 .FooterContent {
--- a/src/components/EmptyState/EmptyState.tsx
+++ b/src/components/EmptyState/EmptyState.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useContext} from 'react';
 import {classNames} from '../../utilities/css';
 import {WithinContentContext} from '../../utilities/within-content-context';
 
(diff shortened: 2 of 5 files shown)


Ipkhchan:
Uploading image to test with our use case on POS Channel Staff! 

spot-add-POS-staff

Ipkhchan:
Our use case is within a `ResourceList` that displays POS staff members. If a shop does not have POS staff members, we want to show an empty state within the resource list that prompts them to add staff rather than just the default resource list empty state that just indicates that the shop has no staff.
Our workaround right now is to conditionally render the empty state vs. the resource list based on if there are staff. It would be nice to pass in an empty state to the resource list instead. Great that know someone's working on this!

Below is the design for the empty state provided by @pauyeung.

Screen Shot 2019-07-17 at 10 37 32 AM
 

Below is the result of tossing our content (img and text) into the modified `EmptyState` implemented by this PR. 

Screen Shot 2019-07-17 at 10 24 51 AM

The `EmptyState` currently requires there to be a button in the content, whereas our design does not have a button within the empty state because our `pageAction` in the top right provides the call-to-action.

@pauyeung perhaps you can comment on this and how this looks?

pauyeung:
@Ipkhchan @chloerice - I modelled my design off of this example in the Vault update. If buttons are required, does that mean there should be double parked primary action in both the page actions and resource list (Ex. Customers, Products)? I think it could be heavy handed
f9706163a5

chloerice:
Thanks for sharing your use cases @Ipkhchan @pauyeung! 

> The EmptyState currently requires there to be a button in the content, whereas our design does not have a button within the empty state because our pageAction in the top right provides the call-to-action.

> I modelled my design off of this example in the Vault update. If buttons are required, does that mean there should be double parked primary action in both the page actions and resource list (Ex. Customers, Products)? I think it could be heavy handed

Having an `action` in the `EmptyState` is [no longer required](https://github.com/Shopify/polaris-react/pull/1583) 🎉. Your page level action is enough in this use case to give merchants a next step 👍.

chloerice:
Work on this functionality will be picked up again at the next Frideations.

playground/images/index.ts
```diff
@@ -0,0 +1 @@
+export {default as FileUploadSpot} from './File-Upload-Spot.png';
```
dleroux: 🔥 ?

src/components/EmptyState/EmptyState.tsx
```diff
@@ -29,105 +30,113 @@ export interface EmptyStateProps {
...
+            <TextContainer>{footerContent}</TextContainer>
+          </div>
+        ) : null;
 
-    const footerContentMarkup = footerContent ? (
-      <div className={styles.FooterContent}>
-        <TextContainer>{footerContent}</TextContainer>
-      </div>
-    ) : null;
+        const headingSize = withinContentContainer ? 'small' : 'medium';
```
dleroux: Maybe moving the DisplayText Size type to `src/types` to use here, there and in tests would be better.

src/components/EmptyState/EmptyState.scss
```diff
@@ -70,14 +69,66 @@ $illustration-width-cropped: calc(
...
 }
 
-.Details {
-  position: relative;
-  z-index: z-index(details, $stacking-order);
-  padding: 0 spacing();
-  width: $detail-width;
+.withinContentContainer {
+  .Section {
+    position: unset;
```
dleroux: do you need unset?

src/components/EmptyState/EmptyState.tsx
```diff
@@ -29,105 +30,113 @@ export interface EmptyStateProps {
...
+  heading,
+  image,
+  largeImage,
+  imageContained,
+  action,
+  secondaryAction,
+  footerContent,
+}: EmptyStateProps) {
+  return (
+    <WithinContentContext.Consumer>
```
BPScott: As you're in a functional component you can use hooks instead of the render prop function

```jsx
const withinContentContainer = useContext(WithinContentContext);

/* stuff using withinContentContainer */
```

instead of 

```jsx
return (
  <WithinContentContext.Consumer>
    {(withinContentContainer) => {
      /* stuff using withinContentContainer */
    }}
  </WithinContentContext.Consumer>
);

File: src/components/EmptyState/EmptyState.scss (hunk 3 above)
Comment (chloerice) [reply src/components/EmptyState/EmptyState.scss]: Yes, because the position is set to relative, when not in a content context, the position has to be unset or it still has a relative position which pushes it to the right a noticeable amount.

File: src/components/EmptyState/EmptyState.tsx (hunk 4 above)
Comment (chloerice) [reply src/components/EmptyState/EmptyState.tsx]: Gotcha, thanks!

File: src/components/EmptyState/EmptyState.tsx (hunk 2 above)
Comment (chloerice) [reply src/components/EmptyState/EmptyState.tsx]: To clarify, you mean I should change `DisplayText.size` to use an enum instead of a union type so that I can use that there and here instead?

File: playground/images/index.ts (hunk 1 above)
Comment (chloerice) [reply playground/images/index.ts]: Thanks for catching, meant to 🔥 this! That was left over from when I'd saved the empty state image locally before figuring out I could reference images attached to GitHub comments 👍.

File: src/components/EmptyState/EmptyState.tsx (hunk 2 above)
Comment (chloerice) [reply src/components/EmptyState/EmptyState.tsx]: (^ Just pushed up that change, but I think it might be a breaking change, at least for anyone consuming `@shopify/polaris` in a TypeScript project)

Comment (dleroux) [reply src/components/EmptyState/EmptyState.tsx]: I guess you are right, it's not worth the breaking change. Didn't realize it would be. We could just share the current type but not sure there's much value.

**Comment by chloerice** [comment]
> This is looking good! On some mobile devices, the horizontal spacing feels tight. Can we add some left/right padding?
> 
> Before	After
> Screen Shot 2019-09-19 at 9 55 06 AM	Screen Shot 2019-09-19 at 9 55 27 AM

Thanks Sara! I'm hesitant to add spacing around the `EmptyState` component directly, even just for the "in content context" variant. We generally handle the spacing of elements within the component, and leave it to the consumer to handle spacing for between and around components (by using structural Polaris components or custom styled elements as parent containers). 

Nesting the `EmptyState` inside of a `Card.Section` takes care of the spacing in the case of a `Card` being the parent container. I'll update the example and playground code!

File: src/components/EmptyState/tests/EmptyState.test.tsx [hunk 5]
```diff
@@ -1,23 +1,51 @@
...
+
 import {mountWithAppProvider} from 'test-utilities/legacy';
+import {WithinContentContext} from '../../../utilities/within-content-context';
 import {EmptyState} from '../EmptyState';
 
 describe('<EmptyState />', () => {
   let imgSrc =
     'https://cdn.shopify.com/s/files/1/0757/9955/files/empty-state.svg';
 
-  describe('primaryAction', () => {
```
chloerice: (There is no `primaryAction` prop, it's just `action`)

PR #1571, opened 2019-05-24
[Navigation] withinContentContext fix

AndrewMusgrave:

Merging master into v4 caused a regression with banners inside of navigation

Adding the withinContentContext provider

### How to 🎩

Percy ⭐️

### DIFF
--- a/src/components/Navigation/Navigation.tsx
+++ b/src/components/Navigation/Navigation.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 
 import Scrollable from '../Scrollable';
 
+import WithinContentContext from '../WithinContentContext';
 import NavigationContext from './context';
 import {UserMenu, Section, Item} from './components';
 import {SectionType} from './types';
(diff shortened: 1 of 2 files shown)


AndrewMusgrave:
Going to skip adding a changelog entry since this change was never released

src/components/Navigation/Navigation.tsx
```diff
@@ -50,18 +51,19 @@ export default class Navigation extends React.Component<Props, never> {
...
     return (
       <NavigationContext.Provider value={context}>
-        <nav className={styles.Navigation}>
-          {contextControlMarkup}
-          <div className={styles.UserMenu}>{userMenu}</div>
-          <Scrollable className={styles.PrimaryNavigation}>
-            {children}
-          </Scrollable>
-        </nav>
+        <WithinContentContext.Provider value={{withinContentContainer: true}}>
```
AndrewMusgrave: This change is only adding this provider

src/components/Navigation/tests/Navigation.test.tsx
```diff
@@ -39,11 +40,11 @@ describe('<Navigation />', () => {
     it('has a child with contentContext', () => {
       const Child: React.SFC<{}> = (_props) => {
         return (
-          <NavigationContext.Consumer>
+          <WithinContentContext.Consumer>
```
AndrewMusgrave: We'll want to use `WithinContentContext` consumer rather than `NavigationContext` because it provides the `withinContentContext` value

src/components/Navigation/tests/Navigation.test.tsx
```diff
@@ -53,12 +54,7 @@ describe('<Navigation />', () => {
         </Navigation>,
       );
 
-      const div = navigation
-        .find(Child)
-        .find('div')
-        .first();
-
-      expect(div.exists()).toBe(true);
+      expect(navigation.find(Child).find('div')).toHaveLength(1);
```
chloerice: 💯

PR #1573, opened 2019-05-25
[App bridge] Add deprecation messages

tmlayton:

Adds deprecation messages for our App Bridge usage.

For example:

Screen Shot 2019-05-24 at 5 24 46 PM

Adds console warnings, a deprecation doc tag, and deprecation rationale in the component documentation.

### 🎩 checklist

* [x] 🎩 in the style guide

### DIFF
--- a/src/components/Loading/README.md
+++ b/src/components/Loading/README.md
@@ -36,7 +36,7 @@ The loading component must be wrapped in the [frame](/components/structure/frame
 
 ---
 
-## Use in an embedded application
+## Use in an embedded application (deprecated)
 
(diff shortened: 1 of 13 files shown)


src/components/Loading/README.md
```diff
@@ -54,6 +54,10 @@ class EmbeddedAppLoadingExample extends React.Component {
 }
 ```
 
+#### Deprecation rationale
```
kaelig: Should this be an h3?

tmlayton: The parser tries fails because it tries to render them as code examples

PR #1575, opened 2019-05-27
Make choicelist title required

### LINKED ISSUE #1566: Legend should be required for ChoiceList
## Issue summary

ChoiceList creates a fieldset and the legend is optional causing accessibility issues.

## Expected behavior

Legend should be required and it can be visually hidden ( already exists )

## Actual behavior

Legends are not needed causing a11y issues.

alex-page:

Fixes #1566 

Makes the title required on `ChoiceList`. This is necessary as it creates a legend for the fieldset. Without a legend the HTML is not semantic and fails WCAG.

As this is a breaking change it will go into v4.


AndrewMusgrave:
All the tests for [choice lists](https://github.com/Shopify/polaris-react/blob/master/src/components/ChoiceList/tests/ChoiceList.test.tsx) will need title now 😄

alex-page:
> I like this change, but I wonder about admin and other consumers. People will need to do work to upgrade their apps, would we consider that a breaking change?

@ry5n yes, under the title it is being added to version 4! 🎉

ry5n:
Wooops! Thanks :P

PR #1579, opened 2019-05-27
[WIP] Removed aria roles from Tab

### LINKED ISSUE #1577: Selecting a tab changes the Panel id causing missing aria-controls on Tab component
## Issue summary

In our tabs there is an unordered list of individual `Tab` components. These have an `aria-controls` that points to a non existent `tabpanel`.

We currently have one `tabpanel` that the content gets rendered into dynamically. We change the ID after the click event happens, this causes problems as you can navigate to the buttons that `aria-control` `tabpanel`s that do not exist. Other examples of accessible tabs have the child containers pre rendered and hidden:
- https://ui.reach.tech/tabs
- https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html
- https://inclusive-components.design/tabbed-interfaces/

How we can solve this problem:
- All buttons `aria-controls` now control a single `Panel` and remove all semantic meaning from tabs. Make the tabs work like navigation.
- Create empty `tabpanel`s for each `aria-controls`

## Expected behavior

The child container with corresponding `id` exists and is hidden until the user selects the button or link that controls that area.

## Actual behavior

There is one child container that gets its `id` changed on selection of the individual tab.

alex-page:

Fixes #1577. In our tabs there is an unordered list of individual Tab components. These have an aria-controls that points to a non existent tabpanel.

Removes aria roles from tabs and makes them work like a navigation.

### How to 🎩

Make sure the tabs work the same way as expected in the user interface.

### 🎩 checklist


alex-page:
This PR still needs the following before review:
- Accessibility expert review
- Top hatting
- I am not sure if `aria-selected` and `aria-controls` should still be included. My gut says yes.

PR #1580, opened 2019-05-27
Create isObjectsEqual to replace _.isEqual

AndrewMusgrave:

Part of removing lodash #991 . Previous PR #1017

Creating a `isEquals` function and using that instead of `_.isObjectsEqual`

### How to 🎩
Test pass? 💯

### DIFF
--- a/src/components/DataTable/DataTable.tsx
+++ b/src/components/DataTable/DataTable.tsx
@@ -1,8 +1,9 @@
 import React from 'react';
-import isEqual from 'lodash/isEqual';
 import debounce from 'lodash/debounce';
 
 import {classNames} from '../../utilities/css';
+
+import isObjectsEqual from '../../utilities/isObjectsEqual';
 import {headerCell} from '../shared';
 import {
   withAppProvider,
@@ -101,7 +102,7 @@ class DataTable extends React.PureComponent<CombinedProps, DataTableState> {
   }
 
   componentDidUpdate(prevProps: Props) {
-    if (isEqual(prevProps, this.props)) {
+    if (isObjectsEqual(prevProps, this.props)) {
       return;
     }
     this.handleResize();
--- a/src/components/DataTable/DataTable.tsx
+++ b/src/components/DataTable/DataTable.tsx
@@ -3,7 +3,7 @@ import debounce from 'lodash/debounce';
 
 import {classNames} from '../../utilities/css';
 
-import isObjectsEqual from '../../utilities/isObjectsEqual';
+import {isObjectsEqual} from '../../utilities/is-objects-equal';
 import {headerCell} from '../shared';
 import {
   withAppProvider,
--- a/src/components/Page/Page.tsx
+++ b/src/components/Page/Page.tsx
@@ -5,7 +5,7 @@ import {
 } from '@shopify/app-bridge/actions';
 
 import {classNames} from '../../utilities/css';
-import isObjectsEqual from '../../utilities/isObjectsEqual';
+import {isObjectsEqual} from '../../utilities/is-objects-equal';
 import {
   transformActions,
   generateRedirect,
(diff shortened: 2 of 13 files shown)


src/utilities/isObjectsEqual.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+  let result = true;
+
+  for (const prop in objA) {
+    if (!objA.hasOwnProperty(prop)) continue;
+    const propA = objA[prop];
+    const propB = objB[prop];
+    const type = typeof propA;
+
+    if (type === 'object') {
```
AndrewMusgrave: This was changed from a switch -> if so a double break isn't needed

src/utilities/isObjectsEqual.ts
```diff
@@ -0,0 +1,51 @@
...
+      const mapped = map.get(objA);
```
AndrewMusgrave: Added for circular references

BPScott:
How come we're unable to do this on master? Can you expand on why this should be v4 only?

AndrewMusgrave:
> How come we're unable to do this on master? Can you expand on why this should be v4 only?

It doesn't *have* to be version 4, it could have targeted master as well. However, I didn't see much value in that. If this was merged into master it would probably only be in one release before `v4`, and this pr by itself doesn't add much value. This is a change that is not being waited on by anyone, and won't affect anyone's workflow not being in master and It seemed to fit with the theme of `v4` so I decided to drop it in here.

BPScott:
My mindset for prepping major releases is to only include breaking changes in them. Consumers may take some time having to upgrade to v4 due to needing to fixup breaking changes, but they can get onto the last v3.x release without much difficulty. As such I think it's worth putting as many changes in the easy-to-get-ahold-of v3 line as possible, and saving v4 for changes that require more time to process.

By putting these smaller, non-critical changes into v4 we delay their inclusion in other projects, because people won't update to v4 immediately upon its release (with the exception of web because we'll be stewarding that through betas).

That said, you're the one doing the work and it's already been done in the v4 branch so I'm happy to merge this (and the css-utils work) into v4 but I'd like to see future non-breaking, small improvement" prs continue to be made against master.

dleroux:
@AndrewMusgrave do you mind fixing the conflicts and I'll 🎩 ?

src/components/ResourceList/components/Item/Item.tsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
 import {HorizontalDotsMinor} from '@shopify/polaris-icons';
 import {createUniqueIDFactory} from '@shopify/javascript-utilities/other';
-import isEqual from 'lodash/isEqual';
 import {classNames} from '../../../../utilities/css';
+import isObjectsEqual from '../../../../utilities/isObjectsEqual';
```
dleroux: `is-objects-equal`

src/components/DataTable/DataTable.tsx
```diff
@@ -1,8 +1,9 @@
 import React from 'react';
-import isEqual from 'lodash/isEqual';
 import debounce from 'lodash/debounce';
 
 import {classNames} from '../../utilities/css';
+
+import isObjectsEqual from '../../utilities/isObjectsEqual';
```
dleroux: is-objects-equal

src/components/Page/Page.tsx
```diff
@@ -1,11 +1,11 @@
 import React from 'react';
-import isEqual from 'lodash/isEqual';
 import {
   Button as AppBridgeButton,
   TitleBar as AppBridgeTitleBar,
 } from '@shopify/app-bridge/actions';
 
 import {classNames} from '../../utilities/css';
+import isObjectsEqual from '../../utilities/isObjectsEqual';
```
dleroux: `is-objects-equal`

AndrewMusgrave:
Sorry my changes didn't get pushed up when git was down 🤦‍♂️

PR #1581, opened 2019-05-27
[v4] Update app bridge example to use useAppBridge

### LINKED ISSUE #457: Shopify App Bridge context consumer for migrating off legacy React context APIs
# Feature request summary

For current embedded apps trying to interact with the App Bridge, the suggestion is to use the legacy `static contextTypes`, as mentioned in the docs: https://polaris.shopify.com/components/structure/app-provider#section-access-to-the-shopify-app-bridge-instance. As of React 16.6, `StrictMode` will [now start warning](https://reactjs.org/blog/2018/10/23/react-v-16-6.html#deprecations-in-strictmode) that it's deprecated.

This request is for a `Context.Consumer` or HOC (or other mechanism) that gives us access to the App Bridge through the now stable context APIs built in React 16.3 as well as the additional `static contextType` added in React 16.6 (https://reactjs.org/blog/2018/10/23/react-v-16-6.html#static-contexttype). Since there's already an `AppProvider` that is required at the top level of Polaris apps, perhaps there just needs to be a corresponding `AppContextConsumer` component that can wrap components needing access to the App Bridge.

## Rationale

The biggest benefit would be to better prepare Polaris, and all downstream consumers of Polaris, for future React releases as they deprecate the legacy context APIs.

### LINKED ISSUE #1507: [v4][Docs] Update docs - consuming appBridge in applications

AndrewMusgrave:

Fixes #1507
Fixes #457

### How to 🎩

Test the example in an embedded app. Maybe this we can pair on that this week @dleroux ?

### DIFF
--- a/UNRELEASED-V4.md
+++ b/UNRELEASED-V4.md
@@ -19,6 +19,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Documentation
 
+- Updated `AppProvider` app bridge example to use `useAppBridge`([#1581](https://github.com/Shopify/polaris-react/pull/1581))
+
 ### Development workflow
 
 ### Dependency upgrades
--- a/src/components/AppProvider/README.md
+++ b/src/components/AppProvider/README.md
@@ -399,26 +399,21 @@ To provide access to your initialized Shopify App Bridge instance, we make it av
 import React from 'react';
 import {render} from 'react-dom';
 import * as PropTypes from 'prop-types';
-import {AppProvider} from '@shopify/polaris';
+import {AppProvider, useAppBridge} from '@shopify/polaris';
 import {Redirect} from '@shopify/app-bridge/actions';
 
-class MyApp extends React.Component {
-  // This line is very important! It tells React to attach the `polaris`
-  // object to `this.context` within your component.
-  static contextTypes = {
-    polaris: PropTypes.object,
-  };
+function MyApp() {
+  const appBridge = useAppBridge();
+  const redirect = Redirect.create(appBridge);
 
-  redirectToSettings() {
-    const redirect = Redirect.create(this.context.polaris.appBridge);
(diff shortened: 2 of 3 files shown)


dleroux:
@AndrewMusgrave Let's do this at our next pairing.

BPScott:
We're pushing people away from using AppBridge at all - it is currently deprecated and will be removed in v5 and we're telling people to go use [`app-bridge-react`](https://github.com/Shopify/app-bridge/tree/master/packages/app-bridge-react). Given that people shouldn't be using AppBridge via Polaris is it worth publicly exposing a new way of consuming it?

AndrewMusgrave:
> Given that people shouldn't be using AppBridge via Polaris is it worth publicly exposing a new way of consuming it?

We'll have to if we're still allowing it to be consumed from polaris (imo) since consumers can't access it from context anymore

BPScott:
Exposing hooks will only work if our consumers are using functional components. Those using class-based components are out of luck. It would be better to expose the `AppBridgeContext`, that way both class and functional components are good.  We could have the example demo it using either a render prop, or using contextType.

See https://help.shopify.com/en/api/embedded-apps/app-bridge/react-components/provider for an example of how the app-bridge-react demos both.

I don't mind if we demo using a render prop in a functional component, or contextType in a class based one (or both). Render prop is a bit more general purpose as it can be used in both functional and class components but context type requires fewer changes from the current example (dictated but not tested):

```js
class MyApp extends React.Component {
  static contextType = AppBridgeContext;

  redirectToSettings() {
    const redirect = Redirect.create(this.context);

    // Go to {appOrigin}/settings
    redirect.dispatch(Redirect.Action.APP, '/settings');
  }
}
```

BPScott:
If we don't expose the useAppBridge() hook then we can use that as a carrot to get people over to using app-bridge-react - "go migrate if you want to use a hooky interface".

PR #1582, opened 2019-05-28
[MODAL] add accessibility label to Modal close button

robin-drexler:
The `CloseButton` inside the `Modal` does not have any label and is recognized by VoiceOver (and probably other screen readers) as `button` only without any context about its purpose.
 
Adds a label to the `CloseButton` to provide context to users with screen readers.
I decided to use a fixed text instead of introducing a new (`accessibilityLabel`) prop because the text should be consistent and this way nobody needs to adjust their code. :)

I'm not familiar with the translation process, though so any help / info would be appreciated.

### How to 🎩

<details>

```jsx
import {Page} from '../src';
import {CloseButton} from '../src/components/Modal/components';

interface State {}

  render() {
    return (
      <Page title="Playground">
        <CloseButton onClick={() => {}} />
      </Page>
    );
  }
}

```

- Open VoiceOver (⌘ + F5)
- Focus the button (press tab)
- Notice how it reads `Close, button` instead of only `button`

</details>

### 🎩 checklist


dpersing:
Hi @robin-drexler! Thanks so much for tackling this. I've gone into the "Details" section for `translation-platform` and submitted a translation request. Could you also add an entry to `UNRELEASED.md`?

robin-drexler:
@dpersing Thanks for reviewing! Added the changelog entry. :)

dpersing:
Hi @robin-drexler! It turns out @alex-page was working on the same fix along with some others in a separate PR (#1565). Would you be open to merging your PR into his to keep things tidier?

UNRELEASED.md
```diff
@@ -20,6 +20,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- Added missing accessibility label to `Modal` close button ([1582](https://github.com/Shopify/polaris-react/pull/1582))
```
```suggestion
- Added missing accessibility label to `Modal` close button ([#1582](https://github.com/Shopify/polaris-react/pull/1582))
```

PR #1584, opened 2019-05-28
[MODAL] remove unnecessary border radius from modal body

robin-drexler:

Nobody seems to be able to remember why it was added in the first place and it is preventing full width body content.

We are currently building a non-white full width content application in React and noticed that the `Modal` body has a border radius that we are not sure it should have.
Usually you don’t notice it, because the backgrounds are white and the content has no borders.

Removes the border radius from the modal body

### How to 🎩

<details>

```jsx
import {Page, Modal} from '../src';

interface State {}

  render() {
    return (
      <Page title="Playground">
        <Modal open large>
          <Modal.Section>hello</Modal.Section>
        </Modal>
      </Page>
    );
  }
}

```

</details>

- Inspect modal
- Notice border radius is gone from `BodyWrapper`

### 🎩 checklist

### DIFF
--- a/src/components/Modal/Modal.scss
+++ b/src/components/Modal/Modal.scss
@@ -3,7 +3,6 @@ $small-width: rem(620px);
 .BodyWrapper {
   display: flex;
   flex-grow: 1;
-  border-radius: border-radius(large);
   overflow-x: hidden;
   -webkit-overflow-scrolling: touch;
 }
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -20,6 +20,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- Removed unnecessary border-radius from `Modal` body ([#1584](https://github.com/Shopify/polaris-react/pull/1584))
 - Fixed inconsistent width depending on your browser/version in `Sheet` ([#1569](https://github.com/Shopify/polaris-react/pull/1569))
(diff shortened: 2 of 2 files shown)


src/components/Modal/Modal.scss
```diff
@@ -3,7 +3,6 @@ $small-width: rem(620px);
 .BodyWrapper {
```
loic-d: Make sure to add an entry in the changelog.
For example https://github.com/Shopify/polaris-react/pull/1580/files#diff-828bfc9e3c18f7cacda75343f4a1d75e

danrosenthal:
ping @Shopify/app-bridge

Just trying to eliminate all possible reasons for this being present. Is there any embedded modal reason for the modal body setting a border radius?

src/components/Modal/Modal.scss
robin-drexler: Do you think it warrants a changelog entry? For now I added the `skip changelog` label because I thought it does not.

loic-d: Not sure. Better to check with the Polaris team.

danrosenthal: Since we're unsure why this was here, I think it warrants a changelog entry. That way, if it causes style regressions, others without context can track down the root cause. This would go under `bug fixes` in the `unreleased` doc.

robin-drexler: 👍

vividviolet:
I can't think of any reason why embedded apps would need the border radius

danrosenthal:
@robin-drexler :shipit:

PR #1586, opened 2019-05-28
[v4][Proposal] Remove css utilities

AndrewMusgrave:
Do we really need a dependency for two one-liners?

cc/ @BPScott thoughts?


src/utilities/css.ts
```diff
@@ -0,0 +1,9 @@
+export function classNames(
```
AndrewMusgrave: Utilities

src/utilities/tests/css.test.ts
```diff
@@ -0,0 +1,26 @@
+import {classNames, variationName} from '../css';
```
AndrewMusgrave: Tests for the utilities

src/utilities/css.ts
```diff
@@ -0,0 +1,9 @@
...
+  ...classes: (boolean | string | undefined | null | 0)[]
```
BPScott: Do we want `0` or `null` in here? I think we should be restrictive with our types and make people pass a restricted set of values to ensure consistency.

Having `boolean` in here and filtering on Boolean would mean that `classNames(true)`  would result in `'true'` which I don't think we want. I think we should exclude that possibility by making the typing `(string | false | undefined)[]`.

AndrewMusgrave: > Do we want 0 or null in here? I think we should be restrictive with our types and make people pass a restricted set of values to ensure consistency.

It's because we use type coercion a lot in the classNames function e.g.

```js
const className = classNames (reactComponent && styles.Class)
``` 

That's not to say we should, and couldn't wrap all the sneaky cases in the boolean constructor. I don't think we have a lot of nullable values other than react components so we could probably 🔥 `null` & `0` if we wanted too.

> Having boolean in here and filtering on Boolean would mean that classNames(true) would result in 'true' which I don't think we want.

Good call 👍 

> I don't think this is a breaking change as it's purely internal dependencies. Can we make this against master instead?

I would probably have to cherry-pick the commits into a new branch since this will include all the `v4` commits. However, my reasoning is the same as for [isEqual](https://github.com/Shopify/polaris-react/pull/1580#issuecomment-496595117)

BPScott: I've poked around this for a bit, it seems we can't remove null and 0 because of how TS handles  `thing && 'string'` as if thing is 0 then it'll get passed through. Thats just how the typing of the language works and it's out of our control because it's language-builtin stuff rather than functions arguments we control.

The following might be a little clearer in the intent rather than inlining everything, it's the same as you current code but changing `boolean` to `false`:

```ts
type Falsey = false | 0 | null | undefined;

export function classNames(...classes: (string | Falsey)[]) {
  return classes.filter(Boolean).join(' ');
}
```

AndrewMusgrave: The type has been extracted 👍

PR #1590, opened 2019-05-28
[v4] Don’t try to size html to 10px

### LINKED ISSUE #431: Components with `rem` unit have layout problem.
_Originally posted by @ry5n in https://github.com/Shopify/polaris/issues/296#issuecomment-393682836_

# Issue summary

 [utilities.scss's $base-font-size](https://github.com/Shopify/polaris/blob/master/src/styles/foundation/utilities.scss#L2) set to `10px` will cause layout problem in `Chrome` in which system's language is  `Chinese or Japanese`.

Because at that env Chrome will limit the min font size to `12px`, `10px` will not take effect. It will cause components with `rem` unit have larger size.

## Expected behavior

Set  [utilities.scss's $base-font-size](https://github.com/Shopify/polaris/blob/master/src/styles/foundation/utilities.scss#L2) to `12px`

## Actual behavior

You can easily recreate the issue by setting the the html element's font-size to 12px in Chrome's devtool.

1. Change the system language to Chinese or Japanese

## Specifications

- Polaris version number: v2.8 and v2.10
- Browser: Chrome
- Device: Macbook pro
- Operating System: macOS High Sierra 10.13.6

---

ry5n:

Fixes #431

**TL;DR**

Polaris sets the font-size on the html element to 62.5% (10px). We do this as a developer convenience, but it causes most elements to be larger than intended in some locales.

**Long version** 

Polaris uses rem units in order to support people who choose a larger or smaller font size at the browser level. However, Shopifolk are used to thinking in `px`. When we inspect CSS in the browser, the rem output is harder to make sense of: how many px is `0.875rem`? But when we set the html element to a font-size of 10px, it’s easier: `1.4rem` is 14px.

However, languages that use ideographic writing systems (e.g. Chinese, Korean, Japanese) need larger font sizes for legibility, so browsers like Chrome have a minimum font size of 12px. This means the font size of html computes to 12px, not 10, and many elements get rendered larger than intended.

This PR removes a small amount of developer convenience in favor of a UI that is sized the same across locales. This PR should not affect the size of rendered text, only the size of interface elements.

### HOW does this PR work?

Because we’re now going with the browser default as our root em, we had some redundant code. The code has been refactored and improved to handle conversions involving ems more accurately, fix a bug in one code path, and be (hopefully) more readable.

## 

### How to 🎩

**Part 1**
1. In Chrome, go to Chrome > Preferences
1. In Appearance section, click “Customize fonts”
1. For “Minimum font size”, change the value from 10px (default) to 12px
1. Log into a test Shopify store (production)
1. Verify that elements are larger than normal, but text sizes are normal (body font size should still be 14px)

**Part 2**
1. Run the local Storybook for this repo, or open it from the Deployment link in this PR
2. With the browser still set to 12px min font-size, view several components
3. Verify that padding, margin, heights and other dimensions are now back to normal
4. Verify that text is not smaller than normal (body font size should still be 14px)

### 🎩 checklist

### DIFF
--- a/src/styles/foundation/utilities.scss
+++ b/src/styles/foundation/utilities.scss
@@ -1,11 +1,11 @@
-$default-browser-font-size: 16px;
-$base-font-size: 10px;
+$root-font-size: 16px;
 
 /// Returns the value in rem for a given pixel value.
 /// @param {Number} $value - The pixel value to be converted.
+/// @param {Number} $current-font-size - For converting ems, defines the pixel value of 1em.
 /// @return {Number} The converted value in rem.
 
-@function rem($value) {
+@function rem($value, $current-font-size: $root-font-size) {
   $unit: unit($value);
 
   @if $value == 0 {
@@ -13,9 +13,11 @@ $base-font-size: 10px;
   } @else if $unit == 'rem' {
     @return $value;
   } @else if $unit == 'px' {
-    @return $value / $base-font-size * 1rem;
+    @return $value * (1rem / $root-font-size);
+    // e.g. 16px × 1rem/16px = 1rem
   } @else if $unit == 'em' {
-    @return $unit / 1em * 1rem;
+    @return $value * ($current-font-size / 1em) * (1rem / $root-font-size);
+    // e.g. 1em × 24px/em = 24px × 1rem/16px = 1.5rem
   } @else {
     @error 'Value must be in px, em, or rem.';
   }
@@ -23,9 +25,10 @@ $base-font-size: 10px;
 
 /// Returns the value in pixels for a given rem value.
 /// @param {Number} $value - The rem value to be converted.
+/// @param {Number} $current-font-size - For converting ems, defines the pixel value of 1em.
 /// @return {Number} The converted value in pixels.
 
-@function px($value) {
+@function px($value, $current-font-size: $root-font-size) {
   $unit: unit($value);
 
   @if $value == 0 {
@@ -33,20 +36,22 @@ $base-font-size: 10px;
   } @else if $unit == 'px' {
     @return $value;
   } @else if $unit == 'em' {
-    @return ($value / 1em) * $base-font-size;
+    @return $value * ($current-font-size / 1em);
+    // e.g. 1em × 24px/em = 24px
   } @else if $unit == 'rem' {
-    @return ($value / 1rem) * $base-font-size;
+    @return $value * ($root-font-size / 1rem);
+    // e.g. 1rem × 16px/rem = 16px
   } @else {
     @error 'Value must be in rem, em, or px.';
   }
 }
 
-/// Returns the value in ems for a given pixel value. Note that this
-/// only works for elements that have had no font-size changes.
-/// @param {Number} $value - The pixel value to be converted.
+/// Returns the value in ems for a given pixel value.
+/// @param {Number} $value - The value (em, rem, or px) to be converted.
+/// @param {Number} $current-font-size - For converting ems, defines the pixel value of 1em.
 /// @return {Number} The converted value in ems.
 
-@function em($value) {
+@function em($value, $current-font-size: $root-font-size) {
   $unit: unit($value);
 
   @if $value == 0 {
@@ -54,9 +59,11 @@ $base-font-size: 10px;
   } @else if $unit == 'em' {
     @return $value;
   } @else if $unit == 'rem' {
-    @return $value / 1rem * 1em * ($base-font-size / $default-browser-font-size);
+    @return $value * ($root-font-size / 1rem) * (1em / $current-font-size);
+    // e.g. 1rem × 16px/rem = 16px × 1em/24px = 0.667em
   } @else if $unit == 'px' {
-    @return $value / $default-browser-font-size * 1em;
+    @return $value * (1em / $current-font-size);
+    // e.g. 16px × 1em/24px = 0.667em
   } @else {
     @error 'Value must be in px, rem, or em.';
   }
--- a/src/styles/foundation/utilities.scss
+++ b/src/styles/foundation/utilities.scss
@@ -1,11 +1,11 @@
 $root-font-size: 16px;
 
-/// Returns the value in rem for a given pixel value.
-/// @param {Number} $value - The pixel value to be converted.
-/// @param {Number} $current-font-size - For converting ems, defines the pixel value of 1em.
+/// Returns the value in rem for a given length value.
+/// Note: converting ems only works for elements that have had no font-size changes.
+/// @param {Number} $value - The value (em, rem, or px) to be converted.
 /// @return {Number} The converted value in rem.
 
-@function rem($value, $current-font-size: $root-font-size) {
+@function rem($value) {
   $unit: unit($value);
 
   @if $value == 0 {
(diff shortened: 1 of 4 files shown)


BPScott:
I'm happy that this is happening but it's a terrifying change as it is very fundamental and I'd be a little scared to merge this into master as I don't think "you need to sweep your codebase for references to rem values and adjust them" is an acceptable ask for a minor release. Can we target the version-4.0.0 branch instead?

Can we also just remove this line if we're going for 16px which is the browser default: https://github.com/Shopify/polaris-react/blob/83fb11a83ebb66c3bb3849c9452eb12ffb65036a/src/styles/global/elements.scss#L19

alex-page (changes requested):
Great work Ryan, this is a big win for internationalisation. Weirdly enough I usually assume 1rem = 16px so this is actually less confusing for me.

As Ben said above there is a bit of redundancy across the code as we are now making the base font size 16px. We can remove the `$default-browser-font-size` and the `$font-size` calculation. We can make `$font-size: 16px;` and that should trickle through the system correctly.

One thing I was unsure about is making the padding `16px` on the sheet examples. I might not understand what happens to the CSS that gets passed to these components from props. If it isn't getting converted to REM it should just use `1rem` value so that the padding scales when the root font size is increased (e.g. text zoom).

I agree with Ben that I would prefer this change to be made in version `4.0.0`, it will likely be a breaking change for some users if they don't change their values or are using variables that are removed.

ry5n:
> I'm happy that this is happening but it's a terrifying change as it is very fundamental and I'd be a little scared to merge this into master as I don't think "you need to sweep your codebase for references to rem values and adjust them" is an acceptable ask for a minor release. Can we target the version-4.0.0 branch instead?
> 
> Can we also just remove this line if we're going for 16px which is the browser default:
> 
> https://github.com/Shopify/polaris-react/blob/83fb11a83ebb66c3bb3849c9452eb12ffb65036a/src/styles/global/elements.scss#L19

That’s a good point—people using raw rems would be affected and that definitely means this should go into 4.0.

> there is a bit of redundancy across the code

When I change this to target 4.0, I’ll look into cleaning this up.

> One thing I was unsure about is making the padding 16px on the sheet examples.

This is an artifact of how our examples work. It’s hard to include Sass (or CSS for that matter) in the examples, since they’re just text in a markdown file. The Sheet component has no content of its own, so we’re writing more custom markup than usual, and sprinkling some CSS into `style` attributes. Those styles were using raw rems, which broke in exactly the way Ben was afraid of. I could re-write these with rems based on 16px, but the thought was that px would be more reliable.

src/styles/foundation/utilities.scss
```diff
@@ -1,62 +1,69 @@
...
   @if $value == 0 {
     @return 0;
   } @else if $unit == 'rem' {
     @return $value;
   } @else if $unit == 'px' {
-    @return $value / $base-font-size * 1rem;
+    @return $value * (1rem / $root-font-size);
+    // e.g. 16px × 1rem/16px = 1rem
   } @else if $unit == 'em' {
-    @return $unit / 1em * 1rem;
```
ry5n: ^ This looks like a bug. This code path (where `$value` is in `em` units) is an invalid Sass expression.

ry5n:
I’ve stripped out some of the redundant code and changed the base branch to version 4.

In refactoring the unit conversion calculations:

- I found at least one bug–hopefully fixed now. I don’t see any tests, which is slightly scary 😄
- The conversion functions didn’t accurately convert ems. In contexts where 1em ≠ 1rem, devs can now provide an optional second parameter with the current font size to get an accurate em value.
- I found some calculations hard to follow. I’ve reformatted the calculations so they’re hopefully easier to understand. Remember stoichiometry from high school? 🧪

src/styles/foundation/utilities.scss
ry5n: (Probably intended to be `$value / 1em * 1rem`)

src/styles/foundation/utilities.scss
```diff
@@ -1,62 +1,69 @@
-$default-browser-font-size: 16px;
-$base-font-size: 10px;
+$root-font-size: 16px;
```
ry5n: I would really like this to be a constant but Sass has no notion of private variables without pain, so I’m leaving it here. (The alternative is hard-coding `16px` into these calculations which I like even less.)

src/styles/global/elements.scss
```diff
@@ -16,7 +16,7 @@ button {
 
 html {
   position: relative;
-  font-size: ($base-font-size / $default-browser-font-size) * 100%;
+  font-size: ($root-font-size / 16px) * 100%;
```
ry5n: Because the `$root-font-size` variable still exists (see above), I’ve opted to keep this line.

src/styles/foundation/utilities.scss
kaelig: You're all good! The way it's declared, it's basically a constant. People can't easily override it (compared to a variable declared using a `!default` flag).

If you wanna make it _look_ private, prefix it with a `_`, it's a very common convention in the Sass community but I don't know if it is at Shopify.

```scss
// Easy to override
$root-font-size: 16px !default;

// Hard to override (have to fiddle with imports)
$root-font-size: 16px;

// Don't override me!
$_root-font-size: 16px;
```

src/styles/foundation/utilities.scss
kaelig: `em` and `rem` are behaving so differently that I think we should `@error` when an `em` value is passed.

ry5n: What do you think of the updated implementation? It fixes this bug, I think

kaelig: The bug fix is valid!

I think we can safely stop allowing passing `em` values to the function:

1. This use-case didn't work before, therefore it's not a working part of the API
1. If people were using `rem(<number>em)`, they'd have raised an issue because it throws `Undefined operation: ""em"/1em times 1rem". on line x at column 13.`

alex-page (approved):
This looks good @ry5n, nice work. I did a quick top hat over the components in firefox. No changes have been seen in percy so this should be safe to merge.

I think the variable as `$root-font-size` is fine. Only `!default` values are encouraged to be overridden in sass.

The bug fix looks good as well, it would be cool to write some tests for this function. Happy to help out, it might be better as another PR to separate concerns.

src/styles/foundation/utilities.scss
```diff
@@ -1,27 +1,31 @@
...
   } @else if $unit == 'rem' {
     @return $value;
   } @else if $unit == 'px' {
-    @return $value / $base-font-size * 1rem;
+    // e.g. 16px × 1rem/16px = 1rem
+    @return $value * (1rem / $root-font-size);
   } @else if $unit == 'em' {
-    @return $unit / 1em * 1rem;
+    // This is a fudge; it assumes 1em == 1rem
+    @return $value / 1em * 1rem;
```
kaelig: Fixing this bug means we'll now have to support it.

Since we know nobody is using this (or they'd have submitted an issue), I vote for removing this part completely:

```diff
- } @else if $unit == 'em' {
-    @return $unit / 1em * 1rem;

# and also:
-   @error 'Value must be in px, em, or rem.';
+   @error 'Value must be in px, or rem.';
```

ry5n: The thing is we’re supporting this API already, and fixing it is low cost.

I’d be all for removing em unit conversion if it wasn’t needed. However, [we convert _to_ ems in our own code](https://github.com/Shopify/polaris-react/blob/6b224cdcfe95f2b45adc0f282e3ddb7757e71b8c/src/styles/shared/breakpoints.scss#L5), so we have to support that. The most we could do is remove converting _from_ em units. That seems like a strangely asymmetric API to me.

Is there really any downside to just shipping this bug fix?

ry5n: Thanks Kaelig. I see where you’re coming from, and I appreciate it. In general I definitely support not adding (and deleting!) code we don’t use.

PR #1592, opened 2019-05-29
Remove export of CombinedProps and export of class

alex-page:

`CombinedProps` are not used across any components so the export is not useful. We are also exporting a bunch of classes that is causing a double up.

Removing the export of `CombinedProps` and unnecessary exports for classes.


alex-page:
There is also a bunch of private components that have exports where they should not. See `Avatar`:
`export class Avatar extends React.PureComponent<CombinedProps, State> {`
~~export~~ `class Avatar extends React.PureComponent<CombinedProps, State>`

The `export` should not be needed as we export the component as the default value:
`export default withAppProvider<Props>()(Avatar);`

alex-page:
@BPScott I have removed the unused export for classes as well.

I noticed that there were some exports that are being used which I thought might be incorrect, however I have left them for now:
```jsx
export class BaseItem extends React.Component<CombinedProps, State>
export class OptionList extends React.Component<CombinedProps, State>
export class SearchField extends React.Component<ComposedProps, never>
export class UnstyledLink extends React.PureComponent<CombinedProps, never>
```

These components fail the build when removing `export` as they are used in `static` definitions:
```jsx
static OptionList = OptionList;
```

BPScott:
> Does this reduce the bundle significantly since we're not double exporting anymore?

No, The code is still there and required, but it makes it clear that these classes/interfaces are only for use inside the file in which they're defined and not elsewhere.

PR #1596, opened 2019-05-29
[v4] Add react-testing/utils/matchers

### LINKED ISSUE #1506: [v4] Testing library/tools exploration
* [react-testing](https://github.com/Shopify/quilt/tree/master/packages/react-testing)
* [enzyme-utilities](https://github.com/Shopify/quilt/tree/master/packages/enzyme-utilities)
* custom matchers e.g. [jest-enzyme](https://www.npmjs.com/package/jest-enzyme)

AndrewMusgrave:

Fixes #1506

React testing [faq](https://github.com/Shopify/quilt/tree/master/packages/react-testing#faq) explains why we should use it over enzyme. If we're using `react-testing`, we won't need to look `jest-enzyme`, `enzyme-utilities`, etc.

* Add react-testing
* Setup setupTestFrameworkScriptFile
  * add matchers
  * add test cleanup
* Create utilities
* Create custom matcher that can be used as an example in the future
* Refactor random test file (radio button) to use react-testing instead of enzyme
* Rename `test-utils/enzyme` to `test-utils/legacy`


AndrewMusgrave:
Thoughts, opinions?

src/components/RadioButton/tests/RadioButton.test.tsx
```diff
@@ -1,11 +1,11 @@
 import React from 'react';
```
AndrewMusgrave: Refactoring this test file to use `react-testing` as an example

src/test-utilities/matchers/index.ts
```diff
@@ -0,0 +1,2 @@
+import '@shopify/react-testing/matchers';
```
AndrewMusgrave: Main import for our matchers

src/test-utilities/index.ts
```diff
@@ -1,2 +1,2 @@
-export * from './enzyme';
```
AndrewMusgrave: Our enzyme utils will need to be imported from `test-utilities/legacy` now

src/test-utilities/matchers/react.ts
```diff
@@ -0,0 +1,25 @@
+import chalk from 'chalk';
+import {Node} from './types';
+
+declare global {
```
AndrewMusgrave: This chuck of TS is so we can extend expect, e.g.
```ts
expect(X). toBeDisabled()
```

src/test-utilities/matchers/react.ts
```diff
@@ -0,0 +1,25 @@
...
+    pass,
+    message() {
+      return pass
+        ? chalk.green(`Expected component not to be disabled, but it was.`)
+        : chalk.red(`Expected component to be disabled, but it wasn't.`);
+    },
+  };
+}
+
+expect.extend({toBeDisabled});
```
AndrewMusgrave: This adds our custom matcher to jest, and why we only need to import our matchers in `test/each-test.ts`

src/test-utilities/react-testing.tsx
```diff
@@ -0,0 +1,89 @@
...
+  themeProvider,
+  frame,
+  ...props
+}: Props) {
+  const childWithProps =
+    Object.keys(props).length > 0
+      ? React.cloneElement(children, props)
+      : children;
+
+  return (
```
AndrewMusgrave: Setups all our global context

tests/each-test.ts
```diff
@@ -0,0 +1,6 @@
+import {destroyAll} from '@shopify/react-testing';
+import '../src/test-utilities/matchers';
```
AndrewMusgrave: Imports our matchers to be used

tests/each-test.ts
```diff
@@ -0,0 +1,6 @@
...
+
+afterEach(() => {
+  destroyAll();
```
AndrewMusgrave: [Test cleanup](https://github.com/Shopify/quilt/tree/master/packages/react-testing#-destroyall)

src/test-utilities/react-testing.tsx
```diff
@@ -22,8 +22,8 @@ interface Providers {
   themeProvider: ThemeProviderContextType;
   frame: FrameContextType;
 }
-type Options = {};
-type Context = DeepPartial<Providers>;
+type Options = DeepPartial<Providers>;
```
AndrewMusgrave: These are the options the consumer provides e.g. `mount(component, options)`

src/test-utilities/react-testing.tsx
```diff
@@ -22,8 +22,8 @@ interface Providers {
...
+type Context = Providers;
```
AndrewMusgrave: This is provided to the render method of `createMount`. Which will currently be the provider contests.

AndrewMusgrave:
> Awesome! What will it take to completely leave enzyme behind, refactoring all the tests like you did in src/components/RadioButton/tests/RadioButton.test.tsx?

We would have to, but I'm not sure it's worth the time. We have enzyme utilities that'll work with the new context. I think moving forward we should exclusively use `react-tesing`. I moved `enzyme` to `testing-utilities/legacy` to hopefully prevent further us 😅 

> Ran build-consumer in web though and tests break all over the place, but I think that has more to do with other things in v4 like context changes etc. Have we been testing branches in web before they merge into v4?

We haven't been. All tests that use icon right now are going to fail, I was hoping to do an alpha release this week and start migrating web but I didn't get it too. So early next week 😄

PR #1599, opened 2019-05-29
[v4] Remove Navigation.UserMenu

### LINKED ISSUE #879: Make sure deprecation versions are correct when releasing the next version
Not sure what would be the best way to address this so figured I would make a note here. When the next version of Polaris gets released the version number in [`Navigation.UserMenu`'s deprecation rationale](https://github.com/Shopify/polaris-react/blob/master/src/components/Navigation/README.md#deprecation-rationale) will need to be validated. It might be good to in the future have some kind of workflow for this.

cc @danrosenthal @tmlayton

### LINKED ISSUE #1485: [v4][Navigation/TopBar] Remove deprecation / withContext

AndrewMusgrave:

Fixes #1485
Fixes #879

 Removing the deprecated `UserMenu` prop from `Navigation`.

@kvendrik How does this look? It this the desired layout after deprecations?


AndrewMusgrave:
cc/ @kvendrik

PR #1602, opened 2019-05-30
[v4] Simplify withinContentContext

### LINKED ISSUE #1521: [v4] Simplify `withinContentContext`

AndrewMusgrave:

Fixes #1521

Changing WithinContentContextType from `{withinContentContext: boolean}` to `boolean`

### DIFF
--- a/src/components/WithinContentContext/context.tsx
+++ b/src/components/WithinContentContext/context.tsx
@@ -1,11 +1,9 @@
 import React from 'react';
 
-export interface WithinContentContextType {
-  withinContentContainer: boolean;
-}
+export type WithinContentContextType = boolean;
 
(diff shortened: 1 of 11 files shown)


src/components/WithinContentContext/context.tsx
```diff
@@ -1,11 +1,9 @@
 import React from 'react';
 
-export interface WithinContentContextType {
-  withinContentContainer: boolean;
-}
+export type WithinContentContextType = boolean;
```
BPScott: Having to define and export such a trivial type just so we can use it in tests feels a bit rubbish.

What do you think to removing this and updating the tests so the TestComponent is `const TestComponent = (_: any) => null;`?

PR #1603, opened 2019-05-30
[Button] Make Button with plain and destructive props red

calumptrck:

The online-store-ui library has a component which uses the Polaris `<Button />` with a red destructive button.

• Change the colour of `<Button plain destructive />` to be red instead of blue when it's passed both the `destructive` and `plain` props.
• Add an example to the component's `README.md`.

| Before | After |
| ------------- | ------------- |
| Screen Shot 2019-05-30 at 1 13 55 PM  | Screen Shot 2019-05-29 at 4 14 38 PM |

### How to 🎩

<details>

```jsx

interface State {}

  render() {
    return (
      <Page title="Playground">
         <Button plain destructive />
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -164,6 +164,13 @@ $spinner-size: rem(20px);
     margin-right: 0;
   }
 
+  &.destructive {
+    color: color('red', 'dark');
+    &:hover {
+      color: color('red', 'darker');
+    }
+  }
+
   &.disabled {
     background: none;
   }
@@ -207,6 +214,7 @@ $spinner-size: rem(20px);
     margin-right: 0;
   }
 }
+
 // stylelint-enable selector-max-specificity
 
 .sizeSlim {
--- a/src/components/Button/README.md
+++ b/src/components/Button/README.md
@@ -187,6 +187,18 @@ Use to render a buttons that look like links where matching the surrounding text
 />
 ```
 
+### Plain destructive button
+
+<!-- example-for: online-store-ui -->
+
+Use for actions that will delete merchant data or be otherwise difficult to recover from. Since they're less prominent, use for less important or less commonly used destructive actions. For example, plain buttons are used as actions in cards.
+
+```jsx
+<Button plain destructive>
+  Remove
+</Button>
+```
+
 ### Primary button
 
 Use to highlight the most important actions in any experience. Don’t use more than one primary button in a section or screen to avoid overwhelming merchants.
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -22,6 +22,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Added `actions` prop to `<Card.Section>` to allow you to easily define header actions in a card section ([#1598](https://github.com/Shopify/polaris-react/pull/1598))
 - Removed transition on tag button hover state [#1337](https://github.com/Shopify/polaris-react/pull/1337)
 - Added `textAlign` prop to Button ([#1576](https://github.com/Shopify/polaris-react/pull/1576))
+- Made `Button` red when given both the `plain` and `destructive` props ([#1603](https://github.com/Shopify/polaris-react/pull/1603))
 
 ### Bug fixes
 


src/components/Button/Button.tsx
```diff
@@ -116,6 +117,7 @@ function Button({
     isDisabled && styles.disabled,
     loading && styles.loading,
     plain && styles.plain,
+    isPlainDestructive && styles.plainDestructive,
```
danrosenthal: Instead of adding a new className, could you style `.plain.destructive` instead?

calumptrck: @danrosenthal Yep, pushed an updated version. Thanks!

src/components/Button/Button.scss
```diff
@@ -149,6 +149,13 @@ $spinner-size: rem(20px);
     margin-right: 0;
   }
 
+  &.destructive {
+    color: color('red');
+    &:hover {
+      color: color('red', 'dark');
```
danrosenthal: @ry5n do you have any thoughts on creating a `destructive` style for a `plain` Buttons. Any potential downsides to this? Omitting it feels like an oversight to me. Also, what do you think about the hover style in this implementation?

src/components/Button/Button.scss
```diff
@@ -149,6 +149,13 @@ $spinner-size: rem(20px);
     margin-right: 0;
   }
 
+  &.destructive {
```
chloerice: @sarahill What do you think color-wise here? Should destructive card actions should match destructive action list items in being dark red? The reason action list item content is dark red is because there wouldn't be enough contrast between the focus/hover background colors, but plain buttons don't have background so could arguably work as is here.

chloerice: Screen Shot 2019-05-31 at 11 36 55 AM

src/components/Button/README.md
```diff
@@ -187,6 +187,18 @@ Use to render a buttons that look like links where matching the surrounding text
 />
 ```
 
+### Plain destructive button
+
+<!-- example-for: online-store-ui -->
+
+Use to render a buttons that look like links where matching the surrounding text color is visually more appropriate. For example in the InlineError component.
```
chloerice: Here, I wouldn't mention the inline error component because using a plain button inline there would make the button indistinguishable. We recommend using plain monochrome buttons for error actions.

```suggestion
Use for actions that will delete merchant data or be otherwise difficult to recover from. Since they're less prominent, use for less important or less commonly used destructive actions. For example, plain buttons are used as actions in cards.
```

src/components/Button/README.md
```diff
@@ -187,6 +187,18 @@ Use to render a buttons that look like links where matching the surrounding text
...
+
+```jsx
```
chloerice: Let's also add a new example to the `Card` component `README` that shows a destructive card action, since that's the intended/main use case 👍

src/components/Button/Button.scss
sarahill: The 2 different reds feel odd. I'm good with using `red dark` for all of these as far as color goes.

src/components/Button/README.md
chloerice: @calumptrck You'll also need to update the action type of `Card.action` from `DisableableAction` to `DestructibleAction` in order to make a card action destructible.

sarahill:
I wonder about the hover state for these. The background color on hover is pretty low contrast against the background of the page or card. Is the color change of the text enough in this case? Also, if we go with `red dark` I assume we'd use `red darker` for the hover? It gets pretty dark but there would be little to no change otherwise.

image

cc: @dpersing @ry5n

src/components/Button/Button.scss
chloerice: Should we have a hover/focus state similar to a regular plain button that darkens the content? 

Current

```scss
  &.destructive {
    color: color('red');
    &:hover {
      color: color('red', 'dark');
    }
  }
```



---

Dark red to darker red might be too drastic of a transition 🤔 

```scss
  &.destructive {
    color: color('red', 'dark');
    &:hover {
      color: color('red', 'darker');
    }
  }
```



---

Maybe something like

```scss
  &.destructive {
    color: color('red', 'dark');
    &:hover {
      color: darken(color('red', 'dark'), 10%);
    }
  }
```

sarahill: Ah I didn't see you commented on this. Related to my comment here https://github.com/Shopify/polaris-react/pull/1603#issuecomment-497770381

I'm mixed on modifying the red darker. I agree it's pretty drastic but it's also another example of where our color system starts to fall short (lack of range). I'm ok with this approach but something we should keep in mind / consider with the color system updates.

chloerice: 💯

chloerice: To clarify, we should go with the third variation above?

chloerice: > I'm mixed on modifying the red darker

Understandable, though I just did a quick search and it doesn't look like `color('red', 'darker')` is used anywhere in the system or in `web`, so it would likely be low risk as far as UI regression is concerned.

dpersing:
@sarahill @chloerice I think the gray background and darker red text for hover/focus is the way to go to make the state change more obvious. This will helpfully also call attention to the control for folks who can't see red and are going to be relying on the text only for the destructive action. I also agree we should avoid adjusting colors with opacity, since that gets away from the palette very quickly, when we should address issues with color in the palette itself.

chloerice:
> @sarahill @chloerice I think the gray background and darker red text for hover/focus is the way to go to make the state change more obvious. This will helpfully also call attention to the control for folks who can't see red and are going to be relying on the text only for the destructive action. I also agree we should avoid adjusting colors with opacity, since that gets away from the palette very quickly, when we should address issues with color in the palette itself.

💯I'll update my suggestion in the review 👍

src/components/Button/Button.scss
```diff
@@ -149,6 +149,13 @@ $spinner-size: rem(20px);
...
+    color: color('red');
```
chloerice:
```suggestion
    color: color('red', 'dark');
```

src/components/Button/Button.scss
chloerice:
```suggestion
      color: color('red', 'darker');
```

calumptrck:
Thank you @chloerice @sarahill @dpersing! Pushed the changes.

Issue #1061, opened 2019-02-20
Remove support for custom Shopify SVG format


amrocha:
Shopify imports SVGs with the following format:

```
{
  viewBox: [string],
  body: [body],
}
```

We've asked around and it seems nobody is aware of a reason why it needs to be this way, and we can switch to a more standard format.

We can't just remove it though, because sewing-kit uses this format in its SVG loader, so that would be a very disruptive change to the rest of Shopify.

The proposed order of steps to remove this is:
 - [x] Add support for SVGR to `polaris-react` (#1042 )
 - [x] Add support for SVGR to `sewing-kit` (https://github.com/Shopify/sewing-kit/pull/1208)
 - [x] Deprecate support for the old format in `polaris-react`
 - [x] Remove support for the old format in `polaris-react` in a major release (#1604)

@BPScott

BPScott:
🎉

PR #1604, opened 2019-05-30
[v4] Deprecate icon source

### LINKED ISSUE #1298: v4 Icon
Recently we've done some nice work around how icons get imported and used that means we can have fewer opinions about icons.

- Icons are no longer part of polaris-react, and instead live in polaris-icons. These icons are exported as react components using SVGr.
- Untrusted icons can now be rendered as an `` by passing a string into `<Icon source="">`

Going forward we want to support only these two styles of icon. This means making the breaking change of dropping support for the current values that the source prop accepts.

Going forward the new typing of `source` shall be `React.SFC<React.SVGProps<SVGSVGElement>> | 'placeholder'  | string;` This will allow us to remove lots of logic in the Icon element.

BPScott:

Remove deprecated functionality.

Fixes #1298 and is the final step in #1061

Icon now only supports a source of 'placeholder' to use a placeholder,
a component that returns an SVG element (i.e. something provided by
polaris-icons or svgr imports) or a string (for untrusted svgs)

- Remove passing a string into source to use a bundled icon
- Remove passing a SvgSource shaped object into source (this was the legacy shopify-style icon imports that we have replaced with using SVGr in sewing-kit)
- Remove passing a React element into source
- Remove the deprecated 'untrusted' prop

### How to 🎩

check tests and an storybook for no regressions

### DIFF
--- a/src/components/Icon/Icon.tsx
+++ b/src/components/Icon/Icon.tsx
@@ -1,59 +1,5 @@
 import React from 'react';
 import {classNames, variationName} from '@shopify/css-utilities';
-import {
-  PlusMinor,
-  AlertMinor,
-  ArrowDownMinor,
-  ArrowLeftMinor,
-  ArrowRightMinor,
-  ArrowUpMinor,
-  ArrowUpDownMinor,
-  CalendarMinor,
-  MobileCancelMajorMonotone,
-  CancelSmallMinor,
-  CaretDownMinor,
-  CaretUpMinor,
-  TickSmallMinor,
-  ChevronDownMinor,
-  ChevronLeftMinor,
-  ChevronRightMinor,
-  ChevronUpMinor,
-  CircleCancelMinor,
-  CircleChevronDownMinor,
-  CircleChevronLeftMinor,
-  CircleChevronRightMinor,
-  CircleChevronUpMinor,
-  CircleInformationMajorTwotone,
-  CirclePlusMinor,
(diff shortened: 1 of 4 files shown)


src/components/Icon/Icon.tsx
```diff
@@ -219,42 +95,21 @@ function Icon({
...
+      <SourceComponent
+        className={styles.Svg}
+        focusable="false"
+        aria-hidden="true"
+      />
     );
-    contentMarkup = <SourceComponent {...defaultIconProps} />;
-  } else if (typeof source === 'function') {
-    const sourceElement = <SourceComponent {...defaultIconProps} />;
     if (React.isValidElement(sourceElement)) {
```
amrocha: Should we throw an error here?

BPScott: I don't think it's worth having this if check in the first place and that we should always do `contentMarkup = sourceElement` - if you pass a function that returns something that isn't a react element you deserve an explosion when trying to render it.

We don't need to do any error handling when "React can't render the result of that function" will happen anyway.

amrocha: I agree, but sometimes it's nice to throw an explicit error with a nicer error message than the default one. I'll leave it up to you

BPScott: Gonna kill it. In our most common use-case of internal use people will be using typescript which shall scowl at them before they even run the code.

PR #1605, opened 2019-05-30
[DataTable] Remove fixed column UX

chloerice:

Fixes #1426 

The absolute positioning of the first column cells is no longer working across supported browsers, specifically Safari iOS (which means it doesn't work in any mobile browser in iOS). After @dleroux and I each explored other ways to accomplish the same UX with different CSS, there wasn't a clear path forward that doesn't cause other parts of the table to break. 

I concluded that the fixed column experience is not worth the hacks or tradeoffs needed to fix in iOS nor the risk of future breaks. Removing the fixed first column enables removal of all the excessive complexity that went into maintaining the default table behavior of row height change etc.

- [x] Removes table collapse/fixed first column 
- [x] Removes set width of first column unless `truncate` is true
- [x] Removes the logic that calculates and updates row heights for non-truncated tables
- [x] Removes the logic that preserves the horizontal scroll position of the table after sort of non-truncated tables

## 

### How to 🎩

<details>

```jsx
import {
  Page,
  Card,
  Link,
  DataTable,
  SortDirection,
  ColumnContentType,
  TableData,
} from '../src/components';

interface State {
  truncate: boolean;
  sortedRows?: TableData[][];
}

  state: State = {
    truncate: false,
  };

  sortCurrency = (
    rows: TableData[][],
    index: number,
    direction: SortDirection,
  ) => {
    return [...rows].sort((rowA: string[], rowB: string[]) => {
      const amountA = parseFloat(rowA[index].substring(1));
      const amountB = parseFloat(rowB[index].substring(1));

      return direction === 'descending' ? amountB - amountA : amountA - amountB;
    });
  };

  sortText = (rows: TableData[][], index: number, direction: SortDirection) => {
    function extractText(cellContent: TableData) {
      return typeof cellContent === 'string'
        ? cellContent.toLowerCase()
        : cellContent.props.children.toLowerCase();
    }

    return [...rows].sort((rowA: TableData[], rowB: TableData[]) => {
      const textA = extractText(rowA[index]);
      const textB = extractText(rowB[index]);

      if (textA > textB) {
        return direction === 'descending' ? -1 : 1;
      }
      if (textA < textB) {
        return direction === 'descending' ? 1 : -1;
      }

      return 0;
    });
  };

  toggleTruncate = () => {
    this.setState(({truncate}) => ({truncate: !truncate}));
  };

  handleSort = (rows: TableData[][], contentTypes: string[]) => (
    index: number,
    direction: SortDirection,
  ) => {
    if (contentTypes[index] === 'text') {
      return this.setState({
        sortedRows: this.sortText(rows, index, direction),
      });
    }
    this.setState({sortedRows: this.sortCurrency(rows, index, direction)});
  };

  render() {
    const {sortedRows, truncate} = this.state;
    const columnContentTypes: ColumnContentType[] = [
      'text',
      'numeric',
      'numeric',
      'numeric',
      'numeric',
    ];

    const initiallySortedRows = [
      [
        <Link url="https://www.example.com">Emerald Silk Gown</Link>,
        '$875.00',
        124689,
        140,
        '$122,500.00',
      ],
      [
        <Link url="https://www.example.com">Mauve Cashmere Scarf</Link>,
        '$230.00',
        124533,
        83,
        '$19,090.00',
      ],
      [
        <Link url="https://www.example.com">
          Navy Merino Wool Blazer with khaki chinos and yellow belt
        </Link>,
        '$445.00',
        124518,
        32,
        '$14,240.00',
      ],
    ];

    return (
      <Page title="Playground">
        <Card
          title="Data table with fixed column UX removed"
          actions={[
            {content: 'Toggle truncate', onAction: this.toggleTruncate},
          ]}
        >
          <DataTable
            truncate={truncate}
            columnContentTypes={columnContentTypes}
            headings={[
              'Product',
              'Price',
              'SKU Number',
              'Net quantity',
              'Net sales',
            ]}
            rows={rows}
            footerContent={`Showing ${rows.length} of ${rows.length} results`}
            sortable={[true, true, false, false, true]}
            initialSortColumnIndex={4}
            onSort={this.handleSort(rows, columnContentTypes)}
          />
        </Card>
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/DataTable/DataTable.scss
+++ b/src/components/DataTable/DataTable.scss
@@ -90,6 +90,7 @@ $breakpoint: 768px;
   white-space: nowrap;
   overflow-x: hidden;
   text-overflow: ellipsis;
+  max-width: $first-column-width;
 }
 
 .Cell-header {
--- a/src/components/DataTable/DataTable.scss
+++ b/src/components/DataTable/DataTable.scss
@@ -20,12 +20,6 @@ $breakpoint: 768px;
   }
 }
 
-.hasFooter {
-  .ScrollContainer {
-    margin-bottom: rem(52px);
-  }
-}
-
 .Navigation {
   display: none;
 }
@@ -61,10 +55,6 @@ $breakpoint: 768px;
       background: color('sky', 'lighter');
     }
   }
-}
-
-.TableFoot {
-  border-bottom: 0;
 }
 
 .Cell {
@@ -148,16 +138,9 @@ $breakpoint: 768px;
   border-bottom: border();
 }
 
-.Cell-footer {
-  @include text-emphasis-normal;
-  position: absolute;
-  top: 100%;
-  left: 0;
-  width: 100%;
-  border-bottom: 0;
+.Footer {
+  padding: spacing();
   background: color('sky', 'light');
   color: color('ink', 'lighter');
-  white-space: unset;
   text-align: center;
-  backface-visibility: hidden; // stop painting on scroll (due to positioning)
 }
--- a/src/components/DataTable/DataTable.scss
+++ b/src/components/DataTable/DataTable.scss
@@ -1,4 +1,4 @@
-$fixed-column-width: rem(145px);
+$first-column-width: rem(145px);
 $breakpoint: 768px;
 
 .DataTable {
@@ -26,7 +26,7 @@ $breakpoint: 768px;
   }
 
   .ScrollContainer {
-    margin-left: rem($fixed-column-width);
+    margin-left: rem($first-column-width);
   }
 }
 
@@ -71,7 +71,7 @@ $breakpoint: 768px;
     position: absolute;
     top: 0;
     bottom: 0;
-    left: $fixed-column-width;
+    left: $first-column-width;
     display: none;
     width: rem(6px);
     background: linear-gradient(
@@ -111,9 +111,10 @@ $breakpoint: 768px;
   @include text-emphasis-normal;
   @include text-breakword;
   position: absolute;
+  z-index: 5;
   top: auto;
   left: 0;
-  width: $fixed-column-width;
+  width: $first-column-width;
   white-space: unset;
   text-align: left;
   backface-visibility: hidden; // stops painting on scroll (due to positioning)
--- a/src/components/DataTable/DataTable.scss
+++ b/src/components/DataTable/DataTable.scss
@@ -6,7 +6,7 @@ $breakpoint: 768px;
   max-width: 100vw;
 }
 
-.collapsed {
+.condensed {
   .Navigation {
     display: flex;
     align-items: center;
@@ -78,7 +78,6 @@ $breakpoint: 768px;
 
 .Cell-firstColumn {
   @include text-emphasis-normal;
-  @include text-breakword;
   text-align: left;
   white-space: normal;
 }
--- a/src/components/DataTable/DataTable.scss
+++ b/src/components/DataTable/DataTable.scss
@@ -7,12 +7,6 @@ $breakpoint: 768px;
 }
 
 .collapsed {
-  .Table {
-    &::after {
-      display: block;
-    }
-  }
-
   .Navigation {
     display: flex;
     align-items: center;
@@ -23,10 +17,6 @@ $breakpoint: 768px;
     @include breakpoint-after($breakpoint, inclusive) {
       justify-content: flex-end;
     }
-  }
-
-  .ScrollContainer {
-    margin-left: rem($first-column-width);
   }
 }
 
@@ -57,29 +47,12 @@ $breakpoint: 768px;
 
 .ScrollContainer {
   overflow-x: auto;
-  // account for a mysterious gap in Safari when not collapsed
-  margin-left: rem(140px);
   -webkit-overflow-scrolling: touch;
 }
 
 .Table {
   width: 100%;
   border-spacing: 0;
-
-  &::after {
-    content: '';
-    position: absolute;
-    top: 0;
-    bottom: 0;
-    left: $first-column-width;
-    display: none;
-    width: rem(6px);
-    background: linear-gradient(
-      to right,
-      rgba(color('black'), 0.12),
-      rgba(color('black'), 0)
-    );
-  }
 }
 
 .TableRow {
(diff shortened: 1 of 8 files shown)


AndrewMusgrave:
Love the 🍏 : 🔴 ratio, going to 🎩 this tonight!

src/components/DataTable/DataTable.scss
```diff
@@ -103,20 +76,17 @@ $breakpoint: 768px;
...
+.Cell-firstColumn {
   @include text-emphasis-normal;
   @include text-breakword;
-  position: absolute;
-  top: auto;
-  left: 0;
-  width: $fixed-column-width;
-  white-space: unset;
+  min-width: $first-column-width;
+  max-width: $first-column-width;
```
dleroux: Do you know why min and max work but not width?

dleroux:
The code looks good. Just a question about the min-width. Also, any way to increase code coverage or is this mostly a Percy job?

chloerice:
> LGTM ✅ 🎉
> Screen Shot 2019-06-14 at 12 50 59 PM
> I did notice the dots are filled in before it reaches the end, not sure if this is intentional and also not related to the PR but thought I'd mention it.

That's intentional, the dots indicate which columns are visible by at least 30%.

chloerice:
> The code looks good. Just a question about the min-width. Also, any way to increase code coverage or is this mostly a Percy job?

I think @amrocha wrote some tests to bump code coverage in data table during Frideation a week or two ago that will bump coverage over 90% when they ship. I'm all for getting to 90%, but none of the code in this PR is coverable by anything but Percy and this bug is keeping all tables from displaying their first column across mobile so I don't think codecov passing should block shipping this.

src/components/DataTable/DataTable.scss
chloerice: Good question! I think I originally took the set width out and had no min/max, but the first column very easily takes up the entire screen on small devices when the content is long, I was playing with adding either a max or a min but forgot to take it back out. What I'm thinking is that normal table behavior should be the default, and if implementations want to have set minimum or maximum we can always enhance with a prop.

src/components/DataTable/DataTable.tsx
```diff
@@ -114,94 +105,35 @@ class DataTable extends React.PureComponent<CombinedProps, DataTableState> {
...
-    ) : null;
+    const headingMarkup = <tr>{headings.map(this.renderHeadings)}</tr>;
 
     const totalsMarkup = totals ? (
       <tr>{totals.map(this.renderTotals)}</tr>
     ) : null;
 
-    const headingMarkup = (
-      <tr>
-        {headings.map((heading, headingIndex) => {
```
chloerice: (extracted this anonymous callback into a class method like the other render{Row} callbacks, see 126 and 246)

src/components/DataTable/DataTable.tsx
```diff
@@ -230,99 +158,48 @@ class DataTable extends React.PureComponent<CombinedProps, DataTableState> {
...
-              {footerMarkup}
             </table>
           </div>
+          {footerMarkup}
         </div>
       </div>
     );
   }
 
-  private tallestCellHeights = () => {
```
chloerice: No more need to calculate/set/update row heights manually anymore. With the absolute positioning removed, normal table wrapping and resizing behavior is restored 🙌

src/components/DataTable/DataTable.tsx
```diff
@@ -230,99 +158,48 @@ class DataTable extends React.PureComponent<CombinedProps, DataTableState> {
...
-      if (left) {
-        scrollContainer.scrollLeft = left;
-      }
-      if (top) {
-        window.scrollTo(0, top);
-      }
-    }
-  };
-
-  private setHeightsAndScrollPosition = () => {
```
chloerice: The scroll position needed to be preserved when the row heights were being managed manually because the horizontal scroll position would be reset on rerender (most notably after clicking a sortable cell heading). This is no longer an issue.

src/components/DataTable/DataTable.tsx
```diff
@@ -435,25 +338,9 @@ class DataTable extends React.PureComponent<CombinedProps, DataTableState> {
     );
   };
 
-  private renderFooter = () => {
```
chloerice: Removing the absolute positioning of the first column cells made positioning the footer content seem egregious. I've changed the footer content to be rendered within a `div` as a sibling to table's scroll container so calculating/setting/updating the footer's height and the scroll container's bottom margin is no longer needed.

src/components/DataTable/DataTable.tsx
```diff
@@ -479,16 +366,6 @@ class DataTable extends React.PureComponent<CombinedProps, DataTableState> {
         () => {
           if (onSort) {
             onSort(headingIndex, newSortDirection);
-
-            if (!truncate && this.scrollContainer.current) {
-              const preservedScrollPosition = {
```
chloerice: (See https://github.com/Shopify/polaris-react/pull/1605/files#r294746700)

src/components/DataTable/components/Cell/Cell.tsx
```diff
@@ -66,13 +59,8 @@ function Cell({
...
 
   const iconClassName = classNames(sortable && styles.Icon);
-
-  const style = {
-    height: height ? `${height}px` : undefined,
-  };
-
   const direction = sorted ? sortDirection : defaultSortDirection;
-  const source = direction === 'ascending' ? CaretUpMinor : CaretDownMinor;
+  const source = direction === 'descending' ? CaretDownMinor : CaretUpMinor;
```
chloerice: (fixes icon to align with sort defaulting to ascending)

chloerice:
@dleroux @AndrewMusgrave After spending tonight working on codecov, I realized and removed a bunch of additional code that was also no longer needed because we are scrapping the fixed first column. Essentially, this component is a lot easier to reason about without that UX 🎉

amrocha:
I have written some tests for Data Table in a local branch last frideations, but haven't had the time to open a PR yet. I'll see if I can do that this friday, but don't let that stop you from shipping this :)

src/components/DataTable/DataTable.tsx
```diff
@@ -230,99 +162,48 @@ class DataTable extends React.PureComponent<CombinedProps, DataTableState> {
...
+
       const tableData = {
-        fixedColumnWidth,
         firstVisibleColumnIndex,
         tableLeftVisibleEdge,
         tableRightVisibleEdge,
       };
 
-      const columnVisibilityData = collapsedHeaderCells.map(
+      const columnVisibilityData = Array.from(headerCells).map(
```
dleroux: Array.from is used a few times here. Judging by this https://www.measurethat.net/Benchmarks/ShowResult/59877looks like spreading is more performant.

chloerice: We have to change `headerCells` from a NodeList to an Array, it's not spreadable unfortunately.

chloerice: You know what scratch that. I just tried it and TS doesn't yell about it anymore!

PR #1606, opened 2019-05-30
[v4][Page] Narrow width page

### LINKED ISSUE #1572: [v4][Page] Change singleColumn prop to narrowWidth
## Issue summary

`Page` has a prop named `singleColumn` and it's very deceptive. Page doesn't control layout and has nothing to do with columns. Tim and I have discussed renaming this prop `narrowWidth` to better represent its use.

Deprecation/Upgrade path
- Add `narrowWidth` to the component API and map it to `singleColumn`
- Remove `singleColumn` from documentation and replace it with `narrowWidth` (we'll have to properly compose our types to continue allowing `singleColumn`)
- Use deprecation warnings to inform consumers that `singleColumn` will be replaced with `narrowWidth` in `v5`

Using this approach consumers reading our documentation will start to use `narrowWidth` instead of `singleColumn`. Consumers will have an entire major version to change `singleColumn` to `narrowWIdth`.

AndrewMusgrave:

Fixes #1572

@tmlayton For documentation, what do you think about sectioning out deprecated props rather than removing them? Where `Deprecated Props` could be our `DeprecatedProps` interface?

Screen Shot 2019-05-30 at 3 02 34 PM

### DIFF
--- a/src/components/Page/Page.tsx
+++ b/src/components/Page/Page.tsx
@@ -22,7 +22,7 @@ export interface Props extends HeaderProps {
   /** Remove the normal max-width on the page */
   fullWidth?: boolean;
   /** Decreases the maximum layout width. Intended for single-column layouts */
-  singleColumn?: boolean;
+  narrowWidth?: boolean;
   /**
    * Force render in page and do not delegate to the app bridge TitleBar action
    * @default false
@@ -32,7 +32,14 @@ export interface Props extends HeaderProps {
   forceRender?: boolean;
 }
 
-export type ComposedProps = Props & WithAppProviderProps;
+export interface DeprecatedProps {
+  /** Decreases the maximum layout width. Intended for single-column layouts
+   * @deprecated As of release 4.0, replaced by {@link https://polaris.shopify.com/components/structure/page#props-narrow-width}
+   */
+  singleColumn?: boolean;
+}
+
+export type ComposedProps = Props & DeprecatedProps & WithAppProviderProps;
 
 const APP_BRIDGE_PROPS: (keyof Props)[] = [
   'title',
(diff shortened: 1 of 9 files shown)


src/components/Page/Page.tsx
```diff
@@ -32,7 +32,14 @@ export interface Props extends HeaderProps {
   forceRender?: boolean;
 }
 
-export type ComposedProps = Props & WithAppProviderProps;
+export interface DeprecatedProps {
```
AndrewMusgrave: If we wanted to, we could parse for this interface in the style guide (see bottom of the screenshot)

src/components/Page/Page.tsx
```diff
@@ -32,7 +32,14 @@ export interface Props extends HeaderProps {
...
+  /** Decreases the maximum layout width. Intended for single-column layouts
+   * @deprecated As of release 4.0, replaced by {@link https://polaris.shopify.com/components/structure/page#props-narrow-width}
```
AndrewMusgrave: We don't currently use these tags in the style guide, but ids should be added to the props none the less

src/components/Page/Page.tsx
```diff
@@ -180,4 +200,4 @@ export class Page extends React.PureComponent<ComposedProps, never> {
   }
 }
 
-export default withAppProvider<Props>()(Page);
+export default withAppProvider<Props & DeprecatedProps>()(Page);
```
AndrewMusgrave: `<Props & DeprecatedProps>` is our public API

src/components/Page/tests/Page.test.tsx
```diff
@@ -383,6 +383,20 @@ describe('<Page />', () => {
       restoreTitleBarCreateMock();
     });
   });
+
+  describe('deprecations', () => {
+    it('warns the singleColumn prop has been renamed', () => {
+      const warningSpy = jest
+        .spyOn(console, 'warn')
+        .mockImplementation(() => {});
```
AndrewMusgrave: Blocking the warning spam in our test output

src/components/Page/Page.tsx
tmlayton: I prefer this as well—using the correct tags and following standards. That way it is there if we want to support it in the future and/or change to tool we used to parse this, we have them.

tmlayton:
> For documentation, what do you think about sectioning out deprecated props rather than removing them? Where Deprecated Props could be our DeprecatedProps interface?

I like it

PR #1610, opened 2019-05-31
[v4] Remove WithRef component and react-compose utility

BPScott:

Removing now unused utilties, and killing a dependant package in the process
Part of #1484

Remove WithRef component and react-compose utility
Remove dependency on @shopify/react-compose package

### How to 🎩

Check tests + storybook for no regressions


AndrewMusgrave:
Part of #1484

PR #1613, opened 2019-06-02
[Code Quality] Replace rem(8px) with spacing(tight)

beefchimi:
I stumbled upon two instances where we were using `rem(8px)` for padding values. Just swapping that out for `spacing(tight)`.

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -130,8 +130,8 @@ $spinner-size: rem(20px);
 .plain {
   @include recolor-icon(color('blue'));
   margin: (-1 * $vertical-padding) rem(-8px);
-  padding-left: rem(8px);
-  padding-right: rem(8px);
+  padding-left: spacing(tight);
+  padding-right: spacing(tight);
   background: transparent;
   border: 0;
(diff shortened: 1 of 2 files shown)


beefchimi:
@ry5n definitely makes sense - I did not replace every instance of `rem(8px)` for this reason _(there are a few spots where the value simply appears to be coincidental)_.

In the case of `Button` however, the `left/right` padding I am overriding is `spacing()`, so it seemed like the relationship between `Button` and `spacing-map` was pre-established:

https://github.com/Shopify/polaris-react/blob/0bcc667eff97ac876c9357a220d4f3cc3f37909c/src/styles/shared/buttons.scss#L19

ry5n:
Ah, I see. That makes sense. You’re good 👍

ry5n:
Thanks @beefchimi! 🎉

chloerice:
> I don’t think every instance of a value (say 8px) should be tied to the spacing function. Does that make sense?

That's good to know when reviewing feature work, in my time on feature teams this is def how we use the `spacing()` function currently (tying everything to the `spacing()` function)

PR #1615, opened 2019-06-03
[TextField] [a11y] Define aria-labelledby only if multiple labels are required

francinen:

Fixes https://github.com/Shopify/polaris-react/issues/1587

The `TextField` component is unnecessarily defining the input's `aria-label` and `aria-labelledby` attributes in cases where the value of the label node is sufficient on its own.

As mentioned in https://github.com/Shopify/polaris-react/issues/1587, "Only one method is needed, and extra methods may cause duplicate reading by assistive tech."

Screen Shot 2019-06-03 at 2 06 48 PM

This pull request implements the recommended behaviour outlined in https://github.com/Shopify/polaris-react/issues/1587.

1. Use only the default <label> with the for/id pair by default, and do not set the `aria-label` and `aria-labelledby` attributes

Screen Shot 2019-06-04 at 1 48 43 PM

Screen Shot 2019-06-03 at 2 26 47 PM

2. If multiple labels are required, set the `aria-labelledby` attribute

Screen Shot 2019-06-04 at 1 52 18 PM

Screen Shot 2019-06-03 at 2 28 07 PM

### How to 🎩

<details>

```jsx

interface State {}

  render() {
    return (
      <Page title="Playground">
         <TextField label="Store name" prefix="Hello" suffix="Goodbye" />
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -310,12 +310,16 @@ class TextField extends React.PureComponent<CombinedProps, State> {
       describedBy.push(`${id}CharacterCounter`);
     }
 
-    const labelledBy = [labelID(id)];
+    const labelledBy: string[] = [];
     if (prefix) {
       labelledBy.push(`${id}Prefix`);
     }
     if (suffix) {
       labelledBy.push(`${id}Suffix`);
+    }
+
+    if (labelledBy.length) {
+      labelledBy.unshift(labelID(id));
     }
 
     const inputClassName = classNames(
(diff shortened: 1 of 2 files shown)


src/components/TextField/TextField.tsx
```diff
@@ -310,14 +310,18 @@ class TextField extends React.PureComponent<CombinedProps, State> {
...
+    const labelledBy: string[] = [];
     if (prefix) {
       labelledBy.push(`${id}Prefix`);
     }
     if (suffix) {
       labelledBy.push(`${id}Suffix`);
     }
 
+    if (labelledBy.length) {
+      labelledBy.unshift(labelID(id));
```
chloerice: I realize this is literally putting the text field label in the same position as it currently would be, but looking at this has me wondering a couple of things @dpersing:

- Does the order matter when a field is labelled by multiple things? In other words, do screen readers read the label text, prefix text content (if any) and the suffix text (if any)? Or do they read the `aria-labelledby` value itself word for word to describe the field?

dpersing: >Does the order matter when a field is labelled by multiple things?

The `id` values are supposed to set the reading order, but since we're kind of doubling up on those reading the `<label>` first seems to be the default. Generally AT will default to reading HTML native features if they're available, then read ARIA features. Generally. In theory. 😂

francinen:
Thank you both for such prompt reviews!

PR #1621, opened 2019-06-04
[v4][PositionedOverlay] Remove component will receive props

AndrewMusgrave:

Part of https://github.com/Shopify/polaris-react/issues/519 and https://github.com/Shopify/polaris-react-deprecated/pull/506

* using mutation observer to watch for changes
* add test support for mutation observer. MutationObserver support was [finally added after 5 years](https://github.com/jsdom/jsdom/issues/639). However, jest [supports node 6](https://github.com/facebook/jest/pull/7763) and won't be updating `jsdom` for a while.

### How to 🎩

Autocomplete examples are a perfect case of resizing from changes. The original issues (https://github.com/Shopify/polaris-react-deprecated/pull/506) also has a playground included.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -9,3 +9,4 @@ node_modules
 /styles.scss
 /styles
 /src/styles/polaris-tokens
+tests/shims/mutation-observer.ts
--- a/src/components/PositionedOverlay/PositionedOverlay.tsx
+++ b/src/components/PositionedOverlay/PositionedOverlay.tsx
@@ -52,6 +52,8 @@ export interface State {
   lockPosition: boolean;
 }
 
+const OBSERVER_CONFIG = {childList: true, subtree: true};
+
 export default class PositionedOverlay extends React.PureComponent<
   Props,
   State
@@ -71,6 +73,13 @@ export default class PositionedOverlay extends React.PureComponent<
 
   private overlay: HTMLElement | null = null;
   private scrollableContainer: HTMLElement | Document | null = null;
+  private observer: MutationObserver;
+
+  constructor(props: Props) {
+    super(props);
+
+    this.observer = new MutationObserver(this.handleMeasurement);
+  }
 
   componentDidMount() {
     this.scrollableContainer = Scrollable.forNode(this.props.activator);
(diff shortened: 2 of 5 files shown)


.eslintignore
```diff
@@ -9,3 +9,4 @@ node_modules
 /styles.scss
 /styles
 /src/styles/polaris-tokens
+tests/shims/mutation-observer.ts
```
AndrewMusgrave: The copied shim

src/components/PositionedOverlay/PositionedOverlay.tsx
```diff
@@ -153,6 +158,8 @@ export default class PositionedOverlay extends React.PureComponent<
   private handleMeasurement = () => {
     const {lockPosition, top} = this.state;
 
+    this.observer.disconnect();
```
AndrewMusgrave: Disconnect before we perform our handle measurements as it would cause an infinite loop

src/components/PositionedOverlay/PositionedOverlay.tsx
```diff
@@ -220,23 +227,29 @@ export default class PositionedOverlay extends React.PureComponent<
...
+              onScrollOut != null &&
+              rectIsOutsideOfRect(
+                activatorRect,
+                intersectionWithViewport(scrollableContainerRect),
+              ),
+            zIndex,
+          },
+          () => {
+            if (!this.overlay) return;
+            this.observer.observe(this.overlay, OBSERVER_CONFIG);
```
AndrewMusgrave: Re-observer after it's complete

tests/setup.ts
```diff
@@ -1,5 +1,6 @@
 import {configure} from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
+import './shims/mutation-observer';
```
AndrewMusgrave: Importing the mutation observer shim

tests/shims/mutation-observer.ts
```diff
@@ -0,0 +1,723 @@
+/* !
+ * Shim for MutationObserver interface
+ * Author: Graeme Yeates (github.com/megawac)
+ * Repository: https://github.com/megawac/MutationObserver.js
```
AndrewMusgrave: Copied the shim from here and added types

Issue #1617, opened 2019-06-03
Yarn audit failure


Introvertuous:
## Issue summary

yarn audit

```
555 vulnerabilities found - Packages audited: 82316
Severity: 305 Low | 234 Moderate | 16 High
```

## Expected behavior

no vulnerabilities

## Actual behavior

```
555 vulnerabilities found - Packages audited: 82316
Severity: 305 Low | 234 Moderate | 16 High
```

Usually yarn audit failures are easy to fix by resetting yarn.lock, and upgrading large devDependencies like babel or jest.   It's not so straightforward here because many of the dependencies come in through @shopify/js-uploader and @shopify/sewing-kit.

## Steps to reproduce the problem

1.  git clone git@github.com:Shopify/polaris-react.git
1.  cd polaris-react
1.  yarn audit

## Specifications

- Polaris version number: 6e654ccf29d09406b1d1fda846ce4e8e46a67c02

chloerice:
Thank you for raising this, @Introvertuous. Triaged and will discuss next steps with the team tomorrow.

chloerice:
Hey @Introvertuous, just quick update--we're working on upgrades now. Most of the audit is noise, as its all development dependencies so there’s no risk to end users.

Introvertuous:
Thank you for the quick response / action. It is true that a lot of the audit is caused via dev dependencies, but I think these should be resolved at some point (possibly as lower priority) because it is still a concern for anyone building polaris, which is what we are doing.

BPScott:
We've fixed the majority of these warnings in #1624 and #1633 Once those two are merged we're down to "Severity: 6 Low | 8 Moderate". and most of the remaining issues are down to js-uploader using hardcoded old versions.

PR #1624, opened 2019-06-05
Update dependency versions to quieten yarn audit

BPScott:

Partial fix for #1617.

These audits are mostly noise. They're complaining about our development dependencies and thus in a place that exposes a security concern for our users. That said it's nice to cut down on noise.

Updating a bunch of our transitive dependency versions

Before:
> 555 vulnerabilities found - Packages audited: 82316 // Severity: 305 Low | 234 Moderate | 16 High

After:
> 78 vulnerabilities found - Packages audited: 93837 // Severity: 70 Low | 8 Moderate

This doesn't fix everything but it gets the low-hanging fruit.

The rest of this can be solved by

- kicking js-uploader to not used fixed old versions for dependencies (while you're fixing this can you publish it to npm so we don't have to rely on a github dependency as I think that slows installs down)
- updating sewing-kit to latest which in turn shall update jest from v23 to v24 which should fix all those complaints about `braces`.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1211,6 +1211,15 @@
     "@jest/types" "^24.7.0"
     "@types/istanbul-lib-coverage" "^2.0.0"
 
+"@jest/test-result@^24.8.0":
+  version "24.8.0"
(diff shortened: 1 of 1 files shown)


BPScott:
All those test failures look unrelated to these changes.

PR #1628, opened 2019-06-05
[v4][WithinContentContext]Export WithinContentContext

### LINKED ISSUE #1303: Revisiting Banner's withinContentContainer context
# Context

`<Banner>` renders differently based upon if it is rendered within a "content container" - a `<Card>` or `<Modal>` component  or not. It does this by consuming a  context called `withinContentContainer` which both `<Card>` and `<Modal>` set. You can see this in action in Storybook by comparing the "default banner" and "banner in a card" stories.

While this is nice as it ensures the correct visual behaviour and "it just works™" in Cards and Modals it does go against the component design tenant of "define variations of components, don't style based on the container your component lives in".

Occasionally there are cases where it is useful to use the "within content" style of a banner not within a Card or Modal, which means that in these cases [awkward hacks have to be used to create a context](https://github.com/Shopify/polaris-icons/blob/1cc7792b6ed0299217bb35110bc7498508898a4e/packages/app-icon-explorer/src/components/IconDetailsPanel/IconDetailsPanel.tsx#L30-L38). This also occurs in Web's StoreSwitcher. I'm fairly sure that this will stop working in v4 as we move from the legacy to modern context API.

This then gets more complicated as there are [two Card-style components in web](https://github.com/Shopify/web/search?utf8=%E2%9C%93&q=withinContentContainer&type=) that mock out the context too.

Having to rely on creating a context to control the layout of a component feels weird and I don't really like it. Alas I wasn't here when this was originally implemented so I'd love a bit more context on the thinking behind this decision at the time. I think @danrosenthal is the man with the answers I seek.

---

So here's some questions:

1. Do we want to expose this context as part of our public API and recommend to people that is the way to render "within content" banners?
2. If not then should we provide some other mechanism of opting into the "within content" style? A prop that defaults to false perhaps: `<Banner inContent>` (name up for debate)
3. If we do want to provide a prop to toggle this behaviour is it confusing to have a prop that only has an effect some of the time? e.g. `<Card><Banner/></Card>` and `<Card><Banner inContent/></Card>` would render identically.
4. If we don't want to expose the context as a public API (NO on 1) and we consider the prop sometimes having no effect confusing (YES on 3) then should we even have the context at all even as an internal mechanism? Dropping this would be a big breaking change and would mean we have to hunt for all Banners-inside-a-Card-or-Modal and add the prop to them as part of the v4 upgrade in web.

---

Personal reckons:

1. I dislike exposing the context as a public API.
2. I like adding a prop to allow people to opt into this in cases where they need the in-content style but not in a Card or Modal.
3. I don't think having a Banner  prop that does nothing when used inside Card and Modal is overly confusing. It's not an ideal situation but I think this is better than making the big backwards-compat break of removing the context all together
4. I think the cost of fixing every Banner in consuming applications is a very large and I while I like ambition I don't have the appetite to take ownership for that at the moment.

---

Do people have any opinions on the above questions and any thoughts our current implementation or on alternative ways of solving this?

[For more context, we started talking about this in slack](https://shopify.slack.com/archives/G71RFHXNC/p1554921724200900)

AndrewMusgrave:

Fixes #1303

Going to skip a changelog entry since we're not documenting it

### DIFF
--- a/src/components/index.ts
+++ b/src/components/index.ts
@@ -312,3 +312,5 @@ export {default as withContext} from './WithContext';
 export {PolarisContext} from './types';
 
 export {DATA_ATTRIBUTE} from './shared';
+
+export {default as WithinContentContext} from './WithinContentContext';


AndrewMusgrave:
> I'd like to keep the components index alphabetical, bit there's a few other bits in there we need to tidy too so not the end of the world

I noticed that as well, I was going to fix it up in a follow-up PR since the exports above this one were 🚫

PR #1643, opened 2019-06-06
Add debounce

### LINKED ISSUE #1519: [v4] Remove debounce

AndrewMusgrave:

Fixes #1519

* Adding @tmlayton debounce/throttle [from polaris-rails](https://github.com/Shopify/polaris-rails/blob/master/admin/lib/utilities.js) and the [tests](https://github.com/Shopify/polaris-rails/blob/258b0d0660b92eecfb6c7b94b829f1bfc3e8d21c/test/javascripts/unit/admin/utilities_test.js) 🙇‍♂️ and typescriptify it
* replace lodash debounce with the custom debounce
* remove debounce from loading

### DIFF
--- a/src/components/Frame/components/Loading/Loading.tsx
+++ b/src/components/Frame/components/Loading/Loading.tsx
@@ -1,5 +1,4 @@
 import React from 'react';
-import debounce from 'lodash/debounce';
 import styles from './Loading.scss';
 
 export interface Props {}
@@ -20,11 +19,6 @@ export default class Loading extends React.Component<Props, State> {
     animation: null,
   };
 
-  private ariaValuenow = debounce(() => {
-    const {progress} = this.state;
-    return Math.floor(progress / 10) * 10;
-  }, 15);
-
   componentDidMount() {
     this.increment();
   }
@@ -44,7 +38,7 @@ export default class Loading extends React.Component<Props, State> {
       transform: `scaleX(${Math.floor(progress) / 100})`,
     };
 
-    const ariaValuenow = this.ariaValuenow();
+    const ariaValuenow = Math.floor(progress / 10) * 10;
 
     return (
       <div className={styles.Loading}>
--- a/UNRELEASED-V4.md
+++ b/UNRELEASED-V4.md
@@ -39,6 +39,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Code quality
 
+- Replaced all occurrences of `_.debounce` with a custom `debounce`/`throttle` function ([#1643](https://github.com/Shopify/polaris-react/pull/1643))
 - Simplified `WithinContentContainer` context type ([#1602](https://github.com/Shopify/polaris-react/pull/1602))
 - Remove `withRef` and `withContext` from `DropZone.FileUpload` ([#1491](https://github.com/Shopify/polaris-react/pull/1491))
 - Updated `OptionList` to no longer use `componentWillReceiveProps`([#1557](https://github.com/Shopify/polaris-react/pull/1557))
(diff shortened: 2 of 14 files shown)


src/components/Frame/components/Loading/Loading.tsx
```diff
@@ -20,11 +19,6 @@ export default class Loading extends React.Component<Props, State> {
     animation: null,
   };
 
-  private ariaValuenow = debounce(() => {
```
AndrewMusgrave: Debounce doesn't return a value anymore. State changes are also triggered from RAF which runs every ~16ms which is longer than the debounce time.

tmlayton:
Do we want to move these debounce and throttle functions to the JS utils library in quilt?

src/components/Frame/components/Loading/Loading.tsx
tmlayton: We should still probably update debounce to return the return value. This was an oversight on my part because all of our uses at the time were side effects.

BPScott:
@michelleyschen was working on a debounce utility in quilt in https://github.com/Shopify/quilt/pull/611 but it looks it stalled somewhat. Putting these two in quilt feels like they'll be useful to others too

michelleyschen:
It totally stalled. But if anyone wanna review it we can keep going!
Also plan on adding a hook version once this is in too.

AndrewMusgrave:
> It totally stalled. But if anyone wanna review it we can keep going! Also plan on adding a hook version once this is in too.

It looks Chris is still waiting for feedback to be addressed. Is there anything you need from us to push it through?

tmlayton:
Feel free to use this code, it is well tested and has been used in prod for the last couple years.

michelleyschen:
I have address everything in the issue, but I don't think Chris had time to follow up.
At this point a fresh new review will likely kick start it again.

BPScott:
> Putting these two in quilt feels like they'll be useful to others too

That said, I'm really liking the batteries-included approach we've been taking with our dependencies lately. I quite like that we've brought a bunch of them into the library and are shrinking our list of external dependencies.

michelleyschen:
In general I also like the batteries-included approach.
But if we CAN figure out a quilt version of this, we do have the benefit of not including a debounce utility twice.

AndrewMusgrave:
> In general I also like the batteries-included approach.
> But if we CAN figure out a quilt version of this, we do have the benefit of not including a debounce > > utility twice in web for the future

It would have the benefit of not include it twice in web, however, every other consuming application would now have an extra dependency. And I don't think a few hundred extra bytes will harm web's bundle size.

As an interim solution should we fix up this debounce to return a value and 🚢so we can 🔥 lodash and decide on if we want to add another dependency when debounce has been added to `@shopify/function-enhancers`?

michelleyschen:
What is the major con for `every other consuming application would now have an extra dependency` other than the extra byte size, since this is what you are introducing embedded anyway.

Since this debounce version had been trial and tested in production, I would actually prefer to just move this into `@shopify/function-enhancers` and kill my PR.

AndrewMusgrave:
> What is the major con for every other consuming application would now have an extra dependency other than the extra byte size, since this is what you are introducing embedded in polaris anyway.

Personal opinion but I tend to prefer packages without (or minimal) dependencies. Dependency hell is a huge problem 😬 and I prefer to keep small utilities included (when it makes sense).

> Since this debounce version had been trial and tested in production, I would actually prefer to just move this into @shopify/function-enhancers and kill my PR.

Let's move forward with this, did you want to handle this @michelleyschen or would you rather I push this through?

michelleyschen:
@AndrewMusgrave I likely won't have the bandwidth to take this one till late next week.
If you can tackle it earlier go ahead, otherwise I will add it to my list of ToDo no matter what.

AndrewMusgrave:
@michelleyschen I can work on adding it tomorrow tonight or Wednesday 😄

PR #1653, opened 2019-06-07
🏗 [ActionMenu] Extract `Page > Header` code into its own component

beefchimi:
**Warning!** ~This PR introduces `1 breaking change`: both `secondaryActions` and `actionGroups` must contain `content` for their individual `action`.~ This issue is now being tracked here: https://github.com/Shopify/polaris-react/issues/1676

> **Update:** This PR now almost completely refactors `Page > Header`. Not much more work will be needed to further break down into reusable components.

Resolves: https://github.com/Shopify/polaris-react/issues/1639

This PR is the first step in attempting to break apart `Page > Header` into more re-usable components. The long-term goal is to get `Page > Header` much more digestible by having it comprised of several small components. This will enable Polaris consumers to be able to build their own "page-style headers".

<details>
<summary><strong>Screenshot: Plain actions</strong></summary>
plain-actions
</details>

<details>
<summary><strong>Screenshot: Roll up actions</strong></summary>
Screen Shot 2019-06-07 at 5 02 20 PM
</details>

<details>
<summary><strong>Screenshot: Wrapping actions</strong></summary>
wrapping-actions
</details>

<details>
<summary><strong>Screenshot: Truncated Breadcrumbs</strong></summary>
truncated-breadcrumb
</details>

### Quick recap of what is happening:
- Created new `ActionMenu` component, responsible for the presentation / logic of the `actions / groups / rollup`.
- Pulled out "rollup action" markup/styles/logic into its own `RollupActions` component.
- Moved all `Header > components` into the `Menu > components` folder.
  - `Action` is now `MenuAction`
  - `ActionGroup` is now `MenuGroup`

### Additional improvements
- `ActionList > Item` now supports `accessibilityLabel`.
- `Breadcrumb` no longer breaks off the screen if the window width is too short and the breadcrumb text is too long.
  - Simply truncating the breadcrumb text. In most cases, this probably won't happen.
- No more double rendering of several components, which were previously being rendered in two places and toggled `display: none` at a breakpoint.
  - `primaryAction`.
  - "roll up actions".
- No more conditional returning different markup in `Header.tsx`.
- Some additional test coverage in a few places.
- Some various layout safety improvements, as much as code cleanup.

### Concerns
- I am now using the same `matchMedia` mobile logic as `Frame`. `Navigation > Item`, `Navigation > Section`, and `Sheet`.
  - Curious if there is a cleaner way to do this, such as sticking the mobile state onto `Context`?

### Remaining work:
- [ ] I have not written the documentation / more thorough README examples for any of the new components. I'd love some help completing that portion of the PR.

### Questions
- I don't actually like the name `Menu`. Is there something better anyone would suggest?
- How do I resolve all the failing CI tests?
  - Translation platform
  - Percy
  - pa11y
  - etc...
- Most questions have since been resolved.

<details>
<summary><strong>Preview questions</strong></summary>
<ul>
  <li>
    <p>Should `RollupActions` also be displaying the `PlainActionGroup`s `details` prop?</p>
  </li>
  <li>
    <p>Whats with the `app-bridge` stuff?</p>
    <ul>
      <li><p>See: `src/utilities/app-bridge-transformers.ts`</p></li>
      <li><p>Is there anything I need to know about how this impacts the `Header` and the changes I’ve made?</p></li>
    </ul>
  </li>
  <li>
    <p>Why does `pagination` get hidden on mobile?</p>
  </li>
  <li>
    <p>Should `PlainAction` really exist as its own component?</p>
    <ul>
      <li><p>Is it possible to roll this into `Button.tsx`?</p></li>
    </ul>
  </li>
  <li>
    <p>I don’t love that we render `rollupActionsMarkup` conditionally in two possible locations based on if `breadcrumbs` are present _( `navigationMarkup` vs `titleMarkup` )_.</p>
  </li>
  <li>
    <p>Seems like we should be able to `position: absolute; top: 0; right: 0;` the `wrapper`, and just utilize the `hasRollupActions` variant to add some `padding-right` equal to the width of the `RollupActions > <button />` _(which should be an easy width to predict, since its just the `icon-size + (button-padding * 2)`)_.</p>
  </li>
</ul>
</details>

### Tophatting

Please read over all the code and give the Playground instance a whirl.

<details>

```jsx
import {
  ActivitiesMajorMonotone,
  AffiliateMajorMonotone,
  AppsMajorMonotone,
} from '@shopify/polaris-icons';
import * as Polaris from '../src';

interface State {}

function noop() {}

const secondaryActions = [
  {
    content: 'Duplicate radness',
    accessibilityLabel: 'action label 1',
    icon: ActivitiesMajorMonotone,
    url: '#',
  },
  {
    content: 'Another one',
    accessibilityLabel: 'action label 2',
    icon: AppsMajorMonotone,
    url: '#',
  },
  {
    content: 'Online Store',
    accessibilityLabel: 'action label 3',
    onClick: noop,
  },
  {
    content: 'Pidgeon Smidgeon',
    onClick: noop,
  },
];
const actionGroups = [
  {
    title: 'Dropdown thing',
    icon: AffiliateMajorMonotone,
    actions: [
      {
        content: 'Share on Facebook',
        accessibilityLabel: 'action label 4',
        helpText: 'Here is some help text',
        role: 'Hello',
      },
      {
        content: 'Share on Twitter',
        helpText: 'Tweet tweet tweet!',
        active: true,
        role: 'Cream',
      },
    ],
    details: 'These are some details for my thing',
  },
  {
    title: 'Promote',
    accessibilityLabel: 'group label one',
    actions: [
      {
        content: 'Drink milk',
        accessibilityLabel: 'action label 5',
      },
      {
        content: 'Eat cream',
        accessibilityLabel: 'action label 6',
        url: '#',
      },
      {
        content: 'Give in to television',
        accessibilityLabel: 'action label 7',
        ellipsis: true,
        onClick: noop,
      },
    ],
  },
];

const breadcrumbsData = [
  {content: 'A longer than average Breadcrumb label', url: '/products'},
];
const paginationData = {
  hasPrevious: true,
  hasNext: true,
};

const shortTitle = null && 'Hello, I am short';
const longTitle =
  'A long Resource title that might wrap onto a second line or even three lines depending on how wide your browser is';

const metadataMarkup = (
  <Polaris.Badge status="attention">Verified</Polaris.Badge>
);

  render() {
    return (
      <Polaris.Page
        // fullWidth
        // singleColumn
        // separator
        title={shortTitle || longTitle}
        secondaryActions={secondaryActions}
        actionGroups={actionGroups}
        breadcrumbs={breadcrumbsData}
        pagination={paginationData}
        titleMetadata={metadataMarkup}
      >
        <Polaris.Card title="Staff accounts">
          <Polaris.Card.Section>
            <Polaris.List>
              <Polaris.List.Item>Felix Crafford</Polaris.List.Item>
              <Polaris.List.Item>Ezequiel Manno</Polaris.List.Item>
            </Polaris.List>
          </Polaris.Card.Section>

          <Polaris.Card.Section subdued title="Deactivated staff accounts">
            <Polaris.List>
              <Polaris.List.Item>Felix Crafford</Polaris.List.Item>
              <Polaris.List.Item>Ezequiel Manno</Polaris.List.Item>
            </Polaris.List>
          </Polaris.Card.Section>
        </Polaris.Card>
      </Polaris.Page>
    );
  }
}
```


src/components/Page/components/Header/Header.tsx
```diff
@@ -91,13 +74,16 @@ class Header extends React.PureComponent<CombinedProps, State> {
...
+      secondaryActions.length > 0 || actionGroups.length > 0;
+
     const className = classNames(
       styles.Header,
       titleHidden && styles['Title-hidden'],
-      pagination && styles['Header-hasPagination'],
-      separator && styles['Header-hasSeparator'],
       breadcrumbs && breadcrumbs.length && styles['Header-hasBreadcrumbs'],
-      this.hasRollup && styles['Header-hasRollup'],
+      separator && styles['Header-hasSeparator'],
```
beefchimi: Is this really the preferred style for variants? I thought we just did: `separator && styles.separator`

danrosenthal: Seems off, I wouldn't follow that convention.

danrosenthal:
## Unknowns:
> There are several areas of this refactor where I think we can improve things even further. As a first pass, I think this is a good start. I would love the opinions of the Polaris team to help me guide this refactor to completion.

Opinions given.

> I have not written the documentation / more thorough README examples for any of the new components. I'd love some help completing that portion of the PR.

I'm happy to help with this, and we also have documentation clinics available every other Friday. We can talk about booking you one for this.

## Questions

> Should RollupActions also be displaying the PlainActionGroups details prop?

What does the `details` prop do?

> Whats with the app-bridge stuff?

App bridge delegates rendering of the Page Header in the context of an embedded application. Leaving this stuff as-is is the best approach.

> See: src/utilities/app-bridge-transformers.ts
> Is there anything I need to know about how this impacts the Header and the changes I’ve made?

If tests continue to pass, you should be good.

> Why does pagination get hidden on mobile?

Good question. I guess for simplicity and screen real estate, since it renders at the bottom of the page.

> Should PlainAction really exist as its own component?
> Is it possible to roll this into Button.tsx?

I think it's fine to keep it as a subcomponent of my proposed `Menu` component. I don't see a lot of value in it as its own component. It doesn't seem that different from the existing `plain` variant of `Button`, so I would not incorporate it into the button component.

> I don’t love that we render rollupActionsMarkup conditionally in two possible locations based on if breadcrumbs are present ( navigationMarkup vs titleMarkup ).
Seems like we should be able to position: absolute; top: 0; right: 0; the wrapper, and just utilize the hasRollupActions variant to add some padding-right equal to the width of the RollupActions > <button /> (which should be an easy width to predict, since its just the icon-size + (button-padding * 2)).

Yeah, this is a bit weird to me too, but is not worth touching as part of this work. Keep the conditional rendering of the rollup actions as a concern of the `Header` component, and don't touch any of that for now.

beefchimi:
@chloerice + @danrosenthal I've updated the PR to implement some of Dan's suggestions for a `Menu` component.

I really wanted to avoid doing this, as it required refactoring several things in order to achieve in a sensible way. At this point, I've pretty much refactored all of `Header.tsx`.

I'd love a thorough review, as well as some assistance resolving the documentation referenced in the OP.

danrosenthal:
Looks like the Percy VRT failures on page are legitimate.

|:eyes:|
|-|
|Screen Shot 2019-06-11 at 11 19 03 AM|

It appears that there is some extra padding at the top of the `Header` and that a separator was added in cases where it does not belong.

danrosenthal:
The translation failure is fine, we can just mark those translations as asynchronous. I've done that.

Accessibility check also seems legitimate, as the button that is causing the failure does not have a name that can be read by screen readers.

Adding an entry in UNRELEASED.MD will fix the failing changelog check.

danrosenthal:
I'm happy to handle the documentation for this component.

src/components/ActionList/ActionList.tsx
```diff
@@ -34,15 +34,15 @@ export default function ActionList({
...
       <Section
         key={section.title || index}
         section={section}
-        onActionAnyItem={onActionAnyItem}
         hasMultipleSections={hasMultipleSections}
         actionRole={actionRole}
+        onActionAnyItem={onActionAnyItem}
       />
-    );
+    ) : null;
```
danrosenthal: Why was this change necessary?

src/components/Menu/components/MenuAction/MenuAction.scss
```diff
@@ -1,28 +1,24 @@
...
   display: inline-flex;
   align-items: center;
+  padding: 0 $menu-action-padding-x;
   min-height: control-height();
-
-  // Negative margin vertically for the amount that is not covered by
-  // the focus overlay
-  margin-top: -0.5 * $difference-between-touch-area-and-backdrop;
-  margin-bottom: -0.5 * $difference-between-touch-area-and-backdrop;
-  padding: 0 $action-horizontal-padding;
```
danrosenthal: Why does this get removed?

src/components/Menu/components/MenuAction/MenuAction.scss
```diff
@@ -87,18 +83,3 @@ $action-horizontal-padding: (1.5 * spacing(tight));
...
-  &::after {
-    top: auto;
-    left: auto;
-    right: auto;
-  }
-
-  &.disabled {
-    @include button-outline-disabled(color('ink', 'lighter'));
-  }
-}
```
danrosenthal: Also curious why this gets removed

beefchimi:
@danrosenthal This is for the `all header elements` example, which you will see that I updated:
https://github.com/Shopify/polaris-react/pull/1653/files#diff-63ec484b9233a6121ad12bc626225d26R186

I made this update because... it wasn't actually using `all header elements`.

As for the padding, that is a result of me removing a CSS style that I can't imagine makes any sense:
https://github.com/Shopify/polaris-react/pull/1653/files#diff-e3007b21594476114884b5991caab8aeL13

Why would be change the `padding-top` of the header when there are secondaryActions? I don't see why the addition of secondary actions should impact the top padding of the Header at all.

src/components/Page/README.md
```diff
@@ -186,12 +186,20 @@ Use for detail pages, which should have pagination and breadcrumbs, and also oft
...
+    {
+      title: 'Promote',
+      actions: [{content: 'Share on Facebook', onAction: () => {}}],
+    },
+  ]}
   pagination={{
     hasPrevious: true,
     hasNext: true,
   }}
+  separator
```
danrosenthal: Ah, I see you added a separator here. That probably accounts for that part of the failing VRT.

src/components/Page/components/Header/Header.tsx
```diff
@@ -138,159 +109,80 @@ class Header extends React.PureComponent<CombinedProps, State> {
...
-): ReadonlyArray<JSX.Element> {
-  return actions.map(({content, ...action}, index) => (
-    <div className={styles.IndividualAction} key={`Action-${content || index}`}>
-      <Action {...action}>{content}</Action>
-    </div>
-  ));
+// TODO: Can we instead get this from the <Frame />?
+// Or perhaps store in Context to be shared across components?
+function isMobileView(): boolean {
+  return navigationBarCollapsed().matches;
```
danrosenthal: Why do you feel it's necessary to move this out of this component and into context?

danrosenthal:
> I don't actually like the name Menu. Is there something better anyone would suggest?

What do you see as the issue with the `Menu` name? Did you have something else in mind?

src/components/Menu/index.ts
```diff
@@ -0,0 +1 @@
+export {default, Props, convertGroupToSection} from './Menu';
```
BPScott: Convention for component exports would be

```
export {default as Menu, Props as MenuProps} from './Menu';
```


Is `convertGroupToSection` part of the public API? I can't see it used anywhere. If not we don't need to export it

src/components/Page/components/Header/types.ts
```diff
@@ -0,0 +1,15 @@
+import {
+  AppBridgeAction,
+  DestructableAction,
+  DisableableAction,
+  LoadableAction,
+} from '../../../../types';
+
+export interface HeaderPrimaryAction
```
BPScott: This type is used both within the Page and Page.Header. In order to avoid potential for circular dependencies due to components referencing each other things like utility function and types should live at the highest level that need them.

Can you pull this up and put it in src/components/Page/types.ts instead?

src/components/ActionList/ActionList.tsx
beefchimi: Nothing within `items: ActionListItemDescriptor[];` is a required prop, so `section.items` could equal `[]`, which meant returning a useless result.

src/components/Menu/components/MenuAction/MenuAction.scss
beefchimi: At least in my own tophatting, these styles were causing misalignment, not fixing it.

src/components/Menu/components/MenuAction/MenuAction.scss
beefchimi: `outline` was never a variant this component accepted

src/components/Page/components/Header/Header.tsx
beefchimi: Maybe its not - I was initially concerned about too many components repeating the exact same code. I'm fine to remove the comment and leave it.... less work 😅

beefchimi:
@danrosenthal addressed the `a11y` issue here:
https://github.com/Shopify/polaris-react/pull/1653/commits/b6c673e55804d5a504aaabcd40abc81eba044f89

That also includes giving `ActionList > Item` the ability to render a `aria-label`. The `a11y` checks seem to be passing now 🥇 

Still working on addressing the final comments. Will update again soon.

src/components/Menu/index.ts
beefchimi: I was using it in a test - but it doesn't actually need to be exported in the `index.ts`. I will remove.

src/components/Page/components/Header/types.ts
beefchimi: I'll move it up to `Page/types.ts`

ry5n:
Is wrapping the secondary actions new behavior?

Wrapping secondary actions

The secondary actions used to roll up into “More actions ▾” based on the available screen space. Sometime last year we made some changes to roll everything up to an ellipsis to make more room on phones, but I think something got lost in translation because the rollup on medium size screens seems to no longer work.

I’m not sure if this PR is the place to address it, but if we’re changing how this works, secondary actions should roll up based on available space, and then transition to a single ellipsis menu on phones.

It would be great to make the rollup behavior reusable—that would be super valuable. For example, the Tabs component has a similar rollup based on screen space. Could the new RollupActions component be used for Tabs too?

beefchimi:
@ry5n The existing `Page > Header` "roll up actions" behaviour was simply based on a window breakpoint. In fact, it didn't even use JS logic to "roll up the actions" - the code simply rendered non-rolledup actions + the rollup popover, and toggled between them with `@media + display: block` in CSS 😢 

The new behaviour at the very least improves this by having the `matchMedia + mobileView` within the JSX.

The wrapping onto a second line behaviour - while not ideal - is actually an improvement on what this component does today. Here is a screenshot from the Styleguide _(when using enough secondaryActions + actionGroups)_:

Screen Shot 2019-06-11 at 2 24 26 PM

I totally agree that the ideal behaviour would maybe be to roll up once a full line width is reached. That change in behaviour is something that should be tracked as a new issue, and tackled as a follow up.

As for re-using RollupActions, I definitely want that as well. Not much point in redundant/slightly-inconsistent code + ui. That being said, I also believe this should be a separate issue + follow up PR.

ry5n:
The current page actions behavior is a regression. I logged it as a separate issue here: https://github.com/Shopify/polaris-react/issues/1660

beefchimi:
In addition to the issue `ry5n` opened, I also opened an issue to track the README: https://github.com/Shopify/polaris-react/issues/1661

@danrosenthal I believe the only comment of yours left unaddressed is:

> determining if the center alignment of the rollup popover is a problem (it should probably be aligned with its activator)

Did you just want me to slap on the `preferredAlignment="left"` prop for `MenuGroup > Popover`? I can do that quickly and fixup into the previous commit.

dleroux:
Whats the extra line about in your screens?

59133476-2b60f000-8946-11e9-8737-ffd2ab102743

beefchimi:
@dleroux 

> Whats the extra line about in your screens?

Scrollable Popover's are supposed to "start at the bottom and scroll to the top" in order to convey that "there is more content here!".

That's what you are seeing in the screenshot... there is content ABOVE what is being shown, and the line you refer to is a divider in the action list sections.

beefchimi:
@danrosenthal the only remaining piece is getting 90% codecov. The issue seems to be because of the `mobile` state logic in `Header.tsx` - not really sure how to change my tests to better cover those lines.

Everything I've done mimics the code in `Sheet.tsx`. I ran the local `test:coverage` on both files to compare... but to be honest I don't fully understand what these results mean.

#### Sheet.test.tsx
sheet

### Header.test.tsx
header

src/components/Menu/Menu.tsx
```diff
@@ -87,11 +87,8 @@ export default class Menu extends React.PureComponent<Props, State> {
     ) : null;
   };
 
-  private handleMenuGroupClose = (group: string) => {
-    this.setState(
-      ({activeMenuGroup}) =>
-        activeMenuGroup === group ? {activeMenuGroup: undefined} : {},
-    );
```
danrosenthal: Simplified this so that we would not need coverage for the null case.

src/components/Menu/components/MenuGroup/MenuGroup.tsx
```diff
@@ -55,8 +55,8 @@ export default class MenuGroup extends React.Component<Props, never> {
   }
 
   private handleClose = () => {
-    const {title, onClose} = this.props;
-    onClose(title);
+    const {onClose} = this.props;
+    onClose();
```
danrosenthal: This was no longer necessary as `onClose` wouldn't be called with a `title` that was not active.

src/components/Menu/tests/Menu.test.tsx
```diff
@@ -31,6 +31,12 @@ describe('<Menu />', () => {
       expect(wrapper.find(MenuAction)).toHaveLength(mockActions.length);
     });
 
+    it('does not render <MenuAction /> when there are no actions', () => {
+      const wrapper = mountWithAppProvider(<Menu {...mockProps} />);
+
+      expect(wrapper.find(MenuAction)).toHaveLength(0);
+    });
+
```
danrosenthal: This and the change below was to handle these partially covered paths

|:computer:|
|-|
|Screen Shot 2019-06-12 at 10 22 59 AM|

src/components/Menu/tests/Menu.test.tsx
```diff
@@ -66,6 +72,12 @@ describe('<Menu />', () => {
       expect(wrapper.find(MenuGroup)).toHaveLength(mockGroups.length);
     });
 
+    it('does not render <MenuGroup /> when there are no actions', () => {
+      const wrapper = mountWithAppProvider(<Menu {...mockProps} />);
+
+      expect(wrapper.find(MenuGroup)).toHaveLength(0);
+    });
+
```
danrosenthal: See above comment

src/components/Page/components/Header/Header.tsx
```diff
@@ -140,8 +140,7 @@ class Header extends React.PureComponent<ComposedProps, State> {
     const primaryActionMarkup = primaryAction ? (
       <div className={styles.PrimaryActionWrapper}>
         {buttonsFrom(primaryAction, {
-          primary:
-            primaryAction.primary === undefined ? true : primaryAction.primary,
```
danrosenthal: Simplified this as we always want `primary` to be `true` for a primary action.

danrosenthal:
Achieving 90% code coverage on this is going to be tough, but there are a couple places you could get more coverage easily.

- https://codecov.io/gh/Shopify/polaris-react/compare/77c17414fd9441a42277f965f1c536dc59a446c4...923c73f3e2cf76d5da5ee1242535cc4245dcfffb/diff#D1-61
  - pass an action without content in an existing test to get this covered
- https://codecov.io/gh/Shopify/polaris-react/compare/77c17414fd9441a42277f965f1c536dc59a446c4...923c73f3e2cf76d5da5ee1242535cc4245dcfffb/diff#D1-72
  - pass a group without a title in an existing test to get this covered

Looks like the rest is hard to test (resizing) or untestable (generated classes for variants). Once the are done, I'll be comfortable merging this with whatever coverage number codecov gets us, even if it's below the 90% threshold.

src/components/ActionList/ActionList.tsx
beefchimi: related: https://github.com/Shopify/polaris-react/issues/1676

PR #1664, opened 2019-06-11
[v4] New structure

AndrewMusgrave:

As discussed we're going to restructure our utilities and context to enable greater scalability and eliminate circular references. 

Unfortunately, I ran into some circular dependencies and this PR grew larger than first intended. I haven't added a changelog entry yet. I plan to squash this PR once ✅ and use it as a base for all the other restructure PR's so we can merge it all into `v4` at once, with a single changelog entry.

This PR is has a lot of noise from
* type changes/fixes
* path changes/fixes
* fixes tests after context changes

### Brief overview
* Restructure intl/scrolllockmananger/themeprovider
  * create context
  * add hook e.g. `useIntl`
  * add tests for hooks
* Update AppProvider/withAppProvider with new contextx
* Remove intl/scrolllockmanager from `createAppProviderContext`
* Update `createPolarisContext`
* Add exports to src/index since they were removed from src/components/index
* Update new/legacy testing utils
* Add `Omit` type

### Questions

Do we really need hooks(e.g. useIntl), it may help us in the feature if we want to add onto our internationalization without adding onto our context. But it might be to much abstraction at the moment?

## 

### How to 🎩

<details>

```jsx

interface State {}

  render() {
    return (
      <Page title="Playground">
      </Page>
    );
  }
}
```

</details>

<details>
<summary>ResourceList in scrollable</summary>

```tsx
import React from 'react';
import {
  Page,
  Modal,
  ResourceList,
  Card,
  FilterType,
  TextStyle,
  Avatar,
} from '../src';

export default class ResourceListExample extends React.Component {
  state = {
    selectedItems: [],
    sortValue: 'DATE_MODIFIED_DESC',
    searchValue: '',
    appliedFilters: [
      {
        key: 'accountStatusFilter',
        value: 'Account enabled',
      },
    ],
  };

    this.setState({searchValue});
  };

  };

  handleSortChange = (sortValue) => {
    this.setState({sortValue});
  };

  };

  renderItem = (item) => {
    const {id, url, name, location, latestOrderUrl} = item;
    const shortcutActions = latestOrderUrl
      ? [{content: 'View latest order', url: latestOrderUrl}]
      : null;
    return (
      <ResourceList.Item
        id={id}
        url={url}
        media={media}
        shortcutActions={shortcutActions}
        persistActions
      >
        <h3>
        </h3>
        <div>{location}</div>
      </ResourceList.Item>
    );
  };

  render() {
    const resourceName = {
      singular: 'customer',
      plural: 'customers',
    };
    const items = [
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1457',
      },
    ];
    const promotedBulkActions = [
      {
        content: 'Edit customers',
      },
    ];
    const bulkActions = [
      {
        content: 'Add tags',
      },
      {
        content: 'Remove tags',
      },
      {
        content: 'Delete customers',
      },
    ];
    const filters = [
      {
        key: 'orderCountFilter',
        label: 'Number of orders',
        operatorText: 'is greater than',
        type: FilterType.TextField,
      },
      {
        key: 'accountStatusFilter',
        label: 'Account status',
        operatorText: 'is',
        type: FilterType.Select,
        options: ['Enabled', 'Invited', 'Not invited', 'Declined'],
      },
    ];
    const filterControl = (
      <ResourceList.FilterControl
        filters={filters}
        additionalAction={{
          content: 'Save',
        }}
      />
    );
    return (
      <Modal
        open
        onClose={() => {}}
        primaryAction={{
          content: 'Add Instagram',
          onAction: () => {},
        }}
        secondaryActions={[
          {
            content: 'Learn more',
            onAction: () => {},
          },
        ]}
      >
        <Card>
          <ResourceList
            resourceName={resourceName}
            items={items}
            renderItem={this.renderItem}
            bulkActions={bulkActions}
            sortValue={this.state.sortValue}
            sortOptions={[
              {label: 'Newest update', value: 'DATE_MODIFIED_DESC'},
              {label: 'Oldest update', value: 'DATE_MODIFIED_ASC'},
            ]}
            onSortChange={(selected) => {
              this.setState({sortValue: selected});
              console.log(`Sort option changed to ${selected}.`);
            }}
            filterControl={filterControl}
          />
        </Card>
      </Modal>
    );
  }
}

```
</details>
### 🎩 checklist


src/components/AppProvider/AppProvider.tsx
```diff
@@ -26,14 +27,14 @@ export default class AppProvider extends React.Component<Props, State> {
...
-    const {theme, children, ...rest} = this.props;
+    const {theme, children, i18n, ...rest} = this.props;
 
     this.state = {
       context: createAppProviderContext({
         ...rest,
         stickyManager: this.stickyManager,
-        scrollLockManager: this.scrollLockManager,
       }),
+      intl: new Intl(i18n),
```
AndrewMusgrave: The start of breaking up our main context 🎉

src/components/AppProvider/AppProvider.tsx
```diff
@@ -65,25 +66,32 @@ export default class AppProvider extends React.Component<Props, State> {
...
+      intl,
+    } = this.state;
 
     return (
-      <AppProviderContext.Provider value={context}>
-        <ThemeProvider theme={theme}>
-          {React.Children.only(children)}
-        </ThemeProvider>
+      <AppProviderContext.Provider value={appProviderContext}>
+        <IntlContext.Provider value={intl}>
```
AndrewMusgrave: Our two new providers

src/components/AppProvider/context.tsx
```diff
@@ -1,13 +1,12 @@
 import React from 'react';
 import {ClientApplication} from '@shopify/app-bridge';
+import Link from '../../utilities/unstyled-link';
 import createPolarisContext from './utilities/createPolarisContext';
-import {Intl, Link, StickyManager, ScrollLockManager} from './utilities';
+import {StickyManager} from './utilities';
 
 export interface AppProviderContextType {
-  intl: Intl;
```
AndrewMusgrave: These have been burned from our context type

src/components/AppProvider/utilities/createAppProviderContext/createAppProviderContext.ts
```diff
@@ -6,28 +6,22 @@ import createApp, {
...
-    linkComponent,
-    apiKey,
-    shopOrigin,
-    forceRedirect,
-    stickyManager,
-    scrollLockManager,
-  }: CreateAppProviderContext = {i18n: {}},
-): AppProviderContextType {
-  const intl = new Intl(i18n);
+export default function createAppProviderContext({
```
AndrewMusgrave: This function will be 🔥 by the time our new structure is finished

src/components/AppProvider/utilities/createPolarisContext/createPolarisContext.ts
```diff
@@ -2,45 +2,42 @@ import {PolarisContext} from '../../../types';
...
 
   const appProvider = appProviderContext
     ? createAppProviderContext(appProviderContext)
     : createAppProviderContext();
-  const theme = themeContext
-    ? createThemeContext(themeContext)
+  const theme = themeProviderContext
+    ? createThemeContext(themeProviderContext)
     : createThemeContext();
+  const intl = new Intl(translations);
```
AndrewMusgrave: New additions since we don't have them in `createPolarisContext` anymore

src/components/ScrollLock/ScrollLock.tsx
```diff
@@ -8,11 +8,13 @@ type CombinedProps = Props & WithAppProviderProps;
 class ScrollLock extends React.Component<CombinedProps, never> {
   componentDidMount() {
     const {scrollLockManager} = this.props.polaris;
+    if (!scrollLockManager) return;
```
AndrewMusgrave: There was a type change so scrollLockManager can default to `null`

src/components/UnstyledLink/UnstyledLink.tsx
```diff
@@ -1,20 +1,9 @@
...
-  external?: boolean;
-  /** Tells the browser to download the url instead of opening it. Provides a hint for the downloaded filename if it is a string value. */
-  download?: string | boolean;
-  /**	Content to display inside the link */
-  children?: React.ReactNode;
-  [key: string]: any;
-}
-
-export type LinkLikeComponent = React.ComponentType<Props> | undefined;
+interface Props extends UnstyledLinkProps {}
```
AndrewMusgrave: Needed for styleguide parsing

src/components/types.ts
```diff
@@ -1,7 +1,11 @@
+import Intl from '../utilities/intl';
+import {ThemeProviderContextType} from '../utilities/theme-provider';
+import ScrollLockManager from '../utilities/scroll-lock-manager';
 import {AppProviderContextType} from './AppProvider';
-import {ThemeProviderContextType} from './ThemeProvider';
 
 export interface PolarisContext extends AppProviderContextType {
+  intl: Intl;
```
AndrewMusgrave: These don't exist on AppProviderContextType anymore

src/test-utilities/legacy.tsx
```diff
@@ -1,6 +1,7 @@
 import {ReactWrapper, CommonWrapper, mount} from 'enzyme';
```
AndrewMusgrave: This adds context to our legacy tests now that it's not in polaris context

src/test-utilities/react-testing.tsx
```diff
@@ -7,24 +7,36 @@ import {
 } from '../components/AppProvider';
```
AndrewMusgrave: This adds context to our new react-testing mounting util

src/types.ts
```diff
@@ -258,3 +258,5 @@ export type DeepPartial<T> = {
       ? ReadonlyArray<DeepPartial<U>>
       : DeepPartial<T[P]>
 };
+
+export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
```
AndrewMusgrave: Omit types, e.g.

```ts
interface A {
    a: number;
    b: number;
}

type B = Omit<A, 'b'>

// type B = {
//    a: number;
//}
```

src/utilities/intl/tests/use-intl.test.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
```
AndrewMusgrave: Simple test making sure our hook returns something. This test possibly may not be needed since hooks get tested with their implementation

src/utilities/scroll-lock-manager/tests/scroll-lock-manager.test.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
```
AndrewMusgrave: ^

src/utilities/theme-provider/tests/use-theme.test.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import isEqual from 'lodash/isEqual';
+import {mountWithContext} from 'test-utilities';
+
+import useTheme from '../use-theme';
+
+describe('useTheme', () => {
+  it('returns context', () => {
```
AndrewMusgrave: ^

src/components/AppProvider/utilities/createPolarisContext/createPolarisContext.ts
```diff
@@ -2,45 +2,42 @@ import {PolarisContext} from '../../../types';
...
   CreateAppProviderContext,
 } from '../createAppProviderContext';
+import Intl, {TranslationDictionary} from '../../../../utilities/intl';
+import ScrollLockManager from '../../../../utilities/scroll-lock-manager';
 
 export interface CreatePolarisContext extends AppProviderProps {
   stickyManager?: StickyManager;
 }
 
-export function createPolarisContext(): PolarisContext;
```
AndrewMusgrave: This type was originally overloaded to prevent a breaking change, we should add a deprecation message

src/index.ts
```diff
@@ -3,3 +3,9 @@ import './configure';
 
 export * from './types';
 export * from './components';
+
+export {createThemeContext} from './utilities/theme-provider';
```
AndrewMusgrave: Exports to keep these in our bundle

src/types.ts
BPScott: I believe TS 3.5 added a built in Omit type, so once you merge master into v4 I think this will complain about conflicts.

We still want to support consumers on earlier versions so I think we want to keep this around but under a different name for the moment

AndrewMusgrave: > I believe TS 3.5 added a built in Omit type

🎉 

> We still want to support consumers on earlier versions so I think we want to keep this around but under a different name for the moment

👍

BPScott:
Next steps from what I can tell (not telling you what to do as you know this already, just writing stuff down so I can get it right in my own head):

- Extract link out of AppProvider and create a hook for the unstyled Link component and use it in UnstyledLink
Extract appBridge out of AppProvider, and update the use-app-bridge hook to use the new format, and update cases where we use usePolaris just for app-bridge to use this new format
- Extract stickymanager from AppProvider

src/components/UnstyledLink/UnstyledLink.tsx
AndrewMusgrave: Should we create an issue in polaris-styleguide and link to that? This is something we'll want to fix eventually and an issue can gather context over time as it progresses

src/index.ts
AndrewMusgrave: I mostly added this because we already exported it. I'm a little iffy on removing our exports from `src/components`. We export everything from `src/components` which I think is coming back to bite us. I'm the future we should definitely not export every single component we have, and selectively export our public API. Technically IMO this is undocumented, and not part of our public API so we can safely remove it, but we do export it. So it could also be seen as a breaking change. What are you thoughts?

To answer your question theme context can be used "independently" from polaris-context but ultimately this is nothing `createPolarisContext` won't be able to handle and could be 🔥 from our exports

src/utilities/unstyled-link/link.ts
```diff
@@ -1,4 +1,4 @@
-import {LinkLikeComponent} from '../../../UnstyledLink';
```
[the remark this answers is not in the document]
AndrewMusgrave: I'm ok combining them into a single file. We'll never have a  `useLink` as it stands since it's a utility class

src/index.ts
```diff
@@ -3,3 +3,9 @@ import './configure';
...
+export {
```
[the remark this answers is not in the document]
AndrewMusgrave: As I mentioned above I think we should move to a position where we selectively export our components since we have a lot of stuff being exported that probably shouldn't be

AndrewMusgrave:
> The constant having to define a name for default imports feels really annoying and prone to mistakes. I think we should stop that and just always use named exports. I've pushed a commit to shuffle us over to that format. https://shopify.slack.com/archives/CCNRSKLTH/p1558462166429200 is a slack convo that suggests that others have came to a similar conclusion.

I'm a huge support for this, so if we're making the change 👍

src/index.ts
BPScott: This is v4, we're free to make breaking changes :)

Looking at https://github.com/search?p=3&q=org%3AShopify+createThemeContext&type=Code It seems like there's a few usages on createThemeContext() by consuming projects in testing setups (next to AppProviderContext) but they can all be replaced with `createPolarisContext()`. so that feels like the api we should be promoted going forward.

I think remove this and tell people using this that they should use `createPolarisContext()` instead.

src/utilities/unstyled-link/link.ts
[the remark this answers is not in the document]
BPScott: In that case how would we reference the link in UnstyledLink?

Currently it uses `usePolaris().link`, but my interpretation of our direction is that we want to move away from one mega hook that exposes all polaris stuff and instead we have multiple hooks that  we use individually. useTheme, useIntl, useScrollLockManager  all of which currently exist and have their own contexts. Then TODO: useAppBridge (which exists but uses the AppProvider context for now and should be refactored to have its own)

BPScott:
> I'm a huge supporter for [named exports], so if we're making the change 👍

Let's say for new utilities we're going to use named exports. Existing components can keep doing export defaults for the particular component in `components/Button/Button.tsx` and I'll write up a fuller proposal for stopping doing that in general.

src/components/AppProvider/AppProvider.tsx
```diff
@@ -65,25 +67,32 @@ export default class AppProvider extends React.Component<Props, State> {
...
+          <ScrollLockManagerContext.Provider value={this.scrollLockManager}>
+            <ThemeProvider theme={theme}>
```
danrosenthal: Should we also be updating `ThemeProvider` to follow the same syntax? Something like `Theme.Provider`? Or do we want to preserve the current syntax to maintain backwards compatibility?

src/index.ts
danrosenthal: > This is v4, we're free to make breaking changes :)

Anything we can do to reduce the surface area of our API makes sense to do here. If consumers don't _really_ need it, don't export it.

src/types.ts
danrosenthal: Omit is pretty trivial to implement on your own. I don't feel this is necessary to export.

src/components/AppProvider/AppProvider.tsx
BPScott: This is an interesting one. In all the other cases here AppProviderContext/IntlContext/ScrollLockManagerContext are react contexts, while the ThemeProvider is a component of our creation, which acts pretty similar to a react context provider but we add that extra div to add the theme custom properties. My gut feel is that we don't want to make this mimic the API of a context because it's not a context.

Moving the ThemeProvider component into utilities/theme feels strange as we're putting a component into utilities but it probably serves our separation of concerns better than keeping the ThemeProvider as a "private" component in components.

src/components/Avatar/tests/Avatar.test.tsx
```diff
@@ -38,10 +38,15 @@ describe('<Avatar />', () => {
       }));
 
       const Avatar = require('../Avatar').default;
+      const AppProvider = require('../../AppProvider').default;
       const Image = require('components').Image;
 
       const src = 'image/path/';
-      const avatar = mountWithAppProvider(<Avatar source={src} />);
+      const avatar = mountWithAppProvider(
```
BPScott: Do we still need this wrapping?

src/components/ThemeProvider/tests/ThemeProvider.test.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import {mountWithAppProvider} from 'test-utilities/legacy';
 import ThemeProvider from '../ThemeProvider';
-import ThemeProviderContext from '../context';
+import {ThemeProviderContext} from '../../../utilities/theme/context';
```
BPScott:
```suggestion
import {ThemeProviderContext} from '../../../utilities/theme';
```

src/test-utilities/react-testing.tsx
```diff
@@ -1,30 +1,52 @@
...
-type Context = Providers;
-interface Props extends Providers {
+interface SimpleProviders {
+  intl: TranslationDictionary | TranslationDictionary[];
+  scrollLockManager: ScrollLockManager;
+  stickyManager: StickyManager;
+  appBridge: AppBridgeOptions;
+  link: LinkLikeComponent;
+}
+type ReturnedContext = ComplexProviders &
```
BPScott: Can we rework this so we build up form multiple types instead of using Discard?

src/utilities/link/context.tsx
```diff
@@ -0,0 +1,4 @@
+import {createContext} from 'react';
+import {Link} from './link';
+
+export const LinkContext = createContext<Link>(new Link());
```
BPScott: You don't need the generic in the <> because TS is already smart enough to work that out based on your argument type

src/utilities/app-bridge/context.tsx
```diff
@@ -0,0 +1,6 @@
+import {createContext} from 'react';
+import {ClientApplication} from '@shopify/app-bridge';
+
+export const AppBridgeContext = createContext<
+  ClientApplication<{}> | undefined
+>(undefined);
```
BPScott: We set the default for AppBridge to be undefined, but use null for sticky-manager and scroll-lock-manager. Are we able to use a consistent type for both of these optional fields?

src/utilities/app-bridge/tests/hooks.tsx
```diff
@@ -1,11 +1,11 @@
 import React from 'react';
```
BPScott: This file should be renamed to hooks.tests.tsx

src/components/Avatar/tests/Avatar.test.tsx
AndrewMusgrave: Unfortunately yes

src/test-utilities/react-testing.tsx
AndrewMusgrave: I can, it'll make it a little more verbose though

src/utilities/app-bridge/context.tsx
AndrewMusgrave: I'll take a look, I left it as is because we're removing it in `v5` but it might be a simple change.

AndrewMusgrave:
@BPScott This should be ready now 💯

src/utilities/app-bridge/context.tsx
BPScott: In wonder if instead we could say createContext for stickymanager and scrollockmanager should both always return a default value of an instance of their respective classes, that way we wouldn't have to deal with "what if the sticky manager was null" logic in components that consume the context provider.

But that can be done in a follow up PR

PR #1665, opened 2019-06-11
Adds support for function component examples and React hooks

kvendrik:
See https://github.com/Shopify/polaris-react/issues/1662

- Adds support for function component examples.
- Adds support for React hooks to the examples.
- Updated the first 5 TextField examples to use hooks instead of class syntax.

To be able to fully support this the styleguide editor will also need to support function syntax and hooks. I have a PR up for that here: https://github.com/Shopify/polaris-styleguide/pull/2818.

### How to 🎩
1. Run `yarn dev`.
1. In StoryBook navigate to the first 5 `TextField` examples and confirm they still work.

### DIFF
--- a/src/components/TextField/README.md
+++ b/src/components/TextField/README.md
@@ -173,24 +173,15 @@ You didn’t enter a store name.
 Use to allow merchants to provide text input when the expected input is short. For longer input, use the auto grow or multiline options.
 
 ```jsx
-class TextFieldExample extends React.Component {
-  state = {
-    value: 'Jaded Pixel',
-  };
-
-  handleChange = (value) => {
-    this.setState({value});
-  };
-
-  render() {
-    return (
-      <TextField
-        label="Store name"
-        value={this.state.value}
-        onChange={this.handleChange}
-      />
-    );
-  }
+function TextFieldExample() {
+  const [value, setValue] = useState('Jaded Pixel');
+  return (
+    <TextField
+      label="Store name"
+      value={value}
+      onChange={(newValue) => setValue(newValue)}
+    />
+  );
 }
 ```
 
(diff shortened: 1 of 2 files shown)


kvendrik:
@BPScott any idea why Percy could be throwing this error? Locally I don't seem to be able to reproduce.

Screen Shot 2019-06-11 at 6 30 26 PM

BPScott:
Aww yess this is exciting!

Master still only targets react 16.4 and I'm not quite sure how that interacts with the storybook demo panel. Can you rebase this atop of the version-4.0.0 branch which is certainly using a hooks-ready version of react to see if that helps.

Once you've got this sorted would you be able to scribble up a sister PR for polaris-styleguide too?

src/components/TextField/README.md
```diff
@@ -211,25 +202,16 @@ class TextFieldExample extends React.Component {
...
-    );
-  }
+function NumberFieldExample() {
+  const [value, setValue] = useState('1');
+  return (
+    <TextField
+      label="Quantity"
+      type="number"
+      value={value}
+      onChange={(newValue) => setValue(newValue)}
```
BPScott: Can this (and all the below onChange usages) be `onChange={setValue}`?

kvendrik:
@bgraham11 will do! And my thoughts exactly haha, have a PR up [here](https://github.com/Shopify/polaris-styleguide/pull/2818).

src/components/TextField/README.md
kvendrik: That would mean react receives all arguments which unfortunately triggers an error 😞

kvendrik:
@BPScott this is good to go for review 🙂

PR #1672, opened 2019-06-12
[v4][TextField] Fix textfield examples

AndrewMusgrave:

These examples were introduced in https://github.com/Shopify/polaris-react/pull/1665

* use usecallback

### Notes

* skipping a changelog entry since one was never added noting the conversion

### DIFF
--- a/.storybook/polaris-readme-loader.js
+++ b/.storybook/polaris-readme-loader.js
@@ -2,6 +2,9 @@
 const chalk = require('chalk');
 const grayMatter = require('gray-matter');
 const MdParser = require('./md-parser');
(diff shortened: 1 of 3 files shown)


kvendrik:
@AndrewMusgrave what is the thought behind wrapping these in `useCallback`?

AndrewMusgrave:
Without it, a new function is created every render causing shallow prop comparison (memo/PureComponent) to fail, causing unnecessary renders.

kvendrik:
@AndrewMusgrave oh yes sorry you're totally right in this case! Good catch!

PR #1674, opened 2019-06-12
[v4] Follow strict naming convention and alphabetize component exports

AndrewMusgrave:

Clean up 🧼

* Alphabetize `src/components/index.ts`
* Use kebab case for our utilities


AndrewMusgrave:
@BPScott 🙇 before I could even request a review 😄

PR #1695, opened 2019-06-17
[v4] Update no unsafe eslint rule

AndrewMusgrave:

We removed the last deprecated lifecycle method 🎉 I was going to add strict mode to storybook but it looks like we use ReactDOM.findDOMNode which is also a nono for strict mode, and will need to be handled but should be an easy fix.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -33,6 +33,7 @@
     "react/button-has-type": "off",
     "react/no-array-index-key": "off",
     "react/jsx-fragments": ["error", "element"],
+    "react/no-unsafe": ["error", {"checkAliases": true}],
     "shopify/jsx-no-complex-expressions": "off",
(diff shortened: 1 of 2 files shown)


BPScott:
I snuck this into eslint-plugin-shopify the other day https://github.com/Shopify/eslint-plugin-shopify/pull/302. Matt's already cut a v29.0.0 with that change in it, and is in the process of [putting it into sewing-kit](https://github.com/Shopify/sewing-kit/pull/1285).

I'd be tempted to wait ill new version of SK is released with that in, and then we get this rule as part of a sewing-kit update.

AndrewMusgrave:
Since you approved this pr, I'll merge this in for now. But create an issue to check for a `sewing-kit` upgrade before we release `v4` to 🔥 this change

PR #1696, opened 2019-06-17
[v4] Removed all usage of findDOMNode

AndrewMusgrave:

Last change before we can enable strict mode 🎉 

```
Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of StyledComponent which is inside StrictMode. Instead, add a ref directly to the element you want to reference.
```

Luckily none of our uses require findDOMNode to find a child element without any context so we can replace them with refs.

### DIFF
--- a/src/components/Focus/tests/Focus.test.tsx
+++ b/src/components/Focus/tests/Focus.test.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, {useRef, useState, useEffect} from 'react';
 import {mountWithAppProvider} from 'test-utilities/legacy';
-import Focus from '../Focus';
+import Focus, {Props} from '../Focus';
+import {Discard} from '../../../types';
 
 describe('<Focus />', () => {
   let requestAnimationFrameSpy: jest.SpyInstance;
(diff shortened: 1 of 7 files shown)


src/components/ResourceList/components/BulkActions/components/BulkActionButton/BulkActionButton.tsx
```diff
@@ -20,14 +19,13 @@ export default class BulkActionButton extends React.PureComponent<
...
   componentDidMount() {
     const {handleMeasurement} = this.props;
     if (handleMeasurement) {
-      const bulkActionButtonNode = findDOMNode(this);
-      const width =
-        (bulkActionButtonNode instanceof Element &&
-          bulkActionButtonNode.getBoundingClientRect().width) ||
-        0;
+      const width = (this.bulkActionButton
+        .current as HTMLButtonElement).getBoundingClientRect().width;
```
AndrewMusgrave: This cast is to increase codecov. We don't optionally render the button so this ref will always exist in componentDidMount lifecycle

src/components/Focus/tests/Focus.test.tsx
```diff
@@ -51,14 +44,27 @@ describe('<Focus />', () => {
...
+        <input />
+      </FocusTestWrapper>,
     );
 
     const input = focus.find('input').getDOMNode();
     expect(input).not.toBe(document.activeElement);
   });
 });
+
+function FocusTestWrapper({children, ...props}: Discard<Props, 'root'>) {
```
AndrewMusgrave: Wrapper component that passes a root to `Focus`

PR #1699, opened 2019-06-18
[Banner] Support disabling screen reader announcements when changing content.

emarchak:

We're using the `Banner` component multiple times on a page to communicate information in a form. When switching context (e.g. changing the form data but keeping the form), we need to update the banners. 

When doing so, the `aria-live` state of the banners is causing them to be read out one after another. I'd like the option of turning `aria-live` off until we need it.

Feel free to slack `@emarchak` for more context on this.

Adding a new prop to the banner component, `stopAnnouncements` that allows devs to set `aria-live` to `off`.

### How to 🎩

1. Set up the playground with the following code. You'll get a warning banner that has a primary action that updates text.
2. Boot up your screen reader, and click the `Update banner` button. 
    - You should hear the new content being read out.
3. Uncomment `// stopAnnouncements` and allow the component to reload.
4. Boot up your screen reader, and click the `Update banner` button. 
   - You should _not_ hear the new content being read out.

<details>

```jsx
import React from 'react';
import {Banner} from '../src';

interface State {
  bannerTitle: string;
  bannerContent: string;
}

  constructor(props) {
    super(props);
    this.state = {
      bannerTitle: 'This is this original banner title.',
      bannerContent: 'This is the original banner content.',
    };
  }

  render() {
    const {bannerTitle, bannerContent} = this.state;
    return (
      <Banner
        title={bannerTitle}
        // stopAnnouncements
        status="warning"
        action={{
          content: 'Update banner',
          onAction: () => {
            this.setState({
              bannerTitle: 'This is a new banner title',
              bannerContent: 'This is some new banner content',
            });
          },
        }}
      >
        {bannerContent}
      </Banner>
    );
  }
}

```

</details>

### 🎩 checklist

* [x] Updated the component's `README.md` with documentation changes.

cc @Shopify/accessibility

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -16,6 +16,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Added the `stopAnnouncements` prop to `Banner`, which disables screen reader announcements when content changes ([#1699](https://github.com/Shopify/polaris-react/pull/1699))
 - Add `selectable` prop to `ResourceList` component ([#1614](https://github.com/Shopify/polaris-react/pull/1614))
 - Allow `Link` and `Button` interactions when rendered as `prefix/suffix` within `<TextField />` ([#1394](https://github.com/Shopify/polaris-react/pull/1394))
 - `ActionList` can now pass a unique `accessibilityLabel` to each `Item` ([#1653](https://github.com/Shopify/polaris-react/pull/1653))
--- a/src/components/Banner/README.md
+++ b/src/components/Banner/README.md
@@ -611,6 +611,7 @@ Banners provide context and assist workflows for merchants with disabilities.
 
 - Critical and warning banners have a `role=”alert”` and are announced by assistive technologies when they appear.
 - All other banners have a `role=”status”` and are read after any critical announcements.
+- All banners have an `aria-live` attribute and are announced by assistive technologies when their content is updated. These announcements can be disabled by using the prop `stopAnnouncements`.
 - Banners use `aria-describedby` to describe their purpose to assistive technologies when they’re announced or receive focus. If a banner has a `title`, then the title content is used for the `aria-describedby`. If the banner doesn’t have a `title`, then all of the banner content is used for the `aria-describedby`.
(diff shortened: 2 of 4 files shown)


UNRELEASED.md
```diff
@@ -18,6 +18,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Support disabling screen reader announcements when changing content in `Banner`. ([#1699](https://github.com/Shopify/polaris-react/pull/1699))
```
```suggestion
- Added the `stopAnnouncements` prop to `Banner`, which disables screen reader announcements when content changes ([#1699](https://github.com/Shopify/polaris-react/pull/1699))
```

kaelig:
Code looks good to me!

How do we feel about the naming of the `stopAnnouncements` prop?

cc @danrosenthal @chloerice, can you chime in? (you've worked on this component before and have context on props naming across the library 👓)

src/components/Banner/README.md
```diff
@@ -611,6 +611,7 @@ Banners provide context and assist workflows for merchants with disabilities.
 
 - Critical and warning banners have a `role=”alert”` and are announced by assistive technologies when they appear.
 - All other banners have a `role=”status”` and are read after any critical announcements.
+- All banners have a `aria-live=”polite”` and are announced by assistive technologies when their content is updated. These announcements can be disabled by using the prop `stopAnnouncements`.
```
dpersing: Tiny suggested tweak:

```suggestion
- All banners have an `aria-live="polite"` attribute and are announced by assistive technologies when their content is updated. These announcements can be disabled by using the prop `stopAnnouncements`.
```

danrosenthal:
Generally speaking, I'm opposed to props which prevent default behavior of a component. This is basically the equivalent of passing `doDefaultThing: false`. My wish would be that we would have offered a more flexible prop on this component that might enable announcing, instead of disabling it from the outset. But, that would be a breaking change, so I think this solution is fine. I don't have strong feelings on the naming used here.

emarchak:
Not sure why Codecov hasn't reported this yet, it's passed when you look at the PR over there. 😕 

https://codecov.io/gh/Shopify/polaris-react/pull/1699

emarchak:
CodeCov isn't reporting correctly, so I rolled this into a new PR at #1719. That seemed to fix the problem.

PR #1702, opened 2019-06-18
[TextField] Improve TextField so that character count is only announced on focus.

emarchak:

We're using the TextField component inside a very long form. When switching context (e.g. changing the form data but keeping the form), we need to update the values of the form..

When doing so, the `aria-live` state of the character counters is causing them to be read out one after another without any context. I'm setting the aria-live state to off unless the TextField has focus.

Feel free to slack `@emarchak` for more context on this. Related to the same problem as #1699.

I'm piggy-backing off of the existing `focus` state in the TextField component to toggle between `off` and `polite` to the character count.

### How to 🎩

1. Set up the playground with the following code. You'll get an input field with a character count and a button that says "Update textfield."
2. Boot up your screen reader, and edit the text in the textfield
   - You should hear the screen reader read out the character count as your typing.
3. Tab out of the text field and click the "Update textfield." button.
   - You should _not_ hear the screen reader read out the new character count.

<details>

```jsx
import React from 'react';
import {TextField, Button} from '../src';

interface State {
  value: string;
}

  constructor(props) {
    super(props);
    this.state = {
      value: 'This is this original value.',
    };
  }

  handleTextChange = (value) => {
    this.setState({value});
  };

  updateTextField = () => {
    this.setState({value: 'This is a new value'});
  };

  render() {
    const {value} = this.state;

    return (
      <React.Fragment>
        <TextField
          label="Input Value name"
          value={value}
          showCharacterCount
          onChange={this.handleTextChange}
        />

        <Button onClick={this.updateTextField}>Update textfield</Button>
      </React.Fragment>
    );
  }
}
```

</details>

### 🎩 checklist

* ~Updated the component's `README.md` with documentation changes~ (Nothing to update)

cc @Shopify/accessibility

### DIFF
--- a/src/components/TextField/tests/TextField.test.tsx
+++ b/src/components/TextField/tests/TextField.test.tsx
@@ -465,6 +465,28 @@ describe('<TextField />', () => {
 
       expect(characterCount.text()).toBe('4/10');
     });
+
+    it('announces updated character count only when input field is in focus', () => {
+      const textField = mountWithAppProvider(
+        <TextField
+          value="test"
+          showCharacterCount
+          label="TextField"
+          id="MyField"
+          onChange={noop}
+        />,
+      );
+
+      expect(
+        textField.find('#MyFieldCharacterCounter').prop<string>('aria-live'),
+      ).toBe('off');
+
+      textField.find('input').simulate('focus');
+
+      expect(
+        textField.find('#MyFieldCharacterCounter').prop<string>('aria-live'),
+      ).toBe('polite');
+    });
   });
 
   describe('type', () => {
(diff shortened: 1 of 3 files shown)


src/components/TextField/tests/TextField.test.tsx
```diff
@@ -379,6 +379,33 @@ describe('<TextField />', () => {
...
+          showCharacterCount
+          label="TextField"
+          id="MyField"
+          onChange={noop}
+        />,
+      );
+
+      expect(quietTextField.find('#MyFieldCharacterCounter').prop<string>('aria-live')).toBe('off');
+
+      const politeTextField = mountWithAppProvider(
```
emarchak: Not sure if there's a more elegant way to force this component to re-render with updated props. The tests didn't update aria-live if I set the prop to focused or gave focus manually.

AndrewMusgrave: We could split these into separate tests. For reference 
```tsx
quietTextField.setProps({...})
```
would cause a re-render with the new props 😄

emarchak: I dug a bit further. We're not updating `this.state.focus` when props update, only during the constructor. Using ` textField.find('input').simulate('focus');` fixed this, and better shows how this feature is supposed to work, too! 🎉

emarchak:
Not sure why Codecov hasn't reported this yet, it's passed when you look at the PR over there. 😕https://codecov.io/gh/Shopify/polaris-react/pull/1702

emarchak:
Opening a new PR to kickstart CodeCov, which appeared to get it responding. #1720.

PR #1703, opened 2019-06-18
[UnstyledLink] Remove unused data-polaris-unstyled

### LINKED ISSUE #1190: Remove unused data-polaris-unstyled prop from all components
As it's not being used anymore, context [here (internal link)](https://shopify.slack.com/archives/C8H54T86Q/p1552504392143900?thread_ts=1552504258.143800&cid=C8H54T86Q).

cdmwebs:

Fixes #1190. I think.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -41,6 +41,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Code quality
 
 - Updated `PositionedOverlay` to no longer use `componentWillReceiveProps`([#1621](https://github.com/Shopify/polaris-react/pull/1621))
+- Remove unused `data-polaris-unstyled` from `UnstyledLink` ([#1190](https://github.com/Shopify/polaris-react/issues/1190))
 
 ### Deprecations
 
--- a/a11y_shitlist.json
+++ b/a11y_shitlist.json
@@ -2,7 +2,7 @@
   "all-components-app-provider--with-i18n-object": [
     {
       "code": "WCAG2AA.Principle4.Guideline4_1.4_1_2.H91.A.NoContent",
-      "context": "<a aria-describedby=\"341\" class=\"Item-Link_2-raD\" tabindex=\"0\" id=\"ResourceListItemOverlay1\" href=\"customers/341\" data-polaris-unstyled=\"true\"></a>",
+      "context": "<a aria-describedby=\"341\" class=\"Item-Link_2-raD\" tabindex=\"0\" id=\"ResourceListItemOverlay1\" href=\"customers/341\"></a>",
       "message": "Anchor element found with a valid href attribute, but no link content has been supplied.",
       "type": "error",
       "typeCode": 1,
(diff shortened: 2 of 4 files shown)


BPScott:
pinging @kvendrik for review as he raised the original issue, and the original slack context thread is now dead (damn you 3 month retention)

kvendrik:
I'd like to remove it but we do have some internal repositories that [rely on this attribute](https://github.com/search?q=org%3AShopify+data-polaris-unstyled&type=Code) in their CSS selectors. To be able to remove it I think we need to update those selectors first, we'd have to create issues in those repositories so the teams can update them. What do you think @tmlayton?

tmlayton:
@kvendrik can you provide a summary of the context since the Slack link is dead? My initial thought is I don’t think we can remove this yet, but if we do it is a breaking change since the selector is part of the public API and would go out with v4 or v5. It would also be nice to provide deprecation notice.

kvendrik:
Yep so the gist is that I still don't really know what `data-polaris-unstyled` historically was used for but we do seem to rely on it [for various reasons](https://github.com/search?q=org%3AShopify+data-polaris-unstyled&type=Code). I think removing it would be a good idea considering that it doesn't seem to have a single, well-defined, use-case. Say a lot of codebases use this for interop fixes we probably just need a better solution for that.

I think the Slack link linked to a conversation between me and @lemonmade where we said we don't know what the attribute was for and that we should likely burn it but as the link expired I have no way to verify. @lemonmade just to confirm, do you recall what `data-polaris-unstyled` was originally used for?

lemonmade:
IIRC the original intent was that we could do a generic `a:not([data-polaris-unstyled])` selector to turn all links in phrasing content to the Polaris-style blue, while avoiding having those styles also apply for the various `a` tags Polaris renders that aren't actually text links (which had `data-polaris-unstyled`.

kvendrik:
@lemonmade ah I see. Was that a fix that we did specifically for Web? What are your thoughts on removing the attribute from Polaris? Any negative implications that you see for Web?

lemonmade:
It was for Polaris more generally, because we couldn't think of a good way of allowing you to just have a collection of regular old `p`/ `a`/ `strong` etc tags, but have them still be properly formatted in the style Polaris prescribes. If I were doing it now I would probably just have a wrapping element that scopes that styling down to only `a`s within that component, not sure whether it's even a case Polaris needs to care about though.

kvendrik:
@lemonmade just to define this, would it be correct to say that:

> `data-polaris-unstyled` was a way for consumers to target Polaris components in their styling with the purpose of excluding them from certain styles.

I'm trying to clearly define the use-case so the team can decide if this is something they want to care about going forward.

lemonmade:
It was not really meant for consumers to rely on, it was mostly just a way for Polaris itself to scope "visual" link styles (the blue text color) only to "text-ey" links (those appearing in passages of text), and to scope those styles to not apply to Polaris components that were anchors under the hood (e.g., to prevent the blue text color on links in the nav).

kvendrik:
Gotcha so something like this would probably be more accurate:

> `data-polaris-unstyled` was a way for Polaris to exclude links in a non-phrasing context from their global link styles. Over time however consumers have started to rely on the attribute to target Polaris components.

Does that sound correct?

lemonmade:
Yeah I think that's accurate 👍

kvendrik:
@tmlayton looks like we've decided on the [definition above](https://github.com/Shopify/polaris-react/pull/1703#issuecomment-528060185). I'll leave it up to the team to decide if the attribute is still useful, how they feel about [consumers using it to target Polaris components](https://github.com/search?q=org%3AShopify+data-polaris-unstyled&type=Code), and what the next step should be.

tmlayton:
Hmm yeah I’d like to remove, but it is a breaking change as we are removing part of the API so want to provide a deprecation notice with a suggested alternative.

danrosenthal:
Hey Chris, any interest in continuing work on this pull request? If so, give me a ping. Based on Tim's feedback, it makes sense to target our version 5.0 branch for this change, following our guidance for deprecation and removal.

cdmwebs:
Sounds good Dan. This was a one-day PR from the Unite Hack Session. Happy to help if I can!

PR #1705, opened 2019-06-18
[Card] Add example for card with footer action + plain button

### LINKED ISSUE #562: Example for card with footer actions with plain button
# Card with footer actions with plain button

This is for a modification of the card component. There's another example which isn't reflected in the style guide but is in our UI kit and exists on Shopify. Could we please add an example to the UI kit that shows the Card with footer actions but with a secondary action and a plain button.

## Components dependencies
https://polaris-styleguide-pr-2346.herokuapp.com/components/structure/card

@tmlayton

sharoonthomas:

Fixes #562

Adds an example of the card where custom buttons are added, similar to how they are used in some parts of Shopify. The example adds a secondary button and a plain button to the footer of the card.

Screenshot 2019-06-18 15 51 48

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -31,6 +31,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Documentation
 
 - Mentioned Polaris icons in the Icon component documentation ([#1693](https://github.com/Shopify/polaris-react/pull/1693))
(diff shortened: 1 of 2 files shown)


sharoonthomas:
Signed CLA

sharoonthomas:
Thanks for the feedback @chloerice. Updated the PR with the changes you suggested.

PR #1706, opened 2019-06-18
[ResourceList] Fix shift-clicking on the checkbox in Firefox

davidcornu:

At present, <kbd>Shift</kbd>-clicking an `Item`'s `Checkbox` in a `ResourceList` with multi-select does not work in Firefox. This is due to a long-standing bug (https://bugzilla.mozilla.org/show_bug.cgi?id=559506) that prevents an `<input type="checkbox"/>` from being checked when <kbd>Shift</kbd>-clicking the `<label>`.

Although our `Checkbox` component wraps the inner `<input>` in a `<label>` https://github.com/Shopify/polaris-react/blob/cb96a7aa749018efe344f2500f7b687783285c88/src/components/Choice/Choice.tsx#L45-L46 
it side-steps this issue by listening to the `<label>`'s `click` event rather than to `change` https://github.com/Shopify/polaris-react/blob/cb96a7aa749018efe344f2500f7b687783285c88/src/components/Checkbox/Checkbox.tsx#L116 https://github.com/Shopify/polaris-react/blob/cb96a7aa749018efe344f2500f7b687783285c88/src/components/Checkbox/Checkbox.tsx#L48-L55

However, `Checkbox`'s custom `onChange` property only passes a boolean to signify whether the checkbox is checked so there's no way of figuring out whether the <kbd>Shift</kbd> key is pressed. To work around this our current implementation listens to `Checkbox`'s inner `<input>`'s `onChange` event by registering a handler on the parent to catch it when it bubbles https://github.com/Shopify/polaris-react/blob/5bf432eed98e1d157a79bccbd5c314ecc002a3cb/src/components/ResourceList/components/Item/Item.tsx#L162
but `onChange` is never triggered on Firefox due to the aforementioned bug.

To fix this I believe there are essentially two options:

1. Expand `Checkbox`'s `onChange` callback signature to either pass along the event or an object containing a `shiftKey`. This would allow `Item` to use it instead of relying on the inner `<input>`'s bubbled `onChange`.<br><br>The only way to make this a backwards-compatible change would be to include a second argument, which has the downside of making `Checkbox`'s `onChange` diverge even further from typical event handlers which take an event object as the first argument.
2. Track the state of the <kbd>Shift</kbd> key separately and rely on that combined with `Checkbox`'s `onChange` to determine whether we should check or uncheck multiple items.

This PR currently implements option (2) by introducing a `KeyWatcher` class track the state of the <kbd>Shift</kbd> key and checking the value of `.isPressed()` instead of `event.nativeEvent.shiftKey` in `handleLargerSelectionArea`.

⚠️ I'm opening a PR with this implementation to spark some discussion around what the best way of fixing this is (and choosing to not fix it at all may be a legitimate outcome). It is by no means complete and has no test coverage at the moment.

### How to 🎩

`dev s` and open up http://localhost:6006/?path=/story/all-components-resource-list--resource-list-with-multiselect. The behaviour should be identical to before with the only difference being that it now works in Firefox.

### 🎩 checklist

### DIFF
--- a/src/components/ResourceList/components/Item/Item.tsx
+++ b/src/components/ResourceList/components/Item/Item.tsx
@@ -161,19 +161,17 @@ class Item extends React.Component<CombinedProps, State> {
           testID="LargerSelectionArea"
         >
           <div onClick={stopPropagation} className={styles.CheckboxWrapper}>
-            <div>
-              <Checkbox
-                testID="Checkbox"
-                id={this.checkboxId}
-                label={label}
-                labelHidden
-                checked={selected}
-                disabled={loading}
-                onChange={() => {
-                  this.handleLargerSelectionArea();
-                }}
-              />
-            </div>
+            <Checkbox
+              testID="Checkbox"
+              id={this.checkboxId}
+              label={label}
+              labelHidden
+              checked={selected}
+              disabled={loading}
+              onChange={() => {
+                this.handleLargerSelectionArea();
+              }}
+            />
           </div>
         </div>
       );
--- a/src/components/ResourceList/components/Item/Item.tsx
+++ b/src/components/ResourceList/components/Item/Item.tsx
@@ -14,6 +14,7 @@ import ButtonGroup from '../../../ButtonGroup';
 import Checkbox from '../../../Checkbox';
 import Button, {buttonsFrom} from '../../../Button';
 import {withAppProvider, WithAppProviderProps} from '../../../AppProvider';
+import EventListener from '../../../EventListener';
 
 import {
   ResourceListContext,
@@ -94,15 +95,7 @@ class Item extends React.Component<CombinedProps, State> {
   private checkboxId = getUniqueCheckboxID();
   private overlayId = getUniqueOverlayID();
   private buttonOverlay = React.createRef<HTMLButtonElement>();
-  private shiftKeyWatcher = new KeyWatcher('Shift');
-
-  componentDidMount() {
-    this.shiftKeyWatcher.attach();
-  }
-
-  componentWillUnmount() {
-    this.shiftKeyWatcher.teardown();
-  }
+  private shiftKeyPressed = false;
 
   shouldComponentUpdate(nextProps: CombinedProps, nextState: State) {
     const {
@@ -305,6 +298,7 @@ class Item extends React.Component<CombinedProps, State> {
         testID="Item-Wrapper"
         data-href={url}
       >
+        {this.renderKeyboardEventListeners()}
         {accessibleMarkup}
         {containerMarkup}
       </div>
@@ -341,11 +335,7 @@ class Item extends React.Component<CombinedProps, State> {
 
   private handleLargerSelectionArea = (event?: React.MouseEvent<any>) => {
     if (event) stopPropagation(event);
-
-    this.handleSelection(
-      !this.state.selected,
-      this.shiftKeyWatcher.isPressed(),
-    );
+    this.handleSelection(!this.state.selected, this.shiftKeyPressed);
   };
 
   private handleSelection = (value: boolean, shiftKey: boolean) => {
@@ -410,6 +400,28 @@ class Item extends React.Component<CombinedProps, State> {
     }
   };
 
+  private renderKeyboardEventListeners(): JSX.Element {
+    return (
+      <>
+        <EventListener event="keydown" handler={this.handleShiftKey} />
+        <EventListener event="keyup" handler={this.handleShiftKey} />
+      </>
+    );
+  }
+
+  private handleShiftKey = (event: KeyboardEvent) => {
+    if (event.key !== 'Shift') return;
+
+    switch (event.type) {
+      case 'keydown':
+        this.shiftKeyPressed = true;
+        break;
+      case 'keyup':
+        this.shiftKeyPressed = false;
+        break;
+    }
+  };
+
   private handleActionsClick = () => {
     this.setState(({actionsMenuVisible}) => ({
       actionsMenuVisible: !actionsMenuVisible,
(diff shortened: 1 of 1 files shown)


davidcornu:
@AndrewMusgrave would appreciate the feedback if you have some time 🙏

src/components/ResourceList/components/Item/Item.tsx
```diff
@@ -419,6 +435,41 @@ function isSelected(id: string, selectedItems?: SelectedItems) {
   );
 }
 
+class KeyWatcher {
```
davidcornu: I believe I can replace this whole thing with `EventListener` https://github.com/Shopify/polaris-react/blob/master/src/components/EventListener/EventListener.tsx

AndrewMusgrave:
Hi @david-english, I actually just started looking into this 20 minutes ago, so this is great! Thanks for pinging me 😄 

From what I can tell it's an [issue with react](https://github.com/facebook/react/issues/6951). Fortunately, click events will work from click but also trigger with spacebar/enter IIFC.  As an alternative approach, we might be able to remove the onChange/stopPropagation from the divs wrapping checkbox and have the click event delegate to the main handle wrapper?

davidcornu:
> From what I can tell it's an [issue with react](https://github.com/facebook/react/issues/6951).

🤔 I can definitely reproduce this issue in vanilla JavaScript by simply having a `<label>` wrap a `<input type="checkbox"/>` with a `change` event listener on it. Shift-clicking the label won't trigger a change event at all. https://jsfiddle.net/oytqvps5/.

kaelig:
👮 🚨 Performance police alert 😅

Is fixing such an edge case in a single browser worth serving additional code to all browsers?

(on top of this, we don't have cross-browser automated tests, so be mindful of the potential manual tophatting costs over time)

AndrewMusgrave:
> Is fixing such an edge case in a single browser worth serving additional code to all browsers? (on top of this, we don't have cross-browser automated tests, so be mindful of the potential manual tophatting costs over time)

I wouldn't say it's an edge case, multi-select is a bulk workflow that's used across index pages in the admin and is a powerful feature. This is just a browser discrepancy/bug?

We won't have to worry about degraded performance from shipping a few extra lines of code. And firefox is a supported browser so we can't simply shake it off.

With that being said I quickly tested out my suggestion, and on first glance, it seems to work. Interaction states are a little wonky and might need to be touched up but we may be able to ship a fix while reducing lines and markup 😄

davidcornu:
> As an alternative approach, we might be able to remove the onChange/stopPropagation from the divs wrapping checkbox and have the click event delegate to the main handle wrapper?

The only thing I don't like about this is we're making `Checkbox` a leaky abstraction (although we're doing this already by listening to the bubbled `onChange` event. Both the options I outlined in the description let `Checkbox` remain responsible for when its state should change.

kaelig:
>We won't have to worry about degraded performance from shipping a few extra lines of code

Agreed, but those few extra lines of code add up over time!

>I wouldn't say it's an edge case, multi-select is a bulk workflow that's used across index pages in the admin and is a powerful feature. This is just a browser discrepancy/bug?

The experience isn't optimal, but the feature still works for Firefox users, as clicking the checkbox itself still works.

---

Back to the original issue: can you add context around the use case for a user to hit <kbd>shift</kbd> + click in the Resource List, rather than a standard left click?

davidcornu:
> but the feature still works for Firefox users, as shift+click on the checkbox itself still works.

This is incorrect. Shift-clicking works on regular checkboxes but ours isn't a regular checkbox. It does however work if you click the row.

resouce_list_shift_click
(recorded on https://polaris.shopify.com/components/lists-and-tables/resource-list)

davidcornu:
> Back to the original issue: can you add context around the use case for a user to hit shift + click in the Resource List, rather than a standard left click?

On multi-select resource lists, this lets you select all the rows between the one you last clicked and the one you shift-click.

kaelig:
My bad – I don't know why I thought it worked when clicking the Checkbox itself!

Would a simpler fix be to have the label not wrap the checkbox?

AndrewMusgrave:
> Would a possible fix be to have the label not wrap the checkbox?

We'll still have to provide some event handlers because the label is a clickable area.

> Web browsers provide the label as a larger clickable area, for example, to select or activate the control. It also ensures that assistive technology can refer to the correct label when presenting a form control. - [w3](https://www.w3.org/WAI/tutorials/forms/labels/)

> The only thing I don't like about this is we're making Checkbox a leaky abstraction

I don't think this is something we have to worry about to much since the event is still handled within the component. And event delegation is a pretty common pattern. And all react events are actually delegated to the document and not attached to nodes!

However, if we were going with option 1/2 I'm leaning towards 2 since we don't currently pass the event object back in any of our events.

cc/ @alex-page @dleroux for some more opinions

src/components/ResourceList/components/Item/Item.tsx
```diff
@@ -159,14 +168,17 @@ class Item extends React.Component<CombinedProps, State> {
...
               <Checkbox
                 testID="Checkbox"
                 id={this.checkboxId}
                 label={label}
                 labelHidden
                 checked={selected}
                 disabled={loading}
+                onChange={() => {
+                  this.handleLargerSelectionArea();
+                }}
```
dleroux: onChange={this.handleLargerSelectionArea};

src/components/ResourceList/components/Item/Item.tsx
```diff
@@ -159,14 +168,17 @@ class Item extends React.Component<CombinedProps, State> {
           testID="LargerSelectionArea"
         >
           <div onClick={stopPropagation} className={styles.CheckboxWrapper}>
-            <div onChange={this.handleLargerSelectionArea}>
+            <div>
```
dleroux: I think this div can 🔥

src/components/ResourceList/components/Item/Item.tsx
dleroux: Yeah, that would be great.

dleroux:
I think if your solution does work Andrew then I don't think I have a preference. What's in the PR is almost what is happening today. I also like that the onChange is back on the checkbox. I say fix the tests, use our listener and go with this unless you feel strongly about your approach Andrew.

I do think this needs to work in Firefox.

src/components/ResourceList/components/Item/Item.tsx
davidcornu: I believe I tried that and hit a TypeScript wall. Will sort it out.

davidcornu: Ah so the reason I did this was because `handleLargerSelectionArea` expects an `event` that it can stop propagation on but `Checkbox`'s `onChange` just passes a boolean. To make this work I made `handleLargerSelectionArea`'s `event` argument optional (https://github.com/Shopify/polaris-react/pull/1706/files#diff-425411853250df4be3bd738e2b2bab02R342) and called it with no arguments from here.

kaelig:
>We'll still have to provide some event handlers because the label is a clickable area.

How so? Clicking the label toggles its checkbox, so I would have assumed the checkbox's change event would be the only event we needed to handle?

kaelig:
>I also like that the onChange is back on the checkbox.

Same here. Feels more "native to the web".

davidcornu:
Did some cleanup in the last 2 commits. Am looking into fixing the tests.

alex-page:
@davidcornu Did you get a chance to look into fixing the tests?

dleroux:
@davidcornu let me know when this is ready to review again.

danrosenthal:
Hey @davidcornu, what's the status on this pull request? Is it still something you're able to work on?

davidcornu:
Hey @danrosenthal. This got lost in the black hole that is my GH inbox. Unfortunately I haven't had the time to pick this up. Happy to help out if someone else wants to take over.

danrosenthal:
@davidcornu looks like no one has had time to pick this up. Going to close it for now until someone else has the bandwidth to take this on.

PR #1709, opened 2019-06-18
[v4] Enable strict mode

AndrewMusgrave:

Fixes: #1709 - This will keep us on track in the future 🤓

### DIFF
--- a/src/components/Page/components/Header/tests/Header.test.tsx
+++ b/src/components/Page/components/Header/tests/Header.test.tsx
@@ -34,7 +34,7 @@ describe('<Header />', () => {
       const warnSpy = jest.spyOn(console, 'warn');
       mountWithAppProvider(<Header {...mockProps} icon="foo" />);
 
-      expect(warnSpy).toHaveBeenCalledTimes(1);
+      expect(warnSpy).toHaveBeenCalled();
       expect(warnSpy).toHaveBeenCalledWith(
         "The icon prop has been removed from Page. Upload an application icon in the Shopify Partners Dashboard 'App setup' section instead.",
       );
--- a/src/test-utilities/tests/legacy.test.tsx
+++ b/src/test-utilities/tests/legacy.test.tsx
@@ -1,0 +1,42 @@
+import React from 'react';
+import {mountWithAppProvider} from '../legacy';
+
+describe('mountWithAppProvider', () => {
+  let consoleSpy: jest.SpyInstance;
+
+  beforeEach(() => {
+    consoleSpy = jest.spyOn(console, 'error');
+    consoleSpy.mockImplementation(() => {});
+  });
+
+  afterEach(() => {
+    consoleSpy.mockRestore();
+  });
+
+  it('renders in strict mode', () => {
+    // eslint-disable-next-line react/no-unsafe
+    class Child extends React.Component {
+      componentWillReceiveProps() {}
+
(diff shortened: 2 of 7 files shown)


BPScott:
Fixes #785

src/components/Page/components/Header/tests/Header.test.tsx
```diff
@@ -34,7 +34,7 @@ describe('<Header />', () => {
       const warnSpy = jest.spyOn(console, 'warn');
       mountWithAppProvider(<Header {...mockProps} icon="foo" />);
 
-      expect(warnSpy).toHaveBeenCalledTimes(1);
+      expect(warnSpy).toHaveBeenCalled();
```
AndrewMusgrave: `React.StrictMode` is causing a second render. I'm not sure why, but I don't think it's a huge deal.

src/test-utilities/tests/legacy.test.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import {mountWithAppProvider} from '../legacy';
+
+describe('mountWithAppProvider', () => {
+  let consoleSpy: jest.SpyInstance;
+
+  beforeEach(() => {
+    consoleSpy = jest.spyOn(console, 'error');
+    consoleSpy.mockImplementation(() => {});
```
AndrewMusgrave: Silences the error

src/components/Page/components/Header/tests/Header.test.tsx
BPScott: It does that intentionally to try and make it side effects obvious as per https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects

.storybook/stories-from-readme.js
```diff
@@ -66,7 +66,9 @@ export function addPlaygroundStory(playgroundModule) {
 function AppProviderDecorator(story) {
   return (
     <div style={{padding: '8px'}}>
-      <AppProvider i18n={en}>{story()}</AppProvider>
+      <React.StrictMode>
```
BPScott: Instead of setting this in the AppProviderDecorator I think we should add a new [global decorator](https://storybook.js.org/docs/addons/introduction/#storybook-decorators) to make sure this applies to all stories. (Admittedly currently AppProvider ends up being applied to everything but at some point that might change if we want to pull in stories from places other than the readmes)

Can you revert this and then in .storybook/config.js under the `addParameters` call add:

```
const StrictModeDecorator = story => <React.StrictMode>{story()}</React.StrictMode>;
addDecorator(StrictModeDecorator);
```

src/components/Page/components/Header/tests/Header.test.tsx
AndrewMusgrave: Neat 😍 

Screen Shot 2019-06-20 at 2 22 55 AM

src/test-utilities/tests/legacy.test.tsx
danrosenthal: Which error is that?

BPScott: The one at the bottom of this file around `expect(consoleSpy).toHaveBeenCalledWith(`. This mockImplementation makes it so the error we're deliberately causing does not get output to the console as we expect it to appear.

PR #1710, opened 2019-06-19
Storybook: Make babel read ts files instead of using ts-loader

BPScott:

A few weeks ago I taught babel-plugin-shopify how to read typescript files with the flick of an option. We should use that instead having to pass TS files into ts-loader then into babel-loader.

- Enable typescript mode for our babel config
- Remove ts-loader from storybook config
- Remove `useBuiltins: entry` from babel config as it is the default so no need to specify it.

### To test:

- Ensure storybook runs without errors (check percy). There will be some console warnings regarding unfound imports of interfaces but they were there before too.

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -62,14 +62,6 @@ module.exports = ({config, mode}) => {
             cacheDirectory: `${cacheDir}/typescript`,
           },
         },
-        {
-          loader: 'ts-loader',
-          options: {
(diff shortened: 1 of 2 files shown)


kaelig:
Just in case, I checked performance, and saw no noticeable changes:

### Before
Screen Shot 2019-06-19 at 8 15 01 AM

### After
Screen Shot 2019-06-19 at 8 14 15 AM

PR #1713, opened 2019-06-19
Update sewing-kit

### LINKED ISSUE #1701: [v4] Check for sewing-kit update before release v4
See #1695

BPScott:
This will now warn on unsafe lifecycle hooks like componentWillReceiveProps

Fixes #1701.

### DIFF
--- a/src/components/Card/tests/Card.test.tsx
+++ b/src/components/Card/tests/Card.test.tsx
@@ -6,8 +6,10 @@ import {contentContextTypes} from '../../../types';
 
 describe('<Card />', () => {
   it('has a child with contentContext', () => {
-    const Child: React.SFC<{}> = (_props, context) =>
-      context.withinContentContainer ? <div /> : null;
+    const Child: React.SFC = (_props, context) => {
+      // eslint-disable-next-line shopify/jest/no-if
+      return context.withinContentContainer ? <div /> : null;
+    };
     Child.contextTypes = contentContextTypes;
 
     const containedChild = mountWithAppProvider(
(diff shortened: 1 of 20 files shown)


src/components/Card/tests/Card.test.tsx
```diff
@@ -6,8 +6,10 @@ import {contentContextTypes} from '../../../types';
 
 describe('<Card />', () => {
   it('has a child with contentContext', () => {
-    const Child: React.SFC<{}> = (_props, context) =>
-      context.withinContentContainer ? <div /> : null;
+    const Child: React.SFC = (_props, context) => {
```
BPScott: ping @cartogram This sort of thing feels like a false positive on `shopify/jest/no-if`. would you be able to look into it?

cartogram: Yeah, it is. Can you copy and paste this example into an issue on the eslint-plugin-shopify?

For now, keep the inline disable and when I fix the rule, the disable comment should be flagged but the no-disable-comments rule and you can remove it then. Thanks Ben!

BPScott: Done! https://github.com/Shopify/eslint-plugin-shopify/issues/311

src/components/TextField/TextField.tsx
```diff
@@ -143,6 +143,7 @@ class TextField extends React.PureComponent<CombinedProps, State> {
   constructor(props: CombinedProps) {
     super(props);
 
+    // eslint-disable-next-line react/state-in-constructor
```
AndrewMusgrave: We could pull this one out of the constructor if we wanted

PR #1715, opened 2019-06-19
[v4] Removed test errors / warning

AndrewMusgrave:

We're starting to get a bunch of errors, and non-deprecation warning in our tests and they're being very noisy 🙅‍♂ 

We currently have two errors left. One is from enzyme that will hopefully be fixed eventually 🤞 and the other is Tabs Item which is placing a ref on a functional component. We can either use forwardRef on UnstyledLink or find an alternative using a ref on a component. Which never would have allowed us to call `.focus` afaik. We'll have a reference to the component so [this](https://github.com/Shopify/polaris-react/blob/77ea29e563ce826870acd090bfed1b24973b8c7a/src/components/Tabs/components/Item/Item.tsx#L23) will never be true. I'll create a follow for us to follow up on this.

* adjusting tests by adding mocks or missing handlers


package.json
```diff
@@ -105,7 +105,7 @@
     "@storybook/addon-viewport": "5.1.0-rc.0",
     "@storybook/react": "5.1.0-rc.0",
     "@storybook/theming": "5.1.0-rc.0",
-    "@types/enzyme": "^3.9.1",
+    "@types/enzyme": "^3.9.3",
```
AndrewMusgrave: In the end, this upgrade didn't fix the enzyme bug but I'm keeping it in to keep us up to date

src/components/Frame/components/Toast/tests/Toast.test.tsx
```diff
@@ -94,14 +94,16 @@ describe('<Toast />', () => {
     });
 
     it('warns that a duration of 10000ms is recommended with action if duration is lower than 10000', () => {
-      const warnSpy = jest.spyOn(console, 'log');
+      const logSpy = jest.spyOn(console, 'log');
```
AndrewMusgrave: log -> warn 😄

src/components/Tabs/components/Item/Item.tsx
```diff
@@ -55,7 +55,9 @@ export default class Item extends React.PureComponent<Props, never> {
     };
 
     const markup = url ? (
-      React.createElement(UnstyledLink, {url, ...sharedProps}, children)
+      <UnstyledLink {...sharedProps} url={url}>
```
AndrewMusgrave: This didn't fix the error but we should have been doing this, to begin with. IIFC we had a type issue in the past forcing us to use `createElement` that's no longer present.

BPScott:
What here is v4 specific? I can't see a reason why this shouldn't be put on master instead?

AndrewMusgrave:
> What here is v4 specific? I can't see a reason why this shouldn't be put on master instead?

@BPScott About half of the errors started to occur with the react version bump

PR #1716, opened 2019-06-20
Autocomplete: Make TextField a subcomponent of AutoComplete

BPScott:

Having it in AutoComplete/components/ComboBox/components/TextField is weird when documented usage is to use `AutoComplete.TextField`.

Also UXPin doesn't like let me define AutoComplete.TextField with how it's currently setup, this will let me link to the TextField subcomponent without weirdness.

Move `AutoComplete/components/ComboBox/components/TextField` to  `AutoComplete/components/TextField`

### How to 🎩

Check the autocomplete still works


src/components/Autocomplete/components/ComboBox/ComboBox.tsx
```diff
@@ -7,7 +7,7 @@ import {PreferredPosition} from '../../../PositionedOverlay';
 import {ActionListItemDescriptor, Key} from '../../../../types';
 import {contextTypes} from '../types';
 import KeypressListener from '../../../KeypressListener';
-import {TextField} from './components';
```
BPScott: Exposing TextField and OptionList on ComboBox is really weird and we shouldn't do it, but removing it now would be a breaking change so lets just leave them there for now even though I'd bet a mars bar on nobody ever using it in that way because it's not documented.

src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -41,10 +40,6 @@ export interface Props {
 
 type CombinedProps = Props & WithAppProviderProps;
 
-function TextField(props: TextFieldProps) {
```
chloerice: 🔥

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
chloerice: With the change in export, the path to the TextField from the [ComboBox test TSX file needs to be updated ](https://circleci.com/gh/Shopify/polaris-react/22335?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).

PR #1718, opened 2019-06-20
Filters component

### LINKED ISSUE #1167: [List filters] Use `Sheet` from Polaris

### LINKED ISSUE #1510: [List filters] Move list filter component to Polaris

tmlayton:

Closes #1167
Closes #1510

Moving the `<Filters />` component to Polaris

<kbd></kbd>

### Documentation

<kbd></kbd>

## 

### How to 🎩

<details>

```jsx

interface State {}

  render() {
    return (
      <Page title="Playground">
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### Known issues

@danrosenthal did some digging and looks like any children which cause a reflow ([here is all of the things that can cause a reflow](https://gist.github.com/paulirish/5d52fb081b3570c81e3a)) during the entering animation will force the animation to skip to the end. For more details see https://github.com/reactjs/react-transition-group/issues/382. For now this is a won’t fix.

### Related todos

The following will be done in other PRs:

- [ ] Move `ResourceList.Item` to its own component named `ResourceItem`
- [ ] Update `ResourceList` examples to use `ResourceFilters`
- [ ] Deprecate `FilterControl`

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -10,6 +10,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### New components
 
+- `Filters`: Use to filter the items of a list or table ([#1718](https://github.com/Shopify/polaris-react/pull/1718))
+
 ### Enhancements
 
 ### Bug fixes
--- a/src/components/Filters/README.md
+++ b/src/components/Filters/README.md
@@ -1,0 +1,482 @@
+---
+name: Filters
+category: Lists and tables
+keywords:
+  - filters
+  - filtering
+  - filter control
+  - resource list
+  - index
+  - list filter
+  - table
+---
+
+# Filters
+
+Filters is a composite component that filters the items of a list or table.
+
+Merchants use filters to:
+
+- view different subsets of items in a list or table
+- filter by typing into a text field
+- filter by selecting filters or promoted filters
+
+The way that merchants interact with filters depends on the components that you decide to incorporate. In its simplest form, filters includes a text field and a set of filters, which can be displayed in different ways. For example, you could show promoted filters and a More button that opens a [sheet](/components/overlays/sheet) containing more filters. What the filters are and how they’re exposed to merchants is flexible.
+
+---
+
+## Accessibility
+
+The filters component relies on the accessibility features of multiple other components:
+
+- [Text field](https://polaris.shopify.com/components/forms/text-field)
+- [Button](https://polaris.shopify.com/components/actions/button)
+- [Popover](https://polaris.shopify.com/components/overlays/popover)
+- [Sheet](https://polaris.shopify.com/components/overlays/sheet)
+- [Collapsible](https://polaris.shopify.com/components/behavior/collapsible)
+
+### Maintain accessibility with custom features
+
+Since custom HTML can be passed to the component for additional actions, ensure that the filtering system you build is accessible as a whole.
+
+All merchants must:
+
+- be able to identify and understand labels for all controls
+- be notified of state changes
+- be able to complete all actions with the keyboard
+
+---
+
+## Best practices
+
+The filters component should:
+
+- help reduce merchant effort by promoting the filtering categories that are most commonly used
+- include no more than 2 or 3 promoted filters
+- consider small screen sizes when designing the interface for each filter and the total number filters to include
+- use children only for content that’s related or relevant to filtering
+
+---
+
+## Content guidelines
+
+### Text field
+
+The text field should be clearly labeled so it’s obvious to merchants what they should enter into the field.
+
+<!-- usagelist -->
+
+#### Do
+
+- Filter orders
+
+#### Don’t
+
+- Enter text here
+
+<!-- end -->
+
+### Filter badges
+
+Use the name of the filter if the purpose of the name is clear on its own. For example, when you see a filter badge that reads **Fulfilled**, it’s intuitive that it falls under the Fulfillment status category.
+
+<!-- usagelist -->
+
+#### Do
+
+- Fulfilled, Unfulfilled
+
+#### Don’t
+
+- Fulfillment: Fulfilled, Unfulfilled
+
+<!-- end -->
+
+If the filter name is ambiguous on its own, add a descriptive word related to the status. For example, **Low** doesn’t make sense out of context. Add the word “risk” so that merchants know it’s from the Risk category.
+
+<!-- usagelist -->
+
+#### Do
+
+- High risk, Low risk
+
+#### Don’t
+
+- High, Low
+
+<!-- end -->
+
+Group tags from the same category together.
+
+<!-- usagelist -->
+
+#### Do
+
+- (Unfulfilled, Fulfilled)
+
+#### Don’t
+
+- (Unfulfilled) (fulfilled)
+
+<!-- end -->
+
+If all tag pills selected: truncate in the middle
+
+<!-- usagelist -->
+
+#### Do
+
+- Paid, par… unpaid
+
+#### Don’t
+
+- All payment status filters selected, Paid, unpa…
+
+<!-- end -->
+
+---
+
+## Examples
+
+### Filtering with a resource list
+
+```jsx
+class FiltersExample extends React.Component {
+  state = {
+    accountStatus: null,
+    moneySpent: null,
+    taggedWith: null,
+    queryValue: null,
+  };
+
+  render() {
+    const {accountStatus, moneySpent, taggedWith, queryValue} = this.state;
+
+    const filters = [
+      {
+        key: 'accountStatus',
+        label: 'Account status',
+        filter: (
+          <ChoiceList
+            title={'Account status'}
+            titleHidden
+            choices={[
+              {label: 'Enabled', value: 'enabled'},
+              {label: 'Not invited', value: 'not invited'},
+              {label: 'Invited', value: 'invited'},
+              {label: 'Declined', value: 'declined'},
+            ]}
+            selected={accountStatus || []}
+            onChange={this.handleChange('accountStatus')}
+            allowMultiple
+          />
+        ),
+        shortcut: true,
+      },
+      {
+        key: 'moneySpent',
+        label: 'Money spent',
+        filter: (
+          <RangeSlider
+            label="Money spent is between"
+            labelHidden
+            value={moneySpent || [0, 500]}
+            prefix="$"
+            output
+            min={0}
+            max={2000}
+            step={1}
+            onChange={this.handleChange('moneySpent')}
+          />
+        ),
+        shortcut: true,
+      },
+      {
+        key: 'taggedWith',
+        label: 'Tagged with',
+        filter: (
+          <TextField
+            label="Tagged with"
+            value={taggedWith}
+            onChange={this.handleChange('taggedWith')}
+            labelHidden
+          />
+        ),
+        shortcut: true,
+      },
+    ];
+
+    const appliedFilters = Object.keys(this.state)
+      .filter((key) => !isEmpty(this.state[key]) && key !== 'queryValue')
+      .map((key) => {
+        return {
+          key,
+          label: disambiguateLabel(key, this.state[key]),
+          onRemove: this.handleRemove,
+        };
+      });
+
+    return (
+      <div style={{height: '568px'}}>
+        <Card>
+          <ResourceList
+            resourceName={{singular: 'customer', plural: 'customers'}}
+            filterControl={
+              <Filters
+                queryValue={queryValue}
+                filters={filters}
+                appliedFilters={appliedFilters}
+                onQueryChange={this.handleChange('queryValue')}
+                onQueryClear={this.handleQueryClear}
+                onClearAll={this.handleClearAll}
+              />
+            }
+            items={[
+              {
+                id: 341,
+                url: 'customers/341',
+                name: 'Mae Jemison',
+                location: 'Decatur, USA',
+              },
+              {
+                id: 256,
+                url: 'customers/256',
+                name: 'Ellen Ochoa',
+                location: 'Los Angeles, USA',
+              },
+            ]}
+            renderItem={(item) => {
+              const {id, url, name, location} = item;
+              const media = <Avatar customer size="medium" name={name} />;
+
+              return (
+                <ResourceList.Item
+                  id={id}
+                  url={url}
+                  media={media}
+                  accessibilityLabel={`View details for ${name}`}
+                >
+                  <h3>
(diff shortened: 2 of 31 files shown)


ry5n:
😍

src/components/ResourceFilters/README.md
```diff
@@ -0,0 +1,322 @@
...
+  - filtering
+  - filter control
+  - resource list
+  - index
+  - list filter
+---
+
+# Resource filters
+
+Resource filters is a composite component that filters the resource list component.
```
tmlayton: We could just call this component `Filters`. It is already entirely possible to use it without a resource list.

src/components/ResourceFilters/ResourceFilters.scss
```diff
@@ -0,0 +1,204 @@
...
+}
+
+.open {
+  &::before,
+  &::after {
+    content: '';
+    position: relative;
+    left: rem(16px);
+    width: calc(100% - #{rem(32px)});
+    height: 1px;
```
tmlayton: Do we always want 1px or should we convert to rem?

src/components/ResourceFilters/ResourceFilters.scss
```diff
@@ -0,0 +1,204 @@
...
+}
+
+.open.first {
+  &::after {
+    content: '';
+    bottom: 0;
+    position: relative;
+    left: rem(16px);
+    width: calc(100% - #{rem(32px)});
+    height: 1px;
```
tmlayton: Same here, px or rem?

src/components/ResourceFilters/ResourceFilters.scss
```diff
@@ -0,0 +1,204 @@
...
+}
+
+.open.last {
+  &::before {
+    content: '';
+    top: 0;
+    position: relative;
+    left: rem(16px);
+    width: calc(100% - #{rem(32px)});
+    height: 1px;
```
tmlayton: And here

src/components/ResourceFilters/ResourceFilters.tsx
```diff
@@ -0,0 +1,503 @@
...
+  label: string;
+  /** The markup for the given filter */
+  filter: React.ReactNode;
+  /** Whether or not the filter should have a shortcut popover displayed */
+  shortcut?: boolean;
+}
+
+export interface Props {
+  /** Currently entered text in the query field */
+  queryValue?: string;
```
tmlayton: Everywhere we were calling this "search" I updated to "query"

src/components/ResourceFilters/README.md
```diff
@@ -0,0 +1,322 @@
+---
+name: Resource filters
+category: Lists and tables
+keywords:
+  - filters
+  - filtering
+  - filter control
+  - resource list
+  - index
+  - list filter
```
tmlayton: Keywords okay?

src/components/ResourceList/index.ts
```diff
@@ -1,8 +1,9 @@
 import ResourceList from './ResourceList';
 
+export {Consumer} from './components';
 export * from './components/FilterControl/types';
 export {Props as FilterControlProps} from './components/FilterControl';
 export {Props} from './ResourceList';
-export {SelectedItems} from './types';
+export {SelectedItems, ResourceListContext} from './types';
```
tmlayton: Exporting this to use the context in filters, specifically to get the `resourceName`

src/components/Sheet/Sheet.tsx
```diff
@@ -72,27 +76,23 @@ class Sheet extends React.Component<ComposedProps, State> {
...
 
     if (frame == null) {
       // eslint-disable-next-line no-console
       console.warn(intl.translate('Polaris.Sheet.warningMessage'));
     }
 
-    if (mobile !== isMobile()) {
-      handleToggleMobile();
-    }
+    this.handleResize();
```
tmlayton: Made this change simply to dry up code and get 100% test coverage

src/components/Sheet/Sheet.tsx
```diff
@@ -39,6 +39,10 @@ export interface Props {
   children: React.ReactNode;
   /** Callback when the backdrop is clicked or `ESC` is pressed */
   onClose(): void;
+  /** Callback when the sheet has completed entering */
+  onEntered?(): void;
+  /** Callback when the sheet has started to exit */
+  onExit?(): void;
```
tmlayton: Added this specifically to solve the issue where the sheet transition would skip directly to the end when I child element had focus.

|Before|After|
|---|---|
|before|after|

Although, there still appears to be an issue with the range slider:

range slider

src/components/ResourceFilters/components/ConnectedFilterControl/ConnectedFilterControl.tsx
```diff
@@ -0,0 +1,198 @@
...
+        </div>
+      </React.Fragment>
+    );
+  }
+
+  private measureProxyButtons() {
+    if (this.proxyButtonContainer.current) {
+      const proxyButtonsWidth: ComputedProperty = {};
+      // this number is magical, but tweaking it solved the problem of items overlapping
+      const tolerance = 52;
```
tmlayton: Just leaving this as-is for now

src/components/ResourceFilters/README.md
BPScott: I was going to say - if this is just for use in ResourceLists should it live at `ResourceList.ResourceFilter`. If this can be used divorced from ResourceList can we see and example of it stand alone?

tmlayton: Yes, if we keep this as-is we will alias it so it is also available at `ResourceList.Filters`. But it will stay as a top-level component for documentation reasons. We are going to do the same with `ResourceList.Item` and move it to a top-level component as `ResourceItem` so we can split up the `ResourceList` documentation.

danrosenthal: I think I'm leaning towards `Filters`, since this really could be used in any list context. Like you said, it is tied to Resource List context, but it doesn't need that in order to function. This component truly is flexible, so lets not use a naming convention that conceals that flexibility.

src/components/ResourceFilters/ResourceFilters.scss
tmlayton: Updated to rems

src/components/Sheet/Sheet.tsx
tmlayton: @danrosenthal did some digging and looks like any children which cause a reflow during the entering animation will force the animation to skip to the end. https://github.com/reactjs/react-transition-group/issues/382

src/components/ResourceFilters/README.md
tmlayton: Okay, I’ve made the change so this is now the `Filters` component and added an example of using it with data table.

BPScott: Nice! For the record - I dislike the ideas of creating aliases to top level components. I think we should avoid it to reduce potential for confusion. People will wonder "what's the difference between ResourceItem and ResourceList.Item" and the answer is there isn't one, they're identical.

Dragging ResourceList-specific subcomponents up into the top level because subcomponent documentation is a bit crap feels like the wrong solution to the problem - long-term I'd rather invest in making the styleguide render pages for public sub-components instead of dragging stuff up to the top level just so they get decent docs.

But that'll take some time so I'm happy to punt on that for the moment.

tmlayton:
I went ahead an updated the component to be called `Filters` and provided an example of using it with data table and resource list.

src/components/Filters/README.md
```diff
@@ -0,0 +1,481 @@
...
+- be notified of state changes
+- be able to complete all actions with the keyboard, efficiently and easily
+
+---
+
+## Best practices
+
+The filters component should:
+
+- help reduce merchant effort by promoting the filters that are most commonly used
```
sadiesaurus:
```suggestion
- help reduce merchant effort by promoting the filtering categories that are most commonly used
```

src/components/Filters/README.md
```diff
@@ -0,0 +1,481 @@
...
+- include no more than 2 or 3 promoted filters
+- consider small screen when deciding the number filters to include in a popover
+- let merchants select more than one filter from within the same category. For example, in the payment filter category, merchants can combine paid, partially paid, and unpaid
```
sadiesaurus: We don't always do this. For example, with order list filters, you can add only one filter from the Status category at a time. 

```suggestion
- In most cases, let merchants select more than one filter from within the same category. For example, from the Payment status filter category, merchants can combine the Payment paid, Payment unpaid, and Payment partially paid filters. 
```

@sarahill Do you think we should mention the rationale for when to not allow merchants to combine filters from the same category, like we do with the Status filters? If yes, what is the rationale?

tmlayton:
> I didn't see the related components section. Did we decide it was best to not include it here?

Yes, the format for related components is to suggest other components if trying to do something related, but we have no other filter component so it didn’t feel right to include.

UNRELEASED.md
```diff
@@ -15,6 +15,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### New components
 
 - `ActionMenu`: Use for display of actions and action groups within the context of a header ([#1653](https://github.com/Shopify/polaris-react/pull/1653))
+- `Filters`: Use to filter the `ResourceList` ([#1718](https://github.com/Shopify/polaris-react/pull/1718))
```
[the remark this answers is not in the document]
tmlayton: How about this?

```suggestion
- `Filters`: Use to filter the items of a list or table ([#1718](https://github.com/Shopify/polaris-react/pull/1718))
```

src/components/Filters/README.md
```diff
@@ -0,0 +1,481 @@
...
+  - filter control
+  - resource list
+  - index
+  - list filter
+  - table
+---
+
+# Filters
+
+Filters is a composite component that filters the items of a list or table.
```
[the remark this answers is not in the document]
tmlayton: I prefer being general because it is a bit more future-proof and leaves it open for creative/flexible implementations if, for example, a consumer has built their own kind of custom list.

src/components/Filters/README.md
```diff
@@ -0,0 +1,483 @@
...
+
+---
+
+## Best practices
+
+The filters component should:
+
+- help reduce merchant effort by promoting the filtering categories that are most commonly used
+- include no more than 2 or 3 promoted filters
+- consider small screen sizes when deciding the number filters to include in a popover
```
tmlayton:
```suggestion
- consider small screen sizes when deciding the interface for each filter and the total number filters to include
```

src/components/Filters/README.md
```diff
@@ -0,0 +1,483 @@
...
+
+---
+
+## Best practices
+
+The filters component should:
+
+- help reduce merchant effort by promoting the filtering categories that are most commonly used
+- include no more than 2 or 3 promoted filters
+- consider small screen sizes when deciding the interface for each filter and the total number filters to include
```
tmlayton: @sadiesaurus how’s this?

src/components/Filters/README.md
tmlayton: Yeah I’m not sure we even want to include this. I think this presumes too much about a consumers data. Specifically, it would be for a filter category that has multiple options.

src/components/Filters/README.md
```diff
@@ -0,0 +1,481 @@
...
+
+- Filter orders
+
+#### Don’t
+
+- Enter text here
+
+<!-- end -->
+
+### Filter badges
```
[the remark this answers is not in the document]
tmlayton: We use both tags and badges in this component. Badges and pills would be interchangeable (both just read-only). Tags have a `X` and are actionable.

src/components/Filters/README.md
```diff
@@ -0,0 +1,481 @@
...
+
+Use the name of the filter if the purpose of the name is clear on its own. For example, when you see a filter badge that reads **Paid**, it’s intuitive that it falls under the Payment status category.
+
+<!-- usagelist -->
+
+#### Do
+
+- Paid, Unpaid
```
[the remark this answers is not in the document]
tmlayton: Maybe we update to 

```
#### Do

- Payment paid, Payment unpaid

#### Don’t

- Payment: Paid, Unpaid

```

src/components/Filters/README.md
sadiesaurus: Looks good! One tiny word change suggestion:

```suggestion
- consider small screen sizes when designing the interface for each filter and the total number filters to include
```

src/components/Filters/README.md
```diff
@@ -0,0 +1,483 @@
...
+
+#### Don’t
+
+- Enter text here
+
+<!-- end -->
+
+### Filter badges
+
+Use the name of the filter if the purpose of the name is clear on its own. For example, when you see a filter badge that reads **Paid**, it’s intuitive that it falls under the Payment status category.
```
sadiesaurus:
```suggestion
Use the name of the filter if the purpose of the name is clear on its own. For example, when you see a filter badge that reads “Fulfilled”, it’s intuitive that it falls under the Fulfillment status category.
```

src/components/Filters/README.md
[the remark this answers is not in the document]
sadiesaurus: Yes, or we could change the example to continue illustrating the point that a single word can be used if the category is implied. So, we could use Fulfilled and Unfulfilled instead. 

```
#### Do

- Fulfilled, Unfulfilled

#### Don’t

- Fulfillment: Fulfilled, Unfulfilled

```

Thoughts?

src/components/Filters/README.md
```diff
@@ -0,0 +1,483 @@
...
+
+- (Unfulfilled, Fulfilled)
+
+#### Don’t
+
+- (Unfulfilled) (fulfilled)
+
+<!-- end -->
+
+If all tag pills selected: truncate in the middle
```
sadiesaurus: We might want to remove this guideline, since this doesn't appear to be what we're doing with order filters? 

Live in the product:
alt

src/components/Filters/README.md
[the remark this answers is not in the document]
tmlayton: I like changing the example

src/components/Filters/README.md
tmlayton: "Do as I say, not as I do" 😆

tmlayton: Both of these are concerns with the badge and tag component. We would need to make this change in tag and then for badge it would be up to the consumer (maybe?). Not great for us to use it differently than we recommend, but also think that is okay.

tmlayton:
Okay going to ship! If there are any other changes to docs we can do them in follow-up PRs.

PR #1722, opened 2019-06-20
[Avatar] Follow content guidelines for dimensions

kaelig:

When referring to dimensions, as seen in the [icons page](https://polaris.shopify.com/design/icons#section-system-icons), we use the multiplication sign and no space between the number and the unit.

Screen_Shot_2019-06-20_at_12_05_45_PM

Uses a multiplication sign instead of an X, and removes space between the number and the unit.


kaelig:
Thank you @sadiesaurus – I’ve updated the pull request!

PR #1725, opened 2019-06-21
[ActionMenu] Move margin-top to Page > Header.scss

beefchimi:
While building out some UI in my app, I realized that I stuck `margin-top` on the `ActionMenu` when in reality it should be up to the consuming parent how to provide spacing.

All I have done is simply move `margin-top` from `ActionMenu.scss` into `Header.scss`. The visual result within Polaris is exactly the same.

I'm keeping the negative `margin-left` as that is specifically for offsetting the action padding - so that spacing logic makes sense within `ActionMenu`.

Below is a gif of my app. This is with the **current ActionMenu** _(with `margin-top`)_. You can see me toggling the `margin-top` on/off. Once this PR merges, I won't have to worry about this spacing discrepancy.

thingy

Also, I have applied the `skip changelog` label, as there is no change for the end user, and the ActionMenu has not been released yet.

### DIFF
--- a/src/components/ActionMenu/ActionMenu.tsx
+++ b/src/components/ActionMenu/ActionMenu.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import classNames from 'classnames';
+import {classNames} from '@shopify/css-utilities';
 
 import {
   ActionListSection,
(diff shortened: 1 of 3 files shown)


src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import classNames from 'classnames';
+import {classNames} from '@shopify/css-utilities';
```
beefchimi: Consistent import for `classNames`

PR #1727, opened 2019-06-24
Move storybook to devDependencies

AndrewMusgrave:
Looks like these were accidentally added as dependencies in #1666


UNRELEASED.md
```diff
@@ -44,6 +44,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Code quality
 
+- Moved Storybook to devDependcies ([#1727](https://github.com/Shopify/polaris-react/pull/1727))
```
kaelig: Not needed

PR #1741, opened 2019-06-26
Fixed toast manager

AndrewMusgrave:

StrictMode found an error in toast and is currently breaking it. `ToastManager` is currently relying on side-effects to work properly

* abstracting some of the logic in useDeepCompare to create useDeepCallback
* change toast manager logic
* add eslint ignore because we're smarter 🤔 

### How to 🎩

Check out toast examples in the deploy 😄 

If you want to see `Toast` not working check out `version-4.0.0` (doesn't currently have a deployment)

### DIFF
--- a/src/components/Frame/components/ToastManager/ToastManager.tsx
+++ b/src/components/Frame/components/ToastManager/ToastManager.tsx
@@ -1,10 +1,12 @@
-import React from 'react';
+import React, {createRef, memo} from 'react';
 import {TransitionGroup, CSSTransition} from 'react-transition-group';
 import {classNames} from '../../../../utilities/css';
 import EventListener from '../../../EventListener';
 import Portal from '../../../Portal';
 import {ToastPropsWithID} from '../../types';
 import Toast from '../Toast';
+import useDeepCompare from '../../../../hooks/use-deep-compare';
+import {useDeepCallback} from '../../../../utilities/use-deep-callback';
 
 import styles from './ToastManager.scss';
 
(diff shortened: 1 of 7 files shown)


src/components/Frame/components/ToastManager/ToastManager.tsx
```diff
@@ -1,77 +1,80 @@
...
+function ToastManager({toastMessages}: Props) {
+  const toastNodes: React.RefObject<HTMLDivElement>[] = [];
 
-  componentDidUpdate() {
-    this.updateToasts();
-  }
-
-  render() {
-    const {toastMessages} = this.props;
+  const updateToasts = useDeepCallback(
```
AndrewMusgrave: We need deep comparison because toastMessages is a complex object

src/components/Frame/components/ToastManager/ToastManager.tsx
```diff
@@ -1,77 +1,80 @@
...
+      });
+    },
+    [toastMessages, toastNodes],
+  );
 
-    const toastsMarkup = toastMessages.map((toast, index) => {
-      this.toastNodes[index] = React.createRef();
+  useDeepCompare(
+    () => {
+      updateToasts();
```
AndrewMusgrave: We need to update toasts in `useEffect` hook so we can access the refs

src/components/Toast/Toast.tsx
```diff
@@ -61,7 +61,8 @@ export default React.memo(function Toast(props: Props) {
         }
       };
     },
-    [appBridge, frame, props],
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [appBridge, props],
```
AndrewMusgrave: We're smarter than eslint in this case. `frame` won't change so this will save us a few renders

src/hooks/use-deep-compare/use-deep-compare.tsx
```diff
@@ -1,22 +1,6 @@
-import {useEffect, useRef} from 'react';
-import isEqual from 'lodash/isEqual';
-
-type EffectCallback = () => void | (() => void | undefined);
-type DependencyList = ReadonlyArray<unknown>;
-type Comparator = (a: DependencyList, b: DependencyList) => boolean;
-
-function useDeepCompareRef(
```
AndrewMusgrave: Abstracting this logic so it can be used in `useDeepCallback` as well

src/utilities/use-deep-callback.tsx
```diff
@@ -0,0 +1,11 @@
+import {useCallback} from 'react';
```
dleroux: I know these are pretty simple but I wonder if adding tests would be worth it. If anything for documentation purposes. What do you think?

dleroux:
Something isn't working as expected. The placement or order of multiple Toasts does not change. In the current implementation it doesn't matter which Toast appears first, the second Toast will always move above it:

toast1

In this PR the second toast might end up pushing up the last:

toast2

src/components/Frame/components/ToastManager/ToastManager.tsx
```diff
@@ -1,77 +1,80 @@
...
-          classNames={toastClasses}
-        >
-          <div ref={this.toastNodes[index]} aria-live="polite">
-            <Toast {...toast} />
-          </div>
-        </CSSTransition>
-      );
-    });
+  const toastsMarkup = toastMessages.map((toast, index) => {
+    const toastNode = createRef<HTMLDivElement>();
```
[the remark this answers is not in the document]
AndrewMusgrave: Ideally, we wouldn't want to create new refs every render, and would rather use `useRef` however hooks need to run in the same order on every render and with a dynamic list (toast messages) that's something we're not able to do 😢I had this working with an alternative approach (using querySelector), but `createRef` can run a few million ops/s so I decided to stick with createRef.

src/utilities/use-deep-callback.tsx
AndrewMusgrave: In general, the approach to testing hooks is by the implementation, rather than internals. So using them in a component, and testing the component. That's not to say we shouldn't test hooks, at least in this situation the hook in so minimal we would almost be testing react internals. If you have a suggestion my ears are open 😄  Right now web is using the test by implementation approach, and not specifically testing hooks. We could always spend some time looking at testing approaches.

src/components/Toast/Toast.tsx
AndrewMusgrave: I was able to remove the comment because we're using deep compare now.

AndrewMusgrave:
@dleroux I changed https://github.com/Shopify/polaris-react/pull/1741/commits/18758ce2434548b9d98571256298b6012390b7f5#diff-aec6ff53816684e4d51d1484b4dc717bR22 to reduce renders since it's a complex object and it fixed it

src/utilities/use-deep-callback.tsx
dleroux: I find the tests they have in quilt useful because they clearly demonstrate what is expected of the hook.

PR #1745, opened 2019-06-26
Rename yarn script `ts` to `type-check`

BPScott:

Every other shopify project I work on uses type-check, and my muscle memory
hates us being different

Renames `yarn run ts` to `yarn run type-check`


package.json
```diff
@@ -32,12 +32,12 @@
   "scripts": {
     "lint": "sewing-kit lint",
     "format": "sewing-kit format",
-    "ts": "tsc --noEmit",
+    "type-check": "sewing-kit type-check",
```
AndrewMusgrave: What does sewing-kit run under the hood?

BPScott: It still runs `tsc --noEmit` under the hood

PR #1753, opened 2019-06-27
Simplify the internals of our styles plugin

BPScott:

This is the preliminary work on my investigation of how we can have a simpler and stricter sass build.

This simplifies the contents of our styles rollup plugin while retaining functionality identical to current build output. The one change is that the final item in `styles` objects that gets provided to js files
does not have a trailing comma.

e.g. 

```
const object = {
foo: "Foo",
bar: "Bar",
};
```

becomes

```
const object = {
foo: "Foo",
bar: "Bar"
};
```

This will make future changes easier to make and understand as there's less indirection and boilerplate.

- Prefer using sync APIs over async ones where possible (cssnano is the only case where a promise is required)
- Use JSON.stringify over an object instead of rolling our own version.
- Hardcode some values instead of them being options. We only ever call this in one place so we can be opinionated in how it works to reduce indirection.

### How to 🎩

- Checkout master do a build and copy it to a master-build folder `git checkout master && yarn && yarn run build && mv build master-build`
- Checkout this branch and do a build `git checkout simpler-sass && yarn && yarn run build`
- Diff the master-build and build folders `diff -ru master-build build`
- Note that the only changes is the removal of  trailing commas on the final item of style objects

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -22,6 +22,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - Renamed `yarn run ts` to `yarn run type-check` to match most other Shopify projects
 - Fixed deprecation notice in build ([#1754](https://github.com/Shopify/polaris-react/pull/1754))
+- Simplified our rollup plugin for sass compilation while retaining identical behaviour ([#1753](https://github.com/Shopify/polaris-react/pull/1753))
 
 ### Dependency upgrades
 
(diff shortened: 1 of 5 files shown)


BPScott:
@tmlayton sorry to be a pain in the ass, but could you give it a check again. I've tweaked some bits as doing avoiding async everywhere made things slightly slower by about a second. I've brought back some async calls and it's nearer its old speed.

BPScott:
Note to self: When doing compare builds docs, talk about the exact commit that this is based off instead of always talking about master

PR #1756, opened 2019-06-28
Removed css transition from popover overlay

AndrewMusgrave:

This component wasn't playing nice when trying to get `react-transition-group` strict mode compatible. `PopoverOverlay` is only animating opacity so this simplifies the animation since `CSSTransition` is not needed.

@dleroux and I 

* Removed `CSSTransitionGroup`
* Animated opacity
* Fixed tests

### How to 🎩

Check out popover examples in the development 🎉 

#### Animation in

Screen Shot 2019-06-28 at 11 39 20 AM

#### Animation out

Screen Shot 2019-06-28 at 11 39 40 AM


AndrewMusgrave:
@dleroux Replaced raf with a forced reflow (and removed test mocks) to reduce test mocking. Going to merge unless you have any objections

PR #1759, opened 2019-07-02
[v4] Switch react-transition-group versions

AndrewMusgrave:

Some of our dependencies use `findDOMNode` which is a strict mode nono 😭 

* enzyme uses `findDOMNode`
* original react-transition-group uses `findDOMNode`

* remove strict mode from our legacy tests
* remove strict mode test
* use a forked version of react-transition-group that removes `findDOMNode`
* add some refs

### How to 🎩

Check out animations in the deployment (Frame, Toast, Modal, Dialog, ReosurceList, and Sheet)

### Notes

* Tests are going to fail till #1756 is merged because we're removing `CSSTransition` from `PopoverOverlay`
* [Put a PR up in `react-transition-group`](https://github.com/reactjs/react-transition-group/pull/514) to suggest an alternative to `findDOMNode` but I'm not sure they'll accept the change since it's an api change and this doesn't seem like a priority


config/typescript/react-transition-group.d.ts
```diff
@@ -0,0 +1,21 @@
+declare module '@material-ui/react-transition-group' {
```
dleroux: Why did we need to declare this module?

src/components/ResourceList/components/BulkActions/BulkActions.scss
```diff
@@ -124,6 +124,10 @@ $bulk-actions-offset-slide-in-start: rem(-40px);
   }
 }
 
+.CheckableContainer {
+  flex: 1 1 0;
```
dleroux: Why was this needed?

src/test-utilities/legacy.tsx
```diff
@@ -151,15 +151,13 @@ export function polarisContextReactWrapper<P, S>(
     }
 
     return (
-      <React.StrictMode>
```
dleroux: Why did you remove Strictmode, wasn't this the reason for this PR?

config/typescript/react-transition-group.d.ts
AndrewMusgrave: Every package we consume needs types, however, most don't include them in the package. Polaris does include types, which is why we don't have a `@types/shopify__polaris-react`. `react-transition-group` does have a types package, which we're still consuming and I'm using to create type definitions for `@material-ui/react-transition-group` which doesn't have a `@types` package

src/test-utilities/legacy.tsx
AndrewMusgrave: Enzyme still uses it 😢 Although `@shopify/react-testing` and storybook don't 😄

PR #1783, opened 2019-07-04
[Tooltip] Implement fix for MouseOver event not firing when cursor enters after leaving a disabled element

### LINKED ISSUE #1782: [Tooltip] Not rendered when entering from a disabled neighbor
## Issue summary

Official React bug report: https://github.com/facebook/react/issues/10109

## Expected behavior

Moving mouse from a disabled, `<Tooltip />`-wrapped button onto an enabled `<Tooltip />`-wrapped button should trigger the display of the active button's tooltip.

## Actual behavior

The tooltip is not displayed. The following GIF shows an example of the bug:

polaris-pagination-tooltip-bug

Render two buttons side-by-side (e.g., the `<Pagination />` component), disable one button, and wrap the other in a `<Tooltip />`.

## Reduced test case

https://codesandbox.io/s/hopeful-moser-8242n

## Specifications

- Polaris version number: 3.18.0
- Browser: Chrome, Firefox (likely others)
- Device: Macbook Pro
- Operating System: macOS Mojave (10.14.5)

```bash
  System:
    OS: macOS 10.14.5
    CPU: x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
    Memory: 876.51 MB / 16.00 GB
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 10.13.0 - ~/.nvm/versions/node/v10.13.0/bin/node
    Yarn: 1.13.0 - ~/.dev/yarn/v1.13.0/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v10.13.0/bin/npm
  Browsers:
    Chrome: 75.0.3770.100
    Firefox: 67.0
    Safari: 12.1.1
  npmPackages:
    react: ^16.4.0 => 16.4.0
    react-dom: ^16.4.0 => 16.4.0
```

rexmac:

Fixes #1782.

MouseEnter events are not triggered when cursor enters from a disabled element.

Uses MouseOver with a "mouseEntered" boolean flag rather than MouseEnter.

#### Before
polaris-tooltio-mouseenter-bug

#### After
polaris-tooltip-mouseenter-fix

<details>

```jsx
import {Page, Tooltip} from '../src';

interface State {}

  render() {
    return (
      <Page title="Playground">
        <button type="button" disabled>
          Previous
        </button>
        <Tooltip content="Next">
          <button type="button">Next</button>
        </Tooltip>
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Tooltip/tests/Tooltip.test.tsx
+++ b/src/components/Tooltip/tests/Tooltip.test.tsx
@@ -31,8 +31,8 @@ describe('<Tooltip />', () => {
     );
   });
 
-  it('renders on mouseEnter', () => {
-    wrapperComponent.simulate('mouseEnter');
+  it('renders on mouseOver', () => {
+    wrapperComponent.simulate('mouseOver');
     expect(findByTestID(tooltip, 'TooltipOverlayLabel').exists()).toBe(true);
   });
 
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -14,6 +14,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- Fixed a bug preventing the display of `Tooltip` when cursor enters from a disabled element ([#1783](https://github.com/Shopify/polaris-react/pull/1783)).
+
 ### Documentation
 
 ### Development workflow
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -118,6 +118,7 @@ export default class Tooltip extends React.PureComponent<Props, State> {
   };
 
   private handleMouseEnter = () => {
+    this.mouseEntered = true;
     this.setState({active: true});
   };
 
@@ -129,10 +130,7 @@ export default class Tooltip extends React.PureComponent<Props, State> {
   // https://github.com/facebook/react/issues/10109
   // Mouseenter event not triggered when cursor moves from disabled button
   private handleMouseEnterFix = () => {
-    if (!this.mouseEntered) {
-      this.mouseEntered = true;
-      this.handleMouseEnter();
-    }
+    !this.mouseEntered && this.handleMouseEnter();
   };
 
   private setAccessibilityAttributes() {
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -43,6 +43,7 @@ export default class Tooltip extends React.PureComponent<Props, State> {
 
   private id = getUniqueID();
   private activatorContainer: HTMLElement | null;
+  private mouseEntered = false;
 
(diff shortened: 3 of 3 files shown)


ry5n:
I asked @dpersing to take a look at this since it has accessibility implications. Keyboard and screen reader users already won’t be able to read the tooltip, so I’m not sure this change makes anything worse. I do wonder if this might open up tooltips to be used in more situations where maybe they shouldn’t be…

dpersing:
Hi @rexmac! Could you include the Playground code you show in the gifs so we can manually text your example? There should be a section of the PR template for it:

``````
<!--
  Give as much information as needed to experiment with the component
  in the playground.
-->

<details>

```jsx

interface State {}

  render() {
    return (
      <Page title="Playground">
      </Page>
    );
  }
}
```

</details>
``````

**Comment by rexmac** [comment]
@dpersing Added. 😃

**Comment by dpersing** [comment]
This looks fine to me @ry5n and @rexmac! Hover/focus tooltips are still weird for folks on mobile and there's a known issue with how tooltips are read by NVDA, but on desktop this appears to work just as I'd expect for the current implementation.

File: src/components/Tooltip/tests/Tooltip.test.tsx [hunk 1]
```diff
@@ -31,8 +31,11 @@ describe('<Tooltip />', () => {
     );
   });
 
-  it('renders on mouseEnter', () => {
-    wrapperComponent.simulate('mouseEnter');
+  it('renders on mouseOver', () => {
+    wrapperComponent.simulate('mouseOver');
```
AndrewMusgrave: Do we need the extra simulate and expect?

AndrewMusgrave: Just saw the commit message was for codecov. Maybe we should break the assertion into its own test and expect mouseOver is called once?

rexmac: Ya, I was testing to see if it would satisfy codecov, which it did. It definitely feels hacky, but I'm not sure how else to satisfy codecov. I think ideally we'd assert that `handleMouseEnter` is only called once, but it's a private method.

>Maybe we should break the assertion into its own test and expect mouseOver is called once?

I don't follow. Can you elaborate?

AndrewMusgrave: Ignore that, I was under the assumption we exposed a mouseEnter handler 😅 These changes may give you codecov
```tsx
  private handleMouseEnter = () => {
    this.mouseEntered = true;
    this.setState({active: true});
  };

  private handleMouseEnterFix = () => {
    !this.mouseEntered && this.handleMouseEnter();
  };
```

rexmac: Nice! I'll give that a shot. :+1:

PR #1785, opened 2019-07-04
[Frame] Make skip to content link an actual link

dleroux:

Fixes https://github.com/Shopify/polaris-react/issues/1746

Uses an actual link for the `skipToContent` link as opposed to a button.

This was done as opposed to using a `<Link>`or a `<Button>` with a URL because there's an issue with react router and anchors. By using an `<a>` 

skiptocontent

## 

### How to 🎩

<details>

```jsx

interface State {}

  render() {
    return (
      <Page title="Playground">
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Frame/Frame.tsx
+++ b/src/components/Frame/Frame.tsx
@@ -4,7 +4,6 @@ import {classNames} from '@shopify/css-utilities';
 import {durationSlow} from '@shopify/polaris-tokens';
 import {CSSTransition} from 'react-transition-group';
 import {navigationBarCollapsed} from '../../utilities/breakpoints';
-import Button from '../Button';
 import Icon from '../Icon';
 import EventListener from '../EventListener';
 import {withAppProvider, WithAppProviderProps} from '../AppProvider';
@@ -72,6 +71,8 @@ class Frame extends React.PureComponent<CombinedProps, State> {
   private contextualSaveBar: ContextualSaveBarProps | null;
 
   private globalRibbonContainer: HTMLDivElement | null = null;
+
+  private mainContentNode = React.createRef<HTMLDivElement>();
 
   getChildContext(): FrameContext {
     return {
@@ -212,13 +213,15 @@ class Frame extends React.PureComponent<CombinedProps, State> {
 
     const skipMarkup = (
       <div className={skipClassName}>
-        <Button
-          onClick={this.handleClick}
+        <a
+          href={`#${APP_FRAME_MAIN}`}
           onFocus={this.handleFocus}
           onBlur={this.handleBlur}
+          onClick={this.handleClick}
+          className={styles.SkipAnchor}
         >
           {intl.translate('Polaris.Frame.skipToContent')}
-        </Button>
+        </a>
       </div>
     );
 
@@ -261,6 +264,8 @@ class Frame extends React.PureComponent<CombinedProps, State> {
           className={styles.Main}
           id={APP_FRAME_MAIN}
           data-has-global-ribbon={Boolean(globalRibbon)}
+          ref={this.mainContentNode}
+          tabIndex={-1}
         >
           <div className={styles.Content}>{children}</div>
         </main>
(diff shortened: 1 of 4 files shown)


src/components/Frame/Frame.tsx
```diff
@@ -212,13 +211,14 @@ class Frame extends React.PureComponent<CombinedProps, State> {
 
     const skipMarkup = (
       <div className={skipClassName}>
-        <Button
-          onClick={this.handleClick}
+        <a
```
dpersing: 👍

src/components/Frame/Frame.tsx
```diff
@@ -351,10 +351,6 @@ class Frame extends React.PureComponent<CombinedProps, State> {
     }
   };
 
-  private handleClick = () => {
```
dpersing: In this case, we do want to preserve the focus handling, since browsers handle this a bit differently for on-page links. Safari in particular doesn't actually move focus by itself.

[You can see this happening most easily with VoiceOver running (video).](https://drive.google.com/open?id=1BwV3roTf3RGRUwcgwCjc3tLE208HRyKH) In this video I'm tabbing to the "Skip to content" link, pressing enter to fire it, and then while keyboard focus does technically skip to the `<main>` container, VoiceOver's cursor moves to the next element in the DOM after the link is hidden again. The next keystroke is shift + tabbing backwards, to show how the keyboard focus order makes sense if focus was in the main section, but the VoiceOver cursor appears to jump down to the end of the nav sidebar.

Safari on iOS behaves the same way.

Let me know if that makes sense! I'm happy to spend some office hours time next week testing this together.

dleroux: I've added focus on the main element. However in Safari Voiceover tells you to click `ctrl-option-spacebar` to follow the link. Doing that works great, pushing `enter` alone does not. Is that as expected?

src/components/Frame/Frame.tsx
```diff
@@ -363,6 +363,14 @@ class Frame extends React.PureComponent<CombinedProps, State> {
     this.setState({skipFocused: false});
   };
 
+  private handleClick = () => {
+    if (this.mainContentNode.current == null) {
+      return;
+    }
+
+    this.mainContentNode.current.focus();
```
AndrewMusgrave: main isn't a focusable element

src/components/Frame/Frame.tsx
dpersing: > Doing that works great, pushing enter alone does not. Is that as expected?

I'm able to activate the link with Enter, as well as use the VO-specific recommended keys. However, either causes a page refresh instead of moving focus.

I'm curious as to why this worked as a button but not a link. Do you have some time before Thursday to test through this together during office hours? @dleroux

src/components/Frame/Frame.tsx
dleroux: The interesting thing is that this works as expected as it ends up moving the focus to the correct location. It looks like chrome moves the focus to the div, therefore, tabbing works great. Safari ends up moving the focus to the heading which is really nice.

@dpersing this came to light when writing tests. The `active element` will never be the `main` element unless we actually add a `tabIndex` to it. Do you see an issue with adding a tabIndex to `main`?

dpersing: > Do you see an issue with adding a tabIndex to main?

That should be fine! It's how focus is managed to modals currently.

src/components/Frame/Frame.tsx
dleroux: adding `tabIndex={-1}` did the trick.

As for jumping out of the frame in the playground, I have a feeling that the playground is doing that because the behavior was the same before this change.

Building to web shows that this is working well there.

beefchimi:
@dleroux I will be making a similar change to a similar component _(originally cloned from this same SkipTo button)_ - curious if you folks are interested in separating this into its own component, rather than being integrated into the `Frame`.

If so, you'd essentially just need to clone:
https://github.com/Shopify/online-store-ui/tree/master/src/components/SkipToAction

dleroux:
That would be great! Do you feel like submitting a PR? 🙏 :)

PR #1790, opened 2019-07-05
[DataTable] Add vertical alignment prop

Mangara:

Fixes https://github.com/Shopify/polaris-react/issues/1789

There is no good way to adjust the vertical alignment of text in `DataTable` cells:

Screen Shot 2019-07-05 at 09 57 45

This PR adds a `verticalAlign` prop to `DataTable` that controls the vertical alignment of content in the cells.

Possible values are `top`, `bottom`, `middle`, and `baseline`, mirroring the [possible CSS values](https://developer.mozilla.org/en-US/docs/Web/CSS/vertical-align#Values_for_table_cells).
To preserve backwards-compatibility, the default value is `top`.

### 🎩 checklist

### DIFF
--- a/src/components/DataTable/DataTable.tsx
+++ b/src/components/DataTable/DataTable.tsx
@@ -9,7 +9,7 @@ import EventListener from '../EventListener';
 import {Cell, CellProps, Navigation} from './components';
 import {measureColumn, getPrevAndCurrentColumns} from './utilities';
 
-import {DataTableState, SortDirection} from './types';
+import {DataTableState, SortDirection, VerticalAlign} from './types';
 import styles from './DataTable.scss';
 
 type CombinedProps = Props & WithAppProviderProps;
@@ -31,6 +31,10 @@ export interface Props {
    * @default false
    */
   truncate?: boolean;
+  /** Vertical alignment of content in the cells.
+   * @default 'top'
+   */
+  verticalAlign?: VerticalAlign;
   /** Content centered in the full width cell of the table footer row. */
   footerContent?: TableData;
   /** List of booleans, which maps to whether sorting is enabled or not for each column. Defaults to false for all columns.  */
@@ -250,6 +254,7 @@ class DataTable extends React.PureComponent<CombinedProps, DataTableState> {
       columnContentTypes,
       defaultSortDirection,
       initialSortColumnIndex = 0,
+      verticalAlign,
     } = this.props;
 
     const {
@@ -283,13 +288,14 @@ class DataTable extends React.PureComponent<CombinedProps, DataTableState> {
         firstColumn={headingIndex === 0}
         truncate={truncate}
         {...sortableHeadingProps}
+        verticalAlign={verticalAlign}
       />
     );
   };
 
   private renderTotals = (total: TableData, index: number) => {
     const id = `totals-cell-${index}`;
-    const {truncate = false} = this.props;
+    const {truncate = false, verticalAlign} = this.props;
 
     let content;
     let contentType;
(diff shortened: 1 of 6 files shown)


src/components/DataTable/DataTable.tsx
```diff
@@ -317,7 +319,7 @@ class DataTable extends React.PureComponent<CombinedProps, DataTableState> {
 
   private defaultRenderRow = (row: TableData[], index: number) => {
     const className = classNames(styles.TableRow);
-    const {columnContentTypes, truncate = false} = this.props;
+    const {columnContentTypes, truncate = false, alignment} = this.props;
```
chloerice:
```suggestion
    const {columnContentTypes, truncate = false, alignment = top} = this.props;
```

Let's either give this a default value, or in cell only give it an alignment class if `alignment` is set.

src/components/DataTable/components/Cell/Cell.tsx
```diff
@@ -44,6 +46,7 @@ function Cell({
   const numeric = contentType === 'numeric';
   const className = classNames(
     styles.Cell,
+    styles[`Cell-${variationName('alignment', alignment)}`],
```
chloerice: I'm unfamiliar with how `variationName` works when the variation passed is undefined. Should this be conditional on `alignment` being truthy so it's not given a class name when there's no value set?

src/components/DataTable/tests/DataTable.test.tsx
```diff
@@ -200,6 +200,67 @@ describe('<DataTable />', () => {
...
+  describe('alignment', () => {
+    it('defaults to undefined', () => {
+      const dataTable = mountWithAppProvider(<DataTable {...defaultProps} />);
+
+      const cells = dataTable.find(Cell);
+
+      cells.forEach((cell) => expect(cell.prop('alignment')).toBeUndefined());
+    });
+
+    it('does not pass the value to headers', () => {
```
chloerice: Let's set `alignment` on `header` and `total` cells as well.

ry5n:
> > Should this also apply to the totals? Or headers?
>
> I think so, for consistencies sake. It might be harder to read if some rows are wrapping and centered while the headings are wrapping and top aligned for instance. @ry5n what do you think?

Good call, especially since this is a top-level component prop 👍 

This keeps it simple while giving us the option for more granular alignment control on the row and cell subcomponents in the future.

ry5n:
> I think alignment is fine here, but if you want to be more explicit you could call the prop verticalAlign and have the union type be the main values that the CSS property takes ('top' | 'middle' | 'bottom' | 'baseline') so that people understand that it's not a flex property, it only applies to inline and inline block elements.

+1 for `verticalAlign`

src/components/DataTable/components/Cell/Cell.tsx
Mangara: It can never be undefined - it gets a default value at the start of the function.

src/components/DataTable/DataTable.tsx
Mangara: The default value is currently set by `Cell`, since it was providing this behaviour before. Do you think we should move it up to `DataTable`? In that case, would it be better if the prop was required? (Iit doesn't seem like we do that?)

Mangara:
> I think so, for consistencies sake. It might be harder to read if some rows are wrapping and centered while the headings are wrapping and top aligned for instance.

👍 

> I'm not sure how much an additional class name affects bundle size (unfortunately, we can't see how greatly the change affects the web bundle from within the shrink-ray check in this repo). If you've noticed it's pretty significant at the page level, I'd say this would be a better way to go.

This was just idle speculation. If it's not a red flag for you, It's probably fine to leave it as-is. This is the way `Stack` works as well.

> I think alignment is fine here, but if you want to be more explicit you could call the prop verticalAlign and have the union type be the main values that the CSS property takes ('top' | 'middle' | 'bottom' | 'baseline') so that people understand that it's not a flex property, it only applies to inline and inline block elements.

👍 

> Let's add an example for this to DataTable/README.md

Looking at the other examples, these are mostly for tables that serve different use cases, not for cosmetic adjustments, like `truncate`. This feels more like a cosmetic adjustment to me. Do you think there's value in showing an example of a "Table with center-aligned rows"?

chloerice:
> Looking at the other examples, these are mostly for tables that serve different use cases, not for cosmetic adjustments, like truncate. This feels more like a cosmetic adjustment to me. Do you think there's value in showing an example of a "Table with center-aligned rows"?

Since the value for union typed props can be toggled through the props list similar to a boolean prop, I think we can skip adding an example 👍.

PR #1796, opened 2019-07-09
[v4] Update version 4 with master

AndrewMusgrave:
### What

Merging master into version 4.

The *few* commits (needed a few merges to resolve ongoing conflicts) has the changes to fix the build/errors after conflicts were resolved.

### How to 🎩

* I usually like to compare v4...this_branch -> master...this_branch but everyone has their own methods.
* tests 💯 ?
* percy ✅ ?


src/components/AppProvider/utilities/createAppProviderContext/createAppProviderContext.ts
```diff
@@ -1,75 +0,0 @@
-import createApp, {
```
AndrewMusgrave: This was added back because of a content change, that I added to `createAppBridge` in `src/utilities/app-bridge`

src/components/ChoiceList/tests/ChoiceList.test.tsx
```diff
@@ -435,8 +435,8 @@ describe('<ChoiceList />', () => {
 
   describe('disabled', () => {
     it('disables choices', () => {
-      const choiceElements = shallowWithAppProvider(
-        <ChoiceList selected={[]} choices={choices} disabled />,
+      const choiceElements = mountWithAppProvider(
+        <ChoiceList selected={[]} choices={choices} disabled title="title" />,
```
AndrewMusgrave: title is required now

src/components/DataTable/components/Cell/tests/Cell.test.tsx
```diff
@@ -125,34 +119,32 @@ describe('<Cell />', () => {
...
       });
 
       it('renders an Icon with an accessibility label indicating the next sort direction is descending', () => {
-        const cell = mountWithAppProvider(
+        const cell = mountWithTable(
           <Cell header firstColumn sortable sorted sortDirection="ascending" />,
         );
 
-        const expectedLabel = cell
-          .instance()
```
AndrewMusgrave: We can't access context like this anymore

src/components/DataTable/components/Cell/tests/Cell.test.tsx
```diff
@@ -125,34 +119,32 @@ describe('<Cell />', () => {
...
-          .context.polaris.intl.translate(
-            'Polaris.DataTable.sortAccessibilityLabel',
-            {
-              direction: 'descending',
-            },
-          );
+        const expectedLabel = cell.app.intl.translate(
```
AndrewMusgrave: I added this to the testing utilities previously so we can access the translations

src/components/DataTable/components/Cell/tests/Cell.test.tsx
```diff
@@ -261,3 +251,14 @@ describe('<Cell />', () => {
     });
   });
 });
+
+function mountWithTable<P>(node: ReactElement) {
+  return mountWithAppProvider<P>(
```
AndrewMusgrave: Mounting just the `Cell` was invalid html which was causing an error to be thrown

src/components/Filters/Filters.tsx
```diff
@@ -194,7 +198,10 @@ class Filters extends React.Component<ComposedProps, State> {
           </button>
           <Collapsible id={collapsibleID} open={filterIsOpen}>
             <div className={styles.FilterNodeContainer}>
-              <Focus disabled={!filterIsOpen || !readyForFocus || !open}>
+              <Focus
+                disabled={!filterIsOpen || !readyForFocus || !open}
+                root={this.focusNode.current}
```
AndrewMusgrave: We require a focus node now since we removed `findDOMNode`

src/components/Filters/Filters.tsx
```diff
@@ -211,6 +218,11 @@ class Filters extends React.Component<ComposedProps, State> {
       </div>
     );
 
+    const filterResourceName = resourceName || {
```
AndrewMusgrave: resourceName can be undefined. Something to note for the future, maybe we should just make the english translation the default?

src/components/Filters/components/ConnectedFilterControl/components/Item/tests/Item.test.tsx
```diff
@@ -1,20 +1,21 @@
...
-import {mountWithAppProvider, trigger} from 'test-utilities';
+import {mountWithAppProvider, trigger} from 'test-utilities/legacy';
 import Item from '../Item';
 
 describe('<Item />', () => {
   it('handles focus', () => {
     const item = mountWithAppProvider(<Item />);
 
-    trigger(item.childAt(0), 'onFocus');
-    expect(item.state().focused).toBe(true);
```
AndrewMusgrave: We can't and shouldn't use state

AndrewMusgrave:
It'll be the last two commits now. I had to merge `version-4.0.0` into this branch to resolve some conflicts

src/components/DataTable/components/Cell/tests/Cell.test.tsx
```diff
@@ -57,12 +53,12 @@ describe('<Cell />', () => {
         />,
       );
 
-      expect(cell.prop('aria-sort')).toBe(sortDirection);
+      expect(cell.find('th').prop('aria-sort')).toBe(sortDirection);
```
BPScott: Should we search based upon the name of the react component instead of the dom element it renders as?

```
expect(cell.find(Cell).prop('aria-sort')).toBe(sortDirection);
```

src/components/DataTable/components/Cell/tests/Cell.test.tsx
BPScott: Reading some of the Slack history in shopify-web the proposed way of testing translations is to check for the english text rather than trying to hook into translations stuff.

I think that makes sense as given this current content if we were to goof and make the `sortAccessibilityLabel` translation empty then this test would still pass as you're basically comparing a variable to itself

src/components/DataTable/components/Cell/tests/Cell.test.tsx
AndrewMusgrave: We would be asserting that the prop we passed to the component has the prop we passed it (which will be true 100% of the time). Now we're asserting the prop is placed on the `th` which was the original intent of the test 😄

src/components/DataTable/components/Cell/tests/Cell.test.tsx
AndrewMusgrave: Makes sense to me, change these tests 👍

AndrewMusgrave:
Had conflicts with version 4 again so this branch needed another merge. So it's the last few commits now

AndrewMusgrave:
More conflicts so I had to do another merge ✌️

src/components/DataTable/components/Cell/tests/Cell.test.tsx
BPScott: Groovy 👍

PR #1810, opened 2019-07-10
[v4] Export TestProvider with defaults and fix typescript react issue

AndrewMusgrave:

Apps consuming polaris don't have a way to use our internal contexts

There's a types issue when reacting context

* Adding defaults to test provider
* Exporting it
* Fix type issues
* export theme provider
* export internal contexts :( see below
* adding a named export to rollup so we can export from `test-utilities/react-testing`
* move within content context to `src/utilities`
* optional strict mode for test provider and default props


src/index.ts
```diff
@@ -6,3 +6,5 @@ export * from './components';
 
 export {UnstyledLinkProps, LinkLikeComponent} from './utilities/link';
 export {createPolarisContext} from './utilities/create-polaris-context';
+
+export {TestProvider} from './test-utilities/react-testing';
```
AndrewMusgrave: Exporting this so consumers can use it

src/test-utilities/react-testing.tsx
```diff
@@ -60,14 +57,20 @@ function noop() {}
...
   children,
-  polaris,
-  themeProvider,
-  frame,
-  intl,
-  scrollLockManager,
-  stickyManager,
-  appBridge,
-  link,
+  themeProvider = createThemeContext(),
```
AndrewMusgrave: Adding default context so we don't have to expose internals

src/test-utilities/react-testing.tsx
```diff
@@ -107,11 +109,6 @@ export const mountWithContext = createMount<Options, Context>({
     appBridge,
     link,
   }) {
-    const polarisContextDefault = createPolarisContext();
```
AndrewMusgrave: Removing this since we don't use it anymore

src/utilities/app-bridge/context.tsx
```diff
@@ -1,6 +1,6 @@
-import {createContext} from 'react';
+import React from 'react';
```
AndrewMusgrave: React doesn't need to be in scope from a babel perspective but in consuming apps I was getting a typescript reference errors with react so I'm thinking this will fix it 🤞

BPScott: I wonder if this is because typescript expects React to be in scope in tsx files. As we never use any JSX would renaming these files to `.ts` fix it too?

config/rollup/index.js
```diff
@@ -50,7 +50,11 @@ module.exports = function createRollupConfig({entry, cssPath}) {
         exclude: 'node_modules/**',
         runtimeHelpers: true,
       }),
-      commonjs(),
+      commonjs({
+        namedExports: {
```
AndrewMusgrave: I had trouble exporting `TestProvider` because of an issue with `createMount`. Here I'm following the docs ([rollup docs](https://rollupjs.org/guide/en/#error-name-is-not-exported-by-module), [github rollup docs](https://github.com/rollup/rollup-plugin-commonjs#custom-named-exports))

src/components/UnstyledLink/UnstyledLink.tsx
```diff
@@ -23,4 +23,4 @@ function UnstyledLink(props: Props) {
   );
 }
 
-export default React.memo(UnstyledLink);
```
AndrewMusgrave: Unfortunately  `enzyme` is having trouble `.find(UnstyledLink)`. We're not going to have a noticeable performance difference from this so I'm removing it for now till we find the root cause/enzyme fixes it.

src/index.ts
```diff
@@ -6,5 +6,11 @@ export * from './components';
 
 export {UnstyledLinkProps, LinkLikeComponent} from './utilities/link';
 export {createPolarisContext} from './utilities/create-polaris-context';
+export {
```
AndrewMusgrave: These are used in web and we can't get around them for now :(

src/test-utilities/react-testing.tsx
```diff
@@ -49,7 +49,7 @@ type ReturnedContext = ComplexProviders &
   };
 type Options = DeepPartial<ComplexProviders> & Partial<SimpleProviders>;
 type Context = ReturnedContext;
-interface Props extends ReturnedContext {
+interface Props extends Partial<ReturnedContext> {
```
AndrewMusgrave: Exporting the provider, now with defaults so props aren't mandatory

src/components/WithinContentContext/context.tsx
```diff
@@ -1,5 +0,0 @@
-import React from 'react';
```
AndrewMusgrave: This was actually a move to `src/utilities`

src/test-utilities/react-testing.tsx
```diff
@@ -52,53 +49,62 @@ type ReturnedContext = ComplexProviders &
...
+  appBridge = null,
+  link = new Link(undefined),
   ...props
 }: Props) {
   const childWithProps =
     Object.keys(props).length > 0
       ? React.cloneElement(children, props)
       : children;
 
+  const Wrapper = strict? React.StrictMode : React.Fragment;
```
AndrewMusgrave: Going to have an optional strict mode so this utility can be used for enzyme tests as well

src/utilities/app-bridge/context.tsx
AndrewMusgrave: I thought that as well, but no 🎲

config/rollup/index.js
```diff
@@ -50,11 +50,7 @@ module.exports = function createRollupConfig({entry, cssPath}) {
         exclude: 'node_modules/**',
         runtimeHelpers: true,
       }),
-      commonjs({
```
AndrewMusgrave: Ended up removing this since it didn't do what I originally thought

src/test-utilities/TestProvider.tsx
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
```
AndrewMusgrave: Broke this apart from `test-utilities/react-testing` since it's not package specific/export issue

src/test-utilities/types.ts
```diff
@@ -0,0 +1,33 @@
+import {ClientApplication} from '@shopify/app-bridge';
```
AndrewMusgrave: Broke out shared types to prevent circular imports

AndrewMusgrave:
Should be ready for 👀 

#### `polaris-styleguide` (with a few breaking changes fixed locally)
* `yarn type-check` ✅ 
* `yarn test` ✅ 
* `yarn server` ✅ 
* quick tophat ✅ 

#### web
* types are good but still, need some works on fixes tests

AndrewMusgrave:
cc/ @alex-page @BPScott

src/test-utilities/TestProvider.tsx
BPScott: I'm not sure what you mean by this? I'm not clear on the value of having this living in its own file

AndrewMusgrave: Sorry I could have been more clear. This was the error
Screen Shot 2019-07-12 at 11 09 55 AM

Screen Shot 2019-07-12 at 11 09 50 AM

Eventually, I decided to move it into a separate file since `TestProvider` isn't test platform-specific. It can be used with `react-testing`, `enzyme`, `react-test-util`, etc.

src/index.ts
```diff
@@ -6,3 +6,11 @@ export * from './components';
...
+  ScrollLockManagerContext as __UNSAFE_SECRET_INTERNAL_SCROLL_LOCK_MANAGER_CONTEXT,
+} from './utilities/scroll-lock-manager';
+export {
+  WithinContentContext as __UNSAFE_SECRET_INTERNAL_WITHIN_CONTENT_CONTEXT,
```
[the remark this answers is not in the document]
AndrewMusgrave: We shouldn't really expose `contexts`. There's a private part of the app, and having this publicly available could harm us in the future. We shouldn't be promoting the use of it. From a design perspective if we decide that `WithinContext` is a variant that could be opted into we should probably expose a prop, rather than our internals. However keeping the export here will allow it to be used, if needed (which was by forked components in web).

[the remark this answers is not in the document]
AndrewMusgrave: As an alternative name we could do with `__UNSTABLE_INTERNAL_WITHIN_CONTENT_CONTEXT`. And we should mention that `_UNSTABLE` is to be used at the consumer's own risk and may receive breaking changes in minor versions

src/test-utilities/TestProvider.tsx
BPScott: Following pairing: This TestProvider shall be the building block of our testing strategy. It is a generic component that can be used by both enzyme and react-testing's mounting systems.

legacy.tsx shall be updated to use this instead of providing its own version.

createPolarisContext shall be removed from our public API and instead 3rd party test configs (e.g. web) should use this TestProvider component as part of their test configuration, for both enzyme and react-testing setups


Related: Can we have a more specific name for this? `PolarisTestContextProvider` or something like that? It's not very clear what the identifier "TestProvider" would do if you're in web's test setup files.

src/index.ts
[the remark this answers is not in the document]
BPScott: On second thoughts keeping this scarey means its' not strictly speaking public API and we can change it up in a minor version. Let's keep it this way;

src/test-utilities/TestProvider.tsx
AndrewMusgrave: > Related: Can we have a more specific name for this? PolarisTestContextProvider or something like that? It's not very clear what the identifier "TestProvider" would do if you're in web's test setup files.

Yup that's definitely possible, it's actually similar to how I'm importing it right now 

```
import {..., TestProvider as PolarisTestProvider ,...} from '@shopify/polaris'
```

src/components/ContextualSaveBar/tests/ContextualSaveBar.test.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import {mount} from 'enzyme';
 import {mountWithAppProvider} from 'test-utilities/legacy';
 import ContextualSaveBar from '../ContextualSaveBar';
-import {FrameContext, createFrameContext} from '../../Frame';
```
AndrewMusgrave: FrameProvider is inside `TestProvider` so we just need to pass our spies through

src/components/DropZone/components/FileUpload/tests/FileUpload.test.tsx
```diff
@@ -102,18 +102,4 @@ describe('<FileUpload />', () => {
     fileUpload.setProps({children: <FileUpload />});
     expect(fileUpload.find(TextStyle).text()).toBe('or drop files to upload');
   });
-
-  it('does not use default action title and hint when props are changed', () => {
```
AndrewMusgrave: This test failed in the conversion and I deleted it since we don't need it anymore. When this component was a class we stored this on state and this was a regression test however we don't follow that pattern in our functional component so 🔥

src/components/UnstyledLink/UnstyledLink.tsx
```diff
@@ -23,4 +23,4 @@ function UnstyledLink(props: Props) {
   );
 }
 
-export default UnstyledLink;
+export default React.memo(UnstyledLink);
```
AndrewMusgrave: I managed to get this working in web so I'm sneakily adding it back 😄

src/index.ts
```diff
@@ -5,12 +5,11 @@ export * from './types';
 export * from './components';
 
 export {UnstyledLinkProps, LinkLikeComponent} from './utilities/link';
-export {createPolarisContext} from './utilities/create-polaris-context';
```
AndrewMusgrave: Just as a quick note in a follow-up PR I can probably remove this entirely from polaris

AndrewMusgrave:
@BPScott 👌

src/test-utilities/TestProvider.tsx
BPScott: I think there's some value in not naming our exports so generically people feel compelled to alias them :)

src/index.ts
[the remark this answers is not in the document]
BPScott: (Totally missed your replies when I commented there, and it sounds like I ignored ya, sorry)

You point about not exposing contexts is a good one. Keeping this scary internal name now gives us a chance to come up with a proper design for this later.


Poking about with naming - unstable and unsafe suggest they might become stable later, but it sounds like we might want to go a different route with how we expose contexts. [React has __SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED](https://github.com/facebook/react/blob/9b0bd43550206e04bfe9ca695e5981eff0e2d03f/packages/react/src/React.js#L104), so perhaps drop that part of the prefix and have `__SECRET_INTERNAL_SCROLL_LOCK_MANAGER_CONTEXT` and `__SECRET_INTERNAL_WITHIN_CONTENT_CONTEXT`?

src/test-utilities/legacy.tsx
```diff
@@ -174,40 +160,13 @@ export function polarisContextReactWrapper<P, S>(
...
-                  </FrameContext.Provider>
-                </LinkContext.Provider>
-              </AppBridgeContext.Provider>
-            </ThemeProviderContext.Provider>
-          </StickyManagerContext.Provider>
-        </ScrollLockManagerContext.Provider>
-      </I18nContext.Provider>
-    );
-  }
+  const appBridge: any = app.appBridge;
```
BPScott: I also just spotted that the appBridge typing in tests is kinda wonky. I'll try and come up with a plan once this and #1828 are merged.

BPScott:
Something that's been gnawing at me over the past two days: The more I think, the more I'm unsure of adding FrameContext into the TestProvider. Without it the TestProvider is an AppProvider but with the ability to set not-usually-accessible items like the sticky and scroll managers (this helps us find a home for TestProvider too, as we could place it in src/AppProvider (this needs a little more thought)). That symmetry feels very useful as it makes it easy to describe what TestProvider does - it's the same as AppProvider, but for use in tests so you can set mocks. It also makes the "mountWithAppProvider" function name a bit more true as it only adds the contexts provided by AppProvider and nothing else. Adding a value for FrameContext into the mix muddies the water and I'm not convinced the extra cognitive overhead to understand TestProvider as being a bit different from AppProvider is worth the slight extra convenience of all our contexts in one place - especially as 90% of the time you don't care about defining a value for FrameContext, you only care about appProvider based things.

For internal testing, we could leverage react-testing's about-to-land [ability to extend configs](https://github.com/Shopify/quilt/pull/788). to create a `mountWithFrame` that extends from `mountWithContext` so we can still have easy config of the Frame context.

What do you think to leaving FrameContext out of TestProvider, renaming TestProvider to TestAppProvider and undoing the changes to ContextualSaveBar, Loading and Toast (eventually those files would be updated to use a mountWithFrame as described above)

src/test-utilities/legacy.tsx
AndrewMusgrave: Yup app bridge is using `{}` in tests as a mock but isn't actually a valid type 😬

AndrewMusgrave:
I would say, the ideal situation for us is to not have to worry about what `mount` we are using and whether it contains context that might not necessarily be relevant but required to what we're testing. I think the ability to extend `createMount` is a cool feature but less needed for us because we have the ability to modify our `createMount` function.

In `polaris` we might not use components with `frame` context a lot however web, embedded apps, etc will since they're creating entire pages.

AndrewMusgrave:
Although if we did decide to have multiple mounts in polaris, I would be ok with that. I still think it's nice to have a `all in mount` mounting function for consumers though so they don't have to worry about what's going on in polaris since it's not really their concern how we handle contexts and which components use it.

BPScott:
> I would say, the ideal situation for us is to not have to worry about what mount we are using and whether it contains context that might not necessarily be relevant but required to what we're testing

To me that feels like many tests will have additional overhead that isn't needed. It won't have any noticeable impact on performance but it makes it a little hard to understand what is relevant and needed for a given test.

Ok, I'm happy to merge this as is as I've got some follow-up work that needs it, and we can continue to go back and forth on this later :)

PR #1811, opened 2019-07-10
[SkeletonPage] render Loading by default

patsissons:

Because `<SkeletonPage />` is typically used as a _loading_ component, it stands to reason that rendering a `<Loading />` component would also be desired within a `<Frame />` so that the page UX includes loading feedback.

To resolve this, I am adding a new optional `loading` prop to `SkeletonPage` that defaults to `false` so that `<Loading />` is only included in a skeleton page when opted in.

## 

### How to 🎩

<details>

```jsx
import {Frame, SkeletonPage} from '@shopify/polaris';

interface State {}

  render() {
    return (
      <Frame>
        <SkeletonPage title="Loading UX" loading>
          <Layout sectioned>
            <SkeletonBodyText />
          </Layout>
        </SkeletonPage>
      </Frame>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/SkeletonPage/README.md
+++ b/src/components/SkeletonPage/README.md
@@ -150,6 +150,48 @@ Use this component to compose a loading version of a page where the page title a
 </SkeletonPage>
 ```
 
+### Page inside a Frame
+
+Use this component to compose a loading version of a page that includes a frame loading bar.
+
+```jsx
+<Frame>
+  <SkeletonPage title="Products" primaryAction secondaryActions={2} loading>
+    <Layout>
+      <Layout.Section>
+        <Card sectioned>
+          <SkeletonBodyText />
+        </Card>
+        <Card sectioned title="Images">
+          <SkeletonBodyText />
+        </Card>
+        <Card sectioned title="Variants">
+          <SkeletonBodyText />
+        </Card>
+      </Layout.Section>
+      <Layout.Section secondary>
+        <Card title="Sales channels">
+          <Card.Section>
+            <SkeletonBodyText lines={2} />
+          </Card.Section>
+          <Card.Section>
+            <SkeletonBodyText lines={1} />
+          </Card.Section>
+        </Card>
+        <Card title="Organization" subdued>
+          <Card.Section>
+            <SkeletonBodyText lines={2} />
+          </Card.Section>
+          <Card.Section>
+            <SkeletonBodyText lines={2} />
+          </Card.Section>
+        </Card>
+      </Layout.Section>
+    </Layout>
+  </SkeletonPage>
+</Frame>
+```
+
 ---
 
 ## Related components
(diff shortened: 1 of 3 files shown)


BPScott:
Passing false into a prop `loading={false}` to opt into non-default behavior feels weird as it goes against convention for [boolean attributes on html elements](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes). When defining interfaces to functions we tend to say "arguments should be positive" to avoid double negative weirdness (compare `enabled=true` to `disabled=false`) but I think we should make an exception for props due to their egronomic similarity to html boolean attributes.

If having a loading bar should be the default then what do you think to turning it off with `<SkeletonPage noLoading>` (or some other negative name) instead of `<SkeletonPage loading={false}>`

AndrewMusgrave:
What if we defaulted `loading` to false to have this behavior opt-in. Or is this a pattern we'll almost always want on pages? cc/ @mirualves for a design perspective

patsissons:
totally agree that `loading={false}` is awkward. I would be totally down to use `noLoading` instead for opting out.

AndrewMusgrave:
cc/ @mirualves 😄

BPScott:
Yo @patsissons, this kinda fell down the back of the sofa, could you either close or rebase atop of master?

patsissons:
oof this definitely fell behind the sofa. I'll rebase and cleanup in a sec.

patsissons:
OK, rebased and updated to be an opt-in feature.

danrosenthal:
Hey @patsissons, this PR is targeting a stale, non-master branch. Should it be?

I'd recommend requesting review from a couple people with context, and removing the draft status, along with updating the target branch, in order to get eyes on this.

patsissons:
@danrosenthal we're now back up to date on this. LMK if you folks want to move forward on this change.

danrosenthal (approved):
I'm good with this change. Going to ping a couple others on it just for a sanity check.

dleroux:
@patsissons is there a reason why we wouldn't want to render this <Loading /> every time <Skeletonpage /> is being rendered?

mirualves:
Late to the party! (sorry, @AndrewMusgrave). I think it makes sense to have the loading and the skeleton page together at all times instead of sometimes having it, and sometimes not.

src/components/SkeletonPage/README.md
```diff
@@ -150,6 +150,48 @@ Use this component to compose a loading version of a page where the page title a
 </SkeletonPage>
 ```
 
+### Page inside a Frame
```
chloerice:
```suggestion
### Page loading inside of Frame
```

chloerice (approved):
This makes sense to me, though I agree with Miru that it might make sense to always show the loading bar 🤔

dleroux:
@patsissons are you still planning on moving with this? Code looks good but if there's no reason not to show the loader every time the page loads it would probably make sense to add it.

alex-page:
Closing due to inactivity

PR #1813, opened 2019-07-10
Hooky withAppProvider

BPScott:

Lots of indentation looks silly and it's a pain in React devtools, lets use hooks instead of render props in withAppProvider so we have a single wrapping component instead of 5.

Uses hooks in withAppProvider's wrapping component to save on indentation and nesting of components in the React devtools.

This didn't play nice with how we overrode isServer in
server-side-rendering tests due to funky mocking. Splitting the mocks
out into separate files fixes this. Strictly speaking this only caused
problems in Avatar as it is a functional component, but I figured lets
preemptively fix ColorPicker's test too

### How to 🎩

Confirm various AppProvider functionality still works:

- setting translations
- setting links
- settings themes
- scrolling / sticky managing

### DIFF
--- a/src/utilities/with-app-provider.tsx
+++ b/src/utilities/with-app-provider.tsx
@@ -1,16 +1,17 @@
-import React, {useContext} from 'react';
+import React from 'react';
 import hoistStatics from 'hoist-non-react-statics';
 import {ClientApplication} from '@shopify/app-bridge';
 import {PolarisContext} from '../components/types';
-import {I18n, I18nContext} from './i18n';
-import {Link, LinkContext} from './link';
+import {I18n, useI18n} from './i18n';
+import {Link, useLink} from './link';
+import {ScrollLockManager, useScrollLockManager} from './scroll-lock-manager';
+import {ThemeProviderContextType, useTheme} from './theme';
 import {
-  ScrollLockManager,
-  ScrollLockManagerContext,
-} from './scroll-lock-manager';
-import {ThemeProviderContextType, ThemeProviderContext} from './theme';
-import {StickyManager, StickyManagerContext} from './sticky-manager';
-import {AppBridgeContext} from './app-bridge';
+  StickyManager,
+  StickyManagerContext,
+  useStickyManager,
+} from './sticky-manager';
+import {useAppBridge} from './app-bridge';
 
 export type ReactComponent<P, C> =
   | React.ComponentClass<P> & C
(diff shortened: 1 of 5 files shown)


src/utilities/with-app-provider.tsx
```diff
@@ -52,66 +52,32 @@ export function withAppProvider<OwnProps>({withinScrollable}: Options = {}) {
...
+      if (!intl || !scrollLockManager || !stickyManager || !link || !theme) {
+        throw new Error(
+          `The <AppProvider> component is required as of v2.0 of Polaris React. See https://polaris.shopify.com/components/structure/app-provider for implementation instructions.`,
         );
       }
-    }
+
+      return <WrappedComponent {...props as any} polaris={polarisContext} />;
+    };
+    WithProvider.contextTypes = WrappedComponent.contextTypes;
```
BPScott: This is just for [legacy context](https://reactjs.org/docs/legacy-context.html) which we no longer us internally but [external users might](https://github.com/Shopify/web/search?q=withAppProvider&unscoped_q=withAppProvider)

BPScott:
Percy changes also occur on the version-4.0.0 branch, and are unrelated to this change

src/utilities/with-app-provider.tsx
```diff
@@ -52,66 +53,32 @@ export function withAppProvider<OwnProps>({withinScrollable}: Options = {}) {
   return function addProvider<C>(
     WrappedComponent: ReactComponent<OwnProps & WithAppProviderProps, C>,
   ): React.ComponentClass<OwnProps> & C {
-    // eslint-disable-next-line react/prefer-stateless-function
-    class WithProvider extends React.Component<OwnProps, never> {
-      static contextTypes = WrappedComponent.contextTypes;
+    const WithProvider: React.FunctionComponent = (props: OwnProps) => {
```
AndrewMusgrave: Will this be inferred or do we need to explicitly set it?

BPScott: `React.FunctionComponent` adds a [few more things](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L503) over a regular function. Notably adding children to the props, and adding some of the static things you can hang off of a component like `WithProvider.contextTypes. In 99% of the time this distinction doesn't matter as we don't use any of those properties but I'm being paranoid here.

Extending from React.Component did [the same sort of thing](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L519)

PR #1815, opened 2019-07-10
[Shopify/polaris-bot] Update post-release web upgrade PR description

chloerice:

The description for the PR automatically opened in `web` after a release doesn't include tophatting instructions. Specifically, the instructions for tophatting with production assets, which is required when upgrading key dependencies like `polaris-react`.

Adds the pre-deploy self-checklist and prod asset tophatting instruction link to the template.

##

### DIFF
--- a/scripts/new-version-pr-generator.js
+++ b/scripts/new-version-pr-generator.js
@@ -46,6 +46,24 @@ See what’s new: https://github.com/Shopify/polaris-react/releases/tag/${versio
 
 ---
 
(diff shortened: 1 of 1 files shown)


kaelig:
This script also opens PRs on Shopify/polaris-styleguide. Can you confirm these instructions only will appear in PRs open in Shopify/web?

chloerice:
> This script also opens PRs on Shopify/polaris-styleguide. Can you confirm these instructions only will appear in PRs open in Shopify/web?

@kaelig Not if the same script is used for both, no. Will make conditional.

PR #1821, opened 2019-07-11
[Skip to content] Fix tabIndex issue

### LINKED ISSUE #1816: v3.19.0 interop regression in web
After deploying v3.19.0 to web we had to revert it in https://github.com/Shopify/web/pull/15881 thanks to issues with the Rails admin interop stuff.

Various rails-based popovers failed to honor what you click on. You can see this in action:

In web: `git checkout 7b9097da` (the last commit with polaris v3.19.0, before the revert)
Run `yarn run sk clean` to make sure you've got no old build remnants lieing around
Edit the `plugins.vendor` section of sewing-kit.config.ts, commenting out the "@shopify/polaris" line so that the polaris tooling doesn't get included in the vendor bundle.

Run `dev server` then go to the existing product and add some "tag1" and "tag2" as tags.
Then go to the add a new product page and try and add one of those new tags to the product by typing "t" in the tags box, and try to click on the "tags1" item in the popover that appears.

The popover closes without adding the tag.

Expected behaviour: the tag is selected.

---

I've spent the day trying to work out what causes this through slow elimination I've gotten quite close but don't have a smoking gun. Here's how to test a given polaris-react commit within web, so we can work out what commit causes the problem.

- Stop your web `dev` server.
- In polaris-react: Run`git checkout f2389d0d` (/whatever commit you want to test) then do `yarn run build-consumer web` to push this into web's node_modules.
- In web: Run `yarn run sk clean && dev server` and follow the above steps to try and add a tag and see if it works.

I've confirmed f2389d0d works as expected for me and there's only commits from two PRs between that and v3.19.0 left https://github.com/Shopify/polaris-react/compare/f2389d0d7...v3.19.0

Happy hunting

dleroux:

In this PR: https://github.com/Shopify/polaris-react/pull/1785 We added focus to the `<main>` element by adding tabIndex={-1}. This broke the Polaris-rails popover. In rails clicking the popover sends the target and hides the popover if the target isn't inside a popover. Adding the tabIndex on the main element made it the target, therefore always closed the popover.

Fixes #1816

Adding an `<a>` tag at the beginning of the content and focusing it instead.

## 

### How to 🎩

To test this properly it is best to do it web.

1. Start web
2. In polaris-react: `yarn build-consumer web`.
3. Tab to the skip to content link and click it. You should skip to the content. Also, use voice over to ensure the correct text is being read.

<details>

```jsx

interface State {}

  render() {
    return (
      <Page title="Playground">
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Frame/Frame.tsx
+++ b/src/components/Frame/Frame.tsx
@@ -50,6 +50,7 @@ export interface State {
 
 export const GLOBAL_RIBBON_CUSTOM_PROPERTY = '--global-ribbon-height';
 export const APP_FRAME_MAIN = 'AppFrameMain';
+export const APP_FRAME_MAIN_ANCHOR_TARGET = 'AppFrameMainContent';
 const APP_FRAME_NAV = 'AppFrameNav';
 const APP_FRAME_TOP_BAR = 'AppFrameTopBar';
 const APP_FRAME_LOADING_BAR = 'AppFrameLoadingBar';
@@ -72,7 +73,7 @@ class Frame extends React.PureComponent<CombinedProps, State> {
 
   private globalRibbonContainer: HTMLDivElement | null = null;
 
-  private mainContentNode = React.createRef<HTMLDivElement>();
+  private skipoToMainContentTargetNode = React.createRef<HTMLAnchorElement>();
 
   getChildContext(): FrameContext {
     return {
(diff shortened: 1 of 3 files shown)


src/components/Frame/Frame.tsx
```diff
@@ -246,6 +247,15 @@ class Frame extends React.PureComponent<CombinedProps, State> {
         />
       ) : null;
 
+    const skipToMainContentTarget = (
+      // eslint-disable-next-line jsx-a11y/anchor-is-valid
+      <a
+        id={APP_FRAME_MAIN_ANCHOR_TARGET}
+        ref={this.skipoToMainContentTargetNode}
```
jgodson: `skipoToMainContentTargetNode` extra `o` in here

PR #1823, opened 2019-07-12
[v4] Change matcher layout

BPScott:

Sorry, I missed the PR where we added react-testing.

I noticed a few things that I dislike:

- ~`The test-utilities folder should only contain things we want to share with the world to make their test harnesses easier to work with Polaris. Matchers should be private to only our tests, we don't want anybody else to use them, so they shouldn't live in the src folder.~ The test-utilities folder should only contain things we need to import in our test files, matchers get imported by each-test.ts, so they should live in the tests folder
- Our own custom matchers are cool but I'd like to try and stick with using the ones `@shopify/react-testing` provides until it is clearly untenable. `toBeDisabled()` can be trivially implemented as `.toHaveReactProps({disabled: true})`, I don't think saving those 20 chars is worth the indirection.

In the future should we add matchers, they should live in the tests folder, not the src folder.

- Moves the registration of  `@shopify/react-testing`'s matchers directly into the `each-test.ts` file. 
- Moves our custom matchers into the `tests` folder.

### How to 🎩
 Check tests pass

### DIFF
--- a/tests/each-test.ts
+++ b/tests/each-test.ts
@@ -1,5 +1,6 @@
 import {destroyAll} from '@shopify/react-testing';
-import '../src/test-utilities/matchers';
+import '@shopify/react-testing/matchers';
+import './matchers';
 
 afterEach(() => {
   destroyAll();
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     "@percy/storybook": "^3.0.2",
     "@shopify/jest-dom-mocks": "^2.1.1",
     "@shopify/js-uploader": "github:shopify/js-uploader",
-    "@shopify/react-testing": "^1.5.3",
+    "@shopify/react-testing": "^1.7.0",
     "@shopify/sewing-kit": "0.88.0",
     "@storybook/addon-a11y": "^5.1.9",
     "@storybook/addon-actions": "^5.1.9",
--- a/src/test-utilities/matchers/index.ts
+++ b/src/test-utilities/matchers/index.ts
@@ -1,2 +1,0 @@
-import '@shopify/react-testing/matchers';
-import './react';
--- a/src/test-utilities/matchers/types.ts
+++ b/src/test-utilities/matchers/types.ts
@@ -1,3 +1,0 @@
-import {Root, Element} from '@shopify/react-testing';
-
-export type Node<Props> = Root<Props> | Element<Props>;
--- a/tests/matchers/index.ts
+++ b/tests/matchers/index.ts
@@ -1,0 +1,11 @@
+import {toBeDisabled} from './props';
+
+declare global {
+  namespace jest {
+    interface Matchers<R> {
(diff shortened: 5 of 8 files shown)


AndrewMusgrave:
> The test-utilities folder should only contain things we want to share with the world to make their test harnesses easier to work with Polaris. Matchers should be private to only our tests, we don't want anybody else to use them, so they shouldn't live in the src folder.

`src/test-utilities` has always been where we keep all of our typed source code, regardless of if we want to expose it to the public or not. `src/test-utilities` if for exactly that purpose, our test utilities not necessarily test utilities we want to expose. That's not too say we can't move them out, it's just our current structure won't allow for that.

> Our own custom matchers are cool but I'd like to try and stick with using the ones @shopify/react-testing provides until it is clearly untenable. toBeDisabled() can be trivially implemented as .toHaveReactProps({disabled: true}), I don't think saving those 20 chars is worth the indirection.

Generally, we want to use the most specific matchers we can. If `toBeDisabled` exists it should be used instead of `toHaveReactProps`. Should we have created it? Who knows 🤷‍♂. It looks the largest [matcher community has created one as well](https://github.com/FormidableLabs/enzyme-matchers/blob/master/packages/enzyme-matchers/src/assertions/toBeDisabled.js). That's not to necessarily say we need it however as I mentioned in the original pr this matcher was created as an example and template for creating matchers in the future. And I don't quite see what we gain from removing it.

> In the future should we add matchers, they should live in the tests folder, not the src folder.

`<rootDir>/tests` and all `tests` folders are generally where we keep tests, not utilities.

BPScott:
Following on from our chat earlier, I've came around on this.

The `tests` folder pulls triple duty - as where we store our test config (setup.ts and each-test.ts) - it's main function IMO, and as a folder for testing our scripts - which is just because there's nowhere better to put these tests.
It also stores our legacy and react-testing files that provide mount functions. But the only reason why they live here rather than in `tests/` is because all the dots in `import X from '../../../tests/utilities` was annoying and this was the least convoluted method to let you import from a non-relative path.


I've moved our matchers into the tests folder as they are only used within the each-test.ts file - this is the same as what other projects like web do, though I've tweaked the layout slightly - instead of extending jest's within every file containing a matchers, I'm following the style of [react-testing's own matchers](https://github.com/Shopify/quilt/tree/master/packages/react-testing/src/matchers) folder where each file exports a file, and its the index file's job to add them all to Jest.
I've kept toBeDisabled as an example of a matcher around for now.

tests/each-test.ts
```diff
@@ -1,5 +1,6 @@
+import '@shopify/react-testing/matchers';
```
AndrewMusgrave: Linting nitpick

```suggestion
import {destroyAll} from '@shopify/react-testing';
import '@shopify/react-testing/matchers';
import './matchers';
```

PR #1824, opened 2019-07-12
[ChoiceList] Connecting the ChoiceList error the input with errors via describedByProp

dleroux:

Fixes https://github.com/Shopify/polaris-react/issues/1589

1. add `ariaDescribedBy` prop to Checkbox and RadioButton
2. remove `aria-describedby` from the ChoiceList fieldset 
3. updated the ChoiceDescriptor with a `describedByError` key to indicate whether to link the Error to the input or not.
4. added some tests

## 

### How to 🎩

Using the playground, ensure the radio buttons `aria-describedby` attribute matches the ID of the error field.

<details>

```jsx
import {Page, ChoiceList} from '../src';

interface State {}

  render() {
    return (
      <Page title="Playground">
        <ChoiceListExample />
      </Page>
    );
  }
}

class ChoiceListExample extends React.Component {
  state = {
    selected: ['hidden'],
  };

  render() {
    const {selected} = this.state;
    return (
      <ChoiceList
        title="Company name"
        choices={[
          {label: 'Hidden', value: 'hidden', describedByError: true},
          {label: 'Optional', value: 'optional'},
          {label: 'Required', value: 'required'},
        ]}
        selected={selected}
        onChange={this.handleChange}
        error="Company name cannot be hidden at this time"
      />
    );
  }

  handleChange = (value) => {
    this.setState({selected: value});
  };
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -5,12 +5,15 @@ import {createUniqueIDFactory} from '@shopify/javascript-utilities/other';
 
 import {withAppProvider, WithAppProviderProps} from '../AppProvider';
 import Choice, {helpTextID} from '../Choice';
+import {errorTextID} from '../InlineError';
 import Icon from '../Icon';
 import {Error, Key} from '../../types';
 
 import styles from './Checkbox.scss';
 
 export interface BaseProps {
+  /** Indicates the ID of the element that describes the checkbox*/
+  ariaDescribedBy?: string;
   /** Label for the checkbox */
   label: React.ReactNode;
   /** Visually hide the label */
@@ -65,6 +68,7 @@ class Checkbox extends React.PureComponent<CombinedProps, never> {
 
   render() {
     const {
+      ariaDescribedBy: ariaDescribedByProp,
       id = getUniqueID(),
       label,
       labelHidden,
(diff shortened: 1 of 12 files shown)


src/components/Checkbox/Checkbox.tsx
```diff
@@ -78,12 +82,15 @@ class Checkbox extends React.PureComponent<CombinedProps, never> {
       value,
     } = this.props;
     const describedBy: string[] = [];
-    if (error) {
-      describedBy.push(`${id}Error`);
+    if (error && typeof error !== 'boolean') {
```
dleroux: If the type is boolean we don't want an area-describeby attribute

src/components/ChoiceList/ChoiceList.tsx
```diff
@@ -19,6 +20,8 @@ export interface ChoiceDescriptor {
   disabled?: boolean;
   /** Additional text to aide in use */
   helpText?: React.ReactNode;
+  /** Indicates that the choice is aria-describedBy the error message*/
+  describedByError?: boolean;
```
dleroux: The reason I went with `describedByError: boolean` as oppose to `describedBy:string` is because ChoiceList is the one generating the error and the id. So I thought to add `describedByError`  to the descriptor was more clear. The choice has 1 error and this choice is described by that error.

PR #1825, opened 2019-07-12
[Indicator] Remove default pulse and add tone

kyledurand:

This change addresses an addition I had to make in web

This change is removing the default to pulse option from the component (could be breaking in some instances but this is much more logical to me). 

It's also adding two other variations to the component for neutral and negative tones.

### How to 🎩

<details>

```jsx
import {Button, Card, Layout, Page, Indicator} from '../src';

interface State {}

  render() {
    return (
      <Page title="Playground">
        <Layout>
          <Layout.Section>
            <Card sectioned>
              <div style={{position: 'relative'}}>
                <div>
                  <Indicator tone="negative" pulse />
                  Bad tings
                </div>
              </div>
            </Card>
          </Layout.Section>
        </Layout>
      </Page>
    );
  }
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Indicator/Indicator.scss
+++ b/src/components/Indicator/Indicator.scss
@@ -8,7 +8,6 @@ $base-position: rem(-4px);
   &::after {
     content: '';
     position: absolute;
-    background-color: color('teal');
     right: $base-position;
     top: $base-position;
     width: $size;
@@ -17,15 +16,36 @@ $base-position: rem(-4px);
   }
 }
 
-.pulseIndicator::before {
+.Indicator-pulse::before {
   z-index: 1;
   animation: bounce 5s ease infinite;
 }
 
-.pulseIndicator::after {
+.Indicator-pulse::after {
   right: $base-position;
   top: $base-position;
   animation: pulse 5s ease infinite;
+}
+
+.Indicator-positive {
+  &::before,
+  &::after {
+    background-color: color('teal');
+  }
+}
+
+.Indicator-neutral {
+  &::before,
+  &::after {
+    background-color: color('sky', 'dark');
+  }
+}
+
+.Indicator-negative {
+  &::before,
+  &::after {
+    background-color: color('red');
+  }
 }
 
 $pip-animation: (
--- a/src/components/Indicator/Indicator.tsx
+++ b/src/components/Indicator/Indicator.tsx
@@ -2,14 +2,20 @@ import * as React from 'react';
 import {classNames} from '@shopify/css-utilities';
 import styles from './Indicator.scss';
 
+type Tone = 'positive' | 'neutral' | 'negative';
+
 export interface Props {
+  // Defaults to positive
+  tone?: Tone;
+  // Sets an animated transparent background
   pulse?: boolean;
 }
 
-export default function Indicator({pulse = true}: Props) {
+export default function Indicator({pulse, tone = 'positive'}: Props) {
   const className = classNames(
     styles.Indicator,
-    pulse && styles.pulseIndicator,
+    pulse && styles['Indicator-pulse'],
+    tone && styles[`Indicator-${tone}`],
   );
 
   return <span className={className} />;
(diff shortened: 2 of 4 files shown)


src/components/Indicator/Indicator.scss
```diff
@@ -17,17 +16,38 @@ $base-position: rem(-4px);
   }
 }
 
-.pulseIndicator::before {
+.Indicator-pulse::before {
```
dleroux: Since the filename is called Indicator, these can all just be: `.pulse`, `.positive`, `.neutral`, `.negative. 

This will end up generating: `.Polaris-Indicator--pulse` instead of  `.Polaris-Indicator__Indicator--pulse`.

src/components/Indicator/Indicator.tsx
```diff
@@ -2,14 +2,20 @@ import * as React from 'react';
...
+type Tone = 'positive' | 'neutral' | 'negative';
+
 export interface Props {
+  // Defaults to positive
+  tone?: Tone;
+  // Sets an animated transparent background
   pulse?: boolean;
 }
 
-export default function Indicator({pulse = true}: Props) {
```
dleroux: Was there a requirement for removing the default of true? Just thinking default of true makes sense based on the nature of the component and technically this could be an unexpected change for people using it. What do you think?

src/components/Indicator/Indicator.tsx
```diff
@@ -2,14 +2,20 @@ import * as React from 'react';
 import {classNames} from '@shopify/css-utilities';
 import styles from './Indicator.scss';
 
+type Tone = 'positive' | 'neutral' | 'negative';
+
```
dleroux: Wondering if `Tone` is a better choice than `Status`. 

Banners have:
`Status = 'success' | 'info' | 'warning' | 'critical'`

Badge has: 
`Status = 'success` | `info` | `attention` | `warning` | `new`

This could be 
`Status = 'neutral' | 'info' | 'warning' | 'critical'`

cc: @sarahill

sarahill:
@kyledurand Hi! Can you provide more context on how you want to use the negative indicator?

src/components/Indicator/Indicator.tsx
sarahill: Agree with using status vs. tone. Keeps the same mental model.

src/components/Indicator/Indicator.tsx
kyledurand: Was wondering about this as well but it caught me off guard when trying to use it. I'm fine with defaulting to true if you think that's best. It's less breaking than defaulting to false

kyledurand:
Sorry for lack of context. I was looking to incorporate this into this list component
image

sarahill:
@kyledurand No worries! Thanks for the screenshot. 

Few questions:
- On the items below that have the indicators, do these show up only when there is a potential risk?
- Do the indicators ever stop pulsing?
- Do the statuses of the list items change/can they change?

I'm trying to determine if this makes sense as an evolution of the `indicator` or if this is a new pattern.

kyledurand:
> On the items below that have the indicators, do these show up only when there is a potential risk?
They can be displayed for low or no risk as well
image
> Do the indicators ever stop pulsing?
They actually do not pulse at all
> Do the statuses of the list items change/can they change?
No, I believe they are static. 

I'm also unsure if this is the right evolution of this component. I'm just going based on the visual similarities. If you think there's something better, or if you think this component should just live in web then I'm fine with shutting this down

sarahill:
For this, I would suggest using an icon to reinforce the risk instead of the dot. There are some accessibility concerns with the dots since it's relying on color alone.

I'd like to hold off making these changes in this component. This feels like a new pattern to me and not an evolution of `Indicator`. I don't want to hold you up on your project so you can use this in Web to move forward. Just keep in mind accessibility and how this might be consumed back to Polaris.

If you want to chat more I'm happy to!

PR #1828, opened 2019-07-12
[v4] Stop repeating typedefs and use undefined as default values for hooks

BPScott:

There's a little bug in how we build the polarisContext in withAppProvider - we get some values that may be nullable from `useAppBridge`, `useStickyManager` and `useScrollLockProvider`, but the typing of PolarisContext suggests they will always be present. In order to avoid this should use existing 
type definitions instead of redefining them multiple times.

We should lean on Typescript's `React.ContextType<SomeContext>` to extract the types from a context.

We should lean on Typescript's `ReturnType` type to let us extract the return type of a function instead having to import the typedefs in addition to a function.

Update useAppBridge, useStickyManager and useScrollLockManager to return
undefined instead of null.

Define WithAppProviderProps in terms of the return types of our hooks

Use `ReturnType<typeof createAppBridge>` instead of redefining
`ClientApplication<{}> | undefined`

Remove create-polaris-context as it is no longer used

### How to 🎩

Check typing, and tests still pass
Ensure scroll locking, and stickyness still work


BPScott:
@AndrewMusgrave Here's the result of our pairing work.

This is going to conflict slightly with #1810. Merge #1810 on monday morning and I'll rebase this on top afterwards.

PR #1831, opened 2019-07-13
[v4] Remove deprecated functionality

BPScott:

Removing deprecated functionality that we said we'd remove in v4

- Remove `secondaryFooterAction` prop on `Card`
- `iconBody` prop on `Navigation`
- `groups` prop on `Select`
- `plain-list` sass mixin

### How to 🎩

Ensure the above items continue to work by checking tests
Search the codebase for other deprecated functionality and ensure the only ones left are things that will go in v5 (namely app-bridge related things)

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -39,8 +39,6 @@ export interface SelectGroup {
 export interface BaseProps {
   /** List of options or option groups to choose from */
   options?: (SelectOption | SelectGroup)[];
-  /** @deprecated List of grouped options to choose from */
-  groups?: (SelectOption | SelectGroup)[];
   /** Label for the select */
   label: string;
   /** Adds an action to the label */
@@ -78,7 +76,6 @@ const getUniqueID = createUniqueIDFactory('Select');
 
 export default function Select({
   options: optionsProp,
-  groups: groupsProp,
   label,
   labelAction,
   labelHidden: labelHiddenProp,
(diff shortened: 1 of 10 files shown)


src/components/Select/Select.tsx
```diff
@@ -39,8 +39,6 @@ export interface SelectGroup {
 export interface BaseProps {
   /** List of options or option groups to choose from */
   options?: (SelectOption | SelectGroup)[];
```
AndrewMusgrave: I'm not positive but I this might require more work, was the intention to have this
```suggestion
  options?: SelectOption[];
```
?

AndrewMusgrave: And if it was we may want to deprecate the type

BPScott: We still want the ability to specify groups which render as `<optgroup>` elements within the `select` in the html, we just don't want to have two props when the options prop can handle both cases. @ry5n you added this deprecation in https://github.com/Shopify/polaris-react-deprecated/pull/1576, can you confirm that was the intent?

Stick the following in your Playground (yay having a function based playground is a nice little thing):


```js
import React from 'react';
import {Page, Select} from '../src';

  return (
    <Page title="Playground">
      <Select
        label="Date range"
        options={[
          {label: 'Today', value: 'today'},
          {label: 'Yesterday', value: 'yesterday'},
          {label: 'Last 7 days', value: 'lastWeek'},
        ]}
        onChange={() => {}}
        value="today"
      />
      <Select
        label="Date range"
        options={[
          {
            title: 'First',
            options: [
              {label: 'Today', value: 'today'},
              {label: 'Yesterday', value: 'yesterday'},
              {label: 'Last 7 days', value: 'lastWeek'},
            ],
          },
          {
            title: 'Second',
            options: [
              {label: 'Today', value: 'today'},
              {label: 'Yesterday', value: 'yesterday'},
              {label: 'Last 7 days', value: 'lastWeek'},
            ],
          },
        ]}
        onChange={() => {}}
        value="today"
      />
    </Page>
  );
}
```

ry5n: The idea was exactly that: not needing 2 props for grouped and ungrouped options. One thing I did badly was not checking to see what other lists still do it with 2 props. I think OptionList and ChoiceList might, so I might have made more work/an inconsistent API with that choice 😭

PR #1839, opened 2019-07-15
Allow Autocomplete.TextField's autoComplete to be overridable

### LINKED ISSUE #1834: Autocomplete TextField overrides autoComplete attribute
## Issue summary

The `Autocomplete.TextField` component forces the `autoComplete` attribute to `false` (`"off"`), which is a value Chrome has been known to ignore.

A fix was added for the base TextField component in #1259 to allow a `string` value for `autoComplete`.

## Expected behavior

Chrome should not be displaying autofill suggestions on top of autocomplete results. Because Chrome does not respect the `autoComplete` attribute having a value of `"off"`, as it should, we should allow setting this attribute to other string values, the same way we allow it to be set for the base `TextField`.

## Actual behavior

Setting the `autoComplete` attribute to an arbitrary string value will be ignored and the `autoComplete` attribute will get a value of `off`.

1. Create an `Autocomplete.TextField` component with an `autoComplete` prop of `"disabled"`
1. Create an `Autocomplete` component with a `textField` prop set to the component created above
1. Render the Autocomplete and observe in Chrome dev tools the `TextField` input has a `autoComplete` attribute of `"off"`, instead of `"disabled"`

[CodeSandbox example](https://codesandbox.io/s/autocompletetextfield-hyobo)

## Reduced test case

## Specifications

- Polaris version number: 3.19.0
- Browser: Chrome
- Device: MacbookPro
- Operating System: macOS

BPScott:

Fixes #1834

Allows users to set Autocomplete.TextField's autoComplete prop that gets passed into TextField

### How to 🎩

In the playground, create an Autcomplete with user-defined autoComplete field, and see that value is applied to the rendered input element. See code below.

<details>

```jsx
import {Page, Autocomplete} from '../src';

interface State {}

  render() {
    const noop = () => {};
    const textField = (
      <Autocomplete.TextField
        label="Tags"
        value=""
        placeholder="Search"
        onChange={noop}
        autoComplete="braaaap"
      />
    );

    return (
      <Page title="Playground">
        <Autocomplete
          textField={textField}
          options={[]}
          selected={[]}
          onSelect={noop}
        />
      </Page>
    );
  }
}
```

</details>

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -23,6 +23,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Fixed React imports in the `Filters` component to use `import * as React` for projects that don't use `esModuleInterop` ([#1820](https://github.com/Shopify/polaris-react/pull/1820))
 - Fixed `tabIndex` on `main` element causing event delegation issues ([#1821](https://github.com/Shopify/polaris-react/pull/1821))
 - Fixed icon color for destructive ActionList items ([#1836](https://github.com/Shopify/polaris-react/pull/1836))
+- Fixed not being able to explictly set `autoComplete` prop on`Autoomplete.TextField`. ([#1839](https://github.com/Shopify/polaris-react/pull/1839))
 
 ### Documentation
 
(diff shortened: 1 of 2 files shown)


UNRELEASED.md
```diff
@@ -23,6 +23,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Fixed React imports in the `Filters` component to use `import * as React` for projects that don't use `esModuleInterop` ([#1820](https://github.com/Shopify/polaris-react/pull/1820))
 - Fixed `tabIndex` on `main` element causing event delegation issues ([#1821](https://github.com/Shopify/polaris-react/pull/1821))
 - Fixed icon color for destructive ActionList items ([#1836](https://github.com/Shopify/polaris-react/pull/1836))
+- Fixed not being able to explictly set `autoComplete` prop on`Autoomplete.TextField`. ([#1839](https://github.com/Shopify/polaris-react/pull/1839))
```
```suggestion
- Fixed not being able to explictly set `autoComplete` prop on`Autocomplete.TextField`. ([#1839](https://github.com/Shopify/polaris-react/pull/1839))
```

BPScott: Haha that would have been unfortunate, thanks!

PR #1840, opened 2019-07-15
Add warning message for v4 removals

AndrewMusgrave:

Some internal types and components are being removed from our export in v4

Adding a message to let consumers know

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -26,6 +26,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Documentation
 
+- Added warning log for internal types and components that are being removed in `v4` ([#1840](https://github.com/Shopify/polaris-react/pull/1840))
 - Added links to App Bridge React component documentation in deprecation notices for embedded components ([#1765](https://github.com/Shopify/polaris-react/pull/1765))
 - Improved link text for App Bridge deprecation notices. [#1802](https://github.com/Shopify/polaris-react/pull/1802)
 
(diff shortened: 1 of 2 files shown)


BPScott:
These sorts of deprecation notices should be actionable - consumers should be able to do something to make deprecation message go away. There's nothing users can do with about this message - every single user of AppProvider (which is everybody, as it is required) will get this and it'll just gunk up their (and their users's) console and they're stuck with it. I think this sort of message is better handled by writing migration documentation.

Usually I'd suggest refactoring so usage of the problematic items trigger warnings (e.g. warn about WithRef when a consumer calls WithRef) but we still use many of these a bunch internally in v3, and there's nothing consumers can do to fix that and it's not worth us refactoring within v3. So I think we're better off just documenting these changes.

AndrewMusgrave:
> These sorts of deprecation notices should be actionable - consumers should be able to do something to make deprecation message go away. There's nothing users can do with about this message - every single user of AppProvider (which is everybody, as it is required) will get this and it'll just gunk up their (and their users's) console and they're stuck with it. I think this sort of message is better handled by writing migration documentation. Usually I'd suggest refactoring so usage of the problematic items trigger warnings (e.g. warn about WithRef when a consumer calls WithRef) but we still use many of these a bunch internally in v3, and there's nothing consumers can do to fix that and it's not worth us refactoring within v3. So I think we're better off just documenting these changes.

I would say this is more of a notice/warning since we're not really deprecating it. We're removing a few things that shouldn't have been there to begin with. We don't have any guidelines for how we handle this so I thought that might be an ok approach but it's a fair point that it might be annoying to always have this in the console. 

I'll be writing up migration/release docs so I'll close this if we think that'll be good enough.

cc/ @dleroux because he's working on the release process and may want to create guidelines for this in the future (we'll probably be removing more exports in v5)

PR #1841, opened 2019-07-15
[v4] Further test and context tidying

BPScott:

Tweaking some of our test and context config to make it a bit less verbose, so you have half a chance of understanding what is happening.

- Simplify args of mountWithAppProvider in legacy tests - There's no need to nest the context values inside another object. This API matches the modern testing API for custom mounts.
-  Stop exporting the Theme context type from utilities/theme - Consuming sites should use `React.contextType<typeof ThemeProviderContext>` like the calling sites of all other contexts we provide
- Rename ThemeProviderContext to ThemeContext  - All our other provided contexts don't have Provider in their name, neither should this
- Make legacy withAppProvider use the modern types - Have a single suite of types for our mounting functions in enzyme and react-testing instead of duplicating them

### How to 🎩

Ensure tests and typechecks pass

### DIFF
--- a/src/test-utilities/legacy.tsx
+++ b/src/test-utilities/legacy.tsx
@@ -91,23 +91,19 @@ interface AppContextOptions {
 }
 
 interface MountWithAppProviderOptions {
-  context?: {
-    themeProvider?: DeepPartial<ThemeProviderContextType>;
-    frame?: DeepPartial<FrameContextType>;
-    intl?: TranslationDictionary | TranslationDictionary[];
-    scrollLockManager?: ScrollLockManager;
-    stickyManager?: StickyManager;
-    appBridge?: ClientApplication<{}> | {};
-    link?: LinkLikeComponent;
-  };
+  themeProvider?: DeepPartial<ThemeProviderContextType>;
+  frame?: DeepPartial<FrameContextType>;
+  intl?: TranslationDictionary | TranslationDictionary[];
+  scrollLockManager?: ScrollLockManager;
+  stickyManager?: StickyManager;
+  appBridge?: ClientApplication<{}> | {};
+  link?: LinkLikeComponent;
 }
 
 export function mountWithAppProvider<P>(
   node: React.ReactElement<P>,
-  options: MountWithAppProviderOptions = {},
+  ctx: MountWithAppProviderOptions = {},
 ): PolarisContextReactWrapper<P, any> {
-  const {context: ctx = {}} = options;
-
   const intlTranslations =
     (ctx.intl && merge(translations, ctx.intl)) || translations;
   const intl = new I18n(intlTranslations);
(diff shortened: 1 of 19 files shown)


src/test-utilities/legacy.tsx
```diff
@@ -76,38 +73,10 @@ function updateRoot(wrapper: AnyWrapper) {
...
-  };
-}
-
 export function mountWithAppProvider<P>(
   node: React.ReactElement<P>,
-  options: MountWithAppProviderOptions = {},
-): PolarisContextReactWrapper<P, any> {
-  const {context: ctx = {}} = options;
-
+  ctx: WithProvidersOptions = {},
```
AndrewMusgrave: Just a thought if we decide to change `WithProvidersOptions` type in the future, that'll require some manual updates in legacy tests. I'd like us to be in a spot where we'll never have to touch legacy stuff again (unless we need to fix a test)

src/test-utilities/legacy.tsx
```diff
@@ -133,9 +102,10 @@ export function mountWithAppProvider<P>(
 
   const link = new Link(ctx.link);
 
-  const appBridge = ctx.appBridge || null;
+  // Yes this is bad typing, we'll fix it later
```
AndrewMusgrave: I don't think we'll need to go and fix this later. It's a little annoying but it doesn't cause us any harm and there isn't to much value going back and changing it since it's legacy now and we're moving away from `enzyme`.

src/utilities/theme/context.tsx
```diff
@@ -1,12 +1,10 @@
 import React from 'react';
 import {Theme} from './types';
 
-export interface ThemeProviderContextType {
+export interface ThemeContextType {
   logo: Theme['logo'] | null;
 }
 
-export const ThemeProviderContext = React.createContext<
```
AndrewMusgrave: I'm a little conflicted about this. The reason why I could see this staying `ThemeProviderContext` is that it's a component context that matches the name of the component; `ThemeProviderContext`, `ResourceListContext`, `FrameContext`, etc. What are your thoughts on this?

BPScott: Per conversation:

IMO The most important part of these contexts is where they get called, in this case components that use `useTheme()`, not where the value came from. it seems odd for the return type of `useTheme` to be `ThemeProviderContext` - why do you care about the component that provided this context, you only care about the value? `ThemeContext` reflects that. In tests we never call the `ThemeProvider` component so trying to hint at the name of the component that provides the context value is misleading anyway.

I'd like to follow the road we started on by splitting the AppProvider up into multiple contexts where one component doesn't necessitate a single context, such a binding feels restrictive.

src/test-utilities/legacy.tsx
BPScott: If we were to ever add a new context into AppProvider, we'd want to add it to both the legacy and modern test harnesses, having to duplicate that feels awkward.

I think the end goal of this is to make our mountWithAppProvider and mountWithContext functions a lot slimmer and instead have all the logic around providing defaults values for contexts done in PolarisTextProvider.

PR #1851, opened 2019-07-16
[Page] Add subtitle prop to Header

chloerice:

Currently, the only support for secondary content in the `Page` `Header` is the `titleMetadata` prop. In use cases that benefit from a subtitle, that prop is not sufficient since the title lays out its children inline. 

This PR adds an optional `subtitle` prop to the `Page`, which accepts a string and renders below the `title` in the `Header`. 

This came out of a pairing session that followed a [Slack discussion](https://shopify.slack.com/archives/C4Y8N30KD/p1563212872385900) around the Billing team's [use case](https://docs.google.com/presentation/d/143hAb4HYvsoSGFjT11-h6CVdCJKsUiY-4XbfssG3Suk/edit#slide=id.g5b180fc265_0_86).

|  Large screen | Small screen  |
|---|---|
|Screen Shot 2019-07-16 at 6 18 32 PM|Screen Shot 2019-07-16 at 6 18 47 PM|

### Input needed

@mirualves @sarahill I'm seeking design input for the `subtitle` styling and placement. 

- I've chosen to implement this as an `h2` using the `Heading` component, as using `medium` `DisplayText` doesn't seem to give the appropriate visual hierarchy.

<details>
<summary>Click to see collapsed screenshot example</summary>

Screen Shot 2019-07-16 at 3 57 57 PM

</details>

- I'm currently working on how best to handle `titleMetadata` layout since `subtitle` placement needs to break the placement of the metadata. I'm thinking the layout should update to place and space `titleMetadata` inline with the `subtitle` instead of the title when present.

## 

### How to 🎩

<details>

```jsx
import {PageDownMajorMonotone} from '@shopify/polaris-icons';
import {Page, Badge} from '../src';

interface State {}

  render() {
    return (
      <Page
        breadcrumbs={[{content: 'Bills', url: '/settings/billing'}]}
        title="Invoice, but this page could have a longer title"
        subtitle="Statement period: May 3, 2019 to June 2, 2019"
        titleMetadata={<Badge status="success">Paid</Badge>}
        secondaryActions={[{content: 'Download', icon: PageDownMajorMonotone}]}
        actionGroups={[
          {
            title: 'Promote',
            actions: [
              {
                content: 'Share on Facebook',
                accessibilityLabel: 'Individual action label',
                onAction: () => {},
              },
            ],
          },
        ]}
        pagination={{
          hasPrevious: true,
          hasNext: true,
        }}
        separator
      >
        <p>Page content</p>
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Page/components/Header/Header.tsx
+++ b/src/components/Page/components/Header/Header.tsx
@@ -11,6 +11,7 @@ import EventListener from '../../../EventListener';
 import {buttonsFrom} from '../../../Button';
 import Breadcrumbs, {Props as BreadcrumbsProps} from '../../../Breadcrumbs';
 import DisplayText from '../../../DisplayText';
+import Heading from '../../../Heading';
 import Pagination, {PaginationDescriptor} from '../../../Pagination';
 import ActionMenu, {hasGroupsWithActions} from '../../../ActionMenu';
 
@@ -20,6 +21,8 @@ import styles from './Header.scss';
 export interface Props {
   /** Page title, in large type */
   title: string;
+  /** Page subtitle, in medium type (stand-alone app use only) */
+  subtitle?: string;
   /** Important and non-interactive status information shown immediately after the title. (stand-alone app use only) */
   titleMetadata?: React.ReactNode;
   /** Visually hide the title (stand-alone app use only) */
@@ -83,6 +86,7 @@ class Header extends React.PureComponent<ComposedProps, State> {
   render() {
     const {
       title,
+      subtitle,
       titleMetadata,
       titleHidden = false,
       icon,
@@ -94,6 +98,7 @@ class Header extends React.PureComponent<ComposedProps, State> {
       actionGroups = [],
       polaris: {intl},
     } = this.props;
+
     const {mobileView} = this.state;
 
     if (icon) {
@@ -123,17 +128,33 @@ class Header extends React.PureComponent<ComposedProps, State> {
         </div>
       ) : null;
 
-    const titleMarkup = (
-      <div className={styles.Title}>
-        <div className={styles.DisplayTextWrapper}>
-          <DisplayText size="large" element="h1">
-            {title}
-          </DisplayText>
-        </div>
+    const titleMetadataMarkup = titleMetadata ? (
+      <div className={styles.TitleMetadata}>{titleMetadata}</div>
+    ) : null;
 
-        {titleMetadata && (
-          <div className={styles.TitleMetadataWrapper}>{titleMetadata}</div>
-        )}
+    const subtitleMarkup = subtitle ? (
+      <Heading element="h2">{subtitle}</Heading>
+    ) : null;
+
+    const titleSecondaryContent =
+      subtitle || titleMetadata ? (
+        <div className={styles.TitleSecondaryContent}>
+          {subtitleMarkup}
+          {titleMetadataMarkup}
+        </div>
+      ) : null;
+
+    const titleClassName = classNames(
+      styles.Title,
+      subtitle && styles.hasSubtitle,
+    );
+
+    const titleMarkup = (
+      <div className={titleClassName}>
+        <DisplayText size="large" element="h1">
+          {title}
+        </DisplayText>
+        {titleSecondaryContent}
       </div>
     );
 
--- a/.storybook/polaris-readme-loader.js
+++ b/.storybook/polaris-readme-loader.js
@@ -115,6 +115,7 @@ import {
   NotificationMajorMonotone,
   OnlineStoreMajorTwotone,
   OrdersMajorTwotone,
+  PageDownMajorMonotone,
   PrintMinor,
   ProductsMajorTwotone,
   ProfileMinor,
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -12,6 +12,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Added a `subtitle` prop to `Page` ([#1851](https://github.com/Shopify/polaris-react/pull/1851))
+
 ### Bug fixes
 
 ### Documentation
--- a/src/components/Page/README.md
+++ b/src/components/Page/README.md
@@ -271,6 +271,23 @@ Use when a primary action functions better as part of the page content instead o
 </Page>
 ```
 
+### Page with subtitle
+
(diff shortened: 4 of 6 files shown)


src/components/Page/components/Header/Header.tsx
```diff
@@ -123,12 +128,19 @@ class Header extends React.PureComponent<ComposedProps, State> {
...
 
+    const subtitleMarkup = subtitle ? (
+      <div className={styles.Subtitle}>
+        <Heading element="h2">{subtitle}</Heading>
+      </div>
+    ) : null;
+
     const titleMarkup = (
       <div className={styles.Title}>
-        <div className={styles.DisplayTextWrapper}>
```
chloerice: I removed the `className` since it wasn't actually being used in any of the stylesheets.

sarahill:
@chloerice When I add `metadata` it looks a little funky. Might have to wrap the `title` and `metadata` together and `subtitle` below. Although, this will affect the hierarchy on mobile.

image

Screen Shot 2019-07-16 at 4 32 29 PM

sarahill:
One more 🔧 . I don't think this would happen but it's worth documenting in best practices do's/dont's. When you've got everything under the sun added to `page`.

image

chloerice:
Thanks @sarahill! The `titleMetadata` funky placement isn't a problem anymore, it sits inline with the `subtitle` when present. Otherwise, sits inline with the `title` as before. Let me know what you think! 

I agree we should make it clear that since the title and subtitle render `h1` and `h2`, those elements shouldn't be used anywhere else on the page. Right now, the prop descriptions for this

||  Large screen | Small screen  |
|---|---|---|
|With `titleMetadata` and `subtitle`|Screen Shot 2019-07-16 at 6 18 32 PM|Screen Shot 2019-07-16 at 6 18 47 PM|
|With just `titleMetadata`|Screen Shot 2019-07-16 at 9 08 02 PM|Screen Shot 2019-07-16 at 9 08 14 PM|
|With just `subtitle`|Screen Shot 2019-07-16 at 9 07 28 PM|Screen Shot 2019-07-16 at 9 07 43 PM|
|With neither `titleMetadata` nor `subtitle`|Screen Shot 2019-07-16 at 9 08 48 PM|Screen Shot 2019-07-16 at 9 08 32 PM|

src/components/Page/components/Header/Header.tsx
```diff
@@ -123,17 +128,33 @@ class Header extends React.PureComponent<ComposedProps, State> {
...
-          <DisplayText size="large" element="h1">
-            {title}
-          </DisplayText>
+    const titleMetadataMarkup = titleMetadata ? (
+      <div className={styles.TitleMetadata}>{titleMetadata}</div>
+    ) : null;
+
+    const subtitleMarkup = subtitle ? (
+      <Heading element="h2">{subtitle}</Heading>
```
dleroux: Maybe a header is not the best for this. I couldn't find the w3c official page on it, but I remembered this: http://html5doctor.com/howto-subheadings. Sounds like because hgroup is obsolete they recommend a p tag.

dleroux:
Code looks good, but I feel like the metadata should be inline with the main title? Semantically I think it would make more sense.

We'll also need to adjust the skeleton components to include this subheading.

ping: @AndrewMusgrave was working on adding an image before the title a few days ago.

brendandeere:
cc: @sarahebbs @dbferreira

src/components/Page/components/Header/Header.tsx
sarahill: Yea looks like `hgroup` was removed from HTML5. Interesting https://www.w3.org/TR/2014/CR-html5-20140204/common-idioms.html#sub-head

brendandeere:
We are hoping to have this available by Monday July 29th? Are there significant blockers which would make that unrealistic? 

cc: @sarahebbs @chloerice

sarahill:
@brendandeere @chloerice @sarahebbs 

I have a couple of questions:

1. Wondering if it would work to make the position of the metadata flexible so it could be inline or below the title?
2. Does it make sense to modify the component or is this only going to be used in the context of billing?

I'm hesitant to introduce the subtitle without clear documentation on how/when to use it. I can see a case for it but if there is the option for a `subtitle` and `metadata`, it's likely there will be a case where these might be used together. In that case:
1. Do we want it to be used in that way?
2. When would it be used that way?
3. How does this adapt to mobile in an elegant way?

It may also be we don't recommend using both which is where my thought of having a flexible position is coming from.

src/components/Page/components/Header/Header.tsx
chloerice: 💯

brendandeere:
> Wondering if it would work to make the position of the metadata flexible so it could be inline or below the title?

This would work fine for us. My understanding is that this would give us the same effect as the "subtitle" presented in this PR?

> Does it make sense to modify the component or is this only going to be used in the context of billing?

@sarahebbs can speak more to the content and UX, but to me this hierarchy of information makes a lot of sense for our use case. We are trying to avoid long titles like `Current billing Cycle: October 20, 2019-November 20, 2019`.

---

Is making the metadata position flexible an option? how much added work is that? I'm hoping we can add a heading to this page in less than a week :)

sarahill:
Yes, same affect. It'd look something like this on both desktop and small screen web.
image

@chloerice what are your thoughts on the approaches? I'll go with your recommendation just throwing thoughts out to consider. If we go with the approach of adding a `subtitle` let's just make sure usage is documented. @sarahebbs we can work on that together if you've got time.

chloerice:
After talking with @AndrewMusgrave and looking at the new `app-bridge-react` components, specifically the `TitleBar`, we think the best way forward will be to:

- Deprecate the `titleMetadata` prop
- Support customization of the `title` by updating the type from `string` to `React.ReactNode`
    - This will enable the flexibility needed across use cases, including titles with images or icons preceding them, titles with status badges, and titles with subtitles
- Add examples for each of the above use cases to encourage best practices (for example, using a `<p>` element instead of an `h2` for subtitles)

I'll be closing this PR and opening a new one that does the above in order to preserve this conversation and code for reference.

chloerice:
> We are hoping to have this available by Monday July 29th? Are there significant blockers which would make that unrealistic?

Regarding timeline, I will do my best to have the new PR ready by the end of the week. But realistically, the changes likely won't be available until August 6th due to summit obligations and vacations.

PR #1854, opened 2019-07-17
[v4][performance] Prevent inappropriate scrollbar toggling on the viewport

kaelig:

Fixes an issue where user-agents that show a scrollbar when content of the page is higher than the viewport would also hide that scrollbar when content is smaller than the viewport, creating a flicker & unnecessary full-page reflows (slow on low-spec devices).

In the screencast below, you can witness that the entire page is jumping when content changes from being smaller than the viewport to being higher than the viewport, due to the introduction of a scrollbar on the right hand side of the page.

scrollbar jump

This is also a problem with modals, where a scrollbar gets introduced if the contents of the page are smaller than the viewport, due to this CSS:

https://github.com/Shopify/polaris-react/blob/a3c2898ac4d210b2eed6b4e91fa5d7c2a8021d3e/src/components/ScrollLock/ScrollLock.scss#L1-L2

scrollbar jump 2

Adds `overflow-y: scroll` to the `html` element (a [classic fix](https://css-tricks.com/eliminate-jumps-in-horizontal-centering-by-forcing-a-scroll-bar/)).

⚠️ This doesn’t impact these users who still get a nice and clean experience in macOS:
- browsers on macOS when “Show scroll bars: When scrolling” is selected in the general system preferences when a mouse or a trackpad is used
- users of a trackpad on macOS with “Show scroll bars: When scrolling, or Automatically based on mouse or trackpad”

### Renderings with various macOS settings (with a mouse plugged in)

Screen Shot 2019-07-16 at 5 09 31 PM
Screen Shot 2019-07-16 at 5 10 03 PM
Screen Shot 2019-07-16 at 5 10 15 PM

—-
## 

### How to 🎩

Compare scrollbar behavior between:
- [Original modal (latest version of Polaris React)](https://polaris-react.herokuapp.com/iframe.html?id=all-components-modal--basic-modal)
- [This PR’s modal](https://polaris-react-pr-1854.herokuapp.com/iframe.html?id=all-components-modal--basic-modal)

</details>

### 🎩 checklist


src/components/ScrollLock/tests/ScrollLock.test.tsx
```diff
@@ -29,23 +29,29 @@ describe('ScrollLock', () => {
...
 
-  it('removes data attribute from the body when it unmounts', () => {
+  it('removes data attribute from the first child element of the body when it unmounts', () => {
     const scrollLock = mountWithAppProvider(<ScrollLock />);
     scrollLock.unmount();
-    expect(document.body.hasAttribute(`${SCROLL_LOCKING_ATTRIBUTE}`)).toBe(
-      false,
-    );
+    expect(
+      document.body.firstElementChild.hasAttribute(
```
kaelig: I’d like to write a test for this, but it fails. I’m not familiar enough with Jest/Enzyme/React to understand why this is failing (this might be why it wasn’t tested at all). Any pointers?

src/components/ScrollLock/tests/ScrollLock.test.tsx
```diff
@@ -29,23 +29,29 @@ describe('ScrollLock', () => {
 
     scrollLockContainer.instance().setState({showScrollLock: false});
 
-    expect(document.body.hasAttribute(`${SCROLL_LOCKING_ATTRIBUTE}`)).toBe(
-      true,
-    );
+    expect(
+      document.body.firstElementChild.hasAttribute(
+        `${SCROLL_LOCKING_WRAPPER_ATTRIBUTE}`,
```
BPScott: nitpick: There's no point to string interpolation if it's a single variable - same goes for other test cases below

```suggestion
        SCROLL_LOCKING_WRAPPER_ATTRIBUTE,
```

kaelig: Thank you for confirming my doubts, I was totally wondering why those were in there in the first place! Will fix as I rework the tests.

src/components/AppProvider/utilities/ScrollLockManager/ScrollLockManager.ts
```diff
@@ -1,6 +1,5 @@
 import {isServer} from '../../../../utilities/target';
 
-export const SCROLL_LOCKING_ATTRIBUTE = 'data-lock-scrolling';
```
dleroux: Looks like this is still being exported in a couple of places. Maybe that's the reason for your failing test? Interestingly I wanted to tophat on web and building consumer failed, which is why I caught this, but my playground ran fine.

src/components/ScrollLock/tests/ScrollLock.test.tsx
dleroux: This worked for me. There was no first child so I created one.
```
describe('ScrollLock', () => {
  it('does not remove the data attribute from the first child element of the body when two scrolllocks are mounted and one unmounts', () => {
    class DummyFrame extends React.Component {
      state = {
        showScrollLock: true,
      };

      render() {
        const {showScrollLock} = this.state;

        // eslint-disable-next-line shopify/jest/no-if
        const scrollLockMarkup = showScrollLock ? <ScrollLock /> : null;

        return (
          <React.Fragment>
            {scrollLockMarkup}
            <ScrollLock />
          </React.Fragment>
        );
      }
    }

    document.body.appendChild(document.createElement('div'));
    const scrollLockContainer = mountWithAppProvider(<DummyFrame />);

    scrollLockContainer.instance().setState({showScrollLock: false});

    const lockedWrapper = document.querySelector(
      `[${SCROLL_LOCKING_WRAPPER_ATTRIBUTE}]`,
    );

    expect(lockedWrapper).toBeTruthy();
  });

  it('adds data attribute to the first child element of the body when it mounts', () => {
    document.body.appendChild(document.createElement('div'));
    mountWithAppProvider(<ScrollLock />);
    const lockedWrapper = document.querySelector(
      `[${SCROLL_LOCKING_WRAPPER_ATTRIBUTE}]`,
    );
    expect(lockedWrapper).toBeTruthy();
  });

  it('removes data attribute from the first child element of the body when it unmounts', () => {
    document.body.appendChild(document.createElement('div'));
    const scrollLock = mountWithAppProvider(<ScrollLock />);
    scrollLock.unmount();
    const lockedWrapper = document.querySelector(
      `[${SCROLL_LOCKING_WRAPPER_ATTRIBUTE}]`,
    );
    expect(lockedWrapper).toBeFalsy();
  });
});
```

dleroux:
🎩 in web and ios and it works!

src/components/AppProvider/utilities/ScrollLockManager/ScrollLockManager.ts
kaelig: I suppose this is a breaking change, then! I'll make 2 PRs, as the modal refactor should go in v4, but the change to `html` can go in v3.

kaelig:
I rebased atop `version-4.0.0`, as it could be seen as a breaking change.

src/components/AppProvider/utilities/ScrollLockManager/ScrollLockManager.ts
BPScott: We used this internally in a few places but it doesn't look like it was ever exported publically. This doesn't need to go to v4 because of this

kaelig: Oh right! I'll edit the changelog and remove the item that mentions this constant. 

However, I now consider this is a breaking change from a CSS point of view, as anything that performs full-page scroll locking (outside of Polaris) should stop doing so on the body and apply it to the html element instead (or they'll have double scrollbars). I'll mention that in the changelog as well. 

If you disagree, I'm happy to consider rebasing onto master again.

BPScott: The CSS behaviour change does feel a bit scary, it's one of those /should/ be fine things but as we're on the cusp of V4 we might as well be safe :) +1 to keeping this in v4

dleroux: You could also not change the name of the constant. We simply moved the attribute from the body to the first child, it was a wrapper before as well.

kaelig: This is a great idea!

I went ahead and renamed the data attribute as well as the name of the constant:

https://github.com/Shopify/polaris-react/pull/1854/commits/cdb175bf565099312ab13b092c405168c0a0d725

UNRELEASED-V4.md
```diff
@@ -20,11 +20,14 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Removed `secondaryFooterAction` prop on `Card`. Pass an array of secondary actions to the `secondaryFooterActions` prop instead. ([#1830](https://github.com/Shopify/polaris-react/pull/1830))
 - Removed `iconBody` prop on `Navigation`. Pass a string into the `icon` prop instead. ([#1830](https://github.com/Shopify/polaris-react/pull/1830))
 - Removed `groups` prop on `Select`. Pass groups to the `options` prop instead. ([#1830](https://github.com/Shopify/polaris-react/pull/1830))
+- Applied scroll locking at the root (`html`) level instead of the `body`, which could impact non-Polaris modals and overlays in consuming apps ([#1854](https://github.com/Shopify/polaris-react/pull/1854))
```
kaelig: Is this a good enough warning?

src/components/ScrollLock/ScrollLock.scss
```diff
@@ -1,10 +1,4 @@
-[data-lock-scrolling] {
-  overflow-y: scroll;
-  margin: 0;
-
-  // stylelint-disable selector-max-attribute
-  [data-lock-scrolling-wrapper] {
-    overflow: hidden;
-    height: 100%;
-  }
+[data-scroll-lock] {
```
kaelig: For consistency, I renamed this to the same name as the file (in snake-case).

src/utilities/scroll-lock-manager/scroll-lock-manager.ts
```diff
@@ -1,8 +1,6 @@
 import {isServer} from '../target';
 
-export const SCROLL_LOCKING_ATTRIBUTE = 'data-lock-scrolling';
-export const SCROLL_LOCKING_WRAPPER_ATTRIBUTE = 'data-lock-scrolling-wrapper';
-export const SCROLL_LOCKING_CUSTOM_PROPERTY = '--scroll-lock-body-padding';
```
kaelig: I couldn't find any usage of this custom property, so I deleted it. Please let me know if I missed something.

AndrewMusgrave:
Could this change target `v5` 🙏 ? We’re in the middle of trying to rollout `v4` so it’s a little late to add breaking changes. And with the already huge change-set of `v4` all breaking changes become increasingly risky, especially during a code red.

kaelig:
It's a breaking change but not going to badly break anything. At most, a custom modal might trigger a double scrollbar to appear on the right hand side of the screen for some users. I'd like to try to push it to v4 rather than wait for v5.

BPScott:
+1 to Kaelig's counterpoint, the surface area of this change is small, and it'd be good to get it in as part of v4 rather than waiting another several months till v5

AndrewMusgrave:
This change isn't in the scope of `v4` and the project has already been delayed. We're in the middle of rolling out and already have release candidates. This is a risky change that does break scroll locking in admin ([gif](https://cl.ly/c8bf611a5852)). So I feel this isn't the right time to push a risky change into something that's already at the finish line and risking regressions in the admin, especially in a code red, and it'll further delay working time on more critical issues. If you don't want to wait months for a `v5`, nothing says it can't be released weeks after v4. However, maybe it's possible to make this change in a backward-compatible/non-breaking way that could be rolled out in a minor version bump?

kaelig:
>This is a risky change that does break scroll locking in admin ([gif](https://cl.ly/c8bf611a5852)).

I don't think this is a regression: I could reproduce this issue in the current admin too. Looks like scroll locking is already missing in modals that don't use Polaris React.

Knowing this, does it change your mind?

AndrewMusgrave:
They [seem to work for me](https://cl.ly/4ebbb591df27) without this change

kaelig:
After a thorough review with @AndrewMusgrave, we have stated that this PR will:

- fix full-page reflows on most modals (since most of the admin is using React)
- fix full-page reflows but cause a double scrollbar on a small proportion of remaining polaris-rails modals, for all Windows users and macOS users who have a mouse plugged
- not have any effect on the old `.modal` ones (which didn’t lock scroll in the first place)

### "regression"
Will happen for all Windows users and macOS users who have a mouse plugged.

`ui_modal` isn't used very widely in the admin so this is a very small regression.

Screen_Shot_2019-07-30_at_5_18_48_PM


@danrosenthal Andrew told me you built this scroll locking mechanism, what are your thoughts on rolling out a similar fix in polaris-rails? Is it worth it, and if so, when should it happen? 

(I'm not too familiar with polaris-rails and its test suite, so I couldn't appreciate how long it'd take me to roll out the same change there)

danrosenthal:
> @danrosenthal Andrew told me you built this scroll locking mechanism, what are your thoughts on rolling out a similar fix in polaris-rails? Is it worth it, and if so, when should it happen?
>
> (I'm not too familiar with polaris-rails and its test suite, so I couldn't appreciate how long it'd take me to roll out the same change there)

I'm unable to commit to this, and can't commit to providing review here, but I would be very concerned about any change to scroll locking that didn't happen in concert with a change in Polaris rails. I built this approach to work inter-operably, so would expect the same change to go out in that library as well.

kaelig:
Thanks @danrosenthal for your input on this! I was just checking to see if we wanted to do this in polaris-rails too. I'll close this pull request as I don't have time to do it in both codebases and then react to potential regressions that would require a rollback.

PR #1855, opened 2019-07-17
[v4] Fix "Function components do not support contextType" error in tests

BPScott:

Filters.test.tsx caused the above error due to us using an old version
of hoist-non-react-statics

[Example of a build containing this error](https://circleci.com/gh/Shopify/polaris-react/23684?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

Update the version of hoist-non-react-statics we use.
Remove a pointless reassign of contextTypes - I'm pretty sure we won't ever need this.

### How to 🎩

Check the test output, see there is no "Function components do not support contextType" error

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,6 +107,7 @@
     "@storybook/theming": "^5.1.9",
     "@types/enzyme": "^3.9.3",
     "@types/enzyme-adapter-react-16": "^1.0.5",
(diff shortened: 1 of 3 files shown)


AndrewMusgrave:
😄

PR #1857, opened 2019-07-17
Revert consistent-rem

AndrewMusgrave:

When attempting to roll out `v4` some major changes were noticed that could affect production. I'll be writing up an issue later with details

Reverting #1590


ry5n:
@AndrewMusgrave Were you able to write up an issue with the details of why we had to revert this? There’s someone asking about it in the public issue: https://github.com/Shopify/polaris-react/issues/431

PR #1864, opened 2019-07-18
[v4] Hook improvements

BPScott:

Tightening up our hooks, so they provide the same assurances that withAppProvider does.
By checking that context values aren't empty within our hooks we no longer need to check the return value of the hooks, e.g. before this PR `useStickyManager()` could return an instance of StickyManager or undefined. Now you'll always get an instance of StickyManager so

```js
const stickyManager = useStickyManager();
if (stickyManager) {
  stickyManager.registerStickyItem(/*...*/);
}
```

can become

```js
const stickyManager = useStickyManager();
stickyManager.registerStickyItem(/*...*/);
```

AND!

Simplifying the LinkContext and UnstyledLink and their usage by storing the custom link component directly in the LinkContext

- The default values for I18nContext and ThemeContext are now undefined by default
- useStickyManager will throw if the context is empty (i.e. you've not wrapped everything in AppProvider)
- useScrollLockManager will throw if the context is empty (i.e. you've not wrapped everything in AppProvider)
- useI18n will throw if the context is empty (i.e. you've not wrapped everything in AppProvider)
- useTheme will throw if the context is empty (i.e. you've not wrapped everything in AppProvider)
- useLink may return undefined as a custom link is optional, and a default will be provided by UnstyledLink
- useAppBridge may return undefined as an app context is optional

__At this point every Context provided by AppProvider is undefined by default__
__Every use hook except useLink and useAppBridge for those corresponding contexts will throw an exception if the context is not defined.__

Remove Link, and instead store a LinkLikeComponent directly on the LinkContext

Remove exporting the LinkLikeComponent type, instead consumers should use `AppProviderProps['linkComponent']` which now more accuratly represents what they need to pass into AppProvider.

### How to 🎩

Check tests, type-check etc.

Check that passing a custom linkComponent still renders as you'd hope with the following Playground:

<details>

```jsx
import React from 'react';
import {AppProvider, Page, UnstyledLink, AppProviderProps} from '../src';

type LinkType = AppProviderProps['linkComponent'];

const LinkComponent: LinkType = ({children, url = '', ...rest}) => {
  return (
    <a href={url} data-custom="hai" {...rest}>
      {children}
    </a>
  );
};

  return (
    <AppProvider i18n={{}} linkComponent={LinkComponent}>
      <Page title="Playground">
        <UnstyledLink url="foo">BRAP</UnstyledLink>
      </Page>
    </AppProvider>
  );
}

```

</details>


src/utilities/scroll-lock-manager/hooks.tsx
```diff
@@ -4,5 +4,10 @@ import {ScrollLockManagerContext} from './context';
 export function useScrollLockManager() {
   const scrollLockManager = useContext(ScrollLockManagerContext);
 
+  if (!scrollLockManager) {
+    throw new Error(
+      'No ScrollLockManager was provided. Your application must be wrapped in an <AppProvider> component. See https://polaris.shopify.com/components/structure/app-provider for implementation instructions.',
```
kaelig: Should error messages follow the same standards as the [console messages](https://github.com/Shopify/polaris-react/blob/master/documentation/Console%20messages.md)?

BPScott: They contain the message, and then a link to find out more. What more do you think is missing. Prefixing the string with "Error" seems somewhat pointless.

BPScott:
After many force-pushes and tweaks to on what I started with, this is now ready for review.

src/utilities/scroll-lock-manager/hooks.tsx
kaelig: I initially thought we prefixed messages with `[Polaris]` (for easy filtering) but it seems like the guidelines don't mention anything about that.

src/test-utilities/react-testing.tsx
```diff
@@ -1,16 +1,17 @@
...
 import {StickyManager} from '../utilities/sticky-manager';
 import {createAppBridge} from '../utilities/app-bridge';
 import {I18n} from '../utilities/i18n';
 import translations from '../../locales/en.json';
 import {merge} from '../utilities/merge';
-import {Link} from '../utilities/link';
 import {PolarisTestProvider} from './PolarisTestProvider';
 import {WithProvidersOptions, WithProvidersContext} from './types';
 
+export {mount};
```
AndrewMusgrave: Curious why we're exporting mount here? Rather than importing it direction from `react-testing`? Was it to clean up imports? (e.g. `import {mountWithContext, mount} ...`)

src/utilities/theme/context.tsx
```diff
@@ -1,10 +1,4 @@
...
 import {Theme} from './types';
 
-export interface ThemeContextType {
-  logo: Theme['logo'] | null;
-}
-
-export const ThemeContext = React.createContext<ThemeContextType>({
-  logo: null,
-});
+export const ThemeContext = React.createContext<Theme | undefined>(undefined);
```
AndrewMusgrave: Curious why we're deciding to make this undefined rather than a base object?

src/test-utilities/react-testing.tsx
BPScott: Yep, exactly. I figure this file should be your one-stop-shop for stuff relating to modern testing bits. Having to export mountWithContext from here, and mount from `@shopify/react-testing` felt like there wasn't much connection between those two. This helps make it clear they're related

src/utilities/theme/context.tsx
BPScott: I'm trying to make consistent behaviours between all the contexts that get provided by the AppProvider (I18nContext, ScrollLockManagerContext, StickyManagerContext, AppBridgeContext, LinkContext, ThemeContext - via ThemeProvider).

Currently we say "you must wrap everything in an AppProvider", and there is no expectation for a component to work if you don't do so. The AppProvider provides the default values for all these contexts. It didn't seem like there was any value in providing a default within ThemeProvider, and then specifying one here as the default of the context too (which would never get used because the default from ThemeProvider would always overwrite it).

---

I'd be interested in talking through changing that assumption and identifying contexts that can and should provide their own defaults instead of being provided by the AppProvider/ThemeProvider instead of setting them within the AppProvider/ThemeProvider (and thus allowing some components to work without being wrapped in an AppProvider/ThemeProvider), but we should think about all these contexts at the same time.

AndrewMusgrave: > I'd be interested in talking through changing that assumption and identifying contexts that can and should provide their own defaults instead of being provided by the AppProvider/ThemeProvider instead of setting them within the AppProvider/ThemeProvider (and thus allowing some components to work without being wrapped in an AppProvider/ThemeProvider), but we should think about all these contexts at the same time.

That's definitely something to look at in the future. For the time being, as long as we throw an error when `!theme` this should be ok. My view was that theme was a constructed object that'll have standard defaults whereas `sticky-manager` has an instance of an object that's specific to the state of the consumer's app. Since we do have the premise that we'll always have `AppProvider` within the react tree and will throw an error when we don't. I'm ok with this 👍

PR #1866, opened 2019-07-22
[Webpack example] Add missing plugin

dleroux:

Fixes https://github.com/Shopify/polaris-react/issues/1863

Our Webpack example does not compile because of a missing plugin: https://babeljs.io/docs/en/babel-plugin-proposal-class-properties

This PR adds it.

## 

### How to 🎩

1. git checkout update-webpack-example
2. cd examples/webpack
3. yarn && yarn start

<details>

```jsx

interface State {}

  render() {
    return (
      <Page title="Playground">
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -14,7 +14,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
-- Fixed accessibility issue with ChoiceList errors not being correctly connected to the inputs ([#1824](https://github.com/Shopify/polaris-react/pull/1824));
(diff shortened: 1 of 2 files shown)


BPScott:
I believe @AndrewMusgrave was doing some work to refactor the examples, including refactoring them to use hooks, which would negate the need for the class properties proposal (as we would never use class based components in the first place). I held fire on fixing this as his work was imminent. I think we should wait for him

dleroux:
Since this is targeting version 3, there’s probably value in adding it here and no harm to v4.

PR #1873, opened 2019-07-23
[ToastManager] Move aria-live to containing div for a11y using NVDA

dleroux:

Fixes https://github.com/Shopify/polaris-react/issues/1215

In order to announce the Toast using the aria-live attribut, NVDA require the element with the aria-live prop to be on screen when the content changes.

Our Toasts are wrapped in a Toast manager. This pr moves the aria-live attribute from the Toast to the Toast Manager    

## 

### How to 🎩

Using NVDA (on windows) navigate to:

http://localhost:6006/?path=/story/all-components-toast--basic-toast

and ensure that the Toast messages are being announced.

<details>

```jsx

interface State {}

  render() {
    return (
      <Page title="Playground">
      </Page>
    );
  }
}
```

</details>

### 🎩 checklist


dleroux:
@devonpmack I wasn't able to get the sound working on NVDA but the tools that write out the content seemed to be working as expected. Can you 🎩 and make sure this does fix our issue?

devonpmack:
@dleroux wrong devon?

PR #1877, opened 2019-07-23
[v4] Updated appbridge example to use new context

### LINKED ISSUE #457: Shopify App Bridge context consumer for migrating off legacy React context APIs

### LINKED ISSUE #1507: [v4][Docs] Update docs - consuming appBridge in applications

AndrewMusgrave:

Fixes #1507
Fixes #457

Also updating the example a bit since it won't currently work

### DIFF
--- a/src/components/AppProvider/README.md
+++ b/src/components/AppProvider/README.md
@@ -480,27 +480,30 @@ To provide access to your initialized Shopify App Bridge instance, we make it av
 ```js
 import React from 'react';
 import {render} from 'react-dom';
-import * as PropTypes from 'prop-types';
-import {AppProvider} from '@shopify/polaris';
+import {AppProvider, AppBridgeContext} from '@shopify/polaris';
 import {Redirect} from '@shopify/app-bridge/actions';
 
 class MyApp extends React.Component {
-  // This line is very important! It tells React to attach the `polaris`
-  // object to `this.context` within your component.
-  static contextTypes = {
-    polaris: PropTypes.object,
-  };
+  static contextType = AppBridgeContext;
 
-  redirectToSettings() {
-    const redirect = Redirect.create(this.context.polaris.appBridge);
+  componentDidMount() {
+    const redirect = Redirect.create(this.context);
 
     // Go to {appOrigin}/settings
     redirect.dispatch(Redirect.Action.APP, '/settings');
   }
+
+  render() {
+    return null;
+  }
 }
 
 render(
-  <AppProvider apiKey="YOUR_APP_API_KEY" i18n={{}}>
+  <AppProvider
+    apiKey="YOUR_APP_API_KEY"
+    shopOrigin="YOUR_SHOP_ORIGIN"
+    i18n={{}}
+  >
     <MyApp />
   </AppProvider>,
   document.querySelector('#app'),
--- a/src/index.ts
+++ b/src/index.ts
@@ -11,5 +11,6 @@ export {
 export {
   WithinContentContext as _SECRET_INTERNAL_WITHIN_CONTENT_CONTEXT,
 } from './utilities/within-content-context';
+export {AppBridgeContext} from './utilities/app-bridge';
 
 export {PolarisTestProvider} from './test-utilities/PolarisTestProvider';
--- a/UNRELEASED-V4.md
+++ b/UNRELEASED-V4.md
@@ -39,6 +39,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Documentation
 
+- Updated `AppProvider` app bridge example to use our `AppBridgeContext`([#1877](https://github.com/Shopify/polaris-react/pull/1877))
+
 ### Development workflow
 
 - Added support for React hooks in Storybook ([#1665](https://github.com/Shopify/polaris-react/pull/1665))


src/index.ts
```diff
@@ -11,5 +11,8 @@ export {
 export {
   WithinContentContext as _SECRET_INTERNAL_WITHIN_CONTENT_CONTEXT,
 } from './utilities/within-content-context';
+export {
+  AppBridgeContext as _SECRET_INTERNAL_APP_BRIDGE_CONTEXT,
```
AndrewMusgrave: I still don't feel the best about exposing this but if we're removing it in `v5` is not too big of a deal, I just don't want us to get in the habit of doing this 😬

BPScott: I guess the alternative is we say we're not going to expose this, and that people need to migrate to app-bridge-react before they move to v4

AndrewMusgrave: I also feel like that approach is a little off since we're deprecating app-bridge. I think this will be ok if we keep this to a minimal in the future 😄

BPScott: Given that we the appBridge context was intended to be public, and we document it and that it is required for people that want to use app-bridge I think this should have a regular name `AppBridgeContext` rather than a scary "if you need to do this then you're probably doing something wrong" name.


For contrast The WithinContentContext was never documented or intended to be public in v3, and we're only exporting it as `_SECRET_INTERNAL_WITHIN_CONTENT_CONTEXT` because some corners of web need it, and we've not yet got a plan for how to expose this in a more controlled manner.

src/components/AppProvider/README.md
```diff
@@ -481,22 +481,27 @@ To provide access to your initialized Shopify App Bridge instance, we make it av
...
   // This line is very important! It tells React to attach the `polaris`
   // object to `this.context` within your component.
-  static contextTypes = {
-    polaris: PropTypes.object,
-  };
+  static contextType = _SECRET_INTERNAL_APP_BRIDGE_CONTEXT;
 
-  redirectToSettings() {
+  componentDidMount() {
     const redirect = Redirect.create(this.context.polaris.appBridge);
```
BPScott: I think this line needs to be updated to:

```suggestion
    const redirect = Redirect.create(this.context);
```
 
https://reactjs.org/docs/context.html#classcontexttype suggests that this.context is set to the value of that particular context when using `contextType` (while the legacy API of `contextTypes` was an object that contained the polaris context, which was an object that contained the appBridge)

src/components/AppProvider/README.md
```diff
@@ -481,22 +481,27 @@ To provide access to your initialized Shopify App Bridge instance, we make it av
...
 import * as PropTypes from 'prop-types';
-import {AppProvider} from '@shopify/polaris';
+import {
+  AppProvider,
+  _SECRET_INTERNAL_APP_BRIDGE_CONTEXT,
+} from '@shopify/polaris';
 import {Redirect} from '@shopify/app-bridge/actions';
 
 class MyApp extends React.Component {
   // This line is very important! It tells React to attach the `polaris`
```
BPScott: This comment is no longer accurate as we're referring to a specific modern context instead of a legacy 'polaris' context object that contained everything.

As this is now idiomatic Context usage I'm tempted to just remove it entirely

AndrewMusgrave:
@BPScott Thanks for the throughout review, I was moving quickly yesterday and made a few mistakes

src/components/AppProvider/README.md
```diff
@@ -481,22 +481,25 @@ To provide access to your initialized Shopify App Bridge instance, we make it av
...
-  // object to `this.context` within your component.
-  static contextTypes = {
-    polaris: PropTypes.object,
-  };
+  static contextType = _SECRET_INTERNAL_APP_BRIDGE_CONTEXT;
 
-  redirectToSettings() {
-    const redirect = Redirect.create(this.context.polaris.appBridge);
+  componentDidMount() {
+    const redirect = Redirect.create(this.context.appBridge);
```
BPScott: the value of `this.context` /is/ the instance of appBridge

```suggestion
    const redirect = Redirect.create(this.context);
```

src/components/AppProvider/README.md
```diff
@@ -481,22 +481,25 @@ To provide access to your initialized Shopify App Bridge instance, we make it av
 import React from 'react';
 import {render} from 'react-dom';
 import * as PropTypes from 'prop-types';
-import {AppProvider} from '@shopify/polaris';
+import {
+  AppProvider,
+  _SECRET_INTERNAL_APP_BRIDGE_CONTEXT,
```
BPScott: You've updated the export name in the index to `AppBridgeContext` but not fixed it here.

PR #1879, opened 2019-07-24
[v4] Improve testing

BPScott:

Reducing duplication between our enzyme `mountWithAppProvider` and react-testing's `mountWithContext` (which I've renamed to `mountWithApp` as Context offers no clue on what sort of context).

Making PolarisTestProvider be able to handle all testing needs, so that mounting helpers can be super simplistic as all complicated stuff is handled by PolarisTestProvider.

Rename `mountWithContext` to `mountWithApp`  as Context offers no clue on what sort of context.

Reduce `mountWithAppProvider` and `mountWithApp` to be very trivial implementations that pass their options through to PolarisTestProvider and make PolarisTestProvider setup all contexts. Shoutout to a new feature in enzyme 3.10 that lets you specify wrapping components on mount which that really straightforward.

It is now up to the mounting functions to provide translations into PolarisTestProvider - Previously PolarisTestProvider tried to use english as a default and that meant that all the english translations got pulled into the standalone build (and thus bundle size increased). This is no longer the case.

Move `createThemeContext` and `createFrameContext` into PolarisTestProvider as that is now the only place they are used.

While doing this I checked what we currently passed to options in our mount helpers and found some things unneeded:

- PolarisTestProvider no longer allows you to overwrite StickyManager or ScrollLockManager. We never needed to pass these into `mountWithAppProvider`, and I highly doubt anybody else needed to either.
- PolarisTestProvider no longer deals with merging translations. We never needed to pass these into `mountWithAppProvider` and I highly doubt anybody else needed to either. It is up to the custom mounting function to provide a default set of translations.

### How to 🎩

Check tests and typechecking pass


src/components/Loading/tests/Loading.test.tsx
```diff
@@ -68,10 +67,6 @@ describe('<Loading />', () => {
 
 function mountWithAppBridge(element: React.ReactElement<any>) {
```
AndrewMusgrave: What do you think about removing this function?

src/test-utilities/legacy.tsx
```diff
@@ -74,47 +70,10 @@ function updateRoot(wrapper: AnyWrapper) {
...
-    stickyManager,
-    appBridge,
-    link,
-  };
-
-  return mount<P>(
-    <PolarisTestProvider {...context}>{node}</PolarisTestProvider>,
-  );
+  return mount<P>(node, {
+    wrappingComponent: PolarisTestProvider,
```
AndrewMusgrave: ...wow this is awesome 😍

src/test-utilities/react-testing.tsx
```diff
@@ -1,80 +1,26 @@
...
+  PolarisTestProvider,
+  WithPolarisTestProviderOptions,
+} from './PolarisTestProvider';
 
 export {mount};
 
-export const mountWithContext = createMount<
-  WithProvidersOptions,
-  WithProvidersContext
+export const mountWithApp = createMount<
```
AndrewMusgrave: I agree that `mountWithContext` may have not been descriptive enough, but I'm not sold on `app` since that not entirely what polaris is. How about `mountWithPolaris`?

BPScott: I was thinking drawing a parallel with `<AppProvider>`,  as mountWithApp is the equivalent of having an element inside `<AppProvider><Frame>BLAH</Frame></AppProvider>`.

mountWithPolaris would be better if we called that PolarisProvider, but alas that horse has bolted.

src/components/Loading/tests/Loading.test.tsx
BPScott: AppBridge stuff is deprecated and will be cleared away in v5, I'm tempted to leave this as-is to keep diff noise down and it'll all be washed away later.

PR #1880, opened 2019-07-24
[Page] Add thumbnail and subtitle to Page

dleroux:

Fixes https://github.com/Shopify/polaris-react/issues/1711 and adds a subtitle prop

Add 2 props to page: Thumbnail and Subtitle.

We opted with props for now until we get time to investigate the best approach to have more composable components.

## 

### How to 🎩

<details>

```jsx
import {PageDownMajorMonotone} from '@shopify/polaris-icons';
import {Page, Badge, Avatar} from '../src';

interface State {}

  render() {
    return (
      <Page
        breadcrumbs={[{content: 'Bills', url: '/settings/billing'}]}
        title="Invoice, but this page could have a longer title"
        subtitle="Statement period: May 3, 2019 to June 2, 2019"
        titleMetadata={<Badge status="success">Paid</Badge>}
        thumbnail={<Avatar customer />}
        secondaryActions={[{content: 'Download', icon: PageDownMajorMonotone}]}
        actionGroups={[
          {
            title: 'Promote',
            actions: [
              {
                content: 'Share on Facebook',
                accessibilityLabel: 'Individual action label',
                onAction: () => {},
              },
            ],
          },
        ]}
        pagination={{
          hasPrevious: true,
          hasNext: true,
        }}
        separator
      >
        <p>Page content</p>
      </Page>
    );
  }
}

```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -12,6 +12,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Added a `subtitle` and `thumbnail` prop to `Page` ([#1880](https://github.com/Shopify/polaris-react/pull/1880))
+
 ### Bug fixes
 
 - Fixed accessibility issue with ChoiceList errors not being correctly connected to the inputs ([#1824](https://github.com/Shopify/polaris-react/pull/1824));
--- a/src/components/Page/README.md
+++ b/src/components/Page/README.md
@@ -185,8 +185,15 @@ Use for detail pages, which should have pagination and breadcrumbs, and also oft
 ```jsx
 <Page
   breadcrumbs={[{content: 'Products', url: '/products'}]}
-  title="Jar With Lock-Lid"
-  titleMetadata={<Badge>Draft</Badge>}
+  title="3/4 inch Leather pet collar"
+  titleMetadata={<Badge status="success">Paid</Badge>}
+  subtitle="Perfect for any pet"
+  thumbnail={
+    <Thumbnail
+      source="https://burst.shopifycdn.com/photos/black-leather-choker-necklace_373x@2x.jpg"
+      alt="Black leather pet collar"
+    />
+  }
   primaryAction={{content: 'Save', disabled: true}}
   secondaryActions={[
     {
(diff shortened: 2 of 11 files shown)


src/components/Page/components/Header/components/PageTitle/PageTitle.scss
```diff
@@ -0,0 +1,29 @@
...
+
+.Title {
+  @include text-breakword;
+}
+
+.Title-hasMetadata {
+  margin-right: spacing(tight);
+  display: inline;
+
+  & > * {
```
dleroux: Since we want the metadata to be inline and DisplayText is block by default I needed to do this :(

Another option would be to simply make this an `H1` and style it myself

src/components/Page/components/Header/components/PageTitle/PageTitle.tsx
```diff
@@ -0,0 +1,70 @@
...
+  /** Page title, in large type */
+  title: string;
+  /** Page subtitle, in regular type*/
+  subtitle?: string;
+  /** Important and non-interactive status information shown immediately after the title. (stand-alone app use only) */
+  titleMetadata?: React.ReactNode;
+  /** thumbnail that precedes the title */
+  thumbnail?:
+    | React.ReactElement<AvatarProps | ThumbnailProps>
+    | React.SFC<React.SVGProps<SVGSVGElement>>;
```
dleroux: How is this for valid types?

src/components/Page/components/Header/components/PageTitle/PageTitle.scss
```diff
@@ -0,0 +1,38 @@
...
+  }
+}
+
+.TitleWithMetadataWrapper {
+  .Title {
+    display: inline;
+    margin-right: spacing(tight);
+
+    // stylelint-disable-next-line selector-max-combinators
+    > * {
```
dleroux: Display block on `DisplayText` needs to `display: inline` if there's meta data

chloerice: I like this a lot. Much better than metadata wrapping to a new line like it does currently. This makes it actually behave like it says it does. 

|  Before | After  |
|---|---|
|metadata-originally-wrapped-to-new-line|metadata-only-SMALL|

src/components/Page/components/Header/components/PageTitle/PageTitle.tsx
chloerice: 💯

PR #1887, opened 2019-07-26
[Autocomplete]Fix keyboard navigation and item selection in Autocomplete combo box.

### LINKED ISSUE #1346: [Autocomplete][v1] Keyboard navigation is laggy and skips options
# Issue summary

If you navigate up / down it will sometimes feel laggy and even skip an element.

## Expected behavior

Each keystroke should navigate to the next option within a reasonable amount of time.

## Actual behavior

CF summary

1. Visit https://polaris.shopify.com/components/forms/autocomplete
1. Click on the search example
1. Try to navigate up and down with your keyboard within options

## Reduced test case

https://codesandbox.io/s/k9446943k3?module=App.js

## Specifications

- Polaris version number: 3.13.0
- Browser: Chrome (latest)
- Device: Macbook Pro
- Operating System: MAC OS

### LINKED ISSUE #1380: [Autocomplete][v1] the enter keystroke event should not bubble up
# Issue summary

When using the keyboard to select an option; pressing enter will trigger the parent form's onSubmit. I think we should call `stopPropagation()` of the event so it does not bubble up to the form.

## Expected behavior

Select the option and **don't** trigger the for submission.

## Actual behavior

Select the option and trigger the for submission.

1. Click on the field
1. Navigate with your keyboard
1. Press enter

## Reduced test case

See https://codesandbox.io/s/oqmom5qrxz

## Specifications

- Are you using the React components? (Y/N): yes
- Polaris version number: 3.13
- Browser: Chrome
- Device: Mac Book Pro
- Operating System: Mac OS

rox163:
Need to fix keyboard navigation and item selection in Autocomplete combo box.
Fixes #1380. Fixes #1346  

Added an EventListener for the `keydown` event and only use the handler to track the Enter key events. This prevents any `submit` events from being triggered on Enter key press inside the combo box. `stopPropagation()` does not seem to work in this case. 
https://stackoverflow.com/questions/895171/prevent-users-from-submitting-a-form-by-hitting-enter  - `The keydown event is preferred over keyup as the keyup is too late to block form submit`

## 

### How to 🎩

1. Use the keyboard to navigate the options in the combo box. See the visual highlight of the options not skip any options. You can use Storybook in [this deployment](https://polaris-react-pr-1887.herokuapp.com/?path=/story/playground-playground--playground) to test it. 

2. Using the Playground code, Use the Enter key to select an option. Option should be selected without any logs in the console. Console log has been added to the outer `<form>` onSubmit action. Compare to the code [here](https://codesandbox.io/s/oqmom5qrxz) where the current production version logs an event when Enter key is pressed to select an option. 

<details>

```jsx
import {Page, Autocomplete, Icon, TextContainer} from '../src';

interface State {}

  options = [
    {value: 'rustic', label: 'Rustic'},
    {value: 'antique', label: 'Antique'},
    {value: 'vinyl', label: 'Vinyl'},
    {value: 'vintage', label: 'Vintage'},
    {value: 'refurbished', label: 'Refurbished'},
  ];

  state = {
    selected: [],
    inputText: '',
    options: this.options,
    loading: false,
  };

  render() {
    const textField = (
      <Autocomplete.TextField
        onChange={this.updateText}
        label="Tags"
        value={this.state.inputText}
        prefix={<Icon source={SearchMinor} color="inkLighter" />}
        placeholder="Search"
      />
    );
    const emptyState = (
      <React.Fragment>
        <Icon source={SearchMinor} />
        <div style={{textAlign: 'center'}}>
          <TextContainer>Could not find any results</TextContainer>
        </div>
      </React.Fragment>
    );
    return (
      <Page title="Playground">
        <form
          style={{height: '225px'}}
          onSubmit={(err) => {
            err.preventDefault();
            console.log('Submit event');
          }}
        >
          <Autocomplete
            options={this.state.options}
            emptyState={emptyState}
            loading={this.state.loading}
            textField={textField}
          />
        </form>
      </Page>
    );
  }

  updateText = (newValue) => {
  };

  filterAndUpdateOptions = (inputString) => {
    if (!this.state.loading) {
      this.setState({loading: true});
    }

    setTimeout(() => {
      if (inputString === '') {
        this.setState({
          options: this.options,
          loading: false,
        });
        return;
      }
      );
      this.setState({
        options: resultOptions,
        loading: false,
      });
    }, 300);
  };

  updateSelection = (selected) => {
    const selectedText = selected.map((selectedItem) => {
      const matchedOption = this.options.find((option) => {
        return option.value.match(selectedItem);
      });
      return matchedOption && matchedOption.label;
    });
    this.setState({selected, inputText: selectedText});
  };
}

```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -48,6 +48,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Fixed accessibility issue with `Tabs` list item presentation role ([#1958](https://github.com/Shopify/polaris-react/pull/1958))
 - Removed `Tooltip` on disabled `Pagination` buttons ([#1963](https://github.com/Shopify/polaris-react/pull/1963))
 - Fixed accessibility labels on `ResourceList.Item` persistent action disclosure icon ([#1973](https://github.com/Shopify/polaris-react/pull/1973))
+- Fixed accessibility issue with `Autocomplete` where keyboard navigation of options was laggy and skipped options([#1887](https://github.com/Shopify/polaris-react/pull/1887))
+- Fixed bug where `Autocomplete` was bubbling up the `Enter` key event unexpectedly ([#1887](https://github.com/Shopify/polaris-react/pull/1887))
 
 ### Documentation
 
--- a/src/components/Autocomplete/components/ComboBox/ComboBox.tsx
+++ b/src/components/Autocomplete/components/ComboBox/ComboBox.tsx
@@ -6,6 +6,7 @@ import Popover from '../../../Popover';
 import {PreferredPosition} from '../../../PositionedOverlay';
 import {ActionListItemDescriptor, Key} from '../../../../types';
 import KeypressListener from '../../../KeypressListener';
+import EventListener from '../../../EventListener';
 import {ComboBoxContext} from './context';
 
 import styles from './ComboBox.scss';
@@ -238,7 +239,7 @@ export default class ComboBox extends React.PureComponent<Props, State> {
             keyCode={Key.UpArrow}
             handler={this.handleUpArrow}
           />
-          <KeypressListener keyCode={Key.Enter} handler={this.handleEnter} />
+          <EventListener event="keydown" handler={this.handleEnter} />
           <KeypressListener
             keyCode={Key.Escape}
             handler={this.handlePopoverClose}
@@ -281,18 +282,20 @@ export default class ComboBox extends React.PureComponent<Props, State> {
     this.handlePopoverOpen;
   };
 
-  private handleEnter = () => {
-    const {selectedOption} = this.state;
-
-    if (this.state.popoverActive && selectedOption) {
-      if (isOption(selectedOption)) {
-        this.handleSelection(selectedOption.value);
-      } else {
-        selectedOption.onAction && selectedOption.onAction();
+  private handleEnter = (event: KeyboardEvent) => {
+    if (event.keyCode === Key.Enter) {
+      const {selectedOption} = this.state;
+      if (this.state.popoverActive && selectedOption) {
+        if (isOption(selectedOption)) {
+          event.preventDefault();
+          this.handleSelection(selectedOption.value);
+        } else {
+          selectedOption.onAction && selectedOption.onAction();
+        }
       }
-    }
 
-    this.handlePopoverOpen;
+      this.handlePopoverOpen;
+    }
   };
 
   private handleFocus = () => {
@@ -407,21 +410,26 @@ export default class ComboBox extends React.PureComponent<Props, State> {
   };
 
   private selectOptionAtIndex = (newOptionIndex: number) => {
-    const {navigableOptions, selectedOption: oldSelectedOption} = this.state;
-    if (!navigableOptions || navigableOptions.length === 0) {
-      return;
-    }
-    const newSelectedOption = navigableOptions[newOptionIndex];
+    this.setState((prevState) => {
+      if (
+        !prevState.navigableOptions ||
+        prevState.navigableOptions.length === 0
+      ) {
+        return prevState;
+      }
+      const newSelectedOption = prevState.navigableOptions[newOptionIndex];
 
-    this.setState(
-      {
+      this.visuallyUpdateSelectedOption(
+        newSelectedOption,
+        prevState.selectedOption,
+      );
+
+      return {
+        ...prevState,
         selectedOption: newSelectedOption,
         selectedIndex: newOptionIndex,
-      },
-      () => {
-        this.visuallyUpdateSelectedOption(newSelectedOption, oldSelectedOption);
-      },
-    );
+      };
+    });
   };
 
   private visuallyUpdateSelectedOption = (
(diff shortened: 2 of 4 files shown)


src/components/Autocomplete/components/ComboBox/ComboBox.tsx
```diff
@@ -430,21 +434,26 @@ export default class ComboBox extends React.PureComponent<Props, State> {
   };
 
   private selectOptionAtIndex = (newOptionIndex: number) => {
-    const {navigableOptions, selectedOption: oldSelectedOption} = this.state;
-    if (!navigableOptions || navigableOptions.length === 0) {
-      return;
-    }
-    const newSelectedOption = navigableOptions[newOptionIndex];
+    this.setState((prevState) => {
```
rox163: This change helps fix the selected index skipping issue by making sure when we use the previous to determine new state, its in the setState function.

UNRELEASED.md
```diff
@@ -14,6 +14,9 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- Fixed accessibility issue with `Autocomplete` where keyboard navigation of options was laggy and skipped options([#1346](https://github.com/Shopify/polaris-react/issues/1346))
```
```suggestion
- Fixed accessibility issue with `Autocomplete` where keyboard navigation of options was laggy and skipped options([#1887](https://github.com/Shopify/polaris-react/pull/1887))
```

UNRELEASED.md
```diff
@@ -14,6 +14,9 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
...
+- Fixed bug where `Autocomplete` was bubbling up the `Enter` key event unexpectedly ([#1380](https://github.com/Shopify/polaris-react/issues/1380))
```
```suggestion
- Fixed bug where `Autocomplete` was bubbling up the `Enter` key event unexpectedly ([#1887]https://github.com/Shopify/polaris-react/pull/1887)
```

UNRELEASED.md
AndrewMusgrave: We reference the pull request that made the changes 😄

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
```diff
@@ -304,18 +306,20 @@ export default class ComboBox extends React.PureComponent<Props, State> {
...
-  private handleEnter = () => {
-    const {selectedOption} = this.state;
-
-    if (this.state.popoverActive && selectedOption) {
-      if (isOption(selectedOption)) {
-        this.handleSelection(selectedOption.value);
-      } else {
-        selectedOption.onAction && selectedOption.onAction();
+  private handleEnter = (event: KeyboardEvent) => {
+    if (event.keyCode === Key.Enter) {
```
AndrewMusgrave: What do you think about having an early return if `if (event.keyCode !== Key.Enter) return;`so we don't have the entire function in a condition?

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
```diff
@@ -569,3 +585,8 @@ function dispatchKeyup(key: Key) {
   const event: KeyboardEventInit & {keyCode: Key} = {keyCode: key};
   document.dispatchEvent(new KeyboardEvent('keyup', event));
 }
+
+function dispatchKeydown(key: Key) {
```
AndrewMusgrave: If we changed the above function signature to 
```
function dispatchKeyBoardEvent(key: Key, keyBoardEvent: string = 'keyup') {
```
we could reduce the code slightly.

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
```diff
@@ -435,10 +435,26 @@ describe('<ComboBox/>', () => {
       );
       comboBox.find(TextField).simulate('click');
       dispatchKeyup(Key.DownArrow);
-      dispatchKeyup(Key.Enter);
```
AndrewMusgrave: Should we add a regression test as well to assert the event doesn't propagate to the forms on submit?

UNRELEASED.md
rox163: oops! haha! sorry! 🤦‍♀

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
rox163: Good point. I feel early returns is more along the lines of our patterns 👍

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
rox163: The issue here is that EventListener works off `window` events whereas KeyboardListener works off `document` events. So not sure its worth trying to make this more generic

AndrewMusgrave: Gotcha didn't notice that. Ignore this comment 🙈

src/components/Autocomplete/components/ComboBox/tests/ComboBox.test.tsx
rox163: If its ok to have a test where the ComboBox is wrapped in `<form>` (since thats where the oSubmit call is made),  then yep i can add tests for those

rox163:
> Awesome stuff looks really good! Thanks for fixing these 😄 Left a few comments, mostly nits. I did notice that pressing enter in the `TextField` still calls the forms `onSubmit`, however, I assume that was intentional/

Yep I felt that makes sense from the ComboxBox perspective so I didnt change it. One press of enter to select an option and the next press of enter triggers the onSubmit as expected.

Issue #1898, opened 2019-07-30
[v4] Ensure no errors/warnings when running tests


BPScott:
Currently running `yarn run test` has some big red text in it. We should fixup these errors and warnings to ensure that our tests are working in the way that we expect.

### Expected Behaviour:

All passes, no errors or warnings raised.

### Actual Behaviour: 

All passes but two warnings (no big red coloring here but just pretend):

```
 PASS  src/components/Navigation/components/Item/tests/Item.test.tsx
  ● Console
    console.error tests/setup.ts:39      Warning: An update to Item inside a test was not wrapped in act(...).
            When testing, code that causes React state updates should be wrapped into act(...):
      
      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */
      
      This ensures that you're testing the behavior the user would see in the browser. Learn more at https://fb.me/react-wrap-tests-with-act
          in Item (at Item.test.tsx:449)
          in RootFinder (created by WrapperComponent)
          in PolarisTestProvider (created by WrapperComponent)
          in WrapperComponent
```

and
```
 PASS  src/components/Tabs/components/Item/tests/Item.test.tsx
  ● Console

    console.error tests/setup.ts:39
      Warning: Function components cannot be given refs. Attempts to access this
 ref will fail. Did you mean to use React.forwardRef()?
      
      Check the render method of `Item`.
          in UnstyledLink (at Item.tsx:58)
          in li (at Item.tsx:67)
          in Item (created by WrapperComponent)
          in RootFinder (created by WrapperComponent)
          in PolarisTestProvider (created by WrapperComponent)
          in WrapperComponent
```

AndrewMusgrave:
For additional context the `Navigation` test's error might be from an [outside source](https://github.com/Shopify/polaris-react/blob/673f08e393e702f761267045492ba65d9e72ca1f/src/components/Navigation/components/Item/tests/Item.test.tsx#L64)

BPScott:
#1900 has fixed the warning in the Navigation Item test

PR #1900, opened 2019-07-31
Use act in Navigation Item test to fix warning

BPScott:

Fixes one of the two warnings in #1898

Wraps a trigger in act.

Migrating this whole file to use modern testing looked like it'd take a while so here's a stopgap.

### How to 🎩

Check Navigation Item test no longer emits the warning detailed in #1898


BPScott:
Yeah,  I figure a quick and dirty fix is enough for this for legacy, and when we move over to react-testing I expect their built-in trigger function to be able to deal with `act`ing.

PR #1906, opened 2019-07-31
[v4] Update frame utilities & associated components

AndrewMusgrave:

`startLoading` and `stopLoading` are never going to change, however `frame` will change everytime `Frame` renders again.  I'm removing frame from the dependency array to reduce unnecessary renders. 

## 

### How to 🎩

Check out loading examples in the deployment 🎉


BPScott:
FrameContext is a boring old object from what I can tell. In that case what if we destructure startLoading and stopLoading out of it and then pass those functions to useEffect's dependencies

```
  const {startLoading, stopLoading} = useContext(FrameContext);
``

Then your dependencies would be `[appBridge, startLoading, stopLoading]` and you'd avoid the need for the disable the linting.


I'm not too familar with useEffect' dependency stuff, would that work?

**Comment by AndrewMusgrave** [comment]
It would, however, `frame` context is initialized as `null` so we can't deconstruct the property until we check for frames existence

**Comment by BPScott** [comment]
Would `const {startLoading, stopLoading} = useContext(FrameContext) || {};` do the trick?

Longer term plans alternative: Should we make the default for FrameContext be an object bunch of noop functions so you never need to check if the value from the context is undefined? Or perhaps we could follow the same concepts as what we have for the contexts provided by AppProvider:

- FrameContext is undefined by default
- Have a useFrame() hook that reads the context and will throw a "you must wrap this component in `<Frame>` error if it undefined, thus meaning useFrame will always return an object
- Update Loading, Toast and ContextualSaveBar to use that new hook

**Comment by BPScott** [comment]
Playing with this locally:

```
type PartialFrameContext = Partial<
  NonNullable<React.ContextType<typeof FrameContext>>
>;

const {startLoading, stopLoading} =
    useContext(FrameContext) || ({} as PartialFrameContext);
```

Looks like it does the job

**Comment by AndrewMusgrave** [comment]
> Looks like it does the job as a dirty stopgap

I think it's a little hacky, I would rather us fix it properly now since it's a rather small change 😄

File: src/components/Loading/Loading.tsx [hunk 1]
```diff
@@ -1,19 +1,19 @@
...
-  const frame = useContext(FrameContext);
+  const {startLoading, stopLoading} = useFrame();
 
   useEffect(
     () => {
-      if (appBridge == null && frame) {
-        frame.startLoading();
+      if (appBridge == null) {
+        startLoading();
       } else if (appBridge != null) {
```
BPScott: If frame is always defined, then we can get ride of the if here and just have `} else {`

src/index.ts
```diff
@@ -11,5 +11,6 @@ export {
   WithinContentContext as _SECRET_INTERNAL_WITHIN_CONTENT_CONTEXT,
 } from './utilities/within-content-context';
 export {AppBridgeContext} from './utilities/app-bridge';
+export {ContextualSaveBarProps, ToastProps} from './utilities/frame';
```
BPScott: I dislike the idea of exporting things from utilities within the index, as I think the relationship between what is ToastProps used for is made blury here.

I think what these props are used for is made much clearer if we export `Props` from `src/components/ContextualSaveBar` via the components index - that way the component and its props are kept together.

src/test-utilities/PolarisTestProvider.tsx
```diff
@@ -1,6 +1,5 @@
 import React from 'react';
-// eslint-disable-next-line shopify/strict-component-boundaries
-import {FrameContext, FrameContextType} from '../components/Frame';
+import {FrameContext, FrameContextType} from '../utilities/frame';
```
BPScott: Can we stop exposing the `FrameContextType` and instead use the `React.ContextType<typeof FrameContext>` inference instead to keep our API surface area small.

src/utilities/frame/context.tsx
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import {createContext} from 'react';
 import {ToastPropsWithID, ToastID, ContextualSaveBarProps} from './types';
 
 export interface FrameContextType {
```
BPScott: Can we stop exposing the `FrameContextType` and instead use the `React.ContextType<typeof FrameContext>` inference instead to keep our API surface area small.

src/utilities/frame/hooks.tsx
```diff
@@ -0,0 +1,27 @@
+import {useContext} from 'react';
+import {useAppBridge} from '../app-bridge';
+import {FrameContext} from './context';
+
+export function useFrame() {
+  const frame = useContext(FrameContext);
+  const appBridge = useAppBridge();
+
+  if (!frame && !appBridge) {
```
BPScott: Should this be or rather than and?

If either the Frame or the AppBridge is defined then you can use this

src/utilities/frame/hooks.tsx
```diff
@@ -0,0 +1,27 @@
...
+    throw new Error(
+      'Your application must be wrapped in an <Frame> component to use: Loading, Toast, or ContextualSaveBar. See https://polaris.shopify.com/components/structure/frame for implementation instructions.',
```
BPScott: I don't think this should mention the Loading/Toast/ContextualSaveBar components explicitly - this is a level above those components and we might add more in the future.

Something like:

```
 'No Frame context was provided. Your component must be wrapped in a <Frame> component, or be used within an embedded application by setting the apiKey and shopOrigin properties on <AppProvider>. See https://polaris.shopify.com/components/structure/frame for implementation instructions.',
```

src/utilities/frame/hooks.tsx
```diff
@@ -0,0 +1,27 @@
...
+    );
+  }
+
+  return (
+    frame || {
```
BPScott: If we're specifying a default here, then we might as well use that default for FrameContext, and thus we don't need to check if value is undefined and potentially throw an error (as it will always be set).

PR #1911, opened 2019-08-01
Move PolarisTestProvider into components

AndrewMusgrave:

Our `test-utilities` folder is being [deleted in `web`](https://github.com/Shopify/web/blob/68750baa3c14bf7390045d260430309d37cdde12/scripts/builder/delete-test-files.sh#L6). I'm temporarily moving `PolarisTestProvider` into `src/components` until node_modules are ignored.

### DIFF
--- a/src/components/PolarisTestProvider/tests/PolarisTestProvider.test.tsx
+++ b/src/components/PolarisTestProvider/tests/PolarisTestProvider.test.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import {mount} from '@shopify/react-testing';
-import {PolarisTestProvider} from '../PolarisTestProvider';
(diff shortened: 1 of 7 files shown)


src/components/PolarisTestProvider/tests/PolarisTestProvider.test.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import {mount} from '@shopify/react-testing';
-import {PolarisTestProvider} from '../PolarisTestProvider';
+import {PolarisTestProvider} from '../..';
```
BPScott: This  doesn't seem right. Don't import from indexes

`import PolarisTestProvider from '../PolarisTestProvider';`

PR #1912, opened 2019-08-02
[v4] Styles build: combine all components into one file

BPScott:

There is no value in having individual files per component in the styles
folder. It is unneeded overhead and people shouldn't be digging into
that folder trying to include a single component's styles in the first place.

Where styles/components.scss previously was a list imports to
per-component files, instead inline the contents of the per-component scss
into styles/components.scss. The listing of the contents is preserved so
the computed css out the other end will be idential.

At this point global.scss and components.scss both contain pure css.
They are only distributed as scss files for the sake of convinience and
ease of @importing (I have future plans to potentially simplify this further but that's a story for another day).

### How to 🎩

Run `yarn run build` and see the `styles/components` folder is no longer generated. See that `styles/components.scss` now includes the contents of  each components' and sub-components' css, with the classnames being the long ones generated using css-modules (note they all start with `Polaris-`.

Check that this works in web's development mode by doing `yarn run build-consumer web` then starting up the web server and browsing arround seeing that the css is still rendered.

### DIFF
--- a/config/rollup/plugins/styles.js
+++ b/config/rollup/plugins/styles.js
@@ -170,35 +170,26 @@ async function generateSass(inputFolder, outputFolder, cssByFile) {
     }),
   );
 
-  // For each file that was referenced in the build, create a copy of the compiled
-  // ouput
-  const componentsCssFilenames = [];
-  cssByFile.forEach(async (compiledCss, filename) => {
+  const writeStylesFolderFilesPromises = [];
+  const componentFilesContent = [];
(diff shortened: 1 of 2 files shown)


BPScott:
> This is great @BPScott love how simple this has become.

I am plotting things for Hackdays and if it goes well you aint seen nuffin yet :D

BPScott:
Yo @beefchimi, your online-store-ui build is pretty similar to ours, this bit of simplification might be up your alley.

PR #1917, opened 2019-08-05
[Color system] Build color variant generation factory

danrosenthal:
Addresses https://github.com/Shopify/polaris-react/issues/2034

### DIFF
--- a/src/components/ThemeProvider/ThemeProvider.tsx
+++ b/src/components/ThemeProvider/ThemeProvider.tsx
@@ -6,6 +6,7 @@ import {
   buildCustomProperties,
 } from '../../utilities/theme';
 import {themeProvider} from '../shared';
+import {useFeatures} from '../../utilities/features';
 
 interface ThemeProviderProps {
   /** Custom logos and colors provided to select components */
@@ -18,10 +19,12 @@ export function ThemeProvider({
   theme: themeConfig,
   children,
 }: ThemeProviderProps) {
+  const {globalTheming} = useFeatures();
   const theme = useMemo(() => buildThemeContext(themeConfig), [themeConfig]);
-  const customProperties = useMemo(() => buildCustomProperties(themeConfig), [
-    themeConfig,
-  ]);
+  const customProperties = useMemo(
+    () => buildCustomProperties(themeConfig, globalTheming),
+    [globalTheming, themeConfig],
+  );
 
   return (
     <ThemeContext.Provider value={theme}>
--- a/src/components/ThemeProvider/ThemeProvider.tsx
+++ b/src/components/ThemeProvider/ThemeProvider.tsx
@@ -19,11 +19,11 @@ export function ThemeProvider({
   theme: themeConfig,
   children,
 }: ThemeProviderProps) {
-  const {globalTheming} = useFeatures();
+  const {unstableGlobalTheming = false} = useFeatures();
   const theme = useMemo(() => buildThemeContext(themeConfig), [themeConfig]);
   const customProperties = useMemo(
-    () => buildCustomProperties(themeConfig, globalTheming),
-    [globalTheming, themeConfig],
+    () => buildCustomProperties(themeConfig, unstableGlobalTheming),
+    [unstableGlobalTheming, themeConfig],
   );
 
   return (
--- a/src/utilities/theme/tests/utils.test.ts
+++ b/src/utilities/theme/tests/utils.test.ts
@@ -71,11 +71,11 @@ describe('buildCustomProperties', () => {
     const colors = buildCustomProperties(theme, true);
     expect(colors).toStrictEqual(
       expect.objectContaining({
-        '--surface': 'hsl(0, 0%, 100%, 1)',
-        '--surface-background': 'hsl(0, 0%, 98%, 1)',
-        '--surface-foreground': 'hsl(0, 0%, 100%, 1)',
-        '--surface-foreground-subdued': 'hsl(0, 0%, 90%, 1)',
-        '--surface-inverse': 'hsl(0, 0%, 0%, 1)',
+        '--p-surface': 'hsl(0, 0%, 100%, 1)',
+        '--p-surface-background': 'hsl(0, 0%, 98%, 1)',
+        '--p-surface-foreground': 'hsl(0, 0%, 100%, 1)',
+        '--p-surface-foreground-subdued': 'hsl(0, 0%, 90%, 1)',
+        '--p-surface-inverse': 'hsl(0, 0%, 0%, 1)',
         '--top-bar-background': '#eeeeee',
         '--top-bar-background-lighter': 'hsl(0, 10%, 100%, 1)',
         '--top-bar-color': 'rgb(33, 43, 54)',
@@ -89,11 +89,11 @@ describe('Colors', () => {
     it('has a default value', () => {
       expect(Colors({colors: {}})).toStrictEqual(
         expect.objectContaining({
-          '--surface': 'hsl(0, 0%, 98%, 1)',
-          '--surface-background': 'hsl(0, 0%, 98%, 1)',
-          '--surface-foreground': 'hsl(0, 0%, 100%, 1)',
-          '--surface-foreground-subdued': 'hsl(0, 0%, 90%, 1)',
-          '--surface-inverse': 'hsl(0, 0%, 0%, 1)',
+          '--p-surface': 'hsl(0, 0%, 98%, 1)',
+          '--p-surface-background': 'hsl(0, 0%, 98%, 1)',
+          '--p-surface-foreground': 'hsl(0, 0%, 100%, 1)',
+          '--p-surface-foreground-subdued': 'hsl(0, 0%, 90%, 1)',
+          '--p-surface-inverse': 'hsl(0, 0%, 0%, 1)',
         }),
       );
     });
@@ -101,11 +101,11 @@ describe('Colors', () => {
     it('creates variants when given a light surface', () => {
       expect(Colors({colors: {surface: '#ffffff'}})).toStrictEqual(
         expect.objectContaining({
-          '--surface': 'hsl(0, 0%, 100%, 1)',
-          '--surface-background': 'hsl(0, 0%, 98%, 1)',
-          '--surface-foreground': 'hsl(0, 0%, 100%, 1)',
-          '--surface-foreground-subdued': 'hsl(0, 0%, 90%, 1)',
-          '--surface-inverse': 'hsl(0, 0%, 0%, 1)',
+          '--p-surface': 'hsl(0, 0%, 100%, 1)',
+          '--p-surface-background': 'hsl(0, 0%, 98%, 1)',
+          '--p-surface-foreground': 'hsl(0, 0%, 100%, 1)',
+          '--p-surface-foreground-subdued': 'hsl(0, 0%, 90%, 1)',
+          '--p-surface-inverse': 'hsl(0, 0%, 0%, 1)',
         }),
       );
     });
@@ -113,11 +113,11 @@ describe('Colors', () => {
     it('creates variants when given a dark surface', () => {
       expect(Colors({colors: {surface: '#000000'}})).toStrictEqual(
         expect.objectContaining({
-          '--surface': 'hsl(0, 0%, 0%, 1)',
-          '--surface-background': 'hsl(0, 0%, 7%, 1)',
-          '--surface-foreground': 'hsl(0, 0%, 13%, 1)',
-          '--surface-foreground-subdued': 'hsl(0, 0%, 10%, 1)',
-          '--surface-inverse': 'hsl(0, 0%, 100%, 1)',
+          '--p-surface': 'hsl(0, 0%, 0%, 1)',
+          '--p-surface-background': 'hsl(0, 0%, 7%, 1)',
+          '--p-surface-foreground': 'hsl(0, 0%, 13%, 1)',
+          '--p-surface-foreground-subdued': 'hsl(0, 0%, 10%, 1)',
+          '--p-surface-inverse': 'hsl(0, 0%, 100%, 1)',
         }),
       );
     });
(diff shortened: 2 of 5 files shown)


BPScott:
Tagging myself for review as I'm excited to learn more and learn about how this fits together

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -1,12 +1,14 @@
 import React from 'react';
 import isEqual from 'lodash/isEqual';
 import {ThemeContext} from '../../utilities/theme';
-import {Theme} from '../../utilities/theme/types';
-import {setColors} from '../../utilities/theme/utils';
+import {Theme, CSSProperties} from '../../utilities/theme/types';
```
danrosenthal: Per @tmlayton's point, this `CSSProperties` typing is not great.

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -1,12 +1,14 @@
...
+import {setColors} from '../../utilities/theme/legacy-utils';
```
danrosenthal: I'm name-spacing all the `topBar` theming concerns under legacy to optimize for deletion while preserving backwards compatibility.

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -1,12 +1,14 @@
...
+import {Colors} from '../../utilities/theme/utils';
```
danrosenthal: This is the new variant generation factory.

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -55,20 +62,26 @@ export class ThemeProvider extends React.Component<ThemeProviderProps, State> {
 
     return (
       <ThemeContext.Provider value={theme}>
-        <div style={styles}>{children}</div>
+        <div style={legacyStyles}>
```
danrosenthal: Adding an extra div with legacy `topBar` styles feels like a nice separation of concerns vs. cramming these all on one div.

src/utilities/color-manipulation.ts
```diff
@@ -8,7 +9,7 @@ export function lightenColor(color: HSLColor | string, lighten = 0) {
   const {lightness} = color;
   const nextLightness = lightness + lighten;
 
-  return {...color, lightness: nextLightness};
+  return {...color, lightness: clamp(nextLightness, 0, 100)};
```
danrosenthal: This fixes a bug where invalid hex colors could be returned if a lightness was outside the valid range.

It could use a regression test.

src/utilities/color-manipulation.ts
```diff
@@ -19,7 +20,20 @@ export function darkenColor(color: HSLColor | string, lighten = 0) {
   const {lightness} = color;
   const nextLightness = lightness - lighten;
 
-  return {...color, lightness: nextLightness};
+  return {...color, lightness: clamp(nextLightness, 0, 100)};
```
danrosenthal: This fixes a bug where invalid hex colors could be returned if a lightness was outside the valid range.

It could use a regression test.

src/utilities/color-transformers.ts
```diff
@@ -147,10 +148,11 @@ function rgbToHsbl(color: RGBAColor, type: 'b' | 'l' = 'b'): HSBLAColor {
   } else if (type === 'b') {
     saturation = delta / largestComponent;
   } else if (type === 'l') {
-    saturation =
+    const baseSaturation =
       lightness > 0.5
         ? delta / (2 - largestComponent - smallestComponent)
         : delta / (largestComponent + smallestComponent);
+    saturation = isNaN(baseSaturation) ? 0 : baseSaturation;
```
danrosenthal: This fixes a bug where saturation could end up as `NaN` and would break color conversion.

It could use a regression test.

src/utilities/color-types.ts
```diff
@@ -36,3 +36,7 @@ export interface HSLAColor extends HSLColor {
   /** Level of transparency */
   alpha: number;
 }
+
+export interface MaybeAlphaHSLAColor extends Omit<HSLAColor, 'alpha'> {
+  alpha?: number;
+}
```
danrosenthal: This ends up being a better typing for the `hslToString` method, as `alpha` is only required for `hsla` notation.

src/utilities/theme/legacy-utils.tsx
```diff
@@ -0,0 +1,117 @@
+import tokens from '@shopify/polaris-tokens';
```
danrosenthal: These are copied over from `utils`, with the only modifications being some type casting and default values to account for the `topBar` theme property going from required to optional.

src/utilities/theme/legacy-utils.tsx
```diff
@@ -0,0 +1,117 @@
...
+import {compose} from '../compose';
+
+import {Theme, ColorsToParse, ThemeVariant, ThemeColors, TopBar} from './types';
+import {needsVariantList} from './legacy-config';
+
+export function setColors(theme: Theme | undefined): string[][] | undefined {
+  let colorPairs;
+  if (theme && theme.colors) {
+    Object.entries(theme.colors).forEach(([colorKey, pairs]) => {
+      const colorKeys = Object.keys(pairs as TopBar);
```
danrosenthal: Typecasting added here.

src/utilities/theme/legacy-utils.tsx
```diff
@@ -0,0 +1,117 @@
...
+      if (colorKey === 'topBar' && colorKeys.length > 1) {
+        colorPairs = colorKeys.map((key: string) => {
+          const {topBar} = theme.colors as ThemeColors;
+          const defaultTopBar: TopBar = {background: ''};
+          const colors = topBar ? topBar : defaultTopBar;
```
danrosenthal: Default topBar added here. In practice, this could would never be reached without a defined `topBar`, but keeping it required has negative impacts on how we write the replacement code.

src/utilities/theme/legacy-utils.tsx
```diff
@@ -0,0 +1,117 @@
...
+          return [constructColorName(colorKey, key), colors[key]];
+        });
+      } else {
+        colorPairs = parseColors([colorKey, pairs as TopBar]);
+      }
```
danrosenthal: Typecasting added here as well. In practice this would never be reached, as other theme properties would not be set when `topBar` was set.

src/utilities/theme/tests/utils.test.ts
```diff
@@ -1,53 +1,120 @@
...
+      '--text-on-dark': 'hsl(210, 9%, 100%, 1)',
+      '--text-on-light': 'hsl(210, 9%, 13%, 1)',
+      '--text-on-opposite': 'hsl(210, 9%, 100%, 1)',
+      '--text-on-surface': 'hsl(210, 9%, 13%, 1)',
+      '--warning': 'hsl(39, 100%, 66%, 1)',
+      '--warning-divider': 'hsl(39, 100%, 66%, 1)',
+      '--warning-icon': 'hsl(39, 100%, 66%, 1)',
+      '--warning-surface': 'hsl(39, 100%, 88%, 1)',
+      '--warning-surface-muted': 'hsl(39, 100%, 98%, 1)',
+      '--warning-text': 'hsl(39, 100%, 30%, 1)',
```
danrosenthal: These values will more than likely change, but testing the output doesn't hurt now.

src/utilities/theme/types.ts
```diff
@@ -17,12 +17,24 @@ export interface ThemeColor {
   [key: string]: string;
 }
 
+export type CSSProperties = {
+  [key: string]: string;
+};
```
danrosenthal: Actually typing for CSS Custom Property syntax would be great.

src/utilities/theme/utils.ts
```diff
@@ -1,115 +1,779 @@
...
+export function Colors(theme: Theme) {
+  const {colors = {}} = theme;
+
+  const isLightTheme = isLight(
+    hslToRgb(colorToHsla(
+      colors == null || colors.surface == null
+        ? DefaultColor.Surface
+        : colors.surface,
+    ) as HSLColor),
+  );
```
danrosenthal: This is a critical piece, as it looks at the surface value and determines whether it is light or dark. Based off this, the generated color variants can respond to whether the applied theme represents light mode or dark mode.

src/utilities/theme/utils.ts
```diff
@@ -1,115 +1,779 @@
...
+  const {
+    surface = DefaultColor.Surface,
+    onSurface = DefaultColor.OnSurface,
+    interactive = DefaultColor.Interactive,
+    branded = DefaultColor.Branded,
+    critical = DefaultColor.Critical,
+    warning = DefaultColor.Warning,
+    highlight = DefaultColor.Highlight,
+    success = DefaultColor.Success,
+  } = colors;
```
danrosenthal: These defaults will get applied using tokens.

src/utilities/theme/utils.ts
```diff
@@ -1,115 +1,779 @@
...
+    ...surfaceColors(surface),
+    ...onSurfaceColors(onSurface),
+    ...interactiveColors(interactive),
+    ...interactiveNeutralColors(onSurface),
+    ...brandedColors(branded),
+    ...criticalColors(critical),
+    ...warningColors(warning),
+    ...highlightColors(highlight),
+    ...successColors(success),
+  });
```
danrosenthal: This builds up a flat object of CSS Custom Properties containing the entire color system generated from the base values.

src/utilities/theme/utils.ts
```diff
@@ -1,115 +1,779 @@
...
+        hue,
+        saturation,
+        lightness: isLightTheme ? 15 : 85,
+      }),
+    };
   }
+}
+
+// TODO: Type name here is weird, we should either really type this so only camelCase key names are valid in the argument
+// or use a more generic name for this type
```
danrosenthal: 👍

src/utilities/theme/utils.ts
```diff
@@ -1,115 +1,779 @@
...
+  'Success/Divider': '#008060',
+  'Success/SurfaceMuted': '#F8FCFB',
+  'Dark/Elevation00': '#111213',
+  'Dark/Elevation01': '#1F2225',
+  'Dark/Elevation02': '#35383B',
+  'Dark/Elevation03': '#484A4D',
+  'Dark/Elevation04': '#606367',
+};
+
+log(colorsFromFigma);
```
danrosenthal: This will all go after development leaves the prototyping phase.

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -55,20 +62,26 @@ export class ThemeProvider extends React.Component<ThemeProviderProps, State> {
...
+      ? legacyColors.reduce(
+          (state, [key, value]) => ({...state, [key]: value}),
+          {},
+        )
       : null;
   }
 }
 
 function setThemeContext(ctx: Theme): Theme {
+  // TODO: consider adding legacy colors to theme context
```
danrosenthal: `TopBar` is unlikely to update, so this may be unnecessary.

src/utilities/theme/types.ts
```diff
@@ -17,12 +17,24 @@ export interface ThemeColor {
...
+
 export interface TopBar extends ThemeColor {
   background: string;
 }
 
 export type ThemeColors = {
-  topBar: TopBar;
+  topBar?: TopBar;
```
danrosenthal: `topBar` becomes optional here.

src/utilities/theme/types.ts
```diff
@@ -17,12 +17,24 @@ export interface ThemeColor {
...
+  surface?: string;
+  onSurface?: string;
+  interactive?: string;
+  branded?: string;
+  critical?: string;
+  warning?: string;
+  highlight?: string;
+  success?: string;
```
danrosenthal: It's unlikely we'll expose all these roles on the `ThemeProvider`, but these are all the new color roles added as part of the Color System project.

alex-page:
Having a way to visibly test all of the generated values in a grid of foreground, background and contrast values would be super valuable and a great way to quickly gauge if this generation is scaling well. This would be a useful component example in storybook.

danrosenthal:
@alex-page 

> Having a way to visibly test all of the generated values in a grid of foreground, background and contrast values would be super valuable and a great way to quickly gauge if this generation is scaling well. This would be a useful example in storybook.
>
> I've seen this done roughly before using something like this: https://github.com/jxnblk/contrast-swatch

The goals for this work are:

- implement MB
- allow theming color across our library of components
- create a naming convention for using color in design and development

The work being done will unlock future accessibility improvements, and doing them in a programmatic way, but implementing those accessibility improvements will not be a direct outcome of the color system project. Let's set up a meeting to chat about the project outcomes, since we initially decided that accessibility improvements wouldn't be included.

src/components/ThemeProvider/ThemeProvider.tsx
danrosenthal:
```suggestion
```

danrosenthal: We don't need to do this.

src/components/ThemeProvider/ThemeProvider.tsx
BPScott: Wrapper divs have caused us some pain before - in UXPin integration and people raising an eyebrow as styleguide's html output includes them. I appreciate it's a nice separation of concerns but I'd like to avoid adding extra nesting if possible. If the only reason for separating them is nice aesthetics could we keep to a single div to reduce the chance people thinking multiple wrapping divs is odd.

src/utilities/theme/utils.ts
```diff
@@ -1,115 +1,779 @@
...
-}
+export function Colors(theme: Theme) {
+  const {colors = {}} = theme;
+
+  const isLightTheme = isLight(
+    hslToRgb(colorToHsla(
+      colors == null || colors.surface == null
+        ? DefaultColor.Surface
+        : colors.surface,
+    ) as HSLColor),
```
BPScott: Having to cast the result of colorToHsla() everywhere makes me think that return type is pretty crap.

Can you investigate what's going on there and make it so it always returns a type of HSLAColor or some other suitable type?

src/utilities/theme/utils.ts
BPScott: I don't think you can make typescript say "a string that fits a particular format", the closest you can do is a set of literal types `type foo = 'a' | 'b' | 'c';` but I don't think you want to write out every possible value here.

You might also be interested in nominal typing: https://basarat.gitbooks.io/typescript/docs/tips/nominalTyping.html as a mechanism to say "this is just a string but we've already checked that it is this particular type of string

src/utilities/theme/utils.ts
```diff
@@ -1,115 +1,779 @@
...
-      colorPairs.push([
-        constructColorName(baseName, key, 'lighter'),
-        lightenToString(color, 15, 15),
-      ]);
-
-      break;
-    default:
+    return {
+      onSurface: hslToString(hslBaseColor),
+      actionOnDark: hslToString({
```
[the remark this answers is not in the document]
danrosenthal: I think this makes sense. I had been reluctant to DRY this code up since we might end up tinkering with `saturation`, but I think it makes sense to handle that case if/when we come to it.

src/utilities/theme/utils.ts
danrosenthal: This one is complicated, since the way we type color is a bit iffy, and will always result as any invalid color being considered a string.

```
function getColorType(color: string): ColorType {
  if (color.includes('#')) {
    return ColorType.Hex;
  } else if (color.includes('rgb')) {
    return ColorType.Rgb;
  } else if (color.includes('rgba')) {
    return ColorType.Rgba;
  } else if (color.includes('hsl')) {
    return ColorType.Hsl;
  } else if (color.includes('hsla')) {
    return ColorType.Hsla;
  } else {
    if (process.env.NODE_ENV === 'development') {
      /* eslint-disable-next-line no-console */
      console.warn('Accepted colors formats are: hex, rgb, rgba, hsl and hsla');
    }
    return ColorType.Default;
  }
}
```


```
export function colorToHsla(color: string): HSLAColor | string {
  const type: ColorType = getColorType(color);
  switch (type) {
    case ColorType.Hex:
      return hexToHsla(color);
    case ColorType.Rgb:
    case ColorType.Rgba:
      return rbgStringToHsla(color);
    case ColorType.Hsl:
    case ColorType.Hsla:
    default:
      return color;
  }
}
```

I think the best I can do here is add a typeguard that the color is not a string, since a string is essentially an invalid color. If this typeguard shows that our type is a string, it will warn to the console that the user provided an invalid color, and early return an empty object.

Thoughts on this? I'll push up the code shortly.

BPScott: If a string is an invalid color perhaps we should be throwing exceptions rather than letting that invalidness permeate through function calls?


```
export function strictColorToHsla(color: string): HSLAColor {
  const type: ColorType = getColorType(color);
  switch (type) {
    case ColorType.Hex:
      return hexToHsla(color);
    case ColorType.Rgb:
    case ColorType.Rgba:
      return rbgStringToHsla(color);
    case ColorType.Hsl:
    case ColorType.Hsla:
      return color
    default:
      throw new Error('hey that's not a valid color object');
  }
}

```

src/utilities/theme/context.tsx
```diff
@@ -1,4 +1,4 @@
 import React from 'react';
-import {Theme} from './types';
+import {Logo} from './types';
 
-export const ThemeContext = React.createContext<Theme | undefined>(undefined);
+export const ThemeContext = React.createContext<Logo | undefined>(undefined);
```
BPScott: This doesn't seem right - My initial thoughts would have expected the values in the context to include all the colors so you can do things like `useContext(ThemeContext)['colors']['surface']`. Is that not desired?

The naming of the stuff within themeing as always been a bit obscure to me - I've never been very clear on what is publicly exposed vs what is internal. Might be good to have a 30min chat so I can understand that better

kaelig:
Not sure this is the right place to bring this up but is there a reason custom properties aren't prefixed with `polaris-`?

src/utilities/theme/context.tsx
danrosenthal: I went ahead and reverted this back to `Theme`, and stopped destructuring colors off the theme we provide through context.

danrosenthal:
> Not sure this is the right place to bring this up but is there a reason custom properties aren't prefixed with polaris-?

Good question @kaelig. I'd love to discuss the pros/cons of this approach.

As I see it:

|pros|cons|
|-|-|
|namespacing so it does not conflict|more characters to type|

The name-spacing is perhaps not necessary, since we are setting our custom properties on the `ThemeProvider` instead of the `body`, so they will not pollute outside the bounds of Polaris.

The character count issue could be solved with a shorter namespace, maybe something like `plrs-`.

Thoughts?

BPScott:
Had a chat with Dan.

Currently we have a single interface `Theme` that is the shape for both the value passed into AppProvider, and then into ThemeProvider. and also the value that gets stored in the ThemeContext (and thus is accessible when you call `useTheme()`.

This feels odd to me as what we pass into AppProvider is all optional, yet we want a concretely shaped object when using useTheme (we don't want to check if values exist before using them we should know they're there). I think there should be two interfaces and within the ThemeProvider we have a function that transforms from one to the other.

How about something like this:

- `ThemeConfig` is the shape of the object that we pass into AppProvider (and then down to ThemeProvider).  All of its values are optional.
- `Theme` is the shape of the object that is stored in the ThemeContext. The majority of its values are required. Any defaults get applied here so we don't need to  check if they exist when using useTheme.
- Within ThemeProvider we call a `generateTheme` function that takes a ThemeConfig (from ThemeProvider's props) and returns a Theme (which is then stored in the ThemeContext)

```
interface ThemeConfig {
  logo?: ThemeLogo; 
  colors?: {
    topBar?: TopBar;
    surface?: string;
    onSurface?: string;
    interactive?: string;
    interactiveNeutral?: string;
    branded?: string;
    critical?: string;
    warning?: string;
    highlight?: string;
    success?: string;
  }
  }
}

// We probably don't want to expose colors on this for the moment as it's not actually useful in JS.
// In the future we might want to add things like spacing to this, at that point we can work out how we want to expose colors / spacing etc on this.
// If we did want colors in this then it'd the generated colors `colors: { onAction: string; actionOnDark: string; /*etc*/ }`
interface Theme {
  logo? ThemeLogo; // A logo is always optional
}

```

src/utilities/theme/types.ts
```diff
@@ -22,14 +22,23 @@ export interface TopBar extends ThemeColor {
...
 };
 
 export interface Theme {
   /** Sets the logo for the top bar and contextual save bar components*/
   logo?: ThemeLogo;
   /** Sets the background color of the top bar component. Complimentary and typography colors are determined programmatically */
-  colors?: ThemeColors;
+  colors: ThemeColors;
```
danrosenthal: This would be considered a breaking change, so we'd want to ship this as part of a `v5.0.0` alpha.

src/utilities/theme/types.ts
danrosenthal: If these might change at all, we'll want to ship this as part of a `v5.0.0` alpha release, in case anyone happens upon them and starts using them for some reason.

src/utilities/theme/types.ts
```diff
@@ -22,14 +22,28 @@ export interface TopBar extends ThemeColor {
...
   logo?: ThemeLogo;
   /** Sets the background color of the top bar component. Complimentary and typography colors are determined programmatically */
   colors?: ThemeColors;
 }
 
+export interface Theme {
+  /** Sets the logo for the top bar and contextual save bar components*/
+  logo?: ThemeLogo;
+}
+
```
danrosenthal: Is there a way to make these type changes backwards compatible, or will this need to be shipped as part of a `v5.0.0` alpha release?

BPScott: Theme and the other types in here are never exposed publicly (check usage references and see what files they get exported from), you're free to name them whatever you wish. The public entry point for using these shapes would be plucking the type from `AppProviderProps['theme']` and we'll be keeping that same prop name. It doesn't matter that we're renaming the type that references internally from `Theme` to `ThemeConfig` because the shape will remain compatible with what it expects currently.

src/components/ContextualSaveBar/README.md
```diff
@@ -91,6 +91,7 @@ Use the save action to provide an opportunity to save changes. Use the discard a
 <div style={{height: '250px'}}>
   <AppProvider
     theme={{
+      colors: {},
```
BPScott: These changes can be reverted now that we've split Theme and ThemeConfig, and colors remains an optional prop in ThemeConfig

src/components/PolarisTestProvider/PolarisTestProvider.tsx
```diff
@@ -114,7 +114,7 @@ export function PolarisTestProvider({
 function noop() {}
 
 function createThemeContext(theme: Partial<Theme> = {}): Theme {
```
BPScott: The arguments of PolarisTestProvider should match that of AppProvider.

The `theme` prop should be of type `ThemeConfig`, and this function should be typed as  `function createThemeContext(theme: ThemeConfig = {}): Theme {`

I think this function can be replaced entirely with your new `buildThemeContext` function now that has been extracted to be sharable

src/components/index.ts
```diff
@@ -46,7 +46,7 @@ export {
   HSBLAColor,
   rgbToHex,
   rgbToHsb,
-  rgbToHsl,
+  rgbToHsla,
```
BPScott: I really hope nobody has been using this publicly, but this is a breaking change

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -1,74 +1,33 @@
...
-
-interface State {
-  theme: Theme;
-  colors: string[][] | undefined;
-}
+import {ThemeConfig} from '../../utilities/theme/types';
+import {
+  buildThemeContext,
+  buildCustomProperties,
+} from '../../utilities/theme/utils';
```
BPScott: This should export from `'../../utilities/theme'`

src/utilities/theme/types.ts
```diff
@@ -11,25 +9,31 @@ export type ThemeLogo = {
...
 
-export interface TopBar extends ThemeColor {
-  background: string;
+// The value that is passed into the ThemeProvider
+export interface ThemeConfig {
+  /** Sets the logo for the top bar and contextual save bar components*/
+  logo?: ThemeLogo;
+  colors?: {
+    /** Sets the background color of the top bar component. Complimentary and typography colors are determined programmatically */
+    topBar?: Record<string, string>;
```
BPScott: It'd be great to if we could identify the valid keys within topBar and identify them here

src/utilities/theme/utils.ts
```diff
@@ -1,115 +1,304 @@
...
+    }),
+    ...overrides(),
+  };
+}
+
+function surfaceColors(
+  color: HSLAColor,
+  lightSurface: boolean,
+): {
+  [key: string]: HSLAColor;
```
BPScott: Don't specify a return type here, you're making the type be a lot wider than it needs to be - this objec has a specific set of keys and it'd be useful to identify that shape rather than be "and object that contain any key".

src/components/ThemeProvider/ThemeProvider.tsx
BPScott: Loving that these these transformation functions now live in utilities/theme, It makes loads more sense!

kaelig:
>The name-spacing is perhaps not necessary, since we are setting our custom properties on the ThemeProvider instead of the body, so they will not pollute outside the bounds of Polaris.

Like you, I'm not sure it's a concern _for now_ since custom property usage is still very new, but it seems like being good CSS citizens, especially for app developers who might wrap their app in the ThemeProvider but also want to use various CSS frameworks.

I'd compare this practice as being good JS citizens by not claiming JavaScript's global scope (although one could argue that the chance to have a war over Polaris's custom property naming is way lower than when jQuery and Prototype had a battle over the `$` variable!).

>The character count issue could be solved with a shorter namespace, maybe something like `plrs-`.

You're spot on with those pros and cons. I agree that `polaris-` would be too long.

I like `plrs-`, although it could be perceived as being too long. Based on feedback from users of the Lightning Design System, `slds-` was a pain to write and bloating their code (it was on classes, mind you).

Alternatives:

- `sp-` (Shopify Polaris)
- `s-` (Shopify)
- `p-` (Polaris)
- `` (nothing, see what happens, we might easily be fine for another couple of years?)

tmlayton:
@BPScott addressed most of your feedback. And pulling a few bits into their own PRs is probably a good idea.

@kaelig @danrosenthal I’m okay with a `p` prefix.

src/utilities/theme/types.ts
tmlayton: Going to leave this one as-is.

src/utilities/theme/types.ts
tmlayton: Updated the interface name here to be `UNSTABLE_colors` so we can make otherwise breaking changes at will.

kaelig:
🎉 🎉 🎉 🎉 🎉 indeed!

PR #1926, opened 2019-08-06
[RangeSlider] Fix range slider focus state

spencercanner:
* The range slider currently doesn't have a visible focus state, which is especially noticeable when it is used without the numeric output bubble 

* Removed a class selector from the `SingleThumb` styles which was preventing the focus styles from being applied to the input
  * The `.RangeSlider` selector was never used anywhere in the component, and was left over from the before the `DualThumb` component existed
* Added an outline to the entire `SingleThumb` input when focused in high contrast mode
* Updated the `SingleThumb` styles to prevent focus from taking over the error track colour
  * This should improve consistency across other input components (e.g. checkbox, select), where it will maintain some error styles but the focused area will receive an indigo outline
* Copied over the focus styles to the `DualThumb` scss to apply consistent focus styles across both components
* Added an outline to the `DualThumb` thumb button when focused in high contrast mode
* Disabled the `DualThumb` buttons to prevent them from receiving focus when the component is disabled
  * The single thumb range slider cannot receive focus in a disabled state, so this improves consistency across range sliders 

### How to 🎩

General case: 
* Tab into the input, and verify the slider receives an indigo border
* Use the mouse to move the slider, and verify it receives the focus state

Error state:
* Update component to be in an error state by adding `error="error"` as a prop
* Verify that focus states will not change the track colour, but the slider will receive an indigo border

<details>

```jsx
import {Page, Card, RangeSlider} from '../src';

  const initialValue = [700, 1000];
  const prefix = '$';
  const min = 0;
  const max = 2000;
  const step = 10;

  const [doubleRangeValue, setDoubleRangeValue] = useState(initialValue);
  const [rangeValue, setRangeValue] = useState(32);

  const handleDoubleRangeSliderChange = useCallback((value) => {
    setDoubleRangeValue(value);
  }, []);

  const handleRangeSliderChange = useCallback(
    (value) => setRangeValue(value),
    [],
  );

  return (
    <Page title="Playground">
      <Card sectioned title="Background color">
        <RangeSlider
          label="Opacity percentage"
          value={rangeValue}
          onChange={handleRangeSliderChange}
          output
        />
      </Card>
      <Card sectioned title="Minimum requirements">
        <RangeSlider
          output
          label="Money spent is between"
          value={doubleRangeValue}
          prefix={prefix}
          min={min}
          max={max}
          step={step}
          onChange={handleDoubleRangeSliderChange}
        />
      </Card>
    </Page>
  );
}
```

</details>

**Mac Chrome:**
Single without focus: 
image
Single with focus: 
image
Single error without focus:
image
Single error with focus:
image

Double without focus:
image
Double with focus:
image
Double error without focus:
image
Double error with focus:
image

**Edge:**
Single without focus: 
image
Single with focus: 
image
Single error without focus:
image
Single error with focus:
image

Double without focus:
image
Double with focus:
image
Double error without focus:
image
Double error with focus:
image

**High contrast mode:**
Single without focus: 
image
Single with focus: 
image
Single error without focus:
image
Single error with focus:
image

Double without focus:
image
Double with focus:
image
Double error without focus:
image
Double error with focus:
image

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -9,6 +9,12 @@
 ### Bug fixes
 
 - Fixed issue with `Filters` component displaying an undesired margin top and bottom on the button element on Safari ([#2292](https://github.com/Shopify/polaris-react/pull/2292))
+- Doesn't render `MenuActions` if no actions are passed to an `actionGroups` item inside `Page` ([#2266](https://github.com/Shopify/polaris-react/pull/2266))
+- Fixed `PositionedOverlay` incorrectly calculating `Topbar.UserMenu` `Popover` width ([#1692](https://github.com/Shopify/polaris-react/pull/1692))
+- Fixed `recolor-icon` Sass mixin to properly scope `$secondary-color` to the child `svg` ([#2298](https://github.com/Shopify/polaris-react/pull/2298))
+- Fixed a regression with the positioning of the `Popover` component ([#2305](https://github.com/Shopify/polaris-react/pull/2305))
+- Fixed Stack Item proportion when shrinking ([#2319](https://github.com/Shopify/polaris-react/pull/2319))
+- Fixed `RangeSlider` focus state style issues ([#1926](https://github.com/Shopify/polaris-react/pull/1926))
 
 ### Documentation
 
--- a/src/components/RangeSlider/components/DualThumb/DualThumb.scss
+++ b/src/components/RangeSlider/components/DualThumb/DualThumb.scss
@@ -14,6 +14,7 @@ $range-thumb-border-error: rem(2px) solid color('red');
 $range-thumb-shadow: (0 0 0 rem(1px) rgba(black, 0.2), shadow(faint));
 $range-thumb-shadow-hover: (0 0 0 rem(1px) rgba(black, 0.4), shadow(faint));
 $range-thumb-shadow-error: 0 0 0 rem(1px) color('red');
+$range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
 
 .Wrapper {
   position: relative;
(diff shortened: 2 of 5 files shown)


danrosenthal:
Hey @spencercanner, sorry for the delay on getting review on this PR, seems like it must have fallen between the cracks.

I've tophatted this change, and it's looking great in the single input range slider. I'd love to see this change made in the dual input range slider as well. The focus state in it currently is pretty gross.

|Dual Input|
|-|
|Screen Shot 2019-09-25 at 10 49 25 AM|

Also pinging @dpersing for her thoughts on the high contrast styles. I wonder if the `button-text-background` color would be more appropriate here. It'd also be great to add high contrast background color to the track so it's visible. But that could be done in another PR if you like.

dpersing:
@danrosenthal @spencercanner Hi! Thanks so much for catching this.

I think we want to maintain a focus state for High Contrast that doesn't rely on color alone, since it's a subtle change. Having a focus indicator on the full input would be ideal, since this mimics the default focus ring behavior.

What tool and browser are you using for the high contrast setting in this case? I ask since we typically test with Edge and the native High Contrast theme settings on Windows, which looks like this for the current example in the style guide:

Default:

Slider in High Contrast

Focus state:

Slider in High Contrast with focus indicator

spencercanner:
@dpersing I was testing with Edge as well, but running it on a VirtualBox Windows VM using the `High Contrast Black` theme. I just checked the polaris styleguide, and I'm seeing this as the focus state:
image 
Maybe the VM isn't providing a good representation of an actual machine?

dpersing:
@spencercanner Gotcha, that is curious. I had the same theme, but I'm using VMware and looking at the version that's currently in the style guide. I'll block out some time later today to take a closer look.

dpersing:
Yeah, here's your version in VMware in High Contrast black.

High Contrast black theme on Windows 10 in VMware

😩

spencercanner:
@danrosenthal I've updated the PR to include the single and dual thumb range slider focus states, and added more detail in the PR description above

@dpersing I think the HC inconsistencies were caused by my VM running an older version of Edge, but I've updated the PR to include HC focus styling for both the single and dual thumb range sliders

danrosenthal:
@spencercanner, I'm going to go ahead and rebase, fix the conflict, and merge this PR.

PR #1927, opened 2019-08-06
[v4] Minor changes to bump test coverage slightly

AndrewMusgrave:

Bump test coverage 🎉

### DIFF
--- a/src/components/DropZone/tests/DropZone.test.tsx
+++ b/src/components/DropZone/tests/DropZone.test.tsx
@@ -3,7 +3,9 @@ import {ReactWrapper} from 'enzyme';
 import {clock} from '@shopify/jest-dom-mocks';
 import {Label, Labelled, DisplayText, Caption} from 'components';
 import {mountWithAppProvider} from 'test-utilities/legacy';
+import {mountWithApp} from 'test-utilities';
 import DropZone from '../DropZone';
+import {DropZoneContext} from '../context';
 
 const files = [
   {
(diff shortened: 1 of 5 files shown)


src/components/DropZone/tests/DropZone.test.tsx
```diff
@@ -308,6 +310,26 @@ describe('<DropZone />', () => {
       const displayText = dropZone.find(DisplayText);
       expect(displayText.contains(errorOverlayText)).toBe(true);
     });
+
+    it('sets type from props on context', () => {
```
AndrewMusgrave: Test to make sure we're setting context

src/components/Frame/Frame.tsx
```diff
@@ -376,10 +376,8 @@ class Frame extends React.PureComponent<CombinedProps, State> {
   };
 
   private handleClick = () => {
-    if (this.skipToMainContentTargetNode.current == null) {
-      return;
-    }
-    this.skipToMainContentTargetNode.current.focus();
+    this.skipToMainContentTargetNode.current &&
```
AndrewMusgrave: `skipToMainContentTargetNode ` will never be null after mount, this is just a shortcut and will give us coverage

src/components/Icon/Icon.tsx
```diff
@@ -56,7 +56,7 @@ export default function Icon({
     );
   } else if (source === 'placeholder') {
     contentMarkup = <div className={styles.Placeholder} />;
-  } else if (typeof source === 'string') {
+  } else {
```
AndrewMusgrave: We don't need an if, elif, elif since we only have 3 type options (function, 'placeholder', string)

PR #1937, opened 2019-08-07
Update examples to v4 & add ts-react-testing example

AndrewMusgrave:

* Updating examples to `v4.0.0-rc.2`
* Adding a new example `create-react-app-ts-react-testing` to align more with our current stack since we're all about promoting typescript (other create-react-app example is enzyme)
* Remove linting from examples since examples are independently run. It doesn't make sense to have linting and ignores that won't apply to the specific examples

### How to 🎩

Startup the examples and look for errors and see if they work 🔍


.eslintignore
```diff
@@ -1,5 +1,5 @@
 node_modules
-/examples/*/build
+/examples
```
AndrewMusgrave: Our examples are to be run independently, we shouldn't link based on our repo's rules.

chloerice: 💯

AndrewMusgrave:
> All updated examples work! Not related to these changes, but the browserify example seems like it's missing files 🤔 Should it only have a build folder?

@chloerice Ben deleted the browserify example, you should be 👌 to delete the build folder 😄

Issue #1943, opened 2019-08-08
[Choice list] "Single choice list with error" should have iOS/Android screenshots


kaelig:
Seen on https://polaris.shopify.com/components/forms/choice-list

![Screen Shot 2019-08-08 at 12 27 12 PM](https://user-images.githubusercontent.com/85783/62731830-e275f800-b9d7-11e9-8686-ff45ea67bbd6.png)

Issue #1683, opened 2019-06-13
[Text field] "Text field with right aligned text" example should have iOS/Android screenshots


kaelig:
Seen on https://polaris.shopify.com/components/forms/text-field

![Screen Shot 2019-06-13 at 11 05 53 AM](https://user-images.githubusercontent.com/85783/59456479-3d6de300-8dcb-11e9-91f7-d0d2dac86f06.png)

see-mack:
We don't use right-aligned text fields on mobile unless it's for Right-To-Left languages. I'll remove it from the android/ios section of the docs.

Issue #1682, opened 2019-06-13
[Range slider] "Dual thumb range slider" example should have iOS/Android screenshots


kaelig:
Seen on https://polaris.shopify.com/components/forms/range-slider

![Screen Shot 2019-06-13 at 11 05 02 AM](https://user-images.githubusercontent.com/85783/59456415-1c0cf700-8dcb-11e9-9388-d7caa125f05b.png)

see-mack:
So we have a dev component that's not used in any app for the slider on iOS but it's not in the iOS UI kit. For android, it doesn't exist as a dev component _or_ in the UI Kit for android.

Issue #1681, opened 2019-06-13
[Button] "Plain destructive button" example should have Android & iOS screenshot


kaelig:
Seen on https://polaris.shopify.com/components/actions/button

![Screen Shot 2019-06-13 at 11 00 19 AM](https://user-images.githubusercontent.com/85783/59456194-aacd4400-8dca-11e9-8d63-94393f4a8560.png)

chloerice:
It looks like we just need to separate the plain destructive variants from the destructive button screenshots. What's the process for making iOS/Android component screenshots? I notice that the images are all a certain dimension. Would it be the wrong approach to create the plain destructive assets by modifying the existing screenshots to be two separate images?

Current screen shots that could be split:

|  iOS | Android  |
|---|---|
|<img width="376" alt="Screen Shot 2019-06-17 at 5 41 43 PM" src="https://user-images.githubusercontent.com/18447883/59639061-16474680-9128-11e9-87f5-1dc9dee7f9b6.png">|<img width="395" alt="Screen Shot 2019-06-17 at 5 40 45 PM" src="https://user-images.githubusercontent.com/18447883/59639068-1e9f8180-9128-11e9-9b8c-aa1a7880571c.png">|

PR #1944, opened 2019-08-08
Add metadata to examples missing mobile content

kaelig:
Some examples are missing mobile examples. This PR hides them from the iOS and Android tabs in the style guide.

Related to: #1681, #1682, #1683, #1943


chloerice:
Hey @kaelig, I've tried running `build-consumer` to the style guide with this branch, but then the build fails in `polaris-styleguide` when I `dev server` with this error

Screen Shot 2019-08-09 at 1 10 11 PM

I don't think it's related to these changes, but `dev up` before `build-consumer` on both branches doesn't help. Gonna see if master fails the build in the style guide as well 👀

chloerice:
Looks like we swapped `react-transition-group` for `@material-ui/react-transition-group` in #1759, I wonder if pending dependency upgrades like this are a limitation in `build-consumer` 🤔

kaelig:
>The current `polaris-react` master fails to build in `polaris-styleguide` too. Looks like we swapped `react-transition-group` for `@material-ui/react-transition-group` in #1759, I wonder if pending dependency upgrades like this are a limitation in `build-consumer` 🤔 

Yes, dependency changes do cause build-consumer to fail.

Try `yarn add @material-ui/react-transition-group` in polaris-styleguide, and then run build-consumer?

PR #1948, opened 2019-08-09
[Filters] Add onQueryFocused callback

joshuajay:
Current functionality in web relies on an `onQueryFocus` callback on the ListFilters component in polaris-next. This is used to instrument the search behaviour, where a focus event on the query field indicates the beginning of the users intent to search.

Tim's excellent PR to replace the polaris-next ListFilters with the proper Filter component from Polaris means we can no longer do this. (see https://github.com/Shopify/web/pull/15425#discussion_r311870970)

Adding a callback to the Filters component to be triggered when the query field is focused.
### 🎩 checklist


joshuajay:
@tmlayton Thanks Tim! Test has been added.

tmlayton:
Will setup a new shipit stack to cut 3.21.1 with this change

PR #1955, opened 2019-08-12
Rename export from use-deep-effect to be useDeepEffect

BPScott:
Follow up to #1938.

That PR renamed the files but didn't rename the exported function

Renames `useDeepCompare` to `useDeepEffect` to match the filename


AndrewMusgrave:
#1938 Should have handled the stragglers if you rebase with masters and we're still missing anything. Feel free to re-ping me 😄

PR #1957, opened 2019-08-12
Add minified styles to npm package

BPScott:

Maintaining our own CDN distribution of our CSS files is a little annoying as it leads to more complex builds. Let's let unpkg handle our CDN styles.

The first step to that is adding minified styles into npm package - that's this PR.
The second step is ripping out our build config that uploads files to sdks.shopifycdn.com and updates our links to unpkg.com files (that'll be a new PR after we do an RC containing this change)

Adds `styles.min.css` to our npm package

### How to 🎩

- Run `yarn run build` and see that `styles.min.css` is generated
- Run `npm pack` then extract the generated tgz and see that styles.min.css is included in the package folder
- See that `styles.min.css` is excluded from git and linting configs
- See that running `yarn run clean` deletes `styles.min.css`

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -7,6 +7,7 @@ node_modules
 /types
 /index.*
 /styles.css
+/styles.min.css
 /styles.scss
 /styles
 /src/styles/polaris-tokens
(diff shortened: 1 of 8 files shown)


BPScott:
Skipping the changelog on this PR, I'll write a changelog entry in the follow up PR that talks about changing URLs

PR #1968, opened 2019-08-14
[ContextualSaveBar] Replace RTG

AndrewMusgrave:

RTG was causing infinite renders with ContextualSaveBar

* adds custom css transition

### How to 🎩

Frame examples 😎


src/components/Frame/components/CSSAnimation/CSSAnimation.tsx
```diff
@@ -0,0 +1,89 @@
+import React, {useRef, useState, useEffect} from 'react';
+// import {durationSlow} from '@shopify/polaris-tokens';
```
dleroux: Forgot this one.

PR #1970, opened 2019-08-14
Reverted isObjectsEqual

AndrewMusgrave:

Reverting #1580 because of an issue that will be linked below


lsit:
Issues noted here https://github.com/Shopify/fbs/issues/5846

PR #1973, opened 2019-08-15
[ResourceList] Make item actions accessible

dleroux:

Fixes **Shortcut actions** part of  https://github.com/Shopify/polaris-react/issues/792

**Shortcut actions**

1. Smaller-breakpoint, persistent shortcut action buttons have no text equivalent for the ... icon. Screen reader users may not be able to tell what the button does.

2. Controls that repeat for persistent shortcut actions for each list item do not have unique text. Non-visual users may have difficulty distinguishing which resource item they relate to as they tab through the controls and hear them read.

For item 1 the wrong attribute was being added to the button so changing it to `aria-label` to `accessibilityLabel` makes the already translated label appear. However @dpersing you suggested a label like: "Actions for Mae Jemison". Since we don't have control over what people put in the list item in order to have that amount of detail we would need to add another prop to ResourceList which is already a bit bloated. ~Therefore I'm following the approach we did elsewhere: If there's is an accessibility label provided it will be "Actions for {accessibilityLabel}" if not it will just be the current translated text "Actions Dropdown"~  This won't actually make sense. For now, it just says  "Actions Dropdown". If this is acceptable then we should keep this on our radar because I believe the ResourceList will be getting a bit of an overhaul soon.

For item 2 we do have that flexibility but it is up to the user to provide the action with the right label. I updated the example to show that.

### How to 🎩

1. Ensure that the labels for these 2 items are correct.
2. Since there is a change to the documentation, `yarn build-consumer polaris-styleguide` to ensure the documentation renders properly.

<details>

```jsx
import React from 'react';
import {Page, ResourceList, Card, Avatar, TextStyle} from '../src';

  return (
    <Page title="Playground">
      <ResourceListExample />
    </Page>
  );
}

  state = {
    selectedItems: [],
  };

  };

  renderItem = (item) => {
    const {id, url, name, location, latestOrderUrl} = item;
    const shortcutActions = latestOrderUrl
      ? [
          {
            content: 'View latest order',
            accessibilityLabel: `View ${name}'s latest order`,
            url: latestOrderUrl,
          },
        ]
      : null;
    return (
      <ResourceList.Item
        id={id}
        url={url}
        media={media}
        shortcutActions={shortcutActions}
        persistActions
      >
        <h3>
        </h3>
        <div>{location}</div>
      </ResourceList.Item>
    );
  };

  render() {
    const resourceName = {
      singular: 'customer',
      plural: 'customers',
    };
    const items = [
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
        latestOrderUrl: 'orders/1456',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
        latestOrderUrl: 'orders/1456',
      },
    ];
    return (
      <React.Fragment>
        <Card>
          <ResourceList
            resourceName={resourceName}
            items={items}
            renderItem={this.renderItem}
            selectable
          />
        </Card>
      </React.Fragment>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -48,10 +48,12 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Updated translation.yml with the new locales path ([#1649](https://github.com/Shopify/polaris-react/pull/1649))
 - Fixed accessibility issue with `Tabs` list item presentation role ([#1958](https://github.com/Shopify/polaris-react/pull/1958))
 - Removed `Tooltip` on disabled `Pagination` buttons ([#1963](https://github.com/Shopify/polaris-react/pull/1963))
+- Fixed accessibility labels on `ResourceList.Item` persistent action disclosure icon ([#1973](https://github.com/Shopify/polaris-react/pull/1973))
 
 ### Documentation
 
 - Updated `AppProvider` app bridge example to use our `AppBridgeContext`([#1877](https://github.com/Shopify/polaris-react/pull/1877))
+- Updated example for the `ResourceList.Item` persistent actions accessibility labels ([#1973](https://github.com/Shopify/polaris-react/pull/1973))
 
 ### Development workflow
 
--- a/a11y_shitlist.json
+++ b/a11y_shitlist.json
@@ -77,24 +77,6 @@
       "selector": "#Popover1 > div > div > ul > li:nth-child(5) > div > button"
     }
   ],
-  "all-components-resource-list--resource-list-with-persistent-item-shortcut-actions": [
-    {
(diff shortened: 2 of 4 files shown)


dpersing:
>Since we don't have control over what people put in the list item in order to have that amount of detail we would need to add another prop to ResourceList which is already a bit bloated.

What do you think about leveraging the `name` prop, like we do for the resource link, if one is provided, then falling back to a generic label if one isn't? We could update the documentation to include guidance around including a unique `name` for each item.

dleroux:
I'll leave this open depending on the timeline of the refactor. If it's too long let's just add `name` to the Item component.

PR #1978, opened 2019-08-16
Removed mocks that are no longer needed

AndrewMusgrave:

At some point, these mocks became redundant, lets strip away this extra code

### DIFF
--- a/src/components/TrapFocus/tests/TrapFocus.test.tsx
+++ b/src/components/TrapFocus/tests/TrapFocus.test.tsx
@@ -10,16 +10,8 @@ import {
 import TrapFocus from '../TrapFocus';
 
 describe('<TrapFocus />', () => {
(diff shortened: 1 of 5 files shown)


src/components/TrapFocus/tests/TrapFocus.test.tsx
```diff
@@ -10,16 +10,8 @@ import {
...
 describe('<TrapFocus />', () => {
-  let requestAnimationFrameSpy: jest.SpyInstance;
-
-  beforeEach(() => {
-    requestAnimationFrameSpy = jest.spyOn(window, 'requestAnimationFrame');
-    requestAnimationFrameSpy.mockImplementation((cb) => cb());
-  });
-
   afterEach(() => {
     (document.activeElement as HTMLElement).blur();
```
AndrewMusgrave: Technically we can remove this as well, how if we're expecting on the activeElement it's good practice.

PR #1774, opened 2019-07-03
Resource item

tmlayton:
### WHY are these changes introduced?

Closes #1768 
Closes #1769
Closes #1770 
Closes #1771
Closes #1772
Closes #1951 

### WHAT is this pull request doing?

- Moves `ResourceItem` to its own component (formerly called `Item` and a subcomponent of `ResourceList`). The component is still available at `ResourceList.Item`
- Updates the `ResourceItem` docs per [the new documentation](https://docs.google.com/document/d/1wuM0bBexvJQvmUsk0EBCbCjyiLxEFz8hdxCYJY3UFNE/edit?ts=5d2c9226)
- Removes the `FilterControl` docs
- Deprecates `FilterControl`
- Updates `ResourceList` examples to use the `Filters` component
- Simplifies `ResourceList` docs and removes case studies for now until they can find a new home in the style guide
- Adds example to `Filters` which uses children content
- Used `inlineLabel` for the sort select in `ResourceList`

### 📸

|Resource item|Resource list|
|---|---|
|screencapture-polaris-myshopify-io-components-lists-and-tables-resource-item-2019-08-16-17_42_46|screencapture-polaris-myshopify-io-components-lists-and-tables-resource-list-2019-08-16-17_44_47|

### 🎩 checklist

* [x] Tested on [mobile](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting.md#cross-browser-testing)
* [x] Tested on [multiple browsers](https://help.shopify.com/en/manual/intro-to-shopify/shopify-admin/supported-browsers)
* [x] Tested for [accessibility](https://github.com/Shopify/polaris-react/blob/master/documentation/Accessibility%20testing.md)
* [x] Updated the component's `README.md` with documentation changes
* [x] [Tophatted documentation](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting%20documentation.md) changes in the style guide

PR #1986, opened 2019-08-18
[Resource list] Hide inline label on small screen

tmlayton:

I switched the sort select to use an inline label in #1774, but did not hide the label on small screen.

|Before|After|
|---|---|
|Screen Shot 2019-08-18 at 3 07 38 PM|Screen Shot 2019-08-18 at 3 07 43 PM|

Tracking `smallScreen` in state and using the existing resize handler to set it. If `smallScreen` is true then hide the inline label. Not adding a changelog entry as the inline label change has not yet been released.

Bonus: bumps the test coverage by 3% for resource list.

### 🎩 checklist

### DIFF
--- a/src/components/ResourceList/ResourceList.tsx
+++ b/src/components/ResourceList/ResourceList.tsx
@@ -40,6 +40,7 @@ interface State {
   selectMode: boolean;
   loadingPosition: number;
   lastSelected: number | null;
+  smallScreen: boolean;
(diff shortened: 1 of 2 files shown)


tmlayton:
> Could you add a changelog entry?

The update to use an inline label hasn’t been released yet so this wouldn’t be a bug fix yet.

PR #1991, opened 2019-08-19
[Text field] Allow min and max to accept a string

### LINKED ISSUE #1988: [TextField] Max and Min have the wrong type
## Issue summary

`min` and `max` should allow for `number|string`, not just `number`.

This prevents using `min` and `max` for a native date input, [which is allowed](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date#max).

tmlayton:

Fixes #1988 

Allows `number | string`, updates the prop description

## 

### How to 🎩

<details>

```jsx

  const [value, setValue] = useState('2008-07-22');

  return (
    <Page title="Playground">
      <TextField
        label="Any time this decade 🙄"
        value={value}
        onChange={setValue}
        type="date"
        min="2010-01-01"
        max="2019-12-31"
      />
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -432,7 +432,10 @@ class TextField extends React.PureComponent<CombinedProps, State> {
     // step / value has.
     const decimalPlaces = Math.max(dpl(numericValue), dpl(step));
 
-    const newValue = Math.min(max, Math.max(numericValue + steps * step, min));
+    const newValue = Math.min(
+      Number(max),
+      Math.max(numericValue + steps * step, Number(min)),
+    );
     onChange(String(newValue.toFixed(decimalPlaces)), this.state.id);
   };
 
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -87,12 +87,12 @@ export interface BaseProps {
   step?: number;
   /** Enable automatic completion by the browser */
   autoComplete?: boolean | string;
-  /** Mimics the behavior of the native HTML attribute, limiting how high the spinner can increment the value */
-  max?: number;
+  /** Mimics the behavior of the native HTML attribute, limiting the maximum value */
+  max?: number | string;
   /** Maximum character length for an input */
   maxLength?: number;
-  /** Mimics the behavior of the native HTML attribute, limiting how low the spinner can decrement the value */
-  min?: number;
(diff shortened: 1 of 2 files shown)


src/components/TextField/TextField.tsx
```diff
@@ -432,7 +432,10 @@ class TextField extends React.PureComponent<CombinedProps, State> {
     // step / value has.
     const decimalPlaces = Math.max(dpl(numericValue), dpl(step));
 
-    const newValue = Math.min(max, Math.max(numericValue + steps * step, min));
+    const newValue = Math.min(
```
BPScott: It's neat that this now allow for doing number-like strings for max and mins (note max and min being strings `"1"` not the number `1`):

```
  const [value, setValue] = useState('1');
  const handleChange = useCallback((newValue) => setValue(newValue), []);

  return (
    <Page title="Playground">
      <TextField
        label="thing"
        value={value}
        onChange={handleChange}
        type="number"
        min="1"
        max="10"
      />
    </Page>
  );
}

```

But it means typescript will no longer complain if you do some weird misconfiguration like setting a max to a string that can't be cast to a number e.g.

```
  const [value, setValue] = useState('0');

  const handleChange = useCallback((newValue) => setValue(newValue), []);

  return (
    <Page title="Playground">
      <TextField
        label="thing"
        value={value}
        onChange={handleChange}
        type="number"
        min="1"
        max="2019-08-10"
      />
    </Page>
  );
}
```

In this case newValue ends up being NaN and we get console warnings. Can we add a check above this to ensure  that max and min can be cast to number and if not return without triggering onChange? Perhaps something like (thought about but not tested):

```diff
- if (onChange == null) {
+ if (onChange == null || isNaN(min)) || isNaN(min)) {
```

tmlayton: > In this case newValue ends up being NaN and we get console warnings

That’s why I prefer this approach. Otherwise, with an early return for not a number the stepper no longer works without feedback to the implementer as to why. We could handle the messaging ourselves, but that is reinventing the wheel.

I also considered `isNan` values defaulting to `Infinity` or `-Infinity` which mimics the behavior of the html number input, but that makes it harder for an implementer to debug if they say, fat-fingered `10-` or something like this.

BPScott: Cool, that makes sense

PR #1992, opened 2019-08-20
[Modal] Set default iframe height when src on different domains 

dleroux:

Fixes https://github.com/Shopify/polaris-react/issues/1985

We are getting cross-origin errors in the modal because the modal assumed the src of the frame was on the same domain and try to get the height of the content.

This only occurs when an admin modal tries to open a standalone app.  

This pull request simply does a try/catch and sets a default height in the catch block. The default height is the same default height used in the embedded app: https://github.com/Shopify/web/blob/1617c9b80602a410347f6512a5ebbb068791bd1c/app/sections/Apps/EmbeddedApp/components/Modal/Modal.tsx#L10

modal-height

## 

### How to 🎩

**In Polaris-react:**

1. Tophat the change in polaris react with the playground below. Ensure that there are no cross-origin error in the console when the modal launches

**In Web**

In order to tophat in web you need to install the Buy-Button app https://github.com/Shopify/buy-button

1. From polaris-react run `yarn build-consumer web`
2. `dev clone buy-button`
3. Follow the instruction the `buy-button` README.md
4. Once web is running with internal routes enabled, go to a product details page, and launch the "Buy button" app from the page actions.
5. Ensure no CORS errors are thrown.
 
<details>

```jsx
import React from 'react';
import {Page, Modal, TextContainer, Button} from '../src';

  return (
    <Page title="Playground">
      <ModalExample />
    </Page>
  );
}

class ModalExample extends React.Component {
  state = {
    active: true,
  };

  render() {
    const {active} = this.state;
    return (
      <div style={{height: '500px'}}>
        <Button onClick={this.handleChange}>Open</Button>
        <Modal
          open={active}
          onClose={this.handleChange}
          primaryAction={{
            content: 'Add Instagram',
            onAction: this.handleChange,
          }}
          secondaryActions={[
            {
              content: 'Learn more',
              onAction: this.handleChange,
            },
          ]}
          src="https://alexpage.com.au/"
        />
      </div>
    );
  }

  handleChange = () => {
    this.setState(({active}) => ({active: !active}));
  };
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -31,6 +31,7 @@ import {
 import styles from './Modal.scss';
 
 const IFRAME_LOADING_HEIGHT = 200;
+const DEFAULT_IFRAME_CONTENT_HEIGHT = 400;
 
 export type Size = 'Small' | 'Medium' | 'Large' | 'Full';
 
@@ -335,9 +336,15 @@ class Modal extends React.Component<CombinedProps, State> {
   private handleIFrameLoad = (evt: React.SyntheticEvent<HTMLIFrameElement>) => {
     const iframe = evt.target as HTMLIFrameElement;
     if (iframe && iframe.contentWindow) {
-      this.setState({
-        iframeHeight: iframe.contentWindow.document.body.scrollHeight,
-      });
+      try {
+        this.setState({
+          iframeHeight: iframe.contentWindow.document.body.scrollHeight,
+        });
+      } catch {
+        this.setState({
+          iframeHeight: DEFAULT_IFRAME_CONTENT_HEIGHT,
+        });
+      }
     }
 
     const {onIFrameLoad} = this.props;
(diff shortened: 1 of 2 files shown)


dleroux:
I wasn't able to get the test working for this but I don't want to slow this down so top-hatting should be sufficient.

dleroux:
@PabloVallejo notice the modal doesn't have a footer is that expected?

src/components/Modal/Modal.tsx
```diff
@@ -335,9 +336,15 @@ class Modal extends React.Component<CombinedProps, State> {
...
     const iframe = evt.target as HTMLIFrameElement;
     if (iframe && iframe.contentWindow) {
-      this.setState({
-        iframeHeight: iframe.contentWindow.document.body.scrollHeight,
-      });
+      try {
+        this.setState({
+          iframeHeight: iframe.contentWindow.document.body.scrollHeight,
+        });
+      } catch (err) {
```
danrosenthal: Should this be prefixed with an `_` since we don't use the `err` var?

```suggestion
      } catch (_err) {
```

dleroux: We don't anywhere else and it doesn't complain because it's required.

BPScott: Because typescript supports [optional catch binding](https://mariusschulz.com/blog/optional-catch-binding-in-typescript) we don't need this variable at all:


```suggestion
      } catch {
```

PabloVallejo:
@dleroux yep, not rendering the footer is an issue of the modal that's addressed in this PR: https://github.com/Shopify/web/pull/17349

PR #1994, opened 2019-08-20
[Dropzone] fix various accessibility issues

dleroux:

Fixes https://github.com/Shopify/polaris-react/issues/762

1. The focus issue is revolved by allowing users to focus the `file input` and turning the button into a simple div. Visually it looks like a button but cannot be tabbed to focused to via keyboard.

2. The dropzone keeps tracks of being focused via its state and communicates it to the fileUpload component via context. The same has been done for hovering and disabling of the dropzone. 

3. We explored adding a hover state to the dropzone. This was challenging because in many cases (like the web image card), once images are uploaded, the outline prop is changed to false to accommodate the design. Having a hover isn't ideal. So this PR only adds a hover state when outline is set to true. (once this is finalized a UX issue will be opened to outline various accessibility issue that may require design changes in the admin.)

## 

### How to 🎩

1. Build the playground with the code below
2. Test hover state at all 4 sizes of the drop zone (with and without the outline prop).
3. Test focus state at all 4 sizes as well (with and without the outline prop). 

<details>
<summary><code>Playground code</code>:</summary>

```jsx
import React from 'react';
import {Page, Stack, DropZone, Banner, List, Caption, Thumbnail} from '../src';

  return (
    <Page title="Playground">
      <DropZoneExample />
    </Page>
  );
}

class DropZoneExample extends React.Component {
  state = {
    files: [],
    rejectedFiles: [],
    hasError: false,
  };

  render() {
    const {files, hasError, rejectedFiles} = this.state;

      <Stack vertical>
        {files.map((file, index) => (
            <Thumbnail
              size="small"
              alt={file.name}
              source={window.URL.createObjectURL(file)}
            />
            <div>
            </div>
          </Stack>
        ))}
      </Stack>
    );

    const errorMessage = hasError && (
      <Banner
        title="The following images couldn’t be uploaded:"
        status="critical"
      >
        <List type="bullet">
          {rejectedFiles.map((file, index) => (
            <List.Item key={index}>
              {`"${
                file.name
              }" is not supported. File type must be .gif, .jpg, .png or .svg.`}
            </List.Item>
          ))}
        </List>
      </Banner>
    );

    return (
      <Stack vertical>
        {errorMessage}
        <DropZone
          accept="image/*"
          type="image"
          outline={false}
            this.setState({
              rejectedFiles,
              hasError: rejectedFiles.length > 0,
            });
          }}
        >
          {uploadedFiles}
          {fileUpload}
        </DropZone>
      </Stack>
    );
  }
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/DropZone/components/FileUpload/FileUpload.scss
+++ b/src/components/DropZone/components/FileUpload/FileUpload.scss
@@ -1,10 +1,59 @@
 @import '../../../../styles/common';
 
 $fileupload-padding: rem(15px);
+$slim-min-height: rem(30px);
+$slim-vertical-padding: ($slim-min-height - line-height(body) - rem(2px)) / 2;
 
 .FileUpload {
   padding: $fileupload-padding;
   text-align: center;
+}
+
+.Button {
+  @include button-base;
+
+  &.disabled {
+    @include base-button-disabled;
+    cursor: not-allowed;
+    box-shadow: none;
+  }
+
+  &.focused {
+    border-color: color('indigo');
+    outline: 0;
+    box-shadow: 0 0 0 1px color('indigo');
+
+    @include high-contrast-button-outline;
+  }
+
+  .sizeSlim {
+    min-height: $slim-min-height;
+    padding: $slim-vertical-padding spacing(base-tight);
+  }
+}
+
+.ActionTitle {
+  color: color('blue');
+  text-decoration: none;
+
+  &:not(.ActionTitle-disabled) {
+    cursor: pointer;
+
+    &:hover,
+    &:active {
+      color: color('blue', 'dark');
+      text-decoration: underline;
+    }
+  }
+}
+
+.ActionTitle-focused {
+  color: color('blue', 'dark');
+  text-decoration: underline;
+}
+
+.ActionTitle-disabled {
+  color: color('ink', 'lightest');
 }
 
 .Image {
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -15,6 +15,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Fixed animation for Modal when being rendered asynchronously ([#2076](https://github.com/Shopify/polaris-react/pull/2076))
 - Updated `TextField` `min` and `max` type from `number` to `number | string` to allow min/max dates ([#1991](https://github.com/Shopify/polaris-react/pull/1991))
 - Fixed item content from overflowing past the container in `Stack` ([#2071](https://github.com/Shopify/polaris-react/pull/2071))
+- Fixed `Dropzone` hover, disabled and focus states ([#1994](https://github.com/Shopify/polaris-react/pull/1994))
 
 ### Documentation
 
--- a/src/components/DropZone/DropZone.scss
+++ b/src/components/DropZone/DropZone.scss
@@ -3,8 +3,8 @@
 $dropzone-padding: rem(15px);
 $dropzone-border-style: dashed;
 $dropzone-border-color: color('sky');
-$dropzone-background: color('white');
 $dropzone-border-radius: border-radius();
+$dropzone-background: color('white');
 $dropzone-min-height-extra-large: rem(205px);
 $dropzone-min-height-large: rem(160px);
 $dropzone-min-height-medium: rem(100px);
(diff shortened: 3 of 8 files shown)


src/components/DropZone/components/FileUpload/FileUpload.scss
```diff
@@ -1,12 +1,63 @@
...
+
+.ActionTitle {
+  color: color('blue');
+  text-decoration: none;
+  cursor: pointer;
+
+  &:not(.ActionTitleDisabled) {
+    &:hover,
+    &:active,
+    &.ActionTitleFocused {
```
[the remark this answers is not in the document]
kaelig: Gotcha - I made a suggestion to slightly simplify the CSS.

src/components/DropZone/components/FileUpload/FileUpload.scss
```diff
@@ -1,12 +1,61 @@
...
+}
+
+.ActionTitle-focused {
+  color: color('blue', 'dark');
+  text-decoration: underline;
+}
+
+.ActionTitle-disabled {
+  color: color('ink', 'lightest');
+}
```
kaelig: I typically avoid `:not()` unless absolutely needed. `:not()` communicates intent correctly but also adds an additional layer of state management.

(this is an opinion, not something I would say I'm definitely right about)

Suggestion: remove `&:not(.ActionTitle-disabled) {`, add `pointer-events: none;` to `.ActionTitle-disabled`.

```scss
.ActionTitle {
  color: color('blue');
  text-decoration: none;
  cursor: pointer;

  &:hover,
  &:active {
    color: color('blue', 'dark');
    text-decoration: underline;
  }
}

.ActionTitle-focused {
  color: color('blue', 'dark');
  text-decoration: underline;
}

.ActionTitle-disabled {
  color: color('ink', 'lightest');
  pointer-events: none;
}
```

kaelig:
### Hit zone size

Hit/hover zone is out of the boundaries of where it should be ([seen here](https://www.chromaui.com/component?appId=5d559397bae39100201eedc1&name=All%20Components%7CDrop%20zone&buildNumber=169&specName=Nested%20drop%20zone)).

Screen_Shot_2019-08-30_at_4_14_56_PM

### Add file window auto-opening a second time

1. Go to the [Nested Drop Zones example](https://www.chromaui.com/component?appId=5d559397bae39100201eedc1&name=All%20Components%7CDrop%20zone&buildNumber=169&specName=Nested%20drop%20zone)
1. With the keyboard, focus the drop zone
1. Hit <kbd>return</kbd>
1. Select file(s) and hit "Open"

#### Actual

Another "add files" window opens.

#### Expected

No window hapens.

src/components/DropZone/components/FileUpload/FileUpload.scss
dleroux: In this particular case I went with the `:not()`  to avoid having to override the hover and active on both `-focused` and `-disabled` and from having to deal with the double classes on `.ActionTitle-focused.ActionTitle-disabled` or relying on the order of css. I find here the intent is clear, we have just the right amount of specificity to not have to rely on order and this will produce the least about of code if I'm not mistaken. Are you good with that?

dleroux:
It looks like that is expected @kaelig. In that particular example, we are nesting `Dropzones`. Inside the parent `Dropzone`, we also nest a `Card` as a parent of the child `Dropzone`. So that orange area is actually the padding of the Card, which part of the drop area of the outer `Dropzone`.

kaelig:
Thanks, I think it's a tiny bit sloppy but it won't affect the UX for most people. 

What about the second issue in my comment? ("Add file window auto-opening a second time")

dleroux:
Didn’t catch that. I’ll take a loom in the morning. If it is existing I’ll open a separate issue for it. We need to rethink the ux of the dropzone as this pr has uncovered a few additional a11y concerns.

dleroux:
It is existing. I created an issue for it: https://github.com/Shopify/polaris-react/issues/2085

kaelig:
>It is existing. I created an issue for it: #2085

I couldn't reproduce the issue in https://polaris-react.herokuapp.com/iframe.html?id=all-components-drop-zone--nested-drop-zone

Are you sure this isn't a regression in this PR?

dleroux:
I reproduced in on the style-guide.

PR #1997, opened 2019-08-20
[TextField] Converting TextField and Resizer to functional components using React Hooks

dleroux:

change `TextField` and `Resizer`  to use react hooks

### How to 🎩

- Loading the Textfield examples in the playground and ensuring they all behave as expected should be sufficient.

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -1,18 +1,14 @@
-import React from 'react';
+import React, {useState, useEffect, useRef, useCallback} from 'react';
 import {addEventListener} from '@shopify/javascript-utilities/events';
-import {createUniqueIDFactory} from '@shopify/javascript-utilities/other';
 import {CircleCancelMinor} from '@shopify/polaris-icons';
 import {VisuallyHidden} from '../VisuallyHidden';
 import {classNames, variationName} from '../../utilities/css';
-
+import {useI18n} from '../../utilities/i18n';
+import {useUniqueId} from '../../utilities/unique-id';
 import {Labelled, Action, helpTextID, labelID} from '../Labelled';
 import {Connected} from '../Connected';
 
 import {Error, Key} from '../../types';
-import {
-  withAppProvider,
-  WithAppProviderProps,
-} from '../../utilities/with-app-provider';
 import {Icon} from '../Icon';
 import {Resizer, Spinner} from './components';
 import styles from './TextField.scss';
@@ -33,12 +29,6 @@ export type Type =
   | 'currency';
 
 type Alignment = 'left' | 'center' | 'right';
-
-interface State {
-  height?: number | null;
-  focus: boolean;
-  id: string;
-}
 
 export interface BaseProps {
   /** Text to display before value */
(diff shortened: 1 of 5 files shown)


src/components/TextField/TextField.scss
```diff
@@ -29,7 +29,7 @@ $stacking-order: (
   flex-wrap: wrap;
 
   > .Input {
-    overflow: auto;
+    overflow: hidden;
```
dleroux: Not sure why but I think because the effects run after the function has run, the Textarea scrollbar was flickering when multiline.

src/components/TextField/TextField.tsx
```diff
@@ -129,394 +120,357 @@ export type TextFieldProps = NonMutuallyExclusiveProps &
...
+  showCharacterCount,
+  align,
+  onClearButtonClick,
+  onChange,
+  onFocus,
+  onBlur,
+}: TextFieldProps) {
+  const intl = useI18n();
+  const [height, setHeight] = useState<number | null>(null);
+  const [focus, setFocus] = useState(focused || false);
```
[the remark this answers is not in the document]
dleroux: I was wondering the same. Looks like it's essentially just an extra level of abstraction that is less explicit, and in the end does the same?

src/components/TextField/TextField.scss
dleroux: Looks like this won't work. The height is set based on the Resizer size, there's always a delay and the TextField scrolling is janky. Needed to add useLayoutEffect in the Resizer.

src/components/TextField/components/Resizer/Resizer.tsx
```diff
@@ -9,77 +9,72 @@ export interface ResizerProps {
...
-      const {currentHeight, onHeightChange} = this.props;
 
       if (newHeight !== currentHeight) {
         onHeightChange(newHeight);
       }
     });
-  };
+  }, [currentHeight, onHeightChange]);
+
+  useLayoutEffect(() => {
```
dleroux: Need to use this so that the height gets adjusted before committing.

src/components/TextField/components/Resizer/Resizer.tsx
```diff
@@ -9,77 +9,72 @@ export interface ResizerProps {
...
+        minimumLinesNode.current.offsetHeight,
       );
-      const {currentHeight, onHeightChange} = this.props;
 
       if (newHeight !== currentHeight) {
         onHeightChange(newHeight);
       }
     });
-  };
+  }, [currentHeight, onHeightChange]);
```
BPScott: This is going to result in creating a new function every time the currentHeight changes. If we store currentHeight in a ref then I think we'd be able to take it out of the dependency array and thus avoid recreating the function

src/components/TextField/components/Resizer/Resizer.tsx
AndrewMusgrave: `useLayoutEffect` doesn't play well SSR, we'll probably want to rethink this -> https://gist.github.com/gaearon/e7d97cdf38a2907924ea12e4ebdf3c85

AndrewMusgrave:
Sorry Dan I didn't get to this today, I'll 🎩 and review in the morning!

src/components/TextField/components/Resizer/Resizer.tsx
AndrewMusgrave: > If we store currentHeight in a ref then I think we'd be able to take it out of the dependency array and thus avoid recreating the function

Your correct 😄 the reference to the ref will never be stale, just make sure not to deconstruct the value off the ref or that'll have to become a dependency

src/components/TextField/TextField.tsx
```diff
@@ -1,18 +1,15 @@
-import React from 'react';
+import React, {useState, useEffect, useRef, useCallback} from 'react';
 import {addEventListener} from '@shopify/javascript-utilities/events';
-import {createUniqueIDFactory} from '@shopify/javascript-utilities/other';
 import {CircleCancelMinor} from '@shopify/polaris-icons';
 import {VisuallyHidden} from '../VisuallyHidden';
 import {classNames, variationName} from '../../utilities/css';
-
+import {useI18n} from '../../utilities/i18n';
+import {useMountedRef} from '../../utilities/use-mounted-ref';
```
dleroux: I brought this over from `@shopify/react-hooks`. I'm thinking it might be useful to add?

src/components/TextField/components/Resizer/Resizer.tsx
dleroux: I opted for his second option since `useEffect` didn't work for us.

src/components/TextField/TextField.tsx
```diff
@@ -1,18 +1,15 @@
-import React from 'react';
+import React, {useState, useEffect, useRef, useCallback} from 'react';
 import {addEventListener} from '@shopify/javascript-utilities/events';
-import {createUniqueIDFactory} from '@shopify/javascript-utilities/other';
 import {CircleCancelMinor} from '@shopify/polaris-icons';
 import {VisuallyHidden} from '../VisuallyHidden';
 import {classNames, variationName} from '../../utilities/css';
-
+import {useI18n} from '../../utilities/i18n';
+import {useMountedRef} from '../../utilities/mounted-ref';
```
dleroux: same as `@shopify/react-hooks`

src/utilities/tests/mounted-ref.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+      const mounted = useMountedRef();
+      spy(mounted.current);
+      return null;
+    }
+
+    mount(<MockComponent />);
+    expect(spy).toHaveBeenCalledWith(true);
+  });
+
+  it('returns a ref with current value as false when the component is un-mounted', async () => {
```
dleroux: This is the same test used in @shopify/react-hooks except we are getting a warning because of the promise.

`Warning: The callback passed to ReactTestUtils.act(...) function must not return anything.`

Any suggestions on how we can test that unmounting returns false without it?

dleroux:
> Sorry Dan I didn't get to this today, I'll 🎩 and review in the morning!

No problem. I applied @BPScott suggestions and brought over `useMountedRef` . Left you a question regarding the test.

src/utilities/tests/mounted-ref.test.tsx
dleroux: Got this to work, thoughts?

`
it('returns a ref with current value as false when the component is un-mounted', () => {
    const spy = jest.fn((_) => {});

    function MockComponent() {
      const mounted = useMountedRef();

      useEffect(() => {
        return () => {
          // eslint-disable-next-line react-hooks/exhaustive-deps
          spy(mounted.current);
        };
      }, [mounted]);

      return <div />;
    }

    const mockComponent = mount(<MockComponent />);

    mockComponent.unmount();

    expect(spy).toHaveBeenCalledWith(false);
  });
`

src/components/TextField/tests/TextField.test.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import {mountWithAppProvider, findByTestID} from 'test-utilities/legacy';
```
AndrewMusgrave: Any chance you can use `react-testing`? We're trying not to use the legacy stuff anymore and *eventually* migrate our old tests away

src/utilities/mounted-ref.ts
```diff
@@ -0,0 +1,13 @@
+import {useRef, useEffect} from 'react';
+
+export function useMountedRef() {
+  const mounted = useRef(true);
```
AndrewMusgrave: Should this be false? 🤔 
```suggestion
  const mounted = useRef(false);
```

src/utilities/mounted-ref.ts
```diff
@@ -0,0 +1,13 @@
...
+
+  useEffect(() => {
+    return () => {
+      mounted.current = false;
```
AndrewMusgrave: Should this be true?? 🤔 
```suggestion
      mounted.current = true;
```

AndrewMusgrave: Maybe I'm missing something?

src/utilities/tests/mounted-ref.test.tsx
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
+import {mount} from '@shopify/react-testing';
```
AndrewMusgrave:
```suggestion
import {mount} from 'test-utilities';
```

src/utilities/tests/mounted-ref.test.tsx
AndrewMusgrave: Are our type files up to date?

src/components/TextField/components/Resizer/Resizer.tsx
```diff
@@ -9,77 +9,77 @@ export interface ResizerProps {
...
-        {minimumLinesMarkup}
-      </div>
-    );
-  }
-
-  private handleHeightCheck = () => {
-    if (this.animationFrame) {
-      cancelAnimationFrame(this.animationFrame);
+  useEffect(() => {
+    return () => {
```
AndrewMusgrave: If we moved the cleanup to `useLayoutEffect`we could kill this hook

src/components/TextField/TextField.tsx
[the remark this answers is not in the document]
AndrewMusgrave: We could probably initialize this as `useState(focused)` or if we wanted to keep the type as boolean we usually use the boolean constructor `useState(Boolean(focused))`, no strong opinions

src/components/TextField/TextField.tsx
```diff
@@ -129,394 +120,356 @@ export type TextFieldProps = NonMutuallyExclusiveProps &
...
+  const prefixRef = useRef<HTMLDivElement>(null);
+  const suffixRef = useRef<HTMLDivElement>(null);
+  const buttonPressTimer = useRef<number>();
+  const mounted = useMountedRef();
+
+  useEffect(() => {
+    if (!inputRef.current) return;
+    if (focused) {
+      inputRef.current.focus();
+    } else if (focused === false) {
```
AndrewMusgrave: Right now if focused isn't true, it'll be false, won't it? We can probably kill this else if with a return in the if statement or an else or a ternary (focused ? ...focus() : ... blur())

src/components/TextField/TextField.tsx
```diff
@@ -129,394 +120,356 @@ export type TextFieldProps = NonMutuallyExclusiveProps &
...
+      );
+
+      onChange(String(newValue.toFixed(decimalPlaces)), id);
+    },
+    [id, max, min, onChange, step, value],
+  );
+
+  const handleButtonRelease = useCallback(() => {
+    clearTimeout(buttonPressTimer.current);
+  }, [buttonPressTimer]);
```
AndrewMusgrave: We can kill this as a dependency

```suggestion
  }, []);
```

src/components/TextField/TextField.tsx
```diff
@@ -129,394 +120,356 @@ export type TextFieldProps = NonMutuallyExclusiveProps &
...
-          onMouseDown={this.handleButtonPress}
-          onMouseUp={this.handleButtonRelease}
-        />
-      ) : null;
-
-    const style = multiline && height ? {height} : null;
-
-    const resizer = multiline ? (
+  const resizer =
+    multiline && mounted.current ? (
```
AndrewMusgrave: I'm a little concerned about this, since in the hook when useEffect runs this will evaluate as false... but I could be being an airhead 😄

src/utilities/mounted-ref.ts
dleroux: This one makes sense that it's false because it's for the clean-up?

AndrewMusgrave: 🤦‍♂ Didn't see it was in the returned function 😄

PR #2000, opened 2019-08-21
[PopoverOverlay] Fix regression introduced by removal of `CSSTransition` 

loic-d:

This PR fixes a regression in `PopoverOverlay` introduced by https://github.com/Shopify/polaris-react/pull/1756

The `onClose` prop was fired when clicking or touching the element activating the `Popover` when that element was not the `activator` received as a prop. 

In this case, the `Popover` would close right away.

It feels like a weird use case but that's how the `ImageEditor` in web is using `Popover`. The `activator` prop is just an empty `div`. The real Popover triggers are rendered higher up in the DOM tree.

Screen Shot 2019-08-21 at 4 31 58 PM

This PR fixes the regression.

In `handleClick`, we prevent `onClose` to be called when the source of the click is a descendant of the `Popover` or the `Popover` is being opened (this one was not working as expected).

We are now properly setting the transition status to `Entering` instead of `Entered` when the `Popover` is opened. It prevents clicks that are not coming from descendants to fire `onClose` while the `Popover` is opening.

## 

### How to 🎩

Test the Playground code **with and without the changes**.

Before: `closePopover` is fired when opening the `Popover`.
**After: `closePopover` won't be fired when opening the `Popover` (fix)**

<details>

```jsx
import React from 'react';
import {ActionList, Button, Popover} from '@shopify/polaris';

export default class PopoverContentExample extends React.Component {
  state = {
    active: false,
  };

  openPopover = () => {
    console.log('openPopover called');
    this.setState({active: true});
  };

  closePopover = () => {
    console.log('closePopover called');
    this.setState({active: false});
  };

  render() {
    return (
      <div>
        <Button onClick={this.openPopover}>Toggle `Popover`</Button>
          <Popover
            active={this.state.active}
            activator={<div />}
            onClose={this.closePopover}
          >
            Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nulla vitae
            orci tellus. Pellentesque non elementum ipsum, at feugiat libero.
            Nulla id mollis nisi. Nam ut ligula mollis, commodo felis ut,
            consequat orci. In in sagittis magna, vitae rutrum mi. Ut facilisis,
            quam eu pulvinar elementum, lorem mauris lobortis leo, a posuere
            erat tortor in ante. Nunc fermentum nulla posuere, tincidunt velit
            id, pharetra velit. Ut congue varius semper. Nullam cursus
            ullamcorper sem, cursus euismod diam mollis id. Nullam sodales nibh
            libero, id auctor lectus condimentum eu. Curabitur dui velit, ornare
            sed vestibulum at, egestas at ante. Proin feugiat eros metus.
            Integer lacinia urna a orci accumsan, eget dignissim turpis pretium.
            Cras feugiat, mi convallis facilisis venenatis, odio lectus
            hendrerit velit, in sollicitudin magna nisi quis lacus. Proin purus
            nunc, posuere eu leo ultrices, hendrerit interdum massa. In non
            felis id justo egestas cursus. Pellentesque euismod quis sem in
            luctus. Nullam ultrices consequat nibh ac cursus. Morbi blandit
            finibus velit, ac scelerisque diam viverra vitae. In a orci orci.
          </Popover>
        </div>
      </div>
    );
  }
}


```

</details>

🎩 In `web`:
- Start the `web` server
- Clone `polaris-react` and checkout `popover-overlay-fix`
- `yarn` then build a consumer for web `yarn run build-consumer web` (still in `polaris-react`)
- Open a Product details page (React) with an image (or upload one)
- Edit the image but do not save it
- Click the `<` (back) button in the left panel, the `X` of the Modal or the 'Undo all' button
- Make sure the leave confirmation is rendered and works as expected. You should also be able to close it.

Web issue for context: https://github.com/Shopify/web/issues/17477

### 🎩 checklist

### DIFF
--- a/src/components/Popover/components/PopoverOverlay/PopoverOverlay.tsx
+++ b/src/components/Popover/components/PopoverOverlay/PopoverOverlay.tsx
@@ -62,6 +62,8 @@ export default class PopoverOverlay extends React.PureComponent<Props, State> {
   };
 
   private contentNode = createRef<HTMLDivElement>();
+  private enteringTimer?: number;
+  private exitingTimer?: number;
 
   changeTransitionStatus(transitionStatus: TransitionStatus, cb?: () => void) {
     this.setState({transitionStatus}, cb);
@@ -81,17 +83,24 @@ export default class PopoverOverlay extends React.PureComponent<Props, State> {
   componentDidUpdate(oldProps: Props) {
     if (this.props.active && !oldProps.active) {
       this.focusContent();
-
-      this.changeTransitionStatus(TransitionStatus.Entered);
+      this.changeTransitionStatus(TransitionStatus.Entering, () => {
+        this.enteringTimer = window.setTimeout(() => {
+          this.setState({transitionStatus: TransitionStatus.Entered});
+        }, durationBase);
+      });
     }
 
     if (!this.props.active && oldProps.active) {
       this.changeTransitionStatus(TransitionStatus.Exiting, () => {
-        setTimeout(() => {
+        this.exitingTimer = window.setTimeout(() => {
           this.setState({transitionStatus: TransitionStatus.Exited});
         }, durationBase);
       });
     }
+  }
+
+  componentWillUnmount() {
+    this.clearTransitionTimeout();
   }
 
   render() {
@@ -128,6 +137,16 @@ export default class PopoverOverlay extends React.PureComponent<Props, State> {
         classNames={className}
       />
     );
+  }
+
+  private clearTransitionTimeout() {
+    if (this.enteringTimer) {
+      window.clearTimeout(this.enteringTimer);
+    }
+
+    if (this.exitingTimer) {
+      window.clearTimeout(this.exitingTimer);
+    }
   }
 
   private focusContent() {
--- a/src/components/Popover/tests/Popover.test.tsx
+++ b/src/components/Popover/tests/Popover.test.tsx
@@ -147,4 +147,39 @@ describe('<Popover />', () => {
     window.dispatchEvent(evt);
     expect(spy).toHaveBeenCalled();
   });
+
+  it('does not call onClose when Popover is opening and trigger was not the activator', () => {
+    class PopoverWithDisconnectedActivator extends React.Component {
+      state = {
+        active: false,
+      };
+
+      handleActivatorClick = () => this.setState({active: true});
+
+      render() {
+        return (
+          <React.Fragment>
+            <button onClick={this.handleActivatorClick}>Activator</button>
+            <Popover
+              active={this.state.active}
+              activator={<div />}
+              onClose={onCloseSpy}
+            />
+          </React.Fragment>
+        );
+      }
+    }
+
+    const onCloseSpy = jest.fn();
+
+    const popoverWithDisconnectedActivator = mountWithAppProvider(
+      <PopoverWithDisconnectedActivator />,
+    );
+
+    popoverWithDisconnectedActivator.find('button').simulate('click');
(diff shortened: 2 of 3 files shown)


src/components/Popover/components/PopoverOverlay/PopoverOverlay.tsx
```diff
@@ -81,8 +81,11 @@ export default class PopoverOverlay extends React.PureComponent<Props, State> {
   componentDidUpdate(oldProps: Props) {
     if (this.props.active && !oldProps.active) {
       this.focusContent();
-
-      this.changeTransitionStatus(TransitionStatus.Entered);
+      this.changeTransitionStatus(TransitionStatus.Entering, () => {
```
loic-d: Setting `Entering` ensures `onClose` is not fired in `handleClick`.

src/components/Popover/components/PopoverOverlay/PopoverOverlay.tsx
```diff
@@ -81,8 +81,11 @@ export default class PopoverOverlay extends React.PureComponent<Props, State> {
...
+        setTimeout(() => {
```
jgodson: I think we’d need to ensure this timeout gets cleared if the component gets unmounted?

tmlayton:
What about the mount lifecycle? While there is no bug, it has the same hole where the state will be `Entered` while it is technically still `Entering`.

tmlayton:
Otherwise looks good! But as @amrocha mentioned needs a test 🤖

src/components/Popover/components/PopoverOverlay/PopoverOverlay.tsx
tmlayton: Yeah probably a good practice to do this

loic-d: Totally. I'll clear the newly added `setTimeout` (entering) and the existing one (exiting).

src/components/Popover/components/PopoverOverlay/PopoverOverlay.tsx
```diff
@@ -81,19 +83,26 @@ export default class PopoverOverlay extends React.PureComponent<Props, State> {
...
-        setTimeout(() => {
+        this.exitingTimer = window.setTimeout(() => {
           this.setState({transitionStatus: TransitionStatus.Exited});
         }, durationBase);
       });
     }
   }
 
+  componentWillUnmount() {
+    this.clearTransitionTimeout();
```
loic-d: Now clearing the transition timers when the component unmounts.

src/components/Popover/tests/Popover.test.tsx
```diff
@@ -147,4 +147,39 @@ describe('<Popover />', () => {
     window.dispatchEvent(evt);
     expect(spy).toHaveBeenCalled();
   });
+
+  it('does not call onClose when Popover is opening and trigger was not the activator', () => {
```
loic-d: Added a test replicating the bug we encountered.

loic-d:
@tmlayton Regarding the mount lifecycle, I believe we should not update the transition status all together. The component can be mounted but not visible `active = false`. WDYT? @AndrewMusgrave might have more context on this one.

loic-d:
@tmlayton @amrocha

I updated the PR:
- Clear transition timer when the component un-mounts
- Add test to make sure the regression was fixed
- Improve release note

If you guys can take another 👀 at it today that would be great!

loic-d:
Fix has been release in `rc5`

BPScott:
Fixed c6d321bc

PR #2060, opened 2019-08-29
[TextField] Fix type for button

chenxsan:

The default `type` for button is [`submit`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type), which means whenever you click the clearButton, it will fire `submit` event on the form.

Here's an example [on codesandbox](https://codesandbox.io/s/pensive-grass-t45xl).

It sets the type to `button` to fix the bug.

### 🎩 checklist


chenxsan:
@danrosenthal done.

tmlayton:
Thanks @chenxsan!

PR #2063, opened 2019-08-30
[WIP][Color system] deprecate sass API

### LINKED ISSUE #2017: [Color system] Deprecate Sass API
#### Border width
- border-width (⚠️ needs token values before deprecation)

#### Borders
- border (⚠️ alias safe to deprecate once border width has tokens)

#### Colors
- color (👍 safe to deprecate, use tokens directly, but maybe should wait until after color system)
- color-multiply (👍 safe to deprecate, copy to project)
- ms-high-contrast-color (⚠️ has foundational color data)

#### Duration
- duration (👍 safe to deprecate, use tokens directly)

#### Easing
- easing (⚠️ needs token values before deprecation)

#### Filters
- filter (👍 safe to deprecate, use tokens directly, weird because tokens use same name as colors)

#### Layout
- layout-width (⚠️ has foundational data, needs token values before deprecation)
- top-bar-height (⚠️ could be considered foundational data)
- mobile-nav-width (⚠️ does math on `dismiss-icon-size`, which could be considered foundational data)
-nav-min-window-corrected (⚠️ references an existing layout width, which would need to be made a token before deprecation)

#### Shadows
- shadow (⚠️ needs token values before deprecation)

#### Spacing
- spacing (👍 safe to deprecate, use tokens directly)

#### Typography
- font-family (👍 safe to deprecate, use tokens directly)
- line-height (⚠️ has foundational data, needs discussion if it should be a token)
- font-size (⚠️ has foundational data, needs discussion if it should be a token)

#### Utilities
- rem (👍 safe to deprecate, no token concerns)
- px (👍 safe to deprecate, no token concerns)
- em (👍 safe to deprecate, no token concerns)
- available-names (⚠️ each `foundation` function depends on this)
- map-extend (👍 only `color` depends on this, safe to deprecate)

#### Z-index
- z-index (⚠️ has foundational data, needs discussion if it should be a token)

### Shared
These are all safe to deprecate (👍) without token concerns excluding `state` and `thumbnail-size`, which need a bit of discussion, and should probably be in `foundation` instead of `shared`. Should also discuss if `control-height` and `control-slim-height` should be considered foundational.

#### Accessibility
- Visually-hidden

#### Breakpoints
- breakpoint
- page-content-breakpoint-before
- page-content-breakpoint-after
- breakpoint-after
- breakpoint-before
- frame-with-nav-when-not-max-width
- page-when-not-max-width
- page-content-when-layout-stacked
- page-content-when-layout-not-stacked
- page-content-when-partially-condensed
- page-content-when-not-partially-condensed
- page-content-when-fully-condensed
- page-content-when-not-fully-condensed
- frame-when-nav-displayed
- frame-when-nav-hidden

#### Button
- high-contrast-button-outline
- button-base
- base-button-disabled
- button-filled
- button-filled-disabled
- button-outline
- button-outline-disabled
- button-full-width
- plain-button-background
- plain-button-backdrop
- unstyled-button

#### Controls
- control-height (⚠️ could be considered foundational data)
- control-slim-height (⚠️ could be considered foundational data)
- control-vertical-padding
- control-icon-transition
- control-backdrop

#### Forms
- unstyled-input
- range-track-selectors
- range-thumb-selectors

#### Icons
- icon-size (⚠️ could be considered foundational data)
- recolor-icon
- color-icon (⚠️ dependencies on `color` and `filter`, ask Dave Newton)

#### Interaction State
- state (⚠️ has external data, replaced by color system, should be considered foundational)

#### Layout
- layout-flex-fix
- safe-area-for
- after-topbar-sheet

#### Links
- unstyled-link

#### Lists
- unstyled-list

#### Page
- page-padding-not-fully-condensed
- page-padding-not-partially-condensed
- page-layout
- page-content-layout
- page-title-layout
- page-header-layout
- page-header-has-navigation
- page-header-without-navigation
- page-header-has-secondary-actions
- page-actions-layout

#### Printing
- when-printing
- when-not-printing
- hidden-when-printing

#### Skeleton
- thumbnail-size (⚠️ could be considered foundational data)
- skeleton-shimmer
- skeleton-content
- skeleton-page-secondary-actions-layout
- skeleton-page-header-has-secondary-actions
- skeleton-page-header-layout

#### Typography
- when-typography-not-condensed (⚠️ foundational data)
- when-typography-condensed (⚠️ foundational data)
- text-style-caption
- text-style-heading
- text-style-subheading
- text-style-input
- text-style-body
- text-style-button
- text-style-button-large
- text-style-display-x-large
- text-style-display-large
- text-style-display-medium
- text-style-display-small
- text-emphasis-placeholder
- text-emphasis-subdued
- text-emphasis-strong
- text-emphasis-normal
- text-breakword
- truncate
- print-hidden

danrosenthal:
- Addresses #2017
- Depends on https://github.com/Shopify/polaris-tokens/pull/76

---

## Todo

- replace sass data that is now in tokens with respective tokens
- split safe deprecations into own PR and :ship:


BPScott:
As we decouple and deprecate our internal Sass API I'd like to consider if we should still be using our helper functions `spacing(tight)` etc, or if we should use variables from polaris-tokens directly (e.g.`$spacing-tight`). Personally I'm leaning towards the latter in most cases - color can get a stay of execution as we know we'll be massively reworking it later.

If we decide on the latter then there's not much value in duplicating those functions if we no longer wish to use them internally. I'd like to see us changing our current code to using tokens directly from polaris-tokens before we deprecated the functions, otherwise we're making extra work for ourselves in the long run.

tmlayton:
> I'd like to consider if we should still be using our helper functions spacing(tight) etc, or if we should use variables from polaris-tokens directly (e.g.$spacing-tight).

The advantage to functions is it gives us a level of indirection. Particularly if we ever want to do something more like an if statement, augment the value, etc without having to update every instance where the value is used. For that reason I’m partial to functions, but could be persuaded.

danrosenthal:
> The advantage to functions is it gives us a level of indirection. Particularly if we ever want to do something more like an if statement, augment the value, etc without having to update every instance where the value is used. For that reason I’m partial to functions, but could be persuaded.

I feel the same way. I like that functions give us the ability to transform a `px` value token to a `rem` value behind the scenes, for example.

Also, we can always remove functions and replace with raw tokens at any point. It's no longer a breaking change, since the API is private now.

src/styles/common/_layout.scss
```diff
@@ -1,3 +1,77 @@
...
+  outer-spacing: (
+    min: spacing(loose),
+    max: spacing(extra-loose),
+  ),
+);
+
+/// Returns the widths of the specified column.
+/// @param {String} $name - The column name.
+/// @return {Number} The width for the column.
+/// @deprecated Pull data from sizes and spacing Polaris tokens
```
danrosenthal: This went in during copy->paste

src/styles/common/_layout.scss
```diff
@@ -1,3 +1,77 @@
...
+  @if type-of($fetched-value) != null {
+    @return $fetched-value;
+  } @else {
+    @error 'Column `#{$name} - #{$value}` not found. Available columns: #{available-names($layout-width-data)}';
+  }
+}
+
+$dismiss-icon-size: 32px;
+
+/// @deprecated Copy function into consuming project
```
danrosenthal: This went in during copy->paste

src/styles/common/_layout.scss
```diff
@@ -1,3 +1,77 @@
...
+@function top-bar-height() {
+  @return rem(56px);
+}
+
+/// @deprecated Copy function into consuming project
```
danrosenthal: This went in during copy->paste

src/styles/common/_layout.scss
```diff
@@ -1,3 +1,77 @@
...
+@function mobile-nav-width() {
+  @return calc(100vw - #{rem($dismiss-icon-size) + spacing() * 2});
+}
+
+/// @deprecated Copy function into consuming project
```
danrosenthal: This went in during copy->paste

BPScott:
> The advantage to functions is it gives us a level of indirection. Particularly if we ever want to do something more like an if statement, augment the value, etc without having to update every instance where the value is used. For that reason I’m partial to functions, but could be persuaded.

It feels bad to say "we're taking away a layer of indirection, you should use our tokens directly" to other teams, but then instantly say for our own usage "using the tokens directly isn't actually good enough, we need a layer of indirection".

The tokens that we provide should be usable out-the-box by anybody and that includes ourselves IMO, otherwise we're not eating our own dogfood and we're signaling that the underlying tokens aren't actually fit enough for general usage.

danrosenthal:
> Do the @deprecated docblocks get surfaced anywhere at build time so people know they should stop using them?

> Should we add an @warn message when these functions/mixins get called to make it very clear in the build output? Or can that be a stage 2 harder-deprecation

Great questions, let's discuss in our meeting tomorrow.

danrosenthal:
Closed in favor of https://github.com/Shopify/polaris-react/issues/1982

Issue #1995, opened 2019-08-20
Post v4 cleanup work


BPScott:
This is a meta issue to collate what I'd like to do as tidy-up work in a post-v4 world (in v4.1.0, v4.2.0 and beyond).

I'd like for us to do three things, in vague order of priority. Much of this conversion work has several quick wins for components where a migration to the new style is easy, we should prioritize that easy work, and punt on the fewer harder migrations. That way our first timebox of work shall feel like good progress and make future planning easier (as there are fewer items to deal with overall)

###  DONE! ~Convert all examples to use hooks instead of classes.~

Shopify is rallying around hooks as the new best practice - we should help ensure people follow that by giving up to date examples.

Searching for "extends React." in README.md files should return no results.

@AndrewMusgrave made a pretty big start on this in https://github.com/Shopify/polaris-react/compare/hookify-examples but I'm not sure how polished that branch is. It might be worth splitting off chunks of that branch into separate PRs (5 or so README files at a time depending on the ease of the comparison) and reviewing them individually so the changes are more digestible.

### Convert existing code to use hooks instead of classes

Hooks are the way forward and we should leverage them. We should prioitize components that use withAppProvider (as we want to remove that HoC), and small components where making the change is easy.

@AndrewMusgrave made a start on this in https://github.com/Shopify/polaris-react/compare/fast-hookifies but I am mot sure how polished that branch is. As above, splitting this out into multiple PRs for ease of review would be useful

### Done ~Migrate our tests from `enzyme` to use `@shopify/react-testing`~

Dedicated issue: https://github.com/Shopify/polaris-react/issues/4274

This is a bunch of work with not much value behind it (aside from enzyme being often low-key annoying), this can probably be deferred and deferred until the end of time.

One the up side @vsumner has wrote a codemod in web that will bulk convert a codebase so that'd be worth investigating.

BPScott:
# Audits for hooks conversion

To help identify where we can focus for quick wins, here's an audit of the difficulty of migrating components (half done, I still need to finish off several files).

Files using withAppProvider are a priority, then easy other class-based components

Difficulty is a 1 to 5 scale, with 1 being easiest, and 5 being hardest.
- 1 is a component with a state or two that need to be convered to use useState
- 2 is a component with a few states and a few callback handlers that need wrapping in useCallback
- 3 is a component with lifecycle hooks (didMount etc) that will need writing a useEffect
- 4 is somewhere between 3 and 5
[...]

PR #2065, opened 2019-08-30
Convert a few simple components to use hooks

BPScott:
*This PR is best viewed if you disable whitespace changes, as the hookification saves a level of indentation, which makes it look like i changed EVERYTHING*

Starting down the path of removing withAppProvider usage into the new world of hooks.
Part of #1995

Remove withAppProvider, and use hooks for the following components:

ActionMenu/RollupAction
Autocomplete
Card
EmptySearchResult
Form
SkeletonPage
TopBar

### How to 🎩

- ActionMeni/RollupAction: Confirm ... popover on with Page with action group example at small screen sizes renders and toggles on click
- Autocomplete: Check interactions on any Autocomplete example
- Card: Confirm secondary actions popover still works on Card
- EmptySearchResult: Check `<EmptySearchResult title="Title" description="Description" withIllustration />` renders in playground
- Form: Check any form example
- SkeletonPage: Check and skeleton page example
- Topbar: check navigation visibility toggling works in Frame example at small screen sizes

### DIFF
--- a/src/utilities/use-toggle.ts
+++ b/src/utilities/use-toggle.ts
@@ -1,0 +1,24 @@
+import {useState, useCallback} from 'react';
+
+export function useToggle(initialState: boolean) {
+  const [state, setState] = useState(initialState);
+  const toggle = useCallback(() => setState((state) => !state), []);
+
+  // cast needed to say this returns a two item array with the items in
+  // their specific positions instead of `(typeof state | typeof toggle)[]`
+  return [state, toggle] as [typeof state, typeof toggle];
+}
+
+export function useForcableToggle(initialState: boolean) {
+  const [state, setState] = useState(initialState);
+
+  const toggles = {
+    toggle: useCallback(() => setState((state) => !state), []),
+    forceTrue: useCallback(() => setState(true), []),
+    forceFalse: useCallback(() => setState(false), []),
+  };
+
+  // cast needed to say this returns a two item array with the items in
+  // their specific positions instead of `(typeof state | typeof toggles)[]`
+  return [state, toggles] as [typeof state, typeof toggles];
+}
--- a/src/utilities/use-toggle.ts
+++ b/src/utilities/use-toggle.ts
@@ -9,7 +9,7 @@ export function useToggle(initialState: boolean) {
   return [state, toggle] as [typeof state, typeof toggle];
 }
 
-export function useForcableToggle(initialState: boolean) {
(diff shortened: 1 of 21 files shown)


src/utilities/use-toggle.ts
```diff
@@ -0,0 +1,24 @@
...
+export function useToggle(initialState: boolean) {
+  const [state, setState] = useState(initialState);
+  const toggle = useCallback(() => setState((state) => !state), []);
+
+  // cast needed to say this returns a two item array with the items in
+  // their specific positions instead of `(typeof state | typeof toggle)[]`
+  return [state, toggle] as [typeof state, typeof toggle];
+}
+
+export function useForcableToggle(initialState: boolean) {
```
amrocha: I think this is a bit of a stretch, I would just do this manually in the file itself and I don't think it comes up often enough to warrant a custom hook. Idk, feels a bit weird, if others think it's fine I'm fine with it though

BPScott: I've saw cases where we want to force set a boolean toggle to false (e.g. when you click a modal's close button / leave focus in popovers.

I think this crops up often enough to make a general purpose thing.

src/utilities/use-toggle.ts
```diff
@@ -0,0 +1,24 @@
+import {useState, useCallback} from 'react';
+
+export function useToggle(initialState: boolean) {
```
dleroux: I can see why you did this, I just wonder it's worth it. It's simple but it requires knowledge about the hook and it's less declarative and just handling the toggle when needed. We aren't saving on much here imo.

BPScott: The same can be said for pretty much all of our custom hooks :)

IMO The value from these small utilities comes from encapsulation, that small amount of complexity saved adds up over the course of several components and its repeated presence means we'll become familiar with it over time for when we have a boolean toggle, which is pretty often.

BPScott: The same can be said for pretty much all of our custom hooks :)

IMO The value from these small utilities comes from encapsulation, that small amount of complexity saved adds up over the course of several components and its repeated presence means we'll become familiar with it over time for when we have a boolean toggle, which is pretty often.

amrocha: I think for a really common usecase like a toggle variable it's fine. I agree that every hook adds to the cognitive load needed to understand the codebase, and we should avoid them

PR #2070, opened 2019-09-03
Update to latest Prettier and reformat

BPScott:

This version of prettier formats hooks a bit more cleanly which is handy
as we're going to be using them a lot

Updates to latest prettier, and runs `yarn run format` over the content.
One manual fix to Sheet.tsx had to be applied as prettier's formatting and ignoring the exhaustive-deps check didn't play nice.

## Tophatting

- Check lint/tests pass
- Review code while supressing whitespace-only changes to make it easier to review (a bit over half of this PR is whitespace changes)

### DIFF
--- a/src/components/Sheet/Sheet.tsx
+++ b/src/components/Sheet/Sheet.tsx
@@ -70,13 +70,11 @@ export function Sheet({
     [mobile],
   );
 
-  useEffect(
-    () => {
-      handleResize();
-    },
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [],
-  );
+  // eslint-disable react-hooks/exhaustive-deps
+  useEffect(() => {
+    handleResize();
+  }, [handleResize]);
+  // eslint-enable react-hooks/exhaustive-deps
 
   return (
     <Portal idPrefix="sheet">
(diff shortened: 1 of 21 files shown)


src/components/Sheet/Sheet.tsx
```diff
@@ -70,13 +70,11 @@ export function Sheet({
...
   );
 
-  useEffect(
-    () => {
-      handleResize();
-    },
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [],
-  );
+  // eslint-disable react-hooks/exhaustive-deps
```
amrocha: What's the issue here? can you disable prettier instead? I'd feel better disabling the styling tool than the linter

BPScott: We were originally disabling the `react-hooks/exhaustive-deps` rule because we were explictly saying the dependency array is empty, while linting things it should contain `handleResize` and prettier didn't like having that disable-next-line comment in the middle of the code.

It looks like vscode not spotting linting config updates means the handleResize had snuck back in when I wasn't looking. After a vscode restart it seems like we can keep to ignoring the exhaustive deps rule without using the block formatting.

---

Not specifying this feels like a potential subtle bug but I'd like to keep this PR focused on formatting updates only

PR #2071, opened 2019-09-03
[Stack] Fix stack item content from overflowing past the container

LauraAubin:

Fixes https://github.com/Shopify/polaris-react/issues/1707

Content inside a Stack item is overflowing past the container.

image

This PR removes the `Item` class.

image

I also tried using `calc(100% - #{spacing($spacing)})` which fixed this specific issue, but according to https://github.com/Shopify/polaris-react/commit/ab23144519d5793e41f0a4933d75f077d972fdae, would reintroduce a bug with Safari.

Another possible solution was to set the `flex shrink` value to 1 (without changing the `min width`), but with multiple items and a smaller container, this caused some overlap:

image

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page, Layout, Card, Stack, Button} from '../src';

  return (
    <Page title="Playground">
      <Layout>
        <Layout.Section>
          <Card title="Title">
            <Card.Section>
              <Stack wrap={false}>
                <img src="https://s3.amazonaws.com/pix.iemoji.com/images/emoji/apple/ios-12/256/snow-capped-mountain.png" />
                <Stack>
                  <p>Short sentence.</p>
                  <p>
                    Long sentence 1. Long sentence 2. Long sentence 3. Long
                    sentence 4. Long sentence 5. Long sentence 6. Long sentence
                    7. Long sentence 8.
                  </p>
                  <Button>Button</Button>
                </Stack>
              </Stack>
            </Card.Section>
          </Card>
        </Layout.Section>
      </Layout>
    </Page>
  );
}

```

</details>

### 🎩 checklist

* [x] Tested on [multiple browsers](https://help.shopify.com/en/manual/intro-to-shopify/shopify-admin/supported-browsers) (Checked Chrome, Safari, Firefox)

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -14,6 +14,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - Fixed animation for Modal when being rendered asynchronously ([#2076](https://github.com/Shopify/polaris-react/pull/2076))
 - Updated `TextField` `min` and `max` type from `number` to `number | string` to allow min/max dates ([#1991](https://github.com/Shopify/polaris-react/pull/1991))
+- Fixed item content from overflowing past the container in `Stack` ([#2071](https://github.com/Shopify/polaris-react/pull/2071))
 
 ### Documentation
 
--- a/src/components/Stack/Stack.scss
+++ b/src/components/Stack/Stack.scss
@@ -104,8 +104,7 @@
 }
 
 .Item {
-  flex: 0 0 auto;
-  min-width: 0;
+  flex: 0 1 auto;
 }
 
 .Item-fill {


dleroux:
So essentially what we are doing is changing flex-shrink to 1 because the default is `flex: 0 1 auto`

I top hatted in various places in web and it looks fine. Still feels a little risky but overall I think it's ok. Let's just remember this one. 

There's a conflict with the unreleased file but apart from that good to ship! Great job!

LauraAubin:
I feel like it would be best to be explicit and set `flex: 0 1 auto`. In the case that this creates future issues, it might make debugging easier if its more obvious whats happening. I'm impartial as its more of a personal preference.

dleroux:
Agreed. Might as well leave the `min-width: 0` as well in that case. I did a blame and it's been there for a while.

LauraAubin:
The min width causes another overflow issue, I briefly tophatted and there appears to be no breaking changes without it.

mbaumbach:
@LauraAubin @dleroux Just an FYI, the removal of the `min-width` does seem to cause a regression in some cases, see #2260. A flex item defaults to have a `min-width` of `auto`, which could cause overflowing issues. Some more context on that here: https://makandracards.com/makandra/66994-css-flex-and-min-width. I'm not sure re-introducing the `min-width` is the appropriate change in all cases either, but figured I'd let you know about the issue I was seeing.

dleroux:
Thanks @mbaumbach!

PR #2073, opened 2019-09-03
[Link] Fix documentation and examples

kaelig:

Fixes https://github.com/Shopify/polaris-styleguide/issues/2942

- Breaks Link monochrome examples in two (as there should only be one code block per example)
- Improves a do/don't section


kaelig:
@mirualves you're the one who flagged this, can you have a look?

Issue #792, opened 2018-06-28
[a11y] Resource list has labeling and focus issues


dpersing:
## Issue summary

The [Resource List](https://polaris.shopify.com/components/lists-and-tables/resource-list#navigation) implementation may present some challenges to screen reader and keyboard-only users.

#### "Hidden" controls are available

- On desktop, checking the bulk checkbox (`Showing 2 customers`) or the bulk button at smaller breakpoints (`Select`) with the keyboard (<kbd>space</kbd>) causes focus to disappear from the page. (It's actually gone to the off-screen checkbox/button that's been replaced by the new visible bulk actions, which are layered using `z-index`.) Pressing <kbd>tab</kbd> causes it to reappear on the visible checked checkbox, but users may be confused about where focus has gone.
- The off-screen bulk checkbox or button is readable and focusable by screen readers, which may cause confusion for non-visual screen reader users.
- If the bulk checkbox is unchecked with the keyboard and no options are checked, focus moves off-screen. Pressing <kbd>tab</kbd> moves focus to the first resource list item link.

```
<div class="Polaris-ResourceList__SelectButtonWrapper">
  <button type="button" class="Polaris-Button Polaris-Button--disabled" disabled="">
    <span class="Polaris-Button__Content">
      <!-- ... -->
      <span>Select</span>
    </span>
  </button>
</div>

<!-- ... -->

<div class="Polaris-ResourceList__BulkActionsWrapper"><!-- ... --></div>
```

```
.Polaris-ResourceList__SelectButtonWrapper {
    position: relative;
    -ms-flex: none;
    flex: none;
}

.Polaris-ResourceList__BulkActionsWrapper {
    position: relative;
    z-index: 2;
    width: 100%;
}
```

#### Labels are unclear

- If there is a mixed state, the visible mixed state checkbox label is not reflected as part of the label that is read on focus. This may cause confusion for sighted keyboard users, especially in the context of the additional off-screen checkbox, as well as challenges for speech activation users.

![Mixed state checkbox](https://user-images.githubusercontent.com/1462085/42062509-9e588bae-7ae2-11e8-9212-43948928c366.png)

```
<label class="Polaris-Choice Polaris-Choice--labelHidden" for="Checkbox131">
  <!-- ... -->
[...]

PR #2077, opened 2019-09-04
[ResourceItem] add name props for accessibility

dleroux:

Fixes part of: https://github.com/Shopify/polaris-react/issues/792

In order to meet accessibility requirements we need to provide the ability to identify the item being rendered.

There is currently and accessibilityLabel prop on the ResourceItem which is used for the entire list items and our examples give `View details for ${name}` as example values.

However, this accessibility is used for the Checkbox which results in:

`Select: View details for ${name}`.

This PR tackled adding the same approach to the disclosure button of the persistentActions, which is what highlighted the issue.

The `name` referred to in our example is not a prop used by the component. It is simply part of the examples. In order to not break the current implementation, I added the name `prop` to the example in order to properly construct the strings. 

I also updated the examples.

## 

### How to 🎩

Using the following playground code ensure that:

1. Using the web inspector ensure the label on the item checkbox reads: `Select: ${name}` or `Deselect: ${name}` accordingly

2. Ensure that the label on the disclosure button (3 dots that reveal the action popover on small screens) has an aria-label of `Actions for ${name}` 

<details>

```jsx
import React from 'react';
import {Page, ResourceList, Card, Avatar, TextStyle, Checkbox} from '../src';

  return (
    <Page title="Playground">
      <ResourceListExample />
    </Page>
  );
}

  state = {
    selectedItems: [],
  };

  };

  renderItem = (item) => {
    return (
      <ResourceList.Item
        id={id}
        url={url}
        media={media}
        name={name}
        shortcutActions={[{content: 'action'}]}
        persistActions
      >
        <h3>
        </h3>
        <div>{location}</div>
      </ResourceList.Item>
    );
  };

  render() {
    const resourceName = {
      singular: 'customer',
      plural: 'customers',
    };
    const items = [
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
      },
    ];
    const promotedBulkActions = [
      {
        content: 'Edit customers',
      },
    ];
    const bulkActions = [
      {
        content: 'Add tags',
      },
      {
        content: 'Remove tags',
      },
      {
        content: 'Delete customers',
      },
    ];
    return (
      <React.Fragment>
        <Card>
          <ResourceList
            resourceName={resourceName}
            items={items}
            renderItem={this.renderItem}
            bulkActions={bulkActions}
          />
        </Card>
      </React.Fragment>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/ResourceItem/tests/ResourceItem.test.tsx
+++ b/src/components/ResourceItem/tests/ResourceItem.test.tsx
@@ -10,6 +10,7 @@ import {
   Checkbox,
   Thumbnail,
   UnstyledLink,
+  Button,
 } from 'components';
 import {ResourceListContext} from '../../../utilities/resource-list';
 import ResourceItem from '../ResourceItem';
@@ -40,6 +41,7 @@ describe('<ResourceItem />', () => {
   const itemId = 'itemId';
   const selectedItemId = 'selectedId';
   const accessibilityLabel = 'link anchor aria-label';
+  const name = 'item name';
 
   const mockSelectableContext = {
     ...mockDefaultContext,
@@ -84,6 +86,50 @@ describe('<ResourceItem />', () => {
       expect(item.find(UnstyledLink).prop('aria-label')).toBe(
         accessibilityLabel,
       );
+    });
+  });
+
+  describe('name', () => {
+    it('is used as the Checkbox label', () => {
+      const item = mountWithAppProvider(
+        <ResourceListContext.Provider value={mockSelectableContext}>
+          <ResourceItem
+            accessibilityLabel={accessibilityLabel}
+            id={itemId}
+            url="https://shopify.com"
+            name={name}
+          />
+        </ResourceListContext.Provider>,
+      );
+
+      const expectedLabel = name;
+
+      expect(item.find(Checkbox).prop('label')).toBe(expectedLabel);
+    });
+
+    it('is used on the disclosure action menu when there are persistent actions', () => {
+      const item = mountWithAppProvider(
+        <ResourceListContext.Provider value={mockSelectableContext}>
+          <ResourceItem
+            accessibilityLabel={accessibilityLabel}
+            id={selectedItemId}
+            url="https://shopify.com"
+            name={name}
+            shortcutActions={[{content: 'action'}]}
+            persistActions
+          />
+        </ResourceListContext.Provider>,
+      );
+
+      const expectedLabel = `Actions for ${name}`;
+
+      expect(
+        item
+          .find(Button)
+          .findWhere(
+            (node) => node.prop('accessibilityLabel') === expectedLabel,
+          ),
+      ).toHaveLength(1);
     });
   });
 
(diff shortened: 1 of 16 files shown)


src/components/ResourceItem/tests/ResourceItem.test.tsx
```diff
@@ -87,6 +89,67 @@ describe('<ResourceItem />', () => {
...
+          <ResourceItem
+            accessibilityLabel={accessibilityLabel}
+            id={itemId}
+            url="https://shopify.com"
+            name={name}
+          />
+        </ResourceListContext.Provider>,
+      );
+
+      const expectedLabel = `Select: ${name}`;
```
chloerice: Use the translation key instead so if it get's updated it automatically stays in sync.

```suggestion
      const expectedLabel = item
        .props()
        .i18n.translate('Polaris.ResourceList.Item.selectItem', {
        accessibilityLabel: name,
      });
```

src/components/ResourceItem/tests/ResourceItem.test.tsx
```diff
@@ -87,6 +89,67 @@ describe('<ResourceItem />', () => {
...
+          <ResourceItem
+            accessibilityLabel={accessibilityLabel}
+            id={selectedItemId}
+            url="https://shopify.com"
+            name={name}
+          />
+        </ResourceListContext.Provider>,
+      );
+
+      const expectedLabel = `Deselect: ${name}`;
```
chloerice:
```suggestion
      const expectedLabel = item
        .props()
        .i18n.translate('Polaris.ResourceList.Item.deselectItem', {
        accessibilityLabel: name,
      });
```

src/components/ResourceItem/tests/ResourceItem.test.tsx
```diff
@@ -87,6 +89,67 @@ describe('<ResourceItem />', () => {
...
+            id={selectedItemId}
+            url="https://shopify.com"
+            name={name}
+            shortcutActions={[{content: 'action'}]}
+            persistActions
+          />
+        </ResourceListContext.Provider>,
+      );
+
+      const expectedLabel = `Actions for ${name}`;
```
chloerice:
```suggestion
      const expectedLabel = item
        .props()
        .i18n.translate('Polaris.ResourceList.Item.actionsDropdownLabel', {
          accessibilityLabel: name,
        });
```

src/components/ResourceItem/tests/ResourceItem.test.tsx
dleroux: This doesn't seem to work. After some discussion with @BPScott, we opted to hardcode the string. Isolates the tests from the translate function.

src/components/ResourceItem/tests/ResourceItem.test.tsx
dleroux: same as above.

src/components/ResourceItem/tests/ResourceItem.test.tsx
chloerice: Ahh, gotcha. I thought we'd set up `mountWithAppProvider` like `web` set up `mountWithAppContext`.

dpersing: Checkboxes report their selected or unselected state by default, so we should avoid including that language in the label. This also helps with translation. Allowing the label to just be `${name}` should provide the support we need. (This is something I've mentioned we should change for the resource list in #792, as well.)

Basically, I'd expect a unique `name` value for each checkbox. If the consumer wants to override the default, they could provide their own in the `accessibilityLabel`. Does that make sense?

dleroux: I've made the change. I left the old `accessibility label` as a fallback, so that there are no unexpected changes for people who had defined it but don't have name yet.

Issue #1761, opened 2019-07-02
Automatic ID generation does not work well with server rendering


lemonmade:
## Issue summary

In the latest edition of "mistakes Chris made several years ago", it's been bugging me that the approach to doing automatic ID generation in Polaris doesn't work properly for server rendering. I think all cases that need an ID generate one using `createUniqueIDFactory`. This creates auto-incrementing IDs (they aren't guaranteed globally unique, but they're fine for our purposes). However, they rely on module-global variables, which means that each request a node server gets will increment those IDs, even though the client will always start them back at 0. This creates a mismatch between client and server markup (not a huge deal, but an annoying warning in dev at the very least).

My proposal for addressing this would be to allow passing some sort of `idFactory` in to Polaris via context. This `idFactory` could default to being module global (e.g., `const IdContext = createContext(idFactory())`), which preserves the 0-config behaviour. However, for our apps, we could provide a fresh `idFactory` for each server render, allowing it to remain consistent with the eventual client markup.

Two things that would be nice to see out of it:

- Make it a shared package so we could use it in other React components
- Make sure it can be reset; our server rendering approach relies on rendering the whole app multiple times, so we need the ability to reset the counter between each pass.

BPScott:
See also #1283

BPScott:
This is still an issue.

BPScott:
I've been low-key working towards a solution for this. Here's a state of the world update

### History:

- Sometimes our usage of createUniqueIDFactory was faulty and it caused incrementing Ids for every render of a component instead of just on initial render
- functional components didn't have a good way of doing unique Ids that remained consistent across multiple renders of that component in the client, never mind keeping consistent between client and server initial renders

Mitigation so far:

- [`useUniqueId()`](https://github.com/Shopify/polaris-react/pull/2079) is a hook created to ensure stable ids across rerenders on the client side. It supports specifying a custom id factory, however that is not exposed from the AppProvider.
- Most components that used `createUniqueIDFactory` have been converted to be functional components that use `useUniqueId`

The final fix for client/server mismatches will be contained within useUniqueId().

Next steps:

- Migrate components that use createUniqueIDFactory to be functional components that use useUniqueId. Ongoing effort for this is tracked in https://github.com/Shopify/polaris-react/issues/1995
- Migrate Banner to be a functional component that uses useUniqueId (for some reason Banner has decided to go its own way and reimplement createUniqueIDFactory

Update our ID handling by either:
[...]

PR #2079, opened 2019-09-05
Add useUniqueId hook and fix several needlessly incrementing ids

BPScott:

We want components to have stable ids between rerenders, otherwise their ids will constantly increment on every render which leads to additional wasteful dom reconciliation.

Currently that's not easily possible with functional components.

Adds a useUniqueId hook that returns an ever incrementing ID when called multiple times but remains constant through rerenders. See the hook tests for examples of how this works in practice.

This hook returns a stable value across multiple rerenders, and can optionally be overridden with a fixed value for components that allow the user to provide their own id.

Use this hook in several functional components that currently have unstable ids that change between rerenders.

The idFactory for this hook is populated in the AppProvider, with eye to eventually allowing consumers to override the idFactory so it can be reset between server renders (making this public and configurable is a follow up step). See #1761 for info where we should go onwards with that. Eventually having this be a standalone library in quilt would be good but I've not got the time to split this out at the moment so I'm doing something just within Polaris for now

### What comes next?

This converts all the currently functional components that use ids to use this hook. Class based components which call their generateId function within a render instead of calling it when initializing a private instance variable also exhibit this behaviour. We could either fix those now, or as part of migrating those components to use hooks - I'm leaning towards the latter.

We will want to make the generator factory we pass into the `UniqueIdFactory` in the AppProvider that gets passed into the  by the consumer so that it can be reset on each server render to complete #1761

### How to 🎩

The following playground contains a bunch of components that trigger rerenderings of a component on a once per second basis thanks to setInterval incrementing a counter. See that only thing that updates within each component is the count content and that any id values remain constant across multiple renderings of the same component.

You can compare this by running this playground on master (but commenting out the `Test` component stuff) and seeing that various id values within the components also update.

<details>

```jsx
import React, {useState, useRef, useEffect} from 'react';
import {
  Page,
  RadioButton,
  ChoiceList,
  FormLayout,
  OptionList,
  Select,
} from '../src';
import {useUniqueId} from '../src/utilities/unique-id';

function noop() {}

  const [count, setCount] = useState(0);

  useInterval(() => {
    setCount((count) => count + 1);
  }, 1000);

  return (
    <Page title="Playground">
      <Test renderCount={count} />
      <hr />
      <Test renderCount={count} />
      <hr />
      <Test renderCount={count} />
      <hr />
      <hr />

      <ChoiceList
        title={count.toString()}
        choices={[
          {label: 'Hidden', value: 'hidden'},
          {label: 'Optional', value: 'optional'},
          {label: 'Required', value: 'required'},
        ]}
        selected={[]}
      />

      <hr />

      <FormLayout.Group helpText="help">Group {count}</FormLayout.Group>

      <hr />

      <hr />

      <OptionList
        title={count.toString()}
        options={[
          {label: 'Hidden', value: 'hidden'},
          {label: 'Optional', value: 'optional'},
          {label: 'Required', value: 'required'},
        ]}
        selected={[]}
        onChange={noop}
      />

      <hr />

      <RadioButton label={count} value="hidden" />

      <hr />

      <Select
        label={count.toString()}
        options={[
          {label: 'Hidden', value: 'hidden'},
          {label: 'Optional', value: 'optional'},
          {label: 'Required', value: 'required'},
        ]}
      />
    </Page>
  );
}

function Test({renderCount}) {
  // const id = useUniqueId();
  const id = useUniqueId();
  const idPrefix = useUniqueId('prefix-');
  const idPrefix2 = useUniqueId('prefix2-');

  return (
    <div>
      {renderCount} :: {id} | {idPrefix} | {idPrefix2}
    </div>
  );
}

function useInterval(callback: () => void, delay: number) {
  const savedCallback = useRef<typeof callback>();

  // Remember the latest callback.
  useEffect(() => {
    savedCallback.current = callback;
  }, [callback]);

  // Set up the interval.
  useEffect(() => {
    function tick() {
      savedCallback.current();
    }
    if (delay !== null) {
      const id = setInterval(tick, delay);
      return () => clearInterval(id);
    }
  }, [delay]);
}
```

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -25,6 +25,6 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Code quality
 
 - Migrated `ActionMenu.RollupAction`, `Autocomplete`, `Card`, `EmptySearchResult`, `Form`, `SkeletonPage` and `TopBar` to use hooks instead of withAppProvider ([#2065](https://github.com/Shopify/polaris-react/pull/2065))
-- Added `useUniqueId` hook that can be used to get a unique id that remains consistent between rerenders and update components to use it where appropriate ([#2079](https://github.com/Shopify/polaris-react/pull/2079))
+- Added `useUniqueId` hook that can be used to get a unique id that remains consistent between rerenders and updated components to use it where appropriate ([#2079](https://github.com/Shopify/polaris-react/pull/2079))
 
 ### Deprecations
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -25,5 +25,6 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Code quality
 
 - Migrated `ActionMenu.RollupAction`, `Autocomplete`, `Card`, `EmptySearchResult`, `Form`, `SkeletonPage` and `TopBar` to use hooks instead of withAppProvider ([#2065](https://github.com/Shopify/polaris-react/pull/2065))
+- Added `useUniqueId` hook that can be used to get a unique id that remains consistent between rerenders and update components to use it where appropriate ([#2079](https://github.com/Shopify/polaris-react/pull/2079))
 
 ### Deprecations
--- a/src/components/RadioButton/RadioButton.tsx
+++ b/src/components/RadioButton/RadioButton.tsx
@@ -42,12 +42,12 @@ export function RadioButton({
   onChange,
   onFocus,
   onBlur,
-  id: providedId,
-  name: providedName,
+  id: idProp,
+  name: nameProp,
   value,
 }: RadioButtonProps) {
-  const id = useUniqueId('RadioButton', providedId);
-  const name = providedName || id;
+  const id = useUniqueId('RadioButton', idProp);
+  const name = nameProp || id;
 
   function handleChange({currentTarget}: React.ChangeEvent<HTMLInputElement>) {
     onChange && onChange(currentTarget.checked, id);
--- a/src/components/RadioButton/RadioButton.tsx
+++ b/src/components/RadioButton/RadioButton.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import {createUniqueIDFactory} from '@shopify/javascript-utilities/other';
+import {useUniqueId} from '../../utilities/unique-id';
 import {Choice, helpTextID} from '../Choice';
 import styles from './RadioButton.scss';
 
@@ -32,8 +32,6 @@ export interface BaseProps {
 
 export interface RadioButtonProps extends BaseProps {}
 
-const getUniqueID = createUniqueIDFactory('RadioButton');
-
 export function RadioButton({
   ariaDescribedBy: ariaDescribedByProp,
   label,
@@ -44,10 +42,13 @@ export function RadioButton({
   onChange,
   onFocus,
   onBlur,
-  id = getUniqueID(),
-  name = id,
+  id: providedId,
+  name: providedName,
   value,
 }: RadioButtonProps) {
+  const id = useUniqueId('RadioButton', providedId);
+  const name = providedName || id;
+
   function handleChange({currentTarget}: React.ChangeEvent<HTMLInputElement>) {
     onChange && onChange(currentTarget.checked, id);
   }
--- a/src/utilities/unique-id/hooks.ts
+++ b/src/utilities/unique-id/hooks.ts
@@ -2,19 +2,19 @@ import {useContext, useRef} from 'react';
 import {UniqueIdFactoryContext} from './context';
 
 /**
- * Returns a unique id that remains consistent across multiple rerenders of the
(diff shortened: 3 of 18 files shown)


src/components/RadioButton/RadioButton.tsx
```diff
@@ -44,10 +42,13 @@ export function RadioButton({
   onChange,
   onFocus,
   onBlur,
-  id = getUniqueID(),
-  name = id,
+  id: providedId,
```
dleroux:
```suggestion
  id: idProp,
```

src/components/RadioButton/RadioButton.tsx
```diff
@@ -44,10 +42,13 @@ export function RadioButton({
...
+  name: providedName,
```
dleroux:
```suggestion
  name: nameProp,
```

src/components/RadioButton/RadioButton.tsx
```diff
@@ -44,10 +42,13 @@ export function RadioButton({
...
   value,
 }: RadioButtonProps) {
+  const id = useUniqueId('RadioButton', providedId);
```
dleroux:
```suggestion
  const id = useUniqueId('RadioButton', idProp);
```

src/components/RadioButton/RadioButton.tsx
```diff
@@ -44,10 +42,13 @@ export function RadioButton({
...
+  const name = providedName || id;
```
dleroux:
```suggestion
  const name = nameProp || id;
```

UNRELEASED.md
```diff
@@ -25,5 +25,6 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Code quality
 
 - Migrated `ActionMenu.RollupAction`, `Autocomplete`, `Card`, `EmptySearchResult`, `Form`, `SkeletonPage` and `TopBar` to use hooks instead of withAppProvider ([#2065](https://github.com/Shopify/polaris-react/pull/2065))
+- Added `useUniqueId` hook that can be used to get a unique id that remains consistent between rerenders and update components to use it where appropriate ([#2079](https://github.com/Shopify/polaris-react/pull/2079))
```
```suggestion
- Added `useUniqueId` hook that can be used to get a unique id that remains consistent between rerenders and updated components to use it where appropriate ([#2079](https://github.com/Shopify/polaris-react/pull/2079))
```

src/utilities/unique-id/tests/hooks.test.tsx
```diff
@@ -0,0 +1,182 @@
...
+    const harness = mountWithApp(<ReRenderingTestHarness />);
+
+    expect(harness.find(HasProp)!.html()).toStrictEqual('count1 :: polaris-1');
+
+    harness.find('button')!.trigger('onClick');
+
+    expect(harness.find(HasProp)!.html()).toStrictEqual('count2 :: polaris-1');
+  });
+
+  it('updates the ID if the overridden ID changes', () => {
```
tmlayton: You should be able to simplify these tests by using `update` instead of providing your own manual button click. https://airbnb.io/enzyme/docs/api/ShallowWrapper/update.html

src/utilities/unique-id/hooks.ts
```diff
@@ -0,0 +1,38 @@
...
+  const uniqueIdRef = useRef<string | null>(null);
+
+  if (!idFactory) {
+    throw new Error(
+      'No UniqueIdFactory was provided. Your application must be wrapped in an <AppProvider> component. See https://polaris.shopify.com/components/structure/app-provider for implementation instructions.',
+    );
+  }
+
+  // If an override was specified, then use that instead of using a unique ID
+  // Hooks can't be called conditionally so this has to go after all use* calls
```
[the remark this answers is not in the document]
BPScott: Yep, see https://reactjs.org/docs/hooks-rules.html

src/utilities/unique-id/hooks.ts
```diff
@@ -0,0 +1,38 @@
...
+ * @param overrideId Defines a fixed value to use instead of generating a unique
+ *   ID. Useful for components that allow consumers to specify a fixed ID.
+ */
+export function useUniqueId(prefix = '', overrideId = '') {
+  const idFactory = useContext(UniqueIdFactoryContext);
+
+  // By using a ref to store the uniqueId for each incovation of the hook and
+  // checking that it is not already populated we ensure that we don't generate
+  // a new ID on ever rerender of a component.
+  const uniqueIdRef = useRef<string | null>(null);
```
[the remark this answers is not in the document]
BPScott: That won't isolate the id value to be within a single component. That'd mean multiple components would end up picking up that same value. This would also break the ability to have unique counters per prefix.

src/utilities/unique-id/tests/hooks.test.tsx
BPScott: These tests use [`@shopify/react-testing`](https://github.com/Shopify/quilt/blob/master/packages/react-testing/README.md) which doesn't offer that. I tried using their setProps utility but I couldn't get the test to fail when deliberately making the error I wanted to guard against

PR #2080, opened 2019-09-05
[Frame] Skip to content target / ref

philmongeon:

I discovered a problem working on a project for experts.shopify.com. 

We were planning on using a `<Toast />` component for some user feedback but the `<Toast />` component relies on the `<Frame />` component to function. 

The `<Frame />` component adds a skip to content link for a11y purposes that is not configurable and does not work the way we need it to work. 

We already have a skip to content link of our own that places the user past our main navigation to the top of the content area. When we add the frame component to the app we end up with 2 skip to content links (ours that does the thing we want and the one created by the frame that does not function correctly).

Making the skip to content link configurable will allow us to use the Polaris components that rely on the frame component while still providing an accessible skip to content link.

Adds a new prop to the `<Frame />` component.

```typescript
/** The RefObject you wish to be focused when clicking the skip to content link */
skipToContentTarget?: React.RefObject<HTMLAnchorElement>;
```

We add a `RefObject` so that on click we can focus the provided `RefObject` and focus the user in the correct place.

<details>
   <summary>[GIF] Skipping to the target element</summary>
   Description of what the gif shows
</details>

### How to 🎩

<details>

```tsx
import React from 'react';
import {Page, Frame} from '../src';

  const targetId = 'SkipToContentTarget';
  const targetRef = React.createRef<HTMLAnchorElement>();

  const skipToMainContentTarget = (
    // eslint-disable-next-line jsx-a11y/anchor-is-valid
    <a id={targetId} ref={targetRef} tabIndex={-1}>
      Focus target
    </a>
  );

  return (
    <Page title="Playground">
      <Frame skipToContentTargetId={targetId} skipToContentTarget={targetRef}>
        {skipToMainContentTarget}
      </Frame>
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Frame/Frame.tsx
+++ b/src/components/Frame/Frame.tsx
@@ -43,6 +43,8 @@ export interface FrameProps {
    * @default false
    */
   showMobileNavigation?: boolean;
+  /** Accepts a ref to the html anchor element you wish to focus when clicking the skip to content link */
+  skipToContentTarget?: React.RefObject<HTMLAnchorElement>;
   /** A callback function to handle clicking the mobile navigation dismiss button */
   onNavigationDismiss?(): void;
 }
@@ -81,7 +83,8 @@ class Frame extends React.PureComponent<CombinedProps, State> {
   private contextualSaveBar: ContextualSaveBarProps | null;
   private globalRibbonContainer: HTMLDivElement | null = null;
   private navigationNode = createRef<HTMLDivElement>();
-  private skipToMainContentTargetNode = React.createRef<HTMLAnchorElement>();
+  private skipToMainContentTargetNode =
+    this.props.skipToContentTarget || React.createRef<HTMLAnchorElement>();
 
   componentDidMount() {
     this.handleResize();
@@ -112,6 +115,7 @@ class Frame extends React.PureComponent<CombinedProps, State> {
       globalRibbon,
       showMobileNavigation = false,
       polaris: {intl},
+      skipToContentTarget,
     } = this.props;
 
     const navClassName = classNames(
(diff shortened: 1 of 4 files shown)


src/components/Frame/Frame.tsx
```diff
@@ -43,6 +43,10 @@ export interface FrameProps {
    * @default false
    */
   showMobileNavigation?: boolean;
+  /** ID of the element you wish to skip to when clicking skip to content link */
+  skipToContentTargetId?: string;
```
philmongeon: optionally this can just be removed and replaced with `this.skipToMainContentTargetNode.current.id` and get the id out of the `RefObject`. My main concern what the id might be `null` if someone doesn't have it so I went with the explicit route. However this 2 props might make it more confusing to use

src/components/Frame/Frame.tsx
```diff
@@ -203,10 +210,14 @@ class Frame extends React.PureComponent<CombinedProps, State> {
       skipFocused && styles.focused,
     );
 
+    const skipTarget = skipToContentTargetId
+      ? skipToContentTargetId
+      : APP_FRAME_MAIN_ANCHOR_TARGET;
```
philmongeon: if we drop the `skipToContentTargetId` prop this could become

```typescript
const skipTarget = skipToContentTarget
      ? skipToContentTarget.current.id
      : APP_FRAME_MAIN_ANCHOR_TARGET;
```

src/components/Frame/Frame.tsx
philmongeon: you know what i think i'm going to just switch to this approach i think the simpler api is better

dleroux:
🎩 works great! I like that this was brought down to 1 prop. Changelog needs an update but apart from that ✅ . Ping me when the that is ready.  Thank you for doing this!

philmongeon:
This works but not in the way I would expect. A bit or weirdness is happening with passing the ref and using it as a ternary to render things. I’ll take a look at fixing it when I have some more time and I’ll ping you then. Thanks !

philmongeon:
@dleroux updated and added changes to readme and 🎩 the styleguide

alex-page:
I tophatted this in chroma -> frame in mac os, firefox nightly.

src/components/Frame/Frame.tsx
```diff
@@ -203,10 +207,14 @@ class Frame extends React.PureComponent<CombinedProps, State> {
       skipFocused && styles.focused,
     );
 
+    const skipTarget = skipToContentTarget
+      ? (skipToContentTarget.current && skipToContentTarget.current.id) || ''
```
dleroux: Wondering if this should default to `APP_FRAME_MAIN_ANCHOR_TARGET`?

src/components/Frame/Frame.tsx
dleroux: Yes just saw that.

Issue #1959, opened 2019-08-13
Proposal: Name components and Props once


BPScott:
Default exports are low-key annoying as their name is "default" so every time you import them you have to give them a name. Instead of naming something once and using it consistently everywhere (and having TS shout at us if we misname it) we have to make sure we use the same name manually in every file (JS doesn't care but its would be confusing for us to have two names for the same concept).

I would like to propose that we stop using default exports and instead use named exports for everything. Additionally we should name `Props` based upon the component to avoid needing to rename props when used in other components to avoid naming clashes. This would save on export renaming, and would lead to simpler usage in sister components and reexporting in component indexes.

We can help enforce this by enabling the `import/no-default-export` eslint rule.

Further reading from someone else who came to the same conclusion: https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/

### The current world

Currently in our components we export the component as the default and its props as `Props` in both `src/components/ComponentName/index.js` and `src/components/ComponentName/ComponentName.js`. We then reexport with a different name in the components index. Additionally in the cases where components are used by other components we have to rename the imports as they would conflict with existing 

```ts
// src/components/Button/Button.tsx
export interface Props {}
export default function Button() {}


// src/components/Button/index.ts
// In reality this is usually done a bit differently as an import then export so it is a little more convoluted
export {default, Props} from './Button';


// Components are reexported in the component index, where we need to name the default and rename the Props 
// src/components/index.ts
export {default as Button, Props as ButtonProps} from './Button';


// Components and their props are often imported in other components, where we need to name the default and rename the Props
// src/components/ResourceList/ResourceList.tsx
import Button, {Props as ButtonProps} from '../Button';
```

### The new world

Use named exports and prefix the Props with the name of the component to avoid clashes (e.g. The Props in the `Button` component becomes `ButtonProps`).

This way there is no need to rename exports when using them

```ts
// src/components/Button/Button.tsx
export interface ButtonProps {}
export function Button() {}


// src/components/Button/index.ts
// A simple reexport of named exports, no renaming needed
export {Button, ButtonProps} from './Button';


// Components are rexported in the component index
// A simple reexport of named exports, no renaming needed
// src/components/index.ts
export {Button, ButtonProps} from './Button';


// Components and their props are often imported in other components,
// A simple import of named exports, no renaming needed
// src/components/ResourceList/ResourceList.tsx
import {Button, ButtonProps} from '../Button';
```

### Sounds neat, how would we do this?

[...]

PR #2083, opened 2019-09-06
Hookify more components

BPScott:
As with other hookify PRs this is probably best viewed with supresed whitespace.

More work towards #1995.

Hookifies RangeSlider, ScrollLock and Topbar's SearchField

### How to 🎩

- Confirm Single / Dual rangeslider examples work
- Confirm Topbar SearchField works, by checking the Frame example (The TopBar example seems to be buggy losing content on blur on master)
- Confirm ScrollLock works by checking modal overlays stop you scrolling the page within web. ScrollLock doesn't work within storybook even on master for whatever reason.


src/components/TopBar/components/SearchField/tests/SearchField.test.tsx
```diff
@@ -2,10 +2,7 @@ import React from 'react';
 import {CircleCancelMinor} from '@shopify/polaris-icons';
 import {ReactWrapper} from 'enzyme';
 import {mountWithAppProvider} from 'test-utilities/legacy';
-// We've got to export a named SearchField otherwise other stuff breaks
```
AndrewMusgrave: Just curious what this was breaking?

src/components/ScrollLock/ScrollLock.ts
```diff
@@ -0,0 +1,21 @@
+import {useEffect} from 'react';
+import {useScrollLockManager} from '../../utilities/scroll-lock-manager';
+import './ScrollLock.scss';
+
+interface ScrollLockProps {}
+
+// Even though this has no args, reference ScrollLockProps so the prop explorer
```
AndrewMusgrave: Will this show up in the style guide without a `Props` interface?

src/components/ScrollLock/ScrollLock.ts
```diff
@@ -0,0 +1,21 @@
+import {useEffect} from 'react';
+import {useScrollLockManager} from '../../utilities/scroll-lock-manager';
+import './ScrollLock.scss';
+
+interface ScrollLockProps {}
```
AndrewMusgrave:
```suggestion
interface Props {}
```

src/components/ScrollLock/ScrollLock.ts
```diff
@@ -0,0 +1,21 @@
...
+
+// Even though this has no args, reference ScrollLockProps so the prop explorer
+// in the styleguide works without warnings about unfound props
+export function ScrollLock(_: ScrollLockProps) {
```
AndrewMusgrave:
```suggestion
export function ScrollLock(_: Props) {
```

src/components/ScrollLock/ScrollLock.ts
AndrewMusgrave: On second thought, looks like we don't have a readme and this is just an internal component.

BPScott: Per my proposal in #1959 we now name Props after the component - see ScrollLockProps above this.

The styleguide was updated to handle this in https://github.com/Shopify/polaris-styleguide/pull/2932

src/components/TopBar/components/SearchField/tests/SearchField.test.tsx
BPScott: It complained about being able to generate the types for TopBar.SearchField - or rather it was ok generating them, but trying to use it in consuming projects caused complaints. exporting SearchField in addition to the default component seemed to make it happy.

I don't understand the reasoning for it, but I seemed like it was a due to a combo of a trying to use a withAppProvider wrapped component as an exposed subcomponent

BPScott: It complained about being able to generate the types for TopBar.SearchField - or rather it was ok generating them, but trying to use it in consuming projects caused complaints. exporting SearchField in addition to the default component seemed to make it happy.

I don't understand the reasoning for it, but I seemed like it was a due to a combo of a trying to use a withAppProvider wrapped component as an exposed subcomponent

src/components/TopBar/components/SearchField/SearchField.tsx
```diff
@@ -33,148 +28,100 @@ export interface SearchFieldProps {
...
-  private handleChange = ({
-    currentTarget,
-  }: React.ChangeEvent<HTMLInputElement>) => {
-    const {onChange} = this.props;
-    onChange(currentTarget.value);
-  };
+  }, [onCancel, onChange]);
+
+  useEffect(() => {
+    if (input.current) {
```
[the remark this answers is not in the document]
BPScott: good call

PR #2089, opened 2019-09-06
[DatePicker] Hookify

amrocha:

We're updating our components to use hooks and functions instead of classes and HoCs like before, this is part of that work

https://github.com/Shopify/polaris-react/issues/1995

Hookifies Day and DatePicker

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/DatePicker/DatePicker.tsx
+++ b/src/components/DatePicker/DatePicker.tsx
@@ -66,12 +66,9 @@ export function DatePicker({
   const [hoverDate, setHoverDate] = useState<Date | undefined>(undefined);
   const [focusDate, setFocusDate] = useState<Date | undefined>(undefined);
 
-  useEffect(
-    () => {
-      setFocusDate(undefined);
-    },
-    [selected],
-  );
+  useEffect(() => {
+    setFocusDate(undefined);
+  }, [selected]);
 
   const handleFocus = useCallback((date: Date) => {
     setFocusDate(date);
--- a/src/components/DatePicker/DatePicker.tsx
+++ b/src/components/DatePicker/DatePicker.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useState, useEffect, useCallback} from 'react';
 import {ArrowLeftMinor, ArrowRightMinor} from '@shopify/polaris-icons';
 import {
   Range,
@@ -11,14 +11,10 @@ import {
   getPreviousDisplayYear,
   getPreviousDisplayMonth,
   Weekdays,
-  isSameDay,
 } from '@shopify/javascript-utilities/dates';
 
-import {
-  withAppProvider,
-  WithAppProviderProps,
-} from '../../utilities/with-app-provider';
 import {Button} from '../Button';
+import {useI18n} from '../../utilities/i18n';
 import {monthName} from './utilities';
 
 import {Month} from './components';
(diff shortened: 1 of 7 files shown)


src/components/DatePicker/DatePicker.tsx
```diff
@@ -299,32 +277,6 @@ function handleKeyDown(event: React.KeyboardEvent<HTMLElement>) {
   }
 }
 
-function isSameSelectedDate(
```
AndrewMusgrave: Are we sure we can remove this logic?

src/components/DatePicker/DatePicker.tsx
```diff
@@ -52,235 +48,217 @@ export interface BaseProps {
...
-      </div>
-    );
-  }
-
-  private handleFocus = (date: Date) => {
-    this.setState({focusDate: date});
-  };
+      setHoverDate(end);
+      setFocusDate(new Date(end));
+      onChange(range);
```
AndrewMusgrave: It looks like `onChange` was previously called when the setState for hover/focus had resolved. Any idea why, and do you think it'll change anything?

src/components/DatePicker/DatePicker.tsx
amrocha: Yeah, I'm not even sure why we're doing this.

This is just doing an actual value comparison, instead of just trusting that if the reference is different then something has changed. I think whoever wrote this was forced to do it like this to avoid re-rendering unnecessarily, but now I can just replace this with the `useEffect` hook in line 69 that only fires if the prop actually changed.

The only case I can think that this doesn't account for is someone selecting the same date twice in a row, and triggering a re-render of the component. But I tried that and it seemed to work fine, so

src/components/DatePicker/DatePicker.tsx
amrocha: Not sure, could just be trying to prevent race conditions. I expect this to be basically equivalent.

Do you think it'd be better to call onChange in a different `useEffect` function?

src/components/DatePicker/DatePicker.tsx
AndrewMusgrave: Looking back the code seems to fix [this issue](https://github.com/Shopify/polaris-react-deprecated/pull/1399) and a quick 🎩 shows it's still 💯 ✅ 

> useEffect hook in line 69 that only fires if the prop actually changed

Funny enough we could have done this in a class based component as well.

```tsx
  useEffect(() => {
    setFocusDate(undefined);
  }, [selected]);
```
Is essentially the same as

```tsx
componentDidUpdate(prevProps) {
 if (!Object.is(prevProps.selected, this.props.selected) 
    this.setState({focusedDate: undefined})
}
```

src/components/DatePicker/DatePicker.tsx
AndrewMusgrave: I think we'll be ok, I was just wondering if anything stood out while you were working on this.

src/components/DatePicker/components/Day/Day.tsx
```diff
@@ -16,84 +13,72 @@ export interface DayProps {
   inHoveringRange?: boolean;
   disabled?: boolean;
   onClick?(day: Date): void;
-  onHover?(day: Date): void;
-  onFocus?(day: Date): void;
+  onHover?(day?: Date): void;
```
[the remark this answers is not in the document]
amrocha: Day is not a public member of the Polaris API, so the only consumers are internal, and this doesn't throw any type errors so that's not a concern.

I think we're safe to make any changes we want to private parts of our API. Do you disagree?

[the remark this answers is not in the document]
BPScott: Good point on this not actually being public, so the breaking change thing is moot.

It looks like we can keep `onFocus?(day: Date)` without any extra changes needed so that'd be easy to do.

I was thinking we could remove the onHover from the empty day but it looks like that has knock of effects for how selecting a range (with `<DatePicker allowRange>`) works so it looks like we need to keep that.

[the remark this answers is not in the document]
amrocha: I'm also wary of making too many changes in this PR, I'm already making a few changes to the flow here and there and I'm not super familiar with the Datepicker so I'm trying to keep everything roughly the same

PR #2090, opened 2019-09-06
[Change log] v4.2.0

chloerice:

Change log for v4.1.1


CHANGELOG.md
```diff
@@ -8,6 +8,24 @@ The format is based on [these versioning and changelog guidelines](https://git.i
 
 ---
 
+## 4.2.0 - 2019-09-06
+
+### Bug fixes
+
+- Fixed animation for Modal when being rendered asynchronously ([#2076](https://github.com/Shopify/polaris-react/pull/2076))
+- Updated `TextField` `min` and `max` type from `number` to `number | string` to allow min/max dates ([#1991](https://github.com/Shopify/polaris-react/pull/1991))
```
kaelig: Should this be under the "Enhancements" section, to make it clearer why this is a minor release and not a patch?

chloerice: This does still seem like a patch to me too. 

The dependency upgrade is just `prettier`, which technically isn't even a dependency (it's not in our `package.json`, it's just in our `yarn.lock`... I guess because one of our other dependencies uses it?) so that doesn't trigger a minor release. 

But since we're adding a type accepted by a prop, according to our versioning guidelines it should be an enhancement and go out as a minor release. I'll move it to the `Enhancements` section.

dleroux:
Once the conflict is fixed ✅

CHANGELOG.md
```diff
@@ -8,6 +8,42 @@ The format is based on [these versioning and changelog guidelines](https://git.i
...
+### Enhancements
+
+- Added support for min/max dates in `TextField` by setting a string on `min` and `max` props ([#1991](https://github.com/Shopify/polaris-react/pull/1991))
+- Made the `title` prop on `Page` optional, supporting continued use of `Page` for structure in apps using the App Bridge React [`TitleBar`](https://github.com/Shopify/app-bridge/tree/master/packages/app-bridge-react/src/components/TitleBar) ([#2082](https://github.com/Shopify/polaris-react/pull/2082))
+
+### Bug fixes
+
+- Fixed inconsistent padding of sections in `Modal` ([#2072](https://github.com/Shopify/polaris-react/pull/2072))
+- Fixed animation for Modal when being rendered asynchronously ([#2076](https://github.com/Shopify/polaris-react/pull/2076))
+- Updated `TextField` `min` and `max` type from `number` to `number | string` to allow min/max dates ([#1991](https://github.com/Shopify/polaris-react/pull/1991))
```
kaelig:
```suggestion
```

Duplicate of line 15

chloerice: 🙏

PR #2092, opened 2019-09-06
Use Storybook's Component Story Format

BPScott:

Storybook CSF format is neat and makes our component generation simpler

Using the CSF format for our readme exports instead of creating our own object shape then passing that through their storiesOf API

### How to 🎩

Check all stories render as before


BPScott:
This gets us using the CSF format.

I had a little play with the docs addon but that'll require a bit of thought to work out the right for how we want to display the readme content, so that can be a job for later

PR #2094, opened 2019-09-06
[ResourceItem] Avoid using withAppProvider

BPScott:

Hookifying the whole class is going to be hard, but hookifying the
wrapper component is pretty easy

This should save 2 levels of nesting - the WithAppProvider, and the context consumer

Remove withAppProvider and a context consumer from ResourceItem

### How to 🎩

Check ResourceItem still renders correctly within a ResourceList

### DIFF
--- a/src/components/ResourceItem/index.ts
+++ b/src/components/ResourceItem/index.ts
@@ -1,3 +1,1 @@
-import ResourceItem, {ResourceItemProps} from './ResourceItem';
-
-export {ResourceItem, ResourceItemProps};
+export {ResourceItem, ResourceItemProps} from './ResourceItem';
--- a/src/components/ResourceItem/tests/ResourceItem.test.tsx
+++ b/src/components/ResourceItem/tests/ResourceItem.test.tsx
@@ -13,7 +13,7 @@ import {
   Button,
 } from 'components';
 import {ResourceListContext} from '../../../utilities/resource-list';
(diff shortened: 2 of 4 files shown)


src/components/ResourceItem/index.ts
```diff
@@ -1,3 +1 @@
-import ResourceItem, {ResourceItemProps} from './ResourceItem';
-
-export {ResourceItem, ResourceItemProps};
+export {ResourceItem, ResourceItemProps} from './ResourceItem';
```
perrupa: 👌

src/components/ResourceItem/tests/ResourceItem.test.tsx
```diff
@@ -13,7 +13,7 @@ import {
   Button,
 } from 'components';
 import {ResourceListContext} from '../../../utilities/resource-list';
-import ResourceItem from '../ResourceItem';
+import {ResourceItem} from '../ResourceItem';
```
perrupa: 🤞 this is the only case of importing it... is there a chance external users would have been depending on the default export here?

BPScott: If they are then they're doing something unstable and unsupported :) Importing from `@shopify/polaris` is the public interface, all other JS is considered private and we're free to change it whenever.

They should be doing `import {ResourceItem} from '@shopify/polaris'`; which is the public export of this component.

Issue #1760, opened 2019-07-02
[Pagination] Add label prop


Tetsuro:
## Feature request summary

Add a `label` prop to the Pagination component. 

## Rationale

Allows for a label to provide more information regarding pagination. For example, we can use it to show `X of X pages`. 

![image](https://user-images.githubusercontent.com/351842/60538856-6a7e2900-9cd9-11e9-9ee6-c887081424c9.png)

## Implementation details

- prop name: `label?` 
- prop type: string

The `aria-live` attribute can be used to read out the label to screen reader users.

![](https://user-images.githubusercontent.com/351842/60367395-195df480-99bc-11e9-8f3c-344d3e90b6aa.gif)

---

🌟 Feature requests that are not yet planned will be closed. We then use the issue’s [:+1: upvotes](https://help.github.com/articles/about-conversations-on-github/) to track and set priorities. See the [contribution guidelines](https://github.com/Shopify/polaris-react/blob/master/.github/CONTRIBUTING.md) for more information.

beefchimi:
For reference, this is implemented in `online-store-ui`:
https://github.com/Shopify/online-store-ui/pull/230

The PR should be relatively easy to follow, as Tetsuro cloned the component in the first commit, then made adjustments in all subsequent commits.

PR #2098, opened 2019-09-09
[Pagination] Added label prop

mitchmaps:

Addresses feature request #1760 following guideline set [here](https://github.com/Shopify/online-store-ui/pull/230).

**TL;DR:** Adds a prop to insert text in between the backwards and forwards pagination buttons.

This PR addresses an issue opened by @Tetsuro who had created a version of the change for [Shopify/online-store-ui](https://github.com/Shopify/online-store-ui) and wanted to see the feature he built introduced into the standard Polaris react library.

The new label prop aims to provide users more context for what the pagination component is representing by adding descriptive text in between the backwards and forwards button. The example @Tetsuro used in the original issue was text indicating how many pages there were to cycle between but the prop could be used for anything a user wants.

#### Example of a Pagination component using the new label prop

_Desktop_
image

_Mobile_
image

## 

### How to 🎩

1. Checkout the branch
2. Run the server
3. Add a Pagination component to the `Playground.tsx` file and pass in a `label` prop
4. Checkout playground on the server and check out the visual changes to the component

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
      <Pagination label="1 of 10" />
    </Page>
  );
}

```

</details>

### 🎩 checklist


src/components/Pagination/Pagination.scss
```diff
@@ -8,6 +8,7 @@ $difference-between-touch-area-and-backdrop: control-height() -
 // This is the adjustment needed to get the icon to appear flush with other
 // content around it.
 $plain-horizontal-adjustment: spacing(tight);
+$labelMinWidth: rem(80px);
```
LauraAubin: Why did you choose `80px` for this? Is this based on the expected min width of text that follows the format `number of number`?

alex-page: I agree with @LauraAubin and would prefer this to be the width of the content with padding.

mitchmaps: That's a much better idea I'll make that change!

src/components/Pagination/Pagination.tsx
```diff
@@ -55,16 +58,17 @@ export function Pagination({
...
-  let label: string;
+  let navLabel: string;
 
   if (accessibilityLabel) {
-    label = accessibilityLabel;
+    navLabel = accessibilityLabel;
   } else {
-    label = intl.translate('Polaris.Pagination.pagination');
+    navLabel = intl.translate('Polaris.Pagination.pagination');
   }
```
dleroux: I know you didn't add this but this might be nicer:

`const navLabel =  accessibilityLabel || intl.translate('Polaris.Pagination.pagination');`

src/components/Pagination/Pagination.tsx
```diff
@@ -163,10 +167,17 @@ export function Pagination({
       />
     ));
 
+  const labelMarkup = label ? (
+    <div className={styles.Label} aria-live="polite">
+      <TextStyle variation="subdued">{label}</TextStyle>
```
dleroux: Should we have it always subdued, of maybe just subdue if `hasNext` and `hasPrevious` are both false 🤷‍♂ 

cc: @sarahill

src/components/Pagination/Pagination.tsx
dleroux: code cov would pass if this line was tested.

dleroux:
@sarahill ui-kit updates?

perrupa:
https://user-images.githubusercontent.com/42747959/64539029-b6f27000-d2eb-11e9-9d7e-712952a3d083.png

Looks like the next and previous button corners aren't rounded as I'd expect when there's a label (when they don't have adjacent buttons). Is that intentional?

src/components/Pagination/Pagination.tsx
mitchmaps: I think that is a good idea, makes sense for the indicator to also be subdued if the actions surrounding it are as well and vice versa 👍

src/components/Pagination/Pagination.tsx
mitchmaps: Good catch! I've made the changes 💯

src/components/Pagination/Pagination.tsx
mitchmaps: ## Label with previous and next
image

## Label without previous and next
image

mitchmaps:
## FIXED: Buttons with a label having rounded corners
image


## Previous style without a label preserved:
image

src/components/Pagination/tests/Pagination.test.tsx
```diff
@@ -78,6 +78,33 @@ describe('<Pagination />', () => {
...
+    });
+  });
+
+  describe('label', () => {
+    it('renders as text', () => {
+      const label = 'test';
+      const pagination = mountWithAppProvider(<Pagination label={label} />);
+      expect(pagination.text()).toContain(label);
+    });
+  });
```
perrupa: Had you considered adding a test to verify that the label is `subdued` when there are no next or previous pages 🤔

src/components/Pagination/Pagination.tsx
```diff
@@ -35,6 +36,8 @@ export interface PaginationDescriptor {
   onNext?(): void;
   /** Callback when previous button is clicked */
   onPrevious?(): void;
+  /** Text to provide more context in between the arrow buttons */
+  label?: string;
```
perrupa: Is this the best prop name for this, it's not a `label` in the form sense of the word, maybe `paginationText` or `currentPageText`?

These aren't the best suggestions, and I'm ok with `label` lacking a better name, I just wonder if this is the _best_ name 😅

perrupa:
Also, might want to add it to the changelog 👍

src/components/Pagination/tests/Pagination.test.tsx
mitchmaps: Done! 👍

src/components/Pagination/Pagination.tsx
```diff
@@ -55,21 +58,26 @@ export function Pagination({
...
+  const previousClassName = classNames(
+    styles.Button,
+    styles.PreviousButton,
+    label && styles.ButtonWithLabel,
+  );
+  const nextClassName = classNames(
+    styles.Button,
+    styles.NextButton,
+    label && styles.ButtonWithLabel,
+  );
```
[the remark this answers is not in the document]
mitchmaps: On it! 👍

src/components/Pagination/Pagination.tsx
mitchmaps: I agree, label could be confusing. How about something like `indicatorText`?

alex-page: I would agree with `paginationText`, `indicatorText` is too vague.

mitchmaps: ^ We chatted about this on slack and decided to stick with label

src/components/Pagination/Pagination.tsx
```diff
@@ -55,21 +58,24 @@ export function Pagination({
...
-  const previousClassName = classNames(styles.Button, styles.PreviousButton);
-  const nextClassName = classNames(styles.Button, styles.NextButton);
+
+  const previousClassName = !label
+    ? classNames(styles.Button, styles.PreviousButton)
+    : classNames(styles.Button);
+
+  const nextClassName = !label
+    ? classNames(styles.Button, styles.NextButton)
+    : classNames(styles.Button);
```
perrupa: This duplication would make adding new classes to buttons awkward. I think it might be simpler to put the condition on the class we're targetting with our condition instead.

```js
const nextClassName = classNames(
  styles.Button,
  !label && styles.NextButton,
);
```

src/components/Pagination/tests/Pagination.test.tsx
```diff
@@ -78,6 +78,45 @@ describe('<Pagination />', () => {
...
+    it('has subdued text without next and previous pages', () => {
+      const label = 'test';
+      const variation = 'subdued';
+      const pagination = mountWithAppProvider(<Pagination label={label} />);
+      expect(
+        pagination
+          .find('.Label')
+          .children()
+          .prop('variation'),
+      ).toStrictEqual(variation);
```
perrupa: Do you need `.find('.Label').children()` over `.find('.TextStyle')`, or was that not working?


Also, is there a reason you put these values in variables? I think the test might read a bit clearer with less indirection, what do you think?

```js
const pagination = mountWithAppProvider(<Pagination label="label" />);
expect(
  pagination
    .find('.TextStyle')
    .prop('variation'),
).toStrictEqual('subdued');
```

src/components/Pagination/Pagination.tsx
mitchmaps: Good catch, this is a better way of doing it 💯

src/components/Pagination/tests/Pagination.test.tsx
mitchmaps: I tried doing it with .Textstyle and it was running into some issues finding the specific child to compare to. Using my own class from the label component seemed to do the trick but if there's a better way to go about it then I think it's something worth changing.

And I agree that might be a little overkill for the variables if they're only used once, a line after they were defined. I think I just put those in to follow the pattern of creating a variable for the component itself to assert.

mitchmaps: Changes have been made 🔥

PR #2104, opened 2019-09-09
[Card] Changed footer actions to be left aligned

### LINKED ISSUE #2075: Card with footer actions tweak
## Issue summary
Update `Card with footer actions` component to display footer actions [button(s)] to be left aligned by default as opposed to right aligned by default. Left aligned is the most common use and intended design of actions on cards.

## Expected behaviour
When using `Card with footer actions` component the buttons on the card are left aligned by default.

## Actual behaviour
When using `Card with footer actions` currently, the buttons are right aligned by default and you only get left aligned buttons when using the `Card with custom footer actions`

This seems odd as most (if not all) cases in the Admin, buttons show up left aligned (Home, Settings, etc). Are all these instances using the `Card with custom footer actions` then to get around this funky default of right aligned buttons? Something worth investigating when fixing this issue.

Also, with this fix, should we keep the `Card with custom footer actions` component and reserve that for right aligned actions?

cc @alex-page  @HYPD @sarahill

mitchmaps:

Closes #2075, @ouellettejordan mentioned that the footer actions for the card component make more sense to be left aligned opposed to right. This was because other versions of Card have the actions on the left so this change is made mainly for consistency.

This PR changes the default style for the primary and secondary footer action props in Card so that they are aligned on the left of the card rather than the right.

Before:
image

After:
image


dleroux:
ping @sarahill ui-kit update needed

UNRELEASED.md
```diff
@@ -12,6 +12,17 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
```
dleroux: Something bizarre going on here.

mitchmaps: Yea I'm not really sure how this happened 😂all of the previous changes were in my unrealeased.md file so when I added what I had changed it just pushed all of them.

Should I just delete all of the other ones in there besides my own since they're already merged into master?

PR #2105, opened 2019-09-09
🔥[AppProvider] Remove unnecessary state

beefchimi:
Discovered some old code while pairing with @AndrewMusgrave and we figured it best to remove.

### Tophat Instructions

This pull request is removing stale code and converting `AppProvider` to a functional component. You'll want to test that the app context stored on the app provider still works as intended. These contexts include; sticky manager (resource list header), scroll lock manager (scrollable), app bridge (some frame components/page/modal), and i18n (the majority of our components)

### DIFF
--- a/src/components/AppProvider/AppProvider.tsx
+++ b/src/components/AppProvider/AppProvider.tsx
@@ -65,7 +65,11 @@ export function AppProvider({
 
   useEffect(() => {
     stickyManager.current.setContainer(document);
-  }, [stickyManager]);
+    // We're using a disable because stickyManager is a ref
+    // Refs are used for values which don't trigger a re-render
+    // and intended to be mutated outside the render data flow
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
 
   return (
     <FeaturesContext.Provider value={features}>
--- a/src/components/AppProvider/AppProvider.tsx
+++ b/src/components/AppProvider/AppProvider.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, {useMemo, useEffect} from 'react';
 import {ThemeConfig} from '../../utilities/theme';
 import {TelemetryContext, TelemetryObject} from '../../utilities/telemetry';
 import {ThemeProvider} from '../ThemeProvider';
+import {useLazyRef} from '../../utilities/use-lazy-ref';
 import {MediaQueryProvider} from '../MediaQueryProvider';
 import {I18n, I18nContext, TranslationDictionary} from '../../utilities/i18n';
 import {
@@ -25,12 +26,6 @@ import {
   globalIdGeneratorFactory,
 } from '../../utilities/unique-id';
 
-interface State {
-  intl: I18n;
-  appBridge: ReturnType<typeof createAppBridge>;
-  link: LinkLikeComponent | undefined;
-}
-
 export interface AppProviderProps extends AppBridgeOptions {
   /** A locale object or array of locale objects that overrides default translations */
   i18n: TranslationDictionary | TranslationDictionary[];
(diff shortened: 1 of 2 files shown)


BPScott:
Neat! It looks like it wouldn't be too hard to rework AppProvider to be a functional component, would you be interested in taking that on?

src/components/AppProvider/AppProvider.tsx
```diff
@@ -85,23 +77,21 @@ export class AppProvider extends React.Component<AppProviderProps, State> {
...
 
   render() {
-    const {theme = {logo: null}, children} = this.props;
-    const {intl, appBridge, link} = this.state;
+    const {i18n, linkComponent, theme = {logo: null}, children} = this.props;
+    const {appBridge} = this.state;
 
     return (
-      <I18nContext.Provider value={intl}>
+      <I18nContext.Provider value={new I18n(i18n)}>
```
sijad: is it a good idea to recreate `I18n` on every render? I think it's better to memorize this value

beefchimi:
@BPScott sadly thats not something I can directly take on, but @AndrewMusgrave and I are planning on doing some pairing to help me build a AppProvider for my own project, so perhaps its something him and I can chip away on in tandem.

src/components/AppProvider/AppProvider.tsx
AndrewMusgrave: `i18n` is an array or object, so it's compared by reference, so it's not going to be easy to memo. We could serialize it, or use a deep compare however it may be more expensive than creating a new `I18n`. And considering `AppProvider` should be at the top level of an application, it's unlikely to see many renders

AndrewMusgrave:
Since @beefchimi is busy if we don't have time for this in our next pairing I'll push it over the finish line.

src/components/AppProvider/AppProvider.tsx
```diff
@@ -22,11 +28,9 @@ import {
   globalIdGeneratorFactory,
 } from '../../utilities/unique-id';
 
-interface State {
-  intl: I18n;
-  appBridge: ReturnType<typeof createAppBridge>;
-  link: LinkLikeComponent | undefined;
-}
+type OnlyChild = ReactElement | ReactFragment | ReactPortal;
```
beefchimi: Copied from https://github.com/Shopify/polaris-react/pull/2121

src/components/AppProvider/AppProvider.tsx
```diff
@@ -35,82 +39,54 @@ export interface AppProviderProps extends AppBridgeOptions {
...
-      appBridge: createAppBridge({shopOrigin, apiKey, forceRedirect}),
-    });
-  }
+  }, []);
 
-  render() {
-    const {theme = {logo: null}, children} = this.props;
-    const {intl, appBridge, link} = this.state;
+  useEffect(() => {
+    console.log('useEffect 2 MOUNT');
```
beefchimi: Remove these console logs... maybe look into preventing this effect on first mount!

src/components/AppProvider/AppProvider.tsx
```diff
@@ -35,82 +39,54 @@ export interface AppProviderProps extends AppBridgeOptions {
...
+export function AppProvider({
+  apiKey,
+  shopOrigin,
+  forceRedirect,
+  i18n,
+  linkComponent,
+  theme = {logo: null},
+  children,
+}: AppProviderProps) {
+  const stickyManager = useRef(new StickyManager());
```
sijad:
```suggestion
  const stickyManager = useMemo(() => new StickyManager());
```

src/components/AppProvider/AppProvider.tsx
```diff
@@ -35,82 +39,54 @@ export interface AppProviderProps extends AppBridgeOptions {
...
+  const scrollLockManager = useRef(new ScrollLockManager());
+  const uniqueIdFactory = useRef(new UniqueIdFactory(globalIdGeneratorFactory));
 
-    const {i18n, apiKey, shopOrigin, forceRedirect, linkComponent} = this.props;
+  const [appBridge, setAppBridge] = useState(
```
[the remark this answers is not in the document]
sijad: @BPScott your other [comment](https://github.com/Shopify/polaris-react/pull/2105#discussion_r323897137) was very informative, can you please share your input for this one too?

src/components/AppProvider/AppProvider.tsx
beefchimi: Going with a custom `useLazyRef` solution instead!

AndrewMusgrave:
Adding another reviewer since I probably should review this myself 😛

src/components/AppProvider/AppProvider.tsx
```diff
@@ -45,89 +40,52 @@ export interface AppProviderProps extends AppBridgeOptions {
...
-    const {
-      theme = {},
-      features = {},
-      UNSTABLE_telemetry,
-      children,
-    } = this.props;
-    const {intl, appBridge, link} = this.state;
+  useEffect(() => {
+    stickyManager.current.setContainer(document);
+  }, [stickyManager]);
```
BPScott: Should this be `stickyManager.current`? I can never remember how we want to handle effects based on refs

src/components/AppProvider/AppProvider.tsx
```diff
@@ -45,89 +40,52 @@ export interface AppProviderProps extends AppBridgeOptions {
...
+  );
 
-    // eslint-disable-next-line react/no-did-update-set-state
-    this.setState({
-      link: linkComponent,
-      intl: new I18n(i18n),
-      appBridge: createAppBridge({shopOrigin, apiKey, forceRedirect}),
-    });
-  }
+  const appBridge = useMemo(
```
BPScott: https://reactjs.org/docs/hooks-reference.html#usememo suggests "You may rely on useMemo as a performance optimization, not as a semantic guarantee. In the future, React may choose to “forget” some previously memoized values and recalculate them on next render".

Will a bad thing happen if we change from using one instance of appBridge to a new one because react dropped the memoised value in the middle of the lifetime of an app? Perhaps sticking this in a ref and updating the ref if any of `[apiKey, shopOrigin, forceRedirect]` changes might work instead (this is a half-baked idea)

Same goes for the i18n instance, but I don't think there will be much of an issue with that using a new instance in the middle of the life of an app

AndrewMusgrave: Nope nothing bad will happen from creating a new instance of appBridge IIRC, we're in the clear 👌

src/components/AppProvider/AppProvider.tsx
AndrewMusgrave: As yes, nice catch. Assuming we want to re-run when the stickyManager changes we would want [ref.current]. However, this was auto-filled because we're using a custom hook. We should be able to omit. I'll omit it and re- 🎩 in web

AndrewMusgrave: Removed the auto ignore and added a comment about why we wouldn't want to use ref's as a dependency

danrosenthal:
Hey @beefchimi @AndrewMusgrave this PR is looking a bit stale. What's the status on it?

beefchimi:
@danrosenthal shucks I thought @AndrewMusgrave actually merged this already 😆 

I'm pretty sure its ready to go? Will need to rebase and resolve the conflict, but other than that I think this is 👍

AndrewMusgrave:
I polished it up, however it's still waiting for a review. Once we get a ✅ we should be able to merge it 😄

src/components/AppProvider/AppProvider.tsx
```diff
@@ -45,89 +40,56 @@ export interface AppProviderProps extends AppBridgeOptions {
...
-    const {
-      theme = {},
-      features = {},
-      UNSTABLE_telemetry,
-      children,
-    } = this.props;
-    const {intl, appBridge, link} = this.state;
+  useEffect(() => {
+    stickyManager.current.setContainer(document);
+    // We're using a disable because stickyManager is a ref
```
beefchimi: What does `We're using a disable` mean?

src/components/AppProvider/AppProvider.tsx
```diff
@@ -45,89 +40,56 @@ export interface AppProviderProps extends AppBridgeOptions {
...
+    // Refs are used for values which don't trigger a re-render
+    // and intended to be mutated outside the render data flow
```
beefchimi: `and are intended to be...`?

PR #2114, opened 2019-09-10
[Modal] Add high contrast border

LauraAubin:

Border is missing on high contrast mode in Windows.

Adds a transparent border to the modal so that in high contrast we can see the modal outline.

Couldn't screenshot since I don't have Windows setup.

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist


LauraAubin:
@dpersing could you tophat this change on Windows with high contrast mode? I don't have Windows setup yet 🙏

alex-page:
> @dpersing could you tophat this change on Windows with high contrast mode? I don't have Windows setup yet 🙏

Let me know if you want help setting this up!

src/components/Modal/components/Dialog/Dialog.scss
```diff
@@ -35,6 +35,10 @@ $large-width: rem(980px);
   background: color('white');
   box-shadow: shadow(layer);
 
+  @media (-ms-high-contrast: active) {
+    border: border(transparent);
```
dpersing: For this line, we'll want to use an actual color instead of `transparent`. Basically, there are two ways to make a border show up:

- Use a transparent border by default in regular CSS
- Use a colored border in the `@media` query

Calling for a transparent border with the `@media` query actually displays a transparent border! Which seems weird, but there we are.

No border visible in high contrast

(Please ignore the highlighted text; that's macOS and VMWare not playing nice together while I tried to take a screenshot.)

In this case, I think we might want to go with the default foreground text color for High Contrast, which would look like something like this:

```CSS
    border: 1px solid ms-high-contrast-color('text');
  }
```

Solid default border using the foreground color in High Contrast

There are some [other examples of that usage in the codebase](https://github.com/Shopify/polaris-react/search?q=-ms-high-contrast&unscoped_q=-ms-high-contrast) that might help.

Let me know too if I can help you get set up with test tools!

chloerice: @LauraAubin your hunch was right! The reason we went with the media query was to be declarative about why the transparent border is there. Let's 🔥the media query and replace with a comment instead.

PR #2119, opened 2019-09-11
[Drop zone][a11y] Fixed border colour to have better contrast

mitchmaps:

Fixes [this](https://github.com/Shopify/polaris-ux/issues/162) issue from `polaris-ux`.

@dpersing pointed out some time ago that the Drop zone component has low contrast (1.16) between the border of the file drop area and the background. This made it difficult for low vision users to distinguish between where they could drop files and where they couldn't.

This PR closes the issue by changing the the colour of the Drop zone border to bring the contrast between it and the background up. Previously the border used the _sky_ colour from the Polaris colour system but has now been changed to using the _ink_ colour with the _lightest_ modifier. This change brings the contrast up to 2.79 which is much improved.

#### Note:
Although using the ink lightest colour doesn't bring the contrast quite up to 3 like mentioned in the issue, the next step up brings it almost up to 5 which I thought might have been too dark but it's easy to switch to that if need be.

#### Before:

image

#### After:

image

#### Bonus, Here's what it looks like with the next colour up, bringing the contrast close to 5:

image


mitchmaps:
CC @sarahill 👀

UNRELEASED.md
```diff
@@ -10,6 +10,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Changed border colour of `Drop zone` to have better contrast from the background ([#2119](https://github.com/Shopify/polaris-react/pull/2119))
```
```suggestion
- Changed border color of `Drop zone` to have better contrast from the background ([#2119](https://github.com/Shopify/polaris-react/pull/2119))
```

🇺🇲

dpersing:
Adding @sarahill to the reviewers since her designer 👀 are warranted for sign-off, but I think this is great! Thank you so much, @mitchmaps!

dpersing:
> making the outline lighter when it's disabled

I love it, @sarahill!

sarahill:
Sweet! do you mind making that change @mitchmaps?

mitchmaps:
@sarahill yep! I'll try to get that done before the end of the day!

mitchmaps:
Okay I actually got it working pretty easily, just set a drop zone to disabled and watch the border go back to the old colour indicating it can't be used 💯

mitchmaps:
Alright, thanks! 🚢

PR #2121, opened 2019-09-11
[AppProvider] Added children type to props interface

AndrewMusgrave:

AppProvider uses `{React.Children.only(children)}` which has not been typed. 

* Adding a children type to the AppProvider prop interface
* Remove `React.Children.only` from `AppProvider` & `ThemeProvider` as it's an arbitary restriction

### 🎩 checklist


BPScott:
Silly question: Why do we need to do `{React.Children.only(children)}` instead of `{children}`?

PR #2129, opened 2019-09-12
Hookify DatePicker, DropZone, and Filters

AndrewMusgrave:

Convert examples to hooks

Converting DatePicker, DropZone, and Filters class-based examples to hooks

### STILL TO DO

I'm making quite a few pr's converting examples to hooks so I anticipate lots of `unreleased.md` conflicts, so I'll add the entry before I merge.

```
- Converted `DatePicker`, `DropZone`, and `Filters` examples to functional components ([#2129](https://github.com/Shopify/polaris-react/pull/2129))
```

### How to 🎩

Make sure the behavior of the example mirrors the style guide 

</details>

### 🎩 checklist

### DIFF
--- a/src/components/DatePicker/README.md
+++ b/src/components/DatePicker/README.md
@@ -48,40 +48,30 @@ Date pickers should:
 Use when merchants need to select a date range close to today (today is the default starting position for the date picker).
 
 ```jsx
-class DatePickerExample extends React.Component {
-  state = {
+function DatePickerExample() {
+  const [{month, year}, setDate] = useState({
     month: 1,
     year: 2018,
-    selected: {
-      start: new Date('Wed Feb 07 2018 00:00:00 GMT-0500 (EST)'),
-      end: new Date('Wed Feb 07 2018 00:00:00 GMT-0500 (EST)'),
-    },
-  };
+  });
+  const [selectedDates, setSelectedDates] = useState({
+    start: new Date('Wed Feb 07 2018 00:00:00 GMT-0500 (EST)'),
+    end: new Date('Wed Feb 07 2018 00:00:00 GMT-0500 (EST)'),
+  });
 
-  render() {
-    const {month, year, selected} = this.state;
+  const handleMonthChange = useCallback(
+    (month, year) => setDate({month, year}),
+    [],
+  );
 
-    return (
-      <DatePicker
-        month={month}
-        year={year}
-        onChange={this.handleChange}
-        onMonthChange={this.handleMonthChange}
-        selected={selected}
-      />
-    );
-  }
-
-  handleChange = (value) => {
-    this.setState({selected: value});
-  };
-
-  handleMonthChange = (month, year) => {
-    this.setState({
-      month,
-      year,
-    });
-  };
+  return (
+    <DatePicker
+      month={month}
+      year={year}
+      onChange={setSelectedDates}
+      onMonthChange={handleMonthChange}
+      selected={selectedDates}
+    />
+  );
 }
 ```
 
--- a/src/components/DropZone/README.md
+++ b/src/components/DropZone/README.md
@@ -82,48 +82,45 @@ The following images couldn’t be uploaded:
 Use to allow merchants to upload files. They can drag and drop files into the dashed area, or upload traditionally by clicking the “Add file” button or anywhere inside the dashed area.
 
 ```jsx
-class DropZoneExample extends React.Component {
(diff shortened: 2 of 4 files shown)


src/components/DatePicker/README.md
```diff
@@ -48,40 +48,32 @@ Date pickers should:
...
+    start: new Date('Wed Feb 07 2018 00:00:00 GMT-0500 (EST)'),
+    end: new Date('Wed Feb 07 2018 00:00:00 GMT-0500 (EST)'),
+  });
+
+  const handleMonthChange = useCallback(
+    (month, year) => setDate({month, year}),
+    [],
+  );
+
+  const {month, year} = date;
```
BPScott: You could  do this in the existing destructuring above

```
const [{month, year}, setDate] = useState({
```

src/components/DropZone/README.md
```diff
@@ -142,69 +139,63 @@ Use to pair with a label for better accessibility.
...
-        title="The following images couldn’t be uploaded:"
-        status="critical"
+  return (
+    <Stack vertical>
+      {errorMessage}
+      <DropZone
+        accept="image/*"
+        type="image"
+        onDrop={(files, acceptedFiles, rejectedFiles) => {
+          setFiles([...files, ...acceptedFiles]);
```
BPScott: Should we make this a callback like in the prior example?

src/components/DropZone/README.md
```diff
@@ -142,69 +139,63 @@ Use to pair with a label for better accessibility.
...
-              source={window.URL.createObjectURL(file)}
-            />
-            <div>
-              {file.name} <Caption>{file.size} bytes</Caption>
-            </div>
-          </Stack>
+function DropZoneWithImageFileUpload() {
+  const [files, setFiles] = useState([]);
+  const [rejectedFiles, setRejectedFiles] = useState([]);
+  const [hasError, setHasError] = useState(false);
```
BPScott: Do we need a state to track this? It seems to be derived from other state info so I think we'd be able to do  `const hasError = rejectedFiles.length > 0`

src/components/DropZone/README.md
```diff
@@ -276,68 +263,65 @@ class DropZoneExample extends React.Component {
...
-              source={window.URL.createObjectURL(file)}
-            />
-            <div>
-              {file.name} <Caption>{file.size} bytes</Caption>
-            </div>
-          </Stack>
+function DropZoneAcceptingSVGFilesExample() {
+  const [files, setFiles] = useState([]);
+  const [rejectedFiles, setRejectedFiles] = useState([]);
+  const [hasError, setHasError] = useState(false);
```
BPScott: As above, do we need a state to track this?

src/components/DropZone/README.md
```diff
@@ -346,53 +330,50 @@ class DropZoneExample extends React.Component {
...
-          </DropZone>
-        </Card>
-      </DropZone>
-    );
-  }
+function NestedDropZoneExample() {
+  const [files, setFiles] = useState([]);
+
+  const handleDropZoneDrop = useCallback(
+    (dropFiles, _acceptedFiles, _rejectedFiles) =>
```
BPScott: If we never use `_acceptedFiles` or `_rejectedFiles` do we need to mention them? There's no other parameter after them that we need

src/components/DropZone/README.md
```diff
@@ -425,64 +406,63 @@ Use for cases with tight space constraints, such as variant thumbnails on the Pr
...
+  const [files, setFiles] = useState([]);
+  const [openFileDialog, setOpenFileDialog] = useState(false);
+
+  const handleDropZoneDrop = useCallback(
+    (dropFiles, _acceptedFiles, _rejectedFiles) =>
+      setFiles([...files, ...dropFiles]),
+    [files],
+  );
+  const toggleOpenFileDialog = useCallback(
+    () => setOpenFileDialog(!openFileDialog),
```
BPScott: use callback style of `setOpenFileDialog`

src/components/DropZone/README.md
AndrewMusgrave: I think it'd be nice to keep them in, so consumers know the parameters are available without skimming the docs

AndrewMusgrave:
Thanks for the suggestions @BPScott 😄  , I was only converting the examples but it looks like there were quite a few quick wins, not sure why hasError was on state 🤷‍♂

PR #2138, opened 2019-09-13
[ResourceList-BulkAction][Checkbox] Handle focus on bulk actions and converts checkbox into a functional component.

dleroux:

Fixes https://github.com/Shopify/polaris-react/issues/792

I believe this is the last of the ResourceList a11y issues outline in [#792].

In order to switch focus between the plain `CheckableButton` checkbox and the ones inside the BulkActions, we needed to keep track of the button that was clicked and which of the 3 possible buttons that would receive focus.

1. The 3 buttons are tracked via context.
2. `handleToggleAll` is only called when the bulk action or plain button are clicked. From there, depending on screen size we focus the right button.
3. To enable focus on the checkbox I needed to forwardRef on the checkbox. According to this: https://reactjs.org/docs/forwarding-refs.html#note-for-component-library-maintainers This would be a breaking changes. Discussed with @AndrewMusgrave and we're not sure we agree in our case. @BPScott thoughts?

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {
  Page,
  ResourceList,
  Card,
  Avatar,
  ResourceItem,
  TextStyle,
} from '../src';

  return (
    <Page>
      <ResourceListExample />
    </Page>
  );
}
  state = {
    selectedItems: [],
  };

  };

  renderItem = (item) => {
    return (
      <ResourceItem
        id={id}
        url={url}
        media={media}
      >
        <h3>
        </h3>
        <div>{location}</div>
      </ResourceItem>
    );
  };

  render() {
    const resourceName = {
      singular: 'customer',
      plural: 'customers',
    };
    const items = [
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
      },
    ];
    const promotedBulkActions = [
      {
        content: 'Edit customers',
      },
    ];
    const bulkActions = [
      {
        content: 'Add tags',
      },
      {
        content: 'Remove tags',
      },
      {
        content: 'Delete customers',
      },
    ];
    return (
      <Card>
        <ResourceList
          resourceName={resourceName}
          items={items}
          renderItem={this.renderItem}
          bulkActions={bulkActions}
        />
      </Card>
    );
  }
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -1,16 +1,11 @@
-import React from 'react';
+import React, {useRef, useImperativeHandle} from 'react';
 import {MinusMinor, TickSmallMinor} from '@shopify/polaris-icons';
-import {createUniqueIDFactory} from '@shopify/javascript-utilities/other';
-
+import {useUniqueId} from '../../utilities/unique-id';
 import {classNames} from '../../utilities/css';
-import {
-  withAppProvider,
-  WithAppProviderProps,
-} from '../../utilities/with-app-provider';
 import {Choice, helpTextID} from '../Choice';
 import {errorTextID} from '../InlineError';
 import {Icon} from '../Icon';
-import {Error, Key} from '../../types';
+import {Error, Key, CheckboxHandles} from '../../types';
 
 import styles from './Checkbox.scss';
 
@@ -44,46 +39,53 @@ export interface BaseProps {
 }
 
 export interface CheckboxProps extends BaseProps {}
-type CombinedProps = CheckboxProps & WithAppProviderProps;
 
-const getUniqueID = createUniqueIDFactory('Checkbox');
-
-class Checkbox extends React.PureComponent<CombinedProps, never> {
-  private inputNode = React.createRef<HTMLInputElement>();
-
-  handleInput = () => {
-    const {onChange, id, disabled} = this.props;
-
-    if (onChange == null || this.inputNode.current == null || disabled) {
-      return;
-    }
-
-    onChange(!this.inputNode.current.checked, id as any);
-    this.inputNode.current.focus();
-  };
-
-  handleKeyUp = (event: React.KeyboardEvent) => {
-    const {keyCode} = event;
-
-    if (keyCode !== Key.Space) return;
-    this.handleInput();
-  };
-
-  render() {
-    const {
+export const Checkbox = React.forwardRef<CheckboxHandles, CheckboxProps>(
+  function Checkbox(
+    {
       ariaDescribedBy: ariaDescribedByProp,
-      id = getUniqueID(),
       label,
       labelHidden,
-      helpText,
       checked = false,
-      error,
+      helpText,
       disabled,
-      onFocus,
-      onBlur,
+      id: idProp,
       name,
       value,
-    } = this.props;
+      error,
+      onChange,
+      onFocus,
+      onBlur,
+    }: CheckboxProps,
+    ref,
+  ) {
+    const inputNode = useRef<HTMLInputElement>(null);
+
+    const id = useUniqueId('Checkbox', idProp);
+
+    useImperativeHandle(ref, () => ({
+      focus: () => {
+        if (inputNode.current) {
+          inputNode.current.focus();
+        }
+      },
+    }));
+
+    const handleInput = () => {
+      if (onChange == null || inputNode.current == null || disabled) {
+        return;
+      }
+      onChange(!inputNode.current.checked, id);
+      inputNode.current.focus();
+    };
+
+    const handleKeyUp = (event: React.KeyboardEvent) => {
+      const {keyCode} = event;
+
+      if (keyCode !== Key.Space) return;
+      handleInput();
+    };
+
     const describedBy: string[] = [];
     if (error && typeof error !== 'boolean') {
       describedBy.push(errorTextID(id));
@@ -123,12 +125,12 @@ class Checkbox extends React.PureComponent<CombinedProps, never> {
         helpText={helpText}
         error={error}
         disabled={disabled}
-        onClick={this.handleInput}
+        onClick={handleInput}
       >
         <span className={wrapperClassName}>
           <input
-            onKeyUp={this.handleKeyUp}
-            ref={this.inputNode}
+            onKeyUp={handleKeyUp}
+            ref={inputNode}
             id={id}
             name={name}
             value={value}
@@ -153,15 +155,11 @@ class Checkbox extends React.PureComponent<CombinedProps, never> {
       </Choice>
       /* eslint-enable jsx-a11y/no-redundant-roles */
     );
-  }
-}
+  },
+);
 
 function noop() {}
 
 function stopPropagation<E>(event: React.MouseEvent<E>) {
   event.stopPropagation();
 }
-
-// Use named export once withAppProvider is refactored away
-// eslint-disable-next-line import/no-default-export
-export default withAppProvider<CheckboxProps>()(Checkbox);
--- a/src/components/ResourceList/ResourceList.tsx
+++ b/src/components/ResourceList/ResourceList.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 
 import debounce from 'lodash/debounce';
 import {EnableSelectionMinor} from '@shopify/polaris-icons';
-
+import {CheckboxHandles} from '../../types';
 import {classNames} from '../../utilities/css';
 import {Button} from '../Button';
 import {EventListener} from '../EventListener';
@@ -16,6 +16,8 @@ import {
   ResourceListContext,
   ResourceListSelectedItems,
   SELECT_ALL_ITEMS,
+  CheckableButtonKey,
+  CheckableButtons,
 } from '../../utilities/resource-list';
 import {Select, SelectOption} from '../Select';
 import {EmptySearchResult} from '../EmptySearchResult';
@@ -42,6 +44,7 @@ interface State {
   loadingPosition: number;
   lastSelected: number | null;
   smallScreen: boolean;
+  checkableButtons: CheckableButtons;
 }
 
 export interface ResourceListProps {
@@ -137,6 +140,7 @@ class ResourceList extends React.Component<CombinedProps, State> {
       loadingPosition: 0,
       lastSelected: null,
       smallScreen: isSmallScreen(),
+      checkableButtons: new Map(),
     };
   }
 
@@ -375,7 +379,6 @@ class ResourceList extends React.Component<CombinedProps, State> {
       polaris: {intl},
     } = this.props;
     const {selectMode, loadingPosition, smallScreen} = this.state;
-
     const filterControlMarkup = filterControl ? (
       <div className={styles.FiltersWrapper}>{filterControl}</div>
     ) : null;
@@ -394,6 +397,7 @@ class ResourceList extends React.Component<CombinedProps, State> {
           paginatedSelectAllText={this.paginatedSelectAllText}
           actions={bulkActions}
           disabled={loading}
+          smallScreen={smallScreen}
         />
       </div>
     ) : null;
@@ -558,6 +562,7 @@ class ResourceList extends React.Component<CombinedProps, State> {
       resourceName,
       loading,
       onSelectionChange: this.handleSelectionChange,
+      registerCheckableButtons: this.handleCheckableButtonRegistration,
     };
 
     return (
@@ -643,6 +648,17 @@ class ResourceList extends React.Component<CombinedProps, State> {
     return this.props.items.slice(min, max + 1).map(resolveItemId);
   };
 
+  private handleCheckableButtonRegistration = (
+    key: CheckableButtonKey,
+    button: CheckboxHandles,
+  ) => {
+    this.setState(({checkableButtons}) => {
+      return {
+        checkableButtons: new Map(checkableButtons).set(key, button),
+      };
+    });
+  };
+
   private handleSelectionChange = (
     selected: boolean,
     id: string,
@@ -723,6 +739,7 @@ class ResourceList extends React.Component<CombinedProps, State> {
       idForItem = defaultIdForItem,
     } = this.props;
 
+    const {checkableButtons} = this.state;
     let newlySelectedItems: string[] = [];
 
     if (
@@ -743,9 +760,24 @@ class ResourceList extends React.Component<CombinedProps, State> {
       this.handleSelectMode(true);
     }
 
+    let checkbox: CheckboxHandles | undefined;
+
+    if (isSmallScreen()) {
+      checkbox = checkableButtons.get('bulkSm');
+    } else if (newlySelectedItems.length === 0) {
+      checkbox = checkableButtons.get('plain');
+    } else {
+      checkbox = checkableButtons.get('bulkLg');
+    }
+
     if (onSelectionChange) {
       onSelectionChange(newlySelectedItems);
     }
+
+    // setTimeout ensures execution after the Transition on BulkActions
+    setTimeout(() => {
+      checkbox && checkbox.focus();
+    }, 0);
   };
 }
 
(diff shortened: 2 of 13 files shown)


src/components/Checkbox/Checkbox.tsx
```diff
@@ -43,125 +44,138 @@ export interface BaseProps {
...
 const getUniqueID = createUniqueIDFactory('Checkbox');
 
-class Checkbox extends React.PureComponent<CombinedProps, never> {
-  private inputNode = React.createRef<HTMLInputElement>();
-
-  handleInput = () => {
-    const {onChange, id, disabled} = this.props;
-
-    if (onChange == null || this.inputNode.current == null || disabled) {
+const ImperitaveCheckbox: RefForwardingComponent<
```
BPScott: I think we'd be able to sidestep this explicit typing stuff if we define this inline rather than a separate function

```
export const Checkbox = React.forwardRef(function Checkbox({
  /* Props here */
}) {
  /* Blah */
});

File: src/components/Checkbox/Checkbox.tsx [hunk 2]
```diff
@@ -43,125 +44,138 @@ export interface BaseProps {
...
+    onChange,
+    onFocus,
+    onBlur,
+  },
+  ref,
+) => {
+  const inputNode = useRef<HTMLInputElement>(null);
+
+  const generatedId = useRef(getUniqueID());
+  const id = idProp || generatedId.current;
```
BPScott: use `useUniqueId` instead of createUniqueIDFactory and ref dancing

src/components/Checkbox/Checkbox.tsx
dleroux: I couldn't get this to work for some reason. I'll give it another go.

dleroux:
After a discussion wit @AndrewMusgrave and @BPScott regarding this being a [breaking change](https://reactjs.org/docs/forwarding-refs.html#note-for-component-library-maintainers) we agreed that in our case we wouldn't consider this a breaking change because even though the type for Checkbox is changing, we did not expose any public API or forwardRef on Checkbox, therefore, this shouldn't affect anything we support today.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -43,47 +38,58 @@ export interface BaseProps {
...
+          inputNode.current.focus();
+        }
+      },
+    }));
+
+    const handleInput = () => {
+      if (onChange == null || inputNode.current == null || disabled) {
+        return;
+      }
+      onChange(!inputNode.current.checked, id as any);
```
BPScott: It looks like `id` is now typed to be `string` so do you need that `as any`?

src/components/ResourceList/tests/ResourceList.test.tsx
```diff
@@ -750,6 +753,118 @@ describe('<ResourceList />', () => {
...
+            (wrap) => wrap.is(CheckableButton) && !wrap.prop('plain'),
+          );
+
+          trigger(deselectAllCheckableButton, 'onToggleAll');
+
+          const checkBox = deselectAllCheckableButton.find(
+            'input[type="checkbox"]',
+          );
+
+          expect(checkBox.instance()).toBe(document.activeElement);
```
AndrewMusgrave: `instance`  returns a react component, however, `getDOMNode` returns the dom element

src/components/ResourceList/tests/ResourceList.test.tsx
```diff
@@ -750,6 +753,118 @@ describe('<ResourceList />', () => {
...
+          const resourceList = mountWithAppProvider(
+            <ResourceList
+              items={itemsWithID}
+              selectedItems={allSelectedIDs}
+              renderItem={renderItem}
+              promotedBulkActions={promotedBulkActions}
+            />,
+          );
+
+          const deselectAllCheckableButton = resourceList.findWhere(
```
AndrewMusgrave: With react-testing you can automatically filter by props in the matcher

src/components/ResourceList/ResourceList.tsx
```diff
@@ -643,6 +650,17 @@ class ResourceList extends React.Component<CombinedProps, State> {
     return this.props.items.slice(min, max + 1).map(resolveItemId);
   };
 
+  private handleCheckableButtonRegistration = (
+    key: CheckableButtonKey,
+    button: CheckableButtonNode,
+  ) => {
+    const {checkableButtons} = this.state;
+    checkableButtons.set(key, button);
```
AndrewMusgrave: Uh-oh we're mutating state here 🚫

src/components/ResourceList/components/BulkActions/BulkActions.tsx
```diff
@@ -27,6 +27,8 @@ const MAX_PROMOTED_ACTIONS = 2;
 export interface BulkActionsProps {
   /** Visually hidden text for screen readers */
   accessibilityLabel?: string;
+  /** Whether to render the small screen BulkActions or not */
+  smallScreen?: boolean;
```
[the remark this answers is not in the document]
dleroux: In this case smallScreen is const random number for the resourceList. I could duplicate the code or add it to context, but I opted for a prop. What do you think?

src/components/ResourceList/components/BulkActions/tests/BulkActions.test.tsx
```diff
@@ -354,5 +354,29 @@ describe('<BulkActions />', () => {
         expect(spy).toHaveBeenCalled();
       });
     });
+
+    describe('smallScreen', () => {
+      it('renders only the large screen bulkactions if smallScreen is false', () => {
+        const bulkActions = mountWithAppProvider(
+          <BulkActions {...bulkActionProps} selectMode />,
+        );
+        const smallGroup = findByTestID(bulkActions, 'smallGroup');
```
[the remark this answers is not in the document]
dleroux: I wanted to use findWhere and use the key, but the keys get striped. Just using Find doesn't guarantee us which will be found.

src/components/ResourceList/ResourceList.tsx
dleroux: 🙈

src/components/index.ts
```diff
@@ -28,7 +28,7 @@ export {Caption, CaptionProps} from './Caption';
 
 export {Card, CardProps} from './Card';
 
-export {Checkbox, CheckboxProps} from './Checkbox';
+export {Checkbox, CheckboxProps, CheckboxHandles} from './Checkbox';
```
BPScott: I'd like to avoid exporting this publicly, people shouldn't need it

src/components/ResourceList/components/BulkActions/tests/BulkActions.test.tsx
```diff
@@ -354,5 +354,29 @@ describe('<BulkActions />', () => {
...
+        const largeGroup = findByTestID(bulkActions, 'largeGroup');
+
+        expect(smallGroup.exists()).toBe(false);
+        expect(largeGroup.exists()).toBe(true);
```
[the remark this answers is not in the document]
dleroux: Same as above. No guarantee which was found.

src/components/ResourceList/tests/ResourceList.test.tsx
dleroux: Interesting how this did pass though.

src/components/index.ts
dleroux: fair enough. I hade to disable strict-component-boundaries in the types file to do this.

dpersing:
👋 @BPScott and @AndrewMusgrave, any chance you've taken a look at this yet?

src/components/ResourceList/tests/ResourceList.test.tsx
```diff
@@ -750,6 +753,118 @@ describe('<ResourceList />', () => {
...
+          const resourceList = mountWithAppProvider(
+            <ResourceList
+              items={itemsWithID}
+              renderItem={renderItem}
+              promotedBulkActions={promotedBulkActions}
+            />,
+          );
+
+          const selectAllCheckableButton = resourceList.findWhere(
+            (wrap) => wrap.is(CheckableButton) && wrap.prop('plain') === true,
```
AndrewMusgrave: No strong feelings, we usually coerce these types as values rather than compare against a boolean but I'll leave the choice up to you 

```suggestion
            (wrap) => wrap.is(CheckableButton) && wrap.prop('plain'),
```

src/components/ResourceList/tests/ResourceList.test.tsx
```diff
@@ -750,6 +753,118 @@ describe('<ResourceList />', () => {
...
+          );
+
+          trigger(selectAllCheckableButton, 'onToggleAll');
+
+          const deselectAllCheckbox = resourceList
+            .findWhere(
+              (wrap) => wrap.is(CheckableButton) && !wrap.prop('plain'),
+            )
+            .find('input[type="checkbox"]');
```
AndrewMusgrave: Just curious why you didn't include this in `findWhere` or before the `findWhere` (both should reduce the number of operations, although probably by a negligible amount)(not asking for a code change, just curious)

src/components/ResourceList/tests/ResourceList.test.tsx
```diff
@@ -750,6 +753,118 @@ describe('<ResourceList />', () => {
...
+          const resourceList = mountWithAppProvider(
+            <ResourceList
+              items={itemsWithID}
+              renderItem={renderItem}
+              selectedItems={allSelectedIDs}
+              promotedBulkActions={promotedBulkActions}
+            />,
+          );
+
+          const deselectAllCheckableButton = resourceList.findWhere(
```
AndrewMusgrave: Seems like we use this logic a lot, might be worth bringing it out into a function

`function deselectAllCheckableButtons(wrapper) {}` | `function selectedAllCheckableButtons(wrapper)`

src/components/ResourceList/tests/ResourceList.test.tsx
```diff
@@ -750,6 +753,118 @@ describe('<ResourceList />', () => {
...
+            (wrap) => wrap.is(CheckableButton) && !wrap.prop('plain'),
+          );
+
+          trigger(deselectAllCheckableButton, 'onToggleAll');
+
+          const selectAllCheckableCheckbox = resourceList
+            .findWhere(
+              (wrap) => wrap.is(CheckableButton) && wrap.prop('plain') === true,
```
AndrewMusgrave: Same comment as a few . above

```suggestion
              (wrap) => wrap.is(CheckableButton) && wrap.prop('plain'),
```

src/components/ResourceList/components/CheckableButton/CheckableButton.tsx
```diff
@@ -9,6 +14,7 @@ export interface CheckableButtonProps {
   label?: string;
   selected?: boolean | 'indeterminate';
   selectMode?: boolean;
+  smallScreen?: boolean;
```
AndrewMusgrave: Should we create an issue to add `RL smallScreen` to media query provider and remove this prop?

src/components/ResourceList/components/CheckableButton/CheckableButton.tsx
```diff
@@ -24,7 +30,28 @@ export function CheckableButton({
...
   disabled,
+  smallScreen,
 }: CheckableButtonProps) {
+  const checkBoxRef = useRef<CheckboxHandles>(null);
+
+  const {registerCheckableButtons} = React.useContext(ResourceListContext);
+
+  let currentKey: CheckableButtonKey = 'plain';
+
+  if (plain) {
```
AndrewMusgrave: Curious what you think about a switch statement here?

```
case plain:
  break;
case smallScreen:
  currentKey = 'bulkSm';
  break;
default:
  currentKey = 'bulkLg';
```

or if there's a condition that lets us remove `currentKey = 'plain';` or default `currentKey` to undefined. Just some written thoughts, don't let this block you from 🐑 🇮🇹

src/utilities/resource-list/types.ts
```diff
@@ -1,3 +1,9 @@
+import {CheckboxHandles} from '../../types';
+
 export type ResourceListSelectedItems = string[] | 'All';
 
+export type CheckableButtonNode = CheckboxHandles;
```
AndrewMusgrave: Do we need this type?

src/components/ResourceList/components/CheckableButton/CheckableButton.tsx
dleroux: If you feel like we should we could do that. I actually don't mind the prop here making this a controllable dumb component.

src/utilities/resource-list/types.ts
dleroux: Not really I guess. I wasn't thinking it made sense just because of the naming, but I can just us the CheckboxHandles. I'll remove it since it brings up the question.

src/components/ResourceList/components/CheckableButton/CheckableButton.tsx
dleroux: in this case, we're switching on 2 different props. I can simply by defaulting to the `else` instead.

src/components/ResourceList/tests/ResourceList.test.tsx
dleroux: 🤔 the findWhere will return the `wrap` based on the filtering condition. I'm after the checkbox inside the wrap. Putting the `find` before would not allow me to `findWhere` based on parents asaik. In any case, this has been simplified, because as you suggested I made the findWhere into a function.

dleroux:
> Using the keyboard it looks like the correct checkbox has focus however it doesn't show visually

Not sure why this is, everything checks out and the input does get focus. For some reason with the keyboard it doesn't trigger the visual change on the checkbox. Tried forcing a reflow and it doesn't work either. Sadly the only thing that does work is `requestAnimationFrame`, so I had to mock it.

The rest of the suggestions have been applied.

dpersing:
> or some reason with the keyboard it doesn't trigger the visual change on the checkbox.

```
.Checkbox-Input_1uEY6 {
    position: absolute !important;
    top: 0;
    clip: rect(1px, 1px, 1px, 1px) !important;
    overflow: hidden !important;
    height: 1px !important;
    width: 1px !important;
    padding: 0 !important;
    border: 0 !important;
}
```

@dleroux I suspect this is because the actual checkbox is visually hidden, and the visually mocked on is getting updated when the user unchecks it. Focus is on the "old" version of the visual checkbox.

If we can avoid hiding the actual checkbox visually, this issue should go away.

AndrewMusgrave:
Is this ready for review again?

dleroux:
Yes it is.

src/components/ResourceList/ResourceList.tsx
```diff
@@ -743,9 +761,24 @@ class ResourceList extends React.Component<CombinedProps, State> {
...
+    } else {
+      checkbox = checkableButtons.get('bulkLg');
+    }
+
     if (onSelectionChange) {
       onSelectionChange(newlySelectedItems);
     }
+
+    // setTimeout ensures execution after the Transition on BulkActions
+    setTimeout(() => {
```
AndrewMusgrave: cc/ @beefchimi similar to what we were looking at yesterday

beefchimi: Similar, except this is able to get away with a `0ms` Timeout 😄 

In this case, is `requestAnimationFrame` an alternative solution the problem? Also... is this needed specifically because of being nested in `react-transition-group`?

dleroux: Yes, it is needed because of `react-transition-group`. They trigger their transition using `setTimeout`. So using `setTimeout` here ensures that theirs is called first. `requestAnimationFrame` was working, but because they are using `setTimeout` we felt it was better here.

PR #2141, opened 2019-09-13
Updated trigger to use act

AndrewMusgrave:

We use `trigger` over 150 so it's going to be easier to update the function rather than always using act

Copying over the trigger function from `enzyme-utilities`. Since we're migrating away from enzyme I don't think it's worth adding it as a dependency and having more enzyme utilities 😄 

### How to 🎩

Watch the tests run errorless 😎

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -10,6 +10,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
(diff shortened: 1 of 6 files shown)


BPScott:
Nice tricks!

Does this mean we can get rid of the rexport of `act` in test-utilities/legacy? I added it in #1900 so we can wrap calls to `trigger` but if that's no longer needed then the rexport could go too

AndrewMusgrave:
It does 😄 I'll update the PR to remove it

PR #2148, opened 2019-09-16
[WIP][Button] Add pressed state to 

AndrewMusgrave:

Fixes #0000 

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -124,6 +124,50 @@ $spinner-size: rem(20px);
   }
 }
 
+.pressed {
+  background: color(sky);
+  border-color: darken(color(sky, dark), 4%);
+  box-shadow: 0 0 0 0 transparent,
+    inset 0 1px 1px 0 rgba(color('ink', 'lighter'), 0.1),
+    inset 0 1px 4px 0 rgba(color('ink', 'lighter'), 0.1);
+
+  &:hover {
+    transition-duration: duration(fast);
+    background: darken(color(sky), 2%);
+    border-color: darken(color(sky, dark), 4%);
+    box-shadow: 0 0 0 0 transparent,
+      inset 0 1px 1px 0 rgba(color('ink', 'lighter'), 0.1),
+      inset 0 1px 4px 0 rgba(color('ink', 'lighter'), 0.1);
+  }
+
+  &:focus {
+    border-color: color('indigo');
+    box-shadow: 0 0 0 1px color('indigo'),
+      inset 0 1px 1px 0 rgba(color('ink', 'lighter'), 0.1),
+      inset 0 1px 4px 0 rgba(color('ink', 'lighter'), 0.1);
+  }
+
+  &:active {
+    background: darken(color(sky), 4%);
+    border-color: darken(color(sky, dark), 4%);
+    box-shadow: 0 0 0 0 transparent,
+      inset 0 1px 1px 0 rgba(color('ink', 'lighter'), 0.1),
+      inset 0 1px 4px 0 rgba(color('ink', 'lighter'), 0.2);
+  }
+
+  &.disabled {
+    background: lighten(color(sky), 4%);
+    color: color('ink', 'lightest');
+    border-color: color(sky, dark);
+    box-shadow: inset 0 1px 1px 0 rgba(color('ink', 'lighter'), 0.1);
+  }
+
+  @media (-ms-high-contrast: active) {
+    color: ms-high-contrast-color('button-text');
+    background: ms-high-contrast-color('button-text-background');
+  }
+}
+
 // The way the designs work, we need to do lots of reaching down to
 // target the content in pseudo-selectors, so we need higher specificity
 // in this case.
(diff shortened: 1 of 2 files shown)


AndrewMusgrave:
Slight delay - Waiting to pair with a designer to hammer out visual states

src/components/Button/Button.tsx
```diff
@@ -61,7 +63,10 @@ export interface ButtonProps {
   ariaControls?: string;
   /** Tells screen reader the controlled element is expanded */
   ariaExpanded?: boolean;
-  /** Tells screen reader the element is pressed */
+  /**
+   * @deprecated As of release 4.X.X, replaced by {@link https://polaris.shopify.com/components/structure/page#props-pressed}
```
AndrewMusgrave: TODO: Edit this while merging

src/styles/shared/_buttons.scss
```diff
@@ -129,8 +129,46 @@
   color: color('white');
 }
 
-@mixin button-outline($outline-color) {
-  background: transparent;
+@mixin button-filled-pressed($button-color, $focus-color) {
+  background: linear-gradient(
```
AndrewMusgrave: Grabbed styles from ToggleButton in web

tmlayton:
Hey @AndrewMusgrave, can we hold off on this until after we make the changes with CSS custom properties and theming in the next few months? Otherwise, let’s pair on reworking this mixin to avoid color manipulation as it makes the code for backwards compatibility kind of gross. Thanks!

BPScott:
I got tagged as this is modifying the public sass API. Eventually we want to stop touching that but we've not got the time to bash out that plan. The change to that looks fine for now.

Removing myself from review as I haven't got the time to look at the rest of this

AndrewMusgrave:
> Hey AndrewMusgrave, can we hold off on this until after we make the changes with CSS custom properties and theming in the next few months? Otherwise, let’s pair on reworking this mixin to avoid color manipulation as it makes the code for backwards compatibility kind of gross. Thanks! 


A few months is a little bit longer to let this go stale, I'll remove all the sass manipulation 😄 

> Can you add an example so folks can easily tophat this change and see it in use in our documentation once it's released?

Good call, I'll include an example. There are a lot of button variants though. The PR description has a complex playground to see the pressed states for 🎩 

@mirualves Want to spend a few minutes going over the visuals Monday for a final check?

src/components/Button/Button.scss
```diff
@@ -7,6 +7,17 @@ $slim-vertical-padding: ($slim-min-height - line-height(body) - rem(2px)) / 2;
...
+$pressed-hover-background: rgb(217, 222, 228);
+$pressed-active-background: rgb(211, 217, 223);
+$primary-pressed-lightest: rgb(63, 78, 174);
+$primary-pressed-light: rgb(60, 75, 167);
+$primary-pressed-dark: rgb(16, 23, 60);
+$primary-pressed-darkest: rgb(11, 15, 39);
+$destructive-pressed-lightest: rgb(176, 43, 19);
+$destructive-pressed-light: rgb(167, 41, 18);
+$destructive-pressed-dark: rgb(117, 4, 10);
+$destructive-pressed-darkest: rgb(93, 3, 8);
```
danrosenthal: I've got some concerns on the impact all these snowflake colors will have on the color system work, and the move to client-side theming. Some context on how these colors were generated would give us a stronger ability to upgrade this to consume the color system. I actually think this was better with the `darken` and `rgba` functions, since we know where the colors started and how they got to the final state. That feels easier to systematically reproduce.

@tmlayton, what do you think about going back to the way it was before, given what I said above?

danrosenthal: Either that, or leave comments on each of these colors to show how they were achieved using programmatic darkening of our color tokens.

dleroux: I agree with Dan here. If there were achieved using the `darken` functions it might be best to leave it. It does make it clear on how to reproduce across themes.

AndrewMusgrave: Added comments on how the colors were achieved

AndrewMusgrave:
> Make sure there's a test passing the pressed property and this should be good to go.

I added a few tests. Historically we haven't tested visuals, is this something we're starting to do?

@danrosenthal I moved the new mixings into `Button.scss` so they're not exposed

danrosenthal:
> I added a few tests. Historically we haven't tested visuals, is this something we're starting to do?

It's not so much about testing visuals, as it is just having coverage on all code paths. I'd even be fine with adding `pressed` to an existing test just to get that code path to execute. The presence of the classNames is less interesting, and can be covered by VRT.

> @danrosenthal I moved the new mixings into Button.scss so they're not exposed

Fantastic!

src/components/Button/Button.tsx
```diff
@@ -39,6 +39,8 @@ export interface ButtonProps {
   textAlign?: TextAlign;
   /** Gives the button a subtle alternative to the default button styling, appropriate for certain backdrops */
   outline?: boolean;
+  /** Demonstrates a pressed state */
```
dleroux:
```suggestion
  /** Gives the button the appearance of being pressed */
```

PR #2153, opened 2019-09-17
[Link] Improve docblocks and readme for links

BPScott:

People are unclear if the `external` prop on links should be used for sites external to shopify, or anything that opens in a new tab.

You should use it for anything that opens in a new tab.

Make it clear external is to be used for an page that opens in a new tab
Make wording and ordering consistent between Link and UnstyledLink

Clarify readme to make it clear the important thing is "you're going anywhere in a new tab", not "you're going outside your current domain"

### DIFF
--- a/src/utilities/link/types.ts
+++ b/src/utilities/link/types.ts
@@ -1,13 +1,13 @@
 export interface LinkLikeComponentProps
   extends React.HTMLProps<HTMLAnchorElement> {
-  /** A destination to link to */
+  /** The url to link to */
   url: string;
-  /** Forces url to open in a new tab */
-  external?: boolean;
-  /** Tells the browser to download the url instead of opening it. Provides a hint for the downloaded filename if it is a string value. */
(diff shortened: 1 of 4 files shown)


src/utilities/link/types.ts
```diff
@@ -1,13 +1,13 @@
...
-  /** A destination to link to */
+  /** The url to link to */
   url: string;
-  /** Forces url to open in a new tab */
+  /**	The content to display inside the link */
+  children?: React.ReactNode;
+  /** Makes the link open in a new tab */
   external?: boolean;
-  /** Tells the browser to download the url instead of opening it. Provides a hint for the downloaded filename if it is a string value. */
+  /** Makes the browser to download the url instead of opening it. Provides a hint for the downloaded filename if it is a string value. */
```
kaelig:
```suggestion
  /** Makes the browser download the url instead of opening it. Provides a hint for the downloaded filename if it is a string value. */
```

PR #2155, opened 2019-09-17
[RenderAfterMount][useMounted] Add utilities

AndrewMusgrave:

Adding two small utilities that'll help with code that needs to be separated from server/client.

Adding `RenderAfterMount` and `useMounted`

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/AfterInitialMount/AfterInitialMount.tsx
+++ b/src/components/AfterInitialMount/AfterInitialMount.tsx
@@ -1,0 +1,14 @@
+import React, {ReactNode} from 'react';
+import {useIsAfterInitialMount} from '../../utilities/use-is-after-initial-mount';
+
+interface Props {
+  children?: ReactNode;
+  fallback?: ReactNode;
+}
+
+export function AfterInitialMount({children, fallback = null}: Props) {
+  const isMounted = useIsAfterInitialMount();
+  const content = isMounted ? children : fallback;
+
+  return <React.Fragment>{content}</React.Fragment>;
+}
--- a/src/components/AfterInitialMount/index.ts
+++ b/src/components/AfterInitialMount/index.ts
@@ -1,0 +1,1 @@
+export {AfterInitialMount} from './AfterInitialMount';
--- a/src/components/AfterInitialMount/test/AfterInitialMount.test.tsx
+++ b/src/components/AfterInitialMount/test/AfterInitialMount.test.tsx
@@ -1,0 +1,42 @@
+import React, {useEffect} from 'react';
+import {mount} from 'test-utilities';
+import {AfterInitialMount} from '../AfterInitialMount';
+
+describe('AfterInitialMount', () => {
+  it('renders fallback before mounting', () => {
+    const spy = jest.fn();
+
+    function Fallback() {
+      spy();
+      return null;
+    }
+
+    mount(<AfterInitialMount fallback={<Fallback />} />);
+    expect(spy).toHaveBeenCalledTimes(1);
+  });
+
+  it('renders children after mounting', () => {
+    const fallbackSpy = jest.fn();
+
+    function Fallback() {
+      useEffect(() => {
+        fallbackSpy(true);
+      }, []);
+
+      return null;
+    }
+
+    function Children() {
+      return null;
+    }
+
+    const afterInitialMount = mount(
+      <AfterInitialMount fallback={<Fallback />}>
+        <Children />
+      </AfterInitialMount>,
+    );
+
+    expect(fallbackSpy).toHaveBeenCalledWith(true);
+    expect(afterInitialMount).toContainReactComponent(Children);
+  });
+});
--- a/src/utilities/tests/use-is-after-initial-mount.test.tsx
+++ b/src/utilities/tests/use-is-after-initial-mount.test.tsx
@@ -1,0 +1,32 @@
+import React from 'react';
+import {mount} from 'test-utilities';
+import {useIsAfterInitialMount} from '../use-is-after-initial-mount';
+
+describe('useIsAfterInitialMount', () => {
+  it('starts off false', () => {
+    let isAfterInitialMountValue: boolean | undefined;
+    function Component() {
+      const isAfterInitialMount = useIsAfterInitialMount();
+      // eslint-disable-next-line shopify/jest/no-if
+      if (isAfterInitialMountValue === undefined)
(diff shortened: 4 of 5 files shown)


src/components/RenderAfterMount/RenderAfterMount.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
```
AndrewMusgrave: Unfortunately, we don't really have a way to test this right now. We'll need to figure out how to have a node environment and jsdom maybe jest [projects](https://jestjs.io/docs/en/configuration#projects-array-string-projectconfig)

AndrewMusgrave:
Thoughts?

src/components/RenderAfterMount/RenderAfterMount.tsx
```diff
@@ -0,0 +1,12 @@
...
+import {useMounted} from '../../utilities/use-mounted';
+
+interface Props {
+  children?: React.ReactNode;
+}
+
+export function RenderAfterMount({children}: Props) {
```
BPScott: Per comment below, feedback depends on what this is for. I wonder if there is a name that more accurately describes the intent more succinctly than the `how`?

Prefixing with `Render` feels a bit superflous, perhaps `AfterMount` (if this is for after *every* mount), 
`AfterInitialMount` (if this is for after the initial mount)

What do you think to accepting a `fallback` prop (defaulting to null) ([like Suspense](https://itnext.io/what-the-heck-is-this-in-react-suspense-c5e641e487a) to allow you to define initial content.

BPScott:
Nice idea! I'd love to see some examples of where we'd use this. I'm sure we've got a fair few but I can't think of any off the top of my head as I've not done much with out meatier components

src/utilities/use-mounted.ts
```diff
@@ -0,0 +1,11 @@
+import {useState, useEffect} from 'react';
+
+export function useMounted() {
```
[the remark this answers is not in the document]
AndrewMusgrave: > It'd be great to add a docblock describing what the hook does and returns (like useUniqueId

Yes, 😍 I've always wished we used jsdocs for our utilities.

`useIsMountedState` and `useIsMountedRef` would have two differently effects. The state version is so we can render empty values on the version then hydrate and immediately render on the client. Which is something we do quite often with dom measurements. Another idea ive been having is the replication of `componentDidMount`. Occasionally we'll have a situation like this.
```
useEffect(() => {
  console.log(someDep)
// eslint-disable-next-line react/exhaustive-deps
}, [])
``` 

Which could be achieved with
```
const hasBeenMounted = useHasBeenMounted()
if (!hasBeenMounted) console.log('componentDidMount')
```

It also wouldn't need cleanup in the effect, since it wouldn't execute, since we're settings state in an unmounting component. Which is similarly the main purpose of `isMountedRef` to preventing `setState` in unmounted components (aka probably memory leaks)

src/components/RenderAfterMount/RenderAfterMount.tsx
AndrewMusgrave: > What do you think to accepting a fallback prop (defaulting to null) (like Suspense to allow you to define initial content.

I like the idea 👍 This is for rendering content after mount, so on the client. For example, cases when we have to use `useLayoutEffect` instead of `useEffect`.

src/utilities/use-mounted.ts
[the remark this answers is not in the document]
BPScott: So option 2 :) Some name that makes that clearer would be handy as on inital glance  `useMounted` would imply it changes if the component is mounted or not

More name riffing -  What do you think to `isAfterInitialMount` and `<AfterInitialMount>`? it says it'll only change the once and it keeps the present tense isThing convention (that wasEverMounted and hasBeenMounted lack) and also you can keep a similar name with the helper component:

src/components/AfterInitialMount/AfterInitialMount.tsx
```diff
@@ -1,12 +1,14 @@
...
+  fallback?: ReactNode;
 }
 
-export function AfterInitialMount({children}: Props) {
+export function AfterInitialMount({children, fallback = null}: Props) {
   const isMounted = useIsAfterInitialMount();
+  const content = isMounted ? children : fallback;
 
-  return isMounted ? children : null;
+  return <React.Fragment>{content}</React.Fragment>;
```
AndrewMusgrave: Typescript was having a hard time understanding the return type of the function was a react element so I wrapped it in a fragment rather than casting

src/components/AfterInitialMount/AfterInitialMount.tsx
```diff
@@ -0,0 +1,18 @@
...
+}
+
+function AfterInitialMount({children, fallback}: Props) {
+  const isMounted = useIsAfterInitialMount();
+  const content = isMounted ? children : fallback;
+
+  return <React.Fragment>{content}</React.Fragment>;
+}
+
+AfterInitialMount.defaultProps = {fallback: null};
```
AndrewMusgrave: Adding default props to make this more testable, since we can't replicate a node env right now we won't be able to assert that null is rendered as a default for fallback however using default props we can assert that the default fallback is null.

BPScott: https://github.com/reactjs/rfcs/pull/107 suggests that defaultProps on functional components will be deprecated at some future point.

I'd prefer we do `function AfterInitialMount({children, fallback= null}: Props) {`

src/components/AfterInitialMount/AfterInitialMount.tsx
BPScott: Had a play with this and while `return {content}` works in react the types are grumpy and casting doesn't seem to make it happier :(

It's annoying but i think this will have to stay for a while

src/components/AfterInitialMount/AfterInitialMount.tsx
BPScott: Chat outcome: doing the js default makes it harder to test but as this is such a trivial thing I don't think much is lost by having it be untestable. Lets use the js default and remove this (and then have the export on the same line as the function declaration)

PR #2152, opened 2019-09-17
[FooterHelp] Delete period from example

selenehinkley:
Screen Shot 2019-09-17 at 10 12 10 AM

This is to align with our content standard in Grammar and mechanics.

Grammar_and_mechanics_-_Shopify_Polaris

FYI @sadiesaurus @VanjaMitrovic

PR #2161, opened 2019-09-18
Update changelog

selenehinkley:
Update changelog with: Fixed FooterHelp component example by deleting a period ([2152](https://github.com/Shopify/polaris-react/pull/2152))

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -17,6 +17,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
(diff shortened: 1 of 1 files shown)


UNRELEASED.md
```diff
@@ -17,6 +17,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- Fixed FooterHelp component example by deleting a period ([2152](https://github.com/Shopify/polaris-react/pull/2152))
```
BPScott: This isn't code related so it's not a bug fix, can you put it in the documentation section.

Also the convention is to include a # before . the issue number (so `#2152`)

selenehinkley:
Skipped changelog entry after chat with @BPScott about the process :)

PR #2166, opened 2019-09-18
Add useLazyRef hook

AndrewMusgrave:

@beefchimi and I added this while working on #2105, I broke this out of the other PR so we can take our time working on 2105 and use this hook in the meantime.

It'll allow for lazy initialization of refs, see jsdocs for a little more information

* Adding hook + jsdoc
* Adding tests

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -45,6 +45,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Code quality
 
+- Added `useLazyRef` hook to use while building components ([#2166](https://github.com/Shopify/polaris-react/pull/2166))
 - Migrated `FilterCreator` to use hooks instead of withAppProvider ([#2156](https://github.com/Shopify/polaris-react/pull/2156))
 - Created a custom error for lack of context providers ([#2136](https://github.com/Shopify/polaris-react/pull/2136))
 - Migrated `ContextualSaveBar` to use hooks instead of `withAppProvider` ([#2091](https://github.com/Shopify/polaris-react/pull/2091))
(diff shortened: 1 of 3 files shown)


BPScott:
Looks solid.

I note that quilt has a react-hooks package that contains a similar implementation: https://github.com/Shopify/quilt/blob/master/packages/react-hooks/src/hooks/lazy-ref.ts

At some point in the future if we end up adding lots of these general-purpose hooks it might be worth thinking about pushing them into quilt.

AndrewMusgrave:
> Looks solid.
>
> I note that quilt has a react-hooks package that contains a similar implementation: https://github.com/Shopify/quilt/blob/master/packages/react-hooks/src/hooks/lazy-ref.ts
>
> At some point in the future if we end up adding lots of these general-purpose hooks it might be > worth thinking about pushing them into quilt. I can see useIsAfterIntitalMount being a useful one > all over the place.

It looks like they have a similar use [mounted](https://github.com/Shopify/quilt/blob/master/packages/react-hooks/src/hooks/mounted-ref.ts) hook as well. It might be beneficial to share our code so we don't duplicate it. Other libraries like online store ui could make use of them as well, what do you think @beefchimi

beefchimi:
@AndrewMusgrave definitely in favour of sharing more hooks!

BPScott:
> It looks like they have a similar use mounted hook as well.

Their hook has cleanup so it toggles if the current component is mounted or not, while I believe yours would set a value after initial mount then never change so they're similar but subtly different

AndrewMusgrave:
Yup the two mounting hooks are fundamentally different, with different uses. However I added a similar one [here](https://github.com/Shopify/polaris-react/blob/4e43848c91ebee722b5e4eb631df9469166ec82b/src/utilities/use-is-mounted-ref.ts), they're almost identically however the quilt version doesn't initialize a false value, which might be an error.

PR #2167, opened 2019-09-18
[FilterValueSelector] Convert to functional component & add isMountedRef

AndrewMusgrave:

Part of #1995

* Converting FilterValueSelector to a functional component
* Add useIsMountedRef hook

### How to 🎩

Playground with filters below

<details>

```jsx
import React from 'react';
import {ResourceList, TextStyle, Avatar, FilterType, Card} from '../src';

export class Playground extends React.Component {
  state = {
    searchValue: '',
    appliedFilters: [
      {
        key: 'accountStatusFilter',
        value: 'Account enabled',
      },
    ],
  };

    this.setState({searchValue});
  };

  };

  renderItem = (item) => {

    return (
        <h3>
        </h3>
        <div>{location}</div>
      </ResourceList.Item>
    );
  };

  render() {
    const resourceName = {
      singular: 'customer',
      plural: 'customers',
    };

    const items = [
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
      },
    ];

    const filters = [
      {
        key: 'orderCountFilter',
        label: 'Number of orders',
        operatorText: 'is greater than',
        type: FilterType.TextField,
      },
      {
        key: 'accountStatusFilter',
        label: 'Account status',
        operatorText: 'is',
        type: FilterType.Select,
        options: ['Enabled', 'Invited', 'Not invited', 'Declined'],
      },
    ];

    const filterControl = (
      <ResourceList.FilterControl
        filters={filters}
        additionalAction={{
          content: 'Save',
        }}
      />
    );

    return (
      <Card>
        <ResourceList
          resourceName={resourceName}
          items={items}
          renderItem={this.renderItem}
          filterControl={filterControl}
        />
      </Card>
    );
  }
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/ResourceList/components/FilterControl/components/FilterValueSelector/FilterValueSelector.tsx
+++ b/src/components/ResourceList/components/FilterControl/components/FilterValueSelector/FilterValueSelector.tsx
@@ -1,13 +1,11 @@
-import React from 'react';
+import React, {useCallback} from 'react';
 import {Select} from '../../../../../Select';
 import {Stack} from '../../../../../Stack';
 import {TextField} from '../../../../../TextField';
-import {
-  withAppProvider,
-  WithAppProviderProps,
-} from '../../../../../../utilities/with-app-provider';
 import {DateSelector} from '../DateSelector';
 import {Filter, AppliedFilter, FilterType, Operator} from '../../types';
+import {useI18n} from '../../../../../../utilities/i18n';
+import {useIsMountedRef} from '../../../../../../utilities/use-is-mounted-ref';
 
 export interface FilterValueSelectorProps {
   filter: Filter;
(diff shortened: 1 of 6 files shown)


src/utilities/tests/use-is-mounted-ref.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  it('returns a false value after unmounting', () => {
+    const spy = jest.fn((_) => {});
+
+    function MockComponent() {
+      const isMounted = useIsMountedRef();
+
+      useEffect(() => {
+        return () => {
+          // eslint-disable-next-line react-hooks/exhaustive-deps
```
AndrewMusgrave: dom nodes need to have their references saved since react could have gotten rid of them before the cleanup happens but this is 👌

tmlayton:
Sorry for the late notice on this since a lot of the work has been done, but the `FilterControl` is being removed in v5 in lieu of the `Filters` component.

tmlayton:
Specifically it was removed in https://github.com/Shopify/polaris-react/pull/2047

src/utilities/use-is-mounted-ref.ts
```diff
@@ -0,0 +1,19 @@
+import {useRef, useEffect, MutableRefObject} from 'react';
+
+/**
+ * Returns a MutatableRefObject containing a boolean value that
+ * represents a components mounted status.
+ * @returns MutableRefObject<boolean> The mounted status
+ */
+export function useIsMountedRef(): MutableRefObject<boolean> {
+  const isMounted = useRef(false);
```
BPScott: Should this be true so the hook returns true on a component's first render?

src/utilities/tests/use-is-mounted-ref.test.tsx
BPScott: can we add a comment inline that says what's being excluded and why?

AndrewMusgrave:
@tmlayton No worries 😄 It was a fairly fast refactor and we still might as well 🚢 this incase we can remove `withAppProvider` before `v5`

src/utilities/use-is-mounted-ref.ts
AndrewMusgrave: The component isn't mounted during the first render, however, it'll be mounted when the effects run? afaik

AndrewMusgrave:
The missing `codecov` is from `FiltersControl` so I'm going to leave it be since it's being removed.

src/utilities/use-is-mounted-ref.ts
BPScott: From http://projects.wojtekmaj.pl/react-lifecycle-methods-diagram/ and https://github.com/donavon/hook-flow it looks like you're right - I guess those "­React updates ­D­O­M " phases are when something goes from not being mounted to is mounted so in that render before first mount a component is not mounted. I was thrown because the equivalent hook in quilt sets this to true, but it seems like that's a bug on their side: https://github.com/Shopify/quilt/blob/master/packages/react-hooks/src/hooks/mounted-ref.ts

dleroux: I had mentioned this to @michenly and it sounds like this is the intention. From slack: `I think you should think of this line more like a constructor of the class and the value will remain true till componentWillUnmount... it is not equal to componentDidMount... I wonder if it should be renamed to useUnMountedRef`,

michenly: Typically where we are using `useMountedRef` from quilt are for checking if something had become un-mounted, so even thought the implementation is not exactly accurate it had been working so far.

I would like to fix it as a bug, but wonder if we would run into any situation where going from false -> true - > false will introduce a flash in UI. So this might need to be a breaking change with a more clear new name to avoid confusion.

src/components/ResourceList/components/FilterControl/components/FilterValueSelector/FilterValueSelector.tsx
```diff
@@ -17,110 +15,94 @@ export interface FilterValueSelectorProps {
...
+      if (!value) {
+        return;
+      }
+
+      onChange(value);
+    },
+    [onChange, onFilterKeyChange, value],
+  );
+
+  if (showOperatorOptions && operatorText!.length !== 0 && !isMounted) {
```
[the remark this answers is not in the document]
AndrewMusgrave: 🤦‍♂ I should create a linting rule for this. I'm guessing this'll come up every now and then

This block will execute during the first render, not after. The different between `useIsMountedRef` & `useIsAfterInitialMount` is that `useIsAfterInitialMount` is going to cause a re-render and `useIsMountedRef` will not.

PR #2170, opened 2019-09-19
[Connected] Preventing loss of TextField focus caused by shifting DOM position

arapehl:

Fixes: https://github.com/Shopify/polaris-react/issues/2169
Fixes: https://github.com/Shopify/online-store-web/issues/4181

Ensuring that the `Connected` component always returns fragments at the same DOM depth so as not to cause a `TextField`'s `input` field to unmount and lose focus.

<details>
  <summary>Introducing a "Search" button (`connectedRight`) while typing causes a loss of focus</summary>
  
</details>

<details>
  <summary>Ensuring maintained focus by always returning a TextField at the same depth in the DOM (what this PR does)</summary>
  
</details>

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/TextField/tests/TextField.test.tsx
+++ b/src/components/TextField/tests/TextField.test.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {ReactElement} from 'react';
 import {mountWithAppProvider, findByTestID} from 'test-utilities/legacy';
 import {InlineError, Labelled, Connected, Select} from 'components';
 import {mountWithApp} from 'test-utilities';
@@ -1026,33 +1026,17 @@ describe('<TextField />', () => {
       );
       const connectedChild = textField
         .find(Connected)
-        .find('div')
-        .first();
+        .prop('children') as ReactElement;
 
-      connectedChild.simulate('click');
+      expect(textField.getDOMNode().querySelector('input')).not.toBe(
+        document.activeElement,
+      );
+
+      connectedChild.props.onClick({});
 
       expect(textField.getDOMNode().querySelector('input')).toBe(
         document.activeElement,
       );
-    });
-
-    it('applies focus variant style `onFocus`', () => {
-      const textField = mountWithAppProvider(
-        <TextField label="TextField" onChange={noop} />,
-      );
-      let connectedInteriorWrapper = textField
-        .find(Connected)
-        .find('div')
-        .first();
-
-      connectedInteriorWrapper.simulate('focus');
-
-      connectedInteriorWrapper = textField
-        .find(Connected)
-        .find('div')
-        .first();
-
-      expect(connectedInteriorWrapper.hasClass('focus')).toBe(true);
     });
   });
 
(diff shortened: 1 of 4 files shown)


src/components/TextField/tests/TextField.test.tsx
```diff
@@ -1042,15 +1039,13 @@ describe('<TextField />', () => {
...
 
       connectedInteriorWrapper.simulate('focus');
 
       connectedInteriorWrapper = textField
         .find(Connected)
-        .find('div')
-        .first();
+        .find('div.TextField');
 
       expect(connectedInteriorWrapper.hasClass('focus')).toBe(true);
```
dleroux: Same as in web we follow their best practice and avoid using classNames in tests. [see here](https://github.com/Shopify/web-foundation/blob/master/Best%20practices/React/Testing.md#implementation-details). `document.activeElement` would be more reliable here.

arapehl: The test is called `applies focus variant style onFocus`. Its whole purpose is to test for the `focus` class name. I was just getting it to work given the change in DOM output caused by this PR. The previous test called `sets focus to the <input /> onClick` checks for `document.activeElement` albeit from a different trigger.

I can ditch the whole test, if you'd like.

danrosenthal: Dan's guidance on asserting `document.activeElement` is probably a better approach here.

src/components/TextField/tests/TextField.test.tsx
```diff
@@ -1040,19 +1040,18 @@ describe('<TextField />', () => {
       const textField = mountWithAppProvider(
```
asalisbury: What is a "focus variant style"?

src/components/TextField/tests/TextField.test.tsx
```diff
@@ -1040,19 +1040,18 @@ describe('<TextField />', () => {
...
-      connectedInteriorWrapper = textField
+      const connectedChild = textField
         .find(Connected)
-        .find('div')
-        .first();
+        .prop('children') as ReactElement;
 
-      expect(connectedInteriorWrapper.hasClass('focus')).toBe(true);
+      act(() => {
+        connectedChild.props.onFocus({});
```
asalisbury: What does this do? I think the assertion passes even if you 🔥 it.

src/components/TextField/tests/TextField.test.tsx
```diff
@@ -1026,10 +1027,9 @@ describe('<TextField />', () => {
       );
       const connectedChild = textField
         .find(Connected)
-        .find('div')
-        .first();
+        .prop('children') as ReactElement;
```
asalisbury: Could we add the inverse assertion above the click? 

```tsx
expect(textField.getDOMNode().querySelector('input')).not.toStrictEqual(
  document.activeElement!,
);
```

arapehl:
So, I ended up removing the `onFocus` test because it doesn't actually test anything apart from setting a `focus` class name. I also added an additional assertion to the `onClick` test to makes sure that it wasn't originally the active element and that the `onClick` caused it to be.

asalisbury:
:shipit:

PR #2172, opened 2019-09-19
Add CODEOWNERS file to notify them of any public Sass API changes

kaelig:
The public Sass API is set to be gradually deprecated.

@tmlayton @BPScott and @danrosenthal should be aware of any changes made to it
to prevent any new helpers from sneaking in.

If anyone feels like they should be added to this list, I'll be happy to add them.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,6 +4,5 @@
 # to prevent any new helpers from sneaking in.
 /src/styles/foundation @BPScott @danrosenthal @tmlayton
 /src/styles/shared @BPScott @danrosenthal @tmlayton
-/src/styles/_common.scss @BPScott @danrosenthal @tmlayton
 /src/styles/foundation.scss @BPScott @danrosenthal @tmlayton
 /src/styles/shared.scss @BPScott @danrosenthal @tmlayton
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,9 @@
+# The public Sass API will be gradually deprecated
+#
+# These people should be aware of any changes made to it
+# to prevent any new helpers from sneaking in.
+/src/styles/foundation @BPScott @danrosenthal @tmlayton
+/src/styles/shared @BPScott @danrosenthal @tmlayton
+/src/styles/_common.scss @BPScott @danrosenthal @tmlayton
```
BPScott: /src/styles/_common.scss is safe as it is not exposed publicly, so we can remove this line.

We now ship pre-compiled sass for our component css which means the only publicly available mixins/functions/variables are those exposed in foundation.scss and shared.scss - which then import things from files in the shared and foundation folders. Check the styles folder that is generated when you do a build.

For context: for internal usage Dan is going to do a PR that copies the current contents of the shared and foundations folders into a new `common` folder which `_common.scss` will then import. That will mean we're free to itterate in private in the common folder without breaking public usage while effectively freezing the foundation and shared files in place for public usage but we don't use them internally.

kaelig: Thank you, that makes sense!

PR #2176, opened 2019-09-20
[Spinner] Fix not being read by screen readers

### LINKED ISSUE #791: [a11y] Spinner is not read by screen readers
## Issue summary

The [Spinner](https://polaris.shopify.com/components/feedback-indicators/spinner#navigation) has some implementation issues which may cause issues for screen reader users. The `<svg>` element is skipped by VoiceOver on iOS and macOS and NVDA with or without the `aria-label` attribute. There are a couple things that contribute to this:
- There is a bug with how `<svg aria-label="...">` is identified by VO. If the `<svg>` is a child of a focusable element, it will read. Otherwise, it won't. Not even `role="img"` can save it, it seems. (shrug emoji)
- `role="status"` functions as a live region, which requires HTML content to be dynamically updated. Since the `<svg>` has no content (and content inside SVGs is not conveyed consistently anyway), no update can be conveyed.

[There's a bit more info on SVG weirdness here](https://simplyaccessible.com/article/7-solutions-svgs/), and [an MDN entry on `status` that might help](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_status_role). (Full disclosure, I tested and wrote most of the findings for the SVG article. 😇)

#### Spinner output with `accessibilityLabel` value added (HTML)

```
<svg viewBox="0 0 44 44" class="Polaris-Spinner Polaris-Spinner--colorTeal Polaris-Spinner--sizeLarge" aria-label="Please wait" role="status">
<!-- ... -->
</svg>
```

## Steps to reproduce

1. Navigate to [Spinner](https://polaris.shopify.com/components/feedback-indicators/spinner#navigation) using Safari with VoiceOver or NVDA with Firefox.
1. Use the screen reader to read the spinner.
1. Give the spinner an accessible label (name) and reread it.

### Expected

- The spinner will be conveyed programmatically.

### Actual

- The spinner is not conveyed programmatically to screen readers.

## Recommendations

Since SVGs are conveyed inconsistently, I'd recommend hiding the SVG itself and using visually hidden text to convey its purpose instead.

- Use `aria-hidden="true"` to hide the SVG itself and remove the `aria-label`. (Add `focusable="false"` just in case the SVG is implemented inside a focusable element/area to avoid bugs in some browsers.)
- Create a visually hidden `<span>` or `<div>` to serve as the text container.

Additional steps will depend on how the spinner should be conveyed, which may depend on the context. This might be a good opportunity for documentation instead of implementation.

1. To make the spinner read dynamically when it appears, without moving focus, additionally use `role="status"`:
  - Load the `<span role="status" class="Polaris-VisuallyHidden">` in the DOM by default.
  - Append/update the `accessibleLabel` when the spinner is displayed. This will trigger reading dynamically.
2. Alternately, if the value doesn't need to be read dynamically and should just be accessed inline, or focus is sent to it for reading, the text can just be visually hidden and doesn't need to be dynamically added.

### Recommended spinner output with `accessibleLabel` valued added, live region version (HTML)

```
<svg aria-hidden="true" focusable="false" viewBox="0 0 44 44" class="Polaris-Spinner Polaris-Spinner--colorTeal Polaris-Spinner--sizeLarge"><!-- ... --></svg>
<span role="status" class="Polaris-VisuallyHidden">Please wait</span>
```

### Recommended spinner output with `accessibleLabel` valued added, "static" or focused version (HTML)

```
<div><!-- send focus to me, for example -->
<svg aria-hidden="true" focusable="false" viewBox="0 0 44 44" class="Polaris-Spinner Polaris-Spinner--colorTeal Polaris-Spinner--sizeLarge"><!-- ... --></svg>
<span class="Polaris-VisuallyHidden">Please wait</span>
</div>
```

## Specifications

- **Component:** Spinner
- **WCAG SCs:** [4.1.3 Status Messages (Level AA)](https://www.w3.org/TR/WCAG21/#status-messages)
- **Testing done with:**
    - VoiceOver/Safari/macOS
    - VoiceOver/Safari/iOS
    - NVDA/Firefox/Windows 10

AndrewMusgrave:

Fixes #791

### Styleguide Screenies

Screen Shot 2019-09-30 at 2 45 18 PM
Screen Shot 2019-09-30 at 2 45 27 PM

* Adding `hasFocusableParent` to spinners public API
* Add a focus management example
* Add tests
* Add accessibility docs section

## 

#### With label

<details>

```jsx
import React from 'react';
import {Page, Spinner} from '../src';

  return (
    <Page title="Playground">
      <Spinner accessibilityLabel="label here" />
    </Page>
  );
}
```

</details>

#### With focusable parent

<details>

```jsx
import React from 'react';
import {Page, Spinner} from '../src';

  return (
    <Page title="Playground">
      <div tabIndex={0}>
        <Spinner hasFocusableParent accessibilityLabel="label here" />
      </div>
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Spinner/README.md
+++ b/src/components/Spinner/README.md
@@ -28,7 +28,7 @@ Spinners are used to notify merchants that their action is being processed. For 
 Use to notify merchants that their requested action is being processed.
 
 ```jsx
-<Spinner size="large" color="teal" />
+<Spinner accessibilityLabel="Spinner example" size="large" color="teal" />
 ```
 
 <!-- content-for: android -->
@@ -50,8 +50,92 @@ Use to notify merchants that their requested action is being processed.
 Smaller than the default spinner.
 
 ```jsx
-<Spinner size="small" color="teal" />
+<Spinner accessibilityLabel="Small spinner example" size="small" color="teal" />
 ```
+
+### Spinner with focus management
+
+Use to direct the focus state from the control to the spinner, to the content.
+
+```jsx
+function SpinnerWithFocusManagement() {
+  const tabs = useRef([
+    {
+      id: 'all-customers',
+      content: 'All',
+      accessibilityLabel: 'All customers',
+      panelID: 'all-customers-content',
+    },
+    {
+      id: 'accepts-marketing',
+      content: 'Accepts marketing',
+      panelID: 'accepts-marketing-content',
+    },
+  ]);
+
+  const [selected, setSelected] = useState(0);
+  const [loading, setLoading] = useState(false);
+  const [value, setValue] = useState('');
+  const [textFieldFocused, setTextFieldFocused] = useState(false);
+
+  useEffect(() => {
+    setTextFieldFocused(!loading);
+  }, [loading]);
+
+  const handleTabChange = useCallback((selectedTab) => {
+    setLoading(true);
+    setSelected(selectedTab);
+    setTimeout(() => {
+      setValue('');
+      return setLoading(false);
+    }, 1500);
+  }, []);
+
+  const handleUrlChange = useCallback((value) => setValue(value), []);
+
+  const handleSubmit = useCallback((_event) => setValue(''), []);
+
+  const label = selected ? 'Marketing' : 'Customers';
+  const sectionMarkup = loading ? (
+    <Spinner
+      accessibilityLabel="Loading form field"
+      hasFocusableParent={false}
+    />
+  ) : (
+    <Form noValidate onSubmit={handleSubmit}>
+      <FormLayout>
+        <TextField
+          value={value}
+          focused={textFieldFocused}
+          onChange={handleUrlChange}
+          label={label}
+        />
+        <Button submit>Submit</Button>
+      </FormLayout>
+    </Form>
+  );
+
+  return (
+    <Card>
+      <Tabs tabs={tabs.current} selected={selected} onSelect={handleTabChange}>
+        <Card.Section title={tabs.current[selected].content}>
+          {sectionMarkup}
+        </Card.Section>
+      </Tabs>
+    </Card>
+  );
+}
+```
+
+---
+
+## Accessibility
+
+<!-- content-for: web -->
+
+SVGs are often conveyed inconsistently to assistive technologies. The `Spinner` component's accessibility is also highly contextual. When the parent component is focusable, you'll need to set the `hasFocusableParent` prop for the appropriate `role` attribute to be applied.
+
+<!-- /content-for-->
 
 ---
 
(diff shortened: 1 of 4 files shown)


src/components/Spinner/Spinner.tsx
```diff
@@ -24,14 +25,22 @@ export interface SpinnerProps {
...
 }
 
 export function Spinner({
   size = 'large',
   color = 'teal',
   accessibilityLabel,
+  hasFocusableParent,
 }: SpinnerProps) {
   const intl = useI18n();
+  const [hasMounted, setHasMounted] = useState(false);
```
AndrewMusgrave: Replace with hook when it's merged

AndrewMusgrave:
@dpersing What are you thoughts on this so far? I'm having trouble testing this on my VM (storybook freezes and crashes virtual box 😓) Seems great on safari!

dpersing:
@AndrewMusgrave Will take a look first thing tomorrow! 💜

src/components/Spinner/README.md
```diff
@@ -55,6 +55,16 @@ Smaller than the default spinner.
 
 ---
 
+## Accessibility
+
+<!-- content-for: web -->
+
+SVGs after often displayed inconsistently. The `Spinner` components accessibility is highly contextual. When the parent component is focusable you'll need to set the `hasFocusableParent` prop for the appropriate `role` attribute to be applied.
```
kaelig:
```suggestion
SVGs after often displayed inconsistently. The `Spinner` components accessibility is highly contextual. When the parent component is focusable you’ll need to set the `hasFocusableParent` prop for the appropriate `role` attribute to be applied.
```

dpersing:
@AndrewMusgrave This is looking good! I tested with VoiceOver/Safari on macOS and iOS, as well as TalkBack/Chrome on Android, and NVDA/Firefox and Narrator/Edge (just for kicks) on Windows 10.

I really like the inclusion of discussing how the component is very contextual. I wonder if it would make sense to update our examples to show some common examples?

src/components/Spinner/README.md
dpersing: @kaelig I think this is a typo.

@AndrewMusgrave What about something like:

```
SVGs are often conveyed inconsistently to assistive technologies. The `Spinner` component's accessibility is also highly contextual. When the parent component is focusable, you'll need to set the `hasFocusableParent` prop for the appropriate `role` attribute to be applied.
```

We can also include this in an accessibility section in the doc, to also provide some context for `role="status"`.

AndrewMusgrave: Yup it was just a typo 😄

AndrewMusgrave:
@dpersing I'll create an issue to follow up with it 😄

dpersing:
@AndrewMusgrave I think we should be good to merge this once the file conflict is addressed!

Issue #2182, opened 2019-09-23
Compiler warning "start value has mixed support"


mbaumbach:
## Issue summary

When building a project using a vanilla `create-react-app` on Polaris 4.3.0, you get the following warning:

```
./node_modules/@shopify/polaris/styles.css (./node_modules/css-loader/dist/cjs.js??ref--6-oneOf-3-1!./node_modules/postcss-loader/src??postcss!./node_modules/@shopify/polaris/styles.css)
Warning

(1436:3) start value has mixed support, consider using flex-start instead
```

## Expected behavior

When building/compiling a project with Polaris there are no warnings.

## Actual behavior

As mentioned above, you get the compiler warning when using the `Card` component. It appears to be due to changes in src/components/Card/Card.scss in Polaris 4.3.0: https://github.com/Shopify/polaris-react/commit/a32e01327c844bbdc1a70599845b8f37db16f59a#diff-719bbc36ca2215599a2260f3e8380fd6R86

The `start` property has mixed support on the `justify-content` attribute, though it does function slightly differently than `flex-start`, so it may be intentional.

## Steps to reproduce the problem

1. Start a `create-react-app` app. (Using the latest create-react-app: 3.1.2)
1. Add Polaris 4.3.0 to the dependencies.
1. Add the Polaris stylesheet at the top of the `index.js` file: `import '@shopify/polaris/styles.css';`
1. Add a `Card` somewhere in the component hierarchy.

## Specifications

- Are you using the React components? (Y/N): Y
- Polaris version number: 4.3.0
- Browser: Chrome 77.0.3865.90 (Official Build) (64-bit)
- Device: MacBook Pro
- Operating System: macOS High Sierra 10.14.6

dleroux:
Thanks for reporting this @mbaumbach.

mbaumbach:
Confirmed this is fixed in Polaris 4.4.0. Feel free to close and thanks for the fix!

PR #2185, opened 2019-09-24
[Card] Improve standards compliance via flex-start

GoodForOneFare:

PostCSS throws warnings about this, which blocks sewing-kit compilation:

```
ModuleWarning: Module Warning (from ./node_modules/postcss-loader/src/index.js):
Warning

(1:1639) start value has mixed support, consider using flex-start instead
    at Object.emitWarning (node_modules/@shopify/sewing-kit/node_modules/webpack/lib/NormalModule.js:158:6)
    at result.warnings.forEach (node_modules/postcss-loader/src/index.js:146:16)
    at Array.forEach (<anonymous>)
    at postcss.process.then (node_modules/postcss-loader/src/index.js:145:27)
```

### DIFF
--- a/src/components/Card/Card.scss
+++ b/src/components/Card/Card.scss
@@ -83,7 +83,7 @@
 
 .Footer {
   display: flex;
-  justify-content: start;
+  justify-content: flex-start;
   padding: 0 spacing() spacing();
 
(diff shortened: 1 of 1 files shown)


mbaumbach:
Addresses #2182

PR #2186, opened 2019-09-24
[ColorPicker] Add hex/alpha text fields

ruairiphackett:

Fixes https://github.com/Shopify/polaris-ux/issues/276

- [x] Add labels
- [x] Add i18n for error messages
- [x] Clean up AlphaField state
- [x] Add ability to enter english colour names into the 'hex' field
- [x] Add tests
- [ ] Request translations once strings are reviewed

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist


tleunen:
- Should the hex color value contain the opacity? (becoming a 8 digits hex instead of 6 when the alpha is enabled)
- Should the `onChange` callback return the hex value alongside its params? Or should the consumer find out a way to convert the hue, brithness, saturation and alpha into an rgba/hex? In most cases, I'd guess the consumer just wants a color value?

src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -112,6 +141,15 @@ export class ColorPicker extends React.PureComponent<ColorPickerProps, State> {
...
   };
 
+  private handleHexChange = (hex: string) => {
+    const {
+      color: {alpha = 1},
+      onChange,
+    } = this.props;
+    const newColor = hexToHsb(hex);
+    onChange({...newColor, alpha});
+  };
```
AndrewMusgrave: Should we prevent invalid hex codes from being added, or maybe our error message could include why it's invalid? What do you think?

src/components/ColorPicker/components/AlphaField/AlphaField.tsx
```diff
@@ -0,0 +1,102 @@
...
+      }
+    }
+  };
+
+  private handleTextChange = (value: string) => {
+    this.setState({percentage: parseInt(value, 10)});
+  };
+}
+
+// Use named export once withAppProvider is refactored away
```
AndrewMusgrave: We're in the process of removing all class components / `withAppProvider`. Could we convert this to a functional component and use the `useI18n` hook?

src/components/ColorPicker/components/AlphaField/AlphaField.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  private handleBlur = () => {
+    const {onChange, alpha} = this.props;
+    const {percentage} = this.state;
+
+    let roundedValue = Math.round(percentage);
+
+    if (roundedValue !== null) {
+      if (roundedValue > 100) {
+        roundedValue = 100;
```
AndrewMusgrave: It might be easier to do something like

```jsx
const roundedValue = clamp(Math.round(percentage), 1, 100);
```

Or something along those lines, what do you think?

src/components/ColorPicker/components/SwatchBackground/SwatchBackground.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export interface SwatchBackgroundProps {
+  color: HSBColor;
+}
+
+export function SwatchBackground(props: SwatchBackgroundProps) {
+  const {color} = props;
+
+  const style = {backgroundColor: hsbToString(color)};
+  return <div style={style} className={styles.SwatchBackground} />;
```
AndrewMusgrave: Just curious why this is a component rather than inline?

src/components/ColorPicker/components/TextPicker/TextPicker.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+    return (
+      <div className={className}>
+        <SwatchBackground color={color} />
+      </div>
+    );
+  };
+}
+
+// Use named export once withAppProvider is refactored away
```
AndrewMusgrave: Same comment as above 😄

src/components/ColorPicker/components/TextPicker/TextPicker.tsx
```diff
@@ -0,0 +1,163 @@
...
+export interface TextPickerProps {
+  color: HSBColor;
+  allowAlpha?: boolean;
+  onChange(hex: string): void;
+}
+
+type CombinedProps = TextPickerProps & WithAppProviderProps;
+
+interface State {
+  valueHasBeenUpdatedOnce: boolean;
```
AndrewMusgrave: We could probably have this as a class property rather than state unless we needed it to trigger a re-render?

AndrewMusgrave (commented):
Thanks for all the work you've put in, it's looking great, we Polaris really appreciates it 😄 

I've noticed your adding props onto state, it's an anti-pattern in react. Converting the components to a functional component should help avoid this 😄 

Adding some tests would be awesome 😄 🙏

src/utilities/color-validation.ts
```diff
@@ -9,3 +13,22 @@ export function isLight({red, green, blue}: RGBColor | RGBAColor): boolean {
...
+  }
+  return false;
+}
+
+export function isHexString(value: string) {
+  return /(^#[0-9A-F]{6}$)|(^#[0-9A-F]{3}$)/i.test(value);
+}
+
+export function isHashlessHex(value: string) {
+  return /(^[0-9A-F]{6}$)|(^[0-9A-F]{3}$)/i.test(value);
```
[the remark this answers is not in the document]
ruairiphackett: Nice!

ruairiphackett:
@AndrewMusgrave  @tleunen thanks a million for looking at this for me!

>Should the hex color value contain the opacity? (becoming a 8 digits hex instead of 6 when the alpha is enabled)

There were a few questions relating to 6/8 digit hex, so I thought I'd bring the discussion to one place.

We went with 6 digit hex to match what is currently used in both the rails and react implementations of the colour picker in the online store (specifically the one used in the theme editor).

Do you think it'd be better to use a 8 digit version?

@mirualves @dpersing Any thoughts on switching to 8 digit hex that would include the alpha value (representing RGBA, instead of the 6 digit hex that represents RGB) or should we keep the 2 separate as 6 digit hex and alpha percentage as they are in [the designs](https://www.figma.com/file/yBEALRv5joQRqvyieFxmAl/Color-picker?node-id=4%3A2) currently?

src/components/ColorPicker/ColorPicker.tsx
ruairiphackett: This was discussed over in the [Figma designs](https://www.figma.com/file/yBEALRv5joQRqvyieFxmAl/Color-picker?node-id=4%3A2). We had error messages implemented first, but then decided to change to prevent invalid hex values being entered. The check is made in [`handleBlur`](https://github.com/Shopify/polaris-react/pull/2186/files#diff-b3d0546d98cde51214257201a3fb16adR104-R122) in the `TextPicker` component, by using the last valid value if an invalid value is entered.

dpersing:
> or should we keep the 2 separate as 6 digit hex and alpha percentage as they are in the designs currently?

I'm in favor of keeping them separate. I don't think we want to assume that most merchants will be comfortable enough with CSS to know how RGBA works. In this case I think we want to do the work of parsing what most merchants will understand, versus what would be most efficient.

@ruairiphackett @mirualves @AndrewMusgrave

package.json
```diff
@@ -192,6 +192,7 @@
     "@types/react-transition-group": "^2.0.7",
     "hoist-non-react-statics": "^3.3.0",
     "lodash": "^4.17.4",
+    "tinycolor2": "^1.4.1",
```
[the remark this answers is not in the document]
ruairiphackett: Good point, I'll remove it and add the [svg colour map](https://www.w3.org/TR/css-color-3/#svg-color) 👍

src/utilities/color-validation.ts
```diff
@@ -9,3 +13,22 @@ export function isLight({red, green, blue}: RGBColor | RGBAColor): boolean {
...
+}
+
+export function isRgbString(value: string) {
+  return IS_RGB_STRING_REGEX.test(value);
```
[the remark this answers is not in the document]
ruairiphackett: I had been using it elsewhere & removed the other uses, good catch.

src/utilities/color-transformers.ts
```diff
@@ -294,3 +305,34 @@ export function colorToHsla(color: string) {
...
+
+export function nameToHex(value: string) {
+  const hex = nameHexMap[normalizeValue(value)];
+  return expandHex(`#${hex}`);
+}
+
+export function rgbStringToHex(value: string) {
+  const rgb = normalizeValue(value).match(
+    RGB_STRING_TO_HEX_REGEX,
+  ) as RegExpMatchArray;
```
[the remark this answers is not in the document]
ruairiphackett: Hey @AndrewMusgrave I'm not sure what type to use here other than string, as we [use the function](https://github.com/Shopify/polaris-react/pull/2186/files/62d6e89075b848941a4bda8111e304f247cf76bd#diff-b3d0546d98cde51214257201a3fb16adR130) before we make any check if it is a valid colour entry.

Would you have any suggestions?

src/components/ColorPicker/components/SwatchBackground/SwatchBackground.tsx
ruairiphackett: I think it can be inlined looking back at it 👍

I'll make the necessary changes

AndrewMusgrave:
Sorry I didn't realize this was ready for review again, re-requesting myself so I'll get to it later 😄

AndrewMusgrave:
> or should we keep the 2 separate as 6 digit hex and alpha percentage as they are in the designs currently?

I'm in favor of keeping them separate. 8 digit hex codes aren't too common among developers from what I've seen so I image merchants without a strong understanding of CSS or color systems may be thrown off.

src/components/ColorPicker/ColorPicker.tsx
AndrewMusgrave: Interesting, so if a user enters `#1234567` they'll have to enter the code again rather than deleting the extra character?

src/utilities/color-transformers.ts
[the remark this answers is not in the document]
AndrewMusgrave: Sorry I didn't realize we were chaing on a match. It looks like the return type is ` RegExpMatchArray | null` for match. It seems valid since match can return null. Rather than casting should we return a default value in case a match was not found?

src/utilities/color-validation.ts
```diff
@@ -1,5 +1,9 @@
+import {names} from 'tinycolor2';
 import {RGBColor, RGBAColor} from './color-types';
 
+export const nameHexMap: Record<string, string> = names;
```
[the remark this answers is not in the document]
AndrewMusgrave: Is the idea that `online-store-web` will consume our color picker?

AndrewMusgrave:
Feel free to ping me when this is ready for a review again 😄

ruairiphackett:
@AndrewMusgrave It isn't quite ready for review yet, getting there 😄

I'll ping you when it's ready. I also want to add tests before the next one 👍

src/utilities/color-validation.ts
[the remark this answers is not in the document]
ruairiphackett: The idea of the change is to make the `ColorPicker` accessible by adding keyboard support. There's no plan for any particular team to consume the new version, but hopefully it'll be used in favour of the old one.

src/utilities/color-transformers.ts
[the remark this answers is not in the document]
ruairiphackett: I added a default return value here if we don't match, but let me know if you think this is the wrong approach. Apologies if I'm misunderstanding your advice on this one.

ruairiphackett:
@AndrewMusgrave This is ready for another review if you get a chance ❤️ 

I changed the 2 subcomponents to functional components & hopefully addressed all the other review comments. I also added tests, but let me know if they don't cover enough 👍

ruairiphackett:
@mirualves @dpersing 

Apologies for the delay in working on the `ColorPicker`, I've had other projects on the go at the same time, so it's been a bit slow going!

This PR is ready for its second code review (@AndrewMusgrave gave a first preliminary one last week).

It adds the text colour input field (`TextPicker`) and alpha input field (`AlphaField`):

https://screenshot.click/11-51-096ou-vcdp2.png

I implemented the same valid colour inputs as were used in the Online Store implementation, but was wondering if we could extend it to accept more colour types as input?

### Currently accepted colour types

3 & 6 digit hex codes:
- `#000000`
- `000000`
- `#000`
- `000`

`rgb` colours:
- `rgb(0,0,0)`

Colour names:
- `red`
- `light green`
- etc. (list here: https://github.com/Shopify/polaris-react/pull/2186/files?file-filters%5B%5D=#diff-a1588a58bf100c9d77410e0aa36c4c1d)

To illustrate its usage, here's a quick video:

https://screenshot.click/11-27-3tqli-mcdca.mp4

### Adding more colour types?

I was wondering if we should add to the valid colour types for the TextPicker, by adding support for:
- `hsl(0, 100%, 50%)`
- `hsb(0, 100%, 50%)`

Also, we could make the `TextPicker` accept colour types that set the alpha value of the colour by adding support for the following colour types:
- `rgba(0, 255, 124, 0.5)`
- `hsla(0, 100%, 50%, 0.5)`

Let me know if you have any thoughts on extending it like this or any other comments 👍

dpersing:
@ruairiphackett I'm thinking for this version to keep the current support, and perhaps split out the extended support into a separate PR for a later date. What do you think, @mirualves?

mirualves:
>  I'm thinking for this version to keep the current support, and perhaps split out the extended support into a separate PR for a later date.

By current support do you mean hex + RGB? I'll only make sure the input field accommodates the 3 digits for each RGB value, like `rgb(196,196,196)`

It really looks great, @ruairiphackett!

ruairiphackett:
>By current support do you mean hex + RGB? I'll only make sure the input field accommodates the 3 digits for each RGB value, like `rgb(196,196,196)`

@mirualves any `rgb` value will currently work, so `rgb(196,196,196)` currently works 👍 

>I'm thinking for this version to keep the current support, and perhaps split out the extended support into a separate PR for a later date.

@dpersing I'll create an issue to track this and the extended support can either be incorporated here or in a later PR depending on what is decided.

mirualves:
> any rgb value will currently work, so rgb(196,196,196) currently works

Awesome! To be clear, I'm referring to the characters inside the input field, without overflowing it:

color-picker

ruairiphackett:
@mirualves Apologies, I misunderstood! That's true, when the transparent value is present the text does overflow when entering most `rgb` values: https://screenshot.click/21-13-o51af-yiwrn.mp4

That behaviour isn't ideal 💯 

Could this be affected by the font type/size being used also? If we make the fix above might we run into the same issue in certain instances anyway? Adding spaces (e.g. `rgb(255, 255, 255)`) could also affect things.

mirualves:
> Could this be affected by the font type/size being used also? 
We can fix that by increasing the width of the entire component. By making it a bit wider, it will accommodate the RGB values, no problem.
I think the font size is fine 👍

ruairiphackett:
>By making it a bit wider, it will accommodate the RGB values, no problem.

Cool. I wonder if we should try to maintain a square shape for the main colour picker, i.e. width equal to height for this component?

mirualves:
> Cool. I wonder if we should try to maintain a square shape for the main colour picker, i.e. width equal to height for this component?

I don't have strong option for it. It seems to work well, regardless if the shape is more squared or not. One argument for keeping the high shorter is to save screen space, but I think we can test both options in context and see how it feels.

See more rectangular examples here: http://casesandberg.github.io/react-color/

locales/en.json
```diff
@@ -28,6 +28,11 @@
       "spinnerAccessibilityLabel": "Loading"
     },
 
+    "ColorPicker": {
+      "textPickerAccessibilityLabel": "HEX color",
+      "alphaFieldAccessibilityLabel": "Alpha"
```
ruairiphackett: @dpersing Any suggestions on what a11y labels should we use here?

The first is used for the hex/colour name/rgb input field and the second is used for the alpha percentage field.

ruairiphackett: I added these as placeholders for now.

ruairiphackett:
@AndrewMusgrave I added a change to the component width when the alpha picker is present. This is to allow a full `rgb` string to be entered into the text field without the text scrolling out of view:



**[Change](https://www.figma.com/file/yBEALRv5joQRqvyieFxmAl/Color-picker?node-id=300%3A572):**

src/utilities/color-validation.ts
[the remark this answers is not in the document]
AndrewMusgrave: I may under an incorrect assumption but I suspect the majority of people don't know color keywords, apart from the basic; red, green, purple. Not the rebeccapurple, forestgreen, papayawhip, etc. Is this someing being requested accessibility wise? @dpersing 

I'm not saying we shouldn't have it, just curious about the context before we increase our bundle 😄

src/utilities/color-transformers.ts
```diff
@@ -10,6 +11,8 @@ import {
 } from './color-types';
 import {compose} from './compose';
 
+const RGB_STRING_TO_HEX_REGEX = /^rgba?[\s+]?\([\s+]?(\d+)[\s+]?,[\s+]?(\d+)[\s+]?,[\s+]?(\d+)[\s+]?/i;
```
AndrewMusgrave: Same issue as the other regex

src/utilities/color-transformers.ts
```diff
@@ -314,3 +325,38 @@ export function colorToHsla(color: string): HSLAColor {
...
+}
+
+export function nameToHex(value: string) {
+  const hex = nameHexMap[normalizeColorString(value)];
+  return expandHex(`#${hex}`);
+}
+
+export function rgbStringToHex(value: string) {
+  const rgb = normalizeColorString(value).match(RGB_STRING_TO_HEX_REGEX) || [
+    'rgb(0,0,0)',
```
AndrewMusgrave:
```suggestion
   undefined,
```

Since we don't use the full capture

src/components/ColorPicker/components/AlphaField/AlphaField.tsx
```diff
@@ -0,0 +1,63 @@
...
+export interface AlphaFieldProps {
+  alpha: number;
+  onChange(alpha: number): void;
+}
+
+export function AlphaField({alpha, onChange}: AlphaFieldProps) {
+  const i18n = useI18n();
+
+  const initialValue = Math.round(alpha * 100);
+  const [percentage, setPercentage] = useState(
```
AndrewMusgrave: We can use lazy initial state here

```tsx
  const [percentage, setPercentage] = useState(() => {
    const initialValue = Math.round(alpha * 100);
    return initialValue === null ? 0 : initialValue,
  })
```

Or a slightly similar approach using the conditional or operator

```tsx
  const [percentage, setPercentage] = useState(() => Math.round(alpha * 100); || 0)
```

However since math.round isn't very expensive we could just it down too

```tsx
  const [percentage, setPercentage] = useState(Math.round(alpha * 100); || 0)
```

src/components/ColorPicker/components/AlphaField/tests/AlphaField.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+    it('calls onChange with the alpha value when percentage is set', () => {
+      const onChangeSpy = jest.fn();
+      const alphaField = mountWithAppProvider(
+        <AlphaField {...mockProps} onChange={onChangeSpy} />,
+      );
+      const textField = alphaField.find(TextField);
+      const percentageValue = 70;
+      const expectedAlpha = percentageValue / 100;
+
+      (textField.find('input') as any).instance().value = percentageValue;
```
AndrewMusgrave:
```suggestion
      textField.find('input').getDOMNode().value = percentageValue;
```

We'll want to get getDOMNode to retrieve the dom node

src/components/ColorPicker/components/AlphaField/tests/AlphaField.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+    it('calls onChange with alpha value of 1 when percentage value over 100 is set', () => {
+      const onChangeSpy = jest.fn();
+      const alphaField = mountWithAppProvider(
+        <AlphaField {...mockProps} onChange={onChangeSpy} />,
+      );
+      const textField = alphaField.find(TextField);
+      const percentageValue = 150;
+      const expectedAlpha = 1;
+
+      (textField.find('input') as any).instance().value = percentageValue;
```
AndrewMusgrave: ^

src/components/ColorPicker/components/AlphaField/tests/AlphaField.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+      textField
+        .find('input')
+        .simulate('change')
+        .simulate('blur');
```
AndrewMusgrave: Why do we need to blur here, and in the below tests?

src/components/ColorPicker/components/TextPicker/TextPicker.tsx
```diff
@@ -0,0 +1,116 @@
...
+  color: HSBColor;
+  allowAlpha?: boolean;
+  onChange(hex: string): void;
+}
+
+export function TextPicker({color, allowAlpha, onChange}: TextPickerProps) {
+  const i18n = useI18n();
+
+  const [text, setText] = useState('');
+  const [lastValidValue, setLastValidValue] = useState('');
```
AndrewMusgrave: We could get *probably*(this is an asumption) get away with this being a ref since we'll never need it to update a component (cause a re-render)

src/components/ColorPicker/components/TextPicker/TextPicker.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const className = classNames(
+    styles.TextPicker,
+    allowAlpha && styles.AlphaAllowed,
+  );
+  const label = i18n.translate(
+    'Polaris.ColorPicker.textPickerAccessibilityLabel',
+  );
+  const valueForDisplay = isHexString(text) ? text.toUpperCase() : text;
+
+  const renderSelectedColorSwatch = () => {
```
AndrewMusgrave: Rather than creating a function we usually create a variable and suffix it with markup, i.e

```tsx
const preficClassNames = classNames(
  styles.TextFieldSwatch,
  allowAlpha && styles.AlphaAllowed,
);
const prefixStyle = {backgroundColor: hsbToString(color)};
const prefixMarkup = (
  <div className={prefixClassNames}>
     <div style={prefixStyle} className={styles.SwatchBackground} />
   </div>
)
```

src/components/ColorPicker/components/TextPicker/TextPicker.tsx
```diff
@@ -0,0 +1,116 @@
...
+    switch (true) {
+      case isHexString(normalizedValue):
+        return expandHex(normalizedValue);
+      case isHashlessHex(normalizedValue):
+        return expandHex(`#${normalizedValue}`);
+      case isRgbString(normalizedValue):
+        return rgbStringToHex(normalizedValue);
+      case isColorName(normalizedValue):
+        return nameToHex(normalizedValue);
+      default:
```
AndrewMusgrave: We can omit the default statement and the function will return undefined 😄

src/components/ColorPicker/components/TextPicker/tests/TextPicker.test.tsx
```diff
@@ -0,0 +1,153 @@
...
+    it("calls onChange with #FFFFFF when 'white' is entered", () => {
+      const onChangeSpy = jest.fn();
+      const textPicker = mountWithAppProvider(
+        <TextPicker {...mockProps} onChange={onChangeSpy} />,
+      );
+      const textField = textPicker.find(TextField);
+      const colorValue = 'white';
+      const expectedHex = nameToHex(normalizeColorString(colorValue));
+
+      (textField.find('input') as any).instance().value = colorValue;
```
AndrewMusgrave: Same comment about getDOMNode and questions about blur

src/components/ColorPicker/ColorPicker.tsx
AndrewMusgrave: I'll leave this up to Miru's best judgment 😄

src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -41,25 +51,40 @@ export class ColorPicker extends React.PureComponent<ColorPickerProps, State> {
...
+        });
       }, 0);
     }
   }
 
   render() {
     const {id, color, allowAlpha} = this.props;
     const {hue, saturation, brightness, alpha: providedAlpha} = color;
-    const {pickerSize} = this.state;
+    const {pickerWidth, pickerHeight} = this.state;
```
AndrewMusgrave: Just curious, do we need to update these values on resize?

src/components/ColorPicker/components/AlphaField/tests/AlphaField.test.tsx
```diff
@@ -0,0 +1,95 @@
+import React from 'react';
+import {mountWithAppProvider} from 'test-utilities/legacy';
+import {Spinner} from 'components';
```
AndrewMusgrave:
```suggestion
```

This should fix CI 😄

AndrewMusgrave (commented):
Thanks for adding tests, everything is looking great 😍

src/utilities/color-validation.ts
```diff
@@ -9,3 +22,24 @@ export function isLight({red, green, blue}: RGBColor | RGBAColor): boolean {
...
+export function isHexString(value: string) {
+  return HEX_REGEX.test(value);
+}
+
+export function isHashlessHex(value: string) {
+  return HASHLESS_HEX_REGEX.test(value);
+}
+
+export function isRgbString(value: string) {
+  return /rgba?\(\s*(0*((25[0-5]|2[0-4]\d|1\d{1,2}|\d\d?)\s*,\s*?)){2}(0*(25[0-5]|2[0-4]\d|1\d{1,2}|\d\d?))\s*,?\s*([01]\.?\d*?)?\s*\)/i.test(
```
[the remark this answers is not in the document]
ruairiphackett: `rgba` isn't actually supported as input in this version (it may be [added later]()), so I will change this to only match `rgb` strings for now.

ruairiphackett:
>I maybe under an incorrect assumption but I suspect the majority of people don't know color keywords, apart from the basic; red, green, purple. Not the rebeccapurple, forestgreen, papayawhip, etc. Is this something being requested accessibility wise? @dpersing

I'm not saying we shouldn't have it, just curious about the context before we increase our bundle 😄

@AndrewMusgrave Good point, I'd be up for either reducing the name map (although that might make it difficult for people to figure out what colour names will work) or removing it completely depending on what people think 💯

src/components/ColorPicker/components/AlphaField/AlphaField.tsx
ruairiphackett: Nice!

src/components/ColorPicker/components/AlphaField/tests/AlphaField.test.tsx
ruairiphackett: `onChange` gets called by `handleBlur` [here](https://github.com/Shopify/polaris-react/pull/2186/files#diff-05e020bf01a1b218631cd41871d91d5dR42), so the `blur` event needs to occur for the `onChangeSpy` to be called. This is also the case with the other instances below.

src/components/ColorPicker/components/TextPicker/TextPicker.tsx
ruairiphackett: Yep, you're right, I'll change it.

src/components/ColorPicker/components/TextPicker/tests/TextPicker.test.tsx
ruairiphackett: Answer about `blur` above, will change to `getDOMNode`.

src/components/ColorPicker/components/AlphaField/tests/AlphaField.test.tsx
ruairiphackett: Oops!

src/components/ColorPicker/ColorPicker.tsx
ruairiphackett: Good point, they should be updated 👍

ruairiphackett:
@AndrewMusgrave Should have those comments addressed now, thanks again

apman:
Hi, this is looking great - I chanced upon your PR because I was also working on adding a hex input to the colour picker for the Rich Text Editor in the Admin (Web) ... Looking forward to using yours instead!

But I just want to add to the comment @tleunen made earlier
> * Should the `onChange` callback return the hex value alongside its params? Or should the consumer find out a way to convert the hue, brithness, saturation and alpha into an rgba/hex? In most cases, I'd guess the consumer just wants a color value?

I think returning RGB values might be a more important issue than just convenience to the consumer, because the RGB<->HSB conversion is NOT exactly one-to-one. If the TextPicker returns HSB and the consumer converts it to RGB for use with their HTML, the resulting RGB value will not always be the same as what was typed into the picker! 
E.g. `#ac4056` will become `#ab3f55` in the round-trip conversion! (maybe not a big deal visually, but if a merchant inputs their corporate colour x, I really don't think we can have it show up as some x +/- 1 approximation next time they open the colour picker on the same selection! 😆)

I realize that changing the return type of the overall ColorPicker component makes the roll-out more complicated, but unless there are other conversion algorithms that don't have this problem, it might be something you should consider.

ruairiphackett:
@apman I've been looking into the issue of the RGB->HSB->RGB round trip conversion & it seems to be solvable by removing some rounding included in the current conversion utility functions. I will add a PR on a sub-branch of this one & let you know when it's up.

>Should the onChange callback return the hex value alongside its params? Or should the consumer find out a way to convert the hue, brithness, saturation and alpha into an rgba/hex? In most cases, I'd guess the consumer just wants a color value?

@tleunen This is a good idea, currently achievable by copying the hex from the text field, but it would make sense to return a more usable colour format alongside the `hsba` colour type 💯 

I'll run it by the Polaris team to see what would be the best way to achieve this.

danrosenthal:
@ruairiphackett, what's the status on this PR? Ready for another round of review from @AndrewMusgrave?

AndrewMusgrave:
iirc there were a few un-answered questions that @mirualves followed up with? e.g. supporting color keywords

mirualves:
@danrosenthal, this is part of the work on the a11y backlog. We paused this while the BOP team work on the color system project, component audit, etc.

ruairiphackett:
@danrosenthal @AndrewMusgrave Yes, this is ready for another review when you have time, I rebased it this morning and addressed some linting changes.

I noticed some new `codecov` issues since I rebased, I'm not 100% sure how to read them or what exactly I need to fix to pass (I think I need to cover the resize event in thee `ColorPicker` tests & it should be OK?). Maybe you can advise in the review @AndrewMusgrave?

>iirc there were a few un-answered questions that @mirualves followed up with? e.g. supporting color keywords

@dpersing, @miualves & myself made the decision (in the 3 comments [starting from here](https://github.com/Shopify/polaris-react/pull/2186#issuecomment-541096698)) to go with the following for this version:
- 3 & 6 digit hex values (with/without preceding `#`)
- Colour keywords 
- `rgb` values.

Further support will be added later and is tracked here: https://github.com/Shopify/polaris-ux/issues/383

src/components/ColorPicker/components/TextPicker/tests/TextPicker.test.tsx
ruairiphackett: `getDOMNode` fails without casting, so I will change (this and all other instances) to 
```
(textField.find('input').getDOMNode() as HTMLInputElement).value = colorValue;
```

AndrewMusgrave: For reference a way to avoid casting here could be to use a type guard. Here's an example using [instanceof](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof) 😄 
```tsx
const input = textField.find(‘input’).getDOMNode();
const value = input instanceof HTMLElement ? ‘’ : input.value;
```

ruairiphackett: 🤔 Using the ternary operator here gives the further error of 
`Tests should not contain conditional statements.eslint(jest/no-if)`

It also didn't seem to allow me to set the input node's value.

ruairiphackett: @AndrewMusgrave I pushed a fix using casting (`(textField.find('input').getDOMNode() as HTMLInputElement).value`) just to get this to a place where it is ready for another review, I'm not ignoring your advice, don't worry! Just haven't been able to get an error free method to work without casting 🤔

tleunen:
What is the status of this new update? Need help with the review? I'm very interested in using the new update in my project instead of our custom picker ;)

ruairiphackett:
Hey @tleunen, this is on hold at the moment awaiting review. I'd love help with a review, I've rebased it a couple of times already to try keep it mergeable, do you want me to rebase it again?

tleunen:
Yes please, to be sure we get the latest, I'll review this today then ;)

ruairiphackett:
@tleunen This has been rebased now.

There are other features being added to this on sub-branches in other PRs, I'm in the progress of rebasing those as well (it might be tomorrow by the time I have them all up to date):
- https://github.com/Shopify/polaris-react/pull/2599
- https://github.com/Shopify/polaris-react/pull/2399
- https://github.com/Shopify/polaris-react/pull/2405
- https://github.com/Shopify/polaris-react/pull/2413

src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -67,25 +110,44 @@ export class ColorPicker extends React.PureComponent<ColorPickerProps, State> {
       />
     ) : null;
 
+    const hexPickerMarkup = (
+      <TextPicker
+        color={color}
+        allowAlpha={allowAlpha}
```
tleunen: Instead of having to pass an `allowAlpha`, which the `TextPicker` shouldn't really care about imo. Could we render them in a grid or even a flex container and let this component take a bit more space than the alpha field?

src/components/ColorPicker/components/AlphaField/AlphaField.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export interface AlphaFieldProps {
+  alpha: number;
+  onChange(alpha: number): void;
+}
+
+export function AlphaField({alpha, onChange}: AlphaFieldProps) {
+  const i18n = useI18n();
+
+  const [percentage, setPercentage] = useState(Math.round(alpha * 100) || 0);
```
tleunen: Should this initial percentage be clamped?

src/components/ColorPicker/components/TextPicker/TextPicker.tsx
```diff
@@ -0,0 +1,114 @@
...
+  color: HSBColor;
+  allowAlpha?: boolean;
+  onChange(hex: string): void;
+}
+
+export function TextPicker({color, allowAlpha, onChange}: TextPickerProps) {
+  const i18n = useI18n();
+
+  const [text, setText] = useState('');
+  const lastValidValue = useRef<string>('');
```
tleunen: nit: no need explicit typing here

src/components/ColorPicker/components/TextPicker/TextPicker.tsx
```diff
@@ -0,0 +1,114 @@
...
+    allowAlpha && styles.AlphaAllowed,
+  );
+  const prefixStyle = {backgroundColor: hsbToString(color)};
+  const prefixMarkup = (
+    <div className={prefixClassNames}>
+      <div style={prefixStyle} className={styles.SwatchBackground} />
+    </div>
+  );
+
+  const coerceToValidUserInput = (value: string) => {
```
tleunen: Move this outside of the component?

src/components/ColorPicker/components/TextPicker/TextPicker.tsx
```diff
@@ -0,0 +1,114 @@
...
+    const normalizedValue = normalizeColorString(value);
+    switch (true) {
```
tleunen: Hmm... Why not grouping Hex and HashLessHex, then having `expandHex` prepending the hash or not?

Can we keep it simple with `if`? 😅

tleunen (commented):
The feature seems to be working fine! Good work! :)

src/components/ColorPicker/ColorPicker.scss
```diff
@@ -36,8 +45,12 @@ $stacking-order: (
   @include checkers;
   position: relative;
   overflow: hidden;
-  height: $picker-size;
-  width: $picker-size;
+  height: $picker-height;
+  width: $picker-width;
+
+  &.AlphaAllowed {
+    width: $picker-with-alpha-width;
```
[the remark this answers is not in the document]
tleunen: Oh I just saw the earlier discussions about that too... I'm not sure increasing the width is really the best approach tho.
I understand why we would want the field to be large enough to fit a full rgb color, but it's main purpose is to render an hex value, so if we don't render the full rgb when the user writes it, I guess it's fine?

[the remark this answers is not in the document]
ruairiphackett: @mirualves What do you think about this approach?

src/components/ColorPicker/ColorPicker.tsx
ruairiphackett: One issue with not passing it is the use of `allowAlpha` in the [TextFieldSwatch](https://github.com/Shopify/polaris-react/blob/8c6f48f5a1213bd18cb541c3841b9968af100130/src/components/ColorPicker/ColorPicker.scss#L187).

src/components/ColorPicker/components/TextPicker/TextPicker.tsx
ruairiphackett: Thanks a million for the review @tleunen!

Apologies in the delay getting back to you.

>Why not grouping Hex and HashLessHex, then having expandHex prepending the hash or not?

I wasn't 100% sure sure what you meant by this comment?

tleunen: Sorry 😅
I meant having `expandHex` knowing if it has to prepend the hash character or not.
Something like this: (and the `isHexString` could also test both scenario)
```
if isHexString(x) or isHashlessHex(x)
  expandHex(x)
```

src/components/ColorPicker/ColorPicker.scss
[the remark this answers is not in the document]
mirualves: As long as the component width adapts to accept RGB colors comfortably, I don't see a problem.

ctrlaltdylan:
Very interested to see this to completion. Anything I can do to help?

alex-page:
@ruairiphackett This PR is excellent and is quite out of date. Are there any smaller pieces you could break apart and ship easily?

ruairiphackett:
>This PR is excellent and is quite out of date. Are there any smaller pieces you could break apart and ship easily?

@alex-page I would like to revisit this when I get some time to work on it, hopefully in the new year.

alex-page:
@ruairiphackett going to close for now but if you want to keep working on this please reopen.

ruairiphackett:
Thanks for the heads up @alex-page

sagar-ranglani:
@ruairiphackett This looks very useful... any updates about this?

ruairiphackett:
For anyone who had interest in this PR, it has been updated and recreated in https://github.com/Shopify/polaris/pull/8294

PR #2193, opened 2019-09-24
[DateSelector] Hookify component

AndrewMusgrave:

Part of #1995

Convert DateSelector to a functional component

### Notes

I noticed while converting this component,the `setState` callback was used quite often however I couldn't figure out why, except when setting state with the date. I'm adding @dleroux since he'll have the most context on this

## 

### How to 🎩

Playground with date filter

<details>

```jsx
import React from 'react';
import {ResourceList, TextStyle, Avatar, FilterType, Card} from '../src';

export class Playground extends React.Component {
  state = {
    searchValue: '',
    appliedFilters: [
      {
        key: 'accountStatusFilter',
        value: 'Account enabled',
      },
    ],
  };

    this.setState({searchValue});
  };

  };

  renderItem = (item) => {

    return (
        <h3>
        </h3>
        <div>{location}</div>
      </ResourceList.Item>
    );
  };

  render() {
    const resourceName = {
      singular: 'customer',
      plural: 'customers',
    };

    const items = [
      {
        id: 341,
        url: 'customers/341',
        name: 'Mae Jemison',
        location: 'Decatur, USA',
      },
      {
        id: 256,
        url: 'customers/256',
        name: 'Ellen Ochoa',
        location: 'Los Angeles, USA',
      },
    ];

    const filters = [
      {
        key: 'accountStatusFilter',
        label: 'Date selector',
        type: FilterType.DateSelector,
        minKey: '0',
        maxKey: '30',
      },
    ];

    const filterControl = (
      <ResourceList.FilterControl
        filters={filters}
        additionalAction={{
          content: 'Save',
        }}
      />
    );

    return (
      <Card>
        <ResourceList
          resourceName={resourceName}
          items={items}
          renderItem={this.renderItem}
          filterControl={filterControl}
        />
      </Card>
    );
  }
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/ResourceList/components/FilterControl/components/DateSelector/DateSelector.tsx
+++ b/src/components/ResourceList/components/FilterControl/components/DateSelector/DateSelector.tsx
@@ -1,13 +1,10 @@
-import React from 'react';
+import React, {useState, useCallback, useRef, memo} from 'react';
 import {CalendarMinor} from '@shopify/polaris-icons';
 import {DatePicker, Months, Year, Range} from '../../../../../DatePicker';
 import {Select} from '../../../../../Select';
 import {TextField} from '../../../../../TextField';
 import {Icon} from '../../../../../Icon';
-import {
-  withAppProvider,
-  WithAppProviderProps,
-} from '../../../../../../utilities/with-app-provider';
+import {useI18n} from '../../../../../../utilities/i18n';
 
 import styles from './DateSelector.scss';
 
@@ -25,17 +22,6 @@ export interface DateSelectorProps {
   onFilterKeyChange(filterKey?: string): void;
 }
 
-interface State {
-  selectedDate?: Date;
-  userInputDate?: string;
-  userInputDateError?: string;
-  datePickerMonth: Months;
-  datePickerYear: Year;
-  initialConsumerFilterKey?: string;
-}
-
-type CombinedProps = DateSelectorProps & WithAppProviderProps;
-
 export enum DateFilterOption {
   PastWeek = 'past_week',
   PastMonth = 'past_month',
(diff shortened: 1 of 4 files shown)


src/components/ResourceList/components/FilterControl/components/DateSelector/DateSelector.tsx
```diff
@@ -49,200 +36,296 @@ export enum DateFilterOption {
...
+  filterMinKey,
+  filterMaxKey,
+  dateOptionType,
+  onFilterValueChange,
+  onFilterKeyChange,
+}: DateSelectorProps) {
+  const now = new Date();
+
+  const {translate} = useI18n();
+  const initialConsumerFilterKey = useLazyRef(() => filterKey);
```
BPScott: What's the value in using useLazyRef here? It looks like filterKey is a plain string. using useLazyRef instead of useRef is useful if the default value is something that would be expensive to compute on every render, but that doesn't seem to be the case here.

src/components/ResourceList/components/FilterControl/components/DateSelector/DateSelector.tsx
```diff
@@ -49,200 +36,296 @@ export enum DateFilterOption {
...
+  filterKey,
+  filterMinKey,
+  filterMaxKey,
+  dateOptionType,
+  onFilterValueChange,
+  onFilterKeyChange,
+}: DateSelectorProps) {
+  const now = new Date();
+
+  const {translate} = useI18n();
```
BPScott: Existing code tends to do `const i18n = useI18n();` (my personal preference) or `const intl = useI18n();` (a hangover from before we renamed this) rather than destructure.

We should stick to one style throughout the codebase. Which one do you prefer?

src/components/ResourceList/components/FilterControl/components/DateSelector/DateSelector.tsx
AndrewMusgrave: Your right, it was a blip, no need to lazyRef here 😄

src/components/ResourceList/components/FilterControl/components/DateSelector/DateSelector.tsx
AndrewMusgrave: I would definitely prefer `i18n` or `intl`, the name clashing seems odd, best to just stick with one. However, we usually deconstruct whenever possible in our code. So I'm wondering if we should use `const {translate} = useI18n()` to be consistent?  I'm wondering if the reason we didn't originally deconstruct was that iirc web was used as a guideline and they can't deconstruct because they need access to the context, however, we don't but we just didn't deconstruct? 🤷‍♂ Either way as long as well consistent I'm ok with it.

BPScott: i18n is an instance of the I18n class, so I'm kinda surprised that we can even destructure it in the first place. I thought destructuring only worked on objects.

To reduce others being suprised by that I'd be tempted to avoid destructuring and stick with `const i18n = useI18n()`.

AndrewMusgrave: > i18n is an instance of the I18n class, so I'm kinda surprised that we can even destructure it  in the first place. I thought destructuring only worked on objects.

In JavaScript, mostly everything is going to be an object, even regex since it's not primitive. Classes themselves are not objects, but functions (typeof class Test {} === 'function') however they create a instance of the function in an object (typeof new Test === 'object'). The problem with destructuring is that it'll remove context.

```js
class Test {
	constructor() {
  	this.num = 5;
  }

	func() {
  	return this.num;
  }

}

// All good
const t = new Test();
t.func();

// Losing context
const {func} = t;
func(); // Bad error ☹️
```

We don't use `this` within `I18n` so we're ok deconstructing. 

> To reduce others being suprised by that I'd be tempted to avoid destructuring and stick with const i18n = useI18n().

Two more points to support both sides of the argument.

Unless your playing with `i18n` you won't know it originated from a class.

If we don't deconstruct we'll be able to add features that do use `this` within having to refactor.

BPScott: Huh neat! It seems like [I18n's translate function already uses `this`](https://github.com/Shopify/polaris-react/blob/master/src/utilities/i18n/I18n.ts#L35) so by that logic this shouldn't even work now. And yet it seems to be manage. That still feels like it's more by luck than judgement though, so I'd still prefer not destructuring

AndrewMusgrave: Oops looks like we are. It's working because translate is class property (not currently part of JavaScript but in the works)

```js
class Test {
  func = () => {}
}
```

Could compile down to something like 

```js
class Test {
  constructor() {
    this.func = () => {}
  }
}
```

Lamda functions lack certain contexts, including `this` and *usually* have the environment they were created in.

```js
class Test {
  func = () => { console.log(this) }
}

const {func} = new Test;
func(); // All ok
```

```js
class Test {
  func() { console.log(this) }
}

// Sorta gross alternative without using class properties still deconstructing
const test = new Test;
const {func} = test;
const boundFunc = func.bind(test);
boundFunc(); // All ok
```

Either way, I'm ok with not deconstructing our classes, it'll help future proof us and reduce the change of errors. We should reach out for other opinions as well.

AndrewMusgrave: Talked it over with @dleroux and we're all on board with not deconstructing

Issue #2100, opened 2019-09-09
DataTable display totals row at the bottom


vict-shevchenko:
## Feature request summary
In `DataTable` component display "Totals" row at the bottom of the table

Introduce an additional parameter to control the position of totals row for `DataTable`. That will allow to display it at the bottom of the table.

## Rationale

Generally, for most users, it is more convenient to see summary fields at the bottom of the table. This pattern is used in most eCommerce shops when doing a checkout.

Like this:

![some shop checkout table](https://miro.medium.com/max/2048/1*XD5tXB9U5Javay42EAJZDw.png)

Please let me know if this is interesting for the project, so I can prepare a pull request.

---

🌟 Feature requests that are not yet planned will be closed. We then use the issue’s [:+1: upvotes](https://help.github.com/articles/about-conversations-on-github/) to track and set priorities. See the [contribution guidelines](https://github.com/Shopify/polaris-react/blob/master/.github/CONTRIBUTING.md) for more information.

PR #2200, opened 2019-09-25
[DataTable] Add ability to control location of Totals row

mitchmaps:
#### TL;DR: 
Added a prop to DataTable which allows control over placement of the Totals row.

Addresses #2100 

Currently the DataTable component displays the `Totals` row at the top of the table which is defintely a valid place to display that information but in some cases it makes more sense from a UX perspective to have that information at the bottom of the table. This creates the visual flow as your eye goes down each column and subconsciously starts to add up the different values. Having the option in Polaris to do both will make the component more flexible and allow devs and designers alike to customize it to fit the context that it's being implemented in.

This PR introduces an additional optional prop to the `DataTable` component called _footerTotals_ (could definitely find a better name) that, when set to true will place the existing Totals row and it's data supplied in other props at the bottom of the data table. Introducing these changes this way makes them backwards compatible with existing implementations of the DataTable because if the prop hasn't been supplied then then the default behaviour will be preserved.

#### Default rendering:

```jsx
<DataTable
  columnContentTypes={[
    'text',
    'numeric',
    'numeric',
    'numeric',
    'numeric',
  ]}
  headings={[
    'Product',
    'Price',
    'SKU Number',
    'Net quantity',
    'Net sales',
  ]}
  rows={rows}
/>
```
image

#### With footerTotals rendering:

```jsx
<DataTable
  columnContentTypes={[
    'text',
    'numeric',
    'numeric',
    'numeric',
    'numeric',
  ]}
  headings={[
    'Product',
    'Price',
    'SKU Number',
    'Net quantity',
    'Net sales',
  ]}
  rows={rows}
  footerTotals
/>
```

image

If you have any suggestions for a better prop name than footerTotals 🦶🔢, let me know 😂

### DIFF
--- a/src/components/DataTable/DataTable.tsx
+++ b/src/components/DataTable/DataTable.tsx
@@ -136,11 +136,11 @@ class DataTable extends React.PureComponent<CombinedProps, DataTableState> {
       condensed && styles.condensed,
     );
 
+    const headingMarkup = <tr>{headings.map(this.renderHeadings)}</tr>;
+
     const totalsMarkup = totals ? (
       <tr>{totals.map(this.renderTotals)}</tr>
     ) : null;
-
-    const headingMarkup = <tr>{headings.map(this.renderHeadings)}</tr>;
 
     const bodyMarkup = rows.map(this.defaultRenderRow);
 
--- a/src/components/DataTable/DataTable.tsx
+++ b/src/components/DataTable/DataTable.tsx
@@ -144,18 +144,12 @@ class DataTable extends React.PureComponent<CombinedProps, DataTableState> {
 
     const bodyMarkup = rows.map(this.defaultRenderRow);
 
-    let footerMarkup;
+    const footerMarkup = footerContent ? (
+      <div className={styles.Footer}>{footerContent}</div>
+    ) : null;
 
-    if (footerContent) {
-      footerMarkup = footerTotals ? (
-        <div className={styles.Footer}>
-          {footerContent}
-          {totalsMarkup}
-        </div>
-      ) : (
-        <div className={styles.Footer}>{footerContent}</div>
-      );
-    }
+    const headerTotalsMarkup = !footerTotals ? totalsMarkup : null;
+    const footerTotalsMarkup = footerTotals ? totalsMarkup : null;
 
     return (
       <div className={wrapperClassName}>
@@ -177,9 +171,10 @@ class DataTable extends React.PureComponent<CombinedProps, DataTableState> {
             <table className={styles.Table} ref={this.table}>
               <thead>
                 {headingMarkup}
-                {totalsMarkup}
+                {headerTotalsMarkup}
               </thead>
               <tbody>{bodyMarkup}</tbody>
+              {footerTotalsMarkup}
             </table>
           </div>
           {footerMarkup}
--- a/src/components/DataTable/DataTable.tsx
+++ b/src/components/DataTable/DataTable.tsx
@@ -144,20 +144,18 @@ class DataTable extends React.PureComponent<CombinedProps, DataTableState> {
 
     const bodyMarkup = rows.map(this.defaultRenderRow);
 
-    const footerMarkup = footerContent ? (
-      <div className={styles.Footer}>{footerContent}</div>
-    ) : null;
+    let footerMarkup;
 
-    // if (footerContent) {
-    //   const footerMarkup = footerTotals ? (
-
-    //   ) : (
-    //     <div className={styles.footer}
-    //     //<div className={styles.Footer}>{footerContent}</div>
-    //   )
-    // }
-
-    // if (footer)
+    if (footerContent) {
+      footerMarkup = footerTotals ? (
+        <div className={styles.Footer}>
+          {footerContent}
+          {totalsMarkup}
+        </div>
+      ) : (
+        <div className={styles.Footer}>{footerContent}</div>
+      );
+    }
 
     return (
       <div className={wrapperClassName}>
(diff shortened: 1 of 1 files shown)


UNRELEASED.md
```diff
@@ -10,6 +10,9 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Added skipToContentTarget prop to Frame component ([#2080](https://github.com/Shopify/polaris-react/pull/2080))
```
```suggestion
```

Remove this

src/components/DataTable/DataTable.tsx
```diff
@@ -32,6 +32,8 @@ export interface DataTableProps {
   headings: string[];
   /** List of numeric column totals, highlighted in the table’s header below column headings. Use empty strings as placeholders for columns with no total. */
   totals?: TableData[];
+  /** Placement of totals row within table, true for t */
+  footerTotals?: boolean;
```
danrosenthal: Can you spend a bit of time looking through the `polaris-react` codebase for any other examples of conditional-layout-related props similar to this? I'd like to understand what else exists so we can try to align naming conventions.

mitchmaps:
Sorry for the early re-request, I didn't see the comment about test coverage! @danrosenthal

src/components/DataTable/DataTable.tsx
danrosenthal: Looking at the `alignFlushLeft` (verb: 'align') prop on `ContextualSaveBar` and the `showHeader` (verb: 'show') prop on `ResourceList`, I think it makes more sense to use a verb-based prop for this.

mitchmaps: I like this idea, makes the prop seem more action oriented just by reading it

mitchmaps:
@danrosenthal I'm not sure how I'd be able to test these sort of changes because the only effect this new prop has on the component is a visual and we wouldn't normally test for that case. What's your take on this?

danrosenthal:
That's a great point @mitchmaps, we don't want to test the visuals on this, we just want to make sure that all code paths are covered by tests.

|Codecov|
|-|
|Screen Shot 2019-09-27 at 12 23 40 PM|

Looking at the coverage report, we see that each of these lines of code is only half covered, meaning only one outcome of the ternary is covered. By passing the footer totals prop in any test, we will get code coverage over these lines. We don't even necessarily need to assert the outcome of passing that prop. We just want to make sure that when our tests run, all possible code paths are executed.

In this case, your test could simply be that the totals markup is rendered **somewhere** in the component when footer totals is passed, visual regression testing will handle whether it is rendered in the **right** place. Alternatively, you could simply find an existing test, pass the prop in that test, and assert nothing about its behavior. This will execute all code paths, and create full coverage on those lines of code.

mitchmaps:
I did consider the other name for the action oriented prop but I figured that `showFooterTotals` made it sound like we were rendering total information about the footer rather than just rendering existing info within the footer. `showTotalsFooter`, to me sounds more like existing information being rendered in a different location but I agree isn't the best name. Maybe even just renaming it to `showTotalsInFooter` would make it more clear. What do you think? @danrosenthal

danrosenthal:
`showTotalsInFooter` 💯

src/components/DataTable/DataTable.tsx
```diff
@@ -146,6 +154,9 @@ class DataTable extends React.PureComponent<CombinedProps, DataTableState> {
       <div className={styles.Footer}>{footerContent}</div>
     ) : null;
 
+    const headerTotalsMarkup = !showTotalsInFooter ? totalsMarkup : null;
+    const footerTotalsMarkup = showTotalsInFooter ? totalsMarkup : null;
```
danrosenthal:
```suggestion
    const footerTotalsMarkup = showTotalsInFooter ? (<tfoot>{totalsMarkup}</tfoot>) : null;
```

mitchmaps: Implemented! I put the <tfoot> in the actual render function to make the conditions there a bit more consistent and readable. Do you think that makes sense?

danrosenthal: Since the `tfoot` should only render when `footerTotalsMarkup` is true, I think it makes more sense to put it in line 158 instead. With your current implementation, the `tfoot` will always render, even when empty. We don't want that.

src/components/DataTable/tests/DataTable.test.tsx
```diff
@@ -155,6 +155,15 @@ describe('<DataTable />', () => {
...
       totalsCells.forEach((total) => expect(total.text()).toBe(''));
     });
+
+    it('renders totals in the footer with a showTotalsInFooter prop', () => {
+      const totals = ['', '', ''];
+      const dataTable = mountWithAppProvider(
+        <DataTable {...defaultProps} totals={totals} showTotalsInFooter />,
+      );
+
+      expect(dataTable.find('thead tr')).toHaveLength(1);
```
danrosenthal: This is going to be true whether or not you pass `showTotalsInFooter` prop. Take a look at the way the markup is output and you'll see that this does not test the thing you want tested. If you make the `tfoot` conditionally rendered, you can change the test to:

```suggestion
      expect(dataTable.find('tfoot')).toHaveLength(1);
```

And it will accurately test the rendering you want.

mitchmaps: Good catch, I see the reasoning behind that now!

mitchmaps: @danrosenthal I think I've got it all set up now

danrosenthal:
Good eye @sarahill!

mitchmaps:
@sarahill , that suggestion makes a lot of sense. I've implemented the changes now:

image

vict-shevchenko:
Thanks a lot, we will now consume these changes in our project.

PR #2203, opened 2019-09-26
[useEventListener][useComponentDidMount] Add useComponentDidMount & useEventListener

AndrewMusgrave:

I needed these while converting dropzone to a functional component

* add hooks
* add tests
* add docs

### How to 🎩

* tests
* try to break them in the playground

### 🎩 checklist

### DIFF
--- a/src/utilities/use-component-did-mount.ts
+++ b/src/utilities/use-component-did-mount.ts
@@ -5,7 +5,7 @@ import {useIsAfterInitialMount} from './use-is-after-initial-mount';
 /**
  * Similarly to the lifecycle hook componentDidMount, useComponentDidMount
  * will be invoked after the component has mounted, and only the initial mount.
- * @param callback Defines a callback to use invoked once the component has
+ * @param callback Defines a callback to invoke once the component has
  * initially mounted.
  */
 export function useComponentDidMount(callback: EffectCallback) {
--- a/src/utilities/use-component-did-mount.ts
+++ b/src/utilities/use-component-did-mount.ts
@@ -1,0 +1,19 @@
+import {useRef} from 'react';
+import {EffectCallback} from '../types';
+import {useIsAfterInitialMount} from './use-is-after-initial-mount';
+
+/**
+ * Similarly to the lifecycle hook componentDidMount, useComponentDidMount
+ * will be invoked after the component has mounted, and only the initial mount.
+ * @param callback Defines a callback to use invoked once the component has
+ * initially mounted.
+ */
+export function useComponentDidMount(callback: EffectCallback) {
+  const isAfterInitialMount = useIsAfterInitialMount();
+  const hasInvokedLifeCycle = useRef(false);
+
+  if (isAfterInitialMount && !hasInvokedLifeCycle.current) {
+    hasInvokedLifeCycle.current = true;
+    return callback();
+  }
+}
--- a/src/utilities/tests/use-component-did-mount.test.tsx
+++ b/src/utilities/tests/use-component-did-mount.test.tsx
@@ -1,0 +1,43 @@
+import React from 'react';
+import {mount} from 'test-utilities';
+import {useComponentDidMount} from '../use-component-did-mount';
+import {useIsAfterInitialMount} from '../use-is-after-initial-mount';
+
+describe('useComponentDidMount', () => {
+  it('invokes your callback after mount', () => {
+    const spy = jest.fn();
+    function Component() {
+      const isAfterInitialMount = useIsAfterInitialMount();
+      useComponentDidMount(() => {
+        spy(isAfterInitialMount);
+      });
+
+      return null;
+    }
+
+    mount(<Component />);
+
+    expect(spy).toHaveBeenCalledWith(true);
+  });
+
+  it('invokes the callback once', () => {
+    const spy = jest.fn();
+    function Component() {
+      useComponentDidMount(spy);
+      return null;
+    }
+
+    const component = mount(<Component />);
+
+    component.setProps({});
+    component.setProps({});
+    component.setProps({});
+    component.setProps({});
+    component.setProps({});
+    component.setProps({});
+
+    component.unmount();
+
+    expect(spy).toHaveBeenCalledTimes(1);
+  });
+});
--- a/src/utilities/tests/use-event-listener.test.tsx
+++ b/src/utilities/tests/use-event-listener.test.tsx
@@ -1,0 +1,75 @@
+import React, {useRef, useEffect} from 'react';
+import {mount} from 'test-utilities';
+import {useEventListener} from '../use-event-listener';
+
+describe('useComponentDidMount', () => {
+  let addEventListenerSpy: jest.SpyInstance;
+  let removeEventListenerSpy: jest.SpyInstance;
+
+  beforeEach(() => {
+    addEventListenerSpy = jest.spyOn(document, 'addEventListener');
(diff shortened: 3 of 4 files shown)


src/utilities/use-component-did-mount.ts
```diff
@@ -0,0 +1,19 @@
...
+import {EffectCallback} from '../types';
+import {useIsAfterInitialMount} from './use-is-after-initial-mount';
+
+/**
+ * Similarly to the lifecycle hook componentDidMount, useComponentDidMount
+ * will be invoked after the component has mounted, and only the initial mount.
+ * @param callback Defines a callback to use invoked once the component has
+ * initially mounted.
+ */
+export function useComponentDidMount(callback: EffectCallback) {
```
AndrewMusgrave: I was pondering about adding this one but it's a clean way to achieve a component did mount without ignoring dependencies in `useEffect`

AndrewMusgrave:
Going to take a quick look in the morning to make sure I didn't see anything

src/utilities/use-component-did-mount.ts
BPScott: My understanding of hooks is that your mental modal changes from component lifecycles "do something if the  component did mount, did update or will unmount" to "do something if this value changes".

"trigger only on the first time a value changes" feels like an odd thing to me - why do we want that instead of changing whenever a value changes?

AndrewMusgrave: Yup, the mental model is definitely different between functional components and class-based components, and lifecycles aren't going to line up. But even in functional components your still going to have code paths that you'll only want to execute once. They'll be less common in a components library, for example, a network is probably the most common. Use cases we might have are measuring dom elements, delay measurements to allow time for styles to load (I think we do this in a few components), maybe registering something to be tracked (I'm thinking `StickyManager`) but this is just an assumption.

> Alternatively is there some way we can represent this better in the calling component - where it would use useEffect, but we make it so the value we pass into useEffect would only be changed once (perhaps it is the result of calling a memoized function)

cc/ @dleroux who mentioned he was interesting in hooks

The reason why I choose to go with `componentDidMount` was that it'll be very clear to anyone that reads it what exactly it's doing.

AndrewMusgrave: Just ccing again so you'll see it in the moring cc/ @dleroux

BPScott: > The reason why I choose to go with componentDidMount was that it'll be very clear to anyone that reads it what exactly it's doing.

I agree on the clarity, however I wonder if being clear in a mental model that we're moving away from is actually that useful? Over time peoples familiarity with the old model will fade. 

> Use cases we might have are measuring dom elements, delay measurements to allow time for styles to load (I think we do this in a few components), maybe registering something to be tracked (I'm thinking StickyManager) but this is just an assumption.

Possibly a better question from my end:

We have `useIsAfterInitialMount`, this `useComponentDidMount` and a proposed `useIsMountedRef` from #2167. They all seem very similar to me. How do these three items differ and in what case would you use each of them? Is there some way that these three things can be combined into a single concept instead of three things that are all slightly different.


---

> I was pondering about adding this one but it's a clean way to achieve a component did mount without ignoring dependencies in useEffect

Ah so `useComponentDidMount(callback)` is near identical to `useEffect(callback, [])` except we're mutating things inside the render function instead of in an effect (As we call the callback immediately). Breaking that rule of "use useEffect for side effects" doesn't seem right to me.

I think passing in am empty array into useEffect is a lot cleaner and will be more understandable compared to using this new hook

src/utilities/use-component-did-mount.ts
```diff
@@ -0,0 +1,19 @@
+import {useRef} from 'react';
+import {EffectCallback} from '../types';
+import {useIsAfterInitialMount} from './use-is-after-initial-mount';
+
+/**
+ * Similarly to the lifecycle hook componentDidMount, useComponentDidMount
+ * will be invoked after the component has mounted, and only the initial mount.
+ * @param callback Defines a callback to use invoked once the component has
```
dleroux:
```suggestion
 * @param callback Defines a callback to invoke once the component has
```

src/utilities/use-component-did-mount.ts
dleroux: > Is there some way that these three things can be combined into a single concept instead of three things that are all slightly different.

That would be great. I might help clarify each of these hooks usage. Even part of the conversations with Andrew on this I have to look at the code of all 3 hooks to make sense of it.

Alternatively, we just stick to `useEffect(callback, [])` and escape the linter every time.

AndrewMusgrave: > I agree on the clarity, however I wonder if being clear in a mental model that we're moving away from is actually that useful? Over time peoples familiarity with the old model will fade. Also when the callback is executed doesn't map to when componentDidUpdate lifecycle in the class component world so I think conflating this may cause more harm than good - we're using the same name but it has a subtly different effect.

From what I understand the mental model for component did mount for the majority of people is it's a lifecycle hook that's called once after the components mounted. It'll never line up directly with a functional component but the logic will still be useful. A block of code that'll execute once, and have access to dom elements. What about maybe `useSingleExecution` or `useSingleEffect` 🤷‍♂ 

> We have useIsAfterInitialMount, this useComponentDidMount and a proposed useIsMountedRef from #2167. They all seem very similar to me. How do these three items differ and in what case would you use each of them? Is there some way that these three things can be combined into a single concept instead of three things that are all slightly different.

>That would be great. I might help clarify each of these hooks usage. Even part of the conversations with Andrew on this I have to look at the code of all 3 hooks to make sense of it.
> Alternatively, we just stick to useEffect(callback, []) and escape the linter every time.

The problem with combining them all into a single hook is at first glace them seem similar but they're effectively different  (some build on top of others) and combining them could cause unnecessary renders on the client.

If we're going with most common uses

useComponentDidMount
It would be great for network requests, for us, possibly dom measurements, incrementing a counter, focusing, etc.

useIsAfterInitialMount
Very useful for triggering a re-render to enhance server-side code with code that couldn't have run on the server, even building other hooks (for example useComponentDidMount)

useIsMounted
Great to be aware of the status of your component (this exists on class components once upon a time, this.mounted). Mostly used to avoid setting the state on unmounted components from async calls but also great for knowing the status of your component and executing code once (similar to componentDidMount) but doesn't require a re-render (i.e doesn't use dom nodes, window, document, etc)

> Ah so useComponentDidMount(callback) is near identical to useEffect(callback, []) except we're mutating things inside the render function instead of in an effect (As we call the callback immediately). Breaking that rule of "use useEffect for side effects" doesn't seem right to me.
> I think passing in am empty array into useEffect is a lot cleaner and will be more understandable compared to using this new hook

The great thing about `useComponentDidMount` is we are obeying the rules for react hooks and not risking stale values. The callback isn't executing within an effect, `useComponentDidMount` is using an effect (isAfterinitialMount) to determine the status of the component than invoking the callback outside of an effect.


One thing I really appreciate about hooks is we're able to abstract stateful logic and use it to build very composable functions to enhance our components.

dleroux:
We're adding a lot of hooks with no documentation except code and the bit of jsdocs comment. I wonder if we should start using `@example` or come up with another way to provide examples other than the tests.

AndrewMusgrave:
> We're adding a lot of hooks with no documentation except tests and the bit of jsdocs comments. I wonder if we should start using @example or come up with another way to provide examples other than the tests.

That's a fair point. I think we're ok at the moment since we're only using these internally however if we expose hooks we'll want usage examples to generate docs with.

PR #2204, opened 2019-09-27
[App Provider] Add features prop

tmlayton:

We want a way to toggle certain features within Polaris React. 

Adding a `features` prop to app provider and creating a hook for settings context.

### 🎩 checklist

### DIFF
--- a/src/components/AppProvider/AppProvider.tsx
+++ b/src/components/AppProvider/AppProvider.tsx
@@ -17,7 +17,7 @@ import {
   StickyManagerContext,
 } from '../../utilities/sticky-manager';
 import {LinkContext, LinkLikeComponent} from '../../utilities/link';
-import {Settings, SettingsContext} from '../../utilities/settings';
+import {Features, FeaturesContext} from '../../utilities/features';
 import {
   UniqueIdFactory,
   UniqueIdFactoryContext,
@@ -28,7 +28,7 @@ interface State {
   intl: I18n;
   appBridge: ReturnType<typeof createAppBridge>;
   link: LinkLikeComponent | undefined;
-  settings?: Settings;
+  features?: Features;
 }
 
 export interface AppProviderProps extends AppBridgeOptions {
@@ -38,8 +38,8 @@ export interface AppProviderProps extends AppBridgeOptions {
   linkComponent?: LinkLikeComponent;
   /** Custom logos and colors provided to select components */
   theme?: Theme;
-  /** For toggling settings */
-  settings?: Settings;
+  /** For toggling features */
+  features?: Features;
   /** Inner content of the application */
   children?: React.ReactNode;
 }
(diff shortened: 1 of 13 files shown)


src/components/AppProvider/AppProvider.tsx
```diff
@@ -95,25 +98,27 @@ export class AppProvider extends React.Component<AppProviderProps, State> {
   }
 
   render() {
-    const {theme = {logo: null}, children} = this.props;
+    const {theme = {logo: null}, settings, children} = this.props;
```
danrosenthal: Do we know if children will receive an update when this changes?

tmlayton: Yep, looks like it.

src/components/AppProvider/AppProvider.tsx
```diff
@@ -36,6 +37,8 @@ export interface AppProviderProps extends AppBridgeOptions {
   linkComponent?: LinkLikeComponent;
   /** Custom logos and colors provided to select components */
   theme?: Theme;
+  /** For toggling settings */
+  settings?: Settings;
```
tmlayton: Are we good with this prop name? Other options could be `features`, `flags`, etc.

danrosenthal: `flags` might be a better indicator of its intended use, but I'm comfortable with `settings`

BPScott: bikeshedding: flags or features feel better to me - "betas" while being our internal name for some of this stuff implies unfinished while settings feels a bit too permanent.

src/utilities/settings/hooks.tsx
```diff
@@ -0,0 +1,6 @@
+import {useContext} from 'react';
+import {SettingsContext} from './context';
+
+export function useSettings() {
+  return useContext(SettingsContext);
```
BPScott: Throw an exception if the context is falsy here - see the other hooks like sticky-manager for an example.

That'll ensure that this hook returns a value, and you won't need to do `settings != null` checks whenever you use the result of this hook

src/utilities/settings/types.ts
```diff
@@ -0,0 +1,3 @@
+export interface Settings {
+  [key: string]: boolean;
```
BPScott: Can this have an explicit shape? Telling consumers "you can toss anything into this object" doesn't feel that helpful when it comes to showing what's available.

danrosenthal: How much more explicit could you get without limiting flexibility? Or would you see us opening up the type each time we add a new key?

tmlayton: Chatted directly about this

BPScott: Context on that chat: I was thinking adding to the type each time you add a new key - doing that feels like a cheap price to pay to ensure we get "isAdafafaf is not a valid key" warnings if we do stuff like `{isAdafafaf} = useSettings();`.

But Tim made a point that we want to be obtuse about this for REASONS for the moment so I'm happy with this.

PR #2206, opened 2019-09-27
[Modal] Pass an activator element to the modal

LauraAubin:
Addresses part of https://github.com/Shopify/polaris-react/issues/795 and https://github.com/Shopify/polaris-react/issues/1932.

Whenever the modal is displayed, the browser focuses the modal and its contents. When the modal is closed (regardless if its done via the escape key or the x button/action elements) the browser does not refocus the button that initially opened it. Instead, we focus the entire body, which can be disorientating for screen readers.

This solution allows you to pass in an `activator` to the modal component, which can be any React element. We're not being super specific with the type of the `activator` because some modals are being opened by custom elements that we cannot always foresee.

Pros: This approach abstracts the creation of the ref away from the consumer entirely.

An alternative solution involves creating the ref yourself as explored here: https://github.com/Shopify/polaris-react/pull/2140.

## 

### How to 🎩

<details>

```jsx
import {Page, Button, Modal, TextContainer} from '../src';

  return (
    <Page title="Playground">
      <ModalExample />
    </Page>
  );
}

function ModalExample() {
  const [active, setActive] = useState(false);

  const handleChange = useCallback(() => setActive(!active), [active]);
  return (
    <div style={{height: '500px'}}>
      <Modal
        activator={<Button onClick={handleChange}>Open</Button>}
        open={active}
        onClose={handleChange}
        primaryAction={{
          content: 'Add Instagram',
          onAction: handleChange,
        }}
        secondaryActions={[
          {
            content: 'Learn more',
            onAction: handleChange,
          },
        ]}
      >
        <Modal.Section>
          <TextContainer>
            <p>
              Use Instagram posts to share your products with millions of
              people. Let shoppers buy from your store without leaving
              Instagram.
            </p>
          </TextContainer>
        </Modal.Section>
      </Modal>
    </div>
  );
}
```

</details>

### 🎩 checklist


src/components/TrapFocus/TrapFocus.tsx
```diff
@@ -84,9 +85,14 @@ export class TrapFocus extends React.PureComponent<TrapFocusProps, State> {
...
-        return focusLastFocusableNode(focusTrapWrapper);
+        return write(() => {
+          focusLastFocusableNode(focusTrapWrapper);
+        });
       }
-      focusFirstFocusableNode(focusTrapWrapper);
+
+      write(() => {
+        focusFirstFocusableNode(focusTrapWrapper);
+      });
```
LauraAubin: I had to wrap both of these in `write` to focus the button for all variations of using the mouse, escape button, and modal action item buttons.

src/components/Modal/Modal.tsx
```diff
@@ -77,6 +77,8 @@ export interface ModalProps extends FooterProps {
   onTransitionEnd?(): void;
   /** Callback when the bottom of the modal content is reached */
   onScrolledToBottom?(): void;
+  /** Element that opens and closes the modal */
```
dleroux:
```suggestion
  /** The element that activates the Modal */
```

src/components/Modal/Modal.tsx
```diff
@@ -306,8 +311,13 @@ class Modal extends React.Component<CombinedProps, State> {
 
     const animated = !instant;
 
+    const trigger = activator
+      ? React.cloneElement(activator, {ref: this.activatorRef})
```
dleroux: As per our discussion, it looks like you can wrap the activator in a `div` with a `ref` and use `focusFirstFocusableNode` to focus the button. As a result, there should be no need to add the `forwardRef` to the Button component :tada:

src/components/Modal/Modal.tsx
```diff
@@ -333,6 +339,10 @@ class Modal extends React.Component<CombinedProps, State> {
     if (this.focusReturnPointNode) {
       write(() => focusFirstFocusableNode(this.focusReturnPointNode, false));
     }
+
+    this.activatorRef &&
+      this.activatorRef.current &&
+      focusFirstFocusableNode(this.activatorRef.current);
```
AndrewMusgrave: We should be able to drop `this.activatorRef &&` since we initialize  activatorRef with `createRef` which is the "sort of equivalent" of `private activatorRef = {current: undefined}`

src/components/TrapFocus/TrapFocus.tsx
AndrewMusgrave: Is this for reflow or a delay? I.e, would [triggering a reflow](https://github.com/Shopify/polaris-react/blob/4199676447d8222560abbc766953c17c3ef92cd2/src/components/Frame/components/CSSAnimation/CSSAnimation.tsx#L78) work as well? Or is it for a delay? If it's a latter we might want to find a more reliable approach.

LauraAubin: This is for a delay. I don't think we need these `focusLastFocusableNode`/`first` calls with the modal since we're focusing the activator on our own, but I think these are called after we focus the activator, and so they change the focus to the body (I'm assuming because the modal has been removed and it can't find it?)

LauraAubin:
I figure since this is an accessibility improvement, we'll eventually want to enforce it. It feels like the best way to start seeing this pattern take effect is to implement it right away in the style guide examples, which is why I didn't make a separate example for this change.

We're also claiming that this works in the accessibility section so I feel like all of the examples should remain consistent with that.

> After a modal is closed, focus returns to the button that launched it

What do y'all think?

dleroux:
What do you think about making the ‘Basic example’ this one. Update all the examples to use this approach. Adding a new example for ‘External trigger’ which would be today’s basic example. In the description we can add a note about ensuring focus gets set on the activator when close for accessibility purposes?

src/components/TrapFocus/TrapFocus.tsx
dleroux: Do we need this change at all?

LauraAubin: Yup, without it we lose focus of the activator

src/components/Modal/README.md
```diff
@@ -684,6 +752,7 @@ See Apple’s Human Interface Guidelines and API documentation about accessibili
 
 - Modals use ARIA `role=”dialog”` to convey to screen reader users that they work like native dialog windows.
 - If you set the `title` prop to give the modal component a heading, then the `title` is used to label the dialog element with `aria-labelledby`. This helps to convey the purpose of the modal to screen reader users when it displays.
+- After a modal is closed, in order to return focus to the button that launched it, pass the button to the modal as an `activator`.
```
LauraAubin: This sentence is very similar to the one in the `keyboard support` section,

> After a modal is closed, focus returns to the button that launched it

I feel like it makes the activator concept stand out if its explained in two places. Removing it from the accessibility section would lose the implementation details, and removing it from the keyboard support section might make it seem like we stopped supporting this feature. What do y'all think?

src/components/TrapFocus/TrapFocus.tsx
LauraAubin: Looks like v.5 is removing the app bridge logic from the modal so we probably won't even need to test in an app if we wait to merge this PR after that release.

I've also asked about removing `focusReturnPointNode` entirely from Modal since it doesn't appear to be used. https://github.com/Shopify/polaris-react/pull/2150#issuecomment-543297427

src/components/Modal/Modal.tsx
```diff
@@ -70,13 +72,13 @@ const getUniqueID = createUniqueIDFactory('modal-header');
 
 class Modal extends React.Component<CombinedProps, State> {
   static Section = Section;
-  focusReturnPointNode: HTMLElement;
```
LauraAubin: @danrosenthal I'm removing this, along with [this line](https://github.com/Shopify/polaris-react/pull/2206/files#diff-eaf22cccaf0b1e13dc39209235b7229fL208-L209) that should never run.

UNRELEASED-v5.md
```diff
@@ -12,6 +12,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Add an activator prop to the `Modal` ([#2206](https://github.com/Shopify/polaris-react/pull/2206))
```
dleroux:
```suggestion
- Added an activator prop to `Modal` so that focus can be returned to it when the `Modal` is closed ([#2206](https://github.com/Shopify/polaris-react/pull/2206))
```

src/components/Modal/Modal.tsx
danrosenthal: I _think_ this should be fine, but I don't have a lot of context on how this code was used.

@tmlayton do you understand how this was used, and if it is now superfluous with these changes?

tmlayton: Yes, this is safe to remove. The previous versions of modal that used the EASDK and App Bridge had to cache the `activeElement` because the modal was rendered in the parent window (different DOM). Then when the modal closed focused was restored to that cached `activeElement`.

PR #2208, opened 2019-09-27
Add discoverability-action

amrocha:

We don't have much observability into what the effect of our changes is, and sometimes that can cause unexpected bugs. This tries to address that.

Adds a github action that runs yarn splash and comments on PRs with their splash zone

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### DIFF
--- a/.github/workflows/discoverability.yml
+++ b/.github/workflows/discoverability.yml
@@ -1,9 +1,6 @@
 name: Discoverability
 
-on:
-  pull_request:
-    branches:
-      - test-splash-action
+on: pull_request
 
 jobs:
   discoverability-action:
--- a/.github/workflows/discoverability.yml
+++ b/.github/workflows/discoverability.yml
@@ -13,7 +13,7 @@ jobs:
       run: yarn
 
     - name: Add a comment with known areas to test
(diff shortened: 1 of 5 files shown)


.github/workflows/discoverability.yml
```diff
@@ -0,0 +1,21 @@
...
+
+jobs:
+  discoverability-action:
+    name: Discoverability action
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+
+    - name: yarn
+      run: |
```
alex-page: The `|` is used when chaining multiple commands but we are running one. I don't think we need to encourage more commands?

src/components/Avatar/Avatar.tsx
```diff
@@ -120,7 +120,7 @@ export function Avatar({
         <svg className={styles.Svg} viewBox="0 0 48 48">
           <text
             x="50%"
-            y="50%"
+            y="60%"
```
alex-page: These changes can be reverted now.

PR #2229, opened 2019-10-03
[Color system] default features to empty object in test environment and export use features hook

danrosenthal:
Some cleanup following https://github.com/Shopify/polaris-react/pull/2204


BPScott:
Just snuck a some extra shuffling in there:

- keep contexts provided by AppProvider together, rather than
  intermingled with ones provided by Frame
- mimic the nesting of contexts used by AppProvider

These changes don't affect anything but it makes it easier to see at a
glance that PolarisTestProvider and AppProvider are basically the same

danrosenthal:
Codecov checks have been slow today. Going to merge without that passing.

PR #2241, opened 2019-10-04
[Decision Log] Add decision log for Autocomplete approach

dleroux:

Following the meeting, we had regarding using a [compound component approach for the Autocomplete](https://github.com/Shopify/polaris-react/issues/1794), the team decided it was time we added a decision log to Polaris. This is it.

### DIFF
--- a/documentation/decisions/01 - Using a Compound Component Pattern for Autocomplete.md
+++ b/documentation/decisions/01 - Using a Compound Component Pattern for Autocomplete.md
@@ -1,0 +1,28 @@
+# Using a compound component patter to build Autocomplete
+
+## Date
+
+October 3, 2019
+
+## Contributors
+
+- Andrew Musgrave
+- Chloe Rice
+- Dan Rosenthal
+- Daniel Leroux
+- Dominic McPhee
+- Sara Hill
+
+## Summary
+
+There is a need for Polaris to provide smaller more reusable components so that Polaris can be used by everyone without needing to fork component. The re-development of the Autocomplete component provided an opportunity for exploration.
+
+## Solution
+
+Following an (https://github.com/Shopify/polaris-react/issues/1794)[exploration] we decided to build the Autocomplete using a compound component pattern. We explored various ways to do this:
+
+1. Having the parent render its children via methods shared via Context: We opted against this because it defied the purpose of having smaller components and cause the component to renderer every time an option as added the Autocompletes results. For example, 100 results cause for 100 calls to the parent.
+
+2. Using `cloneElement`: We opted against this approach because since cloning happens at run time and TypeScript happens at compile time the benefits of using TypeScript were lost. The experience also felt less performant most likely due to the cloning at every render.
+
+3. Having the list of results as a child of the component and the activator text field as a prop: We chose this approach. Overall the approach felt right, the experience not lagging and renders we're kept to the elements being affected only.
--- a/documentation/decisions/01 - Using a Compound Component Pattern for Autocomplete.md
+++ b/documentation/decisions/01 - Using a Compound Component Pattern for Autocomplete.md
@@ -29,4 +29,6 @@ We opted against this for a couple of reasons:
 
 2. Using `cloneElement`: We opted against this approach because since cloning happens at run time and TypeScript happens at compile time the benefits of using TypeScript were lost. The experience also felt less performant most likely due to the cloning at every render.
 
-3. Having the list of results as a child of the component and the activator text field as a prop: We chose this approach. Overall the approach felt right, the experience not lagging and renders we're kept to the elements being affected only.
+3. Having the list of results as a child of the component and the activator text field as a prop
+
+We chose this approach. Overall the approach felt right, the experience felt snappy and renders were kept to the elements being affected only.
--- a/documentation/decisions/01 - Using a Compound Component Pattern for Autocomplete.md
+++ b/documentation/decisions/01 - Using a Compound Component Pattern for Autocomplete.md
@@ -1,4 +1,4 @@
-# Using a compound component patter to build Autocomplete
+# Using a compound component pattern to build Autocomplete
 
 ## Date
 
--- a/documentation/decisions/01 - Using a Compound Component Pattern for Autocomplete.md
+++ b/documentation/decisions/01 - Using a Compound Component Pattern for Autocomplete.md
@@ -27,7 +27,11 @@ We opted against this for a couple of reasons:
 - it defeated the purpose of having smaller components
 - it causes the component to re-render every time an option is added (one call to the parent render method for each option)
 
-2. Using `cloneElement`: We opted against this approach because since cloning happens at run time and TypeScript happens at compile time the benefits of using TypeScript were lost. The experience also felt less performant most likely due to the cloning at every render.
+2. Using `cloneElement`
+
+We opted against this approach because:
+- cloning occurs at run time, while TypeScript runs at compile time, meaning the benefits of using TypeScript were lost
+- the experience also felt less performant most likely due to the cloning at every render
 
 3. Having the list of results as a child of the component and the activator text field as a prop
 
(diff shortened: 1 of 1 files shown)


dleroux:
I've pinged everyone that was involved in the initial meeting that sparked the need to add a decision log to polaris-react. Let me know if this the level of information we are expecting from our logs.

documentation/decisions/01 - Using a Compound Component Pattern for Autocomplete.md
```diff
@@ -0,0 +1,28 @@
...
+- Andrew Musgrave
+- Chloe Rice
+- Dan Rosenthal
+- Daniel Leroux
+- Dominic McPhee
+- Sara Hill
+
+## Summary
+
+There is a need for Polaris to provide smaller more reusable components so that Polaris can be used by everyone without needing to fork component. The re-development of the Autocomplete component provided an opportunity for exploration.
```
danrosenthal:
```suggestion
There is a need for Polaris to provide smaller, more composable components so that Polaris can be extended without needing to fork components. The re-development of the Autocomplete component provided an opportunity for exploration of the compound component pattern.
```

documentation/decisions/01 - Using a Compound Component Pattern for Autocomplete.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Solution
+
+Following an (https://github.com/Shopify/polaris-react/issues/1794)[exploration] we decided to build the Autocomplete using a compound component pattern. We explored various ways to do this:
```
danrosenthal:
```suggestion
Following an [exploration](https://github.com/Shopify/polaris-react/issues/1794), we decided to build the Autocomplete using a compound component pattern. We explored various ways to do this.
```

documentation/decisions/01 - Using a Compound Component Pattern for Autocomplete.md
```diff
@@ -0,0 +1,28 @@
...
+
+1. Having the parent render its children via methods shared via Context: We opted against this because it defied the purpose of having smaller components and cause the component to renderer every time an option as added the Autocompletes results. For example, 100 results cause for 100 calls to the parent.
```
danrosenthal:
```suggestion
1. Having the child delegate rendering to its parent via methods shared through context

We opted against this for a couple of reasons:
- it defeated the purpose of having smaller components
- it causes the component to re-render every time an option is added (one call to the parent render method for each option)
```

documentation/decisions/01 - Using a Compound Component Pattern for Autocomplete.md
```diff
@@ -0,0 +1,28 @@
...
+
+2. Using `cloneElement`: We opted against this approach because since cloning happens at run time and TypeScript happens at compile time the benefits of using TypeScript were lost. The experience also felt less performant most likely due to the cloning at every render.
```
danrosenthal:
```suggestion
2. Using `cloneElement`

We opted against this approach because:
- cloning occurs at run time, while TypeScript runs at compile time, meaning the benefits of using TypeScript were lost
- the experience also felt less performant most likely due to the cloning at every render
```

documentation/decisions/01 - Using a Compound Component Pattern for Autocomplete.md
```diff
@@ -0,0 +1,28 @@
...
+
+3. Having the list of results as a child of the component and the activator text field as a prop: We chose this approach. Overall the approach felt right, the experience not lagging and renders we're kept to the elements being affected only.
```
danrosenthal:
```suggestion
3. Having the list of results as a child of the component and the activator text field as a prop

We chose this approach. Overall the approach felt right, the experience felt snappy and renders were kept to the elements being affected only.
```

documentation/decisions/01 - Using a Compound Component Pattern for Autocomplete.md
danrosenthal: This might need one more sentence around why composable components allow greater flexibility. IE: not needing to add a prop for any new configuration, as it can be handled as children. Adding an `Icon` to an `Option` feels like a good example to explain this, since this approach would not require adding/maintaining another prop.

documentation/decisions/01 - Using a Compound Component Pattern for Autocomplete.md
danrosenthal: > - it defeated the purpose of having smaller components

This needs more detail. Why did it defeat the purpose of having smaller components?

documentation/decisions/01 - Using a Compound Component Pattern for Autocomplete.md
danrosenthal: This leaves out how context is used to share state with children. That's an important detail for this approach.

dleroux:
Feel free to add to this PR as well, as I feel it's more of a collaboration based on our discussion more than my PR per se.

documentation/decisions/01 - Using a Compound Component Pattern for Autocomplete.md
```diff
@@ -0,0 +1,38 @@
+# Using a compound component patter to build Autocomplete
```
chloerice:
```suggestion
# Using a compound component pattern to build Autocomplete
```

documentation/decisions/01 - Using a Compound Component Pattern for Autocomplete.md
```diff
@@ -0,0 +1,38 @@
...
+
+2. Using `cloneElement`
+
+We opted against this approach because:
+- cloning occurs at run time, while TypeScript runs at compile time, meaning the benefits of using TypeScript were lost
+- the experience also felt less performant most likely due to the cloning at every render
+
+3. Having the list of results as a child of the component and the activator text field as a prop
+
+We chose this approach. Overall the approach felt right, the experience felt snappy and renders were kept to the elements being affected only.
```
AndrewMusgrave: It's also very similar to an HTML approach without a lot of magic happening behind the scenes. `Stack` has bit us a couple of times 😬

documentation/decisions/01 - Using a Compound Component Pattern for Autocomplete.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Solution
+
+Following an [exploration](https://github.com/Shopify/polaris-react/issues/1794), we decided to build the Autocomplete using a compound component pattern. We explored various ways to do this.
+
+1. Having the child delegate rendering to its parent via methods shared through context
+
+We opted against this for a couple of reasons:
+- it defeated the purpose of having smaller components
+- it causes the component to re-render every time an option is added (one call to the parent render method for each option)
```
AndrewMusgrave: I wonder if we can go into more detail with this approach? It'll vary depending on architecture and complexity. A child `re-render` won't necessarily cause siblings to re-render. Here's a [sandbox example](https://codesandbox.io/s/amazing-river-6lomm)

documentation/decisions/01 - Using a Compound Component Pattern for Autocomplete.md
```diff
@@ -0,0 +1,38 @@
...
+
+2. Using `cloneElement`
+
+We opted against this approach because:
+- cloning occurs at run time, while TypeScript runs at compile time, meaning the benefits of using TypeScript were lost
+- the experience also felt less performant most likely due to the cloning at every render
+
+3. Having the list of results as a child of the component and the activator text field as a prop
```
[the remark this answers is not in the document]
dleroux: As we discussed we'll need to wire up the context but yes, I'll mention it.

dleroux:
Update: this is not forgotten. Focusing on the prototype which will help solidify this decision.

dleroux:
Going to close this for now until the project picks up again.

PR #2248, opened 2019-10-07
[Popover] Add ariaHaspopup

### LINKED ISSUE #2187: [Popover] Action list accessibility enhancements
## Feature request summary

Add `aria-haspopup` as a prop to the popover, if the intention is to create an action list (`aria-haspopup=”menu”`).

Partially fixes: https://github.com/Shopify/polaris-ux/issues/161

## Rationale

When using a `Popover` with `Action list`, the behaviour of the component should be different from a generic popover: it should behave like a menu and support interaction in a manner that is consistent with user expectations about other menus of that type.

When navigating a menu using a keyboard, it's expected to use "up" and "down" arrow keys instead of "tab" to navigate between selections.

Generic popovers (anything else but `Action list`) should be implemented as a generic expandable/collapsible buttons and containers. In this case, it should not have `aria-haspopup` to take away ambiguity. Keep `aria-expanded`.

We should probably clarify that on the component documentation on the style guide.
cc @sadiesaurus 

---

AndrewMusgrave:

Fixes #2187

* Add prop
* Consume prop
* Use prop were it's needed in doc examples
* Add accessibility docs

### How to 🎩

<details>

```jsx
function PopoverWithActionListExample() {

    [],
  );

  const activator = (
      More actions
    </Button>
  );

  return (
      <Popover
        active={popoverActive}
        activator={activator}
        onClose={togglePopoverActive}
        ariaHaspopop
      >
      </Popover>
    </div>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Popover/README.md
+++ b/src/components/Popover/README.md
@@ -254,6 +254,7 @@ function PopoverFormExample() {
         active={popoverActive}
         activator={activator}
         onClose={togglePopoverActive}
+        ariaHaspopup={false}
         sectioned
       >
         <FormLayout>
@@ -338,6 +339,7 @@ function PopoverLazyLoadExample() {
           active={popoverActive}
           activator={activator}
           onClose={togglePopoverActive}
+          ariaHaspopup={false}
         >
           <Popover.Pane onScrolledToBottom={handleScrolledToBottom}>
             <ResourceList items={staffList} renderItem={renderItem} />
@@ -414,6 +416,10 @@ See Apple’s Human Interface Guidelines and API documentation about accessibili
 <!-- content-for: web -->
 
 Popovers usually contain an [option list](https://polaris.shopify.com/components/lists-and-tables/option-list) or an [action list](https://polaris.shopify.com/components/actions/action-list), but can also contain other controls or content.
+
+A popover can contain many numerous types of content. Whether it's a menu, grid or something entirely different! When `aria-expanded` is applied to an element, `aria-haspopup` will default to `menu`. To assist screen readers you'll need to pass `ariaHaspopup` to `Popover`.
+
+Popovers can contain many types of content. If a popover contains a complex feature such as an ARIA menu, grid, listbox, or similar feature, then use the `ariaHaspopup` prop. This adds `aria-haspopup` to the button, which is helpful information for people who use screen readers.
 
 ### Keyboard support
 
--- a/src/components/Popover/README.md
+++ b/src/components/Popover/README.md
@@ -419,8 +419,6 @@ Popovers usually contain an [option list](https://polaris.shopify.com/components
 
 A popover can contain many numerous types of content. Whether it's a menu, grid or something entirely different! When `aria-expanded` is applied to an element, `aria-haspopup` will default to `menu`. To assist screen readers you'll need to pass `ariaHaspopup` to `Popover`.
 
-Popovers can contain many types of content. If a popover contains a complex feature such as an ARIA menu, grid, listbox, or similar feature, then use the `ariaHaspopup` prop. This adds `aria-haspopup` to the button, which is helpful information for people who use screen readers.
-
 ### Keyboard support
 
 - When a popover opens, focus moves to the popover container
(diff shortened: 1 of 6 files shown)


src/components/Popover/set-activator-attributes.ts
```diff
@@ -0,0 +1,23 @@
+import {AriaAttributes} from 'react';
```
AndrewMusgrave: It seems this chunk of code was never tested. When I was trying to add a few tests I kept getting stale dom nodes. I broke this utility into its own module so we can easily mock it in `Popover`.

src/components/Popover/tests/Popover.test.tsx
```diff
@@ -1,9 +1,33 @@
 import React from 'react';
 import {mountWithAppProvider, findByTestID} from 'test-utilities/legacy';
 import {Popover} from '../Popover';
+import * as SetActivatorAttributes from '../set-activator-attributes';
```
AndrewMusgrave: Importing like this to spyon the module

src/components/Popover/README.md
```diff
@@ -415,6 +417,8 @@ See Apple’s Human Interface Guidelines and API documentation about accessibili
 
 Popovers usually contain an [option list](https://polaris.shopify.com/components/lists-and-tables/option-list) or an [action list](https://polaris.shopify.com/components/actions/action-list), but can also contain other controls or content.
 
+A popover can contain many numerous types of content. Whether it's a menu, grid or something entirely different! When `aria-expanded` is applied to an element, `aria-haspopup` will default to `menu`. To assist screen readers you'll need to pass `ariaHaspopup` to `Popover`.
```
dpersing: A few thoughts here:

- I'd like to remove the "Whether it's..." fragment, since it doesn't reference other patterns or components.
- In reference to `aria-expanded` and `aria-haspopup`, it's [a bit more complicated](https://www.w3.org/TR/wai-aria-1.1/#aria-haspopup). I think we could simplify this by saying something like, "The `ariaHaspopup ` prop adds`aria-haspopup="true"` to the button. This should only be used when the popover contains a complex feature like an ARIA menu, grid, listbox, or similar features."
- We typically want to use people-centered language, so we'd want to say "to assist people who use screen readers" instead of "to assist screen readers".

Let's chat about this one on Tuesday, and involve @sadiesaurus if she's available.

AndrewMusgrave: Hi @sadiesaurus , what are your thoughts on this?

sadiesaurus: Heya Andrew! I agree with Devon's points. I'll make a suggestion for this paragraph which you can feel free to edit.

sadiesaurus: Here's my stab at a rewrite! Though, I'm not certain if I've accurately captured the context, so let me know if something is wrong or missing. 

```suggestion
Popovers can contain many types of content. If a popover contains a complex feature such as an ARIA menu, grid, listbox, or similar feature, then use the `ariaHaspopup` prop. This adds `aria-haspopup="true"` to the button, which is helpful information for people who use screen readers. 
```

AndrewMusgrave: It looks great! I made one small adjustment since you can provide more values than "true" to `ariaHasPopup` (`boolean | 'false' | 'true' | 'menu' | 'listbox' | 'tree' | 'grid' | 'dialog'`)

AndrewMusgrave:
cc/ @dpersing

src/components/Popover/README.md
```diff
@@ -415,6 +417,10 @@ See Apple’s Human Interface Guidelines and API documentation about accessibili
...
+
+Popovers can contain many types of content. If a popover contains a complex feature such as an ARIA menu, grid, listbox, or similar feature, then use the `ariaHaspopup` prop. This adds `aria-haspopup` to the button, which is helpful information for people who use screen readers.
```
sadiesaurus: Was this section meant to be removed?

AndrewMusgrave: It was a bad rebase, thanks for catching this 🙏

Issue #2197, opened 2019-09-25
[Color system] Develop confidence in color factory variant generation by applying colors to a subset of components and sass mixins


danrosenthal:
Using the generated CSS Custom Properties from the ThemeProvider/Color factory, we need to apply the color system to a subset of components and critical mixins in order to gain confidence on the color system.

**Those components are:**
- [x] all typographic components (largely achieved through mixins)
- [x] `TextField`
- [x] `Button`
  - [x] `primary`
  - [x] `basic`
  - [x] `plain`
  - [x] `destructive`
- [x] `Card`
  - [x] `shadow` mixin
- [x] `Frame`
- [x] `TopBar`
  - [ ] transparency for `Navigation` and `Menu` activators
- [x] `Navigation`
  - [x] Backdrop on mobile view to account for transparency
- [x] `Banner`
  - [ ] icons and primary actions in banners need to be addressed
- [x] `Badge`
  - [x] success and info are visually very similar, so the surface values there need some work 
- [x] `Link`
- [x] `ActionList`
- [x] `Popover`

This subset of the system gives us a nice cross-section of different color usage across our components helping us gain confidence on the widest possible application of the color system with the least amount of work.

The remainder of the color application work will be done as part of https://github.com/Shopify/polaris-react/issues/2040.

Additionally, we should use this time to gain confidence on tricky mixins like:

- `color-icon`
- `recolor-icon`

danrosenthal:
This issue can now be closed, as we have completed all critical items from the checklist.

PR #2252, opened 2019-10-07
[Button][Color system] Apply color system

tmlayton:
Co-authored-by: Dan Rosenthal <dan.rosenthal@shopify.com>

Applying the color system to button. Addresses #2197 

Makes use of the new CSS custom properties for color and applies them to button with a fallback to the current value to maintain backwards compatibility. This change should be invisible unless enabling the theming feature.

<summary>

<h3>Playground</h3>

<details>

```tsx
import {Page, Card, List, AppProvider, Checkbox, Stack} from '../src';

const initDarkMode =
  window && window.matchMedia('(prefers-color-scheme: dark)').matches;

  const [darkMode, setDarkMode] = useState(initDarkMode);
  const [masterbrand, setMasterbrand] = useState(false);
  const handleChange = useCallback((newChecked) => setDarkMode(newChecked), []);
  const masterbrandToggle = useCallback(
    (newChecked) => setMasterbrand(newChecked),
    [],
  );

  return (
    <AppProvider
      i18n={{}}
      theme={{
        colors: {
          topBar: {
            background: '#357997',
          },
        },
        // eslint-disable-next-line babel/camelcase
        UNSTABLE_colors: {surface: darkMode ? '#111213' : '#FAFAFA'},
      }}
      features={{unstableGlobalTheming: masterbrand}}
    >
      <Page
        title="Playground"
        titleMetadata={
          <div style={{paddingLeft: 32}}>
            <Stack distribution="trailing" alignment="center">
              <Checkbox
                label="Masterbrand"
                checked={masterbrand}
                onChange={masterbrandToggle}
              />
              <Checkbox
                label="Dark Mode"
                checked={darkMode}
                onChange={handleChange}
                disabled={!masterbrand}
              />
            </Stack>
          </div>
        }
      >
        <Card
          sectioned
          title="Shipment 1234"
          secondaryFooterActions={[{content: 'Edit shipment'}]}
          primaryFooterAction={{content: 'Add tracking number'}}
        >
          <List>
            <List.Item>1 × Isis Glass, 4-Pack</List.Item>
            <List.Item>1 × Anubis Cup, 2-Pack</List.Item>
          </List>
        </Card>
      </Page>
    </AppProvider>
  );
}
```

</details>
</summary>

### 🎩 checklist

### DIFF
--- a/src/styles/_common.scss
+++ b/src/styles/_common.scss
@@ -17,6 +17,7 @@
 @import './foundation/shadows';
 @import './foundation/typography';
 @import './foundation/z-index';
+@import './foundation/focus-ring';
 
 @import './shared/accessibility';
 @import './shared/breakpoints';
--- a/src/styles/foundation/_focus-ring.scss
+++ b/src/styles/foundation/_focus-ring.scss
@@ -3,26 +3,24 @@
 ///
 
 @mixin focus-ring($size: 'base') {
-  $offset: 1px;
-  $stroke: 2px;
+  $offset: rem(1px);
+  $stroke: rem(2px);
   $position: $offset + $stroke;
-  $newRadius: 4px;
-  $newWideRadius: 8px;
   position: relative;
 
   &::after {
     content: var(--p-focus-ring-content, none);
     position: absolute;
-    top: rem(-$position);
-    right: rem(-$position);
-    bottom: rem(-$position);
-    left: rem(-$position);
+    top: -$position;
+    right: -$position;
+    bottom: -$position;
+    left: -$position;
     display: block;
-    border: rem($stroke) solid var(--p-interactive-focus);
+    border: $stroke solid var(--p-interactive-focus);
     @if $size == 'wide' {
-      border-radius: rem($newWideRadius + $position);
+      border-radius: calc(var(--p-border-radius-wide) + #{$position});
     } @else {
-      border-radius: rem($newRadius + $position);
+      border-radius: calc(var(--p-border-radius-base) + #{$position});
     }
   }
 }
--- a/src/styles/foundation/_focus-ring.scss
+++ b/src/styles/foundation/_focus-ring.scss
@@ -20,6 +20,4 @@
       border-radius: calc(var(--p-border-radius-base) + 3px);
     }
   }
-
-  @include high-contrast-button-outline;
 }
--- a/src/styles/foundation/_focus-ring.scss
+++ b/src/styles/foundation/_focus-ring.scss
@@ -3,21 +3,26 @@
 ///
 
 @mixin focus-ring($size: 'base') {
+  $offset: 1px;
+  $stroke: 2px;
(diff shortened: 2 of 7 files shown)


src/styles/foundation/_focus-ring.scss
```diff
@@ -0,0 +1,25 @@
...
+    content: var(--p-focus-ring-content, none);
+    position: absolute;
+    top: rem(-3px);
+    right: rem(-3px);
+    bottom: rem(-3px);
+    left: rem(-3px);
+    display: block;
+    border: rem(2px) solid var(--p-interactive-focus);
+    @if $size == 'wide' {
+      border-radius: calc(var(--p-border-radius-wide) + 3px);
```
tmlayton: Do we want to convert our borders and other thin lines and outer detail to rems?

src/styles/_common.scss
```diff
@@ -17,6 +17,7 @@
 @import './foundation/shadows';
 @import './foundation/typography';
 @import './foundation/z-index';
+@import './foundation/focus-ring';
```
tmlayton: Is this the right spot?

BPScott: Kiiinda. `_common.scss` is only used privately so this isn't exposing anything new, but I think it's confusing to add a file to the foundations folder and it not be referenced by `foundation.scss` (which is an entry point for the public API).

Can you move the `focus-ring.scss` file into the common folder? That'd help make it clearer that it is a new private file distinct from the public api files (foundation and shared)

danrosenthal: Yeah, sorry, in the initial PR, I added this to a new `src/styles/common` directory. But it sounds like that won't work from Ben's comment.

src/styles/foundation/_focus-ring.scss
danrosenthal: Yeah, we do. I think we'll need to add a JavaScript `rem` function.

tmlayton: In this case, since we know the value and this focus ring is only used with theming we could hardcode for now.

tmlayton: Never mind, decided to convert in JS

src/styles/shared/_buttons.scss
```diff
@@ -37,96 +36,121 @@
...
 
   &:focus {
-    border-color: color('indigo');
+    border-color: var(--p-transparent, color('indigo'));
     outline: 0;
-    box-shadow: 0 0 0 1px color('indigo');
+    box-shadow: var(--p-none, 0 0 0 1px color('indigo'));
 
+    @include focus-ring;
     @include high-contrast-button-outline;
```
tmlayton: @dpersing ping me when you’re back and let’s talk about resolving the new focus ring styles with the high contrast outline. I think we can achieve both in a single CSS approach and mixin.

PR #2255, opened 2019-10-07
[Popover ] Fixed incorrect element being focused when closed

### LINKED ISSUE #2188: [Popover] Keyboard navigation enhancements
## Feature request summary

Currently, tabbing forward out of popover closes it, sending focus back to the button that triggers it. The expected behaviour is to send focus to the forward element instead.

Partially fixes: https://github.com/Shopify/polaris-ux/issues/161

## Rationale

**General keyboard Interaction** 
With focus on the button (trigger):
Enter or Space: opens the popover and places focus on the first menu item.
Tab: Moves focus to the first menu item.
Tab (out of popover): Moves focus to the next element on the page.
Shift Tab: Moves focus to the previous menu item or element on the page.

Try the above instructions on https://www.shopify.com/. It behaves as expected.

---

AndrewMusgrave:

Fixes #2188 

* When the popover is closed it'll focus the next focusable element. If a next focusable element is not present the activator will be focused
* Create two focus utilities
* Add tests for utilities
* Add missing focus tests

### How to 🎩

<details>
<summary>With another focusable element</summary>

```jsx
import {Popover, Button, ActionList} from '../src';

    [],
  );

  const activator = (
      More actions
    </Button>
  );

  return (
      <Popover
        active={popoverActive}
        activator={activator}
        onClose={togglePopoverActive}
      >
      </Popover>
      <button>next focusable target</button>
    </div>
  );
}
```

</details>

<details>
<summary>Without another focusable element</summary>

```jsx
import {Popover, Button, ActionList} from '../src';

    [],
  );

  const activator = (
      More actions
    </Button>
  );

  return (
      <Popover
        active={popoverActive}
        activator={activator}
        onClose={togglePopoverActive}
      >
      </Popover>
    </div>
  );
}
```

</details>

### TODO

* Add changelog

### 🎩 checklist

### DIFF
--- a/src/utilities/focus.ts
+++ b/src/utilities/focus.ts
@@ -1,5 +1,42 @@
+import {FOCUSABLE_SELECTOR} from '@shopify/javascript-utilities/focus';
+import {isElementInViewport} from './is-element-in-viewport';
+
+type Filter = (element: Element) => void;
+
 export function handleMouseUpByBlurring({
   currentTarget,
 }: React.MouseEvent<HTMLAnchorElement | HTMLButtonElement>) {
   currentTarget.blur();
+}
+
+export function nextFocusableNode(
+  node: HTMLElement,
+  filter?: Filter,
+): HTMLElement | Element | null {
+  const allFocusableElements = [
+    ...document.querySelectorAll(FOCUSABLE_SELECTOR),
+  ];
+  const sliceLocation = allFocusableElements.indexOf(node) + 1;
+  const focusableElementsAfterNode = allFocusableElements.slice(sliceLocation);
+
+  for (const focusableElement of focusableElementsAfterNode) {
+    if (
+      isElementInViewport(focusableElement) &&
+      (!filter || (filter && filter(focusableElement)))
+    ) {
+      return focusableElement;
+    }
+  }
+
+  return null;
+}
+
+export function focusNextFocusableNode(node: HTMLElement, filter?: Filter) {
+  const nextFocusable = nextFocusableNode(node, filter);
+  if (nextFocusable && nextFocusable instanceof HTMLElement) {
+    nextFocusable.focus();
+    return true;
+  }
+
+  return false;
 }
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -4,6 +4,7 @@
 
 ### Enhancements
 
+- Updated `Popover` to focus the correct element when closed ([#2255](https://github.com/Shopify/polaris-react/pull/2255))
 - Updated the type of the `title` prop in `ChoiceList` from `string` to `ReactNode` ([#2355](https://github.com/Shopify/polaris-react/pull/2355))
 - Added `disabled` prop to `Filters` component ([2389](https://github.com/Shopify/polaris-react/pull/2389))
 
--- a/src/components/Popover/Popover.tsx
+++ b/src/components/Popover/Popover.tsx
@@ -1,12 +1,11 @@
 import React from 'react';
 import {createUniqueIDFactory} from '@shopify/javascript-utilities/other';
-import {
-  focusFirstFocusableNode,
-  findFirstFocusableNode,
-} from '@shopify/javascript-utilities/focus';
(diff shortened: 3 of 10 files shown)


src/utilities/focus.ts
```diff
@@ -1,5 +1,39 @@
...
   currentTarget.blur();
 }
+
+export function nextFocusableNode(
+  node: HTMLElement,
+): HTMLElement | Element | null {
+  let nextElement = node.nextElementSibling;
+
+  while (nextElement) {
+    if (nextElement.matches(FOCUSABLE_SELECTOR)) return nextElement;
```
AndrewMusgrave: I'm going with the `matches` approach that rather a tabIndex so we can ignore disabled elements

src/utilities/focus.ts
```diff
@@ -1,5 +1,39 @@
...
+
+    if (nextElement instanceof HTMLElement) {
```
AndrewMusgrave: findFirstFocusableNode require `HTMLElement` rather than `Element`. Although `nextElementSibling` won't return text nodes so I'm not sure we'll ever get an `Element`

src/utilities/focus.ts
```diff
@@ -1,5 +1,39 @@
...
+
+    nextElement = nextElement.nextElementSibling;
+  }
+
+  return nextElement;
+}
+
+export function focusNextFocusableNode(node: HTMLElement) {
+  const nextFocusable = nextFocusableNode(node);
+  if (nextFocusable && nextFocusable instanceof HTMLElement) {
```
AndrewMusgrave: The `Element` interface doesn't contain `focus` so we need this check

src/utilities/focus.ts
```diff
@@ -1,5 +1,39 @@
...
+    nextFocusable.focus();
+    return true;
```
AndrewMusgrave: Added a boolean return rather than void so we can determine if an element was focused or not

src/utilities/focus.ts
```diff
@@ -1,5 +1,39 @@
...
 export function handleMouseUpByBlurring({
   currentTarget,
 }: React.MouseEvent<HTMLAnchorElement | HTMLButtonElement>) {
   currentTarget.blur();
 }
+
+export function nextFocusableNode(
+  node: HTMLElement,
+): HTMLElement | Element | null {
+  let nextElement = node.nextElementSibling;
```
dleroux: This logic will only work if the nextFocusableElement same tree level or lower correct? The following would not pass

```
 <React.Fragment>
    <div><Popover active activator={<div />} onClose={noop} /></div>
     <button id={id} />
</React.Fragment>
```

I think it's fine since the activator will get focus in the end. I can't think of an inexpensive way to do this either. What do you think?

AndrewMusgrave:
> I'm not sure if this is a new issue or not (since our previous examples didn't have additional elements to manage focus around, but I found while testing this that focus is does not appear to be moved to the popover on iOS when the button is activated. When using VoiceOver with Safari or Chrome on iOS, I had to swipe to the next focusable target button before I could swipe to the popover contents. Can we confirm if this was an existing issue, or if it's introduced by this update? If it's not new, I'll open a new issue, which we should address as soon as possible, since it impacts popover usage across mobile.

@dleroux and I spent some time on iOS voiceover and it appears it's an existing issue. Oddly enough we still have the correct focus order, however it seems voice was following the dom order, and not respecting the currently focused element. Is this many a voiceover bug?

> I wonder if we can also take a look at web to see if anyone is overriding our current focus management logic for popovers, to make sure this update won't cause conflicts?

It would be pretty difficult to check all the popovers since there're hundreds, maybe thousands 😬 however I tested home, and a few other pages as well and everything looked 👍

dpersing:
Created issue #2387 to capture the issue with VoiceOver on iOS.

dleroux:
Is this ready to review again @AndrewMusgrave?

dpersing:
@dleroux Per discussion in Slack last week, yes!

PR #2256, opened 2019-10-07
Fixed spacing

AndrewMusgrave:

Noticed an extra newline and indentation


kaelig:
Unrelated: Any reason why none of those changelog items point to a PR?

dleroux:
I don't think we have a strict convention. Since this is causing confusion, I wonder if we aren't better repeating each item in a separate entry, repeating the PR? @danrosenthal what do you think?

danrosenthal:
I would prefer they stay nested under the item, or deleted in favor of just the top level item

kaelig:
I'm with @danrosenthal on this one, let's keep it as is.

PR #2266, opened 2019-10-08
[ActionMenu] doesn't render MenuActions if no actions or actions array is empty

416serg:

Fixes https://github.com/Shopify/polaris-react/issues/2220

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -10,30 +10,17 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
-- Added `totalItemsCount` prop to `ResourceList` component ([2233](https://github.com/Shopify/polaris-react/pull/2233))
-- Prevents the `Header` primary action label on `Page` from wrapping when the title is too long ([2262](https://github.com/Shopify/polaris-react/pull/2262))
+- Added `pressed` state to `Button` ([#2148](https://github.com/Shopify/polaris-react/pull/2148))
 
 ### Bug fixes
 
-- Doesn't render `MenuActions` if no actions are passed to an `actionGroups` item inside `Page` ([2266](https://github.com/Shopify/polaris-react/pull/2266))
-- Fixed issue with `Stack` where a `Stack.Item` was not getting a minimum width ([2273](https://github.com/Shopify/polaris-react/pull/2273))
-- Fixed issue with `Filters` applying inconsistent border styles to sibling filters and when
-  there is only one filter in the filter list ([2284](https://github.com/Shopify/polaris-react/pull/2284))
-
-### Documentation
-
-- Added accessibility documentation for the drop zone component ([#2243](https://github.com/Shopify/polaris-react/pull/2243))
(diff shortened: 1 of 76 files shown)


src/components/ActionMenu/components/MenuGroup/MenuGroup.tsx
```diff
@@ -30,6 +30,10 @@ export class MenuGroup extends React.Component<MenuGroupProps, never> {
       icon,
     } = this.props;
 
+    if (!actions || (actions && actions.length === 0)) {
```
LauraAubin: I believe you only need to check `actions.length === 0`.

416serg: I believe, since it's not required, someone can pass 

```
{
  title: "Test",
}
```

into `actionGroups` prop and it will still render the element, see https://codesandbox.io/embed/sleepy-austin-4kkxt

LauraAubin: Both `title` and `actions` are required in Typescript and raise an error,

https://github.com/Shopify/polaris-react/blob/5429427022bf09eba8c06007005e6711bf165b3a/src/types.ts#L188-L192

But there doesn't appear to be any errors/warnings in JS. Probably best to leave in the extra checks then 👍

AndrewMusgrave: > I believe, since it's not required, someone can pass

As Laura mentioned above it is a required prop, and using the API without `actions` isn't support, and we usually rely on typescript to catch these errors.

Given that, `if ( actions.length === 0)) return null;` or `if (!actions.length)) return null;` should suffice for a check. With that being said, if you were looking to use the API without actions, we could look into that. What's your use case?

416serg: No particular use case, just wanted to add a safety net for developers that are using Polaris without TS that could potentially add an `actionGroups` item without actions thinking it's a secondary action 🤷‍♂ but since it seems like an overkill, updated to a length check 👍 Thanks for your input!

UNRELEASED.md
```diff
@@ -14,6 +14,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- Doesn't render `MenuActions` if no actions are passed to an `actionGroups` item inside `Page` ([2266](https://github.com/Shopify/polaris-react/pull/2266))
```
LauraAubin: A small nitpick,

```suggestion
- `ActionMenu` doesn't render `MenuActions` if no actions are passed to an `actionGroups` item inside `Page` ([2266](https://github.com/Shopify/polaris-react/pull/2266))
```

416serg: I'm not much of a writer 🤦‍♂

PR #2268, opened 2019-10-09
[Color system] don't output legacy custom properties when global theming is enabled

danrosenthal:


src/utilities/theme/utils.ts
```diff
@@ -14,7 +14,6 @@ export function buildCustomProperties(
 ): CustomPropertiesLike {
   return globalTheming
     ? {
-        ...buildLegacyColors(themeConfig),
```
BPScott: If this is gone do we need the object destructuring? seems like we're creating a new object for nothing.

```
return globalThemeing ? ...buildColors(themeConfig) : buildLegacyColors(themeConfig)
```

PR #2269, opened 2019-10-09
[Color system] move background color and text color from body to theme provider node, and apply color system

danrosenthal:
Addresses #2197

Prior to this pull request, Polaris set a background and a base font color on the body node. With global theming provided by the `ThemeProvider`, this is no longer possible, since the body is outside of the scope of the custom properties added to the theme provider node. This moves global typographic and background styles to the theme provider node, which wraps an entire app thanks to its position in the app provider.

## Impact

This will result in any applications that were using Polaris outside of a `Frame` to lose their full height colored background. This impact can be avoided by wrapping any application in an empty `Frame` component, as it creates a full height container.

### Chroma

This also has the impact of our chroma stories no longer having a full height background. 

(This has resulted in an accessibility failure on disabled `Select`. We should be able to safely ignore that failure, as disabled interface elements are not subject to WCAG contrast guidelines, and because the failure is erroneous.)

|disabled `Select`|
|-|
|Screen Shot 2019-10-09 at 10 53 22 AM|

### Typography

The preferable alternative would be to have each component declare its own typographic styles, and not rely on global styles and the cascade. But, that would be a large undertaking, and we want to put minimal fingerprints on existing code. Moving it from a global style on the body to a global style on the theme provider felt like a decent, though less idealistic, alternative.

### Background color on body

We could set the background color of the body element programmatically from the `ThemeProvider` component to be surface background once that component mounts.

### DIFF
--- a/src/components/ThemeProvider/ThemeProvider.tsx
+++ b/src/components/ThemeProvider/ThemeProvider.tsx
@@ -1,4 +1,4 @@
-import React, {useMemo} from 'react';
+import React, {useMemo, useEffect} from 'react';
 import {
   ThemeContext,
   ThemeConfig,
@@ -7,8 +7,6 @@ import {
 } from '../../utilities/theme';
 import {themeProvider} from '../shared';
 import {useFeatures} from '../../utilities/features';
-
-import styles from './ThemeProvider.scss';
 
 interface ThemeProviderProps {
   /** Custom logos and colors provided to select components */
@@ -28,13 +26,19 @@ export function ThemeProvider({
     [unstableGlobalTheming, themeConfig],
   );
 
+  // We want these values to be `null` instead of `undefined` when not set.
+  // Otherwise, setting a style property to `undefined` does not remove it from the DOM.
+  const backgroundColor = customProperties['--p-surface-background'] || null;
+  const color = customProperties['--p-text-on-surface'] || null;
+
+  useEffect(() => {
+    document.body.style.backgroundColor = backgroundColor;
+    document.body.style.color = color;
+  }, [backgroundColor, color]);
+
   return (
     <ThemeContext.Provider value={theme}>
-      <div
-        style={customProperties}
-        {...themeProvider.props}
-        className={styles.ThemeProvider}
-      >
+      <div style={customProperties} {...themeProvider.props}>
         {children}
       </div>
     </ThemeContext.Provider>
--- a/src/components/ThemeProvider/ThemeProvider.tsx
+++ b/src/components/ThemeProvider/ThemeProvider.tsx
@@ -8,6 +8,8 @@ import {
 import {themeProvider} from '../shared';
 import {useFeatures} from '../../utilities/features';
 
+import styles from './ThemeProvider.scss';
+
 interface ThemeProviderProps {
   /** Custom logos and colors provided to select components */
   theme: ThemeConfig;
@@ -28,7 +30,11 @@ export function ThemeProvider({
 
   return (
     <ThemeContext.Provider value={theme}>
-      <div style={customProperties} {...themeProvider.props}>
+      <div
+        style={customProperties}
+        {...themeProvider.props}
+        className={styles.ThemeProvider}
+      >
         {children}
       </div>
     </ThemeContext.Provider>
--- a/src/components/ThemeProvider/tests/ThemeProvider.test.tsx
+++ b/src/components/ThemeProvider/tests/ThemeProvider.test.tsx
@@ -120,4 +120,19 @@ describe('<ThemeProvider />', () => {
       }),
     );
   });
+
+  it('sets color system properties when global theming is enabled', () => {
+    const themeProvider = mountWithAppProvider(
+      <ThemeProvider theme={{}}>
+        <p>Hello</p>
+      </ThemeProvider>,
+      {features: {unstableGlobalTheming: true}},
+    );
+
+    expect(themeProvider.find('div').props().style).toStrictEqual(
+      expect.objectContaining({
+        '--p-surface-background': 'hsl(0, 0%, 98%, 1)',
+      }),
+    );
+  });
 });
--- a/src/components/ThemeProvider/ThemeProvider.scss
+++ b/src/components/ThemeProvider/ThemeProvider.scss
@@ -1,0 +1,6 @@
+@import '../../styles/common';
+
+.ThemeProvider {
+  background-color: var(--p-surface-background, color('sky', 'light'));
+  @include text-emphasis-normal;
+}
--- a/src/components/ThemeProvider/ThemeProvider.scss
+++ b/src/components/ThemeProvider/ThemeProvider.scss
@@ -1,6 +1,0 @@
-@import '../../styles/common';
-
-.ThemeProvider {
-  background-color: var(--p-surface-background, color('sky', 'light'));
-  @include text-emphasis-normal;
-}
--- a/src/styles/global.scss
+++ b/src/styles/global.scss
@@ -10,7 +10,6 @@
 html,
 body {
   @include text-style-body;
-  @include text-emphasis-normal;
 }
 
 html,
@@ -35,7 +34,6 @@ body {
   min-height: 100%;
   margin: 0;
   padding: 0;
-  background-color: color('sky', 'light');
 }
 
 *,
--- a/src/styles/global.scss
+++ b/src/styles/global.scss
@@ -10,6 +10,7 @@
 html,
 body {
   @include text-style-body;
+  @include text-emphasis-normal;
 }
 
 html,
@@ -34,6 +35,7 @@ body {
   min-height: 100%;
   margin: 0;
   padding: 0;
+  background-color: color('sky', 'light');
 }
 
 *,
--- a/src/styles/shared/_typography.scss
+++ b/src/styles/shared/_typography.scss
@@ -154,7 +154,7 @@ $typography-condensed: em(640px);
 
 @mixin text-emphasis-normal($for-background: null) {
   font-weight: 400;
-  color: color(ink, $for-background: $for-background);
+  color: var(--p-text-on-surface, color(ink, $for-background: $for-background));
 }
 
 @mixin text-breakword {


dfmcphee:
Would we consider this a breaking change then? I imagine most embedded apps wouldn't be using the frame so would need to update.

I'm not a fan of making all the chroma stories worse too. What do you see as the downside to setting the colour on the body on mount?

danrosenthal:
Thanks @dfmcphee!

> Would we consider this a breaking change then? I imagine most embedded apps wouldn't be using the frame so would need to update.

We've never really discussed if design changes were necessarily breaking. It's a bit less cut and dry than API changes. This does feel like a case where it is potentially breaking, at least in the context of embedded apps. Technically they should be wrapped in a `ThemeProvider`, so the background color _should_ continue to fill the entire frame unless the app was particularly short. But I don't feel super confident on that.  🤷‍♂ 

> What do you see as the downside to setting the colour on the body on mount?

I'm not sure there is any downside, and am leaning towards doing it in this PR. Any problems you can think of?

tmlayton:
I pushed a couple commits to set the background and text with JS if the feature flag is enabled and removed the changelog entry as there really is no change now. We’ve been defaulting to no changelog entry as this should be invisible to consumers.

Although I could see an argument for adding entries like "Applied color system to body", "Applied color system to Button", etc to make it easier to sniff out any potential regressions.

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -25,6 +25,18 @@ export function ThemeProvider({
...
   );
+  const backgroundColor = customProperties['--p-surface-background'];
+  const color = customProperties['--p-text-on-surface'];
+
+  useEffect(() => {
+    if (unstableGlobalTheming) {
+      document.body.style.backgroundColor = backgroundColor;
+      document.body.style.color = color;
+    } else {
+      document.body.style.backgroundColor = null;
```
BPScott: If we're not using global theming then will `customProperties['--p-surface-background']` (and text-on-surface) be set? If not then do we this logic toggle at all?

When unstableGlobalTheming changes, then customProperties will change and the keys mentioned will either be a particular color (if unstableGlobalTheming is true)  or undefined (if unstableGlobalTheming is false)

```
useEffect(() => {
  document.body.style.backgroundColor = backgroundColor;
  document.body.style.color = color;
}, [backgroundColor, color])
```

tmlayton: `'--p-surface-background'` in this case is a key lookup that returns the actual value, so yep it works. It sets the style attribute inline on body.

That’s why we need the `unstableGlobalTheming` dependency, so if that value changes we need to set those inline styles to null (remove them) so they fallback to the body styles defined in global.scss.

tmlayton: I think the missing part here is that setting `document.body.style.backgroundColor = undefined` is a no-op and does not remove the inline value, whereas `document.body.style.backgroundColor = null` does remove the value.

I updated the code to be a bit simpler.

src/components/ThemeProvider/tests/ThemeProvider.test.tsx
```diff
@@ -120,4 +120,19 @@ describe('<ThemeProvider />', () => {
...
+
+  it('sets color system properties when global theming is enabled', () => {
+    const themeProvider = mountWithAppProvider(
+      <ThemeProvider theme={{}}>
+        <p>Hello</p>
+      </ThemeProvider>,
+      {features: {unstableGlobalTheming: true}},
+    );
+
+    expect(themeProvider.find('div').props().style).toStrictEqual(
```
tmlayton: I was trying to do an assertion on body style for background color and color here but for some reason it is returning an empty string. I tried forcing an update in case it was a render cycle issue, but no luck.

src/components/ThemeProvider/ThemeProvider.tsx
BPScott: Thats the gotcha I was missing! the `|| null` works a treat

BPScott:
Touching the body feels somewhat scary to me, it feels like we're polluting a scope that up till now we've mostly not wanted to touch.

Question for yous:

Do we consider the global scope - styling on elements outside of an AppProvider  such as html and body - fair game for Polaris to touch?

If not then we're good at encapsulation, and we can handle having multiple AppProvider roots on the whole page.
If we don't then we say we expect only one instance of Polaris on a page.


Having a ThemeProvider create custom properties scoped to a div on itself suggests we want encapsulation.
Overriding the body styles (both here in ThemeProvider, and in global.scss) suggests we are opinionated and don't consider this aspect of encapsulation important.

If we decide we don't care about encapsulating Polaris to entirely within the ThemeProvider and that changing the body background is acceptable then may I ask what is the value of sticking all the customProperties on a div in ThemeProvider - why not add those custom properties in the scope of the html element?

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -26,6 +26,16 @@ export function ThemeProvider({
     [unstableGlobalTheming, themeConfig],
   );
 
+  // We want these values to be `null` instead of `undefined` when not set.
+  // Otherwise, setting a style property to `undefined` does not remove it from the DOM.
+  const backgroundColor = customProperties['--p-surface-background'] || null;
```
tmlayton: This will be better to grab from context once we make the colors available. Something like `theme.surfaceBackground` so if the color role name changes the type check will fail. Otherwise, if we change the color role name now, nothing would fail.

tmlayton:
> Do we consider the global scope - styling on elements outside of an AppProvider such as html and body - fair game for Polaris to touch?

The idealist answer is no, but we’ve already set precedence by having body element and other global styles that fall outside of the app provider. We also practically have to deal with the global scope (JS, Sass, CSS variables).

> Having a ThemeProvider create custom properties scoped to a div on itself suggests we want encapsulation.

Yes, indeed.

> Overriding the body styles (both here in ThemeProvider, and in global.scss) suggests we are opinionated and don't consider this aspect of encapsulation important.

I would say that the theme provider overrides are for the sake of backwards compatibility and that the global.scss overrides suggest that we have been known to make bad decisions from time to time 😆  

> why not add those custom properties in the scope of the html element

I had this thought earlier as well, it would makes things easier. But it feels like an anti-pattern. I think we do care about encapsulation and want to stick with it, unless we need to be pragmatic to handle something like backwards capability.

danrosenthal:
Great feedback @bpscott, and great response @tmlayton. I'm in agreement down to the letter.

An important point I'd like to emphasize. We **do** care about encapsulation, and breaking the bounds of the `AppProvider` is something we want to avoid, ideally. There are some cases where it is hard to avoid, cases like scroll locking or, as Tim said, backwards compatibility when we did something wrong in the past. These cases should be few and far between, and we should probably document why we've made the decisions we've made.

As far as setting the background color of the body, I think we should consider this to be a temporary answer to undoing our past mistakes. In the spirit of backwards compatibility, we should do it **now**. But we should remove it the first chance we can (I think in the next major release), while offering a sensible alternative. That is either a `FullScreenContainer` component, or the existing `Frame`. Once something like that is able to be offered, we kill the body styling (in either v5 or v6). 

I propose calling this practice **incremental idealism**, and I think we're taking a similar tack with burning down the Sass API. We know we've made mistakes in the past, and we want to undo them in the least painful way possible, with a sane upgrade path, and no breaking changes in non-major releases. We incrementally move towards a more idealistic state, where Polaris is contained within the bounds of the AppProvider (and our only API is components).

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -26,6 +26,16 @@ export function ThemeProvider({
...
+  const color = customProperties['--p-text-on-surface'] || null;
+
+  useEffect(() => {
+    document.body.style.backgroundColor = backgroundColor;
+    document.body.style.color = color;
+  }, [backgroundColor, color]);
+
```
danrosenthal: I think we can do all this work inside the existing `useMemo`

src/components/ThemeProvider/tests/ThemeProvider.test.tsx
danrosenthal: This is a big part of the reason we avoided setting these on the body in the first place. Take a look at our `setRootProperty` function to see that it is totally untested. The Jest env seems not to give us the ability to realistically assert on the `body`.

tmlayton:
Yeah I’m not comfortable with this approach, at least not right now. Adding a new full size div to every use of Polaris exposes us to a lot more risk.

BPScott:
> An important point I'd like to emphasize. We do care about encapsulation, and breaking the bounds of the AppProvider is something we want to avoid, ideally. There are some cases where it is hard to avoid, cases like scroll locking or, as Tim said, backwards compatibility when we did something wrong in the past. These cases should be few and far between, and we should probably document why we've made the decisions we've made.
>
> As far as setting the background color of the body, I think we should consider this to be a temporary answer to undoing our past mistakes. In the spirit of backwards compatibility, we should do it now. But we should remove it the first chance we can (I think in the next major release), while offering a sensible alternative. That is either a FullScreenContainer component, or the existing Frame. Once something like that is able to be offered, we kill the body styling (in either v5 or v6).

Strong agree on all of this and that our ideology is: we want to encapsulate as much as possible. Occasionally there will be leakage either because it is unavoidable (e.g. ScrollLock) or because we did something that is hard to rectify without a breaking change. Unavoidable cases should be documented on why we considered leaking acceptable. Where we've evolved our thinking - like color on body here - we aim to remove it when we can make a breaking change (i.e. next major release), and leave mitigations in place till that point.

We're all on the same page, I was being deliberately vague to make sure me taking a stance didn't influence you two.

Total agreement on doing this for backwards compat to maintain existing behaviour for now and removing it in the next major version and telling people if they want a body background they'll need to set it themselves.

---

Agree with Tim that making ThemeProvider  feels more  risk than maintaining the current behaviour of setting a background on body

---


 Small issue with current code - what happens if we  load with globalFeatures off - that'll set the backgroundColor to null - will that override any styling in existing stylesheets? Also note that in storybook we remove the body background color, so that we can see the color underneath, that might affect your testing of this - https://github.com/Shopify/polaris-react/blob/master/.storybook/preview-head.html#L2-L9

danrosenthal:
> Small issue with current code - what happens if we load with globalFeatures off - that'll set the backgroundColor to null - will that override any styling in existing stylesheets?

@tmlayton do you have thoughts on this?

We could set the body background and color using tokens that represent the non-global-theming values, when that was the case (instead of null).

tmlayton:
Nope, no issue. This code sets inline styles and doesn’t change style sheet values. Null will just fallback to the global sass.

PR #2273, opened 2019-10-10
[Stack] Fix min-width issue on Stack.Item

dleroux:

In fixing this issue https://github.com/Shopify/polaris-react/issues/1707 we removed min-width thinking it wasn't needed. Turns out it is. 

fixes: https://github.com/Shopify/polaris-react/issues/2260

Simply bringing back the min-width so that the flex-items takes the space it requires.

We do still have an issue where if `textOverflow: 'ellipsis',` is used on the child of the flex item, overflow is a little off. In which case its probably best to write flexbox for this edge case.

This fixed the bigger issue of the column not appearing.

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {AppProvider, Page, Layout, Card, Stack} from '../src';

  return (
    <Page title="Playground" narrowWidth>
      <Layout sectioned>
        <Card title="Broken 1" sectioned>
          <Stack>
            <div
              style={{
                overflow: 'hidden',
                whiteSpace: 'nowrap',
                textOverflow: 'ellipsis',
              }}
            >
              This should cause hello to wrap and also make sure that my text
              does not overflow my container. The ellipsis is spilling into the
              margin of the card however.
            </div>
            <div>
              <p>This line should wrap</p>
            </div>
          </Stack>
        </Card>
        <Card title="Broken 2" sectioned>
          <Stack alignment="center" wrap={false}>
            <div
              style={{
                overflow: 'hidden',
                whiteSpace: 'nowrap',
                textOverflow: 'ellipsis',
              }}
            >
              This should cause hello to still be visible and my content to be
              overflow hidden when I reach my container width. However, hello is
              hidden entirely and the ellipsis is spilling into the card margin.
            </div>
            <div>
              <p>hello world</p>
            </div>
          </Stack>
        </Card>
      </Layout>
    </Page>
  );
}


```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Stack/Stack.scss
+++ b/src/components/Stack/Stack.scss
@@ -105,6 +105,7 @@
 
 .Item {
   flex: 0 1 auto;
+  min-width: 0;
 }
 
 .Item-fill {
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -14,6 +14,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- Fixed issue with `Stack` where a `Stack.Item` was not getting a minimum width ([2273](https://github.com/Shopify/polaris-react/pull/2273))
+
 ### Documentation
 
 - Added accessibility documentation for the drop zone component ([#2243](https://github.com/Shopify/polaris-react/pull/2243))
--- a/src/components/TextField/README.md
+++ b/src/components/TextField/README.md
@@ -680,32 +680,34 @@ function SeparateValidationErrorExample() {
     : '';
 
   const formGroupMarkup = (
-    <Stack wrap={false} alignment="leading" spacing="tight">
+    <Stack wrap={false} alignment="leading" spacing="loose">
       <Stack.Item fill>
-        <Stack distribution="fill" spacing="tight">
-          <Select
-            labelHidden
-            label="Collection rule type"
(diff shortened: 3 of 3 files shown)


dleroux:
Looks like in some case this is touching the edge. On Percy, but also on ios.

dfmcphee:
> Looks like in some case this is touching the edge. On Percy, but also on ios.

Yeah, that's interesting. It looks like that example is a little wonky though. It uses nested stacks to do something that should be handled by a `FormLayout` instead.

src/components/Stack/Stack.scss
```diff
@@ -105,6 +105,7 @@
 
 .Item {
   flex: 0 1 auto;
+  min-width: 0;
```
LauraAubin: When the stack width is smaller than the contents, the contents overflow each other. Maybe we could make them wrap to the next line? Not sure if this would introduce even more edge cases though

dleroux:
There's still some weirdness happening with this when white space nowrap is set on a children. Overall  I think this is better than we have today since the column is not showing up at all.  I'm thinking if you are happy with the top hat Dom, we ship this for now.

src/components/Stack/Stack.scss
dleroux: Because this is an abstraction of Flexbox we might just want to leave it as is. Wrapping is left up to the props which should generate the right flexbox rules.

PR #2285, opened 2019-10-14
[Color system] add critical action variants, simplify neutral actions, add surface actions

danrosenthal:


src/components/Badge/Badge.scss
```diff
@@ -18,7 +18,7 @@ $pip-spacing: ($height - $pip-size) / 2;
   display: inline-flex;
   align-items: center;
   padding: 0 $horizontal-padding;
-  background-color: var(--p-interactive-neutral-elevation-1, color('sky'));
+  background-color: var(--p-neutral-action-disabled, color('sky'));
```
tmlayton: feels like the wrong role to use here

src/components/Navigation/Navigation.scss
```diff
@@ -166,7 +166,7 @@ $disabled-fade: 0.6;
       --p-none,
       linear-gradient(color(state, hover), color(state, hover))
     );
-    background-color: var(--p-interactive-neutral-elevation-2);
+    background-color: var(--p-neutral-action);
```
tmlayton: I think these should be surface-based roles, more similar to action list.

src/components/Badge/Badge.scss
danrosenthal: I feel the same way, but I'm not sure what actually fits there.

src/components/Navigation/Navigation.scss
danrosenthal: Makes sense to me.

src/components/TopBar/components/SearchField/SearchField.scss
```diff
@@ -99,7 +99,7 @@ $stacking-order: (
 
 .Clear {
   @include recolor-icon(
-    var(--p-interactive-neutral-elevation-4, color('ink', 'lightest'))
+    var(--p-neutral-action-pressed, color('ink', 'lightest'))
```
tmlayton: Also seems like the wrong role, maybe a new role for `neutral-icon`, `neutral-icon-hovered`, `neutral-icon-pressed`?

src/components/Badge/Badge.scss
danrosenthal: Probably a good use-case for the base neutral

```suggestion
  background-color: var(--p-neutral, color('sky'));
```

src/components/TopBar/components/SearchField/SearchField.scss
danrosenthal: I don't think so, since this is in-effect an icon-only button

PR #2295, opened 2019-10-15
[WIP][Color system] apply color system to text field

danrosenthal:
**todo**
- [ ] focus ring


src/components/TextField/TextField.scss
```diff
@@ -60,31 +60,38 @@ $stacking-order: (
...
   }
 }
 
+.readOnly {
+  color: var(--p-text-disabled-on-surface);
+}
+
 .readOnly > .Backdrop {
-  background-color: color('sky', 'lighter');
+  background-color: var(--p-neutral-action-disabled, color('sky', 'lighter'));
```
tmlayton: Feels like the wrong role. `--p-surface-foreground-subdued`? Or maybe another/new role?

tmlayton: Or maybe not, I can see the logic here.

src/styles/foundation/_focus-ring.scss
```diff
@@ -23,6 +23,7 @@
     } @else {
       border-radius: calc(var(--p-border-radius-base) + #{$position});
     }
+    pointer-events: none;
```
tmlayton: Carrying this change over from the existing text field focus implementation

src/components/TextField/TextField.scss
```diff
@@ -106,6 +113,10 @@ $stacking-order: (
   font-weight: inherit;
   appearance: none;
 
+  .hasValue & {
+    color: var(--p-text-on-surface, color('black'));
```
tmlayton: There appears to be a bug in Safari when using `inherit` as a `var` fallback (`var(--p-text-on-surface, inherit)`) will inherit a light ink color instead of the user agent stylesheet which sets input text color to black.

danrosenthal: So strange. Good catch @tmlayton!

danrosenthal:
Force pushed to remove the `[WIP]` from my initial commit, so this would register as no longer being in progress.

PR #2298, opened 2019-10-16
🐛[Sass] Fix recolor-icon mixin

beefchimi:
This PR safely moves `color: $secondary-color` into the `svg` block alongside `fill: $primary-color`.

Previously, the mixin was expected to be used like so:

```scss
.Component {
  color: blue;
  // more styles for the component
  @include recolor-icon(red, green);
}
```

Within the `recolor-icon` mixin, `fill` was scoped to to the `svg` selector... but `color` was outside of it. This meant that in the above example, my `color: blue` will get overridden with `green`, when really my intention is to have `.Component` use `blue` text, while making the primary color of the icon `red` and the secondary color of the icon `green`.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -15,6 +15,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Bug fixes
 
 - Doesn't render `MenuActions` if no actions are passed to an `actionGroups` item inside `Page` ([2266](https://github.com/Shopify/polaris-react/pull/2266))### Documentation
+- Fixed `recolor-icon` Sass mixin to properly scope `$secondary-color` to the child `svg` ([#2298](https://github.com/Shopify/polaris-react/pull/2298))
 
 ### Development workflow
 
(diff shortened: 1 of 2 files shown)


UNRELEASED.md
```diff
@@ -19,6 +19,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Fixed issue with `Filters` applying inconsistent border styles to sibling filters and when
   there is only one filter in the filter list ([2284](https://github.com/Shopify/polaris-react/pull/2284))
 - Added `aria-disabled` to select component content when it is disabled ([#2281](https://github.com/Shopify/polaris-react/pull/2281))
+- Fixed `recolor-icon` SASS mixin to properly scope `$secondary-color` to the child `svg` ([#2298](https://github.com/Shopify/polaris-react/pull/2298))
```
```suggestion
- Fixed `recolor-icon` Sass mixin to properly scope `$secondary-color` to the child `svg` ([#2298](https://github.com/Shopify/polaris-react/pull/2298))
```

kaelig:
:shipit:

PR #2303, opened 2019-10-16
[Frame] hookify Loading

sijad:

related to #1995

change `Frame.Loading` to use react hooks

### 🎩 checklist

### DIFF
--- a/src/components/Frame/components/Loading/Loading.tsx
+++ b/src/components/Frame/components/Loading/Loading.tsx
@@ -1,78 +1,57 @@
-import React from 'react';
-import debounce from 'lodash/debounce';
+import React, {useEffect, useState} from 'react';
 import styles from './Loading.scss';
-
-export interface LoadingProps {}
-
-interface State {
-  progress: number;
-  step: number;
-  animation: number | null;
-}
 
 const INITIAL_STEP = 10;
 const STUCK_THRESHOLD = 99;
 
-export class Loading extends React.Component<LoadingProps, State> {
-  state: State = {
-    progress: 0,
-    step: INITIAL_STEP,
-    animation: null,
-  };
+export function Loading() {
+  const [progress, setProgress] = useState(0.1);
 
-  private ariaValuenow = debounce(() => {
-    const {progress} = this.state;
-    return Math.floor(progress / 10) * 10;
-  }, 15);
+  useEffect(() => {
+    let animation: number;
+    let step = INITIAL_STEP;
+    let currentProgress = 0;
+    let previousProgress = 0;
 
-  componentDidMount() {
-    this.increment();
-  }
+    const increment = () => {
+      if (currentProgress >= STUCK_THRESHOLD) {
+        return;
+      }
 
-  componentWillUnmount() {
-    const {animation} = this.state;
+      currentProgress = Math.min(currentProgress + step, 100);
+      const nextProgress = Math.floor(currentProgress) / 100;
 
-    if (animation != null) {
-      cancelAnimationFrame(animation);
-    }
-  }
+      if (nextProgress !== previousProgress) {
+        setProgress(nextProgress);
+        previousProgress = nextProgress;
+      }
 
-  render() {
-    const {progress} = this.state;
+      step = INITIAL_STEP ** -(currentProgress / 25);
 
-    const customStyles = {
-      transform: `scaleX(${Math.floor(progress) / 100})`,
+      animation = requestAnimationFrame(increment);
     };
 
-    const ariaValuenow = this.ariaValuenow();
+    increment();
 
-    return (
-      <div className={styles.Loading}>
-        <div
-          className={styles.Level}
-          style={customStyles}
-          aria-valuenow={ariaValuenow}
-          aria-valuemin={0}
-          aria-valuemax={100}
-          role="progressbar"
-        />
-      </div>
-    );
-  }
+    return () => {
+      cancelAnimationFrame(animation);
+    };
+  }, []);
 
-  private increment() {
-    const {progress, step} = this.state;
+  const customStyles = {
+    transform: `scaleX(${progress})`,
+  };
 
-    if (progress >= STUCK_THRESHOLD) {
-      return;
-    }
-
-    const animation = requestAnimationFrame(() => this.increment());
-
-    this.setState({
-      progress: Math.min(progress + step, 100),
-      step: INITIAL_STEP ** -(progress / 25),
-      animation,
-    });
-  }
+  return (
+    <div className={styles.Loading}>
+      <div
+        className={styles.Level}
+        style={customStyles}
+        aria-valuenow={progress * 100}
+        aria-valuemin={0}
+        aria-valuemax={100}
+        role="progressbar"
+      />
+    </div>
+  );
 }
--- a/src/components/Frame/components/Loading/tests/Loading.test.tsx
+++ b/src/components/Frame/components/Loading/tests/Loading.test.tsx
@@ -1,17 +1,38 @@
 import React from 'react';
 import {mountWithAppProvider} from 'test-utilities/legacy';
+import {animationFrame} from '@shopify/jest-dom-mocks';
+import {act} from 'react-dom/test-utils';
 import {Loading} from '../Loading';
 
 describe('<Loading />', () => {
-  const loading = mountWithAppProvider(<Loading />);
-
-  it('mounts', () => {
-    expect(loading.exists()).toBe(true);
+  beforeEach(() => {
+    animationFrame.mock();
   });
 
-  it('unmounts safely', () => {
+  afterEach(() => {
+    animationFrame.restore();
+  });
+
+  it('increases over time', () => {
+    const loading = mountWithAppProvider(<Loading />);
+
+    for (let i = 0; i <= 100; i++) {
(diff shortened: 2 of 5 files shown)


sijad:
I'm not sure how tests can improved?

src/components/Frame/components/Loading/Loading.tsx
```diff
@@ -1,78 +1,69 @@
...
+  };
 
-    this.setState({
-      progress: Math.min(progress + step, 100),
-      step: INITIAL_STEP ** -(progress / 25),
-      animation,
-    });
-  }
+  const ariaValuenow = useLazyRef(() =>
+    debounce((progress: number) => {
```
sijad: In this aproach component will only re-rendered when progress changes, unlike the old aproach which would re-render with each `requestAnimationFrame` call. I guess it'd be safe to remove debounce and calculate the valenow inline.

BPScott: This is a a pure function  that doesn't depend on anything in the outer closure, so there's no need to put it inside the component or do any of the lazyRef stuff, just define it as a regular function outside of the component function

---

That said, now that we're only rendering this once per progress change I'm good with just defining this value inline. If I've understood your other changes - where progress is now a number between 0 and 1, rounded to two decimal places this should do:

```
aria-valuenow={progress * 100}
```

src/components/Frame/components/Loading/Loading.tsx
```diff
@@ -1,78 +1,69 @@
...
+export function Loading(_props: LoadingProps) {
+  const [progress, setProgress] = useState(0);
 
-  componentWillUnmount() {
-    const {animation} = this.state;
+  useEffect(() => {
+    let animation: number | undefined;
+    let step = INITIAL_STEP;
+    let currentProgress = 0;
+    let previusProgress = 0;
```
BPScott: typo `previousProgress`

src/components/Frame/components/Loading/Loading.tsx
```diff
@@ -1,78 +1,69 @@
...
 
-    if (animation != null) {
-      cancelAnimationFrame(animation);
-    }
-  }
+    const increment = () => {
+      if (currentProgress >= STUCK_THRESHOLD) {
```
BPScott: You seem to have changed the progress from being an integer between 0 an 100, to a number rounded to 2 decimal places between 0 and 1. I'm ok with that, but make sure that you update the stuck threshold and anything else like how the next step is calculated to account for that.

src/components/Frame/components/Loading/Loading.tsx
```diff
@@ -1,78 +1,69 @@
...
-  componentDidMount() {
-    this.increment();
-  }
+export function Loading(_props: LoadingProps) {
+  const [progress, setProgress] = useState(0);
 
-  componentWillUnmount() {
-    const {animation} = this.state;
+  useEffect(() => {
+    let animation: number | undefined;
```
BPScott: Should animation, step currentProgress and previousProgress be refs? It seems to work ok in this closure but I'm not sure why.

sijad: as we don't need those outside of `useEffect` I moved them all there, a race condition might occurs?

src/components/Frame/components/Loading/Loading.tsx
```diff
@@ -1,78 +1,69 @@
...
-          style={customStyles}
-          aria-valuenow={ariaValuenow}
-          aria-valuemin={0}
-          aria-valuemax={100}
-          role="progressbar"
-        />
-      </div>
-    );
-  }
+      animation = requestAnimationFrame(increment);
```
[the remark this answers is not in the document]
sijad: we should track latest `animation` for calling `cancelAnimationFrame` in un-mounting, I'm not sure if `const animation = increment();` will works.
should we use ref?

src/components/Frame/components/Loading/Loading.tsx
sijad: I'll refactor it and remove `debounce`, just FTR if we move that function outer closure, multiple instance of this component will use same debounced ref and it might cause data leak between those instance (they mill use each other values).

src/components/Frame/components/Loading/Loading.tsx
sijad: `currentProgress` still go from 0 - 100 with not limited decimal point (e.g. 20, 21.332232332, 22.12321, 23.5565, etc)
`nextProgress` and `previousProgress` goes from 0 - 1 with only two decimal point (e.g. 0.12, 0.13, 0.14, 0.15)
with this approach set state called only when `nextProgress` and `previousProgress` are not equal which save redundant re-rendering. if we set `currentProgress` in state, it'd re-render more often (due its decimals changes with each `requestAnimationFrame` call)

AndrewMusgrave:
Sorry for the wait @sijad @BPScott is busy at the moment and won't be able to review this again. If it's ready for another review let me know and I can take look!

sijad:
no worries, take your time
there're few unresolved conversations but I'd be great if you review this again

src/components/Frame/components/Loading/Loading.tsx
AndrewMusgrave: > Should animation, step currentProgress and previousProgress be refs? It seems to work ok in this closure but I'm not sure why.

From the current implementation, it should work as-is from what I can tell. We're creating a closure, recursing and mutating the original values. `setProgress` will be consistently called with the correct values, and when the component is dismounted it'll have the current animation in scope to clear.

src/components/Frame/components/Loading/Loading.tsx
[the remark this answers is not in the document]
AndrewMusgrave: Just chiming in `const animation = increment();` won't keep track of the recursed animation as the assignment will

src/components/Frame/components/Loading/Loading.tsx
```diff
@@ -1,78 +1,61 @@
...
-  private ariaValuenow = debounce(() => {
-    const {progress} = this.state;
-    return Math.floor(progress / 10) * 10;
-  }, 15);
+export interface LoadingProps {}
 
-  componentDidMount() {
-    this.increment();
-  }
+export function Loading(_props: LoadingProps) {
```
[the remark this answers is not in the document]
BPScott: Andrew: I've got the horrible feeling that this exists because not having a `LoadingProps` type might do ugly things to the types finder in the styleguide. Could you give that a check?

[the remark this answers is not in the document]
AndrewMusgrave: Yikes, I thought we used regex to find `{FILE_NAME}Props`. Do we use the ts compiler? I'll take a look and see how it works out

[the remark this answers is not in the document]
AndrewMusgrave: Ahh this is `components/Frame/components/Loading` not `components/Loading`. This component doesn't show in the style guide so we can remove `export interface LoadingProps {}`

sijad:
let me know if there's anything else than can be improved.
I'd also work on tests but as this component uses `requestAnimationFrame` I'm couldn't figure it out how?

AndrewMusgrave:
Mocking can be done a bunch of ways in `jest`, however, for `RAF` you'll probably want to go with one of these two depending on what you're trying to do.

```tsx
describe('suite', () => {
  let requestAnimationFrameSpy: jest.SpyInstance;

  beforeEach(() => {
    requestAnimationFrameSpy = jest.spyOn(window, 'requestAnimationFrame');
    requestAnimationFrameSpy.mockImplementation((cb) => {
      cb();
      // unique id used to cancel rafs
      return 1;
    });
  });

  afterEach(() => {
    requestAnimationFrameSpy.mockRestore();
  });

  ...
  it statements
  ...
})
```

or

```tsx
const origRequestAnimationFrame = window.requestAnimationFrame;
describe('suite', () => {
  afterEach(() => {
    // This might require Object.defineProperty
    window.requestAnimationFrame = origRequestAnimationFrame;
  });

  ...
  it('statement', () => {
    // This might require Object.defineProperty
    window.requestAnimationFrame = someMockRaf;
  })
  ...
})

// Object.defineProperty usage
// Object.defineProperty(window, 'requestAnimationFrame', {
//   configurable: true,
//   writable: true,
//   value: someMockFn,
// })
```

If you have any specific questions I'm happy to help :)

sijad:
@AndrewMusgrave thanks for the tips, I get `Maximum call stack size exceeded` error with recursive `requestAnimationFrameSpy` method, I went and used `animationFrame`

src/components/Frame/components/Loading/tests/Loading.test.tsx
```diff
@@ -1,17 +1,38 @@
...
+  afterEach(() => {
+    animationFrame.restore();
+  });
+
+  it('increases over time', () => {
+    const loading = mountWithAppProvider(<Loading />);
 
-  it('mounts', () => {
-    expect(loading.exists()).toBe(true);
+    for (let i = 0; i <= 100; i++) {
```
sijad: is this a good approach?
we can also change `100` to `100,000` to have a 100% code coverage but it took ~7 seconds in my computer...

sijad:
this is very disappointing, at first I thought this is a community driven project, but now I realized I was wrong.

PR #2312, opened 2019-10-17
[Migration guide] Improve part about using translations in AppProvider

alynioke:
Current migration docs didn't provide clear explanation what kind of `translations` should be used while migrating `<AppProvider>`.  I saw migration examples in different projects - in total, around 4-5 options what developers passed into translations. It wasn't clear which option is correct one. I saw other people looking for examples with translations and being confused what approach to take and what each approach means.

I'm describing options what to put into `translations` and for what case each option is sutable.

### How to 🎩
Is it the right place to put it? Maybe this info should be rather in README of `AppProvider`, and in migration guide there will be enough to insert a link to the updated `AppProvider` docs?  In case if this info is okay/needed in your opinion in the first place.

### DIFF
--- a/documentation/guides/migrating-from-v3-to-v4.md
+++ b/documentation/guides/migrating-from-v3-to-v4.md
@@ -73,9 +73,10 @@ Polaris now supports multiple languages and ships with [many translations](https
 
 // new
 import translations from '@shopify/polaris/locales/en.json';
-
 <AppProvider i18n={translations}>
 ```
+
+If you use [`@shopify/react-i18n`](https://github.com/Shopify/quilt/tree/master/packages/react-i18n) and want to dynamically load translations based on a provided locale, see the [AppProvider README](https://github.com/Shopify/polaris-react/blob/master/src/components/AppProvider/README.md#using-translations) for more information.
 
 ### Autocomplete <a name="polaris-autocomplete"></a>
 
--- a/src/components/AppProvider/README.md
+++ b/src/components/AppProvider/README.md
@@ -430,13 +430,48 @@ function AppProviderWithAllThemeKeysExample() {
 
 Translations are provided in the locales folder. When using Polaris, you are able to import translations from all languages supported by the core Shopify product and consume them through the `i18n` prop.
 
+If a project has only one locale, then you can pass the JSON content from the locale file into `AppProvider`.
+
 ```jsx
+import AppProvider from '@shopify/polaris';
 // en.json is English. Replace with fr.json for French, etc
 import translations from '@shopify/polaris/locales/en.json';
 
-ReactDOM.render(
-  <AppProvider i18n={translations}>{/* App content */}</AppProvider>,
-);
+function App() {
+  return <AppProvider i18n={translations}>{/* App content */}</AppProvider>;
+}
+```
+
+If a project supports multiple locales, then load them dynamically using [`@shopify/react-i18n`](https://github.com/Shopify/quilt/tree/master/packages/react-i18n#translation). This ensures that you load only the translations you need.
+
+```jsx
+import AppProvider from '@shopify/polaris';
+// en.json is English. Replace with fr.json for French, etc
+import translations from '@shopify/polaris/locales/en.json';
+import {useI18n} from '@shopify/react-i18n';
+
+function App() {
+  const [i18n] = useI18n({
+    id: 'Polaris',
+    fallback: translations,
+    translations(locale) {
+      return import(
+        /* webpackChunkName: "Polaris-i18n-[request]", webpackMode: "lazy-once" */ `@shopify/polaris/locales/${locale}.json`
+      ).then((dictionary) => dictionary && dictionary.default);
+    },
+  });
+
+  // i18n.translations is an array of translation dictionaries, where the first
+  // dictionary is the desired language, and the second is the fallback.
+  // AppProvider however expects that the first dictionary is the fallback
+  // and the second is the desired language. Thus we need to reverse the array
+  // to ensure the dictionaries are in the order desired by AppProvider
+  return (
+    <AppProvider i18n={i18n.translations.reverse()}>
+      {/* App content */}
+    </AppProvider>
+  );
+}
 ```
 
 ---
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -21,6 +21,9 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Documentation
 
+- Updated the `AppProvider` section in the Polaris [v3 to v4 migration guide](https://github.com/Shopify/polaris-react/blob/master/documentation/guides/migrating-from-v3-to-v4.md) ([#2312](https://github.com/Shopify/polaris-react/pull/2312))
+- Updated the `Using translations` section in the [AppProvider README](https://github.com/Shopify/polaris-react/blob/master/src/components/AppProvider/README.md#using-translations) ([#2312](https://github.com/Shopify/polaris-react/pull/2312))
+
 ### Development workflow
 
 ### Dependency upgrades


documentation/guides/migrating-from-v3-to-v4.md
```diff
@@ -72,9 +72,64 @@ Polaris now supports multiple languages and ships with [many translations](https
...
+This approach is suitable:
+
+- when project have only one specific locale
+- inside tests when you don't care which translations to use and need to pass any
+
+#### Dynamically
+
+What if your project supports different locales? Don't import all translation files as in example above - it's not optimal, you will end up with all translations in your bundle. What you need is to dynamically choose translations. There are 2 options.
+
+First option: you can copy/paste [translation files](https://github.com/Shopify/polaris-react/tree/master/locales) from Polaris into your project , into `translations` folder in the same place where your `AppProvider` is used. Then internationalization hook will work as usual and automatically will pick them up. You have freedom of controlling all translations, however, manual copy/pasting and updating is required.
```
BPScott: I really dislike this first option - it means consumers have to constantly check for new translations  that are added to polaris, which feels like a complicated waste of time.

Can you remove this, and push people towards the second option?

documentation/guides/migrating-from-v3-to-v4.md
```diff
@@ -72,9 +72,64 @@ Polaris now supports multiple languages and ships with [many translations](https
...
+```jsx
+import en from '@shopify/polaris/locales/en.json';
+
+export function AppProviderWithI18n({children}: Props) {
+  const [i18n] = useI18n({
+    id: 'Polaris',
+    fallback: en,
+    translations: async (locale) => {
+      const translationsLocale = locale && locale !== '' ? locale : 'en';
+      const dictionary = await import(
```
BPScott: What happens if the requested locale is one not yet supported by Polaris? Ideally we want to support every language web supports, but we might miss when one is added.

And slight bikeshed: do we need the dictionaty intermediate variable? 

```ks
return (await import(
    /* webpackChunkName: "Polaris-i18n-[request]" */ `@shopify/polaris/locales/${translationsLocale}.json`
)).default;
```

documentation/guides/migrating-from-v3-to-v4.md
```diff
@@ -72,9 +72,64 @@ Polaris now supports multiple languages and ships with [many translations](https
...
+const [i18n] = useI18n();
+
+return (
+  <AppProvider linkComponent={ErrorLink} i18n={i18n.translations[0]}>
+    {children}
+  </AppProvider>
+);
+```
+
+Second option: translations are loaded from Polaris package dynamically based on locale. In this case you don't need to copy/paste anything, your code will use latest translations.
```
BPScott: Can we add a note in here that points out that this uses `@shopify/react-i18n`

BPScott:
Thinking about this more, I think you're right, there's not much point duplicating this in both the AppProvider readme and in the migration guide - let's put the docs in the AppProvider readme and have the migration guide link to the relevant section

documentation/guides/migrating-from-v3-to-v4.md
alynioke: I just checked it with non-existing locale and I see that fallback translations are used in this case. Also, no worries about errors in this case - in [docs](https://github.com/Shopify/quilt/tree/master/packages/react-i18n#translations) it's written, "The translations function can also throw and react-i18n will handle the situation gracefully"

alynioke:
Hey @BPScott, I applied changes you recommended. 

I also did a change in my code - added `.reverse() `[here](https://github.com/Shopify/polaris-react/pull/2312/files#diff-7a936f09d89747421f1ed11ca7e51421R478), as i understood, without it, fallback didn't work. I saw examples in other projects which take `[0]`, `[1]` or use `reverse` on `i18n.translations` when passing to AppProvider. It didn't seem public knowledge why use one approach or another. So below the code I wrote explanation about why reverse is best suitable way. Could you please check if I have correct understanding, or if I don't introduce duplication? Maybe explanation how `useI18n` hook and AppProvider `i18n` prop work together is somewhere already.

And let me know if I should adjust anything else. Thanks! :)

documentation/guides/migrating-from-v3-to-v4.md
BPScott: Fantastic, thanks for that context - I'm not familiar with how react-i18n works internally.

src/components/AppProvider/README.md
```diff
@@ -431,14 +431,59 @@ function AppProviderWithAllThemeKeysExample() {
 Translations are provided in the locales folder. When using Polaris, you are able to import translations from all languages supported by the core Shopify product and consume them through the `i18n` prop.
 
 ```jsx
-// en.json is English. Replace with fr.json for French, etc
```
BPScott: I think referencing a `translations` variable without showing where it comes from is confusing. Let's keep this as an example of the static translations, then segue into dynamic translations - that way we might not need the explicit headings.

How does something like this sound? (Sorry for only coming up with this after you wrote the first draft):


> Translations are provided in the locales folder. When using Polaris, you are able to import translations from all languages supported by the core Shopify product and consume them through the `i18n` prop.
> 
> At its simplest you can pass the json content from a locale file into `AppProvider`. This approach is suitable when a project has only one specific locale.
> 
> ```jsx
> import AppProvider from "@shopify/polaris";
> // en.json is English. Replace with fr.json for French, etc
> import translations from "@shopify/polaris/locales/en.json";
> 
> function App() {
>   return <AppProvider i18n={translations}>{/* App content */}</AppProvider>;
> }
> ```
> 
> Alternatively if your project supports multiple locales you can load them dynamically using [`@shopify/react-i18n`](https://github.com/Shopify/quilt/tree/master/packages/react-i18n#translation). This will ensure you only load the translations you need instead of potentially loading every set of translations even if you are not using them.
> 
> ```jsx
> import AppProvider from "@shopify/polaris";
> // en.json is English. Replace with fr.json for French, etc
> import translations from "@shopify/polaris/locales/en.json";
> import { useI18n } from "@shopify/react-i18n";
> 
> function App() {
>   const [i18n] = useI18n({
>     id: "Polaris",
>     fallback: translations,
>     translations(locale) {
>       return import(
>         /* webpackChunkName: "Polaris-i18n-[request]", webpackMode: "lazy-once" */ `@shopify/polaris/locales/${locale}.json`
>       ).then(dictionary => dictionary && dictionary.default);
>     }
>   });
> 
>   // i18n.translations is an array of translation dictionaries, where the first
>   // dictionary is the desired language, and the second is the fallback.
>   // AppProvider however expects that the first dictionary is the fallback
>   // and the second is the desired language. Thus we need to reverse the array
>   // to ensure the dictionaries are in the order desired by AppProvider
>   return (
>     <AppProvider i18n={i18n.translations.reverse()}>{/* App content */}</AppProvider>
>   );
> }
> ```

---

I've modified the translations function there, but I've not ran it to test as I've not got an environment setup for that. Would you be able to check it for me - it is based up on a simplified version of the code the [react-i18n babel plugin](https://github.com/Shopify/quilt/blob/3f8f237de601f93501f6a2ad0d63bdbe845ea43c/packages/react-i18n/src/babel-plugin.ts#L30-L48) generates - that's where the "webpacMode" comes from. 

I've removed any typescript types as we want our examples to be plain JS. I've also removed the `const translationsLocale = locale && locale !== '' ? locale : 'en';` check as [react-i18n's types](https://github.com/Shopify/quilt/blob/3f8f237de601f93501f6a2ad0d63bdbe845ea43c/packages/react-i18n/src/manager.ts#L9-L11) suggests that a string will always be passed into the translate function. I've also added a comment explaining why the reverse is needed inline, rather than explaining it in the docs (related - we should fix the ordering to match usei18n in the next major version to avoid this fiddling)

alynioke: agree that fixing the order will help - then there will be no need to understand this reverse "gotcha"

alynioke:
@BPScott applied suggestions. I applied code you wrote in docs to project - it worked! 

Thank you for the thorough review :) Noting for the future, how to format docs for it to be more generic and universal (removing parts which are project-related; adding parts to make examples valid.

let me know if there's anything else.

documentation/guides/migrating-from-v3-to-v4.md
```diff
@@ -73,10 +73,11 @@ Polaris now supports multiple languages and ships with [many translations](https
 
 // new
 import translations from '@shopify/polaris/locales/en.json';
-
 <AppProvider i18n={translations}>
 ```
 
+If you wish to dynamically load translations based on a provided locale see the [AppProvider README](https://github.com/Shopify/polaris-react/blob/master/src/components/AppProvider/README.md#using-translations) for more information.
```
AndrewMusgrave: Should we mention it's only relevant if you use `@shopify/react-i18n`?

src/components/AppProvider/README.md
```diff
@@ -430,13 +430,48 @@ function AppProviderWithAllThemeKeysExample() {
...
+import translations from '@shopify/polaris/locales/en.json';
+import {useI18n} from '@shopify/react-i18n';
+
+function App() {
+  const [i18n] = useI18n({
+    id: 'Polaris',
+    fallback: translations,
+    translations(locale) {
+      return import(
+        /* webpackChunkName: "Polaris-i18n-[request]", webpackMode: "lazy-once" */ `@shopify/polaris/locales/${locale}.json`
```
AndrewMusgrave: I'm not too familiar with `react-i18n`, does this require specific webpack configuration?

src/components/AppProvider/README.md
```diff
@@ -430,13 +430,48 @@ function AppProviderWithAllThemeKeysExample() {
...
+      ).then((dictionary) => dictionary && dictionary.default);
+    },
+  });
+
+  // i18n.translations is an array of translation dictionaries, where the first
```
AndrewMusgrave: `translation dictionaries` is a named for a type we've used but consumers may not know what we're referring too. What do you think about describing it the type?

AndrewMusgrave:
Awesome work, thanks for adding this  @alynioke 🎉  . I've added a content strategist as a reviewer 😄

documentation/guides/migrating-from-v3-to-v4.md
BPScott: We mention that in the README, people will find out soon enough.

src/components/AppProvider/README.md
BPScott: Nothing beyond what sewing-kit already provides

src/components/AppProvider/README.md
BPScott: That feels like an implementation detail of react-i18n, I don't think we need to go into much detail other than "here's what useI18n gives us, here's how we need to tweak it so it works with our internationalisation".

documentation/guides/migrating-from-v3-to-v4.md
AndrewMusgrave: It seems odd to me we're directing people to another page, then letting them know we actually meant we can show you how to use dynamically load translations if you use our internationalization package and our non-opened sourced tool without public docs or webpack (with maybe custom config required). Keeping in mind the majority of projects consuming polaris are not internal.

src/components/AppProvider/README.md
AndrewMusgrave: We may want to add a section here about webpack being required

src/components/AppProvider/README.md
AndrewMusgrave: iirc i18n.translations can include various translations. If we're wanting people to use `react-i18n`, they'll want to use it for their custom translations as well rather than have two i18n dependencies. i18n.translations could be [polaris_fallback, polaris_jp, custom translations]. I'm not sure how much detail we'll want to include, but it might help the consumer become more aware of what's actually happening.

BPScott: I don't think it's our responsibility to describe how react-i18n works here - that's what their docs are for. This is true for the common case that we recommend and if people want to use a more advanced setup then it's up to them to work out how they want to diverge from our recommendations.

src/components/AppProvider/README.md
BPScott: That seems pretty strongly implied. This section is targeted at shopify developers, it's almost guaranteed that they'll be working on a project that uses sewing-kit, and thus webpack and react-i18n will be already preconfigured anyway.

documentation/guides/migrating-from-v3-to-v4.md
sadiesaurus: I think it'd be good to mention `@shopify/react-i18n` up front in this case. Here's an option (the link needs to be added, sorry I couldn't find it): 

```suggestion
If you use [`@shopify/react-i18n`](link) and want to dynamically load translations based on a provided locale,  see the [AppProvider README](https://github.com/Shopify/polaris-react/blob/master/src/components/AppProvider/README.md#using-translations) for more information.
```

src/components/AppProvider/README.md
```diff
@@ -430,13 +430,48 @@ function AppProviderWithAllThemeKeysExample() {
 
 Translations are provided in the locales folder. When using Polaris, you are able to import translations from all languages supported by the core Shopify product and consume them through the `i18n` prop.
 
+At its simplest you can pass the json content from a locale file into `AppProvider`. This approach is suitable when a project has only one specific locale.
```
sadiesaurus: Can we simplify to something like this?

```suggestion
If a project has only one locale, then you can pass the JSON content from the locale file into `AppProvider`. 
```

src/components/AppProvider/README.md
```diff
@@ -430,13 +430,48 @@ function AppProviderWithAllThemeKeysExample() {
...
 
-ReactDOM.render(
-  <AppProvider i18n={translations}>{/* App content */}</AppProvider>,
-);
+function App() {
+  return <AppProvider i18n={translations}>{/* App content */}</AppProvider>;
+}
+```
+
+Alternatively if your project supports multiple locales you can load them dynamically using [@shopify/react-i18n](https://github.com/Shopify/quilt/tree/master/packages/react-i18n#translation). This will ensure you only load the translations you need instead of potentially loading every set of translations even if you are not using them.
```
sadiesaurus:
```suggestion
If a project supports multiple locales, then load them dynamically using [@shopify/react-i18n](https://github.com/Shopify/quilt/tree/master/packages/react-i18n#translation). This ensures that you load only the translations you need.
```

documentation/guides/migrating-from-v3-to-v4.md
BPScott: Sadie brought this up as well. let's add a note in :)

BPScott: Andrew mentioned this too - I'm 👍 on this change

src/components/AppProvider/README.md
BPScott: Much nicer! And thank you for getting rid of my use of "simplest". Describing something as simple is a bold assumption and I should know better :D

BPScott:
@alynioke: Sadie's changes all sound good to me, can you add them, then I'll have a quick chat on Friday with Andrew to see if he'd like to see anything else added. Once I confirm he's happy I'll make any changes and merge this in (as you'll have finished for the weekend by that point) - does that sound OK?

alynioke:
@sadiesaurus thank you for your suggestions! I agree on all and applied them. 
@BPScott - that sounds great! Updated PR. I go to vacations next week, so you merging would be helpful to speed this up.

PR #2315, opened 2019-10-17
Autocomplete v2 test tweaks

dpersing:
- Added a check for having multiple `aria-selected="true"` only if `aria-multiselectable` is on the listbox and set to `true`
- Tweaked some ARIA attribute names in comments

### DIFF
--- a/src/components/AutocompleteV2/tests/Autcomplete.A11Y.tests.tsx
+++ b/src/components/AutocompleteV2/tests/Autcomplete.A11Y.tests.tsx
@@ -2,21 +2,21 @@
 describe('A11Y', () => {
   describe('combobox', () => {
     it.todo('renders the activator wrapped in a div with a role of combobox');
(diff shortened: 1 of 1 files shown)


src/components/AutocompleteV2/tests/Autcomplete.A11Y.tests.tsx
```diff
@@ -2,21 +2,21 @@
 describe('A11Y', () => {
   describe('combobox', () => {
     it.todo('renders the activator wrapped in a div with a role of combobox');
-    it.todo('has an `area-expanded` attribute of false by default');
+    it.todo('has an `aria-expanded` attribute of false by default');
     it.todo(
-      'has an `area-expanded` attribute of true when a ListBox is visible',
```
dleroux: Noticed that this morning 🙈

PR #2318, opened 2019-10-17
[Color system] apply surface and divider colors to modal

tmlayton:
Applying surface and divider colors to modal


danrosenthal:
✅

PR #2321, opened 2019-10-17
Textfield placeholder clarification

dpersing:

Clarifies usage for the `helpText`, `label`, and `placeholder` props, to appear in `polaris-react` docs and in the style guide for the `TextField` component.

* [X] Updates accessibility documentation and and props descriptions for the text field `README.md`
* [x] Adds an entry to `UNRELEASED.md`

## How to 🎩

1. Check out `master` from `polaris-styleguide` to get the changes that support the accessibility section.
1. In another Terminal tab or window, check out this branch from `polaris-react` and [run the instructions for testing in a consuming project](https://github.com/Shopify/polaris-react/#testing-in-a-consuming-project).
1. In the `polaris-styleguide` tab, run `dev up && dev start`.
1. View changes after examples: https://polaris.myshopify.io/components/forms/text-field

### DIFF
--- a/src/components/TextField/README.md
+++ b/src/components/TextField/README.md
@@ -830,32 +830,30 @@ See Apple’s Human Interface Guidelines and API documentation about accessibili
 
 <!-- content-for: web -->
 
-### Structure
-
 Screen readers convey information about text fields automatically through native HTML.
 
 - Use the `disabled` prop to add the HTML `disabled` attribute to the text field.
 - Use the `readOnly` prop to add the HTML `readonly` attribute to the text field.
-- If you use the `type` prop, then some assistive technologies adapt the software keyboard to the current task. This helps merchants with mobility, vision, and cognitive issues to enter information more easily.
+- If you use the `type` prop, then some browsers and assistive technologies adapt the software keyboard to the current task. This helps users with mobility, vision, and cognitive issues to enter information more easily.
 
 Use the `id` prop to provide a unique `id` attribute value for the text field. If you don't provide an `id`, then the component generates one automatically. All text fields need to have unique `id` values.
 
 ### Labeling
 
-The `label` prop is required to convey the purpose of the checkbox to all merchants.
+The `label` prop is required to convey the purpose of the checkbox to all users.
 
-If there are separate visual cues that convey the purpose of the text field to sighted merchants, then the label can be visually hidden with the `labelHidden` prop.
+If there are separate visual cues that convey the purpose of the text field to sighted users, then the label can be visually hidden with the `labelHidden` prop.
 
 When you provide help text via the `helpText` prop or an inline error message via the `error` prop, the help or error content is conveyed to screen reader users with the `aria-describedby` attribute. This attribute causes the content to be read along with the label, either immediately or after a short delay.
 
-Use the `placeholder` prop to provide additional instructions. However, don’t rely on placeholders alone since the content isn’t always conveyed to all merchants.
+Use the `placeholder` prop only to provide formatting instructions that aren't essential. [HTML placeholders can be difficult for users to access.](https://www.nngroup.com/articles/form-design-placeholders/) Don’t rely on placeholders to convey important information since the content isn’t always conveyed to all users.
 
 <!-- usageblock -->
 
 #### Do
 
-- Use the label to provide instructions critical to using the text field
-- Use help text and placeholder text to provide additional, non-critical instructions
+- Use the label as the primary way to provide instructions critical to using the text field
+- Use help text as a secondary way to provide additional, non-critical instructions
 
 #### Don’t
 
@@ -867,8 +865,8 @@ Use the placeholder to provide information that’s required to use the text fie
 
 Text fields have standard keyboard support.
 
-- Merchants who rely on the keyboard expect to move focus to each text field using the <kbd>tab</kbd> key (or <kbd>shift</kbd> + <kbd>tab</kbd> when tabbing backwards)
-- If the `type` is set to `number`, then merchants can use the up and down arrow keys to adjust the value typed into the field
+- Users who rely on the keyboard expect to move focus to each text field using the <kbd>tab</kbd> key (or <kbd>shift</kbd> + <kbd>tab</kbd> when tabbing backwards)
+- If the `type` is set to `number`, then users can use the up and down arrow keys to adjust the value typed into the field
 - Using the `disabled` prop will prevent the text field from receive keyboard focus or inputs
 - The `readOnly` prop allows focus on the text field but prevents input or editing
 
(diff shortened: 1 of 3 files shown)


dpersing:
@kaelig and @selenehinkley Here are proposed changes to clarify the use of labels, help text, and placeholder text for all applications. let me know if you've got questions!

src/components/TextField/README.md
```diff
@@ -830,32 +830,30 @@ See Apple’s Human Interface Guidelines and API documentation about accessibili
...
 
-If there are separate visual cues that convey the purpose of the text field to sighted merchants, then the label can be visually hidden with the `labelHidden` prop.
+If there are separate visual cues that convey the purpose of the text field to sighted users, then the label can be visually hidden with the `labelHidden` prop.
 
 When you provide help text via the `helpText` prop or an inline error message via the `error` prop, the help or error content is conveyed to screen reader users with the `aria-describedby` attribute. This attribute causes the content to be read along with the label, either immediately or after a short delay.
 
-Use the `placeholder` prop to provide additional instructions. However, don’t rely on placeholders alone since the content isn’t always conveyed to all merchants.
+Use the `placeholder` prop only to provide formatting instructions that aren't essential. [HTML placeholders can be difficult for users to access.](https://www.nngroup.com/articles/form-design-placeholders/) Don’t rely on placeholders to convey important information since the content isn’t always conveyed to all users.
```
kaelig: We can probably get away with less repetition?

```suggestion
Use the `placeholder` prop only to provide formatting instructions that aren’t essential, as [users may find difficult or impossible to access placeholders](https://www.nngroup.com/articles/form-design-placeholders/).
```

src/components/TextField/README.md
```diff
@@ -830,32 +830,30 @@ See Apple’s Human Interface Guidelines and API documentation about accessibili
...
 
 <!-- usageblock -->
 
 #### Do
 
-- Use the label to provide instructions critical to using the text field
-- Use help text and placeholder text to provide additional, non-critical instructions
+- Use the label as the primary way to provide instructions critical to using the text field
+- Use help text as a secondary way to provide additional, non-critical instructions
```
kaelig: - Should we provide examples?
- Since _label_ and _help text_ both have a "Do", would it be useful to also provide a "Do" for _placeholder_?

src/components/TextField/README.md
```diff
@@ -830,32 +830,30 @@ See Apple’s Human Interface Guidelines and API documentation about accessibili
...
 <!-- content-for: web -->
 
-### Structure
-
 Screen readers convey information about text fields automatically through native HTML.
 
 - Use the `disabled` prop to add the HTML `disabled` attribute to the text field.
 - Use the `readOnly` prop to add the HTML `readonly` attribute to the text field.
-- If you use the `type` prop, then some assistive technologies adapt the software keyboard to the current task. This helps merchants with mobility, vision, and cognitive issues to enter information more easily.
+- If you use the `type` prop, then some browsers and assistive technologies adapt the software keyboard to the current task. This helps users with mobility, vision, and cognitive issues to enter information more easily.
```
kaelig: With the introduction of "browsers", does "software keyboard" still cover all use cases?

src/components/TextField/TextField.tsx
```diff
@@ -35,15 +35,15 @@ export interface BaseProps {
   prefix?: React.ReactNode;
   /** Text to display after value */
   suffix?: React.ReactNode;
-  /** Hint text to display */
+  /** Non-essential hint text to display in the field */
```
kaelig: 👍 ❤️

src/components/TextField/README.md
dpersing: Works for me!

src/components/TextField/README.md
dpersing: My personal opinion is that placeholders are...not good. I don't want to provide examples where they are deemed useful, because I think folks should use labels and help text instead. I am open to arguments on this topic though. 😂

src/components/TextField/README.md
dpersing: I think so. I was trying to get at software keyboards on mobile, as well as those introduces by things like switch devices. This can probably be clearer, though. 🤔

src/components/TextField/README.md
```diff
@@ -830,32 +830,30 @@ See Apple’s Human Interface Guidelines and API documentation about accessibili
...
 
 Use the `id` prop to provide a unique `id` attribute value for the text field. If you don't provide an `id`, then the component generates one automatically. All text fields need to have unique `id` values.
 
 ### Labeling
 
-The `label` prop is required to convey the purpose of the checkbox to all merchants.
+The `label` prop is required to convey the purpose of the checkbox to all users.
 
-If there are separate visual cues that convey the purpose of the text field to sighted merchants, then the label can be visually hidden with the `labelHidden` prop.
+If there are separate visual cues that convey the purpose of the text field to sighted users, then the label can be visually hidden with the `labelHidden` prop.
```
sadiesaurus: Loving the changes to `users` ❤️

src/components/TextField/README.md
```diff
@@ -830,32 +830,30 @@ See Apple’s Human Interface Guidelines and API documentation about accessibili
...
 
 When you provide help text via the `helpText` prop or an inline error message via the `error` prop, the help or error content is conveyed to screen reader users with the `aria-describedby` attribute. This attribute causes the content to be read along with the label, either immediately or after a short delay.
 
-Use the `placeholder` prop to provide additional instructions. However, don’t rely on placeholders alone since the content isn’t always conveyed to all merchants.
+Use the `placeholder` prop only to provide formatting instructions that aren’t essential, as [users may find difficult or impossible to access placeholders](https://www.nngroup.com/articles/form-design-placeholders/).
```
sadiesaurus: Related to this point, and the following DO examples - the content standards workgroup is still finalizing new content patterns for text fields. Currently the group is planning to recommend that no placeholder text is used unless the text field is related to searching or filtering (is this component ever used in that context?), and that formatting instructions are included as help text instead. Here are the WIP guidelines if you're curious https://docs.google.com/document/d/1eNHv13nrJ562TN-f2UpvZmhstFRh-x_6wzPbgEzQz3s/edit# 

Can we align with that approach here as well?
PS I like having the NN group link there for additional context.

src/components/TextField/README.md
sadiesaurus: I agree with that approach, Devon. And that aligns with the upcoming content pattern that I mentioned in my comment above about not using placeholder text unless it's for a search or filter bar ("Search {x}", or "Filter {x}") because it's such a common pattern. I'm assuming that the text field component isn't used in this way, so using that as a DO example here wouldn't work.

src/components/TextField/README.md
dpersing: @sadiesaurus  Text fields could totally be used for a search or filter interface. What type of text would the search or filter examples use, just "Search" or "Filter" as a visual prompt to duplicate a visually hidden label?

I also really like the phrase "Don’t use placeholder text for comments and notes; use help text instead" in the doc you shared.

sadiesaurus: Yes, the pattern that we're proposing is going to be "Search {x}" or "Filter {x}" (for example "Search themes" or "Filter orders"). I think the plan is to recommend that these placeholders are floating when you type, like here in the Help Center search https://help.shopify.com/

src/components/TextField/README.md
kaelig: Case where a placeholder is useful: disambiguation. Not crucial, but a nice little nudge to write the correct format in the field. For example: "https://" for a URL, or "@username" for a Twitter account. 
I don't have research data on hand for this, but I can say that it definitely helps me reduce the number of errors I do when filling a form.

PR #2322, opened 2019-10-17
4.6.1

tmlayton:
Merging in changes from the v4.6.1 patch release branch.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -11,4 +11,3 @@ node_modules
 /styles.scss
 /styles
 /src/styles/polaris-tokens
-tests/shims/mutation-observer.ts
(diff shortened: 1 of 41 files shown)


kaelig:
Looks like the tags have already been pushed and the package was published? (I thought we were supposed to tag master?)

tmlayton:
Yes, already tagged and released. This is just to get the changelog entry and readme updates in master. The v4.6.1 patch is based on v4.6.0 and I setup a separate deploy stack for it. Generally we tag master, but I’ve done this a few times in the past when we have a non-trivial amount of unreleased changes and only want the urgent fix(es).

kaelig:
Thanks, sounds good!

PR #2325, opened 2019-10-18
[ChoiceList] Allow JSX in a choice's label

jtrollia:

I have a couple of use cases, where I want to increase the font weight of a group of words in a choice's label, in a `<ChoiceList />`. Also, `<Checkbox />` and `<RadioButton />` already allow `ReactNode`s on the `label` prop, while `<ChoiceList />` doesn't. It is also possible in rails.

Screen Shot 2019-10-18 at 11 06 54 AM

Replacing the `string` type of the `label` sub prop by a more permissive `ReactNode` type. Adding tests accordingly.

## 

### How to 🎩

You can just play with the `Playground` snippet below ⬇️ 

<details>

```jsx
import React, {Fragment, useState} from 'react';
import {Page, ChoiceList} from '../src';

const jsxLabel = <b>Two</b>;
const ComponentLabel = () => (
  <Fragment>
    Label <u>one</u>
  </Fragment>
);

const choices = [
  {label: <ComponentLabel />, value: 'one'},
  {label: jsxLabel, value: 'two'},
  {label: 'Three', value: 'three', helpText: 'Some help text'},
];

  const [selected, setSelected] = useState('two');
  const changeMePlz = (value: string[]) => setSelected(value[0]);

  return (
    <Page title="Playground">
      <ChoiceList
        title="Choose a number"
        selected={[selected]}
        choices={choices}
        onChange={changeMePlz}
      />
      ,
    </Page>
  );
}
```

</details>

### 🎩 checklist

* [ ] ~Tested on [mobile](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting.md#cross-browser-testing)~
* [ ] ~Tested on [multiple browsers](https://help.shopify.com/en/manual/intro-to-shopify/shopify-admin/supported-browsers)~
* [ ] ~Tested for [accessibility](https://github.com/Shopify/polaris-react/blob/master/documentation/Accessibility%20testing.md)~
* [ ] ~Updated the component's `README.md` with documentation changes~
* [ ] ~[Tophatted documentation](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting%20documentation.md) changes in the style guide~
* [ ] ~For visual design changes, pinged one of @ HYPD, @ mirualves, @ sarahill, or @ ry5n to update the [Polaris UI kit](https://polaris.shopify.com/resources/polaris-ui-kit)~

### DIFF
--- a/src/components/ChoiceList/tests/ChoiceList.test.tsx
+++ b/src/components/ChoiceList/tests/ChoiceList.test.tsx
@@ -50,6 +50,31 @@ describe('<ChoiceList />', () => {
       });
     });
 
+    it('renders choices with labels containing JSX', () => {
+      const jsxLabel = <b>Two</b>;
+      const ComponentLabel = () => (
+        <React.Fragment>
+          Label <i>one</i>
+        </React.Fragment>
+      );
+
+      choices = [
+        {label: <ComponentLabel />, value: 'one'},
+        {label: jsxLabel, value: 'two'},
+        {...choices[2], helpText: 'Some help text'},
+      ];
+
+      const choiceElements = mountWithAppProvider(
+        <ChoiceList title="Choose a number" selected={[]} choices={choices} />,
+      ).find(RadioButton);
+
+      choiceElements.forEach((choiceElement, index) => {
+        expect(choiceElement.prop('label')).toBe(choices[index].label);
+        expect(choiceElement.prop('value')).toBe(choices[index].value);
+        expect(choiceElement.prop('helpText')).toBe(choices[index].helpText);
+      });
+    });
+
     describe('with valid children property returning node', () => {
       const children = <span>Child</span>;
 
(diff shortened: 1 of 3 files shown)


src/components/ChoiceList/tests/ChoiceList.test.tsx
```diff
@@ -50,6 +50,31 @@ describe('<ChoiceList />', () => {
       });
     });
 
+    it('renders choices with labels containing JSX', () => {
+      const jsxLabel = <b>Two</b>;
+      const ComponentLabel = () => (
+        <Fragment>
```
daggeto: Why do you use `Fragment` here instead of `<>`?  Never used `Fragment` in tests, so curious if it possible at all.

BPScott: `<> ` is syntactic sugar and gets compiled to `<React.Fragment>`. Because of *historical reasons* in Polaris's build we have to use the long-form version and we have an eslint rule to enforce that.

jtrollia: I went fancy with that `<Fragment />` but I could simply use a good old `<div />` as well if it better matches your best practices 👍

PR #2326, opened 2019-10-18
Dependency updates

BPScott:

Keeping up with latest, and unblocking storybook updates, as our old version of sewing-kit was preventing storybook updates

Updates sewing-kit, updates storybook.
Code updates to pass linting as this included a new version of eslint-plugin-shopify

Linting fixes are mostly:

- Removing `eslint-disable-next-line shopify/jest/no-if` lines as that no longer errors if your if statement is inside a function in a test.
- Tweaking ordering of imports
- Fixing violations of https://eslint.org/docs/rules/no-prototype-builtins

### How to 🎩

Ensure lint, tests and  type check all pass


sewing-kit.config.ts
```diff
@@ -19,9 +19,6 @@ export default function sewingKitConfig(
       plugins.jest((config: InitialOptions) => {
         config.roots = [join(__dirname, 'src'), join(__dirname, 'tests')];
 
-        config.setupFiles.push(join(tests, 'setup.ts'));
```
BPScott: Sewing-kit now automatically adds these files to the setupFiles and setupFilesAfterEnv blocks if they exist, so we don't need to be explict here.

scripts/utilities/retry.js
```diff
@@ -3,6 +3,8 @@ function sleep(ms) {
 }
 
 function retry(functionToTry, maxAttempts = 3, delay = 1000) {
+  // We should fix this but I've not got the time at the moment
```
kaelig: Can we make this comment more useful and actionable?

src/components/UnstyledLink/UnstyledLink.tsx
```diff
@@ -7,9 +7,10 @@ import {useLink, LinkLikeComponentProps} from '../../utilities/link';
 // but the props explorer isn't smart enough to work that out
 export interface UnstyledLinkProps extends LinkLikeComponentProps {}
 
+// This does have a display name, but the linting has a bug in it
+// https://github.com/yannickcr/eslint-plugin-react/issues/2324
+// eslint-disable-next-line react/display-name
```
kaelig: Is this needed anymore?

BPScott: Yep, still needed. We're using eslint-plugin-eslint-comments which contains rules than trigger a warning when we've got a disable line that doesn't do anything, so if this was unneeded it would be flagged.

This issue has been [resolved](https://github.com/yannickcr/eslint-plugin-react/pull/2399), and is in v7.15.0 but eslint-plugin-shopify depends up on v7.14.3 exactly.

On eslint-plugin-shopify's master branch that dependency has been bumped, so the next time theres a release of eslint-plugin-shopify this can be fixed up.

scripts/utilities/retry.js
BPScott: Got a suggestion in mind? If I work out how to make it actionable then I might as well just make the change and I don't really want to do that at the moment

kaelig: How about:

FIXME: reasons why this isn't good

src/components/UnstyledLink/UnstyledLink.tsx
kaelig: Ok! Found it a bit strange that the eslint-disable-next-line was moved.

BPScott: There was a version bumb to eslint-plugin-react but not to the version we need - Now it raises a warning on both the React.memo and React.forwardRef - both of which are spurious.

Tbh there's no value in wrapping this in a memo anyway so I'm planning to remove that in a later PR.

scripts/utilities/retry.js
BPScott: Have added a link to the lint rule info page which explains all that.

PR #2327, opened 2019-10-18
[Color system][Portal] Remove style attribute from portal node when custom properties are undefined

danrosenthal:


tmlayton:
I realized when 🎩ing this would prevent us from removing styles when disabling global theming, so updated to instead remove the style attribute.

PR #2329, opened 2019-10-18
[Button] add support for split disclosure button

chloerice:

There is currently no way to expose merchants to related primary actions in the `ContextualSaveBar` or `Page` `Header` because they only support a single primary action. For example, if the primary action is to save a resource like an order, but you can also save the order as a draft. 

This PR adds support for a split button variant. This is the same UX as the split button you are probably familiar with in Gmail. It renders a disclosure button connected right which toggles a popover action list with related actions.

| Split button | Split button with disclosure active |
|---|---|
|Screen Shot 2019-10-18 at 4 33 15 PM|Screen Shot 2019-10-18 at 4 33 46 PM|

<details>
<summary>Other UX considered</summary>

Another possible UX could be that the action list acts as a way to choose which action the primary button renders. You may be familiar with that type of split button if you've ever created a draft PR here on GitHub. The currently selected action has a selected state in the action list. Clicking a related action closes the popover and updates the primary action to be that clicked action. 

cc @ShaniqueS @mirualves 

| Initial primary action | Related primary action selected |
|---|---|
|Primary action|Primary action updated to selected related action|

</details>

|| Split button variants |
|---|---|
|Current|Screen Shot 2019-12-05 at 6 04 33 PM|
|Light mode|Screen Shot 2019-12-05 at 6 04 44 PM|
|Dark mode|Screen Shot 2019-12-05 at 6 04 54 PM|

## 

### How to 🎩

#### To tophat in Chroma deployment

- <a href="https://www.chromaui.com/component?appId=5d559397bae39100201eedc1&name=All%20Components%7CButton&buildNumber=1624&specName=Split%20Button" target="_blank" rel="noopener noreferrer">Click to view the split button example in a new tab</a>

#### To tophat locally
- `git checkout button-connectedSegment && dev up && yarn dev`
  - A new tab or window with the Storybook examples will open automatically 
- Open `polaris-react` in your text editor and copy/paste the Playground code collapsed below into `/playground/Playground.tsx`
  - Click "Playground" in the left navigation of the Storybook to view the playground

<details>

```jsx
import React from 'react';
import {Page, Card, ButtonGroup, Button} from '../src';

  return (
    <Page title="Playground">
      <Card sectioned>
        <ButtonGroup>
          <Button
            onClick={() => console.log('Draft purchase order saved')}
            primary
            connectedDisclosure={{
              actions: [
                {
                  content: 'Save and mark as ordered',
                  onAction: () => console.log('PO saved and marked as ordered'),
                },
              ],
            }}
          >
            Primary
          </Button>
          <Button
            onClick={() => console.log('Draft purchase order saved')}
            connectedDisclosure={{
              actions: [
                {
                  content: 'Save and mark as ordered',
                  onAction: () => console.log('PO saved and marked as ordered'),
                },
              ],
            }}
          >
            Basic
          </Button>
          <Button
            outline
            onClick={() => console.log('Draft purchase order saved')}
            connectedDisclosure={{
              actions: [
                {
                  content: 'Save and mark as ordered',
                  onAction: () => console.log('PO saved and marked as ordered'),
                },
              ],
            }}
          >
            Outline
          </Button>
          <Button
            destructive
            onClick={() => console.log('Draft purchase order saved')}
            connectedDisclosure={{
              actions: [
                {
                  destructive: true,
                  content: 'Save and mark as ordered',
                  onAction: () => console.log('PO saved and marked as ordered'),
                },
              ],
            }}
          >
            Destructive
          </Button>
          <Button
            outline
            destructive
            onClick={() => console.log('Draft purchase order saved')}
            connectedDisclosure={{
              actions: [
                {
                  destructive: true,
                  content: 'Save and mark as ordered',
                  onAction: () => console.log('PO saved and marked as ordered'),
                },
              ],
            }}
          >
            Destructive
          </Button>
        </ButtonGroup>
      </Card>
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -180,7 +180,7 @@ $stacking-order: (
   @include recolor-icon(color('white'));
 
   &.globalTheming {
-    @include recolor-icon(var(--p-icon-on-branded));
+    @include recolor-icon(var(--p-button-text));
   }
 
   &.disabled {
@@ -217,6 +217,10 @@ $stacking-order: (
       $destructive-pressed-darkest
     );
   }
+
+  &.globalTheming {
+    @include recolor-icon(var(--p-button-text));
+  }
 }
 
 .outline {
@@ -229,7 +233,6 @@ $stacking-order: (
 
 .destructive.outline {
   @include button-outline(color('red'));
-  @include recolor-icon(color('red', 'dark'));
 
   &.pressed {
     @include button-outline(
@@ -671,7 +674,19 @@ $stacking-order: (
   border-top-left-radius: 0;
   border-bottom-left-radius: 0;
 
-  &:focus {
+  // Because the outline border color has a 40% opacity, the left border appears darker than the rest of the borders because they are layered over one another. Reducing the opacity to zero for the connected disclosure when not focused gives us the expected border color.
+  &.outline:not(:focus) {
+    border-left-color: rgba(color('ink', 'lighter'), 0);
+  }
+
+  &:focus,
+  &:active {
     z-index: z-index(focused, $stacking-order);
+  }
+
+  &.globalTheming {
+    margin-left: rem(1px);
+    border-top-left-radius: 0;
+    border-bottom-left-radius: 0;
   }
 }
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -63,11 +63,26 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
   }
 }
 
+$stacking-order: (
+  segment: 10,
+  focused: 20,
+);
+
 .Button {
   @include button-base;
 
   &.disabled {
     @include base-button-disabled;
+  }
+
+  &.connectedDisclosure {
+    z-index: z-index(segment, $stacking-order);
+    border-top-right-radius: 0;
+    border-bottom-right-radius: 0;
+
+    &:focus {
+      z-index: z-index(focused, $stacking-order);
+    }
   }
 }
 // stylelint-disable selector-max-combinators
@@ -643,4 +658,20 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
 
 .DisclosureIconFacingUp {
   transform: rotate(-180deg);
+}
+
+.ConnectedDisclosureWrapper {
+  display: flex;
+  align-items: center;
+}
+
+.ConnectedDisclosure {
+  z-index: z-index(segment, $stacking-order);
+  margin-left: -(border-width());
+  border-top-left-radius: 0;
+  border-bottom-left-radius: 0;
+
+  &:focus {
+    z-index: z-index(focused, $stacking-order);
+  }
 }
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -242,6 +242,7 @@ export function Button({
       connectedDisclosure.disabled && styles.disabled,
       styles.iconOnly,
       styles.ConnectedDisclosure,
+      unstableGlobalTheming && styles.globalTheming,
     );
 
     const defaultLabel = i18n.translate(
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -1,20 +1,22 @@
-import React, {useRef} from 'react';
+import React, {useRef, useState, useCallback} from 'react';
 import {CaretDownMinor} from '@shopify/polaris-icons';
+
 import {classNames, variationName} from '../../utilities/css';
 import {handleMouseUpByBlurring} from '../../utilities/focus';
 import {useFeatures} from '../../utilities/features';
 import {useI18n} from '../../utilities/i18n';
 import {UnstyledLink} from '../UnstyledLink';
 import {Icon} from '../Icon';
-import {IconProps} from '../../types';
+import {IconProps, ConnectedDisclosure} from '../../types';
 import {Spinner} from '../Spinner';
+import {Popover} from '../Popover';
+import {ActionList} from '../ActionList';
+
 import styles from './Button.scss';
 
-type Size = 'slim' | 'medium' | 'large';
-
-type TextAlign = 'left' | 'right' | 'center';
-
-type IconSource = IconProps['source'];
+export type Size = 'slim' | 'medium' | 'large';
+export type TextAlign = 'left' | 'right' | 'center';
+export type IconSource = IconProps['source'];
 
 export interface ButtonProps {
   /** The content to display inside the button */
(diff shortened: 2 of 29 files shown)


tmlayton:
I dig this new prop!

src/components/Button/Button.tsx
```diff
@@ -205,9 +211,71 @@ export function Button({
...
+    );
+
+    const {
+      disabled,
+      accessibilityLabel: disclosureLabel = defaultLabel,
+    } = connectedDisclosure;
+
+    const connectedDisclosureActivator = (
+      <button
+        type={type}
```
kyledurand: I think we might be safer by typing this as button vs passing the prop.

BPScott:
@dleroux has some interesting stuff coming up that might allow us to do this split style with a ButtonGroup instead of a needing a new prop, you two should talk

src/components/Button/Button.tsx
chloerice: 💯

chloerice:
> @dleroux has some interesting stuff coming up that might allow us to do this split style with a ButtonGroup instead of a needing a new prop, you two should talk

Thanks @BPScott! Dan reached out to me yesterday 👍 He's got some kinks to work out with the way the context works, but if it get's figured out and folks are on board with the asterisks then I'll end up closing this and opening a new PR adding a split button example to `Button` using the new `ButtonGroup` context.

dleroux:
It’s now ready for review https://github.com/Shopify/polaris-react/pull/2441

dleroux:
It's ready for review again and much simpler https://github.com/Shopify/polaris-react/pull/2441

simon-callsen:
👋 Will this split action work when the button isn't `Primary`? We have a potential use case for this component when it's live that can be primary/not primary. 
image

chloerice:
> 👋 Will this split action work when the button isn't `Primary`? We have a potential use case for this component when it's live that can be primary/not primary.
> image

Hey @simon-callsen 👋, yes the `connectedDisclosure` prop can be used with any button variant 🎉

src/styles/shared/_buttons.scss
```diff
@@ -201,6 +201,7 @@
   border-color: rgba($outline-color, 0.4);
   box-shadow: none;
   color: darken($outline-color, 20%);
+  @include recolor-icon(darken($outline-color, 20%));
```
chloerice: Right now, a destructive outline icon button will have a bright red icon fill and darkened text. This isn't a variant represented in the style guide or in the UI kit, so I think this was just an oversight.

src/styles/shared/_buttons.scss
```diff
@@ -247,6 +248,7 @@
       background: transparent;
       box-shadow: 0 0 0 border-width('base') var(--p-critical-border);
       color: var(--p-critical-link);
+      @include recolor-icon(var(--p-critical-link));
```
chloerice: Again, parity between destructive button icon and text colors. cc @jessebc and @mirualves

src/styles/shared/_buttons.scss
mirualves: On the UI kit, the icon on the basic button is two tones lighter `Ink lighter` while the text is `Ink`, so I think the example on "Before" is actually correct.
Aso, the outline seems lighter than what's in the style guide.
Screen Shot 2019-12-06 at 9 42 26 AM

src/styles/shared/_buttons.scss
mirualves: Yeah, the after makes sense to me. The contrast looks much better in both light and dark mode.

ShaniqueS: 👀 Just a random question, but why are the outline and the text for the dark mode button two separate colors?

src/styles/shared/_buttons.scss
chloerice: Hmmm, those screenshots are from the style guide, is your screenshot an outline button or a monochrome button?

src/components/Button/Button.scss
```diff
@@ -165,7 +180,7 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
   @include recolor-icon(color('white'));
 
   &.globalTheming {
-    @include recolor-icon(var(--p-icon-on-branded));
+    @include recolor-icon(var(--p-button-text));
```
danrosenthal:
```suggestion
    @include recolor-icon(var(--p-icon-on-primary));
```

src/components/Button/Button.scss
```diff
@@ -202,6 +217,10 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
       $destructive-pressed-darkest
     );
   }
+
+  &.globalTheming {
+    @include recolor-icon(var(--p-button-text));
```
danrosenthal:
```suggestion
    @include recolor-icon(var(--p-icon-on-critical));
```

src/styles/shared/_buttons.scss
danrosenthal: This should use 
```suggestion
      @include recolor-icon(var(--p-critical-icon));
```

src/components/Button/Button.scss
```diff
@@ -643,3 +661,31 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
...
+
+.ConnectedDisclosure {
+  z-index: z-index(segment, $stacking-order);
+  margin-left: -(border-width());
+  border-top-left-radius: 0;
+  border-bottom-left-radius: 0;
+
+  // Because the outline border color has a 40% opacity, the left border appears darker than the rest of the borders because they are layered over one another. Reducing the opacity to zero for the connected disclosure when not focused gives us the expected border color.
+  &.outline:not(:focus) {
+    border-left-color: rgba(color('ink', 'lighter'), 0);
```
danrosenthal: Keep in mind, functions like `rgba` and `darken` won't work under the new color system with CSS Custom Properties, since theming happens client side, and these functions are all run at build time.

src/styles/shared/_buttons.scss
danrosenthal: Be careful of relying on functions like `darken` (see previous comment re: rgba/darken). This is also a smell that a color is not accounted for in our palette/color system, since we have to transform existing colors to create it.

src/styles/shared/_buttons.scss
tmlayton: The short answer is for contrast ratios. But it is possible these colors will still be tweaked—it is more important the code is using the proper color role variable.

src/styles/shared/_buttons.scss
mirualves: My screenshot is from the outline monochrome button (style guide). I'm sorry for the late reply!

PR #2330, opened 2019-10-18
Don't memoise UnstyledLink

BPScott:

Removing an unneeded usage of React.memo

Removing an unneeded usage of React.memo - memo here has no meaningful perf affect so lets not bother

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -29,5 +29,6 @@
 - Converted `Focus` into a functional component ([#2540](https://github.com/Shopify/polaris-react/pull/2540))
 - Converted `Tooltip` into a functional component ([#2543](https://github.com/Shopify/polaris-react/pull/2543))
 - Converted `Option` into a functional component ([#2541](https://github.com/Shopify/polaris-react/pull/2541))
+- Removed memoisation of `UnstyledLink` as it has no meaningful effect ([#2330](https://github.com/Shopify/polaris-react/pull/2330))
 
 ### Deprecations
--- a/src/components/UnstyledLink/UnstyledLink.tsx
+++ b/src/components/UnstyledLink/UnstyledLink.tsx
@@ -7,15 +7,8 @@ import {useLink, LinkLikeComponentProps} from '../../utilities/link';
 // but the props explorer isn't smart enough to work that out
 export interface UnstyledLinkProps extends LinkLikeComponentProps {}
 
-// Wrapping forwardRef in a memo gets a name set since
-// https://github.com/facebook/react/issues/16722
-// but eslint-plugin-react doesn't know that just yet
-// eslint-disable-next-line react/display-name
-export const UnstyledLink = React.memo(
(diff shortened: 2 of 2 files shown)


UNRELEASED.md
```diff
@@ -34,6 +34,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Code quality
 
+- Dont memoise UnstyledLink as it has no meaningful effect ([#2330](https://github.com/Shopify/polaris-react/pull/2330))
```
kaelig: This item should follow the changelog guidelines

BPScott:
We also don't know if the render of the custom link provided to AppProvider should even be skipped - that decision sohuld be up to the link component that gets provided, we're being a rather bold by assuming that consumers would want it always skipped.

AndrewMusgrave:
> We also don't know if the render of the custom link provided to AppProvider should even be skipped - that decision should be up to the link component that gets provided, we're being a rather bold by assuming that consumers would want it always skipped.

I don't think we should be promoting the use of impure components, and UnstyledLink has been pure for as we've been accepting customing lists, without any issues regarding this. So I don't think our assuming is a big deal.

I don't disagree that the performance impact is small. I saw render times up to 3ms and usually under 1ms in dev, which'll be much faster. But is there more of a benefit to remove it?

With that being said the longer render times could have been coming from internationalization, maybe dates? Which should be better handled by the consumer.

BPScott:
> But is there more of a benefit to remove it?

See https://github.com/Shopify/polaris-react/issues/2062. My conversations with web-foundation-tech folk have erred towards saying "you shouldn't use PureComponent and React.memo unless you profile it and prove that it helps". This viewpoint is echoed by the wider industry in https://kentcdodds.com/blog/usememo-and-usecallback and https://reacttraining.com/blog/react-inline-functions-and-performance/ which report that blind usage of PureComponent and React.memo can be actually harmful to performance (as you're doing two diffs instead of one).

A lot of our usage of PureComponent and React.memo looks like we've just slapped it on without actually checking for performance improvements and I'd like to go back and reconsider that behaviour.

To quote the second article I linked:

>When I first learned about PureRenderMixin (the thing from earlier versions of React that later became PureComponent) I put in a bunch of measurements and measured my app’s performance. I then added PureRenderMixin to every single component. When I took the optimized set of measurements I was hoping to have a cool story to tell about how much faster everything got.
>
> Much to my surprise, my app got slower 🤔.
>
> Why? Well, think about it. If you have a Component how many diffs are there? If you have a PureComponent how many diffs are there? The answers are “just one” and “at least one and sometimes two”, respectively. If a component usually changes when there’s an update, then a PureComponent will be doing two diffs instead of just one (props and state in shouldComponentUpdate, and then the normal element diff). Which means it’s going to be slower usually but faster occasionally. Apparently, most of my components changed most of the time, so on the whole, my app got slower. Oops.

PR #2335, opened 2019-10-21
[Color system] backwards compatibly implement color system for plain button

danrosenthal:
- Addresses https://github.com/Shopify/polaris-react/issues/2197
- Does not implement color system for `pressed` state for plain `Button`
- ~Adds a focus style for `Link` **outside** of global theming~

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -247,25 +247,56 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
 // stylelint-disable selector-max-specificity
 
 .plain {
-  @include recolor-icon(color('blue'));
+  @include recolor-icon(var(--p-interactive-action, color('blue')));
   margin: (-1 * $vertical-padding) rem(-8px);
   padding-left: spacing(tight);
   padding-right: spacing(tight);
   background: transparent;
   border: 0;
   box-shadow: none;
-  color: color('blue');
+  color: var(--p-interactive-action, color('blue'));
+
+  .Content {
+    font-weight: 400;
+  }
 
   &.pressed,
   &:hover,
   &:focus,
   &:active {
-    @include recolor-icon(color('blue', 'dark'));
     background: transparent;
     border: 0;
     box-shadow: none;
+  }
+
+  // TODO: implement pressed for global theming
+  &.pressed {
+    @include recolor-icon(color('blue', 'dark'));
     color: color('blue', 'dark');
     text-decoration: underline;
+  }
+
+  &:hover {
+    @include recolor-icon(
+      var(--p-interactive-action-hovered, color('blue', 'dark'))
+    );
+    color: var(--p-interactive-action-hovered, color('blue', 'dark'));
+    text-decoration: underline;
+  }
+
+  &:focus {
+    @include recolor-icon(var(--p-interactive-action, color('blue', 'dark')));
+    color: var(--p-interactive-action, color('blue', 'dark'));
+    text-decoration: var(--p-override-none, underline);
+    @include no-focus-ring;
+  }
+
+  &:active {
+    @include recolor-icon(
+      var(--p-interactive-action-pressed, color('blue', 'dark'))
+    );
+    color: var(--p-interactive-action-pressed, color('blue', 'dark'));
+    text-decoration: var(--p-override-none, underline);
   }
 
   &.pressed,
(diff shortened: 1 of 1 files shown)


src/components/Button/Button.scss
```diff
@@ -247,27 +247,58 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
...
   padding-left: spacing(tight);
   padding-right: spacing(tight);
   background: transparent;
   border: 0;
   box-shadow: none;
-  color: color('blue');
+  color: var(--p-interactive-action, color('blue'));
+
+  .Content {
+    font-weight: 400;
```
tmlayton: This applies to both current and new theming?

src/components/Button/Button.scss
```diff
@@ -247,27 +247,58 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
...
   &:hover,
   &:focus,
   &:active {
-    @include recolor-icon(color('blue', 'dark'));
     background: transparent;
     border: 0;
     box-shadow: none;
+  }
+
+  // TODO: implement pressed for masterbrand
```
tmlayton: Does this mean the interaction states? Also, maybe refer to it as global theming.

src/components/Button/Button.scss
```diff
@@ -277,16 +308,29 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
     @include high-contrast-button-outline;
   }
 
+  &:focus:not(:active) > .Content {
```
tmlayton: Side note there are a few places where I did `&:focus { @include focus-ring }` and `&:active { @include no-focus-ring }`. Those can likely be refactored to `&:focus:not(:active) { @include focus-ring }`.

src/components/Button/Button.scss
```diff
@@ -277,16 +308,29 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
...
 
   &.pressed:hover:not(.iconOnly) > .Content,
-  &.pressed:active:not(.iconOnly) > .Content,
-  &:focus:not(.iconOnly) > .Content {
+  &.pressed:active:not(.iconOnly) > .Content {
     @include plain-button-backdrop;
   }
 
+  &:focus:not(.iconOnly) > .Content {
+    @include plain-button-backdrop(
```
tmlayton: This and the below `:active` selector add padding and negative margin to styles to those states which previous had none. I noticed in Firefox this causes the text to wrap if it is at the edge of its container:

firefox wrap

src/components/Link/Link.scss
```diff
@@ -8,15 +8,32 @@
...
   }
+
+  &:focus {
+    text-decoration: var(--p-override-none, underline);
+    outline: none;
+  }
+
+  &:focus:not(:active) {
+    @include focus-ring;
+    @include plain-button-backdrop(
```
tmlayton: Might be the same wrapping issue like above with adding the padding and negative margin to this state which previously did not have it.

tmlayton: Looks like this is only an issue with `Link` and not `Button`. `Button` already has a base padding and negative margin so it doesn’t change the box model during active or focus. I think to achieve these focus and active styles with `Link` it will require using the `before` or `after` pseudo element.

tmlayton: Or maybe we could make the base `Link` box model the same as the focus and hover styles, although that change feels riskier.

tmlayton:
Besides for the `Link` box model issue, 🎩 looks good.

src/components/Button/Button.scss
danrosenthal: It does. This does not represent a change from the existing implementation (since it defaulted to this weight), and is needed to override font weight applied to all buttons.

src/components/Button/Button.scss
danrosenthal:
```suggestion
  // TODO: implement pressed for global theming
```

danrosenthal: This means the 'depressed' state for `Button` that was recently added. I wanted to punt on implementing that until we have designs nailed down, and the right color variants to cover it.

src/components/Link/Link.scss
danrosenthal: Bummer. I was worried this might happen. I think it makes sense to cherry pick out the plain `Button` changes, and implement `Link` later on. Backwards compatibility becomes tricky, because we have to remove the existing browser focus ring on the new styles. But that becomes tricky (impossible?) to do backwards compatibly. We can specify a `outline` style, but this will be somewhat of a regression from what exists today, which is just whatever the browser has as a default.

Let's chat about this once you're online.

danrosenthal: I moved `Link` over to this PR: https://github.com/Shopify/polaris-react/pull/2342, so we can handle focus styles without blocking plain `Button`.

src/components/Button/Button.scss
```diff
@@ -247,27 +247,58 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
...
   &:hover,
   &:focus,
   &:active {
-    @include recolor-icon(color('blue', 'dark'));
     background: transparent;
     border: 0;
     box-shadow: none;
+  }
+
+  // TODO: implement pressed for global theming
```
tmlayton:
```suggestion
  // TODO: implement depressed for global theming
```

danrosenthal:
```suggestion
  // TODO: implement depressed state for global theming
```

src/components/Button/Button.scss
```diff
@@ -247,27 +247,58 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
...
   &:hover,
   &:focus,
   &:active {
-    @include recolor-icon(color('blue', 'dark'));
     background: transparent;
     border: 0;
     box-shadow: none;
+  }
+
+  // TODO: implement depressed for global theming
```
danrosenthal:
```suggestion
  // TODO: implement depressed state for global theming
```

PR #2306, opened 2019-10-16
[Color system] set theme css custom properties on portal node

danrosenthal:
Co-authored-by: Tim Layton <tim.layton@shopify.com>

### WHY are these changes introduced?

This PR enables components rendered in a `Portal` to be themed using the generated CSS custom properties. Otherwise components rendered in a `Portal` are out of scope (rendered outside of the `ThemeProvider`) and do not have access to the CSS custom properties.

#2224 was the first attempt at solving this issue which moved the `Portal` container within the `ThemeProvider`. But this broke `Popover` positioning if an app had a positioned wrapper (e.g `position: relative; top: 50px`, see #2302) and was reverted in #2305.

### WHAT is this pull request doing?

1. Adds the generated CSS custom properties used for global theming to the theme context
1. Sets the CSS custom properties on the `Portal` container

- [x] 🎩

PR #2337, opened 2019-10-22
[Color system] Add tests for adding and removing styles

tmlayton:

Adds a couple more tests for work done in #2306 and #2327 which expect that we are adding and removing the CSS custom property styles to and from the portal node. 

By spying on the `Element` prototype we can get access to instance methods.

### DIFF
--- a/src/components/Portal/tests/Portal.test.tsx
+++ b/src/components/Portal/tests/Portal.test.tsx
@@ -96,15 +96,24 @@ describe('<Portal />', () => {
     }).not.toThrow();
   });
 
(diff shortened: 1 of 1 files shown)


src/components/Portal/tests/Portal.test.tsx
```diff
@@ -96,15 +96,24 @@ describe('<Portal />', () => {
     }).not.toThrow();
   });
 
-  it('renders okay with new theme context', () => {
```
tmlayton: The first test below is now a better version of this deleted test.

PR #2339, opened 2019-10-22
[TextField] delete irrelevant default numeric-range props, prevent unconditional aria-invalid='false'

### LINKED ISSUE #2282: [TextField] `min`, `max`, `step` and `aria-invalid="false"` shouldn't appear in the markup
## Issue summary

Attributes `min`, `max` and `step` should only appear in the generated markup when appropriate.

~It's also possible that we could get rid of `aria-invalid="false"`: @dpersing what do you think?~

Edit: we should also get rid of `aria-invalid="false"`, as per @dpersing's comment below.

## Expected behavior

```html
<input id="PolarisTextField2" class="Polaris-TextField__Input" value="Jaded Pixel">
```

## Actual behavior

```
// These attributes show up in the markup:
min="-Infinity" max="Infinity" step="1" aria-invalid="false"
```

```
<input id="PolarisTextField2" class="Polaris-TextField__Input" min="-Infinity" max="Infinity" step="1" aria-invalid="false" value="Jaded Pixel">
```

1. Go to https://polaris.shopify.com/components/forms/text-field
1. In the code playground, switch to the HTML view

AlexCLeduc:

Fixes #2282. I made a single PR because it was one issue, but maybe it should be split in two

1. Removes default props for `max`, `min`, `step` so they no longer show up in markup unless explicitly passed
2. Prevents `aria-invalid="false"` from being rendered when there is no error

### DIFF
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -144,11 +144,11 @@ export function TextField({
   name,
   id: idProp,
   role,
-  step = 1,
+  step,
   autoComplete,
-  max = Infinity,
+  max,
   maxLength,
-  min = -Infinity,
+  min,
   minLength,
   pattern,
   spellCheck,
@@ -183,6 +183,11 @@ export function TextField({
 
   const normalizedValue = value != null ? value : '';
   const {unstableGlobalTheming = false} = useFeatures();
+
+  const normalizedStep = step != null ? step : 1;
+  const normalizedMax = max != null ? max : Infinity;
+  const normalizedMin = min != null ? min : -Infinity;
+
   const className = classNames(
     styles.TextField,
     Boolean(normalizedValue) && styles.hasValue,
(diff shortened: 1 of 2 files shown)


src/components/TextField/TextField.tsx
```diff
@@ -182,7 +182,12 @@ export function TextField({
   }, [focused]);
 
   const normalizedValue = value != null ? value : '';
+  const normalizedStep = step != null ? step : 1;
+  const normalizedMax = max != null ? max : Infinity;
+  const normalizedMin = min != null ? min : -Infinity;
+
```
dleroux:
```suggestion
```

src/components/TextField/TextField.tsx
```diff
@@ -182,7 +182,12 @@ export function TextField({
   }, [focused]);
 
   const normalizedValue = value != null ? value : '';
+  const normalizedStep = step != null ? step : 1;
+  const normalizedMax = max != null ? max : Infinity;
+  const normalizedMin = min != null ? min : -Infinity;
+  
   const {unstableGlobalTheming = false} = useFeatures();
+
```
dleroux:
```suggestion
```

dleroux:
Thank you for doing this @AlexCLeduc! There was a conflict I tried to resolve using Github but it introduced formatting issues in ci. Ping me when those are updated and I'll approve.

AlexCLeduc:
Thanks @dleroux! I've just rebased over master

PR #2341, opened 2019-10-22
[Modal] Convert the modal to a function

LauraAubin:

Some notable changes,

- The activator now uses `useRef` instead of `createRef`.
- The external activator example needs to call `requestAnimationFrame` when focusing the button.

### How to 🎩

### 🎩 checklist

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -213,7 +213,3 @@ export const Modal: React.FunctionComponent<ModalProps> & {
 };
 
 Modal.Section = Section;
-
-// Use named export once withAppProvider is refactored away
-// eslint-disable-next-line import/no-default-export
-// export default Modal;
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useState} from 'react';
 import {TransitionGroup} from '@material-ui/react-transition-group';
 import {write} from '@shopify/javascript-utilities/fastdom';
 import {focusFirstFocusableNode} from '@shopify/javascript-utilities/focus';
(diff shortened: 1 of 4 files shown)


src/components/Modal/Modal.tsx
```diff
@@ -64,178 +64,156 @@ export interface ModalProps extends FooterProps {
...
-          onLoad={this.handleIFrameLoad}
-          style={{height: `${iframeHeight}px`}}
-        />
-      ) : (
-        <Scrollable
-          shadow
-          className={styles.Body}
-          onScrolledToBottom={onScrolledToBottom}
+const Modal: React.FunctionComponent<CombinedProps> & {
+  Section: typeof Section;
```
LauraAubin: The `Card` explains why this typing is needed:

> TypeScript can't generate types that correctly infer the typing of subcomponents so explicitly state the subcomponents in the type definition. Letting this be implicit works in this project but fails in projects that use generated *.d.ts files.

src/components/Modal/Modal.tsx
```diff
@@ -64,178 +64,156 @@ export interface ModalProps extends FooterProps {
...
+  onScrolledToBottom,
+  activator,
+  onClose,
+  onIFrameLoad,
+  onTransitionEnd,
+}: CombinedProps) {
+  const [iframeHeight, setIframeHeight] = useState(IFRAME_LOADING_HEIGHT);
+
+  const headerId = getUniqueID();
+  const activatorRef = React.useRef<HTMLDivElement>(null);
```
LauraAubin: Should these still be private?

src/components/Modal/Modal.tsx
```diff
@@ -64,178 +64,156 @@ export interface ModalProps extends FooterProps {
...
+  polaris: {intl},
+  onScrolledToBottom,
+  activator,
+  onClose,
+  onIFrameLoad,
+  onTransitionEnd,
+}: CombinedProps) {
+  const [iframeHeight, setIframeHeight] = useState(IFRAME_LOADING_HEIGHT);
+
+  const headerId = getUniqueID();
```
BPScott: getUniqueID/createUniqueIdFactory is  problematic as it's not stable between rerenders (i.e. every time you rerender the component then you get a new Id which causes pointless changes to the DOM). Use the useUniqueId hook instead. See https://github.com/Shopify/polaris-react/pull/2079 for more info

```suggestion
  const headerId = useUniqueId('modal-header');
```

src/components/Modal/Modal.tsx
```diff
@@ -64,178 +64,156 @@ export interface ModalProps extends FooterProps {
...
+  open,
+  instant,
+  sectioned,
+  loading,
+  large,
+  limitHeight,
+  footer,
+  primaryAction,
+  secondaryActions,
+  polaris: {intl},
```
BPScott: A big part of moving towards functional components is that we can remove the withAppProvider HoC that provides this polaris prop. We've got a handful of custom hooks that replace these values of polaris

Remove this and the withAppProvider Hoc and instead do:

```
const i18n = useI18n();

// ...

i18n.translate('Polaris.SomeKey');
```

src/components/Modal/Modal.tsx
```diff
@@ -64,178 +64,156 @@ export interface ModalProps extends FooterProps {
...
+      <Portal idPrefix="modal">
+        <TransitionGroup appear={animated} enter={animated} exit={animated}>
+          {dialog}
+        </TransitionGroup>
+        {backdrop}
+      </Portal>
+    </WithinContentContext.Provider>
+  );
+
+  function handleEntered() {
```
BPScott: These functions need to be wrapped in useCallback so we don't trigger rerenders of the elements that they get passed into

PR #2346, opened 2019-10-22
Stop generating Sass.zip

kaelig:
We're moving the Sass.zip generation to polaris-styleguide.

Another PR is coming in polaris-styleguide to generate the Sass.zip, and remove the need for uploading any assets with the GitHub releases.

This change makes it less convenient to download the Sass.zip for older versions of Polaris. People who want to get an older version of the Sass files will have to go through the trouble of downloading a package via this type of URL:

```
https://registry.yarnpkg.com/@shopify/polaris/-/polaris-POLARIS_VERSION.tgz
```

For example:

https://registry.yarnpkg.com/@shopify/polaris/-/polaris-4.7.0.tgz

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -16,10 +16,10 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Development workflow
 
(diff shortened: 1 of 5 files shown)


kaelig:
Thanks for the review! First, [this PR in polaris-styleguide](https://github.com/Shopify/polaris-styleguide/pull/3065) needs to get merged.

PR #2347, opened 2019-10-23
[Color system] Update surface colors for data table

tmlayton:
Knocking off low hanging fruit for surface colors


src/components/DataTable/DataTable.scss
```diff
@@ -151,7 +151,7 @@ $breakpoint: 768px;
 
 .Cell-total {
   @include text-emphasis-strong;
-  background: color('sky', 'lighter');
+  background: var(--p-surface-foreground, color('sky', 'lighter'));
```
danrosenthal: Since these are `sky, lighter`, not `white`, I think they should be surface-foreground-subdued. Going to add a change that cleans up subdued a bit so that looks truer to its intent.

src/components/DataTable/DataTable.scss
```diff
@@ -162,7 +162,7 @@ $breakpoint: 768px;
 
 .Footer {
   padding: spacing();
-  background: color('sky', 'light');
+  background: var(--p-surface-foreground, color('sky', 'lighter'));
```
danrosenthal: Likewise this should be a subdued foreground. I'm also going to update the text color so the contrast is correct.

src/components/DataTable/DataTable.scss
```diff
@@ -162,7 +162,7 @@ $breakpoint: 768px;
 
 .Footer {
   padding: spacing();
-  background: color('sky', 'light');
-  color: color('ink', 'lighter');
+  background: var(--p-surface-foreground-subdued, color('sky', 'lighter'));
```
tmlayton: I prefer not be be changing fallback values as part of this PR

danrosenthal:
```suggestion
  background: var(--p-surface-foreground-subdued, color('ink', 'lighter'));
```

src/components/DataTable/DataTable.scss
```diff
@@ -162,7 +162,7 @@ $breakpoint: 768px;
 
 .Footer {
   padding: spacing();
-  background: color('sky', 'light');
-  color: color('ink', 'lighter');
+  background: var(--p-surface-foreground-subdued, color('ink', 'lighter'));
```
danrosenthal:
```suggestion
  background: var(--p-surface-foreground-subdued, color('sky', 'light'));
```

PR #2350, opened 2019-10-23
[Color system] Revert button changes

tmlayton:
Co-authored-by: @danrosenthal 

Fixes https://github.com/Shopify/web/issues/20016. Partially reverts 834a741 due to a bug related to https://bugs.webkit.org/show_bug.cgi?id=194332

Partial revert of #2252. It appears using a `var` function in CSS with a fallback to a linear gradient in combination with some other styles on our `Button` component cause an infinite layout loop and pegs the CPU.

### 🎩 checklist

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -127,14 +127,10 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
 }
 
 .primary {
-  --p-button-color: var(--p-branded-action);
-  --p-button-color-hover: var(--p-branded-action-hovered);
-  --p-button-color-active: var(--p-branded-action-pressed);
   @include button-filled(color('indigo'), color('indigo', 'dark'));
   @include recolor-icon(color('white'));
 
   &.disabled {
-    --p-button-color-disabled: var(--p-branded-action-disabled);
     @include button-filled-disabled(color('indigo'));
   }
 
--- a/src/styles/shared/_buttons.scss
+++ b/src/styles/shared/_buttons.scss
@@ -17,15 +17,16 @@
   min-width: $min-height;
   margin: 0;
   padding: $vertical-padding spacing();
-  background: var(
-    --p-neutral-action,
-    linear-gradient(to bottom, color('white'), color('sky', 'lighter'))
+  background: linear-gradient(
+    to bottom,
+    color('white'),
+    color('sky', 'lighter')
   );
-  border: var(--p-override-none, border(dark));
-  box-shadow: var(--p-override-none, shadow(faint));
-  border-radius: var(--p-border-radius-base, border-radius());
+  border: border(dark);
+  box-shadow: shadow(faint);
+  border-radius: border-radius();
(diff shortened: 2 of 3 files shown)


src/components/Button/Button.scss
```diff
@@ -127,14 +127,10 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
 }
 
 .primary {
-  --p-button-color: var(--p-branded-action);
-  --p-button-color-hover: var(--p-branded-action-hovered);
-  --p-button-color-active: var(--p-branded-action-pressed);
```
danrosenthal: We should be setting `--p-button-text` here so that `critical` can set its own text color

src/styles/shared/_buttons.scss
```diff
@@ -36,121 +37,96 @@
...
+    darken($button-color, 2%)
   );
-  border-color: var(--p-override-transparent, $border-color);
-  box-shadow: var(
-    --p-override-none,
-    inset 0 1px 0 0 lighten($button-color, 3%),
-    shadow(faint),
-    0 0 0 0 transparent
-  );
-  color: var(--p-text-on-branded, color('white'));
```
danrosenthal: This should be `--p-button-text` to allow for variability

src/styles/shared/_buttons.scss
```diff
@@ -36,121 +37,96 @@
...
-    background: var(
-      --p-button-color-hover,
-      linear-gradient(to bottom, $button-color, darken($button-color, 5%))
+    background: linear-gradient(
+      to bottom,
+      $button-color,
+      darken($button-color, 5%)
     );
-    border-color: var(--p-override-transparent, $border-color);
-    color: var(--p-text-on-branded, color('white'));
```
danrosenthal: This should be `--p-button-text` to allow for variability

src/styles/shared/_buttons.scss
```diff
@@ -36,121 +37,96 @@
...
-    )
+  background: linear-gradient(
+    to bottom,
+    lighten($button-color, 25%),
+    lighten($button-color, 25%)
   );
-  border-color: var(--p-override-transparent, lighten($button-color, 20%));
+  border-color: lighten($button-color, 20%);
   box-shadow: none;
-  color: var(--p-text-on-branded, color('white'));
```
danrosenthal: This should be `--p-button-text` to allow for variability

src/styles/shared/_buttons.scss
```diff
@@ -36,121 +37,96 @@
...
-    );
+    background: linear-gradient(to bottom, $border-color, $border-color);
+    border-color: $active-color;
+    box-shadow: inset 0 0 0 0 transparent, shadow(faint),
+      0 0 1px 0 $active-color;
   }
 }
 
 @mixin button-filled-disabled($button-color) {
   @include recolor-icon(color('white'));
```
danrosenthal: We should also be handling all `recolor-icon` calls in this file.

PR #2356, opened 2019-10-24
Move color adjustments to JSON file

martenbjork:
**🚩 This is a highly speculative, low-confidence suggestion. Consider it an idea.**

With the new color system, ~10 colors are inputed and ~100 more are generated. These new colors are adjusted versions of the initial ones. These adjustments are hard-coded into `utils.ts`. It means that the *color factory* and the *color blueprint* are tightly coupled. This makes it hard to introduce new tooling around colors — if you want to play around with colors, your only option is to tweak `utils.ts`.

It extracts the adjustments to a separate JSON file. 

Before:

```js
actionOnDark: setLightness(color, 76),
actionOnLight: setLightness(color, 36),
```

After (now in `colorAdjustments.json`):
```js
"ACTION_ON_DARK": {
  "baseColor": "onSurface", // ACTION_ON_DARK is based on onSurface
  "light": {"lightness": 76}, // Apply this modification in light mode
  "dark": {"lightness": 76} // Apply this modification in dark mode
},
"ACTION_ON_LIGHT": {
  "baseColor": "onSurface",
  "light": {"lightness": 36},
  "dark": {"lightness": 36}
},
...
```

Utils.ts reads this file, creates the corresponding colors and returns the same CSS Custom Properties as before. 

### WHAT can we do now?
I can build a separate tool that allows designers (like me or @jessebc) to play with these values. As long as the tool outputs the same JSON file as above, Polaris can use it and generate Custom Properties from it.

### DIFF
--- a/src/utilities/theme/colorAdjustments.json
+++ b/src/utilities/theme/colorAdjustments.json
@@ -1,0 +1,546 @@
+{
+  "SURFACE": {"baseColor": "surface", "light": {}, "dark": {}},
+  "SURFACE_BACKGROUND": {
+    "baseColor": "surface",
+    "light": {"lightness": 98},
+    "dark": {"lightness": 7}
+  },
+  "SURFACE_FOREGROUND": {
+    "baseColor": "surface",
+    "light": {"lightness": 100},
+    "dark": {"lightness": 13}
+  },
+  "SURFACE_FOREGROUND_SUBDUED": {
+    "baseColor": "surface",
+    "light": {"lightness": 90},
+    "dark": {"lightness": 10}
+  },
+  "SURFACE_INVERSE": {
+    "baseColor": "surface",
+    "light": {"lightness": 0},
+    "dark": {"lightness": 100}
+  },
+  "SURFACE_HOVERED": {
+    "baseColor": "surface",
+    "light": {"lightness": 93},
+    "dark": {"lightness": 20}
+  },
+  "SURFACE_PRESSED": {
+    "baseColor": "surface",
+    "light": {"lightness": 86},
+    "dark": {"lightness": 27}
+  },
+  "ON_SURFACE": {"baseColor": "onSurface", "light": {}, "dark": {}},
+  "ACTION_ON_INVERSE": {
+    "baseColor": "onSurface",
+    "light": {"lightness": 76},
+    "dark": {"lightness": 36}
+  },
+  "ACTION_ON_SURFACE": {
+    "baseColor": "onSurface",
+    "light": {"lightness": 36},
+    "dark": {"lightness": 76}
+  },
+  "ACTION_DISABLED_ON_INVERSE": {
+    "baseColor": "onSurface",
+    "light": {"lightness": 66},
+    "dark": {"lightness": 46}
+  },
+  "ACTION_DISABLED_ON_SURFACE": {
+    "baseColor": "onSurface",
+    "light": {"lightness": 46},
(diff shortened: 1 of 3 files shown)


tmlayton:
As the aphorism goes "All problems in computer science can be solved by another level of indirection" 😄

The general idea I like, a data structure that represents each generated role and the manipulations that we map over based on the high level inputs. It’s more approachable for tweaking. The structure itself I would want to think about more. It is also tied to HSL which might change. That is just to say if this moves forward it depends on deciding the exact structure and color space.

kaelig:
I 100% agree with Tim 👍 

---

On a more general level, I'm not against the idea of storing data like this in a JSON file, as long as it comes with a schema to validate against.

In case you end up pursuing this route, here's what a schema looks like: https://github.com/Shopify/polaris-icons/blob/52b9ea555d2c8de7dda04d94bdea34ae78ea42ee/packages/polaris-icons-raw/tests/metadata-schema.json

We'd then be able to use a tool such as [ajv](https://www.npmjs.com/package/ajv) to validate the JSON's file conformance with the schema.

BPScott:
> On a more general level, I'm not against the idea of storing data like this in a JSON file, as long as it comes with a schema to validate against.

Do we have any plans to consume this data in any language other than JS? If not then it might be easier to stick it in a js/ts file and validate it using TS types

kaelig:
I was going to suggest JS or TS but this made me think JSON could be considered a better target:

>We can build separate tooling that allows designers (like me or @jessebc) to play with these values. As long as the tool outputs the same JSON file as above, Polaris can use it and generate Custom Properties from it.

martenbjork:
Thanks for the input, everyone!

> The structure itself I would want to think about more. It is also tied to HSL which might change.

To me, the structure of the JSON file doesn't imply increased confidence a particular color space — it just says that the "knobs" that you can tweak (i.e. the API interface) will be lightness/saturation (and maybe hue). The factory is still free to use apply these changes in any color space (as PR #2361 demonstrates).

---

To rephrase the essence of this PR: It would be good to have an "API" for the color factory so one can play with the adjustments outside the factory itself.

tmlayton:
The structure and values of the data are tied to a color space, though. For example, lightness can’t be defined in the abstract because it translates to different colors in HSL, HSLuv, and Lab. So you need to know which knob you are turning.

src/utilities/theme/colorAdjustments.json
```diff
@@ -0,0 +1,546 @@
...
+  },
+  "neutralActionPressed": {
+    "baseColor": "neutral",
+    "light": {"lightness": 76},
+    "dark": {"lightness": 39}
+  },
+  "branded": {"baseColor": "branded", "light": {}, "dark": {}},
+  "brandedAction": {
+    "baseColor": "branded",
+    "light": {"lightness": 25},
```
tmlayton: Just FYI this `25` lightness comes from manually pulling the lightness from the base jade color. In HSLuv, the lightness is `47.3`.

martenbjork: Ah, that's interesting and weird. I stand corrected. Will read up on the color spaces and pick up the discussion next week!

tmlayton:
Otherwise, yes an API for the factory is great. TBH we will probably go with HSL or HSLuv so the schema is likely fine 😄

tmlayton:
Besides for a rebase to sort out color roles that have since been added, this is good to go.

PR #2359, opened 2019-10-24
[Docs] Add internal hook docs

AndrewMusgrave:

* more awareness
* add-general,  simple examples

Skipping changelog since this is internal

### DIFF
--- a/src/utilities/use-component-did-mount.ts
+++ b/src/utilities/use-component-did-mount.ts
@@ -11,7 +11,7 @@ import {useIsAfterInitialMount} from './use-is-after-initial-mount';
  * function Playground({active}) {
  *  useComponentDidMount(() => {
  *    if (active) {
- *      console.warning(`Active is now deprecated.`);
+ *      console.warning(`Component has mounted.`);
  *    }
  *  });
  *
--- a/src/utilities/use-component-did-mount.ts
+++ b/src/utilities/use-component-did-mount.ts
@@ -3,10 +3,20 @@ import {EffectCallback} from '../types';
 import {useIsAfterInitialMount} from './use-is-after-initial-mount';
 
 /**
- * Similarly to the lifecycle hook componentDidMount, useComponentDidMount
+ * Similarly to the life-cycle method componentDidMount, useComponentDidMount
  * will be invoked after the component has mounted, and only the initial mount.
  * @param callback Defines a callback to invoke once the component has
  * initially mounted.
+ * @example
+ * function Playground({active}) {
+ *  useComponentDidMount(() => {
+ *    if (active) {
+ *      console.warning(`Active is now deprecated.`);
+ *    }
+ *  });
+ *
+ *  return null;
+ * }
  */
 export function useComponentDidMount(callback: EffectCallback) {
   const isAfterInitialMount = useIsAfterInitialMount();
--- a/src/utilities/use-deep-callback.tsx
+++ b/src/utilities/use-deep-callback.tsx
@@ -2,6 +2,25 @@ import {useCallback} from 'react';
 import {EffectCallback, DependencyList, Comparator} from '../types';
 import {useDeepCompareRef} from './use-deep-compare-ref';
 
+/**
+ * A replacement for React.useCallback that'll allow for custom and deep compares.
+ * @see {@link https://reactjs.org/docs/hooks-reference.html#usecallback}
+ * @param callback Accepts a callback that's forwarded to React.useCallback
+ * @param dependencies A dependency array similar to React.useCallback however it utilizes a deep compare
+ * @param customCompare Opportunity to provide a custom compare function
+ * @returns A memoized callback
+ * @example
+ * function ComponentExample() {
+ *  const [, forceUpdate] = useState();
+ *  const obj = {a: 1};
+ *
+ *  const handleClick = useDeepCallback(() => {
+ *    forceUpdate(obj);
+ *  }, [obj]);
+ *
+ *  return <button onClick={handleClick}>Click me</button>;
+ * }
+ */
 export function useDeepCallback(
   callback: EffectCallback,
   dependencies: DependencyList,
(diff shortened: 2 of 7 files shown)


src/utilities/use-component-did-mount.ts
```diff
@@ -3,10 +3,20 @@ import {EffectCallback} from '../types';
...
- * Similarly to the lifecycle hook componentDidMount, useComponentDidMount
+ * Similarly to the life-cycle method componentDidMount, useComponentDidMount
  * will be invoked after the component has mounted, and only the initial mount.
  * @param callback Defines a callback to invoke once the component has
  * initially mounted.
+ * @example
+ * function Playground({active}) {
+ *  useComponentDidMount(() => {
+ *    if (active) {
+ *      console.warning(`Active is now deprecated.`);
```
dleroux:
```suggestion
 *      console.warning(`Component has mounted.`);
```

src/utilities/use-deep-compare-ref.tsx
```diff
@@ -4,6 +4,18 @@ import isEqual from 'lodash/isEqual';
 type DependencyList = ReadonlyArray<unknown>;
 type Comparator = (a: DependencyList, b: DependencyList) => boolean;
 
+/**
+ * useDeepCompare is valuable to keep a consistent dependency list across reference changes
+ * @param dependencies A dependency array like React.useEffect/React.useCallback/React.useMemo
+ * @param comparator A optional function to compare dependencies that'll default to a deep comparison
+ * @returns A dependency list
```
dleroux:
```suggestion
 * @returns A dependency array
```

src/utilities/use-deep-compare-ref.tsx
```diff
@@ -4,6 +4,18 @@ import isEqual from 'lodash/isEqual';
 type DependencyList = ReadonlyArray<unknown>;
 type Comparator = (a: DependencyList, b: DependencyList) => boolean;
 
+/**
+ * useDeepCompare is valuable to keep a consistent dependency list across reference changes
+ * @param dependencies A dependency array like React.useEffect/React.useCallback/React.useMemo
```
dleroux:
```suggestion
 * @param dependencies A dependency array similar to React.useEffect/React.useCallback/React.useMemo
```

src/utilities/use-deep-compare-ref.tsx
```diff
@@ -4,6 +4,18 @@ import isEqual from 'lodash/isEqual';
 type DependencyList = ReadonlyArray<unknown>;
 type Comparator = (a: DependencyList, b: DependencyList) => boolean;
 
+/**
+ * useDeepCompare is valuable to keep a consistent dependency list across reference changes
```
dleroux: changes.
```suggestion
 * Allows for custom or deep comparison of a dependency list. Useful to keep a consistent dependency list across reference changes. 
```

dleroux: Not 100% sold on my suggested description either. I just think that simply outlining the value may not be the best description?

src/utilities/use-deep-effect.tsx
```diff
@@ -2,6 +2,26 @@ import {useEffect} from 'react';
 import {EffectCallback, DependencyList, Comparator} from '../types';
 import {useDeepCompareRef} from './use-deep-compare-ref';
 
+/**
+ * A replacement for React.useEffect that'll allow for custom and deep
+ * compares.
```
dleroux:
```suggestion
 * compares of the dependency list. 
```

src/utilities/use-component-did-mount.ts
```diff
@@ -3,10 +3,20 @@ import {EffectCallback} from '../types';
...
 
 /**
- * Similarly to the lifecycle hook componentDidMount, useComponentDidMount
+ * Similarly to the life-cycle method componentDidMount, useComponentDidMount
  * will be invoked after the component has mounted, and only the initial mount.
  * @param callback Defines a callback to invoke once the component has
  * initially mounted.
+ * @example
+ * function Playground({active}) {
+ *  useComponentDidMount(() => {
```
LauraAubin: Would it be worth expanding on why we would use this vs `useEffect` with an empty dependency array?

src/utilities/use-deep-compare-ref.tsx
```diff
@@ -4,6 +4,18 @@ import isEqual from 'lodash/isEqual';
...
+ * @param dependencies A dependency array like React.useEffect/React.useCallback/React.useMemo
+ * @param comparator A optional function to compare dependencies that'll default to a deep comparison
```
LauraAubin:
```suggestion
 * @param comparator An optional function to compare dependencies that'll default to a deep comparison
```

src/utilities/use-deep-callback.tsx
```diff
@@ -2,6 +2,25 @@ import {useCallback} from 'react';
...
+ * @param customCompare Opportunity to provide a custom compare function
+ * @returns A memoized callback
+ * @example
+ * function ComponentExample() {
+ *  const [, forceUpdate] = useState();
+ *  const obj = {a: 1};
+ *
+ *  const handleClick = useDeepCallback(() => {
+ *    forceUpdate(obj);
+ *  }, [obj]);
```
LauraAubin: I'm not sure I understand what this hook is doing. Would `forceUpdate` only be called once?

src/utilities/use-component-did-mount.ts
AndrewMusgrave: These docs were more meant to explain the hook usage, rather than the trade-offs between react API's. I'm going to leave it out for now, but if this comes up again we can look into adding a reference to the use-effect docs 😄

src/utilities/use-deep-compare-ref.tsx
AndrewMusgrave: Going keep it as a list for the moment since it's how react reference to it.

src/utilities/use-deep-callback.tsx
AndrewMusgrave: Added a new example to hopefully make this clearer 😄

PR #2361, opened 2019-10-25
[Color system] Convert colors to HSLuv

tmlayton:

Use a color space (HSLuv) which maintains human perceived lightness across hues. This is important to maintain color contrast regardless of hue. Also, tweaked colors, added shadow colors, and fixed an issue where types were not generated for the color JSON config. 

- Importing the HSLuv library, converting colors accordingly
- Tweaking the values in the color adjustments config
- Converting the JSON file to a TS file

### 🎩 checklist

- [x] 🎩


src/utilities/theme/utils.ts
```diff
@@ -70,25 +76,25 @@ export function buildColors(theme: ThemeConfig) {
...
-      ...successColors(colorToHsla(success), lightSurface),
+      ...surfaceColors(hexToHsluv(surface), lightSurface),
+      ...onSurfaceColors(hexToHsluv(onSurface), lightSurface),
+      ...interactiveColors(hexToHsluv(interactive), lightSurface),
+      ...neutralColors(hexToHsluv(neutral), lightSurface),
+      ...brandedColors(hexToHsluv(branded), lightSurface),
+      ...criticalColors(hexToHsluv(critical), lightSurface),
+      ...warningColors(hexToHsluv(warning), lightSurface),
+      ...highlightColors(hexToHsluv(highlight), lightSurface),
+      ...successColors(hexToHsluv(success), lightSurface),
```
danrosenthal: This functionally narrows our acceptable color space API from hex, rgba, hsla, down to just hex. Is that what we want?

tmlayton: We can rework this so it continues to support those 3 colors

tmlayton: We can rework this so it continues to support those 3 colors

tmlayton:
I added a [kitchen sink example](http://localhost:6006/?path=/story/playground-playground--kitchen-sink). It’s still a WIP to cherry pick which of the stories we want to show. As of now it is filtering out any stories which match the `exclude` regex.

tmlayton:
image

PR #2364, opened 2019-10-25
Version 5.0.0

BPScott:
[`UNRELEASED-v5.md`](https://github.com/Shopify/polaris-react/pull/2150/files#diff-3903aeb4e6449fc64d8f22f4ef7ddb03)

This branch is the same as https://github.com/Shopify/polaris-react/pull/2150, but renamed from `v5.0.0` to `version-5.0.0` to avoid conflicting with wit the final v5.0.0 tag that will be created on release.


BPScott:
Extra things I'd like to do:

- Consider removing global.scss and instead have those global styles pulled in as a side-effect of importing AppProvider (this could be done in a v4 release, as long as we leave an empty `global.scss` hanging around that old versions of SK can reference without throwing a file not found error)
- Reverse the ordering of the array of dictionaries that i18n has, so the main language is first (currently it is fallback first). This will make it match the ordering that react-i18n's useI18n() uses.
  This will let us remove the explanation and reversing in that consumers have to do
  https://github.com/Shopify/polaris-react/pull/2312/files#diff-7a936f09d89747421f1ed11ca7e51421R464-R470

LauraAubin:
Will look at adding more tests to address codecov.

BPScott:
Stuff for V5:

- https://github.com/Shopify/polaris-react/pull/2938
- https://github.com/Shopify/polaris-react/issues/3009

BPScott:
Now that `react-transition-group` supports strict mode we should start using that instead of `@material-ui/react-transition-group` again - we originally moved away because of the lack of strict mode.

.vscode/settings.json
```diff
@@ -9,38 +9,17 @@
...
-    "styles.{css,min.css,scss}": true,
-    "styles/": true,
-    "types/": true
+    "{build,dist}/": true
   },
   "javascript.validate.enable": false,
   "jest.autoEnable": false,
   "scss.validate": false,
   "search.exclude": {
-    "**/bower_components": true,
```
amrocha, line 31: haha bye bower

BPScott:
This all looks good to to me, but I can't approve my own PR. I'm happy to say lets merge into master and make that our v5 mainline going forward.

https://github.com/Shopify/polaris-react/pull/3113 and https://github.com/Shopify/polaris-react/pull/3115 will need to be rebased, but that's easy.

tmlayton:
Nice! Glad to see this finally going out 🎉

camelmasa:
🎉

PR #2366, opened 2019-10-25
[Color system] opacity (Card, Popover, Modal, Navigation)

danrosenthal:


src/utilities/theme/utils.ts
```diff
@@ -258,6 +259,29 @@ function successColors(color: HSLAColor, lightSurface: boolean) {
...
+      lightness: 8,
+      alpha: 0.05,
+    },
+    opacity: {
+      hue: 0,
+      saturation: 0,
+      lightness: 0,
+      alpha: 0.15,
+    },
+    opacityDeep: {
```
tmlayton:
```suggestion
    backdrop: {
```

src/utilities/theme/utils.ts
```diff
@@ -258,6 +259,29 @@ function successColors(color: HSLAColor, lightSurface: boolean) {
   };
 }
 
+function opacityColors() {
```
tmlayton: Pedantic as this is function name is private, but:

```suggestion
function translucentColors() {
```

src/utilities/theme/utils.ts
```diff
@@ -258,6 +259,29 @@ function successColors(color: HSLAColor, lightSurface: boolean) {
...
+  return {
+    opacityShallow: {
```
tmlayton:
```suggestion
    diffuseShadow: {
```

src/utilities/theme/utils.ts
```diff
@@ -258,6 +259,29 @@ function successColors(color: HSLAColor, lightSurface: boolean) {
...
+      hue: 180,
+      saturation: 5,
+      lightness: 8,
+      alpha: 0.05,
+    },
+    opacity: {
```
tmlayton:
```suggestion
    concentratedShadow: {
```

src/utilities/theme/utils.ts
tmlayton: Though I found myself explaining this as:

```suggestion
    shadowFromAmbientLight: {
```

src/utilities/theme/utils.ts
tmlayton: And like this: 

```suggestion
    shadowFromDirectLight: {
```

tmlayton:
Going to merge this into #2361 so we can test changes on that branch.

PR #2368, opened 2019-10-25
Add (unstable) telemetry support with a hook

amrocha:

We want to gather data around icon usage, and test the concept of gathering data about component usage, before deciding if we want to expand this to other components.

This is a temporary change, so we will remove it in a few weeks, once we have gathered enough data. However, we want feedback on this approach for gathering usage data on Polaris components.

This PR adds a prop to the `AppProvider` that is then accessed by `Icon` (for now, but possibly any component in the future if this approach works) via a hook. `Icon` then logs relevant information to monorail when it is rendered.

In order to gather data, this PR will be accompanied by a PR in web that provides their monorail instance to `AppProvider`, and that we can then use in `polaris-react` to log usage.
Adding

In order to clearly mark this as a temporary change, we marked all public-facing aspects of it "unstable" cc @tmlayton 

## 

### How to 🎩

See the example code

<details>

```jsx
import React from 'react';
import {
  PlusMinor,
  AlertMinor,
  ArrowDownMinor,
  ArrowLeftMinor,
  ArrowRightMinor,
} from '@shopify/polaris-icons';
import {
  Page,
  AppProvider,
  Card,
  ResourceList,
  Avatar,
  TextStyle,
  Icon,
} from '../src';

  return (
    <Page title="Playground">
      <AppProvider
        i18n={{
          Polaris: {
            ResourceList: {
              sortingLabel: 'Sort by',
              defaultItemSingular: 'item',
              defaultItemPlural: 'items',
              showing: 'Showing {itemsCount} {resource}',
              Item: {
                viewItem: 'View details for {itemName}',
              },
            },
            Common: {
              checkbox: 'checkbox',
            },
          },
        }}
        UNSTABLE_telemetry={{
          produce: (schemaId, payload) => {
            console.log(schemaId, payload);
            return true;
          },
        }}
      >
        <Page>
          <Card>
            <Icon source={PlusMinor} />
            <Icon source={AlertMinor} />
            <Icon source={ArrowDownMinor} />
            <Icon source={ArrowLeftMinor} />
            <Icon source={ArrowRightMinor} />
            <ResourceList
              showHeader
              items={[
                {
                  id: 341,
                  url: 'customers/341',
                  name: 'Mae Jemison',
                  location: 'Decatur, USA',
                },
                {
                  id: 256,
                  url: 'customers/256',
                  name: 'Ellen Ochoa',
                  location: 'Los Angeles, USA',
                },
              ]}
              renderItem={(item) => {

                return (
                    <h3>
                    </h3>
                    <div>{location}</div>
                  </ResourceList.Item>
                );
              }}
            />
          </Card>
        </Page>
      </AppProvider>
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/AppProvider/AppProvider.tsx
+++ b/src/components/AppProvider/AppProvider.tsx
@@ -42,7 +42,7 @@ export interface AppProviderProps extends AppBridgeOptions {
   features?: Features;
   /** Inner content of the application */
   children?: React.ReactNode;
-  monorail?: MonorailObject;
+  UNSTABLE_monorail?: MonorailObject;
 }
 
 export class AppProvider extends React.Component<AppProviderProps, State> {
@@ -100,7 +100,7 @@ export class AppProvider extends React.Component<AppProviderProps, State> {
   }
 
   render() {
-    const {theme = {}, features = {}, monorail, children} = this.props;
+    const {theme = {}, features = {}, UNSTABLE_monorail, children} = this.props;
     const {intl, appBridge, link} = this.state;
 
     return (
@@ -112,7 +112,7 @@ export class AppProvider extends React.Component<AppProviderProps, State> {
                 <AppBridgeContext.Provider value={appBridge}>
                   <LinkContext.Provider value={link}>
                     <ThemeProvider theme={theme}>
-                      <MonorailContext.Provider value={monorail}>
+                      <MonorailContext.Provider value={UNSTABLE_monorail}>
                         <MediaQueryProvider>{children}</MediaQueryProvider>
                       </MonorailContext.Provider>
                     </ThemeProvider>
--- a/src/components/AppProvider/AppProvider.tsx
+++ b/src/components/AppProvider/AppProvider.tsx
@@ -42,7 +42,7 @@ export interface AppProviderProps extends AppBridgeOptions {
   features?: Features;
   /** Inner content of the application */
   children?: React.ReactNode;
-  UNSTABLE_monorail?: MonorailObject;
+  UNSTABLE_monorail?: MonorailObject; // eslint-disable-line babel/camelcase
 }
 
 export class AppProvider extends React.Component<AppProviderProps, State> {
@@ -99,6 +99,7 @@ export class AppProvider extends React.Component<AppProviderProps, State> {
     });
   }
 
+  /* eslint-disable babel/camelcase */
   render() {
     const {theme = {}, features = {}, UNSTABLE_monorail, children} = this.props;
     const {intl, appBridge, link} = this.state;
@@ -125,4 +126,5 @@ export class AppProvider extends React.Component<AppProviderProps, State> {
       </FeaturesContext.Provider>
     );
   }
+  /* eslint-enable babel/camelcase */
 }
--- a/src/components/AppProvider/AppProvider.tsx
+++ b/src/components/AppProvider/AppProvider.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import {ThemeConfig} from '../../utilities/theme';
+import {MonorailContext, MonorailObject} from '../../utilities/monorail';
 import {ThemeProvider} from '../ThemeProvider';
 import {MediaQueryProvider} from '../MediaQueryProvider';
 import {I18n, I18nContext, TranslationDictionary} from '../../utilities/i18n';
@@ -41,6 +42,7 @@ export interface AppProviderProps extends AppBridgeOptions {
   features?: Features;
   /** Inner content of the application */
   children?: React.ReactNode;
+  monorail?: MonorailObject;
 }
 
 export class AppProvider extends React.Component<AppProviderProps, State> {
@@ -98,7 +100,7 @@ export class AppProvider extends React.Component<AppProviderProps, State> {
   }
 
   render() {
-    const {theme = {}, features = {}, children} = this.props;
+    const {theme = {}, features = {}, monorail, children} = this.props;
     const {intl, appBridge, link} = this.state;
 
     return (
@@ -110,7 +112,9 @@ export class AppProvider extends React.Component<AppProviderProps, State> {
                 <AppBridgeContext.Provider value={appBridge}>
                   <LinkContext.Provider value={link}>
                     <ThemeProvider theme={theme}>
-                      <MediaQueryProvider>{children}</MediaQueryProvider>
+                      <MonorailContext.Provider value={monorail}>
+                        <MediaQueryProvider>{children}</MediaQueryProvider>
+                      </MonorailContext.Provider>
                     </ThemeProvider>
                   </LinkContext.Provider>
                 </AppBridgeContext.Provider>
--- a/src/components/AppProvider/AppProvider.tsx
+++ b/src/components/AppProvider/AppProvider.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import {ThemeConfig} from '../../utilities/theme';
-import {MonorailContext, MonorailObject} from '../../utilities/monorail';
+import {TelemetryContext, TelemetryObject} from '../../utilities/telemetry';
 import {ThemeProvider} from '../ThemeProvider';
 import {MediaQueryProvider} from '../MediaQueryProvider';
 import {I18n, I18nContext, TranslationDictionary} from '../../utilities/i18n';
@@ -42,7 +42,7 @@ export interface AppProviderProps extends AppBridgeOptions {
   features?: Features;
   /** Inner content of the application */
   children?: React.ReactNode;
-  UNSTABLE_monorail?: MonorailObject; // eslint-disable-line babel/camelcase
+  UNSTABLE_telemetry?: TelemetryObject; // eslint-disable-line babel/camelcase
 }
 
 export class AppProvider extends React.Component<AppProviderProps, State> {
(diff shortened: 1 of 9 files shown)


src/components/Icon/Icon.tsx
```diff
@@ -21,6 +22,29 @@ interface Props extends IconProps {}
...
+  /* eslint-disable babel/camelcase */
+  useEffect(() => {
+    if (unstableMonorail) {
+      let loggedSource;
+      if (typeof source === 'function') {
+        loggedSource = source.name;
+      } else if (source === 'placeholder') {
+        loggedSource = source;
+      } else {
+        loggedSource = 'custom icon string';
```
amrocha: @kaelig this is how I decided to log custom string icons, what do you think? I can't log a filename like I do for the components because this is just a string

src/components/AppProvider/AppProvider.tsx
```diff
@@ -41,6 +42,7 @@ export interface AppProviderProps extends AppBridgeOptions {
   features?: Features;
   /** Inner content of the application */
   children?: React.ReactNode;
+  UNSTABLE_monorail?: MonorailObject; // eslint-disable-line babel/camelcase
```
BPScott: I'm not a fan of tieing this to monorail within the polaris-react codebase by this naming. Talking about internal shopify tooling feels like an odd leakage. Instead what if instead of passing in an instance of monorail we pass in a function that then calls monorail - or any other telemetry aggregator we might want in the future.

By pushing the responsibility of what telemetry producer is used to apps we let apps decide on how they want to handle the results, and any future changes should only need to be made in apps.

What if we set this up so consumers used this something like so:

```ts
// Or however web gets a monorail instance
const monorail = useMonorail();

function telemetryProducer(action: string, payload: Record<string, any> = {}) {
  // this example is lazy as the api of this function happens to map to monorail's api
 // but we could do clever things like batch or manipulate results here
  monorail.produce(action, payload);
}

const app = <AppProvider UNSTABLE_telemetry={telemetryProducer}>{/*...*/}</AppProvider>
```

src/utilities/monorail/hooks.tsx
```diff
@@ -0,0 +1,6 @@
+import React from 'react';
+import {MonorailContext} from './context';
+
+export function useMonorail() {
+  return React.useContext(MonorailContext);
```
BPScott: Check if the result returns a value, and if not return a noop function. That will ensure this function always returns a function. that way we can avoid having to check if the context has a value in https://github.com/Shopify/polaris-react/pull/2368/files#diff-ac192d85916a4a5bf777ae9cb10c1a08R29

src/components/Icon/Icon.tsx
```diff
@@ -21,6 +22,29 @@ interface Props extends IconProps {}
...
+      }
+
+      unstableMonorail.produce('polaris_icons_usage/1.0', {
+        icon_source: loggedSource,
+        color,
```
BPScott: Do we care if the color, backdrop or a11y label change?

src/components/Icon/Icon.tsx
```diff
@@ -21,6 +22,29 @@ interface Props extends IconProps {}
 
 export function Icon({source, color, backdrop, accessibilityLabel}: Props) {
   const i18n = useI18n();
+  const unstableMonorail = useMonorail();
+
+  /* eslint-disable babel/camelcase */
+  useEffect(() => {
+    if (unstableMonorail) {
+      let loggedSource;
+      if (typeof source === 'function') {
```
BPScott: nitpick: I'd be tempted to the "determine the loggedSource based on the source" logic out into a separate function

src/components/Icon/Icon.tsx
amrocha: Any of them changing triggers a re-render which I think we would want to catch. Maybe not though, what do you think @kaelig ?

src/components/AppProvider/AppProvider.tsx
amrocha: Using monorail specifically was an intentional decision to not commit us to supporting a generic telemetry that would work across the board.

This is a holdover from back when we were taking a different approach, making it generic here is simple enough.

src/components/Icon/Icon.tsx
kaelig: Thanks, that's probably the right compromise to not impact performance.

Is there any additional info we could log that would give us a hint on what/where this is rendered?

kaelig:
Should `<AppProvider monorail={}>` be `<AppProvider UNSTABLE_monorail={}>`?

src/components/Icon/Icon.tsx
BPScott: Triggering two trackings of "IconX got rendered" when what happened is its color changed from red to green seems odd to me. Taken to a silly conclusion, an icon that flashes rainbow colors would be considered the most used icon even though it was never hidden / reshown.

BPScott:
@kaelig Check the props of AppProvider: `UNSTABLE_monorail={}`

src/components/Icon/Icon.tsx
kaelig: - color: useful for creative direction to simulate the most recurring use cases in the masterbrand work (including dark mode)
- a11y label: this one's useful to satisfy my curiosity and see if/where we should do some a11y labelling work, or see if certain icons could use a brand new variant because a label is so misaligned with the icon

tmlayton:
`UNSTABLE` prefix looks good, glad to see this pattern helping to get ideas out there quicker.

tmlayton:
Same feels from me for using a more generic term like `telemetry`

src/components/Icon/Icon.tsx
```diff
@@ -21,6 +22,29 @@ interface Props extends IconProps {}
 
 export function Icon({source, color, backdrop, accessibilityLabel}: Props) {
   const i18n = useI18n();
+  const unstableMonorail = useMonorail();
+
+  /* eslint-disable babel/camelcase */
+  useEffect(() => {
+    if (unstableMonorail) {
```
tmlayton: Internally/privately I think it is okay to just call this `monorail`, as the `monorail` object is stable just not the provider API.

src/components/Icon/Icon.tsx
amrocha: I thought the same thing, it'd be nice to know where this icon is being rendered, I don't think so because the code gets minified. I have to try it out running in actual web though, an example in the playground wouldn't be enough here

src/components/Icon/Icon.tsx
amrocha: Right, getting the values could be useful, but what Ben is saying is that if those values change then we would trigger another event for that exact same icon. Does it make sense to do this? Now that I'm thinking about it I don't think so, I'd prefer to only log when icons are initialized or when the source changes

kaelig: Where does the rerendering-with-a-different-color take place?

If it happens rarely, I'm fine with whatever solution. If it happens a lot, then it might be meaningful?

amrocha: I really don't know where it happens, but off the top of my head probably whenever an icon is interacted with (Button, Banner etc) the color or backdrop might change. I think this is going to muddy the data that we're working with so I made it log again only when the source is updated.

src/components/Icon/Icon.tsx
amrocha: I couldn't figure out how to do this in web, I wasn't able to run web with production assets to be able to test :/

src/components/AppProvider/AppProvider.tsx
```diff
@@ -41,6 +42,7 @@ export interface AppProviderProps extends AppBridgeOptions {
   features?: Features;
   /** Inner content of the application */
   children?: React.ReactNode;
+  UNSTABLE_telemetry?: TelemetryObject; // eslint-disable-line babel/camelcase
```
BPScott: We tend to avoid the trailing comment syntax and instead use `eslint-disable-next-line`

src/components/Icon/Icon.tsx
```diff
@@ -19,8 +20,30 @@ const COLORS_WITH_BACKDROPS = [
 // styleguide to generate the props explorer
 interface Props extends IconProps {}
 
+function parseSource(source: string | Function) {
```
BPScott: Put utility functions underneath the component function -show us the important bit first

src/components/Icon/Icon.tsx
```diff
@@ -19,8 +20,30 @@ const COLORS_WITH_BACKDROPS = [
...
+
+  /* eslint-disable babel/camelcase */
+  useEffect(() => {
+    telemetry.produce('polaris_icons_usage/1.0', {
+      icon_source: parseSource(source),
+      color,
+      backdrop,
+      accessibility_label: accessibilityLabel,
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
BPScott: I think we should keep the rules of hooks happy. We should either:

- Track color/backdrop/accessibilityLabel which includes when they change
- Don't track them at all

Currently we're in an odd halfway house where we track initial values but if the color ever changes then we don't track it which feels like we're building an incomplete dataset.

My gut is saying that tracking these will make it harder to sort through the data as it makes it harder to reason about "what icons were loaded" as we have to discount any cases where just the color/backdrop/label were changed

src/utilities/telemetry/hooks.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import {TelemetryContext} from './context';
+
+function noop() {}
+
+export function useTelemetry() {
+  return React.useContext(TelemetryContext) || {produce: noop};
```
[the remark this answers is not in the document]
amrocha: Yep, that's happening but it's also happening when we do have the context. I'm trying to figure out why 🤔

PR #2369, opened 2019-10-26
Sewing-kit bump

BPScott:

This updates our eslint version to 6 and our eslint-plugin-shopify
version to v31, which includes support for the new @typescript-eslint
parser so we can get rid of that resolution hack.

Updates sewing-kit and updates code to adhere to updated rules.

This also changes our vscode config so that the prettier extension no longer formats our `{js,ts}x?` files on save and instead this is handled by the eslint extension, this is because the prettier extension doesn't play nice with eslint 6 (as it uses prettier-eslint under the hood which doesn't like eslint6: https://github.com/prettier/prettier-eslint/pull/236) and its future roadmap will remove support for running eslint formatting (https://github.com/prettier/prettier-vscode/issues/958).

### How to 🎩

linting and tests pass

After doing a `yarn install` restart VSCode and ensure that format on save still works for various file types:
- JS files
- JSX files (don't think we have any of these but whatevs)
- TS files
- TSX files (these 4 have changes to their behaviour as they're now using the eslint extension to format rather than the prettier extension
- scss files (this has no changes but nice to regression test)
- md files (this has no changes but nice to regression test)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -69,6 +69,18 @@
   },
   "overrides": [
     {
+      "files": ["src/**/*.{ts,tsx}"],
+      "extends": ["plugin:shopify/typescript-type-checking"],
+      "parserOptions": {
+        "project": "./tsconfig.json"
+      },
+      "rules": {
+        "@typescript-eslint/prefer-readonly": "off",
+        "@typescript-eslint/no-unnecessary-condition": "off",
+        "@typescript-eslint/unbound-method": "off"
+      }
+    },
+    {
       "files": ["**/*.test.{ts,tsx}"],
       "rules": {
         "jest/no-truthy-falsy": "off",
--- a/.eslintrc
+++ b/.eslintrc
@@ -62,7 +62,8 @@
         "allow": [
           "UNSTABLE_Color",
           "UNSTABLE_colors",
-          "UNSTABLE_cssCustomProperties"
+          "UNSTABLE_cssCustomProperties",
+          "UNSTABLE_telemetry"
         ]
       }
     ]
--- a/.eslintrc
+++ b/.eslintrc
@@ -55,14 +55,23 @@
     "jsx-a11y/mouse-events-have-key-events": "off",
     "jsx-a11y/click-events-have-key-events": "off",
     "jsx-a11y/no-noninteractive-element-interactions": "off",
-    "jsx-a11y/no-noninteractive-element-to-interactive-role": "off"
+    "jsx-a11y/no-noninteractive-element-to-interactive-role": "off",
+    "@typescript-eslint/camelcase": [
+      "error",
+      {
+        "allow": [
+          "UNSTABLE_Color",
+          "UNSTABLE_colors",
+          "UNSTABLE_cssCustomProperties"
+        ]
+      }
+    ]
   },
   "overrides": [
     {
       "files": ["**/*.test.{ts,tsx}"],
       "rules": {
         "jest/no-truthy-falsy": "off",
-        "shopify/jsx-no-hardcoded-content": "off",
         "shopify/no-ancestor-directory-import": "off"
       }
     },
--- a/scripts/splash/index.tsx
+++ b/scripts/splash/index.tsx
@@ -202,7 +202,7 @@ const App = () => {
           <Color dim>
             <Box width={3}>💡</Box>
             <Box>
-              Tip: to disable these reports, run{' '}
+              Tip: to disable these reports, run
               <Text bold>DISABLE_SPLASH=1 yarn dev</Text>
             </Box>
           </Color>
--- a/scripts/splash/index.tsx
+++ b/scripts/splash/index.tsx
@@ -203,7 +203,7 @@ const App = () => {
             <Box width={3}>💡</Box>
             <Box>
               Tip: to disable these reports, run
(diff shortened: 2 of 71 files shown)


.eslintrc
```diff
@@ -55,14 +55,35 @@
     "jsx-a11y/mouse-events-have-key-events": "off",
     "jsx-a11y/click-events-have-key-events": "off",
     "jsx-a11y/no-noninteractive-element-interactions": "off",
-    "jsx-a11y/no-noninteractive-element-to-interactive-role": "off"
+    "jsx-a11y/no-noninteractive-element-to-interactive-role": "off",
+    "@typescript-eslint/camelcase": [
+      "error",
+      {
+        "allow": [
+          "UNSTABLE_Color",
```
BPScott: This is a bit nicer than sprinkling single-line ignores throughout the codebase

.eslintrc
```diff
@@ -55,14 +55,35 @@
...
+        "@typescript-eslint/prefer-readonly": "off",
+        "@typescript-eslint/no-unnecessary-condition": "off",
+        "@typescript-eslint/unbound-method": "off"
+      }
+    },
     {
       "files": ["**/*.test.{ts,tsx}"],
       "rules": {
         "jest/no-truthy-falsy": "off",
-        "shopify/jsx-no-hardcoded-content": "off",
```
BPScott: This is now turned off in tests by default

.eslintrc
```diff
@@ -55,14 +55,35 @@
...
+          "UNSTABLE_Color",
+          "UNSTABLE_colors",
+          "UNSTABLE_cssCustomProperties"
+        ]
+      }
+    ]
   },
   "overrides": [
+    {
+      "files": ["src/**/*.{ts,tsx}"],
```
BPScott: Because some of our ts files aren't included in our tsconfig's files list we can't set this stuff at the global level, which is a little annoying but not the end of the world.

src/components/AppProvider/tests/AppProvider.test.tsx
```diff
@@ -16,7 +16,7 @@ describe('<AppProvider />', () => {
   });
 
   it('updates context when props change', () => {
-    const Child: React.SFC<{}> = () => {
```
BPScott: A lot of the changes in this PR are "that's the default value for the generic, you don't need to specify that"

src/components/Loading/Loading.tsx
```diff
@@ -5,9 +5,6 @@ import {useAppBridge} from '../../utilities/app-bridge';
 
 export interface LoadingProps {}
 
-// This does have a display name, but the linting has a bug in it
-// https://github.com/yannickcr/eslint-plugin-react/issues/2324
-// eslint-disable-next-line react/display-name
```
BPScott: We've updated the version of eslint-plugin-react to a version that fixes this false positive

src/components/UnstyledLink/UnstyledLink.tsx
```diff
@@ -29,3 +25,8 @@ export const UnstyledLink = React.memo(
     );
   }),
 );
+
+// Wrapping forwardRef in a memo means the component does not get a name set
+// So we have to do it ourselves
+// https://github.com/facebook/react/issues/16722
```
BPScott: Bug whack-a-mole. I'm not convinced we should be wrapping this in memo in the first place: https://github.com/Shopify/polaris-react/pull/2330

src/utilities/get.ts
```diff
@@ -11,6 +11,7 @@ export function get<T>(
 
   const keys = Array.isArray(keypath) ? keypath : getKeypath(keypath);
   let acc = obj;
+  // eslint-disable-next-line @typescript-eslint/prefer-for-of
```
BPScott: fixing this causes some other error so gonna ignore it

scripts/splash/index.tsx
```diff
@@ -202,7 +202,7 @@ const App = () => {
           <Color dim>
             <Box width={3}>💡</Box>
             <Box>
-              Tip: to disable these reports, run{' '}
```
kaelig: This causes a regression where a space is missing between `run` and `DISABLE_SPLASH`:

Screen Shot 2019-11-08 at 3 20 51 PM

scripts/splash/index.tsx
```diff
@@ -81,7 +81,7 @@ const Components = ({components, status}) => (
   <React.Fragment>
     {status === 'loading' && (
       <Box marginLeft={4} marginBottom={1}>
-        ⏳{'  '}Please wait during compilation… Beep boop beep 🤖
+        ⏳ Please wait during compilation… Beep boop beep 🤖
```
kaelig: No regression here – all good.

src/components/Frame/components/ToastManager/ToastManager.tsx
```diff
@@ -17,9 +17,6 @@ export interface ToastManagerProps {
   toastMessages: (ToastPropsWithID)[];
 }
 
-// This does have a display name, but the linting has a bug in it
-// https://github.com/yannickcr/eslint-plugin-react/issues/2324
-// eslint-disable-next-line react/display-name
```
kaelig: 🎉 🔥

PR #2372, opened 2019-10-28
Make yarn splash optional

amrocha:

Yarn splash can be really noisy at times

Enables maintainers running `yarn dev` to hide [`yarn splash`](https://github.com/Shopify/polaris-react/tree/master/scripts/splash) reports from the console by running `DISABLE_SPLASH=1 yarn dev` ([#2372](https://github.com/Shopify/polaris-react/pull/2372))

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -116,7 +116,7 @@ module.exports = ({config, mode}) => {
     apply: (compiler) => {
       compiler.hooks.afterEmit.tap('AfterEmitPlugin', (compilation) => {
         process.env.DISABLE_SPLASH ||
-          spawn('yarn splash --storybook', {
+          spawn('yarn splash --show-storybook-tip', {
             shell: true,
             stdio: 'inherit',
           });
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -115,10 +115,11 @@ module.exports = ({config, mode}) => {
   config.plugins.push({
     apply: (compiler) => {
       compiler.hooks.afterEmit.tap('AfterEmitPlugin', (compilation) => {
-        const spawnedProcess = spawn('yarn splash', {
-          shell: true,
-          stdio: 'inherit',
-        });
+        process.env.DISABLE_SPLASH ||
+          spawn('yarn splash', {
+            shell: true,
+            stdio: 'inherit',
+          });
       });
     },
   });
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -116,7 +116,7 @@ module.exports = ({config, mode}) => {
     apply: (compiler) => {
       compiler.hooks.afterEmit.tap('AfterEmitPlugin', (compilation) => {
         process.env.DISABLE_SPLASH ||
-          spawn('yarn splash', {
+          spawn('yarn splash --storybook', {
             shell: true,
             stdio: 'inherit',
           });
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -115,11 +115,10 @@ module.exports = ({config, mode}) => {
   config.plugins.push({
     apply: (compiler) => {
       compiler.hooks.afterEmit.tap('AfterEmitPlugin', (compilation) => {
-        process.env.DISABLE_SPLASH ||
-          spawn('yarn splash --show-storybook-tip', {
-            shell: true,
-            stdio: 'inherit',
-          });
+        spawn('yarn splash --show-disable-tip', {
+          shell: true,
+          stdio: 'inherit',
+        });
       });
     },
   });
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -16,7 +16,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Development workflow
 
-- Made it possible to prevent [“yarn splash”](https://github.com/Shopify/polaris-react/tree/master/scripts/splash) from automatically running alongside `yarn dev`: `yarn dev-no-splash` or `DISABLE_SPLASH=1 yarn dev` ([#2372](https://github.com/Shopify/polaris-react/pull/2372))
+- Enable developers to hide [`yarn splash`](https://github.com/Shopify/polaris-react/tree/master/scripts/splash) reports when running `yarn dev` by running `DISABLE_SPLASH=1 yarn dev` ([#2372](https://github.com/Shopify/polaris-react/pull/2372))
 
 ### Dependency upgrades
 
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -16,7 +16,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Development workflow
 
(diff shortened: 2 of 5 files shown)


scripts/splash/index.tsx
```diff
@@ -154,6 +154,14 @@ const App = () => {
   return (
     <React.Fragment>
       <Box marginBottom={1} flexDirection="column">
+        <Box>
+          <Box width={3}>💡</Box>
+          <Box>
+            Tip: disable <Text bold>yarn splash</Text> by running{' '}
+            <Text bold>yarn dev-no-splash</Text> or setting an environment
+            variable <Text bold>DISABLE_SPLASH=1</Text>
```
kaelig: Isn't it a bit strange to have this message here when the responsibility for this condition is handled by the caller? This means that running `yarn splash` will show this message.

package.json
```diff
@@ -53,6 +53,7 @@
     "prepublish": "in-publish && yarn run build || :",
     "postpublish": "in-publish && npm-run-all new-version-pr-generator",
     "dev": "npm-run-all copy-polaris-tokens storybook",
+    "dev-no-splash": "DISABLE_SPLASH=1 yarn run dev",
```
kaelig: Not sure this warrants adding a new script – it might be okay to tell people to run `DISABLE_SPLASH=1 yarn run dev`?

.storybook/webpack.config.js
```diff
@@ -115,10 +115,11 @@ module.exports = ({config, mode}) => {
   config.plugins.push({
     apply: (compiler) => {
       compiler.hooks.afterEmit.tap('AfterEmitPlugin', (compilation) => {
-        const spawnedProcess = spawn('yarn splash', {
-          shell: true,
-          stdio: 'inherit',
-        });
+        process.env.DISABLE_SPLASH ||
```
kaelig: The environment variable is a great idea 👍

UNRELEASED.md
```diff
@@ -20,6 +20,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Development workflow
 
 - Removed the need to upload assets with each release ([#2346](https://github.com/Shopify/polaris-react/pull/2346))
+- Made running yarn splash along with the storybook optional ([#2372](https://github.com/Shopify/polaris-react/pull/2372))
```
kaelig: We could save people a click by telling them a bit more about this.

Here's a suggestion:

```suggestion
- Made it possible to prevent [“splash”](https://github.com/Shopify/polaris-react/tree/master/scripts/splash) from automatically running alongside `yarn dev`: `DISABLE_SPLASH=1 yarn dev` ([#2372](https://github.com/Shopify/polaris-react/pull/2372))
```

scripts/splash/index.tsx
amrocha: People need to learn about it somehow or we're always going to get messages from people like "hey how do I turn off yarn splash". Even if we let people on Slack know, there's outside contributors and new people joining all the time.

I see what you mean about this showing up when people run the standalone command. I can make it show up when storybook is running, but not when it's run by itself. How about that?

package.json
amrocha: This is the same reason as above, I wanted to document it so that people can find the option easily.

If we don't add a command, how many people are going to dig into the code to figure out they can set an environment variable? Probably not a lot

amrocha:
The reason I made it this way is discoverability.

I think we should keep the message in there for sure. I'd prefer to keep the command in there too, since the storybook output is very noisy and I think people might not find it, but this is redundant if we have the message in there so I'm ok with removing it if you want to.

scripts/splash/index.tsx
kaelig: I totally agree that showing a message is good for awareness 👍 

My question was around whether it's appropriate to have the message living in `scripts/splash/`, when the code that handles what's advertized by the message lives elsewhere (in the webpack config).

Alternatively, would it make more sense to handle the conditional for `DISABLE_SPLASH` in `scripts/splash/index.ts`, so both the message and the condition live under the same roof?

package.json
kaelig: The console message and Splash's README do a great job telling people about the `DISABLE_SPLASH` trick.

I see your point about adding even more discoverability, however having 2 ways of achieving the same thing means we have to document and provide support/answer questions about both.

scripts/splash/index.tsx
amrocha: I tried doing it in the webpack config but webpack overrides stdout so I can't just echo.

I could create a separate script for it but I think that would be adding too much complexity for very little gain

package.json
amrocha: What's the support we need to provide for a yarn command? Seems fairly straightforward to me

kaelig: >What's the support we need to provide for a yarn command?

When yarn splash fails or needs to be disabled for any reason, we'd be giving people 3 ways of disabling it:

1. `DISABLE_SPLASH=1 yarn dev`
1. `DISABLE_SPLASH=1` environment variable + `yarn dev`
1. `yarn dev-no-splash` (probably doesn't work in Windows, btw)

It's not a lot of overhead, but it adds to the overall cognitive load for contributors.

Also, if people start using `yarn dev-no-splash` and we want to remove it later, we'll be disrupting their developer experience (and I really don't want us to have to think about a deprecation path for custom yarn scripts).

Plus, that's less documentation for us to write and keep up to date 🎉 

PS: I agree with the intent behind having it more discoverable, but I'm not sure it's appropriate to do it that way, especially since the console message & README doc are great ways to advertise this feature.

kaelig:
Hey @dleroux, I just merged a3354b8 into this branch to contain the logic into the splash script itself. Can you give this an extra review?

scripts/splash/README.md
```diff
@@ -11,6 +11,8 @@ It answers the question:
 1. Edit files in `src/`, such as components 🧩 and style sheets 🎨.
 2. As you run `yarn dev`, `yarn splash` will run in the background. Keep an eye on the terminal to see the splash zone of your changes in the working directory.
 
+   💡 Tip: to disable these reports, run `DISABLE_SPLASH=1 yarn dev`
```
dleroux: Did you mean to indent this tip?

UNRELEASED.md
```diff
@@ -16,6 +16,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Development workflow
 
+- Made it possible to prevent [“yarn splash”](https://github.com/Shopify/polaris-react/tree/master/scripts/splash) from automatically running alongside `yarn dev`: `yarn dev-no-splash` or `DISABLE_SPLASH=1 yarn dev` ([#2372](https://github.com/Shopify/polaris-react/pull/2372))
```
dleroux: `yarn dev-no-splash` is not an option is it?

dleroux: Apart from that everything looks good. 🎩 as well.

scripts/splash/README.md
kaelig: Yep 👍 (it's nested under the list item)

UNRELEASED.md
kaelig: Thank you, good catch!

PR #2373, opened 2019-10-28
Fix bug that introduced a race condition in consumers

### LINKED ISSUE #2333: Datepicker jump in Firefox
## Issue summary
Since web was upgraded to Polaris 4.3, the overview dashboard, finances summary and finances reports datepickers have had an issue with focusing on select of the days in Firefox. The focus ends up on the bottom of the page, making it impossible to select a day.

See gif and more details in the first part of this issue https://github.com/Shopify/store/issues/10573

Note: I added a temporary fix [here](https://github.com/Shopify/web/pull/19864/files#diff-ed56dee96d87ac7a90075d49476dbff8R36) in web. Happy to add this to Polaris but not sure if it is the right solution and dealing with the actual root problem.

## Expected behavior
You should be able to select a day without the datepicker jumping and closing

## Actual behavior
Oct-21-2019 09-55-44

Revert this web PR https://github.com/Shopify/web/pull/19864 locally (my temporary fix, which cloned the Polaris component and added `{preventScroll: true}` to the `focus` method). Visit `admin/dashboards` using Firefox and see the issue.

## Reduced test case
I haven't been able to create a reduced test case

## Specifications

- Are you using the React components? (Y/N): yes
- Polaris version number: 3.4 and above
- Browser: Firefox
- Device: web
- Operating System: Mac 10.14.6

Paste the results here:

```bash

```

amrocha:

Fixes #2333 

See https://github.com/Shopify/polaris-react/pull/2089#discussion_r322790348

When I converted DatePicker to a functional component using hooks, I decided to simplify the code a bit as well. Since our tests and examples for the DatePicker are not thorough, this introduced a race condition that was hard to test locally. The only browser that triggered this condition consistently was Firefox, and only when also using the `Month` component.

This fixes that by waiting until state changes have resolved to trigger the callback

## 

### How to 🎩

`yarn run build-consumer web` and check out the analytics page on FF, as per the original issue

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -12,6 +12,9 @@
 - Prevented scrolling to `Popover` content in development ([#2403](https://github.com/Shopify/polaris-react/pull/2403))
 - Fixed an issue which caused HSL colors to not display in Edge ([#2418](https://github.com/Shopify/polaris-react/pull/2418))
 - Fixed an issue where the dropzone component jumped from an extra-large layout to a layout based on the width of its container ([#2412](https://github.com/Shopify/polaris-react/pull/2412))
+- Fixed an issue which caused HSL colors to not display in Edge ((#2418)[https://github.com/Shopify/polaris-react/pull/2418])
+- Fixed an issue where the dropzone component jumped from an extra-large layout to a layout based on the width of it's container ([#2412](https://github.com/Shopify/polaris-react/pull/2412))
+- Fixed a race condition in DatePicker ([#2373](https://github.com/Shopify/polaris-react/pull/2373))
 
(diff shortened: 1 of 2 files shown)


amrocha:
Hey @AndrewMusgrave , remember when we were talking about hookifying the DatePicker and we decided that we could simplify the code a bit and execute a callback sequentially instead of waiting for state changes to resolve? Remember when I said
> I expect this to be basically equivalent

And then you said

>I think we'll be ok

??

Well guess what 🙃🙃🙃🙃🙃🙃🙃🙃🙃🙃🙃🙃🙃

PR #2376, opened 2019-10-29
[Modal] Convert modal to function

LauraAubin:

Opening this new PR since the old one https://github.com/Shopify/polaris-react/pull/2341 was closed when the `v5.0.0` branch was deleted.

Some notable changes,

- The activator now uses `useRef` instead of `createRef`.
- The external activator example needs to call `requestAnimationFrame` when focusing the button.

### How to 🎩

### 🎩 checklist

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -213,7 +213,3 @@ export const Modal: React.FunctionComponent<ModalProps> & {
 };
 
 Modal.Section = Section;
-
-// Use named export once withAppProvider is refactored away
-// eslint-disable-next-line import/no-default-export
-// export default Modal;
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useState} from 'react';
 import {TransitionGroup} from '@material-ui/react-transition-group';
 import {write} from '@shopify/javascript-utilities/fastdom';
 import {focusFirstFocusableNode} from '@shopify/javascript-utilities/focus';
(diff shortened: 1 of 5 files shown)


src/components/Modal/Modal.tsx
```diff
@@ -62,181 +59,157 @@ export interface ModalProps extends FooterProps {
...
-
-      const body = loading ? (
-        <div className={styles.Spinner}>
-          <Spinner />
-        </div>
-      ) : (
-        content
-      );
+export const Modal: React.FunctionComponent<ModalProps> & {
+  Section: typeof Section;
```
LauraAubin: The `Card` explains why this typing is needed:

> TypeScript can't generate types that correctly infer the typing of subcomponents so explicitly state the subcomponents in the type definition. Letting this be implicit works in this project but fails in projects that use generated *.d.ts files.

src/components/Modal/Modal.tsx
```diff
@@ -62,181 +59,157 @@ export interface ModalProps extends FooterProps {
...
+  onScrolledToBottom,
+  activator,
+  onClose,
+  onIFrameLoad,
+  onTransitionEnd,
+}: ModalProps) {
+  const [iframeHeight, setIframeHeight] = useState(IFRAME_LOADING_HEIGHT);
+
+  const headerId = useUniqueId('modal-header');
+  const activatorRef = React.useRef<HTMLDivElement>(null);
```
LauraAubin: Should these still be private?

src/components/Modal/tests/Modal.test.tsx
```diff
@@ -337,9 +337,17 @@ describe('<Modal>', () => {
       expect(modal.find(Button).exists()).toBe(true);
     });
 
+    it('does not throw an error when no activator is passed in', () => {
```
LauraAubin: @danrosenthal this should fix codecov on the v5 branch.

src/components/Modal/README.md
```diff
@@ -658,7 +658,7 @@ function ModalExample() {
 
   const handleClose = useCallback(() => {
     setActive(false);
-    button.current.querySelector('button').focus();
+    requestAnimationFrame(() => button.current.querySelector('button').focus());
```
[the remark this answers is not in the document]
AndrewMusgrave: I wonder if it's worth making this work without RAF? It's not very intuitive using it to focus an element and I'm if we're supporting this we'll get questions about it in the future.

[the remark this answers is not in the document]
LauraAubin: I agree that it's not ideal, but since we're going to make the `activator` prop mandatory in the future (give or take a few weeks after the release of v5), this example will be entirely removed. For that short while, think it's okay to leave it as is?

PR #2381, opened 2019-10-30
Update description of imageContained prop (EmptyState component)

martenbjork:
The description is incorrect and seems to be copy-pasted from another prop.

It updates the description.

### DIFF
--- a/src/components/EmptyState/EmptyState.tsx
+++ b/src/components/EmptyState/EmptyState.tsx
@@ -17,7 +17,7 @@ export interface EmptyStateProps {
   image: string;
   /** The image to use for large screens */
   largeImage?: string;
-  /** The image to use for large screens */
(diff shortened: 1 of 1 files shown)


src/components/EmptyState/EmptyState.tsx
```diff
@@ -17,7 +17,7 @@ export interface EmptyStateProps {
   image: string;
   /** The image to use for large screens */
   largeImage?: string;
-  /** The image to use for large screens */
+  /** The image width is contained within its wrapper */
```
kaelig: Would using the active voice like so work as well?

```suggestion
  /** Contain the image within its wrapper */
```

danrosenthal:
Looks like this was fixed upstream. Closing in favor of master.

PR #2384, opened 2019-10-31
Remove references to Sketch

kaelig:
Removes references to Sketch, as we are moving away from it internally.


kaelig:
@HYPD I'm merging this small change before I go on vacation! Feel free to revert if there is better wording for this.

PR #2389, opened 2019-10-31
[Filter] Add 'disabled' prop

PabloVallejo:

Related to https://github.com/Shopify/store/issues/10681

As part of the solution for https://github.com/Shopify/store/issues/10681 we need to disable the `Filters` in the Products Index Page. Currently the filter component doesn't have a prop to disable it.

This PR adds a `disabled` prop to the `Filter` component.

When the `Filter` is disabled:

* Search-bar is disabled
* Actions from `ConnectedFilterControl` are disabled.
* Popover is closed 
* Sheet filter togglers are disabled
* Sheet filter titles are subdued

| Disabled filters        | Disabled sheet     | 
| image | image | 

## 

### How to 🎩

* Paste the playground code and run the project 
* From the playground, see that all the filters are active
* ✅Click the account status button and select any option: See that the filters get disabled
* ✅Click `More filters` and from the Account status filter select any option: See that the main filters get disabled and the filters in the Sidebar get their title subdued!

<details>

```jsx
import {
  Page,
  TextField,
  Avatar,
  Filters,
  Card,
  TextStyle,
  ResourceList,
  ChoiceList,
} from '../src';

function ResourceListFiltersExample() {
  const [accountStatus, setAccountStatus] = useState(null);
  const [moneySpent, setMoneySpent] = useState(null);
  const [taggedWith, setTaggedWith] = useState(null);
  const [queryValue, setQueryValue] = useState(null);

  const handleAccountStatusChange = useCallback(
    (value) => setAccountStatus(value),
    [],
  );
  const handleMoneySpentChange = useCallback(
    (value) => setMoneySpent(value),
    [],
  );
  const handleTaggedWithChange = useCallback(
    (value) => setTaggedWith(value),
    [],
  );
  const handleFiltersQueryChange = useCallback(
    (value) => setQueryValue(value),
    [],
  );
  const handleAccountStatusRemove = useCallback(
    () => setAccountStatus(null),
    [],
  );
  const handleMoneySpentRemove = useCallback(() => setMoneySpent(null), []);
  const handleTaggedWithRemove = useCallback(() => setTaggedWith(null), []);
  const handleQueryValueRemove = useCallback(() => setQueryValue(null), []);
  const handleFiltersClearAll = useCallback(() => {
    handleAccountStatusRemove();
    handleMoneySpentRemove();
    handleTaggedWithRemove();
    handleQueryValueRemove();
  }, [
    handleAccountStatusRemove,
    handleMoneySpentRemove,
    handleQueryValueRemove,
    handleTaggedWithRemove,
  ]);

  const filters = [
    {
      key: 'accountStatus',
      label: 'Account status',
      filter: (
        <ChoiceList
          title="Account status"
          titleHidden
          choices={[
            {label: 'Enabled', value: 'enabled'},
            {label: 'Not invited', value: 'not invited'},
            {label: 'Invited', value: 'invited'},
            {label: 'Declined', value: 'declined'},
          ]}
          selected={accountStatus || []}
          onChange={handleAccountStatusChange}
          allowMultiple
        />
      ),
      shortcut: true,
    },
    {
      key: 'taggedWith',
      label: 'Tagged with',
      filter: (
        <TextField
          label="Tagged with"
          value={taggedWith}
          onChange={handleTaggedWithChange}
          labelHidden
        />
      ),
      shortcut: true,
    },
  ];

  const appliedFilters = [];
  if (!isEmpty(accountStatus)) {
    const key = 'accountStatus';
    appliedFilters.push({
      key,
      label: disambiguateLabel(key, accountStatus),
      onRemove: handleAccountStatusRemove,
    });
  }
  if (!isEmpty(moneySpent)) {
    const key = 'moneySpent';
    appliedFilters.push({
      key,
      label: disambiguateLabel(key, moneySpent),
      onRemove: handleMoneySpentRemove,
    });
  }
  if (!isEmpty(taggedWith)) {
    const key = 'taggedWith';
    appliedFilters.push({
      key,
      label: disambiguateLabel(key, taggedWith),
      onRemove: handleTaggedWithRemove,
    });
  }

  const isDisabled =
    appliedFilters.length > 0 && appliedFilters[0].key === 'accountStatus';

  return (
    <div style={{height: '568px'}}>
      <Card>
        <ResourceList
          resourceName={{singular: 'customer', plural: 'customers'}}
          filterControl={
            <Filters
              queryValue={queryValue}
              filters={filters}
              appliedFilters={appliedFilters}
              onQueryChange={handleFiltersQueryChange}
              onQueryClear={handleQueryValueRemove}
              onClearAll={handleFiltersClearAll}
              disabled={isDisabled}
            />
          }
          items={[
            {
              id: 341,
              url: 'customers/341',
              name: 'Mae Jemison',
              location: 'Decatur, USA',
            },
            {
              id: 256,
              url: 'customers/256',
              name: 'Ellen Ochoa',
              location: 'Los Angeles, USA',
            },
          ]}
          renderItem={(item) => {

            return (
              <ResourceList.Item
                id={id}
                url={url}
                media={media}
              >
                <h3>
                </h3>
                <div>{location}</div>
              </ResourceList.Item>
            );
          }}
        />
      </Card>
    </div>
  );

  function disambiguateLabel(key, value) {
    switch (key) {
      case 'moneySpent':
        return `Money spent is between $${value[0]} and $${value[1]}`;
      case 'taggedWith':
        return `Tagged with ${value}`;
      case 'accountStatus':
        return value.map((val) => `Customer ${val}`).join(', ');
      default:
        return value;
    }
  }

  function isEmpty(value) {
    if (Array.isArray(value)) {
      return value.length === 0;
    } else {
      return value === '' || value == null;
    }
  }
}

  return (
    <Page title="Playground">
      <ResourceListFiltersExample />
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Filters/Filters.tsx
+++ b/src/components/Filters/Filters.tsx
@@ -20,7 +20,7 @@ import {ScrollLock} from '../ScrollLock';
 import {Icon} from '../Icon';
 import {TextField} from '../TextField';
 import {Tag} from '../Tag';
-import {TextStyle} from '../TextStyle';
+import {TextStyle, VariationValue} from '../TextStyle';
 import {Badge} from '../Badge';
 import {Focus} from '../Focus';
 import {Sheet} from '../Sheet';
@@ -50,6 +50,8 @@ export interface FilterInterface {
   filter: React.ReactNode;
   /** Whether or not the filter should have a shortcut popover displayed */
   shortcut?: boolean;
+  /** Whether or not the filter is disabled */
+  disabled?: boolean;
 }
 
 export interface FiltersProps {
@@ -75,6 +77,10 @@ export interface FiltersProps {
   onQueryFocus?(): void;
   /** The content to display inline with the controls */
   children?: React.ReactNode;
+  /** Disable all filters */
+  disabled?: boolean;
+  /** Additional hint text to display below the filters */
+  helpText?: string | React.ReactNode;
 }
 
 type ComposedProps = FiltersProps & WithAppProviderProps;
@@ -126,6 +132,8 @@ class Filters extends React.Component<ComposedProps, State> {
       onQueryClear,
       queryPlaceholder,
       children,
+      disabled = false,
+      helpText,
     } = this.props;
     const {resourceName} = this.context;
     const {open, readyForFocus} = this.state;
@@ -173,7 +181,17 @@ class Filters extends React.Component<ComposedProps, State> {
             aria-expanded={filterIsOpen}
           >
             <div className={styles.FilterTriggerLabelContainer}>
-              <h2 className={styles.FilterTriggerTitle}>{filter.label}</h2>
+              <h2 className={styles.FilterTriggerTitle}>
+                <TextStyle
+                  variation={
+                    this.props.disabled || filter.disabled
+                      ? VariationValue.Subdued
+                      : undefined
+                  }
+                >
+                  {filter.label}
+                </TextStyle>
+              </h2>
               <span className={styles.FilterTriggerIcon}>
                 <Icon source={icon} color="inkLightest" />
               </span>
@@ -196,7 +214,11 @@ class Filters extends React.Component<ComposedProps, State> {
 
     const rightActionMarkup = (
       <div ref={this.moreFiltersButtonContainer}>
-        <Button onClick={this.toggleFilters} testID="SheetToggleButton">
+        <Button
+          onClick={this.toggleFilters}
+          testID="SheetToggleButton"
+          disabled={disabled}
+        >
           {intl.translate('Polaris.Filters.moreFilters')}
         </Button>
       </div>
@@ -212,6 +234,7 @@ class Filters extends React.Component<ComposedProps, State> {
         rightPopoverableActions={this.transformFilters(filters)}
         rightAction={rightActionMarkup}
         auxiliary={children}
+        disabled={disabled}
       >
         <TextField
           placeholder={
@@ -239,6 +262,7 @@ class Filters extends React.Component<ComposedProps, State> {
           }
           clearButton
           onClearButtonClick={onQueryClear}
+          disabled={disabled}
         />
       </ConnectedFilterControl>
     );
@@ -311,6 +335,7 @@ class Filters extends React.Component<ComposedProps, State> {
                 onRemove={() => {
                   filter.onRemove(filter.key);
                 }}
+                disabled={disabled}
               >
                 {filter.label}
               </Tag>
@@ -349,11 +374,18 @@ class Filters extends React.Component<ComposedProps, State> {
       </Sheet>
     );
 
+    const helpTextMarkup = helpText ? (
+      <div id="FiltersHelpText" className={styles.HelpText}>
+        <TextStyle variation="subdued">{helpText}</TextStyle>
+      </div>
+    ) : null;
+
     return (
       <div className={styles.Filters}>
         {filtersControlMarkup}
         {filtersContainerMarkup}
         {tagsMarkup}
+        {helpTextMarkup}
         {backdropMarkup}
         <KeypressListener keyCode={Key.Escape} handler={this.closeFilters} />
       </div>
@@ -446,13 +478,14 @@ class Filters extends React.Component<ComposedProps, State> {
     const transformedActions: PopoverableAction[] = [];
 
     getShortcutFilters(filters).forEach((filter) => {
-      const {key, label} = filter;
+      const {key, label, disabled} = filter;
 
       transformedActions.push({
         popoverContent: this.generateFilterMarkup(filter),
         popoverOpen: this.state[`${key}${Suffix.Shortcut}`],
         key,
         content: label,
+        disabled,
         onAction: () => this.toggleFilterShortcut(key),
       });
     });
--- a/src/components/Filters/README.md
+++ b/src/components/Filters/README.md
@@ -650,3 +650,481 @@ function FiltersExample() {
   }
 }
 ```
+
+### All filters disabled
+
+```jsx
+function DisableAllFiltersExample() {
+  const [taggedWith, setTaggedWith] = useState(null);
+  const [queryValue, setQueryValue] = useState(null);
+
+  const handleTaggedWithChange = useCallback(
+    (value) => setTaggedWith(value),
+    [],
+  );
+  const handleQueryValueChange = useCallback(
+    (value) => setQueryValue(value),
+    [],
+  );
+  const handleTaggedWithRemove = useCallback(() => setTaggedWith(null), []);
+  const handleQueryValueRemove = useCallback(() => setQueryValue(null), []);
+
+  const handleClearAll = useCallback(() => {
+    handleTaggedWithRemove();
+    handleQueryValueRemove();
+  }, [handleQueryValueRemove, handleTaggedWithRemove]);
+
+  const filters = [
+    {
+      key: 'taggedWith',
+      label: 'Tagged with',
+      filter: (
+        <TextField
+          label="Tagged with"
+          value={taggedWith}
+          onChange={handleTaggedWithChange}
+          labelHidden
+        />
+      ),
+      shortcut: true,
+    },
+  ];
+
+  const appliedFilters = !isEmpty(taggedWith)
+    ? [
+        {
+          key: 'taggedWith',
+          label: disambiguateLabel('taggedWith', taggedWith),
+          onRemove: handleTaggedWithRemove,
+        },
+      ]
+    : [];
+
+  return (
+    <div style={{height: '568px'}}>
+      <Card>
+        <ResourceList
+          resourceName={{singular: 'customer', plural: 'customers'}}
+          filterControl={
+            <Filters
+              queryValue={queryValue}
+              filters={filters}
+              appliedFilters={appliedFilters}
+              onQueryChange={handleQueryValueChange}
+              onQueryClear={handleQueryValueRemove}
+              onClearAll={handleClearAll}
+              disabled
+            >
+              <div style={{paddingLeft: '8px'}}>
+                <Button
+                  disabled
+                  onClick={() => console.log('New filter saved')}
+                >
+                  Save
+                </Button>
+              </div>
+            </Filters>
+          }
+          items={[
+            {
+              id: 341,
+              url: 'customers/341',
+              name: 'Mae Jemison',
+              location: 'Decatur, USA',
+            },
+            {
+              id: 256,
+              url: 'customers/256',
+              name: 'Ellen Ochoa',
+              location: 'Los Angeles, USA',
+            },
+          ]}
+          renderItem={(item) => {
+            const {id, url, name, location} = item;
+            const media = <Avatar customer size="medium" name={name} />;
+
+            return (
+              <ResourceList.Item
+                id={id}
+                url={url}
+                media={media}
+                accessibilityLabel={`View details for ${name}`}
+              >
+                <h3>
+                  <TextStyle variation="strong">{name}</TextStyle>
+                </h3>
+                <div>{location}</div>
+              </ResourceList.Item>
+            );
+          }}
+        />
+      </Card>
+    </div>
+  );
+
+  function disambiguateLabel(key, value) {
+    switch (key) {
+      case 'taggedWith':
+        return `Tagged with ${value}`;
+      default:
+        return value;
+    }
+  }
+
+  function isEmpty(value) {
+    if (Array.isArray(value)) {
+      return value.length === 0;
+    } else {
+      return value === '' || value == null;
+    }
+  }
+}
+```
+
+### Some filters disabled
+
+```jsx
+function DisableSomeFiltersExample() {
+  const [taggedWith, setTaggedWith] = useState(null);
+  const [vendor, setVendor] = useState(null);
+  const [queryValue, setQueryValue] = useState(null);
+
+  const handleTaggedWithChange = useCallback(
+    (value) => setTaggedWith(value),
+    [],
+  );
+  const handleQueryValueChange = useCallback(
+    (value) => setQueryValue(value),
+    [],
+  );
+  const handleVendorChange = useCallback((value) => setVendor(value), []);
+
+  const handleTaggedWithRemove = useCallback(() => setTaggedWith(null), []);
+  const handleQueryValueRemove = useCallback(() => setQueryValue(null), []);
+  const handleVendorRemove = useCallback(() => setVendor(null), []);
+
+  const handleClearAll = useCallback(() => {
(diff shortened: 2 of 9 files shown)


src/components/Filters/Filters.tsx
```diff
@@ -171,9 +179,14 @@ class Filters extends React.Component<ComposedProps, State> {
             type="button"
             aria-controls={collapsibleID}
             aria-expanded={filterIsOpen}
+            disabled={disabled}
           >
             <div className={styles.FilterTriggerLabelContainer}>
-              <h2 className={styles.FilterTriggerTitle}>{filter.label}</h2>
+              <h2 className={styles.FilterTriggerTitle}>
+                <TextStyle {...filterTriggerTitleTextStyle}>
```
dleroux: This would be a little less code
`<TextStyle  variation={disabled ? VariationValue.Subdued : undefined}>`

src/components/Filters/README.md
```diff
@@ -650,3 +650,130 @@ function FiltersExample() {
   }
 }
 ```
+
+### Disabling filters
```
dleroux: Disabled filters

dleroux:
Code looks good. Thank you for adding the examples!

One thing... The tags that get generated should probably also be disabled as well because I assume removing them would ### have a negative impact on the filtering in your case.

PabloVallejo:
The way our workflow works is that filters are enabled and once merchants add one particular filter, they get disabled and we should a text that they need to remove that filter in order to use the other ones. 

For this case, having the tags available allows merchants to remove the filter and be able to add more again.

I guess we could disable everything and provide the merchants with an action link/button to remove applied filters enabling them again but removing the tags seems to be clean for them.  We do that now in Rails.

I strongly prefer this solution, but if you'd rather have everything disabled I can add the changes too.

dleroux:
I see, a bit of a special usecase. I wonder if someone involved in the dev/design of this component can chime in. @sarahill, @danrosenthal

dleroux:
I can see someone down the road thinking this isn't intended and breaking your use case.

PabloVallejo:
@dleroux analyzing our use case, we decided that along with the descriptive text on why the filters get disabled when you apply a particular filter, we're going to add a link to reset the filters. 

In that sense, the filters components will be fully disabled once you pass `disabled` prop as `true`. Keeping consistency between the name of the prop and the state of the component.

dleroux:
👍 I think from an accessibility perspective it might be a better choice/ux. Thanks, Pablo.

PabloVallejo:
@dleroux sounds good! Yea, in that way the disabled prop ends up being more semantic!

The change is up and just fixed the merge conflict ✅

src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.tsx
```diff
@@ -183,30 +184,34 @@ export class ConnectedFilterControl extends React.Component<
...
+        onClick={action.onAction}
+        disclosure
+        disabled={this.props.disabled}
+      >
+        {action.content}
+      </Button>
     );
-  });
-}
+  }
```
devonpmack: Is this to stop them from clicking on popover elements if the filter gets disabled while the popover is already open? If so could we just close the popover if the filter is disabled instead?

src/components/Filters/tests/Filters.test.tsx
```diff
@@ -293,6 +293,84 @@ describe('<Filters />', () => {
       expect(spy).toHaveBeenCalledWith('filterOne');
     });
   });
+
+  describe('disabled prop is passed as true', () => {
+    it('filters search field is diabled', () => {
```
devonpmack:
```suggestion
    it('disables filter search field', () => {
```

src/components/Filters/tests/Filters.test.tsx
```diff
@@ -293,6 +293,84 @@ describe('<Filters />', () => {
...
+      const resourceFilters = mountWithAppProvider(
+        <Filters {...mockProps} queryValue="bar" disabled />,
+      );
+
+      expect(resourceFilters.find(TextField).prop('disabled')).toBe(true);
+    });
+
+    it('sheet toggle button is disabled', () => {
```
devonpmack:
```suggestion
    it('disables sheet toggle button', () => {
```

src/components/Filters/tests/Filters.test.tsx
```diff
@@ -293,6 +293,84 @@ describe('<Filters />', () => {
...
+      const resourceFilters = mountWithAppProvider(
+        <Filters {...mockProps} queryValue="bar" disabled />,
+      );
+
+      expect(
+        resourceFilters.find(ConnectedFilterControl).prop('disabled'),
+      ).toBe(true);
+    });
+
+    it('sheet filter toggle button is disabled', () => {
```
devonpmack:
```suggestion
    it('disables sheet filter toggle buttons', () => {
```

src/components/Filters/tests/Filters.test.tsx
```diff
@@ -293,6 +293,84 @@ describe('<Filters />', () => {
...
+        const toggleButton = findById(
+          resourceFilters,
+          `${filter.key}ToggleButton`,
+        );
+
+        expect(toggleButton.prop('disabled')).toBe(true);
+      });
+    });
+
+    it('sheet filters heading text is subdued', () => {
```
devonpmack:
```suggestion
    it('subdues sheet filter heading text', () => {
```

src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.tsx
devonpmack: Is this to stop them from clicking on popover elements if the filter gets disabled while the popover is already open? If so could we just close the popover if the filter is disabled instead?

devonpmack:
When tophatting, it seems like I can't remove the filter after enabling it

image

devonpmack:
image

It would be great if we could also disable the onHover of the drop downs. Right now it changes your cursor to 'link mode' and darkens the button. This makes it seem like you can click it even though you can't. It doesn't really match the way other disabled buttons work in polaris. I understand that this could be hard to fix though.

src/components/Filters/Filters.tsx
```diff
@@ -75,6 +75,8 @@ export interface FiltersProps {
   onQueryFocus?(): void;
   /** The content to display inline with the controls */
   children?: React.ReactNode;
+  /** Disables the filters when set to true */
```
chloerice:
```suggestion
  /** Disable all filters */
```

src/components/Filters/README.md
```diff
@@ -650,3 +650,130 @@ function FiltersExample() {
...
+              queryValue={queryValue}
+              filters={filters}
+              appliedFilters={appliedFilters}
+              onQueryChange={handleQueryValueChange}
+              onQueryClear={handleQueryValueRemove}
+              onClearAll={handleClearAll}
+              disabled
+            >
+              <div style={{paddingLeft: '8px'}}>
+                <Button onClick={() => console.log('New filter saved')}>
```
chloerice: Let's disable the save button too since the filters themselves are disabled.

```suggestion
                <Button disabled onClick={() => console.log('New filter saved')}>
```

src/components/Filters/Filters.tsx
```diff
@@ -171,9 +174,16 @@ class Filters extends React.Component<ComposedProps, State> {
             type="button"
             aria-controls={collapsibleID}
             aria-expanded={filterIsOpen}
+            disabled={disabled}
           >
             <div className={styles.FilterTriggerLabelContainer}>
-              <h2 className={styles.FilterTriggerTitle}>{filter.label}</h2>
+              <h2 className={styles.FilterTriggerTitle}>
+                <TextStyle
+                  variation={disabled ? VariationValue.Subdued : undefined}
```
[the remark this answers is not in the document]
dleroux: Is there a use-case for disabling individual filters?

Good catch on it not being needed since the sheet can't be opened.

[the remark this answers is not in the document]
chloerice: I think so, but if you think it's out of scope for this PR it can just be added when it comes up in the future.

src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.tsx
chloerice: Ahh, I think the diff made this confusing. This was a functional component but Pablo converted it to a class because these two functions both need access to `props.disabled`.

devonpmack: Oops

dleroux:
> It would be great if we could also disable the onHover of the drop downs. Right now it changes your cursor to 'link mode' and darkens the button. This makes it seem like you can click it even though you can't. It doesn't really match the way other disabled buttons work in polaris. I understand that this could be hard to fix though.

Hi @devonpmack are you saying with the filters disabled you were able to open the sheet? These should be clickable if filters aren't disabled, they work as an accordion. Is that no the case for you?

src/components/Filters/Filters.tsx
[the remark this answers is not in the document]
PabloVallejo: To give some context, our workflow is:

* While all filters are active
* Merchant opens the sheet
* Sets a filter by collection: This action disables other filters but the sheet is still open. 

So we'll need to disable filters in the sheet, but we don't need to disable all of them, so adding support for disabling individual filters is the way to go.

src/components/Filters/README.md
```diff
@@ -650,3 +650,130 @@ function FiltersExample() {
   }
 }
 ```
+
+### Disabled filters
```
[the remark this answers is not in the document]
PabloVallejo: Added the example!

PabloVallejo:
@chloerice @dleroux @devonpmack , I just pushed some code that:

* Adds support for disabling individual filters
* Adds a new prop: `HelpText` to Filters & add tests for it
* Fix filter tests

I was chatting IRL with Devon yesterday and he noticed that not being able to remove a filter feels blocking. To add some context, here's the workflow we're trying to replicate:

filters-gif

We have two options to do this:

### 1. Filter tags are available for merchants to remove them

image


### 2. Filter tags are not available and merchants are given a help text with a link to reset them and make them available.

image

Allowing the removal of tags feels more natural in my opinion but I'm open to both options.

PabloVallejo:
@dleroux what @devonpmack meant is that if before disabling the filters you open the sheet, and then disable the filters, you'd see the headers as links.

The sheet can be open before the filters are disabled.

dleroux:
I see what you are saying about leaving the tags enabled. Maybe a designer can weigh in, @sarahill ?

I was thinking if you disable because a request is in progress, removing the tag would defy the purpose. In this case I can see why you would want it.

PabloVallejo:
@dleroux sounds good! I'm going to talk to @sarahill to discuss this one! Thanks 🙏🏼🙏🏼

sarahill:
Hey @PabloVallejo, I've got a few questions so I have some more context:

- What's the reasoning for only 1 filter at a time?
- Does the search count as a filter (could we leave the search enabled)?

PabloVallejo:
👋🏼 @sarahill 

The TL;DR is that because of backend limitation, the collection filter cannot be combined with other filters, therefore:

* We're limiting merchants to add other filters if they filter by collection.
* The search does count as a filter and cannot be combined with the collections filter hence cannot be enabled.

src/components/Filters/Filters.tsx
```diff
@@ -212,6 +233,7 @@ class Filters extends React.Component<ComposedProps, State> {
         rightPopoverableActions={this.transformFilters(filters)}
```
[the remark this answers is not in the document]
chloerice: The `PopoverableAction` interface defined in `ConnectedFilterControl.tsx` also needs to be updated to extend `DisableableAction` instead of `Action`.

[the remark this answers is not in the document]
chloerice: The type of the `action` argument of the `activatorButtonFrom` method needs to be updated to `PopoverableAction` instead of `BaseAction`.

src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.tsx
```diff
@@ -183,30 +184,34 @@ export class ConnectedFilterControl extends React.Component<
...
-    <Button onClick={action.onAction} disclosure>
-      {action.content}
-    </Button>
-  );
+  private popoverFrom(actions: PopoverableAction[]): React.ReactElement<any>[] {
+    return actions.map((action) => {
+      return (
+        <Item key={action.key}>
+          <Popover
+            active={this.props.disabled ? false : action.popoverOpen}
```
chloerice:
```suggestion
            active={action.popoverOpen}
```

I don't think this needs to be conditional since the popover activator is `disabled` when the filter is `disabled`.

PabloVallejo:
Discussing IRL with @dleroux and @sarahill the approach that made the most sense was approach number 2: Disabling filters and tags altogether using the disable prop and giving merchants a link to reset the filters:



I've updated the code accordingly and addressed review comments.

I also did a couple of tests with extracting `Filter` component from the  `ResourceList` but a refactor would be needed in order to allow this as the styles get messed up (see screenshot). So I'll keep things inside of `ResourceList`: 

image


Similarly, I tried adding the `HelpText` after the filters but was not happy with the looks, what do you think @dleroux, @sarahill:

image

Deciding on the placement of the help text would be the last detail and this PR should be good to 🚢

src/components/Filters/Filters.tsx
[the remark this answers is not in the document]
PabloVallejo: Nice suggestion, this allows for disabling individual activators! Added it to the code!

sarahill:
@pablogolan cc: @daniedleroux 

For the text, can we do something like the attached? If it's easier to leave the text below the tags that's ok. I think this reads a little better and then on small screens the text + link can go below.

Screen Shot 2019-11-06 at 10 59 17 AM

dleroux:
@sarahill, yes that would be ideal and in line with what we do in the bulk actions today. However not possible without a new prop. Is this a prop we see value in? an `inlineAction` maybe?

If not then it might be best to merge the disable stuff and then fork it for your use case @PabloVallejo and then delete it when the limitation is no longer.

pablogolan:
@sarahill I think you meant to ping Pablo Vallejo, not me.

pablogolan:
Can someone please remove me from here. I am still getting emails regarding this issue. Thanks!

PabloVallejo:
Hey Pablo, sorry about that, you can unsubscribe:

image


But since the PR was just merged we won't be discussing on it anymore! 👌🏼👍🏼

PR #2397, opened 2019-11-04
[TrapFocus] Focus should stay within the container

LauraAubin:

This PR is fixing `TrapFocus` so that focus remains within a designated container.

The focus flow on each press of tab works like,

> focus container -> focus first element -> focus second element(last element) -> focus first element -> focus second element -> etc

Instead of returning focus back to the container.
 
<kbd></kbd>

### How to 🎩

Use Storybook. Open up a modal and press tab. Once you reach the last element, the next tab should focus the first element in the modal.

### 🎩 checklist

### DIFF
--- a/src/components/Focus/Focus.tsx
+++ b/src/components/Focus/Focus.tsx
@@ -1,3 +1,7 @@
+/*
+`Focus` will automatically focus the first focusable element that it finds in it's children when it is mounted, unless the user specified an `autoFocus` element.
+*/
+
 import React from 'react';
 import {focusFirstFocusableNode} from '@shopify/javascript-utilities/focus';
 import isEqual from 'lodash/isEqual';
--- a/src/components/Focus/Focus.tsx
+++ b/src/components/Focus/Focus.tsx
@@ -1,7 +1,3 @@
-/*
-`Focus` will automatically focus the first focusable element that it finds in it's children when it is mounted, unless the user specified an `autoFocus` element.
-*/
-
 import React from 'react';
 import {focusFirstFocusableNode} from '@shopify/javascript-utilities/focus';
 import isEqual from 'lodash/isEqual';
--- a/src/components/TrapFocus/TrapFocus.tsx
+++ b/src/components/TrapFocus/TrapFocus.tsx
@@ -93,7 +93,7 @@ export class TrapFocus extends React.PureComponent<TrapFocusProps, State> {
       event.preventDefault();
 
       if (event.srcElement === findFirstFocusableNode(focusTrapWrapper)) {
-        return focusLastFocusableNode(focusTrapWrapper);
+        return write(() => focusLastFocusableNode(focusTrapWrapper));
       }
       const firstNode = findFirstFocusableNode(focusTrapWrapper) as HTMLElement;
       write(() => focusFirstFocusableNode(firstNode));
--- a/src/components/TrapFocus/TrapFocus.tsx
+++ b/src/components/TrapFocus/TrapFocus.tsx
@@ -1,10 +1,20 @@
+/*
+A component which allows you to trap keyboard focus inside of a container.
+
+TrapFocus internally employs `Focus` to focus it's first focusable child on mount.
+
+Whenever a `blur` event occurs that would take the user outside the trap, we reset to the first focusable child.
+
+If you want to cease trapping focus, simply cease rendering the trap.
+*/
+
 import React from 'react';
-import {closest} from '@shopify/javascript-utilities/dom';
 import {
   focusFirstFocusableNode,
   findFirstFocusableNode,
   focusLastFocusableNode,
 } from '@shopify/javascript-utilities/focus';
+import {write} from '@shopify/javascript-utilities/fastdom';
 
 import {EventListener} from '../EventListener';
 import {Focus} from '../Focus';
@@ -44,7 +54,7 @@ export class TrapFocus extends React.PureComponent<TrapFocusProps, State> {
 
     return (
       <Focus disabled={this.shouldDisable} root={this.focusTrapWrapper}>
-        <div ref={this.setFocusTrapWrapper}>
+        <div ref={this.setFocusTrapWrapper} style={{background: 'salmon'}}>
           <EventListener event="focusout" handler={this.handleBlur} />
           {children}
         </div>
@@ -68,25 +78,31 @@ export class TrapFocus extends React.PureComponent<TrapFocusProps, State> {
   };
 
   private handleBlur = (event: FocusEvent) => {
-    const {relatedTarget} = event;
-    const {focusTrapWrapper} = this;
+    const {relatedTarget: currentTarget} = event;
+    const {focusTrapWrapper} = this; // the div around the modal
     const {trapping = true} = this.props;
 
-    if (relatedTarget == null || trapping === false) {
-      return;
+    if (currentTarget == null || trapping === false) {
+      // do nothing
+      // return;
     }
 
+    // if we have declared a container
+    // and the currently selected element is outside of that container
+    // and the current target has no parent with the `data-polaris-overlay` property
     if (
       focusTrapWrapper &&
-      !focusTrapWrapper.contains(relatedTarget as HTMLElement) &&
-      !closest(relatedTarget as HTMLElement, '[data-polaris-overlay]')
+      !focusTrapWrapper.contains(currentTarget as HTMLElement) // &&
+      // !closest(currentTarget as HTMLElement, '[data-polaris-overlay]')
     ) {
       event.preventDefault();
 
       if (event.srcElement === findFirstFocusableNode(focusTrapWrapper)) {
+        // !!
         return focusLastFocusableNode(focusTrapWrapper);
       }
-      focusFirstFocusableNode(focusTrapWrapper);
+      const firstNode = findFirstFocusableNode(focusTrapWrapper) as HTMLElement;
+      write(() => focusFirstFocusableNode(firstNode));
     }
   };
 }
--- a/src/components/TrapFocus/TrapFocus.tsx
+++ b/src/components/TrapFocus/TrapFocus.tsx
@@ -82,8 +82,8 @@ export class TrapFocus extends React.PureComponent<TrapFocusProps, State> {
     const {focusTrapWrapper} = this;
     const {trapping = true} = this.props;
 
-    if (currentTarget == null || trapping === false) {
-      // return;
+    if (trapping === false) {
+      return;
     }
 
     if (
--- a/src/components/TrapFocus/TrapFocus.tsx
+++ b/src/components/TrapFocus/TrapFocus.tsx
@@ -1,13 +1,3 @@
-/*
-A component which allows you to trap keyboard focus inside of a container.
-
-TrapFocus internally employs `Focus` to focus it's first focusable child on mount.
-
-Whenever a `blur` event occurs that would take the user outside the trap, we reset to the first focusable child.
-
-If you want to cease trapping focus, simply cease rendering the trap.
-*/
-
 import React from 'react';
 import {
   focusFirstFocusableNode,
--- a/src/components/TrapFocus/TrapFocus.tsx
+++ b/src/components/TrapFocus/TrapFocus.tsx
@@ -54,7 +54,7 @@ export class TrapFocus extends React.PureComponent<TrapFocusProps, State> {
 
     return (
       <Focus disabled={this.shouldDisable} root={this.focusTrapWrapper}>
-        <div ref={this.setFocusTrapWrapper} style={{background: 'salmon'}}>
+        <div ref={this.setFocusTrapWrapper}>
           <EventListener event="focusout" handler={this.handleBlur} />
           {children}
         </div>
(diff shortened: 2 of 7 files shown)


src/components/Focus/Focus.tsx
```diff
@@ -1,3 +1,7 @@
+/*
+`Focus` will automatically focus the first focusable element that it finds in it's children when it is mounted, unless the user specified an `autoFocus` element.
+*/
```
LauraAubin: This definition, and the one for `TrapFocus` used to live in a readme file that was removed. I'd like to resurface these since they really helped me how `Focus` and `TrapFocus` works.

src/components/TrapFocus/TrapFocus.tsx
```diff
@@ -68,25 +78,31 @@ export class TrapFocus extends React.PureComponent<TrapFocusProps, State> {
...
 
+    // if we have declared a container
+    // and the currently selected element is outside of that container
+    // and the current target has no parent with the `data-polaris-overlay` property
     if (
       focusTrapWrapper &&
-      !focusTrapWrapper.contains(relatedTarget as HTMLElement) &&
-      !closest(relatedTarget as HTMLElement, '[data-polaris-overlay]')
+      !focusTrapWrapper.contains(currentTarget as HTMLElement) // &&
+      // !closest(currentTarget as HTMLElement, '[data-polaris-overlay]')
```
LauraAubin: > "If we have a container, and the currently selected element is outside of that container, and the current target has no parent with the `data-polaris-overlay` property".

src/components/TrapFocus/TrapFocus.tsx
```diff
@@ -68,25 +78,31 @@ export class TrapFocus extends React.PureComponent<TrapFocusProps, State> {
...
     ) {
       event.preventDefault();
 
       if (event.srcElement === findFirstFocusableNode(focusTrapWrapper)) {
```
LauraAubin: When is this called?

src/components/TrapFocus/TrapFocus.tsx
```diff
@@ -68,25 +78,31 @@ export class TrapFocus extends React.PureComponent<TrapFocusProps, State> {
   };
 
   private handleBlur = (event: FocusEvent) => {
-    const {relatedTarget} = event;
-    const {focusTrapWrapper} = this;
+    const {relatedTarget: currentTarget} = event;
+    const {focusTrapWrapper} = this; // the div around the modal
```
LauraAubin: This is the `div` around the modal.

src/components/TrapFocus/TrapFocus.tsx
```diff
@@ -68,25 +78,25 @@ export class TrapFocus extends React.PureComponent<TrapFocusProps, State> {
...
 
     if (
       focusTrapWrapper &&
-      !focusTrapWrapper.contains(relatedTarget as HTMLElement) &&
-      !closest(relatedTarget as HTMLElement, '[data-polaris-overlay]')
+      !focusTrapWrapper.contains(currentTarget as HTMLElement)
     ) {
       event.preventDefault();
 
       if (event.srcElement === findFirstFocusableNode(focusTrapWrapper)) {
```
LauraAubin: When is this called?

src/components/TrapFocus/TrapFocus.tsx
```diff
@@ -68,25 +78,25 @@ export class TrapFocus extends React.PureComponent<TrapFocusProps, State> {
...
-      return;
+    if (currentTarget == null || trapping === false) {
+      // return;
     }
 
     if (
       focusTrapWrapper &&
-      !focusTrapWrapper.contains(relatedTarget as HTMLElement) &&
-      !closest(relatedTarget as HTMLElement, '[data-polaris-overlay]')
+      !focusTrapWrapper.contains(currentTarget as HTMLElement)
```
LauraAubin: > "If we have a container, and the currently selected element is outside of that container, and the current target has no parent with the data-polaris-overlay property".

src/components/TrapFocus/TrapFocus.tsx
```diff
@@ -68,25 +78,25 @@ export class TrapFocus extends React.PureComponent<TrapFocusProps, State> {
   };
 
   private handleBlur = (event: FocusEvent) => {
-    const {relatedTarget} = event;
+    const {relatedTarget: currentTarget} = event;
     const {focusTrapWrapper} = this;
```
LauraAubin: This is the `div` around the modal.

src/components/TrapFocus/TrapFocus.tsx
```diff
@@ -68,25 +78,25 @@ export class TrapFocus extends React.PureComponent<TrapFocusProps, State> {
...
 
-    if (relatedTarget == null || trapping === false) {
+    if (trapping === false) {
       return;
     }
 
     if (
       focusTrapWrapper &&
-      !focusTrapWrapper.contains(relatedTarget as HTMLElement) &&
-      !closest(relatedTarget as HTMLElement, '[data-polaris-overlay]')
```
LauraAubin: @AndrewMusgrave Do you think we need this line? I can't think of a situation when we would need to find an element with the property `data-polaris-overlay`?

src/components/Focus/Focus.tsx
dleroux: I'm not sure these comments are the best way to document these 🤔 . Do we know if the readme was removed because they were causing issues with the style guide? Wondering if jsdocs style comments like we use for hooks would make more sense.

src/components/TrapFocus/TrapFocus.tsx
```diff
@@ -1,10 +1,20 @@
+/*
+A component which allows you to trap keyboard focus inside of a container.
+
+TrapFocus internally employs `Focus` to focus it's first focusable child on mount.
+
+Whenever a `blur` event occurs that would take the user outside the trap, we reset to the first focusable child.
+
+If you want to cease trapping focus, simply cease rendering the trap.
+*/
+
```
dleroux: Again I think tests provide this description?

src/components/TrapFocus/TrapFocus.tsx
```diff
@@ -68,25 +78,25 @@ export class TrapFocus extends React.PureComponent<TrapFocusProps, State> {
   };
 
   private handleBlur = (event: FocusEvent) => {
-    const {relatedTarget} = event;
+    const {relatedTarget: currentTarget} = event;
```
dleroux: you probably just want to keep this as `relatedTarget`. Renaming it to currentTarget is a bit confusing since it isn't the curentTarget.

src/components/TrapFocus/TrapFocus.tsx
```diff
@@ -68,25 +78,25 @@ export class TrapFocus extends React.PureComponent<TrapFocusProps, State> {
...
     const {focusTrapWrapper} = this;
     const {trapping = true} = this.props;
 
-    if (relatedTarget == null || trapping === false) {
+    if (trapping === false) {
```
dleroux: Was the extra check causing issues? `relatedTarget == null || `

src/components/TrapFocus/TrapFocus.tsx
dleroux: Hard to tell because I couldn't find the original commit. But if it's not part of the issue I would leave it as to not introduce more issues. It is extra precaution.

src/components/Focus/Focus.tsx
AndrewMusgrave: Iirc we never had readmes. When I added these components we didn't want to expose them. Readme would be the place for these comments, if you want to add some documentation you could try adding a `README_INTERNAL.md` file or something along those lines. I'm *sort of* sure we only look for 'README.md'

src/components/TrapFocus/TrapFocus.tsx
AndrewMusgrave: ^

src/components/TrapFocus/TrapFocus.tsx
AndrewMusgrave: ➕ 1️⃣

src/components/TrapFocus/TrapFocus.tsx
AndrewMusgrave: We'll probably want to keep the check since `relatedTarget` can be `null`. We can slightly change it though. `relatedTarget instanceOf HTMLElement` we are also able to remove the type cast

src/components/TrapFocus/TrapFocus.tsx
AndrewMusgrave: I'm not sure tbh, `d-p-overlay` is legacy but were not 100% confident in removing it. As Dan mentioned it removing it required for this change?

src/components/TrapFocus/TrapFocus.tsx
```diff
@@ -68,25 +78,25 @@ export class TrapFocus extends React.PureComponent<TrapFocusProps, State> {
...
       focusTrapWrapper &&
-      !focusTrapWrapper.contains(relatedTarget as HTMLElement) &&
-      !closest(relatedTarget as HTMLElement, '[data-polaris-overlay]')
+      !focusTrapWrapper.contains(currentTarget as HTMLElement)
     ) {
       event.preventDefault();
 
       if (event.srcElement === findFirstFocusableNode(focusTrapWrapper)) {
-        return focusLastFocusableNode(focusTrapWrapper);
+        return write(() => focusLastFocusableNode(focusTrapWrapper));
```
AndrewMusgrave: For context, as I usually ask, why is RAF needed for this change?

src/components/Focus/Focus.tsx
LauraAubin: Yup it looks like the style guide is hardcoded to `readme.md`:

https://github.com/Shopify/polaris-styleguide/blob/dbc092cd68b0e4c90836995112403894fabc4e73/scripts/parseComponentTypes.js#L12

LauraAubin: FYI this is where I found the definitions https://github.com/Shopify/polaris-react/commit/633c4595880c21216d4cacf2323451b68ca2c79a#diff-475a76122fa5410d0a4098dbc54d3e09

src/components/TrapFocus/TrapFocus.tsx
LauraAubin: Removing it is not required for this change, though I was mostly hoping we could remove it if it was not being used as things appear to work fine without it. But we can keep it if we're not 100% confident in removing it.

src/components/TrapFocus/TrapFocus.tsx
LauraAubin: In find in some cases, `relatedTarget` can be `null` if the next focusable node is outside of the page content, like the browser search bar. In these cases, I think we still want to manage focus and avoid the early `return`.

src/components/TrapFocus/TrapFocus.tsx
```diff
@@ -72,21 +73,24 @@ export class TrapFocus extends React.PureComponent<TrapFocusProps, State> {
...
-    if (relatedTarget == null || trapping === false) {
+    if (trapping === false) {
       return;
     }
 
     if (
       focusTrapWrapper &&
       !focusTrapWrapper.contains(relatedTarget as HTMLElement) &&
-      !closest(relatedTarget as HTMLElement, '[data-polaris-overlay]')
+      (relatedTarget
```
LauraAubin: This ternary is doing the same thing as the previous check `relatedTarget == null`.

src/components/TrapFocus/TrapFocus.tsx
LauraAubin: Without it, focus seems to leave the window content, despite the active element being the first/last focusable node.

src/components/TrapFocus/TrapFocus.tsx
LauraAubin: @dleroux @AndrewMusgrave After checking with Devon, we do want to keep focus within the modal and prevent tabs from focusing browser elements (search bar, tabs). A user can always press the escape key, or navigate to the x button if they want to close the modal.

According to [this](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html) ARIA example, 

> When focus is on the last focusable element in the dialog, moves focus to the first focusable element in the dialog.

You can also see this in action here: https://svinkle.github.io/launchy/

That means we should keep the change to remove the check `relatedTarget == null ` so that in the case that focus leaves the window content making it null, we don't have an early return (since our focus logic comes after this check).

AndrewMusgrave: Still slightly worried about this, since `relatedTarget` has other situations where is can be null. We'll want a solid 🎩  in web when we 🚢 this.

src/components/TrapFocus/TrapFocus.tsx
AndrewMusgrave: `!relatedTarget || !closest(relatedTarget as HTMLElement, '[data-polaris-overlay]')`

Would this work as well rather than the nested conditionals?

AndrewMusgrave:
My second comment should also make code coverage pass as well 😄

src/components/TrapFocus/TrapFocus.tsx
LauraAubin: It's lovely 🎉

src/components/TrapFocus/tests/TrapFocus.test.tsx
```diff
@@ -92,21 +96,47 @@ describe('<TrapFocus />', () => {
...
 
-      trigger(trapFocus.find(EventListener), 'handler', event);
+      afterEach(() => {
+        rafSpy.mockRestore();
+      });
+
+      it('has one focusable node', () => {
+        trigger(trapFocus.find(EventListener), 'handler', {
+          ...event,
+          srcElement: findFirstFocusableNode(trapFocus.getDOMNode()),
```
LauraAubin: If there is one focusable node in a container, then the first and second element are the same so `srcElement` will be the first element in this test.

src/components/TrapFocus/tests/TrapFocus.test.tsx
```diff
@@ -92,21 +96,47 @@ describe('<TrapFocus />', () => {
...
+        });
+
+        expect(event.preventDefault).toHaveBeenCalled();
+      });
+
+      it('it has multiple focusable nodes', () => {
+        trigger(trapFocus.find(EventListener), 'handler', {
+          ...event,
+          srcElement: findLastFocusableNode(trapFocus.getDOMNode()),
```
LauraAubin: If there are **multiple** focusable nodes in a container, then the first and second element are _not_ the same so `srcElement` will be the last element in this test.

LauraAubin:
^ Addressed reviews and added some tests since last approval

PR #2401, opened 2019-11-05
[TextField] Should always have an aria-labelledby property

LauraAubin:

If the text field does not have a `prefix` or `suffix`, then it doesn't have an `aria-labelledby` value. Since `label` is a required prop on TextField, we will always have an id to link to.

Removing the check for an empty array since the labelledby property should have a value regardless if the `TextField` has a prefix or suffix.

### How to 🎩

- Open up Storybook
- Go to the "Modal with primary action" example
- Inspect the modal's TextField
- The TextField should have an `aria-labelledby` matching the label element.

image

- Notice the label's id `PolarisTextField2Label` on the input.

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -14,4 +14,6 @@
 
 ### Code quality
 
+- Changed `aria-labelledby` to always exist on `TextField` ([#2401](https://github.com/Shopify/polaris-react/pull/2401))
+
 ### Deprecations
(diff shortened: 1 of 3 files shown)


dleroux:
🎩 works great! Just thinking we should probably add a test for this. Coverage is fine but this is not covered. What do you think?

LauraAubin:
Probably best to add a test so that this doesn't get changed in the future.

PR #2402, opened 2019-11-05
Add Kitchen sink and Details page stories

tmlayton:

Adds a Kitchen Sink story to the Playground stories. The Kitchen Sink ideally renders _every_ story on a single page, but quite a few stories need to be excluded due to fixed positioning layout and default open states. This is something we can add back in manually to the Kitchen Sink at a later time.

This story is useful for quickly checking global changes, specifically we are using it as we apply and test global theming.

Also adds a Details page story which functions similar to the kitchen sink, but is a crafted page vs a dump of every story.

kitchensink

### 🎩 checklist


tmlayton:
If we decide to move forward with this should we split it into its own bundle? Or maybe generate this story at runtime?

.storybook/config.js
```diff
@@ -109,7 +110,8 @@ function addPlaygroundStory() {
     .addParameters({
       chromatic: {disable: true},
     })
-    .add('Playground', () => <Playground />);
+    .add('Playground', () => <Playground />)
+    .add('Kitchen Sink', () => <KitchenSink />);
```
kaelig: 💅 For consistency with other component stories where the second word starts with a lowercase letter (Choice list, Button group…):
```suggestion
    .add('Kitchen sink', () => <KitchenSink />);
```

danrosenthal:
> If we decide to move forward with this should we split it into its own bundle

+4k is a big increase, so this might make sense

kaelig:
I'm not too concerned about the bundle size (242k -> 246k), however if there is a significant compilation time cost (slowing down `yarn dev` and hot reloading), it'd be good to know.

playground/stories.js
```diff
@@ -0,0 +1,70 @@
+export * from '../src/components/AccountConnection/README.md';
```
kaelig: Is this list manually generated?

BPScott: It is currently, but I've worked out a way to get this built out automatically, just waiting to demo it to Tim

UNRELEASED.md
```diff
@@ -14,6 +14,8 @@
 
 ### Documentation
 
+- Added a Kitchen Sink example to Storybook ([#2402](https://github.com/Shopify/polaris-react/pull/2402))
```
```suggestion
- Added a details page and kitchen sink example to Storybook ([#2402](https://github.com/Shopify/polaris-react/pull/2402))
```

PR #2404, opened 2019-11-06
[Color system] Update loading

tmlayton:
Addresses https://github.com/Shopify/polaris-ux/issues/345

### DIFF
--- a/src/components/Frame/Frame.scss
+++ b/src/components/Frame/Frame.scss
@@ -173,8 +173,11 @@ $skip-vertical-offset: rem(10px);
 
   @include frame-when-nav-displayed {
     .hasTopBar & {
-      top: top-bar-height();
-      z-index: z-index(loading-bar, $fixed-element-stacking-order);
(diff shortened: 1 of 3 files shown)


src/components/Frame/Frame.scss
```diff
@@ -173,8 +173,11 @@ $skip-vertical-offset: rem(10px);
 
   @include frame-when-nav-displayed {
     .hasTopBar & {
-      top: top-bar-height();
-      z-index: z-index(loading-bar, $fixed-element-stacking-order);
+      top: var(--p-override-zero, top-bar-height());
```
LauraAubin: Why are we overriding `rem(56px)` with 0?

danrosenthal: In the new design language, the `Loading` bar moves to the very top of the application. In current-form Polaris, it sits below the `TopBar`. Overriding to `0` moves it into that top position.

PR #2405, opened 2019-11-06
[color-transformers] Remove rgbToHsbl conversion rounding

ruairiphackett:

Fixes issue described here: https://github.com/Shopify/polaris-react/pull/2186#issuecomment-547455558

Added to a sub-branch of the branch [this PR](https://github.com/Shopify/polaris-react/pull/2186) is on.

This PR removes some rounding from the function used to convert `rgb` colours to `hsb`/`hsl`.

This is done to prevent an inaccuracy in the round trip conversion of `rgb`->`hsb`->`rgb`

cc @tleunen @apman


ruairiphackett:
@apman I tested this out on the value you mentioned in [your comment](https://github.com/Shopify/polaris-react/pull/2186#issuecomment-547455558) and got the same hex value `#ac4056` returned after the `rgb`->`hsb`->`rgb` roundtrip:

ruairiphackett:
Here's the conversion using the rounded values (using [this tool](https://www.rapidtables.com/convert/color/hsv-to-rgb.html)):

apman:
🎩looks great! (using this [Playground.txt](https://github.com/Shopify/polaris-react/files/3826875/Playground.txt) to check lots of values)

ruairiphackett:
Nice, thanks for the playground code @apman !

PR #2406, opened 2019-11-06
[Color system] write documentation

danrosenthal:


tmlayton:
Good first pass, we can adjust based on feedback

PR #2410, opened 2019-11-07
Update to latest chromatic package

BPScott:

Keeping up to date, making it easier to do new version bumps in the future now that we use their new package

Move from storybook-chroma to storybook-chromatic as the package is
being unified

### How to 🎩

See that the chroma GH check runs successfully

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,7 @@ jobs:
           paths:
             - node_modules
       - run: yarn run test:percy
-  chromaui:
+  chroma:
     docker:
       - image: circleci/node:10.13.0-browsers
     steps:
@@ -46,7 +46,7 @@ jobs:
           key: v3-npm-deps-{{ arch }}-{{ checksum "yarn.lock" }}-{{ checksum ".yarnclean" }}
           paths:
             - node_modules
-      - run: npx chroma publish --build-script-name='storybook:build'
+      - run: yarn run chromatic publish --build-script-name='storybook:build'
   a11y:
     docker:
       - image: circleci/node:10.13.0-browsers
@@ -69,5 +69,5 @@ workflows:
     jobs:
       - check
       - percy
-      - chromaui
+      - chroma
       - a11y
--- a/yarn.lock
+++ b/yarn.lock
@@ -974,10 +974,10 @@
   dependencies:
     regenerator-runtime "^0.13.2"
 
-"@babel/runtime@^7.0.0", "@babel/runtime@^7.1.2", "@babel/runtime@^7.1.6", "@babel/runtime@^7.3.1", "@babel/runtime@^7.4.2", "@babel/runtime@^7.4.5", "@babel/runtime@^7.5.0", "@babel/runtime@^7.5.5":
-  version "7.6.0"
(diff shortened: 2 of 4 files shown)


.circleci/config.yml
```diff
@@ -46,7 +46,7 @@ jobs:
           key: v3-npm-deps-{{ arch }}-{{ checksum "yarn.lock" }}-{{ checksum ".yarnclean" }}
           paths:
             - node_modules
-      - run: npx chroma publish --build-script-name='storybook:build'
+      - run: npx chromatic publish --build-script-name='storybook:build'
```
kaelig: Are we supposed to run this using npx? I thought this method was only recommended for projects that don't have the package listed in their dependencies?

yarn.lock
```diff
@@ -5580,13 +5599,30 @@ cssom@0.3.x, "cssom@>= 0.3.2 < 0.4.0":
   resolved "https://registry.yarnpkg.com/cssom/-/cssom-0.3.2.tgz#b8036170c79f07a90ff2f16e22284027a243848b"
   integrity sha1-uANhcMefB6kP8vFuIihAJ6JDhIs=
 
+cssom@^0.4.1:
+  version "0.4.1"
+  resolved "https://registry.yarnpkg.com/cssom/-/cssom-0.4.1.tgz#b24111d236b6dbd00cdfacb5ab67a20473381fe3"
+  integrity sha512-6Aajq0XmukE7HdXUU6IoSWuH1H6gH9z6qmagsstTiN7cW2FNTsb+J2Chs+ufPgZCsV/yo8oaEudQLrb9dGxSVQ==
+
+cssom@~0.3.6:
```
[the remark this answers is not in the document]
BPScott: yay transitive dependencies :p

.circleci/config.yml
BPScott: npx will check the local path and installed modules before downloading it, (as `npm run` only looks at scripts, it doesn't pick up binarys like `yarn run`).

I've changed this to `yarn run chromatic...` so we're consistently using yarn

.circleci/config.yml
```diff
@@ -32,7 +32,7 @@ jobs:
           paths:
             - node_modules
       - run: yarn run test:percy
-  chromaui:
+  chroma:
```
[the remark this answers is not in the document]
BPScott: Yeah it's a bit all over the place. Chroma is the company, they make Chromatic (the VRT tool) and the early-access thing ChromaUI (the hosted storybook tool, which just got VRT stuff too), they've decided that two products is silly so once they're done with the early access stuff the "ChromaUI" brand will go away and there will be one product "Chromatic 2.0" which does both storybook hosting and VRT - this explains why this PR changes from storybook-chroma to storybook-chromatic as they've already unified their package so they have one package that handles both products while they're split and eventually storybook-chroma shall go away.

I named this chroma so it matches the name of the check that they add that links to their site, which is "ci/chroma — 264 stories published."

.circleci/config.yml
BPScott: This was a copy+paste from their original docs. npx will check the local path and installed modules before downloading it, (as `npm run` only looks at scripts, it doesn't pick up binarys like `yarn run`) it's not just for if you've not got something in your deps.

I've changed this to `yarn run chromatic...` which is functionally the same as we install their package except now we're consistently using yarn

kaelig:
:shipit:

PR #2413, opened 2019-11-08
[ColorPicker] Add keyboard controls to Slidable

ruairiphackett:

Fixes https://github.com/Shopify/polaris-ux/issues/277

There is currently no way to set a colour value in the `ColorPicker` without the use of mouse/touch controls.

This PR makes the slider on each of the `ColorPicker` `Slidable` elements focusable and allows interaction using the arrow and shift keys.

**Interaction after changes:**

### Still to do

- [ ] Need to decide on what [ARIA role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques) to use for the 2 dimensional main colour picker panel. The `slider` role doesn't work as it is intended for a slider that works on a single axis.
- [ ] If sticking with `slider` role for 1 axis sliders (`HuePicker` & `AlphaPicker`) I need to fix the value for `aria-valuenow`, currently just has a placeholder value of `0`.

## 

### How to 🎩

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -5,6 +5,7 @@
 ### Enhancements
 
 - Added `hideTags` prop to `Filters` ([#2573](https://github.com/Shopify/polaris-react/pull/2573))
+- Added keyboard control to ColorPicker Slidable, HuePicker & AlphaPicker subcomponents ([#2413](https://github.com/Shopify/polaris-react/pull/2413))
 
 ### Bug fixes
 
--- a/src/components/ColorPicker/ColorPicker.tsx
+++ b/src/components/ColorPicker/ColorPicker.tsx
@@ -75,7 +75,11 @@ export class ColorPicker extends React.PureComponent<ColorPickerProps, State> {
         id={id}
         onMouseDown={this.handlePickerDrag}
       >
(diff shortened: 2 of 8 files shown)


ruairiphackett:
@dpersing, I'm using the `shift` key here to allow for a larger jump when combined with an arrow key (jumps 5 'steps' when `shift` is held).

I didn't use the advised `page up`/`page down` as I didn't know of any corresponding keys that could be used in a 2 dimensional slider for a large jump left/right, i.e. if `page up`/`page down`  were used as large steps up/down respectively, then I didn't know what buttons/combinations to use for a large step left/right.

I realise there might be issues with using `shift` however, so I was wondering if you have any advice on this?

dpersing:
Hi @ruairiphackett! 

For one of your to do items ("what ARIA role to use") I'm thinking this would be a good opportunity for `role="application"` on the two-dimensional slider, since it's a generic catchall for things that aren't native controls or document content.

The question about using Shift is an excellent one. My gut says go ahead with it, but I'm going to check on some other work for other multi-directional/dimensional interactions to see if we can have a common pattern. 👍

ruairiphackett:
Thanks @dpersing !

PR #2414, opened 2019-11-08
[Color system] Apply color system to TextStyle

dleroux:
addresses: https://github.com/Shopify/polaris-ux/issues/366

Screen Shot 2019-11-08 at 8 55 37 AM
Screen Shot 2019-11-08 at 8 55 22 AM

- Not sure if this is the right approach but I updated the shadowFromDirectLight since it's should look like a border in dark mode.

- Not the success green. It isn't very visible in light mode and really bright in dark mode. Is that correct?
 
## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page, TextStyle, Stack} from '../src';

  return (
    <Page title="Playground">
      <Stack>
        <TextStyle variation="negative">Negative</TextStyle>
        <TextStyle variation="positive">Positive</TextStyle>
        <TextStyle variation="strong">Strong</TextStyle>
        <TextStyle>Regular</TextStyle>
        <TextStyle variation="subdued">Subdued</TextStyle>
        <TextStyle variation="code">Code</TextStyle>
      </Stack>
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/TextStyle/TextStyle.scss
+++ b/src/components/TextStyle/TextStyle.scss
@@ -4,21 +4,21 @@ $code-padding: 0 spacing(extra-tight);
 $code-font-size: 1.15em;
 
 .variationPositive {
-  color: color('green', 'dark');
+  color: var(--p-success-text, color('green', 'dark'));
 }
 
 .variationNegative {
-  color: color('red', 'dark');
+  color: var(--p-critical-text, color('red', 'dark'));
 }
 
 .variationCode {
   position: relative;
   padding: $code-padding;
   border-radius: border-radius();
-  background-color: color('sky', 'light');
+  background-color: var(--p-surface-background, color('sky', 'light'));
   display: inline-block;
   font-size: $code-font-size;
-  box-shadow: inset 0 0 0 1px color('sky');
+  box-shadow: inset 0 0 0 1px var(--p-shadow-from-direct-light, color('sky'));
 
   &::after {
     content: '';
--- a/src/components/TextStyle/TextStyle.scss
+++ b/src/components/TextStyle/TextStyle.scss
@@ -15,10 +15,10 @@ $code-font-size: 1.15em;
   position: relative;
   padding: $code-padding;
   border-radius: border-radius();
-  background-color: var(--p-surface-background, color('sky', 'light'));
+  background-color: var(--p-surface-foreground-subdued, color('sky', 'light'));
   display: inline-block;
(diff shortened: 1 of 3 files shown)


src/components/TextStyle/TextStyle.scss
```diff
@@ -4,21 +4,21 @@ $code-padding: 0 spacing(extra-tight);
...
-  color: color('red', 'dark');
+  color: var(--p-critical-text, color('red', 'dark'));
 }
 
 .variationCode {
   position: relative;
   padding: $code-padding;
   border-radius: border-radius();
-  background-color: color('sky', 'light');
+  background-color: var(--p-surface-background, color('sky', 'light'));
```
dleroux: Would this need to become `--p-surface-background` or `--p-surface-foreground` depending on where it's used?

tmlayton: I think it should be `--p-surface-foreground-subdued`

tmlayton:
> fixes: Shopify/polaris-ux#366

If you don’t mind changing that to `addresses` as this is only half of the work, still need to do Rails too.

src/components/TextStyle/TextStyle.scss
```diff
@@ -4,21 +4,21 @@ $code-padding: 0 spacing(extra-tight);
...
   display: inline-block;
   font-size: $code-font-size;
-  box-shadow: inset 0 0 0 1px color('sky');
+  box-shadow: inset 0 0 0 1px var(--p-shadow-from-direct-light, color('sky'));
```
tmlayton: I think since this is supposed to emulate a light border it should be `--p-border-subdued-on-surface`

tmlayton:
> Not the success green. It isn't very visible in light mode and really bright in dark mode. Is that correct?

I think for both success and critical, the values should be:

```ts
light: {lightness: 40},
dark: {lightness: 60},
```

This will still be AA even on the most muted surfaces (lightness of 90/10 respectively for light/dark). AAA = a lightness difference of ~62 or more and AA = a lightness difference of ~50 or more, regardless the hue or saturation.

/cc @jessebc 

Screen Shot 2019-11-08 at 10 06 36 PMScreen Shot 2019-11-08 at 10 06 42 PM

jessebc:
Will these text colours propagate through all of the components now?

dleroux:
> Will these text colours propagate through all of the components now?

Yes, I believe they would.

PR #2420, opened 2019-11-11
[Color Sytem] Apply color system to Breadcrumbs

dleroux:
addresses: https://github.com/Shopify/polaris-ux/issues/334

Apply color system to Breacrumbs

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Breadcrumbs/Breadcrumbs.scss
+++ b/src/components/Breadcrumbs/Breadcrumbs.scss
@@ -7,11 +7,6 @@ $icon-size: rem(20px);
 $difference-between-touch-area-and-backdrop: control-height() -
   control-slim-height();
 
-$stacking-order: (
-  focus-backdrop: 10,
-  content: 20,
-);
-
 .Breadcrumb {
   @include text-style-body;
   @include text-emphasis-subdued;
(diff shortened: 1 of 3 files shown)


src/components/Breadcrumbs/Breadcrumbs.scss
```diff
@@ -30,16 +30,16 @@ $stacking-order: (
...
 
   &:hover,
   &:active {
-    color: color('ink');
+    color: var(--p-text-on-surface, color('ink'));
     text-decoration: none;
 
     .Icon {
-      @include recolor-icon(color('ink'));
+      @include recolor-icon(var(--p-icon-on-surface, color('ink')));
```
dleroux: I've used `--p-icon` but this causes an issue in dark mode where the icon doesn't match the text color but it does on light mode. What's the approach to fix this inconsistent:

1. change `iconSubduedOnSurface` to match`textSubduedOnSurface` in `color-adjustments.ts`?
2. or just use the `text[Subdued]OnSurface` for both the icon and the text?

src/styles/shared/_buttons.scss
```diff
@@ -288,7 +288,7 @@
 }
 
 @function plain-button-background() {
-  @return rgba(color('ink'), 0.1);
+  @return var(--p-surface-foreground-subdued, rgba(color('ink'), 0.1));
```
dleroux: This is where my [alpha](https://github.com/Shopify/polaris-ux/issues/385) question came from.

dleroux: That said, should this become the focus rectangle since these are all applied on focus to plain buttons?

src/components/Breadcrumbs/Breadcrumbs.scss
tmlayton: This is fine, we can adjust the colors later.

src/components/Breadcrumbs/Breadcrumbs.scss
```diff
@@ -73,9 +73,12 @@ $stacking-order: (
 }
 
 .Icon {
-  @include recolor-icon(color('ink', 'lighter'));
-  height: $icon-size;
+  @include recolor-icon(
+    var(--p-icon-subdued-on-surface, color('ink', 'lighter'))
+  );
+  z-index: z-index(content, $stacking-order);
```
tmlayton: Why is this needed?

src/styles/shared/_buttons.scss
tmlayton: A couple things here: First to make this match monochromatic links it would be:

```suggestion
  @return var(--p-neutral-action, rgba(color('ink'), 0.1));
```

Second this style should only get applied to the active state, and focus should only get the blue focus ring.

tmlayton:
This component is similar to link, which is more of a medium+ difficulty as it requires turning on and off pseudo elements depending on the theme.

src/components/Breadcrumbs/Breadcrumbs.scss
dleroux: Because before we were using transparency, and the icon was seen through it. The stacking order was applied to the content, but was missed on the icon and wasn't visible.

tmlayton: To keep the existing styling unchanged:

```suggestion
  z-index: var(--p-z-index-content, initial);
```

tmlayton: Actually never mind, let’s drop this change for now.

src/styles/shared/_buttons.scss
tmlayton: Let’s drop this change for now

src/components/Breadcrumbs/Breadcrumbs.scss
```diff
@@ -21,65 +16,68 @@ $stacking-order: (
...
     .Icon {
-      @include recolor-icon(color('ink'));
+      @include recolor-icon(var(--p-icon-on-surface, color('ink')));
     }
   }
-
+  // stylelint-disable selector-max-specificity
   &:focus {
     outline: none;
+    .ContentWrapper {
```
dleroux: Adding this content wrapper to attach the focus ring to.

src/components/Breadcrumbs/Breadcrumbs.scss
```diff
@@ -21,65 +16,68 @@ $stacking-order: (
...
+    var(--p-icon-subdued-on-surface, color('ink', 'lighter'))
+  );
   width: $icon-size;
-
-  // We set negative top/ bottom margin so the icon never sets
-  // the height of the content. We set negative left margin to
-  // compensate for the inset viewbox and to match the Page
-  // component’s secondary actions
+  height: $icon-size;
   margin: (-0.5 * $icon-size) 0 (-0.5 * $icon-size) rem(-8px);
```
dleroux: This does not look like its needed, probably because I'm using flexbox.

src/utilities/theme/color-adjustments.ts
```diff
@@ -175,7 +175,7 @@ export const colorAdjustments: ColorAdjustments = {
   iconSubduedOnSurface: {
     baseColor: 'onSurface',
     light: {lightness: 59.8},
-    dark: {lightness: 88},
+    dark: {lightness: 62},
```
dleroux: We'll need to review this, for now, I made it the same as the text in dark mode.

PR #2433, opened 2019-11-14
Export color factory function

martenbjork:
Other programs should be able to use the color factory.

It adds the `buildColors` to the list of exports.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -12,5 +12,5 @@ export {
   WithinContentContext as _SECRET_INTERNAL_WITHIN_CONTENT_CONTEXT,
 } from './utilities/within-content-context';
 
-export {buildColors as UNSTABLE_BuildColors} from './utilities/theme/utils';
+export {buildColors as UNSTABLE_BuildColors} from './utilities/theme';
 export {AppBridgeContext} from './utilities/app-bridge';
--- a/src/index.ts
+++ b/src/index.ts
@@ -5,8 +5,12 @@ export * from './types';
 
 export * from './components';
 
-export {ScrollLockManagerContext as _SECRET_INTERNAL_SCROLL_LOCK_MANAGER_CONTEXT} from './utilities/scroll-lock-manager';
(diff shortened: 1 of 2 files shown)


martenbjork:
Smallest PR ever. For now, this allows an external program to import the color factory and consume the CSS Custom Properties.

The color factory and the tests are currently built with in CSS Custom Properties in mind. In the future, we may want to separate the factory function from the CSS transformations. That way, you could run the factory and then, separately transform the output to an object, CSS Properties or Tokens — whichever suits your use case.

kaelig:
If this is something we'll host as part of polaris-tokens later on, would it make sense to make it clear this is an "unstable" feature?

martenbjork:
> If this is something we'll host as part of polaris-tokens later on, would it make sense to make it clear this is an "unstable" feature?

Absolutely! Updated. (Using the right conventions?)

src/index.ts
```diff
@@ -11,4 +11,6 @@ export {
 export {
   WithinContentContext as _SECRET_INTERNAL_WITHIN_CONTENT_CONTEXT,
 } from './utilities/within-content-context';
+
+export {buildColors as unstableBuildColors} from './utilities/theme/utils';
```
tmlayton: let's make it _really_ clear :) 

```suggestion
export {buildColors as UNSTABLE_BuildColors} from './utilities/theme/utils';
```

src/index.ts
```diff
@@ -11,4 +11,6 @@ export {
 export {
   WithinContentContext as _SECRET_INTERNAL_WITHIN_CONTENT_CONTEXT,
 } from './utilities/within-content-context';
+
+export {buildColors as UNSTABLE_BuildColors} from './utilities/theme/utils';
```
tmlayton:
```suggestion
export {buildColors as UNSTABLE_BuildColors} from './utilities/theme';
```

and add an export from `theme/index.ts`, this allows us to quickly look at the local index file to have an idea of what we expose publicly

PR #2434, opened 2019-11-14
[Tooltip] Add an example for a light tooltip

LauraAubin:
## Issue summary

I was confusing instances of light colored tooltips with a Popover since it wasn't immediately obvious to me that this prop existed. This PR surfaces that prop via an example in the style guide.

I don't mind if we close this PR if we feel we don't need this addition.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -38,7 +38,7 @@
 - Updated the `PageActions` example ([#2471](https://github.com/Shopify/polaris-react/pull/2471))
 - Fixed spacing of the `Filters` data table example ([#2477](https://github.com/Shopify/polaris-react/pull/2477))
 - Fixed duplicate and unclear prop descriptions of `EmptyState` ([#2477](https://github.com/Shopify/polaris-react/pull/2477))
-- Added an example for a light Tooltip ((#2434)[https://github.com/Shopify/polaris-react/pull/2434])
+- Added an example for a light `Tooltip` ([#2434](https://github.com/Shopify/polaris-react/pull/2434))
 
 ### Development workflow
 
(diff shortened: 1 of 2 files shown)


LauraAubin:
We might not want this after all. Pending discussion to remove `light` prop entirely.

dleroux:
I'm guessing this one won't be needed?

LauraAubin:
> I'm guessing this one won't be needed?

I closed the PR that would remove the `light` prop with [this reason](https://github.com/Shopify/polaris-react/pull/2448#issuecomment-557268520) meaning that this example can still be added.

src/components/Tooltip/README.md
```diff
@@ -74,6 +74,18 @@ Use only when necessary to provide an explanation for an interface element.
 </div>
 ```
 
+### Light tooltip
+
+Changes the background color to white.
```
kaelig: Could this include a usage guideline, rather than being a description of what this example does?

LauraAubin: For sure! I chatted with @sarahill and @sadiesaurus and we came up with,

> Light tooltip draws less attention than the default tooltip. Use only when necessary to provide an explanation for an interface element with less emphasis.

UNRELEASED.md
```diff
@@ -38,6 +38,7 @@
 - Updated the `PageActions` example ([#2471](https://github.com/Shopify/polaris-react/pull/2471))
 - Fixed spacing of the `Filters` data table example ([#2477](https://github.com/Shopify/polaris-react/pull/2477))
 - Fixed duplicate and unclear prop descriptions of `EmptyState` ([#2477](https://github.com/Shopify/polaris-react/pull/2477))
+- Added an example for a light Tooltip ((#2434)[https://github.com/Shopify/polaris-react/pull/2434])
```
```suggestion
- Added an example for a light `Tooltip` ([#2434](https://github.com/Shopify/polaris-react/pull/2434))
```

PR #2435, opened 2019-11-14
[Tooltip] Implement color system

LauraAubin:

Part of https://github.com/Shopify/polaris-ux/issues/368

Mockups [here](https://www.figma.com/file/4dAAt5iFPSaxUKiYVKrkYj/DL%E2%80%93Polaris-(Web)%3A-Components?node-id=2098%3A94).

This introduces the color system to the tooltip.

|-| No theme        | Light mode         | Dark mode  |
|:-:| ------------- |:-------------:| -----|
|Regular tooltip|      image  |      image  |   image  |
|High contrast:|image|image|image|

### How is this accomplished?

Since we will want to deprecate the `light` prop with the rollout of the design language, we don't need to colorize it here. We will also be "inversing" the colors so that the tooltip appears dark on a dark background, and likewise.

### DIFF
--- a/src/components/Tooltip/Tooltip.scss
+++ b/src/components/Tooltip/Tooltip.scss
@@ -7,8 +7,8 @@ $content-max-width: rem(200px);
 .Tooltip {
   margin: $offset-from-activator spacing() spacing();
   opacity: 1;
-  box-shadow: shadow(deep);
-  border-radius: border-radius();
+  box-shadow: var(--p-popover-shadow, shadow(deep));
+  border-radius: var(--p-border-radius-base, border-radius());
   pointer-events: none;
   will-change: opacity, left, top;
   transition: opacity duration() easing(in) duration(fast);
@@ -28,17 +28,17 @@ $content-max-width: rem(200px);
 
 .light {
   .Wrapper {
-    background: color('white');
-    color: color('ink');
+    background: var(--p-surface, color('white'));
+    color: var(--p-text, color('ink'));
   }
 }
 
 .Wrapper {
   position: relative;
   display: flex;
-  background-color: color('ink');
-  border-radius: border-radius();
-  color: color('white');
+  background-color: var(--p-surface, color('ink'));
+  border-radius: var(--p-border-radius-base, border-radius());
+  color: var(--p-text, color('white'));
   max-height: $content-max-height;
 }
 


src/components/Tooltip/Tooltip.scss
```diff
@@ -28,17 +28,19 @@ $content-max-width: rem(200px);
 
 .light {
   .Wrapper {
-    background: color('white');
+    background: var(--p-surface-light-foreground, color('white'));
+    // border: border-width() solid var(--p-border-subdued-on-surface, transparent);
```
LauraAubin: This is similar to the `border` function:

https://github.com/Shopify/polaris-react/blob/01be565bcc4d8c7babdfc8e9fbfac9842fd9536c/src/styles/foundation/_borders.scss#L2

I didn't use the function since it we don't want a border if we are not using the color system. The only difference is that the function falls back on `white`, whereas this uses `transparent`.

LauraAubin: (Using this approach instead of `box-shadow: shadow(deep)` which visually adds a border + a shadow)

src/utilities/theme/tokens.ts
```diff
@@ -10,6 +10,8 @@ const Shadow = {
     '-1px 0px 20px var(--p-shadow-from-ambient-light), 0px 1px 5px var(--p-shadow-from-direct-light)',
   modalShadow:
     '0px 6px 32px var(--p-shadow-from-ambient-light), 0px 1px 6px var(--p-shadow-from-direct-light)',
+  tooltipShadow:
+    '-1px 0px 20px var(--p-shadow-from-ambient-light), 0px 1px 5px var(--p-shadow-from-direct-light)',
```
LauraAubin: Note that this is the same value as `popoverShadow`.

tmlayton: This is fine for now, but let’s think of a better naming convention for this.

LauraAubin:
Ouuu I really like that idea. What about layers of depth like, `onSurface, a_better_name_for_above_surface, foreground`.

Though in isolation that may be confusing without the understanding that "larger shadows are for items further from the surface".

src/utilities/theme/tokens.ts
danrosenthal: I'm not crazy about seeing another token declared from this that is identical to an existing token. Can we just re-use the existing one with the aim of following up later with a better naming convention?

LauraAubin: For sure! Made a note to follow up on the shadow variables in this issue.

LauraAubin:
|-| No theme        | Light mode         | Dark mode  |
|:-:| ------------- |:-------------:| -----|
|Light tooltip|image| image|image |

PR #2436, opened 2019-11-15
Union merge conflicting changes (accept both) in UNRELEASED.md

tmlayton:
This will help reduce the annoyance of conflicts in UNRELEASED.md by defaulting to accept both changes. 

### The catch

After releases, out-of-date branches can potentially re-add entries in UNRELEASED that have already been moved to the CHANGELOG without showing a conflict since this defaults to accepting both changes.

### Workaround

For now, I added a test which fails when entries from UNRELEASED.md are found in CHANGELOG.md. This won't prevent an outdated branch from merging after a release thus re-adding entries in UNRELEASED.md, but it will cause updated branches to fail so the issue gets caught downstream.

### Other options

- Use a [custom driver](http://manpages.ubuntu.com/manpages/bionic/man1/git-merge-changelog.1.html) which is designed just for changelogs. This seems to be a popular choice and is smart about proper ordering and deletions. Although, I know very little about this and would require adding our own script to this project.
- Add a bunch of [whitespace to UNRELEASED.md and put entries on random lines](https://about.gitlab.com/blog/2015/02/10/gitlab-reduced-merge-conflicts-by-90-percent-with-changelog-placeholders/)
- Create an [unreleased folder and each entry becomes its own [markdown] file](https://about.gitlab.com/blog/2018/07/03/solving-gitlabs-changelog-conflict-crisis/)


tmlayton:
> The failing diff output in Jest

Yeah it’s not great. Maybe there is a small lib we can use that spits out better diffs or write a custom matcher. Another quick hack would be to only compare the top `n%` of the changelog since really the failures will come from unreleased entries duplicated in the most recent release.

kaelig:
I love this, thank you ❤️

PR #2441, opened 2019-11-18
[ButtonGroup] Add theme to ButtonGroup and move CSS to Button

dleroux:

addresses part of: https://github.com/Shopify/polaris-ux/issues/382

and also 

fixes: https://github.com/Shopify/polaris-react/issues/1240

1. Keep layout CSS as part of `ButtonGroup` (essentially margins).
2. Adds data attributes to ButtonGroup
3. Moves necessary styles to Button
4. adds the theme to ButtonGroup

### How to 🎩

<details>

```jsx
import {Page, ButtonGroup, Button, Popover, ActionList} from '../src';
  return (
    <Page title="Playground">
      <ButtonGroup segmented>
        <Button>Bold</Button>
        <Button>Italic</Button>
        <Button>Underline</Button>
        <PopoverWithActionListExample />
      </ButtonGroup>
    </Page>
  );
}
function PopoverWithActionListExample() {
    [],
  );
  const activator = (
      More actions
    </Button>
  );
  return (
    <Popover
      active={popoverActive}
      activator={activator}
      onClose={togglePopoverActive}
    >
      <Button>More actions</Button>
    </Popover>
  );
}

```

</details>

<details>
<summary>Copy-paste this code for the split action button:</summary>

```jsx
import {CaretDownMinor} from '@shopify/polaris-icons';
import {Page, ButtonGroup, Button, Popover, ActionList} from '../src';

  return (
    <Page title="Playground">
      <ButtonGroup segmented>
        <Button primary>Underline</Button>
        <PopoverWithActionListExample />
      </ButtonGroup>
    </Page>
  );
}

function PopoverWithActionListExample() {
    [],
  );
  const activator = (
    <Button
      pressed={popoverActive}
      primary
      icon={CaretDownMinor}
      onClick={togglePopoverActive}
    />
  );
  return (
    <Popover
      active={popoverActive}
      activator={activator}
      onClose={togglePopoverActive}
    >
      <Button>Im still rounded because Portal overwrites my context</Button>
    </Popover>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -53,13 +53,13 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
     transition-duration: duration(fast);
     background: linear-gradient(to bottom, $color-light, $color-light);
     border-color: $color-dark;
-    box-shadow: $partial-button-filled-pressed-box-shadow$color-dark;
+    box-shadow: $partial-button-filled-pressed-box-shadow $color-dark;
   }
 
   &:active {
     background: linear-gradient(to bottom, $color-lightest, $color-lightest);
     border-color: $color-dark;
-    box-shadow: $partial-button-filled-pressed-box-shadow$color-darkest;
+    box-shadow: $partial-button-filled-pressed-box-shadow $color-darkest;
   }
 }
 
@@ -70,7 +70,37 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
     @include base-button-disabled;
   }
 }
+// stylelint-disable selector-max-combinators
+[data-buttongroup-segmented] {
+  .Button {
+    border-radius: 0;
+  }
+  > :first-child .Button {
+    border-top-left-radius: var(--p-border-radius-base, border-radius());
+    border-bottom-left-radius: var(--p-border-radius-base, border-radius());
+  }
+  > :last-child .Button {
+    border-top-right-radius: var(--p-border-radius-base, border-radius());
+    border-bottom-right-radius: var(--p-border-radius-base, border-radius());
+  }
+}
 
+[data-buttongroup-connected-top] {
+  > :first-child .Button {
+    border-top-left-radius: 0;
+  }
+
+  > :last-child .Button {
+    border-top-right-radius: 0;
+  }
+}
+
+[data-buttongroup-full-width] {
+  .Button {
+    @include button-full-width;
+  }
+}
+// stylelint-enable selector-max-combinators
 .Content {
   @include text-style-button;
   position: relative;
@@ -132,6 +162,10 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
   --p-button-color-active: var(--p-branded-action-pressed);
   @include button-filled(color('indigo'), color('indigo', 'dark'));
   @include recolor-icon(color('white'));
+
+  &.globalTheming {
+    @include recolor-icon(var(--p-icon-on-branded));
+  }
 
   &.disabled {
     @include button-filled-disabled(color('indigo'));
@@ -493,6 +527,10 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
 
   .Icon:last-child {
     margin-right: rem(-4px);
+  }
+
+  .Icon + .Icon:last-child {
+    margin-left: 0;
   }
 
   .Icon:only-child {
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -528,10 +528,11 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
   .Icon:last-child {
     margin-right: rem(-4px);
   }
-
+  // stylelint-disable selector-max-class, selector-max-combinators, selector-max-specificity
   .Icon + .Icon:last-child {
     margin-left: 0;
   }
+  // stylelint-enable selector-max-class, selector-max-combinators, selector-max-specificity
 
   .Icon:only-child {
     margin-right: 0;
--- a/src/components/ButtonGroup/ButtonGroup.scss
+++ b/src/components/ButtonGroup/ButtonGroup.scss
@@ -12,18 +12,20 @@ $item-spacing: spacing(tight);
   display: flex;
   flex-wrap: wrap;
   align-items: center;
-  margin-top: (-1 * $item-spacing);
-  margin-left: (-1 * $item-spacing);
+  margin-top: calc(#{var(--p-button-group-item-spacing, $item-spacing)} * -1);
+  margin-left: calc(#{var(--p-button-group-item-spacing, $item-spacing)} * -1);
 }
 
 .Item {
-  margin-top: $item-spacing;
-  margin-left: $item-spacing;
+  margin-top: var(--p-button-group-item-spacing, $item-spacing);
+  margin-left: var(--p-button-group-item-spacing, $item-spacing);
 }
 
 .Item-plain {
   &:not(:first-child) {
-    margin-left: $item-spacing + $plain-horizontal-spacing;
+    margin-left: calc(
+      #{var(--p-button-group-item-spacing, $item-spacing)} + #{$plain-horizontal-spacing}
+    );
   }
 
   &:not(:last-child) {
@@ -37,8 +39,6 @@ $item-spacing: spacing(tight);
   margin-top: 0;
   margin-left: 0;
 
-  // This is a violation of our component model, but it’s the cleanest
-  // way to remove the border radii on connected elements.
   .Item {
     position: relative;
     z-index: z-index(item, $stacking-order);
@@ -46,24 +46,7 @@ $item-spacing: spacing(tight);
     margin-left: 0;
 
     &:not(:first-child) {
-      margin-left: -(border-width());
-    }
-
-    // stylelint-disable-next-line selector-max-combinators
-    > * {
-      border-radius: 0;
-    }
-
-    // stylelint-disable-next-line selector-max-combinators
-    &:first-child > * {
-      border-top-left-radius: border-radius();
-      border-bottom-left-radius: border-radius();
-    }
-
-    // stylelint-disable-next-line selector-max-combinators
-    &:last-child > * {
-      border-top-right-radius: border-radius();
-      border-bottom-right-radius: border-radius();
+      margin-left: var(--p-button-group-item-spacing, -1 * border-width());
     }
   }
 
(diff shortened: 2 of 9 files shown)


src/components/Button/Button.scss
```diff
@@ -53,13 +53,13 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
     transition-duration: duration(fast);
     background: linear-gradient(to bottom, $color-light, $color-light);
     border-color: $color-dark;
-    box-shadow: $partial-button-filled-pressed-box-shadow$color-dark;
```
dleroux: Not part of the changed but noticed this while editing the file.

src/components/Button/Button.scss
```diff
@@ -71,6 +71,38 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
   }
 }
 
+.globalTheming.inSegmentedGroup,
```
dleroux: Once we remove the .globalTheming these can be safely removed as well. Specificity issue.

src/components/Button/Button.tsx
```diff
@@ -135,6 +139,27 @@ export function Button({
 
   const i18n = useI18n();
 
+  const buttonGroupContext = useButtonGroupContext();
```
dleroux: Styling logic of the button itself now moved to Button.

src/components/ButtonGroup/components/Item/Item.tsx
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useState} from 'react';
```
dleroux: took the opportunity to convert Item to a functional component.

src/components/Portal/Portal.tsx
```diff
@@ -67,8 +68,14 @@ export class Portal extends React.PureComponent<PortalProps, State> {
   }
 
   render() {
+    const childrenMarkup = (
+      <ButtonGroupContext.Provider value={undefined}>
```
dleroux: @BPScott I was thinking this could be a valid option. I debated adding this to popover because additional buttons would always be inside a popover, however, new overlays could come at play one day and this will guard against these as well.

src/components/Button/Button.tsx
```diff
@@ -124,6 +126,7 @@ export function Button({
 }: ButtonProps) {
   const {unstableGlobalTheming = false} = useFeatures();
   const hasGivenDeprecationWarning = useRef(false);
+  const id = useUniqueId('Button', idProp);
```
BPScott: Having to add an ID to every button feels rather heavy handed, I wonder if there is a way we can avoid this (or if not, can we make the id needed by the context be distinct from the id attribute that gets added to the DOM).

Perhaps instead of having the Buttons be responsible for adding themselves to the context, instead ButtonGroup could trawl through its children to collect the Button elements and adjust their props?

src/components/Portal/Portal.tsx
BPScott: Nice trick! I'll forget all about it in a few weeks so it'd be worth adding a comment about why this is here

src/components/ButtonGroup/components/Item/Item.tsx
```diff
@@ -9,39 +9,18 @@ export interface ItemProps {
...
+  const [focused, setFocused] = useToggle(false);
+
+  const className = classNames(
+    styles.Item,
+    focused && styles['Item-focused'],
+    button.props.plain && styles['Item-plain'],
+  );
+
+  return (
+    <div className={className} onFocus={setFocused} onBlur={setFocused}>
```
BPScott: Will this potentially cause oddness if you try and focus something that is already focussed, or blur something that is already blured? (e.g. if it's already focused causing it to focus again will cause it to toggle to the unfocused state instead of remaining focused)

`useForcableToggle()` and its options for "always set to true" and "always set to false" would be better here.

src/components/Button/Button.tsx
dleroux: What do you think of just using the [Id internally](https://github.com/Shopify/polaris-react/pull/2441/files#diff-80ce11d837170156710fe9e24d078a78R129)? 

Having ButtonGroup look at its children and add props I think could be done by recursively looping through the children and adding a prop to the Button by using cloneElement. Is that what you have in mind? Doing that would limit our implementation to Buttons. Having Buttons register themselves means this can be added to any component using the registration hook. For instance, the first button in BulkAction isn't a Button per se, but with this approach, we can still make use to the group.

dleroux:
@chloerice I provide the playground above to confirm that the split button can be done:

buttonGroup

src/components/Button/Button.tsx
BPScott: I'd be on board with using the id internally, I think that pollutes the DOM a little less.

As an aside how many renders does this approach trigger? Do we pay a price for rendering the button group, then when rendering the button updating the context (does that cause another render)

> Having ButtonGroup look at its children and add props I think could be done by recursively looping through the children and adding a prop to the Button by using cloneElement. Is that what you have in mind? Doing that would limit our implementation to Buttons. Having Buttons register themselves means this can be added to any component using the registration hook.

Something like that yeah, though I'm not sure about its performance characteristics so I don't know if it's a good idea or not. We're changing the registration point from within the child element (e.g. Button) to the container (i.e. ButtonGroup) - we can make that say "look for Buttons and Popovers". We wouldn't be able to account for any component (e.g. something that exists in web) but being able to open that up to ANYTHING can be a child of a ButtonGroup feels a bit too open-ended to me at the moment (though I've not thought about this in depth)

dleroux: > As an aside how many renders does this approach trigger? Do we pay a price for rendering the button group, then when rendering the button updating the context (does that cause another render)

That is a bit of a concern that it renders every time a button renders but the render is negligible I would say because buttons are small and there won't be that many in a group.

dleroux:
Going with Ben's suggestion here and simply moving the selectors to Button using data-attributes.

Using context gets us where we want but the only real advantage is knowing which is the first and last button at the cost of additional renders.

This does allow us to use the data attributes to style whatever component we want to style with round corners. Thanks Ben!

dleroux:
Thank you for such a thorough 🎩@chloerice. That playground 😍 

> I also notice that the theming breaks, but I can't tell if it's because button theming is incomplete 

Are you referring to the icon on primary or was there something else? 

The wrapping does occur today, so this is not solved by this PR :(

I also started removing the [ToggleButton](https://github.com/Shopify/web/pull/21260) from `polaris-next`

yarn.lock
```diff
@@ -1479,12 +1479,12 @@
     "@shopify/ast-utilities" "^0.0.6"
     webpack-virtual-modules "^0.1.12"
 
-"@shopify/react-testing@^1.7.8":
-  version "1.7.8"
-  resolved "https://registry.yarnpkg.com/@shopify/react-testing/-/react-testing-1.7.8.tgz#64b79d8b1cc4791d8e77e3ad5c06dfb7380c88b7"
-  integrity sha512-P4CNcYPjX6OKMe4f6LfyTM0UWlQqinBSoFIWPrWRwAI0qnwKRYnP81WdUANpmeY01SxAIp/WwjKTgpg3sdTnoQ==
+"@shopify/react-testing@^1.8.0":
```
dleroux: yarn.lock wasn't committed with https://github.com/Shopify/polaris-react/pull/2465

chloerice:
> I also started removing the [ToggleButton](https://github.com/Shopify/web/pull/21260) from `polaris-next`

🎉

> > I also notice that the theming breaks, but I can't tell if it's because button theming is incomplete
> 
> Are you referring to the icon on primary or was there something else? 

Yes, the icon on primary, but also the negative margin on segmented button groups seems to disappear 🤔 

Screen Shot 2019-11-25 at 11 32 26 AM

dleroux:
I've fixed the icon color, but the gap is part of the new design.

src/components/ButtonGroup/ButtonGroup.scss
```diff
@@ -48,51 +46,24 @@ $item-spacing: spacing(tight);
...
-    flex: 1 1 auto;
 
-    // stylelint-disable-next-line selector-max-combinators
-    > * {
-      @include button-full-width;
+  &.globalTheming {
+    .Item {
+      // stylelint-disable-next-line selector-max-class, selector-max-specificity
+      &:not(:first-child) {
+        margin-left: spacing(extra-tight);
```
chloerice: Looks like we need to add a new variable at the top for this margin value like [`$segmented-spacing: rem(2px);`](https://www.figma.com/file/4dAAt5iFPSaxUKiYVKrkYj/DL%E2%80%93Polaris-(Web)%3A-Components?node-id=2330%3A262) since `spacing(extra-tight)` is `rem(4px)`.

chloerice:
> I've fixed the icon color, but the gap is part of the new design.

Ooohhh okay! Hmmm, I might need to add a `split` prop to `ButtonGroup` so folks can 🔥 that gap when implementing a split button.

src/components/ButtonGroup/ButtonGroup.scss
dleroux: Could have sworn this was 4px. I changed it now and was able to remove the .'globalTheming' class.

dleroux:
> > I've fixed the icon color, but the gap is part of the new design.
> 
> Ooohhh okay! Hmmm, I might need to add a `split` prop to `ButtonGroup` so folks can 🔥 that gap when implementing a split button.

Now with the smaller gap, I think it looks pretty good. Essentially now the only difference with segmented and non-segmented is that non-segmented keep all their round corners.

dleroux:
Screen Shot 2019-11-25 at 3 10 14 PM

src/components/Button/Button.scss
```diff
@@ -487,6 +521,10 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
     padding-right: spacing(base-tight);
   }
 
+  &.disclosure {
```
dleroux: Noticed while replacing web that when there's iconOnly and Disclosure the spacing is off. Can be seen in the toolbar. Screenshot coming.

dleroux:
You're right @tmlayton. I looked and this is used in card footers, and such. Removed the adjustments on the regular group. Once this is green I'll ship and get an update in a subsequent PR if need be, so I can continue with my next PR which depends on this one.

@jessebc how is this looking? I kept the same spacing as today on the `Basic ButtonGroup`

Screen Shot 2019-11-26 at 7 51 41 AM

clauderic:
🎉

danrosenthal:
This is looking awesome @dleroux, sorry I didn't get around to reviewing before you merged. What do we think about the focus state for these? Should the ring match the shape of the underlying element? I think so.

|focus|
|-|
|Screen Shot 2019-11-26 at 5 06 37 PM|

jessebc:
> @jessebc how is this looking? I kept the same spacing as today on the `Basic ButtonGroup`

I think this is looking very nice. My only question is around the split button. I feel like we should either do a 1px gap on that button, or potentially add a darker line as a visual break instead of the gap. 

>Should the ring match the shape of the underlying element? 
Yes! That'd be great.

dleroux:
@danrosenthal Yes it should match but @tmlayton mentioned we might want to hold off and review 
 the current solution once we've had a chance to implement it a few times.

@jessebc maybe instead of using ButtonGroup for this we might just have to stick with [@chloerice solution](https://github.com/Shopify/polaris-react/pull/2329) of adding a prop to Button. If we want to lessen the gap in this one use case we'd need to add a prop to buttonGroup for 1px or a 2px gap.

PR #2443, opened 2019-11-18
Update Prettier to 1.19.1

BPScott:

Keeping up to date.

https://prettier.io/blog/2019/11/09/1.19.0.html has the lowdown, but stuff of interest is:

- Support for new syntax added in TS 3.7
- Fixing comment moving weirdness in useEffect style functions (see https://prettier.io/blog/2019/11/09/1.19.0.html#fix-moving-comments-in-function-calls-like-useeffect-6270-by-sosukesuzuki)
- Nicer formatting of union types

Updating the package in yarn then running a `yarn run format` to fixup expected changes

Removing a few unneeded prettier-ignores as Prettier no longer does ugly looking things in those cases.

### How to 🎩

Linting passes with no additional changes required

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -25,6 +25,7 @@
 ### Dependency upgrades
 
 - Upgraded to TypeScript `v3.7.2` ([#2237](https://github.com/Shopify/polaris-react/pull/2237))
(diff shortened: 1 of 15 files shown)


amrocha:
Seems fine but looks like you didn't catch all the new format changes so CI failed

BPScott:
Aaand now with the results of running a format *after* a eslint cache clear (by deleting the build folder) so it picks up all files.

PR #2445, opened 2019-11-19
Use CSF format for rendering playground stories

BPScott:

This means we're no longer using the old storiesOf API and pushes the story config for playgrounds in the playground folder instead of the storybook config

Defines playground stories in the playground stories.tsx.

### How to 🎩

Check  playground stories still render as before

### DIFF
--- a/playground/stories.tsx
+++ b/playground/stories.tsx
@@ -1,0 +1,16 @@
+import {Playground} from './Playground';
+import {KitchenSink} from './KitchenSink';
+import {DetailsPage} from './DetailsPage';
+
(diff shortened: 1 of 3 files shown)


playground/stories.tsx
```diff
@@ -0,0 +1,16 @@
+import {Playground} from './Playground';
+import {KitchenSink} from './KitchenSink';
+import {DetailsPage} from './DetailsPage';
+
+// eslint-disable-next-line import/no-default-export, import/no-anonymous-default-export
```
BPScott: If we start using stories files more then we can ignore this in the eslint config for stories.tsx files.

PR #2447, opened 2019-11-19
[Form] Add tabIndex={-1} to implicit submit form

dleroux:

While doing an a11y audit on our Form component I found that our Form component is failing because `ARIA hidden element must not contain focusable elements`.

Our Form contains an implicitSubmit prop which generates a visually hidden submit button with aria-hidden set to true.

adds a tabIndex={-1} to our Form implied submit button. 

## 

### How to 🎩

Using Fast pass on the [Accessibility extension for web](https://chrome.google.com/webstore/detail/accessibility-insights-fo/pbjjkligggfmakdaogkfomddhfmpjeni?hl=en) ensure there are no errors with the playground below.

<details>

```jsx
import {Page, Form, FormLayout, TextField, Button} from '../src';

  return (
    <Page title="Playground">
      <FormExample />
    </Page>
  );
}

class FormExample extends React.Component {
  state = {
    url: '',
  };

  render() {
    const {url} = this.state;

    return (
      <Form noValidate onSubmit={this.handleSubmit}>
        <FormLayout>
          <TextField
            value={url}
            onChange={this.handleChange('url')}
            label="App URL"
            type="url"
          />

          <Button submit>Submit</Button>
        </FormLayout>
      </Form>
    );
  }

  handleSubmit = (event) => {
    this.setState({url: ''});
  };

  handleChange = (field) => {
    return (value) => this.setState({[field]: value});
  };
}
```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -7,6 +7,7 @@
 - Added `external` prop to `ResourceList` ([#2408](https://github.com/Shopify/polaris-react/pull/2408))
 - Added `onMouseEnter` and `onTouchStart` props to `Button` ([#2409](https://github.com/Shopify/polaris-react/pull/2409))
 - Added `ariaHaspopup` prop to `Popover` ([#2248](https://github.com/Shopify/polaris-react/pull/2248))
+- Fixed an accessibility issue where the `Form` implicit submit was still accessible via keyboard ([#2447](https://github.com/Shopify/polaris-react/pull/2447))
 
 ### Bug fixes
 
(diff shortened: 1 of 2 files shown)


dleroux:
@dpersing once you confirm that this in fact the behavior we want I'll update the changelog.

dleroux:
I believe is so that the form does submit by simply clicking enter as oppose to actually clicking on the button. Display none wouldn't have the same effect because the button would not be in the DOM.

If this is a temporary fix, what would the right fix be?

kaelig:
Would it help consumers to see an example in the style guide, or is this an edge case we want to discourage?

dleroux:
> Would it help consumers to see an example in the style guide, or is this an edge case we want to discourage?

I'm not sure it's worth it because it is the default behavior. I'm guessing a safeguard for when people add more than 1 field and no submit button is present, clicking enter would not trigger the on submit.

PR #2448, opened 2019-11-19
[Tooltip] Remove light prop

LauraAubin:

We had some discussion in Slack on whether or not we should keep this `light` prop.

https://shopify.slack.com/archives/CJZP6GX6C/p1573765466012900

TLDR; 
- Dark Tooltips add emphasis to content that may be an indication _not_ to use a Tooltip
- 3 out of 55 Tooltips in Web are light, so this change will be noticed
- The design language has some complications with this prop, i.e "Things that are dark in light-mode might not make sense to be light in dark-mode."

Removes the `light` prop and sets the Tooltip to be light by default.

### Anything for reviewers to consider?

Whether or not we want to make this change.

### DIFF
--- a/src/components/Tooltip/Tooltip.scss
+++ b/src/components/Tooltip/Tooltip.scss
@@ -26,19 +26,12 @@ $content-max-width: rem(200px);
   margin: spacing() spacing() $offset-from-activator;
 }
 
-.light {
-  .Wrapper {
-    background: color('white');
-    color: color('ink');
-  }
-}
-
 .Wrapper {
   position: relative;
   display: flex;
-  background-color: color('ink');
+  background: color('white');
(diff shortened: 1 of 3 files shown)


dleroux:
Unfortunately, we can not remove this prop because it will be a breaking change. We'd need to [deprecate](https://github.com/Shopify/polaris-react/blob/master/documentation/Deprecation%20guidelines.md) in this version and target v5 or maybe v6 depending on how many more minor versions we'll have in v4.

LauraAubin:
Going to close this PR since I don't think we have enough reason to support this change in situations that use the Tooltip amongst other small white elements.

Screen Shot 2019-11-21 at 3 56 01 PM

PR #2449, opened 2019-11-19
[Avatar] Use new color system

kyledurand:

This PR handles the first half of https://github.com/Shopify/polaris-ux/issues/333

I'll follow up with another PR to tackle the non latin name styles

image

Concern:
image

Dark mode colors are a little muddy and two are not passing our contrast standards. 

Using the new color system to match master brand styles in the Avatar component with fallbacks to the original colors.

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page, Avatar, Stack} from '../src';

  return (
    <Page title="Playground">
      <Stack alignment="baseline">
        <Avatar initials="KD" name="a" />
        <Avatar initials="KD" name="b" />
        <Avatar initials="KD" name="c" />
        <Avatar initials="KD" name="d" />
        <Avatar initials="KD" name="e" />
        <Avatar initials="KD" name="f" />
        <div
          style={{borderLeft: '1px solid hsl(0, 0%, 50%)', height: '40px'}}
        />
        <Avatar customer initials="KD" name="e" />
        <Avatar
          initials="JD"
          name="John D"
          source="https://images.unsplash.com/photo-1574157573248-293553e90ead?ixlib=rb-1.2.1&ixid=eyJhcHBfaWQiOjEyMDd9&auto=format&fit=crop&w=1534&q=80"
        />
      </Stack>
      <Stack alignment="baseline">
        <Avatar
          size="small"
          initials="JD"
          name="John D"
          source="https://images.unsplash.com/photo-1574157573248-293553e90ead?ixlib=rb-1.2.1&ixid=eyJhcHBfaWQiOjEyMDd9&auto=format&fit=crop&w=1534&q=80"
        />
        <Avatar
          initials="JD"
          name="John D"
          source="https://images.unsplash.com/photo-1574157573248-293553e90ead?ixlib=rb-1.2.1&ixid=eyJhcHBfaWQiOjEyMDd9&auto=format&fit=crop&w=1534&q=80"
        />
        <Avatar
          size="large"
          initials="JD"
          name="John D"
          source="https://images.unsplash.com/photo-1574157573248-293553e90ead?ixlib=rb-1.2.1&ixid=eyJhcHBfaWQiOjEyMDd9&auto=format&fit=crop&w=1534&q=80"
        />
      </Stack>
      <Stack alignment="baseline">
        <Avatar size="small" customer initials="KD" name="e" />
        <Avatar customer initials="KD" name="e" />
        <Avatar size="large" customer initials="KD" name="e" />
      </Stack>
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Avatar/Avatar.scss
+++ b/src/components/Avatar/Avatar.scss
@@ -38,26 +38,32 @@ $large-size: rem(60px);
 }
 
 .styleOne {
-  background: color('teal');
+  color: var(--p-decorative-one-text, color('white'));
+  background: var(--p-decorative-one-surface, color('teal'));
 }
 
 .styleTwo {
-  background: color('red');
+  color: var(--p-decorative-two-text, color('white'));
+  background: var(--p-decorative-two-surface, color('red'));
 }
 
 .styleThree {
-  background: color('orange');
+  color: var(--p-decorative-three-text, color('white'));
+  background: var(--p-decorative-three-surface, color('orange'));
 }
 
 .styleFour {
-  background: color('green');
+  color: var(--p-decorative-four-text, color('white'));
+  background: var(--p-decorative-four-surface, color('green'));
 }
 
 .styleFive {
-  background: color('blue');
+  color: var(--p-decorative-five-text, color('white'));
+  background: var(--p-decorative-five-surface, color('blue'));
 }
 
 .styleSix {
+  color: color('white');
   background: color('purple');
 }
 
@@ -71,6 +77,7 @@ $large-size: rem(60px);
   left: 50%;
   width: 100%;
   height: 100%;
+  background-color: var(--p-neutral, color('sky'));
   border-radius: $large-size / 2;
   transform: translate(-50%, -50%);
   object-fit: cover;
@@ -91,5 +98,4 @@ $large-size: rem(60px);
 .Svg {
   width: rem(48px);
   height: auto;
-  color: color('white');
 }
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -1,6 +1,7 @@
 import React, {useState, useCallback, useEffect} from 'react';
 
 import {classNames, variationName} from '../../utilities/css';
+import {useFeatures} from '../../utilities/features';
 import {useI18n} from '../../utilities/i18n';
 import {isServer} from '../../utilities/target';
 import {Image} from '../Image';
@@ -16,7 +17,7 @@ enum Status {
   Errored = 'ERRORED',
 }
 
-const STYLE_CLASSES = ['one', 'two', 'three', 'four', 'five', 'six'];
+const STYLE_CLASSES = ['one', 'two', 'three', 'four', 'five'];
 const AVATAR_IMAGES = Object.keys(avatars).map(
   // import/namespace does not allow computed values by default
   // eslint-disable-next-line import/namespace
@@ -50,6 +51,16 @@ export function Avatar({
   accessibilityLabel,
 }: AvatarProps) {
   const i18n = useI18n();
+  const {unstableGlobalTheming = false} = useFeatures();
+
+  function styleClass(name?: string) {
+    const finalStyleClasses = unstableGlobalTheming
+      ? STYLE_CLASSES
+      : [...STYLE_CLASSES, 'six'];
+    return name
+      ? finalStyleClasses[name.charCodeAt(0) % finalStyleClasses.length]
+      : finalStyleClasses[0];
+  }
 
   const [status, setStatus] = useState<Status>(Status.Pending);
 
@@ -138,12 +149,6 @@ export function Avatar({
       {imageMarkUp}
     </span>
   );
-}
-
-function styleClass(name?: string) {
-  return name
-    ? STYLE_CLASSES[name.charCodeAt(0) % STYLE_CLASSES.length]
-    : STYLE_CLASSES[0];
 }
 
 function customerPlaceholder(name?: string) {
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -17,7 +17,7 @@ enum Status {
   Errored = 'ERRORED',
 }
 
-const STYLE_CLASSES = ['one', 'two', 'three', 'four', 'five'];
+export const STYLE_CLASSES = ['one', 'two', 'three', 'four', 'five'];
 const AVATAR_IMAGES = Object.keys(avatars).map(
   // import/namespace does not allow computed values by default
   // eslint-disable-next-line import/namespace
--- a/src/components/Avatar/tests/Avatar.test.tsx
+++ b/src/components/Avatar/tests/Avatar.test.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import {mountWithAppProvider} from 'test-utilities/legacy';
 import {mountWithApp} from 'test-utilities';
 import {Avatar, Image} from 'components';
+import {STYLE_CLASSES} from '../Avatar';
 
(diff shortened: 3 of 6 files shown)


src/utilities/theme/color-adjustments.ts
```diff
@@ -590,7 +592,7 @@ export const colorAdjustments: ColorAdjustments = {
   },
   highlightText: {
     baseColor: 'highlight',
-    light: {lightness: 98},
+    light: {lightness: 38},
```
kyledurand: I'm not sure about leaving this change as part of this PR but this value was well below our contrast standards.

Before | After
---|---
 |

src/utilities/theme/types.ts
```diff
@@ -29,6 +29,8 @@ export interface ThemeConfig {
     warning?: string;
     highlight?: string;
     success?: string;
+    peach?: string;
+    cranberry?: string;
```
kyledurand: Peach and cranberry are what these are called in figma. We can probably think of better names for these

src/components/Avatar/tests/Avatar.test.tsx
```diff
@@ -32,7 +32,7 @@ describe('<Avatar />', () => {
     it('renders an Image component with a customer Avatar if the customer prop is true', () => {
       const avatar = mountWithAppProvider(<Avatar customer />);
       const image = avatar.find(Image);
-      expect(image.prop('source')).toContain('avatar-');
+      expect(image.prop('source')).toContain('customer-placeholder.svg');
```
alex-page:
```suggestion
      expect(image.prop('source')).toBe('customer-placeholder.svg');
```

src/utilities/theme/color-adjustments.ts
```diff
@@ -12,6 +12,8 @@ export enum UNSTABLE_Color {
   Warning = '#FFC453',
   Highlight = '#59D0C2',
   Success = '#008060',
+  Peach = '#FFC8B5',
+  Cranberry = '#FDC7CE',
```
alex-page: What do you think about providing one avatar color and then changing the lightness and hue to generate the palette?

src/utilities/theme/color-adjustments.ts
alex-page: Looking at the banner and how this color is used I am not sure why the color is white. I might be missing some context. Best to wait for @tmlayton or @danrosenthal.

tmlayton: This change seems fine, the lightness of the text for similar roles in 30-40 in light mode and 60-70 in dark mode. They should probably be consistent across the board 30/70, 40/60, or something else.

src/utilities/theme/color-adjustments.ts
tmlayton: I like the idea of providing a single decorative color, then perhaps adjusting based on the same hue delta as our colors to create the contrast/counter/secondary decorative color. In theory, regardless of what color is provided the resulted generated color should look nice.

That means these color roles would be relative adjustments instead of absolute ones, which is doable would just need some refactoring of the factory and new schema in the adjustments to flag this role as a relative adjustment.

Would like to talk through this a bit more though.

src/utilities/theme/types.ts
tmlayton: Agreed, we had some early discussion around these as decorative colors. The word "decorative" is so vague, though. Any thoughts?

src/components/Avatar/Avatar.scss
```diff
@@ -38,27 +38,28 @@ $large-size: rem(60px);
 }
 
 .styleOne {
-  background: color('teal');
+  color: var(--p-warning-text, color('white'));
```
tmlayton: Having success and warning roles used in avatars feels wrong to me, highlight a bit less. But maybe this is a time when it is okay to have an exception to color role usage?

UNRELEASED.md
```diff
@@ -4,6 +4,7 @@
 
 ### Enhancements
 
+- Removed customer images from avatar and updated to use new color system [#2449](https://github.com/Shopify/polaris-react/pull/2449)
```
tmlayton: Should customer images be removed in current Polaris, too? If that’s the case I think the customer images removal should be done in its own PR. We also have not been adding changelog entries for color system changes since they are invisible unless the feature flag is turned on and somewhat unstable.

tmlayton:
> Dark mode colors are a little muddy and two are not passing our contrast standards.

Don’t sweat it too much, I think as is quickly changing make the best call you can, ship, and iterate.

src/components/Avatar/Avatar.scss
tmlayton: Otherwise I wonder if it’s possible, to @alex-page’s point, to take a single decorative color and generate a range of complimentary colors that actually look good regardless of the base color. I think that logic would be based on reverse engineering our use of the color.

src/utilities/theme/types.ts
dleroux: extended, complementary? Could they be bundled under 1 role?

UNRELEASED.md
kyledurand: I think it makes sense to remove them before the masterbrand changes as there's no way to fallback to them once the new icon is added. I'll move that to a new PR and it doesn't have to be merged right away

kyledurand:
I've removed the changes to customer avatars and also now basing all decorative colors off of one base

image

image

Figma:
image

These all match the hsl values from the figma file nearly perfectly. Sometime values might be off by 1 but I can't seem to tell the difference.

alex-page:
Approving the work looks great, happy to keep talking about the naming in the future!

src/components/Avatar/Avatar.tsx
```diff
@@ -16,7 +16,7 @@ enum Status {
   Errored = 'ERRORED',
 }
 
-const STYLE_CLASSES = ['one', 'two', 'three', 'four', 'five', 'six'];
+const STYLE_CLASSES = ['one', 'two', 'three', 'four', 'five'];
```
danrosenthal: This will cause visual regressions, as indicated by the failing Percy test.

We can remove this class conditionally based on the `useFeatures` hook returning `globalTheming === true`, which will preserve backwards compatibility.

kyledurand: Oh nice, I'll make the change in a bit 👍

kyledurand:
Fallback added for the sixth variation:
image

And looping over five with the new system:
image

It's my understanding that this PR will be put on hold until we're more confident in how we should handle decorative coloring

src/components/Avatar/Avatar.scss
```diff
@@ -71,6 +77,7 @@ $large-size: rem(60px);
   left: 50%;
   width: 100%;
   height: 100%;
+  background-color: var(--p-neutral, color('sky'));
```
tmlayton: `--p-surface-background`

tmlayton: Actually maybe this should be neutral

src/utilities/theme/color-adjustments.ts
```diff
@@ -679,4 +680,110 @@ export const colorAdjustments: ColorAdjustments = {
...
+      hue: 52,
+      saturation: 100,
+      lightness: 18,
+    },
+    dark: {lightness: 100},
+  },
+  decorativeTwoSurface: {
+    baseColor: 'decorative',
+    light: {
+      hue: 30,
```
tmlayton: I think this is fine for now just to get something out there, but what I had in mind is these being relative changes in hue based on the role color.

With this approach, it ignores the base decorative color and fixes the color to Shopify’s palette. To get a relative approach working it would take updates to the typings and color factory.

PR #2450, opened 2019-11-20
Added eslint rule to restrict testing utility imports

AndrewMusgrave:

In `v4` we added `@shopify/react-testing` as our main testing library and started to refer to our enzyme code as legacy.

* Add linting rule
* Add ignores

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -62,7 +62,7 @@
         "paths": [
           {
             "name": "test-utilities/legacy",
-            "message": "Please import the modern testing library from test-utilities instead"
+            "message": "Please import the modern testing library from 'test-utilities' instead"
           }
         ]
       }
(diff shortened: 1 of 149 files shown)


.eslintrc
```diff
@@ -56,6 +56,17 @@
     "jsx-a11y/click-events-have-key-events": "off",
     "jsx-a11y/no-noninteractive-element-interactions": "off",
     "jsx-a11y/no-noninteractive-element-to-interactive-role": "off",
+    "no-restricted-imports": [
+      "error",
+      {
+        "paths": [
+          {
+            "name": "test-utilities/legacy",
+            "message": "Please import the modern testing library from test-utilities instead"
```
BPScott:
```suggestion
            "message": "Please import the modern testing library from 'test-utilities' instead"
```

A petty request for quote marks, but other than that. Thanks for looking into this!

PR #2455, opened 2019-11-21
[RangeSlider] Implement the color system

LauraAubin:

Fixes #0000 

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
+++ b/src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
@@ -33,6 +33,22 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
     background-color: transparent;
     cursor: pointer;
   }
+
+  &::after {
+    content: '';
+    position: absolute;
+    height: $range-track-height;
+    width: 100%;
+    background-image: linear-gradient(
+      to right,
+      var(--p-surface-inverse-foreground),
+      var(--p-surface-inverse-foreground) 50%,
+      transparent 50%,
+      transparent 100%
+    );
+    background-size: $range-track-height $range-track-height;
+    border-radius: var(--p-border-radius-base, border-radius());
+  }
 }
 
 .disabled {
@@ -53,7 +69,7 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
 
 .Input {
   --progress-lower: #{color('indigo')};
-  --progress-upper: #{color('sky', 'dark')};
+  --progress-upper: #{var(--p-override-transparent, color('sky', 'dark'))};
   // create-react-app v2 leverages postcss-preset-env to transpile modern CSS
   // into something kinda supported by older browsers. Unfortunatly its
   // custom properties transpiler has a bug that means it doesn't like multiple
@@ -80,7 +96,8 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
     cursor: pointer;
     width: 100%;
     height: $range-track-height;
-    background-color: var(--progress-upper, color('sky', 'dark'));
+    // background-color: red;
+    // background-color: var(--progress-upper, color('sky', 'dark'));
     background-image: linear-gradient(to right, var(--gradient-colors));
     border: none;
     border-radius: $range-track-height;


src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -33,6 +33,22 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
...
+    height: $range-track-height;
+    width: 100%;
+    background-image: linear-gradient(
+      to right,
+      var(--p-surface-inverse-foreground),
+      var(--p-surface-inverse-foreground) 50%,
+      transparent 50%,
+      transparent 100%
+    );
+    background-size: $range-track-height $range-track-height;
```
LauraAubin: We tried omitting the second parameter (the height) and letting it default to `auto`, though this doesn't seem fully supported on Safari.

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -33,6 +33,22 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
...
+
+  &::after {
+    content: '';
+    position: absolute;
+    height: $range-track-height;
+    width: 100%;
+    background-image: linear-gradient(
+      to right,
+      var(--p-surface-inverse-foreground),
+      var(--p-surface-inverse-foreground) 50%,
```
LauraAubin: @danrosenthal @tmlayton I'm not sure if this is the correct color to use. 

For reference this is the Figma example:
Screen Shot 2019-11-21 at 4 53 32 PM

Though the `surface` colors resemble a darker shade:

image

I also considered the `surface border` colors since they're slightly lighter:

Screen Shot 2019-11-21 at 4 56 59 PM

I'm assuming we should closely stick to the naming convention and tweak colors later, curious what y'all think?

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -67,6 +83,33 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
...
+
+      @include rangeSliderThumbPosition(
+        var(--p-range-slider-thumb-size-active, $range-thumb-size)
+      );
+
+      .disabled & {
+        width: $range-thumb-size;
+        height: $range-thumb-size;
+        @include rangeSliderThumbPosition($range-thumb-size);
+      }
```
LauraAubin: Although this seems a little repetitive, I didn't want to use the `:not()` function since it doesn't have great cross browser support.

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -114,51 +170,38 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
...
-      background-color: color('sky', 'dark');
-    }
-
-    @include range-thumb-selectors {
-      background: linear-gradient(
-        color('sky', 'lighter'),
-        color('sky', 'light')
-      );
-      border-color: color('indigo');
-      box-shadow: $range-thumb-shadow-focus;
```
LauraAubin: This doesn't appear to be needed anymore.

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
tmlayton: > I'm assuming we should closely stick to the naming convention and tweak colors later, curious what y'all think?

Yes, stick with the most fitting name and tweak colors later. Although, surface foreground won’t change much. Maybe `--p-surface-inverse-foreground-subdued` (soon to be called `--p-surface-foreground-subdued-inverse`)? Maybe a variant in the neutral role or a disabled variant (since it is kind of like the inactive part of the slider? Also not opposed to a new variant as well if something makes sense.

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -91,19 +142,24 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
...
 
+    @include rangeSliderThumbPosition($range-thumb-size);
+
     &:hover {
-      background: linear-gradient(
-        color('sky', 'lighter'),
-        color('sky', 'light')
+      background: var(
+        --p-interactive,
+        linear-gradient(color('sky', 'lighter'), color('sky', 'light'))
```
tmlayton: Fallback values using `linear-gradient` trigger the Safari infinite re-render bug. This will instead need to be done by conditionally adding a class with the new value, or another way which avoids the gradient fallback.

The `useFeatures` hook can be used to know if global theming is enabled or not.

LauraAubin: I wonder if we can use the fallback values within the `linear-gradient` like,

```scss
background: linear-gradient(
  var(--p-interactive, color('sky', 'lighter')),
  var(--p-interactive, color('sky', 'light'))
);
```

tmlayton: I can’t remember, fairly easy to test in Safari though. Happy to walk you through it

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -114,51 +173,38 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
...
 
     @include range-track-selectors {
-      background-color: color('red', 'light');
+      background-color: var(--p-override-none, color('red', 'light'));
     }
 
     @include range-thumb-selectors {
-      border-color: color('red');
+      background: var(--p-critical, color('white'));
+      border-color: var(--p-critical, color('red'));
```
danrosenthal: Likewise I think `criticalAction` makes sense for these.

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -114,51 +173,38 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
...
       box-shadow: $range-thumb-shadow-error;
     }
   }
 
   .disabled & {
     @include range-track-selectors {
       cursor: auto;
+      background-color: var(--p-border-on-inverse, color('sky', 'dark'));
```
danrosenthal: This feels wrong to me, for the same reasons as using the disabled variant for the track felt wrong in my earlier comment. This should use a disabled border variant in the neutral role, which @dleroux is working on creating. Sync with him and use something like that.

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -114,51 +173,38 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
...
-      box-shadow: $range-thumb-shadow-focus;
+      border-color: var(--p-border-on-surface, color('sky', 'dark'));
+      background: var(--p-surface-background, color('white'));
+
+      &:hover {
+        background: linear-gradient(
+          var(--p-surface-background, color('sky', 'lighter')),
+          var(--p-surface-background, color('sky', 'light'))
+        );
+      }
```
danrosenthal: I would expect the disabled range slider to use `interactiveActionDisabled` for the handle and applied portion of the track

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
danrosenthal: Can you add some more context to this?

LauraAubin: At first I removed it because I was unable to replicate the situation in which this is required. But since I don't fully understand why it was added in the first place, I'll keep it as is.

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -91,19 +145,24 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
...
     border-radius: 50%;
-    background: linear-gradient(color('white'), color('sky', 'lighter'));
+    background: linear-gradient(
+      var(--p-interactive, color('white')),
+      var(--p-interactive, color('sky', 'lighter'))
+    );
     box-shadow: $range-thumb-shadow;
     appearance: none;
-    transition-property: border-color, box-shadow;
+    transition-property: width, height, margin-top, border-color, box-shadow;
```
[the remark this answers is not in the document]
LauraAubin: The thumb is now animated because it's size and position changes on hover, so these properties also have a transition.

danrosenthal:
👋 Hey @LauraAubin, what's the status on this PR? Would be great to see this ready to merge. Anything I can do to help?

LauraAubin:
@danrosenthal thanks for the poke! My main blocker right now is bouncing between a bunch of different PR's so I'll try to get to this one as soon as I can.

Also, I wanted to split the `RangeSlider` work into separate PR's since I was finding that there was too much going on in one. So far I've split out the `track` work, ideally I'd like to have all of the parts completed before asking for review so they can appear to come together but still be easy to review.

danrosenthal:
Thanks for the update! Sounds good, @LauraAubin

PR #2459, opened 2019-11-22
[Color system] nested ThemeProvider

danrosenthal:
# High level context

Nested theme providers should be able to be used to flexibly configure use of color in Polaris components, as well as in consumer's custom components. They should be able to be used to invert the color scheme of the parent context, or just to override global values applied to color roles at a local level.

Closes https://github.com/Shopify/polaris-react/issues/2460

## What this pull request does
- stops the color factory from inferring the color scheme (mode) based on the lightness of the provided surface value in favor of a `colorScheme` property in the `theme` object that accepts `'light' | 'dark'`.
- stops the color factory from generating `-inverse`, `-dark`, and `-light` variants of every color in favor of using nested `<ThemeProvider/>` functionality to create an inverse, light, or dark nested context.
- sets the default `colorScheme` to `'light'` 
- converts enum `UNSTABLE_Color` to object for better ergonomics in code
- adds nested `<ThemeProvider/>` functionality where
  - a parent theme provider:
    - sets a default theme
  - a child theme provider:
    - does not set a default theme
    - inherits `colorScheme` from the parent when no `colorScheme` is provided to the child
    - accepts `'inverse'` as a `colorScheme` and inverts the parent `colorScheme`
    - inherits `UNSTABLE_colors` from the parent when the child color scheme differs from the parent color scheme to enable easily inverting color scheme context without the need to resupply the theme
- applies nested `<ThemeProvider/>` to `Toast`, eliminating the need for inverse color variants in styles
- updates storybook config and color docs config to support these API changes
- refactors the color factory for greater simplicity, moving default argument application to the `ThemeProvider` level

|Light mode Toast|Dark mode Toast|
|-|-|
|Screen Shot 2019-12-06 at 1 29 59 PM|Screen Shot 2019-12-06 at 1 30 11 PM|
|Screen Shot 2019-12-06 at 1 30 38 PM|Screen Shot 2019-12-06 at 1 30 26 PM|

|Playground|
|-|
|Screen Shot 2019-12-13 at 2 03 42 PM|

## Playground

```tsx
import React from 'react';
import {
  ThemeProvider,
  TextContainer,
  Button,
  ActionList,
  Popover,
  Modal,
} from '../src';

  return (
    <TextContainer spacing="tight">
      <Button primary>Save theme</Button>
      <ThemeProvider theme={{UNSTABLE_colors: {primary: '#2E72D2'}}}>
        <Button primary>Save theme</Button>
      </ThemeProvider>
      <ThemeProvider theme={{colorScheme: 'inverse'}}>
        <Popover
          active
          activator={<Button disclosure>More actions</Button>}
          onClose={() => {}}
        >
        </Popover>
        <Modal
          open
          onClose={() => {}}
          primaryAction={{
            content: 'Add Instagram',
            onAction: () => {},
          }}
          secondaryActions={[
            {
              content: 'Learn more',
              onAction: () => {},
            },
          ]}
        >
          <Modal.Section>
            <TextContainer>
              <p>
                Use Instagram posts to share your products with millions of
                people. Let shoppers buy from your store without leaving
                Instagram.
              </p>
            </TextContainer>
          </Modal.Section>
        </Modal>
      </ThemeProvider>
    </TextContainer>
  );
}
```

## Why this pull request does what it does

- nested theme providers should not provide defaults, as they are used to override colors  or color schemes applied by a parent theme provider
- nested theme providers should not set the `body` `color` and `backgroundColor`
- nested theme providers don't need to (redundantly) set `override` tokens
- explicitly setting the color scheme rather than the theme provider inferring it feels like a more sensible API (less magical)
- adding `-inverse`, `-dark`, and `-light` variants to each color variant adds a lot of weight to the system. By using a nested theme provider in components that create an inverse color scheme context (light mode in dark context, or dark mode in light context), we can avoid the complexity of generating all those variants for limited use-cases (`Toast`, `Tooltip` and `ContextualSaveBar` in the new design language).


src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -19,9 +22,18 @@ export function ThemeProvider({
   children,
 }: ThemeProviderProps) {
   const {unstableGlobalTheming = false} = useFeatures();
+  const {nested} = useParent();
```
BPScott: Instead of creating a new context, I think we can reuse the existing ThemeContext - it will be `undefined` if there has been no ThemeContext.Provider providing a value (see `src/utilities/theme/context.tsx`). That should remove the need for ParentContext entirely.

(and small pettiness: I like prefixing boolean values with `is` (or `has` of grammar requires it)

written but untested:

```ts
const rawTheme = useContext(ThemeContext); // may return undefined. Perhaps create a useRawTheme() hook?
// I updated to latest TS the other week, see if optional chaining works? :D
const parentSurface = rawTheme && rawTheme.UNSTABLE_colors ? rawTheme.UNSTABLE_colors.surface : undefined;
const isNested = !!rawTheme;
```

src/components/ThemeProvider/tests/ThemeProvider.test.tsx
```diff
@@ -123,22 +123,22 @@ describe('<ThemeProvider />', () => {
 
   it('sets color system properties when global theming is enabled', () => {
     const themeProvider = mountWithAppProvider(
-      <ThemeProvider theme={{}}>
+      <ThemeProvider theme={{UNSTABLE_colors: {surface: '#ffffff'}}}>
```
danrosenthal: I'm a little concerned about why I had to pass a theme to make this test pass. I might have screwed up how defaults are applied.

src/components/ThemeProvider/tests/ThemeProvider.test.tsx
```diff
@@ -123,22 +123,22 @@ describe('<ThemeProvider />', () => {
...
+      themeProvider.find('div').props().style || {},
     );
+
+    expect(styleKeys).toContain('--p-surface-background');
   });
 
   it('sets color system properties in context when global theming is enabled', () => {
     mountWithAppProvider(
-      <ThemeProvider theme={{}}>
+      <ThemeProvider theme={{UNSTABLE_colors: {surface: '#ffffff'}}}>
```
danrosenthal: Likewise here

src/components/ThemeProvider/tests/ThemeProvider.test.tsx
```diff
@@ -164,4 +164,145 @@ describe('<ThemeProvider />', () => {
       return null;
     }
   });
+
+  // since we mount each `ThemeProvider` withApp, each mounted `ThemeProvider` is a nested one
+  // we may need some sort of escape hatch to rendering a theme provider by default
```
danrosenthal: This is an important piece to call out, as the changes to `ThemeProvider` that allow it to be both a parent and a child, mean that any `ThemeProvider` mounted `withApp` will be a child theme provider (`withApp` also means `withThemeProvider`), so it won't perform any default behavior.

This may be a good reason not to make `ThemeProvider` perform that dual role, as it becomes hard to test. Or it could be a good reason for providing an escape hatch to allow `AppProvider` not to render a `ThemeProvider` by default.

src/components/ThemeProvider/tests/ThemeProvider.test.tsx
```diff
@@ -164,4 +164,145 @@ describe('<ThemeProvider />', () => {
...
+  it.todo('does set overrides');
```
danrosenthal: This is an important piece to call out, as the changes to `ThemeProvider` that allow it to be both a parent and a child, mean that any `ThemeProvider` mounted `withApp` will be a child theme provider (`withApp` also means `withThemeProvider`), so it won't perform any default behavior.

This may be a good reason not to make `ThemeProvider` perform that dual role, as it becomes hard to test. Or it could be a good reason for providing an escape hatch to allow `AppProvider` not to render a `ThemeProvider` by default.

BPScott: Taking a quick look at this test file I can't see any obvious reason why it can't use boring old `mount` instead of `mountWithAppProvider` to sidestep the "`mountWithAppProvider` provides a `ThemeContext` but we don't want that" issue.

.storybook/config.js
```diff
@@ -62,20 +62,16 @@ addDecorator(
           name: 'Enabled - Light Mode',
           props: {
             i18n: enTranslations,
-            features: {
-              unstableGlobalTheming: true,
-              theme: {UNSTABLE_colors: {surface: UNSTABLE_Color.Surface}},
-            },
+            features: {unstableGlobalTheming: true},
+            theme: {mode: 'light'},
```
tmlayton: Feels much better

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -19,28 +21,72 @@ export function ThemeProvider({
   children,
 }: ThemeProviderProps) {
   const {unstableGlobalTheming = false} = useFeatures();
+
+  const parentContext = useContext(ThemeContext);
+  const isParentThemeProvider = parentContext === undefined;
+  const parentMode = parentContext && parentContext.mode && parentContext.mode;
```
tmlayton: reads weird imo

tmlayton:
The docs script will need some updating, too, as it uses the build colors function.

danrosenthal:
> The docs script will need some updating, too, as it uses the build colors function.

I didn't think about that one, but yeah, makes sense. Since `buildColors` is no longer the source of defaults, how would you see that being done? Would you feel like taking a stab at that change?

src/components/ThemeProvider/tests/ThemeProvider.test.tsx
```diff
@@ -164,4 +165,221 @@ describe('<ThemeProvider />', () => {
       return null;
     }
   });
+
+  it('does set overrides', () => {
```
tmlayton: We’ll want to add a test to make sure each of the roles are added, too.

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -19,28 +21,73 @@ export function ThemeProvider({
...
-        themeConfig,
+        processedThemeConfig,
         unstableGlobalTheming ? customProperties : undefined,
       ),
-    [customProperties, themeConfig, unstableGlobalTheming],
+    [customProperties, processedThemeConfig, unstableGlobalTheming],
   );
 
-  // We want these values to be `null` instead of `undefined` when not set.
+  // We want these values to be empty string instead of `undefined` when not set.
```
tmlayton: Probably fine to leave this change, `null` seems to behave the same way but maybe this is safer for browser support? Alternatively, this could be more explicit and call `document.body.style.removeProperty` when the custom property is `undefined`.

src/utilities/theme/utils.ts
```diff
@@ -69,68 +79,33 @@ function hexToHsluvObj(hex: string) {
...
-                }),
-              };
-            }, {}),
-          };
-        }, {}),
-      };
-    },
-    {},
-  );
+  return Object.entries(colors).reduce((acc1, [role, hex]: [Role, string]) => {
```
tmlayton: Going to give this a quick refactor and remove the nested reducers.

src/components/ThemeProvider/ThemeProvider.tsx
BPScott: In TS 3.7 the updated the types of the builtins so null isn't considered valid and it threw an compile exception. I updated this in the update to TS3.7 but it got reverted. https://github.com/Shopify/polaris-react/pull/2237/files#diff-ce476db90622695c5696acdd862caf42

tmlayton: Thanks for the refresher!

src/components/ThemeProvider/tests/ThemeProvider.test.tsx
```diff
@@ -1,8 +1,10 @@
 import React from 'react';
 // eslint-disable-next-line no-restricted-imports
 import {mountWithAppProvider} from 'test-utilities/legacy';
+import {mount} from 'enzyme';
```
BPScott: Having to import mount and mountWithAppProvider from two separate places feels odd considering how similar they are. Can you update `test-utilities/legacy` so it reexports `mount` from enzyme and then you can do

```js
import {mount, mountWithAppProvider} from 'test-utilities/legacy';
```


We follow a similar pattern in `test-utilities` already

src/components/ThemeProvider/tests/ThemeProvider.test.tsx
```diff
@@ -164,4 +165,221 @@ describe('<ThemeProvider />', () => {
...
+    );
+    expect(style).not.toStrictEqual(
+      expect.objectContaining({
+        '--p-critical-surface': 'hsla(0, 0%, 98%, 1)',
+      }),
+    );
+  });
 });
+
+function mountWithGlobalTheming(
```
BPScott: Using mount's wrappingComponent options might be a little cleaner - and it also means that the node you get back  from calling `mount(someNode)` is  the node you passed in as the argument instead of the wrapping context provider. Perhaps something like this (inspired by our implementation of mountWithAppProvider)

```ts
export function mountWithGlobalTheming<P>(
  node: React.ReactElement<P>,
  unstableGlobalTheming: boolean,
) {
  return mount<P>(node, {
    wrappingComponent: FeaturesContext.Provider,
    wrappingComponentProps: {value: {unstableGlobalTheming}},
  });
}
```

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -19,28 +21,73 @@ export function ThemeProvider({
...
   // Otherwise, setting a style property to `undefined` does not remove it from the DOM.
-  const backgroundColor = customProperties['--p-surface-background'] || null;
-  const color = customProperties['--p-text-on-surface'] || null;
+  const backgroundColor = customProperties['--p-surface-background'] || '';
+  const color = customProperties['--p-text-on-surface'] || '';
```
tmlayton: Just realized in the case where the `mode` is different than the parent, the `color` needs to be set on the wrapping `div`, too. Otherwise, all of the text that inherits the text color will be the wrong color.

src/utilities/theme/types.ts
```diff
@@ -36,6 +36,8 @@ export interface RoleColors {
 
 export type Role = keyof RoleColors;
 
+export type Mode = 'light' | 'dark';
```
tmlayton: What do you think about an `'inverse'` type? It’s not really a mode and wouldn’t be a valid type for the root theme provider, but as a consumer this would be easier to use than getting the mode from context and using the inverse function or writing a ternary. If not, the theme hook/context consumer needs to be exported and documented.

src/utilities/theme/types.ts
```diff
@@ -45,6 +47,7 @@ export interface ThemeConfig {
     topBar?: Record<string, string>;
   };
   UNSTABLE_colors?: Partial<RoleColors>;
+  mode?: Mode;
```
tmlayton: What do you think about making `mode` a prop on `ThemeProvider`? So eventually it is something like:

```tsx
<ThemeProvider theme={{primary: '#FF0000'}} mode="light">
```

Maybe there is a potential to align to naming with color theory and the media query which this is referred to as a color scheme.

- https://en.wikipedia.org/wiki/Color_scheme
- https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -19,28 +21,73 @@ export function ThemeProvider({
...
+    parentContext &&
+    parentContext.UNSTABLE_colors &&
+    parentContext.UNSTABLE_colors;
+
+  const {UNSTABLE_colors, mode, ...rest} = themeConfig;
+
+  const childShouldInheritParentColors =
+    !isParentThemeProvider && mode !== undefined && mode !== parentMode;
+
+  const defaultColors = {
```
BPScott: Can we make `UNSTABLE_Color` be this object instead of an enum? That way you can spread the object instead of reconstructing a whole one if you ever want to create a theme that only manipulates a single role. That makes it a lot more composable:

```
const myPalette = {...UNSTABLE_Color, primary: '#0000ff';}
```

is easier than 

```
const myPalette = {
    surface: UNSTABLE_Color.Surface,
    onSurface: UNSTABLE_Color.OnSurface,
    interactive: UNSTABLE_Color.Interactive,
    neutral: UNSTABLE_Color.Neutral,
    primary: '#0000ff',
    critical: UNSTABLE_Color.Critical,
    warning: UNSTABLE_Color.Warning,
    highlight: UNSTABLE_Color.Highlight,
    success: UNSTABLE_Color.Success,
    decorative: UNSTABLE_Color.Decorative,
}
```

src/components/ThemeProvider/tests/ThemeProvider.test.tsx
BPScott: Following up from this, It'd be just as easy to use the new react-testing library by importing  `import {mount} from 'test-utilities';` and it'd mean you're not using a library we want to move away from. Could you do that instead?

src/components/ThemeProvider/tests/ThemeProvider.test.tsx
BPScott: Alternativly using a custom mount from react-testing would be preferable. Happy to pair with ya for these test updates next week.

src/index.ts
```diff
@@ -12,9 +12,11 @@ export {AppBridgeContext} from './utilities/app-bridge';
 
 /* eslint-disable @typescript-eslint/camelcase */
 export {
+  buildCustomProperties as UNSTABLE_buildCustomProperties,
   buildColors as UNSTABLE_buildColors,
   roleVariants as UNSTABLE_roleVariants,
   toCssCustomPropertySyntax as UNSTABLE_toCssCustomPropertySyntax,
   Tokens as UNSTABLE_Tokens,
+  inverse,
```
BPScott: Not a fan of exporting this. As Tim mentioned, being able to specify `mode: 'inverse'` so this flipping can be  encapsulated within the ThemeProvider would be preferable to putting that responsibility on the user.

src/components/ThemeProvider/tests/ThemeProvider.test.tsx
danrosenthal: Pairing next week on the test updates would be great! I will send you an invite

src/components/ThemeProvider/ThemeProvider.tsx
danrosenthal: Makes sense, but I think we should do this with CSS on the ThemeProvider node. That avoids the additional complexity a JavaScript solution would add.

src/utilities/theme/types.ts
danrosenthal: Definitely 👍

src/index.ts
danrosenthal: Agreed

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -19,32 +22,84 @@ export function ThemeProvider({
...
+
+  const className = classNames(
+    'ThemeProvider',
+    unstableGlobalTheming && 'globalTheming',
+  );
 
   return (
     <ThemeContext.Provider value={theme}>
-      <div style={customProperties}>{children}</div>
+      <div style={customProperties} className={className}>
```
tmlayton: Alternatively, this can be done by passing `color` to the style prop which avoids the Sass file and class names.

```suggestion
      <div style={{...customProperties, color}}>
```

BPScott: I'd prefer doing this to avoid the scss file too ^^

src/components/ThemeProvider/ThemeProvider.tsx
tmlayton: @BPScott `theme` rather than `palette`. The word "theme" (what our prop and component is called 🤦‍♂) was completely escaping me yesterday when we were chatting.

danrosenthal:
Setting the `backgroundColor` on the nested `ThemeProvider` `<div/>` ends up with some undesirable effects when an element has `borderRadius` or sets a `margin` on itself.

I ended up just adding `color` to it, to avoid this unsightliness. 

|gross|ugh|puke|
|-|-|-|
|Screen Shot 2019-12-12 at 2 18 38 PM|Screen Shot 2019-12-12 at 2 18 45 PM|Screen Shot 2019-12-12 at 2 19 15 PM|

tmlayton:
> Setting the backgroundColor on the nested ThemeProvider <div/>

Yes, the background color should not be set. The provider itself should be entirely invisible (only provide data) so technically even a div violates this model, but nothing we can do. `display: contents` is a no-go because of borking the a11y tree. Unless HTML introduces a provider I don’t think there is anything we can do.

danrosenthal:
@bpscott and @tmlayton feel like giving this one more look?

src/components/AppProvider/AppProvider.tsx
```diff
@@ -107,6 +107,7 @@ export class AppProvider extends React.Component<AppProviderProps, State> {
       children,
     } = this.props;
     const {intl, appBridge, link} = this.state;
+    const {colorScheme, ...restOfTheme} = theme;
```
BPScott: It seems odd that the theme we pass into the AppProvider differs from the theme we pass into the ThemeProvider.

I think this would be clearer if AppProvider's theme prop was equal to either: The theme prop of ThemeProvider, or the whole props object of ThemeProvider. Then in the future if we added new items we wouldn't need to adjust this bit of passthrough logic. You'll also avoid needing to muck about with new types like `ThemeProviderThemeConfig`.

```jsx
<ThemeProvider theme={theme}>....</ThemeProvider>

// or

<ThemeProvider ...theme>....</ThemeProvider>
```

The second option is teeechnically a breaking change but I don't think we ever supported people using ThemeProvider standalone externally anyway.

src/components/ThemeProvider/tests/ThemeProvider.test.tsx
```diff
@@ -172,4 +173,410 @@ describe('<ThemeProvider />', () => {
...
+
+  it('sets overrides', () => {
+    const wrapper = mountWithGlobalTheming(
+      <ThemeProvider theme={{}}>
+        <p>Hello</p>
+      </ThemeProvider>,
+      {globalTheming: true},
+    );
+
+    const styleKeys = Object.keys(wrapper.find('div')!.props.style || {});
```
BPScott: Update all these new tests to use the `toHaveReactProps` style assertions, e.g.

```ts
expect(themeProvider.find('div')).toHaveReactProps({
  style: expect.objectContaining({
    '--p-override-zero': '0',
  }),
});
```


We should probably collapse them all into one or two tests by putting multiple items in that objectContaining bit, that would make this less boilerplatey

src/index.ts
```diff
@@ -12,9 +12,11 @@ export {AppBridgeContext} from './utilities/app-bridge';
 
 /* eslint-disable @typescript-eslint/camelcase */
 export {
+  buildCustomProperties as UNSTABLE_buildCustomProperties,
```
BPScott: Why does this need to be exposed publicly?

src/utilities/theme/role-variants.ts
```diff
@@ -1,19 +1,17 @@
...
-  Interactive = '#2E72D2',
-  Neutral = '#111213',
-  Primary = '#008060',
-  Critical = '#D82C0D',
-  Warning = '#FFC453',
-  Highlight = '#5BCDDA',
-  Success = '#008060',
-  Decorative = '#FDC7CE',
-}
+export const DEFAULT_COLOR = {
```
BPScott: Slightly tangental but I was chatting with Tim yesterday, what do you think about naming this "list of colors" a Palette?

So this becomes DEFAULT_PALETTE, and buildColors becomes buildPalette.

src/utilities/theme/types.ts
```diff
@@ -36,6 +36,10 @@ export interface RoleColors {
 
 export type Role = keyof RoleColors;
 
+export type ColorScheme = 'light' | 'dark';
+
+export type ThemeProviderColorScheme = ColorScheme | 'inverse' | undefined;
```
BPScott: Do we want something that should act as inherit from the current theme context? Is that what undefined does?

src/utilities/theme/utils.ts
```diff
@@ -41,10 +42,16 @@ export function buildThemeContext(
   themeConfig: ThemeConfig,
   cssCustomProperties?: CustomPropertiesLike,
 ): Theme {
-  const {logo} = themeConfig;
+  const {
+    logo,
+    UNSTABLE_colors,
+    colorScheme = DEFAULT_COLOR_SCHEME,
```
BPScott: It feels like the normalisation to set a default value for the colorScheme should have already happened by this point. It feels like that should happen at the same point we turn custom properties object, that we pass into this function, rather than trying to work it out in here.

src/utilities/theme/role-variants.ts
tmlayton: Probably should be called theme rather than palette (referred to this in another comment). And pascal case for enums? `DefaultTheme` (even though the keys are not pascal case)

tmlayton: And the factor really is a color role variant factory, but that’s a long name :)

src/index.ts
danrosenthal: It doesn't. Good catch

src/utilities/theme/role-variants.ts
danrosenthal: `DefaultTheme` it is

src/utilities/theme/types.ts
danrosenthal: `undefined` currently acts as `inherit`. Should we be more explicit and add that as a type?

src/components/AppProvider/AppProvider.tsx
danrosenthal: While the API is a bit more ergonomic when you just want to set the `colorScheme`, I don't feel the increase in complexity is worth it to pull our `colorScheme` into its own prop on the `ThemeProvider`.

How strongly do you feel this should be done, @tmlayton?

danrosenthal:
Ready for another pass @tmlayton and @BPScott.

@BPScott, I've added context to the description, and asked a couple followup questions on your most recent round of feedback. Let me know if you still want to pair to get a deeper understanding of what's going on here before approving this PR.

danrosenthal:
Ready for another pass @tmlayton and @BPScott.

@BPScott, I've added context to the description, and asked a couple followup questions on your most recent round of feedback. Let me know if you still want to pair to get a deeper understanding of what's going on here before approving this PR.

danrosenthal:
Ready for another pass @tmlayton and @BPScott.

@BPScott, I've added context to the description, and asked a couple followup questions on your most recent round of feedback. Let me know if you still want to pair to get a deeper understanding of what's going on here before approving this PR.

danrosenthal:
While working up a `Playground` for this, I noticed that nested theme providers don't work quite as intended on elements which use a `Portal`. Since the nested theme provider sets color on its own div, elements which don't specify their color incorrectly inherit it from the body, as they're not nested within the `ThemeProvider` `<div />`.

I'll need to work up a mechanism for getting the `Portal` to set the `color` style on its div.

|🎩 |
|-|
|Screen Shot 2019-12-13 at 1 14 33 PM|

## Playground

```tsx
import React from 'react';
import {
  ThemeProvider,
  TextContainer,
  Button,
  ActionList,
  Popover,
  Modal,
} from '../src';

  return (
    <TextContainer spacing="tight">
      <Button primary>Save theme</Button>
      <ThemeProvider theme={{UNSTABLE_colors: {primary: '#2E72D2'}}}>
        <Button primary>Save theme</Button>
      </ThemeProvider>
      <ThemeProvider theme={{colorScheme: 'inverse'}}>
        <Popover
          active
          activator={<Button disclosure>More actions</Button>}
          onClose={() => {}}
        >
        </Popover>
      </ThemeProvider>
      <ThemeProvider theme={{colorScheme: 'inverse'}}>
        <Modal
          open
          onClose={() => {}}
          primaryAction={{
            content: 'Add Instagram',
            onAction: () => {},
          }}
          secondaryActions={[
            {
              content: 'Learn more',
              onAction: () => {},
            },
          ]}
        >
          <Modal.Section>
            <TextContainer>
              <p>
                Use Instagram posts to share your products with millions of
                people. Let shoppers buy from your store without leaving
                Instagram.
              </p>
            </TextContainer>
          </Modal.Section>
        </Modal>
      </ThemeProvider>
    </TextContainer>
  );
}

```

src/components/AppProvider/AppProvider.tsx
tmlayton: I'm okay with that

src/utilities/theme/role-variants.ts
```diff
@@ -1,19 +1,19 @@
...
-  Interactive = '#2E72D2',
-  Neutral = '#111213',
-  Primary = '#008060',
-  Critical = '#D82C0D',
-  Warning = '#FFC453',
-  Highlight = '#5BCDDA',
-  Success = '#008060',
-  Decorative = '#FDC7CE',
-}
+export const DefaultTheme = {
```
tmlayton:
```suggestion
export const DefaultThemeColors = {
```

src/utilities/theme/types.ts
```diff
@@ -54,6 +57,9 @@ export interface Theme {
   /** Sets the logo for the top bar and contextual save bar components*/
   logo?: ThemeLogo;
   UNSTABLE_cssCustomProperties?: string;
+  UNSTABLE_colors?: Partial<RoleColors>;
+  colorScheme?: ColorScheme;
+  rootElementColor?: string;
```
tmlayton:
```suggestion
  textColor?: string;
```

PR #2468, opened 2019-11-25
[DropZone] Add `size` prop to overwrite computed size based on node

loic-d:

We are running into issues with the initial `size` defined in `DropZone`. 

See [`web` issue for context](https://github.com/Shopify/web/issues/19946).

When the component initializes, `size` in `state` is set to `extraLarge`. 
When the component is successfully mounted, `size` will be updated based on the node's dimensions.

During this measuring process, we apply a height of `rem(205px)` (`extraLarge`)

67314907-86e5e180-f4d3-11e9-9c98-aefa9ec4dbe5

This causes the UI to 'jump' when the measuring step completes and the computed value is < `extraLarge`.

Nov-25-2019 16-24-58

This PR adds a new `size` prop to `DropZone`. It overrides the dynamic / computed size. When passing a `size`, we won't calculate dimensions based on current node. Helpful when the dropzone's dimension won't change, for example in the Product Details variants card. 

## 

### How to 🎩

🎩 in `web`.

- `yarn run build-consumer web`
- Checkout `variants-card-jump-issue` in `web`
- Make sure you have a single location on your shop
- Open the Product Details page in React for a product with variants
- Make sure the 'jumping' effect is gone when you load the page (see variants card)

<details>

```jsx
import React from 'react';
import {Page, DropZone} from '../src';

  return (
    <Page title="Playground">
      <DropZone type="image" size="small">
        <div>
          Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vivamus
          tempor interdum diam scelerisque tempus. Sed aliquet eros in elit
          scelerisque, luctus condimentum arcu iaculis. Maecenas malesuada
          mollis justo vel lacinia. Nullam eget velit vel tortor pharetra
          aliquam sit amet nec lorem
        </div>
      </DropZone>
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/DropZone/DropZone.tsx
+++ b/src/components/DropZone/DropZone.tsx
@@ -28,6 +28,8 @@ import {DropZoneContext} from './context';
 
 import {fileAccepted, getDataTransferFiles} from './utils';
 
+import {Size} from './types';
+
 import styles from './DropZone.scss';
 
 export type Type = 'file' | 'image';
@@ -40,7 +42,7 @@ interface State {
   focused: boolean;
   numFiles: number;
   overlayText?: string;
-  size: string;
+  size: Size;
   measuring: boolean;
   type?: string;
 }
@@ -92,6 +94,8 @@ export interface DropZoneProps {
   dropOnPage?: boolean;
   /** Sets the default file dialog state */
   openFileDialog?: boolean;
+  /** Forces unique dropzone container size */
+  size?: Size;
   /** Adds custom validations */
   customValidator?(file: File): boolean;
   /** Callback triggered on click */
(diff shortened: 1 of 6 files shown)


src/components/DropZone/DropZone.tsx
```diff
@@ -40,7 +42,7 @@ interface State {
   focused: boolean;
   numFiles: number;
   overlayText?: string;
-  size: string;
+  size: Size;
```
loic-d: Should we rename the prop? 
Consumer should only use it when they want a 'static' size and opt-out from the computed size based on node.

dleroux:
👋 @loic-d Do you think [this PR](https://github.com/Shopify/polaris-react/pull/2412), which will be the next release, fix your issue?

loic-d:
@dleroux Thanks for following up. 

It seems to work. While measuring, it sets `visibility: hidden` and `min-height: 0` on the container.

Screen Shot 2019-11-26 at 9 27 24 AM

What's the rationale behind this calculation at run-time? Should we allow consumers to opt-out by passing a fixed `size` as prop?

dleroux:
There aren't many examples like this where we don't want the dropzone to be responsive. I was thinking there might be a small performance gain by having an 'initialSize' prop but not sure it's worth the extra prop. What do you think?

loic-d:
I am fine with https://github.com/Shopify/polaris-react/pull/2412 as a fix for our issue.

Thanks for your input!

PR #2482, opened 2019-11-28
[WIP] A quick investigation into linking to figma designs from stories

BPScott:
This demos how you can link to either a page or frame in Figma. We'd
need a thing to map from a nodeId provided in a readme's frontmatter to
a single set of stories, but that would be easy.

To see:

Fire up storybook, visit Button (or any other page), View the designs in the addons panel ('press a' if you can't see it already)

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -21,6 +21,7 @@ module.exports = {
     '@storybook/addon-a11y',
     '@storybook/addon-contexts',
     '@storybook/addon-knobs',
+    'storybook-addon-designs',
   ],
   webpackFinal: (config) => {
     const isProduction = config.mode === 'production';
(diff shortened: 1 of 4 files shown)


HYPD:
**Value**
Keeps design and development together, easier to see any discrepancies, easier to link to the specific file. You’d be able to potentially see states, visual examples, and have designers and developers able to get to the code or visual design from the same place (if pointed to an example story)


**Implications**
1. The UI kit would have to be public (the file), so we could live embed frames/components. Is that ok? (there's always been some intent to do this)
2. Is this the right time, considering a new design language is starting up? Maybe as it's a good testing ground

cc @sarahill @kaelig

HYPD:
Figma now has private embeds https://twitter.com/figmadesign/status/1262829360689496064

alex-page:
Going to close this for now. Awesome concept @BPScott. Feel free to reopen if still relevant.

PR #2486, opened 2019-11-29
[Color System] Adjustments to HSLuv values

jessebc:

Corrects the translated colours from HSL to HSLuv colour space. 

Updates lightness values in the `role-variants.ts` theme file.


src/utilities/theme/role-variants.ts
```diff
@@ -4,13 +4,13 @@ import {RoleVariants} from './types';
 export enum UNSTABLE_Color {
   Surface = '#FAFAFA',
   DarkSurface = '#111213',
-  OnSurface = '#1F2225',
-  Interactive = '#0870D9',
-  Neutral = '#EAEAEB',
+  OnSurface = '#111213',
+  Interactive = '#2E72D2',
+  Neutral = '#111213',
```
danrosenthal: Why such a dark value for neutral?

src/utilities/theme/role-variants.ts
```diff
@@ -186,8 +186,8 @@ export const roleVariants: RoleVariants = {
     {
       name: 'interactiveAction',
       description: 'Used for links and plain buttons.',
-      light: {lightness: 44},
-      dark: {lightness: 65},
+      light: {lightness: 48.6},
+      dark: {lightness: 65, saturation: 100, hue: 247.6},
```
danrosenthal: This change, and the changes to `interactiveActionDisabled`, `interactiveActionHovered`, and  `interactiveActionPressed` mean that the interactive action color is no longer configurable. Is that the intention?

src/utilities/theme/role-variants.ts
```diff
@@ -4,13 +4,13 @@ import {RoleVariants} from './types';
 export enum UNSTABLE_Color {
   Surface = '#FAFAFA',
   DarkSurface = '#111213',
-  OnSurface = '#1F2225',
-  Interactive = '#0870D9',
-  Neutral = '#EAEAEB',
+  OnSurface = '#111213',
```
danrosenthal: This change to `OnSurface` means we're getting failures for the `textOnSurfaceDark` variant. That variant is used in all `destructive` and `primary` buttons as a text value. Those are currently failing the Pally accessibility check for color contrast. We should be able to solve that by making the lightness a bit higher for that variant.

src/utilities/theme/role-variants.ts
```diff
@@ -136,43 +136,43 @@ export const roleVariants: RoleVariants = {
...
-      dark: {lightness: 62},
+      dark: {lightness: 52.1},
     },
     {
       name: 'textOnSurface',
       description: 'For use as a neutral text color.',
-      light: {lightness: 13},
-      dark: {lightness: 100},
+      light: {lightness: 13.1},
+      dark: {lightness: 90.8},
```
danrosenthal: Per my comment about accessibility, this value will need to go lighter to get `textOnSurfaceDark` to pass color contrast.

src/components/Badge/Badge.scss
```diff
@@ -18,7 +18,7 @@ $pip-spacing: ($height - $pip-size) / 2;
   display: inline-flex;
   align-items: center;
   padding: 0 $horizontal-padding;
-  background-color: var(--p-neutral, color('sky'));
+  background-color: var(--p-neutral-action-disabled, color('sky'));
```
danrosenthal: Until we have a better variant for this purpose, I've gone ahead and used the neutral disabled action.

src/utilities/theme/role-variants.ts
jessebc: I'm assuming that these source colours are essentially setting the Hue and that the lightness and saturation are taken care of in the adjustments. Is this incorrect?

src/utilities/theme/role-variants.ts
jessebc: It _shouldn't_. Could this be pointing to surface contrast issue somewhere else?

src/utilities/theme/role-variants.ts
jessebc: I found out that with blue, the Hue is treated differently than it is in other colour spaces. When we make it lighter, it also shifts to be more purple. 

I can try starting it from a "bluer" starting place to see if we can compensate, but if not we may need to lock it down in this way. 

I figured this was ok for our internal needs, and that we could ship a cleaner version for partners.

danrosenthal:
Thinking about this a bit more, and I think the problem is more to do with the fact that we should be using `textOnPrimary` for primary action text, and not `textOnSurfaceDark`. Likewise, we should create a `textOnCritical` variant for use as the critical action text, and not use `textOnSurfaceDark` for that either.

src/utilities/theme/role-variants.ts
jessebc: I'll have to set up a call... because I'm having a hard time connecting the dots on that one now.

src/utilities/theme/role-variants.ts
danrosenthal: Thinking about this a bit more, and I think the problem is more to do with the fact that we should be using textOnPrimary for primary action text, and not textOnSurfaceDark. Likewise, we should create a textOnCritical variant for use as the critical action text, and not use textOnSurfaceDark for that either.

src/utilities/theme/role-variants.ts
danrosenthal: Sorry this was a bit confusing. I ultimately arrived at the conclusion that we should be using a different variant for setting the text color on primary and critical actions.

src/utilities/theme/role-variants.ts
danrosenthal: You're definitely not wrong there. It just felt a bit weird to start with a darker lightness value when all the other inputs essentially start with the light mode value, not the dark mode value.

jessebc: That's fair. The main thing I was trying to do was just make sure all the "greys" in the palette came from the same hue/saturation starting point. We can adjust those down to be a lighter starting point if that keeps things more straightforward.

src/utilities/theme/role-variants.ts
```diff
@@ -318,41 +318,13 @@ export const roleVariants: RoleVariants = {
...
-      light: {lightness: 88},
-      dark: {lightness: 88},
-    },
-    {
-      name: 'iconDisabledOnPrimary',
-      description:
-        'For use as a fill color for icons on disabled primary actions. Not for use in icons on navigation and tabs.',
-      light: {lightness: 78},
-      dark: {lightness: 78},
-    },
```
danrosenthal: These variants are superfluous, as we only style critical and primary buttons with white icons and text, regardless of state.

src/utilities/theme/role-variants.ts
```diff
@@ -318,41 +318,13 @@ export const roleVariants: RoleVariants = {
...
-      light: {lightness: 90},
-      dark: {lightness: 90},
-    },
-    {
-      name: 'textDisabledOnPrimary',
-      description:
-        'For use as a text color on disabled primary actions. Not for use in text on navigation and tabs.',
-      light: {lightness: 90},
-      dark: {lightness: 90},
-    },
```
danrosenthal: These variants are superfluous, as we only style critical and primary buttons with white text, regardless of state.

src/components/Button/Button.scss
```diff
@@ -158,6 +158,7 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
 
 .primary {
   --p-button-color: var(--p-primary-action);
+  --p-button-text: var(--p-text-on-primary);
```
danrosenthal: It would be really useful if we could make `textOnPrimary` use the `dark` config when `primaryAction` is dark, and use the `light` config when `primaryAction` is light. That would open up `Primary` to be user configurable while still preserving accessibility.

I think that's a job for another PR though, as it will require some code changes to `utils.ts`

cc: @tmlayton

danrosenthal:
I've gone ahead and hooked up the text for `primary` and `critical` buttons to use tailor-made variants for each use-case, `textOnPrimary` and `textOnCritical`. This solves the color contrast issues without needing to make any tweaks to `textOnSurface`/`textOnSurfaceDark`.

src/utilities/theme/role-variants.ts
tmlayton: Finding a way for this to work without fixing the hue is ultimately what we'll need to figure out. Can you find a more blue starting place that looks good? Otherwise this could be handled universally as a relative shift of `-7.1`, or we could get more complicated and only do shifts based on hue, or have a specific role variant config for our design language which is different than the factory default.

For now we can ship and think on it async unless there is a great answer now.

tmlayton:
Rebased, happy to merge and continue to think about to handle the hue adjustment for that blue.

PR #2489, opened 2019-11-29
add pre-commit hook to generate new color docs if the source changes

tmlayton:
This will run `yarn colordocs` and add the updated docs when committing a change to `role-variants.ts`

### DIFF
--- a/pre-commit
+++ b/pre-commit
@@ -27,7 +27,7 @@ module PreCommit
   def generate_color_docs
     changed_files.each do |path|
       if COLOR_DOC_SOURCE_FILE_PATHS.include? path
-        puts "Color system changes detected. Generating new docs..."
+        puts "Color system changes detected. Generating new docs (may take a couple minutes)..."
         %x(yarn run colordocs)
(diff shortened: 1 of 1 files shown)


pre-commit
```diff
@@ -22,6 +24,16 @@ module PreCommit
     end
   end
 
+  def generate_color_docs
+    changed_files.each do |path|
+      if COLOR_DOC_SOURCE_FILE_PATHS.include? path
+        puts "Color system changes detected. Generating new docs..."
```
kaelig: It could be useful to mention this may take some time, to manage the user's expectations.

tmlayton:
> Although it's unfortunate it takes so long for the commit (1 minute and 40 seconds)

The script to generate the docs is almost instant, but this needs `yarn build` and `yarn format` to run before and after, respectively.

pre-commit
tmlayton:
```suggestion
        puts "Color system changes detected. Generating new docs (may take a couple minutes)..."
```

PR #2496, opened 2019-12-04
Test coverage commands

tmlayton:

To make it dead simple to generate and open a test coverage report (the `yarn test:coverage` command was also broken).

Adds two new commands:

- `dev test:coverage` will run every test or `dev test:coverage path/to/test` and automatically opens the coverage report in your default browser when finished
- `dev open coverage` for manually opening the coverage report

Two changes to the package scripts:

- Forces Jest to run in CI-mode so tests only run once and a coverage report can be properly generated (this was broken if trying to generate a coverage report in the default watch mode and `--no-watch` does not work as it will _not_ run any tests by default)
- Added `posttest:coverage` which opens the coverage report after `dev test:coverage` or `yarn test:coverage` runs

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

* [x] 🎩’d the CLI


package.json
```diff
@@ -34,7 +34,8 @@
     "format": "sewing-kit format",
     "type-check": "sewing-kit type-check",
     "test": "sewing-kit test",
-    "test:coverage": "yarn test --coverage",
+    "test:coverage": "CI=true yarn test --coverage",
```
LauraAubin: I can see value in also allowing this command to test specific files since it takes a few minutes to generate the report for all of them. I believe you can do this now with `yarn test Modal --coverage`.

LauraAubin: Actually, scratch that! This command works fine if you include the file name at the end like `CI=true yarn test --coverage Card`.

tmlayton: Yes, args get passed through so still possible to do `yarn test:coverage card` or `dev test:coverage card`

UNRELEASED.md
```diff
@@ -43,6 +43,11 @@
 
 ### Development workflow
 
+- Added `dev test:coverage` command ([#2496](https://github.com/Shopify/polaris-react/pull/2496))
```
```suggestion
- Added `dev test:coverage` command ([#2496](https://github.com/Shopify/polaris-react/pull/2496)) to run `yarn test:coverage`
```

UNRELEASED.md
```diff
@@ -43,6 +43,11 @@
...
+- Added `dev open coverage` command ([#2496](https://github.com/Shopify/polaris-react/pull/2496))
```
```suggestion
- Added `dev open coverage` command ([#2496](https://github.com/Shopify/polaris-react/pull/2496)) to run `posttest:coverage`
```

dleroux: Is that what it does? I also wonder if it's not best to put the yarn command first in the changelog. Just because when I read the first 2 entries I wasn't sure what they would do until I read the yarn entries.

dev.yml
```diff
@@ -9,14 +9,16 @@ up:
...
   app: http://localhost:6006/
   shipit: https://shipit.shopify.io/shopify/polaris-react/production
+  coverage: build/coverage/index.html
 
 commands:
   # People new to the project might be tempted to run some
   # common commands, so let’s have them in here.
   # For a list of Common Commands in Projects, check out:
-  # https://github.com/Shopify/dev/wiki/Common-Commands-in-Projects
+  # https://development.shopify.io/tools/dev/getting_started/Common-Commands-in-Projects
```
kaelig: nice catch

UNRELEASED.md
tmlayton: `dev open coverage` runs `open build/coverage/index.html` so there is no yarn alias although I can add one.

PR #2505, opened 2019-12-04
[DropZone][Fileupload] Applied color system

AndrewMusgrave:

Fixes *part of* https://github.com/Shopify/polaris-ux/issues/341

#### Regular

Screen Recording 2020-01-20 at 05 08 PM
Screen Recording 2020-01-20 at 05 08 PM 2

## Dark

Screen Recording 2020-01-20 at 05 10 PM
Screen Recording 2020-01-20 at 05 09 PM

### DIFF
--- a/src/components/DropZone/DropZone.scss
+++ b/src/components/DropZone/DropZone.scss
@@ -2,25 +2,41 @@
 
 $dropzone-padding: rem(15px);
 $dropzone-border-style: dashed;
-$dropzone-border-color: color('ink', 'lightest');
-$dropzone-border-color-disabled: color('sky');
+$dropzone-border-color: var(--p-border, color('ink', 'lightest'));
+$dropzone-border-color-disabled: var(--p-border-disabled, color('sky'));
 $dropzone-border-radius: border-radius();
-$dropzone-background: color('white');
+$dropzone-background: var(--p-surface, color('white'));
 $dropzone-min-height-extra-large: rem(205px);
 $dropzone-min-height-large: rem(160px);
 $dropzone-min-height-medium: rem(100px);
 $dropzone-min-height-small: rem(50px);
 
-$dropzone-overlay-color: color('indigo');
-$dropzone-overlay-color-error: color('red');
-$dropzone-overlay-border-color: color('indigo');
-$dropzone-overlay-border-color-error: color('red');
-$dropzone-overlay-background: color('indigo', 'lighter');
-$dropzone-overlay-background-error: color('red', 'lighter');
+$dropzone-overlay-text-color: var(--p-text, color('indigo'));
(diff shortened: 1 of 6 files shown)


danrosenthal:
👋 Anything I can do to help get this out of draft state and ready for review @AndrewMusgrave? It'd be great to get this one ready to go

tmlayton:
This is looking good! Can you provide some more screenshots with the various states like errors?

src/styles/foundation/_focus-ring.scss
```diff
@@ -1,8 +1,9 @@
 /// Sets the focus ring for an interactive element
 /// @param {String} $size - The size of the border radius on the focus ring.
+/// @param {BorderStyle} $style - The style of the border on the focus ring.
 ///
 
-@mixin focus-ring($size: 'base') {
+@mixin focus-ring($size: 'base', $style: solid) {
```
danrosenthal: Let's keep the focus ring consistent across all elements, that way it's totally obvious to merchants any time an element is in the focused state. Having one-off variations like `dashed` is counter to this consistency. I think in this case, the focus ring can wrap outside the default state dashed border with a `1px` offset.

AndrewMusgrave:
> This'll also need a rebase off master, as the focus ring mixin has had some API changes as part of the Checkbox work to allow it to animate

Yikes the changes didn't play well with dropzone 😬 Updated the pr and included gifs in the description (error states included)

danrosenthal:
Looking at this after your focus ring changes, and I have a couple of thoughts.

1) Maybe the focus ring on the button is good enough. It strikes me as strange to have both elements enter a focus state at the same time.
2) If we're going to keep both rings, maybe the drop zone's focus ring should sit outside the dashed line, not replace it.
3) I don't notice a transition as the focus ring is being applied to the dropzone. What's up there?

AndrewMusgrave:
>Maybe the focus ring on the button is good enough. It strikes me as strange to have both elements enter a focus state at the same time.

This is originally why I left the border as dotted 😛

>If we're going to keep both rings, maybe the drop zone's focus ring should sit outside the dashed line, not replace it.

I still think it looks off, would also be interested in @sarahill opinion on this!

>I don't notice a transition as the focus ring is being applied to the dropzone. What's up there?

The `focus-ring` doesn't play well with `DropZone` because it uses `::after`. Eventually, we could probably redo the markup structure of dropzone, maybe add an overlay to add the focus too, however with the current implementation its a no-go. Dan L and Kyle were going to look at other approaches for the `focus-ring` so 🤞

AndrewMusgrave:
> The upload button getting a focus ring should be enough affordance for a keyboard user. Let's only use that focus ring, and keep the dashed border in its default state during focus.

Unfortunately, the upload button is the `FileUpload` component and won't always be present. When the dropzone is used by itself, it would have no visible focus state.

PR #2510, opened 2019-12-04
[ProgressBar] Applied color system

AndrewMusgrave:

Fixes https://github.com/Shopify/polaris-ux/issues/356

### Screenies

Screen Shot 2019-12-04 at 1 01 05 PM
Screen Shot 2019-12-04 at 1 01 00 PM

### DIFF
--- a/src/components/ProgressBar/ProgressBar.scss
+++ b/src/components/ProgressBar/ProgressBar.scss
@@ -20,7 +20,7 @@
 .ProgressBar {
   overflow: hidden;
   width: 100%;
-  background-color: var(--p-secondary-action, color('sky'));
+  background-color: var(--p-action-secondary, color('sky'));
   border-radius: border-radius();
 
   @media screen and (-ms-high-contrast: active) {
--- a/src/components/ProgressBar/ProgressBar.scss
+++ b/src/components/ProgressBar/ProgressBar.scss
@@ -20,7 +20,7 @@
 .ProgressBar {
   overflow: hidden;
   width: 100%;
-  background-color: var(--p-neutral, color('sky'));
+  background-color: var(--p-secondary-action, color('sky'));
   border-radius: border-radius();
 
   @media screen and (-ms-high-contrast: active) {
@@ -44,7 +44,7 @@
   height: inherit;
   width: 0;
   will-change: width;
-  background-color: var(--p-highlight, color('teal'));
+  background-color: var(--p-highlight-border, color('teal'));
   animation: fillup duration(slowest) easing();
   transition: width duration(slowest) easing();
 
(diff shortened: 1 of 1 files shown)


src/components/ProgressBar/ProgressBar.scss
```diff
@@ -20,7 +20,7 @@
 .ProgressBar {
   overflow: hidden;
   width: 100%;
-  background-color: color('sky', 'base');
+  background-color: var(--p-neutral-action, color('sky'));
```
danrosenthal:
```suggestion
  background-color: var(--p-surface-background, color('sky'));
```

Since this element is not interactive, let's use the surface background value for it.

src/components/ProgressBar/ProgressBar.scss
```diff
@@ -44,7 +44,7 @@
   height: inherit;
   width: 0;
   will-change: width;
-  background-color: color('teal');
+  background-color: var(--p-highlight, color('teal'));
```
danrosenthal: We want to avoid using the root variant, since it is unmodified from the original input.

```suggestion
  background-color: var(--p-highlight-border, color('teal'));
```

While it feels a bit strange using a border for this, it's the closest variant we have for this purpose.

src/components/ProgressBar/ProgressBar.scss
AndrewMusgrave: I was originally going to use `surface-background` however when `ProgressBar` is used on a page, without a card/modal/etc it blends in. 
Screen Recording 2019-12-09 at 01 24 PM

Is this intended?

danrosenthal: Good point.

Using the disabled neutral action for the background ought to do the trick. We've used this in `badge` as well for a similar application.

danrosenthal:
👋 Any help needed from me to get this one ready to merge? It'd be great to have it closed out

AndrewMusgrave:
@danrosenthal Thanks for checking in, sorry for the delay. I've been away since the Thursday before and break and was working on a few high priority issues that week. It should be good now!

src/components/ProgressBar/ProgressBar.scss
```diff
@@ -44,7 +44,7 @@
   height: inherit;
   width: 0;
   will-change: width;
-  background-color: color('teal');
+  background-color: var(--p-highlight-border, color('teal'));
```
danrosenthal:
```suggestion
  background-color: var(--p-border-highlight, color('teal'));
```

src/components/ProgressBar/ProgressBar.scss
```diff
@@ -20,7 +20,7 @@
 .ProgressBar {
   overflow: hidden;
   width: 100%;
-  background-color: color('sky', 'base');
+  background-color: var(--p-secondary-action, color('sky'));
```
danrosenthal:
```suggestion
  background-color: var(--p-action-secondary, color('sky'));
```

PR #2511, opened 2019-12-04
[Scrollable] Applied color system

AndrewMusgrave:

Fixes *partly* https://github.com/Shopify/polaris-ux/issues/362

### Screenies

Screen Shot 2019-12-04 at 12 59 39 PM
Screen Shot 2019-12-04 at 12 59 27 PM


AndrewMusgrave:
@dleroux Took your advice and added a color that can be altered later. How does it look?

src/utilities/theme/role-variants.ts
```diff
@@ -110,6 +110,22 @@ export const roleVariants: RoleVariants = {
...
+        lightness: 0,
+        alpha: 0.15,
+      },
+      dark: {
+        hue: 0,
+        saturation: 0,
+        lightness: 75,
+        alpha: 0.2,
+      },
+    },
```
danrosenthal: Since these are the same as the existing shadows, we don't need to add new ones. Use `shadowFromDirectLight` instead.

Adding new ones just for this use feels like premature optimization, since we have no indication that these will ever be different from existing shadows.

danrosenthal:
Anything I can do to help get this ready to merge? It'd be awesome to get this one in.

src/utilities/theme/role-variants.ts
AndrewMusgrave: They're slightly different. Between light & dark, `shadowFromDirectLight` has the same HSL values. `hintFromDirectLight` has different HSL values for dark, so the hint prop on scrollable will work in dark mode.

AndrewMusgrave:
@danrosenthal  Thanks for checking in, sorry for the delay. I've been away since the Thursday before and break and was working on a few high priority issues that week.

src/utilities/theme/role-variants.ts
danrosenthal: My bad, didn't notice the difference when I first looked!

PR #2160, opened 2019-09-18
[ResourceList] Add support for in context empty state

chloerice:
### WHY are these changes introduced?

Builds on #1570 

Right now our empty states are vastly different from our loading and with-data states. We'd like to make the transition from loading to empty state smoother, as well as make the with-data state more familiar once a feature has been used. For list views, this means having our empty states within the context of the resource list.

### WHAT is this pull request doing?

Currently, `ResourceList` only has an empty state for when there are no results for a filter or search query. This PR adds support for disabling filters and providing markup to render when there are not yet any resources to list. That way the same static content can be rendered in a loading, empty, and with-data state. See the tophatting instructions to view the smooth visual transition between these states using the playground code.

** This is a Foundations Frideations project.

## 

### How to 🎩

🖥 [Local development instructions](https://github.com/Shopify/polaris-react/blob/master/README.md#development)
🗒 [General tophatting guidelines](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting.md)
📄 [Changelog guidelines](https://github.com/Shopify/polaris-react/blob/master/documentation/Versioning%20and%20changelog.md)

View the "Resource list with empty state" example 

*_or_* 

- `git checkout resourcelist-emptystate`
- Copy and paste the playground code below into `playground/Playground.tsx`
- `yarn dev` to run and open Storybook locally
- Click "Playground" in the left navigation of Storybook and use the page's secondary actions to toggle between loading, empty, and with-data states.

<details>
<summary>Copy-paste this code in <code>playground/Playground.tsx</code>:</summary>

```jsx
import React, {useState, useCallback} from 'react';
import {
  Page,
  Layout,
  ResourceList,
  ResourceItem,
  EmptyState,
  Card,
  TextStyle,
  Select,
  Filters,
  Thumbnail,
  Tabs,
  SkeletonBodyText,
  Spinner,
} from '../src';

const allItems = [
  {
    id: 341,
    url: 'files/341',
    name: 'Point of sale staff',
    imgSource:
      'https://user-images.githubusercontent.com/29233959/61383074-fc5e6800-a87b-11e9-8e92-487153efd0dc.png',
    fileType: 'png',
  },
  {
    id: 256,
    url: 'files/256',
    name: 'File upload',
    imgSource:
      'https://user-images.githubusercontent.com/18447883/59945230-fa4be980-9434-11e9-9106-a373f0efbe08.png',
    fileType: 'png',
  },
];

const tabs = [
  {
    id: 'all-files',
[...]

PR #2514, opened 2019-12-05
Revert "Merge pull request #2160 from Shopify/resourcelist-emptystate"

chloerice:
This reverts commit 96db90311b89e0e64544b0be7bf6187ac095861f, reversing
changes made to e1506bdbcd7a32ed02e90d94f41ff41e35ebac0e.

Fixes #0000 

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist


dleroux:
The reason we are reverting this is that there is no fallback for current EmptyStates. List without this prop will lose their current empty state.

chloerice:
> The reason we are reverting this is that there is no fallback for current EmptyStates. List without this prop will lose their current empty state.

Edit: We reverted this enhancement because the current logic that decides whether to show the `emptyState` (when provided) vs the built in no results state requires a change in use of an existing prop (`hasMoreItems`), making it a breaking change. I'll be looking into how to configure this in a way that isn't a breaking change next week.

PR #2521, opened 2019-12-05
Update styleguide links

amrocha:

The styleguide is updating a lot of URLs

Updating the docs to reference the new links so we can remove the redirects over there


amrocha:
Yeah, there's a PR here https://github.com/Shopify/polaris-styleguide/pull/3129 that needs to be merged before this can go out

PR #2524, opened 2019-12-06
[TrapFocus] Convert TrapFocus to a functional component and fix tab focusing

LauraAubin:
### What problem is this solving?

- Converts `TrapFocus` to a functional component
- Fixes the issue where tabbing past the last element should focus the first element (and likewise)

### How is this accomplished?

Add event listeners for `keydown` and focus an element depending if shift is pressed or not. There are two separate functions `handleKeyDownLastElement` and `handleKeyDownFirstElement` because there is a distinction between the first and last element and which should react to the shift key.

### Notes

- Noticing some weird tabbing in Storybook on Safari, though the issue seems to exist on master as well. Pressing tab interacts with the search bar making it difficult to tab through the page content. This also doesn't seem to be an issue with specifically `TrapFocus` because it also opens the search bar with components such as `ButtonGroup`.
- `focusTrapWrapper` is the div around the modal

- The following functions find different scopes:

```js
const component = <div><modal>*</modal></div>;
findFirstFocusableNode(component) = will find the modal
findLastFocusableNode(component) = will find the content (*)
```

### To-do:

- [ ] Add more tests to address codecov
- [ ] 🎩on Safari outside of Storybook
- [ ] Add a test for finding the `firstElement` (if component to be trapped has vs does not have a container)

### DIFF
--- a/src/components/TrapFocus/TrapFocus.tsx
+++ b/src/components/TrapFocus/TrapFocus.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useState, useRef} from 'react';
 import {closest} from '@shopify/javascript-utilities/dom';
 import {
   focusFirstFocusableNode,
@@ -7,6 +7,7 @@ import {
 } from '@shopify/javascript-utilities/focus';
 import {write} from '@shopify/javascript-utilities/fastdom';
 
+import {useComponentDidMount} from '../../utilities/use-component-did-mount';
 import {EventListener} from '../EventListener';
 import {Focus} from '../Focus';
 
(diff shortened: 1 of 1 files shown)


src/components/TrapFocus/TrapFocus.tsx
```diff
@@ -15,81 +18,98 @@ export interface TrapFocusProps {
...
+    if (event.keyCode === Key.Tab && focusTrapWrapper.current) {
+      !event.shiftKey && focusFirstFocusableNode(focusTrapWrapper.current);
+
+      document.removeEventListener('keydown', handleKeyDownLastElement);
+    }
+  };
+
+  const handleKeyDownFirstElement = (event: any) => {
+    if (event.keyCode === Key.Tab && focusTrapWrapper.current) {
+      event.shiftKey && focusFirstFocusableNode(focusTrapWrapper.current);
```
LauraAubin: This should really be `event.shiftKey && focusLastFocusableNode(focusTrapWrapper.current)`. Yet this appears to "skip" the last element that we should be focusing.

It appears that we do focus the last element, and then <kbd>tab</kbd> + <kbd>shift</kbd> is also taken into consideration and back tabs a second time.

A work around for this is to focus the first element, and assuming that <kbd>tab</kbd> + <kbd>shift</kbd> is back tabbing on its own, focuses the last element.

src/components/TrapFocus/TrapFocus.tsx
```diff
@@ -15,81 +18,98 @@ export interface TrapFocusProps {
...
     if (trapping === false) {
       return;
     }
 
+    if (focusTrapWrapper.current) {
+      const firstElement = findFirstFocusableNode(
+        findFirstFocusableNode(
+          focusTrapWrapper.current as HTMLElement,
+        ) as HTMLElement,
+      );
```
LauraAubin: The following functions find different scopes:

```js
const component = <div><modal>*</modal></div>;
findFirstFocusableNode(component) = will find the modal
findLastFocusableNode(component) = will find the content (*)
```

This is why I am finding the first element by first finding the container, then the desired first element. Will definitely want a test for this.

AndrewMusgrave:
Haven't forgotten about this. @dleroux mentioned he was going to suggest some big changes so I'll hold off reviewing till now.

dleroux:
In reviewing your change I noticed why findFirstFiocusableNode wasn't working (modal wrapper div) In testing that I started going into a bit of a rabbit hole. https://github.com/Shopify/polaris-react/pull/2555. Tests are failing, but maybe we can pair on this later this week?

src/components/TrapFocus/TrapFocus.tsx
dleroux: The reason why findFirstFocusableNode finds the modal is because we added a tabIndex=-1 for accessibility so this workaround will not work for all scenarios. I've tried moving the divs around and it looks like ti would work.

LauraAubin: Ah good catch 🙈

LauraAubin:
Closing this PR in favour of https://github.com/Shopify/polaris-react/pull/2555 which is now based off of master.

PR #2527, opened 2019-12-09
[DropZone] Prevent dropzone border from showing on focus when disabled

### LINKED ISSUE #2390: Dropzone ignores outline prop when focused
## Issue summary
Currently, if the Dropzone's `outline` prop is explicitly set to `false`, the dashed outline will still appear when the component is focused in the DOM.

gif

## Expected behavior
The `outline` prop should be respected regardless of whether or not the component is focused.

## Actual behavior
The `outline` prop is being ignored during component focus.

1. Create a `<Dropzone />` component with `outline` set to `false`
2. Tab into the dropzone and see the outline appear when focused

## Steps to fix
It looks like all one would need to do, is wrap the Dropzone styles found here on line `#66`: https://github.com/Shopify/polaris-react/blob/master/src/components/DropZone/DropZone.scss#L66 inside of the `.hasOutline` class.

## Specifications
- Are you using the React components?: Yes
- Polaris version number: 4.7.1
- Browser: Chrome
- Device: MBP
- Operating System: OSX

maxariss:
Fixes #2390

We need to prevent the Dropzone from showing a border when the `outline` prop is false. 

Currently, it is still shown within the `.isDragging` and `.isDisabled` classes despite `outline` being false.

Scoping CSS within `.isDragging` and `.isDisabled` to only show a border when `outline` is true.

### Reproduction steps 🎩
- Add a dropzone component to your playground
- Set the `outline` prop to `false`
- Set `active` to `true` (the `active` prop triggers the `. isDragging` class)

### How to 🎩
- Follow the reproduction steps above
- Ensure no outline is shown around the DropZone

### Screenshots 📸 
gif

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -8,6 +8,7 @@
 
 ### Bug fixes
 
+- Fixed a bug where `DropZone` would show an outline on drag when `outline` is false ([#2527](https://github.com/Shopify/polaris-react/pull/2527))
 - Fixed a bug where `Navigation` calls `onNavigationDismiss` on large screens when focused and the escape key is pressed ([#2607](https://github.com/Shopify/polaris-react/pull/2607))
 - Fixed issue with `Filters` component displaying an undesired margin top and bottom on the button element on Safari ([#2292](https://github.com/Shopify/polaris-react/pull/2292))
 - Doesn't render `MenuActions` if no actions are passed to an `actionGroups` item inside `Page` ([#2266](https://github.com/Shopify/polaris-react/pull/2266))
--- a/src/components/DropZone/DropZone.scss
+++ b/src/components/DropZone/DropZone.scss
@@ -67,7 +67,7 @@ $dropzone-stacking-order: (
   &:not(.isDisabled) {
     background-color: $dropzone-overlay-background;
 
-    &::after {
+    &.hasOutline::after {
       border: border-width(thick) $dropzone-border-style
         $dropzone-overlay-border-color;
 
--- a/src/components/DropZone/DropZone.scss
+++ b/src/components/DropZone/DropZone.scss
@@ -83,6 +83,7 @@ $dropzone-stacking-order: (
   &:not(.isDisabled) {
     background-color: $dropzone-overlay-background;
 
+    // stylelint-disable-next-line selector-max-specificity
     &.hasOutline::after {
       border-color: $dropzone-overlay-border-color;
     }
--- a/src/components/DropZone/DropZone.scss
+++ b/src/components/DropZone/DropZone.scss
@@ -67,7 +67,7 @@ $dropzone-stacking-order: (
   &:not(.isDisabled) {
     background-color: $dropzone-overlay-background;
 
-    &.hasOutline::after {
+    &::after {
       border: border-width(thick) $dropzone-border-style
         $dropzone-overlay-border-color;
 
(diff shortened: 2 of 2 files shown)


dleroux:
Hi @maxariss. Was there a place where this was causing grief? We purposely added the outline for accessibility reasons because when there are files that have been dropped there's nothing indicating a focus on the drop zone. I don't think it's a hard rule but would want to make sure we're removing it for a good reason.

maxariss:
Hey @daniedleroux ! Yeah, it has caused a visual regression on the products page within the ImagesCard and MediaCard. It seems that now when dragging items within the Dropzone, it is causing focus to occur, which results in the following: 

gif

dleroux:
I see. Funny we were just talking about something similar this morning. Maybe on mouse up we blur? That way if the keyboard is being used for navigation the focus does appear. I haven’t looked but I hear we might be doing something similar on Button. What do you think?

maxariss:
Oh, no way! Well, good timing I guess haha. I think that would totally work for our use-case, or blurring on `mouseDown` anyways. 

It might also make sense to use a different/separate style here for the DropZone's focus state, as it feels like maybe the disableable dashed border is a separate concern. 

Maybe we could use a purple border like we do form elements? Or perhaps set the DropZone's background-color to `'purple', 'lighter'`? What do you think ?

dleroux:
> Maybe we could use a purple border like we do form elements?

I guess we'd have to see how that feels when we do have an outline. Dashes turning into an outline might feel a little strong?

maxariss:
>I guess we'd have to see how that feels when we do have an outline. Dashes turning into an outline might feel a little strong?

Ah you're right. That would muddy the waters eh. 

Either way, we should probably still ship this change or tweak the code to blur the DropZone on a mouseDown / dragOver event as it's causing a pretty jarring visual regression in production.

dleroux:
For sure. I’d go for blurring on mouse down if you have the time?

maxariss:
Happy New Years' @dleroux! I'm taking a look :) I might need a hand grepping how the focus / blur logic is handled in here though ~ TBD.

maxariss:
@dleroux I believe I have found a better CSS only approach, which leaves the border as is during focus state, but restricts the border color from changing within `.isDragging` and `.isDisabled` when `.hasOutline` is false (which I believe should be the intended behavior). Let me know what you think of this approach 👍🏻

maxariss:
@AndrewMusgrave adding you as a reviewer since I realized @dleroux is on 🌴Feel free to ping me if you need any context not provided in the PR description 👍

maxariss:
@chloerice @AndrewMusgrave adding you as reviewers since I realized @dleroux is on 🌴Feel free to ping me if you need any context not provided in the PR description 👍

AndrewMusgrave:
👋 How will this affect accessibility? Normally focus borders/outlines aren't something we want to remove :)

maxariss:
@AndrewMusgrave my bad – I should have updated the PR description based on the latest changes. I'm no longer removing the border on focus – I realized we really just need to prevent the border color from changing (thus showing a border) within the `.isDragging` and `.isDisabled` classes, when the `.hasOutline` class is not set. The tophat should still show a border on focus 👍I'll update the PR description as well (whoops).

AndrewMusgrave:
I'm still seeing the outline. What's the exact scenario we're trying to fix?

dleroux:
👋 same comment as @AndrewMusgrave, still seeing the border when dragging in Polaris. However, I'm not seeing it in web anymore. Has it been overwritten there?

maxariss:
Closing this PR based on IRL discussion with @dleroux. Products page found a way to side-step this issue while Polaris revisits the DropZone component states in a forthcoming sprint.

PR #2528, opened 2019-12-09
[Checkbox] apply design language to checkboxes

dleroux:

Partly addresses: https://github.com/Shopify/polaris-ux/issues/336 and https://github.com/Shopify/polaris-ux/issues/357

- Applies color system to checkbox
- Keeps the same logic as today therefore no mark-up change, but added selectors.
- Applies animation
- Radio buttons are not marked as complete in Figma. Showing 2 different styles for mobile and desktop. Not sure if that will remain the same but for now leaving as in. Not sure the reasoning with the 2 different styles. **Even though there are minor changes to button this PR is for Checkboxes**

checkbox

## 

### How to 🎩

<details>

```jsx
import {Page, Checkbox, Stack} from '../src';

  return (
    <Page title="Playground">
      <CheckboxExample />
    </Page>
  );
}

function CheckboxExample() {
  return (
    <React.Fragment>
      <Stack vertical>
        <Checkbox
          label="Basic checkbox"
          checked={checked}
          onChange={handleChange}
        />
        <Checkbox
          label="Error checkbox"
          checked={checked}
          onChange={handleChange}
          error
        />
        <Checkbox
          label="Error checkbox"
          checked={checked}
          onChange={handleChange}
          disabled
        />
      </Stack>
    </React.Fragment>
  );
}


```

</details>

### 🎩 checklist

### DIFF
--- a/documentation/Color system.md
+++ b/documentation/Color system.md
@@ -74,6 +74,7 @@ The interactive role is used to express interactivity in components. It is used 
 | <pre>--p-surface-selected</pre>            | For use as a surface color in selected interactive elements, in components such as option list and resource list.                  | ![](https://www.gifpng.com/64x64/f1f4fe/FFFFFF?border-width=16&border-type=rectangle&border-color=fafafa&text=%20)<br />#f1f4fe | ![](https://www.gifpng.com/64x64/020e22/FFFFFF?border-width=16&border-type=rectangle&border-color=0c0d0e&text=%20)<br />#020e22 |
 | <pre>--p-surface-selected-hovered</pre>    | For use as a surface color in selected interactive elements that are hovered, in components such as option list and resource list. | ![](https://www.gifpng.com/64x64/d4defc/FFFFFF?border-width=16&border-type=rectangle&border-color=fafafa&text=%20)<br />#d4defc | ![](https://www.gifpng.com/64x64/071c3b/FFFFFF?border-width=16&border-type=rectangle&border-color=0c0d0e&text=%20)<br />#071c3b |
 | <pre>--p-surface-selected-pressed</pre>    | For use as a surface color in selected interactive elements that are pressed, in components such as option list and resource list. | ![](https://www.gifpng.com/64x64/b9c9f9/FFFFFF?border-width=16&border-type=rectangle&border-color=fafafa&text=%20)<br />#b9c9f9 | ![](https://www.gifpng.com/64x64/0d2a54/FFFFFF?border-width=16&border-type=rectangle&border-color=0c0d0e&text=%20)<br />#0d2a54 |
+| <pre>--p-icon-on-interactive</pre>         | For use as a fill color for icons on interactive elements.                                                                         | ![](https://www.gifpng.com/64x64/ffffff/FFFFFF?border-width=16&border-type=rectangle&border-color=fafafa&text=%20)<br />#ffffff | ![](https://www.gifpng.com/64x64/ffffff/FFFFFF?border-width=16&border-type=rectangle&border-color=0c0d0e&text=%20)<br />#ffffff |
 
 ---
 
@@ -89,6 +90,9 @@ A secondary interactive color role, for use in secondary and tertiary buttons as
 | <pre>--p-action-secondary-disabled</pre> | Used as a disabled state for secondary buttons                                                                                                | ![](https://www.gifpng.com/64x64/ecedee/FFFFFF?border-width=16&border-type=rectangle&border-color=fafafa&text=%20)<br />#ecedee | ![](https://www.gifpng.com/64x64/202222/FFFFFF?border-width=16&border-type=rectangle&border-color=0c0d0e&text=%20)<br />#202222 |
 | <pre>--p-action-secondary-hovered</pre>  | Used as a hovered state for secondary buttons                                                                                                 | ![](https://www.gifpng.com/64x64/e1e3e5/FFFFFF?border-width=16&border-type=rectangle&border-color=fafafa&text=%20)<br />#e1e3e5 | ![](https://www.gifpng.com/64x64/53565a/FFFFFF?border-width=16&border-type=rectangle&border-color=0c0d0e&text=%20)<br />#53565a |
 | <pre>--p-action-secondary-pressed</pre>  | Used as a pressed state for secondary buttons                                                                                                 | ![](https://www.gifpng.com/64x64/d9dbde/FFFFFF?border-width=16&border-type=rectangle&border-color=fafafa&text=%20)<br />#d9dbde | ![](https://www.gifpng.com/64x64/5f6467/FFFFFF?border-width=16&border-type=rectangle&border-color=0c0d0e&text=%20)<br />#5f6467 |
+| <pre>--p-border-secondary</pre>          | Used for borders on form elements                                                                                                             | ![](https://www.gifpng.com/64x64/b5babf/FFFFFF?border-width=16&border-type=rectangle&border-color=fafafa&text=%20)<br />#b5babf | ![](https://www.gifpng.com/64x64/515458/FFFFFF?border-width=16&border-type=rectangle&border-color=0c0d0e&text=%20)<br />#515458 |
+| <pre>--p-border-secondary-hovered</pre>  | Used for borders on hovered form elements                                                                                                     | ![](https://www.gifpng.com/64x64/989ea4/FFFFFF?border-width=16&border-type=rectangle&border-color=fafafa&text=%20)<br />#989ea4 | ![](https://www.gifpng.com/64x64/515458/FFFFFF?border-width=16&border-type=rectangle&border-color=0c0d0e&text=%20)<br />#515458 |
+| <pre>--p-border-secondary-disabled</pre> | Used for disabled borders on form elements                                                                                                    | ![](https://www.gifpng.com/64x64/d3d6d9/FFFFFF?border-width=16&border-type=rectangle&border-color=fafafa&text=%20)<br />#d3d6d9 | ![](https://www.gifpng.com/64x64/676b6f/FFFFFF?border-width=16&border-type=rectangle&border-color=0c0d0e&text=%20)<br />#676b6f |
 
 ---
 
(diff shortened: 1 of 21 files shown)


src/components/Checkbox/Checkbox.scss
```diff
@@ -1,82 +1,121 @@
...
   @include visually-hidden;
 
-  &:focus,
-  &:active,
-  &:checked,
-  &.Input-indeterminate {
-    // stylelint-disable-next-line selector-max-class
-    + .Backdrop {
-      background: color('indigo');
+  &:not(.globalTheming) {
```
dleroux: Kept the `.globalTheming` and `:not(.globalTheming)` completely separate. Because these are quite different this will make it much easier to remove the old code once this is complete.

src/styles/foundation/_focus-ring.scss
```diff
@@ -27,6 +29,21 @@
   }
 }
 
+@mixin focus-ring-animated($size: 'base', $show: 0) {
+  $easing: cubic-bezier(0.4, 0.22, 0.28, 1);
```
dleroux: I wasn't sure what to do with these easing values. Wonder if we keep them apart like this until we know the final functions needed in the system?

dleroux: And the focus ring around the checkbox and radio are animated which is why I added this mixin.

src/styles/shared/_controls.scss
```diff
@@ -11,7 +11,48 @@
 }
 
 @function control-icon-transition() {
-  @return transform duration(fast) easing(in) easing(out);
+  @return transform duration(fast) easing(in);
+}
+
+@mixin control-backdrop-dl($style: base) {
+  @if $style == base {
```
dleroux: created a new mixin again to keep things simple.

dleroux:
@jessebc are the 2 different styles for mobile and desktop intentional?

dleroux:
@johanstromqvist in speaking with @alex-page he mentioned you don't want to show the focus ring when a user clicks. Is that a hard rule? I feel if the element is focused it should sow the focus ring or it might be confusing/unexpected. Also to get it that behavior done would mean hacking what the browser would do by default which is also a red flag.  

This PR also applies the animation from the prototype but with slight differences because the mark-up is not the same. Let me know how it feels.

dleroux:
I'll fix up the coverage but still looking for feedback on the approach.

documentation/Color system.md
```diff
@@ -75,6 +75,7 @@ The interactive role is used to express interactivity in components. It is used
...
 | <pre>--p-interactive-selected-pressed</pre> | For use as a surface color in selected interactive elements that are pressed, in components such as option list and resource list. `-inverse`, `-light`, and `-dark` variants available. | ![](https://www.gifpng.com/64x64/b9c9f9/FFFFFF?border-width=16&border-type=rectangle&border-color=fafafa&text=%20)<br />#b9c9f9 | ![](https://www.gifpng.com/64x64/0d2a54/FFFFFF?border-width=16&border-type=rectangle&border-color=0c0d0e&text=%20)<br />#0d2a54 |
+| <pre>--p-icon-on-interactive</pre>          | For use as a fill color for icons on critical actions.                                                                                                                                   | ![](https://www.gifpng.com/64x64/ffffff/FFFFFF?border-width=16&border-type=rectangle&border-color=fafafa&text=%20)<br />#ffffff | ![](https://www.gifpng.com/64x64/ffffff/FFFFFF?border-width=16&border-type=rectangle&border-color=0c0d0e&text=%20)<br />#ffffff |
```
kyledurand: I think we can replace `critical actions` with `interactive elements`

documentation/Color system.md
```diff
@@ -236,6 +240,8 @@ Used to decorate elements where color does convey a specific meaning in componen
 | `--p-override-loading-z-index`            | `514`                                                                                             |
 | `--p-button-font-weight`                  | `500`                                                                                             |
 | `--p-non-null-content`                    | `''`                                                                                              |
+| `--p-choice-size`                         | `2rem`                                                                                            |
+| `--p-choice-margin`                       | `0.1rem`                                                                                          |
```
kyledurand: This seems a little odd to me at first glance. These variables seem like they would only be used within the one component, not globally for other components / layouts. I could be wrong though

src/components/Checkbox/Checkbox.scss
```diff
@@ -1,82 +1,121 @@
...
+    top: 0;
+    bottom: 0;
+    left: 0;
+    right: 0;
+    pointer-events: none;
+    @include control-backdrop;
+  }
+
+  &.globalTheming {
+    @include control-backdrop-dl;
```
kyledurand: Is suffixing -dl an existing pattern already? Wondering if there's a better naming convention we could use or even start changing legacy mixin names so we're more clear of which ones should be avoided

src/utilities/theme/role-variants.ts
```diff
@@ -240,6 +240,12 @@ export const roleVariants: RoleVariants = {
       light: {lightness: 82},
       dark: {lightness: 18},
     },
+    {
+      name: 'iconOnInteractive',
+      description: 'For use as a fill color for icons on critical actions.',
```
kyledurand: I think we can replace critical actions with interactive elements here as well

src/utilities/theme/role-variants.ts
```diff
@@ -240,6 +240,12 @@ export const roleVariants: RoleVariants = {
...
+      light: {lightness: 100},
+      dark: {lightness: 100},
```
kyledurand: Seems like one of these values should be adjusted. Usually their values should be close to eachothers inverse color. i.e. if light has a lightness of 80, then dark would have a lightness of 20.

src/styles/foundation/_focus-ring.scss
```diff
@@ -27,6 +29,21 @@
   }
 }
 
+@mixin focus-ring-animated($size: 'base', $show: 0) {
```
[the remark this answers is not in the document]
dleroux: My thinking was this would offer the same features as the non-animated focus ring would.

jessebc:
> @jessebc are the 2 different styles for mobile and desktop intentional?

Yes and no. We were prototyping out 2 different directions in browser to see which direction we felt more confident in. I'll update the figma file to align with our findings.

documentation/Color system.md
dleroux: I think we'll be able to delete this one in the end. This margin doesn't exist today which is why I added it here. It could simply be added via the classname as well. I'll do that.

src/components/Checkbox/Checkbox.scss
dleroux: Changing our current mixins would be a breaking change. What do you think about putting inside

`/polaris-react/src/styles/_common.scss` which is part of our private API so that when we remove the `control-backdrop` mixin, we can simply drop the `-dl` and worry about breaking anyone?

documentation/Color system.md
dleroux: > > @jessebc are the 2 different styles for mobile and desktop intentional?
> 
> Yes and no. We were prototyping out 2 different directions in browser to see which direction we felt more confident in. I'll update the figma file to align with our findings.

The version you are seeing in this PR makes it so that both checkbox and radio share the same CSS if that influences the decision. It probably shouldn't but if it comes to a coin toss.

src/utilities/theme/role-variants.ts
dleroux: In this case, the icon is white in both cases. Do we have other cases like this?

src/components/RadioButton/RadioButton.tsx
```diff
@@ -1,5 +1,7 @@
 import React from 'react';
 import {useUniqueId} from '../../utilities/unique-id';
+import {useFeatures} from '../../utilities/features';
```
BPScott: Really petty nitpick: Can imports at a certain level be alphabetically sorted?

(A autofixable rule for this recently got added to eslint-plugin-import and IIRC other folks were also hoping for it so I expect it'll be added to the shopify eslint config at some point)

src/styles/foundation/_focus-ring.scss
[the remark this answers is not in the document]
BPScott: If we make don't expose this as part of our public API then we're free to only add it when we need it.

documentation/Color system.md
tmlayton: Yes, these variables are temporary.

johanstromqvist:
> @johanstromqvist in speaking with @alex-page he mentioned you don't want to show the focus ring when a user clicks. Is that a hard rule? I feel if the element is focused it should sow the focus ring or it might be confusing/unexpected. Also to get it that behaviour done would mean hacking what the browser would do by default which is also a red flag.
> 
> This PR also applies the animation from the prototype but with slight differences because the mark-up is not the same. Let me know how it feels. The way it's set up it more performant than the prototype.

Thanks for ping @dleroux! No, I wouldn't call it a hard rule, but we are exploring what elements that benefit from it and what don't. There are a lot of interactions with elements where the focus state adds noise instead of signal, and it's common to skip focus states on them. E.g. when mouse clicking a radio, your attention is already there, the action is already finished, giving it a focus state doesn't perform any useful job. Same e.g. with checkbox and button, when the click is done the action is done. I'm not worried about it being unexpected, but it's definitely open for discussion.

I agree that sticking to browser defaults is a win and has a lot of benefits, but not that it's a red flag to change them for good reasons. The default don't reflect best practices from a UX POV, so we should definitely question them.

src/styles/foundation/_focus-ring.scss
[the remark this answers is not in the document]
tmlayton: The focus ring will have to be public in some way, whether that ends up coming from polaris-react or polaris-tokens (although a mixin in tokens is questionable). Teams building custom components with focus styles will want to use the same focus ring.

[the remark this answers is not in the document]
tmlayton: Should the animation be added to all rings?

src/components/Checkbox/Checkbox.scss
```diff
@@ -1,82 +1,121 @@
...
+        @include focus-ring-animated($show: 1);
+      }
+    }
+    // stylelint-disable-next-line selector-max-specificity
+    &:active:not(:disabled),
+    &:checked,
+    &.Input-indeterminate {
+      // stylelint-disable-next-line selector-max-specificity
+      + .Backdrop {
+        @include control-backdrop-dl(active);
```
danrosenthal: Instead of a new design-language-focused mixin, what about creating a second argument to the existing `control-backdrop` mixin like `new_design_language` that defaults to false? You could use that as a switch to enable the new styles.

src/components/Checkbox/Checkbox.scss
```diff
@@ -1,82 +1,121 @@
...
+    @include control-backdrop;
+  }
+
+  &.globalTheming {
+    @include control-backdrop-dl;
+    position: relative;
+    display: block;
+    width: 100%;
+    height: 100%;
+    @include focus-ring-animated;
```
danrosenthal: It doesn't make a lot of sense to me to have a separate mixin for the animated focus ring, since all components that use the focus ring will need it to be animated. I'd like to see one mixin that works for all components which use the focus ring (Button and TextField, for example).

src/components/RadioButton/RadioButton.scss
```diff
@@ -2,56 +2,121 @@
...
   }
 
-  &:disabled {
-    + .Backdrop {
-      @include control-backdrop(disabled);
+  &.globalTheming {
+    &:focus {
+      // stylelint-disable-next-line selector-max-specificity
+      + .Backdrop {
+        @include focus-ring-animated($show: 1);
```
danrosenthal: This API feels a bit strange to me. What about setting it to accept states, similar to how the `control-backdrop` mixin operates. That way you could pass `focused` as a state to indicate that it should be visible.

src/styles/foundation/_focus-ring.scss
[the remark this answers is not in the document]
danrosenthal: I'd like to see this refactored so that all components which use the focus ring can consume the same mixin.

I'd also like to see some indication of whether this mixin works for `Button` and `TextField` as well as `Checkbox` and `Radio`.

src/styles/shared/_controls.scss
danrosenthal: As I indicated before, might it make more sense to stick with the same mixin, but add a property for turning on the new design language?

src/components/Checkbox/Checkbox.tsx
```diff
@@ -141,13 +165,13 @@ export const Checkbox = React.forwardRef<CheckboxHandles, CheckboxProps>(
             onFocus={onFocus}
             onBlur={onBlur}
             onClick={stopPropagation}
-            onChange={noop}
+            onChange={handleBlur}
```
dleroux: This makes the focus ring not appear when clicking with the mouse. With Radio buttons, this will be tricky because blurring the input prevents the arrows from navigating to the next radio in the group.

src/components/Choice/Choice.tsx
```diff
@@ -23,6 +21,10 @@ export interface ChoiceProps {
   helpText?: React.ReactNode;
   /** Callback when clicked */
   onClick?(): void;
+  /** Callback when mouse over */
+  onMouseOver?(): void;
+  /** Callback when mouse out */
+  onMouseOut?(): void;
```
dleroux: This is to allow the label to trigger the hover on the checkbox.

src/styles/foundation/_focus-ring.scss
```diff
@@ -2,8 +2,8 @@
 /// @param {String} $size - The size of the border radius on the focus ring.
 ///
 
-@mixin focus-ring($size: 'base') {
-  $offset: rem(1px);
+@mixin focus-ring($size: 'base', $offset: rem(1px), $show: 0) {
```
dleroux: Needed to add the $offset, because when the element has a border, the `:after` positioning of `top left` changes by the width of the border. This offset will the width of the border + the 1 px gap.

src/styles/shared/_controls.scss
```diff
@@ -11,87 +11,155 @@
 }
 
 @function control-icon-transition() {
-  @return transform duration(fast) easing(in) easing(out);
+  @return transform duration(fast) easing(in);
```
dleroux: This transition wasn't working because of the in and out

src/styles/shared/_controls.scss
```diff
@@ -11,87 +11,155 @@
...
-      right: 0;
-      display: block;
-      background: linear-gradient(
-        to bottom,
-        color('white'),
-        color('sky', 'lighter')
-      );
-      border-radius: border-radius() - 1px;
-    }
+@mixin control-backdrop($style: base, $global-theming: false) {
```
dleroux: added a $global-theming arg.

documentation/Color system.md
```diff
@@ -74,6 +74,7 @@ The interactive role is used to express interactivity in components. It is used
...
 | <pre>--p-surface-selected-pressed</pre>    | For use as a surface color in selected interactive elements that are pressed, in components such as option list and resource list. | ![](https://www.gifpng.com/64x64/b9c9f9/FFFFFF?border-width=16&border-type=rectangle&border-color=fafafa&text=%20)<br />#b9c9f9 | ![](https://www.gifpng.com/64x64/0d2a54/FFFFFF?border-width=16&border-type=rectangle&border-color=0c0d0e&text=%20)<br />#0d2a54 |
+| <pre>--p-icon-on-interactive</pre>         | For use as a fill color for icons on interactive elements.                                                                         | ![](https://www.gifpng.com/64x64/ffffff/FFFFFF?border-width=16&border-type=rectangle&border-color=fafafa&text=%20)<br />#ffffff | ![](https://www.gifpng.com/64x64/ffffff/FFFFFF?border-width=16&border-type=rectangle&border-color=0c0d0e&text=%20)<br />#ffffff |
```
dleroux: This is a white icon on blue. Would this be the right colour to add?

dleroux:
@martenbjork @johanstromqvist I think I got it exactly like your prototype including the checkbox being that much slower than the background. Do you mind pulling this branch and giving the checkbox a test?

martenbjork:
@daniedleroux Super! Starting to look really good. Two tweaks:

- We made some minor teaks to the transition duration and easing values. [Updated values](https://docs.google.com/document/d/1mm8KSWjFuO3eX2Nfx4SsxvIjyqdqNUGONxdfHP3tfTg/edit)

- I realized that the prototype is missing one feature; the outline should become blue, instantaneously when the element is :checked. (Right now, the background simply overflows the gray outline.)

src/styles/shared/_controls.scss
```diff
@@ -11,87 +11,155 @@
...
+        top: calc(var(--control-border-width) * -1);
+        right: calc(var(--control-border-width) * -1);
+        bottom: calc(var(--control-border-width) * -1);
+        left: calc(var(--control-border-width) * -1);
+        border-radius: var(--p-border-radius-base);
+        background-color: var(--p-action-interactive);
+        opacity: 0;
+        transform: scale(0.25);
+        transition: opacity var(--p-duration-1-0-0) linear,
+          transform var(--p-duration-1-0-0) linear;
```
kyledurand: Two things about these two lines. 

1. It looks like the prototype is using 150ms for check mark transition
1. It's also using a bezier curve for the function `cubic-bezier(0.4, 0.22, 0.28, 1)`

src/styles/shared/_controls.scss
```diff
@@ -11,87 +11,155 @@
...
+        content: '';
+        position: absolute;
+        top: calc(var(--control-border-width) * -1);
+        right: calc(var(--control-border-width) * -1);
+        bottom: calc(var(--control-border-width) * -1);
+        left: calc(var(--control-border-width) * -1);
+        border-radius: var(--p-border-radius-base);
+        background-color: var(--p-action-interactive);
+        opacity: 0;
+        transform: scale(0.25);
```
kyledurand: The prototype starts at `scale(0.1)`

dleroux: There have been changes tracked in a document. I'll slack it to you.

dleroux:
I've updated the animation values to match the doc @martenbjork 

I also applied the focus-ring animation to all the components that had the ring. Sadly the animation breaks 2 things:

1. In order to trigger the animation, the ring needs to be present but not visible when not focused. This causes an issue with the `:focus::not(.active) selector`. You can actually see the ring flicker.

2. We cannot use the `blur()` option in the prototype for accessibility reasons. This is especially obvious with radio buttons. Once you click to the first element you can no longer use the arrow keys to navigate to the next elements.

I'm sure we'll find a solution but for now I left it like this so we can all see the issue and not forget it.

danrosenthal:
Sounds like we may just need to wait on the `:focus-visible` selector to get broader browser support before we can achieve the style we want with regards to active vs focus states.

- https://caniuse.com/#search=%3Afocus-visible
- https://css-tricks.com/focusing-on-focus-styles/

alex-page:
I would like to decouple this PR's _focus_ on checkboxes. What do people think of moving the focus ring to another PR?

kyledurand:
I was thinking the same thing @danrosenthal. Let's not let this block us while Dan L is away next week. I can probably take some time aside next week to make sure this approach works with all other components

src/components/Checkbox/Checkbox.scss
```diff
@@ -1,91 +1,173 @@
...
+    &:checked,
+    &.Input-indeterminate {
+      + .Backdrop {
+        @include control-backdrop(active, $global-theming: true);
+      }
+      ~ .Icon {
+        transition-timing-function: var(--p-ease-in);
+        transform: translate(-50%, -50%) scale(1);
+        opacity: 1;
+        transition: opacity var(--p-duration-1-5-0) var(--p-ease),
```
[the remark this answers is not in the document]
dleroux: Yes, i was meant to bring this up yesterday in our pauring

dleroux:
I would opt to leaving it and go from there. Fix the bug after. We’ll need to add animation which will bring this bug. My 2 cents

danrosenthal:
To clarify, I think we are all suggesting you ship the PR as-is, with the focus ring included. Alex was suggesting we reserve feedback on this PR for only checkbox, which I agree with. And Kyle and I were saying that you can ship this, and we can iterate on the focus ring later, with the work you've done as the foundation.

Get this passing CI, and merge the PR as-is.

PR #2530, opened 2019-12-10
[TrapFocus] Don't trap the focus if the relatedTarget is null

LauraAubin:

Closes https://github.com/Shopify/polaris-react/issues/2531

We initially removed this line so that we could trap focus within a container, even if focus leaves the page content (or an iframe).

Although this is not a breaking change on its own, it causes a bug with the style guide which disallows the user to interact with the rest of the page such as the examples drop down and the search bar.

We'll be fixing `TrapFocus` in another PR which handles keyboard inputs instead of handling when focus leaves the container.

## 

### How to 🎩

Follow [these](https://github.com/Shopify/polaris-react/blob/6579045fbbf668b8f1624631cd5c7f609f4bff57/documentation/Tophatting%20documentation.md#how-to--documentation) instructions to test the style guide.

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -6,6 +6,8 @@
 
 ### Bug fixes
 
+- Fixed a regression in `TrapFocus` that prevented focus outside of an `iframe` ([#2530](https://github.com/Shopify/polaris-react/pull/2530))
+
 ### Documentation
 
 ### Development workflow
--- a/src/components/TrapFocus/tests/TrapFocus.test.tsx
+++ b/src/components/TrapFocus/tests/TrapFocus.test.tsx
@@ -8,10 +8,6 @@ import {
   TextField,
   Button,
 } from 'components';
-import {
-  findFirstFocusableNode,
-  findLastFocusableNode,
-} from '@shopify/javascript-utilities/focus';
 import {TrapFocus} from '../TrapFocus';
 
 describe('<TrapFocus />', () => {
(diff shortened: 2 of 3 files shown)


src/components/TrapFocus/tests/TrapFocus.test.tsx
```diff
@@ -96,49 +96,11 @@ describe('<TrapFocus />', () => {
...
-        <TextField label="" value="" onChange={noop} autoFocus />
-      </TrapFocus>,
-    );
-
     const event: FocusEvent = new FocusEvent('focusout', {
       relatedTarget: externalDomNode,
     });
     Object.assign(event, {preventDefault: jest.fn()});
 
-    describe('prevents default when focus moves to an external node', () => {
```
LauraAubin: We'll no longer handle focus when `relatedTarget` is null. Instead, a later PR will handle keyboard inputs such as <kbd>tab</kbd> and focusing elements manually.

UNRELEASED.md
```diff
@@ -6,6 +6,8 @@
 
 ### Bug fixes
 
+- Fixed a bug where focus should not be trapped if focus moves outside of the page content ([#2530](https://github.com/Shopify/polaris-react/pull/2530))
```
```suggestion
- Fixed a regression in ` TrapFocus` that prevented focus outside of an `iframe` ([#2530](https://github.com/Shopify/polaris-react/pull/2530))
```

src/components/TrapFocus/tests/TrapFocus.test.tsx
dleroux: Was this test failing? Should we remove it here or in your next PR?

LauraAubin: It was failing because we no longer run anything past the early return:

https://github.com/Shopify/polaris-react/blob/5731ce8f5e09960100010677c00d17e0407e8395/src/components/TrapFocus/TrapFocus.tsx#L76-L78

Without the early return, if `relatedTarget` was null we would still manage and set focus. So we should return it here.

UNRELEASED.md
LauraAubin: So much better 💜

PR #2532, opened 2019-12-10
[Dropzone] Removed reference to window

AndrewMusgrave:

Removing all reference to `window` on code that can execute on the client

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -6,6 +6,8 @@
 
 ### Bug fixes
 
+- Removed reference to `window` in `DropZone` ([#2532](https://github.com/Shopify/polaris-react/pull/2532))
+
 ### Documentation
 
 ### Development workflow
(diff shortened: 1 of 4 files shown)


BPScott:
Could you say why this is needed? I'm at a complete loss why this might be needed - I'd love to learn what weird case I'm overlooking.

dleroux:
To be honest we are unsure. We had to roll back 4.9 from web this morning because of bugsnag error claiming window was undefined.

Do you know how we could test this with SSR before shipping?

AndrewMusgrave:
> Could you say why this is needed? I'm at a complete loss why this might be needed - I'd love to learn what weird case I'm overlooking.

Was going to ask Dan to paste in the bugsnag link cc/ @dleroux 

We're referencing window on the client `useEventListener(window` which doesn't fly in ssr land. This change will make it easier attaching event listeners to the window in the future

dleroux:
```
node_modules/@shopify/polaris/esnext/components/DropZone/DropZone.js:139
window is not defined
```

BPScott:
Had a chat wilth @dleroux: This is weird, I'm not a  fan and I wonder if using isServer we can avoid adding options to uesEventListener. That said this is internal stuff and we can come back to it later, finding a cleaner solution shouldn't block release of this.

PR #2534, opened 2019-12-11
[Stack] Converted to a functional component

AndrewMusgrave:

### DIFF
--- a/src/components/Stack/Stack.tsx
+++ b/src/components/Stack/Stack.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {memo, NamedExoticComponent} from 'react';
 import {classNames, variationName} from '../../utilities/css';
 import {elementChildren, wrapWithComponent} from '../../utilities/components';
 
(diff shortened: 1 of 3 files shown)


BPScott:
Given my talk in https://github.com/Shopify/polaris-react/issues/2062 about our cargo-culty usage of PureComponent and how I'd like us to stop doing that I don't think we should bother wrapping this in a React.memo

AndrewMusgrave:
For other readers, response to Ben [here](https://github.com/Shopify/polaris-react/pull/2539/#issuecomment-571275058)

src/components/Stack/Stack.tsx
```diff
@@ -32,33 +32,31 @@ export interface StackProps {
...
+    wrap === false && styles.noWrap,
+  );
+
+  const itemMarkup = elementChildren(children).map((child, index) => {
+    const props = {key: index};
+    return wrapWithComponent(child, Item, props);
+  });
+
+  return <div className={className}>{itemMarkup}</div>;
+}) as NamedExoticComponent<StackProps> & {
```
BPScott: In other cases where we need this recasting it's been done at the top of the constant declaration rather than using an `as`, see [Autocomplete for any example](https://github.com/Shopify/polaris-react/blob/master/src/components/Autocomplete/Autocomplete.tsx#L46). Can we keep to that style for consistency?


```js
export const Stack: React.NamedExoticComponent<StackProps> & {
  Item: typeof Item;
} = function memo(function Stack({
//...
```

src/components/Stack/Stack.tsx
```diff
@@ -32,33 +32,31 @@ export interface StackProps {
...
-
-    const itemMarkup = elementChildren(children).map((child, index) => {
-      const props = {key: index};
-      return wrapWithComponent(child, Item, props);
-    });
-
-    return <div className={className}>{itemMarkup}</div>;
-  }
-}
+export const Stack = memo(function Stack({
```
BPScott: Per my thoughts in https://github.com/Shopify/polaris-react/issues/2062, I'm not a fan of us using memo everywhere, at it feels like our usage of that and PureComponent has been very cargo-culty and we've been doing it without measuring its impact.

What do you think to removing the `memo()` wrapper?

src/components/Stack/Stack.tsx
AndrewMusgrave: Typing is always preferred overcasting, however, unfortunately, we can't in this situation. 
```ts
Type 'NamedExoticComponent<StackProps>' is not assignable to type 'NamedExoticComponent<StackProps> & { Item: ({ children, fill }: ItemProps) => Element; }'.
  Property 'Item' is missing in type 'NamedExoticComponent<StackProps>' but required in type '{ Item: ({ children, fill }: ItemProps) => Element; }'
```

src/components/Stack/Stack.tsx
AndrewMusgrave: Left my thoughts in this [post](https://github.com/Shopify/polaris-react/pull/2539#issuecomment-571275058)

src/components/Stack/Stack.tsx
BPScott: well that's a little annoying :(

/me waves fist at React.memo

src/components/Stack/Stack.tsx
BPScott: Thanks for reckons :) Happy to say removing PureComponent / React.memo bits can be kept out of the hookification and they can be looked at as a separate phase.

PR #2537, opened 2019-12-11
[Section] Converted to a functional component

AndrewMusgrave:

### DIFF
--- a/src/components/Navigation/components/Section/tests/Section.test.tsx
+++ b/src/components/Navigation/components/Section/tests/Section.test.tsx
@@ -1,11 +1,11 @@
 import React from 'react';
-
 import {matchMedia, animationFrame} from '@shopify/jest-dom-mocks';
 // eslint-disable-next-line no-restricted-imports
 import {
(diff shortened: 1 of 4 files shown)


src/components/Navigation/components/Section/tests/Section.test.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-
+import {act} from 'react-dom/test-utils';
```
BPScott: What do you think to reexporting act in test-legacy instead of adding imports from `react-dom/test-utils`?

So we'd end up with

```
import {
  findByTestID,
  trigger,
  act,
  mountWithAppProvider,
} from 'test-utilities/legacy';

AndrewMusgrave: I'm ok with that, it will clean up our imports. I'll make the change when I add a test later to bump `codecov`.

AndrewMusgrave:
codecov: `89.47%` 😆

Issue #2062, opened 2019-08-29
Proposal: Audit PureComponent usage and potentially stop that


BPScott:
Part of the post-v4 work is to update our examples to use hooks. @AndrewMusgrave made some fantastic progress on this and @amrocha has been reviewing it and he brought up something interesting: Pervasive usage of useCallback() makes them a bit crap and noisy to read.

Killing the useCallbacks would make for easier to read examples but there's a gotcha: The majority of our components use PureComponent and not using useCallback would mean they always fail their memoisation check (as you end up passing in a new function every time) - meaning a mechanism for supposed performance improvement actually becomes harmful as you always do more work with no chance of hitting the short-circuit path.

A component has two choices and we must be consistent if we want to showcase the correct way to use a component:

- Don't use React.PureComponent/React.memo in the component, and don't use useCallback to wrap functions passed into the component (as there is no shallow compare, it is preferable to avoid the small overhead of memoification)
- Use React.PureComponent/React.memo in the component and use useCallback to wrap functions passed into the component (as the wrapping allows the memoisation to work)

A lot of our components use PureComponent which make it seem like the second way is the right way to go but I'm not convinced that a most of our usage of PureComponent is actually justified.

---

A chat with @TheMallen in #web-foundation-tech suggests that their consensus is to avoid using  `React.PureComponent` and `React.memo` until you know you need them and are sure there is a meaningful performance improvement thanks to profiling information. This approach to profiling and validating before jumping on a path with a different performance tradeoff is echoed by https://kentcdodds.com/blog/usememo-and-usecallback and https://reacttraining.com/blog/react-inline-functions-and-performance/ .

It looks like the overwhelming majority of our PureComponent usage is cargo-culty - we've slapped it onto components without ensuring it actually helps as many of the components in question are leaf nodes that often will have very little complex content inside them (e.g. Avatar, Banner, TextField).

I would like to take some time to audit our usage of PureComponent on all our components to validate our assumption that PureComponent makes a positive difference. If this turns out to be unfounded then we should remove PureComponent on the components that do not benefit from it.

// @cc @chloerice @dleroux @danrosenthal and @tmlayton for our earlier conversations on this

BPScott:
This also has knock-on effects regarding our hookification work.

For instance consider the hookification of `ActionMenu/components/ RollupActions/RollupActions.tsx`.

[...]

PR #2539, opened 2019-12-11
[FormLayout] Converted to a functional component

AndrewMusgrave:


BPScott:
Given my talk in #2062 about our cargo-culty usage of PureComponent and how I'd like us to stop doing that I don't think we should bother wrapping this in a React.memo

AndrewMusgrave:
@BPScott For this pull-request, I think I'm going to keep `React.memo`. It's a migration from a class to a functional component, and since `FormLayout` extends `React.PureComponent` we would be removing functionality.

However I'm not disagreeing that we probably don't need `React.memo` on `FormLayout`, the component is so simple a render is production would be negligible.

Before we start pulling `React.PureComponent & React.Memo` from our components we should probably come up with a decision log, outlining our choice and why so our patterns remain consistent.

As a devil's advocate, the render time of `FormLayout` may be negligible nonetheless using `React.memo` is negligible as well. Do we have benchmarks for the time blocked or saved by using `React.Memo`.

BPScott:
> Before we start pulling React.PureComponent & React.Memo from our components we should probably come up with a decision log, outlining our choice and why so our patterns remain consistent.

Excellent points :) I'm happy to say lets do one migration stage at a time, as long as we make sure we make space for doing the PureComponent audit at some point soon. I really like the idea of documenting this in a decision record. Tangentially I've been thinking about decision records for some other stuff too - how we handle imports/exports (i.e. why we avoid defaults), how we structure subcomponent and utility folders, opinions on how we name attributes (prefering positive namings execpt when defaults result in double negatives, like HTML attributes so `disabled` is fine) and how we handle event handlers (this is kinda up in the air - half our stuff has default noop functions, and half leaves them undefined and checks for the functions existence before calling them) - but they're all stories for another day :)

> As a devil's advocate, the render time of FormLayout may be negligible nonetheless using React.memo is negligible as well. Do we have benchmarks for the time blocked or saved by using React.Memo.

Alas I know of no benchmarks - most stuff I've saw is very dependant on how the components are used - the more complex the props you pass in the longer the diff takes I guess. For honest meaningful figures we'd need to look at real world usage. Though my personal bias leans towards "stuff like FormLayout is so dang simple I'd bet on React.memo doing more harm than good" but I've got absolutely nothing to back that up.

PR #2543, opened 2019-12-11
[Tooltip] Converted to a functional component

AndrewMusgrave:

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -1,9 +1,18 @@
-import React from 'react';
+import React, {
+  memo,
+  useEffect,
+  useState,
+  useRef,
+  RefObject,
+  useCallback,
+} from 'react';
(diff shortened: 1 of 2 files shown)


src/components/Tooltip/Tooltip.tsx
```diff
@@ -28,122 +29,98 @@ export interface TooltipProps {
...
+export function Tooltip({
+  children,
+  content,
+  light,
+  active: originalActive,
+  preferredPosition = 'below',
+  activatorWrapper = 'span',
+}: TooltipProps) {
+  const WrapperComponent: any = activatorWrapper;
+  const {value: active, setTrue: handleFocus, setFalse: handleBlur} = useToggle(
```
LauraAubin: TIL about `useToggle` 😄

LauraAubin:
Looks like a test where `activatorContainer` is null could improve codecov

AndrewMusgrave:
`activatorContainer.current` will never be `null` in `useEffect`, the check appeases typescript and prevents future regression but isn't currently testable.

PR #2237, opened 2019-10-03
Upgrade typescript to 3.7.2

danrosenthal:
[⏫ TypeScript 3.6.3](https://devblogs.microsoft.com/typescript/announcing-typescript-3-6/)

PR #2549, opened 2019-12-13
TypeScript 3.7 (take 2)

BPScott:

#2237 bumped us up to typescript 3.7, but it got reverted in #2501 because it didn't play nice in consuming projects.

This reverts the revert and fixes the issues that caused the problems in consuming projects.

Updates to TS 3.7, fixing any type checking issues.

If you have an uninitialized property that then gets referenced (like by
having a getter for it) then TS emits declarations that break in
consuming projects. Convert a bunch of cases where we created getters for non-existant
properties to use boring old method functions, thus allowing consumer builds to work again..

### How to 🎩

Tests / linting pass

Test this in a consuming project:

- Checkout the latest version of polaris-styleguide
- In polaris-react: `yarn run build-consumer polaris-styleguide`
- Then in the styleguide `yarn run build` and see that the build passes

### DIFF
--- a/src/components/DropZone/tests/DropZone.test.tsx
+++ b/src/components/DropZone/tests/DropZone.test.tsx
@@ -473,6 +473,9 @@ function setBoundingClientRect(size: keyof typeof widths) {
         left: 0,
         bottom: 0,
         right: 0,
+        x: 0,
+        y: 0,
+        toJSON() {},
       };
     });
 }
--- a/src/components/Filters/Filters.tsx
+++ b/src/components/Filters/Filters.tsx
@@ -107,14 +107,6 @@ class Filters extends React.Component<ComposedProps, State> {
   private moreFiltersButtonContainer = createRef<HTMLDivElement>();
   private focusNode = createRef<HTMLDivElement>();
 
-  private get hasAppliedFilters(): boolean {
-    const {appliedFilters, queryValue} = this.props;
-    const filtersApplied = Boolean(appliedFilters && appliedFilters.length > 0);
-    const queryApplied = Boolean(queryValue && queryValue !== '');
-
-    return filtersApplied || queryApplied;
-  }
-
   render() {
     const {
       filters,
@@ -300,7 +292,7 @@ class Filters extends React.Component<ComposedProps, State> {
 
     const filtersDesktopFooterMarkup = (
       <div className={styles.FiltersContainerFooter}>
-        <Button onClick={onClearAll} disabled={!this.hasAppliedFilters}>
+        <Button onClick={onClearAll} disabled={!this.hasAppliedFilters()}>
           {intl.translate('Polaris.Filters.clearAllFilters')}
         </Button>
         <Button onClick={this.closeFilters} primary>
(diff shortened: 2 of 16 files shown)


src/components/Filters/Filters.tsx
```diff
@@ -107,14 +107,6 @@ class Filters extends React.Component<ComposedProps, State> {
   private moreFiltersButtonContainer = createRef<HTMLDivElement>();
   private focusNode = createRef<HTMLDivElement>();
 
-  private get hasAppliedFilters(): boolean {
```
BPScott: ESLint wants us to put private methods below public ones like render().

src/components/ResourceList/ResourceList.tsx
```diff
@@ -329,13 +329,15 @@ class ResourceList extends React.Component<CombinedProps, State> {
     };
   }
 
+  // eslint-disable-next-line @typescript-eslint/member-ordering
```
BPScott: ESLint wants us to put private methods below public ones like render(). However in this file that'll make the diff incredibly noisy so I've surpassed the warning

src/components/ResourceList/components/BulkActions/BulkActions.tsx
```diff
@@ -169,6 +169,7 @@ class BulkActions extends React.PureComponent<CombinedProps, State> {
     }
   }
 
+  // eslint-disable-next-line @typescript-eslint/member-ordering
```
BPScott: ESLint wants us to put private methods below public ones like render(). However in this file that'll make the diff incredibly noisy so I've surpassed the warning

src/components/ResourceList/components/BulkActions/BulkActions.tsx
```diff
@@ -204,6 +206,8 @@ class BulkActions extends React.PureComponent<CombinedProps, State> {
       polaris: {intl},
     } = this.props;
 
+    const actionSections = this.actionSections();
```
BPScott: This got called a whole load of times, so instead of computing the actionSections repeatedly, lets do it once

src/components/ResourceList/components/BulkActions/BulkActions.tsx
```diff
@@ -317,19 +320,16 @@ class BulkActions extends React.PureComponent<CombinedProps, State> {
...
     let combinedActions: ActionListSection[] = [];
 
-    if (this.actionSections && rolledInPromotedActions.length > 0) {
-      combinedActions = [
-        {items: rolledInPromotedActions},
-        ...this.actionSections,
-      ];
-    } else if (this.actionSections) {
-      combinedActions = this.actionSections;
+    if (actionSections && rolledInPromotedActions.length > 0) {
```
BPScott: Same logic here, just prettier reformatting.

BPScott:
Added comments to any changes that I didn't already mention as part of #2237

src/components/DropZone/tests/DropZone.test.tsx
```diff
@@ -473,6 +473,9 @@ function setBoundingClientRect(size: keyof typeof widths) {
         left: 0,
         bottom: 0,
         right: 0,
+        x: 0,
+        y: 0,
+        toJSON() {},
```
dleroux: Why is this needed?

beefchimi: Probably to satisfy the type `Element['getBoundingClientRect']`:
https://github.com/Shopify/polaris-react/pull/2549/files#diff-ba67dc9a18e9f3acebe10b28d3397442R1029

BPScott: TS Updated their types and added these properties https://github.com/Shopify/polaris-react/pull/2237#discussion_r344419536

PR #2550, opened 2019-12-13
[Color system] Update names

tmlayton:

Updates color names per our working sessions in preparation for user testing.

Updates names per [these changes](https://github.com/Shopify/polaris-react/compare/color-system-doc-testing?short_path=7e44770#diff-7e44770a23ffedfb5ca5fa1d9856024b)

## 

### How to 🎩

- Poke around the kitchen sink in light and dark modes to make sure things look okay

* [x] 🎩


src/components/Avatar/Avatar.scss
```diff
@@ -77,7 +77,7 @@ $large-size: rem(60px);
   left: 50%;
   width: 100%;
   height: 100%;
-  background-color: var(--p-neutral, color('sky'));
+  background-color: var(--p-secondary, color('sky'));
```
danrosenthal: We should use this pull request as an opportunity to get rid of the base (unmodified) variants. They've been more confusing than useful.

Update this to use the disabled secondary action like `Badge` uses.

```suggestion
background-color: var(--p-action-secondary-disabled, color('sky'));
```

src/utilities/theme/tokens.ts
```diff
@@ -21,13 +21,13 @@ const Overrides = {
...
-  bannerWarningBorder: buildBannerBorder('--p-warning-border'),
-  bannerCriticalBorder: buildBannerBorder('--p-critical-border'),
+  bannerBorderDefault: buildBannerBorder('--p-border'),
+  bannerBorderSuccess: buildBannerBorder('--p-border-success'),
+  bannerBorderHighlight: buildBannerBorder('--p-border-highlight'),
+  bannerBorderWarning: buildBannerBorder('--p-border-warning'),
+  bannerBorderCritical: buildBannerBorder('--p-border-critical'),
   badgeMixBlendMode: 'luminosity',
-  borderSubdued: `${rem('1px')} solid var(--p-border-subdued-on-surface)`,
+  thinBorderSubdued: `${rem('1px')} solid var(--p-border-subdued)`,
```
tmlayton: This one is a little weird but there was a naming conflict

src/components/Avatar/Avatar.scss
tmlayton: Okay, double checked this was the only one

tmlayton:
> Looks the order of the naming for the `primary`, `critical`, `warning`, `highlight`, and `success` buckets needs to be updated to element first (`action-primary`, `border-critical`).
> 
> The other buckets look good 👍

Yep, was still WIP. Finished now and will go ahead and merge so we can move forward with the user testing sandbox (I’d rather have the updated color docs in master). Please still review for possible missed names and fix forward.

PR #2553, opened 2019-12-14
Enable strictPropertyInitialization TS config

BPScott:

We try to be strict mode compliant but this was disabled. This fixes the
components that fell afoul of that lint. Mostly this is accounting for
refs that may be null

Enable strictPropertyInitialization in our tsconfig

Add correct initializations to properties that were flagged. In most cases this was we were saying a value may be null but actually we never initialized it so it was `undefined`. Initilaizing to null is the fix.

### How to 🎩

- Ensure Modal still renders as you'd expect
- Ensure a component that uses a portal (like Toast or Popover) still renders its portaled content correctly.
- Ensure sticky managed items still work

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -6,10 +6,9 @@
     "declarationDir": "types",
     "declarationMap": false,
     "jsx": "react-native",
-    // strictFunctionTypes & strictPropertyInitialization are implictly
-    // enabled by strict mode but we're not yet compliant, so disable for now
+    // strictPropertyInitialization is implictly enabled by strict mode but
+    // we're not yet compliant, so disable for now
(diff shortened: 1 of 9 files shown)


tsconfig.json
```diff
@@ -6,10 +6,9 @@
     "declarationDir": "types",
     "declarationMap": false,
     "jsx": "react-native",
-    // strictFunctionTypes & strictPropertyInitialization are implictly
-    // enabled by strict mode but we're not yet compliant, so disable for now
+    // strictPropertyInitialization is implictly enabled by strict mode but
```
BPScott:
```suggestion
    // strictFunctionTypes is implictly enabled by strict mode but
```

PR #2555, opened 2019-12-16
[TrapFocus] Fix tab focusing

dleroux:

Currently, `TrapFocus` does not listen for keyboard events such as tabbing. Instead, it handles what happens _after_ we lose focus of the container (`handleBlur`). The problem with this approach is two things:
- If we focus an element outside of the page (such as the browser's search bar), we can't get focus back to the page contents, even if the `activeElement` is what it should be.
- If we prevent focus from ever leaving the page contents (by removing the check `relatedTarget == null`, then we introduce a bug where we rip focus away from the rest of the page if the container was rendered separately such as in an iFrame (see this PR explaining why this change broke the style guide https://github.com/Shopify/polaris-react/pull/2530)

- Functionalize `TrapFocus` and `KeypressListener`
- Remove all of the logic that handles focus on the `blur` event
- Add a key press listener for `tab`
- Add an event listener for `focusin` to handle focusing the container

### How to 🎩

- Follow [these](https://github.com/Shopify/polaris-react/blob/6579045fbbf668b8f1624631cd5c7f609f4bff57/documentation/Tophatting%20documentation.md#how-to--documentation) instructions to tophat with the style guide
- Check that `Popover` is acting as expected (since it also makes use of `TrapFocus`)

#### Modal:
- Click on open button, modal opens
	- Expected result = `Dialog` should be focused
- Pressing tab should,
	- focus the first element after the last element (and the reverse when tabbing backwards)
- Modal is opened, click outside of page, keep tabbing until we,
	- focus the `Dialog` (not the first focusable element)

### 🎩 checklist

### DIFF
--- a/src/components/TrapFocus/tests/TrapFocus.test.tsx
+++ b/src/components/TrapFocus/tests/TrapFocus.test.tsx
@@ -1,6 +1,10 @@
 import React from 'react';
 // eslint-disable-next-line no-restricted-imports
-import {mountWithAppProvider, trigger} from 'test-utilities/legacy';
+import {
+  mountWithAppProvider,
+  trigger,
+  ReactWrapper,
+} from 'test-utilities/legacy';
 import {mountWithApp} from 'test-utilities';
 import {
   EventListener,
@@ -10,7 +14,9 @@ import {
   Button,
 } from 'components';
 import * as focusUtilities from '@shopify/javascript-utilities/focus';
+import * as focusUtils from '../../../utilities/focus';
 import {TrapFocus} from '../TrapFocus';
+import {Key} from '../../../types';
 
 jest.mock('@shopify/javascript-utilities/fastdom', () => ({
   ...require.requireActual('@shopify/javascript-utilities/fastdom'),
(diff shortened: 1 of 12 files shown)


LauraAubin:
When we have focus on the `Dialog`, we cannot <kbd>shift</kbd> + <kbd>tab</kbd> to the last element within the modal. This is because we are not handling a case when the target is the `Dialog`.

https://github.com/Shopify/polaris-react/blob/9adeb940df438cf1fab1d064d667b5991e3239cc/src/components/TrapFocus/TrapFocus.tsx#L82-L90

dleroux:
@LauraAubin I'm guessing you decided to go with this PR? I did a 🎩 and it works great! Do you think you might have time to work on test coverage?

src/components/TrapFocus/tests/TrapFocus.test.tsx
```diff
@@ -169,6 +189,136 @@ describe('<TrapFocus />', () => {
...
+    });
+  });
 });
 
 function noop() {}
+
+function firstNode(element: ReactWrapper) {
+  const elementNode = element.getDOMNode();
+
+  if (Array.isArray(elementNode))
```
LauraAubin: If the `element` has a fragment as the highest parent (in this case `Focus`), then getting the DOM node of this will be an array because you cannot have one element if the fragment contains multiple.

LauraAubin:
Tested this change in Web and it appears to be working!

There are a bunch of forked Modals in Web, but I found one using Polaris in the channels settings. Navigate to `admin/settings/channels` and click "Add sales channels" to bring up the `ChannelsSelectionModal.tsx`.

LauraAubin:
^ Scratch that, that was a Rails page.

This fix is still working, tested with the `InsertImageModal.tsx` from `admin/collections/new` by clicking on the image button in the rich text editor.

PR #2562, opened 2019-12-17
Refactor I18n Class

BPScott:

I was looking at making some tweaks to to the parameter order of i18n class, and found that its had some odd loose typing. So before doing that let's kick its tires a bit and make its type checking stricter.

Rewrite the i18n test suite

Require an translations argument in the constructor, and simplify the
typing of the translations property.
Define translate in a function on i18n instead of having it use an external
function.
Remove need for some exported types by infering the args of I18n instead
Stricter typing for the replacements argument

Fix some call sites now that passing undefined keys into the
replacements object is no longer allowed

## 

### How to 🎩

Tests, linting, type-check passes. Ensure translations are still present on modified components


src/components/DataTable/components/Cell/Cell.tsx
```diff
@@ -61,7 +61,8 @@ export function Cell({
   );
 
   const iconClassName = classNames(sortable && styles.Icon);
-  const direction = sorted ? sortDirection : defaultSortDirection;
+  const direction =
+    sorted && sortDirection ? sortDirection : defaultSortDirection;
```
BPScott: making sure direction can't be `undefined` if `sortDirection` is undefined

src/components/Tag/Tag.tsx
```diff
@@ -18,7 +18,9 @@ export interface TagProps {
 export function Tag({children, disabled = false, onRemove}: TagProps) {
   const i18n = useI18n();
   const className = classNames(disabled && styles.disabled, styles.Tag);
-  const ariaLabel = i18n.translate('Polaris.Tag.ariaLabel', {children});
+  const ariaLabel = i18n.translate('Polaris.Tag.ariaLabel', {
+    children: children || '',
```
BPScott: children could be undefined previously, this ensures that we pass a string into the replacements object as `{children: undefined}` is now a type error

AndrewMusgrave: It seems like an accessible issue to have a tag without any content or accessibility label. We should probably create an issue for this

src/components/PolarisTestProvider/PolarisTestProvider.tsx
```diff
@@ -39,7 +39,7 @@ type MediaQueryContextType = NonNullable<
  */
 export interface WithPolarisTestProviderOptions {
   // Contexts provided by AppProvider
-  i18n?: TranslationDictionary | TranslationDictionary[];
+  i18n?: ConstructorParameters<typeof I18n>[0];
```
AndrewMusgrave: TIL `ConstructorParameters`. So many useful types 😄

src/components/AppProvider/AppProvider.tsx
```diff
@@ -32,8 +32,8 @@ interface State {
 }
 
 export interface AppProviderProps extends AppBridgeOptions {
-  /** A locale object or array of locale objects that overrides default translations */
-  i18n: TranslationDictionary | TranslationDictionary[];
+  /** A locale object or array of locale objects that overrides default translations. If specifying an array then your fallback language dictionaries should come first, followed by your primary language dictionary */
+  i18n: ConstructorParameters<typeof I18n>[0];
```
AndrewMusgrave: I wonder how this will appear in the style guide 🤔 And hopefully not `ConstructorParameters<typeof I18n>[0]` 😬

AndrewMusgrave: Looks like it outputs the correct type 👍

PR #2571, opened 2019-12-18
Color system/filters

tmlayton:

To support coloring using the CSS filter property. This technique is specifically used to color SVGs that are embedded with the `img` tag. We do this with 3rd party SVGs for security reasons.

Converting hex values to filter values and making those available as CSS variables.

## 

### How to 🎩

1) Take a look at the "Details page" under the Playground stories
2) The two sales channel icons are being colored with the filter property when looking at the light and dark color schemes, verify they have the proper color. The color might be off a bit as the math is imperfect (brute force).

### 🎩 checklist

* [x] I have 🎩’d


tmlayton:
Likely not moving forward with this approach and will be using static values for this edge case

PR #2572, opened 2019-12-19
[V5] invert translations order in AppProvider i18n

BPScott:

Previously you had to put the language you wanted at the end of the
array which was not obvious and didn't match the behaviour of
react-i18n.

Now if you pass an array into `<AppProvider i18n>` then the first object
in that array should be your preferred language.

This leads to simpler integration and simpler documentation on our side.

### How to 🎩

Check tests

### DIFF
--- a/UNRELEASED-v5.md
+++ b/UNRELEASED-v5.md
@@ -12,6 +12,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Removed `AppBridge`, `ResourcePicker` and `Loading`, `Modal`, `Page`, `Toast` App Bridge render delegation ([#2046](https://github.com/Shopify/polaris-react/pull/2046))
 - Drop support for iOS 9 ([#2195](https://github.com/Shopify/polaris-react/pull/2195))
 - The styles in global.scss have been moved into AppProvider. This only affects apps using the 'esnext' build; if you import `@shopify/polaris/styles.css` you are unaffected. Any applications using the 'esnext' build no longer needs to import the `@shopify/polaris/esnext/global.scss` file as it does nothing. An empty global.scss is provided in order to not break existing integration with sewing-kit versions < 0.113.0. ([#2392](https://github.com/Shopify/polaris-react/pull/2392))
+- Reversed the precedence of the language dictionaries passed into the `AppProvider`'s `i18n` prop. When passing an array of dictionaries the first dictionary should be your prefered language, followed by any fallback languages. ([]())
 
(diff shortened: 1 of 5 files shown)


UNRELEASED-v5.md
```diff
@@ -12,6 +12,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Removed `AppBridge`, `ResourcePicker` and `Loading`, `Modal`, `Page`, `Toast` App Bridge render delegation ([#2046](https://github.com/Shopify/polaris-react/pull/2046))
 - Drop support for iOS 9 ([#2195](https://github.com/Shopify/polaris-react/pull/2195))
 - The styles in global.scss have been moved into AppProvider. This only affects apps using the 'esnext' build; if you import `@shopify/polaris/styles.css` you are unaffected. Any applications using the 'esnext' build no longer needs to import the `@shopify/polaris/esnext/global.scss` file as it does nothing. An empty global.scss is provided in order to not break existing integration with sewing-kit versions < 0.113.0. ([#2392](https://github.com/Shopify/polaris-react/pull/2392))
+- Reversed the precedence of the language dictionaries passed into the `AppProvider`'s `i18n` prop. When passing an array of dictionaries the first dictionary should be your prefered language, followed by any fallback languages. ([]())
```
AndrewMusgrave:
```suggestion
- Reversed the precedence of the language dictionaries passed into the `AppProvider`'s `i18n` prop. When passing an array of dictionaries the first dictionary should be your prefered language, followed by any fallback languages. ([#2572](https://github.com/Shopify/polaris-react/pull/2572))
```

src/utilities/i18n/tests/I18n.test.ts
```diff
@@ -17,12 +17,17 @@ describe('I18n.translate', () => {
...
-    ]);
+      {one: 'one', two: 'two', three: 'three', fallback: 'fallback'},
+    ];
+
+    const i18n = new I18n(translations);
     expect(i18n.translate('one')).toBe('uno');
     expect(i18n.translate('fallback')).toBe('fallback');
+
+    // Assert the original translations array is not mutated
+    expect(translations[0].two).toBe('deux');
```
AndrewMusgrave: Should this be in a separate test?

BPScott: I don't think it's worth it. I don't see much value in having multiple tests with exactly the same setup but slightly different assertions, I'd rather stick them into one test - that stops you wasting time repeating the setup

PR #2579, opened 2019-12-20
Bring back some old type oddness in i18n to help a release

BPScott:
Don't call toString() on the item in the replacements array as JS can
handle the casting fine itself. This accounts for people passing in a
key of undefined and TS not catching it in consuming apps


BPScott:
Gonna skip the unreleased.md entry for this PR and add it direct to the changelog when I cut the release

PR #2591, opened 2020-01-05
[Subheading] Update documentation

LauraAubin:

Improve documentation.

- Adds a period to the end of the description paragraph.
- The default example does not mention card.
- Removes other mentions of card.

### DIFF
--- a/src/components/Subheading/README.md
+++ b/src/components/Subheading/README.md
@@ -13,18 +13,13 @@ keywords:
   - titles of sub-sections
   - subsection titles
   - titles of subsections
-  - card subheadings
-  - card section headings
-  - headings of card sections
-  - card section subheadings
   - android
   - ios
 ---
 
 # Subheading
 
-Subheadings are used for the title of any sub-sections in top-level page
-sections. Generally, sections of a card use subheadings for their titles
+Subheadings are used for the title of any sub-sections in top-level page sections.
 
 ---
 
@@ -46,9 +41,9 @@ Subheadings should follow the content guidelines for [headings and subheadings](
 
 ## Examples
 
-### Subheading in a card
+### Typographic subheading
 
-Use to structure content in a card.
+Used for the title of any sub-sections in top-level page sections.
 
 ```jsx
 <Subheading>Accounts</Subheading>
@@ -56,13 +51,13 @@ Use to structure content in a card.
 
 <!-- content-for: android -->
 
-![Subheading in a card](/public_images/components/Subheading/android/default@2x.png)
+![typographic subheading](/public_images/components/Subheading/android/default@2x.png)
 
 <!-- /content-for -->
 
 <!-- content-for: ios -->
 
-![Subheading in a card](/public_images/components/Subheading/ios/default@2x.png)
+![typographic subheading](/public_images/components/Subheading/ios/default@2x.png)
 
 <!-- /content-for -->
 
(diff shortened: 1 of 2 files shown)


src/components/Subheading/README.md
```diff
@@ -46,12 +46,27 @@ Subheadings should follow the content guidelines for [headings and subheadings](
...
 ```jsx
-<Subheading>Accounts</Subheading>
+<Card title="Online store dashboard">
+  <Card.Section>
+    <Subheading>Products</Subheading>
+  </Card.Section>
+  <Card.Section>
+    <Subheading>Orders</Subheading>
+  </Card.Section>
+</Card>
```
LauraAubin: TLDR: Should we remove all mentions of `Card` from `Subheading` documentation?

Reason: Title on Card.Section implements this for us.

<hr>

I'm going back and forth here regarding the description of this component.

> Generally, sections of a card use subheadings for their titles.

But the `Card.Section` accepts a `title` prop which also implements `Subheading` if the title is a string like so,

https://github.com/Shopify/polaris-react/blob/21be9c8891e5652a479b8c700c40484ee862561e/src/components/Card/components/Section/Section.tsx#L37-L38

One thing I noticed is that the Card `Section` styles whatever its title is so that it has space away from the content that exists within the section,

https://github.com/Shopify/polaris-react/blob/21be9c8891e5652a479b8c700c40484ee862561e/src/components/Card/Card.scss#L66-L67

If you use the `Subheading` component yourself, like the documentation suggests, you need to implement this styling yourself using a Stack, css, etc.

src/components/Subheading/README.md
```diff
@@ -46,12 +46,27 @@ Subheadings should follow the content guidelines for [headings and subheadings](
 
 ## Examples
 
+### Typographic subheading
+
+Used for the title of any sub-sections in top-level page sections.
```
LauraAubin: Following the same sort of wording as `Heading`.

This also removes the mentions of Card so this example can feel isolated.

src/components/Subheading/README.md
```diff
@@ -24,7 +24,7 @@ keywords:
 # Subheading
 
 Subheadings are used for the title of any sub-sections in top-level page
-sections. Generally, sections of a card use subheadings for their titles
+sections. Generally, sections of a card use subheadings for their titles.
```
LauraAubin: Missing period.

src/components/Subheading/README.md
dleroux: > Should we remove all mentions of Card from Subheading documentation?

I think we probably should. I have a feeling it's because this prop was added after and the example wasn't updated.

AndrewMusgrave: As a rule of thumb examples should always be as simple as possible. Some cases require more complexity, e.g. `Form` needs form elements to have a working example or frame needing top/navigation/etc.

src/components/Subheading/README.md
```diff
@@ -46,35 +41,28 @@ Subheadings should follow the content guidelines for [headings and subheadings](
...
 <!-- content-for: ios -->
 
-![Subheading in a card](/public_images/components/Subheading/ios/default@2x.png)
+![typographic subheading](/public_images/components/Subheading/ios/default@2x.png)
 
 <!-- /content-for -->
 
 ---
 
-## Related components
```
AndrewMusgrave: Would heading be a related component?

LauraAubin: Ouuu good point, I think yes!

PR #2592, opened 2020-01-06
[Tabs] Adds a means to hide tabs if there is only one

jared-dykstra:
Devs had a need to create a HOC `Tabs` component for cases when the number of tabs was dynamic and the desire was to hide the tab UI if there is only one tab.

For example: https://github.com/Shopify/web/pull/22045/files#diff-c2ee73134750ecc64faa9c16571e09cf

This PR adds a new optional prop to `Tabs`.  If undefined or `false`, current behavior is retained.  If `true` and if there is not more than one visible tab, `children` is rendered directly, without being wrapped with the tabs markup.

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page, Card, Tabs} from '../src';

  const [selected, setSelected] = React.useState(0);

  const handleTabChange = React.useCallback(
    [],
  );

  const tabs = [
    {
      id: 'prospects',
      content: 'Prospects',
      panelID: 'prospects-content',
    },
    // {
    //   id: 'second',
    //   content: 'Second',
    //   panelID: 'second-content',
    // },
  ];

  return (
    <Page title="Playground">

      <Card>
        <Tabs
          tabs={tabs}
          selected={selected}
          onSelect={handleTabChange}
          optionalSingleTab
        >
          </Card.Section>
        </Tabs>
      </Card>
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -6,6 +6,8 @@
 
 - Added `hideTags` prop to `Filters` ([#2573](https://github.com/Shopify/polaris-react/pull/2573))
 
+- Added `optionalSingleTab` prop to `Tabs` ([#2592](https://github.com/Shopify/polaris-react/pull/2592))
+
 ### Bug fixes
 
 - Fixed issue with `Filters` component displaying an undesired margin top and bottom on the button element on Safari ([#2292](https://github.com/Shopify/polaris-react/pull/2292))
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -27,6 +27,8 @@ export interface TabsProps {
   fitted?: boolean;
   /** Callback when tab is selected */
   onSelect?(selectedTabIndex: number): void;
+  /** Suppress tab header if there is only one tab */
+  optionalSingleTab?: boolean;
 }
 
 type CombinedProps = TabsProps & WithAppProviderProps;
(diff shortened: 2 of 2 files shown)


jared-dykstra:
@briangilhutton - This comes out of a PR you were recently working on:  https://github.com/Shopify/web/pull/22045

dleroux:
Thanks @jared-dykstra. Although I'm not sure there's value in adding this prop at this time. It adds code for a use case that is seldom needed and probably more explicit if conditionally rendered in the parent. Have you found a need for multiple uses of this?

jared-dykstra:
@dleroux - I don't have enough context to know if there are multiple needs for something like this, unfortunately. 

@qq99 - are you able to provide this additional context?

qq99:
From a quick code search, I found 2 more examples of this kind of thing @ `ImagePickerModal` and `FraudModal` in admin, so that'd be 3 spots that could use this prop

dleroux:
Interesting. I feel like this shouldn't be a concern for tabs, tabs that don't render tabs. That might just be me though.

AndrewMusgrave:
I agree with Dan, I'm not sure this needs to be a concern of `Tabs`. It seems odd to me for `Tabs` to not render `Tabs`. Duplicating the logic where it's needed doesn't seem bad to me as it's rather small

```tsx
const tabsMarkup = tabs.length ? <Tabs {...tabsProps}> : children;
```

jared-dykstra:
@AndrewMusgrave - Minor correction: `tabs.length > 1`  (not falsey)

It doesn't seem like there is much appetite for this PR.  If not, that's OK.  I wanted to propose it anyway with the hopes of eventually removing (OptionalTabs.tsx)[ https://github.com/Shopify/web/pull/22045/files#diff-c2ee73134750ecc64faa9c16571e09cf] that was introduced in the referenced PR.

tmlayton:
Thanks for proposing this. I have the same opinion that this shouldn’t be a tab concern.

dleroux:
Sounds like we're going to hold off on this, I'll close the PR for now. If this comes up more often feel free to bring it up again.

PR #2598, opened 2020-01-07
Be explict about some i18n oddness around undefined keys, and TextField's value

BPScott:

Following up https://github.com/Shopify/polaris-react/pull/2579.

Make passing `{foo: undefined}` into translations replacements a runtime error, in addition to a TS type error for those fun cases where people power through mistyping things using `any` (looking at you enzyme's trigger function that isn't type-safe).

Make TextField's `normalizedValue` be an empty string if you pass anything other than a string into it. Once again this is limited to be a string only by typescript but people using those not-type-safe functions can pass in a number and not be warned. Having normalizedValue be a number mucks up a bunch of things so catch that at runtime too.

Make TextField a little lazier by only working out character count stuff if we're actually about to display it. 

### How to 🎩

tests pass

### DIFF
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -217,36 +217,39 @@ export function TextField({
     </div>
   ) : null;
 
-  const characterCount = normalizedValue.length;
-  const characterCountLabel = maxLength
-    ? i18n.translate('Polaris.TextField.characterCountWithMaxLength', {
-        count: characterCount,
-        limit: maxLength,
-      })
-    : i18n.translate('Polaris.TextField.characterCount', {
-        count: characterCount,
-      });
+  let characterCountMarkup = null;
+  if (showCharacterCount) {
+    const characterCount = normalizedValue.length;
+    const characterCountLabel = maxLength
+      ? i18n.translate('Polaris.TextField.characterCountWithMaxLength', {
+          count: characterCount,
+          limit: maxLength,
+        })
+      : i18n.translate('Polaris.TextField.characterCount', {
+          count: characterCount,
+        });
 
-  const characterCountClassName = classNames(
-    styles.CharacterCount,
-    multiline && styles.AlignFieldBottom,
-  );
+    const characterCountClassName = classNames(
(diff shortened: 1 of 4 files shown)


src/components/TextField/TextField.tsx
```diff
@@ -181,7 +181,10 @@ export function TextField({
     focused ? input.focus() : input.blur();
   }, [focused]);
 
-  const normalizedValue = value != null ? value : '';
+  // Use a typeof check here as Typescript mostly protects us from non-stringy
+  // values but overzealous usage of `any` in consuming apps means people have
+  // been known to pass a number in, so make it clear that doesn't work.
+  const normalizedValue = typeof value === 'string' ? value : '';
```
dleroux: I like this approach.

src/utilities/i18n/I18n.ts
```diff
@@ -33,22 +33,15 @@ export class I18n {
...
           );
         }
 
-        // Bringing back an old bit of oddness to expedite a release.
-        // {foo: undefined} is allowed in calling apps but won't trigger type
-        // warnings and won't get caught in the above check so let it through
-        // for now as JS can handle it ok. We should work out how to be
-        // stricter, or deliberatly allow undefined as a value
-        return replacements[replacement] as string;
+        return replacements[replacement].toString();
```
dleroux: 👍

src/utilities/i18n/tests/I18n.test.ts
```diff
@@ -35,7 +35,16 @@ describe('I18n.translate', () => {
     expect(() => {
       i18n.translate('key', {notString: 'bar'});
     }).toThrow(
-      `No replacement found for key 'string'. The following replacements were passed: 'notString'`,
+      `Error in translation for key 'key'. No replacement found for key 'string'. The following replacements were passed: '{"notString":"bar"}'`,
+    );
+
```
dleroux: Not 100% necessary but it might be best to split these 2 expect statements into 2 tests?

BPScott:
A bit or rebasing to split these changes into isolated commits

PR #2603, opened 2020-01-07
[Card] Custom footer actions example should be right aligned

LauraAubin:

`Card` footer actions are now right aligned by default, but the custom footer example was left aligned.

Wrap the `ButtonGroup` in a `Stack`.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -5,7 +5,6 @@
 ### Enhancements
 
 - Added `hideTags` prop to `Filters` ([#2573](https://github.com/Shopify/polaris-react/pull/2573))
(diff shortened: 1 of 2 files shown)


UNRELEASED.md
```diff
@@ -5,6 +5,7 @@
 ### Enhancements
 
 - Added `hideTags` prop to `Filters` ([#2573](https://github.com/Shopify/polaris-react/pull/2573))
+- Updated `Card` with custom footer actions example to be right aligned [#2603](https://github.com/Shopify/polaris-react/pull/2603))
```
dleroux: Wondering if this should be under documentation?

```suggestion
- Updated `Card` with custom footer actions example to be right-aligned ([#2603](https://github.com/Shopify/polaris-react/pull/2603))
```

PR #2604, opened 2020-01-08
[Color system] Use lambdas to relatively adjust variants

tmlayton:

Decorative colors were fixed and we want them to be variable based on the decorative input color. Likewise, there are minor adjustments to the hue of interactive in darsh mode to keep links looking blue-ish instead of lavender that were set as fixed values.

Adds support for role variants to accept a lambda as well as a number. The decorative variants are now generated based on relative changes to base decorative color as well as the interactive variants.

RE: interactive variants—the slight adjustments are likely not needed for other hues, say a consumer had magenta as their interactive color. But for the sake of simplicity this will make the slight hue and saturation adjustments regardless, which I think is fine.

RE: decorative colors—the color harmony for decorative colors is a custom pentad (set of 5 colors) that looks like this:

With these changes we effectively allow consumers to keep the harmony fixed in place and simply rotate the color wheel.

There is a very slight change to hex values based on the HSLuv math and rounding, but visually the colors are the same:

Screen Shot 2020-01-07 at 11 40 12 PM

And here’s what the decorative sets look like passing in red, green, and blue for the decorative role:

Screen Shot 2020-01-08 at 12 24 32 AM

#### Darsh mode

Screen Shot 2020-01-08 at 2 51 02 PM

### A bit more

Lambdas as a concept for color adjustments are potentially useful in lots of ways. One example is for certain hues in HSLuv, the color can appear more saturated. We could have a lambda that slightly desaturates colors with hues of a certain range.

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page, Avatar, Stack, ThemeProvider} from '../src';

  return (
    <Page title="Playground">
      <Stack vertical>
        <Stack alignment="baseline">
          <Avatar initials="1" name="d" />
          <Avatar initials="2" name="e" />
          <Avatar initials="3" name="a" />
          <Avatar initials="4" name="b" />
          <Avatar initials="5" name="c" />
        </Stack>
        <ThemeProvider theme={{UNSTABLE_colors: {decorative: '#FF0000'}}}>
          <Stack alignment="baseline">
            <Avatar initials="1" name="d" />
            <Avatar initials="2" name="e" />
            <Avatar initials="3" name="a" />
            <Avatar initials="4" name="b" />
            <Avatar initials="5" name="c" />
          </Stack>
        </ThemeProvider>
        <ThemeProvider theme={{UNSTABLE_colors: {decorative: '#00FF00'}}}>
          <Stack alignment="baseline">
            <Avatar initials="1" name="d" />
            <Avatar initials="2" name="e" />
            <Avatar initials="3" name="a" />
            <Avatar initials="4" name="b" />
            <Avatar initials="5" name="c" />
          </Stack>
        </ThemeProvider>
        <ThemeProvider theme={{UNSTABLE_colors: {decorative: '#0000FF'}}}>
          <Stack alignment="baseline">
            <Avatar initials="1" name="d" />
            <Avatar initials="2" name="e" />
            <Avatar initials="3" name="a" />
            <Avatar initials="4" name="b" />
            <Avatar initials="5" name="c" />
          </Stack>
        </ThemeProvider>
      </Stack>
    </Page>
  );
}

```

</details>

### 🎩 checklist

* [x] 🎩


documentation/Color system.md
```diff
@@ -197,13 +197,13 @@ Used to decorate elements where color does convey a specific meaning in componen
...
 | <pre>--p-decorative-one-text</pre>      | For use as a decorative text color that is applied on a decorative surface. | ![](https://www.gifpng.com/64x64/3d2800/FFFFFF?border-width=16&border-type=rectangle&border-color=fafafa&text=%20)<br />#3d2800 | ![](https://www.gifpng.com/64x64/ffffff/FFFFFF?border-width=16&border-type=rectangle&border-color=0c0d0e&text=%20)<br />#ffffff |
-| <pre>--p-decorative-two-surface</pre>   | For use as a decorative surface color.                                      | ![](https://www.gifpng.com/64x64/ffc6b3/FFFFFF?border-width=16&border-type=rectangle&border-color=fafafa&text=%20)<br />#ffc6b3 | ![](https://www.gifpng.com/64x64/cc5814/FFFFFF?border-width=16&border-type=rectangle&border-color=0c0d0e&text=%20)<br />#cc5814 |
+| <pre>--p-decorative-two-surface</pre>   | For use as a decorative surface color.                                      | ![](https://www.gifpng.com/64x64/ffc3ad/FFFFFF?border-width=16&border-type=rectangle&border-color=fafafa&text=%20)<br />#ffc3ad | ![](https://www.gifpng.com/64x64/cc5814/FFFFFF?border-width=16&border-type=rectangle&border-color=0c0d0e&text=%20)<br />#cc5814 |
```
tmlayton: Slight changes in the output hex value due to HSLuv math and rounding

src/utilities/theme/role-variants.ts
```diff
@@ -10,7 +10,7 @@ export const DefaultThemeColors = {
   warning: '#FFC453',
   highlight: '#5BCDDA',
   success: '#008060',
-  decorative: '#FDC7CE',
+  decorative: '#FFC96B',
```
tmlayton: Using the 1st decorative color as the base, this is an orange-ish color.

src/utilities/theme/role-variants.ts
```diff
@@ -574,8 +572,7 @@ export const roleVariants: RoleVariants = {
       name: 'decorativeTwoSurface',
       description: 'For use as a decorative surface color.',
       light: {
-        hue: 30,
-        saturation: 100,
+        hue: (hue) => (((hue - 26) % 360) + 360) % 360,
```
tmlayton: Now this take in the hue from the base decorative color and rotates it by -26 degrees. The modulo operations here are just a fancy one-line way to keep the number between 0 and 360 and wrap according. For example, 370 would become 10 and -10 would become 350.

src/utilities/theme/role-variants.ts
```diff
@@ -599,8 +596,8 @@ export const roleVariants: RoleVariants = {
       name: 'decorativeThreeSurface',
       description: 'For use as a decorative surface color.',
       light: {
-        hue: 144,
-        saturation: 54,
+        hue: (hue) => (((hue + 88) % 360) + 360) % 360,
+        saturation: (sat) => Math.max(sat - 46, 0),
```
tmlayton: Since we are only doing subtraction, this prevents that value from being less than 0.

src/utilities/theme/role-variants.ts
```diff
@@ -648,7 +645,11 @@ export const roleVariants: RoleVariants = {
     {
       name: 'decorativeFiveSurface',
       description: 'For use as a decorative surface color.',
-      light: {},
+      light: {
+        hue: (hue) => (((hue - 52) % 360) + 360) % 360,
+        saturation: (sat) => Math.max(sat - 5, 0),
+        lightness: 84,
```
tmlayton: The lightness for each of the 5 decorative variants is fixed at 84.

tmlayton:
I put this up as a draft because the decorative surface dark adjustments and decorative text variants still need to be updated. If we are okay with this approach I will update the rest.

src/utilities/theme/role-variants.ts
kyledurand: I feel like these magic number rotations should be constants i.e. `orangeHueRotation` or something descriptive like that.

kyledurand:
🎩 looks great as well 👍

src/utilities/theme/role-variants.ts
tmlayton: Curious why you feel the `-26` relative adjustment is different than say a fixed adjustment of `84`? The unit here is degrees and the starting degree of the base color would be unknown so color related names wouldn’t work.

I would instead like the update the role variant names to reflect something about the color, rather than 1, 2, 3, 4, and 5. But so far haven’t come up with any ideas other than adding the degree rotation to the variant name, for example `surfaceDecorativeBase`, `surfaceDecorative88`, `surfaceDecorative124`, `surfaceDecorative308`, `surfaceDecorative334`. Those are brittle though, if we tweak the relative hue adjustment that means we should update the name.

kyledurand: I realized after commenting that a color based name wouldn't work. I originally thought basing the rotation off of the original colour to the new color would be more descriptive but it really wouldn't. 

It's hard to tell what `-26` is doing here at a glance without knowing the lib so I think adding the word `rotation` or `paletteRotation` might help with that. You're right about the fragility of adding a number to the constant though. 

I'm perfectly fine with shipping as is

BPScott: I think `(((hue - 26) % 360) + 360) % 360` can be simplified to `(360 + hue - 26) %360`.

What if we did some currying clever-bollocks:


```js
// A function that returns a function that rotates the hue by a certain amount
function hueRotationFn(rotation) {
return (hue) => (360 + hue + rotation) % 360;
}


// Then in the variants:
hue: hueRotationFn(-26),
```

That way the clamping is encapsulated within the function factory rather than being duplicated on every line. You should be able to do a similar thing with the saturation clamping too

```js
function saturationOffsetFn(offset) {
  // Do we want to clamp this to an upper bound too?
  return (sat) => Math.max(sat + offset, 0),
}
```

tmlayton: > I think (((hue - 26) % 360) + 360) % 360 can be simplified to (360 + hue - 26) % 360.

Nice, good catch

tmlayton: > rather than being duplicated

Yeah I intentionally wanted to stay away from any dependencies for the lambdas, even if that was a function in the same file. Although that’s a gut feel and having a hard time putting that into words.

tmlayton: > Do we want to clamp this to an upper bound too?

It isn’t needed because the only adjustments made are subtractions. Though if abstracting it for general use, yes it should be clamped on the upper bound.

PR #2607, opened 2020-01-08
[Navigation] Fix navigation hide on key press

kyledurand:

There's an issue in production where if any item within the Navigation component has focus and you hit the escape key, the nav will try to hide itself off to the left.
[Screencast](https://screenshot.click/screencast_2020-01-08_15-48-59.mp4)

There's no github issue for this. I randomly stumbled upon it while working on something else.

This pr adds a check to make sure the nav is both active and on a small screen before calling dismiss.

🎩
1. Visit the details page in storybook
1. Click on a nav item
1. While that item is still active, hit escape. Nothing should happen

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Frame/tests/Frame.test.tsx
+++ b/src/components/Frame/tests/Frame.test.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import {CSSTransition} from '@material-ui/react-transition-group';
 import {animationFrame} from '@shopify/jest-dom-mocks';
-import {documentHasStyle} from 'test-utilities';
+import {documentHasStyle, mountWithApp} from 'test-utilities';
 // eslint-disable-next-line no-restricted-imports
 import {mountWithAppProvider, trigger} from 'test-utilities/legacy';
 import {
@@ -9,7 +9,7 @@ import {
   ContextualSaveBar as PolarisContextualSavebar,
   Loading as PolarisLoading,
 } from 'components';
-import Frame from '../Frame';
+import Frame, {FrameProps} from '../Frame';
 import {
   ContextualSaveBar as FrameContextualSavebar,
   Loading as FrameLoading,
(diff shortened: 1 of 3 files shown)


kyledurand:
I'm just on a call right now but I'll add a test before the end of the day

src/components/Frame/tests/Frame.test.tsx
```diff
@@ -41,196 +41,241 @@ describe('<Frame />', () => {
...
 
-  // JSDOM 11.12.0 does not support setting/reading custom properties so we are
-  // unable to assert that we set a custom property
-  // See https://github.com/jsdom/jsdom/issues/1895
-  // eslint-disable-next-line jest/no-disabled-tests
-  it.skip('sets a root property with global ribbon height if passed', () => {
-    mountWithAppProvider(<Frame globalRibbon={<div />} />);
-    expect(documentHasStyle('--global-ribbon-height', '0px')).toBe(true);
-  });
+      const {onNavigationDismiss} = frame.props as FrameProps;
```
tmlayton: Can we avoid this casting?

kyledurand: Not that I know of. I think casting is ok in tests? 

The reason I think we need casting here is because this component is an older class component and I'm using the new test utils on it and thus can't infer the prop types. Open to trying other suggestions though

AndrewMusgrave: It seems the type signature for root-level props are inferred at `root.prop(x: never)`. I'm guessing it's probably intentional to discourage the use of prop manipulation and such on the root, e.g

```tsx
const spy = jest.fn();
const frame = mountWithApp(<Frame navigation={<div /> onNavigationDismiss={spy}} />
frame.trigger('onNavigationDismiss'); // frame.prop('onNavigationDismiss')();
expect(spy).toHaveBeenCalled(1);

// is essentially
const spy = jest.fn();
mountWithApp(<Frame navigation={<div /> />
spy();
expect(spy).toHaveBeenCalled(1);
```

However, in this situation we can omit this line.
```suggestion
```

src/components/Frame/tests/Frame.test.tsx
```diff
@@ -41,196 +41,241 @@ describe('<Frame />', () => {
...
 
-  // JSDOM 11.12.0 does not support setting/reading custom properties so we are
-  // unable to assert that we set a custom property
-  // See https://github.com/jsdom/jsdom/issues/1895
-  // eslint-disable-next-line jest/no-disabled-tests
-  it.skip('sets a root property with global ribbon height if new props are passed', () => {
-    const frame = mountWithAppProvider(<Frame />);
-    frame.setProps({globalRibbon: <div />});
-    expect(documentHasStyle('--global-ribbon-height', '0px')).toBe(true);
+      expect(onNavigationDismiss).not.toHaveBeenCalled();
```
AndrewMusgrave:
```suggestion
      expect(spy).not.toHaveBeenCalled();
```

src/components/Frame/tests/Frame.test.tsx
```diff
@@ -41,196 +41,241 @@ describe('<Frame />', () => {
...
+          showMobileNavigation
+          onNavigationDismiss={spy}
+        />,
+        {mediaQuery: {isNavigationCollapsed: true}},
+      );
+      frame
+        .find('div', {id: 'AppFrameNav'})!
+        .trigger('onKeyDown', {key: 'Escape'});
+
+      const {onNavigationDismiss} = frame.props as FrameProps;
```
AndrewMusgrave: ^

src/components/Frame/tests/Frame.test.tsx
kyledurand: Ah yes lol 👍 

Simply a case of end of day brain here. I'll commit your suggestions and ship tomorrow morning unless other comments arise

PR #2616, opened 2020-01-13
[CustomProperties] Added test for custom properties

AndrewMusgrave:

We have incorrect custom properties

Adding utilities and tests for custom properties

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist


AndrewMusgrave:
Before I start adding tests/polish I was wondering if you have any initial thoughts cc/ @danrosenthal @tmlayton . I noticed we have some incorrect custom properties so I wrote out some utilities and added a test that'll give us more confidence using custom properties moving forward.

tmlayton:
> We have incorrect custom properties

Which ones?

AndrewMusgrave:
They'll fail on [CI now](https://circleci.com/gh/Shopify/polaris-react/37803). Here the [commit](https://github.com/Shopify/polaris-react/pull/2616/commits/c3d641610b22d65a1c7c77d284b0b646845498f4) fixing them.

tmlayton:
> initial thoughts

I like it :)

danrosenthal:
> initial thoughts

> I like it :)

Yes, that!

BPScott:
Loving this.

As an implementation hook - I wonder if we could use postcss and the postcss-scss plugin - which we already use as part of our build, and already have available via sewing-kit to generate the AST to check the custom properties rather than bring in a new dependency?

This utility isn't something we want to distribute so it feels like it should live in the tests folder, like `build.test.js`

AndrewMusgrave:
> As an implementation hook - I wonder if we could use postcss and the postcss-scss plugin - which we already use as part of our build, and already have available via sewing-kit to generate the AST to check the custom properties rather than bring in a new dependency?

One idea I was going to explore next was to make this a module that could be shared between other repo's (polaris-rails, web, core, etc). I'm not too familiar with postcss plugins, could we fail the build if we found incorrect custom properties? What kind of warnings/errors can we surface. And if we wanted to share this to other projects, they would have to use postcss which they may not be using.

> This utility isn't something we want to distribute so it feels like it should live in the tests folder, like build.test.js

100% agree. I placed it in `src/utilities` because I assumed `/tests` didn't transpile typescript code. After a check, it looks like it does, any idea why the build test isn't typescript?

BPScott:
> 100% agree. I placed it in `src/utilities` because I assumed `/tests` didn't transpile typescript code. After a check, it looks like it does, any idea why the build test isn't typescript?

Not a clue. Perhaps it used to not work then at some point we chnaged the TS root directory from `src` to `.`

danrosenthal:
Whoops, I accidentally pushed some of my changes to this branch. ☝️

src/utilities/theme/tokens.ts
```diff
@@ -24,6 +24,7 @@ const Overrides = {
   nonNullContent: "''",
   choiceSize: rem('20px'),
   choiceMargin: rem('1px'),
+  controlBorderWidth: rem('2px'),
```
AndrewMusgrave: Grabbed this from your PR Dan 😄

AndrewMusgrave:
Talked to Dan R. and we're hoping to ship this so we'll have confidence incorrect custom properties won't be used. And I can continue to iterate and add tests to this before we release it as a package that'll be consumed by other libraries during our rollout.

.eslintignore
```diff
@@ -11,3 +11,5 @@ node_modules
 /styles.scss
 /styles
 /src/styles/polaris-tokens
+/scripts/analyze-custom-properties/dist
+custom-properties-data.json
```
BPScott: I'm not a fan of adding new stuff into the root. Can we put this (and analyze-custom-properties/dist is possible) into the build folder?

BPScott:
With `splash` we use babel-node to execute typescript directly instead of needing to an intermediate build step. Yuo can probably do the same here to get rid of a bunch of your build scripting and that dist folder.

Ask @amrocha for more info

AndrewMusgrave:
> With splash we use babel-node to execute typescript directly instead of needing to an intermediate build step. You can probably do the same here to get rid of a bunch of your build scripting and that dist folder.

Most of the build script will be removed when this is a separate package. The other part builds the custom properties we expect. We could just use another runtime for that. I usually use `ts-node`. Is there a reason to prefer `babel-node` or just a preference?

BPScott:
> I usually use ts-node. Is there a reason to prefer babel-node or just a preference?

No strong reason to go one over the other - We're already using babel-node for other stuff so feels like keeping to existing tooling feels handier.

PR #2622, opened 2020-01-15
[DetailsPage] Add context control

kyledurand:

This adds a context control to our details page. Allowing quick comparison from the theme provider logo (present on the frame story page) and context control

## 

### How to 🎩

You just need to checkout the details page and make sure the context control looks good from within the nav on small screens, and in the top bar on larger screens

### 🎩 checklist


tmlayton:
Styling is off on small screens, otherwise looks good!

Screen Shot 2020-01-16 at 9 11 23 PM

config/typescript/styles.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module '*.scss' {
```
BPScott: This should already be covered by our ts config importing https://github.com/Shopify/sewing-kit/blob/master/packages/typescript-configs/definitions/styles.d.ts, do we need this too?

kyledurand: I was having weird typescript issues with this today. Seemed like I needed to add it but maybe there's just something wonky going on in this branch. The CI failure seems to confirm that

danrosenthal:
Need anything else here to get this shipped?

Are you aware of the `Skip changelog` label? It allows that check to pass for changes we don't care about documenting.

kyledurand:
> Need anything else here to get this shipped?

TS doesn't seem to like the addition of the scss file in this directory. Adding the module didn't seem to help either. @BPScott would you be down for pairing on a fix for that?

BPScott:
Had a super speedy chat with Kyle - it was down to playground not being in the tsconfig.json and thus those files not picking up the project config.

As a bonus that file is now checked when you do `yarn run type-check`

PR #2625, opened 2020-01-16
Use export *  in component indexes and subcomponent reexports

### LINKED ISSUE #2619: Proposal: Use export * in component indexes
A bit of a follow on from https://github.com/Shopify/polaris-react/issues/1959

I've been looking at the feasibility of updating our build. Currently we do a complicated dance where we compile using typescript, then transpile that using babel. I've been investigating if instead we could just have babel read our TS files directly, which ends up being a fair bit simpler, however we need to modify our code slightly so Babel can handle type reexports correctly.

We've had this problem for a while - our Storybook build also uses babel to transform our TS files - however we've been able to ignore/suppress the generated warnings till now.

The issue is that we have errors when we reexport types. e.g.

```ts
// src/components/ActionList/index.ts
export {ActionList, ActionListProps} from './ActionList';
```

When compiling with `tsc` it looks over the whole project and it knows that `ActionListProps` is a type and thus should be removed from the compiled output. However babel only looks at a single file at a time so it doesn't know if ActionListProps is a type or value, so it assumes it is a value and leaves the export in. This however causes an issue in bundlers because when webpack/rollup processes this file it tries to find ActionListProps and then fails because it doesn't exist (as it is a type not a value). In webpack this results in those `export ActionListProps was not found in...` warnings, and is a hard error in rollup. Thus we need a way to stop these errors.

I would like to propose changing how we export in component indexes, so that we can give babel the information it needs to properly remove type imports.

TypeScript 3.8 will introduce [type-only imports/exports](https://devblogs.microsoft.com/typescript/announcing-typescript-3-8-beta/#type-only-imports-exports) (final release expected in mid February), which is new syntax that will give babel the information to remove the type exports. So we could rewrite the ActionList index as:

```ts
// src/components/ActionList/index.ts
export {ActionList} from './ActionList';
export type {ActionListProps} from './ActionList';
```

However I feel this repetition is a somewhat annoying. Instead I think we should use star exports, to export everything from the `ActionList.ts` file. This sidesteps the issue entirely as bundlers will take whatever is there instead of us having to repeat the names of components and their props:

```ts
// src/components/ActionList/index.ts
export * from './ActionList';
```

This feels safe to me as the majority of the time components only export themselves and their props. There may be some cases where we would want to tweak the file layout slightly to ensure exports that are only used within a component are not exposed outside the component but those cases are in the small minority.

---

This will fix the majority of webpack/rollup's complaints today, however there are two other cases that the above does not address:

The rexporting of all components in `src/components/index.ts` should move to use the `export type` notation and make two export calls - this duplication is a little meh but it allows us to strongly control what gets exported from library instead of a blanket export - this helps protect us against accidentally exposing values that should be internal.

The reexporting of types in our utilities e.g. `export {Theme, ThemeConfig, CustomPropertiesLike, ColorScheme} from './types';` in `src/utilities/theme/index.ts`. In these cases we should also use the `export type` notation too as there would be no values in that type file, so no repeated reexports.

---

Does this seem like a reasonable proposal? Do you have any concerns?
If not I'll spin up a PR to make the changes to component indexes this/next week.

BPScott:

Updating component indexes to use `export *`. Fixes #2619

### 277 changed files. Really Ben?

Yeah, sorry. It probably best to review this PR by commit by commit.

This PR does a few things.

First, a trio of commits that updates all the component (and subcomponent) files (e.g. `src/components/ActionMenu/ActionMenu.tsx` and `src/components/Frame/components/CSSAnimation.tsx`) to ensure that the component file exports only the same as what the current index exports.

Second, now that all component files and their indexes (e.g. `src/components/ActionMenu/ActionMenu.tsx` and `src/components/ActionMenu/index.ts`) export the same things, update all component indexes to use `export *`.

Thirdly, update all subcomponent folder indexes (e.g. `src/components/Frame/components/index.ts`) to reexport to use `export *`

### How to 🎩

Ensure type-check and tests pass.
Note that the contents of the `utilities` folder and `src/components/index.ts` have not been touched so the public exports accessed by importing `@shopify/polaris` are untouched, with the exception of adding the SheetProps, which seem useful publicly.

Once https://github.com/Shopify/web/pull/22624 has been merged into web to stop them reaching into our internal types, then ensure that web passes a type-check with these changes:

- [in polaris] `yarn run build-consumer web`
- [in web] `yarn run type-check`


src/components/Avatar/Avatar.tsx
```diff
@@ -9,15 +9,15 @@ import {Image} from '../Image';
 import styles from './Avatar.scss';
 import * as avatars from './images';
 
-export type Size = 'small' | 'medium' | 'large';
+type Size = 'small' | 'medium' | 'large';
```
BPScott: This sort of thing crops up a lot. Using this particular case as an example, `Size` was exported here but was never reexported by `src/Avatar/index.ts` and was never used in any other file, so there is no value in exporting it at all.

src/components/Avatar/tests/Avatar.test.tsx
```diff
@@ -116,10 +115,6 @@ describe('<Avatar />', () => {
   });
 
   describe('styleClass', () => {
-    it('defaults to five styles', () => {
```
BPScott: This is testing an internal implementation detail, I didn't want to expose STYLE_CLASSES so it was easiest to remove this test instead of moving files around to avoid exposing STYLE_CLASSES

src/components/Button/tests/Button.test.tsx
```diff
@@ -168,7 +168,7 @@ describe('<Button />', () => {
 
     it('does not render the markup for the icon if none is provided', () => {
       const button = mountWithAppProvider(<Button />);
-      expect(button.find(IconWrapper).exists()).toBe(false);
```
BPScott: IconWrapper shouldn't be exposed outside of Button.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -11,7 +11,7 @@ import {Error, Key, CheckboxHandles} from '../../types';
 
 import styles from './Checkbox.scss';
 
-export interface BaseProps {
+export interface CheckboxProps {
```
BPScott: This crops up in a few files - a BaseProps interface and the component's props that extend from that without adding anything news - so remove BaseProps.

src/components/ChoiceList/tests/ChoiceList.test.tsx
```diff
@@ -3,10 +3,10 @@ import {mountWithApp} from 'test-utilities';
 // eslint-disable-next-line no-restricted-imports
 import {mountWithAppProvider, ReactWrapper} from 'test-utilities/legacy';
 import {RadioButton, Checkbox, InlineError, errorTextID} from 'components';
-import {ChoiceList, ChoiceDescriptor} from '../ChoiceList';
+import {ChoiceList, ChoiceListProps} from '../ChoiceList';
 
 describe('<ChoiceList />', () => {
-  let choices: ChoiceDescriptor[];
+  let choices: ChoiceListProps['choices'];
```
BPScott: Where possible we should look at the values that live on the Props interface  instead of exposing the interfaces the Props use

src/components/DropZone/utils/index.ts
```diff
@@ -1,4 +1,4 @@
-import {DropZoneEvent} from '../types';
+type DropZoneEvent = DragEvent | React.ChangeEvent<HTMLInputElement>;
```
BPScott: This type was only used in this file, so inline it instead of having a separate types.ts file.

src/components/Filters/Filters.tsx
```diff
@@ -180,7 +180,7 @@ class Filters extends React.Component<ComposedProps, State> {
                 <TextStyle
                   variation={
                     this.props.disabled || filter.disabled
-                      ? VariationValue.Subdued
```
BPScott: By not mentioning the enum here we can avoid exposing it in TextStyle.

src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.tsx
```diff
@@ -35,9 +35,7 @@ interface State {
   proxyButtonsWidth: ComputedProperty;
 }
 
-export const FILTER_FIELD_MIN_WIDTH = 150;
-
-export const FILTER_FIELD_CUSTOM_PROPERTY = '--textfield-min-width';
```
BPScott: This was never used anywhere, but because it was exported TS didn't warn us of that.

src/components/Image/tests/Image.test.tsx
```diff
@@ -33,7 +33,7 @@ describe('<Image />', () => {
     it('renders an img with a srcSet', () => {
       const src =
         'https://cdn.shopify.com/s/assets/admin/checkout/settings-customizecart-705f57c725ac05be5a34ec20c05b94298cb8afd10aac7bd9c7ad02030f48cfa0.svg';
-      const srcSet: SourceSet[] = [
+      const srcSet = [
```
BPScott: This typing is unneeded everything all compiles fine without it

src/components/Page/components/Header/Header.tsx
```diff
@@ -2,23 +2,37 @@ import React from 'react';
...
 import {Breadcrumbs, BreadcrumbsProps} from '../../../Breadcrumbs';
-
 import {Pagination, PaginationDescriptor} from '../../../Pagination';
 import {ActionMenu, hasGroupsWithActions} from '../../../ActionMenu';
 
-import {HeaderPrimaryAction} from '../../types';
 import {Title, TitleProps} from './components';
 import styles from './Header.scss';
 
+interface PrimaryAction
```
BPScott: This interface previously lived in the types file but it was only used here, so just inline it

src/components/Page/index.ts
```diff
@@ -1,4 +1,3 @@
 import Page, {PageProps} from './Page';
 
 export {Page, PageProps};
-export {HeaderPrimaryAction} from './types';
```
BPScott: This was never used outside of the current component - people should use `PageProps['primaryAction']` instead

src/components/Popover/Popover.tsx
```diff
@@ -12,10 +12,10 @@ import {PreferredPosition, PreferredAlignment} from '../PositionedOverlay';
 import {Portal} from '../Portal';
 import {portal} from '../shared';
 import {useUniqueId} from '../../utilities/unique-id';
-import {CloseSource, Pane, PopoverOverlay, Section} from './components';
+import {PopoverCloseSource, Pane, PopoverOverlay, Section} from './components';
```
BPScott: Name this import/export correctly where it is initially defined, avoids having to rename it in the `Popover/index.ts` file

src/components/Sheet/tests/Sheet.test.tsx
```diff
@@ -41,9 +41,12 @@ describe('<Sheet />', () => {
       {mediaQuery: {isNavigationCollapsed: true}},
     );
 
-    expect(sheet.find(CSSTransition).props().classNames).toStrictEqual(
-      BOTTOM_CLASS_NAMES,
-    );
+    expect(sheet.find(CSSTransition).props().classNames).toStrictEqual({
+      enter: 'Bottom enterBottom',
```
BPScott: This is a bit meh, but I think it's a bit better than exposing the BOTTOM_CLASS_NAMES RIGHT_CLASS_NAMES or extracting them out into a file that used by both Sheet and Sheet.test.tsx

src/components/Spinner/tests/Spinner.test.tsx
```diff
@@ -75,12 +75,10 @@ describe('<Spinner />', () => {
     it('a large spinner with an unavailable color warns in development', () => {
       process.env.NODE_ENV = 'development';
 
-      const color = 'black' as Color;
```
BPScott: That's not even a valid Color *shrug*

src/components/Collapsible/Collapsible.tsx
```diff
@@ -34,7 +34,7 @@ interface State {
 
 const ParentCollapsibleExpandingContext = createContext(false);
 
-class Collapsible extends React.Component<CollapsibleProps, State> {
+class CollapsibleInner extends React.Component<CollapsibleProps, State> {
```
BPScott: Kill off cases where we export a default value - now things that use AppProvider are exported with a named component, which means that the component index file doesn't need to rename the default export to a named export

src/components/PositionedOverlay/tests/PositionedOverlay.test.tsx
```diff
@@ -124,99 +120,6 @@ describe('<PositionedOverlay />', () => {
     });
   });
 
-  describe('intersectionWithViewport', () => {
```
BPScott: This is testing internal details, presumably because it's hard to to test this using just the component.

Move this function to math.ts and do the tests in math.test.ts to avoid exporting `intersectionWithViewport` from PositionedOverlay.tsx

PR #2636, opened 2020-01-20
Use export * in utility indexes too

BPScott:

Following on from #2619 and #2625 I looked at the utility index files and spotted that in almost all cases our index files rexport everything in the files they reference, or what they references is never used externally, so they can also use `export *` too.

* Updates utility indexes to use `export *` because doing things one way everywhere is easier than doing things one way in `components` and another way in `utilities`

[I did run into some issues](https://github.com/Shopify/polaris-react/pull/2636/files#diff-4eb03b66a1f37179b50db6bcb635f31cR4) with some interfaces that get used as the type that gets pushed into a React context.

I could either have those as a type or export the interface. I figured exporting the interface was an acceptable compromise as "a slightly leaky internal export that doesn't bleed into our public exports" was better than "disabling our linting rule that prefers the usage of interfaces over types".

### How to 🎩

Type check and tests pass

### DIFF
--- a/src/utilities/scroll-lock-manager/scroll-lock-manager.ts
+++ b/src/utilities/scroll-lock-manager/scroll-lock-manager.ts
@@ -2,9 +2,7 @@ import {isServer} from '../target';
 
 export const SCROLL_LOCKING_ATTRIBUTE = 'data-lock-scrolling';
 
-export const SCROLL_LOCKING_WRAPPER_ATTRIBUTE = 'data-lock-scrolling-wrapper';
-
-export const SCROLL_LOCKING_CUSTOM_PROPERTY = '--scroll-lock-body-padding';
+const SCROLL_LOCKING_WRAPPER_ATTRIBUTE = 'data-lock-scrolling-wrapper';
 
 let scrollPosition = 0;
 
--- a/src/utilities/theme/index.ts
+++ b/src/utilities/theme/index.ts
@@ -1,14 +1,10 @@
-export {ThemeContext} from './context';
+export * from './context';
 
-export {useTheme} from './hooks';
+export * from './hooks';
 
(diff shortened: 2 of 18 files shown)


src/utilities/scroll-lock-manager/scroll-lock-manager.ts
```diff
@@ -2,9 +2,7 @@ import {isServer} from '../target';
 
 export const SCROLL_LOCKING_ATTRIBUTE = 'data-lock-scrolling';
 
-export const SCROLL_LOCKING_WRAPPER_ATTRIBUTE = 'data-lock-scrolling-wrapper';
-
-export const SCROLL_LOCKING_CUSTOM_PROPERTY = '--scroll-lock-body-padding';
```
amrocha: are you sure we can remove this?

src/utilities/theme/index.ts
```diff
@@ -1,14 +1,10 @@
-export {ThemeContext} from './context';
+export * from './context';
 
-export {useTheme} from './hooks';
+export * from './hooks';
 
-export {Theme, ThemeConfig, CustomPropertiesLike, ColorScheme} from './types';
+export {ThemeConfig} from './types';
```
amrocha: Why not export * here?

BPScott: There's a bunch of other types  in that file that are used within this folder, but not externally. As most of this is going to be extracted into polaris-tokens pretty soon it didn't seem worth the refactoring to split this into a types and private-types file.

src/utilities/scroll-lock-manager/scroll-lock-manager.ts
BPScott: I checked for usages and there were none outside the current file, I then removed the `export` and it complained that it was never used inside this file

PR #2639, opened 2020-01-21
[Tabs] Remove the visited state styling for tabs with url

moimael:

Fixes https://github.com/Shopify/store/issues/12366

When tabs have a `url` prop, the styling becomes different because a `a` tag is rendered with a different `visited` styling. The selected tab change color when hovered. It doesn't show the same behaviour when the tab has no `url` prop.

This PR remove the visited state styling to make the behaviour consistent.

## 

### How to 🎩

<details>

```jsx
import {Page, Card, Tabs} from '../src';

    [],
  );

  const tabs = [
    {
      id: 'all-customers',
      content: 'All',
      url: 'https://www.google.com',
    },
    {
      id: 'accepts-marketing',
      content: 'Accepts marketing',
    },
  ];

  return (
    <Page title="Playground">
      <Card>
          </Card.Section>
        </Tabs>
      </Card>
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -11,6 +11,7 @@
 
 - Fixed scrolling with scrollbar not working in Popover when content changes on scroll ([#2627](https://github.com/Shopify/polaris-react/pull/2627))
 - Work around a build crash when using create-react-app due to a bug in css parsing in `postcss-custom-properties` ([#2643](https://github.com/Shopify/polaris-react/pull/2643))
+- Removed the `visited` CSS styling for tabs using the `url` prop ([#2639](https://github.com/Shopify/polaris-react/pull/2639))
 
 ### Documentation
 
--- a/src/components/Tabs/Tabs.scss
+++ b/src/components/Tabs/Tabs.scss
@@ -83,10 +83,6 @@ $focus-state-box-shadow-color: rgba(92, 106, 196, 0.8);
     }
   }
 
-  &:visited {
-    color: color('ink', 'lighter');
-  }
-
   &.Tab-selected {
     // stylelint-disable-next-line selector-max-class
     .Title {


moimael:
> 🎩 ✅
> 
> You'll have to update UNRELEASED.md to include your changes.
> 
> Also are there any UX concerns with a link that doesn't have a hover effect? Would it be a better solution to not render the tag if that tab is selected?

Thanks will do! I'm not sure honestly, I can't see any reason why we have this behaviour on tabs right now. Maybe someone with more background can chime in ?

UNRELEASED.md
```diff
@@ -8,6 +8,8 @@
 
 ### Bug fixes
 
+- Remove the `visited` CSS styling for tabs using the `url` prop
```
```suggestion
- Remove the `visited` CSS styling for tabs using the `url` prop ([#2639](https://github.com/Shopify/polaris-react/pull/2639))
```

kaelig:
👋 I removed myself from the list of reviewers, and looping in @dpersing who might have an opinion on the `:visited` state from an accessibility standpoint.

yourpalsonja:
@mirualves you've been doing some work on tabs, wondering what your thoughts are!

mirualves:
This is a symptom of the deeper problem of using the `Tabs` component to convey `Navigation`. I don't see a problem with this fix in particular, but it's not going to solve the underlying accessibility issue.
https://github.com/Shopify/polaris-ux/issues/198

cc @dpersing

dpersing:
>looping in @dpersing who might have an opinion on the :visited state from an accessibility standpoint.

I'd like to move away from using tablists as navigation (see https://github.com/Shopify/polaris-ux/issues/198 for rationale). From that standpoint, I don't think that tabs should have a visited state.

mediumjones:
For context, this PR is attempting to fix a visual discrepancy in the interaction states that we noticed between our component and the [Tabs component](https://polaris.shopify.com/components/navigation/tabs#navigation) example on polaris.shopify.com

**polaris.shopify**
- Selected tab uses Indigo border and Ink/Ink text color signifier
Screen Recording 2020-01-22 at 11 22 05 AM


**Our component**
- Selected tab only uses Indigo border
Screen Recording 2020-01-22 at 11 23 38 AM

I assumed polaris.shopify was correct because the tab is already selected and shouldn't indicate it can be clicked again with a changing hover state.

I'm okay with either as long as it's accessible and we're using it consistently :D

dfmcphee:
It looks like this caused a bit of a regression with the tab styling. Now any tab that I have visited before uses the darker text, and the ones I haven't are still subdued.

Screen Shot 2020-02-24 at 1 54 56 PM

PR #2643, opened 2020-01-21
Update sass to work around CRA build crashes

BPScott:

Fixes https://github.com/Shopify/polaris-react/issues/2629

The current version of CRA' s build uses a version of https://github.com/postcss/postcss-custom-properties that crashes when it comes across ~var()s inside calc()s~ a very specific order of calls: `  top:calc(var(--control-border-width)*-1)`. See https://github.com/Shopify/polaris-react/issues/2629#issuecomment-577319704 for more info.  This has been fixed in the newest version of postcss-custom-properties but that's not made its way into CRA (via postcss-preset-env) and probably won't be for some time.

Flip the order of values (yes I know this sounds silly)

### How to 🎩

Ask @dleroux nicely to see that there are no visual regressions.

- Do a build `yarn run build`
- Create a new CRA project with `yarn create react-app testing`
- Copy the build styles.css into the CRA app and import the styles from CRA's src/ index.js file and see if a build compiles using `yarn run build`

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -10,6 +10,7 @@
 ### Bug fixes
 
 - Fixed scrolling with scrollbar not working in Popover when content changes on scroll ([#2627](https://github.com/Shopify/polaris-react/pull/2627))
+- Work around a build crash when using create-react-app due to a bug in css parsing in `postcss-custom-properties` ([#2643](https://github.com/Shopify/polaris-react/pull/2643))
 
 ### Documentation
 
--- a/src/styles/shared/_controls.scss
+++ b/src/styles/shared/_controls.scss
@@ -110,10 +110,10 @@
       &::before {
         content: '';
         position: absolute;
-        top: calc(var(--control-border-width) * -1);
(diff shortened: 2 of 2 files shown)


dleroux:
It looks like this is working Ben. The same approach is being used here: https://github.com/Shopify/polaris-react/blob/master/src/styles/foundation/_focus-ring.scss#L31 I assume this one will need to change as well?

BPScott:
> The same approach is being used here: https://github.com/Shopify/polaris-react/blob/master/src/styles/foundation/_focus-ring.scss#L31 I assume this one will need to change as well?

Funky. That's been in there for a while and hasn't caused crashes. Along with a bunch of other places, searching our codebase for `calc(var(`.

I've done some digging and it's not `calc(var())` that's the problem and now I'm mad. https://github.com/Shopify/polaris-react/issues/2629#issuecomment-577319704

Changing `calc(var(--control-border-width) * -1);` to `calc(-1 * var(--control-border-width));` should get this to compile.

PR #2646, opened 2020-01-22
Improve focus ring animation

kyledurand:
This PR:
1. Cleans up TextField.scss
1. Changes focus ring from border to box-shadow (smooths transitions)
1. Changes transition type from scale to box-shadow (prevents seeing the box-shadow scale down underneath larger items like text fields.
1. Minor improvements to focus rings border-radius

The focus ring looked good on smaller components such as radio buttons and checkboxes but not so much on larger components, like text fields.

@dleroux will be handling z-index overlapping in a separate PR

image
[Screencast](https://screenshot.click/screencast_2020-01-22_11-56-01.mp4)

## 

### How to 🎩

<details>

```jsx
import {
  Card,
  Page,
  RadioButton,
  Stack,
  TextField,
  Select,
  Checkbox,
} from '../src';

  const [radioChecked, setRadioChecked] = useState(false);
  const [textValue, setTextValue] = useState('');
  const handleRadioChange = useCallback(
    (newChecked) => setRadioChecked(newChecked),
    [],
  );
  const handleTextChange = useCallback((value) => setTextValue(value), []);
  return (
    <Page title="Playground">
      <Card>
        <Card.Section>
          <Stack vertical>
            <Checkbox
              name="name"
              label="label"
              checked={checked}
              onChange={handleChange}
            />
            <RadioButton
              name="name"
              label="label"
              checked={radioChecked}
              onChange={handleRadioChange}
            />
            <TextField
              label="Default"
              onChange={handleTextChange}
              value={textValue}
              autoComplete="off"
            />
            <TextField
              label="Default"
              prefix="$"
              readOnly
              onChange={() => {}}
              autoComplete="off"
            />
            <TextField
              label="Default"
              prefix="$"
              disabled
              onChange={() => {}}
              autoComplete="off"
            />
            <TextField
              label="Error"
              onChange={() => {}}
              autoComplete="off"
              error="Some error"
            />
            <TextField
              label="Weight"
              type="number"
              value="1"
              onChange={() => {}}
              connectedRight={
                <Select
                  value="select"
                  label="Weight unit"
                  onChange={() => {}}
                  labelHidden
                  options={['kg', 'lb']}
                />
              }
            />
          </Stack>
        </Card.Section>
      </Card>
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/DropZone/DropZone.scss
+++ b/src/components/DropZone/DropZone.scss
@@ -176,14 +176,16 @@ $dropzone-stacking-order: (
 
 .globalTheming {
   &.DropZone {
-    @include focus-ring;
-
     &:not(.focused) {
       // stylelint-disable-next-line selector-max-specificity
       &::after {
         @include reset-after;
       }
     }
+  }
+
+  .Container {
+    @include focus-ring($border-width: rem(2px));
   }
 
   &.Overlay {
@@ -204,7 +206,10 @@ $dropzone-stacking-order: (
 
   &.focused {
     &:not(.isDisabled) {
-      @include focus-ring($style: 'focused');
+      // stylelint-disable-next-line selector-max-specificity
+      .Container {
+        @include focus-ring($style: 'focused');
+      }
     }
   }
 }
--- a/src/components/DropZone/DropZone.scss
+++ b/src/components/DropZone/DropZone.scss
@@ -4,7 +4,7 @@ $dropzone-padding: rem(15px);
 $dropzone-border-style: dashed;
 $dropzone-border-color: var(--p-border, color('ink', 'lightest'));
 $dropzone-border-color-disabled: var(--p-border-disabled, color('sky'));
-$dropzone-border-radius: border-radius();
+$dropzone-border-radius: var(--p-border-radius-base, border-radius());
 $dropzone-background: var(--p-surface, color('white'));
 $dropzone-min-height-extra-large: rem(205px);
 $dropzone-min-height-large: rem(160px);
--- a/src/styles/foundation/_focus-ring.scss
+++ b/src/styles/foundation/_focus-ring.scss
@@ -1,7 +1,7 @@
 /// Sets the focus ring for an interactive element
 /// @param {String} $size - The size of the border radius on the focus ring.
-/// @param {String} $border-width - The border width of your element in rems.
 /// @param {String} $style - Focus ring state.
+/// @param {Number} $border-width - The border width of your element in rems.
 ///
 
 @mixin focus-ring($size: 'base', $border-width: rem(0), $style: 'base') {
--- a/src/styles/foundation/_focus-ring.scss
+++ b/src/styles/foundation/_focus-ring.scss
@@ -1,12 +1,17 @@
 /// Sets the focus ring for an interactive element
 /// @param {String} $size - The size of the border radius on the focus ring.
-/// @param {String} $offset - Focus ring offset border width + 1px.
+/// @param {String} $border-width - The border width of your element in rems.
 /// @param {String} $style - Focus ring state.
 ///
 
-@mixin focus-ring($size: 'base', $offset: rem(1px), $style: 'base') {
+@mixin focus-ring($size: 'base', $border-width: rem(0), $style: 'base') {
   $stroke: rem(2px);
-  $position: calc(#{$offset} + #{$stroke});
+  $offset: calc(#{$border-width} + #{rem(1px)});
+  $border-radius: if(
+    $size == 'wide',
+    var(--p-border-radius-wide),
+    var(--p-border-radius-base)
+  );
 
   @if $style == 'base' {
     position: relative;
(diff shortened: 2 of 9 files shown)


dleroux:
The nice thing about the box-shadow is that there is no need for the offset as an argument anymore!

src/styles/foundation/_focus-ring.scss
```diff
@@ -20,27 +20,19 @@
       bottom: calc(-1 * #{$position});
       left: calc(-1 * #{$position});
       display: block;
-      border: $stroke solid var(--p-focused);
-      transition: opacity duration(fast) var(--p-ease),
-        transform duration(fast) var(--p-ease);
-      opacity: 0;
-      transform: scale(0.8);
+      box-shadow: 0 0 0 -#{$offset} var(--p-focused);
```
dleroux: The offset would vary because of the `::after` selector. Without a border on the element, it would be 1px, but if the element has a 2px border lets say, for example checkbox, the offset will be `2px + 1px`

src/styles/foundation/_focus-ring.scss
```diff
@@ -20,27 +20,19 @@
...
+      transition: box-shadow duration(fast) var(--p-ease);
 
       @if $size == 'wide' {
-        border-radius: calc(
-          var(--p-border-radius-wide) + #{$position} - #{$offset} + #{rem(1px)}
```
dleroux: the varying border is why these we necessary.

src/styles/foundation/_focus-ring.scss
dleroux: We should look at the performance here as well because animating the scale is better performant than the box-shadow. The shadow will cause a re-paints but the scaling and opacity will not.

tmlayton:
Let’s get some [LICEcap](https://www.cockos.com/licecap/) or [Kap](https://getkap.co/) rocking and see some gifs! 😄

tmlayton:
Math appears to be off for elements expecting a 1px gap between itself and the focus ring:
Screen Shot 2020-01-22 at 9 42 54 PM

kyledurand:
Sorry for the early ping on this last week folks. I put this up initially just to pair with Dan on it and it automatically pinged reviewers. Anyway, after a rather large refactor, here are some gifs:

Slow mo:
slow

Wide:
Wide

Multiple:
multiple

Dropzone:
Dropzone gif is too big 😞. Here's a still:
image

tmlayton:
Each gif is a gift to me

src/styles/foundation/_focus-ring.scss
```diff
@@ -1,12 +1,17 @@
...
-/// @param {String} $offset - Focus ring offset border width + 1px.
+/// @param {String} $border-width - The border width of your element in rems.
 /// @param {String} $style - Focus ring state.
 ///
 
-@mixin focus-ring($size: 'base', $offset: rem(1px), $style: 'base') {
+@mixin focus-ring($size: 'base', $border-width: rem(0), $style: 'base') {
   $stroke: rem(2px);
-  $position: calc(#{$offset} + #{$stroke});
+  $offset: calc(#{$border-width} + #{rem(1px)});
```
kyledurand: Adding `calc` here allows us to pass in custom properties but by definition we prefer a rem value.

src/styles/foundation/_focus-ring.scss
```diff
@@ -1,12 +1,17 @@
 /// Sets the focus ring for an interactive element
 /// @param {String} $size - The size of the border radius on the focus ring.
-/// @param {String} $offset - Focus ring offset border width + 1px.
+/// @param {String} $border-width - The border width of your element in rems.
```
kyledurand: `@param {string}` doesn't seem to fit anymore. Is this a `Number` now?

src/components/DropZone/DropZone.scss
```diff
@@ -186,6 +184,10 @@ $dropzone-stacking-order: (
     }
   }
 
+  .Container {
+    @include focus-ring($border-width: rem(2px));
```
danrosenthal:
```suggestion
    @include focus-ring($border-width: border-width(thick));
```

src/styles/foundation/_focus-ring.scss
danrosenthal: Yup, this would become `Number`

src/styles/foundation/_focus-ring.scss
```diff
@@ -1,12 +1,17 @@
...
 /// @param {String} $style - Focus ring state.
 ///
 
-@mixin focus-ring($size: 'base', $offset: rem(1px), $style: 'base') {
+@mixin focus-ring($size: 'base', $border-width: rem(0), $style: 'base') {
```
danrosenthal: 👍

src/components/DropZone/DropZone.scss
```diff
@@ -186,6 +184,10 @@ $dropzone-stacking-order: (
     }
   }
 
+  .Container {
+    @include focus-ring($border-width: border-width(thick));
+  }
+
   &.Overlay {
     // Same calculation used for focus-ring
     border-radius: calc(var(--p-border-radius-base) + 3px - 1px + #{rem(1px)});
```
dleroux: I don't think we need the `- 1px + #{rem(1px)}` since this will be fixed/consistent?

danrosenthal: Agreed, this math is pretty weird. Between the mixed `rem/px` units, and subtracting and adding `1`, this could be cleaned up a lot

kyledurand: Good catch. I'll track that issue here among other focus ring related issues: https://github.com/Shopify/polaris-ux/issues/396

dleroux:
I was top hatting when it merged. there issues with the outline buttons. I think the border just needs to be added. I think also the plain buttons are probably best without an offset since their spacing is adjusted.

buttons

danrosenthal:
> I was top hatting when it merged. there issues with the outline buttons. 

Outline buttons appear as intended.

danrosenthal:
> I think also the plain buttons are probably best without an offset since their spacing is adjusted.

That probably makes sense, but can be done in a quick followup, or during polish/QA

kyledurand:
This PR was intended to focus more specifically on the animation on the elements that already have the new focus ring. We still have elements that need focus such as select and segmented buttons. Anything missing can be added to the issue I added above https://github.com/Shopify/polaris-ux/issues/396

PR #2648, opened 2020-01-22
[Radio Button] applying colour system

dleroux:

partly addresses: https://github.com/Shopify/polaris-ux/issues/357

- Added the proper branding to radio button
- Decided on not using the backdrop mixin. This was added complexity with very little to no benefit as a lot of the stuff needed to be overwritten. Once I'm done all the form elements I will most likely delete the mixin all together as I have a feeling only Checkbox will be using it. 

Screen Shot 2020-01-22 at 2 08 26 PM

### How to 🎩

<details>

```jsx
import {Page, Stack, RadioButton} from '../src';

  return (
    <Page title="Playground">
      <RadioButtonExample />
    </Page>
  );
}

function RadioButtonExample() {
  const [value, setValue] = useState('disabled');
  const handleChange = useCallback(
    (_checked, newValue) => setValue(newValue),
    [],
  );
  return (
    <Stack vertical>
      <RadioButton
        label="Accounts are disabled"
        helpText="Customers will only be able to check out as guests."
        checked={value === 'disabled'}
        id="disabled"
        name="accounts"
        onChange={handleChange}
      />
      <RadioButton
        label="Accounts are optional"
        helpText="Customers will be able to check out with a customer account or as a guest."
        id="optional"
        name="accounts"
        checked={value === 'optional'}
        onChange={handleChange}
      />

      <RadioButton
        label="Disabled"
        helpText="Customers will only be able to check out as guests."
        checked={false}
        id="disabled"
        name="accounts2"
        onChange={handleChange}
        disabled
      />
      <RadioButton
        label="Disabled checked"
        helpText="Customers will be able to check out with a customer account or as a guest."
        id="optional"
        name="accounts2"
        checked
        onChange={handleChange}
        disabled
      />
    </Stack>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -64,8 +64,8 @@ export const Checkbox = React.forwardRef<CheckboxHandles, CheckboxProps>(
     const id = useUniqueId('Checkbox', idProp);
     const {
       value: mouseOver,
-      setTrue: forceTrueMouseOver,
-      setFalse: forceFalseMouseOver,
+      setTrue: handleMouseOver,
+      setFalse: handleMouseOut,
     } = useToggle(false);
 
     useImperativeHandle(ref, () => ({
@@ -140,8 +140,8 @@ export const Checkbox = React.forwardRef<CheckboxHandles, CheckboxProps>(
         error={error}
         disabled={disabled}
         onClick={handleInput}
-        onMouseOver={forceTrueMouseOver}
-        onMouseOut={forceFalseMouseOver}
+        onMouseOver={handleMouseOver}
+        onMouseOut={handleMouseOut}
       >
         <span className={wrapperClassName}>
           <input
--- a/src/components/RadioButton/RadioButton.scss
+++ b/src/components/RadioButton/RadioButton.scss
@@ -54,9 +54,78 @@
       border-radius: 50%;
     }
   }
+
+  .Icon {
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%) scale(0);
+    transform-origin: 50% 50%;
+    height: 50%;
+    width: 50%;
+    background: var(--p-icon-on-interactive, color('indigo'));
+    border-radius: 50%;
+    transition: control-icon-transition();
+
+    @media (-ms-high-contrast: active) {
+      background: ms-high-contrast-color('text');
+    }
+  }
 }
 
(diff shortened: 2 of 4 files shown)


dleroux:
@johanstromqvist This is ready for review. The border is animated when it's being checked, however, this was removed from the Checkbox. Should we remove it here as well?

johanstromqvist:
Thanks – looking great! Two things:

1. **Sync issue** When keyboarding the radio all animation is in perfect sync, perceived as one animation as intended. When mouse clicking the illusion breaks because the radio animate at mouse down and the focus ring on mouse up. Ideally there's no focus ring when mousing (as discussed earlier), but if there must be one because of technical constraints then the animations should happen in sync, i.e. on mouse up.

radio-review_async

2. **Duration tweak.**
- Border looks perfect – `100ms` on both check and uncheck
- Inner dot
  - On check: looks like duration for `translate` is currently set to `100ms`. Should be the same as opacity, `150ms`.
  - On uncheck: reversed, both `translate` and `opacity` should be `100ms`

image
👆Hope this clarifies. Let's pair up if not.

---

There's a general timing pattern for tiny interactions animating in and out.
- `150ms` when animating in – puts some confirming emphasis on appear
- `100ms` when animating out – gets quickly out of the way on disappear
- This is yet to be documented, but something that we this far have used on radio button, checkbox and focus ring.
- The timing pattern will likely extend beyond tiny interactions, but with modification.

dleroux:
Cool I’ll update the animation.

As for the focus, we are relying on the browsers focus behavior so we don’t entirely control if that happens on mouse up or down. That said we are still investigating the best way to handle the focus ring to meet the requirements.

johanstromqvist:
There are two separate issues regarding the sync of focus and radio animation.
1. To have or not to have focus ring on mouse click. I think we should have a wholistic cross-disciplinary discussion about that.
2. Given that we have a focus ring, the two animations need to sync. The solution to the issue as I see it, is to give radio button feedback on mouse up as done [here](https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_input_type_radio) (no need to touch the focus ring).

dleroux:
I see. All fixed here: https://5d559397bae39100201eedc1-cshurceeod.chromatic.com/iframe.html?id=all-components-radio-button--default-radio-button&contexts=Global%20Theming=Enabled%20-%20Light%20Mode

johanstromqvist:
Looks and feels great, Dan. Thanks!

PR #2652, opened 2020-01-23
Add Travis config

BPScott:

Time to move away from CircleCI.
Onwards to Travis

Sets up travis jobs. I've split this out into a few more steps.

Travis runs two sets of checks:
- The "branch" check runs the commit in this branch (this is what you're used to happening)
- The "Pull request" check merges a PR into its base branch then runs the checks against that. This protects against those scenarios when tests on master and your PR are green but then when you merge tests fail because master had moved on since the base of your PR

In the Pull request check I've disabled running Chroma, Percy and Shrink-ray because those folks don't have have a good way of showing the state of the world for that merge commit. They still get reported as part of the "branch" check which is the same as the old behavior.

I've had to make a change to the shrink-ray script which breaks it on Circle's config but considering that's about to go away that's not a problem

### How to 🎩

Check out travis CI jobs and see that they pass


scripts/build-shrink-ray.js
```diff
@@ -29,16 +29,19 @@ if (sha) {
...
 }
 
 function setupShrinkRay() {
   console.log('[shrink-ray] Running shrink-ray prebuild script...');
 
-  // fetch latest in BuildKite pipeline
-  execSync('git fetch origin master');
+  // fetch latest in pipeline. Travis does a shallow clone by default,
+  // --unshallow makes sure we can fetch the master commit
+  execSync('git fetch --unshallow origin master:refs/remotes/origin/master', {
```
amrocha: why unshallow?

BPScott: Check the comment I ninjaed in just after this comment

Because travis does a shallow clone there's a chance that the commit at the tip of the master branch isn't included in the cloned commits. --unshallow grabs the full history to make sure we can find that commit when we do the rev-parse bits a few lines down

BPScott:
Updated to include changelog, removal of circle config and other references to Circle

shipit.yml
```diff
@@ -1,20 +1,18 @@
 ci:
+  require:
```
BPScott: Chucking this up to the top of the file as it seems like the most important key

PR #2655, opened 2020-01-23
[RangeSlider] Apply color system to thumb

LauraAubin:

Part of https://github.com/Shopify/polaris-ux/issues/358

This applies the color system to the `RangeSlider's` thumb and output bubble (everything else except the track).

Some noticeable differences:
- The thumb size changes depending on if its active or not
- The output bubble only displays when the thumb is active

### Single thumb:

| Light        | Dark           |
|  image  |  image |

### Dual thumb:

| Light        | Dark           |
|  image |  image |

### How to 🎩

### 🎩 checklist

### DIFF
--- a/src/components/RangeSlider/components/DualThumb/DualThumb.scss
+++ b/src/components/RangeSlider/components/DualThumb/DualThumb.scss
@@ -2,18 +2,27 @@
 @import '../../RangeSlider';
 
 $stacking-order: (
-  output: 10,
-  input: 20,
+  input: 10,
+  output: 20,
 );
 
 $range-wrapper: rem(28px);
 
 $range-thumb-border-error: rem(2px) solid color('red');
 
-$range-thumb-shadow: (0 0 0 rem(1px) rgba(black, 0.2), shadow(faint));
-$range-thumb-shadow-hover: (0 0 0 rem(1px) rgba(black, 0.4), shadow(faint));
-$range-thumb-shadow-error: 0 0 0 rem(1px) color('red');
-$range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
+$range-thumb-shadow: var(
+  --p-override-none,
+  (0 0 0 rem(1px) rgba(black, 0.2), shadow(faint))
+);
+$range-thumb-shadow-hover: var(
+  --p-override-none,
+  (0 0 0 rem(1px) rgba(black, 0.4), shadow(faint))
+);
+$range-thumb-shadow-error: var(--p-override-none, 0 0 0 rem(1px) color('red'));
+$range-thumb-shadow-focus: var(
+  --p-override-none,
+  0 0 0 rem(1px) color('indigo')
+);
 
 .Wrapper {
   position: relative;
@@ -83,41 +92,54 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
 .Thumbs {
   position: absolute;
   z-index: z-index('input', $stacking-order);
+  padding: 0;
   width: $range-thumb-size;
   height: $range-thumb-size;
   border-radius: 50%;
-  border: border-width() solid color('sky', 'lighter');
+  border: border-width() solid var(--p-interactive, color('sky', 'lighter'));
   box-shadow: $range-thumb-shadow;
-  background: linear-gradient(color('sky', 'lighter'), color('sky', 'light'));
+  background: linear-gradient(
+    var(--p-interactive, color('sky', 'lighter')),
+    var(--p-interactive, color('sky', 'light'))
+  );
   -webkit-tap-highlight-color: transparent;
 
   // stylelint-disable-next-line value-no-vendor-prefix
   cursor: -webkit-grab;
+  transition: transform var(--p-duration-1-5-0, duration())
+    var(--p-ease, easing());
 
   &.disabled {
     cursor: not-allowed;
-    border-color: color('sky', 'dark');
+    border-color: var(--p-border-secondary-disabled, color('sky', 'dark'));
+    background: var(--p-border-secondary-disabled, color('white'));
   }
 
   &:hover {
-    background: linear-gradient(color('sky', 'lighter'), color('sky', 'light'));
     box-shadow: $range-thumb-shadow-hover;
   }
 
-  .error & {
-    border-color: color('red');
-    box-shadow: $range-thumb-shadow-error;
+  &:active {
+    transform: scale(var(--p-range-slider-thumb-scale, 1));
   }
 
   &:focus {
     outline: 0;
-    background: linear-gradient(color('sky', 'lighter'), color('sky', 'light'));
-    border-color: color('indigo');
-    box-shadow: $range-thumb-shadow-focus;
+    border-color: var(--p-override-transparent, color('indigo'));
+    box-shadow: var(--p-override-none, $range-thumb-shadow-focus);
 
     @media (-ms-high-contrast: active) {
       outline: 1px solid ms-high-contrast-color('text');
     }
+  }
+
+  .error & {
+    border-color: var(--p-action-critical, color('red'));
+    background: linear-gradient(
+      var(--p-action-critical, color('sky', 'lighter')),
+      var(--p-action-critical, color('sky', 'light'))
+    );
+    box-shadow: $range-thumb-shadow-error;
   }
 }
 
@@ -137,20 +159,31 @@ $range-output-spacing: rem(16px);
 .Output {
   position: absolute;
   z-index: z-index('output', $stacking-order);
-  bottom: $range-thumb-size;
+  bottom: var(--p-range-slider-thumb-size-active, $range-thumb-size);
   opacity: 0;
   visibility: hidden;
   pointer-events: none;
-  transition-property: opacity, visibility;
-  transition-duration: duration();
-  transition-timing-function: easing();
+  transition-property: opacity, visibility, bottom;
+  transition-duration: var(--p-duration-1-5-0, duration());
+  transition-timing-function: var(--p-ease, easing());
   transform: translateX(calc(-50% + #{$range-thumb-size / 2}));
 
   .Thumbs:hover + &,
-  .Thumbs:active + &,
   .Thumbs:focus + & {
+    opacity: var(--p-override-zero, 1);
+    visibility: var(--p-override-none, visible);
+  }
+
+  .Thumbs:active + & {
+    $range-thumb-size-difference: var(
+        --p-range-slider-thumb-size-active,
+        $range-thumb-size
+      ) - var(--p-range-slider-thumb-size-base, $range-thumb-size);
     opacity: 1;
     visibility: visible;
+    bottom: calc(
+      #{var(--p-range-slider-thumb-size-active, $range-thumb-size)} + #{$range-thumb-size-difference}
+    );
   }
 }
 
@@ -160,11 +193,12 @@ $range-output-spacing: rem(16px);
   padding: 0 spacing(tight);
   min-width: $range-output-size;
   height: $range-output-size;
-  background-color: color('ink');
+  background-color: var(--p-surface, color('ink'));
+  box-shadow: var(--p-popover-shadow, --p-override-none);
   border-radius: border-radius();
   transition-property: transform;
-  transition-duration: duration();
-  transition-timing-function: easing();
+  transition-duration: var(--p-duration-1-5-0, duration());
+  transition-timing-function: var(--p-ease, easing());
 
   // stylelint-disable selector-max-specificity
   .Thumbs:hover + .Output &,
@@ -185,5 +219,5 @@ $range-output-spacing: rem(16px);
   flex: 1 1 auto;
   margin: auto;
   text-align: center;
-  color: color('white');
+  color: var(--p-text, color('white'));
 }
--- a/src/components/RangeSlider/components/DualThumb/DualThumb.tsx
+++ b/src/components/RangeSlider/components/DualThumb/DualThumb.tsx
@@ -6,6 +6,7 @@ import {
 } from '@shopify/javascript-utilities/events';
 import isEqual from 'lodash/isEqual';
 import {classNames} from '../../../../utilities/css';
+import {FeaturesContext} from '../../../../utilities/features';
 import {CSS_VAR_PREFIX} from '../../utilities';
 import {RangeSliderProps, DualValue} from '../../types';
 import {Labelled, labelID} from '../../../Labelled';
@@ -38,9 +39,9 @@ enum Control {
   Upper,
 }
 
-const THUMB_SIZE = 24;
-
 export class DualThumb extends React.Component<DualThumbProps, State> {
+  static contextType = FeaturesContext;
+
   static getDerivedStateFromProps(props: DualThumbProps, state: State) {
     const {min, step, max, value, onChange, id} = props;
     const {prevValue} = state;
@@ -61,6 +62,8 @@ export class DualThumb extends React.Component<DualThumbProps, State> {
     };
   }
 
+  context!: React.ContextType<typeof FeaturesContext>;
+
   state: State = {
     value: sanitizeValue(
       this.props.value,
@@ -80,9 +83,14 @@ export class DualThumb extends React.Component<DualThumbProps, State> {
   private setTrackPosition = debounce(
     () => {
       if (this.track.current) {
+        const newDesignLanguage =
+          this.context && this.context.newDesignLanguage;
+        const thumbSize = newDesignLanguage ? 16 : 24;
+
         const {width, left} = this.track.current.getBoundingClientRect();
-        const adjustedTrackWidth = width - THUMB_SIZE;
-        const adjustedTrackLeft = left + THUMB_SIZE / 2;
+        const adjustedTrackWidth = width - thumbSize;
+        const adjustedTrackLeft = left + thumbSize / 2;
+
         this.setState({
           trackWidth: adjustedTrackWidth,
           trackLeft: adjustedTrackLeft,
--- a/src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
+++ b/src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
@@ -2,14 +2,23 @@
 @import '../../RangeSlider';
 
 $stacking-order: (
-  output: 10,
-  input: 20,
+  input: 10,
+  output: 20,
 );
 
-$range-thumb-shadow: (0 0 0 rem(1px) rgba(black, 0.2), shadow(faint));
-$range-thumb-shadow-hover: (0 0 0 rem(1px) rgba(black, 0.4), shadow(faint));
-$range-thumb-shadow-error: 0 0 0 rem(1px) color('red');
-$range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
+$range-thumb-shadow: var(
+  --p-override-none,
+  (0 0 0 rem(1px) rgba(black, 0.2), shadow(faint))
+);
+$range-thumb-shadow-hover: var(
+  --p-override-none,
+  (0 0 0 rem(1px) rgba(black, 0.4), shadow(faint))
+);
+$range-thumb-shadow-error: var(--p-override-none, 0 0 0 rem(1px) color('red'));
+$range-thumb-shadow-focus: var(
+  --p-override-none,
+  0 0 0 rem(1px) color('indigo')
+);
 
 .SingleThumb {
   display: flex;
@@ -70,6 +79,10 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
     var(--progress-upper, transparent) var(--Polaris-RangeSlider-progress, 100%),
     var(--progress-upper, transparent) 100%;
 
+  @mixin rangeSliderThumbPosition($size) {
+    margin-top: calc(-1 * (#{$size} - #{$range-track-height}) / 2);
+  }
+
   &::-ms-tooltip {
     display: none;
   }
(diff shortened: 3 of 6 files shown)


src/components/RangeSlider/components/DualThumb/DualThumb.scss
```diff
@@ -86,39 +95,54 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
...
   &:hover {
-    background: linear-gradient(color('sky', 'lighter'), color('sky', 'light'));
     box-shadow: $range-thumb-shadow-hover;
   }
 
-  .error & {
-    border-color: color('red');
-    box-shadow: $range-thumb-shadow-error;
+  &:active {
+    margin-left: var(--p-range-slider-thumb-offset);
```
LauraAubin: This is needed so that the thumb remains "centered" as it grows in size.

Since it is positioned absolutely, and its width/height increase when active, it grows from the top left corner, causing it to be misaligned with the output bubble.

I tried modifying the `left` value in the JSX, but this would require the `useFeatures` hook, which would require functionalizing this component.

src/components/RangeSlider/components/DualThumb/DualThumb.scss
```diff
@@ -86,39 +95,54 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
   width: $range-thumb-size;
   height: $range-thumb-size;
```
LauraAubin: Thumbs on Safari don't appear to be perfect circles for some reason

image

src/components/RangeSlider/components/DualThumb/DualThumb.scss
danrosenthal: You cant use `Class.contextType` to access `this.context` in a class component.

- https://reactjs.org/docs/context.html#classcontexttype

src/components/RangeSlider/components/DualThumb/DualThumb.scss
tmlayton: That’s weird

tmlayton: In the "Layout" tab in the sidebar in FF there is a box model for each element. Taking a look at it shows that the thumb has a border and padding. For this, the element should be `16x16` with not padding or border.

Screen Shot 2020-01-27 at 3 18 11 PM

tmlayton:
The positioning of the dual thumbs is a bit off as the right thumb leaves a bit of track exposed and the left thumb extends beyond the track when active (bigger).

dual

src/components/RangeSlider/components/DualThumb/DualThumb.scss
```diff
@@ -86,39 +95,54 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
...
-  background: linear-gradient(color('sky', 'lighter'), color('sky', 'light'));
+  background: linear-gradient(
+    var(--p-action-interactive, color('sky', 'lighter')),
+    var(--p-action-interactive, color('sky', 'light'))
+  );
   -webkit-tap-highlight-color: transparent;
 
   // stylelint-disable-next-line value-no-vendor-prefix
   cursor: -webkit-grab;
+  transition-property: width, height, margin;
```
tmlayton: Can we transition scale instead? The pixel jogging from the width/height animation was pretty noticeable to be straight away. Based on its position on the track we can change the transform origin so it appears to scale to the right, center, or left.

tmlayton: Did a quick inspector mock up and scale appears to work well for dual thumbs

scale

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -93,19 +106,25 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
...
     border-radius: 50%;
-    background: linear-gradient(color('white'), color('sky', 'lighter'));
+    background: linear-gradient(
+      var(--p-action-interactive, color('white')),
+      var(--p-action-interactive, color('sky', 'lighter'))
+    );
     box-shadow: $range-thumb-shadow;
     appearance: none;
-    transition-property: border-color, box-shadow;
+    transition-property: border-color, box-shadow, width, height, margin-top;
```
tmlayton: Same thing here, we should be able to use scale.

src/components/RangeSlider/components/DualThumb/DualThumb.scss
LauraAubin: Which approach do you think is cleaner?

I went with the css because the value of the thumb offset is actually the negative difference of the active thumb size value and base value (`-4px`). I felt that modifying the JSX might cause future "where is that value coming from?"

LauraAubin: If we decide to go with the css approach, it'll probably be clearer to do that math instead of hardcoding a new variable `rangeSliderThumbOffset`.

danrosenthal: Aren't we already doing the work in JavaScript for the existing product design language? No sense in doing it in two places. We also try to avoid applying `margin` within our components, unless they are specifically for layout.

LauraAubin: Yup we are, but it doesn't currently account for the thumb changing in size when active.

src/components/RangeSlider/components/DualThumb/DualThumb.scss
LauraAubin: Great suggestion! Scale definitely looks a lot smoother

src/components/RangeSlider/components/DualThumb/DualThumb.scss
LauraAubin: Scratch all this! Looks like we don't need to reposition the thumb to the output bubble if we scale the thumb as per this suggestion https://github.com/Shopify/polaris-react/pull/2655#discussion_r371540536

src/components/RangeSlider/components/DualThumb/DualThumb.scss
LauraAubin: Addressed with https://github.com/Shopify/polaris-react/pull/2655/commits/9c3c7430706d31cbbbe99c7d159d2054c0f07303

src/components/RangeSlider/components/DualThumb/DualThumb.scss
```diff
@@ -2,18 +2,27 @@
 @import '../../RangeSlider';
 
 $stacking-order: (
-  output: 10,
+  output: 20,
```
LauraAubin: I was hesitant about editing these values, especially making `output` and `input` the same. I did this to fix stacking issues such as,

image

This screenshot is comparing the output bubble to a `primary Connected` element. Doing a quick search for `$stacking-order` shows that most values are exactly, or less than 20.

src/components/RangeSlider/components/DualThumb/DualThumb.scss
LauraAubin: Resolved with https://github.com/Shopify/polaris-react/pull/2655/commits/a298747542ebb27e5701658306920208d0b6703d

src/components/RangeSlider/components/DualThumb/DualThumb.scss
```diff
@@ -2,18 +2,27 @@
...
   input: 20,
```
danrosenthal: This could be achieved by inverting these values instead. If you want the output to stack over the input, it only needs a larger value. Make this:

```
input: 10,
output: 20,
```

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -2,14 +2,23 @@
 @import '../../RangeSlider';
 
 $stacking-order: (
-  output: 10,
+  output: 20,
   input: 20,
```
danrosenthal: Same recommendation here. Invert these values instead of making them equal

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -93,19 +106,25 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
...
     border-radius: 50%;
-    background: linear-gradient(color('white'), color('sky', 'lighter'));
+    background: linear-gradient(
+      var(--p-action-interactive, color('white')),
+      var(--p-action-interactive, color('sky', 'lighter'))
+    );
     box-shadow: $range-thumb-shadow;
     appearance: none;
-    transition-property: border-color, box-shadow;
+    transition-property: border-color, box-shadow, margin-top, transform;
```
danrosenthal: Why do we need to transition the `margin-top`?

danrosenthal:
This is getting close! A couple things I'd like to see here:
- can we add tests for the new design language changes in JavaScript?
- can you add the new focus ring to the thumb as part of these changes?

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -106,6 +116,19 @@ export class DualThumb extends React.Component<DualThumbProps, State> {
...
+  componentDidUpdate() {
+    const {globalTheming} = this.state;
+    const unstableGlobalTheming = this.context
+      ? this.context.unstableGlobalTheming
+      : false;
+
+    if (unstableGlobalTheming !== globalTheming) {
+      // eslint-disable-next-line react/no-did-update-set-state
+      this.setState({globalTheming: unstableGlobalTheming});
+      this.setTrackPosition();
```
kyledurand: Can you explain this change? I don't like seeing `setState` in `componentDidUpdate` and it also seems odd to have a state for `globalTheming`

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
LauraAubin: We don't! Accidentally left this behind when we swapped from increasing the width/height of the thumb, to adjusting the scale.

src/components/RangeSlider/components/DualThumb/DualThumb.scss
```diff
@@ -83,42 +92,56 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
 .Thumbs {
   position: absolute;
   z-index: z-index('input', $stacking-order);
+  padding: 0;
```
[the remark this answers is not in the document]
LauraAubin: This does not break the existing implementation, but without it the thumbs are not perfect circles in Firefox. More details [here](https://github.com/Shopify/polaris-react/pull/2655#discussion_r370872578).

src/components/RangeSlider/components/DualThumb/DualThumb.scss
```diff
@@ -83,42 +92,56 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
...
+    var(--p-action-interactive, color('sky', 'lighter')),
+    var(--p-action-interactive, color('sky', 'light'))
+  );
   -webkit-tap-highlight-color: transparent;
 
   // stylelint-disable-next-line value-no-vendor-prefix
   cursor: -webkit-grab;
+  transition-property: transform, margin;
+  transition-duration: duration();
+  transition-timing-function: easing();
```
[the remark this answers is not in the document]
LauraAubin: Do you know if they've been ported to the tokens library yet?

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
LauraAubin: In most places where `unstableGlobalTheming` is being used, its within a functional component and we have access to the `useFeatures` hook. But `DualThumb` is a class based component, and even though you can access `this.context.unstableGlobalTheming` from the render section, this doesn't help change the size of the track because `setTrackPosition()` is what modifies the state values `trackWidth` and `trackLeft`.

So in order to call that function only whenever the theme changes, this is why I call it within a  `componentDidUpdate` only when this condition is true:

```
unstableGlobalTheming !== globalTheming
```

LauraAubin:
Made a note about the focus ring [here](https://github.com/Shopify/polaris-ux/issues/358).

src/components/RangeSlider/components/DualThumb/DualThumb.scss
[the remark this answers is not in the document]
kyledurand: No ported but they already exist in `src/utilities/theme/tokens.ts` so they're usable as 

`var(--p-duration-1-0-0)`
or
`var(--p-duration-1-5-0)`
and
`var(--p-ease)`

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
kyledurand: I've pulled down this branch, and removed the componentDidUpdate life cycle method but can't seem to find any oddities while switching between current and new design language. Can you post a gif of the issue this is fixing?

dleroux: Not sure I follow either, if `this.context.newDesignLanguage` is true, it always returns true inside `setTrackPosition`. Not sure we need the state?

src/components/RangeSlider/components/DualThumb/DualThumb.scss
```diff
@@ -83,42 +92,56 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
...
-  background: linear-gradient(color('sky', 'lighter'), color('sky', 'light'));
+  background: linear-gradient(
+    var(--p-action-interactive, color('sky', 'lighter')),
+    var(--p-action-interactive, color('sky', 'light'))
+  );
   -webkit-tap-highlight-color: transparent;
 
   // stylelint-disable-next-line value-no-vendor-prefix
   cursor: -webkit-grab;
+  transition-property: transform, margin;
```
dleroux: Is margin transitioning? If not you could just do something like this:
`transition: transform var(--p-duration-1-5-0, 0) var(--p-ease, linear)`

We currently don't have a 200ms duration, depending on how this looks maybe we can add it.

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -106,6 +116,19 @@ export class DualThumb extends React.Component<DualThumbProps, State> {
     }
   }
 
+  componentDidUpdate() {
+    const {globalTheming} = this.state;
```
dleroux: If we do need this we might just need:
```
componentDidUpdate() {
    this.context && this.context.newDesignLanguage && this.setTrackPosition();
  }
```

dleroux:
There's a bit of regression introduced. The focus shadow size has changed:

**This PR**
Screen Shot 2020-02-06 at 9 33 05 AM

**Master**
Screen Shot 2020-02-06 at 9 32 50 AM

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
LauraAubin: hmm strange, I'm not having issues anymore when I get rid of `componentDidUpdate`. Thanks for challenging me on this!

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
LauraAubin: Looks like we can remove this.

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
kyledurand: Awesome!

LauraAubin:
We decided that we want the output bubble to follow the same rationale as a tooltip, in which it is light on light and dark on dark.

|        | |
|  image   |   image   |

The box shadow looks a little fuzzy but this is also the case for tooltip and popover. We can tweak `--p-popover-shadow` closer to the release date.

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -182,25 +206,35 @@ $range-output-spacing: rem(16px);
...
 
 .OutputBubble {
   position: relative;
   display: flex;
+  box-shadow: var(--p-popover-shadow, --p-override-none);
   padding: 0 spacing(tight);
   min-width: $range-output-size;
   height: $range-output-size;
-  background-color: color('ink');
+  background-color: var(--p-surface, color('ink'));
```
LauraAubin: Both tooltip and popover use `--p-surface` as the background.

src/components/RangeSlider/components/DualThumb/DualThumb.scss
```diff
@@ -83,42 +92,56 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
...
   width: $range-thumb-size;
   height: $range-thumb-size;
   border-radius: 50%;
-  border: border-width() solid color('sky', 'lighter');
+  border: border-width() solid
+    var(--p-action-interactive, color('sky', 'lighter'));
   box-shadow: $range-thumb-shadow;
-  background: linear-gradient(color('sky', 'lighter'), color('sky', 'light'));
+  background: linear-gradient(
+    var(--p-action-interactive, color('sky', 'lighter')),
```
dleroux:
```suggestion
    var(--p-interactive, color('sky', 'lighter')),
```

src/components/RangeSlider/components/DualThumb/DualThumb.scss
```diff
@@ -83,42 +92,56 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
 .Thumbs {
   position: absolute;
   z-index: z-index('input', $stacking-order);
+  padding: 0;
   width: $range-thumb-size;
   height: $range-thumb-size;
   border-radius: 50%;
-  border: border-width() solid color('sky', 'lighter');
+  border: border-width() solid
+    var(--p-action-interactive, color('sky', 'lighter'));
```
dleroux: This has recently changed to --p-interactive
```suggestion
    var(--p-interactive, color('sky', 'lighter'));
```

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -93,19 +106,25 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
     cursor: -webkit-grab;
     width: $range-thumb-size;
     height: $range-thumb-size;
-    border: border-width() solid color('sky', 'lighter');
+    border: border-width() solid
+      var(--p-action-interactive, color('sky', 'lighter'));
```
dleroux:
```suggestion
      var(--p-interactive, color('sky', 'lighter'));
```

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -93,19 +106,25 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
...
     border-radius: 50%;
-    background: linear-gradient(color('white'), color('sky', 'lighter'));
+    background: linear-gradient(
+      var(--p-action-interactive, color('white')),
```
dleroux:
```suggestion
      var(--p-interactive, color('white')),
```

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -93,19 +106,25 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
...
+      var(--p-action-interactive, color('sky', 'lighter'))
```
dleroux:
```suggestion
      var(--p-interactive, color('sky', 'lighter'))
```

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -93,19 +106,25 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
...
     transition-timing-function: easing();
 
+    @include rangeSliderThumbPosition($range-thumb-size);
+
     &:hover {
       background: linear-gradient(
-        color('sky', 'lighter'),
-        color('sky', 'light')
+        var(--p-action-interactive, color('sky', 'lighter')),
+        var(--p-action-interactive, color('sky', 'light'))
```
dleroux:
```suggestion
        var(--p-interactive, color('sky', 'light'))
```

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -93,19 +106,25 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
...
     transition-duration: duration();
     transition-timing-function: easing();
 
+    @include rangeSliderThumbPosition($range-thumb-size);
+
     &:hover {
       background: linear-gradient(
-        color('sky', 'lighter'),
-        color('sky', 'light')
+        var(--p-action-interactive, color('sky', 'lighter')),
```
dleroux:
```suggestion
        var(--p-interactive, color('sky', 'lighter')),
```

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
dleroux:
```suggestion
      var(--p-interactive, color('sky', 'lighter'))
```

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
dleroux:
```suggestion
      var(--p-interactive, color('white')),
```

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
dleroux:
```suggestion
        var(--p-interactive, color('sky', 'lighter')),
```

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
dleroux:
```suggestion
      var(--p-interactive, color('sky', 'lighter'))
```

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
dleroux:
```suggestion
      var(--p-interactive, color('white')),
```

src/components/RangeSlider/components/DualThumb/DualThumb.scss
```diff
@@ -83,42 +92,56 @@ $range-thumb-shadow-focus: 0 0 0 rem(1px) color('indigo');
...
   box-shadow: $range-thumb-shadow;
-  background: linear-gradient(color('sky', 'lighter'), color('sky', 'light'));
+  background: linear-gradient(
+    var(--p-action-interactive, color('sky', 'lighter')),
+    var(--p-action-interactive, color('sky', 'light'))
```
dleroux:
```suggestion
    var(--p-interactive, color('sky', 'light')),
```

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
dleroux:
```suggestion
      var(--p-interactive, color('white')),
```

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
dleroux:
```suggestion
      var(--p-interactive, color('sky', 'lighter'))
```

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
dleroux:
```suggestion
      var(--p-interactive, color('white')),
```

LauraAubin:
I'll rebase to update the color variable names, and will need to revise the output bubbles background color as the Figma design has changed since this PR was opened.

PR #2657, opened 2020-01-23
Fixed react reference error

AndrewMusgrave:

[Issue](https://github.com/Shopify/web/issues/22816)

🤷‍♂ 

### How to 🎩

Build consumer to web and run the server

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -165,7 +165,7 @@ export function Button({
   );
 
   const disclosureIconMarkup = disclosure ? (
-    <IconWrapper>
+    <span className={styles.Icon}>
       <div
         className={classNames(
           styles.DisclosureIcon,
@@ -174,7 +174,7 @@ export function Button({
       >
         {disclosureIcon}
       </div>
-    </IconWrapper>
+    </span>
   ) : null;
 
   let iconMarkup;
@@ -185,7 +185,7 @@ export function Button({
     ) : (
       icon
     );
-    iconMarkup = <IconWrapper>{iconInner}</IconWrapper>;
+    iconMarkup = <span className={styles.Icon}>{iconInner}</span>;
   }
 
   const childMarkup = children ? (
@@ -348,10 +348,6 @@ export function Button({
   ) : (
     buttonMarkup
   );
-}
-
-function IconWrapper({children}: any) {
-  return <span className={styles.Icon}>{children}</span>;
 }
 
 function isIconSource(x: any): x is IconSource {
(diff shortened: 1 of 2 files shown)


AndrewMusgrave:
cc/ @BPScott in case you have any insight

src/components/Button/Button.tsx
```diff
@@ -263,9 +263,9 @@ export function Button({
         onClick={toggleDisclosureActive}
         onMouseUp={handleMouseUpByBlurring}
       >
-        <IconWrapper>
+        <span className={styles.Icon}>
```
AndrewMusgrave: This is the same markup applied from `IconWrapper`

chloerice: Hmmm yeah, was getting this while trying to `yarn run build-consumer web`. [Fixed by moving the `IconWrapper` component into our usual subcomponent file structure (`/components/IconWrapper`)](https://github.com/Shopify/polaris-react/compare/fix-esnext-error?expand=1)

AndrewMusgrave: It's the weirdest thing since there's nothing invalid about it :/ I was talking to Dan about this and he mentioned `IconWrapper` was added for testing purposes, so we could get by removing it. What do you think?

chloerice: I think it's because of the new way we're exporting (`export *`)

chloerice: We do use it 3 times in Button.tsx, but it's just wrapping it in a span with a class so removing and just using the markup directly makes sense.

BPScott:
Oh! I think I know what this might be - A bug in  `@babel/plugin-transform-react-constant-elements` that Gord found a while ago: https://discourse.shopify.io/t/react-is-not-defined-after-upgrading-to-sewing-kit-0-115-0/5619/1

This didn't crop up before because web was rocking an old version of SK where we disabled that transform for polaris (a newer version enabled it again because I thought all the issues were due to circular dependencies that we've since fixed).

---

Putting this in the components folder is the best solution here

AndrewMusgrave:
> (a newer version enabled it again because I thought all the issues were due to circular dependencies that we've since fixed).

Guess something other than circular dependencies we're causing the bugs, not sure what though 🤷‍♂

src/components/Button/Button.tsx
BPScott: >  was talking to Dan about this and he mentioned IconWrapper was added for testing purposes, so we could get by removing it.

I rejigged the test so we no longer needed to export IconWrapper recently https://github.com/Shopify/polaris-react/commit/bdb8432b1e70695603a0fd5daa2fd02957182c26#diff-6735439d743b77b7470242ae481be318


It's local only to this file now so up you to you folks if we want to replace its usage inline or move it into the components directory

AndrewMusgrave: Just removed it 👍

PR #2661, opened 2020-01-24
[Checkbox, Radio, Banner] Focus ring on tab and not click

dleroux:

Certain elements, when getting the focus ring on focus don't look quite right. There is a CSS selector `focus-visible` that is currently only supported by Firefox that would allow us to show the ring only on keyboard focus. It will take some time for this behavior is available in all our supported browsers.

Currently, Banners, Radio, Checkboxes would benefit from this behavior.

This PR makes use of state to apply the proper style. It would be nice to stick to browser behavior but I think this is an acceptable use of Javascript with very little overhead. **I have not written tests yet, will wait to to see if this is something we want to move forward with first.**

** Percy is complaining about the focus on the banner but I'm thinking something in their rendering is stealing the focus. **

focus

<details>

```jsx
import React, {useRef, useEffect, useState, useCallback} from 'react';
import {Page, Banner, RadioButton, Stack, Checkbox} from '../src';

  return (
    <Page title="Playground">
      <Stack vertical>
        <BannerWithFocusExample />
        <RadioButtonExample />
        <CheckboxExample />
      </Stack>
    </Page>
  );
}

function BannerWithFocusExample() {
  const banner = useRef();

  useEffect(() => banner.current.focus(), []);

  return (
    <Banner
      title="High risk of fraud detected"
      onDismiss={() => {}}
      status="critical"
      ref={banner}
    >
      <p>
        Before fulfilling this order or capturing payment, please review the
        fraud analysis and determine if this order is fraudulent
      </p>
    </Banner>
  );
}

function RadioButtonExample() {
  const [value, setValue] = useState('disabled');

  const handleChange = useCallback(
    (_checked, newValue) => setValue(newValue),
    [],
  );

  return (
    <Stack vertical>
      <RadioButton
        label="Accounts are disabled"
        helpText="Customers will only be able to check out as guests."
        checked={value === 'disabled'}
        id="disabled"
        name="accounts"
        onChange={handleChange}
      />
      <RadioButton
        label="Accounts are optional"
        helpText="Customers will be able to check out with a customer account or as a guest."
        id="optional"
        name="accounts"
        checked={value === 'optional'}
        onChange={handleChange}
      />
    </Stack>
  );
}

function CheckboxExample() {

  return (
    <Checkbox
      label="Basic checkbox"
      checked={checked}
      onChange={handleChange}
    />
  );
}
```
</details>

### DIFF
--- a/src/components/Banner/Banner.scss
+++ b/src/components/Banner/Banner.scss
@@ -38,6 +38,9 @@ $secondary-action-horizontal-padding: 1.5 * spacing(tight);
 
   &:focus {
     outline: none;
+  }
+
+  &.keyFocused {
     box-shadow: var(
       --p-banner-border,
       (
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -28,6 +28,10 @@ import styles from './Banner.scss';
 
 export type BannerStatus = 'success' | 'info' | 'warning' | 'critical';
 
+interface State {
+  showFocus: boolean;
+}
+
 export interface BannerProps {
   /** Title content for the banner. */
   title?: string;
@@ -47,14 +51,39 @@ export interface BannerProps {
   stopAnnouncements?: boolean;
 }
 
-export class Banner extends React.PureComponent<BannerProps, never> {
+export class Banner extends React.PureComponent<BannerProps, State> {
+  state: State = {
+    showFocus: false,
+  };
+
   private wrapper = React.createRef<HTMLDivElement>();
 
   public focus() {
     this.wrapper.current && this.wrapper.current.focus();
+    this.setState({showFocus: true});
   }
 
   render() {
+    const {showFocus} = this.state;
+
+    const handleKeyUp = (evt: React.KeyboardEvent<HTMLDivElement>) => {
+      if (evt.target === this.wrapper.current) {
+        this.setState({showFocus: true});
+      }
+    };
+
+    const handleBlur = () => {
+      this.setState({showFocus: false});
+    };
+
+    const handleMouseUp = ({
+      currentTarget,
+    }: React.MouseEvent<HTMLDivElement>) => {
+      const {showFocus} = this.state;
+      currentTarget.blur();
+      showFocus && this.setState({showFocus: false});
+    };
+
     return (
       <BannerContext.Provider value>
         <WithinContentContext.Consumer>
@@ -100,6 +129,7 @@ export class Banner extends React.PureComponent<BannerProps, never> {
               styles.Banner,
               status && styles[variationName('status', status)],
               onDismiss && styles.hasDismiss,
+              showFocus && styles.keyFocused,
               withinContentContainer
                 ? styles.withinContentContainer
                 : styles.withinPage,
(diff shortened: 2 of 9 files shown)


src/components/Banner/Banner.scss
```diff
@@ -38,6 +38,9 @@ $secondary-action-horizontal-padding: 1.5 * spacing(tight);
 
   &:focus {
     outline: none;
+  }
+
+  &.focus {
```
dleroux: Should change to keuyFocused or something make clear it isn't focused per se.

src/components/Banner/Banner.tsx
```diff
@@ -47,14 +51,41 @@ export interface BannerProps {
...
   }
 
   render() {
+    const {showFocus} = this.state;
+
+    const handleKeyUp = (evt: React.KeyboardEvent<HTMLDivElement>) => {
+      if (evt.target === this.wrapper.current) {
+        this.setState({showFocus: true});
+      } else if (showFocus === true) {
+        this.setState({showFocus: false});
```
kyledurand: While tophatting I couldn't actually end up in this condition, which I think makes sense since, handleBlur looks like it's working. Can we 🔥the else if here?

src/components/Banner/Banner.tsx
```diff
@@ -170,6 +202,8 @@ export class Banner extends React.PureComponent<BannerProps, never> {
                 role={ariaRoleType}
                 aria-live={stopAnnouncements ? 'off' : 'polite'}
                 onMouseUp={handleMouseUp}
+                onKeyUp={handleKeyUp}
+                onBlur={handleBlur}
```
kyledurand: I'm curious what others think of this. The current iteration of banner just uses a transition delay to somewhat prevent the focus ring from showing on click. The result isn't too bad, only really noticeable when you highlight text or on a click and hold.

https://screenshot.click/screencast_2020-01-29_13-28-10.mp4

I wonder if the re-renders added here are necessary

src/components/RadioButton/RadioButton.tsx
```diff
@@ -50,12 +50,24 @@ export function RadioButton({
   const id = useUniqueId('RadioButton', idProp);
   const name = nameProp || id;
   const {unstableGlobalTheming = false} = useFeatures();
+  const inputNode = useRef<HTMLInputElement | null>(null);
```
kyledurand:
```suggestion
  const inputNode = useRef<HTMLInputElement>(null);
```

I think should just work

src/components/Banner/Banner.tsx
dleroux: I think this re-render is pretty cheap. A transition delay could lead to confusion imo.

src/components/Banner/Banner.tsx
dleroux: good catch. I thought it was needed when programmatically focused but blur handles that.

src/components/Banner/Banner.tsx
kyledurand: It is pretty cheap, and if we're doing it on other elements then it probably makes sense to do it on banner as well. I doubt too many people are clicking banners anyway

PR #2663, opened 2020-01-24
Re-enable consumer build web unit tests

amrocha:

We had to disable the web unit tests in the consumer build tests for a while now because their version of node was outdated. Now that web updated their version of node we can enable them again.

Re-enable unit tests step
Update docker image to match web's image
Re-add balancer
Increase timeout to 20
Increase parallelism to match web's (24)


amrocha:
Check it out here: https://buildkite.com/shopify/polaris-react/builds/1483

This is run in a separate branch that runs with 4.11 instead of latest. This is because the test is correctly failing right now, Some of our uncommitted changes break web's unit tests (we should fix that).

The styleguide is also failing, but that's not polaris's fault, that's on the styleguide. We're aware of it and will fix itt.

amrocha:
Using web's balancer worked, gonna merge

PR #2674, opened 2020-01-28
[Color system] Add Storybook knobs for customizing theme

tmlayton:

Adds Storybook knobs for playing around with theme colors

Screen Shot 2020-02-18 at 4 18 45 PM

Uses Storybooks Knobs addon together with the Contexts addon to allow customizing the theme. Instead of the Contexts addon using the `AppProvider`, I created a wrapping component called `AppProviderWithKnobs` which allows the Knobs UI to properly render. Each color is generated from the default colors coming from Polaris Tokens.

### How to 🎩

1. Turn on the new design language, either light or dark
1. open Storybook and press "A" to see the addons panel
1. select the "Knobs" tab
1. change the color values and confirm the examples update with the new colors

Note this works for all examples _except_ the `AppProvider` examples which are already wrapped with an `AppProvider`.

### 🎩 checklist

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -20,6 +20,7 @@ module.exports = {
     '@storybook/addon-notes',
     '@storybook/addon-a11y',
     '@storybook/addon-contexts',
+    '@storybook/addon-knobs',
   ],
   webpackFinal: (config) => {
     const isProduction = config.mode === 'production';
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -2,6 +2,8 @@ import React from 'react';
 import {addParameters, addDecorator} from '@storybook/react';
 import {setConsoleOptions} from '@storybook/addon-console';
 import {withContexts} from '@storybook/addon-contexts/react';
+import {color, withKnobs} from '@storybook/addon-knobs';
+import DefaultThemeColors from '@shopify/polaris-tokens/dist/base.json';
 
 import {AppProvider} from '../src';
 import enTranslations from '../locales/en.json';
@@ -25,6 +27,29 @@ function StrictModeToggle({isStrict = false, children}) {
   return <Wrapper>{children}</Wrapper>;
 }
 
+function AppProviderWithKnobs({newDesignLanguage, colorScheme, children}) {
+  const colors = Object.entries(DefaultThemeColors).reduce(
+    (accumulator, [key, value]) => ({
+      ...accumulator,
+      [key]: color(key, value, 'Theme'),
+    }),
+    {},
+  );
+
+  return (
+    <AppProvider
+      i18n={enTranslations}
+      features={{newDesignLanguage}}
+      theme={{
+        colors,
+        colorScheme,
+      }}
+    >
+      {children}
+    </AppProvider>
+  );
+}
+
 addDecorator(
   withContexts([
     {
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -108,6 +108,7 @@
     "@storybook/addon-actions": "^5.3.2",
     "@storybook/addon-console": "^1.2.1",
     "@storybook/addon-contexts": "^5.3.2",
+    "@storybook/addon-knobs": "^5.3.9",
```
BPScott: Match this version to the one we use for the other extensions - 5.3.2 - to avoid a bunch of yarn noise

.storybook/main.js
```diff
@@ -20,6 +20,7 @@ module.exports = {
     '@storybook/addon-notes',
     '@storybook/addon-a11y',
     '@storybook/addon-contexts',
+    '@storybook/addon-knobs/register',
```
BPScott: I thiiink this can be just `'@storybook/addon-knobs`

package.json
tmlayton: Still pretty noisy

.storybook/preview.js
```diff
@@ -25,6 +27,29 @@ function StrictModeToggle({isStrict = false, children}) {
   return <Wrapper>{children}</Wrapper>;
 }
 
+function AppProviderWithKnobs({newDesignLanguage, colorScheme, children}) {
+  const colors = Object.entries(DefaultThemeColors).reduce(
+    (accumulator, [key, value]) => ({
```
BPScott: I'm not a fan of creating a new object on ever loop, can we add to the existing one insead?

```js
    (accumulator, [key, value]) => {
      accumulator[key] = strToHex(color(key, value, 'Theme'));
      return accumulator;
    },
```

.storybook/preview.js
```diff
@@ -37,34 +62,28 @@ addDecorator(
...
-            theme: {colorScheme: 'dark'},
-          },
+          props: {newDesignLanguage: true, colorScheme: 'dark'},
         },
       ],
     },
   ]),
 );
 
+addDecorator(withKnobs);
```
BPScott: Referencing ` '@storybook/addon-knobs'` instead of ` '@storybook/addon-knobs/register'` in main.js means that we don't need this line anymore - as it is already implictly added by activating the addon by its root instead of the `/register` path

(don't forget to remove the import of withKnobs)

.storybook/preview.js
tmlayton: Maybe, depends on how good your reason is 😄

.storybook/preview.js
tmlayton: Looks like this is needed

.storybook/preview.js
```diff
@@ -2,6 +2,8 @@ import React from 'react';
 import {addParameters, addDecorator} from '@storybook/react';
 import {setConsoleOptions} from '@storybook/addon-console';
 import {withContexts} from '@storybook/addon-contexts/react';
+import {color, withKnobs} from '@storybook/addon-knobs';
+import DefaultThemeColors from '@shopify/polaris-tokens/dist/base.json';
```
BPScott: Gonna need a swift rebase after https://github.com/Shopify/polaris-react/pull/2763 as the path to this has changed slightly (`@shopify/polaris-tokens/dist-modern/themes/base.json` off the top of my head)

tmlayton: Thanks, just did it!

PR #2678, opened 2020-01-28
[Select] Apply design language to Select

dleroux:
**Design still pending, no need to review until [WIP] is removed**

select

partly addresses: https://github.com/Shopify/polaris-ux/issues/363

Applying design language to Select. Current Select shouldn't be affected.

I'm also removing the Design Language modification from Backdrop. Once we roll this out, we'll be able to delete that mixin.

The icons have not been updated. My understanding is that these will be updated when icons get updated.

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Select/Select.scss
+++ b/src/components/Select/Select.scss
@@ -26,7 +26,7 @@ $stacking-order: (
 
   .Icon {
     opacity: var(--override-one, 0.4);
-    @include recolor-icon(var(--p-icon-disabled, color('ink', 'lightest')));
+    @include recolor-icon(var(--p-icon-disabled, color('ink', 'lighter')));
   }
 
   &.globalTheming {
--- a/src/components/Select/Select.scss
+++ b/src/components/Select/Select.scss
@@ -17,7 +17,7 @@ $stacking-order: (
 
 .disabled {
   .Content {
-    color: color('ink', 'lightest');
+    color: var(--p-text-disabled, color('ink', 'lightest'));
   }
 
   .InlineLabel {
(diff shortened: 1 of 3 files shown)


dleroux:
@martenbjork can you ping me here once the design is approved?

dleroux:
In speaking with @danrosenthal we decided to remove the `wip` and adjust if the design changes.

documentation/Color system.md
```diff
@@ -43,19 +43,18 @@ The surface role is used for the backgrounds of the UI. With a light color schem
```
dleroux: Rebasing, not sure why this has changed.

dleroux: We'll hold on shipping this. @danrosenthal will fix this in another PR.

dleroux:
I had removed the Backdrop because it was only useful for the checkbox in the new design language. However, because OptionList has a copy of the checkbox I decided to bring it back until we get the opportunity to remove the OptionList from inside the Autocomplete (the reason why additional props were needed on the fork checkbox).

src/components/Select/Select.scss
```diff
@@ -17,32 +17,30 @@ $stacking-order: (
...
-  // stylelint-disable-next-line selector-max-specificity
-  .Input:focus ~ .Backdrop {
-    @include control-backdrop(focused-error);
+    opacity: var(--override-one, 0.4);
+    @include recolor-icon(var(--p-icon-disabled, color('ink', 'lighter')));
+  }
+
+  &.globalTheming {
+    .Backdrop {
+      border-color: var(--p-border-secondary-disabled);
```
tmlayton: Both secondary and onSurface variants are used for borders around inputs. This should be consistent, one or the other. I lean towards onSurface.

tmlayton: /cc @danrosenthal

danrosenthal: All interactive elements should use the secondary borders, that creates a separation between the decorative (on surface) borders, and functional borders so we can ensure color contrast on interactive elements. We should fix any uses of on surface borders in interactive elements

PR #2681, opened 2020-01-28
[TrapFocus] Fixed focus theft

### LINKED ISSUE #2641: [TrapFocus] causes infinite loop when used in a Dialog
## Issue summary
The `TrapFocus` component appears to be creating an infinite loop. If you look at the stacktrace below, you can see that `Dialog` triggers `focusFirstFocusableNode` which triggers a `focusin` event. This event triggers `handleFocusIn` which calls `focusFirstFocusableNode`. The loop keeps going.

This ends up locking the main thread for a significant time.

Screen Shot 2020-01-21 at 1 55 52 PM

## Expected behavior
After the initial focus, the loop should stop.

This doesn't appear to be happening with all modals. I haven't been able to identify what scenario causes this bug.

cc @devonpmack @AndrewMusgrave

AndrewMusgrave:

Fixes #2641

* Creates focus manager to handle multiple trap focuses
* Create useFocusManager to handle listing and delisting trap focuses and letting the component know when its safe to focus
* Adds new application state (focus manager)
* Adds focus manager to app provider
* Use useFocusManager in TrapFocus
* Tests

### Screenie

Screen Shot 2020-01-28 at 5 58 46 PM

### How to 🎩

Use the performance tab and make sure focus first focusable node is not running rampant. Sample code with two trap focuses below

<details>

```jsx
import React from 'react';
import {Card, TextContainer, TextField, TrapFocus} from '../src';

  return (
    <div>
      <Card sectioned>
        <TrapFocus>
          <TextContainer>
            <TextField label="test" autoFocus />
          </TextContainer>
        </TrapFocus>
        
        <TrapFocus>
          <TextContainer>
            <TextField label="test" autoFocus />
          </TextContainer>
        </TrapFocus>
      </Card>
    </div>
  );
}

```

</details>

### 🎩 checklist


src/components/PolarisTestProvider/PolarisTestProvider.tsx
```diff
@@ -108,9 +122,13 @@ export function PolarisTestProvider({
                   <LinkContext.Provider value={link}>
                     <ThemeContext.Provider value={mergedTheme}>
                       <MediaQueryContext.Provider value={mergedMediaQuery}>
-                        <FrameContext.Provider value={mergedFrame}>
-                          {children}
-                        </FrameContext.Provider>
+                        <FocusManagerContext.Provider
+                          value={mergedFocusManager}
+                        >
```
AndrewMusgrave: Should we just use `FocusManager` and not allow any overrides?

AndrewMusgrave:
Let me know what you think @dleroux I changed it up slightly from what we looked at this morning. Listing and delisting logic now lives inside the hook.

src/components/PolarisTestProvider/PolarisTestProvider.tsx
dleroux: I think for now we probably should. Not sure we'll ever need to override it. We can do it when we do.

AndrewMusgrave: It's a different pattern than we've previously been using, however `FocusManager` has some overhead we would need to replicate in mocks and it's easier to use the component. Since this is a public API let's go with using FocusManager since adding to an API is easier than removing.

AndrewMusgrave:
Codecov messed up and included files that weren't touched.  Coverage is over 90 so I'm going to 🚢

PR #2682, opened 2020-01-28
[Color system] Import Color Factory from Polaris tokens

tmlayton:
Addresses https://github.com/Shopify/polaris-react/issues/2041


documentation/Color system.md
```diff
@@ -15,25 +15,25 @@
...
-| <pre>--p-hint-from-direct-light</pre>    | For use in building shadows scrollables.                                                                                              | ![](https://www.gifpng.com/64x64/000000/FFFFFF?border-width=16&border-type=rectangle&border-color=fafafa&text=%20)<br />#000000 | ![](https://www.gifpng.com/64x64/bababa/FFFFFF?border-width=16&border-type=rectangle&border-color=0c0d0e&text=%20)<br />#bababa |
+|CSS variable|Description|Light mode|Dark mode|
+|---|---|---|---|
+|<pre>--p-background</pre>|For use as a background color, in components such as Page and Frame backgrounds. |![](https://www.gifpng.com/64x64/fafafa/FFFFFF?border-width=16&border-type=rectangle&border-color=fafafa&text=%20)<br />#fafafa|![](https://www.gifpng.com/64x64/0b0c0d/FFFFFF?border-width=16&border-type=rectangle&border-color=0b0c0d&text=%20)<br />#0b0c0d|
```
tmlayton: Looks like all of the changes to output values are the result of rounding RGB values in the color factory. If you look at the rich diff you’ll see the are values off by 1 and visually indistinguishable. We could potentially remove this and see if Theo handles floating point numbers.

danrosenthal: I'm comfortable with the changes as-is. We can always iterate on the color factory for later releases if these are not an accurate representation of intended color variants.

PR #2685, opened 2020-01-29
[Color system] stabilize

danrosenthal:
- stabilizes color system APIs
- updates feature flag and associated class names to `newDesignLanguage`
- removes unused or unnecessary colors
- fixes use of border on interactive elements
- fixes incorrect custom properties
- tokenizes custom properties used as overrides for new design language
- removes unnecessary math from dropzone focus ring border radius calc


src/components/Checkbox/Checkbox.scss
```diff
@@ -147,7 +147,7 @@ $control-size: rem(16px);
   }
 }
 
-.error.globalTheming {
+.error.newDesignLanguage {
   .Backdrop {
     @include control-backdrop(error, $global-theming: true);
```
dleroux: I was going to mention this one should change too but this is removed as part of my PR

src/components/RadioButton/RadioButton.scss
```diff
@@ -86,7 +84,7 @@
     display: block;
     width: 100%;
     height: 100%;
-    border: var(--control-border-width) solid var(--p-border-secondary);
+    border: var(--p-control-border-width) solid var(--p-border-secondary);
```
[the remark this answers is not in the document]
dleroux: Funny I made the same change in my PR before seeing this one.

src/components/ActionList/ActionList.tsx
```diff
@@ -31,10 +31,10 @@ export function ActionList({
     finalSections = sections;
   }
 
-  const {unstableGlobalTheming = false} = useFeatures();
+  const {newDesignLanguage = false} = useFeatures();
```
BPScott: Now that this is a stable name, would it be worth adding it as a required key in the features object that gets defaulted to false in the AppProvider (and PolarisTestProvider), so then all these defaults can be removed and replaced with `const {newDesignLanguage} = useFeatures();` (as it will already be set to false)

src/index.ts
```diff
@@ -10,12 +10,10 @@ export {WithinContentContext as _SECRET_INTERNAL_WITHIN_CONTENT_CONTEXT} from '.
...
 export {AppBridgeContext} from './utilities/app-bridge';
 
-/* eslint-disable @typescript-eslint/camelcase */
 export {
-  buildColors as UNSTABLE_buildColors,
-  roleVariants as UNSTABLE_roleVariants,
-  toCssCustomPropertySyntax as UNSTABLE_toCssCustomPropertySyntax,
-  Tokens as UNSTABLE_Tokens,
-  DefaultThemeColors as UNSTABLE_DefaultThemeColors,
+  buildColors,
```
BPScott: I thought very soon `buildColors`, `roleVariants` and `defaultThemeColors` will all come from polaris-tokens. I figured that'd mean they'd all be removed polaris-react entirely. With that in mind is there much point unprefixing them - all we'd be doing is making it so they can't be removed before v5?

I'm also pretty tempted to say that `toCssCustomPropertySyntax` should either be exposed from polaris-tokens, or not be exposed at all also as I'm not clear why it needs to be used externally.

I'm not sure why Tokens needs to exposed publicly?

danrosenthal: Good call. I'll finish up Tim's [PR incorporating this from Polaris-tokens](https://github.com/Shopify/polaris-react/pull/2682) first, rebase this PR, and then only export what I need for the documentation build script which imports from our compiled javascript.

danrosenthal:
Going to press pause on this until I've completed https://github.com/Shopify/polaris-react/pull/2682. Shouldn't take too long.

src/index.ts
danrosenthal: `Tokens` and `toCssCustomPropertySyntax` are both exposed for generating our Color System documentation. Since ultimately that will find a new home, I think it makes sense to keep them exported as `unstable` so folks don't use them and then find that they're gone (which would be a breaking change).

BPScott: Agree on keeping those tagged as unstable.

For the sake of :fire: we could duplicate `toCssCustomPropertySyntax` in  the color system doc builder then remove it from the exports as it's looks like a pretty trivial implementation

src/components/ActionList/ActionList.tsx
BPScott: Had a quick chat with @danrosenthal, we'll ship this as-is for now, then loop back to this later as constant rebasing is annoying :)

PR #2691, opened 2020-01-30
[ContextualSaveBar] Apply design language

kyledurand:

Fixes https://github.com/Shopify/polaris-ux/issues/344

image
CSB-fadeup

## 

### How to 🎩

1. Hit the details page in storybook
1. Type something into the product title field
1. Dismiss the save bar

Make sure both fade up and in and fade out and down look alright

### 🎩 checklist

### DIFF
--- a/src/components/Frame/Frame.tsx
+++ b/src/components/Frame/Frame.tsx
@@ -179,14 +179,14 @@ class FrameInner extends React.PureComponent<CombinedProps, State> {
 
     const contextualSaveBarClassName = classNames(
       styles.ContextualSaveBar,
-      unstableGlobalTheming && styles['ContextualSaveBar-globalTheming'],
+      newDesignLanguage && styles['ContextualSaveBar-newDesignLanguage'],
     );
 
     const contextualSaveBarMarkup = (
       <CSSAnimation
         in={showContextualSaveBar}
         className={contextualSaveBarClassName}
-        type="fade"
+        type={newDesignLanguage ? 'fadeUp' : 'fade'}
       >
         <ContextualSaveBar {...this.contextualSaveBar} />
       </CSSAnimation>
--- a/src/components/Frame/Frame.tsx
+++ b/src/components/Frame/Frame.tsx
@@ -177,10 +177,15 @@ class FrameInner extends React.PureComponent<CombinedProps, State> {
         </div>
       ) : null;
 
+    const contextualSaveBarClassName = classNames(
+      styles.ContextualSaveBar,
+      unstableGlobalTheming && styles['ContextualSaveBar-globalTheming'],
+    );
+
     const contextualSaveBarMarkup = (
       <CSSAnimation
         in={showContextualSaveBar}
-        className={styles.ContextualSaveBar}
+        className={contextualSaveBarClassName}
         type="fade"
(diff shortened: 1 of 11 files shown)


src/components/Frame/Frame.tsx
```diff
@@ -177,11 +177,16 @@ class FrameInner extends React.PureComponent<CombinedProps, State> {
         </div>
       ) : null;
 
+    const contextualSaveBarClassName = classNames(
+      styles.ContextualSaveBar,
+      unstableGlobalTheming && styles['ContextualSaveBar-globalTheming'],
```
dleroux: Any reason why you're not simply using `.globalTheming`?

dleroux:
What's interesting is the first thing that stood out to me was that the animation was backward. I would have expected the save bar to slide down from the top and leave to the top. Is that just me?

dleroux:
Code looks good. Another thing that's interesting is the fact that the component by itself doesn't show the theming. Is that a problem with the way it's implemented in storybook?

kyledurand:
> Is that a problem with the way it's implemented in storybook

Each component has its own app provider. You'd have to set global theming manually in the readme i.e. https://github.com/Shopify/polaris-react/blob/master/src/components/ContextualSaveBar/README.md#default-contextual-save-bar

kyledurand:
> What's interesting is the first thing that stood out to me was that the animation was backward

The thought behind this is that the changes will be created by a form or change from below the topbar. Those changes push up from where they're being made. I think years of working in and using the web we're used to things sliding on and off screen but with this new design / motion language we're going to be changing that to in and out contextually, if that makes any sense 😄

src/components/Frame/Frame.tsx
kyledurand: Something we're going to be removing at some point so I didn't put much thought behind it 🤷‍♂

dleroux:
code cov would be easily achieved by testing this https://codecov.io/gh/Shopify/polaris-react/compare/5f85fbcb02cef7e349ebd9acab8fd9df98968b07...ad59e372cbf6369491e878c687eab4472c7686fa/diff#D7-107. Although not sure it's really necessary since this will be removed eventually.

kyledurand:
image

High contrast mode has been fixed up. Alex has a PR up for adding the border around buttons here: https://github.com/Shopify/polaris-react/pull/2712

@sarahill

src/components/Frame/components/CSSAnimation/CSSAnimation.scss
```diff
@@ -11,3 +11,20 @@
   opacity: 1;
   pointer-events: auto;
 }
+
+.startFadeUp {
+  opacity: 0;
+  will-change: opacity, transform;
+  transform: translateY(#{rem(25px)});
```
danrosenthal: ✨

src/components/Frame/components/ContextualSaveBar/ContextualSaveBar.scss
```diff
@@ -1,5 +1,5 @@
 @import '../../../../styles/common';
-
+$height: rem(64px);
```
danrosenthal: ✨

PR #2702, opened 2020-02-03
Fix r10 tabs test

### LINKED ISSUE #2694: TabMeasurer tests fail in React >=16.10
## Issue summary

TabMeasurer tests fail in React >=16.10.

I don't know what this means for actual behaviour of the Tabs.

My gut is saying that https://github.com/Shopify/polaris-react/blob/master/src/components/Tabs/components/TabMeasurer/TabMeasurer.tsx#L67-L69 has something to do with this but I haven't dove into it

## Expected behavior

Test pass

## Actual behavior

Tests fail

1. Update `react` and `react-dom` to be `16.10.2` in package.json and do a yarn install
2. Run `yarn run test`
3. Pull a funny face

Ping @AndrewMusgrave / @chloerice / @dleroux

AndrewMusgrave:

Fixes #2694

Removing the `setTimeout` mock

### How to 🎩

1. `git reset HEAD~2 --hard`
2. `dev up # or yarn install`
3. `yarn test TabMeasurer`


AndrewMusgrave:
> This fixes the tests but I'm worried that this is just fixing a symptom rather than a root cause. I'm curious about why we need that setTimeout call in the first place - and why only in development mode?

I'm not sure. IIRC at one point in time, styles loading was delayed. Just a shot in the dark but they may have been needed for measuring.

> Some folks appear to be having problems on the initial render - see the comment on the original issue and https://discourse.shopify.io/t/reloading-ssr-application-causes-polaris-tabs-to-load-differently/6389.

@dleroux Mentioned we've had this issue for a while, so I don't think it's related to the react bump.

> Perhaps the measurement handling should be done with a useLayoutEffect rather than useEffect? (which will need to be wrapped in some custom hook because it doesn't play nice with SSR)

If we do need to delay measurements, in production that would be the way to go 👍 However I don't think that'll fix the test. React uses `setTimeout` in their scheduler and I'm betting our mock breaks it 😄

AndrewMusgrave:
For reference plus created an issue #2706

PR #2709, opened 2020-02-04
[OptionList] Apply design language

dleroux:

Fixes: https://github.com/Shopify/polaris-ux/issues/354

Applying design language to OptionList. Unfortunately because of Autocomplete, the checkbox was forked and make for duplicate code.

optionlist

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Card/Card.scss
+++ b/src/components/Card/Card.scss
@@ -1,7 +1,7 @@
 @import '../../styles/common';
 
 .Card {
-  overflow: hidden;
+  overflow: var(--p-override-visible, hidden);
   background-color: var(--p-surface, color('white'));
   box-shadow: var(--p-card-shadow, shadow());
 
--- a/src/components/OptionList/components/Checkbox/Checkbox.scss
+++ b/src/components/OptionList/components/Checkbox/Checkbox.scss
@@ -3,64 +3,145 @@
 .Checkbox {
   position: relative;
   width: 100%;
+  margin: var(--p-choice-margin, 0);
 }
 
 .Input {
   @include visually-hidden;
+}
+// stylelint-disable selector-max-specificity, selector-max-class, selector-max-combinators, max-nesting-depth
+.Checkbox:not(.newDesignLanguage) {
+  .Input {
+    &:focus,
+    &:active,
+    &:checked {
+      + .Backdrop {
+        background: color('indigo');
+      }
+    }
 
-  &:focus,
-  &:active,
-  &:checked {
-    + .Backdrop {
-      background: color('indigo');
+    &:checked {
+      ~ .Icon {
+        transform: translate(-50%, -50%) scale(1);
+      }
+    }
+
+    &:focus,
+    &:active {
+      + .Backdrop {
+        @include control-backdrop(focused);
+      }
+    }
+
+    &:disabled {
+      + .Backdrop {
+        @include control-backdrop(disabled);
+      }
+
+      ~ .Icon {
+        @include recolor-icon(color('ink', 'lightest'));
+      }
     }
   }
 
-  &:checked {
-    ~ .Icon {
-      transform: translate(-50%, -50%) scale(1);
-    }
-  }
-
-  &:focus,
-  &:active {
-    + .Backdrop {
+  &.active {
+    .Backdrop {
       @include control-backdrop(focused);
     }
   }
 
-  &:disabled {
-    + .Backdrop {
-      @include control-backdrop(disabled);
-    }
-
-    ~ .Icon {
-      @include recolor-icon(color('ink', 'lightest'));
-    }
-  }
-}
-
-.active {
   .Backdrop {
-    @include control-backdrop(focused);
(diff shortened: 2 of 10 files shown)


src/components/Card/Card.scss
```diff
@@ -1,7 +1,7 @@
 @import '../../styles/common';
 
 .Card {
-  overflow: hidden;
+  overflow: var(--p-override-visible, hidden);
```
dleroux: This is causing an issue with the focus ring. I'm thinking since we are since in development we change this to overflow visible to see if this breaks anything. @danrosenthal @kyledurand what do you think?

src/components/OptionList/components/Checkbox/Checkbox.scss
```diff
@@ -3,64 +3,145 @@
...
   position: relative;
   width: 100%;
+  margin: var(--p-choice-margin, 0);
 }
 
 .Input {
   @include visually-hidden;
+}
+// stylelint-disable selector-max-specificity, selector-max-class, selector-max-combinators, max-nesting-depth
+.Checkbox:not(.globalTheming) {
```
dleroux: It's unfortunate but the Checkbox is pretty much duplicated because AutoComplete. Checkboxes should be used for comboboxes, in the new version of Autocomplete this will be able to change.

src/components/OptionList/components/Option/Option.scss
```diff
@@ -116,3 +117,53 @@ $control-vertical-adjustment: rem(2px);
...
+    // stylelint-disable-next-line selector-max-specificity
+    &:hover {
+      box-shadow: none;
+      background: var(--p-surface-hovered);
+    }
+  }
+
+  .disabled {
+    color: var(--p-text-disabled);
+    background: var(--p-action-secondary-disabled);
```
dleroux: @danrosenthal We don't seem to have a `--p-surface-disabled`. Should we be adding one?

dleroux:
@johanstromqvist @martenbjork Can you confirm when the design of this component is finalized? Just a note, these interactive states are quite different than the ones we use on resource list, is that intentional? Also, the disabled item is not in the Figma file.

martenbjork:
Will do! Thanks for the comments on the design, will make sure it gets reviewed.

src/components/Card/Card.scss
kyledurand: I'm a fan over removing this entirely. I think @dfmcphee had to port the card component in web just to remove this property. Might be best to do this in a separate PR though, for change log visibility

danrosenthal: This'll probably be considered a breaking-ish change. So we might want to wait until v5 for it to be the default. I think it makes sense to use the fallback approach until then.

This may have repercussions on other components like Resource List, which goes full width in a card. We might need to add a bunch of extra border radius to that like we've done in polaris rails.

sarahill:
Applied the design to option list. Covered the different states in the Figma file!
https://www.figma.com/file/4dAAt5iFPSaxUKiYVKrkYj/DL%E2%80%93Polaris-Web-Components?node-id=441%3A3720

image

dleroux:
This has changed quite significantly since the last round of design. There's a bump in `polaris-token` needed for the `--p-surface-disabled` property to be available.

Codecov complains about the `newDesignLaguage` feature not being tested but it is, so I'm not sure what's going on there.

src/components/OptionList/components/Checkbox/Checkbox.scss
```diff
@@ -3,64 +3,145 @@
...
   position: absolute;
   top: 50%;
   left: 50%;
-  transform: translate(-50%, -50%) scale(0);
   transform-origin: 50% 50%;
-  transition: control-icon-transition();
+  pointer-events: none;
+
+  @media (-ms-high-contrast: active) {
+    fill: ms-high-contrast-color('text');
```
kyledurand: Why not use `recolor-icon` here as well?

src/components/OptionList/components/Checkbox/Checkbox.tsx
```diff
@@ -28,8 +29,28 @@ export function Checkbox({
   role,
 }: CheckboxProps) {
   const id = useUniqueId('Checkbox', idProp);
+  const {newDesignLanguage = false} = useFeatures();
```
danrosenthal: `newDesignLanguage` is now a required key and defaults to false. You should omit the default assignment here, and in other places where you use this hook.

```suggestion
  const {newDesignLanguage} = useFeatures();
```

src/components/OptionList/components/Checkbox/Checkbox.scss
dleroux: not sure. Not something I touched. Wanted to test with virtual box which I just updated but it won't load :(

PR #2715, opened 2020-02-04
[Color system] Require newDesignLanguage feature flag

danrosenthal:


src/components/AppProvider/AppProvider.tsx
```diff
@@ -100,10 +100,12 @@ export class AppProvider extends React.Component<AppProviderProps, State> {
 
   render() {
     const {theme = {}, features = {}, children} = this.props;
+    const {newDesignLanguage = false, ...restOfFeatures} = features;
```
BPScott: I think we can dodge having to destructure to create some local variables then recombine them into into an object. I believe this has the same effect

```
const {theme = {}, children} = this.props;
const features = {newDesignLanguage: false, ...this.props.features};

// Later
 <FeaturesContext.Provider value={features}>
```

Same can be done in the test provider

src/components/PolarisTestProvider/PolarisTestProvider.tsx
```diff
@@ -59,19 +59,20 @@ const defaultMediaQuery: MediaQueryContextType = {
...
-  appBridge,
-  link,
-  theme = {},
-  mediaQuery,
-  features = {},
-  frame,
-}: PolarisTestProviderProps) {
-  const Wrapper = strict ? React.StrictMode : React.Fragment;
+export function PolarisTestProvider(props: PolarisTestProviderProps) {
+  const {
```
BPScott: I think keeping the original structure and having `features = {}` as a default prop is fine here

PR #2735, opened 2020-02-12
Add new build pipeline

amrocha:

The consumer build test has long been an unstable part of polaris-react's CI. We meant to introduce the current test as a temporary measure to improve shipping confidence, while also working on an improved version of the test that was more reliable. This is that version.

In the first implementation of the test, we duplicated web's Buildkite pipeline in polaris-react, and ran a build there. The problems with this are:
- Duplicated code. If web updates their pipeline, we don't get those updates by default;
- Incompatible dependencies. For several months, web used a different version of node than polaris-react. This meant that the test was impossible to run until web updated their node version.
- Flakiness. Web's pipelines are under constant stress, and people work to make sure they're always working. This wasn't the case in polaris-react, leading to our pipeline breaking and no one being available to fix it

This PR addresses this by building web in web's pipeline, and building polaris-react in polaris-react's pipeline. By separating the environments, we solve the problems outlined above.

The new test works as follows:
1. Build polaris-react in our own pipeline
1. Package the build using `yarn pack`, to create a package similar to what is uploaded to npm
1. Upload this build to Google Cloud Platform. I had prod-eng create a bucket and an authorization key solely for this purpose.
1. Trigger a pipeline run in web that uses the uploaded asset
1. Wait for results

I also created a web pipeline over at https://github.com/Shopify/web/pull/23433. This pipeline runs the same as any other web pipeline, but it downloads and installs the package we uploaded to the bucket earlier. 

Improvements to the test:
- Rename "Consumer Build Test" -> "System Integration Test"
- Rename pipeline "polaris-react" -> "polaris-system-integration-test"
- Pipeline now builds and uploads assets to GCP using a custom bucket and IAM auth key created solely for this purpose
- After uploading asset, triggers a build in web that downloads said asset from GCP and uses it
- Reports results in polaris-react's pipeline 

### How to 🎩

Trigger a new build here:
https://buildkite.com/shopify/polaris-system-integration-test


amrocha:
This PR merely lays out the foundation for the test, but doesn't enable it. I plan to do this in a follow-up PR. The order of operations I intend is:
- Merge https://github.com/Shopify/web/pull/23433
- Merge this PR
- Configure it to run on new commits to master
- Have a meeting with the team to determine how we should run this test: on all branches, just in master, make it required, make it optional, etc.
- Ship a follow-up PR with the agreed upon changes

amrocha:
Note to self: update `branch` part of pipeline config to point to `master` before shipping

amrocha:
@BPScott @AndrewMusgrave @tmlayton @danrosenthal @daniedleroux @kyledurand @alex-page Hey team, this is the PR updating the consumer build test

You don't need to comment on this if you don't feel comfortable giving feedback on it, but I thought it would be good for you to be aware of these changes. If you have any questions feel free to ask!

It seems like there's some interest around this, so I'll also be creating a meeting to go over what it took to ship this, and hopefully giving some insight into how Shopify's infrastructure is managed and orchestrated.

UNRELEASED.md
```diff
@@ -19,6 +19,9 @@
 ### Development workflow
 
 - Added first implementation of custom property validation ([#2616](https://github.com/Shopify/polaris-react/pull/2616))
+- Reworked the yarn splash Github comment and added average splash zone information ([#2649](https://github.com/Shopify/polaris-react/pull/2649))
+- Re-enabled the web unit tests in the consumer build test ([#2663](https://github.com/Shopify/polaris-react/pull/2663))
+- Refactor consumer build test (renamed to system integration test) ([#2735](https://github.com/Shopify/polaris-react/pull/2735))
```
AndrewMusgrave: Were these missed in other pr's or just artifacts from a rebase?

.shopify-build/polaris-system-integration-test.yml
```diff
@@ -0,0 +1,29 @@
...
+  default:
+    docker: gcr.io/shopify-docker-images/apps/production/web-platform-ci-node:10.18.0-yarn-1.21.1
+  styleguide:
+    docker: gcr.io/shopify-docker-images/apps/production/web-platform-ci-node:10.13.0-yarn-1.21.1
+
+shared:
+  build-polaris: &build-polaris
+    label: 'Build polaris'
+    container: default
+    timeout: 35m
```
AndrewMusgrave: `35m` seems long, is this what they use in web?

UNRELEASED.md
amrocha: Rebase snag, I'll fix these up!

.shopify-build/polaris-system-integration-test.yml
amrocha: Nope it's from when I was debugging and didn't want the container to close on me while I was SSH'd in, we can easily make this 5m :D

.shopify-build/polaris-system-integration-test.yml
```diff
@@ -0,0 +1,29 @@
+containers:
+  default:
+    docker: gcr.io/shopify-docker-images/apps/production/web-platform-ci-node:10.18.0-yarn-1.21.1
+  styleguide:
```
BPScott: The styleguide container isn't used anymore, we can 🔥 this

.shopify-build/polaris-system-integration-test.yml
```diff
@@ -0,0 +1,29 @@
...
+    docker: gcr.io/shopify-docker-images/apps/production/web-platform-ci-node:10.13.0-yarn-1.21.1
+
+shared:
+  build-polaris: &build-polaris
+    label: 'Build polaris'
+    container: default
```
BPScott: https://shopify-build.docs.shopify.io/jobs/sandboxes/ says that if you don't specify a container it'll used the one called 'default' so perhaps remove this as we only ever have one?

.shopify-build/polaris-system-integration-test.yml
```diff
@@ -0,0 +1,29 @@
...
+    label: 'Build polaris'
+    trigger: 'web-polaris-builder'
+    build:
+      branch: add-polaris-build-pipeline
+      commit: HEAD
+      env:
+        POLARIS_ARTIFACT_FILENAME: 'shopify-polaris-${BUILDKITE_COMMIT}.tgz'
+
+steps:
+  - <<: *build-polaris
```
BPScott: These steps aren't shared between multiple pipelines, so just put them all inline, to avoid that bit of indirection I reckon ([polaris-styleguide does this](https://github.com/Shopify/polaris-styleguide/blob/master/.shopify-build/polaris-styleguide.yml))

package.json
```diff
@@ -188,6 +188,7 @@
   ],
   "dependencies": {
     "@babel/runtime": "^7.1.6",
+    "@google-cloud/storage": "^4.3.1",
```
BPScott: Should be a dev-dependency

.shopify-build/polaris-system-integration-test.yml
amrocha: I was thinking of reusing these later for tests in other repos like the styleguide but that can come later

.shopify-build/polaris-system-integration-test.yml
```diff
@@ -0,0 +1,29 @@
...
+  styleguide:
+    docker: gcr.io/shopify-docker-images/apps/production/web-platform-ci-node:10.13.0-yarn-1.21.1
+
+shared:
+  build-polaris: &build-polaris
+    label: 'Build polaris'
+    container: default
+    timeout: 5m
+    dependencies:
+      - yarn
```
BPScott: Do [`- yarn: ~` ](https://github.com/Shopify/polaris-styleguide/blob/master/.shopify-build/polaris-styleguide.yml#L26) as a run step and you don't need the dependencies block

package.json
amrocha: I . somehow always forget this

amrocha:
Just waiting for CI to be green before merging

PR #2737, opened 2020-02-12
[DropZone] Allow changing the file if allowMultiple={false} 

devonpmack:
Fix https://github.com/Shopify/polaris-react/issues/1860
Fix https://github.com/Shopify/polaris-react/issues/2621
Fix https://github.com/Shopify/polaris-react/issues/1229
Fix https://github.com/Shopify/store/issues/13016

Removes the `numFiles` state variable. This variable was being used to check if multiple files had already been uploaded and accordingly block the dropzone from allowing more files to be uploaded. From reading the issues, it seems this blocking behaviour is unwanted and should be instead handled outside of the component if needed.

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page, DropZone} from '../src';

  return (
    <Page title="Playground">
      <DropZone
        allowMultiple={false}
        dropOnPage
        accept=" , .csv, text/csv, text/comma-separated-values, text/plain, text/x-csv, application/vnd.ms-excel, application/csv, application/x-csv, text/csv, text/comma-separated-values, text/x-comma-separated-values, text/tab-separated-values"
        onDropAccepted={() => {}}
        onDropRejected={() => {}}
        errorOverlayText="Error"
        disabled={false}
      >
        <DropZone.FileUpload />
      </DropZone>
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -18,6 +18,9 @@
 - Fixed the position property for the backdrop on `Select` from being overwritten by the focus ring ([#2748](https://github.com/Shopify/polaris-react/pull/2748))
 - Fixed `ResourceItem` `Actions` visibility on mouse out ([#2742](https://github.com/Shopify/polaris-react/pull/2742))
 - Fixed initial server / client render mismatch in `Avatar` ([#2751](https://github.com/Shopify/polaris-react/pull/2751))
+- Allow changing the `DropZone` file if allowMultiple is false ([#2737](https://github.com/Shopify/polaris-react/pull/2737))
+- Fixed `DropZone` not supporting new file selection when `allowMultiple` is `false` ([#2737](https://github.com/Shopify/polaris-react/pull/2737)) ([#2737](https://github.com/Shopify/polaris-react/pull/2737))
+- Fixed `DropZone` not supporting new file selection when `allowMultiple` is `false` ([#2737](https://github.com/Shopify/polaris-react/pull/2737))
 
 ### Documentation
 
--- a/src/components/DropZone/DropZone.tsx
+++ b/src/components/DropZone/DropZone.tsx
@@ -76,7 +76,7 @@ export interface DropZoneProps {
   /** Text that appears in the overlay when set in error state */
   errorOverlayText?: string;
   /**
-   * Allows multiple files to be uploaded
+   * Allows multiple files to be uploaded at once
    * @default true
    */
   allowMultiple?: boolean;
@@ -181,7 +181,6 @@ export const DropZone: React.FunctionComponent<DropZoneProps> & {
     setTrue: handleFocus,
     setFalse: handleBlur,
   } = useToggle(false);
-  const [numFiles, setNumFiles] = useState(0);
   const [size, setSize] = useState('extraLarge');
   const [measuring, setMeasuring] = useState(true);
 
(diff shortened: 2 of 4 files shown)


devonpmack:
@chloerice @dleroux  Let me know what you guys think, I'm not confident that this is the right approach to fixing this. It's possible that we might break some apps that are expecting the current blocking behaviour.

dleroux:
I haven't played with Dropzone too much but this feels like it would be a breaking change since there's nothing warning the user that only single files are accepted. In top hatting your PR I notice that the rejected files never get updated because `splice` actually mutates the array.

https://github.com/Shopify/polaris-react/pull/2737/files#diff-7ee058419e736ffe4c643519496b920fR203

I'm not sure what the best way to solve this is. Maybe it needs to be added to a breaking version? @chloerice what do you think?

UNRELEASED.md
```diff
@@ -13,6 +13,7 @@
 - Fixed `TrapFocus` stealing focus from other `TrapFocus`'s ([#2681](https://github.com/Shopify/polaris-react/pull/2681))
 - Fixed focus state color on monochrome `Buttons` ([#2684](https://github.com/Shopify/polaris-react/pull/2684))
 - Fixed container's width on `Modal` ([#2692](https://github.com/Shopify/polaris-react/pull/2692))
+- Allow changing the `DropZone` file if allowMultiple is false ([#2737](https://github.com/Shopify/polaris-react/pull/2737))
```
```suggestion
- Fixed `DropZone` not supporting new file selection when `allowMultiple` is `false` ([#2737](https://github.com/Shopify/polaris-react/pull/2737))
```

dleroux:
The reason I was thinking this is breaking @chloerice is that the onus is now on the consumer to ensure their UI shows only 1 thumbnail. If they haven't done that then someone could see 10 thumbnails, but in fact, only the last one will get updated. Although not that likely, it could be quite frustrating to the user. But you are right not a breaking change per se.

I noticed it while top hatting our examples. Maybe adding a single file upload example with thumbnails 
would help avoid this issue.

devonpmack:
@dleroux I see your point, I can't think of an obvious solution to the problem but I'm assuming not many consumers will be making that mistake.

I added a working example as you said :)

devonpmack:
The example looks like this:

image

Identical to the multiple file dropzone but uploading additional files just replaces the current one.

devonpmack:
Hmm, I'm unable to tophat the documentation:

```
❌ Unable to parse information for property "theme" for component "components/AppProvider/AppProvider" TypeError: Cannot read property 'name' of undefined
    at flattenedTypesArray.reduce (/Users/devon/src/github.com/Shopify/polaris-styleguide/scripts/parseComponentTypes.js:209:34)
    at Array.reduce (<anonymous>)
    at Promise.all.then (/Users/devon/src/github.com/Shopify/polaris-styleguide/scripts/parseComponentTypes.js:208:34)
```

```
./node_modules/@shopify/polaris/index.es.js
Module not found: Can't resolve '@shopify/polaris-tokens/dist-modern' in '/Users/devon/src/github.com/Shopify/polaris-styleguide/node_modules/@shopify/polaris'

./node_modules/@shopify/polaris/index.es.js
Module not found: Can't resolve '@shopify/polaris-tokens/dist-modern/theme/base.json' in '/Users/devon/src/github.com/Shopify/polaris-styleguide/node_modules/@shopify/polaris'

[client] Build failed.
```

dleroux:
@devonpmack I'm checking with @BPScott because at first, I thought it was a version mismatch with tokens but then updating the style guide didn't fix the issue. This is happening on master as well so it's not this PR.

Test coverage is a little low, I have a feeling if you test the disabled line you'll get proper coverage.

devonpmack:
@dleroux I fixed the code coverage :) Is this good to go?

PR #2746, opened 2020-02-14
[Button] New design language 

kyledurand:
This PR is meant to be the final push from partially completed in React to full

https://github.com/Shopify/polaris-ux/issues/371

I'd add some screenshots but I feel that would be ineffective. The best way to 🎩is to 🎩

- Adds the focus ring where necessary according to Figma
image

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -359,9 +359,10 @@ $stacking-order: (
   }
 
   &.destructive {
-    color: color('red', 'dark');
-    &:hover {
-      color: color('red', 'darker');
+    color: var(--p-action-critical, color('red', 'dark'));
+    &:focus,
+    &:active {
+      color: var(--p-action-critical-pressed, color('red', 'darker'));
     }
   }
 
@@ -410,11 +411,15 @@ $stacking-order: (
   // stylelint-disable selector-max-class
   &.newDesignLanguage {
     // redeclaring these styles here for specificity
+    @include recolor-icon(var(--p-action-interactive));
     background: transparent;
     border: 0;
     box-shadow: none;
-    @include recolor-icon(var(--p-interactive));
-    color: var(--p-interactive);
+    color: var(--p-action-interactive);
+
+    &::after {
+      content: none;
+    }
 
     &:hover,
     &:focus,
@@ -430,26 +435,27 @@ $stacking-order: (
     }
 
     &:hover {
-      @include recolor-icon(var(--p-interactive-hovered));
-      color: var(--p-interactive-hovered);
+      @include recolor-icon(var(--p-action-interactive-hovered));
+      color: var(--p-action-interactive-hovered);
       text-decoration: underline;
     }
 
     &:focus {
-      @include recolor-icon(var(--p-interactive));
-      color: var(--p-interactive);
-      text-decoration: none;
+      @include recolor-icon(var(--p-action-interactive));
+      color: var(--p-action-interactive);
       @include no-focus-ring;
     }
 
     &:active {
-      @include recolor-icon(var(--p-interactive-pressed));
-      color: var(--p-interactive-pressed);
-      text-decoration: none;
+      @include recolor-icon(var(--p-action-interactive-pressed));
+      color: var(--p-action-interactive-pressed);
     }
 
-    &:focus:not(:active) > .Content {
-      @include focus-ring($style: 'focused');
+    &:focus:not(:active) {
+      text-decoration: none;
+      > .Content {
+        @include focus-ring($style: 'focused');
+      }
     }
 
     &:focus:not(.iconOnly) > .Content {
@@ -461,17 +467,17 @@ $stacking-order: (
     }
 
     &.destructive {
-      @include recolor-icon(var(--p-interactive-critical));
-      color: var(--p-interactive-critical);
+      @include recolor-icon(var(--p-action-critical));
+      color: var(--p-action-critical);
       &:hover {
-        color: var(--p-interactive-critical-hovered);
+        color: var(--p-action-critical-hovered);
       }
       &:active {
-        color: var(--p-interactive-critical-pressed);
+        color: var(--p-action-critical-pressed);
       }
 
       &.disabled {
-        color: var(--p-interactive-critical-disabled);
+        color: var(--p-action-critical-disabled);
       }
     }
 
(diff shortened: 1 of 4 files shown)


src/styles/foundation/_focus-ring.scss
```diff
@@ -1,18 +1,17 @@
 /// Sets the focus ring for an interactive element
 /// @param {String} $size - The size of the border radius on the focus ring.
 /// @param {String} $style - Focus ring state.
-/// @param {Number} $border-width - The border width of your element in rems.
+/// @param {Number} $border-width - The border width of your element in px.
 ///
 
-@mixin focus-ring($size: 'base', $border-width: rem(0), $style: 'base') {
+@mixin focus-ring($size: 'base', $border-width: '0px', $style: 'base') {
```
kyledurand: I had to change this because `rem(0)` outputs `0` without `px` and that will break the border-radius calc below, resulting in a 0px border radius if `$border-width` is not passed to the mixin

i.e. 
`calc(var(--p-border-radius-base) + 0 + 0.1rem)` will render no border radius
vs
`calc(var(--p-border-radius-base) + 0px + 0.1rem)` will render a proper border radius

src/styles/foundation/_focus-ring.scss
```diff
@@ -1,18 +1,17 @@
...
   $stroke: rem(2px);
-  $offset: calc(#{$border-width} + #{rem(1px)});
+  $offset: calc(-1 * calc(#{$border-width} + #{rem(1px)}));
```
kyledurand: @BPScott can you for-see any issues with nesting a calc?

src/styles/foundation/_focus-ring.scss
```diff
@@ -21,15 +20,15 @@
...
+      right: $offset;
+      bottom: $offset;
+      left: $offset;
       display: block;
       pointer-events: none;
-      box-shadow: 0 0 0 $negative-offset var(--p-focused);
+      box-shadow: 0 0 0 $offset var(--p-focused);
       transition: box-shadow duration(fast) var(--p-ease);
-      border-radius: calc(#{$border-radius} + #{rem(1px)});
+      border-radius: calc(#{$border-radius} + #{$border-width} + #{rem(1px)});
```
kyledurand: The border radius was off on some elements. I've tested this with the kitchen sink and it seems to match every element perfectly

src/styles/shared/_buttons.scss
```diff
@@ -235,13 +236,16 @@
     }
 
     &:focus {
-      box-shadow: 0 0 0 border-width('thick') var(--p-border-secondary);
-      @include no-focus-ring;
+      box-shadow: 0 0 0 border-width('base') var(--p-border-secondary);
```
kyledurand: This was probably out of date and changed in figma, or an oversight. This is the proper border width

src/components/Button/Button.tsx
```diff
@@ -325,6 +329,7 @@ export function Button({
         onKeyPress={onKeyPress}
         onMouseUp={handleMouseUpByBlurring}
         onMouseEnter={onMouseEnter}
+        onMouseLeave={handleMouseLeaveByBlurring}
```
kyledurand: This could be overkill but it prevents the focus ring from appearing when a user clicks and holds on a button and drags off of it.

i.e.
image

src/components/Button/Button.scss
```diff
@@ -410,11 +411,15 @@ $stacking-order: (
   // stylelint-disable selector-max-class
   &.newDesignLanguage {
     // redeclaring these styles here for specificity
+    @include recolor-icon(var(--p-action-interactive));
     background: transparent;
     border: 0;
     box-shadow: none;
-    @include recolor-icon(var(--p-interactive));
-    color: var(--p-interactive);
+    color: var(--p-action-interactive);
```
kyledurand: These must have been missed in a `--p-interactive` find / replace

src/styles/foundation/_focus-ring.scss
danrosenthal: Nested calc seems unnecessary, if this is just about OOO, you can wrap it in parens.
```suggestion
  $offset: calc(-1 * (#{$border-width} + #{rem(1px)}));
```

src/components/Button/Button.tsx
danrosenthal: This feels like overkill

danrosenthal:
Does this include the `pressed` state for Button? I was under the impression that was the remaining outstanding change.

kyledurand:
Pressed states: https://screenshot.click/screencast_2020-02-18_09-32-32.mp4

Monochrome is very subtle in figma as well. Maybe we can do better there

src/styles/foundation/_focus-ring.scss
kyledurand: It definitely needs to be in a calc because `$border-width` could be a custom property

src/components/Button/Button.tsx
kyledurand: Removed 👍

src/styles/foundation/_focus-ring.scss
BPScott: Sorry, this mention totally missed my radar. I'm pretty sure nested calcs are fine.

In the past we've been bitten by [forgetting to wrap math that may involve a custom property in a calc](https://github.com/Shopify/polaris-react/commit/334c3b8a4dc0554f082d58bd15e628923ed1c426#diff-d9cca79bcce0f8de4993d487aa6c3646) and [post-css-env not liking the order of items in a calc when one of the items is a custom property](https://github.com/Shopify/polaris-react/pull/2643) but it doesn't look like this falls into similar traps.

I thiiink Dan might be right, If `$border-width` was a custom property then you'd end up with something like `calc(-1 * (var(--something) + 1px));` looks like legit code - the `var()` is already inside a `calc()`. (Disclaimer this is me assuming things, I've not ran this to confirm)

If you want a bit of extra confidence that we won't be caught by surprise like those last two issues, you can spin up a CRA app and copy polaris's compiled css (the generated `styles.css` you get from doing a `yarn build`) into it and see if it builds without issue.

kyledurand:
> A bunch of these updates to color variants are wrong

So weird, those colors weren't working locally for me. I think I may have to clear my yarn cache

src/styles/foundation/_focus-ring.scss
dleroux: I agree. We're after 0px in the end and 0 is 0.

src/styles/shared/_buttons.scss
```diff
@@ -235,13 +236,16 @@
...
+      @include focus-ring($style: 'focused');
     }
 
     &:active {
       box-shadow: 0 0 0 border-width('base') var(--p-border-secondary);
       background: var(--p-surface-pressed);
+      &::after {
+        box-shadow: none;
```
dleroux: Just a thought... we end up overwriting things about the ring in a few places. It doesn't feel right because it's so far detached for the source. If we change the ring, this will just add tech debt because chances are it will be left behind. Adding more arguments to the focus-ring with defaults would allow us to keep all the focus ring CSS in 1 location.

kyledurand: Good point! I've been chatting with Alex about refactoring button css once the design language has been rolled out. I think that would be a good time to address things like this as we'll have a better idea of how the focus ring needs to be used

PR #2751, opened 2020-02-16
[Avatar] fix initial server / client render mismatch

robin-drexler:

Fixes https://github.com/Shopify/polaris-react/issues/1419 

In every project that uses/renders the Avatar component on server and client, we get a React warning about the initial rendering not being equal.

While mismatches in server-rendered in client-rendered results usually don't seem to be as much of an issue anymore as of react 16, they can be if nodes are missing in the server-rendered output, which I think is precisely the case here.

> Also, React 16 is better at hydrating server-rendered HTML once it reaches the client. It no longer requires the initial render to exactly match the result from the server.
**However, it's dangerous to have missing nodes on the server render as this might cause sibling nodes to be created with incorrect attributes.**

https://reactjs.org/blog/2017/09/26/react-v16.0.html
Example issue: https://github.com/facebook/react/issues/10591

There's another issue. Unfortunately, it seems like React only ever warns for the first mismatch. In some projects, (almost) all pages contain the Avatar component, and since it always produces an error, we might have been missing others.

Instead of not rendering the image on the server and rendering it on the client, we now wait for the component to be mounted to render the image. So it won't be part of the server render, but also not part of the initial client side render anymore. 

This does cause an addional rerender, but I'd assume that's ok(?)
The change should have no effect on the UX/behavior of the component.

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page, Avatar} from '../src';

  return (
    <Page title="Playground">
      <Avatar source="https://example.com/does-not-exist" initials="WE" />
      <Avatar source="http://placekitten.com/200/200" initials="WE" />
    </Page>
  );
}

```

</details>

First avatar will fail to load the image and show the initials.
Second avatar loads and shows image.

### 🎩 checklist

### DIFF
--- a/src/components/Avatar/tests/Avatar-ssr.test.tsx
+++ b/src/components/Avatar/tests/Avatar-ssr.test.tsx
@@ -3,11 +3,13 @@ import React from 'react';
 import {mountWithAppProvider} from 'test-utilities/legacy';
 import {Avatar, Image} from 'components';
 
-jest.mock('../../../utilities/target', () => ({
-  get isServer() {
-    return true;
-  },
-}));
+jest.mock('../../../utilities/use-is-after-initial-mount', () => {
+  return {
+    useIsAfterInitialMount: () => {
+      return false;
+    },
+  };
+});
 
 describe('<Avatar /> Server-side only', () => {
   it('does not render an Image', () => {
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -15,6 +15,7 @@
 - Fixed container's width on `Modal` ([#2692](https://github.com/Shopify/polaris-react/pull/2692))
 - Fixed the position property for the backdrop on `Select` from being overwritten by the focus ring ([#2748](https://github.com/Shopify/polaris-react/pull/2748))
 - Fixed `ResourceItem` `Actions` visibility on mouse out ([#2742](https://github.com/Shopify/polaris-react/pull/2742))
+- Fixed initial server / client render mismatch in `Avatar` ([#2751](https://github.com/Shopify/polaris-react/pull/2751))
 
 ### Documentation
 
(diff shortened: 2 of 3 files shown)


robin-drexler:
I will fix the conflicts in `UNRELEASED.md` after a review. Otherwise I'll have to do that multiple times a day. :D

src/components/Avatar/tests/Avatar-ssr.test.tsx
```diff
@@ -3,11 +3,13 @@ import React from 'react';
 import {mountWithAppProvider} from 'test-utilities/legacy';
 import {Avatar, Image} from 'components';
 
-jest.mock('../../../utilities/target', () => ({
-  get isServer() {
-    return true;
-  },
-}));
+jest.mock('../../../utilities/use-is-after-initial-mount', () => {
```
BPScott: It doesn't feel like this test adds much value anymore, mocking out individual hooks feels like a weird smell. Do we care that much about initial renderings?


Strictly speaking this isn't even testing SSR rendering - it's a bad proxy of it by mocking out a single hook. I wonder if ideally react-testing should provide some version of `mountOnServer` that renders a component using `react-dom/server` (like the hack described in https://github.com/testing-library/react-testing-library/issues/561), but that feels like a meaty problem for another day.


---

On a very petty note, we can save the explicit return:

```
jest.mock('../../../utilities/use-is-after-initial-mount', () => ({
  useIsAfterInitialMount: () => false,
}));
```

robin-drexler: To me it provides the same value as before. Earlier we were pretending to be on a server environment by mocking a constant now we simulate that by mocking a hook. Even though I agree earlier it was more explicit. I also agree that a `renderOnServer` function that doesn't mount would be pretty great for these cases. Do you want me to delete the test?

As for the `petty note`: I generally try to avoid to use single line arrow functions for two reasons. The `()` in `({})` easy to forget, which leads to the function not returning anything. And secondly, if you want to add a second line, it's quite tedious to make that happen and I don't see any advantage of using the single line notation in the first place. 
That being said, it's personal preference and I'd be open to change it.

BPScott: If you think it's still got some value let's keep it :)

Same goes for the petty note bits - you've got a solid rational, and it's personal pref vs idle musings.

Keep it all as it is and :shipit: :)

PR #2753, opened 2020-02-18
[Modal] Fix shadow

kyledurand:
Modal has a much broader shadow in figma, it's also darker. I've added a token and modified the modal shadow to match [Figma](https://www.figma.com/file/4dAAt5iFPSaxUKiYVKrkYj/DL%E2%80%93Polaris-(Web)%3A-Components?node-id=358%3A0)

Before:
image

After:
image

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/utilities/theme/tokens.ts
+++ b/src/utilities/theme/tokens.ts
@@ -9,7 +9,7 @@ const Shadow = {
   popoverShadow:
     '-1px 0px 20px var(--p-shadow-from-ambient-light), 0px 1px 5px var(--p-shadow-from-direct-light)',
   modalShadow:
-    '0px 6px 32px var(--p-shadow-from-ambient-light), 0px 1px 6px var(--p-shadow-from-direct-light)',
+    '0px 26px 80px var(--p-shadow-from-dim-light), 0px 0px 1px var(--p-shadow-from-dim-light)',
   topBarShadow: '0 2px 2px -1px var(--p-shadow-from-direct-light)',
 };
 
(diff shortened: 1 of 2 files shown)


src/utilities/theme/tokens.ts
```diff
@@ -4,12 +4,13 @@ const BorderRadius = {
 };
 
 const Shadow = {
+  shadowFromDimLight: 'rgba(0, 0, 0, 0.2)',
```
danrosenthal: You'll want to add this to tokens instead: https://github.com/Shopify/polaris-tokens/blob/next/formats/utils/color-factory/configs/base.js#L98

kyledurand: Can we get into the habit of adding these locally until design reviews are completed and then moving them to tokens?

It just seems like a round about way of doing things. I'd have to add the token in the package, bump the version and then make the change in Polaris.

I'd prefer adding locally and then removing chunks later on as they become available in tokens

danrosenthal: I'm comfortable with that. But it should have an accompanying issue for the removal here, and a pull request in tokens.

kyledurand: I'll put the PR up in tokens. Shouldn't we have a PR to move a bunch of local tokens? Or is this the only one you think worth moving right now?

danrosenthal: Currently the only tokens that live in polaris tokens are those generated by the color factory, so no, there shouldn't be anything else we need to migrate currently from `tokens.ts`

danrosenthal: As we start to explore customizable tokens for font-size, border-radius, we'll work to move those tokens over, but that work is not ready yet.

PR #2754, opened 2020-02-18
[Popover] Add preferInputActivator prop to Popover

BerkiAttila:
Co-authored-by: Robin Drexler <robin.drexler@shopify.com>

Fixes https://github.com/Shopify/polaris-react/issues/2752
Fixes https://github.com/Shopify/polaris-react/issues/1415

When the activator of a Popover (or PositonedOverlay) contains an input element, the Overlay always uses said element to calculate its position and width/height.

Introduces a `preferInputActivator` prop to the `Popover` (and hence `PositonedOverlay` component). This prop would be `true` by default to not introduce a breaking change. 

   When it's `true`, the behavior does not change. The Overlay will continue to search for an `input`, and if it finds one, use it.  If the consumer sets the prop to `false`, the Overlay will never search for an `input` and always use the `activator` to calculate its dimensions.

## 

### How to 🎩

If you set the `preferActivatorInput` to false the Popover should expand to the size of the Textfield. By removing the property it should only have the width of the nested input node.

Before:
image

After:
image

<details>

```jsx
import React from 'react';
import {Page, Popover, TextField, Checkbox, Icon} from '../src';

  return (
    <Page title="Playground">
      <Popover
        active
        onClose={() => {}}
        fullWidth
        preferredAlignment="left"
        preferInputActivator={false}
        activatorWrapper="span"
        activator={
          <div>
            <TextField
              label="TextField"
              onChange={() => {}}
              prefix={<Icon source={SearchMinor} color="inkLightest" />}
            />
          </div>
        }
      >
        <div>
          <Checkbox label="Salut" checked={false} onChange={() => {}} />
        </div>
      </Popover>
    </Page>
  );
}


```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -16,6 +16,7 @@
 - Fixed the position property for the backdrop on `Select` from being overwritten by the focus ring ([#2748](https://github.com/Shopify/polaris-react/pull/2748))
 - Fixed `ResourceItem` `Actions` visibility on mouse out ([#2742](https://github.com/Shopify/polaris-react/pull/2742))
 - Fixed initial server / client render mismatch in `Avatar` ([#2751](https://github.com/Shopify/polaris-react/pull/2751))
+- Fixed the `Popover` reference position calculation ([#2752](https://github.com/Shopify/polaris-react/issues/2752)) ([#1415](https://github.com/Shopify/polaris-react/issues/1415))
 
 ### Documentation
 
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -7,6 +7,7 @@
 - Added high contrast outline to `ActionList` ([#2713](https://github.com/Shopify/polaris-react/pull/2713))
 - Added high contrast border to `Button` ([#2712](https://github.com/Shopify/polaris-react/pull/2712))
 - Added styled placeholder image to `Avatar` when initials are blank ([#2693](https://github.com/Shopify/polaris-react/pull/2693))
+- Added a `preferInputActivator` prop to `Popover` to allow better positioning of the overlay ([#2754](https://github.com/Shopify/polaris-react/pull/2754))
 
 ### Bug fixes
 
@@ -16,7 +17,6 @@
 - Fixed the position property for the backdrop on `Select` from being overwritten by the focus ring ([#2748](https://github.com/Shopify/polaris-react/pull/2748))
 - Fixed `ResourceItem` `Actions` visibility on mouse out ([#2742](https://github.com/Shopify/polaris-react/pull/2742))
 - Fixed initial server / client render mismatch in `Avatar` ([#2751](https://github.com/Shopify/polaris-react/pull/2751))
-- Fixed the `Popover` reference position calculation ([#2752](https://github.com/Shopify/polaris-react/issues/2752)) ([#1415](https://github.com/Shopify/polaris-react/issues/1415))
 
 ### Documentation
 
--- a/src/components/Popover/Popover.tsx
+++ b/src/components/Popover/Popover.tsx
@@ -133,9 +133,8 @@ export const Popover: React.FunctionComponent<PopoverProps> & {
   }, [activatorNode, setAccessibilityAttributes]);
 
   const portal = activatorNode ? (
-    <Portal idPrefix="popover" testID="portal">
+    <Portal idPrefix="popover">
       <PopoverOverlay
-        testID="popoverOverlay"
         id={id}
         activator={activatorNode}
         preferInputActivator={preferInputActivator}
--- a/src/components/Popover/Popover.tsx
+++ b/src/components/Popover/Popover.tsx
@@ -32,6 +32,10 @@ export interface PopoverProps {
   active: boolean;
   /** The element to activate the Popover */
   activator: React.ReactElement;
+  /**
+   * Use the activator's input element to calculate the Popover position
+   * @default true
+   */
   preferInputActivator?: PopoverOverlayProps['preferInputActivator'];
   /**
    * The element type to wrap the activator with
--- a/src/components/Popover/Popover.tsx
+++ b/src/components/Popover/Popover.tsx
@@ -32,6 +32,7 @@ export interface PopoverProps {
   active: boolean;
   /** The element to activate the Popover */
   activator: React.ReactElement;
+  preferInputActivator?: PopoverOverlayProps['preferInputActivator'];
   /**
    * The element type to wrap the activator with
    * @default 'div'
@@ -71,6 +72,7 @@ export const Popover: React.FunctionComponent<PopoverProps> & {
   active,
   fixed,
   ariaHaspopup,
+  preferInputActivator = true,
   ...rest
 }: PopoverProps) {
   const [activatorNode, setActivatorNode] = useState();
@@ -136,6 +138,7 @@ export const Popover: React.FunctionComponent<PopoverProps> & {
         testID="popoverOverlay"
         id={id}
         activator={activatorNode}
+        preferInputActivator={preferInputActivator}
         onClose={handleClose}
         active={active}
         fixed={fixed}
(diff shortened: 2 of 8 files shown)


src/components/Popover/components/PopoverOverlay/tests/PopoverOverlay.test.tsx
```diff
@@ -143,6 +143,25 @@ describe('<PopoverOverlay />', () => {
...
+        onClose={noop}
+        fixed
+        preferInputActivator={false}
+      >
+        {children}
+      </PopoverOverlay>,
+    );
+
+    expect(
+      popoverOverlay.find(PositionedOverlay).prop('preferInputActivator'),
```
robin-drexler: Unfortunately, we can't check for the correct `Rect` being passed in the render prop because jsdom will always return `0` for all values. So we thought this is the second best way of assuring that the input has been used.

src/components/PositionedOverlay/tests/PositionedOverlay.test.tsx
```diff
@@ -131,6 +132,51 @@ describe('<PositionedOverlay />', () => {
...
+      mountWithAppProvider(
+        <PositionedOverlay
+          {...mockProps}
+          preferInputActivator
+          activator={activator}
+        />,
+      );
+
+      expect(
+        getRectForNodeSpy.mock.calls.some(([node]) => node === input),
```
robin-drexler: Unfortunately, we can't check for the correct `Rect` being passed in the render prop because jsdom will always return `0` for all values. So we thought this is the second best way of assuring that the input has been used.

robin-drexler: Also `getRectForNode` is called multiple times for different elements so we could not just use `toHaveBeenCalled`

src/components/Popover/tests/Popover.test.tsx
```diff
@@ -95,6 +95,21 @@ describe('<Popover />', () => {
     expect(popoverOverlay.prop('preferredAlignment')).toBe('left');
   });
 
+  it("passes 'preferInputActivator' to PopoverOverlay", () => {
+    const popover = mountWithAppProvider(
```
dleroux: small nit, we're trying to switch over the react-testing as much as possible. If you can, change this to use `mountWithApp` instead and avoid the testId.

src/components/PositionedOverlay/PositionedOverlay.tsx
```diff
@@ -195,12 +196,13 @@ export class PositionedOverlay extends React.PureComponent<
           onScrollOut,
           fullWidth,
           fixed,
+          preferInputActivator = true,
         } = this.props;
 
         const textFieldActivator = activator.querySelector('input');
```
dleroux:
```suggestion
        const activator = preferInputActivator ? activator.querySelector('input') : activatorProp
```
Above where you deconstruct the props, I woud just rename `activator` to `activator: activatorProp`

src/components/PositionedOverlay/PositionedOverlay.tsx
```diff
@@ -195,12 +196,13 @@ export class PositionedOverlay extends React.PureComponent<
...
 
         const activatorRect =
```
dleroux:
```suggestion
        const activatorRect = getRectForNode(activator);
```

The reason I am making these suggestions is that we won't do an unnecessary `activator.querySelector('input')` if `preferInputActivator` is false,  but also because by letting people specify `preferInputActivator` and then doing this logic, the component might still work but not return what was intended. Does that make sense?

src/components/PositionedOverlay/tests/PositionedOverlay.test.tsx
dleroux: Makes sense. Maybe @AndrewMusgrave will have opinion about this.

src/components/PositionedOverlay/PositionedOverlay.tsx
robin-drexler: yeah, makes sense. We tried to keep this PR as small as possible to make it easier to review and hence didn't want to touch more than needed to introduce our change. We'll be happy to adjust it, though.

src/components/Popover/tests/Popover.test.tsx
robin-drexler: We actually thought about that, but wanted to keep the test file consistent and not use different testing libraries in one file.
Depending on the effort, we'd try to refactor the entire file.

dleroux: That would be great if you could. We have been mixing though :|

src/components/PositionedOverlay/PositionedOverlay.tsx
robin-drexler: Wouldn't that introduce a breaking change? In it's current state `activatorRect` will always be based on an existing node. If the `input` does not exist, we use the `activator`

With you suggested change, we might end up with `null`. When `preferInputActivator` is `true`, but there's no `input`.

Since `preferInputActivator` defaults to true this would probably break most of the usages where no `TextField` is the `activator`

dleroux: yes my bad. I missed the part we do check for null today. Was trying to get away for querying the for the input if `preferInputActivator` is false.

src/components/Popover/tests/Popover.test.tsx
BerkiAttila: We updated the entire test file, could you please take a look?

src/components/PositionedOverlay/PositionedOverlay.tsx
```diff
@@ -195,14 +196,14 @@ export class PositionedOverlay extends React.PureComponent<
           onScrollOut,
           fullWidth,
           fixed,
+          preferInputActivator = true,
         } = this.props;
 
-        const textFieldActivator = activator.querySelector('input');
+        const preferredActivator = preferInputActivator
+          ? activator.querySelector('input') || activator
```
robin-drexler: We now only query for `input` when `preferredActivator` is `true`.

src/components/Popover/tests/Popover.test.tsx
dleroux: That's awesome thank you! 1 test is failing and there's a merge conflict but apart from that, it looks good!

robin-drexler:
We'll update the conflicting `UNRELEASED.md` as soon as this PR is approved and we intend to ship. The file unfortunately conflicts pretty quickly and often. 😅

src/components/Popover/Popover.tsx
```diff
@@ -32,6 +32,7 @@ export interface PopoverProps {
   active: boolean;
   /** The element to activate the Popover */
   activator: React.ReactElement;
+  preferInputActivator?: PopoverOverlayProps['preferInputActivator'];
```
dleroux:
```suggestion
  /**
   * Use the activator's input element to calculate the Popover position
   * @default true
   */
  preferInputActivator?: PopoverOverlayProps['preferInputActivator'];
```

UNRELEASED.md
```diff
@@ -14,6 +14,7 @@
 - Fixed focus state color on monochrome `Buttons` ([#2684](https://github.com/Shopify/polaris-react/pull/2684))
 - Fixed container's width on `Modal` ([#2692](https://github.com/Shopify/polaris-react/pull/2692))
 - Fixed the position property for the backdrop on `Select` from being overwritten by the focus ring ([#2748](https://github.com/Shopify/polaris-react/pull/2748))
+- Fixed the `Popover` reference position calculation ([#2752](https://github.com/Shopify/polaris-react/issues/2752)) ([#1415](https://github.com/Shopify/polaris-react/issues/1415))
```
```suggestion
```

dleroux: This should probably go into the "Enhancements" section. Also, the link simply points to the PR as opposed to the issue.

- Added a `preferInputActivator` prop to `Popover` to allow better positioning of the overlay ([#2754](https://github.com/Shopify/polaris-react/pull/2754))

robin-drexler: It's technically (also) a bugfix for https://github.com/Shopify/polaris-react/issues/2752, but I have no issues moving it in the enhancements section. Thanks for the hint about linking to the issue. I wasn't aware that it should point to that.

robin-drexler:
@dleroux thanks for your review and suggestions. This PR should be good to go now. Should we wait for a second review or can we :shipit: ? :)

PR #2756, opened 2020-02-19
[Color system] add theme provider readme

danrosenthal:
|:tophat:|
|-|
|screencapture-polaris-myshopify-io-components-structure-theme-provider-2020-02-19-10_04_07|

## Todo

- [ ] add proper links to new documentation
- [ ] add ability for examples to declaratively opt out of rendering the default app provider @BPScott

### DIFF
--- a/src/components/ThemeProvider/README.md
+++ b/src/components/ThemeProvider/README.md
@@ -9,6 +9,7 @@ keywords:
   - color scheme
   - light mode
   - dark mode
+omitAppProvider: true
 ---
 
 # Theme provider
@@ -42,7 +43,7 @@ The app provider component renders a theme provider component and a theme. By de
 
 ### Theme provider with theme colors rendered by the app provider
 
-A custom theme can be passed to the theme prop on the app provider to override one or more of the default theme colors at a global level. 
+A custom theme can be passed to the theme prop on the app provider to override one or more of the default theme colors at a global level.
 
 ```jsx
 <AppProvider
--- a/src/components/ThemeProvider/README.md
+++ b/src/components/ThemeProvider/README.md
@@ -198,7 +198,7 @@ Using the `linear-gradient()` function as a fallback to the `var()` function cau
 
 ```scss
 .Button {
-  background-color: linear-gradient(
+  background-image: linear-gradient(
     to bottom,
     color('indigo'),
     color('indigo', 'dark')
@@ -206,6 +206,7 @@ Using the `linear-gradient()` function as a fallback to the `var()` function cau
 }
 
 .Button.newDesignLanguage {
+  background-image: var(--p-override-none);
   background-color: var(--p-action-primary);
 }
 ```
--- a/src/components/ThemeProvider/README.md
+++ b/src/components/ThemeProvider/README.md
@@ -1,0 +1,230 @@
+---
+name: Theme provider
+category: Structure
+keywords:
+  - theme
+  - provider
+  - color
+  - logo
+  - color scheme
+  - light mode
+  - dark mode
+---
+
+# Theme provider
+
+Use theme provider to share global theme settings throughout the hierarchy of your application. Theme provider is included by default as a child of the [app provider component](https://polaris.shopify.com/components/structure/app-provider) but can be used independently to apply a base theme or to override a parent theme in its children.
+
+---
+
+## Examples
+
+### Theme provider rendered by the app provider
+
+The app provider component renders a theme provider component and a theme. By default, theme falls back to the current design language. The new design language is enabled by passing `{newDesignLanguage: true}` to the `features` prop on the app provider component.
+
+```jsx
+<AppProvider i18n={{}} features={{newDesignLanguage: true}}>
+  <Card
+    title="Shipment 1234"
+    secondaryFooterActions={[{content: 'Edit shipment'}]}
+    primaryFooterAction={{content: 'Add tracking number'}}
+  >
+    <Card.Section title="Items">
+      <List>
+        <List.Item>1 × Oasis Glass, 4-Pack</List.Item>
+        <List.Item>1 × Anubis Cup, 2-Pack</List.Item>
+      </List>
+    </Card.Section>
+  </Card>
+</AppProvider>
+```
+
+### Theme provider with theme colors rendered by the app provider
+
+A custom theme can be passed to the theme prop on the app provider to override one or more of the default theme colors at a global level. 
+
+```jsx
+<AppProvider
+  i18n={{}}
+  features={{newDesignLanguage: true}}
+  theme={{
+    colors: {
+      surface: '#111213',
+      onSurface: '#111213',
+      interactive: '#2e72d2',
+      secondary: '#111213',
+      primary: '#3b5998',
+      critical: '#d82c0d',
+      warning: '#ffc453',
+      highlight: '#5bcdda',
+      success: '#008060',
+      decorative: '#ffc96b',
+    },
+  }}
+>
+  <Card
+    title="Shipment 1234"
+    secondaryFooterActions={[{content: 'Edit shipment'}]}
+    primaryFooterAction={{content: 'Add tracking number'}}
+  >
+    <Card.Section title="Items">
+      <List>
+        <List.Item>1 × Oasis Glass, 4-Pack</List.Item>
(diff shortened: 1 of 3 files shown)


src/components/ThemeProvider/README.md
```diff
@@ -0,0 +1,230 @@
...
+  - color
+  - logo
+  - color scheme
+  - light mode
+  - dark mode
+---
+
+# Theme provider
+
+Theme provider is a component that enables sharing global theme settings throughout the hierarchy of your application. It is included by default as a child of the [app provider component](https://polaris.shopify.com/components/structure/app-provider), but can be used independently to apply a base theme, or to override a parent theme in its children.
```
sadiesaurus:
```suggestion
Use theme provider to share global theme settings throughout the hierarchy of your application. Theme provider is included by default as a child of the [app provider component](https://polaris.shopify.com/components/structure/app-provider) but can be used independently to apply a base theme or to override a parent theme in its children.
```

src/components/ThemeProvider/README.md
```diff
@@ -0,0 +1,230 @@
...
+
+---
+
+## Examples
+
+### Theme provider rendered by the app provider
+
+The app provider component renders a theme provider component and a theme. By default, theme falls back to the current design language. The new design language is enabled by passing `{newDesignLanguage: true}` to the `features` prop on the app provider component.
```
sadiesaurus: Wondering if we might want to move away from using "current" and "new" when talking about design language. Could this be confusing and will it need to be changed in the future?

src/components/ThemeProvider/README.md
```diff
@@ -0,0 +1,230 @@
...
+        <List.Item>1 × Anubis Cup, 2-Pack</List.Item>
+      </List>
+    </Card.Section>
+  </Card>
+</AppProvider>
+```
+
+### Theme provider with theme colors rendered by the app provider
+
+A custom theme can be passed to the theme prop on the app provider to override the default theme colors at a global level. Individual color roles can be overridden, or all of them.
```
sadiesaurus: Does this work?

```suggestion
A custom theme can be passed to the theme prop on the app provider to override one or more of the default theme colors at a global level. 
```

src/components/ThemeProvider/README.md
```diff
@@ -0,0 +1,230 @@
...
+    </Card.Section>
+  </Card>
+</AppProvider>
+```
+
+---
+
+## Consuming theme colors in a component
+
+The theme provider component uses [CSS Custom Properties](https://developer.mozilla.org/en-US/docs/Web/CSS/--*) to share color values with components. For a full list of available CSS Custom Properties, see [these docs](link/to/polaris/tokens/docs).
```
sadiesaurus:
```suggestion
The theme provider component uses [CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/--*) to share color values with components. For a full list of available CSS custom properties, see the [Polaris tokens docs](link/to/polaris/tokens/docs).
```

src/components/ThemeProvider/README.md
```diff
@@ -0,0 +1,230 @@
...
+
+We apply the current design language value to the fallback value of the `var()` function in order to gracefull fall back when `{newDesignLanguage: false}`, as the theme CSS Custom Properties will not be defined.
```
sadiesaurus:
```suggestion
We apply the current design language value to the fallback value of the `var()` function in order to gracefully fall back when `{newDesignLanguage: false}`, as the theme CSS custom properties will not be defined.
```

* Same question here about "current" vs "new" design language

src/components/ThemeProvider/README.md
```diff
@@ -0,0 +1,230 @@
...
+.Card {
+  background-color: var(--p-surface, color('white'));
+  box-shadow: var(--p-card-shadow, shadow());
+  border-radius: var(--p-border-radius-wide, border-radius());
+}
+```
+
+### Avoid using `linear-gradient()` as a `var()` fallback
+
+Using the `linear-gradient()` function as a fallback to the `var()` function causes infinite re-paints in Safari, and should be avoided. As an alternative, use the value provided to `newDesignLanguage` to create a class name for your component that is only present when that value is `true`.
```
sadiesaurus:
```suggestion
Using the `linear-gradient()` function as a fallback to the `var()` function causes infinite re-paints in Safari, and should be avoided. As an alternative, use the value provided to `newDesignLanguage` to create a class name for your component that's only present when that value is `true`.
```

src/components/ThemeProvider/README.md
```diff
@@ -0,0 +1,230 @@
...
+.Button.newDesignLanguage {
+  background-color: var(--p-action-primary);
+}
+```
+
+---
+
+## Creating a component with an inverse color scheme
+
+Some components render themselves with a dark color scheme in a light context, or a light color scheme in a dark context, in order to create contrast with surrounding elements. The theme provider component enables this behavior when `'inverse'` is passed as a color scheme to a nested theme provider wrapping a component.
```
sadiesaurus:
```suggestion
To create contrast with surrounding elements, some components render themselves with a dark color scheme in a light context or a light color scheme in a dark context. The theme provider component enables this behavior when `'inverse'` is passed as a color scheme to a nested theme provider wrapping a component.
```

src/components/ThemeProvider/README.md
```diff
@@ -0,0 +1,230 @@
...
+    <div className="InverseComponent">Component content</div>
+  </ThemeProvider>
+);
+```
+
+---
+
+## Related components
+
+- To share global settings throughout the hierarchy of your application, [use the app provider component](https://polaris.shopify.com/components/structure/app-provider)
```
sadiesaurus: ❤️❤️❤️

src/components/ThemeProvider/README.md
danrosenthal: This feels like a temporary problem, as this content pretty much goes away when the new design language *is* the design language. My vote is this stays as-is for now, but I am open to any language that can add some clarity.

amrocha:
There was a Github outage earlier and some weird race condition on Github's end happened, that's why there's two splash comments above. I went ahead and deleted the stale one, sorry for the confusion friends

src/components/ThemeProvider/README.md
sadiesaurus: Fair point. What if we keep it but change "current" to "old" to be sure that folks don't mistake "current" for "new"?

src/components/ThemeProvider/README.md
```diff
@@ -0,0 +1,230 @@
...
+}
+```
+
+### Avoid using `linear-gradient()` as a `var()` fallback
+
+Using the `linear-gradient()` function as a fallback to the `var()` function causes infinite re-paints in Safari, and should be avoided. As an alternative, use the value provided to `newDesignLanguage` to create a class name for your component that's only present when that value is `true`.
+
+```scss
+.Button {
+  background-color: linear-gradient(
```
dleroux: `linear-gradient` should be added to `background-image` as oppose to `background-color`

danrosenthal:
This is ready to ship, but we'll need to make sure that the style guide does not wrap these examples in an app provider, as it will break the examples.

src/components/ThemeProvider/README.md
```diff
@@ -0,0 +1,232 @@
...
+category: Structure
+keywords:
+  - theme
+  - provider
+  - color
+  - logo
+  - color scheme
+  - light mode
+  - dark mode
+omitAppProvider: true
```
tmlayton: This is the secret sauce for turning off the wrapping `AppProvider` in Storybook and the Style Guide

src/components/ThemeProvider/README.md
```diff
@@ -0,0 +1,232 @@
...
+    </Card.Section>
+  </Card>
+</AppProvider>
+```
+
+---
+
+## Consuming theme colors in a component
+
+The theme provider component uses [CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/--*) to share color values with components. For a full list of available CSS custom properties, see the [Polaris tokens docs](link/to/polaris/tokens/docs).
```
tmlayton: Just noticed this link is broken which causes the Style Guide build to fail (link checker)—might require a patch release. For now this can point to https://github.com/Shopify/polaris-react/blob/master/documentation/Color%20system.md.

PR #2759, opened 2020-02-19
[Frame, Navigation, AppProvider] elevate nested components and utils 

dahukish:
We are currently deep linking into the `esnext` folder for certain parts of the `Frame`, `Navigation` and `AppProvider`. Trying to create a new path that lets us expose subcomponents and utils that are needed to replicate top level components that are to be used with Polaris.

Fixes #0000 

moves sub components and utils to a level that can be accessed via the proper `@shopify/polaris` import path.

```
export {I18n, I18nContext} from './utilities/i18n';
export {useToggle} from './utilities/use-toggle';
export {navigationBarCollapsed} from './utilities/breakpoints';
export {FrameContext, ToastID, ToastPropsWithID} from './utilities/frame';
export {setRootProperty} from './utilities/set-root-property';
export {useMediaQuery} from './utilities/media-query';

export {layer, dataPolarisTopBar} from './components/shared';

export {ToastManager, CSSAnimation} from './components/Frame/components';
export {NavigationContext} from './components/Navigation/context';
export {Secondary} from './components/Navigation/components/Item/components';
```

```
/* Atomic exports */
export * from './atomic';
```

### DIFF
--- a/src/atomic.ts
+++ b/src/atomic.ts
@@ -1,0 +1,12 @@
+export {I18n, I18nContext} from './utilities/i18n';
+export {useToggle} from './utilities/use-toggle';
+export {navigationBarCollapsed} from './utilities/breakpoints';
+export {FrameContext, ToastID, ToastPropsWithID} from './utilities/frame';
+export {setRootProperty} from './utilities/set-root-property';
+export {useMediaQuery} from './utilities/media-query';
+
+export {layer, dataPolarisTopBar} from './components/shared';
+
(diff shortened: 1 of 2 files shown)


dahukish:
here is where they are being used in a shitty way: https://github.com/Shopify/global-nav/tree/master/src/components/Hacks

dahukish:
going to close this in favour of making a PRs for each component in question. 🙏

PR #2761, opened 2020-02-20
[Navigation.Section] added `itemsPrefix` prop

dahukish:

enables the ability to do this:

image

Fixes #0000 

adds a new prop that either renders the passed markup or does nothing. All styling etc is handled outside of Polaris and should be non intrusive to the api of `Section`

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Navigation/components/Section/Section.tsx
+++ b/src/components/Navigation/components/Section/Section.tsx
@@ -29,6 +29,7 @@ export interface SectionProps {
     accessibilityLabel: string;
     onClick(): void;
   };
+  itemsPrefix?: React.ReactNode;
   separator?: boolean;
 }
 
@@ -38,6 +39,7 @@ export function Section({
   action,
   items,
   rollup,
+  itemsPrefix,
   separator,
 }: SectionProps) {
   const {
@@ -170,9 +172,12 @@ export function Section({
     </li>
   );
 
+  const itemsPrefixMarkup = itemsPrefix ? <li>{itemsPrefix}</li> : null;
+
   return (
     <ul className={className}>
       {sectionHeadingMarkup}
+      {itemsPrefixMarkup}
       {sectionItems}
       {activeItemsMarkup}
     </ul>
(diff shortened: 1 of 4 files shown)


BPScott:
So I'm understanding this right: the "do this" in that image is "adding a search field into the navigation at the top of a section between the section heading and the items within the section?

src/components/Navigation/components/Section/Section.tsx
```diff
@@ -170,9 +172,12 @@ export function Section({
     </li>
   );
 
+  const itemsPrefixMarkup = itemsPrefix || null;
```
BPScott: This should be wrapped in an `<li>` if it exists, otherwise we'll end up outputting invalid html (e.g. `<ul><div>blah</div></ul>` is not valid as you can only have lis inside a ul

src/components/Navigation/components/Section/tests/Section.test.tsx
```diff
@@ -253,6 +253,35 @@ describe('<Navigation.Section />', () => {
...
+            disabled: false,
+          },
+        ]}
+        action={{
+          icon: 'placeholder',
+          accessibilityLabel: 'This is a test section',
+          onClick: noop,
+        }}
+        itemsPrefix={
+          <div testID="ItemsPrefix">
```
BPScott: Try and avoid testID where possible, searching based upon a class name for this sort of thing should be fine

dahukish: 👍

src/components/Navigation/components/Section/Section.tsx
dahukish: yes that is very true. I did have an <li /> in ours but was styling that. I will add it back in for validity.

dahukish:
> So I'm understanding this right: the "do this" in that image is "adding a search field into the navigation at the top of a section between the section heading and the items within the section?

Yes we are filtering the stores we pass as `items` in a section.

dahukish:
going to pause this for now as this is blocking other work on global nav. 

next steps: allow passing of ItemConfig to items which contains {preFix, suffix and Items}.

alex-page:
@dahukish going to close but feel free to reopen.

PR #2763, opened 2020-02-20
[Color system] update to latest release of polaris tokens

danrosenthal:


src/utilities/theme/utils.ts
```diff
@@ -1,5 +1,5 @@
 import tokens from '@shopify/polaris-tokens';
-import {colorFactory} from '@shopify/polaris-tokens/color-factory';
+import {colorFactory} from '@shopify/polaris-tokens/dist-modern/color-factory';
```
BPScott: the index in the dist-modern folder reexports the color factory. We should treat that index as the new entry point rather than diving into specific files

```suggestion
import {colorFactory} from '@shopify/polaris-tokens/dist-modern';
```

danrosenthal:
Seeing this in my editor, but `yarn type-check` passes locally, and CI is 🍏.

|vscode?|
|-|
|Screen Shot 2020-02-20 at 5 09 39 PM|

Thoughts @BPScott?

BPScott:
VSCode seems happy to me when looking at that file.

Restart your editor and make sure you're using the ts version that's installed in the project (v3.7.3 - cmd+shift+p and search for "select typescript version") perhaps?

PR #2766, opened 2020-02-21
Update import paths

tmlayton:
This fixes the color docs script. Format change to the select CSS was committed as part of the pre commit hook


scripts/color-system-docs.js
```diff
@@ -1,8 +1,10 @@
 const {resolve: resolvePath} = require('path');
 const {writeFileSync} = require('fs-extra');
-const roleVariants = require('@shopify/polaris-tokens/formats/utils/color-factory/configs/base');
+const {
+  config: roleVariants,
```
kyledurand: Ah, this is where I was getting hung up

tmlayton: yeah this is weird, was not expecting this to change with the TS build

PR #2770, opened 2020-02-24
♻️[Spinner] Do not render <span /> when there is no label

beefchimi:
Just a small thing I noticed while using this component. `accessibilityLabel` is optional - but we are rendering the markup for it regardless. This PR changes the markup to only output the `span + VisuallyHidden` when `accessbilityLabel` is present.

It's possible that maybe the intention was to always render this empty `<span />` so that you can at least toggle the `role="status"` prop. If so, I feel that it would be better to replace the `React.Fragment` with that `span` so we wrap the entire component. Perhaps someone can enlighten me on the best practice for this `role` markup?

### DIFF
--- a/src/components/Spinner/Spinner.tsx
+++ b/src/components/Spinner/Spinner.tsx
@@ -63,14 +63,12 @@ export function Spinner({
 
   const spinnerSVG = size === 'large' ? spinnerLarge : spinnerSmall;
 
-  const spanAttributes = {
-    ...(!hasFocusableParent && {role: 'status'}),
-  };
-
-  const accessibilityLabelMarkup = (isAfterInitialMount ||
-    !hasFocusableParent) && (
-    <VisuallyHidden>{accessibilityLabel}</VisuallyHidden>
-  );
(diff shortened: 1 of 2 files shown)


dleroux:
@AndrewMusgrave looks like you had made changes to this during the a11y project. Any chance you did it because the spinner needed a container on screen in order to be announced by NVDA?

AndrewMusgrave:
@dleroux  I'm sure not about NVDA specifically (although it's likely), however, certain screen readers need elements pre-rendered on the page to correctly announce things. I know this is often the case for jaws. Iirc this was why the span is always rendered

dleroux:
@beefchimi is this span causing you grief? If so you might want to check using NVDA on windows to ensure we are introducing a11y issues by removing it.

dleroux:
I'm going to close this because I confirmed that it is an accessibility issue. A div with a role="status" acts as a live region and it needs to be present in the dom for its content changes to be announced properly. If this is a problem Curtis bring it up to the team and we can discuss.

PR #2777, opened 2020-02-26
[OptionList] Export OptionListProps

shanesmith:

Wanting to reference the type of an OptionList prop we noticed that OptionListProps wasn't being exported. Looking at `src/components/index.ts` it's seemingly just an omission since the props type of other components are exported.

`OptionsListProps` is added to the exports.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -26,6 +26,7 @@
 - Added first implementation of custom property validation ([#2616](https://github.com/Shopify/polaris-react/pull/2616))
 - Refactored consumer build test (renamed to system integration test) ([#2735](https://github.com/Shopify/polaris-react/pull/2735))
 - Added Storybook Knobs for customizing theme ([#2674](https://github.com/Shopify/polaris-react/pull/2674))
+- Export missing OptionListProps ([#2777](https://github.com/Shopify/polaris-react/pull/2777))
 
 ### Dependency upgrades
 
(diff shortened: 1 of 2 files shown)


shanesmith:
@BPScott I've included you as a reviewer since I've noticed you've done some recent work in how things are exported (eg: #2625, #2058..) =)

UNRELEASED.md
```diff
@@ -26,6 +26,7 @@
 - Added first implementation of custom property validation ([#2616](https://github.com/Shopify/polaris-react/pull/2616))
 - Refactored consumer build test (renamed to system integration test) ([#2735](https://github.com/Shopify/polaris-react/pull/2735))
 - Added Storybook Knobs for customizing theme ([#2674](https://github.com/Shopify/polaris-react/pull/2674))
+- Export missing OptionListProps ([#2777](https://github.com/Shopify/polaris-react/pull/2777))
```
BPScott: To match the tense of the other lines in the changelog:

```suggestion
- Exported missing OptionListProps ([#2777](https://github.com/Shopify/polaris-react/pull/2777))
```

shanesmith:
Thanks for the quick review and suggestion! I've updated the PR as requested.

What shall we do about the failing Percy test? I really doubt my change is significant enough to have caused the failure, but I haven't used Percy before so I'm not to sure what to do...

BPScott:
The percy change is a little odd - it's one of the components that used to fail occasionally due to animation changes - the length of the loading bar is loader or shorter depending on how long the page took to load. It looks like the "new" version is what it should be so I dunno what was going on before. 

I've approved the change, and you're good to merge.

shanesmith:
Great, thanks for looking into it, merging away!

PR #2782, opened 2020-02-26
[Heading] Prevent orphans on strings

dfmcphee:

Currently we end up with orphan words when our typography components wrap.

This introduces a new helper to add a `&nbsp;` in between the last two words of a string and then uses that helper in our `Heading` component so that a single word never wraps on its own.

auto-remove-orphans

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Heading/Heading.tsx
+++ b/src/components/Heading/Heading.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import {HeadingTagName} from '../../types';
+import {preventOrphans} from '../../utilities/string';
 import styles from './Heading.scss';
 
 export interface HeadingProps {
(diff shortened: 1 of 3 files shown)


dfmcphee:
One of the questions I have around this is with testing. I don't really want every consumer to have to test against having a `&nbsp;` in a string. Any ideas for making sure consumers won't need to update anything to keep their tests passing?

BPScott:
What happens when the last two words are both long to the point that when combined they wouldn't fit on a single line on smaller screens?

I'm thinking for languages that have longer single words possibly due to a love of compound words like German. e.g. ["Verwerfen Sie alle nicht gespeicherten Änderungen"](https://github.com/Shopify/polaris-react/blob/master/locales/de.json#L77). In that case you'd end up with horizontal scrolling for `gespeicherten Änderungen` when those words individually could fit onto 2 lines without wrapping

dfmcphee:
> What happens when the last two words are both long to the point that when combined they wouldn't fit on a single line on smaller screens?

That's a good question. I guess one potential solution to that would be having a threshold of something like less than 10 characters for inserting.

PR #2788, opened 2020-02-27
[Color system] Upgrade tokens

danrosenthal:


danrosenthal:
Looks like I need to fix tests that are asserting previous color values

PR #2789, opened 2020-02-27
[Navigation] Apply design language 

kyledurand:

Addresses most of: https://github.com/Shopify/polaris-ux/issues/372 minus the **important** note below (There's a separate issue for rails)

#### Important
This PR does not solve the issue of `Collapsible` not animating. That will take some extra effort in debugging and may open a can of worms with that component so I'm going to follow up with that issue.

DL_Nav

Dark mode:
image

High contrast mode:
image

## 

### How to 🎩

Visit the [details page](http://localhost:6006/iframe.html?id=playground-playground--details-page&contexts=New%20Design%20Language=Enabled%20-%20Light%20Mode#) in storybook

### 🎩 checklist

### DIFF
--- a/playground/DetailsPage.tsx
+++ b/playground/DetailsPage.tsx
@@ -55,6 +55,7 @@ export function DetailsPage() {
   const [userMenuActive, setUserMenuActive] = useState(false);
   const [mobileNavigationActive, setMobileNavigationActive] = useState(false);
   const [modalActive, setModalActive] = useState(false);
+  const [navItemActive, setNavItemActive] = useState('');
   const [nameFieldValue, setNameFieldValue] = useState(
     defaultState.current.nameFieldValue,
   );
@@ -210,7 +211,6 @@ export function DetailsPage() {
       contextControl={contextControlMarkup}
     />
   );
-
   // ---- Navigation ----
   const navigationMarkup = (
     <Navigation location="/" contextControl={contextControlMarkup}>
@@ -219,42 +219,103 @@ export function DetailsPage() {
           {
             label: 'Home',
             icon: HomeMajorTwotone,
-            onClick: toggleIsLoading,
+            onClick: () => {
+              toggleIsLoading();
+              setNavItemActive('home');
+            },
+            matches: navItemActive === 'home',
+            url: '#',
           },
           {
             label: 'Orders',
             icon: OrdersMajorTwotone,
-            onClick: toggleIsLoading,
+            onClick: () => {
+              toggleIsLoading();
+              setNavItemActive('orders');
+            },
+            matches: navItemActive === 'orders',
+            url: '#',
           },
           {
             label: 'Products',
             icon: ProductsMajorTwotone,
-            onClick: toggleIsLoading,
+            onClick: () => {
+              toggleIsLoading();
+              setNavItemActive('products');
+            },
+            matches: navItemActive === 'products',
+            url: '#',
+            subNavigationItems: [
+              {
+                label: 'All products',
+                onClick: () => {
+                  toggleIsLoading();
+                  setNavItemActive('all-products');
+                },
+                matches: navItemActive.includes('products'),
+                url: '#',
+              },
+              {
+                url: '#',
+                label: 'Drafts',
+
+                onClick: () => {
+                  toggleIsLoading();
+                  setNavItemActive('drafts');
+                },
+                matches: navItemActive === 'drafts',
+              },
+            ],
           },
           {
             label: 'Customers',
             icon: CustomersMajorTwotone,
-            onClick: toggleIsLoading,
+            onClick: () => {
+              toggleIsLoading();
+              setNavItemActive('customers');
+            },
+            matches: navItemActive === 'customers',
+            url: '#',
           },
           {
             label: 'Analytics',
             icon: AnalyticsMajorTwotone,
-            onClick: toggleIsLoading,
+            onClick: () => {
+              toggleIsLoading();
+              setNavItemActive('analytics');
+            },
+            matches: navItemActive === 'analytics',
+            url: '#',
           },
           {
             label: 'Marketing',
             icon: MarketingMajorTwotone,
-            onClick: toggleIsLoading,
+            onClick: () => {
+              toggleIsLoading();
+              setNavItemActive('marketing');
+            },
+            matches: navItemActive === 'marketing',
+            url: '#',
           },
           {
             label: 'Discounts',
             icon: DiscountsMajorTwotone,
-            onClick: toggleIsLoading,
+            onClick: () => {
+              toggleIsLoading();
+              setNavItemActive('discounts');
+            },
+            matches: navItemActive === 'discounts',
+            url: '#',
           },
           {
             label: 'Apps',
             icon: AppsMajorTwotone,
-            onClick: toggleIsLoading,
+            onClick: () => {
+              toggleIsLoading();
+              setNavItemActive('apps');
+            },
+            matches: navItemActive === 'apps',
+            url: '#',
           },
         ]}
       />
--- a/src/components/Navigation/Navigation.scss
+++ b/src/components/Navigation/Navigation.scss
@@ -2,6 +2,7 @@
 @import './variables';
 
 // Base Navigation styles
+$active-indicator: rem(3px);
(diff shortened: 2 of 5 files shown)


src/components/Select/Select.scss
```diff
@@ -148,6 +148,10 @@ $stacking-order: (
 
 .newDesignLanguage {
   .Backdrop {
+    // 'position' needs to sit below focus-ring since it will be overwritten
+    // with relative when the focus ring style is 'base'
+    // stylelint-disable-next-line order/properties-order
+    position: absolute;
```
danrosenthal: `yarn format` keeps making this change, which breaks the focus ring. You'll need to drop this change.

playground/DetailsPage.tsx
```diff
@@ -55,6 +55,7 @@ export function DetailsPage() {
   const [userMenuActive, setUserMenuActive] = useState(false);
   const [mobileNavigationActive, setMobileNavigationActive] = useState(false);
   const [modalActive, setModalActive] = useState(false);
+  const [navItemActive, setNavItemActive] = useState('');
```
kyledurand: Adding state so we can see active items as well as sub items

src/components/Navigation/Navigation.scss
```diff
@@ -84,6 +85,24 @@ $disabled-fade: 0.6;
...
+      top: 0;
+      bottom: 0;
+      left: -8px;
+      height: 100%;
+      width: $active-indicator;
+      background-color: var(--p-action-primary);
+      border-top-right-radius: var(--p-border-radius-base);
+      border-bottom-right-radius: var(--p-border-radius-base);
+      transform: translateX(-100%);
+      transition: transform var(--p-duration-1-5-0) var(--p-ease);
```
kyledurand: This `before` is the green indicator left of active nav items

src/components/Navigation/Navigation.scss
```diff
@@ -93,8 +112,15 @@ $disabled-fade: 0.6;
...
+
+  .Navigation-newDesignLanguage & {
+    &::before {
+      transform: translateX(0);
+    }
+  }
+
   @media (-ms-high-contrast: active) {
-    background-color: ms-high-contrast-color('selected-text-background');
+    outline: 1px solid ms-high-contrast-color('text');
```
kyledurand: This wasn't working before. As @dleroux mentioned yesterday, chromium edge hijacks background color completely. I've been defaulting to outlines for buttons / tabs / checkboxes etc...

src/components/Navigation/Navigation.scss
```diff
@@ -257,23 +285,39 @@ $secondary-item-font-size: rem(15px);
   margin-bottom: spacing(tight);
   margin-left: nav(icon-size) + spacing(loose);
   overflow-x: var(--p-override-visible, hidden);
+
+  .Navigation-newDesignLanguage & {
+    margin-left: 0;
```
kyledurand: We sub nav items to be full width now

image

src/components/Navigation/Navigation.scss
```diff
@@ -301,7 +346,13 @@ $secondary-item-font-size: rem(15px);
       margin-top: spacing(extra-tight);
       margin-bottom: spacing(extra-tight);
     }
+    // stylelint-disable-next-line selector-max-class, selector-max-combinators
+    .Navigation-newDesignLanguage & {
+      margin-top: spacing(tight);
+      margin-bottom: spacing(tight);
```
kyledurand: Slightly different spacing on nav items

src/components/Navigation/_variables.scss
```diff
@@ -133,6 +149,10 @@ $nav-animation-variables: (
     margin-right: spacing();
     margin-bottom: nav(desktop-spacing);
   }
+  .Navigation-newDesignLanguage & {
+    margin-top: spacing(tight);
+    margin-bottom: spacing(tight);
```
kyledurand: Same spacing change as above to keep the icon centrered

src/components/Navigation/_variables.scss
```diff
@@ -158,6 +178,11 @@ $nav-animation-variables: (
   .SecondaryNavigation & {
     opacity: 0;
     animation: fade-in duration(slowest) 1 forwards;
+    // stylelint-disable-next-line selector-max-combinators
+    .Navigation-newDesignLanguage & {
+      opacity: 1;
+      animation: none;
```
kyledurand: This removes the fade in animation for sub nav items

src/components/Navigation/components/Item/Item.tsx
```diff
@@ -266,6 +266,7 @@ export function Item({
   function getClickHandler(onClick: ItemProps['onClick']) {
     return (event: MouseEvent<HTMLElement>) => {
       const {currentTarget} = event;
+      currentTarget.blur();
```
kyledurand: Blurring to remove the focus ring for nav items on click. Tab order remains so clicking and then tabbing will still bring focus to the next nav item

src/components/Navigation/components/Item/tests/Item.test.tsx
```diff
@@ -45,6 +45,7 @@ describe('<Nav.Item />', () => {
     trigger(item.find(UnstyledLink).first(), 'onClick', {
       preventDefault: jest.fn(),
       currentTarget: {
+        blur: jest.fn(),
```
kyledurand: Mocking the blur func for currentTarget so tests will pass

src/components/Select/Select.scss
kyledurand: I'll fix this up in a sec

src/components/Navigation/_variables.scss
```diff
@@ -41,32 +42,36 @@ $nav-animation-variables: (
...
   }
 
   @include focus-ring;
 
   &:focus {
     @include focus-ring($style: 'focused');
-    background-image: var(
-      --p-override-none,
-      linear-gradient(color(state, hover), color(state, hover))
-    );
```
danrosenthal: Good catch. I wonder if this was causing the same Safari problems in production and we just never noticed...

src/components/Navigation/components/Item/Item.tsx
danrosenthal: This breaks keyboard navigation, removing focus from navigation items when the `enter` key is pressed. This'll have to go. It also breaks in the old design language, as this change isn't scoped to the new DL.

Why not apply the focus ring using the `:focus:not(:active)` selector instead? That should solve the problem without breaking accessibility. If not, we're going to have to live with the ring

|😢 |
|-|
|broken-keyboard-navigation|

src/components/Navigation/Navigation.scss
```diff
@@ -84,6 +85,24 @@ $disabled-fade: 0.6;
...
   @include nav-item-attributes;
+
+  .Navigation-newDesignLanguage & {
+    position: relative;
+    &::before {
+      content: '';
+      position: absolute;
+      top: 0;
+      bottom: 0;
+      left: -8px;
```
dleroux:
```suggestion
     left: -1 * spacing(tight);
```

src/components/Navigation/Navigation.scss
dleroux: I think you might be able to remove this altogether and just leave the default?

src/components/Navigation/components/Item/Item.tsx
dleroux: If that doesn't work, a keyboard focused state like used in checkbox will.

kyledurand:
I can't seem to remove the focus ring using css. I've removed blur for now so maybe we can ship this as is and address that when I'm back or in the design review phase @danrosenthal

dleroux:
Why not just use state like I did with checkbox?

tmlayton:
A couple things:

- Sub nav items currently have a height of 36px but should be 32px
- Animation of the `before` indicator appears to not display when selecting sub nav items (perhaps related to the collapsible animation)

And can you test with really long strings both for parent and child nav items? I recall messing with layout styles in my first bit of work on this and want to make sure I didn't break something.

danrosenthal:
> And can you test with really long strings both for parent and child nav items?

There are no regressions with long navigation items (existing styles don't handle them very well though).

PR #2791, opened 2020-02-27
[Button] Nest split button example in tall enough <div>

sambostock:
The example currently does not render well when the button dropdown is clicked, as it renders past the boundaries of the host div, and is obscured. Due to this, it appears that the button is broken and does nothing.

This follows the pattern in <Autocomplete />'s README and hosts the example in a div with enough height.

|Before|After|
|-|-|
|image|image|

### 🎩 checklist

### DIFF
--- a/src/components/Button/README.md
+++ b/src/components/Button/README.md
@@ -321,15 +321,17 @@ function DisclosureButtion() {
 Use when there is only one primary action but other related actions can be taken.
 
 ```jsx
-<Button
-  primary
-  connectedDisclosure={{
-    accessibilityLabel: 'Other save actions',
-    actions: [{content: 'Save as draft'}],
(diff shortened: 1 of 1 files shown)


UNRELEASED.md
```diff
@@ -13,6 +13,8 @@
 
 ### Documentation
 
+- Improved split button example ([#2791](https://github.com/Shopify/polaris-react/pull/2791))
```
kyledurand: I don’t think we need this added to the change log since it’s not a change to a component. You can just use the skip change log label in this pr to get that check to pass.

sambostock:
@kyledurand Thanks for the tip; I didn't know about the label.

I've added the label, and removed the changelog entry.

PR #2793, opened 2020-02-27
Improve build-consumer.js' robustness

sambostock:

If a developer is following the instructions in the ["Tophatting Documentation" guide](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting%20documentation.md), but has never run `dev up` on `Shopify/polaris-styleguide`, an error will occur, but the build script will continue as if everything is fine.

```
Cleaning up old build...
Creating new build directory...
mkdir: no such file or directory: /Users/sambostock/src/github.com/Shopify/polaris-styleguide/node_modules/@shopify
Copying build to node_modules...
cp: dest is not a directory (too many sources)
Build copied to consuming project. You can now run the consuming app and it will include your changes from Polaris.
✨  Done in 50.34s.
```

This enables the `config.fatal` setting for `shelljs`, so errors are noisy, and then fixes the `mkdir` error by adding the `-p` flag.

### 🎩 checklist

### DIFF
--- a/scripts/build-consumer.js
+++ b/scripts/build-consumer.js
@@ -1,12 +1,14 @@
 /* eslint-disable no-console */
 
 const {resolve} = require('path');
-const {cp, mkdir, rm} = require('shelljs');
+const {config, cp, mkdir, rm} = require('shelljs');
 
 const packageJSON = require('../package.json');
 
 const root = resolve(__dirname, '..');
 const projectDir = process.argv[2];
+
+config.fatal = true;
 
 if (!projectDir) {
   console.log(
--- a/scripts/build-consumer.js
+++ b/scripts/build-consumer.js
@@ -33,7 +33,7 @@ console.log('Cleaning up old build...');
 rm('-rf', projectPolarisDir);
 
 console.log('Creating new build directory...');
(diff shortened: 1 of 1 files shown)


BPScott:
Sounds legit. Longer term, I wonder if we could align on using https://github.com/Shopify/tophat-web for our tophatting-in-other-repos needs instead of having to maintain this ourselves (possibly something @AndrewMusgrave may be interested in)

AndrewMusgrave:
> Sounds legit. Longer term, I wonder if we could align on using https://github.com/Shopify/tophat-web for our tophatting-in-other-repos needs instead of having to maintain this ourselves (possibly something @AndrewMusgrave may be interested in)

It would be awesome if we could have a standard rather than something custom. Unfortunately with the changes moving forward, I might not be able to dedicate time to this. We'll have to see 🤞

sambostock:
I've decided to remove the entry in `UNRELEASED.md`, since it didn't really seem relevant to consumers. It previously said

```markdown
### Code quality

- Improve scripts/build-consumer.js' robustness ([#2793](https://github.com/Shopify/polaris-react/pull/2793))
```

PR #2803, opened 2020-03-02
Fix UNRELEASED

devonpmack:
Not sure what happened but somehow I merged the same message 3 times


devonpmack:
thanks that makes sense. In the future I will re-read the diff after rebasing

Issue #2662, opened 2020-01-24
[Proposal] Remove overflow:hidden on the Card component


dfmcphee:
## Issue summary

Currently we apply `overflow: hidden` to the Card component. This raises problems when you need to use something like `position: sticky` within the Card, because it can't be used on elements with overflow.

We apply this to make sure children get the correct border radius to match the card itself. The most common example is for subdued sections at the end of a card.

I propose we remove the overflow hidden and add these rules to Card.Sections:

```scss
border-bottom-left-radius: border-radius();
border-bottom-right-radius: border-radius();
```

This will solve the majority of use cases and then if any other component needs to stretch the full width/height of a card, they apply the border radius.

The current blanket statement to remove overflow feels heavy handed for those few unique use cases.

Thoughts? Opinions?

dpersing:
NUKE IT! 🔥

beefchimi:
FYI We do exactly this in our "Card" component:
https://github.com/Shopify/online-store-ui/blob/master/src/components/SidebarCard/SidebarCard.scss#L21-L33

Happy to send you a link to see it on our Chroma instance if you'd like.

AndrewMusgrave:
I was actually mentioning the problem with `overflow: hidden` & `position: sticky`  to Dan this morning when we were running into issues with Card and the new `focus-ring`. 100% for removing this.

Also as a note, we can't use `position: sticky` in the resource list header exactly because of overflow hidden in cards 😅

perrupa:
## Samesies
I ran into an issue with this regarding non-polaris tooltips inside a Card.

## My problem
I'm trying to implement a rich-text editor that uses popups and the Card's overflow is constraining the tooltip's contents.

## Solutioning
It would be very handy to override this prop, or maybe only have it on cards that are sectioned? Assuming that's what it's _needed_ for.

## Next Steps
What's blocking us here, lack of a PR to push over the line?

Would @beefchimi 's proposed solution be enough to get rid of the `overflow: hidden` entirely? I feel like it might, unless you were relying on Card having `overflow: hidden` for other behaviours (we shouldn't support those cases).

The worry-wort in me thinks this might necessitate a breaking change, but counter-point is the API hasn't changed just possible side-effects....

Thoughts?

cc// @dfmcphee @AndrewMusgrave @dpersing 🎺

dfmcphee:
I just put up a PR for this at #2806

BPScott:
[...]

PR #2806, opened 2020-03-03
[Card] Removes overflow hidden

dfmcphee:

This addresses #2662. 

It changes the `overflow` on the Card component to `visible` by default.

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page, Layout, Card, List} from '../src';

  return (
    <Page title="Playground">
      <Layout>
        <Layout.Section>
          <Card title="Staff accounts">
            <Card.Section>
              <List>
                <List.Item>Felix Crafford</List.Item>
                <List.Item>Ezequiel Manno</List.Item>
              </List>
            </Card.Section>

            <Card.Section subdued title="Deactivated staff accounts">
              <List>
                <List.Item>Felix Crafford</List.Item>
                <List.Item>Ezequiel Manno</List.Item>
              </List>
            </Card.Section>

            <Card.Section>
              <List>
                <List.Item>Felix Crafford</List.Item>
                <List.Item>Ezequiel Manno</List.Item>
              </List>
            </Card.Section>

            <Card.Section>
              <List>
                <List.Item>Felix Crafford</List.Item>
                <List.Item>Ezequiel Manno</List.Item>
              </List>
            </Card.Section>

            <Card.Section subdued title="Deactivated staff accounts">
              <List>
                <List.Item>Felix Crafford</List.Item>
                <List.Item>Ezequiel Manno</List.Item>
              </List>
            </Card.Section>
          </Card>
        </Layout.Section>
      </Layout>
    </Page>
  );
}


```

</details>

### 🎩 checklist

* [NA] Updated the component's `README.md` with documentation changes
* [NA] [Tophatted documentation](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting%20documentation.md) changes in the style guide
* [NA] For visual design changes, pinged one of @ HYPD, @ mirualves, @ sarahill, or @ ry5n to update the Polaris UI kit

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -5,7 +5,7 @@
 ### Enhancements
 
 - Added high contrast outline to `Popover`, `Card` and `Indicator` ([#2792](https://github.com/Shopify/polaris-react/pull/2792))
-- Changes overflow of `Card` to `visible` ([#2806](https://github.com/Shopify/polaris-react/pull/2806))
+- Removed `overflow: hidden` from `Card` ([#2806](https://github.com/Shopify/polaris-react/pull/2806))
 
 ### Bug fixes
 
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -5,6 +5,7 @@
 ### Enhancements
 
 - Added high contrast outline to `Popover`, `Card` and `Indicator` ([#2792](https://github.com/Shopify/polaris-react/pull/2792))
+- Changes overflow of `Card` to `visible` ([#2806](https://github.com/Shopify/polaris-react/pull/2806))
 
 ### Bug fixes
 
--- a/src/components/Card/Card.scss
+++ b/src/components/Card/Card.scss
@@ -1,7 +1,7 @@
 @import '../../styles/common';
 
 .Card {
-  overflow: var(--p-override-visible, hidden);
+  overflow: var(--p-override-visible, visible);
   background-color: var(--p-surface, color('white'));
   box-shadow: var(--p-card-shadow, shadow());
   outline: border-width(base) solid transparent;
@@ -29,6 +29,8 @@
 
 .Section {
   padding: spacing();
+  border-bottom-left-radius: border-radius();
+  border-bottom-right-radius: border-radius();
 
   @include page-content-when-not-fully-condensed {
     padding: spacing(loose);
--- a/src/components/Card/Card.scss
+++ b/src/components/Card/Card.scss
@@ -1,7 +1,6 @@
 @import '../../styles/common';
 
 .Card {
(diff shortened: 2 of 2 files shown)


src/components/Card/Card.scss
```diff
@@ -29,6 +29,8 @@
 
 .Section {
   padding: spacing();
+  border-bottom-left-radius: border-radius();
```
beefchimi: If a `subdued` section is the first section within a `Card`, you could end up with `top left/right` "corner-bleeding".

src/components/Card/Card.scss
```diff
@@ -1,7 +1,7 @@
 @import '../../styles/common';
 
 .Card {
-  overflow: var(--p-override-visible, hidden);
+  overflow: var(--p-override-visible, visible);
```
BPScott: Visible is the default, so we could remove this line entirely

src/components/Card/Card.scss
dfmcphee: I switched this up a bit to target first and last child and only change the border radius on those. What do you think?

beefchimi: Yep - this is essentially the same solution I used in a similar component. 🚢

BPScott: Nice. Would `.Card > :first-child` / `.Card > :last-child` help protect against cases where people put not-a-section as the first / last item (e.g. there's only a Card.Header for some weird reason) or is that not considered valid ?

beefchimi: The issue is only apparent when the child element has _something_ that could bleed out. In the case of a `Section` - its actually specifically a `Section + subdued`, as it is the darkened `background-color` that is "bleeding out".

If there are other child elements that could be `subdued`, we should cast a wide enough net to catch those as well... but I don't _think_ a `Header` or `Footer` can be `subdued` _(I could be wrong)_. Otherwise, if its just any arbitrary children that the consumer might be passing, then I'd lean towards the onus being on the consumer to style those children to suppress the border-radius bleed.

UNRELEASED.md
```diff
@@ -5,6 +5,7 @@
 ### Enhancements
 
 - Added high contrast outline to `Popover`, `Card` and `Indicator` ([#2792](https://github.com/Shopify/polaris-react/pull/2792))
+- Changes overflow of `Card` to `visible` ([#2806](https://github.com/Shopify/polaris-react/pull/2806))
```
```suggestion
- Removed `overflow: hidden` from `Card` ([#2806](https://github.com/Shopify/polaris-react/pull/2806))
```

AndrewMusgrave:
💯

martenbjork:
Potentially also affecting the Live View in Analytics.

image

PR #2811, opened 2020-03-05
Update references to border color variants

sarahill:

It was unclear when to use borderSecondary vs. the on surface border variants and there are no clear benefits to having both.

Incorporates changes from https://github.com/Shopify/polaris-tokens/pull/115

Full details in [Slack thread](https://shopify.slack.com/archives/CTGNZSQ48/p1582749879050700)

### 🎩 checklist


UNRELEASED.md
```diff
@@ -2,6 +2,8 @@
 
 ### Breaking changes
 
+- Update references to border color variants ([#2811](https://github.com/Shopify/polaris-react/pull/2811))
```
tmlayton: Let’s just remove this entry because we are not considering any of these changes breaking until after we default to the new design language and add a "skip changelog" label

sarahill:
Removed the change log entry and merged in the branch from polaris-tokens

PR #2820, opened 2020-03-10
Simplify next.js example

BPScott:

I like net negative diffs.

Since 9.2 next.js supports css imports out the box, so we don't need the
zeit/next-css package or any of its config anymore in the next.js example

Deleting things

### How to 🎩

- `cd examples/next.js`
- `yarn install && yarn server`
- open in a browser (link will be in your console), and see that the example still renders


tomgs:
Hurray for simplifying! :)

PR #2824, opened 2020-03-10
[Color system] consume new design language in spinner component

danrosenthal:
## What this PR does
- Adds a new `highlight` `color` variant for the new design language
- Aliases all existing `color` variants to `highlight` in the new design language
- Renders an `svg` element directly instead of an `img` for the new design language to enable dynamic theming (filters aren't supported in the color system)

Addresses https://github.com/Shopify/polaris-ux/issues/403

Figma: https://www.figma.com/file/4dAAt5iFPSaxUKiYVKrkYj/DL%E2%80%93Polaris-(Web)%3A-Components?node-id=7913%3A9305

## Screenshots

|old DL|new DL - light|new DL - dark|Animation|
|-|-|-|-|
|Screen Shot 2020-03-10 at 1 19 10 PM|Screen Shot 2020-03-10 at 1 19 18 PM|Screen Shot 2020-03-10 at 1 19 27 PM|spinner-animation|

## Playground

```tsx
import React from 'react';
import {Page, Stack, Heading, Spinner, TextStyle, Button} from '../src';

  return (
    <Page title="Playground">
      <Stack vertical>
        <Heading>Loading buttons</Heading>
        <Button primary loading>
          Loading
        </Button>
        <Button destructive loading>
          Loading
        </Button>
        <Button loading>Loading</Button>
        <Button loading plain>
          Loading
        </Button>
        <Heading>Old design language</Heading>
        <TextStyle variation="subdued">teal</TextStyle>
        <Spinner size="large" color="teal" />
        <Spinner size="small" color="teal" />
        <TextStyle variation="subdued">white</TextStyle>
        <Spinner size="large" color="white" />
        <Spinner size="small" color="white" />
        <TextStyle variation="subdued">ink lightest</TextStyle>
        <Spinner size="large" color="inkLightest" />
        <Spinner size="small" color="inkLightest" />
        <Heading>New design language</Heading>
        <TextStyle variation="subdued">highlight</TextStyle>
        <Spinner size="large" color="highlight" />
        <Spinner size="small" color="highlight" />
      </Stack>
    </Page>
  );
}

```


BPScott:
> Renders an svg element directly instead of an img for the new design language to enable dynamic theming (filters aren't supported in the color system)

Is there anything stopping us from using the inlined version in the existing and new styles? Might be a bit simpler

danrosenthal:
> Is there anything stopping us from using the inlined version in the existing and new styles? Might be a bit simpler

The spinner component in Polaris react uses svg as the source of an image tag, which renders as a data-encoded image. It uses filters to style the color of the image, which we are not supporting in the new design language.

To get around this issue, we will need to duplicate the svg code inside of the component. If newDesignLanguage is turned on, it will render that svg directly, remove the filter, and style the svg's color using the fill property.

BPScott:
Yeah I get that and understand why a new approach is needed for newDesignLanguage. I'm asking is it worth updating the current style to use the inline svg instead of an `Image` tag,  i.e. we always use an inline svg to avoid needing a split in logic. It seems like we'd be able to do the old style color changing with an inline SVG.

danrosenthal:
Sorry, now I understand. I'm happy to do away with the old approach.

src/components/Spinner/tests/Spinner.test.tsx
```diff
@@ -21,24 +20,44 @@ describe('<Spinner />', () => {
   describe('size', () => {
     it('renders a large spinner by default', () => {
       const spinner = mountWithAppProvider(<Spinner />);
-      expect(spinner.find(Image).hasClass('sizeLarge')).toBeTruthy();
+      expect(
```
BPScott: I'm procrastinating by noodling with react-testing. How does this look?

```
const spinner = mountWithApp(<Spinner />);
expect(spinner).toContainReactComponent('span', {
  className: expect.stringMatching(/\bsizeLarge\b/),
});
```

PR #2834, opened 2020-03-13
[Navigation] Only focus items on tab

kyledurand:

This addresses the issue where the nav items were receiving a focus state on click. Now the ring only appears on tab

[Screencast](https://screenshot.click/screencast_2020-03-12_23-19-36.mp4)

I'm setting state when the item receives a key up and the key is tab. The state adds a class which is then removed on blur.

## 

### How to 🎩

Run locally and visit the [details page](http://localhost:6006/iframe.html?id=playground-playground--details-page&contexts=New%20Design%20Language=Enabled%20-%20Light%20Mode)

### 🎩 checklist

### DIFF
--- a/src/components/Navigation/components/Item/Item.tsx
+++ b/src/components/Navigation/components/Item/Item.tsx
@@ -5,6 +5,7 @@ import React, {
   MouseEvent,
   ReactNode,
   Fragment,
+  useCallback,
 } from 'react';
 
 import {classNames} from '../../../../utilities/css';
@@ -12,7 +13,7 @@ import {classNames} from '../../../../utilities/css';
 import {NavigationContext} from '../../context';
 import {Badge} from '../../../Badge';
 import {Icon} from '../../../Icon';
-import {IconProps} from '../../../../types';
+import {IconProps, Key} from '../../../../types';
 import {Indicator} from '../../../Indicator';
 import {UnstyledLink} from '../../../UnstyledLink';
 import {useI18n} from '../../../../utilities/i18n';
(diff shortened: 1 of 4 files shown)


src/components/Navigation/components/Item/Item.tsx
```diff
@@ -87,13 +88,27 @@ export function Item({
...
 
   useEffect(() => {
     if (!isNavigationCollapsed && expanded) {
       setExpanded(false);
     }
   }, [expanded, isNavigationCollapsed]);
 
+  const handleKeyUp = useCallback(
+    (event) => {
+      if (event.keyCode === Key.Tab) {
```
kyledurand: Only set state if item was tabbed into

src/components/Navigation/components/Item/Item.tsx
```diff
@@ -87,13 +88,27 @@ export function Item({
...
+        !keyFocused && setKeyFocused(true);
+      }
+    },
+    [keyFocused],
+  );
+
+  const handleBlur = useCallback(() => {
+    keyFocused && setKeyFocused(false);
```
kyledurand: Only set state if item was focused. This prevents calling set state on ever click off of a nav item

kyledurand:
> Should these changes be isolated to the new design language?

Thought about that but old DL is unaffected by these changes since the nav didn't have any real focus state, just a css change to background and maybe subtle change to text color. I thought it would be overkill to add that extra check

src/components/Navigation/components/Item/Item.tsx
```diff
@@ -87,13 +88,27 @@ export function Item({
...
   const [expanded, setExpanded] = useState(false);
+  const [keyFocused, setKeyFocused] = useState(false);
 
   useEffect(() => {
     if (!isNavigationCollapsed && expanded) {
       setExpanded(false);
     }
   }, [expanded, isNavigationCollapsed]);
 
+  const handleKeyUp = useCallback(
```
tmlayton: A way to test this is to count how many times this is called with and without a new value for `keyFocused` to get an idea of how effective this is.

PR #2855, opened 2020-03-18
[Color system] finish filters

danrosenthal:
Closes https://github.com/Shopify/polaris-ux/issues/343

|:tophat:|:tophat:|
|-|-|
|Screen Shot 2020-03-18 at 10 38 46 AM|Screen Shot 2020-03-18 at 10 39 21 AM|


tmlayton:
When we do design review and polish would be nice to do something about that horizontal rule

PR #2856, opened 2020-03-18
[Filters] only show the "More filters" button if necessary

### LINKED ISSUE #2826: Feature request : making visibility of "More Filters" button in Filters component configurable
## Feature request summary

Making it possible to control/configure the visibility of the More Filters button in Filters component.

image

## Rationale

We are using Polaris components in Shopify-internal applications which assume the screen size is fairly large.
Some of the filters we'd like to have will include multiple control properties. And in such case it's beneficial to not have the "More Filters" button show up all the time.

Would be great to have a prop which controls the default visibility of that button.
Maybe have it show up if the width of Filters component goes below some threshold, but hide it when the control is wide enough?

---

rickitan:

Fixes #2826

The filters component always shows a "More filters" button. If there's only one filter or we can fit all the filters in the container, the button feels unnecessary.

This image describes the problem:

image

Notice how there's only one filter, and we still show the "More filters" button.

This PR introduces the following behaviour:

- If there's no filters marked with `shortcut: true` we show the `More filters` button
- If all the filters are marked with `shortcut: true` and we can fit them in the container, we don't show the `More filters` button
- If all the filters are marked with `shortcut: true` and we **CAN'T** fit them in the container, we *show* the `More filters` button
- If some of the filters are marked with `shortcut: true` but no all, we show the `More filters button`

This shows the new look:
image

The changes are backwards compatible.

### 🎩 checklist


src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.tsx
```diff
@@ -69,13 +70,23 @@ export class ConnectedFilterControl extends React.Component<
...
 
     const className = classNames(
       styles.ConnectedFilterControl,
       rightPopoverableActions && styles.right,
     );
 
+    // If there's only one action, and it fits we don't show the "More filters" button
+    const moreFilterButton =
+      rightPopoverableActions &&
+      rightPopoverableActions.length === 1 &&
```
tleunen: Hmm.. It shouldn't be based on the number 1 ;)

If all actions are "inlined", the "more filters" button becomes useless.
So if `rightPopoverableActions.length === this.getActionsToRender(rightPopoverableActions).length`, then we shouldn't render the button I believe.

src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.tsx
```diff
@@ -24,6 +24,7 @@ export interface ConnectedFilterControlProps {
   rightAction?: React.ReactNode;
   auxiliary?: React.ReactNode;
   disabled?: boolean;
+  hideMoreFiltersButton?: boolean;
```
tleunen: Not sure we really need this prop. I'd remove it

src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.tsx
rickitan: Currently, when you click on `More filters` while being on mobile you get the following:
image

Do you suggest keeping the Desktop behaviour if `rightPopoverableActions.length === this.getActionsToRender(rightPopoverableActions).length` is true? As so:

image

src/components/Filters/Filters.tsx
```diff
@@ -87,6 +87,8 @@ export interface FiltersProps {
   helpText?: string | React.ReactNode;
   /** Hide tags for applied filters */
   hideTags?: boolean;
+  /** Hide "More filters" button **/
+  hideMoreFiltersButton?: boolean;
```
tleunen: You can remove this one because we don't use it ;)

tmlayton:
I like the approach, only show more filters if there are actually more filters! 😄 

There is some polishing work to do. A couple things I noticed from a quick 🎩 are:

- Last button when hiding more filters does not have rounded corners<br>Screen Shot 2020-03-19 at 12 48 54 PM
- Some funky behavior here where more filters shows up even though the button is larger: <br>
measuring

rickitan:
Hi @tmlayton I just addressed you comments. Let me know what you think!

tmlayton:
Looking better, still some styling needed to fix up connected buttons. Noticed the rounded corners on the right side of the "Tagged with" button:

Screen Shot 2020-03-27 at 12 51 36 PM
Screen Shot 2020-03-27 at 12 51 46 PM

rickitan:
Hi tim! @tmlayton 

I just address the border-radius problem. Let me know what you think!

tmlayton:
Looks good!

tmlayton:
For test coverage run `dev test:coverage` or in this case `dev test:coverage ConnectedFilterControl.test.tsx` which will generate and open the report in the browser.

Looking at the [codecov diff](https://codecov.io/gh/Shopify/polaris-react/compare/27842a6389d2ebd073463fa5f8f2d47dcdc3a035...e184e9175813820efac737670ff935a711c6167b/diff) the lines without coverage are flagged in yellow for partial or red for none:

Screen Shot 2020-04-08 at 11 10 00 AM
Screen Shot 2020-04-08 at 11 10 06 AM

Adding tests for this is fairly straightforward. The two cases are when `shouldRenderMoreFiltersButton` is `false` and when only some buttons fit and the loop broken.

UNRELEASED.md
```diff
@@ -6,6 +6,8 @@
 
 ### Bug fixes
 
+- `Filters` only show the "More filters" button if necessary ([#](https://github.com/Shopify/polaris-react/pull/2856)).
```
tmlayton, line 9: Let’s move this to "Enhancements" and update the copy a bit (also missing the PR number):

```suggestion
- Updated `Filters` to only show the "More filters" button if necessary ([#2856](https://github.com/Shopify/polaris-react/pull/2856)).
```

src/components/Select/Select.scss
```diff
@@ -159,11 +163,6 @@ $stacking-order: (
     border-radius: var(--p-border-radius-base);
     background-color: var(--p-surface);
     @include focus-ring($border-width: rem(1px));
-
-    // 'position' needs to sit below focus-ring since it will be overwritten
```
tmlayton, line 163: This change should be reverted, `position` needs to be in this order.

tmlayton: Unfortunately the CLI formatter keeps making this change and not respecting the disabled lint

PR #2863, opened 2020-03-19
[ExceptionList] Fix collapsed exception list when description is undefined

jim-chou-shopify:
Fixes https://github.com/Shopify/marketing-integrations/issues/576.
`ExceptionList` breaks the layout when there is an icon and description is undefined and has a sibling element. 
`ExceptionList` renders icon with absolute position and when there is no message the container has no height.
[DEMO of the problem](https://codesandbox.io/s/zealous-ritchie-s9261)

Apply `min-height: rem(20px)` to `ul` container element.
The min container height is base on the [icon height](https://github.com/Shopify/polaris-react/blob/02c2fd3a6f7fc3680b67eae5986a2e517ccdf085/src/components/ExceptionList/ExceptionList.scss#L27).

**Before**
Screen Shot 2020-03-19 at 3 52 32 PM
**After**
Screen Shot 2020-03-19 at 3 52 20 PM

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -6,6 +6,8 @@
 
 ### Bug fixes
 
+- Fix collapsed `ExceptionList` when description is undefined [#2863](https://github.com/Shopify/polaris-react/pull/2863)
+
 ### Documentation
 
 ### Development workflow
--- a/src/components/ExceptionList/ExceptionList.scss
+++ b/src/components/ExceptionList/ExceptionList.scss
@@ -1,9 +1,12 @@
 @import '../../styles/common';
 
+$icon-height: rem(20px);
+
 .ExceptionList {
   margin: 0;
   padding: 0;
   list-style: none;
+  min-height: $icon-height;
 }
 
 .Item {
@@ -24,7 +27,7 @@
   align-items: center;
   justify-content: center;
   min-width: rem(20px);
-  height: rem(20px);
+  height: $icon-height;
   margin-right: spacing(extra-tight);
 
   @include recolor-icon(var(--p-icon-subdued, color('ink', 'lightest')));


BPScott:
Setting heights on containing elements always feel pretty janky to me. I'd  rather  try and work out some way that the line that the icon is on has some kind of height and it all flows from there. Otherwise things will get weird if users resize their text sizes.


I've not poked about with this but can we make it so the Icon is no longer absolutely positioned, (inline-block perhaps?) that might help give that line some intrinsic height

jim-chou-shopify:
> Setting heights on containing elements always feel pretty janky to me. I'd rather try and work out some way that the line that the icon is on has some kind of height and it all flows from there. Otherwise things will get weird if users resize their text sizes.
> 
> I've not poked about with this but can we make it so the Icon is no longer absolutely positioned, (inline-block perhaps?) that might help give that line some intrinsic height

Regardless of the font size, the icon will probably not change height. 
Using the min height of the icon sort of works the same as making icon not absolute positioned.
Im not sure why icon need to be absolute positioned neither. Maybe worth a try.

AndrewMusgrave:
> Regardless of the font size, the icon will probably not change height.
Using the min height of the icon sort of works the same as making icon not absolute positioned.
Im not sure why icon need to be absolute positioned neither. Maybe worth a try.

👋 Let us know if you're going to take a look. If not this should be good for a review!

jim-chou-shopify:
> > Regardless of the font size, the icon will probably not change height.
> > Using the min height of the icon sort of works the same as making icon not absolute positioned.
> > Im not sure why icon need to be absolute positioned neither. Maybe worth a try.
> 
> 👋 Let us know if you're going to take a look. If not this should be good for a review!

Hey @BPScott @AndrewMusgrave,
I don't think I have time to look into refactoring the component to use relative position.
Do you think the proposed change in the PR is worth reviewing?

BPScott:
Removing my review request as I'm moving off the polaris team

alex-page:
@jim-chou-shopify is this still an issue?

PR #2864, opened 2020-03-20
Revert "[Navigation] Design language animation"

danrosenthal:
Reverts Shopify/polaris-react#2796

https://github.com/Shopify/web/pull/24837

|bug|
|-|
|Screen Shot 2020-03-19 at 8 51 44 PM|

### DIFF
--- a/src/components/Navigation/Navigation.scss
+++ b/src/components/Navigation/Navigation.scss
@@ -282,6 +282,7 @@ $disabled-fade: 0.6;
 $secondary-item-font-size: rem(15px);
 .SecondaryNavigation {
   flex-basis: 100%;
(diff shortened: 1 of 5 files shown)


danrosenthal:
This bug doesn't happen consistently

|:bug:|
|-|
|Screen Shot 2020-03-19 at 9 00 08 PM|

kyledurand:
Might be a bug in collapsible or a check in `showExpanded`. Wonder if it could be related to a global nav change that we're not seeing on the Polaris side: https://github.com/Shopify/web/pull/24465
Digging in

danrosenthal:
I'm going to go ahead and merge this so master is shippable

PR #2871, opened 2020-03-24
[Focus] and [Filters] Fixing initial reference focusing

posth:

Fixes https://github.com/Shopify/polaris-react/issues/2741

There are two issues with the current behavior:
- When toggling open a new filter, it does not focus the first selectable input (it does so after selecting and re-toggling)
- When selecting an input on a newly toggled filter, it will force a focus away

The reference being passed to the `<Focus />` component from `<Filters />` was the direct reference to it's `current` value, which would be a stale reference by the time the component is mounted and rendered, forcing that focused element to be "one step behind."

This PR adds an union type to be passed to the `root` of `<Focus />`, which is the entire `React.RefObject`, and lets the `Focus` component grab it's `current` node instead, so that we are sure that the `current` node we want to focus on is the node value of the current reference we passed.

Choosing to union the type onto the prop for backwards compatibility and not break current consumers.

## 

### How to 🎩

To :tophat:
- Run `yarn dev`
- Select the `Filters` component and then `Filter with resource list`
- This renders a default resource list in the sandbox
- Toggle the `More filters`
- Toggle open the `Account status`
- It will focus on the first selectable input (did not do that before)
- Select another option and it will stay focused on that input (before it would focus the first input)

|filtersFocusFix|
|---|

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -13,6 +13,7 @@
 
 - Fixed `Tag` submitting forms when `onClick` is set ([#2895](https://github.com/Shopify/polaris-react/pull/2895))
 - Fixed `DescriptionList` content overflowing when `term` or `description` have long unbroken words ([#2880](https://github.com/Shopify/polaris-react/pull/2880))
+- Fixed focusing bug on Filters where a newly opened filter would not initially focus the first input, and a newly opened filter would incorrectly focus after an input selection ([#2871](https://github.com/Shopify/polaris-react/pull/2871))
 
 ### Documentation
 
--- a/src/components/Focus/Focus.tsx
+++ b/src/components/Focus/Focus.tsx
@@ -4,7 +4,7 @@ import {focusFirstFocusableNode} from '@shopify/javascript-utilities/focus';
 export interface FocusProps {
   children?: React.ReactNode;
   disabled?: boolean;
-  root: HTMLElement | null;
+  root: React.RefObject<HTMLElement> | HTMLElement | null;
 }
 
 export const Focus = memo(function Focus({
@@ -13,9 +13,24 @@ export const Focus = memo(function Focus({
   root,
 }: FocusProps) {
   useEffect(() => {
-    if (disabled || !root || root.querySelector('[autofocus]')) return;
-    focusFirstFocusableNode(root, false);
+    if (disabled || !root) {
+      return;
+    }
+
+    const node = isRef(root) ? root.current : root;
+
+    if (!node || node.querySelector('[autofocus]')) {
+      return;
+    }
+
+    focusFirstFocusableNode(node, false);
   }, [disabled, root]);
 
   return <React.Fragment>{children}</React.Fragment>;
 });
+
+function isRef(
+  ref: React.RefObject<HTMLElement> | HTMLElement,
+): ref is React.RefObject<HTMLElement> {
+  return (ref as React.RefObject<HTMLElement>).current !== undefined;
+}
--- a/src/components/Focus/tests/Focus.test.tsx
+++ b/src/components/Focus/tests/Focus.test.tsx
@@ -21,11 +21,22 @@ describe('<Focus />', () => {
     expect(document.body).toBe(document.activeElement);
   });
 
-  it('will focus first focusable node', () => {
+  it('will focus first focusable node when passing current node', () => {
     const focus = mountWithAppProvider(
       <FocusTestWrapper>
         <input />
       </FocusTestWrapper>,
+    );
+
+    const input = focus.find('input').getDOMNode();
+    expect(input).toBe(document.activeElement);
+  });
+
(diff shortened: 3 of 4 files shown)


src/components/Focus/tests/Focus.test.tsx
```diff
@@ -45,14 +45,9 @@ describe('<Focus />', () => {
 
 function FocusTestWrapper({children, ...props}: Omit<FocusProps, 'root'>) {
   const root = useRef<HTMLDivElement>(null);
-  const [, setMount] = useState(false);
-
-  useEffect(() => {
```
posth: For testing, this was a false positive.  Forcing the component to remount would cause that stale ref to be properly set and focused.  

Removing this in the mock component with the fix makes the test work matching the component's behavior.

src/components/Focus/Focus.tsx
```diff
@@ -4,7 +4,7 @@ import {focusFirstFocusableNode} from '@shopify/javascript-utilities/focus';
 export interface FocusProps {
   children?: React.ReactNode;
   disabled?: boolean;
-  root: HTMLElement | null;
+  root: React.RefObject<HTMLElement> | HTMLElement | null;
```
posth: I feel like passing the entire ref is safer, but this would break the API of this component - would it make sense to go towards passing the entire ref and add a warning of deprecation for passing only a current element?

UNRELEASED.md
```diff
@@ -8,6 +8,7 @@
 
 - Fixed right padding styling issue with the `Tag` component and remove right padding on a removable `Tag` ([#2860](https://github.com/Shopify/polaris-react/pull/2860)).
 - Fixed secondary navigation spacing when no icon is present ([#2874](https://github.com/Shopify/polaris-react/pull/2874)).
+- Fixed focusing bug on Filters where a newly openned filter would not initially focus the first input, and a newly openned filter would incorrectly focus after an input selection ([#2871](https://github.com/Shopify/polaris-react/pull/2871))
```
loic-d, line 11: `opened`?

src/components/Focus/Focus.tsx
```diff
@@ -13,8 +13,19 @@ export const Focus = memo(function Focus({
...
+    }
+
+    let node: HTMLElement | null;
+    if ('current' in root) {
+      node = (root as React.RefObject<HTMLElement>).current;
+    } else {
+      node = root as HTMLElement;
+    }
+
+    if (!node || node.querySelector('[autofocus]')) return;
```
loic-d, line 27: Personal preference but I am not a fan of this one liner syntax.

posth: Will change it - it will also match the early return above

chloerice:
Hey @posth sorry I missed this, will tophat this afternoon!

src/components/Focus/Focus.tsx
kyledurand: I'm a fan of this and would prefer to remove `HTMLElement` altogether but that can be done in another PR since it would be a breaking change.

src/components/Focus/Focus.tsx
```diff
@@ -13,8 +13,22 @@ export const Focus = memo(function Focus({
...
+    if (disabled || !root) {
+      return;
+    }
+
+    let node: HTMLElement | null;
+    if ('current' in root) {
+      node = (root as React.RefObject<HTMLElement>).current;
+    } else {
+      node = root as HTMLElement;
+    }
```
kyledurand, line 25: Not a fan of this casting. Can you use a custom type guard here instead? https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

kyledurand: Just pseudo coding here but something like this:

```js
function isRef(ref: React.RefObject<HTMLElement> | null): ref is React.RefObject<HTMLElement> {
  return (ref as React.RefObject<HTMLElement>).current !== undefined;
}
```

posth: Much cleaner thank you - made the change to use a type guard instead 👍

src/components/Focus/Focus.tsx
posth: Because it's a breaking change, is there a way in Polaris to have a warning for future deprecations?

src/components/Focus/Focus.tsx
```diff
@@ -13,9 +13,24 @@ export const Focus = memo(function Focus({
...
   }, [disabled, root]);
 
   return <React.Fragment>{children}</React.Fragment>;
 });
+
+function isRef(
+  ref: React.RefObject<HTMLElement> | HTMLElement,
+): ref is React.RefObject<HTMLElement> {
+  return (ref as React.RefObject<HTMLElement>).current !== undefined;
+}
```
kyledurand, line 36: 👏

PR #2873, opened 2020-03-25
Enable Typescript 3.8 and update Sewing-kit to latest

BPScott:

Typescript 3.8 contains new syntax abilities (`export type`) that enable my machinations for drastically simplifying our build process - moving from a two-step thing where we compile our TS files with typescript, then compile that output with babel, to having babel reading the TS files directly.

It's not as simple as "just update typescript" though, as we have to also update all our linting tooling to new versions that understand this new syntax too, which is why we have to update sewing-kit too. 

A sewing-kit update includes updates to our linting rules which are almost entirely autofixable but make this PR look a lot bigger and scarier than what it actually is.

#### Ben, why are you giving us ~400 file PRs to review on hackdays FFS

Yeah sorry. I figure it's a lot cleaner to do this piece by piece, get this pulled into the v5 branch, and then make do my build wrangling there, rather than hacking together an even bigger monstrosity that I then have to pick apart and submit piece by piece later on anyway.

- Updates sewing-kit to the latest version - this is needed because it contains updates to linting configs so that they support the new typescript syntax
- Adds a resolution for prettier v2 - also needed because it's where support for the new syntax was added. [I've updated SK to use prettier v2](https://github.com/Shopify/sewing-kit/pull/1918), but it's not been released just yet. As soon as that gets released we can update to the newest SK  and remove this resolution.
- Performs linting fixes as a result of pulling in an up-to-date version of our linting config. The most notable change here is that it autofixes adding blank lines between the group of imports from packages, and the group of imports that are relative paths. These autofixed lints have been isolated into their own files.
- Updates typescript to v3.8, which required a few small casts
- Turns on the "isolatedModules" option in TypeScript which warns us about usage patterns that won't work when we compile directly with babel. This mostly boils down to "you cant reexport a type without using `export {SomeType}`, you have to do `export type {SomeType}`.

### How to 🎩

It's probably best to code review this on a per-commit basis and you can skip over the "automated update from sk format" commits as that's the same sort of whitespace changes in every file.

Ensure linting, tests and typechecks all pass and storybook looks the same as before.

I've compared the build output of the esnext folder, our index.js, index.es.js and styles.css files with the output from master and they're all functionally identical - just a few changes from babel updates in the index.js and index.es.js files.

### DIFF
--- a/src/components/DatePicker/DatePicker.tsx
+++ b/src/components/DatePicker/DatePicker.tsx
@@ -17,8 +17,8 @@ import {Button} from '../Button';
 import {classNames} from '../../utilities/css';
 import {useI18n} from '../../utilities/i18n';
 import {useFeatures} from '../../utilities/features';
-import {monthName} from './utilities';
 
+import {monthName} from './utilities';
 import {Month} from './components';
 import styles from './DatePicker.scss';
 
--- a/src/components/DatePicker/DatePicker.tsx
+++ b/src/components/DatePicker/DatePicker.tsx
@@ -22,7 +22,8 @@ import {monthName} from './utilities';
 import {Month} from './components';
 import styles from './DatePicker.scss';
 
-export {Range, Months, Year};
+export {Months};
+export type {Range, Year};
 
 export interface DatePickerProps {
   /** ID for the element */
--- a/src/components/index.ts
+++ b/src/components/index.ts
@@ -1,47 +1,68 @@
-export {AccountConnection, AccountConnectionProps} from './AccountConnection';
+export {AccountConnection} from './AccountConnection';
+export type {AccountConnectionProps} from './AccountConnection';
 
-export {ActionList, ActionListProps} from './ActionList';
+export {ActionList} from './ActionList';
+export type {ActionListProps} from './ActionList';
 
-export {ActionMenu, ActionMenuProps} from './ActionMenu';
+export {ActionMenu} from './ActionMenu';
+export type {ActionMenuProps} from './ActionMenu';
 
-export {AppProvider, AppProviderProps} from './AppProvider';
+export {AppProvider} from './AppProvider';
+export type {AppProviderProps} from './AppProvider';
 
-export {Autocomplete, AutocompleteProps} from './Autocomplete';
+export {Autocomplete} from './Autocomplete';
+export type {AutocompleteProps} from './Autocomplete';
 
-export {Avatar, AvatarProps} from './Avatar';
+export {Avatar} from './Avatar';
+export type {AvatarProps} from './Avatar';
 
-export {Backdrop, BackdropProps} from './Backdrop';
+export {Backdrop} from './Backdrop';
+export type {BackdropProps} from './Backdrop';
 
-export {Badge, BadgeProps} from './Badge';
+export {Badge} from './Badge';
+export type {BadgeProps} from './Badge';
 
-export {Banner, BannerProps, BannerStatus} from './Banner';
+export {Banner} from './Banner';
+export type {BannerProps, BannerStatus} from './Banner';
 
-export {Breadcrumbs, BreadcrumbsProps} from './Breadcrumbs';
+export {Breadcrumbs} from './Breadcrumbs';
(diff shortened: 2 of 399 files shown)


src/components/index.ts
```diff
@@ -1,224 +1,299 @@
...
+export type {
   DataTableProps,
   TableData,
   TableRow,
   SortDirection,
   ColumnContentType,
 } from './DataTable';
 
-export {DatePicker, DatePickerProps, Range, Months, Year} from './DatePicker';
+export {DatePicker, Months} from './DatePicker';
```
kaelig: I'm new to this, and find surprising that `Months` is in this export statement (rather than the one below, alongside type exports), being an `enum`. Can you explain why this is the case?

kaelig:
Suggestion: relaxing or disabling `selector-max-class` at the project level. Thoughts?

src/components/DatePicker/DatePicker.tsx
```diff
@@ -22,7 +22,8 @@ import {monthName} from './utilities';
 import {Month} from './components';
 import styles from './DatePicker.scss';
 
-export {Range, Months, Year};
+export {Months};
```
amrocha: What's going on here? Is `Months` not a type even though it's imported from the same place as Year and Range?

src/components/index.ts
BPScott: Broadly speaking you've got two sorts of thing: "values" which have some kind of JS output, and "types" which are only used at compile time and get stripped away from the resultant JS.

[A enum ends up outputting javascript code](https://www.typescriptlang.org/play/#code/KYOwrgtgBAgiCi5oG8CwAoKWBuBDANmMAIwA0GWUehwATBgL4bPoBmYIAxgC4CWA9iCi4ATgHMAkgGdEkABSixALlgIkASijIoFLCODcwIoYsYtF02RDlwrAOmpFi6gNwsA9O6gBaXy24AngAOwLAAKsGhALxQAOSOJLFQAD5xCbSxbugY7Fx8gsLi0gCqIAIgESFyAEYqMJXAmtq6UPqGxoViZtk9FlKl5Q1y8QROsepAA) so it is a value (same as something like `const a =1`, which means it gets exported with all our other values (like constants and functions etc).

BPScott:
> Suggestion: relaxing or disabling selector-max-class at the project level. Thoughts?

This has some rather funky knock on effects as it intersects with a handful of other rules around selector specificity - you'd want to update them all at the same time.

I don't think it's particularly worth loosening the rules as we mostly fall afoul of this because we're wrapping stuff in `.newDesignLanguage` and that concept will go away eventually.

src/components/DatePicker/DatePicker.tsx
BPScott: Bingo. Months is an enum, while Range and Year are both types. See https://github.com/Shopify/javascript-utilities/blob/master/src/dates.ts#L11


Follow up questions: is this annoying and would it be better if Months was a union type like `type Months = 0 |1|2|3|4|5|6|7|8|9|10|11`? Yeah probably, but thar be dragons in the javascript-utilities packages and I want to stop relying on eventually so just stick your fingers in your ears and ignore it for now :D

kaelig:
>I think I would wait for your sk changes to ship before shipping this, we're not in a rush here and that would remove an intermediary step so it seems like there's no drawbacks

One drawback I can see is that this is a large PR that touches a lot of files, meaning the risk of encountering merge conflicts increases quite a bit as time goes by. Not a huge issue, but could be leading to more work than shipping this in 2 steps.

BPScott:
> > I think I would wait for your sk changes to ship before shipping this, we're not in a rush here and that would remove an intermediary step so it seems like there's no drawbacks
> 
> One drawback I can see is that this is a large PR that touches a lot of files, meaning the risk of encountering merge conflicts increases quite a bit as time goes by. Not a huge issue, but could be leading to more work than shipping this in 2 steps.

Yep, that pretty much.  Getting this in now unblocks me doing other stuff, that'll be much easier to shuffle around without all this other noise. The update to the currently-unreleased SK will comparativly empty as there won't be any new linting rules appearing.

PR #2875, opened 2020-03-27
Support consumers using older TS versions

BPScott:

We want to use new and shiny typescript features. But some of our consumers are on old versions and won't understand the type definition files that use new features, which will cause them to fail. Using [`downlevel-dts`](https://github.com/sandersn/downlevel-dts) we can generate multiple type versions, so that consuming apps using older versions can receive and use type definitions that they understand, while we get to use the new shiny.

An example of this new shiny is `export type` as introduced in 3.8.

- Stop moving types from `types/src` to `types/` - TS generates those files at that location, let's just use it.
- Use `downlevel-dts` to generate types compatible with older TS version. The lowest it goes is 3.4 so let's use that.
- Update our types and typeVersions specifications in package.json

Locations: 
- `types/latest/src/...` is where they types for up-to-date (we're using v3.9.x) TS lives
- `types/3.4/src/...` is where the types for older versions lives. `downlevel-dts` doesn't support transforms for any versions older than 3.4

This changes the layout of our distributed files in the types folder, however people should
never be importing content directly from the types folder anyway as it
isn't part of our public API, so I don't consider this a breaking change.

### 🎩 checklist

- Use `build-consumer` to push in these changes into a consuming project that uses an old TS version. `polaris-styleguide` is a good bet.
- Ensure that the con consuming project's `type-check` still passes and you can still access the types from importing `@shopify/polaris` from within that project.

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -5,7 +5,7 @@ const {join, resolve: resolvePath} = require('path');
 
 const {ensureDirSync, writeFileSync, readFileSync} = require('fs-extra');
 const {rollup} = require('rollup');
-const {cp, mv, rm} = require('shelljs');
+const {cp, mv} = require('shelljs');
 const copyfiles = require('copyfiles');
 
 const createRollupConfig = require('../config/rollup');
@@ -20,21 +20,18 @@ const intermediateBuild = resolvePath(root, './build-intermediate');
 const mainEntry = resolvePath(intermediateBuild, './index.js');
 
 const scripts = resolvePath(root, 'scripts');
-const types = resolvePath(root, 'types');
 const tsBuild = resolvePath(scripts, 'tsconfig.json');
 
+const execOptions = {stdio: 'inherit', cwd: root};
+
 execSync(
-  `${resolvePath(
-    root,
-    './node_modules/.bin/tsc',
-  )} --outDir ${intermediateBuild} --project ${tsBuild}`,
-  {
-    stdio: 'inherit',
-  },
+  `yarn run tsc --outDir ${intermediateBuild} --project ${tsBuild}`,
+  execOptions,
 );
 
-mv(resolvePath(root, 'types/src/*'), types);
-rm('-rf', resolvePath(root, 'types/src'));
+// Downlevel type declarations to support consuming apps that use older versions
+// of typescript
+execSync(`yarn run downlevel-dts types/latest types/3.4`, execOptions);
 
 mv(resolvePath(intermediateBuild, 'src/*'), intermediateBuild);
 
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -18,4 +18,6 @@
 
 ### Code quality
 
+- Use `downlevel-dts` to produce compatible type definitions for consuming apps using older TypeScript versions ([#2875](https://github.com/Shopify/polaris-react/pull/2875))
+
 ### Deprecations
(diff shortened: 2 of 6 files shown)


scripts/build.js
```diff
@@ -20,21 +20,18 @@ const intermediateBuild = resolvePath(root, './build-intermediate');
 const mainEntry = resolvePath(intermediateBuild, './index.js');
 
 const scripts = resolvePath(root, 'scripts');
-const types = resolvePath(root, 'types');
 const tsBuild = resolvePath(scripts, 'tsconfig.json');
 
+const execOptions = {stdio: 'inherit', cwd: root};
```
amrocha: nice!

scripts/build.js
```diff
@@ -20,21 +20,18 @@ const intermediateBuild = resolvePath(root, './build-intermediate');
...
-  },
+  `yarn run tsc --outDir ${intermediateBuild} --project ${tsBuild}`,
+  execOptions,
 );
 
-mv(resolvePath(root, 'types/src/*'), types);
-rm('-rf', resolvePath(root, 'types/src'));
+// Downlevel type declarations to support consuming apps that use older versions
+// of typescript
+execSync(`yarn run downlevel-dts types/latest types/3.4`, execOptions);
```
amrocha: Is this supported by TS? Like, does TS recognize when it should load an older version's types?

BPScott:
> The code seems fine. but can you explain why you're doing this?

We've wrote code using TS 3.8, using features only available in TS 3.8. This produces type definitions that contain those features, notable we're using `export type {SomeType}` as of my TS 3.8 update PR. 

Apps that consume us might not be using TS 3.8, (e.g. web is currently using 3.7, polaris-styleguide is using 3.1), so when they come to read our type definitions they get to the bit that says `export type {SomeType}` and it crashes because that consuming app's version of typescript doesn't understand the syntax that we're using.

This has happened before in the past when we've tried to use the `Omit` built-in type that was only added in 3.5 - consuming apps using <TS 3.5 couldn't compile because they couldn't understand our type definitions, so we had to stop doing that.

This PR offers a different approach instead of "just don't use new features" - This says "we'll write TS using new features, and offer up two sets of types - one consumable by latest TS versions, and one that has our types transpiled down into a format that is understandable by older TS versions"
We take advantage of a ["typeVersions" - TS feature that allows multiple versions of types to be defined](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-1.html#version-selection-with-typesversions), and lets the consuming app to pick a set of types based upon its version.

We ship two sets of types:
- `types/latest` which is used by consuming apps using 3.8 and above
- `types/3.4` which is used by consuming apps using less than 3.8


> Can't we just upgrade the TS version? 

This isn't our TS version to update - it's the version of  TS in the consuming apps that is the problem - we can't go and update TS in every app that consumes polaris :)

> Are you worried about making breaking changes for consumers? In that case, should we be deprecating the old version in this PR? 

Using new features is a breaking change before this PR, as we would have had to say. "to use this version of Polaris you must also use this TS version too". However now we're shipping a downleveled version of our types it is no longer a breaking change - consuming apps on older TS versions just use the downleveled version of our types.

Think of how web lets people write modern JS syntax and then ships two bundles - one for evergreen browsers containing latest and greatest syntax, and one that is transpiled to an older syntax to maintain support for older browsers. We're doing the same thing but for types and consuming apps using latest vs older typescript versions.

Or is this something you see us supporting in perpetuity?

We should keep this in perpetuity as it allows us to utilise modern TS features without having to worry too much about breaking compatibility with consuming apps using older TS versions.

PR #2885, opened 2020-04-01
[ResourceList] hides header if showHeader is false

kvendrik:

I'm currently working on a component that I would like to have utilize everything ResourceList has to offer but I don't need the select all functionality.

- It makes `showHeader` an optional prop that is either a boolean or undefined 
- It doesn't render a header if you set `showHeader` to `false` explicitly 

### How to 🎩
1. Render a `ResourceList` like this
```tsx
<ResourceList
  showHeader={false}
  selectable
  items={items}
  renderItem={renderItem}
/>
```
2. Without the changes in this PR it will render a header. With the changes it this PR it will not.

<details>

```jsx
import React from 'react';

import {
  Page,
  ResourceList,
  ResourceItem,
  Card,
  Avatar,
  TextStyle,
} from '../src';

  return (
    <Page title="Playground">
      <ResourceListExample />
    </Page>
  );
}

function ResourceListExample() {
  const items = [
    {
      id: 231,
      url: 'customers/231',
      name: 'Mae Jemison',
      location: 'Decatur, USA',
    },
    {
      id: 246,
      url: 'customers/246',
      name: 'Ellen Ochoa',
      location: 'Los Angeles, USA',
    },
    {
      id: 276,
      url: 'customers/276',
      name: 'Joe Smith',
      location: 'Arizona, USA',
    },
    {
      id: 349,
      url: 'customers/349',
      name: 'Haden Jerado',
      location: 'Decatur, USA',
    },
    {
      id: 419,
      url: 'customers/419',
      name: 'Tom Thommas',
      location: 'Florida, USA',
    },
    {
      id: 516,
      url: 'customers/516',
      name: 'Emily Amrak',
      location: 'Texas, USA',
    },
  ];

  return (
    <Card>
      <ResourceList
        showHeader={false}
        selectable
        items={items}
        renderItem={renderItem}
      />
    </Card>
  );

  function renderItem(item, _, index) {

    return (
      <ResourceItem
        id={id}
        url={url}
        media={media}
        sortOrder={index}
      >
        <h3>
        </h3>
        <div>{location}</div>
      </ResourceItem>
    );
  }
}


```

</details>

### DIFF
--- a/src/components/ResourceList/ResourceList.tsx
+++ b/src/components/ResourceList/ResourceList.tsx
@@ -373,7 +373,7 @@ class ResourceListInner extends React.Component<CombinedProps, State> {
       bulkActions,
       filterControl,
       loading,
-      showHeader = false,
+      showHeader,
       sortOptions,
       sortValue,
       alternateTool,
@@ -469,6 +469,7 @@ class ResourceListInner extends React.Component<CombinedProps, State> {
 
     const headerMarkup = !showEmptyState &&
       (showHeader || needsHeader) &&
+      showHeader !== false &&
       this.listRef.current && (
         <div className={styles.HeaderOuterWrapper}>
           <Sticky boundingElement={this.listRef.current}>
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -2,7 +2,7 @@
 
 ### Breaking changes
 
-- Added functionality to that hides the `ResourceList` when `showHeader` is set to `false` explicitely ([#2885](https://github.com/Shopify/polaris-react/pull/2885)).
+- Added functionality that hides the `ResourceList` when `showHeader` is set to `false` explicitly ([#2885](https://github.com/Shopify/polaris-react/pull/2885)).
 
 ### Enhancements
 
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -2,14 +2,13 @@
 
 ### Breaking changes
 
-- Added functionality that hides the `ResourceList` when `showHeader` is set to `false` explicitly ([#2885](https://github.com/Shopify/polaris-react/pull/2885)).
-
 ### Enhancements
 
 ### Bug fixes
 
 - Fixed right padding styling issue with the `Tag` component and remove right padding on a removable `Tag` ([#2860](https://github.com/Shopify/polaris-react/pull/2860)).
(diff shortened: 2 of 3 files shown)


kyledurand:
What would the repercussions be in web with this change? In my mind you'd have to make set this explicitly in a few places, which is manageable, but what about partners?

kvendrik:
> What would the repercussions be in web with this change? In my mind you'd have to make set this explicitly in a few places, which is manageable, but what about partners?

I don't think this will require any consumer changes. The only diff is that now its falsy by default instead of `false`. None of the prior code seemed to rely on it being false explicitly. So TL;DR AFAIK no consumer changes will be needed but if you set this to `false` explicitly (I assume no consumer has done this so far) the header will hide.

src/components/ResourceList/ResourceList.tsx
```diff
@@ -480,6 +480,7 @@ class ResourceListInner extends React.Component<CombinedProps, State> {
 
     const headerMarkup = !showEmptyState &&
       (showHeader || needsHeader) &&
+      showHeader !== false &&
```
kyledurand, line 483: Ah I missed this check the first time around.

```suggestion
      (showHeader !== false || needsHeader) &&
```

Can we merge this with the line above?

kvendrik: By merging it like that `(showHeader || needsHeader)` would come back as true when `showHeader` is `undefined` which I don't think is what we want. Currently it would come back as `false` when `showHeader` is falsy and we then check if isn't set to `false` explicitly (`showHeader !== false`) as in that case we don't want the header to render.

kyledurand: Ugh, right. Disregard

kvendrik: I do think though that this condition is really confusing. Outside the scope of this PR but it might benefit from some cleanup.

kvendrik:
@AndrewMusgrave correct if the consumer currently has `showHeader` explicitly set to `false` this could be a breaking change for the reason you're describing. I was betting on that consumers probably don't have that but on second thought: lets not do that, I'll change this to it being a breaking change.

tmlayton:
Agree with the breaking change approach, but I would frame the changelog entry as a bug fix. To me this is a breaking bug fix.

PR #2888, opened 2020-04-02
[WIP] Storybook docs playing

BPScott:
Some silly little examples to show what you can do with storybook docs / global types

We can use this as a little testbed for playing with text-only doc generation.

- Button.stories.mdx to show component examples with mdx
- JustWords.stories.mdx to show you can have docs unrelated to a
component
- Some rather hacky stuff to showcase generating docs files from our
markdown format. This is pretty crappy atm but shows we can generate our
on docs html from markdown. Stuff like replacing code blocks with
rendered examples will take more work.

This PR also migrates us away from the deprecated contexts configuration to use the new globalTypes. Unfortunatly this loses us some some functionality.
- globalTypes can not yet be programatically set when loading the preview iframe (this means we can't a11y check dark mode)
- globalTypes lacks a good way to configure a complex object so we can't replace the knobs for configuring the color scheme items (surface, critical etc)

### To Play

Open up the deployed storybook, and look at the new "Playground/Button" "Playground/Just Words" docs views, and hit the docs view from any of our existing readmes.

See https://5d559397bae39100201eedc1-wnkvmeevfo.chromatic.com/?path=/story/playground-0-button-csf--button-default

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -15,7 +15,11 @@ const {BundleAnalyzerPlugin} = require('webpack-bundle-analyzer');
 const enableDocs = !parseInt(process.env.STORYBOOK_DISABLE_DOCS || '0', 10);
 
 module.exports = {
-  stories: ['../playground/stories.tsx', '../src/components/**/*/README.md'],
+  stories: [
+    '../playground/stories.tsx',
+    '../playground/**/*.stories.{tsx,mdx}',
+    '../src/components/**/*/README.md',
+  ],
   addons: [
     {name: '@storybook/addon-essentials', options: {docs: enableDocs}},
     '@storybook/addon-a11y',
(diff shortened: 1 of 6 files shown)


BPScott:
> globalTypes can not yet be programatically set when loading the preview iframe (this means we can't a11y check dark mode)

This has been fixed as of storybook v6.3.0 -https://github.com/storybookjs/storybook/pull/15056

alex-page:
@BPScott going to close for now but feel free to reopen.

BPScott:
Time has moved on, as we've removed much of the color /darkmode behaviour we've actually already implemented most of the globalArgs changes in this PR. The only thin that would remain is the mdx file / doc pages stuff

alex-page:
This is being explored for the new polaris.shopify.com so I think we can call this a wrap. https://github.com/Shopify/polaris/issues/5613

PR #2891, opened 2020-04-03
Update font stack

martenbjork:

When a user only has one Robot weight installed, that weight is applied to all elements. This PR puts Segoe before Roboto in the font stack. If the user has both fonts, it's more likely that Segoe contains all the weights, so it makes sense to put it first.

This is similar to the font stacks of Github, Twitter and others.

03-04-hkmjb-35t18


danrosenthal:
Just needs a changelog entry!

PR #2901, opened 2020-04-07
Enable "importsNotUsedAsValues:error" TypeScript config

### LINKED ISSUE #2896: esnext builds are deoptimized by empty types.js files
## Issue summary
webpack [flags any file without an import/export statement as a CommonJS module](https://github.com/webpack/webpack/blob/ba08e41c3b6efa9495f22ad4f6dc3d1fc8cdf5a7/lib/dependencies/HarmonyDetectionParserPlugin.js#L21-L38).  CommonJS modules cannot be concatenated, which causes them to be wrapped in webpack boilerplate.  This increases asses sizes.

Polaris 4.17.0 ships with multiple empty types.js files (e.g., [features](https://cdn.jsdelivr.net/npm/@shopify/polaris@4.17.0/esnext/utilities/features/types.js)).  For consumers without [sewing-kit's workaround](https://github.com/Shopify/sewing-kit/blob/27e7fd300eb6d857a0d77357f98e4b74f223c93b/packages/sewing-kit/src/tools/webpack/config/rules.ts#L261-L263), these add a little bit of bulk :(

### But types.js files existed before this!
Yeah, but:
* Some index files don't export their types
* Some types files contain non-const enums (which generate exportable code)

Probably other stuff; I haven't looked at this in depth.

### But only sewing-kit consumers should be using the esnext build!

Even if that's true, these modules are causing another breakage for all sewing-kit consumers.

## Expected behavior
All of Polaris to be treated as es modules.

## Actual behavior
Some CommonJS boilerplate 😭 

Note: I haven't verified this because all of my consumers use sewing-kit, but I'm usually right about this sort of thing 🤷‍♀ 

That's pretty involved!  If you need help with that, I can cook something up for sure.

BPScott:

Follow up to #2897
Fixes #2896

By ensuring we mark type-only imports we make sure that we avoid
deoptimisations and export issues if compiled files reference empty
files (which are empty because types have no compile time output)

Sets `importsNotUsedAsValues` to `error` in our tsconfig and change any `import {SomeType}` references to `import type {SomeType}` that it flags.

### How to 🎩

Ensure that type-check passes.
Ensure that the build output does not change, by doing the following:

- Checkout the parent of this PR, do a build and copy the esnext folder
  - `git checkout 27842a6389d2ebd073463fa5f8f2d47dcdc3a035 && yarn build && mv esnext esnext-base`
- Checkout this branch, do a build and compare the contents of the esnext folder to the copy you just created
  - `git checkout ts-importsNotUsedAsValues && yarn build && diff -ru esnext-base esnext`
- Note that the build output is identical


src/components/ResourceList/components/FilterControl/types.ts
```diff
@@ -1,5 +1,5 @@
-import {SelectOption} from '../../../Select';
-import {TextFieldProps} from '../../../TextField';
+import type {SelectOption} from '../../../Select';
+import type {TextFieldProps} from '../../../TextField';
```
alex-page, line 2: There are a few imports that come from the Component name. Should we move them to types?

BPScott: talked about this on slack, but short answer is: No: `TextField` in this case is the public API of the component

PR #2913, opened 2020-04-09
[TopBar] Show secondary menu on small screens

dfmcphee:

Currently we hide the `secondaryMenu` in the top bar on small screens. We need to use this for a feature we are working on, but it needs to show on small screens too.

This PR removes the styles on the secondary menu that was hiding it at a breakpoint. I also added an example to the `TopBar` readme.

This is what it looks like now:

Simulator Screen Shot - iPhone 11 Pro Max - 2020-04-13 at 11 13 25

## 

### How to 🎩

Look at the [example in storybook](http://localhost:6006/?path=/story/all-components-top-bar--top-bar-with-all-of-its-elements) on a small screen and make sure the menu still shows.

### 🎩 checklist

### DIFF
--- a/src/components/TopBar/README.md
+++ b/src/components/TopBar/README.md
@@ -149,11 +149,17 @@ Use to provide structure for the top of an application. Style the top bar compon
 ```jsx
 function TopBarExample() {
   const [isUserMenuOpen, setIsUserMenuOpen] = useState(false);
+  const [isSecondaryMenuOpen, setIsSecondaryMenuOpen] = useState(false);
   const [isSearchActive, setIsSearchActive] = useState(false);
   const [searchValue, setSearchValue] = useState('');
 
   const toggleIsUserMenuOpen = useCallback(
     () => setIsUserMenuOpen((isUserMenuOpen) => !isUserMenuOpen),
+    [],
+  );
+
+  const toggleIsSecondaryMenuOpen = useCallback(
+    () => setIsSecondaryMenuOpen((isSecondaryMenuOpen) => !isSecondaryMenuOpen),
     [],
   );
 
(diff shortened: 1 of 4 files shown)


dfmcphee:
> Just looks a bit off to me

Yea it does doesn't it. I tweaked some of the margins for the search and menu activators. This is what it looks like now.

Screen Shot 2020-04-09 at 4 53 41 PM

dfmcphee:
🎩ed this in web and it looked good there too. Can I get another set of 👀s on this?

src/components/TopBar/README.md
```diff
@@ -224,10 +230,30 @@ function TopBarExample() {
     />
   );
 
+  const secondaryMenuMarkup = (
+    <TopBar.Menu
+      activatorContent={
+        <span>
+          <Icon source={QuestionMarkMajorTwotone} color="white" />
```
kyledurand, line 237: This won't be visible on white top bars

src/components/TopBar/README.md
```diff
@@ -224,10 +230,30 @@ function TopBarExample() {
...
+          <VisuallyHidden>Secondary menu</VisuallyHidden>
```
kyledurand, line 238: Kinda curious why you went with visually hidden over the accessibility label on Icon.

dfmcphee: I actually prefer the accessibility label but for some reason that was still causing the accessibility tests to fail.

dfmcphee:
> Looks good just one note about the icon being white because it won't show up when new design language is true

Good catch. I adjusted this to not need a color set on the icon component and instead apply the correct color in the CSS.

Here are some examples:

Screen Shot 2020-04-13 at 3 53 34 PM
Screen Shot 2020-04-13 at 3 53 25 PM
Screen Shot 2020-04-13 at 3 50 47 PM

src/components/TopBar/TopBar.scss
```diff
@@ -128,7 +129,8 @@ $context-control-expand-after: 1400px;
 }
 
 .SecondaryMenu {
-  @include breakpoint-before(layout-width(page-with-nav), false) {
-    display: none;
+  svg {
+    color: transparent;
+    fill: var(--p-icon, var(--top-bar-color));
   }
```
kyledurand, line 135: We could use recolor-icon for this
```suggestion
  @include recolor-icon(var(--p-icon, var(--top-bar-color)), transparent);
```

PR #2918, opened 2020-04-15
[Autocomplete.Combobox] Convert to functional component

genevieveluyt:

The Autocomplete component causes the following warning if used with server-side rendering:
```
Warning: Prop `aria-controls` did not match. Server: "ComboBox2" Client: "ComboBox1"
```

This is a known problem: #1761

Although this PR does not directly fix that problem, converting the Combobox to a functional component and having it use the `useUniqueId` hook is a prerequisite for fixing the automatic ID generation in the future.

Converts the Autocomplete Combobox component into a functional component and uses the `useUniqueId` hook.

## 

### How to 🎩

Not easily testable without server side rendering 🤔 

<details>

```jsx

import {Page, Autocomplete} from '../src';

  const [input, setInput] = useState('');
  const [selected, setSelected] = useState<string[]>([]);

  const textFieldMarkup = (
    <Autocomplete.TextField
      label="Textfield"
      value={input}
      onChange={setInput}
    />
  );
  const options = ['one', 'two', 'three'].map((item) => {
    return {value: item, label: item};
  });
  return (
    <Page title="Playground">
      <Autocomplete
        options={options}
        selected={selected}
        onSelect={setSelected}
        textField={textFieldMarkup}
        allowMultiple
      />
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Autocomplete/components/ComboBox/ComboBox.tsx
+++ b/src/components/Autocomplete/components/ComboBox/ComboBox.tsx
@@ -60,13 +60,13 @@ export function ComboBox({
   const [selectedOption, setSelectedOption] = useState<
     OptionDescriptor | ActionListItemDescriptor | undefined
   >(undefined);
-  const [selectedIndex, setSelectedIndex] = useState<number>(-1);
-  const [selectedOptions, setSelectedOptions] = useState<string[]>(selected);
+  const [selectedIndex, setSelectedIndex] = useState(-1);
+  const [selectedOptions, setSelectedOptions] = useState(selected);
   const [navigableOptions, setNavigableOptions] = useState<
     (OptionDescriptor | ActionListItemDescriptor)[]
   >([]);
-  const [popoverActive, setPopoverActive] = useState<boolean>(false);
-  const [popoverWasActive, setPopoverWasActive] = useState<boolean>(false);
+  const [popoverActive, setPopoverActive] = useState(false);
+  const [popoverWasActive, setPopoverWasActive] = useState(false);
 
   const id = useUniqueId('ComboBox', idProp);
 
--- a/src/components/Autocomplete/components/ComboBox/ComboBox.tsx
+++ b/src/components/Autocomplete/components/ComboBox/ComboBox.tsx
@@ -72,10 +72,7 @@ export function ComboBox({
   const id = useUniqueId('ComboBox', idProp);
 
   const getActionsWithIds = useCallback(
-    (
-      actions: ActionListItemDescriptor[],
-      before?: boolean,
-    ): ActionListItemDescriptor[] => {
+    (actions: ActionListItemDescriptor[], before?: boolean) => {
       if (before) {
         return navigableOptions.slice(0, actions.length);
       }
@@ -83,10 +80,6 @@ export function ComboBox({
     },
     [navigableOptions],
   );
-
-  const handlePopoverClose = useCallback(() => {
-    forcePopoverActiveFalse();
-  }, [forcePopoverActiveFalse]);
 
   const visuallyUpdateSelectedOption = useCallback(
     (
@@ -193,14 +186,6 @@ export function ComboBox({
     [allowMultiple, selectOptions, selected],
   );
 
-  const handleDownArrow = useCallback(() => {
-    selectNextOption();
-  }, [selectNextOption]);
-
-  const handleUpArrow = useCallback(() => {
-    selectPreviousOption();
-  }, [selectPreviousOption]);
-
   const handleEnter = useCallback(
     (event: KeyboardEvent) => {
       if (event.keyCode !== Key.Enter) {
@@ -219,10 +204,6 @@ export function ComboBox({
     },
     [handleSelection, navigableOptions, popoverActive, selectedIndex],
   );
-
-  const handleFocus = useCallback(() => {
-    forcePopoverActiveTrue();
-  }, [forcePopoverActiveTrue]);
 
   const handleBlur = useCallback(() => {
     forcePopoverActiveFalse();
@@ -333,18 +314,24 @@ export function ComboBox({
         aria-owns={id}
         aria-controls={id}
         aria-haspopup
-        onFocus={handleFocus}
+        onFocus={forcePopoverActiveTrue}
         onBlur={handleBlur}
         tabIndex={0}
       >
-        <KeypressListener keyCode={Key.DownArrow} handler={handleDownArrow} />
-        <KeypressListener keyCode={Key.UpArrow} handler={handleUpArrow} />
+        <KeypressListener keyCode={Key.DownArrow} handler={selectNextOption} />
+        <KeypressListener
+          keyCode={Key.UpArrow}
+          handler={selectPreviousOption}
+        />
         <EventListener event="keydown" handler={handleEnter} />
-        <KeypressListener keyCode={Key.Escape} handler={handlePopoverClose} />
+        <KeypressListener
+          keyCode={Key.Escape}
+          handler={forcePopoverActiveFalse}
+        />
         <Popover
           activator={textField}
           active={popoverActive}
-          onClose={handlePopoverClose}
+          onClose={forcePopoverActiveFalse}
           preferredPosition={preferredPosition}
           fullWidth
           preventAutofocus
--- a/src/components/Autocomplete/components/ComboBox/ComboBox.tsx
+++ b/src/components/Autocomplete/components/ComboBox/ComboBox.tsx
@@ -58,9 +58,6 @@ export function ComboBox({
   onSelect,
   onEndReached,
 }: ComboBoxProps) {
-  const [selectedOption, setSelectedOption] = useState<
-    OptionDescriptor | ActionListItemDescriptor | undefined
-  >(undefined);
   const [selectedIndex, setSelectedIndex] = useState(-1);
   const [selectedOptions, setSelectedOptions] = useState(selected);
   const [navigableOptions, setNavigableOptions] = useState<
@@ -107,7 +104,6 @@ export function ComboBox({
   );
 
   const resetVisuallySelectedOptions = useCallback(() => {
-    setSelectedOption(undefined);
     setSelectedIndex(-1);
     navigableOptions.forEach((option) => {
       option.active = false;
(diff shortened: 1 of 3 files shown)


src/components/Autocomplete/components/ComboBox/ComboBox.tsx
```diff
@@ -53,427 +41,374 @@ export interface ComboBoxProps {
...
+  );
 
-    this.setState({
-      navigableOptions,
-    });
-  }
+  const resetVisuallySelectedOptions = useCallback(() => {
+    setSelectedOption(undefined);
+    setSelectedIndex(-1);
+    navigableOptions &&
```
BPScott, line 118: navigableOptions is always an array according to its type, so I don't think you need to perform this check.

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
```diff
@@ -53,427 +41,374 @@ export interface ComboBoxProps {
...
+      setSelectedIndex(newOptionIndex);
+    },
+    [navigableOptions, selectedOption, visuallyUpdateSelectedOption],
+  );
 
-  getComboBoxId(): string {
-    if (this.state && this.state.comboBoxId) {
-      return this.state.comboBoxId;
+  const selectNextOption = useCallback(() => {
+    if (!navigableOptions || navigableOptions.length === 0) {
```
BPScott, line 143: it looks like navigableOptions is always an array so I don't think you need the first check as it will always be truthy

```suggestion
    if (navigableOptions.length === 0) {
```


You perform this check in a bunch of places, I think they can all be simplified.

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
```diff
@@ -53,427 +41,374 @@ export interface ComboBoxProps {
...
-    this.setState({
-      navigableOptions,
-    });
-  }
+  const resetVisuallySelectedOptions = useCallback(() => {
+    setSelectedOption(undefined);
+    setSelectedIndex(-1);
+    navigableOptions &&
+      navigableOptions.forEach(
+        (option: OptionDescriptor | ActionListItemDescriptor) => {
```
BPScott, line 120: Typescript should be able to infer the type for the function argument based upon the shape of navigableOptions so you shouldn't need to explicitly specify it

```suggestion
        (option) => {
```

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
```diff
@@ -53,427 +41,374 @@ export interface ComboBoxProps {
...
-    const {onSelect, allowMultiple} = this.props;
-    selected && onSelect(selected);
-    if (!allowMultiple) {
-      this.resetVisuallySelectedOptions();
-      this.setState({popoverActive: false, popoverWasActive: false});
-    }
-  };
+  const handleDownArrow = useCallback(() => {
+    selectNextOption();
+    handlePopoverOpen;
```
BPScott, line 207: This is just a reference to a function instead of calling the function. So this line does nothing (in the same way a line containing just `"hi"` does nothing)

Seems like this was a bug in the old code too - if it everything worked before then this can be deleted

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
```diff
@@ -53,427 +41,374 @@ export interface ComboBoxProps {
...
-      this.selectOptionAtIndex(newOptions.indexOf(selectedOption));
-    } else if (selectedIndex > newOptions.length - 1) {
-      this.resetVisuallySelectedOptions();
-    } else {
-      this.selectOptionAtIndex(selectedIndex);
-    }
-  };
+  const handleUpArrow = useCallback(() => {
+    selectPreviousOption();
+    handlePopoverOpen;
```
BPScott, line 212: This is just a reference to a function instead of calling the function. So this line does nothing (in the same way a line containing just `"hi"` does nothing)

Seems like this was a bug in the old code too - if it everything worked before then this can be deleted

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
genevieveluyt: I was wondering about this too 😄

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
```diff
@@ -53,427 +41,374 @@ export interface ComboBoxProps {
...
+  const [selectedOption, setSelectedOption] = useState<
+    OptionDescriptor | ActionListItemDescriptor | undefined
+  >(undefined);
+  const [selectedIndex, setSelectedIndex] = useState<number>(-1);
+  const [selectedOptions, setSelectedOptions] = useState<string[]>(selected);
+  const [navigableOptions, setNavigableOptions] = useState<
+    (OptionDescriptor | ActionListItemDescriptor)[]
+  >([]);
+  const [popoverActive, setPopoverActive] = useState<boolean>(false);
+  const [popoverWasActive, setPopoverWasActive] = useState<boolean>(false);
```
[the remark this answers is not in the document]
genevieveluyt: I removed the previous state and it didn't break any tests. I'm not sure if that logic was there for a specific case not covered by tests though 🤔

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
```diff
@@ -53,467 +42,337 @@ export interface ComboBoxProps {
...
+    setFalse: forcePopoverActiveFalse,
+  } = useToggle(false);
+
+  const id = useUniqueId('ComboBox', idProp);
+
+  const getActionsWithIds = useCallback(
+    (
+      actions: ActionListItemDescriptor[],
+      before?: boolean,
+    ): ActionListItemDescriptor[] => {
```
BPScott, line 78: TS can infer the return type fine, no need to be explicit

```suggestion
    ) => {
```

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
```diff
@@ -53,467 +42,337 @@ export interface ComboBoxProps {
...
-    navigableOptions: [],
-    popoverActive: false,
-    popoverWasActive: false,
-  };
-
-  componentDidMount() {
-    const {options, actionsBefore, actionsAfter} = this.props;
-    const comboBoxId = this.getComboBoxId();
-    let navigableOptions: (OptionDescriptor | ActionListItemDescriptor)[] = [];
+  const handlePopoverClose = useCallback(() => {
```
BPScott, line 87: `forcePopoverActiveFalse` is already a function wrapped in a useCallback, it doesn't seem very useful to repeat that again

Down below you can

```difff
- <KeypressListener keyCode={Key.Escape} handler={handlePopoverClose} />
+ <KeypressListener keyCode={Key.Escape} handler={forcePopoverActiveFalse} />
```

And delete this variable as it is no longer used

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
```diff
@@ -53,467 +42,337 @@ export interface ComboBoxProps {
...
+    },
+    [allowMultiple, selectOptions, selected],
+  );
 
-  private handleBlur = () => {
-    this.setState({popoverActive: false, popoverWasActive: false}, () => {
-      this.resetVisuallySelectedOptions();
-    });
-  };
+  const handleDownArrow = useCallback(() => {
```
BPScott, line 196: `selectNextOption` is already a function wrapped in a useCallback so repeating that doesn't seem that useful

Down below, replace the following and delete this

```diff
- <KeypressListener keyCode={Key.DownArrow} handler={handleDownArrow} />
+ <KeypressListener keyCode={Key.DownArrow} handler={selectNextOption} />
```

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
```diff
@@ -53,467 +42,337 @@ export interface ComboBoxProps {
...
+    selectNextOption();
+  }, [selectNextOption]);
 
-  private handleClick = () => {
-    !this.state.popoverActive && this.setState({popoverActive: true});
-  };
+  const handleUpArrow = useCallback(() => {
```
BPScott, line 200: `selectPreviousOption` is already a function wrapped in a useCallback so repeating that doesn't seem that useful

Down below, replace the following and delete this

```diff
- <KeypressListener keyCode={Key.UpArrow} handler={handleUpArrow} />
+ <KeypressListener keyCode={Key.UpArrow} handler={selectPreviousOption} />
```

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
```diff
@@ -53,467 +42,337 @@ export interface ComboBoxProps {
...
 
-  private selectOptions = (selected: string[]) => {
-    const {onSelect, allowMultiple} = this.props;
-    selected && onSelect(selected);
-    if (!allowMultiple) {
-      this.resetVisuallySelectedOptions();
-      this.setState({popoverActive: false, popoverWasActive: false});
-    }
-  };
+  const handleFocus = useCallback(() => {
```
BPScott, line 223: As above, forcePopoverActiveTrue is already afunction wrapped in a callback, replace usages of `handleFocus` with `forcePopoverActiveTrue` and this can be removed

BPScott:
Just gave this a click around in storybook and seems good to me. I'd appreciate some extra eyes though as I'm not that familiar with the Comobox's behaviour. @dfmcphee / @chloerice would you be able to take some time next week to give this a review too?

BPScott:
Oh and a little expectation setting:  This PR alone won't solve the automatic id mismatches between server and client, but it is a prerequisite of that future fix, which will be inside useUniqueId() - either through adding functionality ourselves or by leveraging https://github.com/facebook/react/pull/17322

genevieveluyt:
Ah ok I misunderstood. That means I'll still need to set the ID manually for the time being?

chloerice:
Hey @genevieveluyt, thanks for working on this! I'll tophat/review this by EOD Wednesday 😊

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
```diff
@@ -53,356 +42,85 @@ export interface ComboBoxProps {
...
-  getActionsWithIds(
-    actions: ActionListItemDescriptor[],
-    before?: boolean,
-  ): ActionListItemDescriptor[] {
-    const {navigableOptions} = this.state;
-
-    if (before) {
-      return navigableOptions.slice(0, actions.length);
-    } else {
+export function ComboBox({
```
chloerice, line 45:
```suggestion

export function ComboBox({
```

BPScott:
Thanks @genevieveluyt!

PR #2928, opened 2020-04-17
Update the background menu color when no theme is set.

CSilivestru:
Currently displays white on white when no colour theme is provided to the the top bar. The defaults provided didn't seem correct so I'm updating them :).

Addresses issue https://github.com/Shopify/destinations/issues/760.

Before:
image

After:
image

image

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

* ~~[ ] Tested for [accessibility](https://github.com/Shopify/polaris-react/blob/master/documentation/Accessibility%20testing.md)~~N/A
* ~~[ ] [Tophatted documentation](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting%20documentation.md) changes in the style guide~~

### DIFF
--- a/src/components/TopBar/README.md
+++ b/src/components/TopBar/README.md
@@ -322,9 +322,11 @@ function TopBarExample() {
   const theme = {
     colors: {
       topBar: {
-        background: '#357997',
-        backgroundLighter: '#6192a9',
-        color: '#FFFFFF',
+        background: '#fff',
+        backgroundLighter: '#F4F6F8',
+        backgroundDarker: '#DFE3E8',
+        border: '#C4CDD5',
+        color: '#212B36',
       },
     },
     logo: {
@@ -370,6 +372,7 @@ function TopBarExample() {
       onChange={handleSearchChange}
       value={searchValue}
       placeholder="Search"
+      showFocusBorder
     />
   );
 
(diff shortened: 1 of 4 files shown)


CSilivestru:
I made a change to include a var within a var to have polaris attempt to use the background lighter theme override if provided.

The reason this didn't work in earlier trials is because the css variable fallback is treated as plain text and doesn't invoke any functions -- see [here](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties#Custom_property_fallback_values).

Because of this, I had to remove the hover variable and place it directly in as the fallback. I'm more than happy to discuss alternatives to this!

src/components/TopBar/README.md
```diff
@@ -370,6 +372,7 @@ function TopBarExample() {
       onChange={handleSearchChange}
       value={searchValue}
       placeholder="Search"
+      showFocusBorder
```
alex-page, line 375: To make it easier for the plus team, I updated the example to use their white theme.

src/components/TopBar/components/Menu/Menu.scss
```diff
@@ -2,11 +2,7 @@
 
 $activator-variables: (
   min-width: rem(36px),
-  border-left: 1px solid rgba(color('black'), 0.42),
   transition: background-color 0.1s,
-  focus-background-color: rgba(color('white'), 0.16),
-  hover-background-color: rgba(color('white'), 0.08),
-  active-background-color: rgba(color('black'), 0.28),
```
alex-page, line 9: These values are used once and blocked multiple theme overrides.

alex-page:
@AlexBlaise you should be able to TopHat this when the tests pass there will be a Chroma check that will have a link to an example.

PR #2936, opened 2020-04-20
[ResourceItem] Remove children from the equality check in shouldComponentUpdate

### LINKED ISSUE #2164: [ResourceItem] Performance issue on shouldComponentUpdate
## Issue summary

When using the `<ResourceList />` component, and passing a custom child component to the `<ResourceList.Item />` that has a loading state attached to an API call, `shouldComponentUpdate` does some [deep comparisons on the props](https://github.com/Shopify/polaris-react/blob/master/src/components/ResourceItem/ResourceItem.tsx#L124) to determine if the item should re-render, which causes a blocking of the rendering of the page for very long periods of time for the specific use-case.

## Expected behavior

`shouldComponentUpdate` optimizes its comparison of possible complex child components.

Deep comparisons in `shouldComponentUpdate` aren't ideal, see [related issue](https://github.com/facebook/react/issues/11107)

#### Use-case example 

This was found in the Shopify `web` admin, blocking for **13 seconds** in this example - locally this comparison is much faster, but still observable.

| Performance profile example of the use-case in production  |
|---|
| deepComparisonProduction |

| `isEqual` starts to go very deep in React `FiberNode`s |
|---|
| propsDeepComparison|

This is difficult because the lag is only very noticeable on the production code of Shopify `web`.

However, here's a sandbox that tries to mimic the same behavior:

https://codesandbox.io/embed/festive-oskar-xq5xz?fontsize=14

| The performance profile of that sandbox looking at the `isEqual` check  |
|---|
| Screen Shot 2019-09-18 at 2 33 24 PM|

Related issue for Shopify admin `web` is https://github.com/Shopify/store/issues/9546

Removing this props deep equal check from the Shopify web production build resolves this issue.

## Specifications

```bash

  System:
    OS: macOS 10.14.5
    CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
    Memory: 125.98 MB / 16.00 GB
    Shell: Unknown
  Binaries:
    Node: 10.13.0 - ~/.nvm/versions/node/v10.13.0/bin/node
    Yarn: 1.15.2 - /usr/local/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v10.13.0/bin/npm
  Browsers:
    Chrome: 76.0.3809.132
    Firefox: 68.0.2
    Safari: 12.1.1
  npmPackages:
    react: ^16.8.6 => 16.8.6
    react-dom: ^16.8.6 => 16.8.6

```

chloerice:

Fixes #2164 

The resource item component does a deep equality check of previous and next props in the `shouldComponentUpdate` lifecycle method. This causes significant performance issues in production because `children` are included in that check. 

The ideal way to fix this forward would be to rewrite `ResourceItem` as a functional component and 🔥`shouldComponentUpdate` entirely. Even more ideally, it shouldn't be handling its own selected state but instead rely on a `selected` boolean prop. But we don't have cycles for that effort right now. 

This PR hot fixes the problem by stripping `children` out of the check to reduce complexity of the comparison down to only the explicit props. 

- [x] Keeping this a draft until I've tested in a production environment, as I need to verify whether the deep comparison remains laggy due to the fact that the optional `media` prop expects either an `Avatar` or a `Thumbnail`.

## 

### How to 🎩

To tophat this change, I made a release candidate and pushed the [`web` upgrade branch to Web staging](https://shipit.shopify.io/shopify/web/staging/deploys/974701) to enable performance analysis in a production environment.

| |  Before | After  |
|---|---|---|
|Render time| 5.2sec  | 2.2sec  |
|Browser freezing| Yes  | No  |
|`shouldComponentUpdate` calls / highest call time| >1000 / 2.2ms|5 / 0.5ms|

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -8,6 +8,8 @@
 
 ### Bug fixes
 
+- Fixed performance of `ResourceItem` due to inclusion of `children` in deep prop comparison within `shouldComponentUpdate` ([#2936](https://github.com/Shopify/polaris-react/pull/2936))
+
 ### Documentation
 
 ### Development workflow
--- a/src/components/ResourceItem/ResourceItem.tsx
+++ b/src/components/ResourceItem/ResourceItem.tsx
@@ -111,15 +111,19 @@ class BaseResourceItem extends React.Component<CombinedProps, State> {
 
   shouldComponentUpdate(nextProps: CombinedProps, nextState: State) {
     const {
+      children: nextChildren,
       context: {selectedItems: nextSelectedItems, ...restNextContext},
       ...restNextProps
     } = nextProps;
+
     const {
+      children,
       context: {selectedItems, ...restContext},
       ...restProps
     } = this.props;
 
     const nextSelectMode = nextProps.context.selectMode;
+
     return (
       !isEqual(this.state, nextState) ||
       this.props.context.selectMode !== nextSelectMode ||


jineshshah36:
Hi, we merged this change it broke a bunch of our `ResourceItem` components. I have put together a codesandbox [here](https://codesandbox.io/s/flamboyant-river-8v1zm?file=/App.js) which exemplifies the problem.

Essentially, if you attempt to update the children of the resource item asynchronously without making any additional prop changes, it will never re-render. In the example I included, the author is "fetched" asynchronously, and when we attempt to re-render with the author info, it doesn't.

I believe excluding children is simply a cure for the symptom, but the real problem is that the code is using deep equals instead of shallow equals. Is there a reason for the deep equals? Shouldn't all props be immutable, therefore not need a deep equals comparison?

PR #2942, opened 2020-04-24
[Page] Adding additionalNavigation prop

dfmcphee:

For a project I am working on currently, we need to make better use of the space between the breadcrumbs and pagination/actions menu. Currently we have some hacks with absolute positioning, but I would rather this be a part of the page component so we can make it more reliable and also re-arrange if needed (probably in the new design language too). My current implementation gets the wrong tab order as well because I need to place the elements inside the page content instead of in the proper place in the header.

This adds a new `additionalNavigation` prop to `<Page>` that accepts a React Node and renders it in the correct space.

This is what it looks like with an avatar in that spot:

Screen Shot 2020-04-24 at 12 21 12 PM

Screen Shot 2020-04-24 at 12 21 21 PM

## 

### How to 🎩

I added an example to our `Page with all header elements` example in the README

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -6,6 +6,8 @@
 
 ### Enhancements
 
+- Added `additionalNavigation` prop to `Page` ([#2942](https://github.com/Shopify/polaris-react/pull/2942))
+
 ### Bug fixes
 
 - Fixed performance of `ResourceItem` due to inclusion of `children` in deep prop comparison within `shouldComponentUpdate` ([#2936](https://github.com/Shopify/polaris-react/pull/2936))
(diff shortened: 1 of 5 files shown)


dfmcphee:
Thanks for the feedback. I adjusted the spacing a bit. This is what it looks like now:

Screen Shot 2020-04-24 at 1 20 05 PM

Simulator Screen Shot - iPhone 8 - 2020-04-24 at 13 19 50

dfmcphee:
> . I'm wondering in the future if we can be a little bit more clever about the positioning of the rollup activator so it's not positioned absolutely and can wrap if need be.

Yeah, that's not a bad idea. I'm not sure if wrapping is best there or what, but would be nice to handle it better.

PR #2944, opened 2020-04-26
KonamiCode

tmlayton:
Saturday night in quarantine... 🙂

<details>
<summary>Try it out in the Playground</summary>

```tsx
import React from 'react';

import {Page, KonamiCode} from '../src';

  return (
    <Page title="Playground">
      <KonamiCode
        handler={() => {
          // eslint-disable-next-line no-alert
          alert('Hello, world!');
        }}
      />
    </Page>
  );
}
```

</details>

- [x] I’ve 🎩’d

### DIFF
--- a/src/components/KonamiCode/KonamiCode.tsx
+++ b/src/components/KonamiCode/KonamiCode.tsx
@@ -1,0 +1,54 @@
+import {useEffect, useState} from 'react';
+import {
+  addEventListener,
+  removeEventListener,
+} from '@shopify/javascript-utilities/events';
+
+import {Key} from '../../types';
+
+export interface KonamiCodeProps {
+  handler(event: KeyboardEvent): void;
+}
+
+export const KONAMI_CODE = [
+  Key.UpArrow,
+  Key.UpArrow,
+  Key.DownArrow,
+  Key.DownArrow,
+  Key.LeftArrow,
+  Key.RightArrow,
+  Key.LeftArrow,
+  Key.RightArrow,
+  Key.KeyB,
+  Key.KeyA,
+];
+
+export function KonamiCode({handler}: KonamiCodeProps) {
+  const keyEvent = 'keydown';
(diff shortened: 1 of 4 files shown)


BPScott:
At some point it'd be ace to get rid of our own key handler stuff and instead make use of https://github.com/Shopify/quilt/blob/master/packages/react-shortcuts, but that's dreams for another day, like killing our own translation bits. They even do "when you press these keys in combination" logic

Using components to drive behaviour is so 2019, this should totally be a hook ;)

```js
useKonamiCode(() => {
  console.log('you pressed the buttons!';
})
```

src/components/KonamiCode/KonamiCode.tsx
```diff
@@ -0,0 +1,54 @@
...
+      } else {
+        setPosition(position + 1);
+      }
+    } else {
+      setPosition(0);
+    }
+  };
+
+  useEffect(() => {
+    addEventListener(document, keyEvent, handleKeyEvent);
```
BPScott, line 47: the addEventListener from javascript-utilities only acts as a wrapper to handle browsers that don't support the `once` and `passive` options, all of which we no longer support (ios safari 9 was the last hangout). Save a dependency and do use `document.addEventListener` and `document.removeEventListener` directly :p

tmlayton:
> Using components to drive behaviour is so 2019, this should totally be a hook ;)

You should do it!

BPScott:
> > Using components to drive behaviour is so 2019, this should totally be a hook ;)
> 
> You should do it!


Boom.

```js
import useShortcut from '@shopify/react-shortcuts';

function useKonamiCode(onMatch, options) {
  return useShortcut(
    ['ArrowUp', 'ArrowUp', 'ArrowDown', 'ArrowDown', 'ArrowLeft', 'ArrowRight', 'ArrowLeft', 'ArrowRight', 'B', 'A'],
    onMatch,
    options
  );
}
```

PR #2949, opened 2020-04-27
[Banner] Fix secondary action only rendering if action also set

### LINKED ISSUE #2946: Banner secondaryAction not rendering if action isn't present
https://codesandbox.io/s/misty-mountain-3fyo0?file=/App.js

Is it by design?

chloerice:

Fixes #2946 

## 

### How to 🎩

<details>

```jsx
import React from 'react';

import {Page, Banner} from '../src';

  return (
    <Page title="Playground">
      <Banner
        title="Some of your product variants are missing weights"
        status="warning"
        secondaryAction={{content: 'Edit variant weights'}}
        onDismiss={() => {}}
      >
        <p>
          Add weights to show accurate rates at checkout and when buying
          shipping labels in Shopify.
        </p>
      </Banner>
    </Page>
  );
}

```

</details>

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -19,6 +19,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - **`Button`:** `loading` no longer sets the invalid `role="alert"` ([#3590](https://github.com/Shopify/polaris-react/pull/3590))
 - Removed `tabIndex=-1` from `Popover` when `preventAutoFocus` is true ([#3595](https://github.com/Shopify/polaris-react/pull/3595))
 - Fixed `Modal` header border color ([#3616](https://github.com/Shopify/polaris-react/pull/3616))
+- Fixed `Banner` `secondaryAction` only rendering if `action` is set ([#2949](https://github.com/Shopify/polaris-react/pull/2949))
 
 ### Documentation
 
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -105,40 +105,29 @@ export const Banner = forwardRef<BannerHandles, BannerProps>(function Banner(
     );
   }
 
-  let actionMarkup;
-  if (action) {
-    if (newDesignLanguage) {
-      actionMarkup = (
-        <div className={styles.Actions}>
-          <ButtonGroup>
-            <div className={styles.PrimaryAction}>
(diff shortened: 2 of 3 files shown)


UNRELEASED.md
```diff
@@ -10,7 +10,7 @@
 
 ### Bug fixes
 
-- Fixed performance of `ResourceItem` due to inclusion of `children` in deep prop comparison within `shouldComponentUpdate` ([#2936](https://github.com/Shopify/polaris-react/pull/2936))
+- Fixed `Banner` `secondaryAction` only rendering if `action` is set ([#2949](https://github.com/Shopify/polaris-react/pull/2949))
```
416serg, line 13: the new change should be a separate line item, instead of overriding the one that is there

src/components/Banner/Banner.tsx
```diff
@@ -171,20 +171,25 @@ export class Banner extends React.PureComponent<BannerProps, State> {
...
-              <div className={styles.Actions}>
-                <ButtonGroup>
-                  <div className={styles.PrimaryAction}>
-                    {buttonFrom(action, {outline: true, size: buttonSizeValue})}
-                  </div>
-                  {secondaryActionMarkup}
-                </ButtonGroup>
-              </div>
-            ) : null;
+            const actionMarkup =
```
416serg, line 184: this probably needs a separate test case to make sure it renderers

416serg (dismissed):
Left a few notes, thanks for picking this up!

src/components/Banner/Banner.tsx
```diff
@@ -106,38 +106,29 @@ export const Banner = forwardRef<BannerHandles, BannerProps>(function Banner(
...
+    ) : null;
 
-            {secondaryAction && (
-              <SecondaryActionFrom action={secondaryAction} />
-            )}
-          </ButtonGroup>
-        </div>
-      );
-    }
+    actionMarkup = (
```
kyledurand, line 124: We don't really need this to be a let anymore, we can just do:
```
const actionMarkup = action || secondaryAction ? (<div className={styles.Actions}>...etc) : null;
```

Then we can also get rid of the if statement above. Might be slightly cleaner 🤷‍♂️

kyledurand (commented):
Looks like a good fix just one comment about removing the `let actionMarkup` and removing the if statement if possible

kyledurand (approved):

PR #2957, opened 2020-04-30
[ResourceList] Trunctate long sortOption label

carolopolo:

Fixes https://github.com/Shopify/store/issues/14834

No matter the viewport size, it truncates the sortOption label if it doesn't fit. Previously, we were only truncating for mobile/small screens. This PR addresses running into the same problem on various screens.

## 

### How to 🎩

<details>

```jsx

import {Avatar, Card, ResourceItem, ResourceList, TextStyle} from '../src';

  const [sortValue, setSortValue] = useState('DATE_MODIFIED_DESC');

  const resourceName = {
    singular: 'customer',
    plural: 'customers',
  };

  const items = [
    {
      id: 341,
      url: 'customers/341',
      name: 'Mae Jemison',
      location: 'Decatur, USA',
    },
    {
      id: 256,
      url: 'customers/256',
      name: 'Ellen Ochoa',
      location: 'Los Angeles, USA',
    },
  ];

  return (
    <Card>
      <ResourceList
        selectable
        selectedItems={[]}
        resourceName={resourceName}
        items={items}
        renderItem={renderItem}
        sortValue={sortValue}
        sortOptions={[
          {
            label:
              'Last shrimpy shrimp created (ascending shrimp) shrimp shrimp shrimp shrimp shrimp',
            value: 'DATE_MODIFIED_DESC',
          },
          {label: 'Oldest update', value: 'DATE_MODIFIED_ASC'},
        ]}
        onSortChange={(selected) => {
          setSortValue(selected);
          console.log(`Sort option changed to ${selected}.`);
        }}
      />
    </Card>
  );

  function renderItem(item) {

    return (
      <ResourceItem
        id={id}
        url={url}
        media={media}
      >
        <h3>
        </h3>
        <div>{location}</div>
      </ResourceItem>
    );
  }
}


```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/ResourceList/ResourceList.scss
+++ b/src/components/ResourceList/ResourceList.scss
@@ -118,11 +118,11 @@ $item-wrapper-loading-height: rem(64px);
   }
 }
 
-// Force long sort options to be truncated
-.SortWrapper,
-.SortWrapper > * {
-  @include breakpoint-before(resource-list(breakpoint-small)) {
-    overflow: hidden;
+.SortWrapper {
+  @include layout-flex-fix;
+
+  > * {
+    max-width: 100%;
   }
 }
 
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -6,6 +6,8 @@
 
 ### Enhancements
 
+- Truncated long sort options in `ResourceList` ([#2957](https://github.com/Shopify/polaris-react/pull/2957)
+
 ### Bug fixes
 
 ### Documentation


dleroux:
Looks like the focus ring gets cut off with the overflow hidden:

Screen Shot 2020-05-04 at 9 55 53 AM
Screen Shot 2020-05-04 at 9 54 59 AM

src/components/ResourceList/ResourceList.scss
```diff
@@ -119,13 +119,65 @@ $item-wrapper-loading-height: rem(64px);
...
+//   item: 10,
+//   focused: 20,
+// );
+
+// .SortWrapper,
+// .SortWrapper > * {
+//   overflow: hidden;
+// }
+
+.SortWrapper {
```
carolopolo, line 144: @beefchimi this is a possible solution... tested on chrome, firefox & safari 

but would love some input 🥬

beefchimi: Okay, so this can actually be fixed rather easily 😄 

You can replace ALL of your changed code in this file with:

```scss
.SortWrapper {
  @include layout-flex-fix;

  // Required to apply styles to the <Labelled /> component
  > * {
    max-width: 100%;
  }
}
```

That is the most simple solution. Two things to potentially alter:

**1. Apply "layout fix" to all children**
Personally, I would apply that `layout-flex-fix` mixin to ALL `children` of `.HeaderContentWrapper`. So, you could instead do:

```scss
.HeaderTitleWrapper,
.CheckableButtonWrapper,
.AlternateToolWrapper,
.SortWrapper,
.SelectButtonWrapper {
  @include layout-flex-fix;
}

.SortWrapper {
  > * {
    max-width: 100%;
  }
}
```

**2. Try to avoid reaching into other components to override styles**
The `> *` selector is specifically targeting the `.Labelled` direct child. It would be preferable if we did not have to apply this snowflake override style... so I'm wondering if `max-width: 100%` can live comfortably on the `.Labelled` selector within `Labelled.scss`. I think its worth experimenting with that change to see if it causes issues any where else `<Labelled />` is used.

**Explanation**
We need to utilize this `layout-flex-fix` mixin on direct children ("flex items") of a `display: flex` parent in order to get truncation to work correctly. This is a very common problem/solution! All that mixin is doing is applying the following styles:

```scss
min-width: 0;
max-width: 100%;
```

This seems... silly, because those prop + values seem obvious and should be default behaviour.... right?!?! The short answer is: **flex box items will not shrink below their minimum content size**. So, if the minimum content size is computed as the full width of a sentence, it cannot be _below that width_ in order to truncate. In comes out handy "flex fix" styling to say _"well actually min-width can be `0`"_.

For a better explanation, our very own @dfmcphee wrote a blog post on it!
https://dfmcphee.com/flex-items-and-min-width-0/

carolopolo: thanks so much @beefchimi! really appreciate that. 

fyi for posterity's sake: 

adding the `max-width: 100%` on the `.Labelled` selector in `Labelled.scss` results in:
image

adding the `@include layout-flex-fix;` to all children of `.HeaderContentWrapper` results in: 
image

so i'm going with your very fist suggestion. thanks again for your input!

beefchimi: We _could_ improve this further with some tweaks to the other `HeaderContentWrapper`, mainly to that `SelectWrapper` so that it has a `min-width` of _at least_ the:
- `width` of a `checkbox` +
- `spacing()`

But just so other reviews know - the label of that "select checkbox" will truncate as well. Its just that Caroline's example is _soooo long_ that it goes beyond the text label and starts encroaching onto the `checkbox` input. Realistically, we probably wouldn't end up in the scenario.

@carolopolo would you mind opening an issue with your 2nd screenshot to track this? You can link to this comment + include any relevant context.

PR #2974, opened 2020-05-08
Create a single entrypoint for public sass API

BPScott:

Consuming applications currently have to reference two files to use our public sass API. Instead we should get people to use a single entry point now [recent (since Nov 2019) versions of sewing-kit no-longer hardcode opinions about what our public api endpoints are named](https://github.com/Shopify/sewing-kit/blob/master/migration-guide.md#01130).

Deprecate using foundation.scss and shared.scss to access our public
sass API. Instead consuming apps should import `_public-api.scss`.

By exposing a single entrypoint we make it slightly easier for consuming
apps to use our public API.

foundation.scss and shared.scss remain present for now to retain
compatability with sewing-kit versions older than 0.113.0 (released
Nov 2019), as older versions hardcoded those paths. They should be
deleted in Polaris v6, and v6 should require using sewing-kit >= 0.113.0

## 

### How to 🎩

Do a build, see that `styles/_public-api.scss` and `esnext/styles/_public-api.scss` are created, and note that they contain the same imports as `foundation.scss` and `shared.scss` combined.

### Who does this affect and how?

Importing `_public-api.scss` is functionally identical to importing `foundation.scss` and `shared.scss`.

#### Consuming applications using recent versions of sewing-kit (>=0.113.0)

Modern SK versions require consuming apps to explicitly configure usage of our public sass api if they wish to use it. [Here is how web does it](https://github.com/Shopify/web/blob/1f4e2ef39790b9002f6b6bd70fed801cf36384f7/sewing-kit.config.ts#L44-L45).

Instead of importing `esnext/styles/foundation.scss` and `esnext/styles/shared.scss`, they should import `esnext/styles/_public-api.scss` instead.

```diff
- join(polarisStyles, 'foundation.scss'),
- join(polarisStyles, 'shared.scss'),
+ join(polarisStyles, '_public-api.scss'),
```

####  Consuming applications using old versions of sewing-kit (< 0.113.0)

These old versions of sewing-kit implicitly configure the import of `esnext/styles/foundation.scss` and `esnext/styles/shared.scss` instead of making the App do it. As those two files continue to exist things will be fine until we remove those files in v6.

At some point before that they will need to update to a version of sewing-kit >=0.113.0 and then configure their autoImport settings per the [sewing-kit migration guide](https://github.com/Shopify/sewing-kit/blob/master/migration-guide.md#01130) where they should point to `esnext/styles/_public-api.scss`.

#### Consumers not using sewing-kit at all

Any other consumer is getting at our API by importing `styles/foundation.scss` and `styles/shared.scss`, they should import `styles/_public-api.scss` instead. (that's not a typo remember that scss will look for partial files prefixed with an underscore)

```diff 
- @import '~@shopify/polaris/styles/foundation';
- @import '~@shopify/polaris/styles/shared';
+ @import '~@shopify/polaris/styles/public-api';
```

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -24,3 +24,5 @@
 - Converted `ComboBox` to a functional component ([#2918](https://github.com/Shopify/polaris-react/pull/2918))
 
 ### Deprecations
+
+- Deprecated `styles/foundation.scss` and `styles/shared.scss` as entry points to the Polaris Sass public API. They have been replaced with a single file `styles/_public-api.scss`. By having a single entry point we make it a little easier for consuming applications to use our public API - you only need to import one file instead of two. Any references to these two files should be replaced with a reference to `_public-api.scss` which lives in the same folder. Consuming applications using sewing-kit should replace references to `esnext/styles/foundation.scss` and `esnext/styles/shared.scss` with a single reference to `esnext/styles/_public-api_.scss`. Note the API itself has not changed - only the mechanism by which you access it. ([#2974](https://github.com/Shopify/polaris-react/pull/2974))
--- a/config/rollup/plugins/styles.js
+++ b/config/rollup/plugins/styles.js
@@ -143,8 +143,8 @@ function generateMinifiedCss(sourceFilePath, css) {
  *   this base file.
  * - components.scss and the components folder: a suite of the compiled css for
  *  every component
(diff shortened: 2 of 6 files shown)


tmlayton:
Just a heads up I won’t have time to look at this until next week.

PR #2975, opened 2020-05-08
[Banner] Fix rounded corners on banners

tmlayton:

Rounded corners on banners were missing

A fallback to `inherit` doesn’t appear to work, also border-radius is different for in-context banner so putting those properties in the if-else.

### DIFF
--- a/src/components/Banner/Banner.scss
+++ b/src/components/Banner/Banner.scss
@@ -10,7 +10,6 @@ $secondary-action-horizontal-padding: 1.5 * spacing(tight);
 @mixin banner-attributes($highlight, $background, $in-page, $tint) {
   transition: box-shadow duration() easing();
   transition-delay: duration(fast);
-  border-radius: var(--p-border-radius-wide, inherit);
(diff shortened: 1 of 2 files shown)


src/components/Banner/Banner.scss
```diff
@@ -10,20 +10,25 @@ $secondary-action-horizontal-padding: 1.5 * spacing(tight);
...
   box-shadow: var(--p-banner-border, none);
 
   @media (-ms-high-contrast: active) {
     outline: 1px solid;
   }
 
   @if $in-page {
+    border-radius: var(
+      --p-border-radius-wide,
+      0 0 border-radius() border-radius()
```
kyledurand, line 22: I tophatted the banners in story book but didn't see a version where the top left and right didn't have a radius. Do we need an example for this as well?

tmlayton: This PR is to fix current banners. All the default (in-page) banners have square corners on top. In-context or in-card banners have all rounded corners.

Screen Shot 2020-05-11 at 11 07 22 AM

PR #2977, opened 2020-05-11
[EmptyState] Switched to ComplexAction as type for the EmptyState actions

felipeleusin:

While working on https://github.com/Shopify/email/pull/4656 we had the need to show a Loading indicator on the empty state primary action.

This switched the type for the Actions in the EmptyState from Action to ComplexAction.

Two questions: Do we want to add something on this to the Playground? Also, it could be just the LoadableAction but I used the ComplexAction because it was the typing for the button.

### 🎩 checklist


felipeleusin:
Cool. Going to update the changelog then

PR #2981, opened 2020-05-13
Update typescript to 3.9 and enable skipLibCheck

BPScott:

Keeping up to date. tyepscript 3.9 is hot off the presses. Speeding up builds

TS 3.9 offers a few speed improvements. Nothing huuuuuge, but it shaves 3
seconds of our build and type-check commands.

Enabling skipLibCheck (as set by our parent TS configs) meanwhile saves
an additional 10 seconds as we don't bother checking TS declaration files
that we don't use. Recently TS's init scripts were updated so now this
is set to true in newly creacted configs so I'm now more confident in
following suit.

On master: type-check took ~38s, build took ~52s
This commit: type-check took ~24.5s, build took ~37s

TypeScript 3.9 Release blog: https://devblogs.microsoft.com/typescript/announcing-typescript-3-9/
PR where TypeScript changed the value of skipLibCheck in new projects to be true: https://github.com/microsoft/TypeScript/pull/37808
More info on what skipLibCheck does: https://swatinem.de/blog/js-tooling/#explaining-skiplibcheck

### How to 🎩

Check build tests and type-check pass. Do a quick tophat in a consuming app.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -12,7 +12,11 @@
 
 ### Development workflow
 
+- Improved speed of type-check and build by enabling TypeScript's `skipLibCheck` option ([#2981](https://github.com/Shopify/polaris-react/pull/2981))
+
 ### Dependency upgrades
+
+- Updated TypeScript to 3.9.2 ([#2981](https://github.com/Shopify/polaris-react/pull/2981))
 
 ### Code quality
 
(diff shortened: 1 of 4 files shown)


UNRELEASED.md
```diff
@@ -12,8 +12,12 @@
 
 ### Development workflow
 
+- Speed up type-check and build by enabling TypeScript's `skipLibCheck` option ([#2981](https://github.com/Shopify/polaris-react/pull/2981))
+
 ### Dependency upgrades
 
+- Update TypeScript to 3.9.2 ([#2981](https://github.com/Shopify/polaris-react/pull/2981))
```
kaelig, line 19: Use past tense in change log entries

PR #2983, opened 2020-05-13
[ResourceList] Make EmptySearchResultWrapper responsive

kaelig:

Fixes an issue where user text zoom settings would be ignored by the empty search results wrapper.

- uses spacing tokens (which output values with `rem` units instead of px)
- `em`-ifies the media query
- uses `min-` instead of `max-`, like most of our media queries
- changes the padding from 60px to 64px (extra-loose-spacing times 2)


alex-page (approved):
Amazing ty @kaelig

Issue #2986, opened 2020-05-14
Tag component doesn't have a removable option


artyrcheek:
## Issue summary

when a onRemove prop is passed to the tag component it adds a `Polaris-Tag--removable` class but it doesn't do anything, and the onRemove is never able to be triggered

## Expected behavior

when an onRemove prop is passed, the tag should show a close icon on the right hand side that should be clickable

## Actual behavior

The tag is added but nothing is change

## Steps to reproduce the problem

1. add a tag component
2. add a onRemove prop

## Reduced test case

https://lpy7b.csb.app/

## Specifications

- Are you using the React components? (Y/N): Y
- Polaris version number: 4.22
- Browser: Chrome
- Device: Macbook pro osx
- Operating System: osx

Or run `npx envinfo --system --binaries --browsers --npmPackages react,react-dom,@shopify/polaris` to provide specifications on your environment including version numbers, browser, device, and operating system.

Paste the results here:

```bash
System:
    OS: macOS Mojave 10.14.6
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 323.41 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 13.2.0 - /usr/local/bin/node
    Yarn: 1.19.2 - /usr/local/bin/yarn
    npm: 6.13.1 - /usr/local/bin/npm
  Browsers:
    Chrome: 81.0.4044.138
    Firefox: 75.0
    Safari: 12.1.2
  npmPackages:
    @shopify/polaris: ^4.17.0 => 4.22.0
    react: ^16.13.1 => 16.13.1
    react-dom: ^16.13.1 => 16.13.1

```

chloerice:
Hey @artyrcheek 👋 Can you link to the sandbox? I can't see the code you've used with the link provided. Did you set both `onClick` and `onRemove` on the `Tag`? [Tags can be removable or clickable but not both](https://codesandbox.io/s/focused-golick-o03v2?file=/App.js).

artyrcheek:
My bad 🤦

chloerice:
> My bad 🤦

Not at all, @artyrcheek, we need to make this clearer! I'll add an example for removable tags and add some clarity to the prop descriptions so it's clear `onClick` and `onRemove` are mutually exclusive.

PR #2987, opened 2020-05-14
[Tag] Clarify that onClick and onRemove are mutually exclusive

chloerice:

 #2986  brought to my attention that there isn't a removable tag example, and that it needs to be documented that clickable tags aren't removable.

This PR adds a removable tag example and clarifies that the remove button doesn't render when `onClick` is set. 

***Note: The pa11y test failure is due to [a bug reported with `Autocomplete`](https://github.com/Shopify/polaris-react/issues/3639).

## 

### How to 🎩

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -10,6 +10,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Documentation
 
+- Added an example for the `onRemove` prop to `Tag` and clarified that no remove button is rendered when `onClick` is set ([#2987](https://github.com/Shopify/polaris-react/pull/2987))
+
 ### Development workflow
 
 ### Dependency upgrades
--- a/src/components/Tag/README.md
+++ b/src/components/Tag/README.md
@@ -33,18 +33,105 @@ Tags should:
 
 ### Default tag
 
-Use to allow merchants to add attributes to, and remove attributes from, an object.
+Use to signify the attributes of an object.
 
 ```jsx
 <Tag>Wholesale</Tag>
 ```
 
-### Remove tag
+### Removable tag
 
 Use to allow merchants to remove attributes from an object.
 
 ```jsx
-<Tag onRemove={() => {}}>Wholesale</Tag>
+function RemovableTagExample() {
+  const allOptions = [
+    {value: 'rustic', label: 'Rustic'},
+    {value: 'antique', label: 'Antique'},
+    {value: 'vinyl', label: 'Vinyl'},
+    {value: 'vintage', label: 'Vintage'},
+    {value: 'refurbished', label: 'Refurbished'},
+  ];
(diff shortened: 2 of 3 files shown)


danrosenthal:
Would it be important to actually remove the tag in the example?

chloerice:
> Would it be important to actually remove the tag in the example?

It's moreso illustrating that setting `onRemove` adds a remove button you can wire up. But we do have a simple tag example in the `Autocomplete` docs I can copy over!

BPScott:
Could we also help enforce this in the Typescript declaration of Tag's props? TextField does similar where [you can have only one of `readonly`, `disabled` or `onChange`)[https://github.com/Shopify/polaris-react/blob/master/src/components/TextField/TextField.tsx#L117]

chloerice:
> Could we also help enforce this in the Typescript declaration of Tag's props? TextField does similar where [you can have only one of `readonly`, `disabled` or `onChange`](https://github.com/Shopify/polaris-react/blob/master/src/components/TextField/TextField.tsx#L117)

Marc [did that](https://github.com/Shopify/polaris-react/blob/24d63e3e9d50cb0fe3c72889e64c8227a980da87/src/components/Tag/Tag.tsx#L23) for `Tag` as well.

BPScott:
Cool, ignore me :)

BPScott:
It feels a little odd for the demo of removable tag to not show any on initial render, but to instead show an autocomplete box.Doesn't fee very useful for our VRT stuff.

How about something like this as an example:

```js
function RemovableTagExample() {
  const [selectedTags, setSelectedTags] = useState([
    'Rustic',
    'Antique',
    'Vinyl',
    'Refurbished',
  ]);

  const removeTag = useCallback(
    (tag) => () => {
      setSelectedTags((previousTags) =>
        previousTags.filter((previousTag) => previousTag !== tag),
      );
    },
    [],
  );

  const tagMarkup = selectedTags.map((option) => (
    <Tag key={option} onRemove={removeTag(option)}>
      {option}
    </Tag>
  ));

  return <Stack spacing="tight">{tagMarkup}</Stack>;
}
```

kyledurand (approved):
Code looks good to me 🥇 Thank Chloe!

chloerice:
@kyledurand Would you mind merging this for me? I don't have admin privileges and the pa11y failures don't have anything to do with this PR (the autocomplete is setting `aria-expanded` on the text field instead of the combobox wrapper, see https://github.com/Shopify/polaris-react/issues/3639).

kyledurand:
I can merge for you but I'd prefer if you could add those two failing ids to our accessibility checks list https://github.com/Shopify/polaris-react/blob/master/scripts/accessibility-check.js#L88

BPScott:
No need to fixup a11y tests skip lists if we've already fixed all our failing a11y tests :thonk:

PR #3002, opened 2020-05-25
[Page] Add ReactNode as an accepted primaryAction prop value

WillsonSmith:
The current `primaryAction` on `Page` only allows for one button, I'm currently working on a project that has a need for more flexible primary actions.

This PR maintains the existing `primaryAction` prop, while adding the option to pass a `ReactNode` instead for customizability.

#### What other approaches were considered?
We had considered building the API/UI for `primaryAction` to suit our specific need. We decided against this because we are not sure it's the right solution for everyone.  This way if the provided option `PrimaryAction` doesn't cover every scenario, a custom option can be implemented. 

## 

### How to 🎩

<details>

```jsx
import React from 'react';

  return (
    <Page
      title="Playground"
      primaryAction={
        <Button
          primary
          connectedDisclosure={{
            accessibilityLabel: 'Other save actions',
            actions: [{content: 'Save as new'}],
          }}
        >
          Save
        </Button>
      }
    >
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Page/Page.tsx
+++ b/src/components/Page/Page.tsx
@@ -16,7 +16,7 @@ import {
   WithAppProviderProps,
 } from '../../utilities/with-app-provider';
 
-import {Header, HeaderProps} from './components';
+import {Header, HeaderProps, isPrimaryAction} from './components';
 import styles from './Page.scss';
 
 export interface PageProps extends HeaderProps {
@@ -166,7 +166,7 @@ class PageInner extends React.PureComponent<ComposedProps, never> {
     return {
       title,
       buttons: transformActions(appBridge, {
-        primaryAction,
+        ...(isPrimaryAction(primaryAction) && {primaryAction}),
         secondaryActions,
         actionGroups,
       }),
--- a/src/components/Page/components/Header/Header.tsx
+++ b/src/components/Page/components/Header/Header.tsx
@@ -43,7 +43,7 @@ export interface HeaderProps extends TitleProps {
   /** Adds a border to the bottom of the page header (stand-alone app use only) */
   separator?: boolean;
   /** Primary page-level action */
-  primaryAction?: PrimaryAction;
+  primaryAction?: PrimaryAction | React.ReactNode;
   /** Page-level pagination (stand-alone app use only) */
   pagination?: PaginationDescriptor;
   /** Collection of breadcrumbs */
@@ -54,6 +54,12 @@ export interface HeaderProps extends TitleProps {
   actionGroups?: MenuGroupDescriptor[];
   /** Additional navigation markup */
   additionalNavigation?: React.ReactNode;
+}
+
+export function isPrimaryAction(
+  x: PrimaryAction | React.ReactNode,
+): x is PrimaryAction {
+  return !React.isValidElement(x) && x !== undefined;
 }
 
 export function Header({
@@ -116,28 +122,8 @@ export function Header({
     />
   );
 
-  const primary =
-    primaryAction &&
-    (primaryAction.primary === undefined ? true : primaryAction.primary);
-
   const primaryActionMarkup = primaryAction ? (
-    <ConditionalWrapper
-      condition={newDesignLanguage === false}
-      wrapper={(children) => (
-        <div className={styles.PrimaryActionWrapper}>{children}</div>
-      )}
-    >
-      {buttonsFrom(
-        shouldShowIconOnly(
-          newDesignLanguage,
-          isNavigationCollapsed,
-          primaryAction,
-        ),
-        {
-          primary,
-        },
-      )}
-    </ConditionalWrapper>
+    <PrimaryActionMarkup primaryAction={primaryAction} />
   ) : null;
 
   const actionMenuMarkup =
@@ -225,6 +211,42 @@ export function Header({
         {primaryActionMarkup}
       </div>
     </div>
+  );
+}
+
+function PrimaryActionMarkup({
+  primaryAction,
+}: {
+  primaryAction: PrimaryAction | React.ReactNode;
+}) {
+  const {isNavigationCollapsed} = useMediaQuery();
+  const {newDesignLanguage} = useFeatures();
+  let content = primaryAction;
+  if (isPrimaryAction(primaryAction)) {
+    const primary =
+      primaryAction.primary === undefined ? true : primaryAction.primary;
+
+    content = buttonsFrom(
+      shouldShowIconOnly(
+        newDesignLanguage,
+        isNavigationCollapsed,
+        primaryAction,
+      ),
+      {
+        primary,
+      },
+    );
+  }
+
+  return (
+    <ConditionalWrapper
+      condition={newDesignLanguage === false}
+      wrapper={(children) => (
+        <div className={styles.PrimaryActionWrapper}>{children}</div>
+      )}
+    >
+      {content}
+    </ConditionalWrapper>
   );
 }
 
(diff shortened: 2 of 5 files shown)


src/components/Page/Page.tsx
```diff
@@ -166,7 +166,9 @@ class PageInner extends React.PureComponent<ComposedProps, never> {
     return {
       title,
       buttons: transformActions(appBridge, {
-        primaryAction,
+        primaryAction: isPrimaryAction(primaryAction)
```
WillsonSmith, line 169: @BPScott Thoughts on how I'm handing the typing here & the helper function in general? Will probably ping you for review too, but was curious on your initial thoughts.

BPScott: This seems alright at first glance. AppBridge integration is going away in v5. The PR has been open for an age and hopefully we'll get around to releasing it in 2 months, so I'd prioritize how to handle this stulf in non-AppBridge-land which you're doing anyway.

src/components/Page/components/Header/Header.tsx
```diff
@@ -116,18 +122,13 @@ export function Header({
     />
   );
 
-  const primary =
-    primaryAction &&
-    (primaryAction.primary === undefined ? true : primaryAction.primary);
+  function getPrimaryActionMarkup() {
```
BPScott, line 125: Nested function definitions tend to feel odd to me - you need to go check for any closed over variables that appear in the scope. Can this be pulled out to the top level?

src/components/Page/components/Header/Header.tsx
```diff
@@ -136,9 +137,24 @@ export function Header({
...
-      )}
-    </ConditionalWrapper>
-  ) : null;
+      );
+    } else {
+      content = primaryAction;
+    }
+
+    return (
+      <ConditionalWrapper
```
BPScott, line 146: If you're writing if statements above then doing an early return of `if (newDesignLanguage) { return content; `}` might be more readable than using `ConditionalWrapper`

src/components/Page/components/Header/Header.tsx
```diff
@@ -56,6 +56,12 @@ export interface HeaderProps extends TitleProps {
   additionalNavigation?: React.ReactNode;
 }
 
+export function isPrimaryAction(
+  x: PrimaryAction | React.ReactNode,
+): x is PrimaryAction {
+  return !React.isValidElement(x) && x !== undefined;
```
BPScott, line 62: This feels a little odd to me - this condition is more "this is not a react node" rather than "this is a primary action", which is alright for now but might get confusing if we want to pass more sorts of things into this. Is there a way to check for action's shape rather than not ReactNode's shape?

src/components/Page/components/Header/Header.tsx
```diff
@@ -344,3 +333,39 @@ function determineLayout({
 
   return layout.slots;
 }
+
+function getPrimaryActionMarkup(
```
BPScott, line 337: A function that returns some JSX is a React component. What do you think to naming it and treating it as such? `PrimaryActionMarkup()` and then calling it like a react component above `const primaryActionMarkup = primaryAction ? <PrimaryActionMarkup ... /> : null`

WillsonSmith: Ah yeah this is the right decision, i'll do this.

src/components/Page/components/Header/Header.tsx
WillsonSmith: I agree that it feels a little odd, the reason I did it this way is there are no _required_ properties on `primaryAction`

src/components/Page/components/Header/tests/Header.test.tsx
```diff
@@ -98,6 +98,16 @@ describe('<Header />', () => {
...
+
+    it('renders a `ReactNode`', () => {
+      const primaryAction = <div>Hello</div>;
+
+      const header = mountWithAppProvider(
+        <Header {...mockProps} primaryAction={primaryAction} />,
+      );
+
+      expect(header.contains(primaryAction)).toBeTruthy();
+    });
```
AndrewMusgrave, line 110: Another cool way this could be handled is using `react-testing` matchers!

```tsx
const PrimaryAction = () => null;

const header = mountWithApp(
  <Header {...mockProps} primaryAction={<PrimaryAction />} />,
);

expect(header).toContainReactComponent(PrimaryAction);
```

src/components/Page/Page.tsx
AndrewMusgrave: There's a fancy way to add optional object params

```tsx
{
        ...(isPrimaryAction(primaryAction) && {primaryAction})
}
```

src/components/Page/components/Header/Header.tsx
```diff
@@ -116,28 +122,8 @@ export function Header({
...
-          newDesignLanguage,
-          isNavigationCollapsed,
-          primaryAction,
-        ),
-        {
-          primary,
-        },
-      )}
-    </ConditionalWrapper>
+    <PrimaryActionMarkup primaryAction={primaryAction} />
```
AndrewMusgrave, line 126: We usually create an `if` statement rather than a component in situations like these, but the separation of concern is nice. Curious what others think?

src/components/Page/components/Header/Header.tsx
```diff
@@ -228,6 +214,42 @@ export function Header({
...
 
+function PrimaryActionMarkup({
+  primaryAction,
+}: {
+  primaryAction: PrimaryAction | React.ReactNode;
+}) {
+  const {isNavigationCollapsed} = useMediaQuery();
+  const {newDesignLanguage} = useFeatures();
+  let content = primaryAction;
+  if (isPrimaryAction(primaryAction)) {
```
AndrewMusgrave, line 225: Nitpick: With an early return we don't have to wrap the entire component 😄 

```suggestion
  let content = primaryAction;
  if (!isPrimaryAction(primaryAction)) return null
  
  const primary = ...
```

src/components/Page/components/Header/Header.tsx
```diff
@@ -228,6 +214,42 @@ export function Header({
   );
 }
 
+function PrimaryActionMarkup({
+  primaryAction,
+}: {
+  primaryAction: PrimaryAction | React.ReactNode;
+}) {
+  const {isNavigationCollapsed} = useMediaQuery();
+  const {newDesignLanguage} = useFeatures();
```
AndrewMusgrave, line 223: Curious if anyone thinks we should pass these in as props?

src/components/Page/components/Header/Header.tsx
WillsonSmith: I never want to return `null` with this component. This is wrapped in the condition because it changes `content` to be the button instead of `primaryAction` (which could be a React node). The `ConditionalWrapper` should always be returned with whatever `content` is as the children.

src/components/Page/components/Header/Header.tsx
WillsonSmith: Yeah I was curious about this too.

src/components/Page/components/Header/Header.tsx
dleroux: I think this is fine.

src/components/Page/components/Header/Header.tsx
dleroux: I'd leave it. It will be easier to clean-up and really doesn't become part of the API 🤷

alex-page:
This is good to merge 🚀

PR #3007, opened 2020-05-27
[TopBar] Animatable theming

kaelig:

References https://github.com/Shopify/destinations/issues/863

Introduces transitions on elements that change color/background-color when the theme of the top bar is switched after the page has loaded. See issue above for the full context.

Note: this doesn't follow the design specification to the letter, as some tweaks were made by @AlexBlaise and I directly in the browser in order to reach a result that wouldn't trigger disproportionate amounts of engineering work.

This PR is safe to revert when we have the ability in `Shopify/web` to determine these conditions server-side rather than client-side: **user has access to a single Plus store in an organization, and the Global nav beta is enabled in this store**. When this can be computed server-side, we'll instantly render the appropriate theme, rather than animating it in.

## 

### How to 🎩

### 🎩 checklist

### DIFF
--- a/src/components/TopBar/components/Menu/Menu.scss
+++ b/src/components/TopBar/components/Menu/Menu.scss
@@ -2,7 +2,9 @@
 
 $activator-variables: (
   min-width: rem(36px),
-  transition: background-color 0.1s,
+  transition: (
+    background-color 0.1s,
+  ),
   focus-opacity: 0.85,
 );
 
--- a/src/components/TopBar/components/SearchDismissOverlay/SearchDismissOverlay.scss
+++ b/src/components/TopBar/components/SearchDismissOverlay/SearchDismissOverlay.scss
@@ -2,7 +2,7 @@
 @import '../../variables';
 
 $entry-iterations: 1;
-$backdrop-color: rgba(33, 43, 54, 0.4);
+$backdrop-color: rgba(color('ink'), 0.4);
 
 .SearchDismissOverlay {
   position: fixed;
--- a/src/components/TopBar/components/SearchField/SearchField.scss
+++ b/src/components/TopBar/components/SearchField/SearchField.scss
@@ -98,6 +98,7 @@ $stacking-order: (
 
   &::placeholder {
     color: var(--p-text, var(--top-bar-color));
+    transition: (duration() - 33ms) color easing() 33ms;
   }
 
   &::-webkit-search-decoration,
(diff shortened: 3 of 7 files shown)


src/components/TopBar/components/SearchDismissOverlay/SearchDismissOverlay.scss
```diff
@@ -2,7 +2,7 @@
 @import '../../variables';
 
 $entry-iterations: 1;
-$backdrop-color: rgba(33, 43, 54, 0.4);
+$backdrop-color: rgba(color('ink'), 0.4);
```
kyledurand, line 5: Can this use a custom property with a color(‘ink’) fallback?

src/components/TopBar/components/SearchField/SearchField.scss
```diff
@@ -98,6 +98,7 @@ $stacking-order: (
 
   &::placeholder {
     color: var(--p-text, var(--top-bar-color));
+    transition: (duration() - 33ms) color easing() 33ms;
```
kyledurand, line 101: Super nit but 33ms can be a sass var.

src/components/TopBar/components/Menu/Menu.scss
```diff
@@ -2,7 +2,10 @@
 
 $activator-variables: (
   min-width: rem(36px),
-  transition: background-color 0.1s,
+  transition: (
+    duration(fast) background-color,
```
kyledurand, line 6: Should background color have easing as well?

src/components/TopBar/components/SearchField/SearchField.scss
kaelig: I was thinking that as well, but wasn't sure where to put it so it'd be available across multiple partials. Where would it make sense for this Sass var to be stored?

src/components/TopBar/components/SearchDismissOverlay/SearchDismissOverlay.scss
kaelig: Theming of the search experience is out of scope for this piece of work, would it be okay to revisit this at a later time, if it becomes needed?

src/components/TopBar/components/Menu/Menu.scss
kaelig: Actually, let me remove the changes I made to this map. The color transition is applied to children, not on the top level activator element.

src/components/TopBar/components/SearchField/SearchField.scss
BPScott: If we need a place to share sass variables across subcomponents we occasionally put an `_variables.scss` file in the component folder. For TopBar we've already got  https://github.com/Shopify/polaris-react/blob/master/src/components/TopBar/_variables.scss

PR #3021, opened 2020-05-30
Add support for LoadableAction in AppBridgeAction

alex-page:

Fixes https://github.com/Shopify/polaris-react/issues/1410

Add support for `LoadableAction` in `AppBridgeAction`. This allows us to have a button with a `loading` prop in a modal for the local delivery app.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -165,7 +165,9 @@ export interface AppBridgeAction
   extends Action,
     DisableableAction,
     DestructableAction,
-    AppBridgeActionTarget {}
+    AppBridgeActionTarget,
+    IconableAction,
+    LoadableAction {}
 
 export interface IconableAction extends Action {
   /** Source of the icon */
--- a/src/types.ts
+++ b/src/types.ts
@@ -166,7 +166,6 @@ export interface AppBridgeAction
     DisableableAction,
     DestructableAction,
     AppBridgeActionTarget,
-    IconableAction,
     LoadableAction {}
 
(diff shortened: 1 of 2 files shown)


src/types.ts
```diff
@@ -165,7 +165,8 @@ export interface AppBridgeAction
   extends Action,
     DisableableAction,
     DestructableAction,
-    AppBridgeActionTarget {}
+    AppBridgeActionTarget,
+    LoadableAction {}
```
alex-page, line 169: I would love to know if this is all I have to do. I couldn't find anything else related to the AppBridgeAction type.

vividviolet: I don't think `loading` is supported in App Bridge modals. It's only supported in the Contextual Save Bar. We would need to make the change to both `app-bridge` and `app-bridge-host` before doing it here. The host package is the one that actually renders the Modal: https://github.com/Shopify/app-bridge/blob/96edcdc3f8ed3d912dba5fd37e4589d72682392c/packages/app-bridge-host/src/components/Modal.tsx#L116

AndrewMusgrave: Since `AppBridge` is deprecated and was supposed to be removed already (Dan & Tim we're aiming for last November iirc) we really shouldn't be added onto the API. If this is just a type change can you hack it in your app with casting?

alex-page: I think it needs to be a part of AppBridge first as @vividviolet said. I will close this PR when I can figure out how to do that 😂

PR #3031, opened 2020-06-06
Use a single template style for POLARIS_VERSION

BPScott:

In our code we replace `{{POLARIS_VERSION}}` and `<%= POLARIS_VERSION %>` with the actual version of polaris.

Having two formats for that is silly. Lets just have one.

- Replaces occurrence of `<%= POLARIS_VERSION %>` with `{{POLARIS_VERSION}}`
- Remove the build step that searches for `<%= POLARIS_VERSION %>` as it is no longer used
- Refactor the tests that prove that the version is replaced so that they give more meaningful output if they fail - they will now list the files that included the template string rather than saying the total number of files was different.

### How to 🎩

Search the codebase for `<%= POLARIS_VERSION %>` and see that no results are found
Ensure tests pass
Do a build and see that the files mentioned in the build test contain all contain the version string (e.g. `4.23.0`) instead of the template (`{{POLARIS_VERSION}}`).

### DIFF
--- a/tests/build.test.ts
+++ b/tests/build.test.ts
@@ -65,42 +65,40 @@ describe('build', () => {
   });
 
   it('replaces occurrences of POLARIS_VERSION', () => {
-    const files = glob.sync('./build/**/*.{js,scss,css}', {
+    const files = glob.sync('./{build,esnext}/**/*.{js,scss,css}', {
       ignore: './build/cache/**',
     });
 
-    const total = files.reduce((acc, file) => {
-      const contents = fs.readFileSync(file, 'utf-8');
-      return acc + Number(contents.includes('POLARIS_VERSION'));
-    }, 0);
-    expect(total).toBe(0);
-  });
+    expect(files).not.toHaveLength(0);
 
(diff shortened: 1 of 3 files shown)


tests/build.test.ts
```diff
@@ -65,42 +65,40 @@ describe('build', () => {
...
+    expect(fileBuckets.includesTemplateString).toHaveLength(0);
+
+    expect(fileBuckets.includesVersion).toStrictEqual([
+      './build/polaris.css',
+      './build/polaris.es.js',
+      './build/polaris.js',
+      './build/polaris.min.css',
+      './build/styles/global.scss',
+      './esnext/configure.js',
+      './esnext/styles/global.scss',
```
kaelig, line 100: This coincides with manual tests.

```console
$ cd esnext
$ ag "4.24.0" -l
configure.js
styles/global.scss
$ ag POLARIS_VERSION

$ cd ../build
$ ag "4.24.0" -l
polaris.css
polaris.min.css
styles/global.scss
polaris.js
polaris.es.js
$ ag POLARIS_VERSION

```

Issue #3009, opened 2020-05-28
Remove the need for an esnext/styles folder


BPScott:
We've got a public sass API. It is built from the contents of [src/styles](https://github.com/Shopify/polaris-react/tree/master/src/styles).

However like our JS it has two built out representations:

- [`styles`](https://unpkg.com/browse/@shopify/polaris@4.22.0/styles/) is a plain version of this ready to be consumed by people not using css-modules.
- [`esnext/styles`](https://unpkg.com/browse/@shopify/polaris@4.22.0/esnext/styles/) is ready to be consumed by folks using css-modules - i.e. folks using sewing-kit.

What would it take to remove this duplication and have a single entrypoint to our public api, so we can stop telling internal and external folks to use two different file paths? Removing this split would also help remove some [build hackery](https://github.com/Shopify/polaris-react/blob/af016fb8b0b782587ade0fb5aa78b8f887a98131/config/rollup/plugins/styles.js#L150-L163).

Two things:

- We need to stop referencing the files in `esnext/styles` within our per-component files in the esnext folder.
- We need to remove any usage of postcss directives in `src/styles`. Currently our only usage of this is a use of `:global` to get at an animation name in [_skeleton.scss](https://github.com/Shopify/polaris-react/blob/af016fb8b0b782587ade0fb5aa78b8f887a98131/src/styles/shared/_skeleton.scss#L20).

To solve the first point #2938 will need to be merged. This will mean that these files won't be referenced by our other scss files in the esnext folder (or the css files if [css support in SK is added](https://github.com/Shopify/sewing-kit/pull/2109)). At which point the distribution of our public API is disconnected from the other files we ship - we can place and manipulate it however.

To solve the second point we can place the name of the animation into a css custom property after it is defined in [global.scss](https://github.com/Shopify/polaris-react/blob/af016fb8b0b782587ade0fb5aa78b8f887a98131/src/styles/global.scss#L66) (AppProvider.scss in v5) (idea inspired by [this article](https://gravitydept.com/blog/keyframe-animations-in-css-modules):

```scss
@keyframes :global(polaris-SkeletonShimmerAnimation) {
  0% { opacity: 0.45; }
  100% { opacity: 0.9; }
}

:root {
  --polaris-shimmer-animation-name: 'polaris-SkeletonShimmerAnimation';
}
```

And then the mixin in _skeleton.scss can be updated to reference the custom property

```scss
@mixin skeleton-shimmer {
[...]

PR #3032, opened 2020-06-06
Unify the output of the styles folder build

BPScott:

A step towards #3009.

We have two locations for our public sass API

- `styles/_public-api.scss` for consumers that don't use css-modules
- `esnext/styles/_public-api.scss` for consumers that do use css-modules (e.g. folks using sewing-kit)

Having two locations is confusing to document, adds complexity to our build, and is ripe for confusing our consumers.

It is this way because in one of our mixins we use css-modules' `:global {}` pseudo-selector to get at a globally defined animation name. If we can remove the need for that then we don't need to strip that selector out at build time and then the contents of the api files in `styles/{shared,foundation}/` and `esnext/styles/{shared,foundation}/` become identical.

Let's unify our mixin files so they don't need to differ between these locations.

By doing this we ca tell all consumers to pull in the sass files via`styles/_public-api.scss` instead of `esnext/styles/_public-api.scss`.

We can't fully delete `esnext/styles` just yet as the raw scss files in the esnext folder still reference the contents of that folder.
Once we start shipping built instead of source scss files in the esnext folder we would be able to remove esnext/styles entirely (see my v5 build plotting PR).

- Updates our scss api to use a custom-property to reference the shimmer animation name instead of `:global`

### How to 🎩

- In storybook ensure that the shimmer effect still applies to skeleton components e.g. SkeletonDisplay text.
- Run a build then run `diff -ru {,esnext/}styles/shared` and `diff -ru {,esnext/}styles/foundation` and note that those two sets of folders are now identical

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -18,6 +18,8 @@
 
 ### Development workflow
 
+- Updated how we reference global animations so we can have one public sass API for all consumers instead one entrypoint (`styles/_public-api.scss`) for consumers using plain scss and one entrypoint (`styles/esnext/_public-api.scss`) for consumers using css modules ([#3032](https://github.com/Shopify/polaris-react/pull/3032))
+
 ### Dependency upgrades
 
 ### Code quality
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -18,7 +18,7 @@
 
 ### Development workflow
 
-- Updated how we reference global animations so we can have one public sass API for all consumers instead one entrypoint (`styles/_public-api.scss`) for consumers using plain scss and one entrypoint (`styles/esnext/_public-api.scss`) for consumers using css modules ([#3032](https://github.com/Shopify/polaris-react/pull/3032))
+- Updated how global animations are referenced, in order to publish a single entrypoint for the public Sass API (`styles/_public-api.scss`), instead of two (`styles/_public-api.scss` for “vanilla” SCSS and `styles/esnext/_public-api.scss` for CSS Modules) ([#3032](https://github.com/Shopify/polaris-react/pull/3032))
 
 ### Dependency upgrades
 
--- a/src/styles/global.scss
+++ b/src/styles/global.scss
@@ -8,13 +8,13 @@
   // This value matches the name of the animation defined below.
   // The animation name is stored in a custom property so that the scss mixins
   // in our public api that reference this animation can do so with
-  // `var(--polaris-SkeletonShimmerAnimationName)` instead of
+  // `var(--polaris-animation-skeleton-shimmer)` instead of
   // `:global(polaris-SkeletonShimmerAnimation)`. This allows us to expose a
   // single set of public sass API files instead of needing distinct sass API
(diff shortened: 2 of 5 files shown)


BPScott:
Updated and now ready for review again.

UNRELEASED.md
```diff
@@ -18,6 +18,8 @@
 
 ### Development workflow
 
+- Updated how we reference global animations so we can have one public sass API for all consumers instead one entrypoint (`styles/_public-api.scss`) for consumers using plain scss and one entrypoint (`styles/esnext/_public-api.scss`) for consumers using css modules ([#3032](https://github.com/Shopify/polaris-react/pull/3032))
```
kaelig, line 21:
```suggestion
- Updated how global animations are referenced, in order to publish a single entrypoint for the public Sass API (`styles/_public-api.scss`), instead of two (`styles/_public-api.scss` for “vanilla” SCSS and `styles/esnext/_public-api.scss` for CSS Modules) ([#3032](https://github.com/Shopify/polaris-react/pull/3032))
```

src/styles/global.scss
```diff
@@ -5,6 +5,16 @@
...
+  // This value matches the name of the animation defined below.
+  // The animation name is stored in a custom property so that the scss mixins
+  // in our public api that reference this animation can do so with
+  // `var(--polaris-SkeletonShimmerAnimationName)` instead of
+  // `:global(polaris-SkeletonShimmerAnimation)`. This allows us to expose a
+  // single set of public sass API files instead of needing distinct sass API
+  // entrypoints for consumers that use css-modules in locals mode (such as scss
+  // files in apps using sewing-kit) and for consumers not using css-modules.
+  // stylelint-disable-next-line value-keyword-case
+  --polaris-SkeletonShimmerAnimationName: polaris-SkeletonShimmerAnimation;
```
kaelig, line 17: Is there a standard for naming custom properties? (I was under the impression kebab-case was king, but I might be wrong)

UNRELEASED.md
BPScott: We don't actually provide any docs on sass api usage in this repo. The path we mention in https://github.com/Shopify/sewing-kit/blob/master/docs/plugins/sass.md#autoinclude can be updated but I'll do that after this gets shipped

src/styles/global.scss
BPScott: Yeah kebab-case is what most of the stuff in ThemeProvider does. Have updated the name.

I've kept this prefixed as `--polaris-` (while the ThemeProvider stuff uses `--p-`) to keep variable prefixes consistent within this file

BPScott:
> By doing this we ca tell all consumers to pull in the sass files viastyles/_public-api.scss instead of esnext/styles/_public-api.scss.

While people totally can do this. I'm not gonna tell folks about it as in v5 I'm strongly considering moving everything into a dist folder (so `dist/styles/_public-api.scss`) to avoid spreading out build files throughout the package and to make integration with https://github.com/shopify/tophat-web easier (as it acts upon a single folder)

PR #3040, opened 2020-06-08
[EmptyState] Set max-width to 400px

ashwinnarayanan2001:
This pull request is changing the max-width to 400px for the Empty State content. 

## 

### How to 🎩

You can check out this on the StoryBoard under the Empty state within context component.

### 🎩 checklist

### DIFF
--- a/src/components/EmptyState/EmptyState.scss
+++ b/src/components/EmptyState/EmptyState.scss
@@ -1,6 +1,8 @@
 @import '../../styles/common';
 
 $detail-width: rem(400px);
+$within-page-detail-width: rem(336px);
+
 $stacking-order: (
   illustration: 0,
   details: 10,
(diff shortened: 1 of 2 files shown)


ashwinnarayanan2001:
@chloerice Is changing the max-width to 400px going to affect existing empty states?

src/components/EmptyState/EmptyState.scss
```diff
@@ -1,6 +1,6 @@
 @import '../../styles/common';
 
-$detail-width: rem(336px);
+$detail-width: rem(400px);
```
kyledurand, line 3: I think we may only want to make this change to in content empty states and centered layout empty states. Looks like it could cause problems with legacy layouts

image

chloerice: This shouldn't break current empty states, as `$detail-width` is just the width of the content left of the illustration (so the illustration might shift a bit, but the illustration itself get's pull left using negative margin so it might not actually be affected at all 🤔). Happy to tophat in `web` to help double check 👍

kyledurand:
Just rebased this to resolve a conflict in Unreleased @ashwinnarayanan2001. Should be good to ship tomorrow morning 👍

PR #3043, opened 2020-06-09
[Navigation] Delete unused `sections` prop

kaelig:

`Navigation` has what seems to be an old `sections` prop that isn't featured or used anywhere.

Rather than deprecating this prop, I suggest deleting it as it wasn't doing anything.

### 🎩 checklist


amrocha:
Hmm, I agree with you that it's unused so we should just delete it, but on the other hand, this is technically a breaking change so clients might have to update 🤔 

I'm not sure what to do here. @BPScott @danrosenthal any thought? @Shopify/polaris-team the rest of the team as well

BPScott:
Ah one of those "is it breaking because we're changing the contract or is it  a bug fix because the contract was busted in the first place" tradeoffs :)

Given that prop was never referenced in code and thus had no runtime behaviour I highly doubt this was ever used (and thus any usage was incorrect). Do a quick audit to see if you can see any consuming apps that used it. When we inevitably find out nobody was using it I'm happy to say delete in a minor version.

kaelig:
>Do a quick audit to see if you can see any consuming apps that used it.

Good idea – I looked at Shopify/web and it seems like they don't use it.

For other apps, the impact to merchant is minimal so I feel good about deleting it without checking.

PR #3057, opened 2020-06-15
[VideoThumbnail] Add video progress indicator

yoshicn:

Currently, `VideoThumbnail` displays a video length to set the expectation about how long video is before user starts playing it. Many users are not finishing watching the video at one session, and we are working to allow them to resume the video at the time they left it off.

As a visual confirmation that video has been started, we are adding video progress indicator to the video thumbnail.

Add video progress indicator to `VideoThumnbail` component, so it can visualize how far user has gone through the video and set the expectation that video will resume from where it has been left off.

Progress indicator is calculated based on `videoLength` and `videoProgress` values, and represents percentage of video users watched.

Screen Shot 2020-07-23 at 6 44 10 PM

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page, MediaCard, VideoThumbnail} from '../src';

  return (
    <Page title="Playground">
      <MediaCard
        title="Turn your side-project into a business"
        primaryAction={{
          content: 'Learn more',
          onAction: () => {},
        }}
        description={`In this course, you’ll learn how the Kular family turned their mom’s recipe book into a global business.`}
        popoverActions={[{content: 'Dismiss', onAction: () => {}}]}
      >
        <VideoThumbnail
          videoLength={80}
          videoProgress={45}
          thumbnailUrl="https://burst.shopifycdn.com/photos/smiling-businesswoman-in-office.jpg?width=1850"
        />
      </MediaCard>
    </Page>
  );
}

```

</details>

### 🎩 checklist

Closes https://github.com/Shopify/start/issues/479

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -10,7 +10,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - Added a `dismissOnMouseOut` prop to `Tooltip` to dismiss Tooltip once pointer is no longer over children ([#3086](https://github.com/Shopify/polaris-react/pull/3086))
 - Added an activator prop to `Modal` so that focus can be returned to it when the `Modal` is closed ([#2206](https://github.com/Shopify/polaris-react/pull/2206))
-- Added an optional `videoProgress` and `showVideoProgress` props to `VideoThumbnail` for video progress indicator ([#3057](https://github.com/Shopify/polaris-react/pull/3057))
+- Added optional `videoProgress` and `showVideoProgress` props to `VideoThumbnail` for video progress indicator ([#3057](https://github.com/Shopify/polaris-react/pull/3057))
 
 ### Bug fixes
 
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -10,10 +10,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - Added a `dismissOnMouseOut` prop to `Tooltip` to dismiss Tooltip once pointer is no longer over children ([#3086](https://github.com/Shopify/polaris-react/pull/3086))
 - Added an activator prop to `Modal` so that focus can be returned to it when the `Modal` is closed ([#2206](https://github.com/Shopify/polaris-react/pull/2206))
-- Removed padding from the details container in `EmptyState` to account for new illustration size ([#3069](https://github.com/Shopify/polaris-react/pull/3069))
-- Added `blueDark` to the list of possible `color` values for an `Icon` with a backdrop ([#3076](https://github.com/Shopify/polaris-react/pull/3076))
-- Improved responsive layout for secondary actions in `Banner` ([#3093](https://github.com/Shopify/polaris-react/pull/3093))
-- Added an optional `videoProgress` prop to `VideoThumbnail` for video progress indicator ([#3057](https://github.com/Shopify/polaris-react/pull/3057))
+- Added an optional `videoProgress` and `showVideoProgress` props to `VideoThumbnail` for video progress indicator ([#3057](https://github.com/Shopify/polaris-react/pull/3057))
 
 ### Bug fixes
 
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -8,6 +8,13 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Added a `dismissOnMouseOut` prop to `Tooltip` to dismiss Tooltip once pointer is no longer over children ([#3086](https://github.com/Shopify/polaris-react/pull/3086))
+- Added an activator prop to `Modal` so that focus can be returned to it when the `Modal` is closed ([#2206](https://github.com/Shopify/polaris-react/pull/2206))
+- Removed padding from the details container in `EmptyState` to account for new illustration size ([#3069](https://github.com/Shopify/polaris-react/pull/3069))
+- Added `blueDark` to the list of possible `color` values for an `Icon` with a backdrop ([#3076](https://github.com/Shopify/polaris-react/pull/3076))
+- Improved responsive layout for secondary actions in `Banner` ([#3093](https://github.com/Shopify/polaris-react/pull/3093))
+- Added an optional `videoProgress` prop to `VideoThumbnail` for video progress indicator ([#3057](https://github.com/Shopify/polaris-react/pull/3057))
+
 ### Bug fixes
 
 ### Documentation
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -8,8 +8,6 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
-- Added a `dismissOnMouseOut` prop to `Tooltip` to dismiss Tooltip once pointer is no longer over children ([#3086](https://github.com/Shopify/polaris-react/pull/3086))
-- Added an activator prop to `Modal` so that focus can be returned to it when the `Modal` is closed ([#2206](https://github.com/Shopify/polaris-react/pull/2206))
 - Added optional `videoProgress` and `showVideoProgress` props to `VideoThumbnail` for video progress indicator ([#3057](https://github.com/Shopify/polaris-react/pull/3057))
 
 ### Bug fixes
--- a/src/components/VideoThumbnail/README.md
+++ b/src/components/VideoThumbnail/README.md
@@ -54,6 +54,28 @@ Use as a play button for a video player within a media card.
 </MediaCard>
 ```
 
+### Video thumbnail with progress
+
+Use to indicate the video’s play progress in relation to its duration.
+
+```jsx
+<MediaCard
+  title="Turn your side-project into a business"
+  primaryAction={{
+    content: 'Learn more',
+    onAction: () => {},
+  }}
+  description={`In this course, you’ll learn how the Kular family turned their mom’s recipe book into a global business.`}
+  popoverActions={[{content: 'Dismiss', onAction: () => {}}]}
+>
+  <VideoThumbnail
+    videoLength={80}
+    videoProgress={45}
+    thumbnailUrl="https://burst.shopifycdn.com/photos/smiling-businesswoman-in-office.jpg?width=1850"
+  />
+</MediaCard>
+```
+
 ---
 
 ## Required components
--- a/src/components/VideoThumbnail/README.md
+++ b/src/components/VideoThumbnail/README.md
@@ -71,6 +71,7 @@ Use to indicate the video’s play progress in relation to its duration.
   <VideoThumbnail
     videoLength={80}
     videoProgress={45}
+    showVideoProgress
     thumbnailUrl="https://burst.shopifycdn.com/photos/smiling-businesswoman-in-office.jpg?width=1850"
   />
 </MediaCard>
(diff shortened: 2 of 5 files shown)


UNRELEASED.md
```diff
@@ -6,6 +6,8 @@
 
 ### Enhancements
 
+- Added an optional `videoProgress` prop to `VideoThumbnail` for video progress indicator [#3057](https://github.com/Shopify/polaris-react/pull/3057)
```
kaelig, line 9:
```suggestion
- Added an optional `videoProgress` prop to `VideoThumbnail` for video progress indicator ([#3057](https://github.com/Shopify/polaris-react/pull/3057))
```

src/components/VideoThumbnail/tests/VideoThumbnail.test.tsx
```diff
@@ -75,6 +75,79 @@ describe('<VideoThumbnail />', () => {
...
+      });
+
+      expect(progressIndicator).toHaveReactProps({
+        style: expect.objectContaining({
+          width: '50%',
+        }),
+      });
+    });
+
+    it('warns when video progress exceeds video length', () => {
```
kaelig, line 138: How do you feel about adding a second test to ensure there's no warning for the production environment?

src/components/VideoThumbnail/VideoThumbnail.tsx
```diff
@@ -27,7 +36,8 @@ export interface VideoThumbnailProps {
 
 export function VideoThumbnail({
   thumbnailUrl,
-  videoLength,
+  videoLength = 0,
+  videoProgress = 0,
```
kaelig, line 40: How do you feel about making the API more explicit and having a separate prop for showing the video progress (such as `showVideoProgress`), rather than relying on `progressValue` for this logic?

chloerice: Having an additional prop won't change the fact that the video progress is dependent on video length also being set. We should be clear in the prop description that it will only render if `videoLength` is set 💯.

src/components/VideoThumbnail/VideoThumbnail.tsx
```diff
@@ -13,8 +14,16 @@ import styles from './VideoThumbnail.scss';
...
   /** URL source for thumbnail image. */
   thumbnailUrl: string;
-  /** Length of video in seconds. */
+  /**
+   * Length of video in seconds.
+   * @default 0
+   */
   videoLength?: number;
+  /**
+   * Video progress in seconds.
```
chloerice, line 23:
```suggestion
   * Video progress in seconds. Displays a progress bar at the bottom of the thumbnail. Only renders when videoLength is also set. 
```

src/components/VideoThumbnail/VideoThumbnail.tsx
kaelig: That's a good point!

I also thought of one case where having a `showVideoProgress` prop could be useful: if for some reason we want to show the progress bar when `videoProgress=0` (default).

yoshicn: Great suggestions!

If we want to add `showVideoProgress`, should we should also add `showVideoLength` to make component's API consistent? However, I feel it might be a bit overkill to maintain two additional props on component (at least based on the use cases we have for VideoThumbnail on web).

I would be totally down to add both props if you guys think they should be added? 🙂

yoshicn:
@chloe-gautreau @kaelig 
Would appreciate if you guys could take a second look at this PR. Last two weeks were very spotty with availability, but now it's becoming more pressing on shipping this as it blocks the progress feature we want to ship for contextual video.

Thanks a lot!

src/components/VideoThumbnail/VideoThumbnail.tsx
kaelig: That's a good question! If there is a use-case for the video length to be hidden from the user, then `showVideoLength` would make sense as a prop. If that's resulting in inconsistencies and poorer experiences, I'm not sure this is something we'd want.

yoshicn: I do not think there is a use-case for hiding video length through separate prop. Video progress does not make a lot of sense without video length displayed.

But to address original question, @kaelig do you feel strong on adding `showVideoProgress` as a prop?

src/components/VideoThumbnail/VideoThumbnail.tsx
```diff
@@ -56,8 +66,30 @@ export function VideoThumbnail({
...
+  ) : null;
+
+  const progressMarkup = progressValue ? (
+    <div className={styles.Progress}>
+      <progress
+        className={styles.ProgressBar}
+        value={progressValue}
+        max="100"
+      />
+      <div className={styles.Indicator} style={{width: `${progressValue}%`}}>
```
kaelig, line 89: If this will ever be changing live (as in: the indicator moves as  while the user plays the video), use `transform: scaleX(<value between 0 and 1>)` instead of `width`, as `transform` animates more efficiently than `width`.

Also set these styles on the indicator:

```scss
  width: 100%; // (rather than 0)
  transform-origin: left;
  transform: scaleX(0);
```

src/components/VideoThumbnail/VideoThumbnail.tsx
kaelig: Gotcha, thanks!

I'm a 7 out of 10 on having a clear `showVideoProgress` in there rather than having logic in the component that "smartly" determines if the video progress should appear. That way, we get a fully declarative API rather than a data-based API (which is typically easier to learn and debug).

yoshicn: Ok great, thanks. Will add new prop asap!

src/components/VideoThumbnail/VideoThumbnail.tsx
yoshicn: Great suggestion! Did not have use-case for live update yet, but definitely better to have it! 👍

yoshicn:
@kaelig addressed the feedback. Would appreciate your 👀 again. Thanks!

yoshicn:
@kaelig just friendly bump in case it got out of radar 😄

src/components/VideoThumbnail/VideoThumbnail.tsx
```diff
@@ -56,10 +72,37 @@ export function VideoThumbnail({
...
 
+  const progressValue = calculateProgress(videoLength, videoProgress);
+  const progressValuePercents = Math.round(progressValue * 100);
+
   const timeStampMarkup = videoLength ? (
-    <p className={styles.Timestamp}>{secondsToTimestamp(videoLength)}</p>
+    <p
+      className={classNames(
+        styles.Timestamp,
+        progressValue && styles.WithProgress,
```
kaelig, line 82: Should this be:

```suggestion
        showVideoProgress && styles.WithProgress,
```

?

UNRELEASED.md
```diff
@@ -18,6 +18,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Enhancements
 
 - Added an activator prop to `Modal` so that focus can be returned to it when the `Modal` is closed ([#2206](https://github.com/Shopify/polaris-react/pull/2206))
+- Added an optional `videoProgress` and `showVideoProgress` props to `VideoThumbnail` for video progress indicator ([#3057](https://github.com/Shopify/polaris-react/pull/3057))
```
kaelig, line 21:
```suggestion
- Added optional `videoProgress` and `showVideoProgress` props to `VideoThumbnail` for video progress indicator ([#3057](https://github.com/Shopify/polaris-react/pull/3057))
```

src/components/VideoThumbnail/README.md
```diff
@@ -54,6 +54,28 @@ Use as a play button for a video player within a media card.
...
+  primaryAction={{
+    content: 'Learn more',
+    onAction: () => {},
+  }}
+  description={`In this course, you’ll learn how the Kular family turned their mom’s recipe book into a global business.`}
+  popoverActions={[{content: 'Dismiss', onAction: () => {}}]}
+>
+  <VideoThumbnail
+    videoLength={80}
+    videoProgress={45}
```
kaelig, line 73:
```suggestion
    videoProgress={45}
    showVideoProgress
```

src/components/VideoThumbnail/VideoThumbnail.scss
```diff
@@ -67,4 +67,31 @@ $start-button-size: 60px;
   background-color: var(--p-surface, color('ink'));
   opacity: 0.8;
   text-align: center;
+
+  &.WithProgress {
+    margin-bottom: spacing(base-tight);
+  }
+}
```
kaelig, line 74: I'm not 100% sure on this one but I believe this is what this should be like:

```suggestion
}

.withProgress {
  margin-bottom: spacing(base-tight);
}
```

src/components/VideoThumbnail/VideoThumbnail.tsx
```diff
@@ -27,7 +41,9 @@ export interface VideoThumbnailProps {
 
 export function VideoThumbnail({
   thumbnailUrl,
-  videoLength,
+  videoLength = 0,
+  videoProgress = 0,
+  showVideoProgress = true,
```
kaelig, line 46: Unless I'm missing something, this should default to `false`, and be set explicitly by the user if they want to show it.

```suggestion
  showVideoProgress = false,
```

src/components/VideoThumbnail/VideoThumbnail.tsx
```diff
@@ -56,10 +72,37 @@ export function VideoThumbnail({
     );
   }
 
+  const progressValue = calculateProgress(videoLength, videoProgress);
+  const progressValuePercents = Math.round(progressValue * 100);
```
kaelig, line 76: Should these calculations only run when needed? (for example, by wrapping them in a condition that checks `showVideoProgress` is true)

src/components/VideoThumbnail/VideoThumbnail.tsx
```diff
@@ -56,10 +72,37 @@ export function VideoThumbnail({
...
+        styles.Timestamp,
+        progressValue && styles.WithProgress,
+      )}
+    >
+      {secondsToTimestamp(videoLength)}
+    </p>
   ) : null;
 
+  const progressMarkup =
+    showVideoProgress && progressValue ? (
```
kaelig, line 90: What if `progressValue = 0` but the developer still wants to show the progress bar?

yoshicn:
@kaelig Thanks for the great feedback. This is ready for the next round 👍

src/components/VideoThumbnail/VideoThumbnail.scss
```diff
@@ -77,8 +78,8 @@ $start-button-size: 60px;
   position: absolute;
   bottom: 0;
   width: 100%;
-  background-color: rgba(color('white'), 0.9);
-  height: spacing(extra-tight);
+  background-color: color('white');
```
kaelig, line 81: Did you mean to remove the semi-transparency of the progress bar?

src/components/VideoThumbnail/VideoThumbnail.scss
```diff
@@ -1,5 +1,6 @@
 @import '../../styles/common';
-$start-button-size: 60px;
+$start-button-size: rem(60px);
```
kaelig, line 2: Nice a11y touch! ❤️

src/components/VideoThumbnail/VideoThumbnail.scss
yoshicn: Yes, this change is intentional. Had to make it last minute to match implementation across all platforms (iOS and Android)

PR #3061, opened 2020-06-17
Ship it task to unpublish the latest version

alex-page:
Co-Authored-By: Kyle Durand <kyledurand@users.noreply.github.com>

Sometimes releases go wrong. This helps by allowing the users to unpublish and remove the latest tag from GitHub and NPM.

Adds a custom task to ShipIt.

### How to TopHat

Check the tasks tab on ShipIt, don't run the task please:
https://shipit.shopify.io/shopify/polaris-react/test-unpublish

### DIFF
--- a/shipit.yml
+++ b/shipit.yml
@@ -12,8 +12,6 @@ ci:
 
 tasks:
   unpublish:
-    override:
-      - echo 'nothing to install'
     variables:
       - name: VERSION
         title: 'Unpublish a version from NPM and remove the tag from GitHub'
--- a/shipit.yml
+++ b/shipit.yml
@@ -12,6 +12,8 @@ ci:
 
 tasks:
   unpublish:
+    override:
+      - echo 'nothing to install'
     variables:
       - name: VERSION
         title: 'Unpublish a version from NPM and remove the tag from GitHub'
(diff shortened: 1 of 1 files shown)


shipit.yml
```diff
@@ -10,6 +10,16 @@ ci:
...
 
+tasks:
+  unpublish:
+    variables:
+      - name: VERSION
+        title: 'Unpublish a version from NPM and remove the tag from GitHub'
+        default: 'v0.0.0'
+    action: 'Unpublish a version'
+    steps:
+      - npm unpublish @shopify/polaris@${VERSION}
```
alex-page, line 21: This needs the tag removal to be added once this script is working correctly:
```
- git push --delete origin $VERSION
```

kaelig:
It's generally considered an antipattern to unpublish a version or a package as it can break other people's code/builds. This is why we're not letting people delete packages on our internal packagecloud instance.

I'd recommend using dist-tags or publishing a new version instead.

alex-page:
@kaelig I completely agree. This is for worst case scenario when releases go wrong.

kaelig:
Our versioning script on polaris-react doesn't restrict people to using `-alpha`, `-beta`, `-rc` and `-next` ([documentation](https://github.com/Shopify/shipit-engine/blob/master/docs/npm-lerna.md)) – it'd be great to put a check in place to prevent people from doing the wrong thing.

>I will make sure this is well documented.

💯 x 💯 to that

PR #3063, opened 2020-06-18
Write sass errors that are valid in dart-sass

BPScott:

I found out public_api doesn't compile in dart-sass: `npx node-sass src/styles/_public-api.scss`. Looks like it is because dart-sass and libsass handles newlines in strings a bit differently.

dart-sass doesn't like multi-line strings (which I kinda agree with - css doesn't like unescaped multiline strings either). Escaping the newlines (`\n`)
results in libsass not displaying them properly. Result: remove
linebreaks.

Also change a block-style comment to be inline so it doesn't get output.

### How to 🎩

- Confirm the public api can be compiled with no output using node-sass: `npx node-sass src/styles/_public-api.scss`
Confirm the public api can be compiled with no output (and no errors!) using dart-sass : `npx sass@latest src/styles/_public-api.scss`

For bonus points, add a usage of these mixins and see that they the error message looks decent in both of the above cases. 

```css
.x {
  content: color(JUNK);
  content: ms-high-contrast-color(JUNK);
  content: shadow(JUNK);
}
```

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -9,7 +9,7 @@
 ### Bug fixes
 
 - Added `flex: 1 1 auto` to `Banner` `.ContentWrapper` CSS selector ([#3062](https://github.com/Shopify/polaris-react/pull/3062))
-- Ensured sass mixins can compile in dart sass ([#3064](https://github.com/Shopify/polaris-react/pull/3063))
+- Ensured Sass mixins can compile in Dart Sass ([#3064](https://github.com/Shopify/polaris-react/pull/3063))
 
 ### Documentation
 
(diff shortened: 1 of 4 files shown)


kaelig:
Awesome!

Can we replace the new lines with `\a` to keep the messages readable?

BPScott:
> Can we replace the new lines with `\a` to keep the messages readable?

huh, TIL. I was trying newline escapes with `\n`. `\a`s added!`

UNRELEASED.md
```diff
@@ -9,6 +9,7 @@
 ### Bug fixes
 
 - Added `flex: 1 1 auto` to `Banner` `.ContentWrapper` CSS selector ([#3062](https://github.com/Shopify/polaris-react/pull/3062))
+- Ensured sass mixins can compile in dart sass ([#3064](https://github.com/Shopify/polaris-react/pull/3063))
```
kaelig, line 12:
```suggestion
- Ensured Sass mixins can compile in dart-sass ([#3064](https://github.com/Shopify/polaris-react/pull/3063))
```

:p

PR #3064, opened 2020-06-18
[Page] action rollup consistent padding and alignment

alex-page:
Co-Authored-By: Dan Ross <hypd@users.noreply.github.com>

Fixes https://github.com/Shopify/deliveries/issues/1102 and fixes https://github.com/Shopify/polaris-react/issues/2739

The ellipsis in the page actions currently does not use the same spacing as the header on the left hand side. This makes the padding around the Page component appear consistent.

Example of issue:
Screen Shot 2020-06-18 at 10 29 46@2x

This PR does:
ellipsis

### How to 🎩

Open up the Page component in StoryBook and test the ellipsis and action locations on desktop and mobile.

### 🎩 checklist

### DIFF
--- a/src/components/Page/components/Header/Header.scss
+++ b/src/components/Page/components/Header/Header.scss
@@ -1,7 +1,7 @@
 @import '../../../../styles/common';
 
 $individual-action-padding-x: (1.5 * spacing(tight));
-$action-menu-rollup-computed-width: icon-size();
+$action-menu-rollup-computed-width: rem(40px);
 
 .Header {
   @include page-header-layout;
@@ -104,7 +104,7 @@ $action-menu-rollup-computed-width: icon-size();
   .mobileView & {
     position: absolute;
     top: spacing(loose) + (control-height() / 4);
-    right: 0;
+    right: spacing(loose);
     margin-top: 0;
 
     @include page-content-when-not-fully-condensed {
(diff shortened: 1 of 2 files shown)


src/components/Page/components/Header/Header.scss
```diff
@@ -104,7 +104,7 @@ $action-menu-rollup-computed-width: icon-size();
   .mobileView & {
     position: absolute;
     top: spacing(loose) + (control-height() / 4);
-    right: 0;
+    right: spacing(loose);
```
alex-page, line 107: The padding is 20px around the header, this makes the rollup action consistent. 

The icon has some padding around it so it does not exactly line up with the edge of the padding. Maybe this would look better with a smaller padding so the icon lines up with the texts spacing?

dfmcphee: It does feel like _slightly_ too much spacing

alex-page:
> I do agree that maybe we should play with the spacing a bit but this is definitely an improvement over what is there now.

I did some tophatting this morning with a smaller right value. My opinion is we should keep the padding as is in this PR. Offsetting the `elipsis` also moves the popover position. When you open the popover it is now outside of the header padding.

Screen Shot 2020-06-22 at 8 33 33 AM

The current padding keeps the clickable area inline with the header spacing and when the popover is open it lines up with the header padding correctly. I could do some CSS to remove the right padding and add it back but that does not feel right to me. This could also be confusing to someone trying to understand why that change was made.

@HYPD @sarahill do either of you have any opinions on how the `elipsis` should be positioned here?

PR #3070, opened 2020-06-23
Update activeElement tests to use modern testing utils

BPScott:

An imminent bump to SK and thus Jest exposes a case where enzyme doesn't handle component mounting to jsdom in a way that results in spec compliant focus management, which then results in tests that assert on document.activeElement failing.
Fortunately react-testing doesn't have this problem, so we can just usethat instead.

Gory details: https://github.com/enzymejs/enzyme/issues/2337#issuecomment-608984530

Migrates a handful of test cases from enzyme to use react-testing

### How to 🎩

 Run tests and see they still pass

### DIFF
--- a/src/components/TrapFocus/tests/TrapFocus.test.tsx
+++ b/src/components/TrapFocus/tests/TrapFocus.test.tsx
@@ -1,10 +1,4 @@
 import React from 'react';
-// eslint-disable-next-line no-restricted-imports
-import {
-  mountWithAppProvider,
-  trigger,
-  ReactWrapper,
-} from 'test-utilities/legacy';
 import {mountWithApp} from 'test-utilities';
 import {
   EventListener,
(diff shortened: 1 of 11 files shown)


src/components/TrapFocus/tests/TrapFocus.test.tsx
```diff
@@ -257,15 +245,15 @@ describe('<TrapFocus />', () => {
 
       listenerMap.keydown({
         keyCode: Key.Tab,
-        target: lastNode(trapFocus),
+        target: trapFocus.find('input')!.domNode,
```
kaelig, line 248: This (and a few other instances) slightly diverges from the intent of the test but I feel good about this change.

BPScott:
> Is there any internal documentation about how to test focus that we could point to in the body of this PR?

Unfortunately there is none that I am aware of. But from working through this stuff it boils down to:

- use `react-testing`
- use `expect(document.activeElement).toBe(someNode)` instead of `expect(someNode).toBe(document.activeElement)` as that way around gives more meaaningful error messages.

PR #3072, opened 2020-06-24
Update sewing-kit to 0.132.2 and storybook to 5.3.19

BPScott:

Keeping up to date. This new SK contains:

- new versions of linting packages - reflecting updated package names - they now live under the `@shopify` namespace
- Jest update to v25

Looks scary but most of this is yarn.lock updates.

- Update references to eslint-plugin-shopify, stylelint-config-shopify,
babel-preset-shopify and postcss-shopify to account for them being
deprecated and moved to new packkages under thee `@shopify` namespace
- Add ignores for some new rules and other linting fixes
- Skip some tests in ComboBox as they cause infinite loops in Jest 25 /
  JSDom 16 and I'm not sure of the root cause.
- Do a `yarn-dedupe`

### How to 🎩

Ensure linting and tests pass, ensure HMR still works in storybook.


.nvmrc
```diff
@@ -1 +1 @@
-v10.13.0
```
BPScott, line 1: node versio bump as a dependency required this version of node.

babel.config.js
```diff
@@ -6,12 +6,10 @@ module.exports = function (api) {
   });
 
   const runtimePreset = isWeb
-    ? ['babel-preset-shopify/web', {modules: false, typescript: true}]
-    : ['babel-preset-shopify/node', {modules: 'commonjs', typescript: true}];
+    ? ['@shopify/babel-preset/web', {modules: false, typescript: true}]
+    : ['@shopify/babel-preset/node', {modules: 'commonjs', typescript: true}];
 
-  // babel-preset-shopify/react only uses HMR if hot is true and the env is
```
BPScott, line 12: The hot option was removed in https://github.com/Shopify/web-foundation/pull/145.
Storybook continues to auto reload without it.

package.json
```diff
@@ -211,8 +210,5 @@
     "hoist-non-react-statics": "^3.3.0",
     "lodash": "^4.17.4",
     "tslib": "^1.9.3"
-  },
```
BPScott, line 214: sewing-kit now uses prettier v2 out the box so we don't need to force the resolution anymore

tsconfig.json
```diff
@@ -8,8 +8,6 @@
     "jsx": "react-native",
     "isolatedModules": true,
     "importsNotUsedAsValues": "error",
-    // strictFunctionTypes is implictly enabled by strict mode but
```
BPScott, line 11: Typescript is chill with having comments in this file, but a recent SK change now tries to read this as plain json, and throws up when we have comments.

I still want to stop disabling strictFunctionTypes eventually but not now.

src/components/ResourceList/components/FilterControl/components/FilterCreator/tests/FilterCreator.test.tsx
```diff
@@ -68,49 +69,46 @@ describe('<FilterCreator />', () => {
   };
 
   it('focuses the activator after adding a filter', () => {
-    const filterCreator = mountWithAppProvider(
+    const filterCreator = mountWithApp(
```
BPScott, line 72: I missed a file in #3070

package.json
```diff
@@ -176,7 +175,7 @@
     "serve": "^11.1.0",
     "shelljs": "^0.8.3",
     "shx": "^0.3.2",
-    "storybook-chromatic": "^3.4.1",
+    "storybook-chromatic": "^3.5.5",
```
BPScott, line 178: This helped dedupe some versions of jest 24

package.json
```diff
@@ -158,11 +158,10 @@
     "postcss-modules-parser": "^1.1.1",
     "postcss-modules-scope": "^2.1.0",
     "postcss-modules-values": "^2.0.0",
-    "postcss-shopify": "^2.2.1",
     "puppeteer": "^1.20.0",
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
-    "react-test-renderer": "^16.3.1",
+    "react-test-renderer": "^16.8.6",
```
BPScott, line 164: Not strictly needed but I though this would help with some test stuff but it didn't. This should be kept in sync with the version of react/react-dom anyway.

package.json
kaelig: Might as well update to v5, what do you think? (the v4 upgrade was painless on polaris-styleguide, I expect it'll be the same for v5)

src/components/Checkbox/Checkbox.scss
```diff
@@ -76,11 +76,11 @@
         @include control-backdrop(active, $global-theming: true);
       }
       ~ .Icon {
+        transition: opacity var(--p-duration-1-5-0) var(--p-ease),
+          transform var(--p-duration-1-5-0) var(--p-ease);
         transition-timing-function: var(--p-ease-in);
```
kaelig, line 81: This will now likely override the timing function specified above (not sure why it's been kept there in the first place)

kaelig: @BPScott I just saw the change in 0da64f2f43a1acf452246e737930cea2e72b6f08 and I don't think this is the right rewrite, as the cascade in the original version would have ignored `transition-timing-function: var(--p-ease-in);`.

This should work:

```diff
+        transition: opacity var(--p-duration-1-5-0) var(--p-ease),
+          transform var(--p-duration-1-5-0) var(--p-ease);
-        transition-timing-function: var(--p-ease-in);
-        transition: opacity var(--p-duration-1-5-0) var(--p-ease),
-          transform var(--p-duration-1-5-0) var(--p-ease);
```

package.json
BPScott: done!

src/components/Checkbox/Checkbox.scss
BPScott: done!

PR #3073, opened 2020-06-24
[OrderDetails] Add OrderDetails component

benjyblack:
### Why
This is part of the Web Dev onboarding guide.

### What
image

# This is not intended to actually be merged.

### DIFF
--- a/src/components/OrderDetails/OrderDetails.tsx
+++ b/src/components/OrderDetails/OrderDetails.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, {useEffect, useState} from 'react';
+import AddressFormatter from '@shopify/address';
 
 import {ThemeProvider} from '../ThemeProvider';
 import {Card} from '../Card';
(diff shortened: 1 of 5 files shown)


src/components/OrderDetails/OrderDetails.tsx
```diff
@@ -0,0 +1,49 @@
...
+            surface: '#020c1d',
+            secondary: '#103560',
+            interactive: '#009973',
+          },
+        }}
+      >
+        <Card title="Order details">
+          <Card.Section title="Unfulfilled">
+            <Stack alignment="center" distribution="equalSpacing">
+              <Stack alignment="center">
```
benjyblack, line 29: Am I overusing Stacks here? Wasn't sure how else to do vertical/horizontal alignment

src/components/OrderDetails/OrderDetails.tsx
```diff
@@ -0,0 +1,49 @@
...
+        theme={{
+          colorScheme: 'dark',
+          colors: {
+            surface: '#020c1d',
+            secondary: '#103560',
+            interactive: '#009973',
+          },
+        }}
+      >
+        <Card title="Order details">
```
aaackerman, line 26: A card optionally takes footer actions like `primaryFooterAction`. In your case, I think it may make sense to use the stack because you also have left-aligned text. I am curious if it's possible to mix and match actions with the sections.

src/components/OrderDetails/OrderDetails.tsx
```diff
@@ -0,0 +1,49 @@
...
+export interface OrderDetailsProps {}
+
+export class OrderDetails extends React.PureComponent<OrderDetailsProps> {
+  render() {
+    return (
+      <ThemeProvider
+        theme={{
+          colorScheme: 'dark',
+          colors: {
+            surface: '#020c1d',
```
aaackerman, line 20: This makes sense because we're using a specific color theme. There's also a way to import colors from polaris if there were other specific colors from the palette we wanted.

src/components/OrderDetails/OrderDetails.tsx
aaackerman, line 26: In production, all of these texts would be translated via the [`i18n` tool](https://github.com/Shopify/quilt/tree/master/packages/react-i18n). 

```
const [i18n] = useI18n();

<Card title={i18n.translate('OrderDetails.heading')}>
```

src/components/OrderDetails/OrderDetails.tsx
```diff
@@ -0,0 +1,49 @@
...
+            secondary: '#103560',
+            interactive: '#009973',
+          },
+        }}
+      >
+        <Card title="Order details">
+          <Card.Section title="Unfulfilled">
+            <Stack alignment="center" distribution="equalSpacing">
+              <Stack alignment="center">
+                <Thumbnail
```
aaackerman, line 30: I suspect the example you posted wasn't using a thumbnail. I think this is a good approach though. :)

src/components/OrderDetails/OrderDetails.tsx
aaackerman: Not sure, this is how i would approach it to avoid extra styling changes. 🤔

dfmcphee: If you want to accomplish a similar layout with a single stack, you can use the `Stack.Item` subcomponent with the `fill` prop to have it take up as much space as possible.

```
<Stack alignment="center">
  <Thumbnail
    source="https://burst.shopifycdn.com/photos/black-leather-choker-necklace_373x@2x.jpg"
    alt="Black choker necklace"
  />
  <Stack.Item fill>
    <TextStyle variation="positive">Enamel pin</TextStyle>
  </Stack.Item>
  <span>$9.99</span>
</Stack>
```

PR #3078, opened 2020-06-24
[ActionMenu] Fix regression with new design language

tmlayton:

`Page` actions were no longer triggering their callbacks with the new design language because the callback was being passed through to `Button` as `onAction`, not `onClick`. TypeScript should have caught this, I have not looked at why this happened.

Destructuring `onAction` from the `rest` props and passing to `onClick` for `Button` or `onAction` for `MenuAction`, depending if the new design language is enabled or not.

Also adds the tests first so you can see the failure then the fix as a separate commit to verify.

- [x] 🎩

### DIFF
--- a/src/components/ActionMenu/ActionMenu.tsx
+++ b/src/components/ActionMenu/ActionMenu.tsx
@@ -86,13 +86,18 @@ export class ActionMenu extends React.PureComponent<ActionMenuProps, State> {
         ) : null;
       }
 
-      const {content, ...rest} = action;
+      const {content, onAction, ...rest} = action;
       return newDesignLanguage ? (
-        <Button key={index} {...rest}>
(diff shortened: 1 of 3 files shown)


BPScott:
Too slow on the approval but as for:

>  TypeScript should have caught this, I have not looked at why this happened.

Button says that both `url` and `onClick` are optional. That means it's fine to omit both of them.  It should probably a be a discriminated union that says "you need to specify either url or onClick". Same as how [TextField handles readonly/disabled/onChange](https://github.com/Shopify/polaris-react/blob/master/src/components/TextField/TextField.tsx#L117-L122)

tmlayton:
Do you know off-hand why did it not complain about `onAction`, which is not a prop on Button?

PR #3090, opened 2020-06-30
[hexToRgb] Add hexToRgb to src/index.ts

keidarcy:

Make  HEX colors possible to convert to other formats.

export hexToRgb from @shopify/polaris

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -7,6 +7,7 @@
 ### Enhancements
 
 - Removed padding from the details container in `EmptyState` to account for new illustration size ([#3069](https://github.com/Shopify/polaris-react/pull/3069))
+- Added `hexToRgb` to `src/index`, make HEX color possible to convert to other formats. ([#3090](https://github.com/Shopify/polaris-react/pull/3090))
 
 ### Bug fixes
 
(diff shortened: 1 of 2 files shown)


BPScott:
Hi @xyyolab, Thanks for the PR however I don't think we should accept it.

My current opinion is that these color utilities shouldn't be public exports, they should be internal utilities only. I don't believe we document or recommend their usage anywhere. When we add something to our public API surface area we make a promise to keep supporting it and it makes refactoring any internals that depend upon those utility functions much harder. Because of that I want to keep out public API surface area as small as possible - ideally only components and types for their props. There are a few exceptions to this but they're mostly down to accidents rather down to explicit decisions that something needs to be public.

I believe the only reason why these color utility functions are public currently is that they were publicly exported accidentally at some point and then when I moved some imports around it turned out that somebody was already depending on them so they had to stick around. I've not yet had the time and strength to work our how to remove these in a major version. I think the effort would be pretty low but my focus has been elsewhere.


@AndrewMusgrave thoughts?

BPScott:
Seems my esteemed colleage is busy. So gonna close this for reasons stated above - these color utility functions shouldn't be public and the only reason they are currently is due to oversight and then not wanting to break backwards compatibility.

PR #3094, opened 2020-07-03
Use main react-transition group

BPScott:
WIP - Still need to test this

We started using a fork of react-transition-group because the original didn't support strict mode.

Now the original supports strict-mode so let's start using it again.

Migrate from `@material-ui/react-transition-group` to `react-transition-group`


nicolad:
out of curiosity: why not `react-spring` ? 
one big advantage is hooks API

src/components/Frame/Frame.tsx
```diff
@@ -135,7 +135,7 @@ class FrameInner extends React.PureComponent<CombinedProps, State> {
     const navigationMarkup = navigation ? (
       <TrapFocus trapping={mobileNavShowing}>
         <CSSTransition
-          findDOMNode={this.findNavigationNode}
+          nodeRef={this.navigationNode}
```
AndrewMusgrave, line 138: Glad they made the API accept refs. Wasn't a fan of the function approach

AndrewMusgrave:
> out of curiosity: why not react-spring ?

This pull request isn't adding `react-transition-group`. It migrating from a forked version 😄

dleroux:
> out of curiosity: why not `react-spring` ?
> one big advantage is hooks API

We explored it a while back, but felt it wasn't worth the file size to add to Polaris at the time.

Issue #390, opened 2018-08-02
Header of resource-list


zheren2018:
# Issue summary

## Steps to reproduce the problem

1.
1.
1.

## Specifications

- Are you using the React components? (Y/N):
- Polaris version number:
- Browser:
- Device:
- Operating System:

---

⚓️ We’re not accepting pull requests at this time<br />
🗒 We _are_ accepting issue reports and feature requests<br />
🌟 Feature requests that we can’t get to right away will be closed. However, we track all requests and will use it to set priorities. See the [contribution guidelines](https://github.com/Shopify/polaris/blob/master/.github/CONTRIBUTING.md) for more information.

kyleboss:
This would be an amazing addition! That Orders table would be incredibly useful. 

Since it's been a year, I was wondering if the feedback was enough to see if we can bring this in to a future version of Polaris? 😄

eugene-kim:
Hi @tmlayton and @chloerice (or anyone on the Shopify team, really),

Does Shopify plan on adding something like the Shopify admin's orders list to Polaris?

Without header functionality on the resource list, at times I feel as if I'm forced to choose between UI that's too noisy or potentially unclear. If the Shopify admin orders list didn't have headers, the UI would not be nearly as user friendly as it stands today.

tmlayton:
Hey @eugene-kim, I agree this is useful and an established pattern in the Shopify admin although there is no plan to add this at the moment.

/cc @dfmcphee

eugene-kim:
Hey @tmlayton , understood. Thanks for the quick response!

camelmasa:
I'm waiting this useful feature too 😃 

For now, We built UI using React Hooks and DataTable component like below.

![movie](https://user-images.githubusercontent.com/189824/88028150-3a6b2780-cb73-11ea-814e-b4a0b13cfe6b.gif)

jamieroyce:
@camelmasa could you post example of how you implemented the example you show? I am trying to workaround no column headers in ResourceList as well and I like what you did.

camelmasa:
@jamieroyce Our code like this.  

```jsx
    <Card>
      {checkedOrders.length !== 0 ? (
        <div
          style={{
            paddingTop: '8px',
            paddingBottom: '8px',
            paddingLeft: '14px',
            borderBottom: '0.1rem solid var(--p-border, #c4cdd5)',
          }}
        >
          <ButtonGroup segmented>
            <Button onClick={reject}>Reject</Button>
[...]

PR #3097, opened 2020-07-06
Added unstable index table

AndrewMusgrave:

Adds the WIP IndexProvider & IndexTable to Polaris so they can be consumed by others. An unstable tab is being appended to the export since the API & name are unstable and are likely to change before they're included in a minor or major release.

* Adding experimental components

## 

### How to 🎩

```

```

### DIFF
--- a/src/components/experimental/IndexProvider/IndexProvider.tsx
+++ b/src/components/experimental/IndexProvider/IndexProvider.tsx
@@ -1,0 +1,205 @@
+import React, {useRef, useMemo, useCallback} from 'react';
+
+import {useI18n} from '../../../utilities/i18n';
+
+import {
+  IndexContext,
+  IndexRowContext,
+  IndexSelectionChangeContext,
+  SELECT_ALL_ITEMS,
+  IndexProviderProps,
+  SelectionType,
+  Range,
+} from './utilities/index';
+
+type HandleSelectionChange = (
+  selectionType: SelectionType,
+  toggleType: boolean,
+  selection?: string | Range,
+  sortOrder?: number,
+) => void;
+
+export function IndexProvider({
+  children,
+  resourceName: passedResourceName,
+  loading,
+  onSelectionChange = noop,
+  selectedItemsCount,
+  itemCount,
+  hasMoreItems,
+}: IndexProviderProps) {
+  const selectable = Boolean(selectedItemsCount);
+
+  const i18n = useI18n();
+  const lastSelected = useRef<number | null>(null);
+
+  const defaultResourceName = {
+    singular: i18n.translate('Polaris.IndexProvider.defaultItemSingular'),
+    plural: i18n.translate('Polaris.IndexProvider.defaultItemPlural'),
+  };
+
+  const selectMode = selectedItemsCount === 'All' || selectedItemsCount > 0;
+
+  const handleSelectionChange: HandleSelectionChange = useCallback(
+    (
+      selectionType: SelectionType,
+      toggleType: boolean,
+      selection?: string | Range,
+      sortOrder?: number,
+    ) => {
+      const prevSelected = lastSelected.current;
+
+      if (SelectionType.Multi && typeof sortOrder === 'number') {
+        lastSelected.current = sortOrder;
+      }
+
+      if (
+        selectionType === SelectionType.Single ||
+        (selectionType === SelectionType.Multi &&
+          (typeof prevSelected !== 'number' || typeof sortOrder !== 'number'))
+      ) {
+        onSelectionChange(SelectionType.Single, toggleType, selection);
+      } else if (selectionType === SelectionType.Multi) {
+        const min = Math.min(prevSelected as number, sortOrder as number);
+        const max = Math.max(prevSelected as number, sortOrder as number);
+        onSelectionChange(selectionType, toggleType, [min, max]);
+      } else if (
+        selectionType === SelectionType.Page ||
+        selectionType === SelectionType.All
+      ) {
+        onSelectionChange(selectionType, toggleType);
+      }
+    },
+    [lastSelected, onSelectionChange],
+  );
+
+  let bulkSelectState: boolean | 'indeterminate' | undefined = 'indeterminate';
+  if (!selectedItemsCount || selectedItemsCount === 0) {
+    bulkSelectState = undefined;
+  } else if (
+    selectedItemsCount === SELECT_ALL_ITEMS ||
+    selectedItemsCount === itemCount
+  ) {
+    bulkSelectState = true;
+  }
+
+  const resourceName = passedResourceName
+    ? passedResourceName
+    : defaultResourceName;
+
+  const paginatedSelectAllText = getPaginatedSelectAllText();
+
+  const bulkActionsLabel = getBulkActionsLabel();
+  const bulkActionsAccessibilityLabel = getBulkActionsAccessibilityLabel();
+
+  const contextValue = useMemo(
+    () => ({
+      itemCount,
+      selectMode,
+      selectable,
+      resourceName,
+      loading,
+      paginatedSelectAllText,
+      hasMoreItems,
+      bulkActionsLabel,
+      bulkActionsAccessibilityLabel,
+      bulkSelectState,
+      selectedItemsCount,
+    }),
+    [
+      itemCount,
+      selectMode,
+      selectable,
+      resourceName,
+      loading,
+      paginatedSelectAllText,
+      hasMoreItems,
+      bulkActionsLabel,
+      bulkActionsAccessibilityLabel,
+      bulkSelectState,
+      selectedItemsCount,
+    ],
+  );
+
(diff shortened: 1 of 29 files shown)


AndrewMusgrave:
Liting rules seem to be broken for this directory - looking for a fix but other than that it's ready for 👀

src/components/experimental/IndexProvider/utilities/hooks.ts
```diff
@@ -0,0 +1,31 @@
+import {useContext} from 'react';
+
+import {
+  IndexContext,
+  IndexRowContext,
+  IndexSelectionChangeContext,
+} from './context';
+
+export function useIndexRow() {
```
dleroux, line 9: what do you think of being a little more explicit with the naming and do `useIndexRowContext`, `useIndexContext`, ...

src/components/experimental/IndexProvider/IndexProvider.tsx
```diff
@@ -0,0 +1,205 @@
...
+  IndexContext,
+  IndexRowContext,
+  IndexSelectionChangeContext,
+  SELECT_ALL_ITEMS,
+  IndexProviderProps,
+  SelectionType,
+  Range,
+} from './utilities/index';
+
+type HandleSelectionChange = (
```
dleroux, line 15: Should this type be used in your `IndexSelectionChangeContext` This one has a `sortOrder` and the other a `position`

src/components/experimental/IndexProvider/IndexProvider.tsx
```diff
@@ -0,0 +1,205 @@
...
+      selectMode,
+    }),
+    [selectMode],
+  );
+
+  return (
+    <IndexContext.Provider value={contextValue}>
+      <IndexRowContext.Provider value={rowContextValue}>
+        <IndexSelectionChangeContext.Provider
+          value={handleSelectionChange as any}
```
dleroux, line 135: as any?

src/components/experimental/IndexProvider/tests/IndexProvider.test.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
```
dleroux, line 1: Were you planning on adding more tests for this?

src/components/experimental/IndexProvider/tests/use-index-row.test.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
```
dleroux, line 1: Same question here. More tests coming?

src/components/experimental/IndexTable/IndexTable.scss
```diff
@@ -0,0 +1,320 @@
...
+}
+
+$loading-panel-height: rem(53px);
+
+.LoadingPanel {
+  position: absolute;
+  z-index: z-index(loading-panel, $index-table-stacking-order);
+  top: 0;
+  left: 0;
+  display: flex;
```
dleroux, line 48: no need to change this but the other day I learned about: 
```
display: grid;
place-items: center;
```

to center stuff, which I thought was neat.

src/components/experimental/IndexTable/IndexTable.scss
```diff
@@ -0,0 +1,320 @@
...
+  box-shadow: shadow();
+}
+
+.Table {
+  width: 100%;
+  min-width: 100%;
+  border-collapse: collapse;
+}
+
+.Table--scrolling {
```
dleroux, line 63: In order for the styles to be properly generated in Polaris, it would be `.Table-scrolling`. Same for the other selectors.

dleroux: Just a thought, you could have also use `td:first-of-type` and `td:nth-of-type(2)`

src/components/experimental/IndexTable/IndexTable.scss
```diff
@@ -0,0 +1,320 @@
...
+.TableHeading {
+  padding: spacing();
+  text-align: left;
+  font-weight: 500;
+  white-space: nowrap;
+  border: 0;
+}
+
+.TableHeading--first {
+  position: sticky;
```
dleroux, line 130: ❤️

dleroux:
Might be a good idea to add the design language before shipping?

src/components/experimental/IndexTable/IndexTable.scss
```diff
@@ -0,0 +1,320 @@
...
+  border-collapse: collapse;
+}
+
+.Table--scrolling {
+  .TableCell--first,
+  .TableCell--second,
+  .TableHeading--first,
+  .TableHeading--second {
+    visibility: visible;
+    background-color: color('white');
```
dleroux, line 69: I think this background colour should be added even without scrolling. If you scroll fast you can see the text behind the cell and then disappears when the class is applied.

dleroux:
So after our conversation, I breezed through the remainder of the PR.

Since this is a new approach for experimental components, what sort of guidelines should we follow when reviewing? What does experimental mean?

dleroux:
@BPScott @chloerice  just pinged you both on this because I'd like your thoughts on this approach of adding experimental components.

src/components/index.ts
```diff
@@ -128,6 +128,17 @@ export {Icon} from './Icon';
 export {Image} from './Image';
 export type {ImageProps} from './Image';
 
+export {
+  IndexTable as UnstableIndexTable,
+  IndexProvider as UnstableIndexProvider,
+  useIndexTableRowHovered as unstableUseIndexTableRowHovered,
```
dleroux, line 134:
```suggestion
  useIndexTableRowHovered as UnstableUseIndexTableRowHovered,
```

AndrewMusgrave:
@BPScott @chloerice We'll message you next week internally with more context 😄

src/components/experimental/IndexProvider/utilities/types.ts
```diff
@@ -0,0 +1,31 @@
+export type IndexSelectedItems = string[] | 'All';
+
+export const SELECT_ALL_ITEMS = 'All';
+
+export enum SelectionType {
```
BPScott, line 5: I've been having some confusing feelings about exported enums recently when compared to using union types. 

I can't yet fully articulate it but I think it boils down to: "importing a new object instead of using a string value is slight extra complication. It adds surface area to our API and it is slightly harder to teach enums compared to 'onSelectionChange's first argument is a string that is either `all`, `page` `multi` or `single`". It's not a big difference (which is why I'm finding it hard to have a strong opinion) but I think I'm slightly leaning towards union types.

We tend to use union types in almost all other places (PopoverCloseSource, and DatePicker are the current exceptions, and [I'm looking at removing that in DatePicker](https://github.com/Shopify/polaris-react/pull/3113)) - e.g Button's size prop is `slim | medium | large` not `enum ButtonSize { Slim = "slim", Medium = "medium", Large = large}1


What do you think about using a union type `type SelectionType = 'all' | 'page' | 'multi' | single` instead?

- You can't iterate through all possible options at runtime (but we don't need to do this)
-

BPScott:
I'm liking the experimental folder and exporting components/props/types with an unstable prefix in `src/components/index.ts` as a way to bucket these things as not suitable for general use!

@amrocha had a similar need yesterday - We had to bring back [ResouceList's FilterControl](https://github.com/Shopify/polaris-react/pull/3116) in v5 - we hoped to fully remove it and push people towards using `Filters` but there's one rather pervasive usage in web that can't be migrated to `Filters` that we didn't want to block release for. We couldn't fork the component into web as it relies upon private context, so restoring the component but with a scary export name was deemed better than exporting the ResourceList context.

These two cases sound kinda similar:

- Unstable (and thus may have breaking changes in a minor version) because it's experimental and we're working out the kinks
- Unstable (and thus may be removed in a minor version) because we removed 99% of it's external usage but there's one last knotty usage and we can't just fork the component

Is there a naming system to combine them that makes sense? I've had a bit of a think and nothing is coming to mind - keeping that separation of "on its way in" and "on its way out" feels important. What do ya reckon?

chloerice:
> @BPScott @chloerice just pinged you both on this because I'd like your thoughts on this approach of adding experimental components.

Hey y'all! I don't have strong opinions on format (though I agree with Ben, I like the Unstable naming convention). I'd love to understand what the goals are of the experimental components. In my opinion, if we're sharing this in the public package even experimental components should be documented. Given that they unstable, if the goal is transparency the documentation while experimental could act as a decision log as things are changed/improved/stabilized.

Alexzanderk:
Guys, do you have any upd on this?

LukaszWiktor:
I came here from #390. I'm also looking for a ResourceList-like component with a header.

GhostApps:
@Alexzanderk @LukaszWiktor you want to follow #3646 for updates seems close, don't know why its not been merged yet

AndrewMusgrave:
Closing as this has been added in https://github.com/Shopify/polaris-react/pull/3646

PR #3098, opened 2020-07-06
[v5] Remove withAppProvider

BPScott:

We're down to having a handful of components using withAppProvider. Let's make the final push to remove it.

- Convert Page to be a functional component
- Convert other usages of withAppProvider to use bespoke wrappers. These components are large and complex so a migration to being functional components will take time. This is a minimal approach, that is functionally very similar to withAppProvider but less convoluted. 

### How to 🎩

Confirm affected components look as expected in storybook


src/components/DataTable/DataTable.tsx
```diff
@@ -441,4 +438,7 @@ class DataTableInner extends React.PureComponent<
   };
 }
 
-export const DataTable = withAppProvider<DataTableProps>()(DataTableInner);
+export function DataTable(props: DataTableProps) {
```
amrocha, line 441: Huh, why did you do it this way?

src/components/Page/Page.tsx
```diff
@@ -18,46 +14,26 @@ export interface PageProps extends HeaderProps {
...
+
+  return (
+    <div className={className}>
+      {headerMarkup}
+      <div className={styles.Content}>{children}</div>
+    </div>
+  );
 }
-
-export const Page = withAppProvider<PageProps>()(PageInner);
```
amrocha, line 63: Was this even using AppProvider before?

amrocha:
>These components are large and complex so a migration to being functional components will take time. This is a minimal approach, that is functionally very similar to withAppProvider but less convoluted.

Gotcha, makes sense to me and this lets us move forward and remove with-app-provider 👍

src/components/Page/Page.tsx
BPScott: It was in v4. As part of ripping out app-bridge this component stopped using the `polaris` prop that withAppBridge provided, so yeah it not actually using what withAppProvider provided in the v5 branch.

I could have left this change as just removing the `withAppProvider` wrapping and having Page export a class component, but it was easy enough to migrate to a functional component so why not :)

PR #3101, opened 2020-07-07
Update browser config

AndrewMusgrave:

Update to the shared browser config

### DIFF
--- a/UNRELEASED-v5.md
+++ b/UNRELEASED-v5.md
@@ -26,6 +26,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Dependency upgrades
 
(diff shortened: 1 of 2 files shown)


BPScott:
Nice!

We shouldn't need to specify `@shopify/browserslist-config` directly in package.json as it's already a transitive dependency of sewing-kit (and this config like eslint config / stylelint config is the one case that we do want to rely on the transitive dependency instead of being explicit)

BPScott:
Oh I've just pushed a commit to the v5 branch that [fixes up that package.json/yarn.lock noise](https://github.com/Shopify/polaris-react/commit/d51727b6cbf9a795e64a8578c9fdc0c544c32a57) rebase and you can get rid of those changes.

BPScott:
Took the liberty of acting upon the changes requested so I can get this merged. Thanks!

AndrewMusgrave:
@BPScott Thanks 🙇

PR #3102, opened 2020-07-09
[v5] Reduce dependency on javascript-utilities 

BPScott:

The `@shopify/javascript-utilities` package is OOOOLD, deprecated and doesn't export which makes it hard to tree shake.

Many of the utilities provided by this package can be either replaced with browser-native implementations (given browser support has moved on since the time that package was written) or inlined into our repo at minimal cost.

This PR doesn't fully remove our dependency on `@shopify/javascript-utilities` - only dependencies on the date helpers remain, but that'll require a slightly more involved refactor, which I'll do in a follow up PR.

Replace several of our dependencies on `@shopify/javascript-utilities` with alternatives. See each commit for specific details

#### Use native addEventListener/removeEventListener
The wrappers were needed for browsers that don't support all the "once"
and "passive" options, but now all browsers we support can use those
options.

See:
- https://github.com/Shopify/javascript-utilities/blob/master/src/events.ts
- https://caniuse.com/#feat=once-event-listener
- https://caniuse.com/#feat=passive-event-listener

#### Inline focus utilities

- Rename our already existing `findFirstFocusableNode` to
  `findFirstFocusableNodeIncludingDisabled`
- Copy implementations of findFirstFocusableNode and
  focusFirstFocusableNode into utilities

#### Use native closest

The wrapper was needed for browsers that didn't support
domNode.closest() but now all browsers we support have this
functionality

See
- https://github.com/Shopify/javascript-utilities/blob/master/src/dom.ts
- https://caniuse.com/#feat=element-closest

#### Remove usage of compose as it is not tree-shakeable

Ok this isn't related to the javascript-utilities package but it does hurt treeshakability

#### Use requestAnimationFrame instead of fastdom

The "fastdom" helpers are just wrappers around this

See
- https://github.com/Shopify/javascript-utilities/blob/master/src/fastdom.ts

#### Inline nodeContainsDescendant usage

See:
- https://github.com/Shopify/javascript-utilities/blob/master/src/dom.ts
- https://developer.mozilla.org/en-US/docs/Web/API/Node/contains

### How to 🎩

Ensure tests pass. Click around the affected components.


src/components/DescriptionList/DescriptionList.tsx
```diff
@@ -15,17 +14,17 @@ export interface DescriptionListProps {
   items: Item[];
 }
 
-const getUniqueTermKey = createUniqueIDFactory(`Term`);
-const getUniqueDescriptionKey = createUniqueIDFactory(`Description`);
-
 export function DescriptionList({items}: DescriptionListProps) {
+  // There's no good key to give React so using the index is a last resport.
+  // we can't use the term/description value as it may be a react component
+  // which can't be stringified
```
amrocha, line 20: Why not generate unique IDs here too?

BPScott: keys only need to be unique within the current loop so there's no extra value gained by using a globally unique value instead of a locally unique value (i.e. the loop index)

PR #3104, opened 2020-07-11
[ AppProvider] Update AppProvider Testing Components section for version 4

jprosevear:

The sample test app is updated to use PolarisTestProvider but this documentation snippet still refers to the old 3.x functions.

Changing the AppProvider docs

### 🎩 checklist


BPScott:
Good spot, thanks!

PR #3105, opened 2020-07-13
Re-optimize empty search SVG file

kaelig:
As noted by @BPScott, the empty-search SVG file can be further optimized.

The optimization degrades slightly the visual quality of the illustration, in a way that's not perceivable to users.


kaelig:
Going further down in precision degrades the image perceivably (from experience, I don't think we'll be able to squeeze that much more out of this illustration).

PR #3107, opened 2020-07-14
Build: Avoid repeating babel config in rollup config

BPScott:
 Not repeating myself

Rollup will now use the config in babel.config.js as a base and augment
when needed


BPScott:
@amrocha hey neat, I found a way to avoid repeating all our babel config in our rollup config

PR #3108, opened 2020-07-14
Remove dependency on @shopify/javascript-utilities

BPScott:

Removes dependency on deprecated package

Date functions have been moved into utilities/dates, though they are
only used within the DatePicker component.

This goes to highlight the terrible ergonomics of the Weekday/Month enums. People already use DatePicker by passing in numbers to `month` and `weekStartsOn` number,  which is what our example does e.g. `<DatePicker year={2019} month={0} weekStartsOn={0} />` instead of an enum value e.g. `<DatePicker year={2019} month={Months.January} weekStartsOn={Weekdays.Sunday} />` (note that they would have had to import `Weekdays` from the javascript-utilities package as we never exposed it). Because of that I'm highly tempted to convert these enums into just `number` to avoid needing to export additional enums that are just nice-to-haves (I tried using a union type `0|1|2|3...` but that means consumers have do do annoying casting which doesn't feel worth it - they never had to befor and didn't have problems).

General rule - no enums in props - use union types instead (note that this the case for most other places e.g. Button size, DatePicker and PopoverCloseSource are the two exceptions.

### How to 🎩

 Check functionality of DatePicker

### DIFF
--- a/UNRELEASED-v5.md
+++ b/UNRELEASED-v5.md
@@ -35,6 +35,6 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Removed `withAppProvider` higher-order component. ([#3098](https://github.com/Shopify/polaris-react/pull/3098))
 - Removed several dependencies on the deprecated `@shopify/javascript-utilities` library ([#3102](https://github.com/Shopify/polaris-react/pull/3102))
 - Removed dependency on `@shopify/useful-types`
-- Remove dependency on `@shopify/javascript-utilities` ([#3108](https://github.com/Shopify/polaris-react/pull/3108))
+- Removed dependency on `@shopify/javascript-utilities` ([#3108](https://github.com/Shopify/polaris-react/pull/3108))
 
 ### Deprecations
--- a/UNRELEASED-v5.md
+++ b/UNRELEASED-v5.md
@@ -35,5 +35,6 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Removed `withAppProvider` higher-order component. ([#3098](https://github.com/Shopify/polaris-react/pull/3098))
 - Removed several dependencies on the deprecated `@shopify/javascript-utilities` library ([#3102](https://github.com/Shopify/polaris-react/pull/3102))
 - Removed dependency on `@shopify/useful-types`
+- Remove dependency on `@shopify/javascript-utilities` ([#3108](https://github.com/Shopify/polaris-react/pull/3108))
 
 ### Deprecations
--- a/src/components/DatePicker/DatePicker.tsx
+++ b/src/components/DatePicker/DatePicker.tsx
@@ -1,20 +1,19 @@
 import React, {useState, useEffect, useCallback, useMemo} from 'react';
 import {ArrowLeftMinor, ArrowRightMinor} from '@shopify/polaris-icons';
+
+import {Button} from '../Button';
+import {classNames} from '../../utilities/css';
 import {
-  Range,
-  Months,
-  Year,
+  Month as Months,
+  Weekday,
   isDateAfter,
   isDateBefore,
   getNextDisplayYear,
   getNextDisplayMonth,
   getPreviousDisplayYear,
   getPreviousDisplayMonth,
-  Weekdays,
-} from '@shopify/javascript-utilities/dates';
-
-import {Button} from '../Button';
-import {classNames} from '../../utilities/css';
+} from '../../utilities/dates';
+import type {Range, Year} from '../../utilities/dates';
 import {useI18n} from '../../utilities/i18n';
 import {useFeatures} from '../../utilities/features';
 
@@ -22,6 +21,7 @@ import {monthName} from './utilities';
 import {Month} from './components';
 import styles from './DatePicker.scss';
 
+// Export as Months for public facing backwards compat
 export {Months};
 export type {Range, Year};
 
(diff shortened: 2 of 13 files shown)


src/components/DatePicker/DatePicker.tsx
```diff
@@ -1,27 +1,27 @@
 import React, {useState, useEffect, useCallback, useMemo} from 'react';
 import {ArrowLeftMinor, ArrowRightMinor} from '@shopify/polaris-icons';
+
+import {Button} from '../Button';
+import {classNames} from '../../utilities/css';
 import {
-  Range,
-  Months,
-  Year,
+  Month as Months,
```
kyledurand, line 7: This dance with the renaming and export seems a bit confusing

src/components/DatePicker/components/Month/tests/Month.test.tsx
```diff
@@ -11,7 +10,7 @@ describe('<Month />', () => {
   describe('title', () => {
     it('passes the abbreviated value to the title prop of Weekday', () => {
       const month = mountWithAppProvider(
-        <Month month={0} year={2018} weekStartsOn={Weekdays.Monday} />,
+        <Month month={0} year={2018} weekStartsOn={1} />,
```
kyledurand, line 13: So this is where we're avoiding casting now? I still prefer the enum as a consumer I think. It's no super clear that 1 represents Monday

src/utilities/dates.ts
```diff
@@ -0,0 +1,186 @@
...
+  November = 10,
+  December = 11,
+}
+
+export interface Range {
+  start: Date;
+  end: Date;
+}
+
+export type Year = number;
```
kyledurand, line 31: This doesn't seem particularly helpful 😄

src/utilities/dates.ts
```diff
@@ -0,0 +1,186 @@
...
+  const firstOfMonth = new Date(year, month, 1);
+  const firstDayOfWeek = firstOfMonth.getDay();
+  const weeks: Week[] = [[]];
+
+  let currentWeek = weeks[0];
+  let currentDate = firstOfMonth;
+
+  const orderedWeekday = getOrderedWeekdays(weekStartsOn);
+  for (let i = 0; i < orderedWeekday.indexOf(firstDayOfWeek); i++) {
+    currentWeek.push(null);
```
kyledurand, line 51: Does this render blank squares in the date picker?

src/utilities/dates.ts
```diff
@@ -0,0 +1,186 @@
...
+    return false;
+  }
+  const {start, end} = range;
+
+  return Boolean(
+    (start && isSameDay(start, day)) || (end && isSameDay(end, day)),
+  );
+}
+
+export function isSameDay(day1: Date, day2: Date) {
```
kyledurand, line 92: I think we have some already established date helpers in quilt: https://github.com/Shopify/quilt/tree/master/packages/dates#issameday

src/components/DatePicker/DatePicker.tsx
BPScott: Yeah it's horrible :( In this initial commit i was trying to maintain backwards compatibility and satisfy 3 things:
- linting says to prefer singular enums (so Month, not Months), so name it right in the first place
- We have a public export called Months, and that API compatibility should be kept (for the moment, with renaming that in either a follow up either in this PR or in a new one)
- DatePicker already has an import called Month

Looking at this has made me realize that our DatePicker API isn't that nice and it's i18n support leaves something to be desired.

I chatted with @kaelig and we were thinking:

- Replace typings in DatePicker that reference `Month`, `Year` and `Weekday` with plain old `number`. As you mention below Year isn't a useful abstraction. And the `Month` / `Weekday` enums don't add a great deal either:  `Month.February` instead of `1` and `Weekday.Monday` instead of `1` as alternatives are slightly nicer but not nice enough to warrant having to have an extra export.
- the Month and Weekdays enums don't a lot of value - remove them

src/components/DatePicker/components/Month/tests/Month.test.tsx
BPScott: Turns out I was a little wrong with my original assertions.

Doing it either of these ways was always possible in consumers (and continues to be so for the moment). However we never exported the `Weekdays` enum, consumers had to get at it via `import {Weekdays} from '@shopify/javascript-utilities/date';`. which was practically never done (`Weekday.SOMEDAY` was used 3 times in web, two of which are are in the tests of a forked DatePicker).

This change didn't need to occur, but it means I can avoid import `Weekdays` within this test file

src/utilities/dates.ts
BPScott: Yep. The original dates file from javascript-utilities had this function but for some reason we were [using a duplicated version](https://github.com/Shopify/polaris-react/pull/3108/files#diff-5ef27a053fac0e45adb0486dbca21da4L80), so I updated that callsite to use this version instead

src/utilities/dates.ts
BPScott: True, however this is a somewhat lazy conversion as I've copied https://github.com/Shopify/javascript-utilities/blob/master/src/dates.ts and removed the functions we weren't using. Happy to move to other libs in follow up PRs, but I didn't want to spend too much time on this right now.

UNRELEASED-v5.md
```diff
@@ -35,5 +35,6 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Removed `withAppProvider` higher-order component. ([#3098](https://github.com/Shopify/polaris-react/pull/3098))
 - Removed several dependencies on the deprecated `@shopify/javascript-utilities` library ([#3102](https://github.com/Shopify/polaris-react/pull/3102))
 - Removed dependency on `@shopify/useful-types`
+- Remove dependency on `@shopify/javascript-utilities` ([#3108](https://github.com/Shopify/polaris-react/pull/3108))
```
BPScott, line 38:
```suggestion
- Removed dependency on `@shopify/javascript-utilities` ([#3108](https://github.com/Shopify/polaris-react/pull/3108))
```

PR #3113, opened 2020-07-15
[v5] Date picker missing translation

BPScott:

Follow up to #3108, where I discovered that the API of DatePicker is clunky without adding much type safety, so we could strip it back a little. #3121 made breaking changes to types, and this PR fixes a case where translations weren't being used for an aria label.

Replaces enum usage with translation keys


src/components/DatePicker/components/Day/Day.tsx
```diff
@@ -74,7 +75,9 @@ export const Day = memo(function Day({
     (focused || selected || today || date === 1) && !disabled ? 0 : -1;
   const ariaLabel = [
     `${today ? i18n.translate('Polaris.DatePicker.today') : ''}`,
-    `${Month[day.getMonth()]} `,
+    `${i18n.translate(
```
BPScott, line 78: So uh, it turns out here we were just using the enum key name (which would always be english) instead of actually translating month name.

src/components/DatePicker/components/Weekday/Weekday.tsx
```diff
@@ -21,7 +20,7 @@ export const Weekday = memo(function Weekday({
   );
 
   return (
-    <div aria-label={WeekdayEnum[label]} className={className}>
```
BPScott, line 24: So uh, it turns out here we were just using the enum key name (which would always be english) instead of actually translating weekday name.

I've updated this so we pass in the translated key into the `label` prop from Month.tsx.

src/utilities/dates.ts
```diff
@@ -125,29 +98,29 @@ export function getNewRange(range: Range | undefined, selected: Date): Range {
   return {start: selected, end: selected};
 }
 
-export function getNextDisplayMonth(month: Month): Month {
-  if (Month[month] === Month[11]) {
+export function getNextDisplayMonth(month: number): number {
+  if (month === 11) {
```
BPScott, line 102: "what about numbers bigger than 11" I hear you cry. Well it's always been possible to pass numbers bigger than 11 into this function and it's not been a problem yet :/

src/components/DatePicker/components/Day/Day.tsx
kyledurand: Good catch!

UNRELEASED-v5.md
```diff
@@ -13,6 +13,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
...
 - Moved styles from `global.scss` to `AppProvider`. This change only affects applications using the `esnext` build (applications importing `@shopify/polaris/styles.css` aren’t affected), who no longer need to import the `@shopify/polaris/esnext/global.scss` file. An empty `global.scss` was kept in, to ensure applications using sewing-kit \<v0.113.0 still build ([#2392](https://github.com/Shopify/polaris-react/pull/2392))
 - Reversed the precedence of the language dictionaries passed into the `AppProvider`’s `i18n` prop. When passing an array of dictionaries the first dictionary should be your prefered language, followed by any fallback languages. ([#2572](https://github.com/Shopify/polaris-react/pull/2572))
+- Removed `Year` type export (used by the DatePicker's props). Replace its usage with `number`. ([#3113](https://github.com/Shopify/polaris-react/pull/3113))
+- Removed the `Month` enum export (used by the DatePicker's props). Replace its usage with a number from 0 to 11, representing the number of the month in question - `Month.January` bcomes `0`, `Month.December` becomes `11` etc. ([#3113](https://github.com/Shopify/polaris-react/pull/3113))
```
kyledurand, line 17:
```suggestion
- Removed the `Month` enum export (used by the DatePicker's props). Replace its usage with a number from 0 to 11, representing the number of the month in question - `Month.January` becomes `0`, `Month.December` becomes `11` etc. ([#3113](https://github.com/Shopify/polaris-react/pull/3113))
```

src/components/DatePicker/components/Weekday/Weekday.tsx
kaelig: Worth mentioning in the release notes?

BPScott:
Done a merry dance with this so that we could keep the same translations requests rather than canceling a bunch then making new ones. #3121 contains all the type changey stuff, this PR is now just the translation fix

PR #3116, opened 2020-07-15
Add hidden filter control

amrocha:

We removed this component a year ago and thought we would have removed all instances of it in web by the time we released 5.0. This has not been the case, partly because the new `Filters` component which is the recommended upgrade path does not support all the features that `FilterControl` does.

Let's add this feature to the `Filters` component to enable these upgrades to happen, but in the meantime, to enable 5.0 to be released, let's add the `FilterControl` component back, and keep it undocumented, and add a warning that we might remove this component in a minor version bump in the future.

I tried adding this in web as a forked component but it relies on internal exports, so we'd have to expand the public API of polaris to do that. At that point I figured it'd be easier to add the component back.

Add the component back that was removed here https://github.com/Shopify/polaris-react/pull/2047/files
Add a severe deprecation notice

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist


locales/de.json
```diff
@@ -75,8 +75,8 @@
       }
     },
     "DiscardConfirmationModal": {
-      "title": "Verwerfen Sie alle nicht gespeicherten Änderungen",
-      "message": "Wenn Sie Änderungen verwerfen, werden alle Änderungen gelöscht, die Sie seit dem letzten Speichern vorgenommen haben.",
+      "title": "Verwirf alle nicht gespeicherten Änderungen",
```
amrocha, line 78: I checked these out from master, this won't show up in the diff once the base branch is rebased on master

locales/de.json
```diff
@@ -162,6 +162,15 @@
         "moreActionsActivatorLabel": "Weitere Aktionen",
         "warningMessage": "Um eine bessere Benutzererfahrung zu bieten. Es sollte nur maximal {maxPromotedActions} beworbene Aktionen geben."
       },
+      "FilterCreator": {
```
amrocha, line 165: These are the old keys that were deleted

src/components/index.ts
```diff
@@ -211,7 +211,15 @@ export {ResourceItem} from './ResourceItem';
 export type {ResourceItemProps} from './ResourceItem';
 
 export {ResourceList} from './ResourceList';
-export type {ResourceListProps} from './ResourceList';
+export type {
+  ResourceListProps,
+  FilterControlProps,
```
amrocha, line 216: Maybe it makes more sense to just copy these types in web, that way we don't expand the public API of Polaris. Thoughts?

PR #3121, opened 2020-07-20
[v5] Date picker type juggling, take 2

BPScott:
This PR is the same as #3113, except without requiring translations, as we're time constrained and want to test these breaking changes without waiting for the translations.

---

Follow up to #3108, where I discovered that the API of DatePicker is clunky without adding much type safety, so we could strip it back a little.

Remove usage of enums and replace them with plain numbers as it results
in a smaller API surface area without having a negative affect on type
strictness.

- BREAKING: Removed export of `Year` type and `Months` enum
- DatePickerProps['month'] was an enum `Months` (representing the numbers
  0 to 11), it is now a plain `number` instead
- DatePickerProps['weekStartsOn'] was an enum `Weekdays` (representing
the numbers 0 to 6), it is now a plain `number` instead
- DatePickerProps['year'] was type `Year`, it is now `number`
  (Year was a type alias of number anyway)

Taking a look at web it seems like most people passed numbers into DatePicker instead of using the enums anyway so I'm looking at this as "removing a corner case that was rarely used" instead of a deep change of functionality.

This always worked and still does:

```jsx
import {DatePicker} from '@shopify/polaris';

<DatePicker year={2019} month={0} weekStartsOn={0} />
```

The following no longer works and will need to be changed to the above:

```jsx
import {Weekdays} from '@shopify/javascript-utilities';
import {DatePicker, Months} from '@shopify/polaris';

<DatePicker year={2019} month={Months.January} weekStartsOn={Weekdays.Sunday} />
```

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -15,14 +15,16 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Reversed the precedence of the language dictionaries passed into the `AppProvider`’s `i18n` prop. When passing an array of dictionaries the first dictionary should be your prefered language, followed by any fallback languages. ([#2572](https://github.com/Shopify/polaris-react/pull/2572))
 - Removed `centeredLayout` prop in `EmptyState`. All layouts within the new design language context will be center aligned ([#3111](https://github.com/Shopify/polaris-react/pull/3111))
 - Updated types of `DatePicker` component - `month`,`year` `weekStartsOn` are now typed as plain `number` - functionality remains identical as the former types effectivly ended up being aliases of `number` anyway ([#3113](https://github.com/Shopify/polaris-react/pull/3113))
-- Removed `Year` type export (used by the DatePicker's props). Replace its usage with `number`. ([#3113](https://github.com/Shopify/polaris-react/pull/3113))
-- Removed the `Month` enum export (used by the DatePicker's props). Replace its usage with a number from 0 to 11, representing the number of the month in question - `Month.January` becomes `0`, `Month.December` becomes `11` etc. ([#3113](https://github.com/Shopify/polaris-react/pull/3113))
+- Removed `Year` type export (used by the DatePicker's props). Replace its usage with `number`. ([#3121](https://github.com/Shopify/polaris-react/pull/3121))
+- Removed the `Month` enum export (used by the DatePicker's props). Replace its usage with a number from 0 to 11, representing the number of the month in question - `Month.January` becomes `0`, `Month.December` becomes `11` etc. ([#3121](https://github.com/Shopify/polaris-react/pull/3121))
 
 ### Enhancements
 
 - Added an activator prop to `Modal` so that focus can be returned to it when the `Modal` is closed ([#2206](https://github.com/Shopify/polaris-react/pull/2206))
 
 ### Bug fixes
+
+- Fixed case where `DatePicker` did not translate the month name in an aria label ([#3121](https://github.com/Shopify/polaris-react/pull/3121))
 
 ### Documentation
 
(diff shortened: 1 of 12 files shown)


UNRELEASED.md
```diff
@@ -14,6 +14,9 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Moved styles from `global.scss` to `AppProvider`. This change only affects applications using the `esnext` build (applications importing `@shopify/polaris/styles.css` aren’t affected), who no longer need to import the `@shopify/polaris/esnext/global.scss` file. An empty `global.scss` was kept in, to ensure applications using sewing-kit \<v0.113.0 still build ([#2392](https://github.com/Shopify/polaris-react/pull/2392))
 - Reversed the precedence of the language dictionaries passed into the `AppProvider`’s `i18n` prop. When passing an array of dictionaries the first dictionary should be your prefered language, followed by any fallback languages. ([#2572](https://github.com/Shopify/polaris-react/pull/2572))
 - Removed `centeredLayout` prop in `EmptyState`. All layouts within the new design language context will be center aligned ([#3111](https://github.com/Shopify/polaris-react/pull/3111))
+- Updated types of `DatePicker` component - `month`,`year` `weekStartsOn` are now typed as plain `number` - functionality remains identical as the former types effectivly ended up being aliases of `number` anyway ([#3113](https://github.com/Shopify/polaris-react/pull/3113))
```
kaelig, line 17: It'd be worth pointing out that this is being consistent with the JavaScript native typing for these things.

UNRELEASED.md
```diff
@@ -14,6 +14,9 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
...
+- Removed `Year` type export (used by the DatePicker's props). Replace its usage with `number`. ([#3113](https://github.com/Shopify/polaris-react/pull/3113))
+- Removed the `Month` enum export (used by the DatePicker's props). Replace its usage with a number from 0 to 11, representing the number of the month in question - `Month.January` becomes `0`, `Month.December` becomes `11` etc. ([#3113](https://github.com/Shopify/polaris-react/pull/3113))
```
kaelig, line 19: Can you mention that this fixes a bug where the name of the day wouldn't get translated?

src/components/DatePicker/DatePicker.tsx
```diff
@@ -42,12 +38,14 @@ export interface DatePickerProps {
   disableDatesAfter?: Date;
   /** The selection can span multiple months */
   multiMonth?: boolean;
-  /** First day of week. Sunday by default */
-  weekStartsOn?: Weekday;
+  /** First day of week, from 0 to 6. 0 is Sunday, 1 is Monday ... 6 is Saturday
+   * @default 0
+   */
```
kaelig, line 43: Not sure it matters or if typedoc will like this, but multiline comment blocks typically are typically formatted like this:

```suggestion
  /**
   * First day of week, from 0 to 6. 0 is Sunday, 1 is Monday ... 6 is Saturday
   * @default 0
   */
```

UNRELEASED.md
BPScott: I thought "functionality remains identical as the former types effectivly ended up being aliases of `number` anyway" would cover that?

UNRELEASED.md
BPScott: Added to bug fixes section

UNRELEASED.md
BPScott: I'll expand on this change more when I add it to the migration notes - gonna rebase the migration notes PR once I merge this.

kaelig: My feeling is that this sentence covers the "local" reason for it, whereas mentioning the JS API itself would give some  "global" legitimacy?

PR #3123, opened 2020-07-20
Remove some enums and public types with no value

BPScott:

Shrinking our API surface area

Remove `TypeOf` enum, `GeneralObject`, `DeepPartial`,
`EffectCallback`, `DependencyList` and `Comparator` types from being publicly exported.

These should always have been internal only, but people occasionally forget that everything in `src/types.ts` is publicly exported.

None of these are ever used in web or in our component APIs, and I highly doubt anywhere else is using them as they are never documented, so I'm gonna skip mentioning them in the migration guide.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -22,6 +22,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Updated types of `DatePicker` component - `month`,`year` `weekStartsOn` are now typed as plain `number` - functionality remains identical as the former types effectivly ended up being aliases of `number` anyway ([#3113](https://github.com/Shopify/polaris-react/pull/3113))
 - Removed `Year` type export (used by the DatePicker's props). Replace its usage with `number`. ([#3121](https://github.com/Shopify/polaris-react/pull/3121))
 - Removed the `Month` enum export (used by the DatePicker's props). Replace its usage with a number from 0 to 11, representing the number of the month in question - `Month.January` becomes `0`, `Month.December` becomes `11` etc. ([#3121](https://github.com/Shopify/polaris-react/pull/3121))
+- Removed the `TypeOf` enum, and `GeneralObject`, `DeepPartial`, `EffectCallback`, `DependencyList` and `Comparator` type exports - these were for internal use, and were never documented for external use. ([#3123](https://github.com/Shopify/polaris-react/pull/3123))
 
 ### Enhancements
 
--- a/src/utilities/color-transformers.ts
+++ b/src/utilities/color-transformers.ts
@@ -213,32 +213,25 @@ export function hexToRgb(color: string) {
   return {red, green, blue};
 }
 
-export enum ColorType {
-  Hex = 'hex',
-  Rgb = 'rgb',
-  Rgba = 'rgba',
-  Hsl = 'hsl',
-  Hsla = 'hsla',
-  Default = 'default',
-}
+type ColorType = 'hex' | 'rgb' | 'rgba' | 'hsl' | 'hsla' | 'default';
 
 function getColorType(color: string): ColorType {
   if (color.includes('#')) {
-    return ColorType.Hex;
+    return 'hex';
   } else if (color.includes('rgb')) {
-    return ColorType.Rgb;
+    return 'rgb';
   } else if (color.includes('rgba')) {
-    return ColorType.Rgba;
+    return 'rgba';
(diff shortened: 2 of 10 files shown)


UNRELEASED.md
```diff
@@ -22,6 +22,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Updated types of `DatePicker` component - `month`,`year` `weekStartsOn` are now typed as plain `number` - functionality remains identical as the former types effectivly ended up being aliases of `number` anyway ([#3113](https://github.com/Shopify/polaris-react/pull/3113))
 - Removed `Year` type export (used by the DatePicker's props). Replace its usage with `number`. ([#3121](https://github.com/Shopify/polaris-react/pull/3121))
 - Removed the `Month` enum export (used by the DatePicker's props). Replace its usage with a number from 0 to 11, representing the number of the month in question - `Month.January` becomes `0`, `Month.December` becomes `11` etc. ([#3121](https://github.com/Shopify/polaris-react/pull/3121))
+- Removed the `TypeOf` enum, and `GeneralObject`, `DeepPartial`, `EffectCallback`, `DependencyList` and `Comparator` type exports - these were for internal use, and were never documented for external use.
```
kaelig, line 25: Link to PR?

src/utilities/color-transformers.ts
```diff
@@ -213,32 +213,25 @@ export function hexToRgb(color: string) {
...
 
-export enum ColorType {
-  Hex = 'hex',
-  Rgb = 'rgb',
-  Rgba = 'rgba',
-  Hsl = 'hsl',
-  Hsla = 'hsla',
-  Default = 'default',
-}
+type ColorType = 'hex' | 'rgb' | 'rgba' | 'hsl' | 'hsla' | 'default';
```
kaelig, line 216: Not feeling strongly about this change, but in VS Code, I think enums are a tad easier to trace back ("show all implementations/references").

src/utilities/use-deep-effect.tsx
```diff
@@ -1,9 +1,11 @@
 import {useEffect} from 'react';
 
-import type {EffectCallback, DependencyList, Comparator} from '../types';
-
 import {useDeepCompareRef} from './use-deep-compare-ref';
 
+export type DependencyList = ReadonlyArray<unknown>;
```
kaelig, line 5: Does this need to be exported?

src/utilities/color-transformers.ts
BPScott: Over the past week I've been rallying against enums as I've learned that [how they get expressed at runtime is pretty clunky when compared to using union types](https://www.typescriptlang.org/play/#code/KYOwrgtgBAKsDOAXAouCBlRAnAliA5lAN4CwAUFFAGoCGANmMAIxQC8UA5AAwcA05lWg2AAmNpyZ9yAX3LkAZmBABjRDgD2IKIgSJMuAgAocOiAC5Yu1JH158ASmICo8AO4nlACyjHTj0hSUUMo08MCWSNYY2HYAdEKMTBZYwIhgWFocNBwA3M6UIWERKGi2BPH0jCLJqemZAEa5zrJkLeQ6SGX4hnCRpTHlCcz2eWTkAPTj1OhyZIgAngAO4b2IAKogGiAwS+Hs3BwAPhIcs4oqapraul0ivsDmxRtbO8v+zm4e3vcQ74EFoXCBxqaQynGyoyCwUBJxBdU4jUhUBabTmNwG+DuBxG5CAA), and in addition to that they're not treeshakable as defining an enum doesn't emit code that is unambiguously pure - because [enums may be extended later](https://www.typescriptlang.org/play/?ssl=10&ssc=40&pln=1&pc=1#code/KYOwrgtgBAkgLsCBnKBvAsAKClAagQwBsxgBGKAXigHIAGagGixwOOACZKbTGsBfLFlCRYCZGijM8REgGYu1dtX6DMAYwD2IJBsLAAdIQ0BzABTxESfaxKkGoy9ZnBZASiA)(I'm filing this under "thanks, I hate it").

In some magical ideal world it's be great if enums could compile down to simple objects, or `const enum` was usable in a single file in isolatedModules mode, but that's not how things work currently unfortunately.

UNRELEASED.md
BPScott: Rebased and added

src/utilities/use-deep-effect.tsx
BPScott: fixed.

PR #3133, opened 2020-07-22
Updates to react imports to help tree-shaking

BPScott:

I found that the existing shipped version of polaris isn't very tree-shakable.

You can see this in action using using the delightfully named [agadoo](https://github.com/Rich-Harris/agadoo) library which makes a sample consuming app that imports your package but doesn't use any of its imports (i.e. it just does `import @shopify/polaris-react`). In an ideal world the resulting output would be empty as all your library gets tree-shaken away - no superflous imports of things you're not using bulking up your bundle. 

Unfortunately that's not what happens. Running the following on master to create a comparison between out library code and the shaken output shows that only about 1/3rd of our total library size gets tree-shaken (dist/index.es.js is ~16000 lines, test.txt is ~10000 lines):

```
rm -f test.txt && yarn build && npx agadoo dist/index.es.js > test.txt; code --diff dist/index.es.js test.txt
```

I've worked out the reason that lots of our code sticks around: Rollup considers our usage of `React.PureComponent` and `React.Component as potentially side-effecty as it's reading a property of an object (and that might be a getter and thus trigger side effects) - see the ["treeshake.propertyReadSideEffects" section in the rollup docs](https://rollupjs.org/guide/en/#danger-zone) for an example. [Here is a minimal reproduction demonstrating this](https://rollupjs.org/repl/?version=2.22.2&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmltcG9ydCUyMFJlYWN0JTJDJTIwJTdCQ29tcG9uZW50JTdEJTIwZnJvbSUyMCdyZWFjdCclM0IlNUNuaW1wb3J0JTIwKiUyMGFzJTIwUmVhY3ROcyUyMGZyb20lMjAncmVhY3QnJTNCJTVDbiU1Q24lMkYlMkYlMjBSZWZlcmVuY2luZyUyMGElMjBwcm9wZXJ0eSUyMG9uJTIwdGhlJTIwZGVmYXVsdCUyMGV4cG9ydCUyMGlzJTIwbm90JTIwdHJlZS1zaGFrZW4lNUNuY2xhc3MlMjBEZWZhdWx0RG90Q29tcG9uZW50JTIwZXh0ZW5kcyUyMFJlYWN0LkNvbXBvbmVudCUyMCU3QiU3RCUzQiU1Q24lNUNuJTJGJTJGJTIwUmVmZXJlbmNpbmclMjBhJTIwbmFtZWQlMjBleHBvcnQlMjBpcyUyMHRyZWUlMjBzaGFrZW4lNUNuY2xhc3MlMjBOYW1lZEV4cG9ydENvbXBvbmVudCUyMGV4dGVuZHMlMjBDb21wb25lbnQlMjAlN0IlN0QlM0IlNUNuJTVDbiUyRiUyRiUyMFJlZmVyZW5jaW5nJTIwcHJvcGVydHklMjBvbiUyMGElMjBuYW1lc3BhY2UlMjBleHBvcnQlMjBpcyUyMHRyZWUtc2hha2VuJTVDbmNsYXNzJTIwTnNEb3RDb21wb25lbnQlMjBleHRlbmRzJTIwUmVhY3ROcy5Db21wb25lbnQlMjAlN0IlN0QlM0IlNUNuJTVDbiUyRiUyRiUyMEluJTIwUG9sYXJpcyUyMHdlJTIwY3VycmVudGx5JTIwaW1wb3J0JTIwQ29tcG9uZW50JTJGUHVyZUNvbXBvbmVudCUyMGluJTIwdGhlJTIwJTYwRGVmYXVsdERvdENvbXBvbmVudCU2MCUyMHN0eWxlLiUyMFRoYXQlMjBicmVha3MlMjB0cmVlLXNoYWtpbmclMjBpbiUyMGNvbnN1bWluZyUyMGFwcHMlMjBiZWNhdXNlJTIwcHJvcGVydHklMjBhY2Nlc3MlMjBtYXklMjBoYXZlJTIwc2lkZS1lZmZlY3RzJTIwLWFzJTIwdGhlJTIwcHJvcGVydHklMjB5b3UncmUlMjBhY2Nlc3NpbmclMjBtYXklMjBiZSUyMGElMjBnZXR0ZXIuJTVDbiUyRiUyRiUyMFdlJTIwbmVlZCUyMHRvJTIwY2hhbmdlJTIwaG93JTIwd2UlMjBhY2Nlc3MlMjB0aGlzJTIwLSUyMFVzaW5nJTIwdGhlJTIwTmFtZWRFeHBvcnRDb21wb25lbnQlMjBhcHByb2FjaCUyMGlzJTIwYmVzdCUyMElNTyU1Q24lNUNuJTVDbiUyMiUyQyUyMmlzRW50cnklMjIlM0F0cnVlJTdEJTVEJTJDJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmVzJTIyJTJDJTIybmFtZSUyMiUzQSUyMm15QnVuZGxlJTIyJTJDJTIyYW1kJTIyJTNBJTdCJTIyaWQlMjIlM0ElMjIlMjIlN0QlMkMlMjJnbG9iYWxzJTIyJTNBJTdCJTdEJTdEJTJDJTIyZXhhbXBsZSUyMiUzQW51bGwlN0Q=)

The good news is that's not too bad to fix - replace usage of properties on the React object (e.g. React.Component)  with named exports:

```diff
- import React from 'react';
+ import React, {Component, PureComponent} from 'react';

-export class SomeComponent extends React.Component {}
+ export class SomeComponent extends Component {}

-export class SomePureComponent extends React.PureComponent {}
+ export class SomePureComponent extends PureComponent {}
```

Somewhat annoyingly this doesn't seem to have much impact when used within webpack but it feels like a step in the right direction, and prepares us for using [new jsx transforms](https://babeljs.io/blog/2020/03/16/7.9.0#a-new-jsx-transform-11154-https-githubcom-babel-babel-pull-11154) once they're ready for usage.

- Replaces usage of `<React.Fragment>` with the shorthand syntax `<>` - As we control our build then we don't have to worry about consumers doing odd stuff with this.
- Replace usage of `React.Component` and `React.PureComponent` with named exports `Component` and `PureComponent`
- Replace usage of several other React.SOMETHING items with named exports (these have less of an impact than the above but it feels nicer for consistency)
- Tweak how we define the `Tokens` variable in our utilities creating an object from others is potentially sideeffecty (for similar "that property access might be a getter reasons), which allows it and all the functions used to construct it to be tree shakable
- Update to latest babel and rollup versions to get slightly better shaking of React `createContext` usage

These changes have been split into individual commits to make them easier to review.

After all this we're not quite ideal, but running the same test as above, our post-treeshaking attempted output has gone from ~10000 lines to ~600 lines. Most of what remains is either typescript enums which aren't tree-shakable, and some style definitions that have lead me to raise an issue with Rollup.

## 

### How to 🎩

- On master `yarn clean && yarn build && mv dist dist-orig`
- On this branch: `yarn clean && yarn build`
- Compare the output of master and this branch `diff dist-orig/index.es.js dist/index.es.js` and see that the changes are just some extra PURE annotation comments and using named imports instead of React.SOMETHING (e.g. `PureComponent` instead of `React.PureComponent`)
- Check the tree-shaken output of a new build: `rm -f test.txt && yarn build && npx agadoo dist/index.es.js > test.txt; code test.txt` and note that is now tiny compared to the result on master.


BPScott:
> Can we link to the rollup issue for context? 

Once I write it up sure :p ([here's the minimal reproduction](https://rollupjs.org/repl/?version=2.22.2&shareable=JTdCJTIybW9kdWxlcyUyMiUzQSU1QiU3QiUyMm5hbWUlMjIlM0ElMjJtYWluLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMiUyRiUyRiUyMEFzJTIwaXQlMjBpcyUyMGElMjBzcHJlYWQlMjBhcmd1bWVudCUyMGNsYXNzZXMlMjBpcyUyMGFsd2F5cyUyMGFuJTIwYXJyYXkuJTVDbiUyRiUyRiUyMEklMjB3b3VsZCUyMGV4cGVjdCUyMHNsaWRlQ2xhc3NlcyUyMHRvJTIwYmUlMjB0cmVlLXNoYWtlbiUyMGF3YXklNUNuJTVDbmZ1bmN0aW9uJTIwY2xhc3NOYW1lcyguLi5jbGFzc2VzKSUyMCU3QiU1Q24lMjAlMjByZXR1cm4lMjBjbGFzc2VzLmZpbHRlcihCb29sZWFuKS5qb2luKCclMjAnKSUzQiU1Q24lN0QlNUNuJTVDbmNvbnN0JTIwZXhhbXBsZSUyMCUzRCUyMGNsYXNzTmFtZXMoJ2ZpcnN0JyUyQyUyMCdzZWNvbmQnKSUzQiU1Q24lNUNuJTVDbiUyRiUyRiUyMENvdW50ZXIlMjBFeGFtcGxlJTIwLSUyMElmJTIwYXJyYXklMjBpcyUyMGRlZmluZWQlMjBpbmxpbmUlMjB0aGVuJTIwaXQlMjBpcyUyMHNoYWtlbiUyMGF3YXklNUNuJTVDbmZ1bmN0aW9uJTIwY2xhc3NOYW1lczIoKSUyMCU3QiU1Q24lMjAlMjBjb25zdCUyMGNsYXNzZXMlMjAlM0QlMjAlNUInZmlyc3QnJTJDJTIwJ3NlY29uZCclNUQlM0IlNUNuJTVDdHJldHVybiUyMGNsYXNzZXMuZmlsdGVyKEJvb2xlYW4pLmpvaW4oJyUyMCcpJTNCJTVDbiU3RCU1Q24lNUNuY29uc3QlMjBleGFtcGxlMiUyMCUzRCUyMGNsYXNzTmFtZXMyKCklM0IlNUNuJTVDbiU1Q24lMkYlMkYlMjBDb3VudGVyJTIwRXhhbXBsZSUyMC0lMjBBJTIwc2lsbHklMjB3cmFwcGVyJTIwZnVuY3Rpb24lMjB0aGF0JTIwbWFrZXMlMjBSb2xsdXAlMjByZWFsaXNlJTIwY2xhc3NlcyUyMGlzJTIwYWx3YXlzJTIwYW4lMjBhcnJheSU1Q24lNUNuJTJGJTJGJTIwRnJvbSUyMGElMjBydW50aW1lJTIwcGVyc3BlY3RpdmUlMjB0aGlzJTIwaXMlMjBzaWxseSUyMGJlY2F1c2UlMjAlNjBhcmdzJTYwJTIwaXMlMjBhJTIwc3ByZWFkJTIwc28lMjBpdCUyMGlzJTIwYWxyZWFkeSUyMGFuJTIwYXJyYXklNUNuZnVuY3Rpb24lMjBhcnIoLi4uYXJncyklMjAlN0IlMjByZXR1cm4lMjAlNUIuLi5hcmdzJTVEJTIwJTdEJTNCJTVDbmZ1bmN0aW9uJTIwY2xhc3NOYW1lczMoLi4uY2xhc3NlcyklMjAlN0IlNUNuJTVDdHJldHVybiUyMGFycihjbGFzc2VzKS5maWx0ZXIoQm9vbGVhbikuam9pbignJTIwJyklM0IlNUNuJTdEJTVDbiU1Q25jb25zdCUyMGV4YW1wbGUzJTIwJTNEJTIwY2xhc3NOYW1lczMoJ2ZpcnN0JyUyQyUyMCdzZWNvbmQnKSUzQiU1Q24lNUNuJTVDbiU1Q24lMjIlMkMlMjJpc0VudHJ5JTIyJTNBdHJ1ZSU3RCU1RCUyQyUyMm9wdGlvbnMlMjIlM0ElN0IlMjJmb3JtYXQlMjIlM0ElMjJlcyUyMiUyQyUyMm5hbWUlMjIlM0ElMjJteUJ1bmRsZSUyMiUyQyUyMmFtZCUyMiUzQSU3QiUyMmlkJTIyJTNBJTIyJTIyJTdEJTJDJTIyZ2xvYmFscyUyMiUzQSU3QiU3RCU3RCUyQyUyMmV4YW1wbGUlMjIlM0FudWxsJTdE_)

> is there a way we can make sure we don't accidentally do React.* in the future

There's nothing ESLinty as far as I can tell, but I can add a build test that will shout if you do `extends React.Component` / `extends React.PureComponent` as that's where most of the issues arise from.

src/utilities/theme/tokens.ts
```diff
@@ -49,12 +48,6 @@ const Overrides = {
   badgeFontWeight: '500',
 };
 
-export const Tokens = {
-  ...BorderRadius,
-  ...Shadow,
-  ...Overrides,
-};
```
amrocha, line 56: Why is this a problem?

UNRELEASED.md
```diff
@@ -24,4 +24,6 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Code quality
 
+- Updated how we use React.Component and React.PureComponent to help treeshakability ([#3133](https://github.com/Shopify/polaris-react/pull/3133))
```
kaelig, line 27: There's a bit more that's covered in this PR, I think this item would benefit from summarizing the other changes as well.

BPScott:
>> prepares us for using new jsx transforms once they're ready for usage.

> How does this PR do this? We still end up importing something from React so it's basically the same afaik

Currently you require a `React` import in your code for jsx to work, this is because the existing Babel jsx transform stuff provided by `@babel/preset-react` doesn't try and do anything clever. So in e.g. files with functional components and hooks we need to do `import React, {useState} from 'react';`. With the new transfoms it is smart and injects that import when needed so we could just write `import {useState} from 'react';`.

So when we come to move to the new transform we can get rid of all the case where we reference `React` just for the jsx. Currently on master there's still loads of cases where we references React.SOMETHING that we'd have to modify to a named import so we can remove the default export. This PR does a bunch that "stop doing React.THING, use a named import" work so when the time comes to leverage the new transform we'll have less tidying work to do

src/utilities/theme/tokens.ts
BPScott: `const Tokens = {...BorderRadius, ...Shadow, ...Overrides}` is also considered potentially sideeffecty by rollup - because the spread results in calls to properties, which may be a getter and thus trigger a side effect. Which means that these variables and everything required to build them - lots of our color manipulation utilities don't get tree-shaken.

BPScott:
Build test added, and the final runtime usages of React.THING are removed - with the exception of React.createElement and React.Fragment, which are produced by the JSX transform.

BPScott:
And an issue submitted upstream: https://github.com/rollup/rollup/issues/3690

scripts/build-validate.js
```diff
@@ -20,6 +20,16 @@ function validateStandardBuild() {
   assert.ok(fs.existsSync('./dist/index.es.js'));
   assert.ok(fs.existsSync('./dist/styles.css'));
 
+  // Assert it uses named exports rather than properties from the React default
+  // export to help tree-shaking.
+  // React.createElement and React.Fragment are the allowed exceptions
```
kaelig, line 25: Curious about why we'd allow exceptions, can you say more?

src/components/ResourceList/components/BulkActions/BulkActions.tsx
```diff
@@ -67,11 +67,11 @@ interface State {
...
-  appear: classNames(styles.Slide, styles['Slide-appear']),
-  appearActive: classNames(styles.Slide, styles['Slide-appearing']),
-  enter: classNames(styles.Slide, styles['Slide-enter']),
-  enterActive: classNames(styles.Slide, styles['Slide-entering']),
-  exit: classNames(styles.Slide, styles['Slide-exit']),
+  appear: `${styles.Slide} ${styles['Slide-appear']}`,
+  appearActive: `${styles.Slide} ${styles['Slide-appearing']}`,
+  enter: `${styles.Slide} ${styles['Slide-enter']}`,
+  enterActive: `${styles.Slide} ${styles['Slide-entering']}`,
+  exit: `${styles.Slide} ${styles['Slide-exit']}`,
```
kaelig, line 74: Is using `classNames` impacting the tree-shakability?

scripts/build-validate.js
```diff
@@ -20,6 +20,16 @@ function validateStandardBuild() {
...
+  const esModuleContent = fs.readFileSync('./dist/index.es.js', 'utf-8');
```
kaelig, line 26: The error message is very cryptic when validating against this file (see below), and doesn't contain any actionable feedback.

IMHO the best developer experience would be an ESLint rule with autofix, but I don't know if it's possible.

```
assert.js:85
  throw new AssertionError(obj);
  ^

AssertionError [ERR_ASSERTION]: Input A expected to strictly deep-equal input B:
+ expected - actual

- [
-   'React.PureComponent'
- ]
+ []
    at validateStandardBuild (/Users/kaelig/src/github.com/Shopify/polaris-react/scripts/build-validate.js:31:10)
    at Object.<anonymous> (/Users/kaelig/src/github.com/Shopify/polaris-react/scripts/build-validate.js:11:1)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:742:12)
    at startup (internal/bootstrap/node.js:283:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:743:3)
child_process.js:656
    throw err;
    ^

Error: Command failed: node scripts/build-validate.js
    at checkExecSyncError (child_process.js:616:11)
    at execSync (child_process.js:653:13)
    at run (/Users/kaelig/src/github.com/Shopify/polaris-react/scripts/build.js:5:22)
    at Object.<anonymous> (/Users/kaelig/src/github.com/Shopify/polaris-react/scripts/build.js:27:1)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:742:12)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

src/components/ResourceList/components/BulkActions/BulkActions.tsx
BPScott: Whoops, that was me testing something. Didn't want to commit this.  I've reset this back to what it was originally.

But yes, usage of `classNames` isn't treeshakable because [rollup isn't treating the spread parameters as an array](https://github.com/rollup/rollup/issues/3690).

This isn't a big deal as it only affects cases that aren't inside a called function which is pretty minimal and hopefully rollup can be patched so it will just work.

scripts/build-validate.js
BPScott: This tests against the compiled file. This means that the compiled file may include them and that folks may end up writing them manually (though 99% of the time won't need to as you should just use JSX). We need these exceptions because these are what the JSX transform uses. When we write:

```js
const x = <div>Hi <>you</></div>
```

it gets compiled to:

```
const x = /*#__PURE__*/React.createElement("div", null, "Hi ", /*#__PURE__*/React.createElement(React.Fragment, null, "you"));
```

--- 

These asserts accept a final arg where you can write a description. I'll flesh that out

scripts/build-validate.js
```diff
@@ -20,6 +20,16 @@ function validateStandardBuild() {
...
+  const unwantedReactUsageMatches =
+    esModuleContent.match(/React\.(?!createElement|Fragment)[A-Za-z0-9]+/g) ||
+    [];
+
+  assert.deepStrictEqual(unwantedReactUsageMatches, []);
```
alex-page, line 31: It looks like you can add a message as the third argument. This might help with what @kaelig is suggesting above.
```suggestion
  assert.deepStrictEqual(unwantedReactUsageMatches, [], `Please use named exports from React. Found these failures:\n${unwantedReactUsageMatches}`);
```

BPScott:
Gonna merge this as-is, and address "messages when build tests fail are bad" in a follow-up

PR #3136, opened 2020-07-23
Ignore excluded files when running build-consumer

BPScott:

`build-consumer` currently doesn't work because it doesn't handle excluded files in package.json's files array.

Ignores excluded files when running build-consumer
 
This means these excluded files will be copied across when running
build-consumer but this won't have any negative impact other than being
not optimally efficient.

### How to 🎩

Run build-consumer and see it completes (vs throwing an error on master). Assuming you've got the styleguide checked out: `yarn run build-consumer polaris-styleguide`

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -14,6 +14,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Development workflow
 
+- Fix `build-consumer` script to handle excludes in package.json's files array ([#3136](https://github.com/Shopify/polaris-react/pull/3136))
+
 ### Dependency upgrades
 
 ### Code quality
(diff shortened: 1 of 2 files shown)


UNRELEASED.md
```diff
@@ -14,6 +14,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Development workflow
 
+- Fix `build-consumer` script to handle excludes in package.json's files array ([#3136](https://github.com/Shopify/polaris-react/pull/3136))
```
kaelig, line 17:
```suggestion
- Fixed `build-consumer` script to handle excludes in package.json's `files` array ([#3136](https://github.com/Shopify/polaris-react/pull/3136))
```

BPScott: Past tense always gets me!

PR #3138, opened 2020-07-28
DatePicker improvements

### LINKED ISSUE #2374: [DatePicker] Add more examples
DatePicker only has one really simple example. This is a problem because our automated tests rely on examples, and because consumers rely on them to learn how to use components.

By adding more examples we'll make our codebase better, and help consumers use Polaris better.

BPScott:

Polish, and adding extra examples for a ranged, multi-month and disabled dates functionality in DatePicker

Fixes #2374

- Replace usage of 'isSameDate' with 'isSameDay' as they are identical
- Remove isSameDay helper
- Add extra examples for ranged DatePicker

### How to 🎩

Check out the new date picker examples

### DIFF
--- a/src/utilities/dates.ts
+++ b/src/utilities/dates.ts
@@ -134,14 +134,6 @@ export function isDateBefore(date: Date, dateToCompare: Date) {
   return date.getTime() < dateToCompare.getTime();
 }
 
-export function isSameDate(source: Date, target: Date) {
-  return (
-    source.getFullYear() === target.getFullYear() &&
-    source.getMonth() === target.getMonth() &&
-    source.getDate() === target.getDate()
-  );
-}
-
 const WEEKDAYS: number[] = [0, 1, 2, 3, 4, 5, 6];
 
 export function getOrderedWeekdays(weekStartsOn: number): number[] {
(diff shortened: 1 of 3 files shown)


src/utilities/dates.ts
```diff
@@ -134,14 +134,6 @@ export function isDateBefore(date: Date, dateToCompare: Date) {
   return date.getTime() < dateToCompare.getTime();
 }
 
-export function isSameDate(source: Date, target: Date) {
```
BPScott, line 137: Compare with isSameDay:

https://github.com/Shopify/polaris-react/blob/88868bc44ba464e30827b4a4b49b0f5d89f83f38/src/utilities/dates.ts#L65-L71

I believe this snuck in while I was killing of javascript-utilities - I coped this function from `@shopify/javascript-utilities` without spotting that we already had an equivalent function already in this file

alex-page:
This might close https://github.com/Shopify/polaris-react/issues/2374. What do you think @BPScott?

BPScott:
> This might close #2374. What do you think @BPScott?

Just adding a ranged example would be a lackluster fix for that.
I've added some extra examples for multi-month,and disabled dates - which the cover the props without examples.

PR #3153, opened 2020-08-04
[ResourceList] Add conditional rendering of select all labels

carolopolo:
- Currently, the `Select all resources...` button label only ever says, `Select all {itemsLength}+ {resourceNamePlural} in your store`, regardless of the situation:
before

- The `ResourceList` should be able to know when it has any filters applied to it, so the button label can be more informative & accurate. With this PR, if the list is filtered, then the label reads `Select all {itemsLength}+ {resourceNamePlural} in this filter`:
after

- Fixes https://github.com/Shopify/store/issues/16809, when combined with https://github.com/Shopify/web/pull/29878

- This PR adds an optional `isFiltered` prop to the `ResourceList`, which conditionally renders the select all resources button label. 

## 

### How to 🎩

- Paste & run the Playground below and verify that when you check the select all checkbox, you see the button label `Select all 2+ customers in this filter`

<details>

```jsx

import {Card, ResourceList, ResourceItem, TextStyle, Avatar} from '../src';

  const [selectedItems, setSelectedItems] = useState([]);

  const resourceName = {
    singular: 'customer',
    plural: 'customers',
  };

  const items = [
    {
      id: 341,
      url: 'customers/341',
      name: 'Mae Jemison',
      location: 'Decatur, USA',
    },
    {
      id: 256,
      url: 'customers/256',
      name: 'Ellen Ochoa',
      location: 'Los Angeles, USA',
    },
  ];

  const promotedBulkActions = [
    {
      content: 'Edit customers',
    },
  ];

  const bulkActions = [
    {
      content: 'Add tags',
    },
    {
      content: 'Remove tags',
    },
    {
      content: 'Delete customers',
    },
  ];

  return (
    <Card>
      <ResourceList
        resourceName={resourceName}
        items={items}
        renderItem={renderItem}
        selectedItems={selectedItems}
        onSelectionChange={setSelectedItems}
        bulkActions={bulkActions}
        selectable
        isFiltered
        hasMoreItems
      />
    </Card>
  );

  function renderItem(item) {

    return (
      <ResourceItem
        id={id}
        url={url}
        media={media}
      >
        <h3>
        </h3>
        <div>{location}</div>
      </ResourceItem>
    );
  }
}
```

</details>

### 🎩 checklist


UNRELEASED.md
```diff
@@ -13,6 +13,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- Added optional `isFiltered` prop to `ResourceList` to conditionally render more informative select all button label ([#3153](https://github.com/Shopify/polaris-react/pull/3153))
```
posth, line 16: Could move this to the `Enhancements` section since it's an addition

locales/en.json
```diff
@@ -152,7 +152,9 @@
       "loading": "Loading {resource}",
       "selected": "{selectedItemsCount} selected",
       "allItemsSelected": "All {itemsLength}+ {resourceNamePlural} in your store are selected.",
+      "allFilteredItemsSelected": "All {itemsLength}+ {resourceNamePlural} in this filter are selected.",
```
posth, line 155: :bike: 🏠 `...on this filter in your store are selected` ?

If only to make sure merchant's are aware it's all customers in their store

src/components/ResourceList/tests/ResourceList.test.tsx
```diff
@@ -1197,6 +1198,90 @@ describe('<ResourceList />', () => {
...
+      );
+
+      resourceList.find(BulkActions)!.find(Button)!.trigger('onClick');
+
+      expect(
+        resourceList.find(BulkActions)!.prop('paginatedSelectAllText'),
+      ).toBe('All 2+ customers in this filter are selected.');
+    });
+
+    it('renders `selectAllItems` label if false', () => {
```
posth, line 1244: :bike: 🐑 For these tests, we could just not provide it (not explicitly passing in `false` as a prop), and test that default behavior, ie: `renders selectAllItems label if isFiltered prop is not provided`

locales/en.json
carolopolo: i think what we have is fine personally, but i don't have a super strong opinion. this label only comes up when somebody clicks directly the button that says `Select all XY+ resources in this filter`, so i feel that they know what they've selected at this point 😁

UNRELEASED.md
```diff
@@ -10,6 +10,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - Added optional `videoProgress` and `showVideoProgress` props to `VideoThumbnail` for video progress indicator ([#3057](https://github.com/Shopify/polaris-react/pull/3057))
 - Enabled much easier tree-shaking in consuming apps by having a multi-file build instead of a single-file build ([#3137](https://github.com/Shopify/polaris-react/pull/3137))
+- Added optional `isFiltered` prop to `ResourceList` to conditionally render more informative select all button label ([#3153](https://github.com/Shopify/polaris-react/pull/3153))
```
loic-d, line 13: Looking at the custom fork we have in Products (`ResourceListWithHeader`). It looks like it's already supporting this feature with the `filtersApplied` prop. I feel we should match the prop's name to make it consistent. 
What do you think?

locales/en.json
loic-d: I would keep what you have which matches the custom fork `ResourceListWithHeader` used in Products.
`All {itemsLength}+ {resourceNamePlural} in this filter are selected.`

src/components/ResourceList/tests/ResourceList.test.tsx
```diff
@@ -1197,6 +1198,88 @@ describe('<ResourceList />', () => {
...
+
+      expect(
+        resourceList.find(BulkActions).prop('paginatedSelectAllAction'),
+      ).toStrictEqual({
+        content: 'Select all 2+ customers in your store',
+        onAction: expect.any(Function),
+      });
+    });
+
+    it('renders `allItemsSelected` label if not passed and all items are selected', () => {
```
loic-d, line 1263: Didn't we have a test for this use case already?

UNRELEASED.md
carolopolo: i went with this name because `filtersApplied` doesn't seem like a boolean to me, but rather, an array of applied filters. but i'm not strongly opinionated, we can change it if you believe it's more accurate!

loic-d: Let's let Polaris folks decide based on the conventions they try to enforce in this repo.
cc/ @alex-page

alex-page: Hey @loic-d I would agree with @carolopolo. From my understanding we use `isX` a few times in Polaris.
https://github.com/Shopify/polaris-react/blob/master/src/components/Avatar/Avatar.tsx#L49

PR #3167, opened 2020-08-11
[New DL] Change nav font size and width. Decrease vertical space in page flow

kyledurand:

Decreases font-size from 1.6rem to 1.4rem for the new design language

As per the updated Figma doc: https://www.figma.com/file/2lVS321btZUNFYkSK5X3y5/Design-language-QA?node-id=2%3A401

Before:
localhost_6006_iframe html_id=playground-playground--details-page contexts=New%20Design%20Language=Enabled%20-%20Light%20Mode (3)

After:
localhost_6006_iframe html_id=playground-playground--details-page contexts=New%20Design%20Language=Enabled%20-%20Light%20Mode (4)

Revert navigation width back to original:
localhost_6006_iframe html_id=playground-playground--details-page contexts=New%20Design%20Language=Enabled%20-%20Light%20Mode

## 

### How to 🎩

Check out the details page, as well as the frame examples including the frame with offset example.
Test the contextual save banner on the details page by typing in the first text field

### 🎩 checklist


playground/DetailsPage.tsx
```diff
@@ -41,6 +32,17 @@ import {
   Thumbnail,
 } from '../src';
 
+import {
```
BPScott, line 35: Is this because of icon magic imports failing?

You probably want to update these lists:

- For storybook: https://github.com/Shopify/polaris-react/blob/master/.storybook/polaris-readme-loader.js#L193
- For the styleguide examples: https://github.com/Shopify/polaris-styleguide/blob/master/client/demo-app/Demo/Demo.tsx#L11 (not strictly needed as you're not editing any README files that would then be consumed by the styleguide, but it should be kept in sync with the code in polaris-readme-loader.js

playground/DetailsPage.tsx
```diff
@@ -41,6 +32,17 @@ import {
...
+  HomeMajorMonotone,
```
BPScott, line 36: How come this is needed? New iconography?

kyledurand: Yep we had to update them to make tophatting more inclusive for design. Here's a screenshot
image

playground/images.tsx
```diff
@@ -0,0 +1,192 @@
+import React from 'react';
+
+export function HomeMajorMonotone() {
```
BPScott, line 3: Nitpick: @kaelig probably has some thoughts on how some of these can be simplified - you probably don't need the`g`/`defs` stuff, or width/height delcarations on the svg files

src/components/Card/Card.scss
```diff
@@ -12,6 +12,13 @@
   @include page-content-when-not-fully-condensed {
     border-radius: var(--p-border-radius-wide, border-radius());
   }
+
+  &.newDesignLanguage {
```
BPScott, line 16: Nitpick: Thoughts on moving this up  to be closer to the margin-top override above?

Not tested this but something like:

```
.Card {
  + .Card {
    margin-top: spacing(loose);

    &.newDesignLanguage {
      margin-top: spacing()
    }
  }
}
```

src/components/Layout/Layout.scss
```diff
@@ -50,10 +54,18 @@ $relative-size: $primary-basis / $secondary-basis;
   margin-top: spacing(loose);
   margin-left: spacing(loose);
 
+  .newDesignLanguage & {
+    margin-top: spacing();
```
BPScott, line 58: Seeing a bunch of this spacing stuff is making me think we should have css-tokens for spacing (which IMO we wanted to do anyway), which would mitigate a needing these extra style declarations, but that's a problem for another day :)

src/styles/foundation/_layout.scss
```diff
@@ -60,7 +56,7 @@ $layout-width-data: (
 $dismiss-icon-size: 32px;
 
 @function top-bar-height() {
-  @return rem(56px);
+  @return var(--p-top-bar-height, rem(56px));
```
BPScott, line 59: Currently it's possible to do stuff like `some-property: top-bar-height() - 10px;` and it'll work because it returns a plain number. 

This change will break that behaviour, which feels pretty scary. It doesn't look like we use it in that way internally but external folk might.

If this is super useful and simplifies a bunch of stuff (looks like it'd help further simplify some of the bits I touched in https://github.com/Shopify/polaris-react/pull/3158/files) then it should be a new function. That new function should live in a new file, and be imported in _common, but not _public-api so we don't end up exposing it publicly (and thus would be free to make this sort of change in the future because we don't need to care about external consumers).

kyledurand: Excellent point. I don't think this is necessary for this PR so I can revert

PR #3168, opened 2020-08-11
Storybook - Add ENV variables to enable the new design language

kyledurand:

Since we're mostly working in the new design language I think it makes sense to default to it to avoid having to click twice every time the page refreshes after a change. Maybe its too early to make this change since we're public facing. Would like to hear some thoughts.

Consideration: I'm fairly sure this wouldn't affect the styleguide components but not 100%. @kaelig would know for sure

image

## 

### How to 🎩

Open a story and toggle the different modes

### 🎩 checklist

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -55,6 +55,8 @@ function AppProviderWithKnobs(
   );
 }
 
+const preferredTheme = process.env.STORYBOOK_NEWDESIGNLANGUAGE || 'purpler';
+
 const withContextsDecorator = withContexts([
   {
     title: 'Strict Mode',
@@ -70,15 +72,17 @@ const withContextsDecorator = withContexts([
     params: [
       {
         name: 'Disabled',
-        default: true,
+        default: preferredTheme === 'purpler',
         props: {newDesignLanguage: false},
       },
       {
         name: 'Enabled - Light Mode',
+        default: preferredTheme === 'light',
         props: {newDesignLanguage: true, colorScheme: 'light'},
       },
       {
         name: 'Enabled - Dark Mode',
+        default: preferredTheme === 'dark',
         props: {newDesignLanguage: true, colorScheme: 'dark'},
       },
     ],
--- a/README.md
+++ b/README.md
@@ -100,6 +100,26 @@ We have created example applications to document some of the ways you could incl
 
 We use Storybook to create a simple, hot-reloading playground for development on these components. You can edit the `playground/Playground.tsx` file to import the components you are working on, and run `yarn dev` in order to start the development server. Please do not commit your work on the playground so that it remains pristine for other developers to work on.
 
+### Testing the new design language
+
+Old design language (default):
+
+```bash
+yarn dev
+```
+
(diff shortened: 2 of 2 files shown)


kaelig:
Perhaps an environment variable would be preferable as it would help a lot with local development? As in: default to the new DL in light mode if `NEW_DESIGN_LANGUAGE=light`?

BPScott:
Exciting but scary! You're right this would have no effect on the styleguide.

We should also update [the a11y tests](https://github.com/Shopify/polaris-react/blob/master/scripts/pa11y.js#L68-L71) as they assume that the default is the old style DL. As this PR is currently it'll result in the pa11y tests running with the new DL twice.

Gut feel is we should hold off on this for the moment, it doesn't feel like we're at the tipping point of "new DL is our new default when thinking about stuff" juuuust yet. We're close though, perhaps another 2 or 3 weeks (counterpoint: come at me with arguments of what -if any - big changes still need to happen and predicted timelines). Till we hit that tipping point it feels like we should keep with the current expected merchant per Alex's comment.


> default to the new DL in light mode if NEW_DESIGN_LANGUAGE=light?

The logic for settings these defaults are done in preview.js which is ran in the browser. I sorting out the plumbing to pass a CLI env variable through to there might be more hassle than it's worth.

kaelig:
@bpscott TIL Storybook has a super [easy way to configure environment variables](https://storybook.js.org/docs/react/configure/environment-variables). All it takes is:

```
const theme = process.env.STORYBOOK_NEWDESIGNLANGUAGE || 'purpler';

addDecorator(
  withContexts([
    {
      title: 'Strict Mode',
      components: [StrictModeToggle],
      params: [
        {name: 'Disabled', props: {isStrict: false}},
        {name: 'Enabled', default: true, props: {isStrict: true}},
      ],
    },
    {
      title: 'New Design Language',
      components: [AppProviderWithKnobs],
      params: [
        {
          name: 'Disabled',
          default: theme === 'purpler',
          props: {newDesignLanguage: false},
        },
        {
          name: 'Enabled - Light Mode',
          default: theme === 'light',
          props: {newDesignLanguage: true, colorScheme: 'light'},
        },
        {
          name: 'Enabled - Dark Mode',
          default: theme === 'dark',
          props: {newDesignLanguage: true, colorScheme: 'dark'},
        },
      ],
    },
  ]),
);
```

Then, run:

```
STORYBOOK_NEWDESIGNLANGUAGE=dark yarn dev
```

kyledurand:
Hey @kaelig that looks awesome and much better than changing our public facing storybook too early. I'll give that a go when I have some free time

kaelig:
Hey @kyledurand I just rebased and pushed the changes I suggested – give it a try and let me know how this looks!
[`0fb0a29` (#3168)](https://github.com/Shopify/polaris-react/pull/3168/commits/0fb0a293f4b637410751caafe655a5d991ab9dc0)

kyledurand (commented):
Heck yes! Thanks @kaelig !

alex-page (approved):

PR #3169, opened 2020-08-11
Use esnext as custom mainfield for esnext build

BPScott:

sewing-kit v0 now uses `esnext` as the custom mainField, to match the
behaviour of sewing-kit v1. Update to match that.

This is not a breaking change as sewing-kit v0 still has an alias that forces
the use of the esnext folder, so this mainField doesn't do anything
currently, it's just for future compatibility, once SK v0 drops support
for Polaris <5.0 - which will now be "dropping support for < 5.2".

See https://github.com/Shopify/sewing-kit/pull/2279

Renames field in package.json

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -13,6 +13,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Labelled component now breaks on long lines of text, regardless of presence of naturally breaking characters (hyphens, whitespace, etc.) ([#3156](https://github.com/Shopify/polaris-react/pull/3156))
 - Added optional `isFiltered` prop to `ResourceList` to conditionally render more informative select all button label ([#3153](https://github.com/Shopify/polaris-react/pull/3153))
 - Exported `PositionedOverlay` component for use in consuming applications ([#3161](https://github.com/Shopify/polaris-react/pull/3161))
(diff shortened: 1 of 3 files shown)


UNRELEASED.md
```diff
@@ -13,6 +13,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Labelled component now breaks on long lines of text, regardless of presence of naturally breaking characters (hyphens, whitespace, etc.) ([#3156](https://github.com/Shopify/polaris-react/pull/3156))
 - Added optional `isFiltered` prop to `ResourceList` to conditionally render more informative select all button label ([#3153](https://github.com/Shopify/polaris-react/pull/3153))
 - Exported `PositionedOverlay` component for use in consuming applications ([#3161](https://github.com/Shopify/polaris-react/pull/3161))
+- Update to use `esnext` as a custom mainField instead of `sewing-kit:esnext` to match updated sewing-kit behaviour ([#3169](https://github.com/Shopify/polaris-react/pull/3169))
```
kaelig, line 16:
```suggestion
- Update to use `esnext` as a custom mainField instead of `sewing-kit:esnext` to match updated sewing-kit behavior ([#3169](https://github.com/Shopify/polaris-react/pull/3169))
```

🇺🇸

PR #3171, opened 2020-08-12
[Tabs] Loosen content type

KwanLam17:

We want to add badges to the Tabs content. In our use-case, our tab opens up a list and we would like a badge to show the # of items in that list.

Fixes https://github.com/Shopify/polaris-react/issues/3162

* Loosen the type of Tabs content to allow for React.ReactNode instead of only strings
* Add Tabs example with Badge in content

## 

### How to 🎩

See that the below code type checks.

<details>

```jsx

import {Page, Card, Tabs, Badge} from '../src';

    [],
  );

  const tabs = [
    {
      id: 'all-customers',
      content: (
        <span>
          All <Badge status="info">New</Badge>
        </span>
      ),
    },
    {
      id: 'accepts-marketing',
      content: (
        <span>
          Accepts marketing <Badge status="new">2</Badge>
        </span>
      ),
    },
  ];

  return (
    <Page title="Playground">
      <Card>
          </Card.Section>
        </Tabs>
      </Card>
    </Page>
  );
}
```

</details>

### DIFF
--- a/src/components/Tabs/README.md
+++ b/src/components/Tabs/README.md
@@ -183,3 +183,50 @@ Also known as [Segmented controls](https://developer.apple.com/design/human-inte
 ![Fixed tabs on iOS](/public_images/components/Tabs/ios/fixed@2x.png)
 
 <!-- /content-for -->
+
+### Tabs with badge content
+
+Use to inform a piece of information about the tabs.
+
+```jsx
+function TabsWithBadgeExample() {
+  const [selected, setSelected] = useState(0);
+
+  const handleTabChange = useCallback(
+    (selectedTabIndex) => setSelected(selectedTabIndex),
+    [],
+  );
+
+  const tabs = [
+    {
+      id: 'all-customers-fitted',
+      content: (
+        <span>
+          All <Badge status="new">10+</Badge>
+        </span>
+      ),
+      accessibilityLabel: 'All customers',
+      panelID: 'all-customers-fitted-content',
+    },
+    {
+      id: 'accepts-marketing-fitted',
+      content: (
+        <span>
+          Accepts marketing <Badge status="new">4</Badge>
+        </span>
+      ),
+      panelID: 'accepts-marketing-fitted-Ccontent',
+    },
+  ];
+
+  return (
+    <Card>
+      <Tabs tabs={tabs} selected={selected} onSelect={handleTabChange} fitted>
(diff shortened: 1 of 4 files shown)


src/components/Tabs/types.ts
```diff
@@ -4,7 +4,7 @@ export interface TabDescriptor {
   /** A destination to link to */
   url?: string;
   /** Content for the tab */
-  content: string;
+  content: React.ReactNode;
```
amireynoso, line 7: I think it wouldn't hurt to add a test ensuring React nodes are rendered correctly when passed as `content`.

src/components/Tabs/README.md
```diff
@@ -183,3 +183,50 @@ Also known as [Segmented controls](https://developer.apple.com/design/human-inte
 ![Fixed tabs on iOS](/public_images/components/Tabs/ios/fixed@2x.png)
 
 <!-- /content-for -->
+
+### Tabs with badge content
+
+Use to inform a piece of information about the tabs.
```
amireynoso, line 189: I'm not sure about this example, it sounds like we're defining a UX pattern (which I'm not sure has been validated by the Polaris team). 

We could probably make the example a little bit more generic (e.g. passing down bolded text?) just to exemplify that you can pass any React node. How are other similar components that allow for any React node exemplified?

BPScott: > I'm not sure has been validated by the Polaris team

The polaris team isn't here to be some overarching judge who says what you can and can't do. Us "validating" things that product teams want to do is basically saying "the company doesn't trust product teams, then need some oversight" which is a straight up overzealous and untrue nonsense :) We trust product teams, you don't need our "blessing" - you can get a UX review from anybody, not just polaris - it's just that we're usually a bit more readily available for that sort of review work.

---

I'm happy with the description test, though I wonder if a more realistic example might be to show the number of items in that will be shown within the tab, something like:

- `All <Badge status="info">10+</Badge>`
-  `Accepts marketing <Badge status="info">5</Badge>`


What do ya think?

amireynoso: Really appreciate the context, Ben! 🙏

KwanLam17: Agreed Ben, the number of items is definitely more realistic! Will update.

BPScott:
All looks great to me! Merge when you're ready :shipit:

PR #3178, opened 2020-08-13
Being able to target element position inside the Frame content

alexandcote:
We have an issue with Safari 10. `height: 100%` doesn't work as expected because all the parent element doesn't set `height: 100%`. To fix that, we can use a `position: absolute` trick. To be able to use this trick, we need to set `position: relative` on the Frame content.

This is a screenshot of the AppBridge Frame on safari 10.
image

Setting `position: relative` on the Frame content so we can use `position: absolute` inside a children.
```
position: absolute;
top: 0;
left: 0;
right: 0;
bottom: 0;
```

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -8,6 +8,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Added `position: relative` to content container within Frame ([#3178](https://github.com/Shopify/polaris-react/pull/3178))
+
 ### Bug fixes
 
 ### Documentation
--- a/src/components/Frame/Frame.scss
+++ b/src/components/Frame/Frame.scss
@@ -218,6 +218,7 @@ $skip-vertical-offset: rem(10px);
 }
 
 .Content {
+  position: relative;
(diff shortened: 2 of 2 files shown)


UNRELEASED.md
```diff
@@ -8,6 +8,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Changed `Frame` to let consumers take the entire space available. ([#3178](https://github.com/Shopify/polaris-react/pull/3178)).
```
kyledurand, line 11:
```suggestion
- Added `position: relative` to content container within Frame ([#3178](https://github.com/Shopify/polaris-react/pull/3178))
```
Maybe something like this to be less ambiguous

src/components/Frame/Frame.scss
```diff
@@ -218,6 +218,7 @@ $skip-vertical-offset: rem(10px);
 }
 
 .Content {
+  position: relative;
```
beefchimi, line 221: I agree with @kyledurand - this change is innocent enough and shouldn't cause problems. I have not tophatted, but I trust that this is a viable solution.

PR #3179, opened 2020-08-18
Export `colorToHsla` (Shopify#3033)

### LINKED ISSUE #3033: [Color Picker] colorToHsla not exported
Hi folks, when I was trying to use color picker and wanted to convert between Hex and HSBAColor realised that `colorToHsla` is not exported at https://github.com/Shopify/polaris-react/blob/master/src/index.ts#L17 here, although it's exported in https://github.com/Shopify/polaris-react/blob/master/src/utilities/color-transformers.ts#L299. Is there any specific reason for that?

## Expected behavior

`colorToHsla` utility function should be exported as well, given it's quite handy.

## Actual behavior

Although other helper methods such as `rgbToHex` are exported, `colorToHsla` is not exported.

manzhikov:

Fixes #3033 

Export `colorToHsla` function for use in consuming applications

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -12,6 +12,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- Fixed export function `colorToHsla` ([#3179](https://github.com/Shopify/polaris-react/pull/3179))
+
 ### Documentation
 
 ### Development workflow
--- a/src/index.ts
+++ b/src/index.ts
@@ -14,6 +14,7 @@ export type {
   HSBLAColor,
 } from './utilities/color-types';
 export {
+  colorToHsla,
   rgbToHex,
   rgbToHsb,
   rgbToHsl,


BPScott:
Per my response to #3090, which also dealt with color functions:


My current opinion is that these color utilities shouldn't be public exports, they should be internal utilities only. I don't believe we document or recommend their usage anywhere. When we add something to our public API surface area we make a promise to keep supporting it and it makes refactoring any internals that depend upon those utility functions much harder. Because of that I want to keep out public API surface area as small as possible - ideally only components and types for their props. There are a few exceptions to this but they're mostly down to accidents rather down to explicit decisions that something needs to be public.

I believe the only reason why these color utility functions are public currently is that they were publicly exported accidentally at some point and then when I moved some imports around it turned out that somebody was already depending on them so they had to stick around. I've not yet had the time and strength to work our how to remove these in a major version. I think the effort would be pretty low but my focus has been elsewhere.

So I'm going to close this for reasons stated above - these color utility functions shouldn't be public and the only reason they are currently is due to oversight and then not wanting to break backwards compatibility.

PR #3181, opened 2020-08-18
[Tabs] Bold in new design language

tmlayton:

Fixes https://github.com/Shopify/polaris-ux/issues/410. Font weight is supposed to be 500

Screen Shot 2020-08-18 at 12 32 43 PM

### DIFF
--- a/src/components/Tabs/Tabs.scss
+++ b/src/components/Tabs/Tabs.scss
@@ -303,6 +303,7 @@ $focus-state-box-shadow-color: rgba(0, 0, 0, 0.8);
 .newDesignLanguage {
   .Title {
     padding: spacing(tight) spacing();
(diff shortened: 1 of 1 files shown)


src/components/Tabs/Tabs.scss
```diff
@@ -318,6 +319,14 @@ $focus-state-box-shadow-color: rgba(0, 0, 0, 0.8);
 
   .Tab {
     padding: spacing(tight) spacing(extra-tight);
+
+    &:hover,
+    &:active,
+    &:focus {
```
kyledurand, line 325: Why the specificity here? This should just work directly under `.Tab` unless we override the font weights on focus, active, and hover?

tmlayton: There is

PR #3184, opened 2020-08-20
Migrate to storybook v6

BPScott:

Keeping up to date. Prepping for a future where docs are much more easily leveraged and are consumable within SB.

Upating to SB 6

We're stuck with knobs and contexts for the moment, as the modern
args/toolbars don't support a way of deeplinking to a specific setup
(e.g. "Show me this story with the new DL enabled")

### How to 🎩

Run SB and see that it works:

- AppProvider / Frame stories don't get double wrapped in an app provider
- Everything renders the same according to chromatic
- Toggling strict mode / new DL still works, and you can deep link to a specific story with the new DL enabled.


.storybook/main.js
```diff
@@ -29,49 +23,24 @@ module.exports = {
...
+                '../build/cache/storybook/markdown',
+              ),
             },
           },
           {
             loader: `${__dirname}/polaris-readme-loader.js`,
           },
         ],
       },
-      {
```
BPScott, line 49: v6 as configures TS, and images to work out the box

.storybook/polaris-readme-loader.js
```diff
@@ -44,13 +48,15 @@ export function ${example.storyName}() {
...
-  decorators: [withA11y],
-  parameters: {
-    notes: ${JSON.stringify(example.description)},
-    percy: {skip: ${JSON.stringify(!testIndividualExamples)}},
-  }
+
+${example.storyName}.storyName = ${JSON.stringify(example.name)};
+${example.storyName}.args = {omitAppProvider: ${readme.omitAppProvider}};
+${example.storyName}.parameters = {
+  layout: '${hasFullscreenLayout ? 'fullscreen' : 'padded'}',
```
BPScott, line 55: The new layout parameter controls if a story renders flush to the edge ("full screen") or has a 1rem of padding ("padded"), which removes the need for our PaddingDecorator helper

.storybook/preview.js
```diff
@@ -7,40 +7,25 @@ import DefaultThemeColors from '@shopify/polaris-tokens/dist-modern/theme/base.j
 import {AppProvider} from '../src';
 import enTranslations from '../locales/en.json';
 
-addParameters({
-  options: {
-    showRoots: true,
```
BPScott, line 12: This is now true by default so no need to configure it

package.json
```diff
@@ -110,30 +110,27 @@
     "@shopify/sewing-kit": "^0.132.2",
     "@shopify/shrink-ray": "^2.3.1",
     "@shopify/splash": "^0.0.8",
-    "@storybook/addon-a11y": "^5.3.19",
-    "@storybook/addon-actions": "^5.3.19",
+    "@storybook/addon-a11y": "^6.0.12",
     "@storybook/addon-console": "^1.2.1",
     "@storybook/addon-contexts": "^5.3.19",
```
BPScott, line 115: Contexts is deprecated and doesn't have a v6 version, but we're stuck with it as the new args/toolbars approach of configuring the "Strict Mode" and "Design Language" toggle buttons doesn't support deeplinking. I've raised a bug to request support for that so we can move away: https://github.com/storybookjs/storybook/issues/11604

kaelig:
In the "Docs" tab, "All examples" stories are adding noise at the top of the page. What do you think about removing them from the docs tab? ([this section of Storybook's documentation](https://storybook.js.org/docs/html/writing-docs/docs-page#story-level) make me think it's possible to do so)

BPScott:
Found an option for that by spelunking the source code, because it wasn't documented :)

kaelig:
Ha! Good find 🕵️

kaelig:
Can you also update the changelog?

PR #3193, opened 2020-08-20
[ActionMenu] Convert to functional component

kyledurand:

I have some upcoming changes in this component and they will be easier to do as a functional component vs PC

## 

### How to 🎩

http://localhost:6006/iframe.html?id=playground-playground--details-page&contexts=New%20Design%20Language=Enabled%20-%20Light%20Mode#

Toggle action menus:
image

### 🎩 checklist


src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -110,17 +90,25 @@ export class ActionMenu extends PureComponent<ActionMenuProps, State> {
...
-  };
+  }
 
-  private handleMenuGroupToggle = (group: string) => {
-    this.setState(({activeMenuGroup}) => ({
-      activeMenuGroup: activeMenuGroup ? undefined : group,
-    }));
-  };
+  function handleMenuGroupToggle(group: string) {
+    setActiveMenuGroup(activeMenuGroup ? undefined : group);
```
kyledurand, line 96: This was a callback before but I don't think it needed to be. Any objections to removing it?

alex-page: No objections here.

src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -23,48 +23,28 @@ export interface ActionMenuProps {
...
-      </div>
-    );
-  }
-
-  // eslint-disable-next-line @shopify/react-no-multiple-render-methods
-  private renderActions = () => {
-    const {newDesignLanguage} = this.context || {};
-    const {actions = [], groups = []} = this.props;
-    const {activeMenuGroup} = this.state;
+  function renderActions() {
```
BPScott, line 47: Extract this out so it - A definition of a react component inside another react component is odd, and probably bad for performance as it means the function gets recreated on every render.

Could do `function Actions() {}`, then below you can replace `renderActions()` with `<Actions/>` (possibly with some props needed)

src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -110,17 +90,25 @@ export class ActionMenu extends PureComponent<ActionMenuProps, State> {
...
-  private handleMenuGroupToggle = (group: string) => {
-    this.setState(({activeMenuGroup}) => ({
-      activeMenuGroup: activeMenuGroup ? undefined : group,
-    }));
-  };
+  function handleMenuGroupToggle(group: string) {
+    setActiveMenuGroup(activeMenuGroup ? undefined : group);
+  }
+
+  function handleMenuGroupClose() {
```
BPScott, line 99: Nitpick: I tend to put these callbacks up towards the top next to the useState stuff. 

Also I think this and handleMenuGroupToggle should be wrapped in a useCallback() to memoise it if you're passing it into another react component

src/components/ActionMenu/ActionMenu.tsx
BPScott: If this gets wrapped in a useCallback (which I think it should be so you don't trigger rerenders of MenuGroup because you pass in a new function every rerender) then you'll want to use the callback style notation so you don't end up creating a new function every time activeMenuGroup changes

kyledurand:
@BPScott thanks for the comments! I think I've addressed them all

PR #3199, opened 2020-08-26
[Banner] Refactor to a functional component and fix its `id` server/client mismatch

Tom-Bonnike:

We’re getting server/client `id` mismatches on the Admin dashboard on pages that use `Banner`s. This is because the current way of generating unique `id`s is not foolproof.

The first commit is refactoring `Banner` to a functional component since we can’t use hooks in class components.
The second commit fixes the `id` server/client mismatch by using the `useUniqueId` hook instead of the `Banner`’s custom `uniqueID()` util.

### How to 🎩

We can’t 🎩  the bug fix on the Playground since it’s not server-side rendered.

Here‘s how I tested it:
- Go to a page on your local `web` that renders a `Banner`; I used `/admin/settings/payments`. Notice how there’s a warning in the console that looks like this one:
Screenshot 2020-08-26 at 15 53 03

- Checkout the `fix-mismatch-banner-ids` branch on `polaris-react`
- Run `yarn build` then `yarn build-consumer web`
- Restart your `web` server then refresh the page on `/admin/settings/payments`, notice how there’s no more warning in the console

I think relying on automated testing for the rest is OK here as I’m not really changing any behaviour/styles, just implementation details: the unit tests & visual regression tests are passing.
Let me know if you still want me to go through the regular 🎩 checklist.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -15,6 +15,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Bug fixes
 
 - Fix `Button` css in a `connectedTop` or `fullWidth` `ButtonGroup` ([#3215](https://github.com/Shopify/polaris-react/pull/3215)).
+- Fixed `Banner`’s `id` being mismatched on server VS client ([#3199](https://github.com/Shopify/polaris-react/pull/3199)).
 
 ### Documentation
 
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -18,9 +18,10 @@ import {
   CircleDisabledMajorFilled,
 } from '@shopify/polaris-icons';
 
-import {BannerContext} from '../../utilities/banner-context';
 import {classNames, variationName} from '../../utilities/css';
+import {BannerContext} from '../../utilities/banner-context';
 import {useFeatures} from '../../utilities/features';
+import {useUniqueId} from '../../utilities/unique-id';
 import type {
   Action,
   DisableableAction,
@@ -73,7 +74,7 @@ export const Banner = forwardRef<BannerHandles, BannerProps>(function Banner(
   const {newDesignLanguage} = useFeatures();
   const withinContentContainer = useContext(WithinContentContext);
   const buttonSizeValue = withinContentContainer ? 'slim' : undefined;
-  const id = uniqueID();
+  const id = useUniqueId('Banner');
   const {
     wrapperRef,
     handleKeyUp,
@@ -176,12 +177,6 @@ export const Banner = forwardRef<BannerHandles, BannerProps>(function Banner(
     </BannerContext.Provider>
   );
 });
-
-let index = 1;
-
-function uniqueID() {
-  return `Banner${index++}`;
-}
 
 function SecondaryActionFrom({action}: {action: Action}) {
   if (action.url) {
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -1,4 +1,10 @@
-import React, {PureComponent, createRef} from 'react';
+import React, {
+  forwardRef,
+  useRef,
+  useState,
+  useContext,
+  useImperativeHandle,
+} from 'react';
 import {
   CancelSmallMinor,
   CircleTickMajorTwotone,
@@ -12,9 +18,9 @@ import {
   CircleDisabledMajorFilled,
 } from '@shopify/polaris-icons';
 
-import {FeaturesContext} from '../../utilities/features';
 import {BannerContext} from '../../utilities/banner-context';
 import {classNames, variationName} from '../../utilities/css';
+import {useFeatures} from '../../utilities/features';
 import type {
   Action,
   DisableableAction,
@@ -31,10 +37,6 @@ import {WithinContentContext} from '../../utilities/within-content-context';
 import styles from './Banner.scss';
 
 export type BannerStatus = 'success' | 'info' | 'warning' | 'critical';
-
-interface State {
-  showFocus: boolean;
-}
 
 export interface BannerProps {
   /** Title content for the banner. */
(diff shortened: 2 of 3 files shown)


src/types.ts
```diff
@@ -328,3 +328,7 @@ export enum Key {
 export interface CheckboxHandles {
   focus(): void;
 }
+
+export interface BannerHandles {
+  focus(): void;
+}
```
Tom-Bonnike, line 334: Maybe CheckboxHandles and BannerHandles should be a single type? Not sure what other kinds of methods we’ll expose on components.

src/components/Banner/Banner.tsx
```diff
@@ -55,202 +59,195 @@ export interface BannerProps {
...
+          {headingMarkup}
+          {contentMarkup}
+        </div>
+      </div>
+    </BannerContext.Provider>
+  );
+});
 
-function secondaryActionFrom(action: Action) {
+function SecondaryActionFrom({action}: {action: Action}) {
```
Tom-Bonnike, line 250: Functions returning JSX are React components, so I made it one.

src/components/Banner/Banner.tsx
```diff
@@ -55,202 +59,195 @@ export interface BannerProps {
...
-                <Button
-                  plain
-                  icon={CancelSmallMinor}
-                  onClick={onDismiss}
-                  accessibilityLabel="Dismiss notification"
-                />
-              </div>
-            ) : null;
+  return (
+    <BannerContext.Provider value>
```
Tom-Bonnike, line 217: Should we just ditch this? 🤔

src/components/Banner/Banner.tsx
```diff
@@ -55,202 +59,214 @@ export interface BannerProps {
...
-  static contextType = FeaturesContext;
-  context!: React.ContextType<typeof FeaturesContext>;
-
-  state: State = {
-    showFocus: false,
-  };
+interface UseBannerPropsForStatusResult {
+  iconColor: IconProps['color'];
+  defaultIcon: IconProps['source'];
+  ariaRoleType?: 'alert';
```
Tom-Bonnike, line 65: wonder if this should default to `status` 🤔

UNRELEASED.md
```diff
@@ -12,6 +12,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- Fix `Banner`’s `id` server/client mismatch ([#3199](https://github.com/Shopify/polaris-react/pull/3199)).
```
Tom-Bonnike, line 15: Moving from a class to a functional component is probs not worth an entry in `code quality`?

src/components/Banner/Banner.tsx
```diff
@@ -55,202 +59,210 @@ export interface BannerProps {
...
-                ? styles.withinContentContainer
-                : styles.withinPage,
-              newDesignLanguage && styles.newDesignLanguage,
-            );
+  useImperativeHandle(bannerRef, () => ({
+    focus: () => {
+      wrapperRef.current?.focus();
+      setShouldShowFocus(true);
+    },
+  }));
```
Tom-Bonnike, line 125: I tried it and it seems to work great.
You can paste this in the Playground:

```js

  const bannerRef = React.useRef()

  React.useEffect(() => {
    bannerRef.current?.focus()
  }, [])

  return (
    <Page title="Playground">
       <Banner ref={bannerRef} title="Some banner title" />
    </Page>
  );
}

```

BPScott:
Yay thanks for doing this! I've not had time to review yet, but shall try and get to it early next week

chloerice (approved):
Thanks for contributing, @Tom-Bonnike, the code looks great! I tried to tophat in `web`, but in order to get `dev up` working I had to reset railgun so I didn't have any banners visible because my data is so bare. The tophat in Chroma looks solid with and without the new design lang enabled 🚢

UNRELEASED.md
BPScott, line 15: Nitpick: Our convention is to use past case in changelogs

```suggestion
- Fixed `Banner`’s `id` being mismatched on server vs client ([#3199](https://github.com/Shopify/polaris-react/pull/3199)).
```

src/components/Banner/Banner.tsx
```diff
@@ -55,202 +59,210 @@ export interface BannerProps {
...
-  static contextType = FeaturesContext;
-  context!: React.ContextType<typeof FeaturesContext>;
-
-  state: State = {
-    showFocus: false,
-  };
+interface BannerAttributes {
+  iconColor: IconProps['color'];
+  defaultIcon: IconProps['source'];
+  ariaRoleType?: 'alert';
```
BPScott, line 65:
```suggestion
  ariaRoleType: 'status' | 'alert';
```

src/components/Banner/Banner.tsx
```diff
@@ -55,202 +59,210 @@ export interface BannerProps {
...
-              onDismiss && styles.hasDismiss,
-              showFocus && styles.keyFocused,
-              withinContentContainer
-                ? styles.withinContentContainer
-                : styles.withinPage,
-              newDesignLanguage && styles.newDesignLanguage,
-            );
+  useImperativeHandle(bannerRef, () => ({
+    focus: () => {
+      wrapperRef.current?.focus();
```
BPScott, line 122: Yay new features! (this is all good, I'm just excited that we're starting to dip our toes into these super-modern features)

src/components/Banner/Banner.tsx
```diff
@@ -55,202 +59,210 @@ export interface BannerProps {
...
-              title,
-              children,
-              status,
-              onDismiss,
-              stopAnnouncements,
-            } = this.props;
-            let color: IconProps['color'];
-            let defaultIcon: IconProps['source'];
-            let ariaRoleType = 'status';
+function useBannerFocus(bannerRef: React.Ref<BannerHandles>) {
```
BPScott, line 116: Extracting the focusabiility into a hook is really cool!

src/types.ts
BPScott: I think it's fine to have two interfaces that happen to have the same shape but serve different purposes - if we ever need extra handles on a CheckBox but not a Banner we'd end up having to refactor back into two so I think merging them is premature.

src/types.ts
```diff
@@ -328,3 +328,7 @@ export enum Key {
 export interface CheckboxHandles {
   focus(): void;
 }
+
+export interface BannerHandles {
```
BPScott, line 332: Everything in src/types is publicly exported, I'd like to avoid placing extra stuff in here if at all possible (as then changing /moving this stuff becomes a breaking change).

If BannerHandles is only used within the Banner component, can it be defined in there too please?

src/components/Banner/Banner.tsx
```diff
@@ -55,202 +59,210 @@ export interface BannerProps {
...
-    showFocus: false,
-  };
+interface BannerAttributes {
+  iconColor: IconProps['color'];
+  defaultIcon: IconProps['source'];
+  ariaRoleType?: 'alert';
+}
 
-  private wrapper = createRef<HTMLDivElement>();
+function useBannerAttributes(status: BannerProps['status']): BannerAttributes {
```
BPScott, line 68: Thoughts on passing a `newDesignLanguage` boolean arg into this and then it can be a regular function instead of a hook?

BPScott, line 68: Nitpick: The first function in our component files tends to be the component function itself, so having `useBannerAttributes`, and `useBannerFocus` above `Banner` means you have to scroll half the file to see the most important function, which feels weird. Can we have Banner be the first function, followed by useBannerAttributes/useBannerFocus?

BPScott (changes requested):
Note to self: When telling people you'll review something at the start of the week remember that the first half of your week is full of meetings.

Thanks for doing this and sorry for the delay!

I've dropped a few comments inline. The important one that needs to be fixed is adding the `role=status` to non-critical/non-warning banners as you pointed out. Turns out we did that before but it got lost in the refactoring :)

src/components/Banner/Banner.tsx
Tom-Bonnike: Damn; Totally missed that. 😓  Reviewed the component without the diff and realized it was missing a default haha. Thank you!

src/types.ts
Tom-Bonnike: Right. I put it here to match the Checkbox’s but agree. 👍

src/components/Banner/Banner.tsx
Tom-Bonnike: Sure thing! I’m not a big fan of relying on hoisting, but I see that’s something we do a lot at Shopify! Maybe I’ll start liking it at some point 😄

Tom-Bonnike: ⚠️  Opinions ahead 😄 

IMO passing hook values as arguments/props is kind of an anti-pattern since hooks are great “dependency injectors”. Reducing the number of attributes you need to pass around leads to better APIs and less “cognitive load” (obviously the complexity of that hook isn’t such a big issue here, but I still tend to always do it).
I guess it also has the benefit of allowing you to re-use that hook in other places without having to care about its internal implementation, and allows for simpler composition with other hooks.

One could argue this leads to more calls then needed but `useFeatures` only reads from context, which is cheap. Even if it somehow computed something expensive, it should be handled inside it (like with a `useMemo`), it should not be the consumer’s responsibility to think about performance.

---

I don’t feel strongly about it so happy to make the change, though!

src/components/Banner/Banner.tsx
Tom-Bonnike: Fixed it. Assertions added in 5548b91fbb00c974d7dd3f5e0a8ed8cc72610848.

Tom-Bonnike:
Thanks for the reviews! Applied your comments @BPScott, good catch on that aria role 🙈

src/components/Banner/tests/Banner.test.tsx
```diff
@@ -68,18 +85,6 @@ describe('<Banner />', () => {
...
-    expect(banner.find(Icon).prop('source')).toBe(CircleDisabledMajorTwotone);
-    expect(banner.find(Icon).prop('color')).toBe('redDark');
-  });
-
-  it('uses a default icon', () => {
-    const banner = mountWithAppProvider(<Banner />);
-    expect(banner.find(Icon).prop('source')).toBe(FlagMajorTwotone);
-    expect(banner.find(Icon).prop('color')).toBe('inkLighter');
-  });
-
```
Tom-Bonnike, line 82: (Just moved that up)

src/components/Banner/Banner.tsx
BPScott: Cool, that all makes sense to me :)

> Reducing the number of attributes you need to pass around leads to better APIs and simpler usage.

It leads to simpler usage but IIMO it can potentially hide the complexity around what something depends upon. With "pure" functions dependencies are simple - the arguments of the functions; once you move away from that by using hooks for DI then IMO that relationship can sometimes get a bit harder to see and now folks have to understand the additional complexity of knowing "you also need to provide this context" which isn't apparent from skimming over an argument list - so there's a tradeoff between "easier if you've already got the context provider setup" and "harder to setup the first time as you need to work out how the context provider does its thing". That's not an issue here as there's already an expectation of "you need to wrap everything in an AppProvider" for polaris-react components, but might be something to consider of when building other APIs :)

This particular case is very low-key - its a single arg, and the context will always already be provided so there's nothing new to teach there - so I don't have strong opinions either way. I'm happy to leave it as it is :)

BPScott (approved):
Thanks for being patient while I was on holiday these past few days. Fantastic work, looks all great to me! 

Polaris convention is that you merge your own PRs once you're happy with them (the polaris-team being the ones that merge would imply we're some kind of gatekeepers which we absolutely don't want to encourage - everyone should feel like they own this codebase) so :shipit: when you're happy


> good catch on that aria role

There's a very real chance I would have totally missed it if not for your initial pondering around that, so thank you for pointing it out!

src/components/Banner/Banner.tsx
Tom-Bonnike: > It leads to simpler usage but IIMO it can potentially hide the complexity around what something depends upon.

Right, you got me on that one. 😄 

I made the change!

Tom-Bonnike: Right, interesting comparison with classes. 

> Maybe I’ll start liking it at some point 😄

I think I’ve already started liking it tbh. Until recently, I used arrow functions everywhere, so this was out of the question. Now that I tend to default to function declarations, my brain probably needs to adjust to this being possible 😄 

I think somehow I don’t (didn’t) like the idea of using something that you… haven’t seen before? But it’s kind of a stupid way to think about it.

Among other things, functions allow you to hide “implementation details”, and putting “details” at the bottom makes sense since… they’re less important 😄 
They’re in the same file too, so jumping to them when needed is not a big hassle.

There‘s definitely value in knowing that the first function definition will always be the component, so you can skim through files quickly.

Tom-Bonnike:
Cool. I’ll merge as soon as CI passes. I’m guessing y’all are still responsible for releasing the new versions?

PR #3207, opened 2020-08-27
[Page header] Prototype

kyledurand:

⚠️ This PR is not safe to ship

This pr explores some of the page header changes being considered [here](https://www.figma.com/file/DOLJoC3CbldApU26mIIDiK/Page-Component?node-id=0%3A1) fixes https://github.com/Shopify/polaris-ux/issues/431

Notes:
These changes within the figma file represent some pretty challenging layout changes. Here are some thoughts on how we can accomplish them
1. Rolling up actions as screen size decreases
- Measure the actions container width and each action within it. 
- Split the array of given actions, the ones that cannot fit will be placed into the more actions menu group. The others are rendered beside
- Debounce the measurer and listen for page resizing to make sure there's no overflow when the page width decreases
2. Middle text truncation. 
- Considerably more difficult because we don't know how many characters we have to remove from the middle of the text
- We don't know the width of the individual characters, and the characters vary in width.
- This may not be possible without a set character limit and then base the middle truncation on that value

## 

### How to 🎩

[With secondary meta data](https://5d559397bae39100201eedc1-dkozgvkuas.chromatic.com/iframe.html?id=playground-playground--details-page&contexts=New%20Design%20Language%3DEnabled%20-%20Light%20Mode&viewMode=story)

[Without secondary meta data](https://5d559397bae39100201eedc1-pxactspmyh.chromatic.com/iframe.html?id=playground-playground--details-page&contexts=New%20Design%20Language%3DEnabled%20-%20Light%20Mode&viewMode=story)

### 🎩 checklist

### DIFF
--- a/playground/DetailsPage.tsx
+++ b/playground/DetailsPage.tsx
@@ -2,6 +2,7 @@ import React, {useCallback, useRef, useState, SVGProps} from 'react';
 import {
   CirclePlusMinor,
   DuplicateMinor,
+  PrintMinor,
   ViewMinor,
 } from '@shopify/polaris-icons';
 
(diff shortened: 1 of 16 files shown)


playground/DetailsPage.tsx
```diff
@@ -424,21 +424,14 @@ export function DetailsPage() {
...
         {
           content: 'View',
           icon: ViewMinor,
         },
       ]}
       actionGroups={[
-        {
-          title: 'Promote',
-          actions: [{content: 'Share on Facebook'}],
-        },
```
kyledurand, line 441: Removing these temporarily until we can confidently roll these actions up as screen sizes decrease

src/components/ActionMenu/components/SecondaryAction/SecondaryAction.scss
```diff
@@ -0,0 +1,25 @@
+@import '../../../../styles/common';
+
+$breakpoint: 1080px;
+
+.SecondaryAction {
+  // Hacks
+  // Eventually pull in button-base and modify
+  // stylelint-disable declaration-no-important
+  button {
```
kyledurand, line 9: These are temporary hacks but the safer way to do this would be to import button-base and do some overrides

PR #3214, opened 2020-09-02
Update _border-radius.scss

tleunen:

Border radius should use the new css variables based on this document: https://github.com/Shopify/polaris-react/blob/master/documentation/Color%20system.md#overrides

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -12,6 +12,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- Fix `border-radius` scss function to use the new css variables when defined ([#3214](https://github.com/Shopify/polaris-react/pull/3214)).
+
 ### Documentation
 
 ### Development workflow
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -12,8 +12,6 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
-- Fix `border-radius` scss function to use the new css variables when defined ([#3214](https://github.com/Shopify/polaris-react/pull/3214)).
-
 ### Documentation
 
 ### Development workflow
--- a/src/styles/foundation/_border-radius.scss
+++ b/src/styles/foundation/_border-radius.scss
@@ -1,6 +1,6 @@
 $border-radius-data: (
-  base: 3px,
-  large: 6px,
+  base: var(--p-border-radius-base, 3px),
+  large: var(--p-border-radius-wide, 6px),
 );
 
 /// Returns the border radius of the specified size.


tleunen:
@BPScott - Could you quickly review this as a code owner? :) 

Also, someone must approve the percy check I believe.

alex-page (changes requested):
@tleunen something odd is happening here. The CSS variable is getting applied when the old design language is being used, despite the fallback existing. I will try and investigate today/tomorrow.

BPScott:
This is a breaking change. Currently these mixins output plain values, which means that in sass you can write things like `border-radius() + 1px`, which would compile to 4px. With this change that same code would result in a silen error as `var(--something) * 2` isn't a valid sass math operation, it might be "valid" CSS from a string concatenation perspective though so sass will put that into your stylesheet without any compile time error (it'd result in valid and working css if you wrapped it in a `calc` to push the math to the client side but that wasn't required before).

We actually do this in out own control-backdrop mixin  and some other places so the visual changes are probably due to that mixin now outputing invalid css. You can see this in action if you do a `yarn build` and investigate dist/styles.css. Look for `Polaris-ColorPicker__MainColor` (added whitespace here for readability:

```

.Polaris-ColorPicker__MainColor{
  background-image:linear-gradient(45deg, var(--p-surface, #dfe3e8) 25%, transparent 25%), linear-gradient(-45deg, var(--p-surface, #dfe3e8) 25%, transparent 25%), linear-gradient(45deg, transparent 75%, var(--p-surface, #dfe3e8) 75%), linear-gradient(-45deg, transparent 75%, var(--p-surface, #dfe3e8) 75%); 
  background-size:1.6rem 1.6rem;
  background-position:0 0, 0 0.8rem, 0.8rem -0.8rem, -0.8rem 0;
  position:relative; overflow:hidden; 
  height:16rem; 
  width:16rem; 
  /* 
    The original source code for this line was: border-radius: var(--p-border-radius-base, border-radius() + 1px);
    This outputted CSS is not valid
 */
  border-radius:var(--p-border-radius-base, var(--p-border-radius-base, 3px)1px); 
  cursor:pointer; 
}

```




Alas I don't think this is something we can fix at a global level while supporting the two systems. You'll have to update all your usages of the mixin with `var(--p-border-radius-base, border-radius())` type things - same as how you need to replace replace the usage of the `color` mixin with that sort of wrapping.

tleunen:
Oh, you're right Ben and we cannot do this at this level so I'll close this PR.

What is the plan for these mixins once we completely move to variables then? Are we going to delete them entirely?

BPScott:
> What is the plan for these mixins once we completely move to variables then? Are we going to delete them entirely?

🔥 🔥 🔥 

My hope is to remove a load of our mixins that return token-like data (e.g. `color` uses data from polaris-tokens, `border-radius` has the data still stored in this repo) in a future major version and instead we should move our tokens usage over to be css variable driven. As you've spotted this is already happening with the color and border-radius variables thanks to the new DL. There are other mixins like `spacing()` that we haven't needed to adjust as part of the new DL work, but should eventually follow a similar pattern once we create css variables for their values.

PR #3216, opened 2020-09-04
Add a `suffix` to an Action List Item

tleunen:

Some consumers would like to have a way to provide a "suffix" to an item and this PR adds this functionality.

Screen Shot 2020-09-04 at 9 47 36 AM

### 🎩 checklist

### DIFF
--- a/src/components/ActionList/components/Item/Item.tsx
+++ b/src/components/ActionList/components/Item/Item.tsx
@@ -22,6 +22,7 @@ export function Item({
   onAction,
   icon,
   image,
+  suffix,
   disabled,
   external,
   destructive,
@@ -38,7 +39,7 @@ export function Item({
     newDesignLanguage && styles.newDesignLanguage,
   );
 
-  let imageElement = null;
+  let imageElement: React.ReactNode | null = null;
 
   if (icon) {
     imageElement = (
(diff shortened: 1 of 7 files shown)


alex-page (approved):

src/components/ActionList/components/Item/Item.tsx
```diff
@@ -68,11 +69,15 @@ export function Item({
   );
 
   const badgeMarkup = badge && (
-    <span className={styles.BadgeWrapper}>
+    <span className={styles.Suffix}>
       <Badge status={badge.status}>{badge.content}</Badge>
     </span>
   );
 
+  const suffixMarkup = suffix && (
```
kyledurand, line 77: What does this look like when both a badge and suffix element are passed in?

kyledurand (approved):
Code looks good just one question around aesthetics

src/components/ActionList/components/Item/Item.tsx
tleunen: Yeah I actually tested this scenario, but I should have provided screenshots to help reviewers.
image

In the end, my idea was to eventually deprecate/remove the badge, and the icon, and use more generic `prefix`/`suffix` allowing to add anything instead. But I didn't want to create any breaking change at the moment.

kyledurand: Awesome, looks good and makes sense 👍

PR #3221, opened 2020-09-04
[AppProvier] Fix erroneous code examples.

colshacol:
Fixed code examples with incorrect imports.

In AppProvider's README.md...

```
import AppProvider from '@shopify/polaris';
```

has been changed to:

```
import { AppProvider } from '@shopify/polaris';
```

Reasoning: AppProvider is not the default export. I was able to figure this out quite easily, but could cost another developer tons of debugging time to refer to the docs and see that they are doing nothing wrong but still err'ing out.

### DIFF
--- a/src/components/AppProvider/README.md
+++ b/src/components/AppProvider/README.md
@@ -472,7 +472,7 @@ Translations are provided in the locales folder. When using Polaris, you are abl
 If a project has only one locale, then you can pass the JSON content from the locale file into `AppProvider`.
 
 ```jsx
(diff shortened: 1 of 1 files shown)


AndrewMusgrave (commented):
Thank you for your contribution, it looks great 🎉  Would you minding singing the CLA & adding an entry in `unreleased.md`. Re-request or ping me for a another review when your ready!

alex-page:
Fixed in b604e2dbb9f550cbe68fce73e89ce71c3c486387

PR #3222, opened 2020-09-07
[TextField] Support `inputMode` attribute

Tom-Bonnike:
# WHY are these changes introduced?

I have noticed the `TextField` doesn’t support [the `inputMode` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode). Thought I’d add support for it as I discussed this attribute twice in PRs this week. Here‘s [a good article](https://css-tricks.com/finger-friendly-numerical-inputs-with-inputmode/) about it.

Passing down the `inputMode` prop to the `TextField`’s `input`.

TextField without an `inputMode` attribute:
Screenshot 2020-09-08 at 10 46 11

TextField with `inputMode="numeric"`
Screenshot 2020-09-08 at 10 45 29

☝️ This doesn’t work on `master`, the prop is swallowed.

### 🎩 checklist

### DIFF
--- a/src/components/TextField/README.md
+++ b/src/components/TextField/README.md
@@ -872,6 +872,7 @@ Text fields have standard keyboard support.
 - If the `type` is set to `number`, then merchants can use the up and down arrow keys to adjust the value typed into the field
 - Using the `disabled` prop will prevent the text field from receive keyboard focus or inputs
 - The `readOnly` prop allows focus on the text field but prevents input or editing
+- The `inputMode` prop can be used to bring up a relevant keyboard for merchants on mobile; it’s passed down to the input as an [`inputmode` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode)
 
 #### Automatically focusing
 
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -38,6 +38,16 @@ type Type =
 
 type Alignment = 'left' | 'center' | 'right';
 
+type InputMode =
+  | 'none'
+  | 'text'
+  | 'decimal'
+  | 'numeric'
+  | 'tel'
+  | 'search'
+  | 'email'
+  | 'url';
+
 interface NonMutuallyExclusiveProps {
   /** Text to display before value */
   prefix?: React.ReactNode;
@@ -95,6 +105,8 @@ interface NonMutuallyExclusiveProps {
   minLength?: number;
   /** A regular expression to check the value against */
   pattern?: string;
+  /** Choose the keyboard that should be used on mobile devices */
+  inputMode?: InputMode;
   /** Indicate whether value should have spelling checked */
   spellCheck?: boolean;
   /** Indicates the id of a component owned by the input */
@@ -155,6 +167,7 @@ export function TextField({
   min,
   minLength,
   pattern,
+  inputMode,
   spellCheck,
   ariaOwns,
   ariaControls,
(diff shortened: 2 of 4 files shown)


src/components/TextField/TextField.tsx
```diff
@@ -405,6 +418,7 @@ export function TextField({
     maxLength,
     spellCheck,
     pattern,
```
Tom-Bonnike, line 420: I’d argue this should default to `[0-9]*` on `number` inputs to show the right keyboard. It has better support than `inputMode`.

src/components/TextField/TextField.tsx
```diff
@@ -405,6 +418,7 @@ export function TextField({
...
+    inputMode,
```
Tom-Bonnike, line 421: I’d argue this should default to `decimal` on `number` inputs. I think it’s only iOS Safari that doesn’t show the right keyboard by default on `number` inputs, though.

src/components/TextField/TextField.tsx
```diff
@@ -38,6 +38,16 @@ type Type =
...
 
+type InputMode =
+  | 'none'
+  | 'text'
+  | 'decimal'
+  | 'numeric'
+  | 'tel'
+  | 'search'
+  | 'email'
+  | 'url';
```
Tom-Bonnike, line 49: TypeScript `HTMLInputElement.inputMode` types it as `string | undefined` which is not great, but at least we wouldn’t need to update that list if more input modes are added.

src/components/TextField/README.md
```diff
@@ -872,6 +872,7 @@ Text fields have standard keyboard support.
 - If the `type` is set to `number`, then merchants can use the up and down arrow keys to adjust the value typed into the field
 - Using the `disabled` prop will prevent the text field from receive keyboard focus or inputs
 - The `readOnly` prop allows focus on the text field but prevents input or editing
+- The `inputMode` prop can be used to bring up a relevant keyboard for merchants on mobile
```
Tom-Bonnike, line 875: On a somewhat related topic… I see there’s a `KeyboardAccessories` component here: “https://polaris.shopify.com/components/forms/keyboard-accessories#section-best-practices”, but there’s only documentation in the repo? Am confused.

src/components/TextField/TextField.tsx
Korri: I don't think there is a "safe" default for input type `number` as even the broader `decimal` compared to `numeric`  mentions "Devices may or may not show a minus key (-)." 😬   [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode).

src/components/TextField/README.md
Korri: I feel like this should more explicitly state that it is just passing the property though, and maybe link to the doc about this property on [developer.mozilla.org](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode) for example.

Tom-Bonnike: Didn’t see this done for the other attributes but yeah, makes total sense, I’ll do that

src/components/TextField/TextField.tsx
Tom-Bonnike: Oh damn.

src/components/TextField/README.md
Tom-Bonnike: Done!

Korri (approved):

chloerice (approved):

PR #3238, opened 2020-09-11
[Filters] Fix spacing issues with filters button group

### LINKED ISSUE #3163: [new DL] Filters "More filters" button lacks spacing
## Issue summary

In the new DL at small screen sizing the "more filters button" should have some spacing, but it does not

Screen Shot 2020-08-10 at 2 53 11 PM

It looks like the root of this issue is in `ConnectedFilterControl` - a subcomponent of `Filters`.

## Expected behavior

A space between the filter box and more filters button as "connected" items are no longer visually connected in the new DL

## Actual behavior

The filterbox and more filters button run into each other

## Reduced test case

Visit https://polaris-react.herokuapp.com/?path=/story/all-components-filters--some-filters-disabled, enable the new DL and resize your browser so the filers are replaced with the "More filters" button

ashwinnarayanan2001:
# WHY are these changes introduced?

Fixes #3163  [https://github.com/Shopify/polaris-react/issues/3163]

Fixes the issue of the filter customers textfield and more filters button when the screen is small. 

Before
Screen Shot 2020-09-11 at 11 46 29 AM

After
Screen Shot 2020-09-14 at 2 21 59 PM

Screen Shot 2020-09-14 at 2 22 12 PM

### 🎩 checklist

### DIFF
--- a/src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
+++ b/src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
@@ -57,8 +57,8 @@ $stacking-order: (
 }
 
 .newDesignLanguage {
-  .CenterContainer + .RightContainer {
-    margin-left: spacing(extra-tight);
+  .CenterContainer {
+    margin-right: spacing(extra-tight);
   }
 }
 
--- a/src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
+++ b/src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
@@ -58,7 +58,7 @@ $stacking-order: (
 
 .newDesignLanguage {
   .CenterContainer {
-    margin-right: spacing(extra-tight);
+    margin-right: spacing(tight);
   }
 }
 
--- a/src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
+++ b/src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
@@ -57,8 +57,9 @@ $stacking-order: (
 }
 
 .newDesignLanguage {
-  .CenterContainer {
-    margin-right: spacing(tight);
+  .CenterContainer + .RightContainer,
+  .CenterContainer + .MoreFiltersButtonContainer {
+    margin-left: spacing(extra-tight) / 2;
   }
 }
 


src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
```diff
@@ -57,8 +57,8 @@ $stacking-order: (
 }
 
 .newDesignLanguage {
-  .CenterContainer + .RightContainer {
```
kyledurand, line 60: I have a feeling we'll still need this

src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
```diff
@@ -57,8 +57,8 @@ $stacking-order: (
...
-    margin-left: spacing(extra-tight);
+  .CenterContainer {
+    margin-right: spacing(extra-tight);
```
kyledurand, line 61: Looks like this should be using spacing(tight) or spacing() to equal the space between filters and the save button

kyledurand (changes requested):
Close but I think this would benefit from a bit more spacing to match the space between filters and the save action.

I'll pull this down and take a closer look

src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
ashwinnarayanan2001: The approach I took was applying the margin-right to the CenterContainer, so that we can treat all of the other buttons as a group (no matter the size of the browser, there will always be a spacing to the button groups and the centerContainer). The sizing issue with the More Filters button occurred because when the size is reduced, the moreFilters is apart of its own container (the margin-left was originally only applied to the right container).

src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
ashwinnarayanan2001: I think that's better as well!

src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
kyledurand: Right but if there are no filters buttons then the search bar wouldn't span the full width of its container. I think it's safer to use the `+` selector and apply a left margin when they exist

kyledurand:
I'm wondering if we want to actually tighten the space between the search field and filter buttons to show their relationship (also matches the connected field spacing)

image

image

@jessebc Could you lend us your design eyes here?

jessebc:
Yes!

The reason they are shown differently in Figma today is because Figma doesn't do negative auto-layout spacing. Segmented buttons should look like this. 

Screen Shot 2020-09-11 at 2 40 27 PM

I think we can want to stop connecting allll the fields though. 

So instead of this:

Screen Shot 2020-09-11 at 2 41 48 PM

we would do something like this:

Screen Shot 2020-09-11 at 2 41 36 PM

alex-page:
I am creating an issue https://github.com/Shopify/polaris-react/issues/3243 to fix up segmented buttons, because this will be a bigger change.

Lets get this fix in to create the gap between the search and buttons.

src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
alex-page: We could also do `only-child` to revert the `margin-right`.

kyledurand (approved):
Looks good! 👍

alex-page (approved):
Nice work Ashwin!

PR #3263, opened 2020-09-16
[ButtonGroup] Fix Segmented Button Gap Fix

### LINKED ISSUE #3243: [new DL] Segmented buttons should not have a gap
> The reason they are shown differently in Figma today is because Figma doesn't do negative auto-layout spacing. Segmented buttons should look like this.
> 
> Screen Shot 2020-09-11 at 2 40 27 PM
> 
> I think we can want to stop connecting allll the fields though.
> 
> So instead of this:
> 
> Screen Shot 2020-09-11 at 2 41 48 PM
> 
> we would do something like this:
> 
> Screen Shot 2020-09-11 at 2 41 36 PM

ashwinnarayanan2001:
Fixes https://github.com/Shopify/polaris-react/issues/3243

This PR is fixing the gap between segmented buttons.

After 

Button group with segmented buttons
Screen Shot 2020-09-16 at 3 10 27 PM

Filters (Some filters disabled)
Screen Shot 2020-09-16 at 3 11 16 PM

## 

### How to 🎩

</details>

### 🎩 checklist

### DIFF
--- a/src/utilities/theme/tokens.ts
+++ b/src/utilities/theme/tokens.ts
@@ -38,7 +38,7 @@ export const Tokens = {
   textFieldSpinnerOffset: rem('2px'),
   textFieldFocusRingOffset: rem('-4px'),
   textFieldFocusRingBorderRadius: rem('7px'),
-  buttonGroupItemSpacing: rem('2px'),
+  buttonGroupItemSpacing: rem('-1px'),
   contextualSaveBarHeight: '64px',
   duration100: '100ms',
   duration150: '150ms',
--- a/documentation/Color system.md
+++ b/documentation/Color system.md
@@ -271,7 +271,7 @@ Used to decorate elements where color does convey a specific meaning in componen
 | `--p-text-field-spinner-offset`           | `0.2rem`                                                                                                     |
 | `--p-text-field-focus-ring-offset`        | `-0.4rem`                                                                                                    |
 | `--p-text-field-focus-ring-border-radius` | `0.7rem`                                                                                                     |
-| `--p-button-group-item-spacing`           | `0.2rem`                                                                                                     |
+| `--p-button-group-item-spacing`           | `-0.1rem`                                                                                                    |
 | `--p-contextual-save-bar-height`          | `64px`                                                                                                       |
 | `--p-duration-1-0-0`                      | `100ms`                                                                                                      |
 | `--p-duration-1-5-0`                      | `150ms`                                                                                                      |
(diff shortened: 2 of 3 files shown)


src/utilities/theme/tokens.ts
```diff
@@ -38,7 +38,7 @@ export const Tokens = {
   textFieldSpinnerOffset: rem('2px'),
   textFieldFocusRingOffset: rem('-4px'),
   textFieldFocusRingBorderRadius: rem('7px'),
-  buttonGroupItemSpacing: rem('2px'),
+  buttonGroupItemSpacing: rem('-1px'),
```
alex-page, line 41: As we are no longer using the `--p-button-group-item-spacing` variable maybe we should remove it?

kyledurand: Oh we were when we were pairing. Ashwin it looks like the connected filters changes didn't get pushed up

kyledurand:
One thing I noticed when we were working on this is that the focus ring is now getting clipped when tabbing through connected items. 

[Here's a screencast at 25% speed](https://screenshot.click/screencast_2020-09-16_15-18-49.mp4)

You can see the left border of the focus ring getting clipped as focus shifts right, and vice versa when shifting left

[Here's a screencast at 100% speed](https://screenshot.click/screencast_2020-09-16_15-19-11.mp4)

Still pretty noticeable. We might be able to do some funky things with z-index when connected items are focused to fix this. I'll explore later today or tomorrow

src/utilities/theme/tokens.ts
ashwinnarayanan2001: There were no changes for the ConnectedFilterControl component. The --p-button-group-item-spacing variable is still being used in the component. It is still necessary because it's used as a fallback (without the new design language).

alex-page: Thanks for the context!

alex-page:
@ashwinnarayanan2001 this also needs to be in `polaris-rails`. If you want to ship this PR and close https://github.com/Shopify/polaris-react/issues/3243 that is all good. Just make sure an issue is made for tracking this work in rails 🚀

kyledurand (approved):
I just checked the focus states in the old dl and the issue is happening [there](https://screenshot.click/screencast_2020-09-21_13-54-10.mp4) as well.

I think this is fine to ship as is but we should create an issue to follow up with a fix since it's more noticeable in the new design language.

PR #3268, opened 2020-09-17
Move discard modal out of theme provider

### LINKED ISSUE #3219: [New DL] Modal opened by contextual save bar is using inverted palette
## Issue summary

A user opens a confirmation modal from the contextual save bar the colors of the modal are inverted.

## Expected behavior

When a user gets a confirmation modal from the contextual save bar the color should be not inverted.

## Actual behavior

image (1)

dfmcphee:

This fixes #3219 

It moves the discard modal out of the `ContextualSaveBar`'s theme provider so that it doesn't inherit the same dark theme.

This is what it looks like after this change:

Screen Shot 2020-09-17 at 4 23 44 PM

## 

### How to 🎩

Use the details page view in storybook and make a change to the title and then try to discard with the context bar

### DIFF
--- a/playground/DetailsPage.tsx
+++ b/playground/DetailsPage.tsx
@@ -143,6 +143,7 @@ export function DetailsPage() {
       }}
       discardAction={{
         onAction: handleDiscard,
+        discardConfirmationModal: true,
       }}
     />
   ) : null;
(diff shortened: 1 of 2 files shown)


kyledurand (approved):
Good find! I actually didn't know we had a discard confirmation modal in Polaris

alex-page (approved):

PR #3273, opened 2020-09-18
[Frame] Allow for custom navigation width

### LINKED ISSUE #3212: Allow for custom navigation width
## Feature request summary
Allow consumers to define a custom navigation width. Consumers can already pass the Frame a custom navigation component, yet the Frame is opinionated on this component's width.

## Rationale

The Vault uses a custom nav component, which is much narrower (89px) than the [default nav (240px)](https://github.com/Shopify/polaris-react/blob/master/src/styles/foundation/_layout.scss#L1). 

Up until Polaris v5, we managed to [override the `$navigation-width` variable](https://github.com/Shopify/u2/blob/master/app/ui/styles/_polaris-overrides.scss) and the Frame would use our custom value to define the content width.

I can't exactly pinpoint how or why, but now that the global styles are provided by the `AppProvider` component, the override is not taking effect anymore.

Recognizing that what we were doing was a hack and seeing that `AppProvider` already takes size-related theme options (e.g. `frameOffset`), my proposal is to allow for an optional `navigation` object, which would take a `width` value that can then be used to replace [`layout-width(nav)`](https://github.com/Shopify/polaris-react/blob/master/src/components/Frame/Frame.scss#L168).

**Example usage:**

image

--------

I'm not entirely sure this approach would work, but I'm happy to set up a proof of concept and work on the feature if you're aligned. I'd also love to hear alternative approaches! 🙏

tatianau:

Fixes #3212

Allowing consumers to define a custom navigation width. Consumers can already pass to the Frame a custom navigation component, yet the Frame is opinionated on this component's width.

Screen Shot 2020-09-18 at 11 24 55 AM

This adds a `navWidth` property to the theme provider which 
- generates a custom property
- which is used to make the existing `$navigation-width` dynamic based on the provided width
- existing `$navigation-width` plays the same role as currently: pushes the frame's main content and its sub-components to the right by the pixel value provided

#### Before
Fixed navigation width

#### After
Allowing for custom navigation width

## 

### Another approach I explored
<details>
<summary>Context</summary>

@alex-page [proposed](https://shopify.slack.com/archives/CQK6FNLJK/p1599074350061600?thread_ts=1599074063.061200&cid=CQK6FNLJK) to slightly change direction by **removing existing Frame navigation styles and making the navigation be a 100% width of its parent container**.

🏗 [Branch with exploration](https://github.com/Shopify/polaris-react/compare/3212-allow-for-custom-nav-width)

After exploring this option, even though this approach is how the layout ideally should be structured, **I'm not opening PR with these changes for the following reasons**:
- ❌ Making this happen requires removing fixed position from the nav which changes the document flow 
- ❌ Because of that, it’s a breaking change, and it can potentially negatively impact consumers
	- Consumers will need to add a wrapper to the nav with their own custom width
	- Changes in the document flow might break the layout of the pages that are currently relying on the fixed position of the nav being removed from regular flow. 
	- One example of this could be Vault legacy rails pages that worked with the fixed position but to make them work with a different position will require foundational structure changes. I can imagine other consumers might run into this problem too.
	- Changing the nav width while also keeping the nav ‘fixed’ on the page as user scrolls can be achieved with `position: sticky` support of which is [not the highest](https://caniuse.com/?search=position%20sticky)
- ❌ It’s too foundational change for us to make
	- It’s a foundational change in the Polaris layout, and we think we’d need support from the Polaris team to ensure we’re not only delivering the best long-term solution, but also that this approach works for most consumers

🔗 [Slack thread](https://shopify.slack.com/archives/CQK6FNLJK/p1599074063061200) for more context

</details>

### How to 🎩

</details>

### 🎩 checklist

### DIFF
--- a/src/styles/foundation/_layout.scss
+++ b/src/styles/foundation/_layout.scss
@@ -1,4 +1,4 @@
-$navigation-width: 240px !default;
+$navigation-width: calc(var(--p-nav-width) / 10 * 1rem) !default;
 
 ////
 /// Layout
@@ -21,7 +21,7 @@ $layout-width-data: (
     base: rem(240px),
   ),
   nav: (
-    base: rem($navigation-width),
+    base: $navigation-width,
   ),
   page-with-nav: (
     base: rem(769px),
--- a/src/styles/shared/_breakpoints.scss
+++ b/src/styles/shared/_breakpoints.scss
@@ -1,6 +1,6 @@
 $page-max-width: layout-width(primary, max) + layout-width(secondary, max) +
   layout-width(inner-spacing);
-$frame-with-nav-max-width: layout-width(nav) + $page-max-width;
+$frame-with-nav-max-width: calc(var(--p-nav-width) + $page-max-width);
 
 $stacked-content: em(
   layout-width(primary, min) + layout-width(secondary, min) +
@@ -18,7 +18,7 @@ $not-condensed-min-page: $not-condensed-content + $not-condensed-outer-spacing;
 $partially-condensed-min-page: $partially-condensed-content +
   $partially-condensed-outer-spacing;
 
-$nav-size: em(layout-width(nav));
+$nav-size: calc(var(--p-nav-width) / 16 * 1em);
 $nav-min-window: em(layout-width(page-with-nav));
 
 @function breakpoint($value, $adjustment: 0) {
--- a/src/utilities/theme/utils.ts
+++ b/src/utilities/theme/utils.ts
@@ -26,7 +26,13 @@ export function buildCustomProperties(
   newDesignLanguage: boolean,
   tokens?: Record<string, string>,
 ): CustomPropertiesLike {
-  const {colors = {}, colorScheme, config, frameOffset = 0} = themeConfig;
+  const {
+    colors = {},
+    colorScheme,
+    config,
+    frameOffset = 0,
+    navWidth = 240,
+  } = themeConfig;
   const mergedConfig = mergeConfigs(base, config || {});
 
   return newDesignLanguage
@@ -34,10 +40,14 @@ export function buildCustomProperties(
         ...colorFactory(colors, colorScheme, mergedConfig),
         ...tokens,
         frameOffset: `${frameOffset}px`,
+        navWidth: `${navWidth}`,
       })
     : {
         ...buildLegacyColors(themeConfig),
-        ...customPropertyTransformer({frameOffset: `${frameOffset}px`}),
+        ...customPropertyTransformer({
+          frameOffset: `${frameOffset}px`,
+          navWidth: `${navWidth}`,
+        }),
       };
 }
 
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -8,6 +8,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Added a theme prop for `navWidth` ([#3273](https://github.com/Shopify/polaris-react/pull/3273))
+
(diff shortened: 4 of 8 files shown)


src/styles/foundation/_layout.scss
```diff
@@ -1,4 +1,4 @@
-$navigation-width: 240px !default;
+$navigation-width: calc(var(--p-nav-width) / 10 * 1rem) !default;
```
tatianau, line 1: I'm not a fan of this code but I think this is the only way to convert it to rems - please correct me if I'm wrong 🙏

Custom properties work at runtime, so using [existing sass utilities](https://github.com/Shopify/polaris-react/blob/master/src/styles/foundation/_utilities.scss#L8) to convert the value to rems like this `rem(var(--p-nav-width))` isn't possible because at the time sass compiles to css `var(--p-nav-width)` doesn't actually hold a value to make maipulations and unit conversions with.

src/styles/shared/_breakpoints.scss
```diff
@@ -18,7 +18,7 @@ $not-condensed-min-page: $not-condensed-content + $not-condensed-outer-spacing;
 $partially-condensed-min-page: $partially-condensed-content +
   $partially-condensed-outer-spacing;
 
-$nav-size: em(layout-width(nav));
+$nav-size: calc(var(--p-nav-width) / 16 * 1em);
```
tatianau, line 21: For the [same reason](https://github.com/Shopify/polaris-react/pull/3273/files#r491062443), I'm doing it here.

src/utilities/theme/utils.ts
```diff
@@ -26,18 +26,28 @@ export function buildCustomProperties(
   newDesignLanguage: boolean,
   tokens?: Record<string, string>,
 ): CustomPropertiesLike {
-  const {colors = {}, colorScheme, config, frameOffset = 0} = themeConfig;
+  const {
+    colors = {},
+    colorScheme,
+    config,
+    frameOffset = 0,
+    navWidth = 240,
```
tatianau, line 34: Default `240` plaris nav value [`$navigation-width`](https://github.com/Shopify/polaris-react/pull/3273/files#diff-9085450f12ece7394e7c8d60ccced256R1) used to hold.

src/styles/foundation/_layout.scss
dfmcphee: Although this will work here, doesn't it break the [calculations](https://github.com/Shopify/polaris-react/blob/688d6b9cecb17eea66277e22d6c75d3699c4d7ad/src/styles/shared/_breakpoints.scss#L58) in the media query helpers below?

I think this has been the major blocker for us trying to do this previously. Since those breakpoint helpers are used in Sass throughout different codebases, we can't move this to a runtime calc without breaking those.

BPScott: Yep, changing these from raw values to properties is a breaking change. [See this other PR that wanted to do a similar thing for reasons](https://github.com/Shopify/polaris-react/pull/3214#issuecomment-686611525)

tatianau: Thank you so much, appreciate the context! 🙏 

---

Would you consider an [alternative solution](https://github.com/Shopify/polaris-react/compare/3212-allow-for-custom-nav-width) to the problem? 

It's removing existing Frame navigation styles and making the navigation be a 100% width of its parent container. It also has some tradeoffs. It's also breaking because consumers would need to add a wrapper with their own width but maybe it could work?

```css
/* consumer's wrapper around navigation */
.Wrapper {
  @include breakpoint-after(nav-min-window-corrected()) {
    width: rem(100px);
  }
}
```

PR #3284, opened 2020-09-21
Reduce CI a11y test time by using Storybook provided axe runner

BPScott:
## Why

A11y tests used to take ~12min. Before this PR they took ~25mins. With this PR they're back down to ~13mins.

Replacing pally with using the axe runner that is already loaded within Storybook's a11y addon serves two purposes:

- Speeds up our tests
- Our CI tests now reflect the same errors you see in Storybook's a11y panel, leading to easier debugging. As a bonus it seems to catch more issues than our previous accessibility testing.

# What

Replace pa11y with running the axe runner that Storybook provides.

# Bonus

This PR previously included several changes that fixed the newly raised issues. They were removed in https://github.com/Shopify/polaris-react/pull/3284/commits/9a3d01400af174fc790c7ebee7e67f9c749c7129. They can be addressed in separate follow-up PRs.


amrocha:
Interesting that a lot of the slowdown was due to storybook. I thought there was something weird when I tried running the tests locally and it took over 20 minutes, but was confused by the history since there didn't seem to be a smoking gun, and since the slowdown coincided with a move to Travis's free plan that was my first guess.

I like the idea of using the all components pages to test here. As long as the error reporting is accurate, I think we should go for that.

amrocha:
We should also add timeouts to these jobs. If we had a timeout then this would never have happened in the first place.

BPScott:
I did some pairing with @dominic earlier around a11y stuff. 
The axe stuff is ace and it catches issues that pally didn't and it lining up with what storybook reports already is great, however I don't really want this to bloom into "and now we have to fix these issues" or have to add ignorelists to get green builds.

I'm thinking incremental improvements are still good. I'm gonna open a PR that disables docs (which gets us half-way to our old speed) which we can merge right away. Then have a follow up with a PR to move to axe, @dfmcphee is going to look at fixing the issues that axe raises, and once those are fixed we can merge the axe PR which gets us the rest of the way there.

BPScott:
Have rebased this atop master so now it's a lot smaller. Changes have been split out and merged elsewhere:

Already merged:
- Storybook update - #3305
- Disabling storybook docs when doing the a11y tests - #3292 and #3304
- Fixing axe errors on due to poor contrast on AppProvider, Frame and TopBar pages - #3293

Pending:

- Fixing axe errors in Datepicker due to bad role configuration - #3303

BPScott:
Had a bit of slack time so I've came back to this.

Have build a basic "here's a list of expected errors". It doesn't list out the expected errors inline but it tells you the page, and you can take that URL and go to the storybook and that will tell you error in the ally panel.

dfmcphee:
I reverted all the component changes here. We can follow up with fixes separately.

BPScott:
haha it works!

On the other hand 68 existing errors awwww maaaaan :(

alex-page (approved):
This is a great first step, I think we can clean up logging and exceptions in follow ups.

PR #3302, opened 2020-09-23
Fix focus state for ResourceItem

alex-page:
# [View the changes](https://5d559397bae39100201eedc1-kixnhcavmb.chromatic.com/iframe.html?id=all-components-resource-list--all-examples&contexts=New%20Design%20Language%3DEnabled%20-%20Light%20Mode)

Fixes https://github.com/Shopify/polaris-react/issues/3257

focus states

### How to 🎩

[View the changes](https://5d559397bae39100201eedc1-kixnhcavmb.chromatic.com/iframe.html?id=all-components-resource-list--all-examples&contexts=New%20Design%20Language%3DEnabled%20-%20Light%20Mode)

### 🎩 checklist


alex-page:
@kyledurand @AndrewMusgrave any ideas why this isn't showing? I thought it might be related to `z-index` or a parent with `overflow: hidden` but I can't seem to find the spot.

alex-page:
Known issue with this PR, going to make a follow up to fix it:

Screen Shot 2020-09-23 at 4 29 14 PM

src/components/ResourceItem/ResourceItem.scss
```diff
@@ -318,19 +323,6 @@ $resource-list-item-variables: (
...
-
-  // stylelint-disable selector-max-specificity
-  &.focused:not(:active):not(.focusedInner) {
-    @include breakpoint-after(resource-list-item(breakpoint-small)) {
-      @include focus-ring($style: 'focused');
-    }
-
-    @include breakpoint-before(resource-list-item(breakpoint-small), false) {
-      box-shadow: inset 0.2rem 0 0 var(--p-focused);
-    }
```
AndrewMusgrave, line 331: Did the ndl mocks remove this? iirc we wanted to keep the left onset border on small screens

AndrewMusgrave (approved):
In addition to the bug you mentioned, it's also going to break against the edges of the screen

src/components/ResourceItem/ResourceItem.scss
alex-page: That makes sense, let me make sure this is good.

kyledurand (approved):
For sure an improvement. I think we can improve on this further but this should be fine for launch

alex-page:
Remove the gap on the box shadow and inset the top line of the first items - @jessebc

src/components/ResourceList/ResourceList.tsx
```diff
@@ -400,11 +400,7 @@ export const ResourceList: ResourceListType = function ResourceList<ItemType>({
   const renderItemWithId = (item: ItemType, index: number) => {
     const id = idForItem(item, index);
 
-    return (
-      <li key={id} className={styles.ItemWrapper}>
-        {renderItem(item, id, index)}
-      </li>
-    );
+    return renderItem(item, id, index);
   };
```
alex-page, line 404: Interested to talk about this. It could be really dangerous for accessibility, but I feel like removing the wrapper from the parent component enables the focus `z-index` to be managed without passing state between components.

kyledurand: Dangerous as in if someone uses the resource list but not the resource list item and rolls their own?

alex-page: Yeah exactly, any custom `ResourceListItem` needs to be wrapped in a `li`.

src/components/ResourceItem/ResourceItem.scss
```diff
@@ -343,3 +335,21 @@ $resource-list-item-variables: (
...
+    @include focus-ring($style: 'focused');
+    z-index: resource-list-item(clickable-stacking-order);
+  }
+
+  & + .ListItem {
+    border-top: border('divider');
+  }
+
+  * + ul > &:first-of-type.focused:after {
+    top: 0.1rem;
```
kyledurand, line 353:
```suggestion
    top: rem(1px);
```

But is it safer to avoid weirdness if we remove `position: relative` and just use margin or padding?

src/components/ResourceItem/ResourceItem.scss
```diff
@@ -343,3 +335,21 @@ $resource-list-item-variables: (
     background-color: var(--p-surface-selected-pressed);
   }
 }
+
+.ListItem {
```
kyledurand, line 339: Does this not need to be nested under `.newDesignLanguage`?

src/components/ResourceItem/ResourceItem.scss
```diff
@@ -343,3 +335,21 @@ $resource-list-item-variables: (
...
+  position: relative;
+  @include focus-ring($border-width: rem(0px), $size: 'wide');
```
kyledurand, line 341:
```suggestion
  @include focus-ring($size: 'wide');
```

Shouldn't need to provide the border width because the default is rem(0px)

kyledurand (commented):
🎩 looks great, just a few questions around styling

src/components/ResourceItem/ResourceItem.scss
alex-page: I experimented with that and didn't get the same result as I expected. I can try again.

src/components/ResourceList/ResourceList.tsx
dfmcphee: I think this will be ok for most cases, but I do worry a bit about it too. Even if someone uses the `ResourceItem` but wraps it in a div or other non list element, they would break the structure of the list. Doing a quick poke around in web I don't see anyone doing that though. Do we always recommend to only render a `ResourceItem` from renderItem?

alex-page: Looking at the example files we do that ^

dfmcphee: What about if we thre a warning if the returned component isn't a ResourceItem?

alex-page: Yeah that would make me feel a little bit better. Do we have any examples of this?

dfmcphee: We have a utility function `isElementOfType` that could be used to check against.

jessebc:
I think we should keep the border-radius to 2px on the focus ring. 

Screen Shot 2020-09-28 at 12 52 34 PM

src/components/ResourceList/ResourceList.tsx
```diff
@@ -400,11 +401,18 @@ export const ResourceList: ResourceListType = function ResourceList<ItemType>({
...
+    const itemContent = renderItem(item, id, index);
+
+    if (
+      process.env.NODE_ENV === 'development' &&
+      !isElementOfType(itemContent, ResourceItem)
+    ) {
+      console.warn(
+        '<ResourceList /> renderItem function should return a <ResourceItem />.',
+      );
+    }
```
alex-page, line 413: Prints a warning if the child does not return a `ResourceItem`. I wonder if we should move this outside and just check the first item and not in the loop. It only runs in development.

src/components/ResourceList/tests/ResourceList.test.tsx
```diff
@@ -58,13 +58,20 @@ describe('<ResourceList />', () => {
...
+      process.env.NODE_ENV = 'development';
+      const warningSpy = jest
+        .spyOn(console, 'warn')
+        .mockImplementation(() => {});
       const resourceList = mountWithAppProvider(
         <ResourceList items={itemsWithID} renderItem={renderCustomMarkup} />,
       );
-      expect(resourceList.find('li').first().children().html()).toBe(
-        '<p>title 1</p>',
+      expect(resourceList.find('li').first().text()).toBe('title 1');
```
alex-page, line 69:
```suggestion
      expect(resourceList.find('li').first().children().html()).toBe(
        '<p>title 1</p>',
```

src/components/ResourceList/ResourceList.tsx
alex-page: @dfmcphee @kyledurand I have implemented the warning would appreciate some fresh eyes 👁️

dfmcphee (approved):
This looks good! 👍

kyledurand (approved):
This looks amazing 🚢

PR #3304, opened 2020-09-23
Only disable Storybook docs when running a11y tests

BPScott:

@amrocha had good idea: (paraphrasing) "Docs are still useful, what if we only disable them during the a11y test?"

I keep fogetting that SB has decent support for environment variables.

Only disable docs when running a11y tests

### How to 🎩

- See that a11y test speed does not regress back down to ~45mins
- See that opening storybook normally / in chromatic you still have access to the docs panel

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -8,10 +8,16 @@ const postcssShopify = require('@shopify/postcss-plugin');
 // eslint-disable-next-line node/no-extraneous-require, import/no-extraneous-dependencies
 const {BundleAnalyzerPlugin} = require('webpack-bundle-analyzer');
 
+// Enabling docs means the preview panel takes an extra 2ish seconds to load
+// This usually isn't a big deal, except when we're running all of our stories
+// through our a11y tests, and a 2s delay over several hundred stories adds up.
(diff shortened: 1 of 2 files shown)


.storybook/main.js
```diff
@@ -8,10 +8,17 @@ const postcssShopify = require('@shopify/postcss-plugin');
 // eslint-disable-next-line node/no-extraneous-require, import/no-extraneous-dependencies
 const {BundleAnalyzerPlugin} = require('webpack-bundle-analyzer');
 
+// Enabling docs means the preview panel takes an extra 2ish seconds to load
+// This usually isn't a big deal, except when we're running all of our stories
+// through our a11y tests, and a 2s delay over several hundred stories adds up.
+// This is an escape hatch to disable docs only wheile we're running a11y tests
+const enableDocs = !parseInt(process.env.STORYBOOK_DISABLE_DOCS || '0', 10);
+console.log(process.env.STORYBOOK_DISABLE_DOCS, enableDocs);
```
amrocha, line 16: remove this log and we can merge

amrocha (approved):

PR #3308, opened 2020-09-24
[Page header] Single row layout

### LINKED ISSUE #3286: [New DL] Update page header layout to match Figma
[Figma](https://www.figma.com/file/DOLJoC3CbldApU26mIIDiK/Page-Component?node-id=471%3A0)

Rails | React
---|---
complete | ready for review

- [x] Vertically compress space
- [x] Rollup actions that do not fit in the given space

kyledurand:

Fixes #3286 

image
image
image
image
image

## 

### How to 🎩

Locally: http://localhost:6006/iframe.html?id=playground-playground--details-page&contexts=New%20Design%20Language%3DEnabled%20-%20Light%20Mode&viewMode=story

Chroma: https://5d559397bae39100201eedc1-qayohpojms.chromatic.com/

### 🎩 checklist

### DIFF
--- a/documentation/Color system.md
+++ b/documentation/Color system.md
@@ -242,6 +242,7 @@ Used to decorate elements where color does convey a specific meaning in componen
 | ----------------------------------------- | ------------------------------------------------------------------------------------------------------------ |
 | `--p-border-radius-base`                  | `0.4rem`                                                                                                     |
 | `--p-border-radius-wide`                  | `0.8rem`                                                                                                     |
+| `--p-border-radius-full`                  | `50%`                                                                                                        |
 | `--p-card-shadow`                         | `0px 0px 5px var(--p-shadow-from-ambient-light), 0px 1px 2px var(--p-shadow-from-direct-light)`              |
 | `--p-popover-shadow`                      | `-1px 0px 20px var(--p-shadow-from-ambient-light), 0px 1px 5px var(--p-shadow-from-direct-light)`            |
 | `--p-modal-shadow`                        | `0px 26px 80px var(--p-shadow-from-dim-light), 0px 0px 1px var(--p-shadow-from-dim-light)`                   |
--- a/src/components/ActionMenu/components/Actions/Actions.tsx
+++ b/src/components/ActionMenu/components/Actions/Actions.tsx
@@ -6,10 +6,10 @@ import type {
   MenuActionDescriptor,
   MenuGroupDescriptor,
 } from '../../../../types';
-import {Button} from '../../../Button';
 import {ButtonGroup} from '../../../ButtonGroup';
 import {MenuGroup} from '../MenuGroup';
 import {MenuAction} from '../MenuAction';
+import {SecondaryAction} from '../SecondaryAction';
 
 import styles from './Actions.scss';
 
@@ -58,9 +58,9 @@ export function Actions({actions = [], groups = []}: Props) {
 
     const {content, onAction, ...rest} = action;
     return newDesignLanguage ? (
-      <Button key={index} onClick={onAction} {...rest}>
+      <SecondaryAction key={index} onClick={onAction} {...rest}>
         {content}
-      </Button>
+      </SecondaryAction>
     ) : (
       <MenuAction
         key={`MenuAction-${index}`}
--- a/src/components/ActionMenu/components/Actions/Actions.tsx
+++ b/src/components/ActionMenu/components/Actions/Actions.tsx
@@ -1,4 +1,4 @@
-import React, {useCallback, useState} from 'react';
+import React, {useEffect, useCallback, useRef, useState} from 'react';
 
 import {sortAndOverrideActionOrder} from '../../utilities';
 import {useFeatures} from '../../../../utilities/features';
@@ -20,21 +20,68 @@ interface Props {
   groups?: MenuGroupDescriptor[];
 }
 
+interface DividedActions {
+  secondaryActions: MenuActionDescriptor[];
+  menuGroups: MenuGroupDescriptor[];
+}
+
 export function Actions({actions = [], groups = []}: Props) {
   const {newDesignLanguage} = useFeatures();
+  const actionsLayoutRef = useRef<HTMLDivElement>(null);
+  const menuGroupRef = useRef<HTMLDivElement>(null);
+  const [actionWidths, setActionWidths] = useState<number[]>([]);
   const [activeMenuGroup, setActiveMenuGroup] = useState<string | undefined>(
     undefined,
   );
-
+  const [
+    showableActions,
+    setShowableActions,
+  ] = useState<DividedActions | null>();
+  const handleOffsetWidth = useCallback(
+    (width: number) =>
+      setActionWidths((actionWidths) => [...actionWidths, width]),
+    [],
+  );
   const handleMenuGroupToggle = useCallback(
     (group: string) => setActiveMenuGroup(activeMenuGroup ? undefined : group),
     [activeMenuGroup],
   );
 
+  useEffect(() => {
+    // We may not want this check 🤔
+    if (actionWidths.length === 0) return;
+    const overriddenActions = sortAndOverrideActionOrder([
+      ...actions,
+      ...groups,
+    ]);
+    setShowableActions([{content: 'sup'}]);
+    console.log({overriddenActions, showableActions});
+    const showable = [];
+    const additional = [];
+    overriddenActions.forEach((action, index) => {
+      if ('title' in action) return null;
+
+      if (index === 0) {
+        showable.push(action);
+      }
+      if (index === 1) {
+        additional.push(action);
+      }
+      console.log('hide / show stuff');
+      console.log({actionWidths, showable, additional});
+      console.log(actionsLayoutRef?.current?.offsetWidth);
+      console.log(menuGroupRef?.current?.offsetWidth);
+      // if title push to menu group
+      // if action and fittable push to secondary
+      console.log({action});
+    });
+  }, [actions, actionsLayoutRef, actionWidths, groups, menuGroupRef]);
+
   const handleMenuGroupClose = useCallback(
     () => setActiveMenuGroup(undefined),
     [],
   );
+
   const menuActions = [...actions, ...groups];
 
   const overriddenActions = sortAndOverrideActionOrder(menuActions);
@@ -44,21 +91,27 @@ export function Actions({actions = [], groups = []}: Props) {
       const {title, actions, ...rest} = action;
 
       return actions.length > 0 ? (
-        <MenuGroup
-          key={`MenuGroup-${index}`}
-          title={title}
-          active={title === activeMenuGroup}
-          actions={actions}
-          {...rest}
-          onOpen={handleMenuGroupToggle}
-          onClose={handleMenuGroupClose}
-        />
+        <div key={`MenuGroup-${index}`} ref={menuGroupRef}>
+          <MenuGroup
+            title={title}
+            active={title === activeMenuGroup}
+            actions={actions}
+            {...rest}
+            onOpen={handleMenuGroupToggle}
+            onClose={handleMenuGroupClose}
+          />
+        </div>
       ) : null;
     }
 
     const {content, onAction, ...rest} = action;
     return newDesignLanguage ? (
-      <SecondaryAction key={index} onClick={onAction} {...rest}>
+      <SecondaryAction
+        key={index}
+        onClick={onAction}
+        {...rest}
+        getOffsetWidth={handleOffsetWidth}
+      >
         {content}
       </SecondaryAction>
     ) : (
@@ -72,9 +125,16 @@ export function Actions({actions = [], groups = []}: Props) {
   });
 
   return (
-    <div className={styles.ActionsLayout}>
+    <div className={styles.ActionsLayout} ref={actionsLayoutRef}>
       {newDesignLanguage ? (
-        <ButtonGroup>{actionMarkup}</ButtonGroup>
+        <ButtonGroup noWrap={newDesignLanguage}>
+          {showableActions?.secondaryActions?.map((action) => (
+            <SecondaryAction key={action.content} {...action}>
+              {action.content}
+            </SecondaryAction>
+          ))}
+          {actionMarkup}
+        </ButtonGroup>
       ) : (
         actionMarkup
       )}
--- a/src/components/ActionMenu/components/Actions/Actions.tsx
+++ b/src/components/ActionMenu/components/Actions/Actions.tsx
@@ -103,8 +103,7 @@ export function Actions({actions = [], groups = []}: Props) {
         setHiddenActions((hiddenActions) => [...hiddenActions, action]);
       }
     });
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [availableWidth]);
+  }, [actionWidths, actions, availableWidth, newDesignLanguage]);
 
   const className = classNames(
     styles.ActionsLayout,
--- a/src/components/ActionMenu/components/Actions/Actions.tsx
+++ b/src/components/ActionMenu/components/Actions/Actions.tsx
@@ -1,12 +1,15 @@
 import React, {useEffect, useCallback, useRef, useState} from 'react';
+import debounce from 'lodash/debounce';
 
 import {sortAndOverrideActionOrder} from '../../utilities';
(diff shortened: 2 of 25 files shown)


documentation/Color system.md
```diff
@@ -242,6 +242,7 @@ Used to decorate elements where color does convey a specific meaning in componen
 | ----------------------------------------- | ------------------------------------------------------------------------------------------------------------ |
 | `--p-border-radius-base`                  | `0.4rem`                                                                                                     |
 | `--p-border-radius-wide`                  | `0.8rem`                                                                                                     |
+| `--p-border-radius-full`                  | `50%`                                                                                                        |
```
kyledurand, line 245: This is useful for icon only buttons
image

src/components/ActionMenu/components/Actions/Actions.tsx
```diff
@@ -35,32 +56,92 @@ export function Actions({actions = [], groups = []}: Props) {
...
+  useEffect(() => {
+    if (!menuGroupRef.current) return;
+    menuGroupWidthRef.current = menuGroupRef.current?.offsetWidth;
+  }, [menuGroupRef]);
+
+  useEffect(() => {
+    if (!newDesignLanguage || actionWidths.length === 0 || availableWidth === 0)
+      return;
+
+    let currentAvailableWidth = availableWidth;
```
kyledurand, line 87: Copy available width to we can manipulate it through the loop below and set the final available width after

src/components/ActionMenu/components/Actions/Actions.tsx
```diff
@@ -35,32 +56,92 @@ export function Actions({actions = [], groups = []}: Props) {
...
+    setShowableActions([]);
+    setHiddenActions([]);
```
kyledurand, line 89: Reset showable / hidden every time the available width changes. This is on first render and resize

src/components/ActionMenu/components/Actions/Actions.tsx
```diff
@@ -35,32 +56,92 @@ export function Actions({actions = [], groups = []}: Props) {
...
+
+    actions.forEach((action, index) => {
+      if (
+        actionWidths[index] + menuGroupWidthRef.current + ACTION_SPACING <=
+        currentAvailableWidth
+      ) {
+        currentAvailableWidth -= actionWidths[index];
```
kyledurand, line 96: Reduce current available width by action width

src/components/ActionMenu/components/Actions/Actions.tsx
```diff
@@ -35,32 +56,92 @@ export function Actions({actions = [], groups = []}: Props) {
...
+        setShowableActions((showableActions) => [
+          ...(showableActions || []),
+          action,
+        ]);
```
kyledurand, line 100: Merging showable actions

src/components/ActionMenu/components/Actions/Actions.tsx
```diff
@@ -35,32 +56,92 @@ export function Actions({actions = [], groups = []}: Props) {
...
+      } else {
+        currentAvailableWidth = 0;
```
kyledurand, line 102: I'm doing this because when there's a single item that can't fit we roll up the remaining items. If we don't set this to 0 then we might see actions jumping around in different positions on different viewport sizes. 

i.e. if there is an action in the array that is smaller than a previous action that couldn't fit, it may show up in its place

src/components/ActionMenu/components/Actions/Actions.tsx
```diff
@@ -35,32 +56,92 @@ export function Actions({actions = [], groups = []}: Props) {
...
+        setHiddenActions((hiddenActions) => [...hiddenActions, action]);
+      }
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [availableWidth]);
```
kyledurand, line 107: These calculations only run when the available width changes

src/components/ActionMenu/components/Actions/Actions.tsx
```diff
@@ -35,32 +56,92 @@ export function Actions({actions = [], groups = []}: Props) {
...
+
+  const className = classNames(
+    styles.ActionsLayout,
+    newDesignLanguage && styles.newDesignLanguage,
+  );
+
+  const menuActions = [...actions, ...groups];
   const overriddenActions = sortAndOverrideActionOrder(menuActions);
```
kyledurand, line 115: This is a good candidate for a refactor but I don't have all of the context on it. In the future I'd like to remove this function, and render actions and groups separately. As far as I can tell this function merges the actions and groups and make sure the items are rendered before the groups in proper sequence. It seems like a lot of added complexity for not much gain

src/components/ActionMenu/components/Actions/Actions.tsx
```diff
@@ -35,32 +56,92 @@ export function Actions({actions = [], groups = []}: Props) {
...
+            actions={[...hiddenActions, ...actions]}
+            {...rest}
+            onOpen={handleMenuGroupToggle}
+            onClose={handleMenuGroupClose}
+          />
+        </div>
       ) : null;
     }
-
+    if (showableActions && showableActions.length >= 0) return null;
```
kyledurand, line 134: If there are showable actions, we don't render this below but it's necessary to render them at first to get their widths

src/components/ActionMenu/components/Actions/Actions.tsx
```diff
@@ -72,12 +153,26 @@ export function Actions({actions = [], groups = []}: Props) {
   });
 
   return (
-    <div className={styles.ActionsLayout}>
+    <div className={className} ref={actionsLayoutRef}>
       {newDesignLanguage ? (
-        <ButtonGroup>{actionMarkup}</ButtonGroup>
+        <ButtonGroup segmented>
+          {showableActions && showableActions.length > 0
+            ? showableActions.map((action) => (
```
kyledurand, line 160: I meant to extract this out to a constant but I will do that in a follow up commit

src/components/ActionMenu/components/SecondaryAction/SecondaryAction.scss
```diff
@@ -0,0 +1,24 @@
+@import '../../../../styles/common';
+
+$breakpoint: 1080px;
+
+.SecondaryAction {
+  // stylelint-disable declaration-no-important
+  button {
```
kyledurand, line 7: This is the simplest way I could think of to hack buttons for this one component. An alternative would be to include the outline button and do a lot of overrides. I think this way is easier to comprehend but open to the alternative as well

src/components/ActionMenu/components/SecondaryAction/SecondaryAction.tsx
```diff
@@ -0,0 +1,33 @@
...
+  ...rest
+}: SecondaryAction) {
+  const {newDesignLanguage} = useFeatures();
+  const secondaryActionsRef = useRef<HTMLSpanElement>(null);
+
+  useEffect(() => {
+    if (!getOffsetWidth || !secondaryActionsRef.current || !newDesignLanguage)
+      return;
+
+    getOffsetWidth(secondaryActionsRef.current?.offsetWidth);
```
kyledurand, line 25: This passes the rendered offset width back to the actions component. Much safer than adding refs within a loop in my opinion

src/components/Page/components/Header/Header.scss
```diff
@@ -52,6 +52,15 @@ $action-menu-rollup-computed-width: rem(40px);
   flex: 0 0 auto;
   margin-left: auto;
   line-height: 1;
+
+  .newDesignLanguage & {
+    margin-left: spacing(extra-tight);
+    // stylelint-disable-next-line selector-max-combinators
+    button {
```
kyledurand, line 59: More button hacks to match pagination in figma

src/components/Page/components/Header/Header.scss
```diff
@@ -142,11 +155,46 @@ $action-menu-rollup-computed-width: rem(40px);
   display: flex;
   align-content: flex-start;
   align-items: center;
+
+  .newDesignLanguage & {
+    align-items: flex-start;
+    flex: 1 1 auto;
+
+    @include page-content-when-not-partially-condensed {
+      flex-basis: 50%;
```
kyledurand, line 164: Wouldn't mind pairing with someone on this to see if we can remove flex-basis altogether and make these containers a little more fluid

src/components/Page/components/Header/Header.tsx
```diff
@@ -349,10 +368,15 @@ function determineLayout({
...
+          <>
+            {actionMenuMarkup}
+            {primaryActionMarkup}
+          </>
+        ),
         slot4: paginationMarkup,
-        slot5: actionMenuMarkup,
-        slot6: primaryActionMarkup,
+        slot5: additionalMetaDataMarkup,
+        slot6: null,
```
kyledurand, line 379: I think it should be safe to remove slot6 now but that can be done in a follow up

src/components/ActionMenu/components/Actions/Actions.tsx
alex-page: Want to make an issue for this and put it into fast follows/v6? Would love to track this tech debt.

alex-page (approved):
This is looking good. Happy to merge this and iterate. I did notice two small issues.

When setting the page the `fullWidth` it felt like the title and badges where wrapping too early:
fullwidth-wrapping-too-early

The mobile layout breadcrumbs look like they could be moved to the left more.
Screen Shot 2020-09-29 at 10 59 45 AM

src/components/ActionMenu/components/Actions/Actions.tsx
kyledurand: Added it as part of https://github.com/Shopify/polaris-ux/issues/470

tleunen:
Would love to see these "borderless" buttons available for everybody, not only in this page header.

alex-page:
> Would love to see these "borderless" buttons available for everybody, not only in this page header.

Totally agree this might be a nice follow up in v6/fast follows. Feel free to add a card to that column in the backlog.

kyledurand:
> Would love to see these "borderless" buttons available for everybody, not only in this page header.

Was speaking with Jesse about this yesterday and we're noodling with the idea of having borders across all viewport sizes. We're not sure if borderless should be a pattern yet but if this pattern remains then we should for sure add the variation for v6

tleunen:
We have quite a few instances of borderless buttons in Email and Online Store, that's why I was asking ;)

alex-page (approved):
This is looking good to me. One minor thought is that the header on page should line up with the topbar icons. Maybe we can remove this padding?

Screen Shot 2020-09-30 at 1 38 48 PM

kyledurand:
@alex-page Part of that extra padding comes from story book. If you remove the 1rem padding on `<body>` it looks like this

image

We can definitely reduce the padding to have the page actions line up with avatar / menu actions but that might be out of scope for this pr. Can we add that as either a fast follow or lower priority change?

alex-page:
Can we add that as either a fast follow or lower priority change?

Yep! 🚢

src/components/ActionMenu/components/Actions/Actions.tsx
```diff
@@ -35,32 +56,92 @@ export function Actions({actions = [], groups = []}: Props) {
     () => setActiveMenuGroup(undefined),
     [],
   );
-  const menuActions = [...actions, ...groups];
 
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
dfmcphee, line 60: It would be better to avoid disabling this if we can. I imagine this is complaining because if the `setAvailableWidth` changed it would make sure this callback gets updated to include the new reference.

src/components/ActionMenu/components/Actions/Actions.tsx
```diff
@@ -35,32 +56,92 @@ export function Actions({actions = [], groups = []}: Props) {
...
+        setShowableActions((showableActions) => [
+          ...(showableActions || []),
+          action,
+        ]);
+      } else {
+        currentAvailableWidth = 0;
+        setHiddenActions((hiddenActions) => [...hiddenActions, action]);
+      }
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
dfmcphee, line 106: Same with this one. I know we only want it to run when the available width changes, but it also needs to be updated if any of the dependencies like `setHiddenActions` change.

src/components/ActionMenu/components/Actions/Actions.tsx
```diff
@@ -35,32 +56,92 @@ export function Actions({actions = [], groups = []}: Props) {
...
+        setAvailableWidth(actionsLayoutRef.current.offsetWidth);
+      },
+      20,
+      {leading: false, trailing: true, maxWait: 40},
+    ),
+    [actionsLayoutRef],
+  );
+
+  useEffect(() => {
+    if (!actionsLayoutRef.current) return;
```
dfmcphee, line 74: This is for future consideration but it would be nice to see these split into little hooks that you include to make this a bit more readable and easier to follow what all the effects are for.

dfmcphee (approved):
A few small comments on the code but 🎩 looks really good.

src/components/ActionMenu/components/Actions/Actions.tsx
kyledurand: This is because we're passing in debounce and useCallback can't determine the dependencies within. I don't think we can get around this or at least we weren't able to here as well: https://github.com/Shopify/polaris-react/blob/master/src/components/MediaQueryProvider/MediaQueryProvider.tsx#L19-L31

kyledurand: Would useCallBack work the same if we re-ordered them like

```tsx
  const handleResize = debounce(
    useCallback(() => {
      if (!newDesignLanguage || !actionsLayoutRef.current) return;
      setAvailableWidth(actionsLayoutRef.current.offsetWidth);
    }, [newDesignLanguage]),
    20,
    {leading: false, trailing: true, maxWait: 40},
  );
```

dfmcphee: Ah right, I remember this from our pairing now. I think it isn't quite the same with that because handleResize would still setup a new debounce on each render. Probably fine to ignore and keep your disabling 👍

tleunen: Hooks can only be inside hooks, or at the root of a component ;)

You can replace `useCallback` with `useMemo(() => debounce(...` and it should be able to determine the dependencies I believe.

src/components/ActionMenu/components/Actions/Actions.tsx
```diff
@@ -35,32 +56,92 @@ export function Actions({actions = [], groups = []}: Props) {
...
+    setAvailableWidth(actionsLayoutRef.current.offsetWidth);
+  }, [actionsLayoutRef]);
+
+  useEffect(() => {
+    if (!menuGroupRef.current) return;
+    menuGroupWidthRef.current = menuGroupRef.current?.offsetWidth;
+  }, [menuGroupRef]);
```
tleunen, line 81: A `ref` never changes, its content (`current`) changes, so you don't need to have it explicit.

src/components/ActionMenu/components/Actions/Actions.tsx
kyledurand: I see the same warning with use memo

I could just pass in a function...

```tsx
  const debouncedSetWidth = debounce(
    function () {
      if (!newDesignLanguage || !actionsLayoutRef.current) return;
      setAvailableWidth(actionsLayoutRef.current.offsetWidth);
    },
    20,
    {leading: false, trailing: true, maxWait: 40},
  );

  const handleResize = useCallback(debouncedSetWidth, [debouncedSetWidth]);
```

tleunen: Sorry I'm a bit late, but I wanted to clarify something :)
You still see a warning, but it's a different one.

Here's the warning you see with `useCallback` and debounce:

> React Hook useCallback received a function whose dependencies are unknown. Pass an inline function instead.eslintreact-hooks/exhaustive-deps

Here's the kind of warning you'll see with `useMemo` (if some deps are missing):

> React Hook useMemo has a missing dependency: 'newDesignLanguage'. Either include it or remove the dependency array.eslintreact-hooks/exhaustive-deps

`newDesignLanguage` is not really expected to change in the lifetime of the app, so it's "ok", but otherwise this wouldn't do what we want it to do imo ;)

image

PR #3310, opened 2020-09-24
Link: Add `rel` attribute for <Link /> component.

aud:
This allows users to pass in their own `rel` prop to the Link component,
instead of using the default when the external prop is provided (`noreferrer noopener`).

The default `rel` for links when the `external` prop is provided is `noreferrer noopener`. We have a case in Shopify/athena-flex where we want to allow the referrer to be passed through.

Allowing the `rel` attribute on links to be set through the `rel` prop on the Link component. If its unset, it'll default to the current state: `noreferrer noopener`, so backwards compatibility should be fine.

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -9,6 +9,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Enhancements
 
 - Hide `ActionMenu`, `Actions`, `RollupActions` menu popover overlays when printing ([#3277](https://github.com/Shopify/polaris-react/pull/3277))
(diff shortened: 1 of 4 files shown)


alex-page:
@aud as of now I am unsure why we would want this added flexibility. I am going to close the PR but feel free to reopen the PR and add context on what this would allow for your use case.

> We have a case in Shopify/athena-flex where we want to allow the referrer to be passed through.

Can you provide an example?

aud:
👍🏻 Yeah this is no longer relevant. IIRC the case was that we wanted to pass in a referrer on a link from Shopify/athena-flex which Minerva would validate the source of and save some contextual "access based" information. I believe we solved this another way in the end.

PR #3313, opened 2020-09-24
Add action list item prefix and fix content which should take as much as possible

tleunen:

This PR does 2 things:
- Makes the "content" takes all the space, so the suffix/badge will always go right aligned. This is because it is already the behavior when the item has an icon/image. So this makes it be consistent with or without icons/images.
- Adds a `prefix` prop, in the same way `suffix` works, allowing any React nodes. This PR doesn't remove any other prop such as icon or image, but they could also be removed in favor of `prefix`. Ideally, we would remove all these extra props in a major release (v6?) and keeps the component flexible with these 2 props: prefix/suffix.

Does it make sense? Anything missing?

image


BPScott (commented):
Not tested but approach seems sensible to me.

Add an example in ActionList's readme would be super helpful.

I'll defer to @alex-page / @kyledurand for UI thoughts

kyledurand (approved):
I think this makes sense in making the component more flexible 👍

We should remove the image prop for v6 and keep a note of these deprecations somewhere. @alex-page do we have a list going? Or should this just be added in the v6 column [here](https://github.com/orgs/Shopify/projects/779)

alex-page:
Yep, create an issue and add it to the column. Best place to track this now!

PR #3320, opened 2020-09-25
[WIP][PhoneField] Phone field component (New)

ashwinnarayanan2001:

The changes introduced is creating a PhoneField component that handles and formats international phone numbers. 

This pull request is creating the PhoneField component. This component does not have any autoformatting yet (will be added later on as a feature request). A decision made for this component is that it will handle cases that have an area code (e.g. '+16132558999' will be accepted and recognized as a Canadian phone number, NOT '6132558999'). Another feature request will be added later to handle this.

NOTE: This PR is ready for a technical review and has completed its design review. Unit test files have not yet been written for this component (once the technical review is completed) and documentation (README.md file) has not been fully completed yet.

PhoneField_Component1
PhoneField_Component2

## 

### How to 🎩

To tophat this component and analyze the logic, this document can help: https://docs.google.com/document/d/1OwbE-PsoaAOg4adWVzIjwYlFzpRAgIOOlKUduz-3sFY/edit?usp=sharing. This document explains the purpose of important function in the component, and edge cases when handling phone numbers. 

You can also tophat this component by going into Polaris Storybook, and under the 'PhoneField' component, use one variation of the component. You can also tophat this component in playground. 

<details>
<summary>To test the code with the testashwin NPM package: <code>playground/Playground.tsx</code>:</summary>

```jsx
import allCountries from 'testashwin/index.json';

import {
  Page,
  TextField,
  PhoneField,
  PhoneFieldProps,
  Card,
  EmptyState,
} from '../src';

function checkValidPhoneNumber(
  phoneNum: string,
  selectedCountryObject: Country,
) {
  let validPhoneNumber = true;

  if (!phoneNum.startsWith(selectedCountryObject.countryCode)) {
    validPhoneNumber = false;
  }

  if (phoneNum === '' || phoneNum === '+') {
    validPhoneNumber = false;
  }

  const selectedCountryMaxDigits = selectedCountryObject.displayFormat.reduce(
    (accumulator: number, currentValue: number) => accumulator + currentValue,
  );

  if (
    phoneNum.length !==
    selectedCountryMaxDigits + selectedCountryObject.countryCode.length
  ) {
    validPhoneNumber = false;
  }

  // Check if 'area code' of phone number exists for countries with same country code
  return validPhoneNumber;
}

  return (
    <Page title="Playground">
      <PhoneField
        labelName="Mobile Phone Number"
        countries={allCountries}
        searchBar
        // validator={checkValidPhoneNumber}
        prefix
      />
    </Page>
  );
}


```

</details>

<details>
<summary>To test the code without the 'testashwin' NPM package: <code>playground/Playground.tsx</code>:</summary>

```jsx
import allCountries from 'testashwin/index.json';

import {Country} from '../src/components/PhoneField';
import {
  Page,
  TextField,
  PhoneField,
  PhoneFieldProps,
  Card,
  EmptyState,
} from '../src';

const countries: PhoneFieldProps['countries'] = [
  {
    image: '🇺🇸',
    countryName: 'United States',
    countryCode: '+1',
    displayFormat: [3, 3, 4],
    areaCodes: [201, 202, 408, 409, 412],
  },
  {
    image: '🇨🇦',
    countryName: 'Canada',
    countryCode: '+1',
    displayFormat: [3, 3, 4],
    areaCodes: [403, 416, 613, 579, 604, 613, 867],
  },
  {
    image: '🇮🇳',
    countryName: 'India',
    countryCode: '+91',
    displayFormat: [5, 5],
  },
  {
    image: '🇨🇳',
    countryName: 'China',
    countryCode: '+45',
    displayFormat: [3, 4, 4],
  },
  {
    image: '🇦🇷',
    countryName: 'Argentina',
    countryCode: '+54',
    displayFormat: [2, 4, 4],
  },
  {
    image: '🇦🇺',
    countryName: 'Australia',
    countryCode: '+61',
    displayFormat: [3, 3, 3],
  },
  {
    image: '🇦🇹',
    countryName: 'Austria',
    countryCode: '+43',
    displayFormat: [3, 3, 4],
  },
];

function checkValidPhoneNumber(
  phoneNum: string,
  selectedCountryObject: Country,
) {
  let validPhoneNumber = true;

  if (!phoneNum.startsWith(selectedCountryObject.countryCode)) {
    validPhoneNumber = false;
  }

  if (phoneNum === '' || phoneNum === '+') {
    validPhoneNumber = false;
  }

  const selectedCountryMaxDigits = selectedCountryObject.displayFormat.reduce(
    (accumulator: number, currentValue: number) => accumulator + currentValue,
  );

  if (
    phoneNum.length !==
    selectedCountryMaxDigits + selectedCountryObject.countryCode.length
  ) {
    validPhoneNumber = false;
  }

  // Check if 'area code' of phone number exists for countries with same country code
  return validPhoneNumber;
}

  return (
    <Page title="Playground">
      <PhoneField
        labelName="Mobile Phone Number"
        countries={countries}
        searchBar
        // validator={checkValidPhoneNumber}
        prefix
      />
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/PhoneField/PhoneField.tsx
+++ b/src/components/PhoneField/PhoneField.tsx
@@ -1,0 +1,345 @@
+import React, {useState, useEffect, useCallback} from 'react';
+
+import {TextField} from '../TextField';
+import {Button} from '../Button';
+import {Popover} from '../Popover';
+import {ActionList} from '../ActionList';
+
+import styles from './PhoneField.scss';
+
+export interface Country {
+  /** The country flag */
+  image: string;
+  /** Country name */
+  countryName: string;
+  /** Country area code */
+  countryCode: string;
+  countryAlphaCode?: string;
+  /** Phone number display format */
+  displayFormat: number[];
+  /** Possible area codes for a country. Used to distinguish between countries with same country codes */
+  areaCodes?: number[];
+  population?: number;
+  formatter?(): void;
+}
+
+export interface PhoneFieldProps {
+  /** Placeholder text  */
+  placeholder?: string;
+  /** Textfield label  */
+  labelName?: string;
+  /** Hides the label for accessibility */
+  labelHidden?: boolean;
+  /** Is textfield optional */
+  optional?: boolean;
+  /** Error message */
+  errorMessage?: string;
+  /** Country list in dropdown */
+  countries: Country[];
+  /** The search bar in the dropdown is present */
+  searchBar?: boolean;
+  /** Validator for phone number */
+  validator?(number: string, country: Country): boolean;
+}
+
+export function PhoneField({
+  errorMessage,
+  placeholder,
+  labelName,
+  labelHidden,
+  optional,
+  countries,
+  validator,
+  searchBar,
+}: PhoneFieldProps) {
+  const [phoneNumber, setPhoneNumber] = useState('');
+  const [validPhoneNumber, setValidPhoneNumber] = useState(true);
+  const [popoverActive, setPopoverActive] = useState(false);
+  const [clicked, setClicked] = useState(false);
+
+  // Conduct research on which country appears first
+  const [searchBarText, setSearchBarText] = useState('');
+  const allCountries = countries.map(
+    ({countryName, image, countryCode}, index) => ({
+      content: `${image}  ${countryName} (${countryCode})`,
+      onAction: () => handleSelected(index),
+    }),
+  );
+
+  const maxCountryPopulationObj = useCallback((countryArr: Country[]) => {
+    const filteredObj = countryArr.filter(
+      (countryObj) => 'population' in countryObj,
+    );
+
+    if (filteredObj.length > 0) {
+      let selectedCountryObj = filteredObj[0];
+
+      filteredObj.forEach((filteredObj) => {
+        if (filteredObj.population && selectedCountryObj.population) {
+          if (filteredObj.population > selectedCountryObj.population)
+            selectedCountryObj = filteredObj;
+        }
+      });
+
+      return selectedCountryObj;
+    }
+
+    return countryArr[0];
+  }, []);
+
+  const [selectedCountryObject, setSelectedCountryObject] = useState(
+    maxCountryPopulationObj(countries),
+  );
+
+  const [formattedPhoneNumber, setFormattedPhoneNumber] = useState(
+    `${selectedCountryObject.countryCode}`,
+  );
+
+  const [countryOptions, setCountryOptions] = useState(allCountries);
+
+  const handleClicked = useCallback((bool) => setClicked(bool), [setClicked]);
+  /** Callback function for handling when the popover is clicked */
+  const togglePopoverActive = useCallback(() => {
+    setPopoverActive((popoverActive) => !popoverActive);
+    setSearchBarText('');
+    setCountryOptions(allCountries);
+  }, [allCountries]);
+
+  /* Checks if there is at least one country with 'area code object' */
+  function checkAreaCodeKeyExists(countryArr: Country[]) {
+    const filteredCountryArr = countryArr.filter(
+      (countryObj) => 'areaCodes' in countryObj,
+    );
+
+    return filteredCountryArr.length !== 0;
+  }
+
+  // Given a countryArr, where the countryCodes are the same, and we are retrieving the area codes
+  function retrieveAreaCodeMatches(countryArr: Country[], phoneNum: string) {
+    // If we have pass in '+120255, we obtain '202'
+    // If we have pass in '+120', we obtain '20'
+    // If we have pass in '+1202', we obtain '202'
+    // console.log('This is the country array being passed in');
+    // console.log(countryArr);
+    const retrieveAreaCodeFromPhoneNum =
+      countryArr[0].countryCode.length + countryArr[0].displayFormat[0] <=
+      phoneNum.length
+        ? phoneNum.slice(
+            countryArr[0].countryCode.length,
+            countryArr[0].countryCode.length + countryArr[0].displayFormat[0],
+          )
+        : phoneNum.slice(countryArr[0].countryCode.length);
+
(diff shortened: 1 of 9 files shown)


src/components/PhoneField/PhoneField.tsx
```diff
@@ -0,0 +1,338 @@
...
+
+export interface PhoneFieldProps {
+  /** Placeholder text  */
+  placeholder?: string;
+  /** Textfield label  */
+  labelName?: string;
+  /** Hides the label for accessibility */
+  labelHidden?: boolean;
+  /** Is textfield optional */
+  optional?: boolean;
```
BPScott, line 34: I don't think this adds much value - if people want to add "optional" text at the end of their string then they should add it in the labelName. Also this kind of string concatenation doesn;t really work with internationalisation - some we'd need to translate the "optional" and there's not guarentee its location will be in a meaningfull place in non-english languages

src/components/PhoneField/PhoneField.tsx
```diff
@@ -0,0 +1,338 @@
...
+  /** Error message */
+  errorMessage?: string;
+  /** Country list in dropdown */
+  countries: Country[];
+  /** The search bar in the dropdown is present */
+  searchBar?: boolean;
```
BPScott, line 40: Do we have a use case for showing/hiding the search bar? If possible I'd like to avoid adding options with not existing use case other than "it seemed handy"

src/components/PhoneField/PhoneField.tsx
```diff
@@ -0,0 +1,338 @@
...
+    [retrieveCountries],
+  );
+
+  useEffect(() => {
+    if (!validator) return;
+    setValidPhoneNumber(validator(phoneNumber, selectedCountryObject));
+  }, [phoneNumber, selectedCountryObject, validator]);
+
+  const activator = (
+    // eslint-disable-next-line @shopify/jsx-no-hardcoded-content
```
BPScott, line 291: The strings in here need to be added to the translation files.

BPScott (commented):
Nice work! This is component requires some reasonably meaty data and with that comes performance implications.

I've got a bunch of questions....

What is our expectation of where the country data should live? Do we:

- Have zero opinions, it is entirely up to apps to source this data. (at which point "to use this component you also need this data i don't care where you get it from and also you have to pass it into the item every time" sound like a poor usage experience). If this is the case do we have opinions on where they should source the data from?
- Bundle this list of countries into polaris and it "just works"

Would we ever expect each app to customize list of countries available to them, what is the value in being able to do this? What is the value in having to pass the country list into every PhoneNumber component? If there isn't any value would it be easier to define hardcode this data within the component rather than need it to be passed in?

---

Are there any other more optimal ways of storing the data json - 30kb is a respectably chunky dataset, is there any way that could be reduced, either by changing how we structure our data or shaving off unneeded parts?

---

We'd want the country names to be internationalized (e.g. Germany => Deutchland). Getting people to provide all possible translations here based upon the names of the countries available feels rough, so it might be best if we own the country list, and then we can own the translations too.

At that point I'm starting to get a more worried about the performance impact - The JSON data is 30kb which is pretty beefy, and the translations will be moreso. The translations will be a particular problem - we currently put all our translations into a single file, that means consumers would pay the price of downloading all the country name translations even if they never used the PhoneNumber component. That's rather rubbish. We could fix that up by investigating per-component lazy loaded translations and using react-i18n like what apps do, but I'm not sure of how that'd work with consuming apps - we'd need to make sure no special webpack config is required for consumers and that things "just work" in both apps that do and don't provide their own i18n providers. 

That's starting to sprawl outside the scope of this component though. Getting some more concrete numbers of "how much extra size do the translations add" would be useful before we make a call on if that extra weight even when not using PhoneNumber is a dealbreaker or not.

---

I've not looked too deeply at the react implementation but one thing i noticed was that the search stuff doesn't have any keyboard navigation - I'd expect I'd be able to move from the searchfield to the list of options and then through the options by pressing the down arrow key and select items using enter. The interactivity of this feels close to `Autocomplete` - but with search results that remain open and don't overlay would it be worth investigating if there isis there some way we could leverage that component to piggy back off it's behaviour instead of reimplementing something similar?

---

 A lot of the functions that deal with filtering the country list are tucked inside the component. That makes them impossible to unit test without having load the component interact with it (which is really slow compared to testing what functions return). Extract out identifyPhoneNumberCountry so it is outside of the component (probably in a utils.ts file) and you can test how that behaves much faster.

plentz:
Is the idea to replace https://github.com/Shopify/web/tree/master/app/components/PhoneField?

related https://github.com/Shopify/intl-globalization/issues/114

src/components/PhoneField/PhoneField.tsx
```diff
@@ -0,0 +1,338 @@
...
+import {Popover} from '../Popover';
+import {ActionList} from '../ActionList';
+
+import styles from './PhoneField.scss';
+
+export interface Country {
+  /** The country flag */
+  image: string;
+  /** Country name */
+  countryName: string;
```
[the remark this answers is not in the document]
cejaekl: Yep.  Please use [country-service](https://github.com/Shopify/country-service#countries) for translated country names.  That provides 262 names in 537 locales (140K potential strings, far too much to add to your bundle).

src/components/PhoneField/PhoneField.tsx
cejaekl: Be careful about auto-constructing labels by piecing strings together.  Depending on the language, `(optional)` may need to adjust, e.g. in gender, to match what it's modifying.

ashwinnarayanan2001:
> https://github.com/Shopify/web/tree/master/app/components/PhoneField

He

ashwinnarayanan2001:
> Is the idea to replace https://github.com/Shopify/web/tree/master/app/components/PhoneField?
> 
> related [Shopify/intl-globalization#114](https://github.com/Shopify/intl-globalization/issues/114)

Yes, the idea is to replace that component.

PR #3343, opened 2020-09-30
Bug fix: adds extra padding to dismissable banner

416serg:
Screen Shot 2020-09-30 at 3 09 45 PM

`hasDismiss` style wasn't being applied to `.newDesignLanguage` since it had more cascading points for the regular padding. This works but not sure if the implementation is the best

Closes https://github.com/Shopify/store/issues/17695

When fixed, here are the styles that apply
Screen Shot 2020-09-30 at 3 12 11 PM


alex-page:
@416serg @dleroux found some other areas and fixed it in his PR. Thank you though really appreciate this.

PR #3359, opened 2020-10-01
[DescriptionList] Add spacing prop to DescriptionList

dleroux:

The DescriptionList is used for the [Email modal in web](https://github.com/Shopify/shopify/issues/261981) and from rails to react there's a significant difference in the spacing being used. 

Adds a prop to allow for tighter spacing in the description list.

Current:

Screen Shot 2020-10-01 at 4 20 52 PM

Spacing: tight
Screen Shot 2020-10-01 at 4 19 50 PM

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/DescriptionList/DescriptionList.scss
+++ b/src/components/DescriptionList/DescriptionList.scss
@@ -18,9 +18,17 @@ $breakpoint: 550px;
   @include text-emphasis-strong;
   padding: spacing() 0 spacing(tight);
 
+  .spacingTight & {
+    padding: spacing(tight) 0 spacing(extra-tight);
+  }
+
   @include page-content-breakpoint-after($breakpoint) {
     flex: 0 1 25%;
     padding: spacing() spacing() spacing() 0;
+
+    .spacingTight & {
+      padding: spacing(tight) spacing(tight) spacing(tight) 0;
+    }
 
     // stylelint-disable-next-line selector-max-class, selector-max-combinators
     .Description + & + .Description {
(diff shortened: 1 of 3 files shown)


src/components/DescriptionList/DescriptionList.scss
```diff
@@ -41,6 +53,10 @@ $breakpoint: 550px;
     flex: 1 1 51%;
     padding: spacing() 0;
 
+    .spacingTight & {
+      padding: spacing(tight) 0;
+    }
```
kyledurand, line 58: Do you still need to add styles for loose spacing?

alex-page (approved):
I was wondering if there was a way to fix this without as many overrides but it looks good.

src/components/DescriptionList/DescriptionList.scss
kyledurand: Oh it looks like loose is the default. That seems odd to me. Do we need a middle ground or do you think this is ok for now?

dleroux: Loose is the default. Didn't want to make it a boolean prop so add loose. Not needed but allows us to add more in the future without breaking. thoughts?

dleroux: open to suggestions though

kyledurand (approved):
Looks good to me 👍

dleroux:
> I was wondering if there was a way to fix this without as many overrides but it looks good. Trust your work Dan.

I could group them but I found this makes it clearer looking at the css.

PR #3365, opened 2020-10-02
Use React.Children.toArray to remove child component key warning logs

alex-page:

Fixes warning logs seen on this page http://polaris-react.herokuapp.com/?path=/story/all-components-resource-list--all-examples

Generating the key with `React.Children.toArray`

### 🎩 checklist

### DIFF
--- a/src/components/ResourceList/ResourceList.tsx
+++ b/src/components/ResourceList/ResourceList.tsx
@@ -706,7 +706,7 @@ export const ResourceList: ResourceListType = function ResourceList<ItemType>({
       aria-busy={loading}
     >
       {loadingOverlay}
-      {items.map(renderItemWithId)}
+      {React.Children.toArray(items.map(renderItemWithId))}
     </ul>
   ) : null;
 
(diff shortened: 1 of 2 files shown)


src/components/ResourceList/ResourceList.tsx
```diff
@@ -706,7 +706,7 @@ export const ResourceList: ResourceListType = function ResourceList<ItemType>({
       aria-busy={loading}
     >
       {loadingOverlay}
-      {items.map(renderItemWithId)}
+      {React.Children.toArray(items.map(renderItemWithId))}
```
kyledurand, line 709: The reason `React.Children.toArray` works here is because it changes the child keys internally: https://github.com/facebook/react/blob/7dc9745427046d462506e9788878ba389e176b8a/packages/react/src/ReactChildren.js#L369

I can't foresee any issues with this but I could be mistaken. We do have a render order prop on the resource list with header in web that _might_ get in the way

dfmcphee (approved):

kyledurand (approved):

PR #3370, opened 2020-10-02
[Button] Fix icon from flashing and width from changing when transitioning to a loading state

LauraAubin:

Fixes https://github.com/Shopify/web/issues/31389

- The icon flashes a black square right before the loading spinner is shown
- The width of the button changes when the loading spinner is rendered

|Before|After|
|-|-|
|Before |After|

- Visually hide the icon when `loading` is true since we only care to see the spinner in this state
- The spinner is always rendered but `visually: hidden` is added and removed based on the state of `loading`

## 

### How to 🎩

This problem can be replicated with,

```jsx
import React, { useState } from "react";
import { Button } from "@shopify/polaris";

export default function App() {
  const [loading, isLoading] = useState(false);

  return (
    <div>
      <Button icon="bank" loading={loading} onClick={() => isLoading(!loading)}>Add product</Button>
      <Button onClick={() => isLoading(false)}>Stop loading</Button>
    </div>
  );
}

```

### 🎩 checklist

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -138,6 +138,10 @@ $stacking-order: (
   }
 }
 
+.Hidden {
+  visibility: hidden;
+}
+
 .Spinner {
   position: absolute;
   top: 50%;
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -185,7 +185,11 @@ export function Button({
     ) : (
       icon
     );
-    iconMarkup = <span className={styles.Icon}>{iconInner}</span>;
+    iconMarkup = (
+      <span className={classNames(styles.Icon, loading && styles.Hidden)}>
+        {iconInner}
+      </span>
+    );
   }
 
   const childMarkup = children ? (
@@ -204,7 +208,9 @@ export function Button({
         )}
       />
     </span>
-  ) : null;
+  ) : (
+    <span />
+  );
 
   const content =
     iconMarkup || disclosureIconMarkup ? (
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -10,6 +10,10 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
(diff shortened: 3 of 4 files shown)


src/components/Button/Button.scss
```diff
@@ -138,6 +138,10 @@ $stacking-order: (
   }
 }
 
+.Hidden {
+  visibility: hidden;
```
LauraAubin, line 142: At first I tried using the `visually-hidden` mixin, but this doesn't preserve the space that the icon is taking up, regardless if it is hidden or not.

src/components/Button/Button.tsx
```diff
@@ -185,7 +185,11 @@ export function Button({
     ) : (
       icon
     );
-    iconMarkup = <span className={styles.Icon}>{iconInner}</span>;
+    iconMarkup = (
+      <span className={classNames(styles.Icon, loading && styles.Hidden)}>
+        {iconInner}
+      </span>
+    );
```
LauraAubin, line 192: I went with hiding the icon to preserve the width of the button, instead of removing it entirely.

alex-page (approved):
Thanks @LauraAubin

alex-page:
@LauraAubin merge master into your branch some builds were broken. Sorry.

src/components/Button/Button.tsx
```diff
@@ -204,7 +208,9 @@ export function Button({
         )}
       />
     </span>
-  ) : null;
+  ) : (
+    <span />
```
LauraAubin, line 212: Fixes the button width from changing when the spinner is shown or not.

kyledurand: Hey Laura! It looks like this change broke pagination buttons 

image

Would you be able to find an alternate fix before the next release goes out later this week?

dleroux: The empty span is breaking buttons that are icon only. Noticed on the banner dismissible icons

kyledurand: Thanks Dan, Laura fixed this up yesterday [here](https://github.com/Shopify/polaris-react/pull/3392)

PR #3394, opened 2020-10-07
Bug fix: content and dismiss overlap for banner

416serg:
### Before

Screen Shot 2020-10-06 at 8 17 52 PM

### After

Screen Shot 2020-10-06 at 8 18 42 PM

### DIFF
--- a/src/components/Banner/Banner.scss
+++ b/src/components/Banner/Banner.scss
@@ -131,10 +131,6 @@ $secondary-action-horizontal-padding: 1.5 * spacing(tight);
     $newDesignLanguageOffset: $newDesignLanguageSpacing - spacing();
     padding: spacing() spacing() $newDesignLanguageSpacing;
 
-    &.hasDismiss {
-      padding-right: calc(#{2 * spacing()} + var(--p-icon-size));
-    }
-
     .ContentWrapper {
       margin-top: $newDesignLanguageOffset;
     }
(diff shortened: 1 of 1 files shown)


src/components/Banner/Banner.scss
```diff
@@ -131,10 +131,6 @@ $secondary-action-horizontal-padding: 1.5 * spacing(tight);
     $newDesignLanguageOffset: $newDesignLanguageSpacing - spacing();
     padding: spacing() spacing() $newDesignLanguageSpacing;
 
-    &.hasDismiss {
-      padding-right: calc(#{2 * spacing()} + var(--p-icon-size));
-    }
```
416serg, line 136: Removing from under `.withinContentContainer` since it wasn't being applied to `.withinPage`

src/components/Banner/Banner.scss
```diff
@@ -157,11 +153,6 @@ $secondary-action-horizontal-padding: 1.5 * spacing(tight);
 
   .Ribbon {
     padding-right: $intermediate-spacing;
-
-    // stylelint-disable-next-line selector-max-combinators, selector-max-class
-    .newDesignLanguage & {
-      padding-right: spacing();
-    }
```
416serg, line 164: I don't think this worked because that's not the right syntax so just cleaning up a bit, same styling was applied [here](https://github.com/Shopify/polaris-react/blob/0daad919bcedebcd99cf27be1cea97c666543e01/src/components/Banner/Banner.scss#L146) as part of https://github.com/Shopify/polaris-react/pull/3355

alex-page (approved):

dleroux (approved):
🎉 Thanks @416serg. I rebased to confirm the alignment will be fixed with [this PR](https://github.com/Shopify/polaris-react/pull/3392) and it does!

PR #3395, opened 2020-10-07
Add ref type to modal activator

dleroux:

In order to improve accessibility on our Modal, this adds the ability to include a ref as the activator type on Modal.

We recently added the activator prop in order to return focus to it once the modal closes. However, it's not always possible to pass in the element and consumers are left to manage their own focus, which is in most cases overlooked. By adding a ref as a possible type, the Modal can still be responsible for focusing the correct node when closing.

- Adds React.RefObject as a possible prop type for the activator
- updates test
- updates documentation

## 

### How to 🎩

Storybook contains all the examples. Ensure they all work as expected.

### 🎩 checklist

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -55,8 +55,8 @@ export interface ModalProps extends FooterProps {
   onTransitionEnd?(): void;
   /** Callback when the bottom of the modal content is reached */
   onScrolledToBottom?(): void;
-  /** The element to activate the Modal */
-  activator?: React.ReactElement;
+  /** The element or the RefObject that activates the Modal */
+  activator?: React.RefObject<HTMLElement> | React.ReactElement;
 }
 
 export const Modal: React.FunctionComponent<ModalProps> & {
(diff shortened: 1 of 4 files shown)


alex-page (approved):

src/components/Modal/README.md
```diff
@@ -641,11 +641,70 @@ function ModalWithScrollListenerExample() {
...
+    </div>
+  );
+}
+```
+
+### Modal without an activator prop
+
+<!-- example-for: web -->
+
+Use an external activator when technical limitations prevent you from passing the activator as element or a ref. Make sure to focus the activator on close when choosing this approach.
```
dleroux, line 707:
```suggestion
Use an external activator when technical limitations prevent you from passing the activator as an element or a ref. Make sure to focus the activator on close when choosing this approach.
```

src/components/Modal/Modal.tsx
```diff
@@ -211,4 +219,10 @@ export const Modal: React.FunctionComponent<ModalProps> & {
   );
 };
 
+function isRef(
+  ref: React.RefObject<HTMLElement> | React.ReactElement,
+): ref is React.RefObject<HTMLElement> {
+  return Object.prototype.hasOwnProperty.call(ref, 'current');
```
kyledurand, line 225: Such a nice way to do a type guard 👍 

Just out of curiosity, does `ref.hasOwnProperty('current')` not work on react elements?

kyledurand (approved):
This looks great 👍

src/components/Modal/Modal.tsx
dleroux: I think it would but we have this rule enabled: https://eslint.org/docs/rules/no-prototype-builtins

PR #3430, opened 2020-10-08
Update search input and results to new design language

martenbjork:
- Adjusts width of search results so that they match the search field
- Makes the backdrop cover 100% of the screen (behind the text input + results)
- Adds a background to the search results. We currently recommend that users pass in a Card as a background, but this change adds a "forced" background to the results. This is because the NDL requires 1) a light mode shadow around the component and 2) dark mode contents of the component. This is quite tricky to achieve so it makes sense for the Polaris component to take care of that logic.

search-ndl

### How to 🎩

View the changes locally in Storybook and go to the Details page.

### 🎩 checklist

### DIFF
--- a/src/components/TopBar/components/SearchField/SearchField.scss
+++ b/src/components/TopBar/components/SearchField/SearchField.scss
@@ -5,7 +5,6 @@ $icon-size: rem(20px);
 $input-height: rem(34px);
 $new-input-height: rem(36px);
 $clear-button-width: $icon-size + spacing();
-$new-search-max-width: rem(580px);
 
 $stacking-order: (
   backdrop: 1,
@@ -25,6 +24,7 @@ $stacking-order: (
 }
 
 .SearchField-newDesignLanguage {
+  z-index: z-index(modal, $fixed-element-stacking-order);
   max-width: $new-search-max-width;
 
   .Backdrop {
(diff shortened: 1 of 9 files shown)


src/components/TopBar/components/SearchField/SearchField.scss
```diff
@@ -25,6 +24,7 @@ $stacking-order: (
 }
 
 .SearchField-newDesignLanguage {
+  z-index: z-index(modal, $fixed-element-stacking-order);
```
kyledurand, line 27: Does this maybe need modal + 1? i.e if a modal is open will it interfere with the modal back drop at all? I'm guessing no but want to make sure

alex-page (approved):
Clean, thanks @martenbjork 🏌️‍♂️

kyledurand (approved):
🙌  Looks great!

src/components/TopBar/components/SearchField/SearchField.scss
martenbjork: Hmm, interesting thought. I can't see how two modals could be open at the same time, unless there is a bug (or one is triggered asynchronously). Are there any other cases?

kyledurand: I was thinking that maybe a modal could be opened before the search field was activated. and the active modal background might interfere with the search field in some way. I just checked in core on one modal and it's not possible so I think you're good 👍

PR #3444, opened 2020-10-09
[new DL] Update user menu to accept a colorScheme prop

martenbjork:

Allows theming of the contents of `Popover` and `Menu`. This allows the `Topbar.UserMenu` to dark mode for its content

Makes the `Popover`, `Menu` and `Topbar.UserMenu` components accept a colorScheme prop which sets the color scheme for their contents. Apologies for the prop drilling.

Screen Shot 2020-10-09 at 10 37 37 AM

### 🎩 checklist

### DIFF
--- a/src/components/ThemeProvider/ThemeProvider.tsx
+++ b/src/components/ThemeProvider/ThemeProvider.tsx
@@ -12,7 +12,7 @@ import {useFeatures} from '../../utilities/features';
 
 type OriginalColorScheme = Required<ThemeConfig['colorScheme']>;
 type Inverse = 'inverse';
-type InversableColorScheme = OriginalColorScheme | Inverse;
+export type InversableColorScheme = OriginalColorScheme | Inverse;
 
 // TS 3.5+ includes the built-in Omit type which does the same thing. But if we
 // use that then we break consumers on older versions of TS. Consider removing
--- a/playground/DetailsPage.tsx
+++ b/playground/DetailsPage.tsx
@@ -182,6 +182,7 @@ export function DetailsPage() {
       initials="D"
       open={userMenuActive}
       onToggle={toggleUserMenuActive}
+      colorScheme="dark"
     />
   );
 
--- a/src/components/Popover/Popover.tsx
+++ b/src/components/Popover/Popover.tsx
@@ -7,6 +7,7 @@ import React, {
   AriaAttributes,
 } from 'react';
 
+import {InversableColorScheme, ThemeProvider} from '../ThemeProvider';
 import {
   findFirstFocusableNodeIncludingDisabled,
   focusNextFocusableNode,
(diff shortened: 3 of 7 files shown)


src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -12,7 +12,7 @@ import {useFeatures} from '../../utilities/features';
 
 type OriginalColorScheme = Required<ThemeConfig['colorScheme']>;
 type Inverse = 'inverse';
-type InversableColorScheme = OriginalColorScheme | Inverse;
+export type InversableColorScheme = OriginalColorScheme | Inverse;
```
martenbjork, line 15: Is it OK to put the export here or should it live somewhere else?

alex-page: I think if you put it here it's public. Can you import it without that?

alex-page (approved):

src/components/ThemeProvider/ThemeProvider.tsx
martenbjork: I need to import it in the other components that accept a `colorScheme` with type `InversableColorScheme`.

kyledurand: We'll have to wait for @BPScott to come on line to be sure but I think this is ok. I think only items exported from `src/components/index.ts` are made public. So to make this public you'd need to export the type in that file as well

alex-page (approved):

src/components/ThemeProvider/ThemeProvider.tsx
BPScott: @kyledurand you're right. This is fine as is. 

Feels like we're on the cusp of crosstalk where "public" means two different things: "Available to other components within this library" and "available as an export of `@shopify/polaris`".

`components/ThemeProvider/index.ts` reexports everything so it is available within the library internally by doing `import {InversableColorScheme} from 'BLAH/components/ThemeProvider';` as Marten has already done.

If you *really* need a type to be externally available, so consumers can do `import {InversableColorScheme} from '@shopify/polaris';` you'd need to export it from `src/components/index.ts` as Kyle says, but in this case (and most other cases) you don't shouldn't need to do that.

We don't need to expose most types publicly anyway, because the majority of the time consumers don't care about a specific type, but about how they're exposed from some props. They don't care about `InversableColorScheme`, they care about "Whatever type PopoverProps['colorScheme'] desires because I'm trying to render a Popover" (which just so happens to be `InversableColorScheme` at the moment).

Thus people can still get at this indirectly:

```
import {PopoverProps} from '@shopify/polaris'
const popoverColorScheme: PopoverProps['colorScheme'] = 'inverse'
```

PR #3449, opened 2020-10-09
[new DL][Skeleton Page] Ensure Skeleton Page title is the same font size as title rendered in Page component

francinen:

Fixes https://github.com/Shopify/polaris-react/issues/3414

When the new design language is enabled, the page title shrinks when the Skeleton Page component is replaced by the Page component.

Kapture 2020-10-07 at 14 08 22

The font size for the page title should be consistent between the Skeleton Page component and the Page component.

This PR implements a solution similar to the one introduced by https://github.com/Shopify/polaris-react/pull/3468. Instead of using JavaScript to conditionally render small or large Display Text when the navigation is collapsed, the Skeleton Page component now relies purely on CSS to set different font sizes based on specific breakpoints.

|New design language|Current design language (no regressions)
|---|---
|page-title-skeleton-after|page-title-skeleton-prev-design-lang

### How to 🎩

#### Testing in Storybook
1. Run `dev s` in the Polaris React repo to open Storybook
2. Look at the Skeleton Page examples and ensure there are no regressions

|Page title defined|Page title undefined
|---|---
|image|image

#### Testing against a Shopify Web
1. Check out this branch and configure Shopify Web to consume your local Polaris React repo
2. Visit a page that uses the Skeleton Page component (e.g., /admin/apps, /admin/products, /admin/dashboards)

|Viewport width|Page title defined|Page title undefined|
|---|---|---
|Mobile|skeleton-page-title-defined-mobile|skeleton-page-title-null-mobile
|Desktop|skeleton-page-title-defined-wide|skeleton-page-title-null-wide

### 🎩 checklist


francinen:
Does anyone know why small Display Text is being rendered when the navigation is collapsed, even though we're passing in "large" as the size prop?

image

image

cc @kyledurand

kyledurand:
> Does anyone know why small Display Text is being rendered when the navigation is collapsed

I wonder if this is because the media query doesn't yet know the size of the viewport. @BPScott would probably know the answer for sure. 

Ben could this be caused by SSR? @francinen does this ever get display text large when the nav is collapsed? Or does it flash small and then switch to large?

francinen:
@kyledurand When the navigation is collapsed, the display text switches from small to large as the Skeleton Page is replaced by the Page component.

BPScott:
> I wonder if this is because the media query provider doesn't yet know the size of the viewport. @BPScott would probably know the answer for sure.
> Ben could this be caused by SSR? @francinen does this ever get display text large when the nav is collapsed? Or does it flash small and then switch to large?

Yeah this sounds right to me. When the HTML is rendered on the server-side it doesn't doesn't have any concept of media-queries or `window.matchMedia` (or even `window`!). You're using `{isNavigationCollapsed} = useMediaQuery()`, taking a look at [`MediaQueryProvider`](https://github.com/Shopify/polaris-react/blob/master/src/components/MediaQueryProvider/MediaQueryProvider.tsx#L15-L17) and following that into [`breakpoints.tsx`](https://github.com/Shopify/polaris-react/blob/master/src/utilities/breakpoints.ts#L19-L23) we see that sets `isNavigationCollapsed` is based on a media query, and on the server where no `window.matchMedia` is available it defaults to returning "false" - same as on a large screen (>768px) in a browser.

If our aim here is to match what the Page Header Title component then it looks like we're good. But there might be a better way (depending on how much effort you want to put in here)...

To get up on a pedestal for a moment: IMO MediaQueryProvider is an anti-pattern because it causes issues like this where there ends up being html mismatches between the client and the server. Changing html content based upon browser width has the potential to make the browser do lots of work and we should trry to avoid it where possible, instead we should rely on good old CSS, and where feasible always render the same HTML but show/hide it with CSS media queries.


Ideally instead of doing `<DisplayText size={newDesignLanguage ? (displaySize = isNavigationCollapsed ? 'large' : 'small') : 'large'}>` in both Page Header Title and in SkeletonPage we should create a new css class and style it with our scss mixins, so that we don't need to use `useMedia()` at all.

```scss
.TitleText {
  @include text-style-display-large;

  .newDesignLanguage  {
    @include frame-when-nav-displayed {
      @include text-style-display-small;
  }
}
```

francinen:
@BPScott The CSS approach sounds a lot more scalable. I can give that a try later today. Thank you!

francinen:
@kyledurand @BPScott This PR is officially ready for review

src/components/SkeletonPage/SkeletonPage.tsx
```diff
@@ -101,14 +106,22 @@ function renderSecondaryActions(actionCount: number) {
...
     ) : (
       <DisplayText size="large" element="h1">
-        {title}
+        <TextStyle variation="strong">{title}</TextStyle>
       </DisplayText>
     );
-  return <div className={styles.Title}>{titleContent}</div>;
+  }
+
+  return <SkeletonDisplayText size="large" />;
```
BPScott, line 126: Does this `<SkeletonDisplayText size="large">` have the same visual height as the new design language title at all viewport sizes?

I've not confirmed this but I'm guessing it doesn't as `<DisplayText size="large">` matches `<SkeletonDisplayText size="large">` and wrote our own logic instead of using `<DisplayText size="large">` in Page. (Alternatively `<SkeletonDisplayText size="large">` and thus `<DisplayText size="large">` matches the new header css in Page in which case we should just use DisplayText instead of writing custom css in Page)

We should make sure this Skeleton content is the same height so we don't trigger layout shifts here too.

kyledurand: I have the same concern too, it looks like this change would only match what we have in `Title` if the `title` was passed in to skeleton page. Otherwise you'd always be getting display text which I don't think matches perfectly.

francinen: Some of the gifs added to the PR description show the transition from Skeleton Display Text to Display Text:

skeleton-page-title-null-mobile

skeleton-page-title-null-wide

alex-page (approved):

francinen:
@BPScott or @kyledurand Are either of you able to re-review this PR? I responded to the question re: the Skeleton Display Text height here: https://github.com/Shopify/polaris-react/pull/3449#discussion_r505549812

kyledurand (approved):
Code looks good 👍 

Looks like there might still be some content jumpiness in the second gif but this is definitely a solid improvement 👏

Thanks @francinen

src/components/SkeletonPage/SkeletonPage.tsx
BPScott: Looks like `<SkeletonDisplayText size="large">` uses the `display-large` line heights for its heights - [28px base, 32px when-typography-not-condensed](https://github.com/Shopify/polaris-react/blob/e290a8c751db794cc35916698402364c16588aa5/src/styles/foundation/_typography.scss#L44-L47), while Page title / newDesignLanguageTitle uses 28px at all sizes) so these are indeed different.

That looks about right with the jump seen in those gifs.

Would you be interested in fixing this up so the Skeleton height matches real content so we don't get page jumps there? (uugh it looks like we currently get a jump in that first gif because the actions are taller than the title but that feels slightly out of scope for this ticket).

A div with a class like this should do the job (written but not tested):

```
// duplicated from SkeletonDisplayText.scss
$skeleton-display-text-max-width: rem(120px);

.newDesignLanguageSkeletonTitle {
  @include skeleton-shimmer;
  @include skeleton-content;

  max-width: $skeleton-display-text-max-width;
  height: rem(28px);
}
```

BPScott (commented):
Add a comment inline about Skeleton heights with a possible solution.

@kyledurand / @alex-page: Any strong thoughts on if this should be done in this PR or if this can be done in a separate PR?

francinen:
Thanks, @BPScott! How does this look?

skeleton-title

alex-page (approved):
This looks good to me. @BPScott @kyledurand anything stopping us from merging?

kyledurand:
Nope 🚢 !

BPScott:
There's still a bit of a vertical jump (note the top bar of skeleton tab bar in the gif) but i think that's from the action buttons being taller than the title text now, which feels like a separate issue. Happy to 🚢  as is! Thanks for putting up with us @francinen!

BPScott (approved):

PR #3458, opened 2020-10-09
[Page] Fix up print styles

### LINKED ISSUE #3457: [Page] Header print styles
When printing the Page header should hide actions:

|Expected-ish|Actual|
|---|---|
|Screen Shot 2020-10-09 at 12 07 06 PM|Screen Shot 2020-10-09 at 12 03 05 PM|

tmlayton:

Fixes #3457. When rebuilding Page for the new design language I stripped much of the extra styling, wrappers, etc in an effort simplify the component to a basic conditional layout of primitives and subcomponents. Although a few of those wrappers handle the hiding of actions, etc when printing. 

Printing is frequently used on order pages.

Hiding elements that should not be printed.

|Legacy|Before|After|
|---|---|---|
|Screen Shot 2020-10-09 at 12 54 44 PM|Screen Shot 2020-10-09 at 12 56 03 PM|Screen Shot 2020-10-09 at 12 54 50 PM|

### 🎩 checklist

* [x] 🎩

### DIFF
--- a/src/components/Page/components/Header/Header.scss
+++ b/src/components/Page/components/Header/Header.scss
@@ -56,6 +56,8 @@ $action-menu-rollup-computed-width: rem(40px);
   &.newDesignLanguage {
     margin-right: spacing();
   }
+
+  @include print-hidden;
 }
 
 .PaginationWrapper {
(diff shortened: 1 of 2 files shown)


kyledurand (approved):

alex-page (approved):
🙌 I can add this to 5.7.0 if you want @tmlayton

PR #3463, opened 2020-10-10
Export `NewDesignLanguageColor` type

whizkydee:

Exports the `NewDesignLanguageColor` type so consumers can easily access it for internal component typings.


BPScott:
Heya @whizkydee, could you give an example of where you're finding a need for `NewDesignLanguageColor`?

I'm wondering if you'd be able to use `IconProps['color']` instead?

We tend to avoid exporting internal types unless really required because then it makes it hard to make non-breaking refactoring changes changes. Instead ideally consumers should try and rely on plucking types off component Props types.

whizkydee:
@BPScott You make a good point! However, exporting the type might be useful for typing custom component props, but might not be that big of a deal since we don't have many of them.

I didn't realize that `IconProps` was exposed though, a simple `Exclude<IconProps['color'], Color>` makes sense and would suffice for my use case. Feel free to close this!

alex-page:
Closing for now, but lets see if it pops up again!

PR #3469, opened 2020-10-13
[new DL] Change backdrop properties

martenbjork:

- Makes the global search backdrop visible in the DetailsPage
- Changes the background of the backdrop to transparent (when new DL is on)
- Adjusts the z-index of the backdrop. Before, the overlay lived inside the search results. Due to recent changes, the search results has a background, which caused the overlay to overlay that very background. So, I've moved the overlay outside the results wrapper.


martenbjork:
BRB, with a cleaner PR.

PR #3477, opened 2020-10-13
[TextField] Added new prop that disables spinner

ehasrouni:

Fixes https://github.com/Shopify/web/pull/32659 ; In Admin and Pos Channel, we want to hide the spinner when entering a Pin (4 digit number).

This PR doesn't display/hides the spinner if the input type is 'number' and if the step="0".
Before (top field) & After (bottom field)
image

### How to 🎩

In the readme.md, I copy pasted the code for "Number field", but added the step="0" prop to hide the spinner.
```
### Number field Without Spinner

Use when input text should be a pin.
<code>
```jsx
function NumberFieldExample() {
  const [value, setValue] = useState('1');

  const handleChange = useCallback((newValue) => setValue(newValue), []);

  return (
    <TextField
      label="Quantity"
      type="number"
      step="0"
      value={value}
      onChange={handleChange}
    />
  );
}
```
</code>
</details>

### 🎩 checklist


src/components/TextField/TextField.tsx
```diff
@@ -43,6 +43,7 @@ type InputMode =
   | 'text'
   | 'decimal'
   | 'numeric'
+  | 'numeric-no-spinner'
```
kyledurand, line 46: This isn't a valid input type, unless it's super new and I don't know about it. I'm worried it will break accessibility. Can we add a prop to hide the spinner instead?

ehasrouni: Hey Kyle! Initially, I thought about doing a prop to hide the spinner, but was told that perhaps a new type would be better? https://shopify.slack.com/archives/CQK6FNLJK/p1602612476356400 I might've misunderstood what Ben meant, thank you :)

dfmcphee:
An alternative approach to this change is to use `[0-9]*` in the `pattern` attribute already available on `TextField` to tell it to only accept numbers. This will display the correct keyboard on mobile without displaying the spinner.

Example: https://codesandbox.io/s/gracious-fire-jkckq?file=/App.js

ehasrouni:
> An alternative approach to this change is to use `[0-9]*` in the `pattern` attribute already available on `TextField` to tell it to only accept numbers. This will display the correct keyboard on mobile without displaying the spinner.
> 
> Example: https://codesandbox.io/s/gracious-fire-jkckq?file=/App.js

Thank you! I just tried it and it seems to work (partially) when the input type isn't set; without setting the input type to number, it's possible to enter letters in the field, as soon as I set the input type to number, the spinner is back

dfmcphee:
Ah, ok. I think in that case you would either have to go with this change but still pass a valid number type to the underlying input.

One more approach that I think I actually prefer is that if you set the `step` prop to `0` it hides the spinner since the spinner won't actually do anything when set to 0 anyway.

ehasrouni:
> Ah, ok. I think in that case you would either have to go with this change but still pass a valid number type to the underlying input.
> 
> One more approach that I think I actually prefer is that if you set the `step` prop to `0` it hides the spinner since the spinner won't actually do anything when set to 0 anyway.

Updated the PR with the changes suggested :)

UNRELEASED.md
```diff
@@ -8,6 +8,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Added `step='0'` check to `Textfield` to disable spinner with input type 'number' ([#3477](https://github.com/Shopify/polaris-react/pull/3477))
```
dfmcphee, line 11:
```suggestion
-  Updated `Textfield` with a `type` of `number` to not render a spinner if step is set to `0` ([#3477](https://github.com/Shopify/polaris-react/pull/3477))
```

dfmcphee (commented):
Just one suggestion for the change log but otherwise this works great. Would be good to add a test for this as well.

dfmcphee (approved):
Nice. Looks good! Thanks for adding that.

kyledurand (approved):
Love this solution. Thanks @ehasrouni!

PR #3483, opened 2020-10-14
[Page] Only re-render actions when needed

dfmcphee:

This improves the rendering performance of the page actions.

Rather than storing everything in state, this moves some of the variables to refs and only uses state for the rollup and showable actions. This way all the calculations are done and then the actions get set which will cause the render.

## 

### How to 🎩
Try the details page in Storybook. It should behave the same as before but a little more performant.

### 🎩 checklist


kyledurand (approved):
Beautiful 👏

PR #3494, opened 2020-10-15
🎨 [Button] Simplify code by utilizing more of UnstyledButton

beefchimi:
**The goal of this PR is to consolidate concerns between `UnstyledButton` and `Button`.**

### Preamble

Not too long ago, [I posted this message on Slack](https://shopify.slack.com/archives/CJJD1Q7PT/p1602599400307500), which got answered by @dfmcphee - revealing that a [`UnstyledButton`](https://github.com/Shopify/polaris-react/pull/3406) had just got merged to help folks utilize the mechanics of `<Button />` while applying their own styling.

`UnstyledButton` will allow me to do away with a custom "action" component I crafted, and instead consume the Polaris component while passing my own styles + `children`. There will be some more work to be done however... and I am planning to propose a few API additions / component tweaks in the near future.

Before I do that, I thought it best to dig into what `UnstyledButton + Button` give me. I realized there was some repeated code between the two files, so I set about reducing the redundancy. While at it, I realized `Button.test.tsx` probably wasn't updated after `UnstyledButton` was introduced, so I went through the tests and did my best to clean them up.

### Changes

While this PR is mostly just cleaning up some code, there are a few changes worth documenting:
- The `loading` prop was not used by `UnstyledButton`, but some of the side-effects of that prop looked relevant to include. So, I've added `loading` to `UnstyledButton` so that both `role` and `aria-busy` can be toggled.
- ~`Button.tsx` did not apply `data-polaris-unstyled-button` to the `<a />` or `UnstyledLink`, but `UnstyledButton` did. I've removed some redundant code paths inside `Button.tsx` and the `data-polaris-unstyled-button` prop now gets applied.~

### Questions

(1) Do we really want `[key: string]: any;` as part of `UnstyledButtonProps`?
- I realize this is done so we can pass any arbitrary prop, such as `data-whatever`.
- But is this a bit too free-wheelin'?

(2) Should we pass `onClick` when `url`?
- Does this make sense? What is the execution order?

(3) Why do we use `toBeFalsey` when a value is `undefined`? Should we not be more explicit?

(4) Does the way I am extending `UnstyledButtonProps` within `Button.tsx` affect how descriptions get surfaced in the styleguide?

~(5) Do we always want to apply `data-polaris-unstyled-button` to the `<a />` and `UnstyledLink`?`~
- ~`Button.tsx` did not use this attribute, so I am wondering if we should have a `boolean` prop for `unstyledLink`?~

### Playground

<details>
<summary><strong>Copy + Paste</strong></summary>

```jsx
import React from 'react';

import {Page, Button, Stack} from '../src';

  const handleOnClick = () => console.log('onClick');

  return (
    <Page title="Playground">
      <Stack wrap>
        <Button onClick={handleOnClick}>Basic onClick</Button>
        <Button onClick={handleOnClick} disabled>
          Basic onClick and disabled
        </Button>
        <Button onClick={handleOnClick} loading>
          Basic onClick and loading
        </Button>
        <Button url="https://shopify.com" onClick={handleOnClick}>
          With URL
        </Button>
        <Button url="https://shopify.com" onClick={handleOnClick} disabled>
          With URL and disabled
        </Button>
        <Button url="https://shopify.com" onClick={handleOnClick} loading>
          With URL and loading
        </Button>
      </Stack>
    </Page>
  );
}
```

</details>

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -2,7 +2,10 @@ import React, {useRef, useState, useCallback} from 'react';
 import {CaretDownMinor} from '@shopify/polaris-icons';
 
 import {classNames, variationName} from '../../utilities/css';
-import {handleMouseUpByBlurring} from '../../utilities/focus';
+import {
+  handleMouseUpByBlurring,
+  MouseUpBlurHandler,
+} from '../../utilities/focus';
 import {useFeatures} from '../../utilities/features';
 import {useI18n} from '../../utilities/i18n';
 import {Icon} from '../Icon';
@@ -91,14 +94,37 @@ export interface ButtonProps {
   onTouchStart?(): void;
 }
 
+interface CommonButtonProps {
+  id: ButtonProps['id'];
+  className: string;
+}
+
+interface InteractiveButtonProps
+  extends CommonButtonProps,
+    Pick<
+      ButtonProps,
+      | 'accessibilityLabel'
+      | 'onClick'
+      | 'onFocus'
+      | 'onBlur'
+      | 'onMouseEnter'
+      | 'onTouchStart'
+    > {
+  onMouseUp: MouseUpBlurHandler;
+}
+
 const DEFAULT_SIZE = 'medium';
 
 export function Button({
   id,
+  children,
   url,
   disabled,
+  external,
+  download,
+  submit,
   loading,
-  children,
+  pressed,
   accessibilityLabel,
   ariaControls,
   ariaExpanded,
@@ -111,8 +137,6 @@ export function Button({
   onKeyUp,
   onMouseEnter,
   onTouchStart,
-  external,
-  download,
   icon,
   primary,
   outline,
@@ -120,11 +144,9 @@ export function Button({
   disclosure,
   plain,
   monochrome,
-  submit,
   size = DEFAULT_SIZE,
   textAlign,
   fullWidth,
-  pressed,
   connectedDisclosure,
 }: ButtonProps) {
   const {newDesignLanguage} = useFeatures();
@@ -226,7 +248,6 @@ export function Button({
       </span>
     );
 
-  const type = submit ? 'submit' : 'button';
   const ariaPressedStatus = pressed !== undefined ? pressed : ariaPressed;
 
   const [disclosureActive, setDisclosureActive] = useState(false);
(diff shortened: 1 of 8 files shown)


src/components/UnstyledButton/UnstyledButton.tsx
```diff
@@ -6,20 +6,22 @@ import {UnstyledLink} from '../UnstyledLink';
...
   /** Forces url to open in a new tab */
   external?: boolean;
   /** Tells the browser to download the url instead of opening it. Provides a hint for the downloaded filename if it is a string value */
   download?: string | boolean;
+  /** Allows the button to submit a form */
+  submit?: boolean;
+  /** Disables the button, disallowing merchant interaction */
+  disabled?: boolean;
+  /** Replaces button text with a spinner while a background action is being performed */
+  loading?: boolean;
```
dfmcphee, line 24: Do you think this one makes sense here? I was debating this too. It does seem like it could be useful functionality to share, but also feels like a visual element that could behave strangely depending how the button is styled.

dfmcphee: Never mind, I see now this only handles the a11y portion in UnstyledButton. That makes a lot of sense to me.

src/components/Button/Button.tsx
```diff
@@ -1,38 +1,56 @@
...
 
 import styles from './Button.scss';
 
 type Size = 'slim' | 'medium' | 'large';
 type TextAlign = 'left' | 'right' | 'center';
 type IconSource = IconProps['source'];
 
-export interface ButtonProps {
+export interface ButtonProps
+  extends Pick<
```
dfmcphee, line 25: So I just tested this in the styleguide and it doesn't show any of these props in the docs with how this is setup. I think the better way to do this would be to have a `BaseButtonProps` interface that both `UnstyledButton` and `Button` extend for their props similar to how we do the [action types](https://github.com/Shopify/polaris-react/blob/master/src/types.ts#L86).

beefchimi: Incredible 🙏  thank you so much for testing this! I've gone ahead and pushed a new commit with your suggestion 😺

dfmcphee (approved):
This is looking really great now!

A few answers to your questions:

> (1) Do we really want `[key: string]: any;` as part of `UnstyledButtonProps`?
> 
> * I realize this is done so we can pass any arbitrary prop, such as `data-whatever`.
> * But is this a bit too free-wheelin'?

I think this is ok. We do the same thing on `UnstyledLink` as well. If you need to add a data attribute or aria attribute that we don't support, I think you should be able to with this component. Otherwise we run the risk of people still needing to use their own `<button>` still.



> (2) Should we pass `onClick` when `url`?
> 
> * Does this make sense? What is the execution order?

Ya, we need this for some link tracking and other events. `onClick` should take place before the navigation.



> (3) Why do we use `toBeFalsey` when a value is `undefined`? Should we not be more explicit?

Great question. I actually just fixed a bunch of these in web too. It would be much better to be more explicit about these. We actually have a tslint rule against it there.

> (4) Does the way I am extending `UnstyledButtonProps` within `Button.tsx` affect how descriptions get surfaced in the styleguide?

Just tested this in the style guide and it works great the way you have it setup now. 👍

PR #3495, opened 2020-10-15
Add type-check to build

BPScott:

`yarn run type-check` currently fails thanks to errors in test files. This is not picked up in CI as we don't currently run `type-check` on CI. We run `build` but the TypeScript compilation in that does not run over test files - as we don't want to ship d.ts files for tests.

We should check for these errors in test files on CI.

- Add running `yarn run type-check` to build
- Fix TS issues in test files so `yarn run type-check` passes.

## 

### How to 🎩

- Check CI fails on the first commit, that introduces type-check into CI
- Check CI passes on the second commit that fixes the issues

### DIFF
--- a/src/components/BulkActions/BulkActions.tsx
+++ b/src/components/BulkActions/BulkActions.tsx
@@ -6,7 +6,12 @@ import {CSSTransition, Transition} from 'react-transition-group';
 import {classNames} from '../../utilities/css';
 import {useI18n} from '../../utilities/i18n';
 import {clamp} from '../../utilities/clamp';
-import type {DisableableAction, Action, ActionListSection} from '../../types';
+import type {
+  BadgeAction,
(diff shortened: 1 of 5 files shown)


src/components/BulkActions/BulkActions.tsx
```diff
@@ -17,7 +22,7 @@ import {CheckableButton} from '../CheckableButton';
 import {BulkActionButton} from './components';
 import styles from './BulkActions.scss';
 
-type BulkAction = DisableableAction;
+type BulkAction = DisableableAction & BadgeAction;
```
BPScott, line 25: The test uses an action that is a BadgeAction. Looks like it was added deliberatly so make sure the actual types match.

AndrewMusgrave (approved):
:shipit:

PR #3517, opened 2020-10-19
Allow type override for components like StackItem

henrytao-me:

Fixes https://github.com/Shopify/app-extension-libs/issues/996

This change allows Argo to bypass default type check when rendering to avoid nested StackItem.

According to discussion here https://github.com/Shopify/app-extension-libs/issues/996#issuecomment-710437088, adding `__type__` check is a generic way to support Argo usage without having a Argo and Polaris depended on each other. 

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -8,6 +8,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Allow type override for components, ex: StackItem...
+
 ### Bug fixes
 
 ### Documentation
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -8,8 +8,6 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
-- Allow type override for components, ex: StackItem...
-
 ### Bug fixes
 
 ### Documentation
--- a/src/utilities/components.tsx
+++ b/src/utilities/components.tsx
@@ -44,7 +44,9 @@ export function isElementOfType<P>(
     return false;
   }
 
-  const {type} = element;
+  const {type: defaultType} = element;
+  const overrideType = element.props?.__type__;
+  const type = overrideType || defaultType;
   const Components = Array.isArray(Component) ? Component : [Component];
 
   return Components.some(
--- a/src/utilities/components.tsx
+++ b/src/utilities/components.tsx
@@ -45,6 +45,8 @@ export function isElementOfType<P>(
   }
 
   const {type: defaultType} = element;
+  // Type override allows components to bypass default wrapping behavior. Ex: Stack, ResourceList...
+  // See https://github.com/Shopify/app-extension-libs/issues/996#issuecomment-710437088
   const overrideType = element.props?.__type__;
   const type = overrideType || defaultType;
   const Components = Array.isArray(Component) ? Component : [Component];


alanthai (approved):
🎩 and working

src/utilities/components.tsx
```diff
@@ -44,7 +44,9 @@ export function isElementOfType<P>(
     return false;
   }
 
-  const {type} = element;
+  const {type: defaultType} = element;
```
vividviolet, line 47: Can we add a unit test?

UNRELEASED.md
```diff
@@ -8,6 +8,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Allow type override for components, ex: StackItem...
```
vividviolet, line 11: Do we want to publicly call this out? Should it be an internal feature?

src/utilities/components.tsx
```diff
@@ -44,7 +44,9 @@ export function isElementOfType<P>(
...
+  const overrideType = element.props?.__type__;
+  const type = overrideType || defaultType;
```
BPScott, line 49: Duplicating my comment on https://github.com/Shopify/remote-ui/pull/29

__type__ is very vague. The code in polaris-react isn't clear that this is due to argo-related functionality. I worry that in 3 month when people come to look at the code in polaris-react they won't have a clue what the extra logic is for and thus might be deleted.

Could this be named more explicitly? A longer name more explicit name would also reduce the potential for accidental naming clashes and thus accidental triggering of the logic in polaris-react. `__remoteUiChildType__` perhaps?

I'd also like to see a comment in here that describes what this does and why as it a very special one-off case and it is very important that in the future we understand why this exists so it doesn't get removed.

BPScott (changes requested):

src/utilities/components.tsx
alanthai: I think we wanted to keep it decoupled from remote UI itself. Not sure that anyone would want to piggy back. Although, we're trying to move away from this pattern in the future, so it might not be the worst to be very specific. +1 on the commenting though.

UNRELEASED.md
henrytao-me: Good call. I will skip the changelog for this PR.

src/utilities/components.tsx
henrytao-me: As Alan mentioned, we should eventually come up with something else. Can I keep the name and just put comment, link to the original issue instead? @BPScott

lemonmade: IMO, this is actually a Polaris issue that is not isolated to usage in Argo. The following React component has the same problem as Argo's usage of Polaris does, and could be solved with the same approach:

```tsx
const MyStackItem = ({children}) => <Stack.Item fill>{children}</Stack.Item>
```

Our eventual preference is to no longer rely on the existing implementation of `Stack` or other Polaris components that have this kind of child wrapping issue, so I think it's better if, in the Polaris codebase, we describe why we need this feature more generally than just "for Argo".

src/utilities/components.tsx
henrytao-me: There is no existing unit tests for components. I would prefer to skip it for now.

BPScott (approved):

PR #3527, opened 2020-10-20
Investigation junk for ThemeProvider

BPScott:

We've saw some memory leakage / high render time issues. Our hunch is this is due to the ThemeProvider's being expensive due to useMemo() stuff not actually memoising things properly - every render busts the memo cache so we end up doing all the work on every render instead of using a cached value. This also means we add a new item to useMemo's lookup table one every render too.

- High render time issues due to doing lots of computation on every render of a ThemeProvider
- High memory usage due to adding new data to a lookup table on every render of a ThemeProvider

You can kinda see this in a Playground. 

- Adds some debug logging so you can see the issue

### How to 🎩

Using the following playground:

- Disable "Strict mode" toolbar to reduce console noise from  strict mode triggering double renders (and thus doubling the amount of console logging)
- Click the first button
  - Note that `trigger processedThemeConfig`, `trigger buildCustomProperties` and `trigger buildThemeContext` logs are triggered each time you click the first button.

In ThemeProvider edit the dependencies on the useMemo that builds `processedThemeConfig` and remove `themeConfig` from the dependency array and repeat the above steps and note that no console logs are called each time you click the first button.  

This suggests that *there's something happening where `themeConfig` is being considered as changed between renders where it should have the same value*. This shouldn't happen.

If we can fix that then memory leaks should be solved. However to fixup higher render times we'll need to work out how to stop initial slowness. I think @martenbjork was saying buildThemeContext was where that slowness was.

IDEA: Having 3 interdependent useMemos feels like like overkill - could we refactor this code to be all placed in a single useMemo that returns both `theme` and `style`?

TODO: Once we fix this we should investigate if the useMemo cache is shared between multiple instances of a ThemeProvider so that we can utilize even more caching. this might involve abstracting the "working out all the theme stuff" out into a hook. (I thinks this might mean that the cache can be shared between multiple renderings of ThemeProvider if it isn't already)

```

import {Page, ThemeProvider, Button} from '../src';

  const [baseTheme, setBaseTheme] = useState<any>('light');

  const toggle = useCallback(() =>
    setBaseTheme(baseTheme == 'light' ? 'dark' : 'light'),
  );

  return (
    <Page title="Playground">
      <Button onClick={toggle}>First</Button>
      <ThemeProvider theme={{colorScheme: baseTheme}}>
        <Button>Second</Button>
        <ThemeProvider theme={{colorScheme: 'inverse'}}>
          <Button>Third</Button>
        </ThemeProvider>
      </ThemeProvider>
    </Page>
  );
}
```

### DIFF
--- a/src/components/ThemeProvider/ThemeProvider.tsx
+++ b/src/components/ThemeProvider/ThemeProvider.tsx
@@ -40,12 +40,12 @@ export function ThemeProvider({
   const isParentThemeProvider = parentContext === undefined;
 
   const processedThemeConfig = useMemo(() => {
-    console.log(
-      'trigger processedThemeConfig',
-      parentContext,
-      themeConfig,
-      isParentThemeProvider,
-    );
+    // console.log(
+    //   'trigger processedThemeConfig',
+    //   parentContext,
+    //   themeConfig,
+    //   isParentThemeProvider,
+    // );
 
     const parentColorScheme =
       parentContext && parentContext.colorScheme && parentContext.colorScheme;
@@ -66,7 +66,7 @@ export function ThemeProvider({
   }, [parentContext, themeConfig, isParentThemeProvider]);
 
   const customProperties = useMemo(() => {
-    console.log('trigger buildCustomProperties');
+    // console.log('trigger buildCustomProperties');
 
     return buildCustomProperties(
       processedThemeConfig,
@@ -81,7 +81,7 @@ export function ThemeProvider({
   const color = customProperties['--p-text'] || '';
 
   const theme = useMemo(() => {
-    console.log('trigger buildThemeContext');
+    // console.log('trigger buildThemeContext');
 
     return {
       ...buildThemeContext(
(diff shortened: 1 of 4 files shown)


src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -87,7 +102,7 @@ export function ThemeProvider({
   const style = {...customProperties, ...(!isParentThemeProvider && {color})};
 
   return (
-    <ThemeContext.Provider value={{...theme, textColor: color}}>
```
BPScott, line 90: Previously we stored a new value in ThemeContext on every render. This stops `parentContext` in the above useMemo dependencies  changing on every update, so now it's just `themeConfig` breaking the memoisation.

BPScott:
FAO @kyledurand @dfmcphee @martenbjork @alex-page

alex-page:
Thanks @BPScott this branch was very useful!

PR #3529, opened 2020-10-20
[UnstyledLink] Add new window warning to unstyled link, pass translation to custom Link

### LINKED ISSUE #3476: [a11y] External unstyled link causes change of context without warning
## Issue summary

Related to: https://github.com/Shopify/polaris-ux/issues/146

A solution to this issue was already found and deployed for the `<Link>` component (https://github.com/Shopify/polaris-react/pull/1247), but the problem persists for a `<UnstyledLink>` (https://github.com/Shopify/polaris-react/blob/master/src/components/UnstyledLink/UnstyledLink.tsx)

## Expected behavior

If a `<UnstyledLink>` is going to open in a new tab, then a screen reader should notify the user that the link will open in a new tab.

## Actual behavior

The user is not notified that the link will open in a new tab.

## Specifications

- Polaris version number: 5.7.0

## Fix?

Since an Unstyled Link likely should not have [the external icon, like the Link has](https://github.com/Shopify/polaris-react/pull/1247/files#diff-54ce4d51798a73c9699af609ffeb83195ebdefc49ce0ebec4f44432d97c383b1R43), then it should have some `<VisuallyHidden>` text instead:

```
<VisuallyHidden>
  <span>{ polaris.intl.translate('Polaris.Common.newWindowAccessibilityHint') }</span>
</VisuallyHidden>
```

mzlock:

Fixes https://github.com/Shopify/polaris-react/issues/3476

When a link will open in a new tab, a screen-reader user should be warned.

- Adds a visually hidden warning when an `UnstyledLink` is an external link
- If the Polaris instance has a custom `LinkComponent`, pass the translated `newWindowAccessibilityHint` text to that custom link component, so it can provide the same warning

### How to 🎩

Create a `UnstyledLink` with the prop `external: true`. This will render a link with `target="_blank" rel="noopener noreferrer"`. Inside of the link should be a `<span class="visually-hidden">{ warningText }</span>`.

@BPScott - what's the best way to top-hat Polaris changes like this?


BPScott:
Heya @mzlock, I've thought about this a bit more and I don't think this is the right direction. Two things jump out at me.

First, the `LinkComponent` that is passed into app provider and referenced in `UnstyledLink` from is expected to be basically an extension of a `<a>` element and we've always taught people this though [its interface](https://github.com/Shopify/polaris-react/blob/00808ec3ab893a39af5ebc9b52a525be8916d227/src/utilities/link/types.ts) and [examples](https://github.com/Shopify/polaris-react/blob/master/src/components/AppProvider/README.md#with-linkcomponent)

Passing a prop into LinkComponent that isn't a valid attribute for an `<a>` tag is going to result in loooots of invalid html for people who are already using this link component and are spreading the props into it as in our example usage - their HTML willl end up including a bunch of `<a href="BLAH" newWindowHint="(opens a new window)" >...` which isn't good.

Because of that I don't think the passing this prop through is a good idea. Web can duplicate our translation string.

---

Secondly, Polaris already specify the "opens in a new window)" screen-readable text within Link - [it's attached to the aria-label of the icon](https://github.com/Shopify/polaris-react/blob/344c23bb087a1148fe6380f7c7e7c4261acc2e2e/src/components/Link/Link.tsx#L41-L43).

By adding this text to the UnstyledLink as well we end up duplicating this text which is undesired. You can see this in action in the in [the external link story in Storybook for this PR](https://5d559397bae39100201eedc1-hgnqzxkfqg.chromatic.com/?path=/story/all-components-link--external-link). The source for this looks like the following, note the duplication in VisuallyHidden and in the aria-label:

Screen Shot 2020-10-22 at 11 19 31 AM

We could "fix" this by removing that label text from the Link icon and saying it is now the responsibility of UnstyledLink / the LinkComponent people provide but that feels too scary to do in a minor release - the risk of creating a worse experience for lots of people feels super high. If we did that, we should do it in a major release and clearly communicate what people need to change for an accessible experience. As a tradeoff this would mean that this accessibility is now at the mercy of consuming apps. Would we rather to the right thing and be accessible in 85% of cases automatically (by keeping this how it is now), or risk people not handling this label in their app and thus be not accessible 100% of the time?

BPScott:
Would be interested to hear @dpersing, @svinkle and @dfmcphee's takes on this.

dpersing:
@BPScott I'd recommend splitting the difference here. I'd like to support the existing `external` prop for any and all versions of the [Link component](https://polaris.shopify.com/components/navigation/link), since it displays an icon by default (this helps keyboard users and users with cognitive issues). However, instead of using `aria-label`, which is iffy on non-focusable elements like the SVG in question, use @mzlock's recommendation of just usually visually hidden text as part of the link text. The icon can be treated as "decorative".

BPScott:
Cool, I'm happy with that direction, but we should wait for a major version as this is going to be a breaking change in how we expect people to work with LinkComponent.

BPScott:
pinging @alex-page  and @kyledurand  as this can be something for Polaris v6

dpersing:
Any insights into why the support wasn't included through app provider initially?

BPScott:
Not a clue, sorry https://github.com/Shopify/polaris-react/pull/1247 is the PR where it was added

kyledurand (approved):

alex-page (approved):
Looks great to me, I have created a `v6.0.0` branch and pointed this PR at it. Thanks a ton @mzlock keep making these awesome PR's we really appreciate them.

BPScott (changes requested):
This needs to remove the newWindowHint message from the `Link` component so that we don't have duplicate "(open in new window)" lines read out from Link and then UnstyledLink directly inside it.

It will also need documentation to explain to consumers that they need to update their AppProvider linkComponent to handle the newWindowHint - they need to consume it and render it as hidden text, inistead of just passing the prop into into  an `a` tag / a react-router/next.js Link component etc

This might also be a good time to revisit the type of LinkComponent as its ergonomics are still a bit rubbish as casting always seems to be required when consuming with react-router / next/js's router.

mzlock:
@BPScott - I removed that prop. So, just confirming, each app that uses a custom `LinkComponent` should manage their own translations for the new window hint in their custom `LinkComponent`?

Any other changes that you need me to push up for this?

alex-page:
@mzlock this is looking good to me @BPScott any final thoughts?

BPScott:
Chatted with @alex-page:

As we want to get v6 out the door ASAP we're not accepting extra new stuff that may result in breaking changes to consuming apps at this stage.

This shall end up in v7. Sorry.

alex-page:
We are investigating links, external links and much more. Going to close this PR as it is out of date and share it with the team. Thanks @mzlock

PR #3533, opened 2020-10-20
memoize buildCustomProperties function

tmlayton:

This function takes somewhere between 5 and 10ms on a fast machine in a sandbox, multiply that by many theme providers, slow machines, and production environments and times go up significantly. 

Memoizing `buildCustomProperties`

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

* [x] 🎩

### DIFF
--- a/src/utilities/theme/utils.ts
+++ b/src/utilities/theme/utils.ts
@@ -212,13 +212,13 @@ function parseColors([baseName, colors]: [
 }
 
 function memoize(fnToMemoize: Function) {
-  const cache: {[key: string]: any} = {};
+  const cache: Map<string, any> = new Map();
   return function (...args: any[]) {
-    const key = [fnToMemoize.name, JSON.stringify(args)].join('|');
-    if (!cache[key]) {
-      cache[key] = fnToMemoize(...args);
+    const key = JSON.stringify([fnToMemoize.name, args]);
+    if (cache.get(key) === undefined) {
+      cache.set(key, fnToMemoize(...args));
     }
-    return cache[key];
+    return cache.get(key);
   };
 }
 
--- a/src/utilities/theme/utils.ts
+++ b/src/utilities/theme/utils.ts
@@ -21,7 +21,7 @@ interface CustomPropertiesConfig extends ThemeConfig {
   colorScheme: ColorScheme;
 }
 
-export function buildCustomProperties(
+export function buildCustomPropertiesNoMemo(
   themeConfig: CustomPropertiesConfig,
   newDesignLanguage: boolean,
   tokens?: Record<string, string>,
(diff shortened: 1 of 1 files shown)


BPScott:
How does this work with server-side rendering where the server is a long running process? Will the map keep getting larger over time without ever freeing up any memory?

[The one place buildCustomProperties is used](https://github.com/Shopify/polaris-react/blob/master/src/components/ThemeProvider/ThemeProvider.tsx#L60-L64) is is already wrapped in a `useMemo` (which does handle occasionally throwing out old value to release memory pressure) so it doesn't seem like this adds much value.


https://github.com/Shopify/polaris-react/pull/3527 demonstrates that the useMemo around is buildCustomProperties is always cachebusted because because the memo wrapped around processedThemeConfig is messed up. By fixing the dependencies of the call to useMemo in ThemeProvider as detailed in that PR we sidestep needing to add this 4th! set of memoisation.

dfmcphee:
> [The one place buildCustomProperties is used](https://github.com/Shopify/polaris-react/blob/master/src/components/ThemeProvider/ThemeProvider.tsx#L60-L64) is is already wrapped in a `useMemo` (which does handle occasionally throwing out old values to release memory pressure) so it doesn't seem like this adds much value.

The difference is that `useMemo` will only memoize it for that component instance, and since we have ThemeProviders all over the page, they won't share that cached value unless we move that up to a higher level component. Really we only want to ever run that function for the default values once for the default light and dark theme configs (we don't even use the customization right now). This could even be in a build process as it should never change dynamically.

BPScott:
Does useMemo still act that way if you extract the call into a hook (which is then used by multiple instances of a single component)?

I was wondering about that last night: 

> TODO: Once we fix this we should investigate if the useMemo cache is shared between multiple instances of a ThemeProvider so that we can utilize even more caching. this might involve abstracting the "working out all the theme stuff" out into a hook. (I thinks this might mean that the cache can be shared between multiple renderings of ThemeProvider if it isn't already)

alex-page (approved):

tmlayton:
> How does this work with server-side rendering where the server is a long running process? Will the map keep getting larger over time without ever freeing up any memory?

No, it won’t grow to more than a handful of values so should be fine.

> Does useMemo still act that way if you extract the call into a hook (which is then used by multiple instances of a single component)?

I don't know—if it does I'm in favor of that approach and we can use this until then.

BPScott:
Had a play with "will hooks let us share between component instances" and the answer is no: https://codesandbox.io/s/youthful-andras-3ltyt?file=/src/App.js. It still ends up memoizing once per component instance rather than sharing between items (if it shared between component instances I'd expect calledHook to only be triggered once).

All very good. Ignore me :)

tmlayton:
Thanks for looking into it @BPScott! I certainly would have preferred if we could have done this the React way.

Issue #3448, opened 2020-10-09
[ThemeProvider] Don't output variables if colorScheme is the same as parent context


martenbjork:
Scenario:

— Parent: User-controlled color scheme.
— Child: Always dark.

If the parent switches to dark, you will have identical variables repeated twice which seems suboptimal.

PR #3550, opened 2020-10-22
Optimize ThemeProvider custom properties output

dfmcphee:
This is an alternative approach to https://github.com/Shopify/polaris-react/pull/3546.

Currently we output the custom properties on each level of `ThemeProvider` even if they are identical to the parent.

This changes the ThemeProvider to compare the generated custom properties with the parent context and only output them if they differ.

### How to 🎩

I committed a change to the Details Page example to try this out. If you change to the new design language light mode, you should see only one top level output of the custom properties since the AppProvider's ThemeProvider and the ThemeProvider wrapping the Details Page frame match. If you switch to dark mode you should see it output on both and that the page stays in light mode because of the nested ThemeProvider's custom properties.

### 🎩 checklist

### DIFF
--- a/src/components/ThemeProvider/ThemeProvider.tsx
+++ b/src/components/ThemeProvider/ThemeProvider.tsx
@@ -6,6 +6,7 @@ import {
   ThemeConfig,
   buildThemeContext,
   buildCustomProperties,
+  toString,
   Tokens,
 } from '../../utilities/theme';
 import {useFeatures} from '../../utilities/features';
@@ -84,7 +85,18 @@ export function ThemeProvider({
     }
   }, [backgroundColor, color, isParentThemeProvider]);
 
-  const style = {...customProperties, ...(!isParentThemeProvider && {color})};
+  let style;
+
+  if (isParentThemeProvider) {
+    style = customProperties;
+  } else if (
+    !isParentThemeProvider &&
+    parentContext!.cssCustomProperties !== toString(customProperties)
+  ) {
+    style = {...customProperties, ...{color}};
+  } else {
+    style = {color};
+  }
 
   return (
     <ThemeContext.Provider value={{...theme, textColor: color}}>
--- a/src/components/ThemeProvider/ThemeProvider.tsx
+++ b/src/components/ThemeProvider/ThemeProvider.tsx
@@ -101,13 +101,14 @@ export function ThemeProvider({
     !isParentThemeProvider &&
     parentContext!.cssCustomProperties !== toString(customProperties)
   ) {
-    style = {...customProperties, ...{color}};
+    customProperties.color = color;
+    style = customProperties;
   } else {
     style = {color};
   }
 
   return (
-    <ThemeContext.Provider value={{...theme}}>
+    <ThemeContext.Provider value={theme}>
       <div style={style}>{children}</div>
     </ThemeContext.Provider>
   );
--- a/src/components/ThemeProvider/ThemeProvider.tsx
+++ b/src/components/ThemeProvider/ThemeProvider.tsx
@@ -101,8 +101,7 @@ export function ThemeProvider({
     !isParentThemeProvider &&
     parentContext!.cssCustomProperties !== toString(customProperties)
   ) {
-    customProperties.color = color;
-    style = customProperties;
+    style = {...customProperties, ...{color}};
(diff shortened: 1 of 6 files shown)


kyledurand (approved):
Love it 🥇 

Makes things much less confusing for consumers

BPScott:
Tangental: In [#3527](https://github.com/Shopify/polaris-react/pull/3527/files#diff-392ad55aee8e3d195de49a37182b22b9bfde6d4915a8a01fa57b013cab64e437R87) I identified that we're recreating the object we store in the ThemeContext on every rerender, and found a smallish fix for it (with a further suggesting that we should replace those multiple useMemo calls with a single useMemo call). Is it worth working at caching this response here or should we deal with that in a separate PR?

dfmcphee:
> Is it worth working at caching this response here or should we deal with that in a separate PR?

Good call @BPScott. I want to do a bit of perf testing with this change so it would make sense to try to get that fixed in here as well so we can test it all together.

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -84,10 +93,21 @@ export function ThemeProvider({
...
+    parentContext!.cssCustomProperties !== toString(customProperties)
+  ) {
+    style = {...customProperties, ...{color}};
+  } else {
+    style = {color};
+  }
 
   return (
-    <ThemeContext.Provider value={{...theme, textColor: color}}>
+    <ThemeContext.Provider value={{...theme}}>
```
BPScott, line 110: Now merging textColor in has been solved up there we don't need to spread and thus create a whole new object anymore

```suggestion
    <ThemeContext.Provider value={theme}>
```

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -84,10 +93,21 @@ export function ThemeProvider({
...
-  const style = {...customProperties, ...(!isParentThemeProvider && {color})};
+  let style;
+
+  if (isParentThemeProvider) {
+    style = customProperties;
+  } else if (
+    !isParentThemeProvider &&
+    parentContext!.cssCustomProperties !== toString(customProperties)
+  ) {
+    style = {...customProperties, ...{color}};
```
BPScott, line 104: Rather than create an new object, perhaps we could assign to the already existing one

```suggestion
    customProperties.color = color;
    style= customProperties;
```

dfmcphee: This ones throws an error when modified. `Uncaught TypeError: Cannot assign to read only property 'color' of object '#<Object>'`. I believe it is because `customProperties` is the memoized value above and is now readonly. I guess we do have to create a new object here to add it.

dfmcphee:
Alrighty, this should be ready for a real review now. Following a similar approach to what @BPScott did in #3527 I was able to get this memoizing properly.

One thing I noticed was that the parent context from the `AppProvider` was always changing in storybook, but when I moved to a new create-react-app to test, it didn't have that issue.

I tested in that app with something like this:

```

import {Page, ThemeProvider, Button, Card} from '@shopify/polaris';
const FIRST_THEME = {colorScheme: 'light'};

export default function App() {
  const [baseTheme, setBaseTheme] = useState('light');

  const toggle = () => setBaseTheme(baseTheme === 'light' ? 'dark' : 'light');

  const secondTheme = {colorScheme: baseTheme};
  return (
    <Page title="Playground">
      <Button onClick={toggle}>Toggle child theme</Button>
      <ThemeProvider theme={FIRST_THEME}>
        <Card sectioned>First themed card</Card>
        <ThemeProvider theme={secondTheme}>
          <Card sectioned>Second themed card</Card>
        </ThemeProvider>
      </ThemeProvider>
    </Page>
  );
}
```

The 3 theme providers have their values memoized on first render and when clicking the toggle button, only the most nested provider reruns the memoization.

The custom properties also only output when the `secondTheme` is in dark mode.

dfmcphee:
One other thing I noticed in that example. If you declare `firstTheme` inside the `App` function, or declare it inline in the props, it will also try to rememoize every render because that object gets recreated each time. This is something we should look out for in web if we can too.

dfmcphee:
This fixes #3448

Issue #3017, opened 2020-05-29
subnavigationitems onclick doesn't work


medst:
## Issue summary

subnavigationitems onclick doesn't work

## Expected behavior

i create subnavigation items, but when i click on subnavigation item, the onClick method do not work, it must print 3434 on screen, but nothing happen

code :

import React from 'react';
import {Navigation} from '@shopify/polaris';
import {HomeMajorMonotone, OrdersMajorTwotone, ProductsMajorTwotone} from '@shopify/polaris-icons';

export default class App extends React.Component {
  
  hey(e){
    console.log(3434)
  }
  
  render() {
    return (
      <Navigation location="/">
  <Navigation.Section
    items={[
      {
        url: '/',
        label: 'Home',
        icon: HomeMajorMonotone,
        subNavigationItems: [
          {
            url: '/',
            label: 'home 2',
            onClick: this.hey
          }
        ],
        onClick: this.hey
      },
      {
        url: '/orders',
        label: 'Orders',
        icon: OrdersMajorTwotone,
        badge: '15',
      },
      {
        url: '/hey',
        label: 'Products',
        icon: ProductsMajorTwotone,
      },
    ]}
  />
</Navigation>
    );
  }
}

## codesandbox url

https://codesandbox.io/s/trusting-lake-5xl3g?fontsize=14&hidenavigation=1&theme=dark

## Specifications

- Are you using the React components? (Y/N): Y
- Polaris version number:  4.23
- Browser: chrome
- Device: pc
- Operating System: windows 10

medst:
any help ?

PR #3560, opened 2020-10-24
[Navigation] Fix SubNavigationItem onClick

### LINKED ISSUE #3150: Subnavigation's `onClick` not work
## Issue summary
Subnavigation's `onClick` not work.
say code like this
```
subNavigationItems: [
                {
                  url: "/profile",
                  label: "profile",
                  onClick: e => {
                    e.defaultPrevented();
                    alert(1);
                    console.log(1);
                  }
                }]
```

## Expected behavior
get `alert` and `console.log`, rather than anchor tag like behavior.

## Reduced test case

[CodeSandbox Example](https://codesandbox.io/s/kind-leaf-7rt9v?file=/App.js)

## Specifications

- Polaris version number: ^4.20.1
- Browser: Google Chrome
- Device: Mac
- Operating System: macOS

dkershner6:

Fixes #3150 #3017 

The onclick handler does nothing, despite the prop being present.

- The onNavigationDismiss in the SubNavigationItem was removed, as it is handled by the main Item. This is confirmed through a test, and when I called it in the subitem, it got called twice.
- Test is added so this does not regress.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -10,6 +10,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- OnClick handling now functions for SubNavigationItem
+
 ### Documentation
 
 ### Development workflow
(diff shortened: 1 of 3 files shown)


alex-page:
Hey @dkershner6 we are trying to keep the events passed in opinionated and that is why we don't offer this flexibility. If you think this is an issue let me know.

PR #3561, opened 2020-10-26
🚚 [Utilities] Move isNewDesignLanguageColor to a new file

beefchimi:
Just happened to notice this while working on some other code. I'm moving a utility function out of `/types` and into its own `/utilities` file.


beefchimi:
@dleroux perhaps you could take a quick peek at this?

dleroux (approved):

PR #3566, opened 2020-10-26
Update sewing-kit and typescript

BPScott:

Keeping up to date, prepping for the future

Updates sewing-kit to 0.140.0
Updates typescript to 4.0.3

The typescript bump lets us cache TS type info when running type-checks without emiting any d.ts files - this means repeated runs of `yarn run type-check` are faster as it can use the cache. Repeated runs now take ~7s instead of ~20s).
The SK update pulls us onto newest major versions of eslint 7.x and stylelint 13.7.x. In particular this means better support for optional chaining and we fix a few false positives in react hooks lints.

### How to 🎩

Ensure tests pass

### DIFF
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -18,7 +18,7 @@ import {Connected} from '../Connected';
 import {Error, Key} from '../../types';
 import {Icon} from '../Icon';
 
-import {Resizer, Spinner} from './components';
+import {Resizer, Spinner, SpinnerProps} from './components';
 import styles from './TextField.scss';
 
 type Type =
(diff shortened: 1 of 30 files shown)


src/components/TextField/TextField.tsx
```diff
@@ -315,15 +315,15 @@ export function TextField({
...
+  const handleButtonPress: SpinnerProps['onMouseDown'] = useCallback(
+    (onChange) => {
       const minInterval = 50;
       const decrementBy = 10;
       let interval = 200;
 
       const onChangeInterval = () => {
         if (interval > minInterval) interval -= decrementBy;
-        onChange();
+        onChange(0);
```
BPScott, line 326: Fixing up the types in Spinner highlighted that this was wrong - onChange should take a number as an argument. This is still not doing what we'd hope the code should do - holding the step buttons doesn't increment the value - but at least it is properly typed now. I briefly tried fixing this but didn't get it working. I think a proper fix would be better as a separate PR.

alex-page (approved):

alex-page:
Thanks @BPScott looks like there was some lint/test failures but the changes look good.

PR #3576, opened 2020-10-28
[Page actions] Only measure actions on horizontal resize

kyledurand:

Fixes https://github.com/Shopify/polaris-react/issues/3575

Checking the window width before measuring actions and updating the layout

## 

### How to 🎩

Details page. Add content, resize the screen vertically. There should be no content jump

It might help to add a console log on line 115 of Actions.tsx to see when updates are being made

### 🎩 checklist


alex-page (approved):

kyledurand:
On closer inspection this won't work. The useEffect calling the measurer is running on every keystroke regardless of whether or not that tiggers a page resize. I'll have to rethink this approach

kyledurand:
Going to close this for now as it doesn't seem to address the issue at hand. Digging deeper into it the problem seems to live on the web side. Some higher order function is triggering re-render of subcomponents. I'll have to revisit this next week

PR #3582, opened 2020-10-29
Replace Travis with GitHub actions

alex-page:

Removes travis which is taking on average ~50minutes to complete each build with GitHub actions which is taking ~10minutes.

- [x] Reduces run time of checks to around [~10minutes](https://github.com/alex-page/polaris-react/runs/1324328263)
- [x] Caches yarn dependencies and prefers them
- [x] Moves our secrets for percy and chroma to https://github.com/Shopify/polaris-react/settings/secrets
- [x] Use `process.env.GITHUB_SHA` for shrink ray
- [ ] Replace shrink rays `process.env.TRAVIS_JOB_WEB_URL`

### Tophatting

Check the actions running in my fork https://github.com/alex-page/polaris-react/actions


.github/workflows/test-on-push.yml
```diff
@@ -0,0 +1,165 @@
...
+      id: yarn-cache-dir-path
+      run: echo "::set-output name=dir::$(yarn cache dir)"
+    - uses: actions/cache@v2
+      id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
+      with:
+        path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+        key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-yarn-
+    - run: yarn --prefer-offline
```
alex-page, line 24: This is repeated for each job. We could run [matrixes](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategy) to clean this up for each script that needs to be ran.

README.md
```diff
@@ -1,6 +1,6 @@
 # [Polaris React](https://polaris.shopify.com/)
 
-[![npm version](https://img.shields.io/npm/v/@shopify/polaris.svg?style=flat)](https://www.npmjs.com/package/@shopify/polaris) [![Travis build status](https://travis-ci.com/Shopify/polaris-react.svg?branch=master)](https://travis-ci.com/Shopify/polaris-react) [![codecov](https://codecov.io/gh/Shopify/polaris-react/branch/master/graph/badge.svg?token=IKyeKcpRs1)](https://codecov.io/gh/Shopify/polaris-react) [![PRs Welcome](https://img.shields.io/badge/PRs-welcome-brightgreen.svg)](https://github.com/Shopify/polaris-react/blob/master/.github/CONTRIBUTING.md#your-first-pull-request)
```
alex-page, line 3: I still need to add something here to replace travis. Or do we not feel like that is valuable?

dev.yml
```diff
@@ -2,7 +2,7 @@ name: polaris-react
 up:
   - node:
       yarn: v1.13.0
-      version: v10.15.0 # to be kept in sync with .nvmrc and .travis.yml
+      version: v10.15.0 # to be kept in sync with .nvmrc
```
alex-page, line 5:
```suggestion
      version: v10.15.0 # to be kept in sync with .nvmrc and .github/workflows/test-on-push.yml
```

dev.yml
```diff
@@ -2,7 +2,7 @@ name: polaris-react
 up:
   - node:
       yarn: v1.13.0
-      version: v10.15.0 # to be kept in sync with .nvmrc and .travis.yml
+      version: v10.15.0 # to be kept in sync with .nvmrc and .github/workflows/test-on-push.yml
```
alex-page, line 5:
```suggestion
      version: v10.15.0 # to be kept in sync with .nvmrc
```

.github/workflows/test-on-push.yml
```diff
@@ -0,0 +1,165 @@
+name: Super ultra fast test, lint and visual regression
```
alex-page, line 1: What should it be called?

shipit.yml
```diff
@@ -1,6 +1,6 @@
 ci:
   require:
-    - 'Travis CI - Branch'
+    - ''
```
alex-page, line 3:
```suggestion
    - 'Super ultra fast test, lint and visual regression'
```

shipit.yml
```diff
@@ -12,7 +12,7 @@ ci:
 
 merge:
   require:
-    - 'Travis CI - Branch'
+    - ''
```
alex-page, line 15:
```suggestion
    - 'Super ultra fast test, lint and visual regression'
```

.github/workflows/test-on-push.yml
```diff
@@ -0,0 +1,165 @@
...
+
+on: push
+
+jobs:
+  type_check:
+    name: 'Type Check and Build'
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
```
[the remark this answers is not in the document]
BPScott: Alternatively: Rather than spinning up 5 vms - one per job what if we combine some of these jobs into one - running linting, testing and building sequentially in a single job instead of one thing per VM.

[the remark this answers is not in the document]
alex-page: Yeah I wanted to combine them but was unsure which tests could be combined. I'll definitely do that ^

.github/workflows/test-on-push.yml
```diff
@@ -0,0 +1,165 @@
+name: Super ultra fast test, lint and visual regression
+
+on: push
```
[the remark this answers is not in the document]
alex-page: Do you have a suggestion on how this should work?

kyledurand:
This sounds amazing. Is there a way I can tophat this? I don't have much context around creating this type of config

.github/workflows/ci-on-push.yml
```diff
@@ -0,0 +1,71 @@
...
+      id: yarn-cache-dir-path
+      run: echo "::set-output name=dir::$(yarn cache dir)"
+    - uses: actions/cache@v2
+      id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
+      with:
+        path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+        key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-yarn-
+    - run: yarn --prefer-offline
```
alex-page, line 21: We could turn this into a `uses` and point to a local action to stop duplication.

.github/workflows/ci-on-push.yml
```diff
@@ -0,0 +1,71 @@
...
+    - run: yarn run type-check && yarn run build
+    - run: yarn run lint
+    - run: yarn run check:custom-properties
+    - run: yarn run test --coverage && yarn run codecov
+    - run: node ./scripts/build-shrink-ray.js
```
alex-page, line 26: This runs synchronously, I experimented with matrices but that creates multiple containers.

.github/workflows/ci-on-push.yml
```diff
@@ -0,0 +1,71 @@
...
+    - env:
+        CHROMATIC_PROJECT_TOKEN: ${{ secrets.CHROMATIC_TOKEN }}
+      run: yarn run chromatic publish --build-script-name='storybook:build'
```
alex-page, line 29: I was thinking of adding this as the last step, I could separate it and make sure it `needs` test and lint to run.

.github/workflows/ci-on-push.yml
```diff
@@ -0,0 +1,71 @@
...
+        restore-keys: |
+          ${{ runner.os }}-yarn-
+    - run: yarn --prefer-offline
+    - env:
+        STORYBOOK_DISABLE_DOCS: 1
+      run: |
+        yarn run storybook:build --quiet
+        node ./scripts/accessibility-check.js
+
+  visual_regression:
```
alex-page, line 53: We could also add `needs` here to make sure test and lint completes before running the visual regression. It might slow things down.

shipit.yml
```diff
@@ -12,7 +14,8 @@ ci:
 
 merge:
   require:
-    - 'Travis CI - Branch'
+    - 'Test, lint and deploy'
+    - 'Accessibility Tests'
```
alex-page, line 18: I feel like we don't need visual regression for this step in shipit. Should it be added to ignore?

.github/workflows/test-on-push.yml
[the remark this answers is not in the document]
BPScott: Not sure, would need to investigate. It might be a combination of push and pull_request. Note how the "push" job has triggered actions alex-page/polaris-react but there's been no action triggered within the context of this repo. I think that means we wouldn't be able to enforce CI passing on PRs from forks atm as nothing gets mentioned in the list of checks at the bottom of this PR.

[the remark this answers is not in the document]
alex-page: Yeah I tested this and you are right. I created a fork of [alex-page/harmonograph-xy/](https://github.com/alex-page/harmonograph-xy/) and the pull request did not trigger the `on: push`.

.github/workflows/ci-on-push.yml
BPScott: I think synchronous is fine. Each of these steps are pretty quick to run, we don't save much time by making them run in parallel

typecheck & build: 1m26s
lint: 1m53s
custom properties check: 8s
tests with coverage: 3m6s
shrink ray: 1m36s

That adds up to 8mins (with a cold cache, we could get some speedups by caching the build folder I think). If we did one thing per container (as travis does now) then total time is the time of the slowest item: 3mins. Which sounds good but meanwhile the a11y tests still take 10mins so we'd still be waiting for that even after all these run in parallel.

I don't think the extra complication and extra resources (1 action container vs 5) is worth the time "saved".


---

Question: Does a step item still run if the prior item failed? For instance currently we run linting then tests - if linting fails do the tests still run. I'm wondering if I commit something that fails both tests and linting am i alerted to both failures or just the linting?

alex-page: No they all stop if one fails. Same with matrices. The only way to get around that is separate containers from my understanding.

alex-page: The other option is making multiple `.yml` files and treating them as individual workflows.

BPScott: Looks like you can modify that behaviour: https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#job-status-check-functions

```yaml
- if: ${{ !cancelled() }}
  run: yarn run type-check && yarn run build
- if: ${{ !cancelled() }}
  run: yarn run lint
- if: ${{ !cancelled() }}
  run: yarn run check:custom-properties
// etc
```

alex-page:
Closing for https://github.com/Shopify/polaris-react/pull/3739

PR #3583, opened 2020-10-29
[Page actions] Always render single action

kyledurand:

Fixes https://github.com/Shopify/polaris-react/issues/3581

https://screenshot.click/screencast_2020-10-29_09-16-24.mp4

If there is only one action or group, always render it before rolling up to a default menu group

### How to 🎩

~~I'll add a chroma link~~ I can't because I can't modify the details page 🤦 

On the details page: Remove all secondary actions except for duplicate, remove both action groups. 
Make sure that only duplicate appears and not the `More actions` dropdown

### 🎩 checklist

### DIFF
--- a/src/components/ActionMenu/components/Actions/Actions.tsx
+++ b/src/components/ActionMenu/components/Actions/Actions.tsx
@@ -72,20 +72,24 @@ export function Actions({actions = [], groups = []}: Props) {
       return;
     }
 
+    const actionsAndGroups = [...actions, ...groups];
+
+    if (actionsAndGroups.length === 1) {
+      setShowableActions(actionsAndGroups);
+      return;
+    }
+
     let currentAvailableWidth = availableWidthRef.current;
     let newShowableActions: MenuActionDescriptor[] = [];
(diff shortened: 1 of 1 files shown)


alex-page (approved):

src/components/ActionMenu/components/Actions/Actions.tsx
```diff
@@ -73,17 +73,19 @@ export function Actions({actions = [], groups = []}: Props) {
...
 
     let currentAvailableWidth = availableWidthRef.current;
-
     let newShowableActions: MenuActionDescriptor[] = [];
     let newRolledUpActions: (MenuActionDescriptor | MenuGroupDescriptor)[] = [];
+    const actionsAndGroups = [...actions, ...groups];
 
-    [...actions, ...groups].forEach((action, index) => {
+    actionsAndGroups.forEach((action, index) => {
+      const isSingleAction = actionsAndGroups.length === 1;
```
dfmcphee, line 81: Do we even need to do this loop at all if there is only one action? Could we instead just set it as the showable action in that case?

alex-page (approved):

PR #3590, opened 2020-10-30
✨ [Button] Add new role prop

beefchimi:
Following up on https://github.com/Shopify/polaris-react/pull/3494 - I am adding a `role` prop to both the `UnstyledButton` and `Button` components.

We utilize this in `online-store-ui` to set a `menuitem` value for our `RTE > Toolbar` items.

We could have gotten away with passing any arbitrary prop to `UnstyledButton`, since it accepts `[key: string]: any;` in its `Props` interface... but! We do apply a `role="alert"` if `loading`... so it is best to explicitly define a `role` prop on the interface so that we can pass a custom value.

### Questions

- Do I need to update the `Button > README` to include some information on this new `role` prop?
- Should we pass `role` to `UnstyledLink`?
  - I ask because we did not set `role="alert"` when `loading`
  - If we disallow `role` on `UnstyledLink`, should be log a `dev` warning to the `console` _(like we do with `ariaPressed`)_?
- Should we concatenate roles? Looks like multiple space separated roles are valid
  - https://act-rules.github.io/rules/674b10

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -69,6 +69,7 @@ type ActionButtonProps = Pick<
   | 'submit'
   | 'disabled'
   | 'loading'
+  | 'role'
   | 'ariaControls'
   | 'ariaExpanded'
   | 'ariaPressed'
@@ -90,6 +91,7 @@ export function Button({
   loading,
   pressed,
   accessibilityLabel,
+  role,
   ariaControls,
   ariaExpanded,
   ariaPressed,
@@ -285,6 +287,7 @@ export function Button({
     submit,
     disabled: isDisabled,
     loading,
+    role,
     ariaControls,
     ariaExpanded,
     ariaPressed: ariaPressedStatus,
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -52,6 +52,7 @@ interface CommonButtonProps
     ButtonProps,
     | 'id'
     | 'accessibilityLabel'
+    | 'role'
     | 'onClick'
     | 'onFocus'
     | 'onBlur'
@@ -69,7 +70,6 @@ type ActionButtonProps = Pick<
   | 'submit'
   | 'disabled'
   | 'loading'
-  | 'role'
   | 'ariaControls'
   | 'ariaExpanded'
   | 'ariaPressed'
@@ -271,6 +271,7 @@ export function Button({
     id,
     className,
     accessibilityLabel,
+    role,
     onClick,
     onFocus,
     onBlur,
@@ -287,7 +288,6 @@ export function Button({
     submit,
     disabled: isDisabled,
     loading,
-    role,
     ariaControls,
     ariaExpanded,
     ariaPressed: ariaPressedStatus,
--- a/src/components/UnstyledButton/UnstyledButton.tsx
+++ b/src/components/UnstyledButton/UnstyledButton.tsx
@@ -24,6 +24,7 @@ export function UnstyledButton({
   loading,
   pressed,
   accessibilityLabel,
+  role,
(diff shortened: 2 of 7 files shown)


dleroux:
Code looks good. 

> Should we pass role to UnstyledLink

Any reason why we shouldn't?

> Should we concatenate roles?

I think we should if it's allowed. With the push for a11y, these might come in useful.

dleroux (approved):

src/components/UnstyledButton/UnstyledButton.tsx
```diff
@@ -58,6 +61,9 @@ export function UnstyledButton({
   };
   const interactiveProps = {
     ...commonProps,
+    role: loading
+      ? [...new Set([ARIA_ROLE_ALERT, ...(role?.split(' ') || [])])].join(' ')
```
beefchimi, line 65: There are a bunch of different ways I could have done this, but essentially I am just making sure `alert` doesn't end up in the `role` value twice.

beefchimi:
@dleroux I've addressed your comments in some new commits. Mind taking another look and letting me know what you think?

cc/ @emma-boardman in case you want to review

src/components/UnstyledButton/UnstyledButton.tsx
BPScott: Following some a11y investigation I think we were going to remove adding a role if you're loading as it causes invalid aria states anyway. So it might be easier to just get rid of that entirely. @dfmcphee can probably talk more about that.

dfmcphee: Yep, this is one of the things flagged because alert isn't a valid role for a button. I think we should remove this default behaviour altogether.

Screen Shot 2020-11-03 at 10 15 42 AM

beefchimi: 👍  I will push a new commit that 🔥 the `alert`.

Would we want `alert` on the `<a />` by any chance? Or just kill it entirely?

dfmcphee: > Would we want `alert` on the `<a />` by any chance? Or just kill it entirely?

I would say kill it entirely. Having a loading link seems like a bit of an anti pattern anyway.

beefchimi: Alright folks - new commit has been pushed that removes `alert`!

dfmcphee: I think you are now getting a fail because it is expecting those two failures listed in the check. You can remove them here: https://github.com/Shopify/polaris-react/blob/master/scripts/accessibility-check.js#L103-L104

beefchimi: Excellent 😄  thank you!

dfmcphee (approved):

src/components/Button/Button.tsx
```diff
@@ -52,6 +52,7 @@ interface CommonButtonProps
     ButtonProps,
     | 'id'
     | 'accessibilityLabel'
+    | 'role'
```
DanGodfrey, line 55: Late to the party, but thoughts on triggering a warn if the `role` isn't in a defined list of `Button` appropriate roles in order to take away some foot guns?

beefchimi: We could maintain an `enum` of roles, and then subset `enums` for particular HTML elements. This would mean TypeScript indicates when an invalid value is passed. But I'm not sure thats really the job of Polaris to validate something like this. The spec defining what values are available to `role` could be updated to include more options in the future, so the burden of maintaining that would be on Polaris.

Would be curious to hear what someone on Polaris thinks of this idea.

PR #3595, opened 2020-11-02
[Popover ] Fix for preventAutofocus and remove focus onClose

dleroux:

I'm currently working on an alternate implementation of Autocomplete and need a few minor changes done to Polaris. Although these will not impact the current Autocomplete they will help fix issues there as well should it be needed.

I'm also using this PR to export the `labelId` function, which I also need for a11y on my implementation of Autocomplete.

1. export the `labelId` function from label. This will be used to assign an `aria-labelledby` prop to the list/
2. added a `preventFocusOnClose` prop to popover to prevent the textfield to gain focus when tab out of the field.
3. remove the `tabindex=-1` when preventAutoFocus is enabled. This triggered the focus to that div when scrolling the popover which is not the desired behavior for popover and can be unexpected outside popover. 
 

### How to 🎩

For the popover add the `preventFocusOnClose` to a popover and ensure the activator doesn't get focus on close.

### 🎩 checklist


kyledurand (approved):
Code and 🎩 look good to me 👍 

Thanks Dan!

PR #3598, opened 2020-11-03
[UnstyledLink] Fix mismatch for url / href

kyledurand:

Fixes https://github.com/Shopify/polaris-ux/issues/503

This took a bit of digging but I realized after changing `<LinkComponent />` to a regular `<a />` and passing props in I was able to see that url wasn't being transformed to `href={url}` which was why it wasn't getting an href in web. 

image
image

Building this branch for web fixes the skip to content issue
image

## 

### 🎩 checklist

### DIFF
--- a/src/components/UnstyledLink/UnstyledLink.tsx
+++ b/src/components/UnstyledLink/UnstyledLink.tsx
@@ -17,7 +17,7 @@ export const UnstyledLink = memo(
   forwardRef<unknown, UnstyledLinkProps>(function UnstyledLink(props, _ref) {
     const LinkComponent = useLink();
     if (LinkComponent) {
-      return <LinkComponent {...unstyled.props} {...props} />;
+      return <LinkComponent {...unstyled.props} {...props} href={props.url} />;
     }
 
     const {external, url, ...rest} = props;
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -13,6 +13,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Bug fixes
 
 - Removed extra bottom border on the `DataTable` and added curved edges to footers ([#3571](https://github.com/Shopify/polaris-react/pull/3571))
+- Fixed a bug where `UnstyledLink` wasn't properly converting the url prop to href ([#3598](https://github.com/Shopify/polaris-react/pull/3598))
 
 ### Documentation
 
--- a/src/components/UnstyledLink/tests/UnstyledLink.test.tsx
+++ b/src/components/UnstyledLink/tests/UnstyledLink.test.tsx
@@ -15,6 +15,16 @@ describe('<UnstyledLink />', () => {
       expect(anchorElement).toHaveLength(1);
     });
 
+    it('converts url prop to href', () => {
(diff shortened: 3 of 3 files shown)


src/components/UnstyledLink/UnstyledLink.tsx
```diff
@@ -17,7 +17,7 @@ export const UnstyledLink = memo(
   forwardRef<unknown, UnstyledLinkProps>(function UnstyledLink(props, _ref) {
     const LinkComponent = useLink();
     if (LinkComponent) {
-      return <LinkComponent {...unstyled.props} {...props} />;
+      return <LinkComponent {...unstyled.props} {...props} href={props.url} />;
```
kyledurand, line 20: @BPScott does this break links? I'm not sure what changed on our end in the last couple of weeks but after some debugging this morning I found that the default skip to content link on frame was being marked as not safe in web [here](https://github.com/Shopify/web/blob/master/packages/@shopify/safe-redirect/index.ts#L66-L68). We can Tuple if you need more context

alex-page (approved):
Woah, thanks for digging into this!

src/components/UnstyledLink/UnstyledLink.tsx
BPScott: Yo yeah, I think this has the potential for weirdness. It might not *break* react-router / next-router links at runtime but it will cause the typing to get weird as LinkComponent expects the url to be in the `url` prop but now you're pumping it into the href too which is confusing.

Was just chatting with @dfmcphee about this. I not looked at this deeply but I *thiiiink* the better fix for this to identify where we're passing "href" into UnstyledLink, and change it so that it gets passed as the URL instead. Dom mentioned UnstyledButton might be doing something unexpected. Commenting out https://github.com/Shopify/polaris-react/blob/master/src/components/UnstyledButton/UnstyledButton.tsx#L12 might give some insight into where we're passing arbitrary stuff. Having that escape hatch of "any prop is fine here" feels like a problem to me. 

Alternativly is there something silly in web's link component where it is render

BPScott (changes requested):
Not sure what change is needed, but I think the root of this issue lies elsewhere, and this approach makes the typing of LinkComponent more confusing than what it already is per comment above.

Throwing a "Request changes" to block merge as I'd like to learn a bit more and see if we can find a deeper root cause, as I think adding extra props to LinkComponent has the potential to cause confusion for consumers :)

src/components/UnstyledLink/UnstyledLink.tsx
kyledurand: Thanks Ben! I put up this alternative PR in the meantime which is much safer: https://github.com/Shopify/polaris-react/pull/3599. Going to close this one in a bit unless anyone has any other ideas

BPScott: Part of me wonders if some of the stuff in Web's link component stuff isn't realising anchor links  "#whatever" should be treated as a plain a tag and that url needs to be converted to href.

kyledurand:
Ok with the alt PR up and a potential fix on the way on the web side I'm going to close this

PR #3599, opened 2020-11-03
[Frame] Change skip to content from button to anchor tag

kyledurand:

Alternative to https://github.com/Shopify/polaris-react/pull/3598

Buttons use `Link` under the hood and web will not render an href for any links that may contain JS. This link does not need to use React router so why not just make it a plain anchor?

### DIFF
--- a/src/components/Frame/Frame.scss
+++ b/src/components/Frame/Frame.scss
@@ -258,5 +258,20 @@ $skip-vertical-offset: rem(10px);
   &.focused {
     pointer-events: all;
     opacity: 1;
+    > a {
+      @include focus-ring($style: 'focused');
+    }
+  }
+
+  > a {
+    @include button-base;
+    @include text-style-button;
+    @include focus-ring($border-width: rem(1px));
+    color: var(--p-text);
+
+    &:focus {
+      border-color: none;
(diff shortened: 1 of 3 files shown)


BPScott:
Web is being a buggy consumer because it's LinkComponent isn't handling anchor links properly. The stubborn approach here is to say "that's a bug in web, the fix should be in web". Which is what Shopify/web#34273 does. That fix should be applied to web to avoid any future buggy behaviour in any links in web.

Case closed right? But thinking a bit more about this - there might still be some value in merging this PR anyway. This skip link is always going to be an anchor so it is never going to need to be handled by any routing library specify component - which is why we let people configure LinkComponent. By merging this we protect against consumers accidentally misconfiguring themselves in the future, which seems useful.

Is that value of that extra bit of defensiveness worth the extra complexity? If so let's go for this in addition to the fix in web!

kyledurand:
> Is that value of that extra bit of defensiveness worth the extra complexity? If so let's go for this in addition to the fix in web!

I prefer the web solution because it doesn't mess with button styles like this PR does; but I'm going to leave this open for now in case there's safety concern we didn't think of.

alex-page (approved):

PR #3600, opened 2020-11-03
[Popover] Focus first focusable node

### LINKED ISSUE #3602: [Popover] Focus first focusable child
## Feature request summary

Opening the `Popover` currently set's the focus on the `Popover` container instead of trying to focus the first focusable element. 

It would be a nice improvement to try to find a focusable element and set the focus on it. If nothing focusable is present then keep the current behaviour.

## Rationale

Our use case is the timeline component used in multiple places (`CustomerDetails`, `OrderDetails`, `DiscountDetails`) in the Admin. The timeline has 2 popovers with a search `TextField` at the top so when the popover is activated the TextField should be focused immediately without an extra `tab` or `click`.

So far this functionality was implemented by a fork of the `Popover` (polaris-next) but we would like to switch back to the original component.

---

attila-berki:

Fixes #3602

By using the existing utility function `findFirstFocusableNode` try to find the first focusable child node and set the focus on it when the `Popover` is activated.

If no focusable nodes are present keep the current focus on the container.

This PR extends the `focusContent` functionality of the `PopoverOverlay` component by searching for the first focusable node and setting the focus on it if available.
Simply enhancing the focus to look for the first focusable node would change the behaviour of the component so I decided to add a new prop `autofocusTarget` which can have three options: `none`, `container`, `first-node`.

- `none` is equal with the current `preventAutofocus`:`true` prop so I would suggest to gradually replace it
- `container` would be default option and would work the same way as the `Popover` works today
- `first-node` would be the new functionality which looks for the first focusable node.

What do you think?

The approach is already used by the forked [Popover](https://github.com/Shopify/web/blob/master/packages/%40shopify/polaris-next/components/Popover/PopoverOverlay.tsx#L97-L98) component in the Admin.

Before:
popover-no-focus

After (notice the search field is focused automatically)
popover-focus

## 

### How to 🎩

<details>

```jsx

import {ActionList, Button, Page, Popover, TextField} from '../src';

    [],
  );

  const activator = (
      Open
    </Button>
  );

  return (
    <Page title="Playground">
      <Popover
        active={popoverActive}
        activator={activator}
        onClose={togglePopoverActive}
        autofocusTarget="first-node"
      >
        <Popover.Pane fixed>
          <TextField label="" placeholder="Search" onChange={() => {}} />
        </Popover.Pane>
        <Popover.Pane>
          <ActionList
            items={[
              {content: 'Online store'},
              {content: 'Facebook'},
              {content: 'Shopify POS'},
            ]}
          />
        </Popover.Pane>
      </Popover>
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Popover/components/PopoverOverlay/PopoverOverlay.tsx
+++ b/src/components/Popover/components/PopoverOverlay/PopoverOverlay.tsx
@@ -1,6 +1,7 @@
 import React, {PureComponent, Children, createRef} from 'react';
 import {durationBase} from '@shopify/polaris-tokens';
 
+import {findFirstFocusableNode} from '../../../../utilities/focus';
 import {InversableColorScheme, ThemeProvider} from '../../../ThemeProvider';
 import {classNames} from '../../../../utilities/css';
 import {
@@ -24,6 +25,8 @@ export enum PopoverCloseSource {
   FocusOut,
   ScrollOut,
 }
+
+export type PopoverAutofocusTarget = 'none' | 'first-node' | 'container';
 
 enum TransitionStatus {
   Entering = 'entering',
@@ -49,6 +52,7 @@ export interface PopoverOverlayProps {
   hideOnPrint?: boolean;
   onClose(source: PopoverCloseSource): void;
   colorScheme?: InversableColorScheme;
+  autofocusTarget?: PopoverAutofocusTarget;
 }
 
 interface State {
@@ -157,10 +161,20 @@ export class PopoverOverlay extends PureComponent<PopoverOverlayProps, State> {
   }
 
   private focusContent() {
-    if (this.props.preventAutofocus) {
-      return;
+    const {autofocusTarget = 'container', preventAutofocus} = this.props;
+
+    if (preventAutofocus) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        'Deprecation: The preventAutofocus prop has been deprecated. Use autofocusTarget: "none" instead. Read more at [https://github.com/Shopify/polaris-react/issues/3602]',
+      );
     }
-    if (this.contentNode == null) {
+
+    if (
+      preventAutofocus ||
+      autofocusTarget === 'none' ||
+      this.contentNode == null
+    ) {
       return;
     }
 
@@ -169,9 +183,17 @@ export class PopoverOverlay extends PureComponent<PopoverOverlayProps, State> {
         return;
       }
 
-      this.contentNode.current.focus({
-        preventScroll: process.env.NODE_ENV === 'development',
-      });
+      const focusableChild = findFirstFocusableNode(this.contentNode.current);
+
+      if (focusableChild && autofocusTarget === 'first-node') {
+        focusableChild.focus({
+          preventScroll: process.env.NODE_ENV === 'development',
+        });
+      } else {
+        this.contentNode.current.focus({
+          preventScroll: process.env.NODE_ENV === 'development',
+        });
+      }
     });
   }
 
@@ -191,6 +213,7 @@ export class PopoverOverlay extends PureComponent<PopoverOverlayProps, State> {
       hideOnPrint,
       colorScheme,
       preventAutofocus,
+      autofocusTarget,
     } = this.props;
 
     const className = classNames(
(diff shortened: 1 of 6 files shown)


attila-berki:
I found a use case where the automatic focus doesn't seems to be a good idea
image

To avoid these side effects I will extend the implementation with a boolean flag, something like `focusFirstNode` which will toggle the new functionality defaulting to `false`

src/components/Popover/components/PopoverOverlay/PopoverOverlay.tsx
```diff
@@ -169,7 +170,8 @@ export class PopoverOverlay extends PureComponent<PopoverOverlayProps, State> {
         return;
       }
 
-      this.contentNode.current.focus({
+      const focusableChild = findFirstFocusableNode(this.contentNode.current);
```
sylvhama, line 173: If you look at all existing storybook stories you will see that `findFirstFocusableNode` will always find a node and focus it. 
Screen Shot 2020-11-04 at 2 03 02 PM

I don't think that's an issue but are okay to change the default behaviour of the component?

sylvhama (approved):
Code LGTM and the fix works.
I have one interrogation for the Polaris team in my comment.

src/components/Popover/components/PopoverOverlay/PopoverOverlay.tsx
attila-berki: I think we should make it an optional feature as it can look weird in some cases to have a focus ring around a button for example.

sylvhama: `preventAutofocus` could be deprecated and replaced by an enum prop? 
E.g. it could be `defaultFocus`, by default it would be `Container`, we could set `FirstNode` or `None`.

attila-berki: Sounds good, although I am not familiar with the deprecation process 🙂 
I would propose another name for it, maybe `autoFocusTarget`. What do you think?

sylvhama: Sounds good, we can try to contact polaris tomorrow to see if that's doable.

sylvhama (approved):
Works fine, I used this Playground to test the focus of different elements, it all worked.

I just have few improvement suggestions. 

```tsx

import {
  Page,
  Popover,
  Select,
  TextField,
  Button,
  FormLayout,
  PopoverAutofocusTarget,
} from '../src';

  return (
    <Page title="Playground">
      <PopoverFormExample />
    </Page>
  );
}

function PopoverFormExample() {
  const [tagValue, setTagValue] = useState('');

    [],
  );

  const handleTagValueChange = useCallback((value) => setTagValue(value), []);

  const activator = (
      Filter
    </Button>
  );

  return (
    <div style={{height: '280px'}}>
      <Popover
        active={popoverActive}
        activator={activator}
        onClose={togglePopoverActive}
        ariaHaspopup={false}
        sectioned
        autofocusTarget={PopoverAutofocusTarget.FirstNode}
      >
        <FormLayout>
          <Select label="Show all customers where:" options={['Tagged with']} />
          <TextField
            label="Tags"
            value={tagValue}
            onChange={handleTagValueChange}
          />
          <Button size="slim">Add filter</Button>
        </FormLayout>
      </Popover>
    </div>
  );
}
```

src/components/Popover/components/PopoverOverlay/PopoverOverlay.tsx
```diff
@@ -25,6 +26,12 @@ export enum PopoverCloseSource {
   ScrollOut,
 }
 
+export enum PopoverAutofocusTarget {
+  None,
```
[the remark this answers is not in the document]
attila-berki: It feels redundant but I think we should deprecate the `preventAutofocus` instead. It might take a while to achieve that but having both options will allow a gradual refactor.

src/components/Popover/components/PopoverOverlay/PopoverOverlay.tsx
```diff
@@ -157,7 +161,9 @@ export class PopoverOverlay extends PureComponent<PopoverOverlayProps, State> {
   }
 
   private focusContent() {
-    if (this.props.preventAutofocus) {
+    const {autofocusTarget = 'container'} = this.props;
+
+    if (this.props.preventAutofocus || autofocusTarget === 'none') {
```
tmlayton, line 166: Refactor this to include the condition below `this.contentNode == null`?

src/components/Popover/components/PopoverOverlay/PopoverOverlay.tsx
tmlayton: Here are the deprecation guidelines: https://github.com/Shopify/polaris-react/blob/master/documentation/Deprecation%20guidelines.md

tmlayton (approved):
Looks fine to me. I’d be in favor of deprecating `preventAutofocus` as well.

PR #3604, opened 2020-11-05
Update discoverability-action to latest version

alex-page:


.github/workflows/discoverability.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       run: yarn
 
     - name: Add a comment with known areas to test
-      uses: shopify/discoverability-action@average-impact-version-patch1
+      uses: shopify/discoverability-action@v0.0.13
```
esnunes, line 16: The discoverability-action package.json shows version [0.0.12](https://github.com/Shopify/discoverability-action/blob/master/package.json#L8) is that a problem?

alex-page: It shouldn't matter because we don't `npm i` the package.

amrocha (approved):
Thanks Alex!

PR #3609, opened 2020-11-09
Remove 10px padding on details page

alex-page:

When trying to work on https://github.com/Shopify/polaris-react/issues/3438, the details page has a 10px padding added by storybook. This adds an unnecessary speed bump to replicating production pages.

Removing the stupid added 10px from storybook.

### DIFF
--- a/.storybook/polaris-readme-loader.js
+++ b/.storybook/polaris-readme-loader.js
@@ -71,7 +71,6 @@ ${example.storyName}.parameters = {
     minHeight: '720px',
     borderBottom: '1px solid #000',
     marginBottom: '8px',
-    padding: '10px',
   }}>
   <Heading>${example.name}</Heading>
   <${example.storyName}Component />
--- a/.storybook/polaris-readme-loader.js
+++ b/.storybook/polaris-readme-loader.js
@@ -71,6 +71,7 @@ ${example.storyName}.parameters = {
     minHeight: '720px',
     borderBottom: '1px solid #000',
     marginBottom: '8px',
+    padding: '10px',
   }}>
   <Heading>${example.name}</Heading>
   <${example.storyName}Component />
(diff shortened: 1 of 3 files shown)


BPScott:
Would a better fix be updating the "toggle fullscreen" logic at https://github.com/Shopify/polaris-react/blob/6f0cfe98b04680eb824eaaa53c5285e5719c7978/.storybook/polaris-readme-loader.js#L39-L41

alex-page:
> Would a better fix be updating the "toggle fullscreen" logic at
> 
> https://github.com/Shopify/polaris-react/blob/6f0cfe98b04680eb824eaaa53c5285e5719c7978/.storybook/polaris-readme-loader.js#L39-L41

I think you are probably right! Thanks @BPScott

.storybook/polaris-readme-loader.js
```diff
@@ -71,6 +71,7 @@ ${example.storyName}.parameters = {
     minHeight: '720px',
     borderBottom: '1px solid #000',
     marginBottom: '8px',
+    padding: '10px',
```
amrocha, line 74: Looks like you're adding the padding here. Does this not show up if you're looking at it in the fullscreen layout?

amrocha:
@BPScott Ben get out of here you're on vacation we got this

.storybook/polaris-readme-loader.js
alex-page: Moving to `fullscreen` removed the padding on the "All" component pages. So I added it back here.

amrocha: Gotcha. Is the intent to run all previews on fullscreen now? If so, then I think I like your solution, and I'd get rid of the `hasFullscreenLayout` conditional variable. Otherwise, if it's on a per-component basis, then updating the place Ben mentioned makes more sense

amrocha (approved):

alex-page:
@BPScott @amrocha I have cleaned this up to only affect Playground pages.

kaelig (approved):
Makes total sense! :shipit: 


>Removing the ~stupid~ added 10px from storybook.

🤔👎

PR #3615, opened 2020-11-10
[WIP][Badge][New DL] Fix badge size and shape

### LINKED ISSUE #3614: [New DL] Badge shape and size is slightly off
## Issue summary

With the new design language, badges are slightly the wrong size and shape.

## Expected vs actual behavior

Badges should match the line height of body text (20px) but are 22px tall instead:

Screen Shot 2020-11-10 at 9 41 13 AM

Badges should’t be taller than wide, but this can happen:

Screen Shot 2020-11-10 at 9 41 51 AM

The small badge variant actually renders larger (taller) than the normal badge! It’s intended to be 18px tall:

Screen Shot 2020-11-10 at 9 43 33 AM

## Reduced test case

## Specifications

- Polaris version number: whatever Admin is using as of Nov 10, 2020
- Browser: Chrome 86
- Device: Mac
- Operating System: macOS 10.15.7

ry5n:
Since the new DL, badge size and shape is slightly off. See #3614 for details.

When complete, fixes #3614

This PR gets started with the changes I’d recommend. It’s a draft and will need to be tested more thoroughly.

### What’s this NOT doing?

We also need to fix the main nav, specifically the element `.Polaris-Navigation__Badge`. I’d remove any margins on this element and instead do something like this (pseudocode!):

```scss
// Small screens
.Polaris-Navigation__Badge {
  display: flex;
  min-height: rem(40px);
  align-items: center;
}

// Large screens
.Polaris-Navigation__Badge {
  min-height: rem(36px);
}
```

What this does is make the centering of the badge work independently of the height of the badge itself.

There might be other places where we need to make changes in admin.

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Badge/Badge.scss
+++ b/src/components/Badge/Badge.scss
@@ -2,9 +2,8 @@
 
 $height: rem(20px);
 $small-height: rem(18px);
-$extra-small-height: rem(16px);
 $horizontal-padding: spacing(tight);
-$vertical-padding: rem(3px);
+$vertical-padding: rem(2px);
 
 $pip-size: rem(10px);
 $pip-spacing: ($height - $pip-size) / 2;
@@ -19,12 +18,14 @@ $pip-spacing: ($height - $pip-size) / 2;
   @include pip-color(var(--p-icon, color('ink', 'lightest')));
   display: inline-flex;
   align-items: center;
+  min-width: $height;
+  min-height: $height;
   padding: $vertical-padding $horizontal-padding;
   background-color: var(--p-surface-neutral, color('sky'));
   border: var(--p-override-zero, border-width(thick)) solid color('white');
   border-radius: $height;
   font-size: rem(13px);
-  line-height: $extra-small-height;
+  line-height: 1;
   color: var(--p-text, color('ink', 'light'));
   font-weight: var(--p-badge-font-weight, 400);
 
@@ -36,7 +37,8 @@ $pip-spacing: ($height - $pip-size) / 2;
 
 .sizeSmall {
   font-size: font-size(caption, large-screen);
-  line-height: $small-height;
+  min-width: $small-height;
+  min-height: $small-height;
 }
 
 .statusSuccess {


src/components/Badge/Badge.scss
```diff
@@ -2,9 +2,8 @@
 
 $height: rem(20px);
 $small-height: rem(18px);
-$extra-small-height: rem(16px);
```
ry5n, line 5: There is no extra-small badge size

src/components/Badge/Badge.scss
```diff
@@ -2,9 +2,8 @@
...
 $horizontal-padding: spacing(tight);
-$vertical-padding: rem(3px);
+$vertical-padding: rem(2px);
```
ry5n, line 6: Padding creates a visual buffer in cases where the badge text wraps, but isn’t primarily responsible for the height of the badge.

src/components/Badge/Badge.scss
```diff
@@ -19,12 +18,14 @@ $pip-spacing: ($height - $pip-size) / 2;
   @include pip-color(var(--p-icon, color('ink', 'lightest')));
   display: inline-flex;
   align-items: center;
+  min-width: $height;
+  min-height: $height;
```
ry5n, line 22: Here we set the desired height of the badge, without preventing it from growing if the text wraps

src/components/Badge/Badge.scss
```diff
@@ -19,12 +18,14 @@ $pip-spacing: ($height - $pip-size) / 2;
   @include pip-color(var(--p-icon, color('ink', 'lightest')));
   display: inline-flex;
   align-items: center;
+  min-width: $height;
```
ry5n, line 21: Prevent the badge from being taller than it is wide. In practice, this means it will be circular if the text inside is really narrow

src/components/Badge/Badge.scss
```diff
@@ -19,12 +18,14 @@ $pip-spacing: ($height - $pip-size) / 2;
...
+  min-height: $height;
   padding: $vertical-padding $horizontal-padding;
   background-color: var(--p-surface-neutral, color('sky'));
   border: var(--p-override-zero, border-width(thick)) solid color('white');
   border-radius: $height;
   font-size: rem(13px);
-  line-height: $extra-small-height;
+  line-height: 1;
```
ry5n, line 28: Keep the line-height narrow and relative to the font-size helps avoid making the badge too tall, and also keeps it more compact in cases where the badge text itself wraps

ry5n:
There’s a problem with this approach in the old Polaris style in that the badges have white borders, which are also taken into account by the `min-height` setting, causing them to be too narrow. We’ll need to make adjustments for this or avoid this approach

ry5n:
These badges look good now, but with the changes in this PR, they would end up non-circular. This bears a little more thought.

Screen Shot 2020-11-10 at 2 27 48 PM

alex-page:
@ry5n closing this PR as the badge spacing has been updated and it matches the body text height.

PR #3616, opened 2020-11-11
[Modal] Use correct color for header border

tmlayton:
This is supposed to be using the divider color. `border('divider')` returns `var(--p-divider)` with the new design language.


alex-page (approved):

maxariss:
🙌

PR #3619, opened 2020-11-11
[ActionList] Align ActionList ::before indicator with OptionList

### LINKED ISSUE #3611: [ActionList] ::before "indicator" not aligned to edge of container
Since the `new design language` rollout, I've been seeing several instances of a "gap" to the left of the `ActionList > Item` "indicator":

wups

emma-boardman:

Fixes #3611

image

ActionList indicator styles were [updated within Figma to use the same styles as OptionList](https://github.com/Shopify/polaris-react/issues/3258#issuecomment-694931240)

This PR fixes the alignment issue in #3611 by moving the OptionList indicator styles into a shared mixin that can be used by both the OptionList and the ActionList. 

Action list after:

Action list after changes

Option list after:

Option list after changes

### 🎩 checklist

* [x] Tested on [mobile](Galaxy S5 emulator on Chrome)
* [x] Tested on [multiple browsers](Chrome on MacOS, Safari on MacOS, Firefox on MacOS, Edge on WindowsVM)
* [x] Tested for [accessibility](WHCM on Edge)

### DIFF
--- a/src/components/ActionList/ActionList.scss
+++ b/src/components/ActionList/ActionList.scss
@@ -3,7 +3,6 @@
 $image-size: rem(20px);
 $item-min-height: rem(40px);
 $item-vertical-padding: ($item-min-height - line-height(body)) / 2;
-$active-indicator-width: rem(3px);
 
 .ActionList {
   list-style: none;
@@ -228,7 +227,7 @@ $active-indicator-width: rem(3px);
 
     // Only show when the button is selected, not active
     [aria-selected='true'] > &::before {
-      @include list-single-selection-state;
+      @include list-selected-indicator;
     }
 
     // Active, hovered, focused (:active:hover:focus)
--- a/src/components/ActionList/ActionList.scss
+++ b/src/components/ActionList/ActionList.scss
@@ -228,17 +228,7 @@ $active-indicator-width: rem(3px);
 
     // Only show when the button is selected, not active
     [aria-selected='true'] > &::before {
-      content: '';
-      position: absolute;
-      top: 0;
-      bottom: 0;
-      left: -1 * spacing(extra-tight);
-      height: 100%;
-      width: $active-indicator-width;
-      background-color: var(--p-interactive);
-      border-top-right-radius: var(--p-border-radius-base);
-      border-bottom-right-radius: var(--p-border-radius-base);
-      transform: translateX(-100%);
+      @include list-single-selection-state;
     }
 
     // Active, hovered, focused (:active:hover:focus)
--- a/src/components/OptionList/components/Option/Option.scss
+++ b/src/components/OptionList/components/Option/Option.scss
@@ -3,7 +3,6 @@
 $min-height: control-height();
 $control-size: rem(16px);
 $control-vertical-adjustment: rem(2px);
-$selected-option-identifier-width: rem(3px);
 
 .Option {
   @include unstyled-button;
@@ -164,7 +163,7 @@ $selected-option-identifier-width: rem(3px);
     }
 
     &.select::before {
-      @include list-single-selection-state;
+      @include list-selected-indicator;
     }
   }
 }
--- a/src/components/OptionList/components/Option/Option.scss
+++ b/src/components/OptionList/components/Option/Option.scss
@@ -164,16 +164,7 @@ $selected-option-identifier-width: rem(3px);
     }
 
     &.select::before {
-      background: var(--p-interactive);
-      content: '';
-      position: absolute;
-      top: 0;
-      left: -1 * spacing(tight);
-      height: 100%;
-      display: block;
-      width: $selected-option-identifier-width;
-      border-top-right-radius: var(--p-border-radius-base);
-      border-bottom-right-radius: var(--p-border-radius-base);
+      @include list-single-selection-state;
     }
   }
 }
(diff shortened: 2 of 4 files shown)


src/styles/shared/_interaction-state.scss
```diff
@@ -24,3 +24,16 @@
   }
   background-image: $backgrounds;
 }
+
+@mixin list-single-selection-state {
+  background: var(--p-interactive);
```
beefchimi, line 29: I prefer the explicitness of `background-color` if we know that we are dealing with exactly that - a `color` and not some other `background-*` property.

src/styles/shared/_interaction-state.scss
```diff
@@ -24,3 +24,16 @@
...
+  content: '';
```
beefchimi, line 30: Very nitpicky but if we are dealing with pseudo elements, I prefer to see `content` come first, since it is so vitally important to the pseudo element rendering.

src/styles/shared/_interaction-state.scss
```diff
@@ -24,3 +24,16 @@
   }
   background-image: $backgrounds;
 }
+
+@mixin list-single-selection-state {
```
beefchimi, line 28: Not sure how I feel about this mixin name. Might I suggest some variation of:
- `action-list-indicator`
- `list-pseudo-indicator`
- `list-selected-indicator`

src/components/ActionList/ActionList.scss
```diff
@@ -228,17 +228,7 @@ $active-indicator-width: rem(3px);
 
     // Only show when the button is selected, not active
     [aria-selected='true'] > &::before {
-      content: '';
-      position: absolute;
-      top: 0;
-      bottom: 0;
-      left: -1 * spacing(extra-tight);
-      height: 100%;
-      width: $active-indicator-width;
```
beefchimi, line 237: You no longer use this variable within the file, and therefor it should be removed.

src/components/OptionList/components/Option/Option.scss
```diff
@@ -164,16 +164,7 @@ $selected-option-identifier-width: rem(3px);
...
 
     &.select::before {
-      background: var(--p-interactive);
-      content: '';
-      position: absolute;
-      top: 0;
-      left: -1 * spacing(tight);
-      height: 100%;
-      display: block;
-      width: $selected-option-identifier-width;
```
beefchimi, line 174: You no longer use this variable within the file, and therefor it should be removed

src/styles/shared/_interaction-state.scss
```diff
@@ -24,3 +24,16 @@
...
+  background: var(--p-interactive);
+  content: '';
+  position: absolute;
+  top: 0;
+  left: -1 * spacing(tight);
+  height: 100%;
+  display: block;
+  width: rem(3px);
```
beefchimi, line 36: Should this be tied to `border-width(thicker)`?

src/styles/shared/_interaction-state.scss
```diff
@@ -24,3 +24,16 @@
   }
   background-image: $backgrounds;
 }
+
+@mixin list-single-selection-state {
+  background: var(--p-interactive);
+  content: '';
+  position: absolute;
+  top: 0;
+  left: -1 * spacing(tight);
```
beefchimi, line 33: This here is the critical piece - `spacing(tight)` needs to match exactly what the `padding-left/right` is on the consuming "list" _(`ActionList/OptionList`)_.

Right now, coincidentally, they both use `spacing(tight)`. Just to be a little more explicit / future-proof, perhaps we make this an argument?

```scss
@mixin list-selected-indicator($offset: spacing(tight)) {
  left: -1 * $offset;
}
```

beefchimi (commented):
Some comments. Not sure if this is mentioned in the OP but its worth pointing out that the reason this was not aligned correctly was because the `left` property was using `spacing(extra-tight)` instead of `spacing(tight)`.

FYI The `ActionList` code used `transform: translateX(-100%);`, which is not present in your new mixin. I'm wondering if this was because the "indicator" was intended to `transition` when "selected"?

emma-boardman:
I love all this feedback, thank you! 

> Some comments. Not sure if this is mentioned in the OP but its worth pointing out that the reason this was not aligned correctly was because the `left` property was using `spacing(extra-tight)` instead of `spacing(tight)`.
> 
> FYI The `ActionList` code used `transform: translateX(-100%);`, which is not present in your new mixin.

Interestingly, the `transform: translateX(-100%)` is what requires the `left` to be offset by `spacing(extra-tight)` instead of `spacing(tight)`. If we update the `ActionList` `left` to have `spacing(tight)`, the `translateX` pushes it outside the `Item`. 

image

>  I'm wondering if this was because the "indicator" was intended to `transition` when "selected"?

I agree - what led me to not include it was that there is no other `transform`, nor any `transition` property within this file. That led me to think that maybe we don't transition this, and therefore that it isn't needed. However, I'm basing that on my limited knowledge of CSS transitions. Can we trigger a transition from outside this styling scope? Wouldn't it need to transition from another `transform` style applied to the` ::before `within this `.scss` file?

beefchimi (approved):
Just needs a rebase and you should be good to go.

Always best to get a review from someone on Polaris. @BPScott are you available to review?

alex-page (approved):
Thanks @emma-boardman ✨

PR #3621, opened 2020-11-11
[color-transformers > rgbToHsbl] Fix brightness precision of HSB conversions by increasing from 2 decimal digits to 4

MLDimitry:

I was originally investigating this issue in Shopify web https://github.com/Shopify/shopify/issues/265949

and found that the color-transformers on Polaris would convert both hex values `#888888` and `#878787` to the same HSB value `{ hue: 0, saturation: 0, brightness: 0.53 }`. The calculating actually looked correct, `#888888` was being converted to `0.5333` and `#878787` was being converted to `0.5294`. But the final step of the conversion limits these to only two decimal places, which resulted in the brightness values being rounded to `0.53` in both cases and losing their precision.

The Shopify Web ticket has more details on the original issue.

This PR fixes the `color-transformers` issue by increasing the brightness value's precision from 2 decimal points to 4. 

It also adds two tests for the failing conversions from the Shopify web issue for `rgb(135, 135, 135)` and `rgb(136, 136, 136)` to ensure they're both translated to the correct expected values.

# Questions for reviewers

* Is this change a bug fix? Or would we consider changing the number of decimal digits as a breaking change?
* Thoughts on updating Polaris version in Shopify web? Should we create a workaround/short term fix in Shopify web for now instead of updating Polaris during the holiday critical phase?

### 🎩 checklist

* [x] Tested on [multiple browsers](https://help.shopify.com/en/manual/intro-to-shopify/shopify-admin/supported-browsers) **_(Firefox and Chrome)_**
* [ ] ~Updated the component's `README.md` with documentation changes~ **_(Didn't see any README for utilities)_**

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -20,7 +20,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Removed `tabIndex=-1` from `Popover` when `preventAutoFocus` is true ([#3595](https://github.com/Shopify/polaris-react/pull/3595))
 - Fixed `Modal` header border color ([#3616](https://github.com/Shopify/polaris-react/pull/3616))
 - Fixed `TopBar` search clear button alignment on iOS ([#3618](https://github.com/Shopify/polaris-react/pull/3618))
-- Fixes HSB brightness conversion by increasing precision from 2 decimals to 4
+- Fixed HSB brightness conversion by increasing precision from 2 decimals to 4
 
 ### Documentation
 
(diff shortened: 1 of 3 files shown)


alex-page (approved):

kyledurand (approved):
👍

MLDimitry:
@alex-page @kyledurand Any thoughts on these two questions before I merge:

* Is this change a bug fix? Or would we consider changing the number of decimal digits as a breaking change? Our type hasn't changed so I lean towards bug fix, since consumers shouldn't rely on `toFixed(2)` implementation details
* Since Polaris UNRELEASED.md contains more than just bug fixes, I don't think we want to update the version in Shopify/web directly, would you agree? Or should I work to cut a new release?

kyledurand:
> Is this change a bug fix? Or would we consider changing the number of decimal digits as a breaking change?

Could be considered a bug fix but I can't see it as a breaking change. If you need this available in web right away then we could cut a release with just this commit in it.

MLDimitry:
> If you need this available in web right away then we could cut a release with just this commit in it.

Yes I think that would be best, then I can tophat the fix on web staging before bumping there.

I'm going to kick off the merge this morning. If you guys can help me cut the release once it's merged, or point me to some docs on the process so I can release it myself, that'd be great.

PR #3629, opened 2020-11-12
[Filters] Add semantic markup to header

AndrewMusgrave:

* Adding semantic HTML

### Screenie
Screen Shot 2020-11-12 at 11 33 58 AM

### How to 🎩

* Chroma should be 💯

### DIFF
--- a/src/components/Filters/Filters.tsx
+++ b/src/components/Filters/Filters.tsx
@@ -187,7 +187,7 @@ class FiltersInner extends Component<CombinedProps, State> {
             aria-expanded={filterIsOpen}
           >
             <div className={styles.FilterTriggerLabelContainer}>
-              <h2 className={styles.FilterTriggerTitle}>
+              <h3 className={styles.FilterTriggerTitle}>
                 <TextStyle
                   variation={
                     this.props.disabled || filter.disabled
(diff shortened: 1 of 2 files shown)


src/components/Filters/Filters.tsx
```diff
@@ -187,7 +187,7 @@ class FiltersInner extends Component<CombinedProps, State> {
             aria-expanded={filterIsOpen}
           >
             <div className={styles.FilterTriggerLabelContainer}>
-              <h2 className={styles.FilterTriggerTitle}>
+              <h3 className={styles.FilterTriggerTitle}>
```
alex-page, line 190: Are they always going to follow a h2? Is this known?

AndrewMusgrave: They should. The filters content opens into a sheet (portal), so app adjacent. If we make the sheet title an `h2`. Making the filter headers `h3` should remain semantic 👌

alex-page: Interesting, I wonder if there is an example where they would be used in another context. That makes sense though.

AndrewMusgrave: Nope! The filters components render all of the markup, so unless it changes in Polaris that'll always be the case 😄

kyledurand (approved):
Looks good to me

ashwinnarayanan2001 (approved):
Looking good.

emmaorhun (approved):
👌

PR #3631, opened 2020-11-12
[Tooltip] Add accessibility label to tooltip

AndrewMusgrave:

* adding a label
* adding role

### Screenshot of voiceover text
Screen Shot 2020-11-12 at 12 02 41 PM

### How to 🎩

* Chroma + screen reader 🎉

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -30,6 +30,8 @@ export interface TooltipProps {
    * @default 'span'
    */
   activatorWrapper?: string;
+  /** Visually hidden text for screen readers */
+  accessibilityLabel?: string;
 }
 
 export function Tooltip({
@@ -40,6 +42,7 @@ export function Tooltip({
   active: originalActive,
   preferredPosition = 'below',
   activatorWrapper = 'span',
+  accessibilityLabel,
 }: TooltipProps) {
   const WrapperComponent: any = activatorWrapper;
   const {value: active, setTrue: handleFocus, setFalse: handleBlur} = useToggle(
@@ -78,6 +81,7 @@ export function Tooltip({
         preferredPosition={preferredPosition}
         activator={activatorNode}
         active={active}
+        accessibilityLabel={accessibilityLabel}
         onClose={noop}
         light={light}
         preventInteraction={dismissOnMouseOut}
--- a/src/components/Tooltip/components/TooltipOverlay/TooltipOverlay.tsx
+++ b/src/components/Tooltip/components/TooltipOverlay/TooltipOverlay.tsx
@@ -32,8 +32,21 @@ export function TooltipOverlay({
   accessibilityLabel,
 }: TooltipOverlayProps) {
   const i18n = useI18n();
+  const markup = active ? (
+    <PositionedOverlay
+      active={active}
+      activator={activator}
+      preferredPosition={preferredPosition}
+      preventInteraction={preventInteraction}
+      render={renderTooltip}
+    />
+  ) : null;
 
-  const renderTooltip: PositionedOverlayProps['render'] = (overlayDetails) => {
+  return markup;
+
+  function renderTooltip(
+    overlayDetails: Parameters<PositionedOverlayProps['render']>[0],
+  ) {
     const {measuring, desiredHeight, positioning} = overlayDetails;
 
     const containerClassName = classNames(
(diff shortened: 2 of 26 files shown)


src/components/Tooltip/Tooltip.tsx
```diff
@@ -88,6 +91,8 @@ export function Tooltip({
       onMouseLeave={handleMouseLeave}
       onMouseOver={handleMouseEnterFix}
       ref={setActivator}
+      role="tooltip"
+      aria-label={accessibilityLabel}
```
alex-page, line 95: I wonder if we should be making the aria-label the read text or the content.

AndrewMusgrave: That's a good question. I added the prop for increased flexibility. Occasionally people ask for larger tooltips because they put large amounts of text in it. Although not promoted it does happen, I was thinking it might be better if the label is more concise. Happy to make the change to content if you think that's more worthwhile. 

One thing to note, changing it to content wouldn't be an API change so we could always add the accessibility prop later if we felt it's needed.

kyledurand: I like the idea of reading out the content

AndrewMusgrave: 👍  Ill make the change

AndrewMusgrave: Looks like content was changed to a react node 2-3 weeks ago so we can't use content anymore. Updated the pr according & adding more gifs below

AndrewMusgrave:
### Without accessibility label
Screen Recording 2020-11-17 at 03 30 59 PM

### With accessibility label
Screen Recording 2020-11-17 at 03 29 19 PM

yourpalsonja:
Hey @alex-page @kyledurand can you take another look at this?

kyledurand (approved):
Code looks good to me 👍

kyledurand:
Alex is off until the 7th while moving to BC and Dom is working heads down on live view. Do you have anyone on your team that could help review this?

src/components/Tooltip/components/TooltipOverlay/TooltipOverlay.tsx
```diff
@@ -16,10 +17,15 @@ export interface TooltipOverlayProps {
...
+  accessibilityLabel?: string;
   onClose(): void;
 }
 
-export class TooltipOverlay extends PureComponent<TooltipOverlayProps, never> {
+type CombinedProps = TooltipOverlayProps & {
+  i18n: ReturnType<typeof useI18n>;
+};
+
+class TooltipOverlayInner extends PureComponent<CombinedProps, never> {
```
BPScott, line 28: Possibly volunteering you for bonus work, but it looks pretty straightforward to rewrite this to be a functional component

AndrewMusgrave: I don't mind 😄  I'll have to move this to an enhancement rather than bug fix so it's not accidentally shipped though.

src/components/Tooltip/components/TooltipOverlay/TooltipOverlay.tsx
```diff
@@ -21,45 +21,21 @@ export interface TooltipOverlayProps {
   onClose(): void;
 }
 
-type CombinedProps = TooltipOverlayProps & {
-  i18n: ReturnType<typeof useI18n>;
-};
-
-class TooltipOverlayInner extends PureComponent<CombinedProps, never> {
```
AndrewMusgrave: I removed the pure component aspect since it's unlikely the `activator` prop will be memorized and we're *probably* getting very little gain from it.

BPScott: [Death to pure components!](https://github.com/Shopify/polaris-react/issues/2062)

src/components/Tooltip/components/TooltipOverlay/TooltipOverlay.tsx
```diff
@@ -16,44 +17,38 @@ export interface TooltipOverlayProps {
...
-        render={this.renderTooltip}
-      />
-    );
-  };
-
-  // eslint-disable-next-line @shopify/react-no-multiple-render-methods
-  private renderTooltip: PositionedOverlayProps['render'] = (
-    overlayDetails,
-  ) => {
+  function renderTooltip(
```
BPScott, line 47: Should this be wrapped in a useCallback so we don't redefine the function every time?

AndrewMusgrave: I left it as is because positioned overlay props are unlikely to be pure. Don't have a strong opinion here though, happy to make the change if you think it's valuable! @BPScott

AndrewMusgrave: I'm going on vacation but if you're ok with this would you mind hitting squash and merge for me so other teams aren't blocked 🙏

alex-page (approved):
🚢

PR #3632, opened 2020-11-12
[Filters] Announce applied filters

AndrewMusgrave:

* Making the applied tags a live region

### Screenshot of voiceover
Screen Shot 2020-11-12 at 12 14 51 PM

### How to 🎩

* Chroma + screen reader 🦐

### DIFF
--- a/src/components/Filters/Filters.tsx
+++ b/src/components/Filters/Filters.tsx
@@ -1,4 +1,4 @@
-import React, {Component, createRef} from 'react';
+import React, {Component, createRef, Fragment} from 'react';
 import {
   SearchMinor,
   ChevronUpMinor,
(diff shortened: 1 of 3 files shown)


alex-page (approved):

src/components/Filters/Filters.tsx
```diff
@@ -348,10 +349,13 @@ class FiltersInner extends Component<CombinedProps, State> {
       </div>
     );
 
-    const tagsMarkup =
-      !hideTags && appliedFilters && appliedFilters.length ? (
-        <div className={styles.TagsContainer}>
-          {appliedFilters.map((filter) => {
+    const shouldHideTagsContainer =
+      !appliedFilters || appliedFilters.length < 1;
```
kyledurand, line 353:
```suggestion
      appliedFilters?.length < 1;
```
Haven't tested this but I'm assuming appliedFilters can be either null or empty array so I _think_ this should work

kyledurand (approved):
👍

src/components/Filters/Filters.tsx
AndrewMusgrave: Tests fail with the change so I guess it's almost but not equivalent 😄

PR #3640, opened 2020-11-16
[color-transformers > Hsbl] Increase precision of hue, saturation, lightness, and alpha

MLDimitry:

Fixes https://github.com/Shopify/shopify/issues/265949

I originally solved part of this problem by increasing the precision of brightness to 4 decimal places in https://github.com/Shopify/polaris-react/pull/3621

What I realized testing my solution in production was I had missed the fact that other colour values would also lose precision in their hue, saturation, lightness, and alpha. This PR increases the precision of the other fields and also adds a function for more accurate rounding to decimals.

**Before - Note color gets replaced after pasting**
Color-Before

**After - Note color doesn't get replaced**
Color-After

This PR improves the fix for color-transformers by also increasing precision of other HSBLA values.


src/utilities/roundNumberToDecimalPlaces.ts
```diff
@@ -0,0 +1,5 @@
+// precise rounding https://www.jacklmoore.com/notes/rounding-in-javascript/
+export function roundNumberToDecimalPlaces(value: number, decimals: number) {
+  // @ts-ignore - string concatenation represents exponential number notation
+  return Number(Math.round(value + 'e' + decimals) + 'e-' + decimals);
+}
```
MLDimitry, line 5: On my continued exploration of floating point fun, I found that `toFixed(2)` (which uses `Math.round` for it's rounding) does not accurately round all floating point numbers.

Here's an interesting [blog post on rounding](https://medium.com/swlh/how-to-round-to-a-certain-number-of-decimal-places-in-javascript-ed74c471c1b8) which led me to [this solution](https://www.jacklmoore.com/notes/rounding-in-javascript/) which converts the numbers to exponential notation for more accurate rounding.

MLDimitry:
The 2 Percy snapshot failures look like an improvement, since video thumbnails now load, when previously they didn't? Could someone more familiar take a look?

MLDimitry:
Noticing this still fails for #A76850 so investigating that.

MLDimitry:
I was pushing updates to get failing tests/linting/TypeScript issues fixed and they look like they all are now.

This is ready for review, and if someone could also take a look at the two failing Percy screenshots (I think they can be approved but hoping for 2nd more contextful opinion), that'd be great.

MLDimitry:
Bumping this for a review, hopefully y'all have some time around the Alpha Polaris launch. Thanks! 

cc @kyledurand @BPScott @kaelig @AndrewMusgrave

AndrewMusgrave (approved):
Thanks for the articles! Makes sense to me, nice job 💯  You'll want someone from the new design language to take a look as well though

src/utilities/roundNumberToDecimalPlaces.ts
```diff
@@ -0,0 +1,7 @@
+// precise rounding https://www.jacklmoore.com/notes/rounding-in-javascript/
```
BPScott, line 1: That context on why toFixed is really useful and I'd hate for it to get lost to history. Lets put it in as a comment.

```suggestion
// Because everything is a float in JS, Number.toFixed sometimes rounds in the
// "wrong" direction because of float imprecision. For instance:
// `(1.005).toFixed(2)` is `1.00`, NOT `1.01` because 1.005 in floating point is
// actually 1.004999995. By using exponentiation tricks here we can work around
// this imprecision, so `roundNumberToDecimalPlaces(1.005)` returns the expected
// value of `1.01`
// See https://www.jacklmoore.com/notes/rounding-in-javascript/
```

BPScott (changes requested):
Great sleuthing!

Two little requests for comment adjusting, then we're good to go.

MLDimitry:
Thanks for the comments, reviews, and suggestions. I've applied them and resolved new conflicts. This is ready for another review when folks can.

BPScott (approved):
Thanks! :shipit:

PR #3644, opened 2020-11-18
[PortalsManager] Expose component and add ability to customize container

dfmcphee:

Some consuming applications need more control over how the portals get rendered.

This refactors the `PortalsManager` to allow it to be used in a consuming application to set an alternative element for the portals to render into. It also exposes the component in our exports.

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist


src/utilities/portals/PortalsContainer.tsx
```diff
@@ -1,15 +0,0 @@
...
-  const portalsManagerContext = useContext(PortalsManagerContext);
-
-  if (!portalsManagerContext) {
-    return null;
-  }
-
-  const {setContainerNode} = portalsManagerContext;
-
-  return <div id="PolarisPortalsContainer" ref={setContainerNode} />;
-}
```
clauderic, line 15: You could keep this component, would help avoid duplication:

```ts
import {forwardRef} from 'react';

interface Props {}

export const PortalsContainer = forwardRef<Props, HTMLDivElement>((_props, ref) => {
  return <div id="PolarisPortalsContainer" ref={ref} />;
}
```

src/components/PolarisTestProvider/PolarisTestProvider.tsx
```diff
@@ -104,14 +102,17 @@ export function PolarisTestProvider({
...
                           <FrameContext.Provider value={mergedFrame}>
                             {children}
                           </FrameContext.Provider>
                         </FocusManager>
-                        <PortalsContainer />
-                      </PortalsManagerProvider>
+                        <div
+                          id="PolarisPortalsContainer"
+                          ref={portalsContainerRef}
+                        />
```
clauderic, line 114:
```suggestion
                        <PortalsContainer ref={portalsContainerRef} />
```

src/components/AppProvider/AppProvider.tsx
```diff
@@ -103,10 +102,13 @@ export class AppProvider extends Component<AppProviderProps, State> {
...
                     <MediaQueryProvider>
-                      <PortalsManagerProvider>
+                      <PortalsManager container={this.portalsContainerRef}>
                         <FocusManager>{children}</FocusManager>
-                        <PortalsContainer />
-                      </PortalsManagerProvider>
+                        <div
+                          id="PolarisPortalsContainer"
+                          ref={this.portalsContainerRef}
+                        />
```
clauderic, line 110:
```suggestion
                        <PortalsContainer ref={this.portalsContainerRef} />
```

src/components/PortalsManager/PortalsManager.tsx
```diff
@@ -0,0 +1,23 @@
...
+  return (
+    <PortalsManagerContext.Provider
+      value={{
+        container,
+      }}
+    >
+      {children}
+    </PortalsManagerContext.Provider>
+  );
+}
```
clauderic, line 23: You could also move the complexity of rendering the `PortalsContainer` here if you wanted to avoid duplication. 

```ts
export interface PortalsManagerProps {
  children: React.ReactNode;
  container?: RefObject<PortalsContainerElement>;
}

export function PortalsManager({children, container}: PortalsManagerProps) {
  const portalContainerRef = useRef<PortalsContainerElement>(null);

  return (
    <PortalsManagerContext.Provider
      value={{
        container: container ?? portalContainerRef,
      }}
    >
      {children}
      {container ? null : <PortalContainer ref={portalContainerRef} />}
    </PortalsManagerContext.Provider>
  );
}
```

clauderic (commented):
Looks good overall, this is aligned with what I had in mind. Just a few suggestions open for discussion

src/components/Portal/Portal.tsx
```diff
@@ -21,19 +19,18 @@ export function Portal({
...
+  const portalId = useUniqueId(idPrefix !== '' ? idPrefix : 'Portal');
 
   useEffect(() => {
     if (isMounted) {
       onPortalCreated();
     }
   }, [onPortalCreated, isMounted]);
 
-  return portalsContext && portalsContext.portalsContainerRef
+  return portalsContext && portalsContext.container
```
clauderic, line 30: Should an error be thrown if the portalsContext is undefined?

src/components/PortalsManager/PortalsManager.tsx
```diff
@@ -0,0 +1,38 @@
...
+      ? container.current
+      : portalsContainerRef.current;
+    setCurrentContainer(newContainer);
+  }, [container, portalsContainerRef]);
+
+  return (
+    <PortalsManagerContext.Provider
+      value={{
+        container: currentContainer,
+      }}
```
clauderic, line 32: This needs to be memoized, otherwise the children nested within `PortalsManager` may needlessly re-render if the parent of `PortalsManager` re-renders.

As a rule of thumb, make sure context providers always pass down memoized values.

src/components/PortalsManager/PortalsManager.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const [currentContainer, setCurrentContainer] = useState<
+    PortalsContainerElement
+  >(null);
+
+  useEffect(() => {
+    const newContainer = container
+      ? container.current
+      : portalsContainerRef.current;
+    setCurrentContainer(newContainer);
+  }, [container, portalsContainerRef]);
```
clauderic, line 26: `portalsContainerRef` is a ref, therefore it is referentially equal across re-renders and is not a valid dependency.

clauderic, line 26: This is un-necessary, you should avoid derived state as it makes this component much harder to reason about.

```suggestion
```

src/components/PortalsManager/PortalsManager.tsx
```diff
@@ -0,0 +1,38 @@
...
+export interface PortalsManagerProps {
+  children: React.ReactNode;
+  container?: RefObject<PortalsContainerElement>;
+}
+
+export function PortalsManager({children, container}: PortalsManagerProps) {
+  const portalsContainerRef = useRef<PortalsContainerElement>(null);
+  const [currentContainer, setCurrentContainer] = useState<
+    PortalsContainerElement
+  >(null);
```
clauderic, line 19: You'll need to change this from a ref to a state, because the context provider will not update if PortalsManager doesn't re-render for some other reason when the ref is populated. When you switch it to use state, it guarantees that PortalsManager will re-render when the container node changes.

```suggestion
  const [portalContainerElement, setPortalContainerElement] = useState<
    PortalsContainerElement
  >(container);
```

src/components/PortalsManager/PortalsManager.tsx
```diff
@@ -0,0 +1,38 @@
...
+  }, [container, portalsContainerRef]);
+
+  return (
+    <PortalsManagerContext.Provider
+      value={{
+        container: currentContainer,
+      }}
+    >
+      {children}
+      {container ? null : <PortalsContainer ref={portalsContainerRef} />}
```
clauderic, line 35:
```suggestion
      {container ? null : <PortalsContainer ref={setPortalContainerElement} />}
```

src/components/PortalsManager/PortalsManager.tsx
```diff
@@ -0,0 +1,38 @@
...
+    PortalsContainerElement
+  >(null);
+
+  useEffect(() => {
+    const newContainer = container
+      ? container.current
+      : portalsContainerRef.current;
+    setCurrentContainer(newContainer);
+  }, [container, portalsContainerRef]);
+
```
clauderic, line 27:
```suggestion
const currentContainer = container ?? portalContainerElement;
const contextValue = useMemo(() => ({container: currentContainer}, [currentContainer]));
```

src/components/Portal/Portal.tsx
dfmcphee: Good call. Added a hook to use the context that will error if it isn't there like we do for our other providers.

dfmcphee:
Thanks for the review @clauderic. I applied another round of changes based on your feedback. Can you take another look when you have a chance?

src/components/Portal/Portal.tsx
```diff
@@ -11,30 +11,26 @@ export interface PortalProps {
...
-
 export function Portal({
   children,
   idPrefix = '',
   onPortalCreated = noop,
 }: PortalProps) {
   const isMounted = useIsMountedRef();
-  const portalsContext = useContext(PortalsManagerContext);
+  const portalsManager = usePortalsManager();
+  const {container} = portalsManager ? portalsManager : {container: null};
```
clauderic, line 21: Don't think this is needed, `portalsManager` should always return `{container: PortalsContainerElement}`:

Screen Shot 2020-11-30 at 11 32 01 AM

clauderic (approved):
Looks great 👍  Just a final comment then should be good to go 🚀

PR #3647, opened 2020-11-18
[Tabs] Fix aria-require-* issues & listitem semantic issue

jaxee:

Fixes https://github.com/Shopify/web/issues/34726 - aria-require-child / parent rule
Fixes https://github.com/Shopify/web/issues/34493 - list elements not being used semantically
Fixes https://github.com/Shopify/web/issues/34925 - screen reader tab amount (ex. 1 of 7)

No visual changes to the component.
Added `role='presentation'` to the list elements within the unordered list. This removes semantic meaning from the list element which solves three issues;
- li elements not being used semantically
- aria-require-child/parent rule
- screen reader tab amount (before: tab 1 of 1, after: tab 1 of 7)

***Voice over***
The screen reader now reads `Tab x of x`:
voiceover

### How to 🎩

<details>

```jsx
import React, { useCallback, useState } from "react";

import {Page, Tabs, Card} from '../src';

    []
  );

  const tabs = [
    {
      id: "all-customers",
      content: "All",
      accessibilityLabel: "All customers",
      panelID: "all-customers-content"
    },
    {
      id: "accepts-marketing",
      content: "Accepts marketing",
      panelID: "accepts-marketing-content"
    },
    {
      id: "repeat-customers",
      content: "Repeat customers",
      panelID: "repeat-customers-content"
    },
    {
      id: "prospects",
      content: "Prospects",
      panelID: "prospects-content"
    },
    {
      id: "prospects2",
      content: "Prospects",
      panelID: "prospects-content2"
    },
    {
      id: "prospects3",
      content: "Prospects",
      panelID: "prospects-content3"
    },
    {
      id: "prospects4",
      content: "Prospects",
      panelID: "prospects-content4"
    }
  ];

  return (
    <Page title="Playground">
      
      <Card>
          </Card.Section>
        </Tabs>
      </Card>
    </Page>
  );
}
```
</details>

Copy and paste the above code into your local `Playground.tsx` file. Run your server.
- Test that you can still navigate the tabs correctly with your keyboard
- Test the screen reader, the expected behaviour is that the screen reader will say `Tab 1 of x` instead of `Tab 1 of 1`
- Run the `Accessibility Insights` Chrome extension, ensure the aria-require-* rule does not come up as well as the listitem semantic rule

### 🎩 checklist

* [x] Tested on [multiple browsers](https://help.shopify.com/en/manual/intro-to-shopify/shopify-admin/supported-browsers) (Tested on Chrome, FireFox and Safari)


AndrewMusgrave (approved):
Nice work! This will close a few issues 😍

BPScott:
In #3653 @kyledurand changed these roles to be `tab` instead of `presentation`. I assume because they're rendered inside a   ul with the role `tablist`. What's the better approach?

BPScott:
Kinda bamboozled about why no CI has ran for this PR :/ We should make sure that works before merging this, as I expect this will remove a bunch of expected a11y errors

jaxee:
@BPScott I explained why I think this approach is better in [this thread](https://github.com/Shopify/polaris-react/pull/3653#discussion_r530544424)

I will go fix the merge conflicts and hopefully that will force CI to run properly?

AndrewMusgrave:
> Kinda bamboozled about why no CI has ran for this PR :/ We should make sure that works before merging this, as I expect this will remove a bunch of expected a11y errors

It's quite a few checks that haven't run so I'm guessing it was a github issue. Hopefully, it was a one time issue and pushing again will let them work as expected 🤞

BPScott:
Eeey CI is running!

It is failing because a handful of expected a11y test failures are no longer occurring 🎉 : https://travis-ci.com/github/Shopify/polaris-react/jobs/448195889#L893

You'll need to remove the mentioned lines from this array: https://github.com/Shopify/polaris-react/blob/master/scripts/accessibility-check.js#L88

(We fail when these errors no longer occur, so we know we have to remove them from the list. This avoids a situation where we make a fix, don't remove the "ignore this error" line, and then reintroduce the issue again at a later date)

BPScott (approved):
Nice one! Thanks for digging into this and putting up with my questions!

Looks like we're doing what the [MDN examples](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role) offer, as per https://github.com/Shopify/polaris-react/pull/3653#discussion_r530544424
 the `role=tab` is applied to the button/link directly inside this li that we're giving the role of presentation

ginabak:
This PR also: 

Fixes https://github.com/Shopify/polaris-react/issues/3657 and https://github.com/Shopify/polaris-react/issues/2442 🥳

PR #3651, opened 2020-11-19
[Filters] Removed id from proxy buttons

AndrewMusgrave:

* Fixes internally linked issue (see below)
* Fixes accessibility failures in https://github.com/Shopify/polaris-react/pull/3646

Remove id's from proxy buttons (button that are being used to measure)

### How to 🎩

* Load a filters example
* Check if the id's are removed from the proxy markup


BPScott:
Yay A11y issue fixing up!

CI is failing because it expects you to remove no-longer occurring errors from the expected errors list.

https://travis-ci.com/github/Shopify/polaris-react/jobs/443941780#L893 details what is removed
https://github.com/Shopify/polaris-react/blob/master/scripts/accessibility-check.js#L88 is array you need to remove items from

AndrewMusgrave:
Can't say I've ever had a code fail before it was too accessible 😛

BPScott:
I think I prefer the approach in #3653 where "proxy" is passed in a second option object instead of being jammed into the action object.

kyledurand (approved):
I did my [PR](https://github.com/Shopify/polaris-react/pull/3653) pretty quickly and haven't updated tests or anything yet. Happy to let you ship this, with Bens comment addressed, and rebase mine after. Planning to tackle the rest of these next week.

ashwinnarayanan2001 (approved):

AndrewMusgrave:
Thanks, @kyledurand @BPScott 🙌

PR #3652, opened 2020-11-20
[ThemeProvider] Add prop to force custom properties on elements that render outside of context

kyledurand:

Fixes https://github.com/Shopify/polaris-react/issues/3634

The react render tree has theme provider wrapping portals so that portals can receive that context. 

They're doing sort of the right thing in terms of not rendering the custom properties when the color schemes match i.e. a theme provider with dark color scheme wrapping a popover with a dark color scheme won't render the custom properties because it's determining that it doesn't have to based on context. The problem is that it renders outside of where those properties are set in the DOM

This adds a `rendersOutsideOfContext` prop specifically for portal items that depend on context but also render outside of it.

I've thought of alternate approaches as well:
1. Wrapping this around portal provider. This doesn't work because it's still outside of where the context is set. `rendersOutsideOfContext` needs to be set directly on the element that is comparing against context
2. There's no way to be clever about this inside of the ThemeProvider unless we directly checked the DOM

Open to other suggestions

### How to 🎩

Playground included:

<details>

```jsx

import {ActionList, Frame, Button, Popover, ThemeProvider, Toast} from '../src';

  const [, setToastActive] = useState(true);

    [],
  );

  const activator = (
      More actions
    </Button>
  );

  return (
    <Frame>
      <h1>dark / dark</h1>
      <ThemeProvider theme={{colorScheme: 'dark'}}>
          <Popover
            active={popoverActive}
            activator={activator}
            onClose={togglePopoverActive}
            colorScheme="dark"
          >
          </Popover>
        </div>
      </ThemeProvider>
      <h1>light / light</h1>
      <ThemeProvider theme={{colorScheme: 'light'}}>
          <Popover
            active={popoverActive}
            activator={activator}
            onClose={togglePopoverActive}
            colorScheme="light"
          >
          </Popover>
        </div>
      </ThemeProvider>
      <h1>light / dark</h1>
      <ThemeProvider theme={{colorScheme: 'light'}}>
          <Popover
            active={popoverActive}
            activator={activator}
            onClose={togglePopoverActive}
            colorScheme="dark"
          >
          </Popover>
        </div>
      </ThemeProvider>
      <h1>dark / light</h1>
      <ThemeProvider theme={{colorScheme: 'dark'}}>
          <Popover
            active={popoverActive}
            activator={activator}
            onClose={togglePopoverActive}
            colorScheme="light"
          >
          </Popover>
        </div>
      </ThemeProvider>
      <Toast onDismiss={() => setToastActive(false)} content="sup" />
    </Frame>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -26,7 +26,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Added focus styles to `CloseButton` in `Modal` ([#3628](https://github.com/Shopify/polaris-react/pull/3628))
 - Fixed `Filters` duplicated `ConnectedFilter` ids ([#3651](https://github.com/Shopify/polaris-react/pull/3651))
 - Fixed `Banner` `secondaryAction` only rendering if `action` is set ([#2949](https://github.com/Shopify/polaris-react/pull/2949))
-- Added a `rendersOutsideOfContext` to `ThemeProvider` for elements that render outside of parent context ([#3652](https://github.com/Shopify/polaris-react/pull/3652))
+- Added a `rendersOutsideOfAppFrame` to `ThemeProvider` for elements that render outside of the DOM tree to their parent context ([#3652](https://github.com/Shopify/polaris-react/pull/3652))
 
 ### Documentation
 
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -26,6 +26,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Added focus styles to `CloseButton` in `Modal` ([#3628](https://github.com/Shopify/polaris-react/pull/3628))
 - Fixed `Filters` duplicated `ConnectedFilter` ids ([#3651](https://github.com/Shopify/polaris-react/pull/3651))
 - Fixed `Banner` `secondaryAction` only rendering if `action` is set ([#2949](https://github.com/Shopify/polaris-react/pull/2949))
+- Added a `rendersOutsideOfContext` to `ThemeProvider` for elements that render outside of parent context ([#3652](https://github.com/Shopify/polaris-react/pull/3652))
 
 ### Documentation
 
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -26,7 +26,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Added focus styles to `CloseButton` in `Modal` ([#3628](https://github.com/Shopify/polaris-react/pull/3628))
 - Fixed `Filters` duplicated `ConnectedFilter` ids ([#3651](https://github.com/Shopify/polaris-react/pull/3651))
 - Fixed `Banner` `secondaryAction` only rendering if `action` is set ([#2949](https://github.com/Shopify/polaris-react/pull/2949))
-- Added a `rendersOutsideOfAppFrame` to `ThemeProvider` for elements that render outside of the DOM tree to their parent context ([#3652](https://github.com/Shopify/polaris-react/pull/3652))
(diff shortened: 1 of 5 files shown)


kyledurand:
cc @clauderic for suggestions. Dom mentioned you ran into a similar issue with online store web. Would this PR fix the [hack](https://github.com/Shopify/online-store-web/pull/8803) you had to add?

clauderic:
@kyledurand This would indeed allow us to remove the hack that we were forced to add. I do have a couple questions though.

- Naming is hard, but I'm not sure I fully understand what "context" `rendersOutsideOfContext` is referring to. For us, the issue isn't so much that it doesn't have access to React context, it's that this is rendered in a different DOM tree.
- Does this new option really belong under the `theme` config, or should it be introduced as a prop on `ThemeProvider`?

BPScott:
Having this as a separate prop rather than as theme config sounds good.

I'd also be interested in a better name for this "rendersOutsideOfContext" speaks to "what this does" but doesn't quite address why this is needed

kyledurand:
> For us, the issue isn't so much that it doesn't have access to React context, it's that it is nested in a subtree that has a parent ThemeContext but is rendered in a different DOM tree where the CSS variables are not hydrated

Yeah! 💯  correct. I'll rename this and move it to a direct prop thanks @clauderic and @BPScott

kyledurand:
Moved to prop called `rendersOutsideOfAppFrame` and added a test

UNRELEASED.md
```diff
@@ -23,6 +23,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Fixed `Modal` header border color ([#3616](https://github.com/Shopify/polaris-react/pull/3616))
 - Added focus styles to `CloseButton` in `Modal` ([#3628](https://github.com/Shopify/polaris-react/pull/3628))
 - Fixed `Banner` `secondaryAction` only rendering if `action` is set ([#2949](https://github.com/Shopify/polaris-react/pull/2949))
+- Added a `rendersOutsideOfContext` to `ThemeProvider` for elements that render outside of parent context ([#3652](https://github.com/Shopify/polaris-react/pull/3652))
```
clauderic, line 26: I think @BPScott's concerns around naming still apply even with this rename:
> "rendersOutsideOfContext" speaks to "what this does" but doesn't quite address why this is needed

Perhaps a prop along the lines of `renderCustomPropertiesDeclaration` that defaults to `false` would better speak to why this is needed?

kyledurand: Oh, I forgot to update unreleased. The actual name of the prop is now `rendersOutsideOfAppFrame` which I guess doesn't directly tell the user what the prop does. We could go with your suggestion or also be super concise and call it something like `forceRenderCustomProperties`

dfmcphee: I agree this should be something more explicit like `forceRenderCustomProperties` or `alwaysRenderCustomProperties`. I think that will be more clear to understand for consumers.

clauderic: +1 for `alwaysRenderCustomProperties` 🚢

kyledurand: Love it, probably better to stay away from the word force as well. I've updated

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -27,12 +27,18 @@ interface ThemeProviderThemeConfig extends Discard<ThemeConfig, 'colorScheme'> {
 interface ThemeProviderProps {
   /** Custom logos and colors provided to select components */
   theme: ThemeProviderThemeConfig;
+  /**
+   * For use in components such as portals which render outside of parent context
+   * Forces the render of custom properties
```
clauderic, line 32:
```suggestion
   * By default, Polaris avoids re-declaring custom properties within the same React tree
   * This prop ensures that the CSS custom properties are always rendered. This is useful
   * for components such as portals that render outside of the root DOM node
```

clauderic (approved):

dfmcphee (approved):

PR #3653, opened 2020-11-20
Fix up some automated a11y checks

kyledurand:

Part of https://github.com/Shopify/polaris-react/issues/3565

- Fixes up some components that have aria labels that don't actually apply
- Remove duplicate ids from READMEs
- Only apply the multiline role to text field when applicable since aria-multiline=false is not necessary

## 

### How to 🎩

Check the a11y tab in the chroma link for affected components:

- dual thumb range slider
- resource list
- scrollable
- select
- spinner
- tabs
- text field

### 🎩 checklist


src/components/TextField/TextField.tsx
```diff
@@ -430,8 +430,10 @@ export function TextField({
     'aria-activedescendant': ariaActiveDescendant,
     'aria-autocomplete': ariaAutocomplete,
     'aria-controls': ariaControls,
-    'aria-multiline': normalizeAriaMultiline(multiline),
     'aria-expanded': ariaExpanded,
+    ...(normalizeAriaMultiline(multiline) && {
```
BPScott, line 434: Instead of doing object spread stuff would it be possible to change `normalizeAriaMultiline` so if it is undefined, it will return undefined instead of [changing to false](https://github.com/Shopify/polaris-react/blob/62c9cbb8ad45cd9e95639f862966d07222cc93c3/src/components/TextField/TextField.tsx#L530-L531)?

src/components/Scrollable/Scrollable.tsx
```diff
@@ -130,6 +130,8 @@ export class Scrollable extends Component<ScrollableProps, State> {
             {...scrollable.props}
             {...rest}
             ref={this.setScrollArea}
+            // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
```
BPScott, line 133: Adding a little comment here too say why we know better than eslint would be handy for the future

src/components/Tabs/components/Tab/Tab.tsx
```diff
@@ -127,7 +127,7 @@ export function Tab({
   );
 
   return (
-    <li className={tabContainerClassNames} ref={node}>
+    <li className={tabContainerClassNames} ref={node} role="tab">
```
BPScott, line 130: In #3647 @jaxee changed this to `role="presentation"`. Which is better? I'm guessing `role=tab` is more correct?

BPScott (commented):
Love that big block of removing expected errors :D

Dropped a few little things inline

src/components/Tabs/components/Tab/Tab.tsx
AndrewMusgrave, line 130: @jaxee Has a pull request adding `role=presentation` - I think we'll want `role=tab` on the div

src/components/TextField/TextField.tsx
AndrewMusgrave, line 434: How do you feel about changing the return type of the first case instead? https://github.com/Shopify/polaris-react/blob/a5cb04685a8404201cf15957bfdf2dedbd7f6504/src/components/TextField/TextField.tsx#L529

src/components/Scrollable/Scrollable.tsx
```diff
@@ -130,6 +130,8 @@ export class Scrollable extends Component<ScrollableProps, State> {
...
+            tabIndex={0}
```
AndrewMusgrave, line 134: Did we want scrollable in the tab order or should this be -1? (focusable but not in the tab order)

src/components/Tabs/components/Tab/Tab.tsx
jaxee: @BPScott It's a little more complicated than this.

The `role="tab"` is present on the [Unstyled Link / Button](https://github.com/Shopify/polaris-react/blob/master/src/components/Tabs/components/Tab/Tab.tsx#L115) - which is a child of the list element. I believe that having the `role="tab"` on the link / button to be more correct than having it on the list element. 

With this change - another a11y error will pop up saying that the button / unstyled link doesn't have a parent with the "tablist" role. 

Totally open to discussion on this 😄

src/components/ResourceList/ResourceList.tsx
```diff
@@ -675,10 +675,10 @@ export const ResourceList: ResourceListType = function ResourceList<ItemType>({
 
   const loadingOverlay = loading ? (
     <>
-      <div className={styles.SpinnerContainer} style={spinnerStyle}>
+      <li className={styles.SpinnerContainer} style={spinnerStyle}>
```
AndrewMusgrave, line 678: Unfortunately, this markup isn't in the UL list 😢

src/components/Tabs/components/Tab/Tab.tsx
jaxee: @AndrewMusgrave see [this thread](https://github.com/Shopify/polaris-react/pull/3653#discussion_r530537915) 😄

AndrewMusgrave: From what I understand we'll want both of them. The `ul` has a role of `tablist` so the list items are presentational (e.g role=presentation) and the anchor will have `role=tab`

AndrewMusgrave: Mozilla actually uses [tabs as their example for the presentation role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_presentation_role) 😄

jaxee: Awesome! My PR does exactly what Mozilla is doing 😄

BPScott: #3647 has been merged, this can be removed if you rebase atop latest master

BPScott: #3647 has been merged, this can be removed if you rebase atop latest master

kyledurand: Awesome work gang!

kyledurand:
Splitting these up into separate PRs

PR #3659, opened 2020-11-25
[TopBar] Add accessibility label prop

ashwinnarayanan2001:
This pull request fixes #Shopify/polaris-ux#454. It adds an accessibilityLabel prop to the UserMenu and Menu components of the TopBar, and correctly tests that this is working.

## 

### How to 🎩
In order to test this, open this branch locally and go to the DetailsPage on Polaris storybook. Using 'VoiceOver' from System Preferences and when you tab on the 'UserMenu', it should say 'User menu'.

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist


src/components/TopBar/components/UserMenu/tests/UserMenu.test.tsx
```diff
@@ -12,6 +12,7 @@ describe('<UserMenu />', () => {
     initials: '',
     open: false,
     onToggle: noop,
+    accessibilityLabel: 'User menu',
```
kyledurand, line 15: This can be removed as we typically only use defaultProps for non-optional additions that have to be included in every test. This also isn't testing anything, so you'll need to add a test, similar to the one for Menu.test.tsx somewhere in this file

BPScott:
This doesn't seem like it addresses Shopify/polaris-ux#454? That issue talks about items having unneeded alt text on logos and aria labels on Avatars, yet this issue adds the ability to add an aria-label to `TopBar.UserMenu` - an unrelated field. Is that the right issue?


This PR allows consumers to be able to set the aria-label that appears on the "toggle" button that causes the off-screen menu to fly out and overlay on small screens. What is the use-case for allowing per-app customization of that? In what cases would the behaviour of this button need to be described differently on a per-app basis? If there isn't any then do we need this to be a configurable prop, instead should we define the aria-label text we want to use for this button our translations, to avoid increasing our public API surface area and putting a burden onto consumers?

BPScott (commented):
This doesn't seem like it addresses Shopify/polaris-ux#454? That issue talks about items having unneeded alt text on logos and aria labels on Avatars, yet this issue adds the ability to add an aria-label to TopBar.UserMenu - an unrelated field. Is that the right issue?

This PR allows consumers to be able to set the aria-label that appears on the "toggle" button that causes the off-screen menu to fly out and overlay on small screens. What is the use-case for allowing per-app customization of that? In what cases would the behaviour of this button need to be described differently on a per-app basis? If there isn't any then do we need this to be a configurable prop, instead should we define the aria-label text we want to use for this button our translations, to avoid increasing our public API surface area and putting a burden onto consumers?

kyledurand:
Hey @BPScott Ashwin is away and I have context on this

It does target the correct issue. The problem is that screen readers read out button (avatar label and text) content by default unless an aria label is set. So on this menu in web it would read something like 🎤 Button: Avatar with initials KD Kyle Durand 🎤 , which gives no context about what the button actually does.

In adding a custom prop for a label we could pass in something more useful like 🎤  Button: User menu 🎤. It would ignore the avatar label and user name.

There's no problem with the avatar label when it's in the right context.

alex-page (approved):

kyledurand (approved):

PR #3661, opened 2020-11-27
[Navigation] Add aria attributes for sub nav items

kyledurand:

Fixes https://github.com/Shopify/polaris-ux/issues/448

Always render sub nav items so that aria-controls has an id to reference
Adds aria-expanded true/false when nav items have sub items
Add area-controls attribute with ID and corresponding id to Collapsible

Converted Navigation Item tests to use `mountWithApp` and added aria tests
Added missing tests for Secondary

### How to 🎩

Check out the details page or any navigation examples and make sure that only nav items with sub navs have the two attributes mentioned above. Make sure that aria-expanded is true when the sub nav menu is expanded, and false when it isn't

### 🎩 checklist

### DIFF
--- a/src/components/Navigation/Navigation.scss
+++ b/src/components/Navigation/Navigation.scss
@@ -265,9 +265,12 @@ $disabled-fade: 0.6;
 $secondary-item-font-size: rem(15px);
 .SecondaryNavigation {
   flex-basis: 100%;
-  margin-bottom: spacing(tight);
   margin-left: nav(icon-size) + spacing(loose);
   overflow-x: var(--p-override-visible, hidden);
+
+  &.isExpanded {
+    margin-bottom: spacing(tight);
+  }
 
   .Navigation-newDesignLanguage & {
     margin-left: 0;
--- a/src/components/Navigation/components/Item/Item.tsx
+++ b/src/components/Navigation/components/Item/Item.tsx
@@ -16,6 +16,7 @@ import {Indicator} from '../../../Indicator';
 import {UnstyledLink} from '../../../UnstyledLink';
 import {useI18n} from '../../../../utilities/i18n';
 import {useMediaQuery} from '../../../../utilities/media-query';
+import {useUniqueId} from '../../../../utilities/unique-id';
 import styles from '../../Navigation.scss';
 
 import {Secondary} from './components';
@@ -83,6 +84,7 @@ export function Item({
 }: ItemProps) {
   const i18n = useI18n();
   const {isNavigationCollapsed} = useMediaQuery();
+  const secondaryNavigationId = useUniqueId('SecondaryNavigation');
   const {location, onNavigationDismiss} = useContext(NavigationContext);
   const [expanded, setExpanded] = useState(false);
   const [keyFocused, setKeyFocused] = useState(false);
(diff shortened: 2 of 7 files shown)


src/components/Navigation/components/Item/Item.tsx
```diff
@@ -230,19 +232,20 @@ export function Item({
 
   let secondaryNavigationMarkup: ReactNode = null;
 
-  if (subNavigationItems.length > 0 && showExpanded) {
```
kyledurand, line 233: This is the scariest thing for me in this PR. This makes sub menus and their items render in the DOM, though hidden. I did this so that the aria-controls id will have a valid reference ID but not sure that's super necessary to see until the sub menu has actually rendered.

This broke the Nav in web the last time I tried this, for adding animations but Alex's fixes to [Collapsible](https://github.com/Shopify/polaris-react/pull/3606) has fixed that.

Anyone see any potential issues in doing this?

src/components/Navigation/Navigation.scss
```diff
@@ -261,10 +261,13 @@ $disabled-fade: 0.6;
 $secondary-item-font-size: rem(15px);
 .SecondaryNavigation {
   flex-basis: 100%;
-  margin-bottom: spacing(tight);
   margin-left: nav(icon-size) + spacing(loose);
   overflow-x: var(--p-override-visible, hidden);
 
+  &.isExpanded {
+    margin-bottom: spacing(tight);
```
AndrewMusgrave, line 268: Nice!

src/components/Navigation/components/Item/Item.tsx
AndrewMusgrave: > This makes sub menus and their items render in the DOM, though hidden. I did this so that the aria-controls id will have a valid reference ID but not sure that's super necessary to see until the sub menu has actually rendered.

Hmm, interesting! I'm not sure about aria-control, but certain screen readers require elements to always be in the DOM for certain attributes (e.g aria-live)

> Anyone sees any potential issues in doing this?

As long as the `Collapsible` changes roll out ok and collapsible hides the children from keyboard access I don't foresee any isssues.

AndrewMusgrave (commented):
🎩  LTGM ✅

AndrewMusgrave (approved):

BPScott (approved):
I was a little scared by the size of this PR but then I looked and found out most of it was migrating tests to use `@shopify/react-testing`!

I've not got much context on this but it looks like it does what you say it should :shipit:

src/components/Navigation/components/Item/components/Secondary/Secondary.tsx
```diff
@@ -7,12 +7,17 @@ import styles from '../../../../Navigation.scss';
...
-export function Secondary({children, expanded}: SecondaryProps) {
-  const id = useUniqueId('SecondaryNavigation');
+export function Secondary({id, children, expanded}: SecondaryProps) {
+  const uid = useUniqueId('SecondaryNavigation');
   return (
-    <Collapsible id={id} open={expanded}>
+    <Collapsible
+      id={id || uid}
+      open={expanded}
+      transition={{duration: '0ms', timingFunction: 'linear'}}
```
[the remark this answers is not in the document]
kyledurand: Collapsible default to an animation so I'm cancelling that here with 0ms.

st-baonguyen:
@kyledurand @AndrewMusgrave 
Excuse me! I have a case about expanded navigation. 
I have 2 item in subNavigation, and i want to expanded subNavigation when I click main menu and subNavigation, I tried by use `expanded and onToggleExpandedState property` but it is not expanded, When I click subNavigation, subMenu is unexpanded and it just expanded when `selected of main menu = true`. But I don't want use selected property. Pls help me. Thank you so much

kyledurand:
@st-baonguyen can you create a playground or file an issue with more context?

st-baonguyen:
@kyledurand Sorry about this delay. 
I cann't share my source but I had create a playground on Codesanbox. Pls check it. Thank you
Could you check at line 40. 
`Expended = true`, subNavigation doesn't expended, but with `selected = true`, subNavigation expended.
And I don't know what is Expended used for?.
https://codesandbox.io/s/navigation-sandbox-forked-svqgdn?file=/Nav.js

kyledurand:
I see what you mean @st-baonguyen, thanks for the playground. 

The Nav was ported over from the original rails implementation in ~2017, and we needed to copy the functionality from there. 

We might have changed the way the nav works along the way and `expanded` is no longer needed. We'd have to check every implementation and prop combination used internally to be sure and we don't have the bandwidth for that right now. 

Another way to achieve the expanded state is to set `matches: true`, or `matchPaths: ['<path1>', <path2>]`. 

This will also give you the selected style, however, which is what we want for our use case.

PR #3666, opened 2020-11-30
[Button] Add iconPosition support to button component

robertmeessen:

Fixes https://github.com/Shopify/polaris-react/issues/3665

In this web feature request https://github.com/Shopify/intl-shipping-methods/issues/910 it is proposed to have a connected button to paste tracking numbers directly into the text field. The clipboard icon is displayed after the button content. See below:

image
 
Add iconPosition support to display icons before or after the button content.

Before / After:

image

### 🎩 checklist

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -43,6 +43,8 @@ export interface ButtonProps extends BaseButton {
   monochrome?: boolean;
   /** Icon to display to the left of the button content */
   icon?: React.ReactElement | IconSource;
+  /** Position of the Icon, defaults to left */
+  iconPosition?: 'left' | 'right';
   /** Disclosure button connected right of the button. Toggles a popover action list. */
   connectedDisclosure?: ConnectedDisclosure;
 }
@@ -104,6 +106,7 @@ export function Button({
   onMouseEnter,
   onTouchStart,
   icon,
+  iconPosition = 'left',
   primary,
   outline,
   destructive,
@@ -193,8 +196,9 @@ export function Button({
     iconMarkup || disclosureIconMarkup ? (
       <span className={styles.Content}>
         {spinnerSVGMarkup}
-        {iconMarkup}
+        {iconPosition === 'left' && iconMarkup}
         {childMarkup}
+        {iconPosition === 'right' && !disclosureIconMarkup && iconMarkup}
         {disclosureIconMarkup}
       </span>
     ) : (
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -41,10 +41,10 @@ export interface ButtonProps extends BaseButton {
   plain?: boolean;
   /** Makes `plain` and `outline` Button colors (text, borders, icons) the same as the current text color. Also adds an underline to `plain` Buttons */
   monochrome?: boolean;
-  /** Icon to display to the left of the button content */
+  /** Icon to display before or after the button content */
   icon?: React.ReactElement | IconSource;
-  /** Position of the Icon, defaults to left */
-  iconPosition?: 'left' | 'right';
+  /** Icon position relative to button content, defaults to before */
+  iconPosition?: 'before' | 'after';
   /** Disclosure button connected right of the button. Toggles a popover action list. */
   connectedDisclosure?: ConnectedDisclosure;
 }
@@ -106,7 +106,7 @@ export function Button({
   onMouseEnter,
   onTouchStart,
   icon,
-  iconPosition = 'left',
+  iconPosition = 'before',
   primary,
   outline,
   destructive,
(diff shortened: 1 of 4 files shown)


kyledurand:
Hey Robert, can you share more context on why this change is important?

robertmeessen:
Hey Kyle, sure I just updated my description above. This PR is to fulfill the design proposal of this feature request: https://github.com/Shopify/intl-shipping-methods/issues/910

robertmeessen:
Regarding the percy results it seems I made something wrong with the icon spacings for existing disclosure buttons. Will look into it tomorrow.

src/components/Button/Button.tsx
```diff
@@ -43,6 +43,8 @@ export interface ButtonProps extends BaseButton {
   monochrome?: boolean;
   /** Icon to display to the left of the button content */
   icon?: React.ReactElement | IconSource;
+  /** Position of the Icon, defaults to left */
+  iconPosition?: 'left' | 'right';
```
BPScott, line 47: Fun question: how does the ordering of the icon work with right-to-left languages? Is there some language we can use here that doesn't assume left-to-right languages?

BPScott (commented):
Can you add an example to the Button README so we can get a demo of this?

src/components/Button/Button.tsx
robertmeessen: Good point!

robertmeessen: @BPScott What do you think about "before" and "after" instead? I also added both cases to the readme. Can you approve the percy build?

src/components/Button/README.md
```diff
@@ -339,6 +339,51 @@ function StretchedDisclosureButton() {
...
+    </Button>
+  );
+}
+```
+
+### Button with icon after content
+
+<!-- example-for: web -->
+
+Adds an icon after the button content
```
kaelig, line 368:
```suggestion
Adds an icon after the button content.
```

src/components/Button/README.md
```diff
@@ -339,6 +339,51 @@ function StretchedDisclosureButton() {
 }
 ```
 
+### Button with icon
+
+<!-- example-for: web -->
+
+Adds an icon before the button content
```
kaelig, line 346:
```suggestion
Adds an icon before the button content.
```

BPScott:
Heya @robertmeessen, thanks for your patience.

Button is one of those horrible components that is so fundamental and has so many combinations of props it's suprisingly tough to change in way that works everywhere. This gets doublely fun as @beefchimi is also poking at Button improvements in #3709. I've spent this afternoon chatting with @aveline and I think we've got a plan.

We've been thinking about disclosure icons and how really they're just a specific case of icons at the end of text, and if there's a way this could be done in a more general sense. How does this sound:

- Instead of `iconPosition` have a new prop called `iconSuffix` that shall always place the icon after the text - where the disclosure icon is now.

This would help @beefchimi's goal of having a being able to customize the disclosure button - as he wants to add select as an option - instead it he could do `<Button iconSuffix={SelectMinor}>` rather than needing to add a new value to `disclosure`

Then in @beefchimi's Button PR, he could take that work, combine it with his work that removes Disclosure's animations and we end up at a point where `<Button iconSuffix="CaretUpMinor">` and <Button disclosure="up">` are identical, and where `<Button iconSuffix="CaretDownMinor">` and <Button disclosure="down">` are identical. At which point we could start thinking about deprecating the `disclosure` prop.

beefchimi:
@BPScott I am very enthusiastic about your plan, and actually had that in mind as a longer term vision for `Button` 😄 

With this in mind, I'm happy to start reverting some commits in my PR if you desire?

robertmeessen:
> @BPScott I am very enthusiastic about your plan, and actually had that in mind as a longer term vision for `Button` 😄
> 
> With this in mind, I'm happy to start reverting some commits in my PR if you desire?

dleroux:
Not to muddy the waters here, and it might just be me but this:

Screen Shot 2020-12-16 at 9 24 18 AM

Was very confusing to me. I think it's because we read right to left, I saw this as "paste clipboard". My first thought was will clicking the clipboard copy and clicking paste, paste? While the other way around feel more like the icon is a bullet of some kind 🤷

BPScott:
Had a big conversation with @robertmeessen, @beefchimi and @aveline: We've came to the conclusion that:

- Having an icon as a suffix of the text feels odd when scanning the line - as a suffix it feels like it is a distinct action from the rest of the text.
- Disclosure icons are separate to general-purpose suffix icons, so we shouldn't try and merge those two concepts together at the moment (though in the future we might want a `disclosureIcon` prop that hints it should be specifically for hinting at disclosures rather than as a general-purpose thing)


Thus we don't really want either iconSuffix (to add an arbitrary icon) or iconPosition (to allow for controlling the icon position). @robertmeessen shall go back to his designers and this pushback and check if it is ok to update his designs to the icon is a suffix (i.e. use the current polaris-react behaviour)

robertmeessen:
Update from design: I can go without the icon suffix feature so this PR can be closed.

PR #3668, opened 2020-11-30
Add accessibility label to link component

ErickTamayo:

Fixes https://github.com/Shopify/polaris-ux/issues/458 

Adds the `accessibilityLabel` prop to the `Link` component

## 

### How to 🎩

- Add the `accessibilityLabel` prop to the `Link` component and observe the `aria-label` prop is added to the component

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

* [X] Tested for [accessibility](https://github.com/Shopify/polaris-react/blob/master/documentation/Accessibility%20testing.md)
* [X] Updated the component's `README.md` with documentation changes
* [X] [Tophatted documentation](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting%20documentation.md) changes in the style guide


src/components/Link/tests/Link.test.tsx
```diff
@@ -123,4 +123,32 @@ describe('<Link />', () => {
       });
     });
   });
+
+  describe('accessibilityLabel', () => {
+    it('passes prop to the button url is not provided', () => {
+      const mockAccessibilityLabel = 'mock accessibility label';
+      const link = mountWithAppProvider(
```
ashwinnarayanan2001, line 130: It may be best to use the mountWithApp that has been already imported at the top of the file.

src/components/Link/tests/Link.test.tsx
```diff
@@ -123,4 +123,32 @@ describe('<Link />', () => {
...
+        <Link accessibilityLabel={mockAccessibilityLabel} />,
+      );
+
+      expect(link.find('button').prop('aria-label')).toBe(
```
ashwinnarayanan2001, line 134: We'll have to fix our expect line once we've made the change with mountWithApp. We may want to say something like: expect(link).toContainReactComponent('button', {'aria-label': mockAccessibilityLabel, });

src/components/Link/tests/Link.test.tsx
```diff
@@ -123,4 +123,32 @@ describe('<Link />', () => {
...
+        mockAccessibilityLabel,
+      );
+    });
+
+    it('passes prop', () => {
+      const mockAccessibilityLabel = 'mock accessibility label';
+
+      const link = mountWithAppProvider(
```
ashwinnarayanan2001, line 142: We'll also want to apply the same changes that was mentioned above.

src/components/UnstyledLink/tests/UnstyledLink.test.tsx
```diff
@@ -57,5 +57,17 @@ describe('<UnstyledLink />', () => {
       ).find('a');
       expect(anchorElement.prop('download')).toBeFalsy();
     });
+
+    describe('accessibilityLabel', () => {
+      it('passes prop', () => {
+        const mockAccessibilityLabel = 'mock accessibility label';
+        const anchorElement = mountWithAppProvider(
```
ashwinnarayanan2001, line 64: Similarly, we'll want to use mountWithApp, and do the same changes that was suggested for Link.test.tsx.

ashwinnarayanan2001 (changes requested):
Looks good! Just a few changes that need to be made for the test files.

src/utilities/link/types.ts
```diff
@@ -8,6 +8,8 @@ export interface LinkLikeComponentProps
   external?: boolean;
   /** Makes the browser download the url instead of opening it. Provides a hint for the downloaded filename if it is a string value. */
   download?: string | boolean;
+  /** Descriptive text to be read to screenreaders */
+  accessibilityLabel?: string;
```
BPScott, line 12: LinkLikeComponentProps / UnstyledLink are a funky corners of our setup, as they act as the interface for customizing how client-side routing gets handled. Because of that I think we want to stick as as close to behaving like a boring old `<a>` tags as possible, as evidenced by the props extending from `React.HTMLProps<HTMLAnchorElement>`

Because of that, in this one case of `UnstyledLink` I think we should stick to the api that already exists and we should use `aria-label`.

src/components/UnstyledLink/UnstyledLink.tsx
```diff
@@ -20,11 +20,18 @@ export const UnstyledLink = memo(
       return <LinkComponent {...unstyled.props} {...props} />;
     }
 
-    const {external, url, ...rest} = props;
+    const {external, url, accessibilityLabel, ...rest} = props;
     const target = external ? '_blank' : undefined;
     const rel = external ? 'noopener noreferrer' : undefined;
     return (
-      <a target={target} {...rest} href={url} rel={rel} {...unstyled.props} />
+      <a
```
BPScott, line 27: Per other comment, Let's keep with using `aria-label` as the prop here, which I believe means we don't need any of this - as it'll already be covered by the `{...rest}`

BPScott (changes requested):
It's very non-obvious, but I think we should keep UnstyleLink accepting `aria-label` instead of renaming for the sake of it, see comments inline.

ErickTamayo:
Closed in favour of https://github.com/Shopify/polaris-react/pull/3691

PR #3669, opened 2020-11-30
[Tabs][a11y]: Fix a11y keyboard interaction (when focus overrides selected state)

### LINKED ISSUE #3638: [New DL] Focused selected Tab has gray underline instead of green
## Issue summary

When the selected Tab is focused, it has a gray underline instead of a green one.

Screen Shot 2020-11-16 at 3 07 24 PM

## Expected behavior

When focused, the selected Tab should have a green underline.

## Actual behavior

When focused, the selected Tab has a gray underline.

https://screenshot.click/20-11-r7as7-l0gcl.mp4

1. Select a Tab
2. Tab to focus on it

## Reduced test case

Ensure that ``` <AppProvider i18n={en} features={{ newDesignLanguage: true }}>```
https://codesandbox.io/s/suspicious-liskov-2twtr?file=/index.js

## Specifications

- Polaris version number: 5.10.0
- Browser: Chrome
- Device: Mac
- Operating System: MacOS 10.15.7

ginabak:
Fixes #3638 (when selected Tab is focused (on keyboard), it has a grey underline instead the primary green)

A11y related: when auditing [online-store-ui](https://github.com/Shopify/online-store-ui/)'s `PanelCard` component (see Initial Audit [#2291](https://github.com/Shopify/online-store-ui/issues/2291)) we noticed that during the manual keyboard testing, when the selected tab is focused, the `::before` underline was grey instead of the primary green `--p-action-primary` [colour](https://github.com/Shopify/polaris-react/blob/master/documentation/Color%20system.md#primary). 

#### BEFORE 
beforeTabs

#### AFTER
afterTabs

Please note the following: 

- the changes follow the `pre` new design language traits: 
     - focus state is a lighter hue from the selected state instead of grey (pre changes)
     - the grey underline is **_only_** during a hover state
preDesignTab

### How to 🎩

1. Test new changes after pulling branch in [this playground](http://localhost:6006/iframe.html?id=all-components-tabs--all-examples&contexts=New%20Design%20Language%3DEnabled%20-%20Light%20Mode&viewMode=story) (post `dev s`) 

### 🎩 checklist

     * tested on Google Pixel 2 XL (with TalkBack on Android)✅
     * tested on Windows10 Edge ✅
     * tested on FireFox ✅
     * tested on Chrome ✅
     * tested on Safari ✅
     * tested Voiceover on MacOS (Safari) ✅
     * tested JAWs on Windows (Chrome) ✅
     * tested Windows High Contrast (Edge) ❓ (going to post screenshot - not sure what the best is possibly a thicker border)
     * tested Zoom on MacOS (FireFox) ❓(see comment below)


UNRELEASED.md
```diff
@@ -29,6 +27,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Fixed `Banner` `secondaryAction` only rendering if `action` is set ([#2949](https://github.com/Shopify/polaris-react/pull/2949))
 - Added a `alwaysRenderCustomProperties` to `ThemeProvider` for elements that render outside of the DOM tree to their parent context ([#3652](https://github.com/Shopify/polaris-react/pull/3652))
 - Increased precision of hue, saturation, lightness, and alpha in HSBLA `color-transformers` (https://github.com/Shopify/polaris-react/pull/3640)
+- Fixed keyboard interactions for the `Tab` component ([#3650](https://github.com/Shopify/polaris-react/pull/3650))
+- Fixed keyboard interaction when selected Tab was focused and rendering the wrong `::before` colour ([#3669](https://github.com/Shopify/polaris-react/pull/3669))
```
ginabak, line 31: This was my unreleased portion. I just moved the other ones and change the verb tenses to be `past tense` as per the [Polaris Documentation Guideline](https://github.com/Shopify/polaris-react/blob/master/documentation/Versioning%20and%20changelog.md)

src/components/Tabs/Tabs.scss
```diff
@@ -328,11 +328,10 @@ $focus-state-box-shadow-color: rgba(0, 0, 0, 0.8);
   .Tab {
     padding: spacing(tight) spacing(extra-tight);
 
-    &:focus .Title::before {
-      background: var(--p-border-hovered);
-    }
-
     &:active .Title::before {
+      background: var(--p-action-primary-pressed);
```
ginabak, line 332: I made the color to be `--p-action-primary-pressed` so it matched the format PRE new design language as you can see in this gif:

preDesignTab

src/components/Tabs/Tabs.scss
```diff
@@ -343,6 +342,11 @@ $focus-state-box-shadow-color: rgba(0, 0, 0, 0.8);
   }
 
   .Tab-selected {
+    outline: 3px solid transparent;
```
ginabak, line 345: the reason I added this was for Windows High Contrast Mode:

This was BEFORE: 
Capture

This is what it looks like AFTER:


{3E68B6E5-DDEE-4D6A-AD7B-01026F4465A3} png

ginabak:
🎩 on Zoom on MacOS (FireFox):
ZoomTabs

... not sure how one accesses the three dots?! 🤯

src/components/Tabs/Tabs.scss
kyledurand: I would just wrap this in a media query so it only applies to high contrast mode. We can also be explicit about the color we use here by including our high contrast color mixin

```suggestion
    @media (-ms-high-contrast: active) {
      outline: 3px solid ms-high-contrast-color('text');
    }
```

`ms-high-contrast-color('text')` this will return the highest contrast value according to whatever the background color is set to. If the background is black, text will always be white etc... so we use this for borders / outlines and such.

kyledurand (approved):
This looks awesome! I just left a comment around the high contrast stuff. Let me know if you have any questions about it 😄 

Thanks Gina!

arapehl (approved):
Apart from Kyle's media query suggestion, everything else LGTM.

ginabak:
With the high contrast media-query, this is what it looks like and looks much more clean in my opinion! 👀 
YAY!

{8BF34ED7-1577-4136-A9B1-1BE28F7424DE} png

src/components/Tabs/Tabs.scss
ginabak: oooh, I like this 👀 😄

src/components/Tabs/Tabs.scss
```diff
@@ -328,13 +328,12 @@ $focus-state-box-shadow-color: rgba(0, 0, 0, 0.8);
   .Tab {
     padding: spacing(tight) spacing(extra-tight);
 
+    &:active .Title::before {
+      background: var(--p-action-primary-pressed);
+    }
```
beefchimi, line 333: Nit: Space between blocks

src/components/Tabs/Tabs.scss
```diff
@@ -343,6 +342,9 @@ $focus-state-box-shadow-color: rgba(0, 0, 0, 0.8);
   }
 
   .Tab-selected {
+    &:focus .Title::before {
+      background: var(--p-action-primary);
+    }
```
beefchimi, line 347: Nit: Space between blocks.

src/components/Tabs/Tabs.scss
```diff
@@ -351,6 +353,7 @@ $focus-state-box-shadow-color: rgba(0, 0, 0, 0.8);
       }
 
       @media (-ms-high-contrast: active) {
+        outline: 3px solid ms-high-contrast-color('text');
```
beefchimi, line 356: Where does the `3px` come from? Is this informed by somewhere else in Polaris? If so, is there a way to share that code? If not, you should at least wrap `3pm` in the `rem()` function.

beefchimi (commented):
Quick comments.

src/components/Tabs/Tabs.scss
ginabak: this was not informed by Polaris, I just did a quick search and I see on `ActionList.scss` that they use `1px`: 

https://github.com/Shopify/polaris-react/blob/f1e8dcbfc049cca0fa1e9fa2323dcdc3c10b5897/src/components/ActionList/ActionList.scss#L209

src/components/Tabs/Tabs.scss
```diff
@@ -351,6 +355,7 @@ $focus-state-box-shadow-color: rgba(0, 0, 0, 0.8);
       }
 
       @media (-ms-high-contrast: active) {
+        outline: rem(3px) solid ms-high-contrast-color('text');
```
ginabak, line 358: @beefchimi 👀 how's this?

beefchimi (approved):
Have not been able to tophat, but code looks good 👍

src/components/Tabs/Tabs.scss
DanGodfrey: We seem to be using this approach, media query + HC mixin, in a lot of places so I am all in favour of using it for consistency. 

That said, it comes with some tradeoffs. The media query only works in Microsoft browsers so a user using WHCM in Chrome would still see the before experience. 

The issue with the transparent outline is that even though the CSS property visually has no impact on the user, it is still rendered by the browser even when its not needed.

DanGodfrey (commented):
🎉 awesome work 🙌

emma-boardman:
> In the GIF above, I'm using arrow keys but when I get to the last one it goes back to the first tab ... 🤭
> ... not sure how one accesses the three dots?! 🤯 OR ... if that's okay? I have a feeling it's not.

I think this is ok - I would expect the three dots to be accessible via the `tab` key, rather than an arrow key. The arrow keys in this pattern is for moving specifically between the tab titles (the elements with `role=tab`). The three dots are button, rather than a tab title. 

If I'm tabbing through the element in your gif as a whole, my first tab would focus the first `role=tab` element (which I can then navigate within using the arrow keys), and my second tab would focus the three-dot button (which I can then open using enter or space).

PR #3674, opened 2020-12-02
[Filters] Make Filters Query Text Field Optional [#SOPS-81]

umar-k:

Whenever a `Filters` component is used, it always includes the search text input field. 
Text search doesn't make sense in all contexts that a `Filters` can be used in.

Related Feature Request
https://github.com/Shopify/polaris-react/issues/3671

This PR adds an optional boolean prop named `hideQueryField` which hides the query text field.

Here is the how the Filters component currently looks:
Current view of Filters Component

Here is what the Filters component looks with the query text field hidden:
Filters Component with the query field hidden

## 

### How to 🎩

You can use any of the existing `Filters` examples and add the `hideQueryField` prop. 
This should hide the query filter text field.

**NOTE: I tried tophatting the documentation but I don't see my new example (`Filters with query field hidden`) showing as an example in the drop-down. I'm not sure if this is expected or if I'm doing something wrong.**

### 🎩 checklist

### DIFF
--- a/src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
+++ b/src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
@@ -78,6 +78,19 @@ $stacking-order: (
   }
 }
 
+.RightContainer.queryFieldHidden {
+  .Item:first-of-type > * > * {
+    border-top-left-radius: var(
+      --p-border-radius-base,
+      border-radius()
+    ) !important;
+    border-bottom-left-radius: var(
+      --p-border-radius-base,
+      border-radius()
+    ) !important;
+  }
+}
+
 .RightContainerWithoutMoreFilters {
   .Item:last-child > * > * {
     border-top-right-radius: var(
--- a/src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
+++ b/src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
@@ -123,7 +123,7 @@ $stacking-order: (
 }
 
 .MoreFiltersButtonContainer.onlyButtonVisible * {
-  border-radius: var(--p-border-radius-base) !important;
+  border-radius: var(--p-border-radius-base, border-radius()) !important;
 }
 
 .Wrapper {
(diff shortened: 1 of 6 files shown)


src/components/Filters/tests/Filters.test.tsx
```diff
@@ -85,6 +85,20 @@ describe('<Filters />', () => {
     expect(onQueryFocus).toHaveBeenCalledTimes(1);
   });
 
+  it('hides the TextField when "hideQueryField" is "true"', () => {
```
kyledurand, line 88:
```suggestion
  it('does not render the TextField when "hideQueryField" is "true"', () => {
```

src/components/Filters/tests/Filters.test.tsx
```diff
@@ -85,6 +85,20 @@ describe('<Filters />', () => {
...
+    const filters = mountWithAppProvider(
+      <Filters {...mockProps} hideQueryField />,
+    );
+
+    expect(filters.find(TextField).exists()).toBe(false);
+  });
+
+  it('shows the TextField when "hideQueryField" is falsy', () => {
```
kyledurand, line 96:
```suggestion
  it('renders the TextField when "hideQueryField" is false', () => {
```

src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
```diff
@@ -78,6 +78,19 @@ $stacking-order: (
   }
 }
 
+.RightContainer.queryFieldHidden {
+  .Item:first-of-type > * > * {
```
kyledurand, line 82: Can we be more specific to what we're targeting here?

kyledurand (commented):
Code looks good just some small comments.

It looks like you cloned Polaris to make this change, which makes tophatting a bit more challenging. Would you mind creating a branch within the repo?

src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
umar-k: I updated this to `.Item > div > button` since that's the structure of all of our items.

I also updated all of the existing `.Item > * > *` selectors to `.Item > div > button` as well so that if the DOM structure changes, all of these selectors break. Hopefully this makes it easier to avoid any UI regressions.

PR #3683, opened 2020-12-07
[Filters] Make Filters Query Text Field Optional [#SOPS-81]

umar-k:

Whenever a `Filters` component is used, it always includes the search text input field. 
Text search doesn't make sense in all contexts that a `Filters` can be used in.

Related Feature Request
https://github.com/Shopify/polaris-react/issues/3671

Original PR that was closed:
https://github.com/Shopify/polaris-react/pull/3674

This PR adds an optional boolean prop named `hideQueryField` which hides the query text field.

Here is the how the Filters component currently looks:
Current view of Filters Component

Here is what the Filters component looks with the query text field hidden:
Filters Component with the query field hidden

## 

### How to 🎩

You can use any of the existing `Filters` examples and add the `hideQueryField` prop. 
This should hide the query filter text field.

**NOTE: I tried tophatting the documentation but I don't see my new example (`Filters with query field hidden`) showing as an example in the drop-down. I'm not sure if this is expected or if I'm doing something wrong.**

### 🎩 checklist

### DIFF
--- a/src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
+++ b/src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
@@ -51,8 +51,8 @@ $stacking-order: (
 
   &.right.newDesignLanguage {
     .CenterContainer * {
-      border-top-right-radius: var(--p-border-radius-base); // check
-      border-bottom-right-radius: var(--p-border-radius-base); // check
+      border-top-right-radius: var(--p-border-radius-base);
+      border-bottom-right-radius: var(--p-border-radius-base);
     }
   }
 }
@@ -94,8 +94,8 @@ $stacking-order: (
 
 .newDesignLanguage .RightContainer {
   .Item:first-of-type > div > button {
-    border-top-left-radius: var(--p-border-radius-base); // check
-    border-bottom-left-radius: var(--p-border-radius-base); // check
+    border-top-left-radius: var(--p-border-radius-base);
+    border-bottom-left-radius: var(--p-border-radius-base);
   }
 }
 
--- a/src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
+++ b/src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
@@ -78,6 +78,19 @@ $stacking-order: (
   }
 }
 
+.RightContainer.queryFieldHidden {
+  .Item:first-of-type > * > * {
+    border-top-left-radius: var(
+      --p-border-radius-base,
+      border-radius()
+    ) !important;
+    border-bottom-left-radius: var(
+      --p-border-radius-base,
+      border-radius()
+    ) !important;
+  }
+}
+
 .RightContainerWithoutMoreFilters {
   .Item:last-child > * > * {
     border-top-right-radius: var(
--- a/src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
+++ b/src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
@@ -123,7 +123,7 @@ $stacking-order: (
 }
 
 .MoreFiltersButtonContainer.onlyButtonVisible * {
-  border-radius: var(--p-border-radius-base) !important;
+  border-radius: var(--p-border-radius-base, border-radius()) !important;
 }
 
 .Wrapper {
--- a/src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
+++ b/src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
@@ -5,6 +5,7 @@
 // stylelint-disable selector-max-combinators
 // stylelint-disable selector-max-specificity
 // stylelint-disable selector-max-compound-selectors
+// stylelint-disable selector-max-type
 // 🐦🐀
 $stacking-order: (
   item: 10,
@@ -67,7 +68,7 @@ $stacking-order: (
 .RightContainer {
   display: flex;
 
-  .Item > * > * {
+  .Item > div > button {
     margin-right: var(--p-button-group-item-spacing, 0);
     margin-left: var(--p-override-none, -1px);
     border-radius: 0 !important;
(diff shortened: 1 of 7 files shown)


src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
```diff
@@ -67,7 +68,7 @@ $stacking-order: (
 .RightContainer {
   display: flex;
 
-  .Item > * > * {
+  .Item > div > button {
```
kyledurand, line 71: Are these always rendered in a button group? I wonder if we could use our segmented data attribute here i.e.
```
.Item > [data-buttongroup-segmented='true'] {
  .Button {}
}
```

src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
```diff
@@ -78,8 +79,21 @@ $stacking-order: (
   }
 }
 
+.RightContainer.queryFieldHidden {
+  .Item:first-of-type > div > button {
+    border-top-left-radius: var(
+      --p-border-radius-base,
+      border-radius()
+    ) !important;
```
kyledurand, line 87: With all of this specificity do we still need `!important`?

src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.tsx
```diff
@@ -112,7 +115,7 @@ export class ConnectedFilterControl extends Component<
     const moreFiltersButtonContainerClassname = classNames(
       styles.MoreFiltersButtonContainer,
       actionsToRender.length === 0 &&
-        newDesignLanguage &&
+        (newDesignLanguage || queryFieldHidden) &&
```
kyledurand, line 118: Why do we need this?

src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.tsx
```diff
@@ -148,9 +151,11 @@ export class ConnectedFilterControl extends Component<
         {proxyButtonMarkup}
         <div className={styles.Wrapper}>
           <div className={className} ref={this.container}>
-            <div className={styles.CenterContainer}>
-              <Item>{children}</Item>
-            </div>
+            {children ? (
```
kyledurand, line 154: Maybe add a test for this

kyledurand (commented):
This is looking pretty good 👏

Just some minor comments

src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
umar-k: Looks like we never actually use the Polaris `ButtonGroup` component in `Filters`.

We create individual `Button`s using the `activatorButtonFrom` function on `Line 234` in `ConnectedFilterControl.tsx` and then use CSS to override the border-radius / spacing

src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
umar-k: Good suggestion! 
Had to change a few of the selectors but was able to remove all of the `!important` selectors 😄

src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.tsx
umar-k: If the query field is hidden AND only the `More Filters` button is showing, we have to round the left borders.
Previously we didn't have to do that because the in the old design, the `More Filters` button would have been up against the text field.

**With Query Field Hidden:**
image



**Without Query Field Hidden:**
image

src/components/Filters/components/ConnectedFilterControl/ConnectedFilterControl.scss
umar-k: I think it would make sense to use a `ButtonGroup` so that we can delete a lot of the CSS in here.
I can try to tackle that in a future PR 👍

BPScott:
I'm not a designer, but what if we keep the filter buttons aligned to the right? For some pulling them over to the left feels odd if the input element isn't there.

umar-k:
> I'm not a designer, but what if we keep the filter buttons aligned to the right? For some pulling them over to the left feels odd if the input element isn't there.

@BPScott I talked about this with the Designer on our team. 

We were going to with with right-aligning the filters but decided to go with left-aligned because once you start applying filters, it looks like they're floating in a lot of empty space.
image

vs

image

kyledurand:
Ben is back on Monday, can this wait for his final review?

umar-k:
Yeah for sure 👍 
Sorry, didn't realize Ben was away

src/components/Filters/components/ConnectedFilterControl/tests/ConnectedFilterControl.test.tsx
```diff
@@ -202,6 +202,24 @@ describe('<ConnectedFilterControl />', () => {
       disabled: true,
     });
   });
+
+  it('does not render CenterContainer when no child element is "null"', () => {
+    const connectedFilterControl = mountWithAppProvider(
+      <ConnectedFilterControl>{null}</ConnectedFilterControl>,
+    );
+
+    expect(connectedFilterControl.find('.CenterContainer')).toHaveLength(0);
```
BPScott, line 211: One glorious day we'll complete the move from enzyme to `@shopify/react-testing` and we'll be able to write things like `expect(connectedFilterControl).not.toContainReactComponent('div', {className: 'CenterContainer'})` and it will be glorious.

Today is not that day though.

(Ignore this, nothing to do here, this is just me rambling)

BPScott (approved):
Looks all good to me :shipit: 

Some of that CSS looks brittle and scary but is basically the same as what we did before so that's not on you. Agree with the thoughts of revisiting that so it can leverage ButtonGroup (and probably enhancing ButtonGroup with `connectedLeft` / `connectedRight` props to match the existing `connectedTop`) but that's a story for another PR.

Thanks for the contribution, merge when you're ready!

umar-k:
Thanks for the reviews @kyledurand @BPScott 👍 !!
Had to resolve merge conflict in `UNRELEASED.md`. Will merge soon as all tests pass again

PR #3685, opened 2020-12-07
[WIP] [OptionList] [a11y] Switch singleSelect to input type='radio'

### LINKED ISSUE #3358: a11y - Enhance OptionList accessibility
## Issue summary

When `allowMultiple=false`, we're rendering DOM elements that aren't the best choice in terms of accessibility.

## Expected behavior

Since the most semantic way of doing this is either rendering a `select` or an `input[type=radio]`, I believe we need to adjust the code so that instead of rendering the `button` we have a new `RadioBox` component that piggybacks a lot from the code already present on `Checkbox`.

## Actual behavior

Currently, an `OptionList` that's not `allowMultiple` will render a `button` for each one of the options being listed, but it has no indication for SRS (Screen Reading Software) that the specific option is the one selected.

On the other hand, if `allowMultiple=true`, what gets rendered in place is a `Checkbox` component, which makes a lot of sense when it comes to the DOM semantics.

Mount the element toggling the `allowMultiple` to see the different DOM that gets rendered and using a SRS try to pinpoint which of the options are selected.

## Reduced test case

https://codesandbox.io/s/optimistic-golick-qi7pg?file=/App.js

## Specifications

- Polaris version number: latest
- Browser: Chrome/Safari
- Device: Desktop
- Operating System: Mac OS

emma-boardman:
Fixes #3358 

### What this PR is doing

- Converts the singleSelect `OptionList` to an `<input type="radio">` for improved accessibility. (the component currently renders a list of buttons). 

### Questions

- is this the right approach?  
- should the label for the RadioButton be moved into the component, to avoid selectors such as ` [type='radio'] + label` in the Option.tsx file? this would make the styling tidier, but hamper the readability of the `allowMultiple ?` structure, which currently highlights parallels between the media/label markup. It would also mean the label styles would be repeated in two files. Maybe these could be a mixin? 
- still need to add disabled styles and tidy up old styles

### 🎩 checklist

### DIFF
--- a/src/components/OptionList/components/Checkbox/Checkbox.tsx
+++ b/src/components/OptionList/components/Checkbox/Checkbox.tsx
@@ -62,7 +62,6 @@ export function Checkbox({
         checked={checked}
         disabled={disabled}
         className={inputClassName}
-        aria-checked={checked}
         onChange={onChange}
         onBlur={handleBlur}
         onKeyUp={handleKeyUp}
--- a/src/components/OptionList/components/Option/Option.scss
+++ b/src/components/OptionList/components/Option/Option.scss
@@ -134,20 +134,15 @@ $control-vertical-adjustment: rem(2px);
   }
 
   .SingleSelectOption {
-    // stylelint-disable-next-line selector-max-combinators, selector-max-compound-selectors
-    [type='radio'] + label {
-      @include focus-ring;
+    position: relative;
+    @include focus-ring;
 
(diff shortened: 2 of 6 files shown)


beefchimi (commented):
I started reading over the `OptionList` component to better understand it. Turns out, this has some rather complicated areas of code. I started jotting down a few notes as I went along, but decided it might be best for us to just pair on some parts of this PR and we can better understand it together.

How about we save this for our pairing session later in the week?

PR #3686, opened 2020-12-07
[Button] Add new prop to support `aria-describedby`

### LINKED ISSUE #3684: [Button] Add new prop to support `aria-describedby`
## Feature request summary

The `Button` component currently does not support `aria-describedby`.  

## Rationale

The `Button` component supports an `aria-label`, which describes the essence of the button. In specific scenarios, the `aria-label` may become overwhelming with too much information. The extra information would be better conveyed using `aria-describedby`.

For example in Email, we have a button for section duplication with an `aria-label: "Duplicate section"`. Adding more context (ex: `Position ${section.index}. Type {section.type}.`) to the `aria-label` would become overwhelming. Since the extra information is mostly a description, `aria-describedby` is more appropriate. 

Moreover, I think it is good to note that SRS have a small break in between announcing the label and description.

image

---

ben10code:

Fixes #3684 

`Button` currently supports `aria-label` via `accessibilityLabel`, but does not support `aria-describedby`. 

Introduces a new prop `ariaDescribedBy` to the `Button` component to support `aria-describedby`.

### 🎩 checklist


alex-page (approved):
This looks good, I would worry though that this is fixing the wrong problem. Buttons shouldn't have a lot of text in them.

arthurgouveia:
> this is fixing the wrong problem. Buttons shouldn't have a lot of text in them.

Thanks @alex-page, but the thing is that depending on the button location & UI constraints, the `aria-describedby` will serve to add the context necessary for that button to users of Screen Reading Software. Visually, this will not impact the Button.

arthurgouveia (approved):

PR #3688, opened 2020-12-07
[TextField] Include focus-ring on clear button && Fix clear button in popovers

voiid:

Fixes https://github.com/Shopify/polaris-react/issues/3687

Fixes https://github.com/Shopify/web/issues/34755

1. Adds missing focus-ring for clear button when using the new design language.
2. Fixes an issue with clear buttons when used in popovers. 
[The popover overlay `handleClick` function](https://github.com/Shopify/polaris-react/blob/6c39f4a00d7431d275d1441230814dca04f0be07/src/components/Popover/components/PopoverOverlay/PopoverOverlay.tsx#L248) determines whether or not a click was inside the popover to decide if it should close the popover or not. The clear button would inadvertently cause this to happen because the button disappears as soon as it is clicked, causing the `isDescendent` check to fail and the popover to be closed. This change ensures that the `clearButton` remains in the DOM after being clicked.

## 

### How to 🎩

<details>

```jsx

import {
  AppProvider,
  Page,
  Button,
  Popover,
  TextField,
  FormLayout,
} from '../src';

  const [value, setValue] = useState('');
  const [active, setActive] = useState(false);

  const toggleActive = () => setActive((active) => !active);

  return (
    <AppProvider features={{newDesignLanguage: true}}>
      <Page title="Playground">
        <Popover
          active={active}
          activator={<Button onClick={toggleActive}>Click me</Button>}
          onClose={toggleActive}
        >
          <Popover.Pane sectioned>
            <FormLayout>
              <TextField
                clearButton
                onClearButtonClick={() => setValue('')}
                label="label"
                labelHidden
                type="search"
                placeholder="Some placeholder"
                onChange={setValue}
                value={value}
              />
            </FormLayout>
          </Popover.Pane>
        </Popover>
        <TextField
          clearButton
          onClearButtonClick={() => setValue('')}
          label="label"
          labelHidden
          type="search"
          placeholder="Some placeholder"
          onChange={setValue}
          value={value}
        />
      </Page>
    </AppProvider>
  );
}
```

</details>

### 🎩 checklist

* [X] Tested on [multiple browsers](https://help.shopify.com/en/manual/intro-to-shopify/shopify-admin/supported-browsers)
* [X] Tested for [accessibility](https://github.com/Shopify/polaris-react/blob/master/documentation/Accessibility%20testing.md)
* [X] Updated the component's `README.md` with documentation changes
* [X] [Tophatted documentation](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting%20documentation.md) changes in the style guide
* [N/A] For visual design changes, pinged one of @ HYPD, @ mirualves, @ sarahill, or @ ry5n to update the Polaris UI kit

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -35,6 +35,18 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Removed `aria-multiselectable` from OptionList ([#3729](https://github.com/Shopify/polaris-react/pull/3729))
 - Replaced `button` with `div` in `RangeSlider` for correct semantics when using `role="slider"` ([#3730](https://github.com/Shopify/polaris-react/pull/3730))
 - Replaced `React.Fragment` with `li` in `ResourceList` spinner for valid markup ([#3732](https://github.com/Shopify/polaris-react/pull/3732))
+- Replace `button` with `div` in `RangeSlider` for correct semantics when using `role="slider"` ([#3730](https://github.com/Shopify/polaris-react/pull/3730))
+- Added `flush` prop to `Card.Section` ([#3645](https://github.com/Shopify/polaris-react/pull/3645))
+- Added `stretchContent` prop for `Button` ([#3664](https://github.com/Shopify/polaris-react/pull/3664))
+- Extracted `TagsWrapper` from `Filters` for testability ([#3688](https://github.com/Shopify/polaris-react/pull/3688))
+
+### Bug fixes
+
+- Fixed `ClearButton` in `TextField` unintentionally closing `Popover` when clicked ([#3688](https://github.com/Shopify/polaris-react/pull/3688))
+- Added focus styles to `ClearButton` in `TextField` ([#3688](https://github.com/Shopify/polaris-react/pull/3688))
+- Added dependency list to useImperativeHandle in `Banner` ([#3478](https://github.com/Shopify/polaris-react/pull/3478))
+- Internationalize `Badge` labels ([#3655](https://github.com/Shopify/polaris-react/pull/3655))
+- Aligned the `::before` 'indicator' to edge of container for `ActionList` ([#3619](https://github.com/Shopify/polaris-react/pull/3619))
 
 ### Documentation
 
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -35,18 +35,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Removed `aria-multiselectable` from OptionList ([#3729](https://github.com/Shopify/polaris-react/pull/3729))
 - Replaced `button` with `div` in `RangeSlider` for correct semantics when using `role="slider"` ([#3730](https://github.com/Shopify/polaris-react/pull/3730))
 - Replaced `React.Fragment` with `li` in `ResourceList` spinner for valid markup ([#3732](https://github.com/Shopify/polaris-react/pull/3732))
-- Replace `button` with `div` in `RangeSlider` for correct semantics when using `role="slider"` ([#3730](https://github.com/Shopify/polaris-react/pull/3730))
-- Added `flush` prop to `Card.Section` ([#3645](https://github.com/Shopify/polaris-react/pull/3645))
-- Added `stretchContent` prop for `Button` ([#3664](https://github.com/Shopify/polaris-react/pull/3664))
-- Extracted `TagsWrapper` from `Filters` for testability ([#3688](https://github.com/Shopify/polaris-react/pull/3688))
-
-### Bug fixes
-
-- Fixed `ClearButton` in `TextField` unintentionally closing `Popover` when clicked ([#3688](https://github.com/Shopify/polaris-react/pull/3688))
-- Added focus styles to `ClearButton` in `TextField` ([#3688](https://github.com/Shopify/polaris-react/pull/3688))
-- Added dependency list to useImperativeHandle in `Banner` ([#3478](https://github.com/Shopify/polaris-react/pull/3478))
-- Internationalize `Badge` labels ([#3655](https://github.com/Shopify/polaris-react/pull/3655))
-- Aligned the `::before` 'indicator' to edge of container for `ActionList` ([#3619](https://github.com/Shopify/polaris-react/pull/3619))
+- Fixed clear button in `TextField` unintentionally closing `Popover` when clicked ([#3688](https://github.com/Shopify/polaris-react/pull/3688))
+- Added focus styles to the clear button in `TextField` ([#3688](https://github.com/Shopify/polaris-react/pull/3688))
 
 ### Documentation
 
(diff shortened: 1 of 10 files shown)


LauraAubin:
It looks like when you click the clear button and it becomes visually hidden, its is still tab focusable. Not sure if this is intended but I wonder if we could apply a tab index of -1 whenever we are visually hiding this button?

src/components/Filters/Filters.tsx
```diff
@@ -430,7 +431,7 @@ class FiltersInner extends Component<CombinedProps, State> {
         <div className={styles.Filters}>
           {filtersControlMarkup}
           {filtersContainerMarkup}
-          {tagsMarkup}
+          <span testID="tagsContainer">{tagsMarkup}</span>
```
voiid, line 434: I would love suggestions for how to better target the `TagsWrapper` to test that it renders as a `VisuallyHidden` component. The test below that matches this change broke with the TextField change because it was looking for `VisuallyHidden` _anywhere_ in the Filters component which was quite brittle. 

I did try adding the testID to `TagsWrapper` directly and the test did pass, however Typescript doesn't like it due to the prop not existing on the `VisuallyHidden` component.

The bigger picture fix would probably be to refactor this large components into more easily testable pieces but that's out of the scope of this work.

voiid:
> It looks like when you click the clear button and it becomes visually hidden, it is still tab focusable. Not sure if this is intended but I wonder if we could apply a tab index of -1 whenever we are visually hiding this button?

@LauraAubin Sorry about that, had to change the implementation due to a visual regression on mobile. Fix pushed up.

src/components/Filters/Filters.tsx
LauraAubin: Maybe we could test the logic that uses `VisuallyHidden` instead?

https://github.com/Shopify/polaris-react/blob/46c2c4536979ae58bebc66505540725dae88c4ee/src/components/Filters/Filters.tsx#L359

So we could compare how many `VisuallyHidden` elements there are by changing `appliedFilters`?

voiid: By that, do you mean diff the amount of `VisuallyHidden` elements before/after passing `appliedFilters`? 

My main concern with the original test was how brittle it was due to it checking the entirety of the `Filters` component for instances of `VisuallyHidden` when this logic only impacts a small part of the component. Narrowing down the change to when `appliedFilters` changes would definitely be a huge improvement, just a little concerned by how many different markup elements we're checking against which could result in false flags (either this test passing when it shouldn't if there are changes that could break this, or other false failures like this where the component is still working as intended).

src/components/Filters/Filters.tsx
```diff
@@ -231,6 +231,7 @@ class FiltersInner extends Component<CombinedProps, State> {
       <div ref={this.moreFiltersButtonContainer}>
         <Button
           onClick={this.toggleFilters}
+          id="SheetToggleButton"
```
BPScott, line 235: We need to avoid fixed IDs. Using a fixed ID means we can end up with accessibility warnings about "you have duplicate ids on the page" if you ever render two Filters components on one page.

Can you remove this id and target the element in a different way?

src/components/Filters/components/TagsWrapper/TagsWrapper.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+import {VisuallyHidden} from '../../../VisuallyHidden';
+
+interface Props {
+  children: React.ReactNode;
+  shouldHide: boolean;
```
BPScott, line 7: Gonna be real annoying as I was the one who suggested this in the first place:

Could we rename this prop to `hidden` as we tend to use that tense in other places (titleHidden, labelHidden, ViauallyHidden etc)

src/components/Filters/components/TagsWrapper/tests/TagsWrapper.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+describe('<TagsWrapper />', () => {
+  it('renders visually hidden component when shouldHide is true', () => {
+    const tagsWrapper = mountWithApp(
+      <TagsWrapper shouldHide>
+        <MockChild />
+      </TagsWrapper>,
+    );
+
+    expect(tagsWrapper).toContainReactComponentTimes(VisuallyHidden, 1);
```
BPScott, line 17: How strongly do you feel about needing the count in there? I don't think it adds much value (same for the other assertions in this file).

```suggestion
    expect(tagsWrapper).toContainReactComponent(VisuallyHidden);
```

src/components/Filters/components/TagsWrapper/tests/TagsWrapper.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+    expect(tagsWrapper).toContainReactComponentTimes(MockChild, 1);
```
BPScott, line 18: Is it worth asserting that MockChild is inside the VisuallyHidden? Something like:

```suggestion
    expect(tagsWrapper.find(VisuallyHidden)).toContainReactComponent(MockChild);
```

UNRELEASED.md
```diff
@@ -13,6 +13,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- Fixed `ClearButton` in `TextField` unintentionally closing `Popover` when clicked ([#3688](https://github.com/Shopify/polaris-react/pull/3688))
+- Added focus styles to `ClearButton` in `TextField` ([#3688](https://github.com/Shopify/polaris-react/pull/3688))
```
BPScott, line 17: Do we need a changlog entry for the Filters changes?

BPScott (commented):
Some little nitpicks inline.

Remove the `SheetToggleButton` id and rename the `shouldHide` prop and i think this is good to go!

UNRELEASED.md
```diff
@@ -10,9 +10,12 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - Added `flush` prop to `Card.Section` ([#3645](https://github.com/Shopify/polaris-react/pull/3645))
 - Added `stretchContent` prop for `Button` ([#3664](https://github.com/Shopify/polaris-react/pull/3664))
+- Extracted `TagsWrapper` from `Filters` for testability ([#3688](https://github.com/Shopify/polaris-react/pull/3688))
```
voiid, line 13: @BPScott Wasn't sure where this should go. Let me know if it needs to be moved/re-worded. :)

BPScott: This feels like a "Code quality" entry - it doesn't have any effect on external users (functionality remains identical) but makes our codebase a bit nicer.

BPScott (commented):
Code looks good. Looks like unit tests are failing thanks to some weirdness in your UNRELEASED.md. It seems like you've got a handful of entries for things that aren't this issue and that are already mentioned in CHANGELOG.md

voiid:
> Code looks good. Looks like unit tests are failing thanks to some weirdness in your UNRELEASED.md. It seems like you've got a handful of entries for things that aren't this issue and that are already mentioned in CHANGELOG.md

I was running into some build failures unrelated to my changes and the automatic conflict resolution went a little crazy when I rebased with master. Should be resolved now. 😅

LauraAubin (approved):

BPScott (approved):
Fantastic, thank you for your contribution! :shipit:

PR #3695, opened 2020-12-09
Adding focus-visible styles with polyfill

dfmcphee:

Currently we use a hack for not showing focus rings on things like buttons by blurring on `mouseup`. This works, but it takes adding this behaviour specifically to every component that needs it.

This adds the `focus-visible` polyfill so that we can start taking a CSS approach to this that will eventually be supported by browsers and we can remove the polyfill at that point.

Right now this would be about a __1kb__ bundle increase, but I think we could get it to about the same once we start removing the custom click/blur code.

Here is an example of clicking or tabbing to a tooltip activator before this change:

Screen Shot 2020-12-09 at 1 15 56 PM

And after this change clicking would show this:
Screen Shot 2020-12-09 at 1 16 21 PM

And tabbing to would show this:
Screen Shot 2020-12-09 at 1 16 13 PM

This adds the polyfill package as well as some default `focus-visible` styles for all elements.

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Tooltip/README.md
+++ b/src/components/Tooltip/README.md
@@ -68,8 +68,8 @@ Use only when necessary to provide an explanation for an interface element.
 
 ```jsx
 <div style={{padding: '75px 0'}}>
-  <Tooltip content="This order has shipping labels.">
-    <Link>Order #1001</Link>
+  <Tooltip active content="This order has shipping labels.">
+    <TextStyle variation="strong">Order #1001</TextStyle>
   </Tooltip>
 </div>
 ```
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -6,8 +6,6 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
-- Added `flush` prop to `Card.Section` ([#3645](https://github.com/Shopify/polaris-react/pull/3645))
-- Added `stretchContent` prop for `Button` ([#3664](https://github.com/Shopify/polaris-react/pull/3664))
 - Added `focus-visible` polyfill and default styles ([#3695](https://github.com/Shopify/polaris-react/pull/3695))
 
 ### Bug fixes
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -6,6 +6,10 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Added `flush` prop to `Card.Section` ([#3645](https://github.com/Shopify/polaris-react/pull/3645))
+- Added `stretchContent` prop for `Button` ([#3664](https://github.com/Shopify/polaris-react/pull/3664))
+- Added `focus-visible` polyfill and default styles ([#3695](https://github.com/Shopify/polaris-react/pull/3695))
+
 ### Bug fixes
 
 - Fixed virtual cursor leaving dialog in `Modal`, `Navigation` and `Sheet` ([#3931](https://github.com/Shopify/polaris-react/pull/3931))
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "@types/react": "^16.9.12",
     "@types/react-dom": "^16.9.4",
     "@types/react-transition-group": "^4.4.0",
+    "focus-visible": "^5.2.0",
     "lodash": "^4.17.4",
     "react-transition-group": "^4.4.1"
   },
--- a/src/components/AppProvider/AppProvider.scss
+++ b/src/components/AppProvider/AppProvider.scss
@@ -62,6 +62,29 @@ body {
   box-sizing: border-box;
 }
 
+:focus:not(:focus-visible) {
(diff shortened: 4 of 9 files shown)


src/components/Tooltip/README.md
```diff
@@ -69,7 +69,7 @@ Use only when necessary to provide an explanation for an interface element.
 ```jsx
 <div style={{padding: '75px 0'}}>
   <Tooltip active content="This order has shipping labels.">
-    <Link>Order #1001</Link>
+    <TextStyle variation="strong">Order #1001</TextStyle>
```
alex-page, line 72: Should we revert this example change?

dfmcphee: I think it is actually useful to have one example that isn't a link here. The link never showed this being an issue because it had its own focus styles.

alex-page (approved):
Looks great, the [NPM package](https://github.com/WICG/focus-visible/blob/master/src/focus-visible.js) is nice and clean as well.

dleroux:
Globally, I think we should just remove the focus styles altogether. Since our styles are not being applied to the element directly, but to the `::after`  selector using the focusing ring, this could cause en element to have 2 focus styles:

Screen Shot 2020-12-09 at 2 22 46 PM

dfmcphee:
> Globally, I think we should just remove the focus styles altogether. Since our styles are not being applied to the element directly, but to the `::after` selector using the focusing ring, this could cause en element to have 2 focus styles:

The problem with this is that for components not getting a specific focus style, just a tabindex like a tooltip activator, they would never have a focus style reapplied.

I think the real problem here is we have two different ways of applying our focus styles. This one adds a box-shadow directly to the focused element, whereas or `focus-ring` mixin applied it to an `::after`. The reason I didn't use the same approach for the global style is that it assumes too much by applying `position: relative` to everything. I wish our focus-ring was a little more flexible to work on elements.

dleroux:
> it assumes too much by applying position: relative

I think that is the reality of the focus ring. We need to make a conscious decision when applying styles to all focusable elements. I don't recall us adding it directly to an element. Do we do that anywhere?

dfmcphee:
> I think that is the reality of the focus ring. We need to make a conscious decision when applying styles to all focusable elements. I don't recall us adding it directly to an element. Do we do that anywhere?

Not currently but that is part of the issue. We end up with inconsistent browser default focus styles on some elements because we don't do this (like the tooltip example above). In my opinion it would be better to have a better base focus style that we can apply globally that respects focus-visible. We could then tweak those focus styles where we need to in specific components, but we also wouldn't leave anything out then.

The alternative is not apply anything globally but fix this up in each individual case but it is going to be difficult to do that in all cases.

dfmcphee:
I pushed up one more commit showing what this would look like without any global styles and only fixing the tooltip activator focus for now.

alex-page:
I think it's worth fixing globally and then going through the components and addressing any technical debt. It will remove a lot of code and hopefully make it a better experience when developing new components or patterns as they will have focus styles by default.

beefchimi:
I implemented a CSS-only solution to using `:focus-visible` in `online-store-ui`, which looks something like this:

```scss
@mixin interaction-focus-ring($width: rem(0)) {
  @include focus-ring($border-width: $width);

  &:focus {
  }

  &:focus:not(:focus-visible) {
    @include interaction-focus-ring-suppress;
  }

  &:focus-visible {
  }
}
```

So far this seems to have worked nicely. Curious what more the polyfill is affording you?

dleroux:
That shouldn't work anywhere but chrome right now: https://caniuse.com/mdn-css_selectors_focus-visible

dfmcphee:
Yeah, exactly. That code would only work in browsers that support `focus-visible` whereas the polyfill would let you also use `[data-focus-visible-added]` as a selector to make that code work in all browsers.

beefchimi:
@daniedleroux seems to work as intended in the browsers I've tested. You can play around in our [Chroma instance](https://5d5b0560de84090020ac0b31-tfzaviljft.chromatic.com/iframe.html?id=playground-playground--playground&viewMode=story) if you like - but you'd kind of need to know what components this code is applied to.

dfmcphee:
> Edit: I think the part that might not be obvious is the use of `:not(:focus-visible)`. I picked up this technique from here:
> https://developer.paciellogroup.com/blog/2018/03/focus-visible-and-backwards-compatibility/

I was actually following that implementation as well when looking at this. This is only backwards compatible in the sense that you will still get focus styles in all browsers, not that they will only be applied with keyboard focus.

For example, take a look a [this codepen](https://codepen.io/dfmcphee/pen/bGwwKzw) in Chrome, Safari, and Firefox.

In Chrome where `focus-visible` is supported, clicking the example heading will not show the custom focus ring, but tabbing to it will. In Safari and Firefox the focus ring shows up both when clicking and tabbing.

kyledurand (approved):

PR #3697, opened 2020-12-09
[Link] Add underline prop

ashwinnarayanan2001:
### WHY are these changes introduced? ### WHAT is this pull request doing?
These changes are introduced to fix Shopify/polaris-ux#455. This pull request adds a 'underline' prop to the Link component, that adds a text-decoration underline and keeps the same link color. 

Screen Shot 2020-12-09 at 4 39 19 PM

## 

### How to 🎩

<details>

```jsx
import React from 'react';

import {Page, Link} from '../src';

  return (
    <Page title="Playground">
      <Link url="https://help.shopify.com/manual" underline>fulfilling orders</Link>
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Link/Link.tsx
+++ b/src/components/Link/Link.tsx
@@ -21,6 +21,8 @@ export interface LinkProps {
   external?: boolean;
   /** Makes the link color the same as the current text color and adds an underline */
   monochrome?: boolean;
+  /** Adds text decoration underline to the link*/
+  underline?: boolean;
   /** Callback when a link is clicked */
   onClick?(): void;
 }
@@ -32,6 +34,7 @@ export function Link({
   external,
   id,
   monochrome,
+  underline,
 }: LinkProps) {
   const i18n = useI18n();
   let childrenMarkup = children;
@@ -63,6 +66,7 @@ export function Link({
           styles.Link,
           shouldBeMonochrome && styles.monochrome,
           newDesignLanguage && styles.newDesignLanguage,
+          underline && styles.underline,
         );
 
         return url ? (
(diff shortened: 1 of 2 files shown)


src/components/Link/Link.tsx
```diff
@@ -21,6 +21,8 @@ export interface LinkProps {
   external?: boolean;
   /** Makes the link color the same as the current text color and adds an underline */
   monochrome?: boolean;
+  /** Adds text decoration underline to the link*/
+  underline?: boolean;
```
alex-page, line 25: This might be controversial but I think the underline link should be the default as it's the most accessible. Having a prop to remove the underline in specific use cases where the user understands it is interactive e.g. a list would be better. Would love to know others thoughts?

kyledurand: I was wondering this as well. I think that's the right decision in the long run. It might be worth adding an underline by default and building Polaris for web to see how much it would be affected by that change. 

If it's quite a bit then we might want to save a change like that for v6 before we fully default to all links being underlined.

BPScott: +1 to "underline by default, allow removing as an option" as giving a better out the box behaviour. By giving a underline prop we just force the problem downstream and now every consumer has to remember to add the underline prop if they want to have better accessibility.

Agree that this feels like a pretty big design change and that it might be best saved for a major version.

alex-page: I would be happy to make this PR a part of v6 and make links underlined by default.

kyledurand:
Should we close this in favour of https://github.com/Shopify/polaris-react/pull/3705 or 🚢  and reverse the prop in v6?

My gut says we should close it as it creates extra work for teams to add these props that they'd have to remove when we ship v6, although I guess we could do that easily enough. Thoughts?

PR #3709, opened 2020-12-14
🐛 + ✨ [Button] Bug fixes, small improvements, new "select" disclosure

beefchimi:
> **Tophat:** I recommend using the provided `Playground` snippet to more easily tophat each combination of `<Button />`

<details>
<summary><strong>Playground code:</strong></summary>

```tsx
import {OnlineStoreMajor} from '@shopify/polaris-icons';

import {Button, Card, ChoiceList, Layout, Page} from '../src';

const childrenOptions = {
  none: undefined,
  short: 'Short text',
  long: 'This long line of text will wrap unless truncate is true',
};
const sizeOptions = {
  none: undefined,
  slim: 'slim',
  medium: 'medium',
  large: 'large',
};
const textAlignOptions = {
  none: undefined,
  left: 'left',
  center: 'center',
  right: 'right',
};
const disclosureOptions = {
  none: undefined,
  up: 'up',
  down: 'down',
  select: 'select',
};
const toggleOptions = [
  'primary',
  'destructive',
  'outline',
  'fullWidth',
  'plain',
  'monochrome',
  'icon',
  'stretchContent',
  'connectedDisclosure',
  'url',
  'external',
  'download',
  'submit',
  'disabled',
  'loading',
  'pressed',
  'role',
  'ariaExpanded',
];

const mockConnectedDisclosure = {
  actions: [
    {
      content: 'Save and mark as ordered',
    },
  ],
};
const mockRole = 'menuitem';
const mockUrl = 'https://shopify.ca';

function mockEvent(label = '') {
  return () => console.log(label);
}

  const [children, setChildren] = useState(['none']);
  const handleChildrenChange = useCallback((value) => setChildren(value), []);

  const [size, setSize] = useState(['none']);
  const handleSizeChange = useCallback((value) => setSize(value), []);

  const [textAlign, setTextAlign] = useState(['none']);
  const handleTextAlignChange = useCallback((value) => setTextAlign(value), []);

  const [disclosure, setDisclosure] = useState(['none']);
  const handleDisclosureChange = useCallback(
    (value) => setDisclosure(value),
    [],
  );

  const [toggles, setToggles] = useState(['icon']);
  const handleToggleChange = useCallback((value) => setToggles(value), []);
  const isPropToggled = (value: string) => toggles.includes(value);

  return (
    <Page title="Button Configuration">
      <Layout>
        <Layout.Section oneThird>
          <Card title="Component">
            <Card.Section title="Result">
              <Button
                id="ButtonAllProps"
                ariaControls="ControlTarget"
                accessibilityLabel="Button a11y text"
                size={sizeOptions[size]}
                textAlign={textAlignOptions[textAlign]}
                disclosure={disclosureOptions[disclosure]}
                icon={isPropToggled('icon') ? OnlineStoreMajor : undefined}
                primary={isPropToggled('primary')}
                destructive={isPropToggled('destructive')}
                outline={isPropToggled('outline')}
                fullWidth={isPropToggled('fullWidth')}
                plain={isPropToggled('plain')}
                monochrome={isPropToggled('monochrome')}
                stretchContent={isPropToggled('stretchContent')}
                connectedDisclosure={
                  isPropToggled('connectedDisclosure')
                    ? mockConnectedDisclosure
                    : undefined
                }
                url={isPropToggled('url') ? mockUrl : undefined}
                external={isPropToggled('external')}
                download={isPropToggled('download')}
                submit={isPropToggled('submit')}
                disabled={isPropToggled('disabled')}
                loading={isPropToggled('loading')}
                pressed={isPropToggled('pressed')}
                role={isPropToggled('role') ? mockRole : undefined}
                ariaExpanded={isPropToggled('ariaExpanded')}
                onClick={mockEvent('onAction')}
                onFocus={mockEvent('onFocus')}
                onBlur={mockEvent('onBlur')}
                onKeyPress={mockEvent('onKeyPress')}
                onKeyUp={mockEvent('onKeyUp')}
                onKeyDown={mockEvent('onKeyDown')}
                onMouseEnter={mockEvent('onMouseEnter')}
                onTouchStart={mockEvent('onTouchStart')}
              >
                {childrenOptions[children]}
              </Button>
            </Card.Section>
            <Card.Section title="Output">
              <p id="ControlTarget">
                <strong>Boolean toggles:</strong> {toggles.join(', ')}
              </p>
            </Card.Section>
          </Card>
        </Layout.Section>

        <Layout.Section oneThird>
          <Card title="Children" sectioned>
            <ChoiceList
              title=""
              choices={Object.keys(childrenOptions).map((prop) => ({
                label: childrenOptions[prop] ?? 'None',
                value: prop,
              }))}
              selected={children}
              onChange={handleChildrenChange}
            />
          </Card>

          <Card title="Size" sectioned>
            <ChoiceList
              title=""
              choices={Object.keys(sizeOptions).map((prop) => ({
                label: sizeOptions[prop] ?? 'None',
                value: prop,
              }))}
              selected={size}
              onChange={handleSizeChange}
            />
          </Card>

          <Card title="Text alignment" sectioned>
            <ChoiceList
              title=""
              choices={Object.keys(textAlignOptions).map((prop) => ({
                label: textAlignOptions[prop] ?? 'None',
                value: prop,
              }))}
              selected={textAlign}
              onChange={handleTextAlignChange}
            />
          </Card>

          <Card title="Disclosure" sectioned>
            <ChoiceList
              title=""
              choices={Object.keys(disclosureOptions).map((prop) => ({
                label: disclosureOptions[prop] ?? 'None',
                value: prop,
              }))}
              selected={disclosure}
              onChange={handleDisclosureChange}
            />
          </Card>
        </Layout.Section>

        <Layout.Section oneThird>
          <Card title="Toggles" sectioned>
            <ChoiceList
              allowMultiple
              title=""
              choices={toggleOptions.map((prop) => ({
                label: prop,
                value: prop,
              }))}
              selected={toggles}
              onChange={handleToggleChange}
            />
          </Card>
        </Layout.Section>
      </Layout>
    </Page>
  );
}
```

</details>

**This PR makes a few small adjustments to the `Button` component:**
- ~Enable `color` transitions, allowing for `newDesignLanguage` "icons" to inherit `color` from their parent and transition accordingly.~
- Revise use of `transition`, as it was inconsistent.
  - Avoid `transition` for "old design language"
  - Enable `transition` for "new design language"
  - Remove use of `--p-override-none`
- Fix issue with `Button` text remaining visible when `newDesignLanguage + loading + plain`.
- No longer do a `180deg` spin on the `disclosure` when toggling between `up/down`.
- Right-align the `disclosure` arrow when working with `textAlignLeft`.
- No longer set `border-width: 0` on "filled" buttons, which was resulting in layout shift when toggling certain boolean props.
- A few minor code adjustments to improve readability.
- New `"select"` option for `disclosure`.

### Screenshots

<details>
<summary><strong>1. Revised transitions:</strong></summary>

transitions

</details>

<details>
<summary><strong>2. `newDesignLanguage + loading + plain`</strong></summary>

button-plain-loading

</details>

<details>
<summary><strong>3. No longer spin `disclosure`:</strong></summary>

disclosure-up-down

</details>

<details>
<summary><strong>4. Right-align `disclosure`:</strong></summary>

disclosure-right

</details>

<details>
<summary><strong>5. Prevent layout-shift from `border-width`:</strong></summary>

border-width-layout

</details>

<details>
<summary><strong>6. Select disclosure:</strong></summary>

Screen Shot 2020-12-15 at 4 06 28 PM

</details>

---

### Questions:

**1. Should we remove the left/right margins when "only disclosure"?**
Its a low-effort fix... but maybe we want to discourage "only disclosure"?

<details>
<summary><strong>Screenshot:</strong></summary>

</details>

**~2. Should `stretchContent` instead be a `textAlign="justified"` option?~**
~Another low-effort fix. Also... why do we need to qualify the styles with `:not(.loading)`?~
Since resolved. Marked for deprecation.

**3. Why were we using `--p-override-none`?**
This was used on `transition-duration` and `transition-timing-function`. Seems to suggest we want to disable transitions in "new design language" but enable for "old design language"? This PR reverses that behaviour while completely removing `--p-override-none` from `Button.scss`. Am I misunderstanding something? Should I revert?

**4. What is with the `.monochrome.outline > ::before` for?**
Seems like its used to provide a "background". So what I don't understand is: why not use `background-color`?

**5. Do we want to document/implement logic for "incompatible props"?**
Several prop combinations are incompatible, such as: `plain + connectedDisclosure` or `primary + destructive`.

**6. Should the deprecated `ariaPressed` not have been removed by now?**
Am I able to get rid of this prop from the `BaseButton` type?

**7. Should we swap `icon` and `disclosure` for `rtl` languages?**
My instinct is "yes". Have we received any feedback about this?

---

### Pre-existing issues (discovered while tophatting):

1. `loading` does not affect `connectedDisclosure`
2. Undesired styles when using: `outline + disabled + connectedDisclosure`
3. Undesired styles when using: `primary + outline`
4. There are surely a few more faulty prop combinations... but I ran out of time tophatting.

cc/ @emma-boardman in case you have some extra time to help tophat.
cc/ @dleroux since you have touched some of the changed code
cc/ @dfmcphee since you also have some expertise/opinions on this component

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -10,19 +10,25 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - Updated `OptionList` selected styles ([#3633](https://github.com/Shopify/polaris-react/pull/3633))
 - Added the ability to hide the clear filter button on the filter component ([#3049](https://github.com/Shopify/polaris-react/pull/3049))
-- **`Popover`:** New `autofocusTarget` prop to enhance autofocus options ([#3600](https://github.com/Shopify/polaris-react/pull/3600))
+- Right-align `disclosure` when using `textAlignLeft` for `Button` ([#3709](https://github.com/Shopify/polaris-react/pull/3709))
+- Remove all transitions from `Button` ([#3709](https://github.com/Shopify/polaris-react/pull/3709))
+- New `select` option for `disclosure` in `Button` ([#3709](https://github.com/Shopify/polaris-react/pull/3709))
 - Conveyed `DatePicker` more clearly to screen readers ([#3660](https://github.com/Shopify/polaris-react/pull/3660))
-- Added `accessibilityLabels` prop to `Pagination` ([#3667] (https://github.com/Shopify/polaris-react/pull/3667))
+- Added `accessibilityLabels` prop to `Pagination` ([#3667](https://github.com/Shopify/polaris-react/pull/3667))
+- New `autofocusTarget` prop to enhance autofocus options on `Popover` ([#3600](https://github.com/Shopify/polaris-react/pull/3600))
 - Added ability to hide query text field in `Filters` component using `hideQueryField` prop ([#3674](https://github.com/Shopify/polaris-react/pull/3674))
 - Added `tabIndex` to `Scrollable` for keyboard focus ([#3744](https://github.com/Shopify/polaris-react/pull/3744))
 - Added accessibility label prop to `UserMenu` and `Menu` subcomponents in `TopBar` ([#3659](https://github.com/Shopify/polaris-react/pull/3659))
 - Add `aria-label` to the `Loading` bar in `Frame` ([#3770](https://github.com/Shopify/polaris-react/pull/3770))
-- **`Button`:** New `ariaDescribedBy` prop for `<button />` ([#3664](https://github.com/Shopify/polaris-react/pull/3686))
 - Updated `Collapsible` to be a functional component ([#3779](https://github.com/Shopify/polaris-react/pull/3779))
 - Coverted `TooltipOverlay` to a functional component ([#3631](https://github.com/Shopify/polaris-react/pull/3631))
+- New `ariaDescribedBy` prop for `Button` ([#3664](https://github.com/Shopify/polaris-react/pull/3686))
 
 ### Bug fixes
 
+- `plain` variant `children` no longer remain visible while `loading` for `Button` ([#3709](https://github.com/Shopify/polaris-react/pull/3709))
+- No longer spin `disclosure` 180deg when toggling between `up` and `down` on `Button` ([#3709](https://github.com/Shopify/polaris-react/pull/3709))
+- Prevent layout shift when toggling “filled” variants on `Button` ([#3709](https://github.com/Shopify/polaris-react/pull/3709))
 - Fixed `FocusManager` from tracking inactive items that prevented trap focusing([#3630](https://github.com/Shopify/polaris-react/pull/3630))
 - Added escape keybind to `Tooltip` ([#3627](https://github.com/Shopify/polaris-react/pull/3627))
 - Removed extra bottom border on the `DataTable` and added curved edges to footers ([#3571](https://github.com/Shopify/polaris-react/pull/3571))
@@ -69,4 +75,5 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Deprecations
 
+- `stretchContent` has been replaced by `textAlign="left" + fullWidth` for `Button` ([#3709](https://github.com/Shopify/polaris-react/pull/3709))
 - Deprecated `Popover`'s prop `preventAutofocus`. Use `autofocusTarget` instead ([#3602](https://github.com/Shopify/polaris-react/issues/3602))
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -97,8 +97,11 @@ $stacking-order: (
   justify-content: center;
   align-items: center;
   min-width: 1px;
-  width: 100%;
   min-height: 1px;
+
+  .Button:not(.plain) & {
+    width: 100%;
+  }
 }
 
 .textAlignLeft {
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -144,7 +144,8 @@ $stacking-order: (
     margin-left: spacing(extra-tight);
   }
 
-  .textAlignLeft &:last-child {
+  // stylelint-disable-next-line selector-max-class, selector-max-specificity
+  .fullWidth.textAlignLeft &:last-child:not(:only-child) {
     margin-left: auto;
   }
 
@@ -608,11 +609,6 @@ $stacking-order: (
   .Icon:last-child {
     margin-right: rem(-4px);
   }
-  // stylelint-disable selector-max-class, selector-max-combinators, selector-max-specificity
-  &:not(.textAlignLeft) .Icon + .Icon:last-child {
-    margin-left: 0;
-  }
-  // stylelint-enable selector-max-class, selector-max-combinators, selector-max-specificity
 
   .Icon:only-child {
     margin-right: 0;
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -50,7 +50,6 @@ $partial-button-filled-pressed-box-shadow: inset 0 0 0 0 transparent,
 
   &:focus,
   &:hover {
-    transition-duration: duration(fast);
     background: linear-gradient(to bottom, $color-light, $color-light);
     border-color: $color-dark;
     box-shadow: $partial-button-filled-pressed-box-shadow $color-dark;
@@ -130,12 +129,6 @@ $stacking-order: (
 }
 
 .Icon {
-  transition: color duration() easing();
-
-  .newDesignLanguage & {
-    transition: none;
-  }
-
   // This compensates for the typical icon used in buttons being
   // inset within its bounding box.
   margin-left: -(spacing(extra-tight));
@@ -155,6 +148,13 @@ $stacking-order: (
 
   + *:not(.Icon) {
     margin-left: spacing(extra-tight);
+  }
+
+  .newDesignLanguage & {
+    // stylelint-disable-next-line selector-max-combinators
+    svg {
+      transition: fill duration() easing();
+    }
   }
 }
 
@@ -273,7 +273,6 @@ $stacking-order: (
 .loading,
 .newDesignLanguage.loading {
   position: relative;
-  transition: border-color duration() easing();
 
   // stylelint-disable-next-line selector-max-class
   &,
@@ -294,7 +293,6 @@ $stacking-order: (
   @include pressed-box-shadow;
 
   &:hover {
-    transition-duration: duration(fast);
     background: $pressed-hover-background;
     border-color: $pressed-border-color;
     @include pressed-box-shadow;
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -141,8 +141,7 @@ $stacking-order: (
     margin-left: spacing(extra-tight);
   }
 
-  // stylelint-disable-next-line selector-max-class, selector-max-specificity
-  .fullWidth.textAlignLeft:not(.stretchContent) &:last-child {
+  .textAlignLeft &:last-child {
     margin-left: auto;
   }
 
@@ -607,7 +606,7 @@ $stacking-order: (
     margin-right: rem(-4px);
   }
   // stylelint-disable selector-max-class, selector-max-combinators, selector-max-specificity
-  .Icon + .Icon:last-child {
+  &:not(.textAlignLeft) .Icon + .Icon:last-child {
     margin-left: 0;
   }
   // stylelint-enable selector-max-class, selector-max-combinators, selector-max-specificity
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -731,16 +731,6 @@ $stacking-order: (
 }
 // stylelint-enable selector-max-specificity
 
-.DisclosureIcon {
-  transition-property: transform;
-  transition-duration: duration(slow);
-  transition-timing-function: easing(out);
-}
-
-.DisclosureIconFacingUp {
-  transform: rotate(-180deg);
-}
-
 .ConnectedDisclosureWrapper {
   display: flex;
 }
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -98,12 +98,17 @@ $stacking-order: (
   justify-content: center;
   align-items: center;
   min-width: 1px;
+  width: 100%;
   min-height: 1px;
 }
 
 .textAlignLeft {
   justify-content: flex-start;
   text-align: left;
+
+  .Content {
+    justify-content: flex-start;
+  }
 }
 
 .textAlignCenter {
@@ -114,11 +119,14 @@ $stacking-order: (
 .textAlignRight {
   justify-content: flex-end;
   text-align: right;
+
+  .Content {
+    justify-content: flex-end;
+  }
 }
 
-.stretchContent:not(.loading) > .Content {
+.stretchContent > .Content {
   justify-content: space-between;
-  width: 100%;
 }
 
 .Icon {
@@ -138,6 +146,11 @@ $stacking-order: (
     // spacing on the right of the button)
     margin-right: -(spacing(tight));
     margin-left: spacing(extra-tight);
+  }
+
+  // stylelint-disable-next-line selector-max-class, selector-max-specificity
+  .fullWidth.textAlignLeft:not(.stretchContent) &:last-child {
+    margin-left: auto;
   }
 
   + *:not(.Icon) {
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -500,6 +500,10 @@ $stacking-order: (
     &.disabled {
       color: var(--p-text-disabled);
       background: none;
+
+      &.loading {
+        color: transparent;
+      }
     }
 
     &.iconOnly {
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -124,6 +124,10 @@ $stacking-order: (
 .Icon {
   transition: color duration() easing();
 
+  .newDesignLanguage & {
+    transition: none;
+  }
+
   // This compensates for the typical icon used in buttons being
   // inset within its bounding box.
   margin-left: -(spacing(extra-tight));
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -155,7 +155,7 @@ $stacking-order: (
   .newDesignLanguage & {
     // stylelint-disable-next-line selector-max-combinators
     svg {
-      transition: fill duration() easing();
+      transition: fill duration(fast) easing();
     }
   }
 }
@@ -675,7 +675,7 @@ $stacking-order: (
 
     &::before {
       content: '';
-      transition: opacity duration() easing();
+      transition: opacity duration(fast) easing();
       position: absolute;
       top: 0;
       right: 0;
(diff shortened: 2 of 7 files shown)


BPScott:
Heya @beefchimi, you might want to check out https://github.com/Shopify/polaris-react/pull/3666#issuecomment-745690844

src/styles/shared/_buttons.scss
```diff
@@ -31,9 +31,6 @@
   cursor: pointer;
   user-select: none;
   text-decoration: none;
-  transition-property: background, border, box-shadow;
-  transition-duration: var(--p-override-none, duration());
```
dleroux, line 35: What's the timeline for removing these? Won't that break apps using the component if they haven't enabled the new design language?

beefchimi: @dleroux was this ever working correctly though? There was a comment in the code about doing _something a certain way_ in order to animate the `background: linear-gradient`, but as far as I know, you can't animate gradients. So we had some properties animating while others were not, adding up to a disjointed experience.

Worth mentioning though - the properties that were animating correctly were things like `border-color` and `box-shadow`, and I think they were hard to notice in the first place. So the experience wasn't terribly inconsistent, unless you were _really paying attention_.

Essentially, I don't really see the drawback of doing this for old design language users. But I completely defer to Polaris here - you folks are going to have way better context than I will 😆

dleroux: Ha yes, It was a safari bug IIRC. @danrosenthal and @tmlayton were involved in that.

> you folks are going to have way better context than I will.

I do have context but I'm not on Polaris anymore :)

src/types.ts
```diff
@@ -94,7 +94,7 @@ export interface BaseButton {
   /** Tells screen reader the controlled element is expanded */
   ariaExpanded?: boolean;
   /**
-   * @deprecated As of release 4.7.0, replaced by {@link https://polaris.shopify.com/components/structure/page#props-pressed}
+   * @deprecated As of release 4.7.0, replaced by {@link https://polaris.shopify.com/components/actions/button#prop-pressed}
```
beefchimi, line 97: This was pointing to the wrong link. But also - as mentioned in my `Questions` section - should we not have removed this already?

src/styles/shared/_buttons.scss
danrosenthal: No issues related to the Safari bug here. That was related to the `linear-gradient` function being passed as a `var` function fallback property.

beefchimi:
@BPScott / @dleroux this is ready for final review 👍🏻

beefchimi:
@robertmeessen + @BPScott this should be ready for final review.

Is there someone that reviews the `percy/codecov` changes and approves them?

robertmeessen (approved):
🎩  Looks good to me, thanks @beefchimi

UNRELEASED.md
```diff
@@ -8,13 +8,19 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - Updated `OptionList` selected styles ([#3633](https://github.com/Shopify/polaris-react/pull/3633))
 - Added the ability to hide the clear filter button on the filter component ([#3049](https://github.com/Shopify/polaris-react/pull/3049))
+- **`Button`:** Right-align `disclosure` when using `textAlignLeft` ([#3709](https://github.com/Shopify/polaris-react/pull/3709))
+- **`Button`:** More consistent transitions ([#3709](https://github.com/Shopify/polaris-react/pull/3709))
+- **`Button`:** New `select` option for `disclosure` ([#3709](https://github.com/Shopify/polaris-react/pull/3709))
```
alex-page, line 13:
```suggestion
- Right-align `disclosure` when using `textAlignLeft` for `Button` ([#3709](https://github.com/Shopify/polaris-react/pull/3709))
- More consistent transitions in `Button` ([#3709](https://github.com/Shopify/polaris-react/pull/3709))
- New `select` option for `disclosure` in `Button` ([#3709](https://github.com/Shopify/polaris-react/pull/3709))
```

src/components/Button/Button.scss
```diff
@@ -136,12 +144,23 @@ $stacking-order: (
...
+  .newDesignLanguage & {
+    // stylelint-disable-next-line selector-max-combinators
+    svg {
+      transition: fill duration() easing();
+    }
+  }
 }
 
-.Hidden {
+.hidden {
```
alex-page, line 163: Any reason for change here?

BPScott:
Thanks for putting up with me. Gonna look at this later today :)

src/styles/shared/_buttons.scss
```diff
@@ -76,6 +72,9 @@
     border: 1px solid var(--p-border-neutral-subdued);
     border-top-color: var(--p-border-subdued);
     border-bottom-color: var(--p-border-shadow-subdued);
+    transition-property: color, background, border, box-shadow;
+    transition-duration: duration();
+    transition-timing-function: easing();
```
alex-page, line 77: Lets remove all transitions. This is not adding any benefit to the user experience right now.

UNRELEASED.md
beefchimi: Will make the change, but I do think its worth considering this style in the future. We write things this way in `online-store-ui` and I find it much easier to parse the Release Notes and identify the components that are relevant to the consuming app.

As a consumer, when I'm going through `Breaking changes`, I can quickly look at the prefixed components and identify what is relevant to my app.

src/styles/shared/_buttons.scss
alex-page: We will spend some time looking into component transitions in the future, for now lets remove them.

beefchimi: I agree that the `duration` was too slow. 👍🏻 

Also agree with the sentiment that for subtle changes, motion can make them _even more subtle_, therefor lessening the discoverability of the change. However, after speeding up the `duration`, I'm not sure that point is as relevant anymore?

If we are committed to completely removing the button transitions - I'd ask that we consider offering some CSS custom property values that consumers can override. That way, Polaris can default to no transition, but consumers can easily override.

If that is a path we want to pursue - and if there is a non-insignificant time investment - I would then ask if we can defer that to a follow-up PR? Getting this PR merged and available to `online-store-ui` would help us unblock ourselves.

src/components/Button/tests/Button.test.tsx
```diff
@@ -434,4 +445,19 @@ describe('<Button />', () => {
       });
     });
   });
+
+  describe('stretchContent', () => {
+    it('sets variant class', () => {
+      const warningSpy = jest
```
BPScott, line 451: Perform mocking and reseting of mocks in a beforeEach / afterEach method rather than within the test block. Eg. https://github.com/Shopify/polaris-react/blob/3d2da225b7c484e446dd1522cab0db0dfbfdbb87/src/components/Modal/components/Dialog/tests/Dialog.test.tsx#L10-L16

This is because if the mockRestore is within the test block and an expectation fails then the mockResore() won't be called and that mock will potentially leak in other test cases, which can lead to hard to diagnose cascading errors - SomeOther.test.tsx failing because it is inadvertently using a mock that was set but not cleared up in Button.test.tsx is hecking confusing (I've been bit by this before, it's a great way to waste an afternoon)

src/types.ts
BPScott: Good spot! Yeah this should have ideally been removed in v5.0.0 but we missed that window. We should try again in v6.0.0

BPScott (commented):
Looking very tasty.

Something we talked about a while back was creating this layout:

Screen Shot 2021-01-06 at 1 15 37 PM

I can create that with:

```jsx
<Button
  icon={OnlineStoreMajor}
  disclosure="select"
  fullWidth
  textAlign="center"
  stretchContent
>
  Some Text
</Button>
```

But i need stretchContent push the disclosure and icon to the edges of the box.

Is this a style of layout we want to support? I don't think we can currently create that if we deprecate and remove stretchContent?

BPScott:
The plot thickens: Things are stretched and the icon and disclosure are at the edges of their boxes out when textAlign is left, but not when right or centered:

Screen Shot 2021-01-06 at 1 27 22 PM

Produced by:

```
<Button
  icon={OnlineStoreMajor}
  disclosure="select"
  fullWidth
  textAlign="left"
>
  align=left
</Button>
<br />
<Button
  icon={OnlineStoreMajor}
  disclosure="select"
  fullWidth
  textAlign="center"
>
  align=center
</Button>
<br />
<Button
  icon={OnlineStoreMajor}
  disclosure="select"
  fullWidth
  textAlign="right"
>
  align=right
</Button>
```

Should we handle that? My initial gut feel is that this is weird, but sitting on it more this might actually be reasonable behaviour - if you get rid either the icon or the disclosure then things feel lopsided.

Perhaps the fix is is "the discource icon is always flush to the right hand edge" and "When you talk about alignment, you consider the icon text as a single block and you align that at the same time"?

BPScott:
Another nitpicky thing:

When specifying textAlign="left" the padding of the disclosure icon changes. It has no effect when set to any other value. I'm thinking the following buttons should all render visually identically

Screen Shot 2021-01-06 at 1 45 50 PM

Produced by:

```
<Button disclosure="select">align</Button>
<br />
<Button disclosure="select">align</Button>
<br />
<Button disclosure="select" textAlign="left">
  align
</Button>
<br />
<Button disclosure="select" textAlign="center">
  align
</Button>
<br />
<Button disclosure="select" textAlign="right">
  align
</Button>
```

src/styles/shared/_buttons.scss
alex-page: I think this PR can remove the transitions, instead of making them consistent. Let me know if you want me to do the work ^

beefchimi: I can easily remove the transitions - I was hoping however we could compromise and go the route of CSS Custom Properties so that Polaris can default to no transitions, but I - a consumer - can more easily enable transitions without digging into underlying markup.

Removing transitions altogether invalidates one of my approaches to utilizing Polaris' `Button`:
https://github.com/Shopify/online-store-ui/pull/2143

To be honest I didn't love that solution anyways _(for various reasons)_, its a very gross way of customizing a component, but it did save me from copying much of `Button.tsx` over into my codebase.

Removing transitions forces me into the other approach:
https://github.com/Shopify/online-store-ui/pull/2199

At this point the PR isn't actually providing me any value. Debatable if we want to continue with pushing this forward?

alex-page: Why do the online-store buttons require transitions? Right now I am saying that by default all buttons should not have them?

beefchimi:
@BPScott 

> Something we talked about a while back was creating this layout:

Oh I had thought that we didn't want to support that layout anymore? If we do want to support it, I suppose we could make this what happens when you use: `textAlign="center" + fullWidth + disclosure`.

> Things are stretched and the icon and disclosure are at the edges of their boxes out when textAlign is left, but not when right or centered:

This was also discussed during our meeting. I was under the impression we were happy with this behaviour, but I can change it. What you want to see is: `disclosure + fullWidth` ALWAYS results in a right-aligned `disclosure`?

I don't have super strong opinions here because I don't actually think all of these prop combinations are going to be recommended patterns. This kind of leads me back to my question about combo-prop validation and how we define logic around what props can be combo'd together. At the end of the day I think we should just let the consumer do what they want and not be so defensive, but the trade off there is "do we then have to spend time making every combination look good?".

> When specifying textAlign="left" the padding of the disclosure icon changes.

Ah yes good catch, that would be a [result of this commit](https://github.com/Shopify/polaris-react/pull/3709/commits/0f1f69be789ba696859975c04945c34ba16f8312) where we decided to no longer scope the `disclosure` alignment to `fullWidth`. I think this is now a valid case to go back to the added specificity.

> Perform mocking and reseting of mocks in a beforeEach / afterEach method

heheh... I was just copying this style of test authoring: https://github.com/Shopify/polaris-react/blob/master/src/components/Button/tests/Button.test.tsx#L175

I will fix them both! Since there are two of them, I did the mocking at the beginning of the test suite rather than in each `describe`. Did you prefer I do this twice in each `describe`?

As always, thanks for the info :) this will no doubt save me from trouble in the future.

### Closing thoughts

So I spent some time trying to accommodate all of your feedback... and it is very hard! No matter what I always end up with _some prop combination_ revealing an undesirable result. I think in order to do this correctly, I need to introduce a `.hasDisclosure` variant class. I feel like this is going down a dubious path. `Button` has ended up trying to do too many things and resulted in too many conditions to account for.

In order to really go through each scenario and craft the best possible combination, it will take me more time than I have available right now. So before attempting to solve everything, the questions I'd ask are:
1. Is the new result worse/same/better than before?
2. Is this a prop combination we would ever want to see in the wild?

src/styles/shared/_buttons.scss
beefchimi: > all buttons should not have them

I wouldn't have expected that opinion to be extended to consumers. I'm left wondering how we determine what styles are open to theme customization while others - such as transitions - are not? How can contributors better understand what fits into theme customization?

I think its fair to say that we have unique concerns for our app. While we conform to the system for much of what we do, we have our own set of problems to solve and therefor our own UI library to achieve that experience.

We utilize motion in much of our UI to enhance the experience. When you remove motion from some of our actions, it begins to feel out of place in respect to the rest of the UI.

I've misinterpreted systems-of-systems, as I wouldn't have expected us to be unable to utilize transitions for our actions. This feels like the wrong decision to me. What led me to this PR was an effort to remove a completely custom UI component in favour of consuming `Polaris.Button` and theming it. I think consumer's will be incentivized to keep building custom components so they can have better control over the desired styles.

When consumers write their own components, where do we look to inform ourselves about how to author certain UI details, such as motion? The currently styleguide doesn't really tell us very much about authoring our own motion styling, and it doesn't seem to include the context you provided from Johan:
https://polaris.shopify.com/design/polaris-is-changing#section-motion

alex-page: I think this is getting a bit into the weeds. We don't want consumers modifying the button transitions yet. We need to do exploration into how buttons should feel and animate.

UNRELEASED.md
alex-page: I like this format. I think we should be thinking about these sorts of things in a seperate PR or issue though.

BPScott:
> Oh I had thought that we didn't want to support that layout anymore? If we do want to support it, I suppose we could make this what happens when you use: textAlign="center" + fullWidth + disclosure.

...

> This was also discussed during our meeting. I was under the impression we were happy with this behaviour, but I can change it. What you want to see is: disclosure + fullWidth ALWAYS results in a right-aligned disclosure?

My bad for forgetting outcomes of that meeting :)


> In order to really go through each scenario and craft the best possible combination, it will take me more time than I have available right now. So before attempting to solve everything, the questions I'd ask are:
> 
>Is the new result worse/same/better than before?
> Is this a prop combination we would ever want to see in the wild?


You're right, that this is indeed better than before I do want to be mindful of your time as getting lots of these combos to play nice is a time-consuming game of whack-a-mole. Let's call this good to merge now and we can revisit the center aligning at some point in the future if it becomes an issue again.

BPScott (approved):
Amazing work, thank you for picking through component full of edge cases and working out how to make improvements!

:shipit:

alex-page:
Thank you @beefchimi 🚀

PR #3721, opened 2020-12-16
Format a11y error messages and remove skip

alex-page:

Fixes #0000 

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -8,6 +8,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - Update `OptionList` selected styles ([#3633](https://github.com/Shopify/polaris-react/pull/3633))
 - Added the ability to hide the clear filter button on the filter component ([#3049](https://github.com/Shopify/polaris-react/pull/3049))
+- **`OptionList`:** Export `OptionListOptionDescriptor` and `OptionListSectionDescriptor` ([#3741](https://github.com/Shopify/polaris-react/pull/3741))
 - **`Popover`:** New `autofocusTarget` prop to enhance autofocus options ([#3600](https://github.com/Shopify/polaris-react/pull/3600))
 - Added ability to hide query text field in `Filters` component using `hideQueryField` prop ([#3674](https://github.com/Shopify/polaris-react/pull/3674))
 
--- a/src/components/OptionList/OptionList.tsx
+++ b/src/components/OptionList/OptionList.tsx
@@ -27,7 +27,7 @@ export interface OptionDescriptor {
   media?: React.ReactElement<IconProps | ThumbnailProps | AvatarProps>;
 }
 
-interface SectionDescriptor {
+export interface SectionDescriptor {
   /** Collection of options within the section */
   options: OptionDescriptor[];
   /** Section title */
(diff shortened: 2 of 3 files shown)


scripts/accessibility-check.js
```diff
@@ -1,221 +1,91 @@
...
+    .map(fail => ` - ${chalk.bgRed.bold(fail.impact)} ${fail.description}\n   ${url}`);
+}
 
-      if (!isSkippedStory) {
-        const idParam = `id=${encodeURIComponent(story.id)}`;
-        memo.push(
-          idParam,
-          `${idParam}&contexts=Global%20Theming=Enabled%20-%20Light%20Mode`,
-          // Dark mode has lots of errors. It is still very WIP so ignore for now
-          // `${idParam}&contexts=Global%20Theming=Enabled%20-%20Dark%20Mode`,
```
alex-page, line 36: These links were not correctly resolving the `New Design Language`.

scripts/accessibility-check.js
```diff
@@ -1,221 +1,91 @@
...
-          url,
-          errorCount: result.violations.length,
-          error: JSON.stringify(result.violations, null, 2),
-        });
-      } catch (error) {
-        return Promise.resolve({
-          type: 'ERROR',
-          url,
-          error: JSON.stringify(error, null, 2),
-        });
```
alex-page, line 72: We don't need to resolve a promise inside `p-map`. Removing the different types as the user only needs to know if there is an error caught or if the a11y tests failed.

scripts/accessibility-check.js
```diff
@@ -1,221 +1,91 @@
...
-    await initialPage.close();
+  return urls;
+}
 
-    const storyUrls = stories.reduce((memo, story) => {
-      // There is no need to test the Playground, or the "All Examples" stories
-      const isSkippedStory =
-        story.kind === 'Playground/Playground' || story.name === 'All Examples';
+const formatMessage = (id, violations) => {
+  const url = chalk.underline.blue(`http://localhost:6006/iframe.html?id=all-components-${id}`);
```
alex-page, line 41: Added the URL for finding issues quickly, might be a bit too much information.

scripts/accessibility-check.js
```diff
@@ -1,221 +1,94 @@
...
+    else {
+      console.log(chalk.bold('✅ Accessibility is :chefkiss:'));
     }
-
-    process.exit(1);
-  } catch (err) {
-    console.log(err);
+  } catch (error) {
+    console.error(error.message);
     process.exit(1);
```
alex-page, line 92: `process.exit(1);` is causing error responses from CI

BPScott: Presumably if there is an error when executing the script we'd want it the job to fail? Otherwise how would we notice that it isn't working?

alex-page: This didn't do a good job at highlighting the problem. No matter what happens it ends with `process.exit(1)`.

https://github.com/Shopify/polaris-react/blob/a334af33d0c9a16babedb1d8ec9efd9a96690035/scripts/accessibility-check.js#L206-L220

BPScott: If there are no errors then it bails out early here: https://github.com/Shopify/polaris-react/blob/a334af33d0c9a16babedb1d8ec9efd9a96690035/scripts/accessibility-check.js#L172-L178

PR #3727, opened 2020-12-16
Address accessibility issues in AutoComplete and TextField

alex-page:

Fixes accessibility issues related to autocomplete
1. ARIA attribute is not allowed: aria-expanded="false"
2. ARIA attribute is not allowed: aria-multiline="false"

1. Fixed with moving `role="combobox"` from the `div` to the `input`
2. Fixed by removing `multiline` in `TextField` if it is undefined or false (only added when needed )

Followed these two examples:
- [WCAG example of combobox](https://www.w3.org/TR/wai-aria-1.2/#example-6) from @dleroux original PR
- [Gov UK autocomplete](https://alphagov.github.io/accessible-autocomplete/examples/) component

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -30,6 +30,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Added focus ring to disclosure tab when tabbing with keyboard([#3675](https://github.com/Shopify/polaris-react/pull/3675))
 - Fixed windows high contrast mode on hover within disclosure menu and displaying active state upon click for `::before` ([#3675](https://github.com/Shopify/polaris-react/pull/3675))
 - Removed `aria-selected` from `ActionList` as it is not a selectable list ([#3725](https://github.com/Shopify/polaris-react/pull/3725))
+- Moved `aria-role="combobox"` in `Autocomplete` from the `div` to the `input` ([#3727](https://github.com/Shopify/polaris-react/pull/3727))
+- Removed `aria-multiline` in `Input` when false or undefined ([#3727](https://github.com/Shopify/polaris-react/pull/3727))
 
 ### Documentation
 
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -430,8 +430,8 @@ export function TextField({
     'aria-activedescendant': ariaActiveDescendant,
     'aria-autocomplete': ariaAutocomplete,
     'aria-controls': ariaControls,
-    'aria-multiline': normalizeAriaMultiline(multiline),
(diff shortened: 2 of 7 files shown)


UNRELEASED.md
```diff
@@ -30,6 +30,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Added focus ring to disclosure tab when tabbing with keyboard([#3675](https://github.com/Shopify/polaris-react/pull/3675))
 - Fixed windows high contrast mode on hover within disclosure menu and displaying active state upon click for `::before` ([#3675](https://github.com/Shopify/polaris-react/pull/3675))
 - Removed `aria-selected` from `ActionList` as it is not a selectable list ([#3725](https://github.com/Shopify/polaris-react/pull/3725))
+- Moved `Autocomplete` `aria-role="combobox"` to the `input` and removed `aria-multiline` when it if false ([#3727](https://github.com/Shopify/polaris-react/pull/3727))
```
alex-page, line 33:
```suggestion
- Moved `aria-role="combobox"` in `Autocomplete` from the `div` to the `input` ([#3727](https://github.com/Shopify/polaris-react/pull/3727))
- Removed `aria-multiline` in `Input` when false or undefined ([#3727](https://github.com/Shopify/polaris-react/pull/3727))
```

src/components/TextField/TextField.tsx
```diff
@@ -430,8 +430,10 @@ export function TextField({
     'aria-activedescendant': ariaActiveDescendant,
     'aria-autocomplete': ariaAutocomplete,
     'aria-controls': ariaControls,
-    'aria-multiline': normalizeAriaMultiline(multiline),
     'aria-expanded': ariaExpanded,
+    ...(multiline && {
+      'aria-multiline': normalizeAriaMultiline(multiline),
+    }),
```
kyledurand, line 436: I did something similar in my a11y checks PR but got a comment from @BPScott and Andrew asking if we could just change the return type in the normalize function instead of the spread / check combo. 

```ts
function normalizeAriaMultiline(multiline?: boolean | number) {
  if (!multiline) return undefined;

  return Boolean(multiline) || multiline > 0
    ? {'aria-multiline': true}
    : undefined;
}
```

Then you could replace these lines with 

```suggestion
    ...normalizeAriaMultiline(multiline),
```

kyledurand: There are also two tests that would have to be fixed because they're expecting to receive `aria-multiline="false"` which is incorrect

alex-page: Ahhh that is awesome, I will do that. Thanks @kyledurand ✨

PR #3728, opened 2020-12-16
Investigate a11y fails

BPScott:

Investigating then fixing why the a11y test is failing on master

Ignoring warnings that people introduced.
Fix a case where transient errors would not be output, leading to a poor debugging experience

### To test

a11y check CI job should pass


alex-page (approved):
❤️

PR #3730, opened 2020-12-17
RangeSlider - Use a non semantic element for role="slider"

alex-page:

Fixes accessibility issue `ARIA role slider is not allowed for given element`

Replacing `button` with `div`. When using `role="slider"` it should not be applied to a `button` element. [Following the examples on MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_slider_role#Example_2_Text_Values) we replace `button` with `div` update the types and add a `tab-index` for focus management. 

### 🎩 checklist


kyledurand (approved):
Nice! This works well 👍

PR #3737, opened 2020-12-17
Make `utilities/focus` helpers available to consumers

### LINKED ISSUE #3738: Make helpers from `src/utilities/focus.ts` available to consumers
## Feature request summary

Make helpers from `src/utilities/focus.ts` available to consumers.

## Rationale

There are a lot of helpful helpers in `src/utilities/focus.ts`. They could definitely benefit other Shopify projects. :)

---

ben10code:

Fixes #3738 

This PR makes the helpers in `utilities/focus.ts` publicly available to consumers. 

### How to 🎩

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -10,6 +10,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Added the ability to hide the clear filter button on the filter component ([#3049](https://github.com/Shopify/polaris-react/pull/3049))
 - **`Popover`:** New `autofocusTarget` prop to enhance autofocus options ([#3600](https://github.com/Shopify/polaris-react/pull/3600))
 - Added ability to hide query text field in `Filters` component using `hideQueryField` prop ([#3674](https://github.com/Shopify/polaris-react/pull/3674))
+- Export helpers from `utilities/focus.ts` to be publicly available to consumers ([#3737](https://github.com/Shopify/polaris-react/pull/3737))
 
 ### Bug fixes
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -33,3 +33,15 @@ export {
   toCssCustomPropertySyntax as UNSTABLE_toCssCustomPropertySyntax,
   Tokens as UNSTABLE_Tokens,
 } from './utilities/theme';
+
+export {
+  findFirstFocusableNode,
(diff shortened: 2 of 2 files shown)


BPScott:
I'm generally a bit wary of exposing our utilities. it means that if we ever change our internals we're stuck with having to support a bunch of functions that we no longer use.

Many of these were copied from https://github.com/shopify/javascript-utilities. If you find them useful then you should either use that package or add these bits of functionality into a project in quilt or copy what you need into your own app.

tleunen:
I wasn't aware some of these focus utils were available in the [shared JS libs](https://github.com/Shopify/javascript-utilities/blob/master/src/focus.ts). I think it would indeed make more sense for a consumer app to import them from `@shopify/javascript-utilities` than from `polaris`.
@ben10code - If some functions are missing from the js utilities, feel free to add them there instead.

> Many of these were copied from https://github.com/shopify/javascript-utilities. If you find them useful then you should either use that package or add these bits of functionality into a project in quilt or copy what you need into your own app.

Any reason why Polaris wouldn't use the shared shopify utils instead of copying them? Is it to keep Polaris as "self-sufficient" as much as possible?

BPScott:
> Any reason why Polaris wouldn't use the shared shopify utils instead of copying them? Is it to keep Polaris "self-sufficient" as much as possible?

Per the comment at the top of its readme `@shopify/javascript-utilities` is deprecated. It also only exports commonJS content which means we don't treeshake away any unused utility functions. As part of v5 I removed our dependency on `@shopify/javascript-utilities` and chose to inline the functions we needed as it turned out we had a bunch of focus management helpers already and creating an extra package in quilt didn't seem worth the overhead. 

See https://github.com/Shopify/polaris-react/pull/3102 and https://github.com/Shopify/polaris-react/pull/3108 for some of those changes.

BPScott:
A small further clarification on:

> I'm generally a bit wary of exposing our utilities. it means that if we ever change our internals we're stuck with having to support a bunch of functions that we no longer use.

I see `@shopify/polaris` as a component library first and foremost. We're shouldn't be responsible for maintaining low-level helpers - that's what the packages in Quilt are for.

tleunen:
Makes sense! Thanks @BPScott

ben10code:
Thanks for the insight Ben! Closing this PR and issue :)

PR #3740, opened 2020-12-18
Remove /.github from .prettierignore

alex-page:
Co-Authored-By: Ben Scott <227292+BPScott@users.noreply.github.com>

Not necessary to disable Prettier in this directory.

Removing the disable of prettier in `/.github/`


BPScott (approved):
:shipit:

PR #3741, opened 2020-12-18
🏷️ [OptionList] Export types

beefchimi:
We have a need for the `OptionDescriptor` and `SectionDescriptor` in `online-store-web`, so I am exporting both of these types for consumption.

Since both of these type names are somewhat generic... I've renamed their exports. Please let me know if this is following convention!

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -8,6 +8,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - Update `OptionList` selected styles ([#3633](https://github.com/Shopify/polaris-react/pull/3633))
 - Added the ability to hide the clear filter button on the filter component ([#3049](https://github.com/Shopify/polaris-react/pull/3049))
+- **`OptionList`:** Export `OptionListOptionDescriptor` and `OptionListSectionDescriptor` ([#3741](https://github.com/Shopify/polaris-react/pull/3741))
 - **`Popover`:** New `autofocusTarget` prop to enhance autofocus options ([#3600](https://github.com/Shopify/polaris-react/pull/3600))
 - Added ability to hide query text field in `Filters` component using `hideQueryField` prop ([#3674](https://github.com/Shopify/polaris-react/pull/3674))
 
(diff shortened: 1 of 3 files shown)


kyledurand (approved):

BPScott:
Heya @beefchimi, Could you tell me a bit more about your use-case?

I tend to be pretty adverse to exporting arbitrary types because it can hamper our ability to do internal refactors. Usually you don't really care about the type in isolation, but you care because you're about to pass something into the a set of props - in this case OptionList's `options or `sections`.

Because what people usually really care about is what they pass into they component (in this case `OptionList`) I tend to prefer it if the type you want can be be extracted from the component's props type (in this case `OptionListProps`) - which is already exported.

In this case things get a little fiddly as the `options` Prop is typed as `options?: OptionDescriptor[];` but we can knock off the nullability and the arrayness (ugh made up word) :

```ts
import {OptionsListProps} from "@shopify/polaris";

// NonNullable is a built-in type, demoing how we can knock off the `| undefined` from `OptionListProps['options']`.
// see https://www.typescriptlang.org/docs/handbook/utility-types.html
// Equivalent to `OptionDescriptor[]`
type Options = NonNullable<OptionListProps['options']>;

// Helper for transforming "SomeType[]" into just "SomeType"
// Also available from @shopify/useful-types if you don't want to define it yourself
type ArrayElement<T> = T extends (infer U)[] ? U : never;

//  Equivalent to `OptionDescriptor`. You can see this by pasting this into VSCode and hovering over "Option"
type Option = ArrayElement<NonNullable<OptionListProps['options']>>;
```

Thus doing `type Option = ArrayElement<NonNullable<OptionListProps['options']>>;` would allow you to get at this type without us needing to export it. Would that suit your use-case?

beefchimi:
@BPScott this is the PR where I wanted access to the "non-array" `options`:
https://github.com/Shopify/online-store-web/pull/8965/files#diff-68a74c3d067351e56684fcba3b47d71c2abce7b847a30f4a69cae25a3fc23dbeR3

I did not know about this `ArrayElement` technique! This should definitely solve my problem, and overall help me simplify a few things across a couple of code bases. Very nice piece of TypeScript 😃  Thank you for sharing!

BPScott:
Nice! It looks like in most cases there you don't even need the `ArrayElement` cleverness. Mostly you're referencing OptionListOptionDescriptor[] so you'd be fine with `OptionListProps['options']` and an occasional `NonNullable<OptionListProps['options']>`

PR #3752, opened 2020-12-22
Get base_ref for Percy in pull request

ashwinnarayanan2001:

For Percy to accurately check the changes in version 6.0.0 we have to set the base branch to the `github.base_ref` variable. To do this we needed to run Percy on the `pull_request` event to access this value. In doing this it made sense to combine the two CI files and access the `pull_request` events.

- Combine the CI and add a rule for `pull_request` and `push` events
- Add the environment variable for Percy so that it runs on the base branch


kyledurand (approved):
I don't have a lot of context on this kind of change but the parts I do understand look good 👍

BPScott:
Contrasting the checks that appear in https://github.com/Shopify/polaris-react/pull/3739 vs this PR:

- The "Shrink-Ray: Net Diff" report is no longer generated correctly. It is stuck on "pending". Probably because using pull_request causes the action to checkout and use a merge commit instead of the actual one.
- The "Storybook Publish" check is no longer present. For the same reason. The chromatic docs suggest that it should be ran on a push event not a pull_request per https://www.chromatic.com/docs/cli#usage-in-a-github-action

PR #3769, opened 2021-01-04
[Icon] Remove new DL fallbacks and references

ashwinnarayanan2001:
These changes are introduced as part of polaris-v6, where all reference to the newDesignLanguage are being removed. Fixes Shopify/polaris-ux#567.

This pull request removes all references to the newDesignLanguage in the Icon component on polaris-react.
## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist


ashwinnarayanan2001:
One question I had was if removing the 'warning when an untrusted SVG is used with a color option from the new design language' was the right decision (Icon.tsx)

alex-page (approved):
🌈

src/components/Icon/Icon.tsx
```diff
@@ -46,31 +43,11 @@ export function Icon({source, color, backdrop, accessibilityLabel}: Props) {
...
-  ) {
-    // eslint-disable-next-line no-console
-    console.warn(
-      'Recoloring external SVGs is not supported with colors in the new design language. Set the intended color on your SVG instead.',
-    );
-  }
-
   const className = classNames(
     styles.Icon,
     color && styles[variationName('color', color)],
```
BPScott, line 48: Do we want people to still be able to specify color props like `white` and `orange` in the new DL or do we want to force people to use roles like `critical` and `warning`?

I'm thinking the latter.

If people agree we should remove  the strings in `Color` from IconProps over here https://github.com/Shopify/polaris-react/blob/master/src/types.ts#L57, and then you can remove all the css classes for `.colorWhite` `.colorOrange` etc. Also check if Color is used anywhere else and if not so remove that type.

BPScott (commented):
Got a question about legacy colors that might give you a bit more work to do, see inline:

src/components/Icon/Icon.tsx
```diff
@@ -1,8 +1,6 @@
 import React from 'react';
 
 import {classNames, variationName} from '../../utilities/css';
-import {isNewDesignLanguageColor} from '../../utilities/color-new-design-language';
```
BPScott, line 4: Is this function used anywhere else? If not we should delete its definition too.

kyledurand: I think we may still need this and I'm thinking it should be renamed to something like `isValidIconColor` Then we can still use it in the component below. That way if people pass in something like `orange` we can issue a console warning

BPScott: When have props that are strings but only some values are valid we usually just use typescript to assert people are using a valid string. E.g. [Button Size](https://github.com/Shopify/polaris-react/blob/3d2da225b7c484e446dd1522cab0db0dfbfdbb87/src/components/Button/Button.tsx#L31) can be 'small', 'medium' or 'large' and we type it as such.

As far as I remember we've never double checked at runtime too.

I'm proposing that the [`color` prop](https://github.com/Shopify/polaris-react/blob/master/src/types.ts#L57) should have the type `NewDesignLanguageColor` instead of `Color | NewDesignLanguageColor` and that way if people pass in 'orange' then typescript will shout at them, and thus we don't need a runtime check.

alex-page:
@ashwinnarayanan2001 going to close this one as it had a lot of overlapping problems with Spinner. Thank you though @emmaorhun and I used this PR to fix up spinner and icon https://github.com/Shopify/polaris-react/pull/3778 and it helped a lot 💯

PR #3770, opened 2021-01-05
[Loading] Functional component, add aria-label

alex-page:

In https://github.com/Shopify/polaris-react/pull/3749 new a11y issues have emerged due to the loading bar not having a label. This PR adds a label and converts the component to a functional component.

> Ensures every ARIA progressbar node has an accessible name
> aria-label attribute does not exist or is empty
> https://dequeuniversity.com/rules/axe/4.1/aria-progressbar-name?application=axeAPI

Converting the loading bar to a functional component, add a label that is translated and removes technical debt around rAF and debounce.

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist


src/components/Frame/components/Loading/Loading.tsx
```diff
@@ -1,79 +1,41 @@
...
-    const {progress, step} = this.state;
+  const [progress, setProgress] = useState(0);
+  const [step, setStep] = useState(INITIAL_STEP);
 
+  useEffect(() => {
     if (progress >= STUCK_THRESHOLD) {
       return;
     }
 
-    const animation = requestAnimationFrame(() => this.increment());
```
alex-page, line 71: I don't know why we are using `requestAnimationFrame` for incrementing the [value on a transform](https://github.com/Shopify/polaris-react/pull/3770/files#diff-21fb7fa504ae13f32acd177042514b61f7cde459c035fe20b4879d54c1fd1cceL45). From my understanding the transform should be doing transitions on the GPU so we wouldn't be getting a performance improvement by waiting for the next frame.

src/components/Frame/components/Loading/Loading.tsx
```diff
@@ -1,79 +1,41 @@
...
-  state: State = {
-    progress: 0,
-    step: INITIAL_STEP,
-    animation: null,
-  };
-
-  private ariaValuenow = debounce(() => {
-    const {progress} = this.state;
-    return Math.floor(progress / 10) * 10;
-  }, 15);
```
alex-page, line 27: Removed all this sillyness for a [CSS transition delay](https://github.com/Shopify/polaris-react/pull/3770/files#diff-5cd4dd79a5c5206628e3b059fadc942878686fcb3830fd86d2de08d8de774661R17).

aveline (approved):

BPScott (changes requested):
Running this in storybook I get about 50 console warnings per second of "Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render." . Adding a `console.log('f')` into the useEffect I see that logged 50 times per each of the above warnings. Looking in task manager this pegs a one of my CPU cores at 100% usage when that tab is active.

So you've managed to make a hot infinite loop. Whoops.

I assume the requestAnimationFrame stuff was there to limit how often the update gets ran and to stop stress-testing CPUs (so the update runs at a max 60 times per second rather than AS MANY AS THE CPU CAN MANAGE AAAAAH).


---


Given that this loading bar is a massive lie anyway why do we even need javascript at all? Could we do all this with a CSS animation that runs once, that uses a cubic-bezier curve to give the "fast at the start but slows over time" effect (the suttering effect is probably also possible but might require an animation with a few more keyframes to give the stop then start effect): The following CSS seems to get us most of the way there:

```css
.Level {
  width: 100%;
  height: 100%;
  transform: scaleX(1);
  transform-origin: 0;
  background-color: var(--p-icon-success, color('teal'));

  animation-name: loading-animmate;
  animation-duration: 30s;
  animation-iteration-count: 1;
  animation-timing-function: cubic-bezier(0.27, 0.93, 0.46, 0.73);

  @media screen and (-ms-high-contrast: active) {
    background-color: ms-high-contrast-color('selected-text-background');
  }
}

@keyframes loading-animmate {
  from {
    transform: scaleX(0);
  }
  to {
    transform: scaleX(0.98);
  }
}

```

src/components/Frame/components/Loading/Loading.tsx
```diff
@@ -1,79 +1,50 @@
...
-
-    this.setState({
-      progress: Math.min(progress + step, 100),
-      step: INITIAL_STEP ** -(progress / 25),
-      animation,
+    requestAnimationFrame(() => {
+      const newProgress = Math.min(progress + step, 100);
+      setAnimating(true);
+      setProgress(newProgress);
+      setStep(10);
```
alex-page, line 26: This is not the nice UX we previously had but the infinite loop is fixed.

src/components/Frame/components/Loading/Loading.tsx
```diff
@@ -1,79 +1,50 @@
...
       return;
     }
 
-    const animation = requestAnimationFrame(() => this.increment());
-
-    this.setState({
-      progress: Math.min(progress + step, 100),
-      step: INITIAL_STEP ** -(progress / 25),
-      animation,
+    requestAnimationFrame(() => {
```
alex-page, line 22: This `rAF` is needed to trigger the transition otherwise it sets the value to `10` and does not animate between 0 and 10.

src/components/Frame/components/Loading/Loading.tsx
kyledurand: I forget what the previous behaviour is but getting stuck at 10 isn't great. Can we store this value in a ref and increment by 10 if stuck to give the illusion that the page is still making progress loading? 

Maybe we could also do something to reduce the increment past 80% by half each time so it never reaches 100?

alex-page: I added some magic, take a look and let me know what you think!

src/components/Frame/components/Loading/Loading.scss
```diff
@@ -14,7 +14,7 @@ $height: rem(3px);
   height: 100%;
   transform-origin: 0;
   background-color: var(--p-icon-success, color('teal'));
-  transition: transform duration(slowest) easing(out);
+  transition: transform duration(slowest) easing(out) 15ms;
```
beefchimi, line 17: I actually think the default `linear` feels more natural here.

This `delay` seems a bit magical. Is it an arbitrary number because it _just happens to work nicely_, or is it informed by _something_?

beefchimi (approved):
Extremely minor feedback. I do echo what @BPScott said - I suspect there is probably a more elegant solution that leans more on CSS `keyframes`. But that can always be a future enhancement.

For anyone curious, here is the entire life of the "loading":

<details>
<summary><strong></strong></summary>

```
begin useEffect (progress, animating) 0 false
step 9.9
progress + step 9.9
begin useEffect (progress, animating) 0 true
RETURN!
begin useEffect (progress, animating) 9.9 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 9.9 false
step 8.91
progress + step 18.810000000000002
begin useEffect (progress, animating) 9.9 true
RETURN!
begin useEffect (progress, animating) 18.810000000000002 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 18.810000000000002 false
step 8.019
progress + step 26.829
begin useEffect (progress, animating) 18.810000000000002 true
RETURN!
begin useEffect (progress, animating) 26.829 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 26.829 false
step 7.217099999999999
progress + step 34.0461
begin useEffect (progress, animating) 26.829 true
RETURN!
begin useEffect (progress, animating) 34.0461 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 34.0461 false
step 6.49539
progress + step 40.54149
begin useEffect (progress, animating) 34.0461 true
RETURN!
begin useEffect (progress, animating) 40.54149 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 40.54149 false
step 5.845851
progress + step 46.387341000000006
begin useEffect (progress, animating) 40.54149 true
RETURN!
begin useEffect (progress, animating) 46.387341000000006 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 46.387341000000006 false
step 5.2612659
progress + step 51.648606900000004
begin useEffect (progress, animating) 46.387341000000006 true
RETURN!
begin useEffect (progress, animating) 51.648606900000004 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 51.648606900000004 false
step 4.735139309999999
progress + step 56.383746210000005
begin useEffect (progress, animating) 51.648606900000004 true
RETURN!
begin useEffect (progress, animating) 56.383746210000005 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 56.383746210000005 false
step 4.261625379
progress + step 60.64537158900001
begin useEffect (progress, animating) 56.383746210000005 true
RETURN!
begin useEffect (progress, animating) 60.64537158900001 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 60.64537158900001 false
step 3.835462841099999
progress + step 64.4808344301
begin useEffect (progress, animating) 60.64537158900001 true
RETURN!
begin useEffect (progress, animating) 64.4808344301 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 64.4808344301 false
step 3.4519165569899997
progress + step 67.93275098709
begin useEffect (progress, animating) 64.4808344301 true
RETURN!
begin useEffect (progress, animating) 67.93275098709 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 67.93275098709 false
step 3.106724901291
progress + step 71.039475888381
begin useEffect (progress, animating) 67.93275098709 true
RETURN!
begin useEffect (progress, animating) 71.039475888381 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 71.039475888381 false
step 2.7960524111618996
progress + step 73.8355282995429
begin useEffect (progress, animating) 71.039475888381 true
RETURN!
begin useEffect (progress, animating) 73.8355282995429 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 73.8355282995429 false
step 2.51644717004571
progress + step 76.35197546958861
begin useEffect (progress, animating) 73.8355282995429 true
RETURN!
begin useEffect (progress, animating) 76.35197546958861 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 76.35197546958861 false
step 2.2648024530411393
progress + step 78.61677792262975
begin useEffect (progress, animating) 76.35197546958861 true
RETURN!
begin useEffect (progress, animating) 78.61677792262975 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 78.61677792262975 false
step 2.038322207737025
progress + step 80.65510013036678
begin useEffect (progress, animating) 78.61677792262975 true
RETURN!
begin useEffect (progress, animating) 80.65510013036678 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 80.65510013036678 false
step 1.8344899869633224
progress + step 82.4895901173301
begin useEffect (progress, animating) 80.65510013036678 true
RETURN!
begin useEffect (progress, animating) 82.4895901173301 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 82.4895901173301 false
step 1.6510409882669905
progress + step 84.14063110559708
begin useEffect (progress, animating) 82.4895901173301 true
RETURN!
begin useEffect (progress, animating) 84.14063110559708 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 84.14063110559708 false
step 1.485936889440292
progress + step 85.62656799503738
begin useEffect (progress, animating) 84.14063110559708 true
RETURN!
begin useEffect (progress, animating) 85.62656799503738 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 85.62656799503738 false
step 1.3373432004962624
progress + step 86.96391119553364
begin useEffect (progress, animating) 85.62656799503738 true
RETURN!
begin useEffect (progress, animating) 86.96391119553364 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 86.96391119553364 false
step 1.203608880446636
progress + step 88.16752007598028
begin useEffect (progress, animating) 86.96391119553364 true
RETURN!
begin useEffect (progress, animating) 88.16752007598028 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 88.16752007598028 false
step 1.0832479924019722
progress + step 89.25076806838226
begin useEffect (progress, animating) 88.16752007598028 true
RETURN!
begin useEffect (progress, animating) 89.25076806838226 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 89.25076806838226 false
step 1
progress + step 90.25076806838226
begin useEffect (progress, animating) 89.25076806838226 true
RETURN!
begin useEffect (progress, animating) 90.25076806838226 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 90.25076806838226 false
step 1
progress + step 91.25076806838226
begin useEffect (progress, animating) 90.25076806838226 true
RETURN!
begin useEffect (progress, animating) 91.25076806838226 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 91.25076806838226 false
step 1
progress + step 92.25076806838226
begin useEffect (progress, animating) 91.25076806838226 true
RETURN!
begin useEffect (progress, animating) 92.25076806838226 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 92.25076806838226 false
step 1
progress + step 93.25076806838226
begin useEffect (progress, animating) 92.25076806838226 true
RETURN!
begin useEffect (progress, animating) 93.25076806838226 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 93.25076806838226 false
step 1
progress + step 94.25076806838226
begin useEffect (progress, animating) 93.25076806838226 true
RETURN!
begin useEffect (progress, animating) 94.25076806838226 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 94.25076806838226 false
step 1
progress + step 95.25076806838226
begin useEffect (progress, animating) 94.25076806838226 true
RETURN!
begin useEffect (progress, animating) 95.25076806838226 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 95.25076806838226 false
step 1
progress + step 96.25076806838226
begin useEffect (progress, animating) 95.25076806838226 true
RETURN!
begin useEffect (progress, animating) 96.25076806838226 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 96.25076806838226 false
step 1
progress + step 97.25076806838226
begin useEffect (progress, animating) 96.25076806838226 true
RETURN!
begin useEffect (progress, animating) 97.25076806838226 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 97.25076806838226 false
step 1
progress + step 98.25076806838226
begin useEffect (progress, animating) 97.25076806838226 true
RETURN!
begin useEffect (progress, animating) 98.25076806838226 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 98.25076806838226 false
step 1
progress + step 99.25076806838226
begin useEffect (progress, animating) 98.25076806838226 true
RETURN!
begin useEffect (progress, animating) 99.25076806838226 true
RETURN!
onTransitionEnd
begin useEffect (progress, animating) 99.25076806838226 false
```

</details>

I do find it curious that we re-log the same `progress` value more often than I would expect. I couldn't quite wrap my head around why that is happening.

src/components/Frame/components/Loading/Loading.scss
alex-page: The delay came from the previous `setTimeout` so it is definitely magicial.

alex-page:
> a more elegant solution that leans more on CSS keyframes

I agree, the current downside of that is it doesn't tell the user any information about `the page is loading` or `the page is 50%` complete. So controlling the `progress` in JS is useful for pushing that information to screenreaders.

src/components/Frame/components/Loading/Loading.scss
beefchimi: Just in case you maybe accidentally changed the wrong thing:

My comment was to change the `timing-function` to be `linear`. Your latest commit made a change to the `duration` - which is maybe fine, I didn't test the speed difference.

src/components/Frame/components/Loading/Loading.tsx
```diff
@@ -1,79 +1,47 @@
...
+  }, [progress, animating]);
+
+  const customStyles = {
+    transform: `scaleX(${Math.floor(progress) / 100})`,
+  };
+
+  return (
+    <div
+      className={styles.Loading}
+      aria-valuenow={progress}
```
BPScott, line 34: Taking a read of https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-valuenow_attribute

> When the rendered value cannot be accurately represented as a number, authors SHOULD use the aria-valuetext attribute in conjunction with aria-valuenow to provide a user-friendly representation of the range's current value. For example, a slider may have rendered values of small, medium, and large. In this case, the values of aria-valuenow could range from 1 through 3, which indicate the position of each value in the value space, but the aria-valuetext would be one of the strings: small, medium, or large.


Seeing as we're just increasing the percentage over time rather than being in relation to anything meaningful, I'd argue that loading does not have accurate numerical representation.

If that's the case, what if we stop using`aria-valuenow` (and aria-valuemin / aria-valuemax`) and instead set `aria-valuetext="Loading in progress"` or some other not-pretending-there-is-an-exact-value text. Then we'd be able to move to the CSS animation approach for the visual effect.

src/components/Frame/components/Loading/Loading.tsx
```diff
@@ -1,79 +1,47 @@
...
+export function Loading() {
+  const i18n = useI18n();
 
-  private increment() {
-    const {progress, step} = this.state;
+  const [progress, setProgress] = useState(0);
+  const [animating, setAnimating] = useState(false);
 
-    if (progress >= STUCK_THRESHOLD) {
+  useEffect(() => {
```
BPScott, line 15: The old code handled cancelling the animation frame on unmount. Do we need to do the same thing in the new world?

src/components/Frame/components/Loading/Loading.tsx
alex-page: Yeah I agree, but lets make a new issue or PR to do that. That is outside of the scope of this fix.

BPScott (approved):
Looking pretty tasty. No longer pegging a CPU core which is always a good sign ;)

One question about cancelling animation frames if the component gets unmounted in the moment between a frame is requested but not yet triggered. Not sure how to test that though :/

If we want to have a new PR that replaces this visual with a CSS animation, then this is good to go.

src/components/Frame/components/Loading/Loading.tsx
alex-page: I don't think so because we stop the `useEffect` and don't trigger an animation unless there is still progress to be made.

PR #3773, opened 2021-01-05
Use vendor-provided actions for Chromatic and Percy

BPScott:

`pull_request` builds checkout use a merge commit (of your branch into the base branch's tip) instead of just your branch's tip.

This means that Chromatic / Percy need too be taught about that foible and have them set custom commit information. Rather than doing that ourselves lets use the vendor-provided actions that do that for us.

See https://github.com/percy/storybook-action and https://github.com/chromaui/action

Use Chromatic and Percy actions for their respective steps

### TO TEST

Note that the "Storybook Publish" check that links to the published storybook is now being correctly set.
Note that Percy's PR / base branch info is now being set: https://percy.io/Shopify/polaris-react/builds-next/8256288/unchanged/469473165?browser=chrome&viewLayout=side-by-side&viewMode=new&width=1280 -> Click the "i" icon in the left panel next to the build number

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,7 +126,6 @@
     "babel-preset-shopify": "^21.0.0",
     "chalk": "^2.4.2",
     "change-case": "^3.1.0",
-    "chromatic": "^5.0.0",
     "codecov": "^3.8.1",
     "commander": "^4.1.0",
     "copyfiles": "^2.1.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -126,7 +126,6 @@
     "babel-preset-shopify": "^21.0.0",
     "chalk": "^2.4.2",
     "change-case": "^3.1.0",
-    "chromatic": "^5.0.0",
```
BPScott, line 129: No need for chromatic anymore as CI uses its own latest version.

alex-page (approved):

PR #3775, opened 2021-01-05
Improve accessibility HCM mixin to work in other browsers

### LINKED ISSUE #3774: Update ms-high-contrast-outline to work on all browsers
## Issue summary

There is currently a mixin in `_accessibility.scss` used to provide high contrast outlines in Windows environment. It currently uses a media query `@media(-ms-high-contrast: active)`, but this latter only works in Edge and not recommended for production use. Moreover, the colours seem deprecated as well.

More information can be found here:  https://developer.mozilla.org/en-US/docs/Web/CSS/@media/-ms-high-contrast

My suggestion would be to remove the media query and to update the color to `transparent`. 

## Expected behavior

The mixin should work in all browsers that support Microsoft's HCM (Firefox, Edge, IE).

## Actual behavior

The mixin is using a media query that only works in Edge and not recommended for production use. Because of that, extra redundant code is needed to provide an HCM outline in browsers other than Edge.

ben10code:

Fixes #3774

The mixin `ms-high-contrast-mode` in `_accessibility.scss` uses a media query that is non-standard and only supported by Microsoft Edge (version < 18). Alternatively, `forced-colors` media query could be used, but it is not optimal since it requires a beta flag to be manually enabled in Firefox (version > 81). The current color used (windowText) seems to also be deprecated.

The use of just a transparent outline seems to be commonly recommended solution to support a wider range of browsers, without disrupting non-high-contrast environments. 

More information can be found here: 
https://drafts.csswg.org/css-color-4/#deprecated-system-colors (windowText being a deprecated system color)
https://developer.mozilla.org/en-US/docs/Web/CSS/@media/-ms-high-contrast (non-standard media query)

Slack conversations leading up to this PR:
https://shopify.slack.com/archives/C4Y8N30KD/p1607455945447700

### Screenshots

**Non-high-contrast environment (undisturbed)**
image

**Firefox without this PR's changes**
image

**Firefox (same with Edge and IE) with this PR's changes**
image

1. Removes non-standard media query used.
2. Updates color to `transparent` instead of `windowText`, which is deprecated.

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -40,7 +40,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Fixed clear button in `TextField` unintentionally closing `Popover` when clicked ([#3688](https://github.com/Shopify/polaris-react/pull/3688))
 - Added focus styles to the clear button in `TextField` ([#3688](https://github.com/Shopify/polaris-react/pull/3688))
 - Increased contrast of navigation text color ([#3742](https://github.com/Shopify/polaris-react/pull/3742))
-- Removed `-ms-high-contrast` media query from `ms-high-contrast-outline` as it is non-standard and updated the outline color from `windowText` to `tranpsarent` ([#3775](https://github.com/Shopify/polaris-react/pull/3775)).
+- Removed `-ms-high-contrast` media query from `ms-high-contrast-outline` as it is non-standard and updated the outline color from `windowText` to `transparent` ([#3775](https://github.com/Shopify/polaris-react/pull/3775)).
 
 ### Documentation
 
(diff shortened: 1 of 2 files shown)


alex-page:
@ben10code this looks good, before I kick it open on my windows PC to test do you have any screenshots you can provide?

alex-page (approved):

UNRELEASED.md
```diff
@@ -40,6 +40,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Fixed clear button in `TextField` unintentionally closing `Popover` when clicked ([#3688](https://github.com/Shopify/polaris-react/pull/3688))
 - Added focus styles to the clear button in `TextField` ([#3688](https://github.com/Shopify/polaris-react/pull/3688))
 - Increased contrast of navigation text color ([#3742](https://github.com/Shopify/polaris-react/pull/3742))
+- Removed `-ms-high-contrast` media query from `ms-high-contrast-outline` as it is non-standard and updated the outline color from `windowText` to `tranpsarent` ([#3775](https://github.com/Shopify/polaris-react/pull/3775)).
```
kaelig, line 43:
```suggestion
- Removed `-ms-high-contrast` media query from `ms-high-contrast-outline` as it is non-standard and updated the outline color from `windowText` to `transparent` ([#3775](https://github.com/Shopify/polaris-react/pull/3775)).
```

alex-page (approved):

PR #3778, opened 2021-01-05
[WIP] Remove newDesignLanguageColor type and more newDL fallbacks in Spinner and Icon

emmaorhun:

Fixes https://github.com/Shopify/polaris-ux/issues/539
Fixes Shopify/polaris-ux#567

AND in the process of doing so found a bunch of extra bulk to remove, like how the newDesignLanguageColor type was not necessary anymore and removed a lot of the instances where we were manually passing colors around like in the Spinner and Icon component.

*  remove new DL references in the Spinner component
*  remove newDesignLanguageColor type (Spinner color has a constant color set to 'highlight' and we are no longer manually setting or passing colors into Icon)
*  remove newDL tests and anything that was dependant on manual color setting

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist


BPScott:
> we are no longer manually setting or passing colors into Icon)

I don't think that's quite right. I believe we still want to be able to pass color roles into icon.

so `<Icon color="critical" source={Blah} />` and `<Icon color="warning" source={Blah} />` and the other color roles from https://github.com/Shopify/polaris-react/blob/8c31a954ba8a2742d0ea95720f3252015025c394/src/types.ts#L39-L46 are allowed. But we don't want the "old" colors like "red" and "orange" from https://github.com/Shopify/polaris-react/blob/8c31a954ba8a2742d0ea95720f3252015025c394/src/types.ts#L1-L37.

BPScott:
Ref #3769 for an alternative PR for Icon updates. Not sure which of you two are wrangling this

alex-page (approved):

src/types.ts
```diff
@@ -1,42 +1,4 @@
...
-  | 'yellowLighter'
-  | 'yellow'
-  | 'yellowDark'
-  | 'orange'
-  | 'redLighter'
-  | 'red'
-  | 'redDark'
-  | 'purple';
-
-export type NewDesignLanguageColor =
```
BPScott, line 39: For weird legacy never-got-around-to-fixing-it reasons everything in this file is exported publicly. So `Color` remains exported but now has a very different shape. Will that more problems than "Color is no longer exported" from this file?

I'm thinking it might, and it'd be easier to keep this named as `NewDesignLanguageColor`.

src/components/Spinner/Spinner.tsx
```diff
@@ -35,33 +23,14 @@ export interface SpinnerProps {
 
 export function Spinner({
   size = 'large',
-  color = 'teal',
```
BPScott, line 38: Over in https://github.com/Shopify/polaris-react/issues/3487 there was talk of needing colors on a Spinner. Thoughts on keeping this hanging around for the moment?

BPScott (commented):
Gah you merged just before I made comments!

src/types.ts
alex-page: Hey @BPScott I don't want to forget about this. As this is a breaking change I think we should be deprecating `NewDesignLanguageColor`. Should we also make the `Color` export private so it cannot be used externally??

PR #3779, opened 2021-01-06
[Collapsible] Fix animation

kyledurand:

Fixes a bit of animation jank when toggling quickly

## 

### How to 🎩

Chroma should be fine

### 🎩 checklist


alex-page (approved):

src/components/Collapsible/Collapsible.tsx
```diff
@@ -30,165 +24,83 @@ export interface CollapsibleProps {
...
-    if (target === this.node.current) {
-      this.setState({animationState: 'idle', height: null});
+  }, [open, isOpen]);
+
+  // If collapsible defaults to open, set an initial height
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useEffect(() => {
+    if (!open || !collapisbleContainer.current) return;
+    setHeight(collapisbleContainer.current.scrollHeight);
+  });
```
BPScott, line 77: No dependency array here means it runs on every rerender. An empty array means only run on initial mount, is that what you want?

From https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects

BPScott (approved):
Looks good to me! One little thing to check inline ->

One of the more fiddly class components bites the dust!

src/components/Collapsible/Collapsible.tsx
kyledurand: Thanks Ben! That makes sense, I always get these mixed up 🤦‍♂️

PR #3787, opened 2021-01-07
[Label, Labelled] Change `label` prop type to `React.ReactNode`

Tom-Bonnike:

The current type for `label` is `string`, which doesn’t allow us to add anything else than unstyled text to a label. This would have greatly improved https://github.com/Shopify/web/pull/35964, we currently workaround it by passing an empty label and re-implementing fields’ labels, which is not ideal… I added some examples of what a `React.ReactNode` `label` enables in the tophatting section below.

This PR updates the `Label` and `Labelled`’s `label` prop types to `React.ReactNode` and makes the necessary change anywhere that prop is drilled down (which most form fields seem to do). I think I changed every occurence, I checked and I couldn’t find any code manipulating the label and expecting it to be a string, there isn’t any TypeScript error. 😌

The `helpText` prop is already typed with `React.ReactNode`, I don’t see any reason not to use it for the `label` as well, except that block elements cannot be children of `label` tags, which polaris consumers may not realize; I expected React to log an error for invalid DOM nesting but I couldn’t have it show up 🤔 

### How to 🎩

<details>

```jsx
import React from 'react';
import {QuestionMarkMajor} from '@shopify/polaris-icons';

import {Page, TextField, Icon} from '../src';

function Marker() {
  return <span style={{color: 'red'}}>*</span>;
}

function noop() {}

  return (
    <Page title="Playground">
      <TextField onChange={noop} label="Regular label" />
      <TextField
        onChange={noop}
        label={
          <>
            Label as Fragment <Marker />
          </>
        }
      />
      <TextField
        onChange={noop}
        label={['Label as array', ' ', <Marker key="marker" />]}
      />
      <TextField
        onChange={noop}
        label={
          <span style={{display: 'flex'}}>
            Label with icon &nbsp;
            <Icon source={QuestionMarkMajor} />
          </span>
        }
      />
    </Page>
  );
}

```

</details>

Every field should render its label as expected:
Screenshot 2021-01-07 at 11 58 32

### 🎩 &nbsp;checklist

None of these are applicable.

### DIFF
--- a/src/components/DropZone/DropZone.tsx
+++ b/src/components/DropZone/DropZone.tsx
@@ -35,7 +35,7 @@ type Type = 'file' | 'image';
 
 export interface DropZoneProps {
   /** Label for the file input */
-  label?: string;
+  label?: React.ReactNode;
   /** Adds an action to the label */
   labelAction?: LabelledProps['action'];
   /** Visually hide the label */
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -6,6 +6,24 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Updated `OptionList` selected styles ([#3633](https://github.com/Shopify/polaris-react/pull/3633))
+- Added the ability to hide the clear filter button on the filter component ([#3049](https://github.com/Shopify/polaris-react/pull/3049))
+- Right-align `disclosure` when using `textAlignLeft` for `Button` ([#3709](https://github.com/Shopify/polaris-react/pull/3709))
+- Remove all transitions from `Button` ([#3709](https://github.com/Shopify/polaris-react/pull/3709))
+- New `select` option for `disclosure` in `Button` ([#3709](https://github.com/Shopify/polaris-react/pull/3709))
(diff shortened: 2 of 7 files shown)


axelerator (approved):

laurenmayers (approved):

src/components/DropZone/DropZone.tsx
```diff
@@ -35,7 +35,7 @@ type Type = 'file' | 'image';
 
 export interface DropZoneProps {
   /** Label for the file input */
-  label?: string;
+  label?: React.ReactNode;
```
alex-page, line 38: @Tom-Bonnike I might be really silly but should this say `React.ReactNode | string` or does `React.ReactNode` accept a string and render it out?

Tom-Bonnike: `React.ReactNode` is anything that can be rendered by React:
```js
    type ReactText = string | number;
    type ReactChild = ReactElement | ReactText;

    interface ReactNodeArray extends Array<ReactNode> {}
    type ReactFragment = {} | ReactNodeArray;
    type ReactNode = ReactChild | ReactFragment | ReactPortal | boolean | null | undefined;
```

Which is also why that change is non-breaking!

alex-page (approved):

PR #3794, opened 2021-01-07
Undo high contrast media query name change

AndrewMusgrave:

The media query name was changed in https://github.com/Shopify/polaris-react/pull/3675 Sounds like it was supposed to be changed back in https://github.com/Shopify/polaris-react/pull/3775 but may have been missed. This change would need to be included in a major version as its a breaking change.

I'm omitting a changelog entry because this change hasn't been released yet.

Undoing a name change

### DIFF
--- a/src/components/DatePicker/DatePicker.scss
+++ b/src/components/DatePicker/DatePicker.scss
@@ -185,7 +185,7 @@ $range-end-border-radius: rem(30px);
     &:hover {
       background: var(--p-interactive-hovered);
       color: var(--p-text-on-interactive);
-      @include ms-high-contrast-outline;
+      @include high-contrast-outline;
     }
 
     @include focus-ring;
--- a/src/components/DropZone/DropZone.scss
+++ b/src/components/DropZone/DropZone.scss
@@ -184,7 +184,7 @@ $dropzone-stacking-order: (
     }
 
     &:hover {
-      @include ms-high-contrast-outline;
+      @include high-contrast-outline;
     }
   }
 
--- a/src/components/Filters/Filters.scss
+++ b/src/components/Filters/Filters.scss
@@ -135,7 +135,7 @@ $list-filters-footer-height: rem(70px);
     }
 
     &:hover {
-      @include ms-high-contrast-outline;
+      @include high-contrast-outline;
     }
   }
 }
--- a/src/components/Tabs/Tabs.scss
+++ b/src/components/Tabs/Tabs.scss
@@ -13,7 +13,7 @@ $focus-state-box-shadow-color: rgba(0, 0, 0, 0.8);
 }
 
 @mixin focus-ring-override {
-  @include ms-high-contrast-outline {
+  @include high-contrast-outline {
     @include reset-before-pseudo-content;
   }
 }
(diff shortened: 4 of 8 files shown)


ginabak (commented):
👋 Hello! I have a question from #3775 @ben10code, when we removed `-ms-high-contrast`, were we supposed to remove all consumptions of `-ms-high-contrast`?

When I did a global search I found 33 results that still consume -ms-high-contrast 😨.

See here: https://github.com/search?q=%22-ms-high-contrast%22+repo%3AShopify%2Fpolaris-react&type=code 

For example: https://github.com/Shopify/polaris-react/blob/ccd60944bd5a6a2ed509f8c12f846c52a83b902e/src/components/ActionList/ActionList.scss#L207

ginabak (approved):
Besides my question above: LGTM! 🥳

ben10code:
> 👋 Hello! I have a question from #3775 @ben10code, when we removed `-ms-high-contrast`, did we want to remove all consumptions of `-ms-high-contrast`? OR just remove it from `ms-high-contrast-outline`?
> 
> When I did a global search I found 33 results that still consume `-ms-high-contrast`. 😅
> 
> See here: https://github.com/search?q=%22-ms-high-contrast%22+repo%3AShopify%2Fpolaris-react&type=code
> 
> For example:
> 
> https://github.com/Shopify/polaris-react/blob/ccd60944bd5a6a2ed509f8c12f846c52a83b902e/src/components/ActionList/ActionList.scss#L207

They don't use the mixin that we updated directly, and I don't think we have

> 👋 Hello! I have a question from #3775 @ben10code, when we removed `-ms-high-contrast`, did we want to remove all consumptions of `-ms-high-contrast`? OR just remove it from `ms-high-contrast-outline`?
> 
> When I did a global search I found 33 results that still consume `-ms-high-contrast`. 😅
> 
> See here: https://github.com/search?q=%22-ms-high-contrast%22+repo%3AShopify%2Fpolaris-react&type=code
> 
> For example:
> 
> https://github.com/Shopify/polaris-react/blob/ccd60944bd5a6a2ed509f8c12f846c52a83b902e/src/components/ActionList/ActionList.scss#L207

They are not directly consuming the updated mixin, so I think it should be fine. 

> 👋 Hello! I have a question from #3775 @ben10code, when we removed `-ms-high-contrast`, did we want to remove all consumptions of `-ms-high-contrast`? OR just remove it from `ms-high-contrast-outline`?
> 
> When I did a global search I found 33 results that still consume `-ms-high-contrast`. 😅
> 
> See here: https://github.com/search?q=%22-ms-high-contrast%22+repo%3AShopify%2Fpolaris-react&type=code
> 
> For example:
> 
> https://github.com/Shopify/polaris-react/blob/ccd60944bd5a6a2ed509f8c12f846c52a83b902e/src/components/ActionList/ActionList.scss#L207

Hi 👋 ! It should be fine. Since we updated the mixin, we have to ensure that it's not a breaking change for consumers and it is not the case. The results from your search are not directly using the mixin that we updated. 

However, that's a good catch, I think it's worth creating an issue and mention the drawbacks of using the `-ms-high-contrast` media query.

kyledurand (approved):

alex-page (approved):

PR #3796, opened 2021-01-08
[Page, Header] remove new DL

kyledurand:

Fixes https://github.com/Shopify/polaris-ux/issues/556

Removes new DL conditions and merges styles with legacy for Page and Header components

## 

### How to 🎩

Opening as a draft because I recommend considerable tophatting. 

Start with the page examples in storybook and then move to the details page
On the details page, remove / modify all page props in as many configurations you have time for. 
Test both at as many break points as possible

### 🎩 checklist

### DIFF
--- a/src/components/Page/components/Header/Header.scss
+++ b/src/components/Page/components/Header/Header.scss
@@ -59,6 +59,7 @@ $action-menu-rollup-computed-width: rem(40px);
 }
 
 .PaginationWrapper {
+  flex: 0 0 auto;
   display: flex;
   justify-content: flex-end;
   margin-left: spacing(extra-tight);
--- a/src/components/Page/components/Header/Header.scss
+++ b/src/components/Page/components/Header/Header.scss
@@ -83,12 +83,7 @@ $action-menu-rollup-computed-width: rem(40px);
 .AdditionalNavigationWrapper {
   display: flex;
   flex: 1 0 auto;
-  margin: 0 spacing(tight) 0 0;
   justify-content: flex-end;
-
-  @include page-content-when-not-fully-condensed {
-    margin: 0 spacing(extra-loose);
-  }
 
   @include print-hidden;
 }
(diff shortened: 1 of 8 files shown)


alex-page:
There is a few Percy changes, unsure if they should be there or not.

kyledurand:
Looks like the badge might be wrapping early, or always. I'll double check that. The rest seem to be unrelated (Card spacing it seems like)

src/components/Page/components/Header/Header.scss
```diff
@@ -52,57 +50,41 @@ $action-menu-rollup-computed-width: rem(40px);
...
-    flex: 0 0 auto;
-    margin-right: spacing();
-  }
+  margin-right: spacing();
 
   @include print-hidden;
 }
 
 .PaginationWrapper {
-  flex: 0 0 auto;
```
alex-page, line 68: Should this stay?

src/components/Page/components/Header/Header.scss
```diff
@@ -52,57 +50,41 @@ $action-menu-rollup-computed-width: rem(40px);
...
-      // stylelint-disable-next-line selector-max-combinators, selector-max-specificity
-      &:hover,
-      &:active,
-      &:focus {
-        border: 1px solid var(--p-border-neutral-subdued);
-      }
+  button {
+    // stylelint-disable declaration-no-important
+    border: 1px solid var(--p-border-neutral-subdued) !important;
+    box-shadow: none !important;
```
alex-page, line 70: I assume the `!important` is needed now, is there a way we can clean this up. Happy to ship and fix later.

alex-page (approved):

src/components/Page/components/Header/Header.scss
kyledurand: It's needed because there was specificity before with the added class. Not sure we can remove it and not sure that re-working button will fix it either. I'm fairly fine with leaving !important for now though since this is wrapped in `.PaginationWrapper`. Should be fairly safe

src/components/Page/components/Header/Header.scss
kyledurand: Added this back. I think it was removed by accident so good catch 👍

kyledurand:
There was a reference to .newDesignLanguage in Title.scss which was causing the wrapping of the badge. Fixed that up

alex-page (approved):
This is looking good on Percy and locally. Thanks @kyledurand

PR #3832, opened 2021-01-13
Fix button focus offset

alex-page:

The latest release has a regression for filled button focus styles brought up by @whizkydee

This PR adds back the offset and removes some duplicated code.

### How to 🎩

1. Open the button examples and focus through them

### 🎩 checklist

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -698,9 +698,9 @@ $stacking-order: (
     }
 
     &.newDesignLanguage {
-      @include focus-ring($border-width: border-width('base'));
       border-color: currentColor;
       box-shadow: 0 0 0 border-width('base') currentColor;
+      @include focus-ring($border-width: rem(2px));
 
       &:focus {
         @include focus-ring($style: 'focused');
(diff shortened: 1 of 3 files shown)


src/components/Button/Button.scss
```diff
@@ -698,9 +698,9 @@ $stacking-order: (
     }
 
     &.newDesignLanguage {
-      @include focus-ring($border-width: border-width('base'));
       border-color: currentColor;
       box-shadow: 0 0 0 border-width('base') currentColor;
+      @include focus-ring($border-width: rem(2px));
```
whizkydee, line 703: We could use the thick variant of `border-width` here as well.

```suggestion
      @include focus-ring($border-width: border-width('thick'));
```

whizkydee (approved):
Left a non-blocking comment, but otherwise, this is good to go. Tophat checks out!

src/components/Button/Button.scss
alex-page: This is a good idea, I noticed we don't do it in lots of places. I think a separate PR to clean that up would be useful.

PR #3834, opened 2021-01-13
Fix up secondary action button

alex-page:

Fixes https://github.com/Shopify/polaris-ux/issues/520

Adding a text color so destructive actions are also reset.

### How to 🎩

1. Go to the details page and the `View` button should not have white text.

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/ActionMenu/components/SecondaryAction/SecondaryAction.scss
+++ b/src/components/ActionMenu/components/SecondaryAction/SecondaryAction.scss
@@ -9,6 +9,7 @@ $button-spacing: rem(12px);
   button {
     @include focus-ring($border-width: rem(1px));
     background: transparent !important;
+    color: var(--p-text) !important;
     box-shadow: none !important;
     border-radius: var(--p-border-radius-base) !important;
     padding-left: $button-spacing;
--- a/playground/DetailsPage.tsx
+++ b/playground/DetailsPage.tsx
@@ -534,6 +534,7 @@ export function DetailsPage() {
         ...actions,
         {
           content: 'View',
+          destructive: true,
           onAction: () => {
             // eslint-disable-next-line no-console
             console.log(previewValue);


src/components/ActionMenu/components/SecondaryAction/SecondaryAction.scss
```diff
@@ -9,6 +9,7 @@ $button-spacing: rem(12px);
   button {
     @include focus-ring($border-width: rem(1px));
     background: transparent !important;
+    color: var(--p-text) !important;
```
alex-page, line 12: This is kind of weird as it makes the button look not destructive. Should we remove the ability for destructive secondary actions?

kyledurand: Can primary be destructive still? If so then I think we should remove primary from the allowed props here

dleroux: There are instances where the secondary action is destructive. Pretty sure I've need that.

dleroux:
While I was tophatting your other PR I just happen to use this one for comparison. This one has the opposite problem.

Screen Shot 2021-01-14 at 7 46 27 AM

src/components/ActionMenu/components/SecondaryAction/SecondaryAction.scss
avocadomayo: @alex-page this is what you're referring to, right? The "View" action which is secondary doesn't have a differentiating colour?

Screen Shot 2021-01-13 at 4 43 13 PM

And seconding @dleroux - I've needed a destructive secondary action in `ActionMenu` on several occasions

alex-page: Yeah exactly the `View` button looks like the other secondary actions. We might need to get some additional UX opinions but I wonder if the balance a destructive button in the page would bring too much attention, if that is the case should it look the same or is there a way to make it look destructive without it being so distracting.

alex-page:
> While I was tophatting your other PR I just happen to use this one for comparison. This one has the opposite problem.
> 
> Screen Shot 2021-01-14 at 7 46 27 AM

This is fixed in master, I will merge it into this branch now.

tomoyukikashiro:
@dleroux Can this PR be merged ? I need destructive secondaryAction button!

PR #3837, opened 2021-01-13
[ToastManager] Change aria-live to assertive for accessibility

tarynmusgrave:

Fixes https://github.com/Shopify/web/issues/34584 

Toast messages using aria-live set to 'polite' are ignored by screen reader. There is discussion in the linked PR as for the decision made to choose this solution over extending the time a toast spends on screen. 

### How to 🎩

### 🎩 checklist

- [X] Tested for [accessibility](https://github.com/Shopify/polaris-react/blob/master/documentation/Accessibility%20testing.md)

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -13,6 +13,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Fixed an incorrect translation key for `accessibilityLabel` in `Tooltip`([#3843](https://github.com/Shopify/polaris-react/pull/3843))
 - Fix shadows on filled `Button`s not touching the bottom edge ([#3841](https://github.com/Shopify/polaris-react/pull/3841))
 - Adjust `Thumbnail` icon color to be subdued ([#3846](https://github.com/Shopify/polaris-react/pull/3846))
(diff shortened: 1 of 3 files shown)


kyledurand (commented):
Wondering if we could default to assertive and add a prop that allows consumers to change to polite?

cc @alex-page and @kaelig do you think `assertive` might be too much for all popovers? Just thinking out loud 🤷‍♂️

kaelig:
Thanks for the ping! I'd recommend going for assertive as suggested in this pull request, and adding a prop only if we receive feedback that this creates issues in specific instances.

AndrewMusgrave:
I have the same thoughts as @kaelig . API is easy to add but difficult to remove. Since we think this is the best, accessible choice let's roll with it and extend the API based on feedback.

AndrewMusgrave (commented):
LGTM ✅ I'll 🎩 web in the morning

kaelig (approved):
I haven't had the chance to tophat but it makes sense. This may be something we consider as a "bug fix" rather than an announcement, but that's a nit pick!

AndrewMusgrave (approved):
🎩 💯

PR #3849, opened 2021-01-15
[AppProvider & ThemeProvider] remove ndl

alex-page:

Fixes https://github.com/Shopify/polaris-ux/issues/531

- [x] Makes the `theme` property optional, light color scheme is the default
- [x] Removes legacy topbar theming
- [x] Removes legacy theming functions
- [ ] Fixes ThemeProvider tests

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist


BPScott:
i'm probably gonna be pretty full for the rest of today. I'll get to this on Monday :)

src/components/AppProvider/AppProvider.tsx
```diff
@@ -89,7 +89,7 @@ export class AppProvider extends Component<AppProviderProps, State> {
     const {theme = {}, children} = this.props;
 
     const {intl, link} = this.state;
-    const features = {newDesignLanguage: false, ...this.props.features};
+    const features = {...this.props.features};
```
BPScott, line 92: No need to spread and create a new object anymore. We can get rid of this variable and do the following below

```
<FeaturesContext.Provider value={this.props.features}>
```

src/components/Filters/Filters.tsx
```diff
@@ -390,6 +390,7 @@ class FiltersInner extends Component<CombinedProps, State> {
 
     const filtersContainerMarkup = isNavigationCollapsed ? (
       <Sheet
+        accessibilityLabel={moreFiltersLabel}
```
BPScott, line 393: This Sheet stuff seems unrelated to this PR? Pull it out to a separate PR?

src/components/PolarisTestProvider/PolarisTestProvider.tsx
```diff
@@ -77,14 +77,12 @@ export function PolarisTestProvider({
 
   const uniqueIdFactory = new UniqueIdFactory(globalIdGeneratorFactory);
 
-  const features = {newDesignLanguage: false, ...featuresProp};
+  const features = {...featuresProp};
```
BPScott, line 80: No need to spread anymore and create a new object anymore. We can remove this declaration and avoid renaming the features prop above

```diff
- features: featuresProp = {},
+ features = {}
```

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -26,7 +25,7 @@ interface ThemeProviderThemeConfig extends Discard<ThemeConfig, 'colorScheme'> {
 
 interface ThemeProviderProps {
   /** Custom logos and colors provided to select components */
-  theme: ThemeProviderThemeConfig;
+  theme?: ThemeProviderThemeConfig;
```
BPScott, line 28: When would we not passing a theme into a theme provider make sense?

src/components/ThemeProvider/tests/ThemeProvider.test.tsx
```diff
@@ -1,31 +1,14 @@
 import React, {useContext} from 'react';
-import {createMount} from 'test-utilities';
+import {mountWithApp} from 'test-utilities/react-testing';
```
BPScott, line 2: The `test-utilities` index reexports everything from `test-utilities/react-testing` for the shortest naming

```suggestion
import {mountWithApp} from 'test-utilities';
```

src/utilities/theme/utils.ts
```diff
@@ -23,34 +22,28 @@ interface CustomPropertiesConfig extends ThemeConfig {
...
+  });
 }
 
 export function buildThemeContext(
   themeConfig: ThemeConfig,
   cssCustomProperties?: CustomPropertiesLike,
 ): Theme {
   const {logo, colors = {}, colorScheme} = themeConfig;
-  const {topBar, ...newDesignLanguageColors} = colors;
+  const {...colorValues} = colors;
```
BPScott, line 42: No need to spread this to create a new value. Below we can use colors directly


```diff
- colors: colorValues,
+ colors
```

BPScott (commented):
Quick first pass

src/components/ThemeProvider/ThemeProvider.tsx
alex-page: If nothing is passed in then it returns the default CSS variables. What do you think about that assumption?

BPScott: I would have expected a theme without an config should be a non-operation:

```
<ThemeProvider theme={{colorScheme: 'dark'}}>
  <ThemeProvider>
    <Button>I would expect this button to still be in a dark scheme</Button>
  </ThemeProvider>
</ThemeProvider>
```

Forcing a color scheme of light seems to breakt that

BPScott:
Dont' forget removing mentions of new DL in the ThemeProvider/AppProvider readmes

BPScott:
I've updated low-hanging fruit on the ThemeProvider and AppProvider readmes. 


What do you think we should do about the AppProvider's "with theme" and "with theme using all keys" examples? They currently document the legacy theming of the topbar. Should we remove them and instead point people at the ThemeProvider documentation or should we rework them to describe modern theming?

alex-page:
> What do you think we should do about the AppProvider's "with theme" and "with theme using all keys" examples? They currently document the legacy theming of the topbar. Should we remove them and instead point people at the ThemeProvider documentation or should we rework them to describe modern theming?

I think we should be removing the ability to theme the `TopBar` with the `theme`. You can see the new way to theme `TopBar` in `Top bar themed for Plus`. The only useful information would be the `logo`, `colorScheme` or `color` overrides.

kyledurand:
> I would hope to deprecate the theming of the TopBar using the previous API.

I agree but that would break the current contextual save bar in web since it uses the logo and not contextControl. Not a deal breaker just something we'd have to update in web before this ships

src/components/Popover/components/PopoverOverlay/PopoverOverlay.tsx
```diff
@@ -51,7 +51,7 @@ export interface PopoverOverlayProps {
   fixed?: boolean;
   hideOnPrint?: boolean;
   onClose(source: PopoverCloseSource): void;
-  colorScheme?: InversableColorScheme;
+  colorScheme?: NonNullable<ThemeProviderProps['theme']>['colorScheme'];
```
kyledurand, line 54: Curious what `NonNullable` is doing here if the prop is optional?

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -121,19 +93,13 @@ export function ThemeProvider({
...
-  colorScheme: InversableColorScheme | undefined,
-  parentColorScheme: OriginalColorScheme | undefined,
-) {
+  colorScheme: ThemeConfig['colorScheme'],
+  parentColorScheme?: Theme['colorScheme'],
+): Theme['colorScheme'] {
   if (colorScheme == null) {
     return parentColorScheme || 'light';
-  } else if (isInverseColorScheme(colorScheme)) {
+  } else if (colorScheme === 'inverse') {
```
kyledurand, line 102: ✨  Nice and simple, I'm typically a fan of typeguards but this makes more sense in this case

kyledurand (approved):
Lots changes but the code and readmes look good.

src/components/ThemeProvider/ThemeProvider.tsx
BPScott: This is still a typeguard, it's just one that TS can work out automatically instead of us having  to be explicit.

Inside this if block the type of `colorScheme` is still narrowed down `'light'| 'dark' | 'inverse'` to `'light' | 'dark'`  (`undefined` was already narrowed away thanks to the prior if block)

src/components/Popover/components/PopoverOverlay/PopoverOverlay.tsx
BPScott: Note where the Generic ends (spacing added for emphasis) 

```
NonNullable<ThemeProviderProps['theme']>    ['colorScheme']
```

`ThemeProviderProps['theme']` is  `ThemeConfig | undefined`. The NonNullable knocks off the undefined bit of that so we have an object and thus can access the type of `ThemeConfig['colorScheme']` (which is itself potentially undefined)

BPScott:
I would hope to deprecate the theming of the TopBar using the previous API.

That's what this PR does. Here's where you rip out being able to define `{colors: {topBar: {background: '#225062'}}` in the theme config: https://github.com/Shopify/polaris-react/pull/3849/files#diff-d20bc9ba53e6c4c404e50adf23cb39bf5fe6525effc3acd98867dacd454826b0L84


This is the configuration that the ["With Theme"](https://github.com/Shopify/polaris-react/blob/1c32f482c955e42b27b2b214af53062d46ad853d/src/components/AppProvider/README.md) example demonstrates

alex-page:
I think we should remove the logo and other parts in the theme provider that are no longer used. We should fix the contextual save bar in web to resolve this issue https://github.com/Shopify/web/issues/32812

src/utilities/features/types.ts
```diff
@@ -3,6 +3,5 @@ export interface FeaturesConfig {
 }
 
 export interface Features {
-  newDesignLanguage: boolean;
   [key: string]: boolean;
```
BPScott, line 6: TODO: Remove this too so that people who use the newDesignLanguage prop will be alerted that it has gone away.

Features should be an empty object: `export interface Features {}`

Which isn't very useful now but let's keep it around in case we want to add toggles in the future

BPScott (commented):
see inline TODO comment

BPScott (approved):
Pending final percy run 🎉

PR #3725, opened 2020-12-16
Actionlist remove aria selected

alex-page:
### WHY are these changes introduced?

Fixes accessibility issue `ARIA attribute is not allowed: aria-selected="true"`

### WHAT is this pull request doing?

Removing `aria-selected` as the `ActionList` should be a list of actions and the `OptionList` is a selectable list.

### 🎩 checklist

* [x] Tested on [mobile](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting.md#cross-browser-testing)
* [x] Tested on [multiple browsers](https://help.shopify.com/en/manual/intro-to-shopify/shopify-admin/supported-browsers)
* [x] Tested for [accessibility](https://github.com/Shopify/polaris-react/blob/master/documentation/Accessibility%20testing.md)
* [x] Updated the component's `README.md` with documentation changes
* [x] [Tophatted documentation](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting%20documentation.md) changes in the style guide
* [x] For visual design changes, pinged one of @ HYPD, @ mirualves, @ sarahill, or @ ry5n to update the Polaris UI kit

PR #3854, opened 2021-01-15
[ActionList] Remove child selector from .active styles

emma-boardman:

I noticed when tophatting the Online Store Editor bump to Polaris 5.13.1 that the indicator styles were missing from the `.active` `ActionList` buttons.

image

This can be resolved by removing a stray `>` child selector in `ActionList.scss`
- #3725 removed the aria-selected from  `<li>` in `ActionList`
- The `aria-selected` was used as a selector in the styles, and required the `>` child selector to target the `button` within the `li`
- The `.active` class is applied directly to the `button`, so the `>` child selector is not needed (and is preventing the indicator styles from displaying) 

**Before:**

image

**After**

image

- removes  `>` child selector in `ActionList.scss`

### 🎩 checklist


kyledurand (approved):
Thanks Emma! I'm wondering if they were removed on purpose but  I'll let Alex answer that. 

Can you add un unreleased entry under bug fix for this?

beefchimi (approved):
Looks like the visual tests from https://github.com/Shopify/polaris-react/pull/3725 did not pick up on the "indicator" not rendering because of the switch to `.active`. Does anyone know why this is? I would have expected this example to have failed:
http://polaris-react.herokuapp.com/?path=/story/all-components-action-list--action-list-with-destructive-item

alex-page (approved):

PR #3865, opened 2021-01-18
Remove public color export

alex-page:

Fixes https://github.com/Shopify/polaris-react/issues/3848

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/TextField/TextField.scss
+++ b/src/components/TextField/TextField.scss
@@ -16,11 +16,11 @@ $stacking-order: (
   position: relative;
   display: flex;
   align-items: center;
-  color: color('ink', 'lighter');
+  color: var(--p-text);
   cursor: text;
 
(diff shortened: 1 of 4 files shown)


BPScott (changes requested):
This is approach is gonna have some weird knock on effects. Now anything that uses IconProps from base (like IconAction) won't have a colour which seems like a problem.

Originally IconProps was moved into the root types.ts to avoid circular dependencies. Though now we have `import type` that issue might have gone away these days.


---

If we want to remove Color from being a public export we should change the `export * from ./types` line in index.ts and replace it with an explicit list of named exports instead of using the namespace export

PR #3867, opened 2021-01-18
[Scrollable] Add focusable for when child content is not focusable

alex-page:

Fixes https://github.com/Shopify/web/issues/36623 where focus is lost when moving between elements in a modal.

Adding an optional property to make the `Scrollable` component focusable.

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Scrollable/Scrollable.tsx
+++ b/src/components/Scrollable/Scrollable.tsx
@@ -134,7 +134,7 @@ export class Scrollable extends Component<ScrollableProps, State> {
             {...rest}
             ref={this.setScrollArea}
             // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
-            tabIndex={focusable ? 0 : -1}
+            tabIndex={focusable ? 0 : undefined}
           >
             {children}
           </div>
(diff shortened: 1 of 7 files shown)


kyledurand (approved):
Probably worth adding this to the changelog

src/components/Scrollable/Scrollable.tsx
```diff
@@ -131,7 +134,7 @@ export class Scrollable extends Component<ScrollableProps, State> {
             {...rest}
             ref={this.setScrollArea}
             // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
-            tabIndex={0}
+            tabIndex={focusable ? 0 : -1}
```
dleroux, line 137: I think in this case we might want `undefined` since -1 is a thing. I needed something similar here: https://github.com/Shopify/polaris-react/pull/3595/files#diff-3a6c8eaf1de81279fa78d47ad8cd27a34087c7c29881ac4c9a1c881858132c08L214

For context on that change I linked to... when inside a Combobox/autocomplete pattern, we can't have anything steal focus from the text field. In the current implementation, we watch for it and then set the focus back on the field if it wasn't intentional. That results in a poor ux for screen reader users as the lost and gain of focus creates lots of unnecessary noise.

dleroux, line 137:
```suggestion
            tabIndex={focusable ? 0 : undefined}
```

andrewmcgov (approved):
Should this prop be added to `Modal` too (maybe in a separate PR) to cover the event that the modal contents have no focusable elements but the `Scrollable` is still rendered?

PR #3898, opened 2021-01-21
[Scrollable] Fix scrolled to bottom checking for sub-pixel values

### LINKED ISSUE #3225: Modal > `onScrolledToBottom` broken in some screen resolutions
## Issue summary

While working in `shopify/web`, I've noticed that the `onScrolledToBottom` in `Modal` does not work in some screen sizes. 

Moreover, I think it is good to note that I have a 1440p monitor. It works fine in all screen sizes for my coworker with a 1080p monitor.

Here is a little video showing the issue. 

- First try: 1953 px x 1050px. Works ✅ 
- Second try: 2059px x 1083 px. Works ✅ 
- Third try: 2097px x 1108px. ❌ 
- Fourth try (not in the video): 2097 px x 997 px ❌ 

https://drive.google.com/file/d/1928qia8wK4v5enksp5b869Vvjr5a3Jmi/view?usp=sharing

## Expected behavior

`onScrolledToBottom` should be triggered in all resolutions when the bottom of the modal is reached.

## Actual behavior

`onScrolledToBottom` is not triggered in all resolutions. 

https://drive.google.com/file/d/1928qia8wK4v5enksp5b869Vvjr5a3Jmi/view?usp=sharing

## Reduced test case

Unfortunately, I am unable to reproduce this issue in a sandbox with just a modal. It works just fine. However, when I copy the code from the sandbox into web, it doesn't work (in web). 

## Specifications

- Polaris version number: 5.2.1
- Browser: Version 85.0.4183.83 (Official Build) (64-bit)
- Device: MacBook Pro (15-inch, 2019)
- Operating System: macOS Catalina V10.15.3

### LINKED ISSUE #3607: Popover.Pane doesn't handle onScrolledToBottom properly
## Issue summary

[Popover.Pane](https://polaris.shopify.com/components/overlays/popover) component doesn't handle the `onScrolledToBottom` event properly on some screens.

Digging further into this problem. I noticed that [Popover.Pane](https://polaris.shopify.com/components/overlays/popover) can be composed of the Scrollable component and that component has a `handleScroll` handler where it checks if it has scrolled to the bottom (`const hasScrolledToBottom = scrollHeight - scrollTop === clientHeight;`)

This condition doesn't become true thus not triggering the event due to `scrollHeight - scrollTop` coming very close but not equal to `clientHeight`. When I tested it, `scrollHeight - scrollTop  = 295.2` and `clientHeight = 295`

## Expected behavior

ezgif-com-gif-maker

## Actual behavior

ezgif-com-gif-maker-1

The code is taken from the official Polaris example on 'Popover with lazy loaded list'
It works on certain screens. Not really sure what is the root cause.

I tried it on the following screens:

1600 x 900 (Works)
2560 x 1440 (Doesn't work)
1920 x 1080 (Doesn't work)

## Reduced test case

https://codesandbox.io/s/incsl?module=App.js

rorz:

Fixes https://github.com/Shopify/polaris-react/issues/3607, Fixes https://github.com/Shopify/polaris-react/issues/3225

Certain resolutions, OSes, and browser zoom-level combinations cause `scrollTop` to be a sub-pixel value ([StackOverflow example](https://stackoverflow.com/questions/52470549/subpixel-scroll-issue-cant-set-scrolltop-properly-on-chrome-69)), which does not satisfy the strict-equality check on the `hasScrolledToBottom` calculation of the `Scrollable` component.

In order to catch these cases, and fire the `onScrolledToBottom` callback prop, I've made the strict-equality check a less-than-or-equal to operator.

Notes: 
- "less-than" might seem like a counter-intuitive operator here, but I didn't want to modify the original calculation and it is the appropriate comparison
- I experimented using `Math.ceil(scrollTop)` to catch sub-pixel values that are _under_ the threshold, but this introduces another issue whereby scrolling up from the "at bottom" state can cause a re-fire of the callback. In testing, it seems to me that `clientHeight` is increased by a pixel at zoom levels where `scrollTop` is under the threshold anyway.

You can reproduce the original issue on Mac / Chrome by setting your browser zoom level to 67%, 75%, or 125%.

## 

### How to 🎩

<details>

**Important** — To emulate a breaking resolution, in certain cases you need to set the browser zoom level (e.g. Chrome on Mac) to 67%, 75%, or 125%...

```jsx

import {Page} from '../src';
import {Scrollable} from '../src/components';

  const [atBottom, setAtBottom] = useState(false);
  return (
    <Page title="Playground">
      <Scrollable
        style={{height: 100, background: 'yellow'}}
        onScrolledToBottom={() => setAtBottom(true)}
      >
        <div
          style={{
            background: 'blue',
            height: 150,
          }}
        />
      </Scrollable>
      {atBottom && (
        <h1
          style={{color: 'red', background: '#ccc'}}
          onClick={() => setAtBottom(false)}
        >
          BOTTOMED_OUT (click to reset)
        </h1>
      )}
    </Page>
  );
}
```

</details>

### 🎩 checklist

- ~Tested on [mobile](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting.md#cross-browser-testing)~
- ~Tested for [accessibility](https://github.com/Shopify/polaris-react/blob/master/documentation/Accessibility%20testing.md)~
- ~Updated the component's `README.md` with documentation changes~
- ~[Tophatted documentation](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting%20documentation.md) changes in the style guide~
- ~For visual design changes, pinged one of @ HYPD, @ mirualves, @ sarahill, or @ ry5n to update the Polaris UI kit~

### DIFF
--- a/src/components/Scrollable/Scrollable.tsx
+++ b/src/components/Scrollable/Scrollable.tsx
@@ -160,7 +160,7 @@ export class Scrollable extends Component<ScrollableProps, State> {
     const shouldTopShadow = Boolean(shadow && scrollTop > 0);
 
     const canScroll = scrollHeight > clientHeight;
-    const hasScrolledToBottom = scrollHeight - scrollTop === clientHeight;
+    const hasScrolledToBottom = scrollHeight - scrollTop <= clientHeight;
 
     if (canScroll && hasScrolledToBottom && onScrolledToBottom) {
       onScrolledToBottom();


anasyusef:
@rorz I applied your patch to check how it would work out. I found a small issue with the calculation. For some values of `scrollTop`, it went over `clientHeight` by a slight amount, which would cause the condition to fail. To get around it, I just added 1 to `scrollTop`, to account for that amount. Check out the screenshot attached.
image

alex-page:
@rorzis this still an issue? I can't replicate it with this codesandbox. https://codesandbox.io/s/floral-sun-rzebp1?file=/index.js

Feel free to reopen if it is an issue.

PR #3910, opened 2021-01-25
[AutoComplete] Rebuild autocomplete internals to use new accessible components

voiid:

Fixes https://github.com/Shopify/polaris-react/issues/3811

Fixes accessibility issues on all uses of the Autocomplete component.

This PR includes the following:
- [x] Introduction of new `ComboBox` and `ListBox` components built by @dleroux with help from @AndrewMusgrave. These are currently subcomponents of Autocomplete for now, but our intention is to get these in a state (mostly missing documentation) to become top level components in the future. These follow the aria 1.2 spec.
- [x] Storybook story bug fixes to better showcase the component. New stories introduced for more complete documentation of the components capabilities.
- [x] New and improved Autocomplete component which supports much more intuitive visual representations for different states, much improved keyboard navigation and screen reader support.

### Visual change audit
To maintain full backwards compatibility we did our absolute best to match all of the visuals of the component 1 to 1. There are a few intentional changes to help differentiate between hover and selection states but otherwise, everything should match.

The link below contains a comprehensive before and after comparison of just about every prop combination.

https://docs.google.com/spreadsheets/d/12X7igZB6jN3mzuhaXL9WK2SlK2HvRhJyAepCJkZRB7M/edit?userstoinvite=andrew.musgrave@shopify.com&ts=6011e9fc#gid=0

### Render graphs

* It's hard to tell the whole story of render times in screen shots - these shots are mainly displaying the render amounts which can we been above the flame graph to the right

|Before|After|
|-|-|
|old-1|new-1|
|old-2|new-2|
|old-3|new-3|
|old-4|new-4|
|old-5|new-5|

### Test coverage

The test coverage is near perfect. The failure stems from the old combobox being added as additions

|8/8 Old|8/8 Old|8/10 Old - 2/10 New|
|-|-|-|
|Image 2021-02-01 at 12 25 49 PM|Image 2021-02-01 at 12 26 53 PM|Image 2021-02-01 at 12 28 33 PM|

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist


locales/en.json
```diff
@@ -8,7 +8,8 @@
       "labelWithInitials": "Avatar with initials {initials}"
     },
     "Autocomplete": {
-      "spinnerAccessibilityLabel": "Loading"
+      "spinnerAccessibilityLabel": "Loading",
+      "ellipsis": "{content}…"
```
AndrewMusgrave, line 12: We're moving the ellipsis prop to a translated string for internationalization

src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -1,23 +1,29 @@
...
+  MappedOption,
+  MappedAction,
+} from './components';
 
 export interface AutocompleteProps {
   /** A unique identifier for the Autocomplete */
   id?: string;
   /** Collection of options to be listed */
-  options: ComboBoxProps['options'];
+  options: ComboBoxOldProps['options'];
```
AndrewMusgrave, line 20: We're keep the interface the same to prevent breaking changes

src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -42,10 +48,9 @@ export interface AutocompleteProps {
 // generated *.d.ts files.
 
 export const Autocomplete: React.FunctionComponent<AutocompleteProps> & {
-  ComboBox: typeof ComboBox;
-  TextField: typeof TextField;
+  ComboBox: typeof ComboBoxOld;
```
AndrewMusgrave, line 51: ^

src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -42,10 +48,9 @@ export interface AutocompleteProps {
...
+  TextField: typeof ComboBox.TextField;
```
AndrewMusgrave, line 52: Both text fields extend the base Polaris textfield without adding additional API's

src/components/Autocomplete/components/ComboBox/ComboBox.tsx
```diff
@@ -1,377 +1,153 @@
-import React, {useState, useEffect, useCallback} from 'react';
+import React, {useState, useCallback, useMemo, Children} from 'react';
```
AndrewMusgrave, line 1: This component is a rebuild not a refactor so feel free to ignore the diff

src/components/Autocomplete/components/ComboBoxOld/ComboBox.tsx
```diff
@@ -0,0 +1,377 @@
+import React, {useState, useEffect, useCallback} from 'react';
```
AndrewMusgrave, line 1: This is the old combobox - nothing has been changed

src/components/Autocomplete/components/ComboBoxOld/tests/ComboBox.test.tsx
```diff
@@ -0,0 +1,661 @@
+import React from 'react';
```
AndrewMusgrave, line 1: ^

src/components/Autocomplete/components/MappedAction/MappedAction.tsx
```diff
@@ -0,0 +1,122 @@
+import React, {useMemo} from 'react';
```
AndrewMusgrave, line 1: We created a new action component to map the difference between the old & new API

src/components/Autocomplete/components/MappedOption/MappedOption.tsx
```diff
@@ -0,0 +1,51 @@
+import React, {memo} from 'react';
```
AndrewMusgrave, line 1: We create a new option component to map the differences between the old + new API

src/test-utilities/react-testing.tsx
```diff
@@ -1,13 +1,18 @@
 import React from 'react';
-import {createMount, mount} from '@shopify/react-testing';
+import {
+  createMount,
+  mount,
+  Element as ReactTestingElement,
```
AndrewMusgrave, line 5: Renaming because `Element` from the DOM is global

tests/setup.ts
```diff
@@ -11,9 +11,7 @@ const IGNORE_ERROR_REGEXES = [
   /React does not recognize the `%s` prop on a DOM element/,
 ];
 
-const IGNORE_WARN_REGEXES: RegExp[] = [
-  /Deprecation: <FilterControl \/> is deprecated\. This is a private component, do not use it\. This component might be removed in a minor version update\. Use <Filters \/> instead\./,
-];
+const IGNORE_WARN_REGEXES: RegExp[] = [/Deprecation:.*/];
```
AndrewMusgrave, line 14: Changed this to a more general deprecation warning to remove test clutter.

alex-page (approved):
This looks good to me. I would strongly recommend getting these child components down to the global level as soon as possible. The tech debt of supporting them isn't high value.

dleroux:
Thank you much for doing this @voiid and @AndrewMusgrave 

2 things I noticed while tophatting:

1. when the list is scrollable using the keys keeps the list scrolled at the top. Was this an intentional change? I'd expect the focus to move the last visible item before scrolling.

scrollIntoView

2. I see that we are still using actual inputs in the list when it's selectable. I was under the impression that ListBox should not contain any focusable elements. I believe the reasoning is that it can steal focus from the field and interfere with screen readers. I see that is not happening though so we might be ok?

AndrewMusgrave:
@dleroux The keyboard focus was an intentional change. We've fixed some bugs in the interaction and reverting to how it's currently in the autocomplete.

As for the checkbox @voiid reached out to @HYPD for confirmation on visuals. The new styles weren't a great indication that multi-select was possible. So we've reverted to the old visuals. The checkbox has a role of presentation so it'll be skipped by screen readers & if you've top-hatted that it cannot gain focus I think we've in the clear!

@voiid anything else you wanted to add?

voiid:
@dleroux @AndrewMusgrave Just about covers it! Both changes were made to reduce friction with getting this change out the door.

For scrolling, I was thinking that we could revisit the scrolling behaviour with the team when we do the work to move the `ComboBox` and `ListBox` to the global level. Having it match the existing functionality for now is an effort to reduce any friction with getting this particular change out the door.

For checkboxes: The reason this change in particular was important is that while the new version communicated multiple selection very well once selections were made, having the visual checkboxes show does a better job of letting users know that multiple selection is possible _before_ a selection is made. As Andrew mentioned it uses the presentation role so we should be safe on that front. The one negative here is that there could be an expectation that pressing the space key should select the option (in this case, it will actually enter a space into the text field and trigger a search change) but this matches the existing behaviour of the component (though the original had a weird bug that would only check on blur) so we thought it was acceptable for this version of the component. Something we should explore with usability testing for sure though!

dleroux:
I just checked in the admin where we do use the Autocomplete and it does arrow down the list before scrolling into view. Where are you seeing this behavior?

dleroux (approved):
I have not been able to tophat in web but in Polaris it works great 🎉

src/components/ComboBox/README.md
```diff
@@ -0,0 +1,421 @@
...
+
+The `ComboBox` component is a component that implements part of the [Aria 1.2 combobox](https://www.w3.org/TR/wai-aria-practices-1.2/#combobox) specs on a TextField and a popover containing a ListBox. Like `Autocomplete`, the `ComboBox` allows merchants to quickly search through and select from large collections of options.
+
+---
+
+## Best practices
+
+The ComboBox component should:
+
+- Be clearly labeled so it’s obvious to the merchant what type of options will be available
```
AndrewMusgrave, line 22: Same comment about `obvious ` 😄

src/components/ComboBox/README.md
```diff
@@ -0,0 +1,421 @@
...
+          placeholder="Search customers"
+        />
+      }
+    >
+      {listBoxMarkup}
+    </ComboBox>
+  );
+}
+```
+
```
AndrewMusgrave, line 363: Should we add an example for `allowMultiple` & `onScrolledToBottom`?

src/components/ComboBox/README.md
```diff
@@ -0,0 +1,421 @@
...
+- [Accessible development on iOS](https://developer.apple.com/accessibility/ios/)
+
+<!-- /content-for -->
+
+<!-- content-for: web -->
+
+### Structure
+
+The `ComboBox` component is based on the [ARIA 1.2 combobox pattern](https://www.w3.org/TR/wai-aria-practices-1.1/#combobox). It is a combination of a single-line
+`TextField` and a `Popover`. The current implementation expects a [`ListBox`] component to be used.
```
AndrewMusgrave, line 399:
```suggestion
The `ComboBox` component is based on the [ARIA 1.2 combobox pattern](https://www.w3.org/TR/wai-aria-practices-1.1/#combobox). It is a combination of a single-line `TextField` and a `Popover`. The current implementation expects a [`ListBox`] component to be used.
```

src/components/ComboBox/README.md
```diff
@@ -0,0 +1,421 @@
...
+
+The `ComboBox` popover displays below the text field or other control by default so it is easy for merchants to discover and use. However, you can change the position with the `preferredPosition` prop.
+
+ComboBox features can be challenging for merchants with visual, motor, and cognitive disabilities. Even when they’re built using best practices, these features can be difficult to use with some assistive technologies. Merchants should always be able to search, enter data, or perform other activities without relying on the combobox.
```
AndrewMusgrave, line 403:
```suggestion
`ComboBox` features can be challenging for merchants with visual, motor, and cognitive disabilities. Even when they’re built using best practices, these features can be difficult to use with some assistive technologies. Merchants should always be able to search, enter data, or perform other activities without relying on the combobox.
```

src/components/ComboBox/README.md
```diff
@@ -0,0 +1,421 @@
...
+
+<!-- usageblock -->
+
+#### Do
+
+Use combobox as progressive enhancement to make the interface easier to use for most merchants.
```
AndrewMusgrave, line 409: I think these are lists

```suggestion
- Use combobox as progressive enhancement to make the interface easier to use for most merchants.
```

src/components/ComboBox/README.md
```diff
@@ -0,0 +1,421 @@
...
+
+#### Don’t
+
+Require that merchants make a selection from the combobox to complete a task.
```
AndrewMusgrave, line 413: ^

```suggestion
- Require that merchants make a selection from the combobox to complete a task.
```

src/components/ComboBox/README.md
```diff
@@ -0,0 +1,421 @@
...
+
+<!-- end -->
+
+### Keyboard support
+
+- Give the combobox text input keyboard focus with the <kbd>tab</kbd> key (or <kbd>shift</kbd> + <kbd>tab</kbd> when tabbing backwards)
```
AndrewMusgrave, line 419:
```suggestion
- Give the combobox's text input keyboard focus with the <kbd>tab</kbd> key (or <kbd>shift</kbd> + <kbd>tab</kbd> when tabbing backwards)
```

AndrewMusgrave (commented):
Looked over the docs for `Autocomplete`, `ComboBox` & `ListBox`. They're looking great 😍

katihoeschen:
Thanks for looping me in, @voiid. These docs are in great shape! I'm dropping my feedback below, but it’s mostly questions and thinking out loud. I'll be OOO Thursday and Friday this week, but we have a meeting on Monday in case you want to chat through any of this.

ListBox - intro:
- Typo: `part the` → `part of the`.
- Is `composition` a common enough concept that it won’t slow folks down? I don’t see mention of it in the current Polaris components.
- Super nitpicky: `it is possible` seems to imply difficulty. Maybe `Composition lets you customize` or `you can`?

ListBox - content guidelines:
- Typo: `an Listbox` → `a Listbox`
- Can you include an example of a clear and descriptive item?

ComboBox - intro:
- Super nitpicky: `The ComboBox component is a component that implements` → `The ComboBox component implements`

ComboBox - accessibility > structure, Autocomplete - accessibility > structure:
- Would it be helpful to include recommended reasons or examples of using `PreferredPosition` to change the component position? Given the preceding sentence, it seems risky not to add some guardrails about changing the position.

Applies to all 3 docs:
- Examples section: For the basic implementation, it might be worth calling out what makes it basic so the reader doesn’t have to compare against the other examples.
- Accessibility section: I can see both pros and cons of having accessibility best practices called out in a separate section, but have you considered including that information in the relevant sections above instead? That might bring more attention to accessibility direction and make it seem less optional.
- Throughout: I see the component names both in all lowercase letters (`listbox`, `combobox`) and camel caps (`ListBox`, `ComboBox`). Is there a logic for using one vs. the other?

src/components/ListBox/README.md
```diff
@@ -0,0 +1,188 @@
...
+category: Lists and tables
+keywords:
+  - list
+  - listbox
+  - interactive list
+---
+
+# ListBox
+
+The `ListBox` component is a list component that implements part the [Aria 1.2 ListBox specs](https://www.w3.org/TR/wai-aria-practices-1.2/#Listbox). It presents a list of options and allows users to select one or more of them. Through composition, it is possible to customize the presentation of these lists by using sections, headers or custom elements.
```
katihoeschen, line 12:
```suggestion
The `ListBox` component is a list component that implements part of the [Aria 1.2 ListBox specs](https://www.w3.org/TR/wai-aria-practices-1.2/#Listbox).
```

katihoeschen, line 12: If you need more structure than the standard component offers, use composition to customize the presentation of these lists by using sections, headers, or custom elements.

PR #3912, opened 2021-01-25
Remove #AppFrameMainContent link and update SkipToContent link to target #AppFrameMain instead

henryyi:

References https://github.com/Shopify/global-nav/issues/775

The "Skip to Content" link by default targets a hidden anchor element within Frame via `#AppFrameMainContent`. This presents an accessibility concern since the focus does not advance to actual visible content item. This is also undesirable when Voice Over is activated since it will cause it to focus on and describe the hidden anchor.
By targeting the main container (`#AppFrameMain`) instead the focus gets moved to the first focusable (or "voiceable") element inside of `main`.

Remove the `#AppFrameMainContent` link from Frame. Update `Skip to Content` to target #AppFrameMain.

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Frame/Frame.tsx
+++ b/src/components/Frame/Frame.tsx
@@ -1,4 +1,4 @@
-import React, {PureComponent, createRef} from 'react';
+import React, {PureComponent, createRef, MouseEvent} from 'react';
 import {MobileCancelMajor} from '@shopify/polaris-icons';
 import {durationSlow} from '@shopify/polaris-tokens';
 import {CSSTransition} from 'react-transition-group';
@@ -62,7 +62,6 @@ interface State {
 const GLOBAL_RIBBON_CUSTOM_PROPERTY = '--global-ribbon-height';
 
 const APP_FRAME_MAIN = 'AppFrameMain';
-
 const APP_FRAME_NAV = 'AppFrameNav';
 const APP_FRAME_TOP_BAR = 'AppFrameTopBar';
 const APP_FRAME_LOADING_BAR = 'AppFrameLoadingBar';
@@ -79,8 +78,6 @@ class FrameInner extends PureComponent<CombinedProps, State> {
   private contextualSaveBar: ContextualSaveBarProps | null = null;
   private globalRibbonContainer: HTMLDivElement | null = null;
   private navigationNode = createRef<HTMLDivElement>();
-  private skipToMainContentTargetNode =
-    this.props.skipToContentTarget || createRef<HTMLAnchorElement>();
 
   componentDidMount() {
     this.handleResize();
@@ -367,9 +364,12 @@ class FrameInner extends PureComponent<CombinedProps, State> {
     this.setState({skipFocused: false});
   };
 
-  private handleClick = () => {
-    this.skipToMainContentTargetNode.current &&
-      this.skipToMainContentTargetNode.current.focus();
+  private handleClick = (event: MouseEvent<HTMLAnchorElement>) => {
+    const {skipToContentTarget} = this.props;
+    if (skipToContentTarget && skipToContentTarget.current) {
+      skipToContentTarget.current.focus();
+      event?.preventDefault();
+    }
   };
 
   private handleNavigationDismiss = () => {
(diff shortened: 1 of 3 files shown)


src/components/Frame/Frame.tsx
```diff
@@ -221,11 +219,12 @@ class FrameInner extends PureComponent<CombinedProps, State> {
 
     const skipTarget = skipToContentTarget?.current
       ? skipToContentTarget.current.id
-      : APP_FRAME_MAIN_ANCHOR_TARGET;
+      : APP_FRAME_MAIN;
 
     const skipMarkup = (
       <div className={skipClassName}>
         <a
+          className="hy_link_in_frame"
```
kyledurand, line 227: Doesn't look like you need this

src/components/Frame/Frame.tsx
```diff
@@ -262,15 +261,6 @@ class FrameInner extends PureComponent<CombinedProps, State> {
...
       ) : null;
 
-    const skipToMainContentTarget = skipToContentTarget ? null : (
-      // eslint-disable-next-line jsx-a11y/anchor-is-valid
-      <a
-        id={APP_FRAME_MAIN_ANCHOR_TARGET}
-        ref={this.skipToMainContentTargetNode}
-        tabIndex={-1}
-      />
-    );
```
kyledurand, line 272: Removing this would be a bit of a breaking change. Are you sure we don't need this anymore?

kyledurand (commented):
Tophat looks fine here. Can you build for web and test on shop1 as well?

See: [testing in a consuming project](https://github.com/Shopify/polaris-react#testing-in-a-consuming-project)
Basically just dev up in web and then run:
`yarn run build-consumer web` in polaris-react

src/components/Frame/Frame.tsx
henryyi: sorry test code...

src/components/Frame/Frame.tsx
BPScott: Yeah this seems half-done

Either: there's still a need for a skipToMainContentTargetNode prop - in which case we need to do something with iit OR it's not needed anymore, in which case we need to comment on the prop saying it's deprecated and docuementation detailing that too., 
 

What happens if a users specifies a `skipToContentTarget` prop - we no longer emmbed that an.

henryyi: From what I can tell this is only used as a target for Skip To Content. I did a global search for `AppFrameMainContent` and only see it referenced in Polaris and GlobalNav (which I updated here https://github.com/Shopify/global-nav/pull/859 as well).
However with that said please let me know if there are other places I should check.

henryyi: Also the reason this was removed is because focus was still being given to the anchor since it's technically a valid focusable element within `<main>`. So by remove this it should go to a visible element in the content.

henryyi: @BPScott yes I left the prop since it was needed according to this PR https://github.com/Shopify/polaris-react/pull/2080

henryyi:
Minor update to remove the unused target ref. I also added prevent default after focusing on custom targets.

I also considered keeping the target ref and placing it on `<main>` and using `focusFirstKeyboardFocusableNode` to focus on an item in `<main>` when the skipToContent link is clicked. However this makes the Voice Over experience worst since it will skip over non-focusable elements that it would usually read. For example if we just link to `#AppFrameMain` then the screen reader to read the first `<h1>` in `<main>`.

Top hatted in shop1 on chrome, ff and safari (macOS)

mathildebuenerd (commented):
Looks nice!
🎩 in Storybook on the _Frame in an application_ component with Voiceover, and it works as expected. After clicking the skip link we are moved to the first focusable element in the main, in this case the h1

BPScott:
Removing myself from the reviewers list as I've moved off the polaris team

src/components/Frame/Frame.tsx
alex-page: Reading through this I think this change is fine. `skipToContentTarget` either points towards `APP_FRAME_MAIN` or the provided ref prop. This removes the private `skipToMainContentTargetNode` and the link that is rendered. Is the risk that someone incorrectly referenced that element?

alex-page (approved):

PR #3917, opened 2021-01-26
Default to new Design Language in storybook

BPScott:

Getting ready for v6. By defaulting to the new DL in storybook we get to compare the version-6.0.0 branch against main so we can check for regressions.

Change the default rendering in Storybook to use the new design language.

This is going to do a number on percy on this PR but accepting everything here as a new baseline, then comparing v6 against this new baseline will help us catch regressions.

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -55,7 +55,7 @@ function AppProviderWithKnobs(
   );
 }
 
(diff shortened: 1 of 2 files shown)


kyledurand (approved):
One step closer 🦶

alex-page (approved):

PR #3931, opened 2021-01-27
[Modal, Frame(Nav), Sheet] Trapped virtual cursor inside dialogs

AndrewMusgrave:

Keyboard focus is trapped whoever virtual cursors are not

### Current keyboard trapping

|Modal|Sheet|Nav|
|-|-|-|
|key-modal|key-sheet|key-nav|

### Before/After virtual cursor navigation

||Before|After|
|-|-|-|
|<b>Modal</b>|vc-modal-old|key-modal|
|<b>Sheet</b>|vc-sheet-old|vc-sheet|
|<b>Nav</b>|vc-nav-old|vc-nav|

* Make using of the aria-modal & role attribute to trap virtual cursors in dialogs

### How to 🎩

* Loading storybook and play around modal/nav/sheet with a virtual cursor

### DIFF
--- a/src/components/Frame/Frame.tsx
+++ b/src/components/Frame/Frame.tsx
@@ -135,6 +135,8 @@ class FrameInner extends PureComponent<CombinedProps, State> {
           classNames={navTransitionClasses}
         >
           <div
+            aria-modal
+            role="dialog"
             ref={this.navigationNode}
             className={navClassName}
             onKeyDown={this.handleNavKeydown}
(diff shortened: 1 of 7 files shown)


src/components/Frame/Frame.tsx
```diff
@@ -135,6 +135,9 @@ class FrameInner extends PureComponent<CombinedProps, State> {
           classNames={navTransitionClasses}
         >
           <div
+            aria-modal
+            role="dialog"
```
kyledurand, line 139: This makes sense on mobile but does it make things confusing at all when the nav isn't hidden?

kyledurand (approved):
This is awesome Andrew, just one question about mobille vs desktop label / role

src/components/Frame/Frame.tsx
AndrewMusgrave: Nice catch, I thought this was mobile only markup. I made a change to conditionally apply the attributes & added tests for it.

kyledurand (approved):
Nice, additions look good 🥇

PR #3952, opened 2021-02-01
Add prop docs to generated storybook docs pages

BPScott:

Storybook generates docs pages for us. Let's teach it about where to pull component docs from.

Splitting off part of the functionality of #2888 which is general investigative PR

Teaching storybook what what component corresponds to each README file.

### How to 🎩

- open storybook
- Load docs page for any story generated by a README file (which is everything but the playground)
- See that a props table is generated and displayed.


kyledurand (approved):
Code looks good. There was an a11y check error though. Seemed unrelated so I just re-ran it

BPScott:
Worked out how to stop ally tests being a pain so :shipit:

PR #3961, opened 2021-02-03
Convert some simple components to be functional

BPScott:

Death to class components!

Converting the following components to be functional components:

- Connected Item
- Filter ConnectedFilterControl Item
- List
- Tabs List

### DIFF
--- a/src/components/List/List.tsx
+++ b/src/components/List/List.tsx
@@ -1,4 +1,4 @@
-import React, {NamedExoticComponent} from 'react';
+import React from 'react';
 
 import {classNames, variationName} from '../../utilities/css';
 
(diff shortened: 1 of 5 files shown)


src/components/List/List.tsx
```diff
@@ -1,4 +1,4 @@
-import React, {PureComponent} from 'react';
+import React, {NamedExoticComponent} from 'react';
```
kyledurand, line 1: Were you planning on doing something with this?

kyledurand (approved):
🎩  and code look good. Just that NamedExoticComponent that you can 🔥

src/components/List/List.tsx
BPScott: ooh whoops. That was a hangover from a slightly different trick to tempt the types into playing nice that wasn't needed. Removed!

PR #3962, opened 2021-02-03
[a11y]: Remove outline & border uses of ms-high-contrast

### LINKED ISSUE #3826: [a11y]: remove all instances of -ms-high-contrast
## Issue summary
There are many instances of `-ms-high-contrast`
See here: https://github.com/search?q=%22-ms-high-contrast%22+repo%3AShopify%2Fpolaris-react&type=code

According [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/-ms-high-contrast): 

> This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future

More on the limitations of `-ms-high-contrast`, [here](https://developer.paciellogroup.com/blog/2016/12/windows-high-contrast-mode-the-limited-utility-of-ms-high-contrast/).  

For even MORE context, see [this slack conversation](https://shopify.slack.com/archives/C03T74ZTD/p1607460676013100). 

## What should we do?  / Potential solutions

We must remove the media query since it only properly works in Edge and not recommended for production. 

For the use of an outline, this fix below is okay: 
From: 

```javascript
    outline: 1px solid ms-high-contrast-color('text');
  }
```

to 

```javascript
outline: $border-width solid transparent;
```

However, in this [interesting article](https://blogs.windows.com/msedgedev/2020/09/17/styling-for-windows-high-contrast-with-new-standards-for-forced-colors/), we know that we can use a media query ... 

```CSS
@media (forced-colors: active) {
  /* Styles here only apply to Forced Colors Mode */
}
```
that basically does the same thing as `-ms-high-contrast: active` **_HOWEVER_**, 😭 it does NOT work for firefox. 

In [this article ](https://hacks.mozilla.org/2020/07/adding-prefers-contrast-to-firefox/), it suggests creating a custom media query called `prefers-contrast`, which is quite laborious. 

We must then remove all `@media (forced-colors: active)` in the Polaris repo too. It is currently used in certain places as a temporary fix. 

There are still many instances of `-ms-high-contrast` in the Polaris repo. 
See here: https://github.com/search?q=%22-ms-high-contrast%22+repo%3AShopify%2Fpolaris-react&type=code

According [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/-ms-high-contrast):

> This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future

More on the limitations of `-ms-high-contrast`, [here](https://developer.paciellogroup.com/blog/2016/12/windows-high-contrast-mode-the-limited-utility-of-ms-high-contrast/).

ginabak:

(partly) Fixes #3826 : it only "partly" fixes the issue since this PR only removes the `outline` and `border` uses of  `-ms-high-contrast`. 

Inspired by this [interesting article](https://blogs.windows.com/msedgedev/2020/09/17/styling-for-windows-high-contrast-with-new-standards-for-forced-colors/), which states that 
```CSS
@media (forced-colors: active) {
  /* Styles here only apply to Forced Colors Mode */
}
```
basically does the same thing as `-ms-high-contrast: active` HOWEVER, 😭 it doesn’t work for firefox. See [this article ](https://hacks.mozilla.org/2020/07/adding-prefers-contrast-to-firefox/)which talks about creating a custom media query called prefers-contrast, which is quite laborious. This is a temporary fix and will require thought as to how to account for firefox in high contrast mode. 

There are still many instances of `-ms-high-contrast` in the Polaris repo. 
See here: https://github.com/search?q=%22-ms-high-contrast%22+repo%3AShopify%2Fpolaris-react&type=code

According [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/-ms-high-contrast):

> This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future

More on the limitations of `-ms-high-contrast`, [here](https://developer.paciellogroup.com/blog/2016/12/windows-high-contrast-mode-the-limited-utility-of-ms-high-contrast/).  

<details>
      <summary>Summary of your gif(s)</summary>

### Avatar 
avatar-high-contrast

### Banner 
banner-high-contrast

### Contextual Save Bar 
contextualSaveBar-high-contrast

### ProgressBar
progressbar-ms-high-contrast

### Toast
toast-high-contrast

### Tag
tag-high-contrast

#### SingleThumb
| Before        | After           |
|singlethumb-before|singlethumb-AFTERFIX-high-contrast

#### DualThumb
(the changes are subtle, but before, there are 2 outlines, after there's only 1)

| Before        | After           |
|dual-thumb-BEFORE-high-contrast|dual-thumb-AFTER-high-contrast

</details>

### How to 🎩

1. Go to VM & set to High Contrast
2. Compare side by side with http://polaris-react.herokuapp.com/?path=/story/playground-playground--playground 
3. There should be NO visual changes except with RangeSlider (Dual & Single noted in the before and after above)

### 🎩 checklist

OUTLINE not border

### DIFF
--- a/src/components/ActionList/ActionList.scss
+++ b/src/components/ActionList/ActionList.scss
@@ -52,15 +52,14 @@ $item-vertical-padding: ($item-min-height - line-height(body)) / 2;
   cursor: pointer;
   padding: $item-vertical-padding spacing(tight);
   border-radius: var(--p-border-radius-base);
-  border-top: 1px solid var(--p-surface); // 1px gap between elements
+  border-top: 1px solid transparent var(--p-surface); // 1px gap between elements
+  outline: 1px solid transparent;
 
   &:hover {
     background-color: var(--p-surface-hovered);
     text-decoration: none;
 
-    @media (-ms-high-contrast: active) {
-      outline: 1px solid ms-high-contrast-color('text');
-    }
+    @include high-contrast-outline($border-width: border-width(thicker));
   }
 
   &.active {
@@ -79,6 +78,7 @@ $item-vertical-padding: ($item-min-height - line-height(body)) / 2;
 
   &:focus:not(:active) {
     @include focus-ring($style: 'focused');
+    @include high-contrast-outline($border-width: border-width(thicker));
   }
 
   &.destructive {
--- a/src/components/ActionList/ActionList.scss
+++ b/src/components/ActionList/ActionList.scss
@@ -52,7 +52,6 @@ $item-vertical-padding: ($item-min-height - line-height(body)) / 2;
   cursor: pointer;
   padding: $item-vertical-padding spacing(tight);
   border-radius: var(--p-border-radius-base);
-  border-top: 1px solid transparent var(--p-surface); // 1px gap between elements
   outline: 1px solid transparent;
 
   &:hover {
--- a/src/components/ActionList/ActionList.scss
+++ b/src/components/ActionList/ActionList.scss
@@ -52,14 +52,15 @@ $item-vertical-padding: ($item-min-height - line-height(body)) / 2;
   cursor: pointer;
   padding: $item-vertical-padding spacing(tight);
   border-radius: var(--p-border-radius-base);
-  border-top: 1px solid transparent var(--p-surface); // 1px gap between elements
-  outline: 1px solid transparent;
+  border-top: 1px solid var(--p-surface); // 1px gap between elements
 
   &:hover {
     background-color: var(--p-surface-hovered);
     text-decoration: none;
 
-    @include high-contrast-outline($border-width: border-width(thicker));
+    @media (-ms-high-contrast: active) {
+      outline: 1px solid ms-high-contrast-color('text');
+    }
   }
 
   &.active {
@@ -78,7 +79,6 @@ $item-vertical-padding: ($item-min-height - line-height(body)) / 2;
 
   &:focus:not(:active) {
     @include focus-ring($style: 'focused');
-    @include high-contrast-outline($border-width: border-width(thicker));
   }
 
   &.destructive {
--- a/src/components/ActionList/ActionList.scss
+++ b/src/components/ActionList/ActionList.scss
@@ -52,8 +52,7 @@ $item-vertical-padding: ($item-min-height - line-height(body)) / 2;
   cursor: pointer;
   padding: $item-vertical-padding spacing(tight);
   border-radius: var(--p-border-radius-base);
-  padding-top: 1px solid transparent;
-  outline: 1px solid transparent;
+  border: 1px solid transparent;
 
   &:hover {
     background-color: var(--p-surface-hovered);
--- a/src/components/ActionList/ActionList.scss
+++ b/src/components/ActionList/ActionList.scss
@@ -52,6 +52,7 @@ $item-vertical-padding: ($item-min-height - line-height(body)) / 2;
   cursor: pointer;
   padding: $item-vertical-padding spacing(tight);
   border-radius: var(--p-border-radius-base);
+  margin-top: 1px;
   outline: 1px solid transparent;
 
   &:hover {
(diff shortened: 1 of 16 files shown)


src/components/ActionList/ActionList.scss
```diff
@@ -52,7 +52,6 @@ $item-vertical-padding: ($item-min-height - line-height(body)) / 2;
   cursor: pointer;
   padding: $item-vertical-padding spacing(tight);
   border-radius: var(--p-border-radius-base);
-  border-top: 1px solid transparent var(--p-surface); // 1px gap between elements
```
alex-page, line 55: `border-top: 1px solid transparent; // 1px gap between elements` is probably still wanted here

src/components/CheckableButton/CheckableButton.scss
```diff
@@ -43,10 +43,6 @@ $chekbox-label-margin: rem(20px);
     background: var(--p-action-secondary-pressed);
   }
 
-  @media (-ms-high-contrast: active) {
-    border: border-width(base) solid;
-  }
```
alex-page, line 48: Do we need the high contrast outline here?

src/components/RangeSlider/components/DualThumb/DualThumb.scss
```diff
@@ -98,9 +98,6 @@ $range-thumb-border-error: rem(2px) solid color('red');
     @include focus-ring($style: 'focused');
     outline: 0;
     border-color: var(--p-override-transparent);
-    @media (-ms-high-contrast: active) {
-      outline: 1px solid ms-high-contrast-color('text');
-    }
```
alex-page, line 103: Does this need the high contrast mixin? I think the `border-color: var(--p-override-transparent);` should be removed

alex-page (commented):
Looking good some quick comments!

src/components/ActionList/ActionList.scss
ginabak: 👋🏽 hi hi! When the `border-top` is added, it looks like this: 

WITH transparent border top


When it should look like this: 

withOUT transparent border top

src/components/RangeSlider/components/DualThumb/DualThumb.scss
ginabak: I removed the high-contrast mixin, and you're right, removing border-colour!

src/components/CheckableButton/CheckableButton.scss
ginabak: I removed it as it didn't add any visual changes 😄

sarahill (approved):
Visually looks good 👍 Thanks for this update!

src/components/ActionList/ActionList.scss
alex-page: I see, we need the border top so that the elements have a gap between them and does not cause the mouse to flicker when scrolling. This is only needed when outsidte of high contrast. If we can override it later that would be best.

ginabak: oooh hi hi! Sorry - just got around to this. Added a `margin` and I think that would fix this!

src/components/ActionList/ActionList.scss
```diff
@@ -52,7 +52,7 @@ $item-vertical-padding: ($item-min-height - line-height(body)) / 2;
   cursor: pointer;
   padding: $item-vertical-padding spacing(tight);
   border-radius: var(--p-border-radius-base);
-  margin-top: 1px;
+  padding-top: 1px solid transparent;
```
alex-page, line 55:
```suggestion
  border-top: 1px solid transparent;
```

src/components/ActionList/ActionList.scss
ginabak: This is what it looks like with the added border: 
ActionLIst

focused state:
ActionListFocused

alex-page (approved):

PR #3966, opened 2021-02-03
[TextField] Add default behaviour if onClearButtonClick is not provided

ryanwilsonperkin:

It would be nice if `onClearButtonClicked` wasn't required if it only needed to provide the same behaviour as calling `onChange('')`. This would allow us to get "clearing" functionality by only needing to add `clearButton={true}`.

Adds new behaviour to the function that handles "clear button" presses. Previously if no `onClearButtonClick` function was provided, pressing the clear button would have no effect. Now if no value is provided, it will try to fallback to calling the `onChange` function with an empty string, which is likely to be what the user expected as default behaviour.

This means that we can save developers a step in not having to pass a second function to `TextField`. As a comparison, if you previously wrote this:

```tsx
function Component() {
  const [textFieldValue, setTextFieldValue] = useState('Jaded Pixel');
  
  const handleTextFieldChange = useCallback(
    (value) => setTextFieldValue(value),
    [],
  );

  const handleClearButtonClick = useCallback(
    (value) => setTextFieldValue(value),
    [],
  );

  return (
    <TextField
      label="Store name"
      value={textFieldValue}
      onChange={handleTextFieldChange}
      clearButton
      onClearButtonClick={handleClearButtonClick}
    />
  );
}
```

you can now avoid the second function and instead write:

```tsx
function Component() {
  const [textFieldValue, setTextFieldValue] = useState('Jaded Pixel');
  
  const handleTextFieldChange = useCallback(
    (value) => setTextFieldValue(value),
    [],
  );

  return (
    <TextField
      label="Store name"
      value={textFieldValue}
      onChange={handleTextFieldChange}
      clearButton
    />
  );
}
```

to get the same behaviour. (Note: this doesn't change the behaviour of `onClearButtonClick` itself. If provided it will still take precedence).

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -7,6 +7,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Enhancements
 
 - Added `focus-visible` polyfill and default styles ([#3695](https://github.com/Shopify/polaris-react/pull/3695))
+- Add default onChange behaviour to `TextField` when `clearButton` is provided
 
 ### Bug fixes
 
--- a/src/components/TextField/README.md
+++ b/src/components/TextField/README.md
@@ -785,7 +785,38 @@ function TextFieldWithClearButtonExample() {
     [],
   );
 
-  const handleClearButtonClick = useCallback(() => setTextFieldValue(''), []);
+  return (
+    <TextField
+      label="Store name"
+      value={textFieldValue}
+      onChange={handleTextFieldChange}
+      clearButton
+    />
+  );
+}
+```
+
+### Text field with clear button and custom clearing behaviour
+
+<!-- example-for: web -->
+
+Customize the behaviour of hitting the clear button.
+
+```jsx
+function TextFieldWithCustomClearButtonExample() {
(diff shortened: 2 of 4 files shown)


ryanwilsonperkin:
This PR is currently failing the codecov check because there's no test for the path where neither `onChange` nor `onClearButtonClicked` is provided. I tried to write a test for this case but I'm having a hard time figuring out what a relevant assertion would be to make. The basic setup looks like this:

```tsx
it('does nothing when clear button is clicked if neither onClearButtonClick or onChange is provided', () => {
  const textField = mountWithAppProvider(
    <TextField
      id="MyTextField"
      label="TextField"
      type="search"
      value="test value"
      disabled
      clearButton
    />,
  );
  findByTestID(textField, 'clearButton').simulate('click');
});
```

(note: the `disabled` prop is required, because TextField requires one of `readOnly`, `disabled`, or `onChange` to be provided)

Typically we'd pass in a spy and assert that it was called, but we're kinda testing the _absence_ of that here. Anyone have any thoughts on what a relevant assertion would be? Is it cool to just have a test with _no_ assertions to exercise the case and ensure it doesn't error?

PR #3999, opened 2021-02-12
Updated Peer Dependencies to accept React 17.x.x

eran-pinhas:
React 17 is out and it there's no API change between 16 and 17

In my environment it seems to work as before

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,8 +89,8 @@
     "react-transition-group": "^4.4.1"
   },
   "peerDependencies": {
-    "react": "^16.9.0",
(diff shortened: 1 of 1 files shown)


vfonic:
This small PR was opened half a year ago. Is there no one from the Shopify development team that could have a look?

/cc @rdott @AndrewMusgrave @BPScott @kaelig @dleroux @tmlayton @danrosenthal @alex-page

BPScott:
Fixed in #4432. This will go out as part of the upcoming v7.0.0 release

PR #4005, opened 2021-02-17
[TextStyle] Fix an issue where inline code would be hard to select

kaelig:

The `::after` pseudo element goes on top of the code, making it impossible to select when clicking/selecting or double clicking on it.

It adds `pointer-events: none` to the pseudo element that covers the code, so that clicking on code isn't impacted by it.


kyledurand (approved):
👍

PR #4013, opened 2021-02-22
[IndexTable] Remove `in your store` from IndexTable bulk selection

lhoffbeck:

Fixes [shopify/web issue #38565](https://github.com/Shopify/web/issues/38565)

The IndexTable component assumes that the resource present in the list is a root-level resource, which means that the messaging when selecting all items in the list assumes that you're selecting all of that resource type across the entire store. This assumption may not always hold, which is what we're bumping up against in web.

This PR removes the `in your store` assumption from the `IndexTable` bulk action strings.

### 🎩 checklist


kyledurand:
Does this actually improve the UX of the component? What if we removed `in your store` from those translation keys instead? Would the user still be able to assume the correct context? 

Just trying to maintain simplicity in the code base and translations that we get locked into.

lhoffbeck:
@kyledurand this makes sense to me, double checking with the product person that asked me to make the change

kyledurand (approved):
👏

PR #4019, opened 2021-02-25
Ensure @charset is at the top of generated css

### LINKED ISSUE #4014: Invalid `@charset` in `dist/styles.css`
## Issue summary
Looking at the latest Polaris distribution, [in `dist/styles.css`](https://unpkg.com/browse/@shopify/polaris@6.0.1/dist/styles.css), I see a rogue charset in the middle of the file:

polaris 2021-02-24 00-19-43

Per [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/@charset):

> The @charset CSS at-rule specifies the character encoding used in the style sheet. It must be the first element in the style sheet and not be preceded by any character

## Expected behavior
`@charset` should be the first line.

## Actual behavior
`@charset` is in the middle of the file, which causes `esbuild` to warn:

```
 > node_modules/@shopify/polaris/dist/styles.css:1953:0: warning: "@charset" must be the first rule in the file
    1953 │ @charset "UTF-8";
         ╵ ~~~~~~~~
   node_modules/@shopify/polaris/dist/styles.css:1950:0: note: This rule cannot come before a "@charset" rule
    1950 │ .Polaris-Truncate{ display:block; overflow:hidden; white-space:nowrap; text-overflow:ellipsis; }
         ╵ ^
```

BPScott:

Fixes #4014 

Our generated css file was not handling the consolidation and hoisting of `@charset` declarations.

Pulls the `@charset` declaration up to the top of the file. Will error if there is are multiple competing charsets though that shouldn't ever happen because prettier should force all our source files to be utf-8.

### How to 🎩

- `yarn run build` and check `dist/styles.css` and note that `@charset "UTF-8";` is at the top of the file and not half way down.

If you'd like to be more thorough compare output against main:

- Create a build of main: `git checkout main && yarn clean && yarn build  && mv dist dist-orig`
- Create a build from this branch: `git checkout fix-charset && yarn clean && yarn build`
- Diff the old and new dist folders and note that the only change is the location of the charset in `dist/styles.css`: `diff -ru dist-orig dist`

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -34,6 +34,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Fixed `CheckableButton` missing border when focused ([#3987](https://github.com/Shopify/polaris-react/issues/3987))
 - Removed all `outline` and `border`instances of `-ms-high-contrast` as it is non-standard ([#3962](https://github.com/Shopify/polaris-react/pull/3962)).
 - Fixed `Autocomplete` popover height not being calculated correctly ([#4015](https://github.com/Shopify/polaris-react/pull/4015)).
+- Ensured `@charset` declaration is the first thing in our styles.css file ([#4019](https://github.com/Shopify/polaris-react/pull/4019))
 
 ### Documentation
 
--- a/config/rollup/plugin-styles.js
+++ b/config/rollup/plugin-styles.js
@@ -107,10 +107,16 @@ export function styles({
       );
     }
 
-    const css = bundleModuleIds
+    let css = bundleModuleIds
       .filter((id) => id in cssByFile)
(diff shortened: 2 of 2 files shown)


alex-page (approved):
Thanks @BPScott

GoodForOneFare (approved):
👍  Output is as described. I want to endlessly bikeshed, but this approach is more concise than anything I'd come up with.

BPScott:
> I want to endlessly bikeshed

Yeah this is a little hacky. There's probably a smart way to run the single file through postcss again to do this with a proper AST but it's going to be slower than this and I don't think we'll ever run into a case where this this will cause errors - If we end up with multiple file encodings our fix should be "write your source files in UTF8" rather than handle multiple encodings in the build.

PR #4030, opened 2021-03-03
Update tokens to v2.21.1

BPScott:

Gives us esm support for legacy tokens, and lets us remove some deep
imports inside dist-modern.

- Update `@shopify/polaris-tokens` to v2.21.0
- Import from `@shopify/polaris-tokens/dist-modern` where possible
- Use named imports in storybook manager config (the default export still works but is deprecated)


kyledurand (approved):
👏

alex-page (approved):
💎

PR #4033, opened 2021-03-04
[IndexTable] Fix UX / jank when resizing

### LINKED ISSUE #3991: [IndexTable] Sticky column headers are misaligned
## Issue summary
On the IndexTable component, the sticky column headers are misaligned with the column content after resizing the browser window. It looks like the problem is that the sticky column header widths are calculated on pageload and aren't updated on resize.

Related web issue [#37885](https://github.com/Shopify/web/issues/37885)

https://user-images.githubusercontent.com/3619012/107650520-918a3400-6c4c-11eb-9ab6-ebd36f777e16.mp4

## Expected behavior
Sticky column headers should align with columns after window resize

## Actual behavior
Column headers are misaligned with columns after resize

1. Open up an indextable instance (e.g. /admin/collections on shopify admin)
2. Resize your browser window
3. Scroll the indextable until the sticky headers appear

## Specifications
```bash
System:
    OS: macOS 10.15.7
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 223.27 MB / 32.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 10.15.0 - ~/.nvm/versions/node/v10.15.0/bin/node
    Yarn: 1.13.0 - ~/.dev/yarn/v1.13.0/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v10.15.0/bin/npm
  Browsers:
    Chrome: 88.0.4324.150
    Chrome Canary: 90.0.4414.0
    Firefox: 85.0.1
    Firefox Developer Edition: 85.0
    Safari: 13.1.3
  npmPackages:
    react: ^16.9.0 => 16.9.0 
    react-dom: ^16.9.0 => 16.9.0 
```

lhoffbeck:
Resolves #3991.

The IndexTable currently has some UX, jank, and performance issues when resizing. This PR aims to address a lot of these by making UX and performance fixes to make this resize buttery-smooth.

**1) Fixed sticky header alignment issues**
All the sticky headers are sized on load and not resized when the window width changes.

👀 _before (watch the Location column)_ 👀 

https://user-images.githubusercontent.com/3619012/109985851-31c6fc00-7cd3-11eb-9f50-962d805a235e.mp4

👀 _after (watch the Location column)_ 👀 

https://user-images.githubusercontent.com/3619012/109988962-0b569000-7cd6-11eb-8954-89f70faf2d5c.mp4

**2) Fixed first (non-sticky) column alignment issues when resizing browser**
First column header doesn't resize when window does.

👀 _before_ 👀 

https://user-images.githubusercontent.com/3619012/109986360-a7cb6300-7cd3-11eb-9041-1be1628149eb.mp4

👀 _after_ 👀 

https://user-images.githubusercontent.com/3619012/109989074-26c19b00-7cd6-11eb-8dd7-18b5990b2015.mp4

**3) Fixed first sticky header misalignment **
First column is out of place when you load the page and scroll down

👀 _before (watch the Name column)_ 👀 

https://user-images.githubusercontent.com/3619012/109985568-e44a8f00-7cd2-11eb-9a77-af1e8cd65cf2.mp4

👀 _after_ 👀 

https://user-images.githubusercontent.com/3619012/109989399-7607cb80-7cd6-11eb-9247-1d9e6112de1e.mp4

**4)  Fix flickering scrollbar **
Scrollbar flickers when resizing window large->small. The issue here was that the width of `.ScrollBarContent` was set by `--p-scroll-bar-content-width`, so when you resize smaller unless `--p-scroll-bar-content-width` is set on a bare-metal resize event (i.e. not by a throttled/debounced function), the scrollbar will be larger than the content when you size the window down and will flicker. Fixed this by setting this width to 0 when resizing and only actually setting `--p-scroll-bar-content-width` once resizing completes.

👀 _before_ 👀 

https://user-images.githubusercontent.com/3619012/109987181-67201980-7cd4-11eb-9cf2-5c1e0108be86.mp4

👀 _after_ 👀 

https://user-images.githubusercontent.com/3619012/109989526-93d53080-7cd6-11eb-8d69-ef8af87d4597.mp4

**5)  Fix sticky header not resizing smoothly **
Watch the upper right corner of the sticky header. As you resize, the width of the header goes out of shape with the rest of the card, this happens because the debounce period is at a slower framerate than it should be. Fix was to update to 60FPS (and converted to use `throttle` instead of `debounce` with `maxWait` for clarity).

👀 _before (watch upper right corner of sticky header)_ 👀 

https://user-images.githubusercontent.com/3619012/109987935-0ba25b80-7cd5-11eb-8081-29f2217fde36.mp4

👀 _after_ 👀 

https://user-images.githubusercontent.com/3619012/109989890-e9114200-7cd6-11eb-95d4-789d305ce41f.mp4

**6)  Performance stuff **
- converted the resize handler to use `useCallback` instead of `useMemo`, we really don't need to memoize anything for this
- cache the column header / sticky column header DOM lookups instead of performing them on every resize event
- increased the debounce period on the scrollbar resize event so it fires ~600% less frequently

## 

### How to 🎩

<details>

```

import React from 'react';

import {Card, IndexTable, Page, TextStyle, useIndexResourceState} from '../src';

  return (
    <Page title="Playground">

      <div style={{ marginBottom: 1000 }}>
        <SimpleIndexTableExample />
      </div>
    </Page>
  );
}


function SimpleIndexTableExample() {
  const customers = [
    {
      id: '3411',
      url: 'customers/341',
      name: 'Mae Jemison',
      location: 'Decatur, USA',
      orders: 20,
      amountSpent: '$2,400',
    },
    {
      id: '2561',
      url: 'customers/256',
      name: 'Ellen Ochoa',
      location: 'Los Angeles, USA',
      orders: 30,
      amountSpent: '$140',
    },
  ];
  const resourceName = {
    singular: 'customer',
    plural: 'customers',
  };

  const {
    selectedResources,
    allResourcesSelected,
    handleSelectionChange,
  } = useIndexResourceState(customers);

  const rowMarkup = customers.map(
    ({id, name, location, orders, amountSpent}, index) => (
      <IndexTable.Row
        id={id}
        key={id}
        selected={selectedResources.includes(id)}
        position={index}
      >
        <IndexTable.Cell>
        </IndexTable.Cell>
        <IndexTable.Cell>{location}</IndexTable.Cell>
        <IndexTable.Cell>{orders}</IndexTable.Cell>
        <IndexTable.Cell>{amountSpent}</IndexTable.Cell>
      </IndexTable.Row>
    ),
  );

  return (
    <Card>
      <IndexTable
        resourceName={resourceName}
        itemCount={customers.length}
        selectedItemsCount={
          allResourcesSelected ? 'All' : selectedResources.length
        }
        onSelectionChange={handleSelectionChange}
        headings={[
          {title: 'Name'},
          {title: 'Location'},
          {title: 'Order count'},
          {title: 'Amount spent'},
        ]}
      >
        {rowMarkup}
      </IndexTable>
    </Card>
  );
}


```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/IndexTable/IndexTable.tsx
+++ b/src/components/IndexTable/IndexTable.tsx
@@ -1,6 +1,7 @@
 import React, {useRef, useState, useEffect, useCallback, useMemo} from 'react';
 import {EnableSelectionMinor} from '@shopify/polaris-icons';
 import debounce from 'lodash/debounce';
+import throttle from 'lodash/throttle';
 import {CSSTransition} from 'react-transition-group';
 import {durationFast} from '@shopify/polaris-tokens';
 
@@ -52,6 +53,8 @@ export interface TableHeadingRect {
 }
 
 const SCROLL_BAR_PADDING = 4;
+const SIXTY_FPS = 1000 / 60;
+const SCROLL_BAR_DEBOUNCE_PERIOD = 300;
 
 function IndexTableBase({
   headings,
@@ -91,6 +94,10 @@ function IndexTableBase({
 
   const [isSmallScreenSelectable, setIsSmallScreenSelectable] = useState(false);
 
+  const tableHeadings = useRef<HTMLElement[]>([]);
+  const stickyTableHeadings = useRef<HTMLElement[]>([]);
+  const stickyHeaderWrapperElement = useRef<HTMLDivElement>(null);
+  const stickyHeaderCheckboxElement = useRef<HTMLDivElement>(null);
   const stickyHeaderElement = useRef<HTMLDivElement>(null);
   const scrollBarElement = useRef<HTMLDivElement>(null);
   const scrollingWithBar = useRef(false);
@@ -118,21 +125,57 @@ function IndexTableBase({
     );
   }, [handleSelectionChange, selectedItemsCount]);
 
-  const resizeTableHeadings = useCallback(() => {
-    if (!tableElement.current || !scrollableContainerElement.current) {
-      return;
-    }
+  const calculateFirstHeaderOffset = useCallback(() => {
+    return condensed
+      ? tableHeadingRects.current[0].offsetWidth
+      : tableHeadingRects.current[0].offsetWidth +
+          tableHeadingRects.current[1].offsetWidth;
+  }, [condensed]);
 
-    const measuredTableHeadingRects = Array.from(
-      tableElement.current.querySelectorAll('[data-index-table-heading]'),
-    ).map((heading) => ({
-      offsetWidth: heading instanceof HTMLElement ? heading.offsetWidth : 0,
-      offsetLeft: heading instanceof HTMLElement ? heading.offsetLeft : 0,
-    }));
-    const boundingRect = scrollableContainerElement.current.getBoundingClientRect();
-    tablePosition.current = {top: boundingRect.top, left: boundingRect.left};
-    tableHeadingRects.current = measuredTableHeadingRects;
-  }, []);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const resizeTableHeadings = useCallback(
+    throttle(() => {
+      if (!tableElement.current || !scrollableContainerElement.current) {
+        return;
+      }
+
+      const boundingRect = scrollableContainerElement.current.getBoundingClientRect();
+      tablePosition.current = {top: boundingRect.top, left: boundingRect.left};
+
+      tableHeadingRects.current = tableHeadings.current.map((heading) => ({
+        offsetWidth: heading.offsetWidth || 0,
+        offsetLeft: heading.offsetLeft || 0,
+      }));
+
+      if (tableHeadings.current.length === 0) {
+        return;
+      }
+
+      // update left offset for first column
+      if (tableHeadings.current.length > 1)
+        tableHeadings.current[1].style.left = `${tableHeadingRects.current[0].offsetWidth}px`;
+
+      // update the min width of the checkbox to be the be the un-padded width of the first heading
+      if (stickyHeaderCheckboxElement?.current) {
+        const elementStyle = getComputedStyle(tableHeadings.current[0]);
+        const boxWidth = tableHeadings.current[0].offsetWidth;
+        stickyHeaderCheckboxElement.current.style.minWidth = `calc(${boxWidth}px - ${elementStyle.paddingLeft} - ${elementStyle.paddingRight} + 2px)`;
+      }
+
+      // update sticky header min-widths
+      stickyTableHeadings.current.forEach((heading, index) => {
+        let minWidth = 0;
+        if (index === 0) {
+          minWidth = calculateFirstHeaderOffset();
+        } else if (tableHeadingRects.current.length > index) {
+          minWidth = tableHeadingRects.current[index].offsetWidth;
+        }
+
+        heading.style.minWidth = `${minWidth}px`;
+      });
+    }, SIXTY_FPS),
+    [calculateFirstHeaderOffset],
+  );
 
   const resizeTableScrollBar = useCallback(() => {
     if (scrollBarElement.current && tableElement.current && tableInitialized) {
@@ -143,14 +186,24 @@ function IndexTableBase({
     }
   }, [tableInitialized]);
 
-  const handleResize = useMemo(
-    () =>
-      debounce(() => {
-        resizeTableHeadings();
-        resizeTableScrollBar();
-      }, 50),
-    [resizeTableHeadings, resizeTableScrollBar],
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const debounceResizeTableScrollbar = useCallback(
+    debounce(resizeTableScrollBar, SCROLL_BAR_DEBOUNCE_PERIOD, {
+      trailing: true,
+    }),
+    [resizeTableScrollBar],
   );
+
+  const handleResize = useCallback(() => {
+    // hide the scrollbar when resizing
+    scrollBarElement.current?.style.setProperty(
+      '--p-scroll-bar-content-width',
+      `0px`,
+    );
+
+    resizeTableHeadings();
+    debounceResizeTableScrollbar();
+  }, [debounceResizeTableScrollbar, resizeTableHeadings]);
 
   const handleScrollContainerScroll = useCallback(
     (canScrollLeft, canScrollRight) => {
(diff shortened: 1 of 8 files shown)


lhoffbeck:
**NOTE** that I didn't add any test cases for this--this behavior hinges on being able to grab dimensions from the dom, which isn't possible with our test setup (afaik).

src/components/IndexTable/IndexTable.tsx
```diff
@@ -197,7 +250,16 @@ function IndexTableBase({
     scrollingContainer.current = false;
   }, []);
 
-  useEffect(() => resizeTableHeadings(), [headings, resizeTableHeadings]);
+  useEffect(() => {
+    tableHeadings.current = getTableHeadings();
```
lhoffbeck, line 254: Curious to get other input on this; I'm setting `tableHeadings` / `stickyTableHeadings` on a `ref` instead of in a `state` var since we don't need to do a re-render after it happens; as long as we set it in this `useEffect` we're covered whenever it's used by the resizer.

src/components/IndexTable/IndexTable.tsx
```diff
@@ -118,21 +125,57 @@ function IndexTableBase({
...
+        offsetLeft: heading.offsetLeft || 0,
+      }));
+
+      if (tableHeadings.current.length === 0) {
+        return;
+      }
+
+      // update left offset for first column
+      if (tableHeadings.current.length > 1)
+        tableHeadings.current[1].style.left = `${tableHeadingRects.current[0].offsetWidth}px`;
```
kmdavis, line 156: have you considered moving these inline styles to the css file, with css vars for the necessary data?

src/components/IndexTable/IndexTable.tsx
```diff
@@ -118,21 +125,57 @@ function IndexTableBase({
...
+  const calculateFirstHeaderOffset = useCallback(() => {
+    return condensed
+      ? tableHeadingRects.current[0].offsetWidth
+      : tableHeadingRects.current[0].offsetWidth +
+          tableHeadingRects.current[1].offsetWidth;
+  }, [condensed]);
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const resizeTableHeadings = useCallback(
+    throttle(() => {
```
kmdavis, line 137: one edge case to consider: this throttle will get reinitialized every time the callback gets reinitialized (which might not be an issue, since it's only dep is another callback). if this becomes an issue, one solution would be to store the throttle in a ref, so it doesn't change; the downside is that anything it references would also need to be a ref, but that might already be the case.

src/components/IndexTable/IndexTable.tsx
```diff
@@ -118,21 +125,57 @@ function IndexTableBase({
...
+      // update sticky header min-widths
+      stickyTableHeadings.current.forEach((heading, index) => {
+        let minWidth = 0;
+        if (index === 0) {
+          minWidth = calculateFirstHeaderOffset();
+        } else if (tableHeadingRects.current.length > index) {
+          minWidth = tableHeadingRects.current[index].offsetWidth;
+        }
+
+        heading.style.minWidth = `${minWidth}px`;
```
kmdavis, line 174: this could cause N layouts, where N = number of columns. if you can find a way to store list data in a css var, you could build up a set of css vars and then set them 1 time on a common parent, and avoid a bunch of layouts. not sure if it's possible though.

src/utilities/sticky-manager/sticky-manager.ts
```diff
@@ -25,26 +25,28 @@ interface StickyItem {
   ): void;
 }
 
+const SIXTY_FPS = 1000 / 60;
```
kmdavis, line 28: does polaris have a constants file? might be good to move this there if it exists

src/components/IndexTable/IndexTable.tsx
lhoffbeck: I didn't! What benefit does this give us (besides making it so we're less handsy with setting style props)? Might prevent some layout shifts I guess?

kmdavis: depending on implementation, it can prevent some layout shifts. basically, instead of setting inline styles on several elements (which can cause a layout) you set inline css vars on 1 common parent element, causing 1 layout.

it also moves the presentation logic into css, where it belongs. JS is only responsible then for providing data that css needs.

it also makes it easier/possible to override the implementation in either a 3rd party app, and/or via composition

lhoffbeck: Ah yeah that's what I figured. Our issue is that the number of columns is dynamic... which makes solving this non-programmatically difficult. Good insights though!

src/components/IndexTable/IndexTable.tsx
```diff
@@ -580,6 +646,7 @@ function IndexTableBase({
         className={stickyHeadingClassName}
         key={heading.title}
         style={headingStyle}
+        data-index-table-sticky-heading="1"
```
AndrewMusgrave, line 649:
```suggestion
        data-index-table-sticky-heading
```

Nitpick: if we're not using the `1` we could default it to true

AndrewMusgrave (commented):
Love how smooth this is looking 😍

src/utilities/sticky-manager/sticky-manager.ts
```diff
@@ -1,4 +1,4 @@
-import debounce from 'lodash/debounce';
+import throttle from 'lodash/throttle';
```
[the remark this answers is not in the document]
lhoffbeck: Ahhh gotcha, I was wondering why we were doing the `debounce` equivalent instead of just using `throttle`, keeping the bundle down makes sense. I'll replace it:)

src/components/IndexTable/tests/IndexTable.test.tsx
```diff
@@ -16,6 +18,11 @@ import {ScrollContainer} from '../components';
 import {SelectionType} from '../../../utilities/index-provider';
 import {AfterInitialMount} from '../../AfterInitialMount';
 
+jest.mock('../utilities', () => ({
+  ...(jest.requireActual('../utilities') as any),
```
AndrewMusgrave, line 22: Why was casting needed here?

AndrewMusgrave (approved):
LGTM! ✅

src/components/IndexTable/tests/IndexTable.test.tsx
lhoffbeck: bad copy paste 🙈

PR #4034, opened 2021-03-04
[ActionList] Remove focus ring on click #3649

### LINKED ISSUE #3649: [ActionList] focus-ring shows on click
The pattern in `polaris-react` is to visually render a `focus-ring` only on keyboard/programmatic `:focus`. However, `ActionList` appears to be rendering its `focus-ring` on mouse click:

skrabakl-shopify:

Fixes #3649 

Applied the same approach used on Polaris Button component. 

### How to 🎩

https://user-images.githubusercontent.com/67908111/110001290-850d6e80-7d14-11eb-8b34-0379f1263774.mov

### 🎩 checklist

### DIFF
--- a/src/components/ActionList/ActionList.scss
+++ b/src/components/ActionList/ActionList.scss
@@ -137,6 +137,11 @@ $item-vertical-padding: ($item-min-height - line-height(body)) / 2;
   margin-left: spacing();
 }
 
+.ContentBlock,
+.ContentBlock span {
+  display: block;
+}
+
 .Text {
   @include layout-flex-fix;
   flex: 1 1 auto;
--- a/src/components/ActionList/ActionList.scss
+++ b/src/components/ActionList/ActionList.scss
@@ -138,9 +138,7 @@ $item-vertical-padding: ($item-min-height - line-height(body)) / 2;
 }
 
 .ContentBlock,
-.Prefix,
-.Text,
-.Content {
+.ContentBlockInner {
   display: block;
 }
 
--- a/src/components/ActionList/ActionList.scss
+++ b/src/components/ActionList/ActionList.scss
@@ -138,7 +138,9 @@ $item-vertical-padding: ($item-min-height - line-height(body)) / 2;
 }
 
 .ContentBlock,
-.ContentBlock span {
+.Prefix,
+.Text,
+.Content {
   display: block;
 }
 
--- a/src/components/ActionList/components/Item/Item.tsx
+++ b/src/components/ActionList/components/Item/Item.tsx
@@ -62,7 +62,7 @@ export function Item({
 
   const contentMarkup = helpText ? (
     <span className={styles.ContentBlock}>
-      <span>{contentText}</span>
+      <span className={styles.ContentBlockInner}>{contentText}</span>
       <TextStyle variation="subdued">{helpText}</TextStyle>
     </span>
   ) : (
--- a/src/components/ActionList/components/Item/Item.tsx
+++ b/src/components/ActionList/components/Item/Item.tsx
@@ -8,6 +8,7 @@ import {UnstyledLink} from '../../../UnstyledLink';
 import {Badge} from '../../../Badge';
 import {TextStyle} from '../../../TextStyle';
 import styles from '../../ActionList.scss';
+import {handleMouseUpByBlurring} from '../../../../utilities/focus';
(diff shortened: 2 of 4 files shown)


src/components/Popover/Popover.scss
```diff
@@ -17,6 +17,7 @@ $content-max-width: rem(400px);
   will-change: opacity;
   opacity: 0;
   transition: opacity duration() easing(in);
+  transform: translate3d(0, 0, 0);
```
kyledurand, line 20: What do you need this for?

src/components/ActionList/components/Item/Item.tsx
```diff
@@ -81,12 +82,12 @@ export function Item({
   const textMarkup = <div className={styles.Text}>{contentMarkup}</div>;
 
   const contentElement = (
-    <div className={styles.Content}>
+    <span className={styles.Content}>
```
kyledurand, line 85: You can't have block level elements inside of inline elements. https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elements#conceptual_differences

kyledurand (changes requested):
A couple odd things I think need to be left out of this PR

src/components/ActionList/components/Item/Item.tsx
dleroux: That’s the reason we changed it to a span.

src/components/Popover/Popover.scss
dleroux: Was this left in from our different attempts?

src/components/ActionList/components/Item/Item.tsx
kyledurand: Oh, but `textMarkup` from above is still a `div`?

src/components/Popover/Popover.scss
skrabakl-shopify: Yes this was left in for the debugging of the popover issue - I'll remove that

src/components/ActionList/components/Item/Item.tsx
skrabakl-shopify: I went in and fix all nested divs for this

src/components/ActionList/components/Item/Item.tsx
```diff
@@ -40,30 +41,30 @@ export function Item({
...
     );
   } else if (image) {
     prefixMarkup = (
-      <div
+      <span
         role="presentation"
         className={styles.Prefix}
         style={{backgroundImage: `url(${image}`}}
-      />
+      ></span>
```
kyledurand, line 57:
```suggestion
      />
```
Tests are failing because this can be a self closing tag

kyledurand (commented):
Getting closer!

Because these elements have changed to inline (thanks for doing this btw) it breaks the layout of action list items with helper text

image

Should look like

image

So we'll need to add some css i.e. `display: block` to these spans that were changed in order to preserve this layout.

src/components/ActionList/ActionList.scss
```diff
@@ -137,6 +137,11 @@ $item-vertical-padding: ($item-min-height - line-height(body)) / 2;
   margin-left: spacing();
 }
 
+.ContentBlock,
+.ContentBlock span {
```
kyledurand, line 141:
```suggestion
.ContentBlock,
.Prefix,
.Text,
.Content {
```

Targeting html elements can be fragile. Let's be explicit to make sure this doesn't break in the future

skrabakl-shopify: Ok so this broke it again :D I'll check what's happening but I'm guessing the class specificity &&|| the missing inner spanon .ContentBlock  is breaking flex or something

skrabakl-shopify: So actually adding .Prefix, .Text, .Content broke it since those already had defined flex display. Dropping those from the scss fixed it, and instead of using HTML tag span for the ContentBlock I added one more class `ContentBlockInner`

dleroux (approved):
Thanks for doing this and following through @skrabakl-shopify!

kyledurand (approved):
🎉 This looks great, thank you @skrabakl-shopify !

skrabakl-shopify:
/shrimpit

Awesome!

PR #4037, opened 2021-03-05
DropZone - show plural message when allowMultiple is true

sylvhama:

Fixes https://github.com/Shopify/polaris-react/issues/3586

I have created a singular and a plural sentence for each call to action in DropZone.
I have based my changes on [this thread in Slack](https://shopify.slack.com/archives/C7TJQLVC7/p1614733227142400).
TLDR: We don't know how many files should be upload, so we don't have any `count` value. Using a random `count` value would be problematic in languages that have several plural rules. So we hope each translator will find a general sentence in both cases.

Also I could not use [scope](https://github.com/Shopify/quilt/tree/main/packages/react-i18n#dynamic-translation-keys), it seems not available in Polaris 🤔 

### How to 🎩

Run `yarn dev` and browse DropZone stories.

### 🎩 checklist

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -96,17 +96,28 @@
       "secondaryAction": "Continue editing"
     },
     "DropZone": {
-      "overlayTextFile": "Drop file to upload",
-      "overlayTextImage": "Drop image to upload",
-      "errorOverlayTextFile": "File type is not valid",
-      "errorOverlayTextImage": "Image type is not valid",
-      "FileUpload": {
+      "single": {
+        "overlayTextFile": "Drop file to upload",
+        "overlayTextImage": "Drop image to upload",
         "actionTitleFile": "Add file",
         "actionTitleImage": "Add image",
+        "actionHintFile": "or drop file to upload",
+        "actionHintImage": "or drop image to upload",
+        "labelFile": "Upload file",
+        "labelImage": "Upload image"
+      },
+      "allowMultiple": {
+        "overlayTextFile": "Drop files to upload",
+        "overlayTextImage": "Drop images to upload",
+        "actionTitleFile": "Add files",
+        "actionTitleImage": "Add images",
         "actionHintFile": "or drop files to upload",
         "actionHintImage": "or drop images to upload",
-        "label": "Upload file"
-      }
+        "labelFile": "Upload files",
+        "labelImage": "Upload images"
+      },
+      "errorOverlayTextFile": "File type is not valid",
+      "errorOverlayTextImage": "Image type is not valid"
     },
     "EmptySearchResult": {
       "altText": "Empty search results"
--- a/src/components/DropZone/DropZone.tsx
+++ b/src/components/DropZone/DropZone.tsx
@@ -35,7 +35,7 @@ import {
 } from './utils';
 import styles from './DropZone.scss';
 
-export type Type = 'file' | 'image';
+export type DropZoneFileType = 'file' | 'image';
 
 export interface DropZoneProps {
   /** Label for the file input */
@@ -52,7 +52,7 @@ export interface DropZoneProps {
    * Whether is a file or an image
    * @default 'file'
    */
-  type?: Type;
+  type?: DropZoneFileType;
   /** Sets an active state */
   active?: boolean;
   /** Sets an error state */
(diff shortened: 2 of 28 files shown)


src/components/DropZone/utils/index.ts
```diff
@@ -57,3 +57,9 @@ function isChangeEvent(
 ): event is React.ChangeEvent<HTMLInputElement> {
   return Object.prototype.hasOwnProperty.call(event, 'target');
 }
+
+export const defaultAllowMultiple = true;
+
+export function pluralSuffix(isPural: boolean) {
+  return isPural ? 's' : '';
+}
```
kyledurand, line 65: This would break in non-latin languages wouldn't it?

kyledurand (changes requested):
Thanks @sylvhama! 

Can you update the translations to use one / other keys? I'm also concerned about the suffix function but otherwise this is great

locales/en.json
```diff
@@ -97,15 +97,24 @@
     },
     "DropZone": {
```
husshegz, line 98: Would it make it easier to organize this as such:

```
    "DropZone": {
      "FileUpload": {
        "allowMultiple": {
          "file": {
            "overlayText": "Drop files to upload"
          },
          "image": {
            "overlayText": "Drop images to upload"
          }
        },
        "single": {
          "file": {
            "overlayText": "Drop file to upload"
          },
          "image": {
            "overlayText": "Drop image to upload"
          }
        }
      }
    }
```

then do something like this at the top of the file `DropZone`

```
  const translationNameSpace = 'Polaris.DropZone';
  const allowMultipleKey = allowMultiple ? 'allowMultiple' : 'single';
  const fileType = capitalize(type);

  const translationKey = `${translationNameSpace}.${allowMultipleKey}.${fileType}`;

  const overlayTextWithDefault =
    overlayText === undefined
      ? i18n.translate(`${translationKey}.overlayText`)
      : overlayText;
```

sylvhama: I have no strong opinion about it, my goal was to preserve existing translations as much as I can.

paulisloud:
Bumping this, as it is important to a project of ours. The only conflict is in UNRELEASED.md, so should be an easy resolution. Could this be merged for the next release?

src/components/DropZone/utils/index.ts
husshegz, line 65: I think this could get merged if we make the change to remove the letter `s` appending and organize the translations differently. Appending the letter `s` would break in some languages. @sylvhama do you have time to tackle this ? I can try to push some changes to have this fixed :)

sylvhama:
Sorry for the noise, I had rebased on `master`. I changed the JSON structure, we no have an `allowMultiple` and a `single` group.

husshegz (approved):
Code ✅ 
Tested it with different `type` and `allowMultiple` and works great ✅

src/components/DropZone/DropZone.tsx
```diff
@@ -27,10 +27,15 @@ import {useToggle} from '../../utilities/use-toggle';
...
+import {
+  fileAccepted,
+  getDataTransferFiles,
+  defaultAllowMultiple,
+  createAllowMultipleKey,
+} from './utils';
 import styles from './DropZone.scss';
 
-type Type = 'file' | 'image';
+export type Type = 'file' | 'image';
```
kyledurand, line 38: If you're going to export this I think you may need a more specific name

kyledurand (approved):
This looks great 🙌  

Just one comment around naming being more explicit:

`export type Type = 'file' | 'image';`

src/components/DropZone/DropZone.tsx
husshegz: I like `FileType` 
Other suggestions 🤔 
`ResourceType` - `MediaType` - `AssetType` - `DocumentType` - `UploadFileType`

sylvhama: Let's go with `DropZoneFileType`

sylvhama:
I will merge this on Monday morning (EST).

sylvhama:
> I will merge this on Monday morning (EST).

Actually no, I will wait for all translations first so this might take few days.

PR #4052, opened 2021-03-09
[Avatar] Expose `onError` hook

whizkydee:

Due to surface area constraints, in our app, we use custom a user menu — the activator is a standalone avatar within a button which toggles a `Popover`. The popover includes an avatar and the full name of the user in question, as opposed to displaying the full name alongside the avatar directly on the top bar.

For Identity users that haven't uploaded an avatar, we attempt to grab an image from gravatar based on their `uuid` which eventually resolves to their email — that request can either return the image or `404` if the account doesn't exist.

The problem with that is each time the user menu activator is clicked, the popover is re-rendered which results in extraneous requests to the same url that previously failed in an attempt to retrieve the gravatar.

This adds an `onError` hook to the `Avatar` component, which provides consumers with an avenue to make decisions when the HTTP request for an avatar fails. In our case, we'll simply prevent subsequent requests to the same url.

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page, Avatar} from '../src';

  return (
    <Page title="Playground">
      <Avatar source="invalidUrl" onError={() => console.log('bummer')} />
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -31,6 +31,8 @@ export interface AvatarProps {
   customer?: boolean;
   /** URL of the avatar image which falls back to initials if the image fails to load */
   source?: string;
+  /** Callback fired when the image fails to load  */
+  onError?(): void;
   /** Accessible label for the avatar image */
   accessibilityLabel?: string;
 }
@@ -38,6 +40,7 @@ export interface AvatarProps {
 export function Avatar({
   name,
   source,
+  onError,
   initials,
   customer,
   size = 'medium',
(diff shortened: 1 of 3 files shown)


src/components/Avatar/Avatar.tsx
```diff
@@ -31,13 +31,16 @@ export interface AvatarProps {
   customer?: boolean;
```
whizkydee, line 31: Thought — Should this be the default behaviour of the `Avatar` component?

alex-page:
@whizkydee I am wondering if the error could be handled in your app before using the component? It feels like the wrong level of abstraction. Interested to know what blockers exist there?

whizkydee:
@alex-page The problem is we have know way of knowing that the image errored, so we can't handle it. This hook is meant to inform us.

alex-page (approved):
Happy to ship this, but I still think this check should happen before rendering the `Avatar` component.

kyledurand (approved):
I think this makes sense to add but also agree with Alex that the best UX would be to determine if the image request 404'd before trying to render the Avatar. 

Sounds like your situation is a bit different with the re-rendering on click though so I think if this works for your use case then 🚢

PR #4053, opened 2021-03-10
[TextField] Replace autocomplete off with nope

dfmcphee:

Currently we use the attribute `autocomplete="off"` on the `<TextField>` component when passing in `autoComplete={false}`.

This causes issues in Chrome because of this bug: https://bugs.chromium.org/p/chromium/issues/detail?id=468153

This changes the attribute to be `autocomplete="nope"` which looks strange but does work across browsers.


gcornhill (approved):
Thank you!

alex-page (approved):

yourpalsonja:
Checked and works for me in Chrome 88. Just wanted to surface this SO answer, [it seems that Chromium "fixed" this and then broke it again.](https://stackoverflow.com/questions/57367813/2019-chrome-76-approach-to-autocomplete-off)

scripts/accessibility-check.js
```diff
@@ -1,107 +1,21 @@
...
-  return violations
-    .map((fail) => {
-      const message = fail.nodes
-        .map((error) => formatFailure(error))
-        .join('\n');
-      return `${chalk.red(fail.impact)} => ${url}\n${message}`;
-    })
-    .join('\n');
-};
+const {storybookA11yTest} = require('@shopify/storybook-a11y-test');
```
alex-page, line 2: Unblocked by https://github.com/Shopify/quilt/pull/1764

alex-page (approved):

utkarshsaxena93:
This change is breaking auto complete in the theme editor. Is there a recommendation to fix it? https://screenshot.click/21-04-yqnd4-01tux.mp4 

`autocomplete: off` works in Chrome for me Version 89.0.4389.114 (Official Build) (x86_64)

dfmcphee:
> This change is breaking auto complete in the theme editor. Is there a recommended way to fix it? https://screenshot.click/21-04-yqnd4-01tux.mp4
> 
> `autocomplete: off` works correctly for me in Chrome Version 89.0.4389.114 (Official Build) (x86_64)

We saw one other case of this recently. Chrome seems to handle these in very strange ways where sometimes it can depend on the form around it and names and ids on the inputs as well. A work around for now is to use `autoComplete="OFF"` on the text field.

alex-page:
If you use `OFF` it doesn't hit that `if` statement.

utkarshsaxena93:
Yes please. it will be cumbersome to update all text fields in the theme editor (and other places where autocomplete may not have broken similar to how it did for the editor)

PR #4057, opened 2021-03-12
[Setting toggle] Replace instances of enable/disable

jjgali:
For the setting toggle component, replaced recommendation to use `enable`/`disable` with `activate`/`deactivate`.

Addresses issue https://github.com/Shopify/polaris-styleguide/issues/3872 

- Replaced `enable`/`disable` with `activate`/`deactivate`
- Replaced `enabled`/`disabled` with `activated`/`deactivated`

_Note_: The changes mentioned apply only to content guidelines. There are instances within the code base (like props) where we have not addressed this because it is a breaking change.

### DIFF
--- a/src/components/SettingToggle/README.md
+++ b/src/components/SettingToggle/README.md
@@ -27,10 +27,10 @@ on or off.
 
 Settings toggles should:
 
-- Include different body content for the enabled and disabled states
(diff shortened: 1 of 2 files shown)


kyledurand (approved):
Nice, thanks @jjgali 👍

alex-page (approved):

PR #4058, opened 2021-03-15
[Link] Fixes inconsistent "external" icon

mmargel:

This PR updates the `Link` component to show the external icon when the `children` prop is a number. Currently, it only shows the icon if `children` is a string. This leads to inconsistent behaviour, where some numerical links will show the icon and others will not.

The following code would show the external icon, since the `children` component is a string, even though it's a number.
```js
<Link url="..." external>3</Link>
```

On the other hand, this code would not show the external icon, despite it rendering an identical link (icon notwithstanding):
```js
<Link url="..." external>{3}</Link>
```

This is something that could occur normally when generating links with dynamic content (e.g. adding a link to a table, where some links are internal and some are external).

This PR adds `|| typeof children === 'number'` to the check that determines whether to append the external icon. Otherwise, the functionality of this component is unchanged (specifically, if the child is a react component, no icon will be rendered).

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist


alex-page:
@mmargel this is very odd that we have this rule in our link. I wonder if we could remove it completely. I don't know why we need to check what gets passed into a link.

.changeset/pretty-knives-rhyme.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/polaris': minor
+---
+
+Link add external link for non component based children
```
lgriffee, line 5:
```suggestion
Add external link icon for non component based children
```

lgriffee (approved):

sam-b-rose:
> this is very odd that we have this rule in our link. I wonder if we could remove it completely. I don't know why we need to check what gets passed into a link.

I agree with @alex-page here. Seems like we shouldn't be concerned that the `children` type is given that `external` is explicitly set.

I would be in favor of removing the `&& typeof children !== '...'` entirely 👍

polaris-react/src/components/Link/Link.tsx
```diff
@@ -40,8 +40,7 @@ export function Link({
 }: LinkProps) {
   const i18n = useI18n();
   let childrenMarkup = children;
-
-  if (external && typeof children === 'string') {
+  if (external && typeof children !== 'object') {
```
sam-b-rose, line 43: Can we just check `external` here and remove `&& typeof children !== 'object'`?

sam-b-rose (changes requested):

polaris-react/src/components/Link/Link.tsx
alex-page: I thought about this. The reason we do this is because if a react node is passed in we don't know what content will render or if we should append an external icon. It's a bit weird... 

I would be happy removing this check in a breaking change as it could significantly change some user interfaces.

PR #4070, opened 2021-03-22
[Card] Add overflow handling

lhoffbeck:

Our Card component currently allows child content to overflow the `border-radius`. This is usually obscured since we regularly use `Sectioned` cards (which apply their own `border-radius`) but for non-sectioned cards this may cause square borders to appear.

** 👀  Trivial example 👀  **
([link](https://codesandbox.io/s/cocky-jackson-u8uu3))

Screen Shot 2021-03-22 at 10 08 19 AM

** 👀  Real life example 👀  **
In the polaris documentation for the IndexTable component, `IndexTable with bulk actions`:

Screen Shot 2021-03-22 at 9 58 59 AM

Applies an `overflow: hidden` where the card has a defined `border-radius`.

## 

### How to 🎩

👀  [CODESANDBOX LINK here](https://codesandbox.io/s/cocky-jackson-u8uu3)

### 🎩 checklist

### DIFF
--- a/src/components/Card/Card.scss
+++ b/src/components/Card/Card.scss
@@ -11,6 +11,7 @@
 
   @include page-content-when-not-fully-condensed {
     border-radius: var(--p-border-radius-wide, border-radius());
+    overflow: hidden;
   }
 }
 
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -9,6 +9,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Bug fixes
 
 - Update `IndexTable` to select row when clicked ([#4062](https://github.com/Shopify/polaris-react/issues/4062))
+- Hide `Card` border overflow when not sectioned ([#4070](https://github.com/Shopify/polaris-react/issues/4070))
 
 ### Documentation
 


AndrewMusgrave:
> I'm fairly sure this breaks things but I don't remember just what. Adding a few more people to the review who might be able to share some context

`overflow: hidden` was removed because it break `position: sticky`. I thought the pull request fixed the border-radius issue though 🤔

AndrewMusgrave:
Is the content within a section?

lhoffbeck:
> overflow: hidden was removed because it break position: sticky
aww man :( 

> Is the content within a section?
@AndrewMusgrave It's not. If we'd prefer to leave this `overflow` off the `Card` I can just fix the usages

src/components/Card/Card.scss
```diff
@@ -11,6 +11,7 @@
 
   @include page-content-when-not-fully-condensed {
     border-radius: var(--p-border-radius-wide, border-radius());
+    overflow: hidden;
```
alex-page, line 14:
```suggestion
    overflow: hidden;
```
As of right not it's recommended for the child content to have the border-radius or `overflow: hidden;`, shipping this would potentially break multiple card layouts in the admin.

lhoffbeck: Alright, will close. Thanks Alex :)

nickpresta:
FWIW, we're encountering a similar issue in an experience we're building:

https://github.com/Shopify/web/pull/40361#issuecomment-804929619

The TL;DR is that we have a horizontal scrolling container inside a Polaris `Card`, rendering a bunch of `AnalyticsCard` components. There is a hover state on the `AnalyticsCard` which results in the background "spilling" out of the Polaris `Card`.

Adding `overflow: hidden` to the parent container of the `AnalyticsCard` doesn't do anything since the border-radius is applied to the Polaris `Card`. Of course, we could copy the Polaris `Card` styles to our own container and essentially "copy" those styles but that seems just as bad as adding `overflow: hidden` to our particular component.

PR #4090, opened 2021-03-29
FormLayout - refactor how negative margin top is cancelled

sylvhama:

When a `FormLayout` is wrapped in another element such as a `Form`, its negative `margin-top` might cause issues e.g. [this example](https://codesandbox.io/s/mystifying-kirch-mtrl9?file=/App.js) with a `TextContainer`.

I am refactoring the way we "cancel" the `margin-top` of the very first `Item`. There is no more negative value.

### How to 🎩

<details>

```jsx
import React from 'react';

import {Page, TextContainer, Form, FormLayout, TextField} from '../src';

  return (
    <Page title="Playground">
      <TextContainer>
        <p>
          Without negative margin top we should see space between me and the
          form.
        </p>
        <Form onSubmit={() => {}}>
          <FormLayout>
            <TextField label="Store name" onChange={() => {}} />
            <TextField type="email" label="Account email" onChange={() => {}} />
          </FormLayout>
        </Form>
      </TextContainer>
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/FormLayout/FormLayout.scss
+++ b/src/components/FormLayout/FormLayout.scss
@@ -26,7 +26,7 @@ $item-min-size: rem(220px);
   margin-left: spacing(loose);
   max-width: calc(100% - #{spacing(loose)});
 
-  &:not(.First) {
+  & + & {
     margin-top: spacing();
   }
 
--- a/src/components/FormLayout/FormLayout.scss
+++ b/src/components/FormLayout/FormLayout.scss
@@ -3,7 +3,6 @@
 $item-min-size: rem(220px);
 
 .FormLayout {
-  margin-top: -1 * spacing();
   margin-left: -1 * spacing(loose);
 }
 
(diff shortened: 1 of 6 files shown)


src/components/FormLayout/FormLayout.scss
```diff
@@ -24,10 +23,13 @@ $item-min-size: rem(220px);
 
 .Item {
   flex: 1 1 $item-min-size;
-  margin-top: spacing();
   margin-left: spacing(loose);
   max-width: calc(100% - #{spacing(loose)});
 
+  &:not(.First) {
+    margin-top: spacing();
+  }
```
alex-page, line 31: Could we do...
```suggestion
  & + .Item {
    margin-top: spacing();
  }
```

sylvhama: Will try that! it would be much cleaner, thanks!

sylvhama:
When using `FormLayout.Group` we have a margin issue (see percy report). I guess there is no way to simply fix that, I am gonna close this PR for now. Maybe in a far future we could use [flexbox gap)(https://web.dev/flexbox-gap/) instead of margin.

PR #4111, opened 2021-04-08
[Button] Fix button types

pedrodurek:

I'm working on a task in `Shopify/web` where I call the preventDefault function (event handler), but it's not returning the appropriate types for Button callbacks.

Fixing button types (callbacks)

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -38,17 +38,17 @@ export interface BaseButton {
   /** Indicates the ID of the element that describes the button */
   ariaDescribedBy?: string;
   /** Callback when clicked */
-  onClick?(): void;
+  onClick?: React.MouseEventHandler<HTMLButtonElement>;
   /** Callback when button becomes focussed */
-  onFocus?(): void;
+  onFocus?: React.FocusEventHandler<HTMLButtonElement>;
   /** Callback when focus leaves button */
-  onBlur?(): void;
+  onBlur?: React.FocusEventHandler<HTMLButtonElement>;
   /** Callback when a keypress event is registered on the button */
-  onKeyPress?(event: React.KeyboardEvent<HTMLButtonElement>): void;
+  onKeyPress?: React.KeyboardEventHandler<HTMLButtonElement>;
   /** Callback when a keyup event is registered on the button */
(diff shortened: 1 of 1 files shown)


alololox (approved):

BPScott:
Hi. Have an old issue: https://github.com/Shopify/polaris-react/issues/521

In particular note [@lemonmade's comment](https://github.com/Shopify/polaris-react/issues/521#issuecomment-592656727) that points out that while this is good for use on web it is fundamentally incompatible in other worlds like Argo and exposing these dom events, and potentially relying on them is going to be a big problem for Argo.

clauderic:
Hey @BPScott, 

I think it'd be worthwhile to re-visit this discussion. @lemonmade's comments around how this is fundamentally incompatible with Argo make sense and I generally share those concerns. I also agree we should be discouraging consumers from relying on these events and opening issues instead for the things that are currently hard to implement.

With that said, I fundamentally disagree with the current strategy of type obfuscation. It's one thing not to document the fact that events are available in order to discourage use; on the other hand, type obfuscation has real consequences with regards to the type safety of consuming applications. 

This recently caused a bug in production for us on the Online Store because we passed a handler that accepted optional arguments to a `Button` handler, which you would assume is safe to do based on the type declaration of the `Button` props, but is not safe in practice.

```tsx
function foo(optionalArg?: boolean) {
  if (optionalArg) {
    throw new Error();
  }
}

<Button onClick={foo} /> // Typed as () => void, but is in fact (event: React.SyntheticEvent) => void

// On click, error is thrown because `foo` is called with `event`
```

Were these handlers properly typed, TypeScript would have caught this issue. If the event is there, the types should reflect this, regardless of whether or not we want to discourage use.

BPScott:
> I think it'd be worthwhile to re-visit this discussion

Happy to rediscuss and provide context, though I'm not on the polaris team these days so the final decision will fall to @Shopify/polaris-team. :)

PR #4114, opened 2021-04-12
[Popover] center full width popovers on small viewports

spencercanner:
Full width popovers have a max width of `100vw - 3.2rem`, which results in a popover that is smaller than the viewport on narrow screens. Since no horizontal margins are applied, the popover defaults to being left aligned on the screen. 

Narrow screen with an open popover that is left aligned

Given the popover is smaller than the viewport width and has rounded corners, it should be centered aligned on the screen regardless of the preferred alignment. 

* Add in auto horizontal margins to full width popovers to ensure that it is centered on smaller viewports 

Narrow screen with an open popover that is center aligned

## 

### How to 🎩

* Verify that full width popovers still take up the full width on larger viewports (positioned above and below the activator)
* Verify that the popover is centered above/below the activator when its width is smaller

<details>

```jsx

import {Button, Page, Popover} from '../src';

  const [popover1Active, setPopover1Active] = useState(false);
  const [popover2Active, setPopover2Active] = useState(false);

  const togglePopover1 = () => {
    setPopover1Active(!popover1Active);
  };

  const togglePopover2 = () => {
    setPopover2Active(!popover2Active);
  };

  const activatorBelow = (
    <Button onClick={togglePopover1}>Button with popover below</Button>
  );

  const activatorAbove = (
    <Button onClick={togglePopover2}>Button with popover above</Button>
  );

  return (
    <Page title="Playground">
      <Popover
        active={popover1Active}
        onClose={togglePopover1}
        fullWidth
        activator={activatorBelow}
        sectioned
      >
        <p>some text</p>
      </Popover>
      <br />
      <br />
      <Popover
        active={popover2Active}
        onClose={togglePopover2}
        fullWidth
        preferredPosition="above"
        activator={activatorAbove}
        sectioned
      >
        <p>some text</p>
      </Popover>
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -12,6 +12,8 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 
 ### Bug fixes
 
+- Center full width `Popover` on small viewports ([#4114](https://github.com/Shopify/polaris-react/pull/4114))
+
 ### Documentation
 
 ### Development workflow
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -12,7 +12,7 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 
 ### Bug fixes
 
-- Center full width `Popover` on small viewports ([#4114](https://github.com/Shopify/polaris-react/pull/4114))
+- Centered full width `Popover` on small viewports ([#4114](https://github.com/Shopify/polaris-react/pull/4114))
 
 ### Documentation
 
--- a/src/components/Popover/Popover.scss
+++ b/src/components/Popover/Popover.scss
@@ -42,7 +42,7 @@ $vertical-motion-offset: rem(-5px);
 }
 
 .fullWidth {
-  margin: $visible-portion-of-arrow auto 0;
+  margin: $visible-portion-of-arrow auto 0 auto;
 
   .Content {
     max-width: none;
@@ -53,7 +53,7 @@ $vertical-motion-offset: rem(-5px);
   margin: spacing() spacing(tight) $visible-portion-of-arrow;
 
   &.fullWidth {
-    margin: 0 auto $visible-portion-of-arrow;
+    margin: 0 auto $visible-portion-of-arrow auto;
   }
 }
 
--- a/src/components/Popover/Popover.scss
+++ b/src/components/Popover/Popover.scss
@@ -42,8 +42,7 @@ $vertical-motion-offset: rem(-5px);
 }
 
 .fullWidth {
-  margin: 0;
-  margin-top: $visible-portion-of-arrow;
+  margin: $visible-portion-of-arrow auto 0;
 
   .Content {
     max-width: none;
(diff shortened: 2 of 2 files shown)


kyledurand (commented):
Code looks good but I'm wondering if we really need this to be centered, especially if it overrides the alignment prop. Can you ping a designer on this change?

spencercanner:
we made this change for a component we use in channel apps, which leverages a full width popover on smaller viewports. since the popover has a border radius, we wanted to give a bit of margin between the edge of the screen and the start of the popover
| current | desired |
|---|---|
|  |  |

since the `preferredAlignment` prop doesn't currently have an effect on the alignment for full width popovers, we opted to force center alignment on smaller viewports so there would be some margin on both sides. this change shouldn't have an impact on the alignment for non-full width popovers.  

chatted with my team's designer (@stephentjoa) and we're happy to align with the Polaris pattern for popovers, so if the left alignment is expected then we can close this. but happy to get this ready to merge otherwise 

cc @sarahill

sarahill:
I'm ok with the center alignment but agree with Kyle it does feel a little odd if it overrides the alignment prop. That being said I recognize you mentioned the alignment prop doesn't affect full-width popovers.

One question I have: do you know if it gets weird on mid breakpoints?

If this is the desired affect and it doesn't negatively affect other areas using the component I'm ok with the change

spencercanner:
The alignment prop doesn't have an impact on full width popovers, and this change will only apply to full width popovers, so there shouldn't be any negative impact on existing any popovers. The only change that should happen is shown in the gifs below

Before: 
old-popover

After: 
new-popover

sarahill:
Thanks for that context! I'm cool with that 🚢

UNRELEASED.md
```diff
@@ -12,6 +12,8 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 
 ### Bug fixes
 
+- Center full width `Popover` on small viewports ([#4114](https://github.com/Shopify/polaris-react/pull/4114))
```
kyledurand:
```suggestion
- Centered full width `Popover` on small viewports ([#4114](https://github.com/Shopify/polaris-react/pull/4114))
```

src/components/Popover/Popover.scss
```diff
@@ -42,8 +42,7 @@ $vertical-motion-offset: rem(-5px);
 }
 
 .fullWidth {
-  margin: 0;
-  margin-top: $visible-portion-of-arrow;
+  margin: $visible-portion-of-arrow auto 0;
```
kyledurand: CSS shorthands are always harder to read than long form. Using the three property shorthand instead of 4 is especially difficult. Can you add the fourth value so it's less 🤯 ?

src/components/Popover/Popover.scss
```diff
@@ -54,7 +53,7 @@ $vertical-motion-offset: rem(-5px);
   margin: spacing() spacing(tight) $visible-portion-of-arrow;
 
   &.fullWidth {
-    margin: 0 0 $visible-portion-of-arrow;
+    margin: 0 auto $visible-portion-of-arrow;
```
kyledurand: Same here please 😄

kyledurand (approved):
Looks good to me just left a couple of non-blocking comments around conventions and past tense in UNRELEASED.

PR #4123, opened 2021-04-16
[DropZone] Make drag-over state match design

### LINKED ISSUE #4084: Update Dropzone drag-over state to intended styling?
## Issue summary

The visual design for the Dropzone drag-over state ([see Figma](https://polaris.shopify.com/resources/resources#section-figma)) is different from what’s implemented in Polaris React. This shows the discrepancy for the small Dropzone size, but similar inconsistency exists for the other sizes:

Screen Shot 2021-03-24 at 10 50 45 AM

There’s also a weird border effect on drag that looks like 2 dotted borders overlapping, but not lining up:
Screen Shot 2021-03-24 at 10 52 05 AM

## Expected behavior

The Figma UI kit shows the dragover state like this, for the 3 sizes:
Screen Shot 2021-03-24 at 10 55 40 AM

1. Visit https://polaris.shopify.com/components/actions/drop-zone
1. Drag a file from a computer’s file system over the dropzone, without releasing the mouse

nicolleromero:

Fixes #4084

- Fixes border alignment and updates color of `DropZone` to match current design.
- Adds `interactive` color for `Icon` component.
- Fixes minor bug for `NestedDropZoneExample`.

---

Large `DropZone` drag
**BEFORE**
<details>
<p align="center">
Large DropZone drag demo
</p>
</details>

**AFTER**
<details>
<p align="center">
Large DropZone drag demo
</p>
</details>

---
Nested `DropZone` drag
(Fixes a bug where the inner nested drop zone was not available during a dragging event.)
**BEFORE**
<details>
<p align="center">
Nested DropZone drag demo
</p>
</details>

**AFTER**
<details>
<p align="center">
Nested DropZone drag demo
</p>
</details>

---

Small `DropZone` drag
**BEFORE**
<details>
<p align="center">
Small DropZone drag demo
</p>
</details>

**AFTER**
<details>
<p align="center">
Small DropZone drag demo
</p>
</details>


src/components/DropZone/DropZone.scss
```diff
@@ -132,13 +140,11 @@ $dropzone-stacking-order: (
...
-  border-radius: calc(
-    var(--p-border-radius-base) + #{border-width(thick)} + #{rem(1px)}
-  );
+  border: border-width(thick) $dropzone-border-style var(--p-interactive);
   text-align: center;
-  color: var(--p-text);
-  background-color: var(--p-surface-hovered);
+  color: var(--p-interactive);
+  background-color: var(--p-surface-selected);
+  pointer-events: none;
```
nicolleromero, line 147: Fixes a bug where the inner nested drop zone was not available during a dragging event.

src/components/Icon/Icon.tsx
```diff
@@ -11,6 +11,7 @@ type Color =
   | 'critical'
   | 'warning'
   | 'highlight'
+  | 'interactive'
```
nicolleromero, line 14: I noticed that #4112 adds `interactive` to `Icon` as well – will rebase after that's merged. cc @kyledurand

kyledurand: Oops totally forgot about that. I'll get it shipped asap

src/components/Icon/Icon.scss
```diff
@@ -66,6 +66,14 @@
   }
 }
 
+.colorInteractive {
+  @include recolor-icon(var(--p-interactive));
+
+  &::before {
+    background-color: var(--p-surface-selected);
+  }
```
kyledurand, line 74: We can 🔥 this unless you need it for some reason. We're trying to do away with the backdrop prop on Icon which is what this `before` background is for. If you do need this for whatever reason we should add a specific color in polaris-tokens based on interactive vs surface

kyledurand (approved):
Thank you for fixing this! Also the nested bug and the border shift. I just checked out your branch and this looks great!

Just pinging Alex on this to double check with the pointer-events change.

I think we'll have to manually approve the visual regression check as it doesn't work on branches outside of the repo

src/components/Icon/Icon.scss
nicolleromero: No problem, either way. 😊 I was matching the Figma design that has the `--p-surface-selected` as the arrow color, but it's a subtle difference.

nicolleromero: Ah, I see the layer is set to pass through in Figma and isn't needed. Removed – all set! 🔥

nicolleromero:
@kyledurand Please let me know if there is anything else I should address for this PR – thanks! 😊

kyledurand:
I think this is good to go! @alex-page can you take another look? Looks like there's a conflict in Unreleased.md that needs to be fixed up as well.

alex-page (approved):
Looks great thanks!

PR #4140, opened 2021-04-26
[EmptyState] Remove legacy styles from empty state within page context

szuchnik:

Empty state was not centering its components within page context, due to styles from before `new design language` still being applied. The repositioning of the image seems to be an echo of the time when EmptyState was laid out in two columns.

The image above is the previous state, below you can see how this component will look after the update.

Fixes #0000 

Remove condition applying styles from old design language and associated style rules. Add documentation for page context usecase.

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist


kyledurand (commented):
Overall I think these changes are good, but one important thing to note is that all empty states should be nested within cards if they're going onto a page. This is the new pattern after the new design language shipped. 

I think the changes are worth making but I wouldn't include the story example because that's not a pattern we want to see start to emerge / deviate.

PR #4142, opened 2021-04-26
Add print styles to multiple components

matallo:

Added print styles to multiple components, based on previously existing print styles.

- Card
- Heading
- Layout
- Layout.Section
- Subheading
- TextStyle

- Adding print media query to aforementioned styles.

### How to 🎩

1. Navigate to a view where these components exist
2. Open Developer tools
1. Show print styles via "command + P" or devtools command "Emulate CSS print"

Alternatively you can print the page via browser menu.

### 🎩 checklist


src/components/Card/Card.scss
```diff
@@ -7,11 +7,19 @@
 
   + .Card {
     margin-top: spacing();
+
+    @media print {
+      margin-top: -1 * spacing(tight);
+    }
```
alex-page, line 13: Do you have an example of why the negative margin is needed for print?

matallo (commented):
🎩 and looks good.

Before (print styles)

torontontero myshopify com_admin_checkouts_20333427064854


After (print styles)

shop1 shopify e13q carlos-matallin us spin dev_admin_checkouts_1 (3)


I checked overall mobile experience and looks good too. The rest of testing shouldn't be affected by this change as it only adds support for `@print` styles.




I briefly discussed the possibility of moving all the styles to a `print.scss` file but in the long term it may be harder to maintain.

src/components/Card/Card.scss
alex-page: What about other areas of the admin where there is a list of cards, like the home page. Would that be negatively affected?

matallo: This is a good point. In that case it is not affected because, cards are wrapped with `CardList_Item` thus `Card + Card` won't apply. I checked in the code and here are not many places where that can happen and in those case it will look like the screenshot attached only when printing, otherwise it won't apply either.

src/components/Layout/Layout.scss
```diff
@@ -13,11 +13,30 @@ $relative-size: $primary-basis / $secondary-basis;
   align-items: flex-start;
   margin-top: -1 * spacing();
   margin-left: -1 * spacing(loose);
+
+  @media print {
+    padding: 0;
+    margin-top: -1 * spacing();
```
CSilivestru, line 19: Wouldn't this affect every layout component in use? Just confirming this broad change isn't too risky outside the scope we're working towards on our specific work that prompted this change :)

CSilivestru (commented):
LGTM but I do have one minor concern :). Once that's answered, I will approve!

CSilivestru:
Oh forgot to mention you will need an UNRELEASED entry :)

src/components/Layout/Layout.scss
matallo: Just tested and these are redundant, inherited from the Orders print styles. Actually [a coupe of lines above](https://github.com/Shopify/polaris-react/pull/4142/files#diff-5d39bc98db083e35f0f49470a0e013f5f91b3be3b890d0cde5af37718f85fcefR14) there are the same styles for the `.Layout` component without print styles. Well spotted!

alex-page (approved):
Thanks!

PR #4147, opened 2021-04-29
Remove codecov

alex-page:

Feels sensible 🤷‍♂️

Removes codecov dependency and the script to uploads coverage files.


tmlayton:
If the team doesn’t find it useful then makes sense, otherwise agree with @BPScott.

PR #4148, opened 2021-04-29
Remove integration test

alex-page:

The test often fails and teams tophat RC's in web all the time 

Removes integration test and related encrypted secrets


BPScott (approved):
Slightly sad to see this go Feels like a shame that we never really invested in making it as stable as possible. 

Removal changes look good for this repo.

Other things to decommission as part of this change:

- Logic to handle this in web. Searching for `POLARIS_ARTIFACT_FILENAME` and the polaris-react-system-integration-test.yml build would be good places to start
- This uploaded to a special polaris GCP bucket, that should be removed

PR #4153, opened 2021-04-29
[Modal] add noScroll prop

pt8o:

Needed to fix https://github.com/Shopify/core-issues/issues/23913

We're building a component that has its own sticky toolbar header, and then the main body content is wrapped in a `<ScrollableContainer>`, so that's how the user scrolls through the modal contents. The problem is, if the browser window is short enough, it'll squish the modal contents such that an outer scrollbar appears, which belongs to Polaris's `<Modal>`. So we have an outer container and an internal container, each of which will only partially scroll the modal contents 😬

You can see a video in the above issue.

`<Modal>` component now has the prop `noScroll` which when true wraps the body content in a `div` instead of the `Scrollable` component

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -9,6 +9,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Add `variableHeight` prop to `DropZone` so children control its height ([#4136](https://github.com/Shopify/polaris-react/pull/4136))
 - Add print styles to `Card`, `Heading`, `Layout`, `Layout.Section`, `Subheading`, `TextStyle` components ([#4142](https://github.com/Shopify/polaris-react/pull/4142))
 - Add `fullWidth` prop to `ColorPicker` so the color picker can take the full width ([#4152](https://github.com/Shopify/polaris-react/pull/4152))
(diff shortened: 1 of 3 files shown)


src/components/Scrollable/Scrollable.scss
```diff
@@ -9,6 +9,7 @@ $shadow-top: inset 0 $shadow-size $shadow-size (-1 * $shadow-size)
 .Scrollable {
   -webkit-overflow-scrolling: touch;
   position: relative;
+  overflow: hidden;
```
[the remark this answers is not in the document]
pt8o: Yeah good call, I was overthinking it. `Scrollable` has a `shadow` prop that makes some styles that I didn't want to repeat, but obviously if the content isn't scrollable then there's no need for a shadow.

src/components/Modal/Modal.scss
```diff
@@ -5,7 +5,7 @@ $small-width: rem(620px);
 .BodyWrapper {
   display: flex;
   flex-grow: 1;
-  overflow-x: hidden;
+  overflow: hidden;
```
WillsonSmith, line 8: why does this need to be changed?

pt8o: Whoops, leftover change from my prev idea

PR #4173, opened 2021-05-06
[KeypressListener] Prevent listener attaching/detaching on rerender

emma-boardman:

Prevents listeners being detached / reattached on every rerender 

- Includes react-hooks as a dependency 
  - `useIsomorphicEffect` is in v1.13.0
- assigns `handler` event to a ref 
- wraps `handleKeyEvent` in `useCallback`
- leverages `useIsomorphicLayoutEffect` to update `handlerRef`
- adds `useEffect` dependency array

### 🎩 checklist


emma-boardman:
Storybook error:

```
ERROR in ./node_modules/@shopify/react-hooks/build/esm/hooks/debounced.mjs 21:18-26
Can't import the named export 'useState' from non EcmaScript module (only default export is available)
 @ ./node_modules/@shopify/react-hooks/build/esm/hooks/index.mjs
 @ ./node_modules/@shopify/react-hooks/build/esm/index.mjs
 @ ./node_modules/@shopify/react-hooks/index.mjs
 @ ./src/components/KeypressListener/KeypressListener.tsx
 @ ./src/components/KeypressListener/index.ts
 @ ./src/components/index.ts
 @ ./src/index.ts
 @ ./.storybook/preview.js
 @ ./.storybook/preview.js-generated-config-entry.js
 @ multi ./node_modules/@storybook/core/dist/server/common/polyfills.js ./node_modules/@storybook/core/dist/server/preview/globals.js ./.storybook/storybook-init-framework-entry.js ./node_modules/@storybook/addon-docs/dist/frameworks/common/config.js-generated-other-entry.js ./node_modules/@storybook/addon-docs/dist/frameworks/react/config.js-generated-other-entry.js ./node_modules/@storybook/addon-actions/dist/preset/addDecorator.js-generated-other-entry.js ./node_modules/@storybook/addon-actions/dist/preset/addArgs.js-generated-other-entry.js ./node_modules/@storybook/addon-backgrounds/dist/preset/addDecorator.js-generated-other-entry.js ./node_modules/@storybook/addon-backgrounds/dist/preset/addParameter.js-generated-other-entry.js ./node_modules/@storybook/addon-a11y/dist/a11yRunner.js-generated-other-entry.js ./node_modules/@storybook/addon-a11y/dist/a11yHighlight.js-generated-other-entry.js ./node_modules/@storybook/addon-knobs/dist/preset/addDecorator.js-generated-other-entry.js ./.storybook/preview.js-generated-config-entry.js ./.storybook/generated-stories-entry.js ./node_modules/webpack-hot-middleware/client.js?reload=true&quiet=false&noInfo=true
```

Screenshot 2021-05-06 at 12 00 07

**Investigation notes:** 

- Can't replicate in OSUI Storybook if I bump the react-hooks dependency to 1.13.0 and include `useIsomorphicLayoutEffect`. 
   - OSUI Storybook is `v6.2.7`
   - Polaris Storybook is `v6.1.11`

src/components/KeypressListener/KeypressListener.tsx
```diff
@@ -15,18 +16,27 @@ export function KeypressListener({
...
-  };
+  const handlerRef = useRef(handler);
+
+  const handleKeyEvent = useCallback(
+    (event: KeyboardEvent) => {
+      if (event.keyCode === keyCode) {
+        handlerRef.current(event);
+      }
+    },
+    [handler, keyCode],
```
clauderic, line 27: `handler` should not be addd as a dependency of `handleKeyEvent`, we already keep track of the latest value in `handlerRef`; otherwise the listeners will be removed and re-added whenever it changes which could be every single render if `handler` is not memoized by the consumer.

```suggestion
    [keyCode],
```

src/components/KeypressListener/KeypressListener.tsx
```diff
@@ -15,18 +16,27 @@ export function KeypressListener({
...
+    handlerRef.current = handler;
+  }, [handler]);
 
   useEffect(() => {
     document.addEventListener(keyEvent, handleKeyEvent);
     return () => {
       document.removeEventListener(keyEvent, handleKeyEvent);
     };
-  });
+  }, [keyEvent, handleKeyEvent]);
```
clauderic, line 39: I think what you have here is a sufficient improvement over what we had before, though it's worth noting you could push this even further and avoid detaching and re-attaching listeners when the `keyEvent` changes by also keeping track of the latest value of `keyEvent`;

Something like:

```ts
const tracked = useRef({handler, keyEvent};

useIsomorphicLayoutEffect(() => {
  tracked.current = {handler, keyCode};
}, [handler, keyCode]);

const handleKeyEvent = useCallback(
    (event: KeyboardEvent) => {
      const {handler, keyCode} = tracked.current;
      
      if (event.keyCode === keyCode) {
        handler(event);
      }
    },
    [],
  );
```

.storybook/main.js
```diff
@@ -117,6 +117,12 @@ module.exports = {
       ...extraRules,
     ];
 
+    config.module.rules.push({
+      test: /\.mjs$/,
+      include: /node_modules/,
+      type: 'javascript/auto',
+    });
```
clauderic, line 124: Could you leave context on the PR as to how this relates to the change you made / why it was required? 🙏

clauderic (approved):

.storybook/main.js
emma-boardman: This corrected the storybook error documented here: https://github.com/Shopify/polaris-react/pull/4173#issuecomment-833435279

@kaelig are you able to elaborate on why we need to add this?

BPScott (dismissed):
That webpack fix is not the right one, as it any consuming app will need to make it too. Thus this is a breaking change for lots of consumers which is bad. 

I've put a possible solution inline.

src/components/KeypressListener/KeypressListener.tsx
```diff
@@ -15,18 +16,27 @@ export function KeypressListener({
...
+  const handleKeyEvent = useCallback(
+    (event: KeyboardEvent) => {
+      if (event.keyCode === keyCode) {
+        handlerRef.current(event);
+      }
+    },
+    [keyCode],
+  );
+
+  useIsomorphicLayoutEffect(() => {
```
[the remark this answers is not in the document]
clauderic: @BPScott the snippet you have above delegates the responsibility of ensuring that the `handler` prop is memoized to the consumer of `KeypressListener`, otherwise the `useEffect` callback will run on every single render if `handler` is not memoized (since it is a dependency of `handleKeyEvent`, which is a dependency of the effect).

Components such as `KeypressListener` shouldn't assume that their consumers will always remember to memoize the handlers they pas to them, and should optimize for their consumers instead. 

The pattern introduced in this PR ensures that listeners aren't attached and re-attached on every single render regardless of whether the consumer remembers to memoize their handlers

yarn.lock
```diff
@@ -2381,11 +2381,9 @@
...
 
 "@shopify/react-hooks@^1.11.2":
-  version "1.11.2"
-  resolved "https://registry.yarnpkg.com/@shopify/react-hooks/-/react-hooks-1.11.2.tgz#216421d52abc94c5c8e5b8821ff7800da6949110"
-  integrity sha512-KVbBxdOXcShtj+PJrfq5FW0T6DkqXkt3Q4YKf3NlGrBB0vprrN71Zs4MAyUhCqJp1B7qfKM8dgHcikmIxtDp4Q==
-  dependencies:
-    tslib "^1.14.1"
+  version "1.13.1"
+  resolved "https://registry.yarnpkg.com/@shopify/react-hooks/-/react-hooks-1.13.1.tgz#0756205f29e5320920f9dcaae8b057bb34f0e21f"
+  integrity sha512-i1/HKdpsuNXRi5MoZp2nciw0UtqAWEmV25Moq689X8GrfQLAGuxu97HV1V0TUx9PODlyfJsfuHyepEG+TRSjNA==
```
clauderic, line 2386: Should this change be reverted?

UNRELEASED.md
```diff
@@ -14,6 +14,10 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Added `requiredIndicator` prop to `Label`, `Labelled`, `Select` and `TextField` ([#4119](https://github.com/Shopify/polaris-react/pull/4119))
 - Add `small` prop to `Modal` so that width can be decreased to 380px ([#4177](https://github.com/Shopify/polaris-react/pull/4177))
 - Add `status` prop to `IndexTable.Row` to allow table rows to specify background colors([#4146](https://github.com/Shopify/polaris-react/pull/4146))
+- Added `useIsomorphicLayout` ([#4173](https://github.com/Shopify/polaris-react/pull/4173))
+- Prevent `KeypressListener` attaching/detaching on every render ([#4173](https://github.com/Shopify/polaris-react/pull/4173))
+- Added `useIsomorphicLayoutEffect` hook ([#4173](https://github.com/Shopify/polaris-react/pull/4173))
+- Prevented `KeypressListener` attaching/detaching on every render ([#4173](https://github.com/Shopify/polaris-react/pull/4173))
```
clauderic, line 20: Looks like the release notes are duplicated, bad rebase perhaps?

```suggestion
- Added `useIsomorphicLayoutEffect` hook ([#4173](https://github.com/Shopify/polaris-react/pull/4173))
- Optimized how often `KeypressListener` attaches/detaches event listeners on the document ([#4173](https://github.com/Shopify/polaris-react/pull/4173))
```

Also, `useIsomorphicLayoutEffect` isn't exported for consumers, but is mentioned in the changelog, which feels a bit odd. 

We already have `useIsomorphicLayoutEffect` in a lot of places, ideally it should just come from `@shopify/react-hooks` in the long-term. I don't think it should be exported from `@shopify/polaris` for that reason, consumers can import it directly from `@shopify/react-hooks` in the future rather than from `@shopify/polaris`. 

Which brings me to the following question: if this is something that probably shouldn't be exported, is it worth mentioning in the changelog?

emma-boardman: good catch, and valid point. updated to remove duplicates & `useIsomorphic` entry ✅

emma-boardman:
@BPScott I'm getting failures in the [git workflow](https://github.com/Shopify/polaris-react/actions/runs/836195877). Is that something to be worried about? It passes the PR CI builds 🤔 

> .github/workflows/size-limit.yml#L1`size` is not a valid event name

BPScott (commented):
Thanks for putting up with my slow/busyness. My original concerns have been addressed.

I've not ran the code to confirm behaviour, but I'm no longer blocking this :)

emma-boardman:
@kaelig 

I created some performance flame charts, on main and on this branch. However, I'm not sure how indicative they are of the memoization impact. In both instances `EventListener` is `< 0.1ms of 0.1ms`, even with multiple listeners rendering. The `rendered at` figures for the feature branch look smaller, for what that's worth. 

Does this help, or are there other ways I can measure the performance impact? 

|Main | Feature |
| --- | --- |
|main with multiple| branch with multiple|

kaelig:
Thank you for looking into this! The React devtools flamecharts only show components. You'll need to look at the Chrome DevTools flamecharts to see more granular data and you can also compare the number of resulting event listeners, possibly even looking at memory footprint impact when encountering a worst case scenario (very big listener function, lots of renders)... But that starts to get a bit overkill so you probably don't have to go that deep. 

Feel free to spend 30 more minutes at most on this — the PR looks good on principle and I thought flamecharts would have been a nice addition but they're not a requirement!

kaelig (approved):
I haven't been able to assess the real world impact of such performance improvements but I support this PR on principle :shipit:

emma-boardman:
> Thank you for looking into this! The React devtools flamecharts only show components. You'll need to look at the Chrome DevTools flamecharts to see more granular data and you can also compare the number of resulting event listeners, possibly even looking at memory footprint impact when encountering a worst case scenario (very big listener function, lots of renders)... But that starts to get a bit overkill so you probably don't have to go that deep.
> 
> Feel free to spend 30 more minutes at most on this — the PR looks good on principle and I thought flamecharts would have been a nice addition but they're not a requirement!

I ran some experiments in Chrome devtools, and I couldn't see any evidence for this having reduced the number of resulting event listeners 😬  However, I'm probably doing it wrong 😅   I'm going to merge this and dig more into effective performance testing at a later date! Thank you for the review and for bringing my attention to flamecharts and profiling 😁

PR #4189, opened 2021-05-06
Remove storybook file preview-head.html for percy

alex-page:

This file was used for Percy which was removed

Removing legacy files


BPScott (approved):
Follow up to #4167

PR #4191, opened 2021-05-06
Upgrade to node v12

alex-page:

When upgrading dependencies node v10 has reached EOL.


kyledurand (approved):
Looks good to me. Does it make sense to make a temp v7 label to easily reference things like this when we release it?

alex-page:
@kyledurand love that, this is also going into v7 release branch.

PR #4192, opened 2021-05-06
Add sizelimit for package size coverage

alex-page:

We removed shrink ray as it is focused on website/application size

Using size-limit to monitor the library size over time


BPScott (commented):
LGTM, check is failing because siize-limit isn't yet installed on main which is expected

PR #4204, opened 2021-05-13
[Filters] Add filter query read only prop

Leann88:

A SFN application is introducing filter limits to into its indices. As the search query is a filter, we would like to set search query as read only when the filter limit has been reached. The current disable prop is not sufficient as it will disable all filters, not just the search query. Disabling the TextField also prevents the field from being clearable.

- Adding a prop to readOnly the filter query field
- add test
- add docs

## 

### How to 🎩

- Run storybook

### 🎩 checklist


AndrewMusgrave (commented):
I helped author the code so I shouldn't review, but it LGTM to me ✅

PR #4229, opened 2021-05-26
[Button] Fix slim buttons height inconsistencies for mixed icons and text

itwasmattgregg:

When slim buttons are used in a button group with mixed icons and text the size of the buttons are different by 1px. This solution fixes this but I'm not sure it's the right solution. I'm open to any suggestions. Maybe it should be a change on ButtonGroup.

Examples:
Screen Shot 2021-05-26 at 1 20 34 PM

https://codesandbox.io/s/gifted-gates-c9zlp?file=/App.js

It seems the icon is just slightly larger at `2rem` than the text. Adjusting the padding on the top from 0.3rem to 0.2rem here seems to help.

### 🎩 checklist

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -339,7 +339,7 @@ $stacking-order: (
     padding-left: spacing(base-tight);
     padding-right: spacing(base-tight);
   }
-  
+
   &.sizeSlim {
     margin-top: rem(2px);
   }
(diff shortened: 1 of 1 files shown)


kyledurand (changes requested):
Thanks for taking this on but I think we'll need to find a different solution here. Because this change is nested under `iconOnly` we'll still see the issue when there's an icon as well as text. 

My gut feeling is that we'll need to make sure the content container heights are the same between the slim / slim with icon buttons. That may be easier said than done though so let me know if you want to pair on this

itwasmattgregg:
Closing this. It appears this was fixed somehow. I'm not seeing the issue anymore.

PR #4233, opened 2021-05-27
[Button] Fix low contract on primary outlined button

### LINKED ISSUE #4054: Low contrast on primary action hover when using outline prop
## Issue summary
A `primaryAction` button with `outline` has very low contrast on hover.  

## Expected behavior
The button should have dark text so that the text is visible and it passes contrast standards. 

## Actual behavior
Button in hovered state:
Screen Shot 2021-03-10 at 2 12 47 PM

Add an `outline: true` prop to a `primaryAction` on a Polaris component and observe on hover

## Reduced test case
Sandbox: https://codesandbox.io/s/gifted-keller-c05mg?file=/App.js

## Specifications

```bash
 System:
    OS: macOS 11.2.2
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 925.33 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 12.20.1 - ~/.nvm/versions/node/v12.20.1/bin/node
    Yarn: 1.22.5 - ~/.dev/yarn/1.22.5/bin/yarn
    npm: 6.14.10 - ~/.nvm/versions/node/v12.20.1/bin/npm
  Browsers:
    Chrome: 89.0.4389.82
    Firefox Developer Edition: 84.0
    Safari: 14.0.3
  npmPackages:
    @shopify/polaris: 6.0.1 => 6.0.1
    react: ~16.14.0 => 16.14.0
    react-dom: ~16.14.0 => 16.14.0
```

cphelefu:

A primaryAction button with `outline` has very low contrast on hover.

Fixes #4054 

* Adds the color property to the `hover`, `focus` and `active` CSS selectors for an outlined button.

#### AFTER THE FIX
After the fix

### DIFF
--- a/src/styles/shared/_buttons.scss
+++ b/src/styles/shared/_buttons.scss
@@ -134,7 +134,6 @@
   &:focus {
     border: border-width() solid var(--p-border);
     box-shadow: none;
-    color: var(--p-text);
     @include focus-ring($style: 'focused');
   }
 
--- a/src/styles/shared/_buttons.scss
+++ b/src/styles/shared/_buttons.scss
@@ -140,7 +140,6 @@
   &:active {
     border: border-width() solid var(--p-border);
     box-shadow: none;
-    color: var(--p-text);
     background: var(--p-surface-pressed);
     &::after {
       box-shadow: none;
--- a/src/styles/shared/_buttons.scss
+++ b/src/styles/shared/_buttons.scss
@@ -127,18 +127,21 @@
   &:hover {
     border: border-width() solid var(--p-border);
     box-shadow: none;
+    color: var(--p-text);
     background: var(--p-surface-hovered);
   }
 
   &:focus {
     border: border-width() solid var(--p-border);
     box-shadow: none;
+    color: var(--p-text);
     @include focus-ring($style: 'focused');
   }
 
   &:active {
     border: border-width() solid var(--p-border);
     box-shadow: none;
+    color: var(--p-text);
     background: var(--p-surface-pressed);
     &::after {
       box-shadow: none;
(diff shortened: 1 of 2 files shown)


src/styles/shared/_buttons.scss
```diff
@@ -127,18 +127,21 @@
...
     border: border-width() solid var(--p-border);
     box-shadow: none;
+    color: var(--p-text);
     background: var(--p-surface-hovered);
   }
 
   &:focus {
     border: border-width() solid var(--p-border);
     box-shadow: none;
+    color: var(--p-text);
```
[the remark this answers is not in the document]
cphelefu: Thank you. Removed this line.

src/styles/shared/_buttons.scss
```diff
@@ -139,6 +140,7 @@
   &:active {
     border: border-width() solid var(--p-border);
     box-shadow: none;
+    color: var(--p-text);
```
dleroux, line 143: I think it can also be removed here

dleroux: Also note, below under the `pressed` state there's a `color: var(--p-button-text);` Which I don't think exists

cphelefu: @dleroux the main cause of this error is overwriting/inheritance of CSS properties when two classes are used.. i.e. in  `<Button primary outline>Text</Button>` `.outline:hover` color is inherited from the `.primary:hover` class since it is not defined for `.outline:hover`. Fix add color to `.outlined:hover`.
Since color is also not defined in `:focus` and `:active` this will likely occur again for these if the class used along `.outline` has a color defined in these selectors. [Thats why I initially added `color` in these selectors too].

Didnt remove `var(--p-button-text);` because it is used in other parts of the `_button.scss` and is outside the scope of the bug. Maybe we create another bug for it to be removed?

PR #4261, opened 2021-06-14
Expose Card compound component props

pedrodurek:

Right now, in order to get Card section props we have to use the following workaround `Parameters<typeof Card.Section>['0'];`. This works but it's too hacky and not ideal for readability.

This PR aims to expose all Card compound components Props, so instead of using `Parameters` helper to extract component's props, we can simply import them: `CardSectionProps`, `CardHeaderProps`, `CardSubsectionProps`.

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Card/Card.tsx
+++ b/src/components/Card/Card.tsx
@@ -13,6 +13,12 @@ import {Popover} from '../Popover';
 import {Header, Section, Subsection} from './components';
 import styles from './Card.scss';
 
+export type {
+  SectionProps as CardSectionProps,
+  HeaderProps as CardHeaderProps,
+  SubsectionProps as CardSubsectionProps,
+} from './components';
+
 export interface CardProps {
   /** Title content for the card */
   title?: React.ReactNode;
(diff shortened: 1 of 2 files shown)


src/components/Card/Card.tsx
```diff
@@ -13,6 +13,12 @@ import {Popover} from '../Popover';
 import {Header, Section, Subsection} from './components';
 import styles from './Card.scss';
 
+export type {
+  SectionProps as CardSectionProps,
```
BPScott, line 17: Polaris tends to avoid renaming when reexporting as it having the same concept available with two different names is confusing. Can you rename HeaderProps/SectionProps and SubsectionProps at their source so this line can be just `export type {CardHeaderProps, CardSectionProps, CardSubSectionprops} from ...`

pedrodurek: Isn't it going to cause confusion when importing them?
```tsx
import type {HeaderProps} from '@shopify/polaris';
``` 
In my view, this ☝️ implies that we have a `Header` component in place and we're importing its Prop types. The same applies to `SectionProps` and `SubSectionprops`. Alternatively, I can rename its actual types.

alololox (approved):
LGTM 💃

BPScott (approved):

PR #4264, opened 2021-06-16
[TextField] - Add monospaced prop

alexanderMontague:

- For our particular use case, we use a `<TextField />` to display an access token that needs to be clearly interpreted by the user
- The use of our current default font on `<TextField />` leads to some mix ups including not being able to differentiate certain characters like I (uppercase i) and l (lowercase L).
- This introduces a `monospaced` prop that applies a monospaced font to the `<TextField />` allowing for easy character distinction

Example - Top is default, bottom is with `monospaced` prop:
image

- adds a `monospaced` prop to `<TextField />`
- This applies the `.Monospaced` css class to the input
  -  The font-families applied try to use the `SF Mono` font first, while falling back to system defaults as per our design suggestions
- I followed a suggestion to take a look at how `<TextStyle />` conditionally applies fonts/colours but did not think a whole `variation` prop was needed right now. If we think this would be useful instead I can change it, but would I also support the other variations currently on `<TextStyle />`?

## 

### How to 🎩

<details>

```jsx
import React from 'react';

  return (
    <Page title="Playground">
      <TextField value="test 1IlLI1" />
      <TextField monospaced value="test 1IlLI1" />
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/TextField/TextField.scss
+++ b/src/components/TextField/TextField.scss
@@ -310,3 +310,7 @@ $stacking-order: (
     margin-top: 0;
   }
 }
+
+.monospaced {
+  font-family: font-family($family: 'monospace');
+}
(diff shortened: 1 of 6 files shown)


src/components/TextField/TextField.scss
```diff
@@ -310,3 +310,8 @@ $stacking-order: (
     margin-top: 0;
   }
 }
+
+.Monospaced {
```
kyledurand, line 314: Bit of a nitpick but our variations are usually lowerCamelCase and I'm not sure if we have conventions but I'd prefer if we matched the property name here as well

```suggestion
.monospace {
```

kyledurand (commented):
This looks awesome. Just a couple of comments. Also would you be able to add an example to the readme?

PR #4266, opened 2021-06-17
[BulkActions] - Allowing promoted actions to be rendered as menus

### LINKED ISSUE #4246: [BulkActions] - Add an action group as a promotedAction
## Feature request summary

We'd like to be able to make `promotedActions` evolve into `action groups`, so that we can offer more meaningful actions, group them into menus and keep the UX organized.  

Screen Shot 2021-06-03 at 10 17 29 PM

https://user-images.githubusercontent.com/10034981/120738389-c81ae180-c4bd-11eb-8926-d82790f0f651.mp4

## Rationale

It seems that Polaris advises limiting `promotedActions` up to 2 actions, and as part of a project initiative, we would need to add a third one to the `BulkActions` component that is being used on the `Orders index` page. We also know that more actions will be added in the near future and they would benefit from this addition. 

Our team’s suggestion is that we add the ability to make `promoted actions` evolve into `action groups`, similarly to what we do on the `Page component`.  With this addition, we would be able to give merchants more options to improve their workflow, grouping these actions by subject, for example. 

#### Potential solution

I thought of changing the `promotedActions` prop to start accepting a [`MenuGroupDescriptor []`](https://github.com/Shopify/polaris-react/blob/main/src/types.ts#L176) as we do for `ActionGroups` on the `Page` component, and then dynamically render either the single button or the menu depending on the props' type. This option would be trickier to implement but should scale better as we could have for example two menus at the same time.

A simpler alternative would be to add an extra prop to render only one menu and limit promoted actions to 1 instead of 2. 

I’m working on a first draft PR but I would love to hear thoughts on that. 

---

thiagoloschi:

Closes #4246 

Currently, promoted actions are single buttons and are limited to only 2 actions in the `BulkAction` component. With this change, we're proposing that each one of these buttons can become a menu where we would group relevant actions by subject. This can be seen [on this card](https://github.com/Shopify/core-issues/issues/24023) with the real use case where it would be first used.

<details>
  <summary>What it looks like now 🎩</summary>
 <b>On large screens</b>
  <br />
  promoted action as a menu, with an open popover
 <b>On small screens</b>
  <br />
   promoted action menu rolled into the general actions menu when in small screens
 <b>when the two buttons are menus</b>
  <br />

image
image
    </details>

https://user-images.githubusercontent.com/10034981/122357313-a12ec780-cf21-11eb-9590-867902cb8870.mp4

To make this possible, we're basically changing the `promotedActions` prop to also accept values of type `MenuGroupDescriptor`, which are essentially the type of the menus that we'll be rendering. Then where we define what the `promotedActions` markup will look like, we're mapping through the actions, checking for their type through a type guard function and deciding based on it whether we should render a `BulkActionButton` or a `BulkActionMenu` (which is also being introduced in this PR).

The other change introduced on this PR was for smaller screens, where promoted actions will get rolled into a general `Actions` menu. Here we still want the actions on the menus to be grouped together, so we're making sure they work as `ActionListSection`, similarly to how the prop `actions` is being treated.

##### Notes
1. This should be backwards compatible as we're just extending the capabilities of promoted actions. 
2. Regular button actions and newly added menu actions can coexist.
3. We can have as many menus as we'd like (the warning for keeping promoted actions up to 2 actions will still be thrown though)

## 

### How to 🎩

<details>

```jsx

import {Page, UnstableBulkActions, Button, Badge, Frame} from '../src';

  const toggleActive = useCallback(() => setActive((active) => !active), []);

  const bulkActionProps = {
    bulkActions: ['button 3', 'button 4', 'button 5'],
    promotedActions: [
      {
        title: 'button1',
        actions: [
          {
            content: 'action1',
          },
          {
            content: 'action2',
          },
          {
            content: 'action3',
            badge: {content: 'new', status: 'new'},
            helpText: 'Short help text',
          },
        ],
      },
      {
        content: 'button 2',
      },
    ],
    actions: [
      {
        content: 'button 7',
      },
      {
        content: 'button 6',
      },
    ],
    paginatedSelectAllText: 'paginated select all text string',
    selected: true,
    accessibilityLabel: 'test-aria-label',
    label: 'Test-Label',
    disabled: false,
    smallScreen: false,
    selectMode: true,
  };

  return (
    <Page title="Playground">
      <UnstableBulkActions {...bulkActionProps} />
    </Page>
  );
}
```

</details>

### 🎩 checklist

- Updated the component's `README.md` with documentation changes (there's no readme for this component)

### DIFF
--- a/src/components/BulkActions/BulkActions.tsx
+++ b/src/components/BulkActions/BulkActions.tsx
@@ -11,6 +11,7 @@ import type {
   DisableableAction,
   Action,
   ActionListSection,
+  MenuGroupDescriptor,
 } from '../../types';
 import {ActionList} from '../ActionList';
 import {Popover} from '../Popover';
@@ -19,10 +20,10 @@ import {ButtonGroup} from '../ButtonGroup';
 import {EventListener} from '../EventListener';
 import {CheckableButton} from '../CheckableButton';
 
-import {BulkActionButton} from './components';
+import {BulkActionButton, BulkActionMenu} from './components';
 import styles from './BulkActions.scss';
 
-type BulkAction = DisableableAction & BadgeAction;
+export type BulkAction = DisableableAction & BadgeAction;
 
 type BulkActionListSection = ActionListSection;
 
@@ -42,7 +43,7 @@ export interface BulkActionsProps {
   /** List is in a selectable state */
   selectMode?: boolean;
   /** Actions that will be given more prominence */
-  promotedActions?: BulkAction[];
+  promotedActions?: (BulkAction | MenuGroupDescriptor)[];
   /** List of actions */
   actions?: (BulkAction | BulkActionListSection)[];
   /** Text to select all across pages */
@@ -178,6 +179,28 @@ class BulkActionsInner extends PureComponent<CombinedProps, State> {
     }
   }
 
+  private rolledInPromotedActions() {
+    const {promotedActions} = this.props;
+    const numberOfPromotedActionsToRender = this.numberOfPromotedActionsToRender();
+
+    if (
+      !promotedActions ||
+      promotedActions.length === 0 ||
+      numberOfPromotedActionsToRender >= promotedActions.length
+    ) {
+      return [];
+    }
+
+    const rolledInPromotedActions = promotedActions.map((action) => {
+      if (instanceOfMenuGroupDescriptor(action)) {
+        return {items: [...action.actions]};
+      }
+      return {items: [action]};
+    });
+
+    return rolledInPromotedActions.slice(numberOfPromotedActionsToRender);
+  }
+
   // eslint-disable-next-line @typescript-eslint/member-ordering
   componentDidMount() {
     const {actions, promotedActions} = this.props;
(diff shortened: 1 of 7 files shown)


charlottepoon:
cc: @sarahill

src/components/BulkActions/BulkActions.tsx
```diff
@@ -555,6 +585,12 @@ function instanceOfBulkActionArray(
   return actions.length === validList.length;
 }
 
+function instanceOfMenuGroupDescriptor(
+  action: MenuGroupDescriptor | BulkAction,
+): action is MenuGroupDescriptor {
+  return (action as MenuGroupDescriptor).title !== undefined;
```
AndrewMusgrave, line 591: If you wanted a type safe way (without casting), we can use the `in` operator

```
return 'title' in action
```

src/components/BulkActions/components/BulkActionMenu/BulkActionMenu.tsx
```diff
@@ -0,0 +1,42 @@
...
+import {Popover} from '../../../Popover';
+import {ActionList} from '../../../ActionList';
+import {BulkActionButton} from '../BulkActionButton';
+import type {MenuGroupDescriptor} from '../../../../types';
+
+interface BulkActionsMenuProps extends MenuGroupDescriptor {
+  isNewBadgeInBadgeActions: boolean;
+}
+
+export function BulkActionMenu({
```
AndrewMusgrave, line 12: It shouldn't need a lot, but would you mind adding a few tests 🙏

src/components/BulkActions/components/BulkActionMenu/BulkActionMenu.tsx
```diff
@@ -0,0 +1,42 @@
...
+  title,
+  actions,
+  isNewBadgeInBadgeActions,
+}: BulkActionsMenuProps) {
+  const [isVisible, setVisible] = useState<boolean>(false);
```
AndrewMusgrave, line 17: Small nit: we have a `useToggle` hook

src/components/BulkActions/tests/BulkActions.test.tsx
```diff
@@ -283,6 +296,96 @@ describe('<BulkActions />', () => {
...
+          disabled: false,
+        };
+        const bulkActions = mountWithAppProvider(
+          <BulkActions {...bulkActionProps} />,
+        );
+        const bulkActionButtons = bulkActions.find(BulkActionButton);
+        expect(bulkActionButtons).toHaveLength(4);
+        const bulkActionMenus = bulkActions.find(BulkActionMenu);
+        expect(bulkActionMenus).toHaveLength(2);
+        warnSpy.mockRestore();
```
AndrewMusgrave, line 348: I know this pattern was already used in this file, but should we move this + spyOn to a beforeEach/afterEach?

AndrewMusgrave (commented):
This is looking really good! 💯

src/components/BulkActions/components/BulkActionMenu/tests/BulkActionMenu.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+      bulkActionButton!.trigger('onAction');
+      const actionList = bulkActionMenu.find(ActionList);
+      actionList!.trigger('onActionAnyItem');
+
+      expect(bulkActionMenu).toContainReactComponent(Popover, {
+        active: false,
+      });
+    });
+  });
+});
```
thiagoloschi, line 73: image

Germanika (approved):
Code lgtm! @AndrewMusgrave do you think it would make sense to add an example of this to the IndexTable storybook?

AndrewMusgrave (approved):
LGTM ✅   @Germanika a new example totally makes sense!!

PR #4270, opened 2021-06-21
[Button] Fix disabled button text in Safari

### LINKED ISSUE #4228: Primary button has incorrect enabled state in Safari
## Issue summary

The enabled state of the primary button does not have white text **on Safari**. This works fine in Chrome. I did not test other browsers.

I could not replicate on polaris.shopify.com.

## Expected behavior

This is what happens in Chrome:

Disabled:
chrome disabled state

Enabled:
chrome enabled state

## Actual behavior

Disabled (correct):
save disabled state

Enabled:
save enabled state

1. Log into Shopify admin using Safari
2. Go to a product details page
3. Make an edit. 
4. Notice that the save button changes state but the text is hard to read.

## Reduced test case

Again, I am not able to reproduce on polaris.shopify.com or in the sandbox.

## Specifications

- Polaris version number: This is in production, live on /admin rights now.
- Browser: Safari
- Device: Mac
- Operating System: macOS

nicolleromero:

Fixes #4228 

Fixes a bug in Safari that prevents the enabled state of the primary button from changing to white text when previously disabled.

Adds a key to the Text of the button component to force a re-paint when the primary button changes from disabled to enabled in Safari. Forces the browser to change the button text from grey to white when enabled.

https://user-images.githubusercontent.com/25596676/122698888-2275ab00-d1fd-11eb-857e-19e85fcb54c2.mov

### Steps to Reproduce the Issue

1. Open this [Sandbox](https://codesandbox.io/s/polaris-disabled-button-bug-in-safari-dyfwn?file=/App.js) in Safari.
2. Uncheck "Disabled".
3. Button text will be gray instead of white.

OR

1. Check out this branch (`fix-disabled-button-text`).
2. Create a file at /playground/Playground.tsx.
3. Add the following code to that file:
```
import React from 'react';

import {Button, Checkbox, Page} from '../src';

  const [disabled, setDisabled] = React.useState(true);

  return (
    <Page title="Playground">
      <Checkbox
        label="Disabled"
        checked={disabled}
        onChange={() => setDisabled(!disabled)}
      />
      <div>
        <Button disabled={disabled} primary>
          Broken in safari?
        </Button>
      </div>
    </Page>
  );
}

```

### 🎩 checklist

- [X] Tested on [mobile](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting.md#cross-browser-testing)
- [X] Tested on [multiple browsers](https://help.shopify.com/en/manual/intro-to-shopify/shopify-admin/supported-browsers)
- [X] Updated the component's `README.md` with documentation changes


UNRELEASED.md
```diff
@@ -15,6 +15,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Bug fixes
 
+- Fix bug in Safari where button text is gray instead of white for enabled `Button` ([#4270](https://github.com/Shopify/polaris-react/pull/4270))
```
kyledurand, line 18: Can we elaborate a bit here? Maybe mention that it happens during a state change from disabled to enable and vice versa

kyledurand (approved):
Amazing. Great find!

runmad:
Thanks for fixing this!!! Nice find.

PR #4280, opened 2021-06-26
TextField added event return to onChange and updated the test

brilliant-ember:
As part of fixing this https://github.com/Shopify/shipping/issues/3083 we need to pass the event to `addressLine` in web, and to that end the polaris TextField is being updated to allow passing the event. 
Next part of the above issue fix is to use the event to prevent the curser from jumping to the end.

Fixes #0000 

Allow TextField to pass the input event so that components that need it can access it.

## 

### How to 🎩
https://user-images.githubusercontent.com/23269759/123495307-394b4180-d5f1-11eb-9e61-eee7f5bdfecd.mp4

🖥 [Local development instructions]()

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -13,6 +13,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Rename and expose Card compound components types ([#4261](https://github.com/Shopify/polaris-react/pull/4261))
 - Add `monospaced` prop to `TextField` component ([#4264](https://github.com/Shopify/polaris-react/pull/4264))
 - Add base tight spacing option to `Stack` component([#4273](https://github.com/Shopify/polaris-react/pull/4273))
+- Add the ability to retian the event.target from the `TextField` (https://github.com/Shopify/polaris-react/pull/4280 and https://github.com/Shopify/shipping/issues/3083)
 
 ### Bug fixes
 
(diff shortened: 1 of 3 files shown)


alex-page:
@brilliant-ember we have intentionally not included the browser event api in this component. Is there an alternative approach you could try in your implementation? How are teams doing this in `shopify/web`?

brilliant-ember:
> @brilliant-ember we have intentionally not included the browser event api. Is there an alternative approach you could try in your implementation? How are teams doing this in `shopify/web`?

other places don't use redux, this problem seems to happen when redux state store updates the textfield and there's a component-local state changing (loading spinner) The addressLine.tsx in web is used in multiple places but not all of them use redux, the shippingLabels page uses redux and that's where the problem is observed. So far I am talking with a senior FED who worked in that code part to see if we can somehow have a specific solution for the redux page, but if that's not possible I can't think of a solution other than somehow passing the event from polaris and programmatically place the curser using the selectedStart/End stored in the event.

lucasfrosty:
Hi 👋 i'm the FED that @brilliant-ember mentioned.

Unfortunately using Redux for state management all the time (even for small fields inside modals) is a required thing on our application because of some reasons that are not worth going through right now. However we did some investigation today and find a workaround for this that won't require dealing with this cursor position hack... Basically we're gonna use local state, but that will be just a copy of the field contained in the Redux store, and we'll keep both in sync all the time, that way we'll update the store accordingly, and the cursor issue won't be a thing anymore since we'll be actually using local state instead.

So, to summarize, this PR will no longer be necessary.

PR #4292, opened 2021-07-07
[ThemeProvider] Add example for custom config

kyledurand:
image

Addresses https://shopify.slack.com/archives/C4Y8N30KD/p1625670603198300?thread_ts=1625609553.186000&cid=C4Y8N30KD

Adds another example to our theme provider readme where we pass it a custom config to achieve custom colors

## 

http://localhost:6006/?path=/story/all-components-theme-provider--theme-provider-with-a-color-scheme-and-custom-config-nested-within-an-app-provider

### DIFF
--- a/src/components/ThemeProvider/README.md
+++ b/src/components/ThemeProvider/README.md
@@ -141,6 +141,74 @@ A theme provider can be nested within the theme provider rendered by the app pro
 </AppProvider>
 ```
 
+### Theme provider with a color scheme and custom config nested within an app provider
+
+You can pass in a custom config to the theme provider to override default HSL values. Find the lightness value you need using www.hsluv.org
+
+```jsx
+<AppProvider i18n={{}}>
+  <TextContainer>
+    <Card
+      title="Shipment 1234"
+      secondaryFooterActions={[{content: 'Edit shipment'}]}
+      primaryFooterAction={{content: 'Add tracking number'}}
+    >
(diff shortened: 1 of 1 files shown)


alex-page (approved):

src/components/ThemeProvider/README.md
```diff
@@ -141,6 +141,74 @@ A theme provider can be nested within the theme provider rendered by the app pro
...
+            {
+              name: 'text',
+              description: 'For use as a text color.',
+              light: {lightness: 100},
+              dark: {lightness: 100},
+              meta: {
+                figmaName: 'Text/Default',
+              },
+            },
+          ],
```
d9su, line 191: Where can one read more about these config values and understand what each means?

d9su: Should we add some documentation around what `lightness` means, and what `figmaName` means, etc? Or are those considered domain knowledge?

kyledurand: I feel like between the color docs file and figma it's easy enough to find what you're looking for. 

Lightness, saturation, and hue on the other hand may be intuitive to some more than others. I'm not sure where exactly the documentation for that would live off the top of my head though.

Feel free to create a PR in a place you see fit and ping me as a reviewer

d9su (approved):
Thanks for adding this example!

PR #4295, opened 2021-07-08
Remove remaining css fallbacks

kyledurand:
Fixes: https://github.com/Shopify/polaris-react/issues/4296

### DIFF
--- a/src/components/MessageIndicator/MessageIndicator.scss
+++ b/src/components/MessageIndicator/MessageIndicator.scss
@@ -15,6 +15,6 @@ $indicator-position: rem(-3px);
   width: $indicator-size;
   height: $indicator-size;
   border-radius: 100%;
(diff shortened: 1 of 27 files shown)


src/components/MessageIndicator/MessageIndicator.scss
```diff
@@ -15,6 +15,6 @@ $indicator-position: rem(-3px);
   width: $indicator-size;
   height: $indicator-size;
   border-radius: 100%;
-  background-color: var(--p-icon-highlight, color('teal'));
-  border: solid rem(2px) var(--p-background, var(--top-bar-background));
```
kyledurand, line 19: @alex-page I'm thinking about adding these back to this PR, they might be used by plus or other teams but we should be able to remove them in v7. What do you think?

alex-page (approved):

PR #4306, opened 2021-07-12
[Test modernization] Avatar, Backdrop, Badge, Banner

rdott:

Updating existing tests to use the new modern framework and eliminate stale dependencies.

Part of test modernization, updating tests using `{mountWithAppProvider} from 'test-utilities/legacy'` to `{mountWithApp} from 'test-utilities'`

### DIFF
--- a/src/components/Avatar/tests/Avatar.test.tsx
+++ b/src/components/Avatar/tests/Avatar.test.tsx
@@ -1,28 +1,30 @@
 import React from 'react';
-// eslint-disable-next-line no-restricted-imports
-import {mountWithAppProvider} from 'test-utilities/legacy';
 import {mountWithApp} from 'test-utilities';
 import {Avatar, Image} from 'components';
 
 describe('<Avatar />', () => {
   describe('intials', () => {
-    it('renders intials if the image is not provided', () => {
-      const avatar = mountWithAppProvider(<Avatar initials="DL" />);
-      expect(avatar.find('span[role="img"] span svg')).toHaveLength(1);
+    it('renders initials if the image is not provided', () => {
+      const avatar = mountWithApp(<Avatar initials="DL" />);
+
+      expect(avatar).toContainReactComponent('span', {
+        className: 'Initials',
+      });
+      expect(avatar).toContainReactText('DL');
+      expect(avatar).not.toContainReactComponent(Image);
     });
   });
 
   describe('source', () => {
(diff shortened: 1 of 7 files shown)


src/components/Avatar/tests/Avatar.test.tsx
```diff
@@ -1,28 +1,25 @@
...
 import {mountWithApp} from 'test-utilities';
 import {Avatar, Image} from 'components';
 
 describe('<Avatar />', () => {
   describe('intials', () => {
     it('renders intials if the image is not provided', () => {
-      const avatar = mountWithAppProvider(<Avatar initials="DL" />);
-      expect(avatar.find('span[role="img"] span svg')).toHaveLength(1);
+      const avatar = mountWithApp(<Avatar initials="DL" />);
+      expect(avatar).toContainReactComponent('text', {children: 'DL'});
```
voiid, line 9: This tests the implementation details of `Avatar` pretty closely. Maybe a `toContainReactText` would be better suited for what you're testing here?

src/components/Backdrop/tests/Backdrop.test.tsx
```diff
@@ -1,24 +1,25 @@
...
 import {Backdrop} from '..';
 
 describe('<Backdrop />', () => {
   describe('onDismiss()', () => {
     it('is called when the backdrop is clicked', () => {
       const spy = jest.fn();
-      const backdrop = mountWithAppProvider(<Backdrop onClick={spy} />);
-      findByTestID(backdrop, 'Backdrop').simulate('click');
+      const backdrop = mountWithApp(<Backdrop onClick={spy} />);
+      backdrop.find('div')!.trigger('onClick');
```
voiid, line 11: Simple enough component that I don't think it's a real issue but ideally we wouldn't target elements this vaguely since it can be quite brittle. If another `div` were added to this component it would start failing and you've have to reverse engineer which random div it was triggering the action on. Again, not _reaaally_ an issue for this particular component because of how small it is but I'd love for us to think about how we can target these in the future.

voiid (approved):
LGTM. Few minor comments

src/components/Backdrop/tests/Backdrop.test.tsx
rdott: Targeted the <div> using the `onClick` prop.

alex-page (approved):

PR #4307, opened 2021-07-12
[Layout][Heading] add ID prop to Heading, Layout.AnnotatedSection

brendanrygus:
* In the SFN application, we want to link users from the home page to a section of the Settings page
* The `AnnotatedSection` subcomponents allow us to break up content of a Settings page, but do not support hash links
* The implicit layout used by `Layout` assumes that `Section` or `AnnotatedSection` will always be the direct child node of the parent, which does not allow consumers to wrap their content section in a labelled div

Example of broken layout with a wrapper div around the shipping section:
Screen Shot 2021-07-12 at 4 16 35 PM

* Adds an optional ID prop to the `Heading` component
* Adds an optional ID prop to the `Layout.AnnotatedSection` component, passed through to the primary Heading

### 🎩 checklist

**I couldn't get this working locally 🤔, but the changes seem pretty straightforward**
- [ ] ~For visual design changes, ping @ sarahill to update the Polaris UI kit~

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -6,7 +6,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
-- [Layout] Add `id` prop for deeplinking to `AnnotatedSection` components
+- Added `id` prop to `Layout` and `Heading` for hash linking ([#4307](https://github.com/Shopify/polaris-react/pull/4307))
 
 ### Bug fixes
 
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -6,6 +6,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- [Layout] Add `id` prop for deeplinking to `AnnotatedSection` components
+
 ### Bug fixes
 
 ### Documentation
--- a/src/components/Heading/Heading.tsx
+++ b/src/components/Heading/Heading.tsx
@@ -12,8 +12,14 @@ export interface HeadingProps {
   element?: HeadingTagName;
   /** The content to display inside the heading */
   children?: React.ReactNode;
+  /** A unique identifier for the heading, used for reference in anchor links  */
+  id?: string;
 }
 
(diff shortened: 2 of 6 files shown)


brendanrygus:
> Within the annotated section there's a `<Heading />` component with the section title. Would it make more sense to add the id there? I think that would be better for screen readers

Thanks @kyledurand, I had wondered about that as well. I was focusing more on the scroll position on my first pass. I'll push a quick update

src/components/Heading/Heading.tsx
```diff
@@ -12,8 +12,14 @@ export interface HeadingProps {
   element?: HeadingTagName;
   /** The content to display inside the heading */
   children?: React.ReactNode;
+  /** A unique identifier for the heading, used for reference in anchor links  */
+  id?: string;
```
brendanrygus, line 16: @kyledurand Turns out the Heading didn't take an ID prop, so I updated the component and type docs.
If this looks good, let me know if there are any examples we should add to the README / code sample docs

UNRELEASED.md
```diff
@@ -6,6 +6,9 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- [Layout] Add `id` prop for hash linking to `AnnotatedSection` components
+- [Heading] Add `id` prop
```
kyledurand, line 10: Can we change this to something like:
```suggestion
- Added `id` prop to `Layout` and `Heading` for hash linking ([#4307](https://github.com/Shopify/polaris-react/pull/4307))
```

To match our [guidelines](https://github.com/Shopify/polaris-react/blob/main/documentation/Versioning%20and%20changelog.md#user-content-component-name-formats)

kyledurand (commented):
This is looking good. Can you add a quick test to heading as well? Just to make sure we don't remove the id in the future and cause a regression.

brendanrygus:
Thanks for the speedy reviews @kyledurand!

This is my first PR here - I noticed that the contributing guide says: 
> “We’ll review your pull request and either merge it, request changes to it, or close it with an explanation.”

Let me know if you need anything else from my end - squashing commits, second approvals etc. - before it can be merged

PR #4308, opened 2021-07-13
Remove dangling default

atesgoral:

Ran into a parsing issue while experimenting with dart-sass and running sass-migrator:

```
Error: expected "}".
   ╷
17 │   monospace !default},
   │             ^
   ╵
  src/styles/foundation/_typography.scss 17:13  root stylesheet
Migration failed!
```

Removing dangling `!default` that causes the parsing issue. I don't know what it's supposed to do. According to @bpscott:

> This looks like a straight up bug and we can remove this ...

### 🎩 checklist

### DIFF
--- a/src/styles/foundation/_typography.scss
+++ b/src/styles/foundation/_typography.scss
@@ -14,7 +14,7 @@ $font-family-data: (
   Consolas,
   Liberation Mono,
   Menlo,
-  monospace !default},
+  monospace},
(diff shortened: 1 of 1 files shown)


kyledurand (approved):
I was always confused by this !default. I don't think it ever worked as intended and I'm not sure what that intention was. Seems safe to remove

PR #4313, opened 2021-07-14
[Test Modernization]: Add test modernization to Item, Panel, List, Tab, TabMeasurer

kpinzas-sh:
**WHY are these changes introduced?**

I am updating existing tests for Item, Panel, List, Tab, TabMeasurer components to use the new testing framework.

**WHAT is this pull request doing?**

Part of test modernization (https://docs.google.com/spreadsheets/d/1GBuEZbOpVYJLNISK7gL69DU8rCxocn5L5GYaKtPXPbU/edit#gid=1498187033), updating tests using {mountWithAppProvider} from 'test-utilities/legacy' to {mountWithApp} from 'test-utilities'.

First time doing this, so if you have any suggestion, it would be really appreciate it.

### DIFF
--- a/src/components/Tabs/components/List/tests/List.test.tsx
+++ b/src/components/Tabs/components/List/tests/List.test.tsx
@@ -1,6 +1,5 @@
 import React from 'react';
-// eslint-disable-next-line no-restricted-imports
-import {mountWithAppProvider} from 'test-utilities/legacy';
+import {mountWithApp} from 'test-utilities';
 
 import {List} from '../List';
 import {Item} from '../../Item';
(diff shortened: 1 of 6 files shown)


src/components/Tabs/components/List/tests/List.test.tsx
```diff
@@ -85,18 +88,17 @@ describe('<List />', () => {
...
     });
 
     it('renders the provided content', () => {
-      const list = mountWithAppProvider(
+      const list = mountWithApp(
         <List {...mockProps} disclosureTabs={disclosureTabs} />,
       );
-      const firstItem = list.find(Item).first();
-      expect(firstItem.contains('Repeat customers')).toBe(true);
+      expect(list).toContainReactText('Repeat customers');
```
rdott, line 101: Would 
```expect(list.find(Item)).toContainReactText('Repeat customers');```
Better translate from the old test?

rdott (approved):
One comment, nothing blocking. Looks good!

src/components/Tabs/components/List/tests/List.test.tsx
kpinzas-sh: Makes sense! I will go for that option! Thanks for the suggestion!

PR #4329, opened 2021-07-22
[IndexTable] Add support for non-selectable table when selectable prop is false

### LINKED ISSUE #4076: [IndexTable] selectable prop being set to false has no effect
## Issue summary

On `IndexTable`, the boolean prop `selectable` is exposed to enable / disable selectability on `IndexTable`. However, setting `selectable` to `false` does not disable selectability.

## Expected behavior

By setting `selectable` to `false` on `IndexTable`, items should not be selectable in `IndexTable`

## Actual behavior

Setting `selectable` to false allows for items to be selected.

## Reduced test case

Sample `IndexTable` example:

```js
import React from "react";
import {
  TextStyle,
  Card,
  IndexTable,
  useIndexResourceState
} from "@shopify/polaris";

export default function SimpleIndexTableExample() {
  const customers = [
    {
      id: "3411",
      url: "customers/341",
      name: "Mae Jemison",
      location: "Decatur, USA",
      orders: 20,
      amountSpent: "$2,400"
    },
    {
      id: "2561",
      url: "customers/256",
      name: "Ellen Ochoa",
      location: "Los Angeles, USA",
      orders: 30,
      amountSpent: "$140"
    }
  ];
  const resourceName = {
    singular: "customer",
    plural: "customers"
  };

  const {
    selectedResources,
    allResourcesSelected,
    handleSelectionChange
  } = useIndexResourceState(customers);

  const rowMarkup = customers.map(
    ({ id, name, location, orders, amountSpent }, index) => (
      <IndexTable.Row
        id={id}
        key={id}
        selected={selectedResources.includes(id)}
        position={index}
      >
        <IndexTable.Cell>
        </IndexTable.Cell>
        <IndexTable.Cell>{location}</IndexTable.Cell>
        <IndexTable.Cell>{orders}</IndexTable.Cell>
        <IndexTable.Cell>{amountSpent}</IndexTable.Cell>
      </IndexTable.Row>
    )
  );

  return (
    <Card>
      <IndexTable
        selectable={false}
        resourceName={resourceName}
        itemCount={customers.length}
        selectedItemsCount={
          allResourcesSelected ? "All" : selectedResources.length
        }
        onSelectionChange={handleSelectionChange}
        headings={[
          { title: "Name" },
          { title: "Location" },
          { title: "Order count" },
          { title: "Amount spent" }
        ]}
      >
        {rowMarkup}
      </IndexTable>
    </Card>
  );
}
```

## Specifications

- Polaris version number: 6.2.0
- Browser: Chrome
- Device: Desktop
- Operating System: Ubuntu

ryanwilsonperkin:

Fixes #4076

Introduces a variant of the `IndexTable` that has no checkboxes in the leftmost column when the `selectable` prop is false. This is inline with the naming of the prop and makes the table usable for resources where the user may not need to support selection, and is instead just using the table as a resource-specific version of a DataTable.

- Updates the IndexProvider to pass through the `selectable` prop into the context. This was one of the allowed props on the context but wasn't actually being passed into the context object itself. Instead the context was deriving its own value of "selectable" to mean "there are more than 0 items present" which is renamed to "hasSelections"
- Updates previous uses of the term "selectable" where it meant "has bulk actions" to be "actionable" instead to indicate intent, and to differentiate it from "selectable" which indicates "rows in this table should be able to be selected"
- Refactored the `renderHeading` inline-component to have a simpler signature by separating out a new `renderCheckboxCell` inline-component. Previously, `renderHeading` would return a 2-tuple of JSX on the first heading, and a single JSX otherwise, which was a bit confusing and harder to manage when checkbox is optional. Instead we'll do a simple map with `renderHeading` and then insert the checkbox content at the front of the array if desired.

A screenshot of the approach rendered with many columns so that it scrolls horizontally. Note the lack of checkbox:

image

#### TODO
- Still needs some additional work in order to get the styling of the sticky columns correct when a checkbox is not rendered. Currently our code makes a lot of assumptions that the first column will be there, and uses it to calculate the offset width for the second column. When the checkbox isn't present, the sticky column calculation doesn't work.

### How to 🎩

Copy-paste the code from the parent ticket into the Playground in order to test this out.

### 🎩 checklist

### DIFF
--- a/src/components/IndexProvider/IndexProvider.tsx
+++ b/src/components/IndexProvider/IndexProvider.tsx
@@ -18,6 +18,7 @@ export function IndexProvider({
   itemCount,
   hasMoreItems,
   condensed,
+  selectable,
 }: IndexProviderProps) {
   const {
     paginatedSelectAllText,
@@ -26,7 +27,6 @@ export function IndexProvider({
     resourceName,
     selectMode,
     bulkSelectState,
-    selectable,
   } = useBulkSelectionData({
     selectedItemsCount,
     itemCount,
--- a/src/components/IndexTable/IndexTable.tsx
+++ b/src/components/IndexTable/IndexTable.tsx
@@ -70,6 +70,7 @@ function IndexTableBase({
     bulkSelectState,
     resourceName,
     bulkActionsAccessibilityLabel,
+    selectable,
     selectMode,
     paginatedSelectAllText,
     itemCount,
@@ -290,7 +291,7 @@ function IndexTableBase({
     }
   }, [condensed, isSmallScreenSelectable]);
 
-  const selectable = Boolean(
+  const actionable = Boolean(
     (promotedBulkActions && promotedBulkActions.length > 0) ||
       (bulkActions && bulkActions.length > 0),
   );
@@ -674,7 +675,7 @@ function IndexTableBase({
   }
 
   function getPaginatedSelectAllAction() {
-    if (!selectable || !hasMoreItems) {
+    if (!actionable || !hasMoreItems) {
       return;
     }
 
--- a/src/components/IndexTable/IndexTable.tsx
+++ b/src/components/IndexTable/IndexTable.tsx
@@ -568,7 +568,7 @@ function IndexTableBase({
     const isSecond = index === 0;
     const headingContentClassName = classNames(
       styles.TableHeading,
-      isSecond && styles['TableHeading-second'],
+      selectable && isSecond && styles['TableHeading-second'],
     );
 
     const stickyPositioningStyle =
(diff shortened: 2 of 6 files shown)


ryanwilsonperkin:
Got a good start on this so far, but looks like it'll still need a fair bit of refactoring in the `IndexTable` component in order to ensure that it makes the correct column(s) sticky when scrolling horizontally. Right now it's hardcoded to always assume that the first two columns should be sticky, and calculates the offset accordingly. When `selectable=false` only the first column should be sticky, because there's no initial checkbox column anymore.

Brydom:
@ryanwilsonperkin happy to provide support on this if needed! Looking forward to this change.

brendanatshopify:
👀 Just a heads up that the Shop channel team is working on a solution to this at https://github.com/Shopify/polaris-react/pull/4376 that branches off @chloerice's work

PR #4348, opened 2021-07-29
[useIndexResourceState] Expose SelectionType enum

### LINKED ISSUE #4346: IndexTable deselection after selected resource removal
## Feature request summary
If you have a bulk action that deletes a resource from an index table, the list of selected resources is not updated when using the [`useIndexResourceState` hook](https://github.com/Shopify/polaris-react/blob/main/src/utilities/use-index-resource-state.ts). It would be nice to expose a callback that could be used to reset or provide a new selection state after some action.

## Rationale
Perhaps there's an established pattern for doing this that I've missed. My naive approach might look something like this:

```tsx
import {useState, useCallback} from 'react';
...
export function useIndexResourceState<T extends Record<string, unknown>>(
  resources: T[],
  {
    selectedResources: initSelectedResources = [],
    allResourcesSelected: initAllResourcesSelected = false,
    resourceIDResolver = defaultResourceIDResolver,
  }: {
    selectedResources?: string[];
    allResourcesSelected?: boolean;
    resourceIDResolver?: ResourceIDResolver<T>;
  } = {
    selectedResources: [],
    allResourcesSelected: false,
    resourceIDResolver: defaultResourceIDResolver,
  },
) {
  const [selectedResources, setSelectedResources] = useState(
    initSelectedResources,
  );
  const [allResourcesSelected, setAllResourcesSelected] = useState(
    initAllResourcesSelected,
  );

  const handleSelectionReset = (selectedResources: string[] = []) => {
    setSelectedResources(selectedResources);
    setAllResourcesSelected(false);
  };

  ...

  return {
    selectedResources,
    allResourcesSelected,
    handleSelectionChange,
    handleSelectionReset,
  };
}

```

CameronGorrie:

Closes: https://github.com/Shopify/polaris-react/issues/4346

Removing resources from an IndexTable requires updating the selection. There is a `handleSelectionChange` callback that can be used for doing so but we don't expose the `SelectionType` enum for TS consumers. Another approach could be to expose a separate callback for resetting the selection but this shouldn't be necessary, unless for some reason we don't want to expose the enum, in which case this approach makes sense.

Simply exports `SelectionType`

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -8,6 +8,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - Added `id` prop to `Layout` and `Heading` for hash linking ([#4307](https://github.com/Shopify/polaris-react/pull/4307))
 - Added `external` prop to `Navigation.Item` component ([#4310](https://github.com/Shopify/polaris-react/pull/4310))
+- Added `SelectionType` export from `useIndexResourceState` hook ([#4348](https://github.com/Shopify/polaris-react/pull/4348))
 
 ### Bug fixes
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -51,7 +51,10 @@ export {
 
 export {ScrollLockManagerContext as _SECRET_INTERNAL_SCROLL_LOCK_MANAGER_CONTEXT} from './utilities/scroll-lock-manager';
 export {WithinContentContext as _SECRET_INTERNAL_WITHIN_CONTENT_CONTEXT} from './utilities/within-content-context';
-export {useIndexResourceState} from './utilities/use-index-resource-state';
+export {
+  useIndexResourceState,
+  SelectionType,
+} from './utilities/use-index-resource-state';
 
 export {
   toCssCustomPropertySyntax as UNSTABLE_toCssCustomPropertySyntax,


gavin-hinchley:
@CameronGorrie This looks like a good change. I'm running into this issue now, wanting to call `handleSelectionChange()` but don't have an exportable definition for `SelectionType` to use and need to hardcode the enum value.
Any idea on if it can be shipped soon?

CameronGorrie:
I think the encapsulation approach [suggested here](https://github.com/Shopify/polaris-react/issues/4346#issuecomment-895501285) might be preferable, but I'm not sure. cc/ @AndrewMusgrave

AndrewMusgrave:
To be honest, it probably should have been a string union rather than an enum so we wouldn't have to deal with this type of issue!

I still think the encapsulation makes sense for certain situations, e.g. the one defined in the other issue, for resetting unavailable ids. And handling common use-cases. The `useIndexResourceState` is probably missing some callbacks that's causing the need to expose the enum. For example `clearSelectedIds ` should probably be added to the hook.

@gavin-hinchley What issue are you running into?

gavin-hinchley:
> What issue are you running into?

In our situation, we 
* click a bulk action on some selected items
*  open a confirmation model
*  once confirmed, we perform to bulk operation
*  when finished ensure no items are selected in the table. 

To unselect items we are calling:

`handleSelectionChange('all' as any, false);`

Perhaps there is a better way to unselect all items?

jtwee:
We're running into this same issue again. Is there any headway on which approach might move forward? (between this and [this](https://github.com/Shopify/polaris-react/issues/4346#issuecomment-895501285))

AndrewMusgrave:
@jtwee There's a [pull request in progress](https://github.com/Shopify/polaris/pull/5429) to handle de-selecting resources. If you need the changes asap you could probably implemented an effect yourself when your resources change & filter out the selected ids that no longer exist. And define the enum yourself.

```
enum SelectionType {
  All = 'all',
  Page = 'page',
  Multi = 'multi',
  Single = 'single',
}
```

AndrewMusgrave (approved):
There're a few more callback we'll probably want to add to the hook in the future, but this is good short term imo. Eventually we'll probably want to change it to a union type rather than enum to make this easier on consumers, but that'll be a breaking change and needs to happen in a major release 😓

PR #4349, opened 2021-07-29
Update to node 12.22.0 and react-testing v3.2.0

BPScott:

Preparing for futures around updating to support react 17 and building/testing/linting using sewing-ki-next. 

- Update `@shopify/react-testing` to v3.2.0. This adds support for React 17 so testing react 17 will be easier (along with other ongoing efforts in enzyme removal)
- Update node version to 12.22.0. Released on 2021-03-30, this is the first version of node where esmodules are considered stable. We're not relying on that yet and won't be for some time but our build tools will soon, and it's easier to make this change now so we're clear in the future. Strictly speaking `@shopify/react-testing` requires 12.14.0, but I figured we could jump a few more versions as it will unlock being able to leverage future sewing-kit-next improvements down the line without additional changes.


chloerice (approved):
🚀 🙏🏽

PR #4361, opened 2021-08-05
[IndexTable] Allow multiple triggers of `data-primary-link` when not navigating away

### LINKED ISSUE #4345: Using `data-primary-link` only works once when not redirecting to a different page
## Issue summary

The index table `data-primary-link` prop does not trigger a click/redirect after the first time. We think this is because of the `isNavigating.current` assignment on l101 of `Row.tsx`.

I'm not entirely sure of what the purpose is with `isNavigating`, but, it looks like it could almost be removed? Alternatively, it may just be able to be moved down below l110 after the if/else.

## Expected behavior

It should keep triggering after the first click.

## Actual behavior

It does not work after the first click.

1. Use `data-primary-link` setup in `IndexTable`
2. Click the link to add a URL param or use an `onClick` handler to stay on the same page
3. Link no longer works

## Reduced test case

https://codesandbox.io/s/unselectable-index-table-forked-utd6z

brydonm:

Fixes #4345  

This PR allows multiple triggers of `data-primary-link`. Previously, only one trigger would be allowed (assuming this was done as its main intention would be to navigate pages.) This change allows you to use an `onClick` with `data-primary-link` more than once.

### DIFF
--- a/src/components/IndexTable/components/Row/Row.tsx
+++ b/src/components/IndexTable/components/Row/Row.tsx
@@ -107,9 +107,10 @@ export const Row = memo(function Row({
         new MouseEvent(event.type, event.nativeEvent),
       );
     } else {
-      isNavigating.current = false;
       handleInteraction(event);
     }
+
+    isNavigating.current = false;
   };
 
   const RowWrapper = condensed ? 'li' : 'tr';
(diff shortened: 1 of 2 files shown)


alex-page (approved):
This looks good. I would love to see a test for this functionality.

@AndrewMusgrave has a lot more context if this could potentially break anything.

src/components/IndexTable/components/Row/Row.tsx
```diff
@@ -107,9 +107,10 @@ export const Row = memo(function Row({
         new MouseEvent(event.type, event.nativeEvent),
       );
     } else {
-      isNavigating.current = false;
```
AndrewMusgrave, line 110: @lhoffbeck iirc you made a change *some what* recently regarding `isNavigating`. Does this change still work with your use-case?

lhoffbeck: LGTM 👍. My change was to enable selecting by row when there isn't a `primaryLinkElement`, verified that @Brydom's change doesn't negatively affect the behavior.

Thanks for pinging @AndrewMusgrave!

TheSecurityDev:
I'm having this issue, and I'm wondering when this will be merged?

chloerice:
Closing this, as the primary link should navigate. There is a lack of support for interactive content that isn't the primary link because the Row swallows click events. I'll open a PR to fix that.

PR #4365, opened 2021-08-06
[IndexTable] Console deprecation warning when using `loading` prop with `React.StrictMode`

### LINKED ISSUE #4364: Console deprecation warning when using `loading` prop on `IndexTable`
1. Use `React.StrictMode`
2. Use `IndexTable` with `loading` to `true`

## Reduced test case

https://codesandbox.io/s/unselectable-index-table-forked-hytn0?file=/App.js

brydonm:

Fixes #4364 

Uses `nodeRef` instead of direct children with `CSSTransition` component.


AndrewMusgrave (approved):
Thanks for the contribution 🎉

PR #4369, opened 2021-08-09
[TextField] Fully removed clear button when there is no text

AndrewMusgrave:

This was noticed durning a bug hunt

Removing the complexity around changing the clear buttons state so the clear button won't be visible to screen readers when the field is empty

### Screenies
|Before|After|
|-|-|
|Screen Shot 2021-08-09 at 12 06 07 PM|Screen Shot 2021-08-09 at 11 38 29 AM|

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -18,7 +18,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Fixed a bug in `Banner` where loading state wasn't getting passed to `primaryAction` ([#4338](https://github.com/Shopify/polaris-react/pull/4338))
 - Fixed `Popover` not correctly positioning itself ([#4357](https://github.com/Shopify/polaris-react/pull/4357))
 - Fixed a bug `TextField` where Safari would render the incorrect text color ([#4344](https://github.com/Shopify/polaris-react/pull/4344))
-- Fixed screen readers reading out the clear button in `TextField` when when there is no input ([#4369](https://github.com/Shopify/polaris-react/pull/4369))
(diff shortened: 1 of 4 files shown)


src/components/TextField/TextField.scss
```diff
@@ -231,10 +231,6 @@ $stacking-order: (
   &:disabled {
     cursor: default;
   }
-
-  &.ClearButton-hidden {
-    @include visually-hidden;
-  }
```
AndrewMusgrave, line 237: Is there a reason we need to hide the button rather than remove it? We could always add `aria-hidden` if so.

UNRELEASED.md
```diff
@@ -18,6 +18,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Fixed a bug in `Banner` where loading state wasn't getting passed to `primaryAction` ([#4338](https://github.com/Shopify/polaris-react/pull/4338))
 - Fixed `Popover` not correctly positioning itself ([#4357](https://github.com/Shopify/polaris-react/pull/4357))
 - Fixed a bug `TextField` where Safari would render the incorrect text color ([#4344](https://github.com/Shopify/polaris-react/pull/4344))
+- Fixed screen readers reading out the clear button in `TextField` when when there is no input ([#4369](https://github.com/Shopify/polaris-react/pull/4369))
```
dleroux, line 21:
```suggestion
- Fixed screen readers reading out the clear button in `TextField` when there is no input ([#4369](https://github.com/Shopify/polaris-react/pull/4369))
```

dleroux (approved):

PR #4372, opened 2021-08-09
✨ [Popover] Now observes overlay attributes

beefchimi:
This PR replaces the previous effort: https://github.com/Shopify/polaris-react/pull/4303

We are now _always observing_ the Overlay `attributes`. We feel that the performance cost will not be that severe, and it is better to take this first step than to introduce another prop to the API.

Our main ask here is having control over how the `Popover` re-renders _(re-sizes/positions itself)_. Eventually, we think a better solution is to offer a mechanism to force a re-render. In the words of @clauderic

> ...the best way to support consumers that have complex use-cases would be to add an imperative API to force the popover to be re-measured from the consumer, something like:

```tsx
function ConsumerComponent() {
  const popoverRef = useRef();

  return (
    <>
      <Button onClick={forceMeasurePopover}>Click me</Button>
      <Popover ref={popoverRef} />
    </>
  );

  function forceMeasurePopover() {
    popoverRef.currrent.forceUpdateMeasurement();
  }
}
```


clauderic (approved):

beefchimi:
Closed in favour of: https://github.com/Shopify/polaris-react/pull/4385

PR #4374, opened 2021-08-10
[Nav/Item] Delegate onClick prop to UnstyledLink for secondaryAction

CSilivestru:

Related to https://github.com/Shopify/core-workflows/issues/26, we would like to instrument events in the admin nav to start tracking all clicks. Currently, we can do that fairly easily except for the case of using a `secondaryAction` on a `Nav/Item` component. To my knowledge, this is only used for the "View Storefront" icon but we may add more secondary actions in the future.

This allows a click handler to be passed in and used from the parent `Item` component when specifying a secondary action icon.

Prop chains to `UnstyledLink` when create the secondary action components.

### How to 🎩

- Run this Polaris version in local web
- Modify [this object] to contain an `onClick` prop with a function that does something
- Load web and click on the View Storefront icon. Notice the click handler executed

Please let me know if there's a better way to tophat this as I don't normally work in react :).
Also, the current documentation for `Item` does not specify prop requirements for `secondaryAction` so I didn't modify the docs to include this -- should this be updated?

### 🎩 checklist

- ~~[ ] Updated the component's `README.md` with documentation changes~~
- ~~[ ] [Tophatted documentation](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting%20documentation.md) changes in the style guide~~
- ~~[ ] For visual design changes, ping @ sarahill to update the Polaris UI kit~~

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -6,6 +6,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 ### Enhancements
 
+- Added optional `onClick` key to `secondaryAction` on `Nav/Item` component ([#4374](https://github.com/Shopify/polaris-react/pull/4374))
 - Added `id` prop to `Layout` and `Heading` for hash linking ([#4307](https://github.com/Shopify/polaris-react/pull/4307))
 - Added `external` prop to `Navigation.Item` component ([#4310](https://github.com/Shopify/polaris-react/pull/4310))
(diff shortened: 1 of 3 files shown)


kyledurand (commented):
This looks good and thanks for adding the missing tests. Just be sure to add this change to unreleased.md

CSilivestru:
Ah yes thanks for the reminder! Done :D

PR #4385, opened 2021-08-11
⚗️ [Popover] Expose imperative relayout function

beefchimi:
> **Warning!** This is currently a work-in-progress

This PR continues to try and find a solution to my struggles with getting `Popover` to re-render its size dimensions while toggling nested `Collapsible` components.

**There have been two previous attempts:**
1. https://github.com/Shopify/polaris-react/pull/4303
2. https://github.com/Shopify/polaris-react/pull/4372

The final attempt has ended in numerous Polaris tests _(which use `Popover` somewhere under the hood)_ stalling in what appears to be an endless re-render cycle. Something about the `attributes` MutationObserver is causing certain tests to never complete.

**Some additional conversation can be found here:**
https://shopify.slack.com/archives/C4Y8N30KD/p1628608394090200

Before I start fighting through each of these tests to attempt a resolution, I want us to first entertain this alternative approach.

### Solution

This was first suggested by @clauderic - which is to offer an imperative API which enables consumer's to dictate exactly when they might need to re-render the `Overlay` dimensions.

I have wrapped `Popover` with `forwardRef` and exposed a `forceReLayout()` function via `useImperativeHandle`. Consumer's can retrieve the `Popover` instance by way of `ref`, and call `popoverRef.current.forceReLayout();` whenever they need to update the dimensions.

This actually calls through multiple component layers to get at the code which handles updating measurements:
`Popover > PopoverOverlay > PositionedOverlay`.

It certainly is not the most elegant solution, but it hopefully leads us to a more explicit result. Rather than forcing the `MutationObserver` to watch for additional attributes and potentially incur _some_ ramification, we can put control in the hands of consumer's and request a layout update on command.

### Reviewers

I very much appreciate any review / advice reviewers can offer, as we are in desperate need of this functionality in Online Store. Some questions to consider:

- What do you think of this approach?
- What can I do to improve the solution?
- Are there pros / cons compared to the previous PR attempts?

Please tophat this PR using the following snippet:

<details>
<summary><strong>Playground snippet:</strong></summary>

```tsx
import React, {useCallback, useRef, useState} from 'react';

import {ActionList, Button, Collapsible, Page, Popover} from '../src';
import type {
  ActionListProps,
  ButtonProps,
  CollapsibleProps,
  PopoverHandles,
} from '../src';

interface CollapsibleSectionProps {
  id: string;
  title: string;
  items: ActionListProps['items'];
  open: CollapsibleProps['open'];
  onAction: ButtonProps['onClick'];
}

const POLARIS_COLLAPSIBLE_DISABLE_TRANSITION = {
  timingFunction: 'unset',
  duration: '1ms',
};

const popoverWrapperStyles = {
  width: 300,
  maxHeight: '50vh',
};

const pageItems: CollapsibleSectionProps['items'] = [
  {content: 'Import'},
  {content: 'Export'},
  {content: 'Something else'},
  {content: 'Shorter content'},
  {content: 'Spring loaded'},
  {content: 'Explore'},
  {content: 'Save'},
  {content: 'Delete'},
  {content: 'Very long thing'},
  {content: 'Final'},
];

const productItems: CollapsibleSectionProps['items'] = [
  {content: 'Another type of something else'},
  {content: 'Lots of stuff that gets written within this element'},
  {
    content:
      'It looks like when I use the fluidContent prop, I need to specify both a max-width and a max-height',
  },
  {
    content:
      'Otherwise, I will end up with a page full of Popover once it opens.',
  },
  {content: 'We need to provide sensible limits to our max Popover size'},
  {content: 'Final'},
];

  const popoverRef = useRef<PopoverHandles>(null);

  function layoutOverlay() {
    popoverRef.current?.forceReLayout();
  }

  const popoverActiveToggle = useCallback(
    [],
  );
  const popoverActiveClose = () => setPopoverActive(false);

  const popoverActivator = (
    <Button onClick={popoverActiveToggle} disclosure>
      More actions
    </Button>
  );

  const [openPages, setOpenPages] = useState(true);
  const toggleOpenPages = useCallback(() => {
    setOpenPages((openPages) => !openPages);
    layoutOverlay();
  }, []);

  const [openProducts, setOpenProducts] = useState(false);
  const toggleOpenProducts = useCallback(() => {
    setOpenProducts((openProducts) => !openProducts);
    layoutOverlay();
  }, []);

  return (
    <Page
      title="Prototype for Popover + Collapsible"
      subtitle="This code explores how we can use the Popover component to render dynamic content"
      primaryAction={{content: 'Primary action', disabled: true}}
    >
      <Popover
        ref={popoverRef}
        active={popoverActive}
        activator={popoverActivator}
        onClose={popoverActiveClose}
      >
        <div style={popoverWrapperStyles}>
          <CollapsibleSection
            id="PagesCollapsible"
            title="Pages"
            items={pageItems}
            open={openPages}
            onAction={toggleOpenPages}
          />

          <CollapsibleSection
            id="ProductsCollapsible"
            title="Products"
            items={productItems}
            open={openProducts}
            onAction={toggleOpenProducts}
          />
        </div>
      </Popover>
    </Page>
  );
}

function CollapsibleSection({
  id,
  title,
  items,
  open,
  onAction,
}: CollapsibleSectionProps) {
  return (
    <>
      <Button
        primary
        fullWidth
        textAlign="left"
        disclosure={open ? 'up' : 'down'}
        ariaControls={id}
        ariaExpanded={open}
        onClick={onAction}
      >
        {title}
      </Button>
      <Collapsible
        id={id}
        open={open}
        transition={POLARIS_COLLAPSIBLE_DISABLE_TRANSITION}
      >
        <ActionList items={items} />
      </Collapsible>
    </>
  );
}
```

</details>


src/components/Popover/Popover.tsx
```diff
@@ -83,28 +85,66 @@ export interface PopoverProps {
...
-}: PopoverProps) {
+};
+*/
+
+export interface PopoverHandles {
+  forceReLayout(): void;
+}
+
+interface PopoverSubcomponents {
+  // TODO: How can we make these required props?
```
beefchimi, line 100: This appears to be a difficult type to wrangle... and there major issues with me setting these to `optional (?)`, or is there another way I can type this to keep these subcomponents as required?

chloerice: Hey Curtis 👋🏽 I played around with this and ended up 🔥 the PopoverSubcomponents interface and trying [the second answer to this StackOverflow question](https://stackoverflow.com/questions/63453536/jsx-dot-notation-with-forwardref-typescript/68678190#68678190) and it works locally. 

```jsx
const PopoverNamespace = Object.assign(Popover, {Pane, Section});

export {PopoverNamespace as Popover};
```

I'm having trouble running `web` right now so I haven't been able to test on a consuming app (the comment on line 83 I'm guessing means this was a problem in the past 🤔 )

src/components/PositionedOverlay/PositionedOverlay.tsx
```diff
@@ -158,6 +158,14 @@ export class PositionedOverlay extends PureComponent<
     );
   }
 
+  forceReLayout() {
+    // Wait a single animation frame before re-measuring.
+    // Consumer's may also need to setup their own timers for
+    // triggering forceReLayout() `children` use animation.
+    // Ideally, forceReLayout() is fired at the end of a transition event.
+    requestAnimationFrame(() => this.handleMeasurement());
```
clauderic, line 166: nit:
```suggestion
    requestAnimationFrame(this.handleMeasurement);
```

beefchimi:
@clauderic + @chloerice thanks so much for the review!

I've addressed all of your feedback and I think this should be ready for final review 😄

src/components/Popover/Popover.tsx
```diff
@@ -78,128 +80,149 @@ export interface PopoverProps {
...
-        source === PopoverCloseSource.EscapeKeypress) &&
-      activatorNode
-    ) {
-      const focusableActivator =
-        findFirstFocusableNodeIncludingDisabled(activatorNode) ||
-        findFirstFocusableNodeIncludingDisabled(activatorContainer.current) ||
-        activatorContainer.current;
-      if (!focusNextFocusableNode(focusableActivator, isInPortal)) {
-        focusableActivator.focus();
+    useImperativeHandle(ref, () => {
```
beefchimi, line 122: For anyone wondering - I've barely changes this file, but since we now wrap with a `forwardRef`, the indentation/formatting got updated and therefor its hard to really parse exact lines of code that have changed 😭

PR #4386, opened 2021-08-11
Modernized tests - Scrollable, ScrollTo, ScrollLock, Select, SettingToggle, Sheet, Spinner, and Sticky components

rdott:

Updating existing tests to use the new modern framework and eliminate stale dependencies.

For the following component tests:

Scrollable
ScrollTo
ScrollLock
Select
SettingToggle
Sheet
Spinner
Sticky

Part of test modernization, updating tests using `{mountWithAppProvider} from 'test-utilities/legacy'` to `{mountWithApp} from 'test-utilities'`


src/components/Scrollable/tests/Scrollable.test.tsx
```diff
@@ -26,41 +24,39 @@ describe('<Scrollable />', () => {
...
-      <Scrollable>{children}</Scrollable>,
-    );
-    expect(scrollable.contains(children)).toBe(true);
+    const scrollable = mountWithApp(<Scrollable>{children}</Scrollable>);
+    expect(scrollable).toContainReactComponent('div', {children});
   });
 
   it('provides scrollToPosition callback to children', () => {
-    const Child: React.SFC = (_) => (
+    const Child: React.FunctionComponent = (_) => (
```
kpinzas-sh, line 32: TIL

src/components/Select/tests/Select.test.tsx
```diff
@@ -1,64 +1,72 @@
...
   });
 
   describe('onBlur()', () => {
     it('is called when the select is blurred', () => {
       const spy = jest.fn();
-      const element = mountWithAppProvider(
+      const element = mountWithApp(
         <Select label="Select" options={[]} onBlur={spy} onChange={noop} />,
       );
-      element.find('select').simulate('focus');
```
kpinzas-sh, line 45: Here is not necessary to call onFocus?

kpinzas-sh (approved):
Looks good to me! I just left a question and I see that the CI is complaining around conflicts because this merge fest that we are having these days 🤪

src/components/Select/tests/Select.test.tsx
rdott: I don't think it's needed to focus on the event as we're just testing the callback when onBlur is triggered.

PR #4394, opened 2021-08-12
Allow subnav items to execute provided onClick handlers

CSilivestru:

Similar to https://github.com/Shopify/polaris-react/pull/4374, the sub-navigation items accept an `onClick` property but do not actually use it. This PR allows a consumer to provide a custom click handler that will be executed when a subnav item is clicked. 

This PR is to enable the subnav item to first execute `onNavigationDismiss` and then any other provided click handler, similar to the [Item click handler](https://github.com/Shopify/polaris-react/blob/main/src/components/Navigation/components/Item/Item.tsx#L336)

## 

### How to 🎩

- Run this Polaris version in local web
- Modify a subnavigation item to contain an onClick prop with a function that does something
- Load web and click on that subnav item. Notice the click handler is executed

### 🎩 checklist


src/components/Navigation/components/Item/tests/Item.test.tsx
```diff
@@ -279,6 +279,73 @@ describe('<Nav.Item />', () => {
...
+        .find(Secondary)!
+        .find('a')!
+        .trigger('onClick', {
+          preventDefault: jest.fn(),
+          currentTarget: {
+            getAttribute: () => '/admin/draft_orders',
+          },
+        });
+
+      expect(spy).toHaveBeenCalled();
```
kyledurand, line 313: Might be worth using `toHaveBeenCalledTimes(1)` here. I think nav item is recursive in rendering itself as a sub nav item.

kyledurand (approved):
Minor comments on testing but I think they're worthwhile changes

voiid (approved):
LGTM. Do you know if there's impact for this change in web?

PR #4402, opened 2021-08-17
Migrate to use sewing-kit-next for lint and test runs

BPScott:
(creating a draft for the moment, I still need to fixup a few tests and confirm build/storybook still works as expected before this is ready for review)

sewing-kit was never great for libraries. sewing-kit-next is better. 

Eventually I want to use skn for build output (it uses rollup under the hood same as polaris's existing build), but the first step to that is configuring skn for test and linting runs.

Replace sewing-kit with sewing-kit-next.

Under the hood, we've updated from jest 25 to 27.
We also updated from babel 7.12 to 7.15, and @shopify/babel-plugin now uses the latest version.

Currently skn does not provide out-the-box config jest transformers for scss and image files, so create them ourselves for now.

### How to 🎩

- See that lint and tests still pass.
- Check that build output remains the same as before.
- Confirm storybook still loads


alex-page (approved):
Awesome work @BPScott

PR #4413, opened 2021-08-19
[Test Modernization] Switch to using mountWithApp on ColorPicker tests.

lucabezerra:

Modernizes tests to use the new modern framework (`{mountWithApp}` from `test-utilities`).

Updates tests using `{mountWithAppProvider}` from `test-utilities/legacy` to use `{mountWithApp}` from `test-utilities`

### DIFF
--- a/src/components/ColorPicker/tests/ColorPicker-ssr.test.tsx
+++ b/src/components/ColorPicker/tests/ColorPicker-ssr.test.tsx
@@ -22,7 +22,7 @@ describe('<ColorPicker /> Server-side only', () => {
       <ColorPicker color={red} onChange={noop} />,
     );
 
-    colorPicker!.find(Slidable)!.find('div')!.trigger('onMouseDown', {
+    colorPicker.find(Slidable)!.find('div')!.trigger('onMouseDown', {
       type: 'mousedown',
       clientX: 1,
       clientY: 1,
--- a/src/components/ColorPicker/tests/ColorPicker-ssr.test.tsx
+++ b/src/components/ColorPicker/tests/ColorPicker-ssr.test.tsx
@@ -1,6 +1,5 @@
 import React from 'react';
-// eslint-disable-next-line no-restricted-imports
-import {mountWithAppProvider} from 'test-utilities/legacy';
+import {mountWithApp} from 'test-utilities';
 
 import {Slidable} from '../components';
 import {ColorPicker} from '../ColorPicker';
@@ -19,11 +18,15 @@ jest.mock('../../../utilities/target', () => ({
 
 describe('<ColorPicker /> Server-side only', () => {
   it('does not attach the touchmove handler to the window', () => {
-    const colorPicker = mountWithAppProvider(
+    const colorPicker = mountWithApp(
       <ColorPicker color={red} onChange={noop} />,
     );
 
-    colorPicker.find(Slidable).first().simulate('mousedown');
+    colorPicker!.find(Slidable)!.find('div')!.trigger('onMouseDown', {
+      type: 'mousedown',
+      clientX: 1,
+      clientY: 1,
+    });
 
     const touch = {clientX: 0, clientY: 0};
     const event = new TouchEvent('touchmove', {
--- a/src/components/ColorPicker/tests/ColorPicker.test.tsx
+++ b/src/components/ColorPicker/tests/ColorPicker.test.tsx
@@ -26,9 +26,10 @@ describe('<ColorPicker />', () => {
           <ColorPicker color={red} onChange={spy} />,
         );
 
-        colorPicker
-          .findAll(Slidable)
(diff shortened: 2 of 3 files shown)


src/components/ColorPicker/tests/ColorPicker-ssr.test.tsx
```diff
@@ -19,11 +18,15 @@ jest.mock('../../../utilities/target', () => ({
 
 describe('<ColorPicker /> Server-side only', () => {
   it('does not attach the touchmove handler to the window', () => {
-    const colorPicker = mountWithAppProvider(
+    const colorPicker = mountWithApp(
       <ColorPicker color={red} onChange={noop} />,
     );
 
-    colorPicker.find(Slidable).first().simulate('mousedown');
+    colorPicker!.find(Slidable)!.find('div')!.trigger('onMouseDown', {
```
BPScott, line 25: colorPicker shall always be defined and will never be null so you don't need the first  `!` to manually assert that.

```suggestion
    colorPicker.find(Slidable)!.find('div')!.trigger('onMouseDown', {
```

src/components/ColorPicker/tests/ColorPicker.test.tsx
```diff
@@ -13,26 +12,31 @@ const red = {
...
         const spy = jest.fn();
-        mountWithAppProvider(<ColorPicker color={red} onChange={spy} />)
-          .find(Slidable)
-          .at(SlidableType.BrightnessSaturation)
-          .simulate('mousedown');
+        const colorPicker = mountWithApp(
+          <ColorPicker color={red} onChange={spy} />,
+        );
+
+        colorPicker.findAll(Slidable)[SlidableType.Color].trigger('onChange', {
```
BPScott, line 29: The enum feels really like a really weird use here. What do you think to removing it and using plain old numbers for indexes?

```suggestion
        colorPicker.findAll(Slidable)[0].trigger('onChange', {
```

src/components/ColorPicker/tests/ColorPicker.test.tsx
```diff
@@ -91,46 +124,50 @@ describe('<ColorPicker />', () => {
...
 
-      colorPicker
-        .find(Slidable)
-        .at(SlidableType.BrightnessSaturation)
-        .simulate('mousedown');
-
-      const mouseMoveListener = colorPicker
-        .find(EventListener)
-        .filterWhere((listener) => listener.prop('event') === 'mousemove');
+      colorPicker!.find(Slidable)!.find('div')!.trigger('onMouseDown', {
```
BPScott, line 131: No need for that first non-null assertion as colorPicker will never be null

```suggestion
      colorPicker.find(Slidable)!.find('div')!.trigger('onMouseDown', {
```

src/components/ColorPicker/tests/ColorPicker.test.tsx
```diff
@@ -91,46 +124,50 @@ describe('<ColorPicker />', () => {
...
 
-      colorPicker
-        .find(Slidable)
-        .at(SlidableType.BrightnessSaturation)
-        .simulate('mousedown');
-
-      const touchMoveListener = colorPicker
-        .find(EventListener)
-        .filterWhere((listener) => listener.prop('event') === 'touchmove');
+      colorPicker!.find(Slidable)!.find('div')!.trigger('onMouseDown', {
```
BPScott, line 149:
```suggestion
      colorPicker.find(Slidable)!.find('div')!.trigger('onMouseDown', {
```

src/components/ColorPicker/tests/ColorPicker.test.tsx
```diff
@@ -91,46 +124,50 @@ describe('<ColorPicker />', () => {
...
     });
 
     it('prevents default for touchmove events if dragging the slider', () => {
-      const colorPicker = mountWithAppProvider(
+      const colorPicker = mountWithApp(
         <ColorPicker color={red} onChange={noop} />,
       );
 
-      colorPicker.find(Slidable).first().simulate('mousedown');
+      colorPicker!.find(Slidable)!.find('div')!.trigger('onMouseDown', {
```
BPScott, line 166:
```suggestion
      colorPicker.find(Slidable)!.find('div')!.trigger('onMouseDown', {
```

BPScott (commented):
Nice!

What do you think to removing that enum and using plain numbers for indexes?

There's a few places where you're using `!` to assert that colorPicker won't be null but typescript knows that already so you can remove it

lucabezerra:
> Nice!
> 
> What do you think to removing that enum and using plain numbers for indexes?
> 
> There's a few places where you're using `!` to assert that colorPicker won't be null but typescript knows that already so you can remove it

@BPScott Thanks for the comments! I didn't like the enum either, but at the same time I thought it had some value in terms of giving names to indexes to make it clearer what they represented. I've removed it anyways :)

BPScott (approved):
:shipit:

PR #4414, opened 2021-08-19
Update ts and storybook

kyledurand:


BPScott:
Looks like @lucabezerra just modernised the icon/image tests too in https://github.com/Shopify/polaris-react/pull/4418. Perhaps we merge that PR, then rebase this?

src/components/Icon/tests/Icon.test.tsx
```diff
@@ -22,7 +27,7 @@ describe('<Icon />', () => {
 
     it('renders a React Element when source is given a React Stateless Functional Component', () => {
       const element = mountWithApp(<Icon source={PlusMinor} />);
-      expect(element).toContainReactComponentTimes(PlusMinor, 1);
+      expect(element).toHaveReactProps({source: PlusMinor});
```
BPScott, line 30: I think the original line is the right intent. `expect(element).toHaveReactProps` asserts what we passed into the element (i.e. what we did on the line above), but doesn't test anything about the internals of Icon

BPScott (approved):
I haven't ran storybook locally as I'm in the middle of a load of other stuff but if CI is happy then I'm happy

Add a changelog entry for the Icon behaviour change, address the test comment down below then :shipit:

PR #4424, opened 2021-08-24
sewing-kit-next build

BPScott:

Use sewing-kit-next for our build process, rather than having a bespoke rollup config

Being a showcase for the modern (albeit getting the breaking changes in as part of a major version bump.
Migrating from a bespoke rollup build to use sewing-kit-next to perform the build.

Parts of this are still a bit lumpy but will be able to be improved without any breaking change in the near future.

This PR replaces our bespoke rollup build with using sewing-kit-next (which still uses rollup under the hood). This doesn't look too different at the moment - we still need bespoke code for css compilation for instance and rollup plugins for image handliing, but in the near future this code shall be extracted into a reusable skn plugin that configures rollup in an opinionated way.

Previously package content was built into `dist`, and temp, cache and storybook output was placed into `build`. The sewing-kit-next plugin we're using is opinionated and places package content into the build folder. Thus we have to find somewhere else for those temp/cache/storybook files to live. The following changes have been made:

- The temp, cache and storybook files folder moves from `/build` to `/build-internal`
- The package content folder moves from `/dist` to `/build`

This change of the package content location means that consumers who previously imported styles.css now need to use the new path

```diff
- @import '@shopify/polaris/dist/styles.css';
+ @import '@shopify/polaris/build/esm/styles.css';
```


scripts/build.js
```diff
@@ -4,24 +4,8 @@ const {resolve} = require('path');
...
-// Run a Rollup build to generate JS and styles
-run(`yarn run rollup -c`);
-
-// Copy documentation into the docs folder
-run(`yarn run copyfiles './src/**/*.md' './dist/docs' --up=1`);
-
-// Copy sass files that make up our public API into the styles folder
-run(`yarn run copyfiles './src/styles/**/*.scss' './dist/styles' --up=2`);
+// Run Sk build to generate everything
+run(`yarn run skn build`);
```
alex-page, line 8: Love this clean up with the `preAndPostBuildPlugin`

BPScott:
Running into some surprises with the style.css load order being changed around, I'll be digging into that, but the rest is seems to be good

kyledurand (approved):
Love this Ben! Thanks for walking me through it all ❤️

PR #4426, opened 2021-08-24
use setup-node's built-in caching

BPScott:

A few weeks ago setup-node gained the ability to handle caching of node_modules content on its own instead of requiring users to do it by configuring the cache. See https://github.com/actions/setup-node#caching-packages-dependencies

Let's use that instead to make our CI configuration simpler.

Replace actions/cache with using actions/setup-node's `cache` option

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,18 +12,7 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: '10.24.0'
-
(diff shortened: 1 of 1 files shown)


alex-page (approved):

BPScott:
You can see this working by looking at the CI for the second commit and seeing that it has found a cache and has restored it https://github.com/Shopify/polaris-react/pull/4426/checks?check_run_id=3407938631#step:3:23

PR #4434, opened 2021-08-25
 [Test Modernization] Modernizing Filters and Form tests.

lucabezerra:
# Disclaimer

~There is one test in the `Filters.test.tsx` suite that wasn't modernized. It seems like we've found a bug on `react-testing` regarding using `.find()` to search for components in a tree that contains React Fragments (it can't find the component if it's a few levels deep). Decided to submit the PR anyways as all the other tests have been modernized.~

UPDATE: disclaimer no longer needed, as the aforementioned suite has already been modernized here: https://github.com/Shopify/polaris-react/pull/4458

Modernizes tests to use the new modern framework (`{mountWithApp}` from `test-utilities`).

Updates tests using `{mountWithAppProvider}` from `test-utilities/legacy` to use `{mountWithApp}` from `test-utilities`


alex-page (approved):
@lucabezerra do we have an issue or PR to address the `find` fix? I don't want to forget about this legacy test. Happy to merge this in its current state.

PR #4476, opened 2021-09-17
[TextField] Add maxHeight prop

dantecatalfamo:

The TextField allows multiline values, but if a long input is created, the component can become extremely tall. `maxHeight` gives developers an option to cap the height of the element.

I'm creating a page that needs to have a transcript pasted into a textfield, it's possibly very long and required a lot of scrolling to get to the bottom of the form. If I could limit the size of the TextForm component that wouldn't be a problem.

Adds a `maxHeight` prop to the TextField component, which adds a `max-height` property to the component's inline style. It only appears if `multiline` is present, as it's not required otherwise.

image


chloerice (approved):
Thanks for contributing @dantecatalfamo!! 🚀

dantecatalfamo:
Thanks for the fixup!

PR #4502, opened 2021-09-30
Migration guide updates

BPScott:

Improving the migration guide

Updating the migration guide

- Group build changes together
- Explicitly call out the styles.css and public API as the things people
  configure rather than talking about a whole directory
- Remove mentions of cjs/esm/types output as nobody should be referencing
  those deep file paths anyway
- Use the correct version of the react peer dependency and fix the
  reason for the peer dep bump - as it was the reason for the 16.8 bump in v4.


voiid (approved):

kyledurand (approved):
✨ Thanks Ben

PR #4505, opened 2021-10-07
[ActionList] add arrow navigation and menu role support

### LINKED ISSUE #4419: [Popover] Content not discoverable via screen reader browse mode
## Issue summary

The revealed content for popover components appear visually next to the activator control. However, the content is dynamically injected at the bottom of the DOM, outside of the `main` region landmark.

As a result, when using [virtual/browse navigation mode](https://tink.uk/understanding-screen-reader-interaction-modes/), screen reader users are unable to find the revealed content. For example, when using JAWS or NVDA on Windows, using the ⬇️ key to move to the popover content moves the virtual cursor to the next item on the screen, not the popover content as expected.

In addition, the current implementation includes shifting focus to a `div` container without a label which creates a "group." The user then needs to move "into" the group container to discover content. Some environments miss conveying this detail which results in content being skipped over on ⬇️ key press.

Another addition, <kbd>Esc</kbd> seems to be partially supported. When the popover is dismissed, content is removed from view, yet focus shifts to another area of the screen. This would be unexpected.

<details>
  <summary>Screenshots</summary>

Admin draft order screen. Collect payment button is highlighted with its popover content revealed.

</details>

#### HTML

```html
<main>
  <!-- … -->

  <button
    class="Polaris-Button_r99lw Polaris-Button--primary_7k9zs"
    type="button"
    aria-controls="Polarispopover75"
    aria-owns="Polarispopover75"
    aria-expanded="false"
    tabindex="0"
  >
    <span class="Polaris-Button__Content_xd1mk">
      <span class="Polaris-Button__Text_yj3uv">Collect payment</span>
      <!-- … -->
    </span>
  </button>

  <!-- … -->
</main>

<!-- … -->

<div
  class="
    Polaris-PositionedOverlay_2y1hl
    Polaris-Popover__PopoverOverlay_emiux
    Polaris-Popover__PopoverOverlay--open_4ccww
  "
  style="top: 784px; left: 1065.92px; width: 155.906px"
>
  <!-- … -->
  <div
    id="Polarispopover78"
    tabindex="-1"
    class="Polaris-Popover__Content_xd1mk"
    style="height: 98px"
  >
    <!-- … -->
    <div class="Polaris-ActionList_z4041">
      <div class="Polaris-ActionList__Section--withoutTitle_1vpsz">
        <ul class="Polaris-ActionList__Actions_17m67">
          <li>
            <button type="button" class="Polaris-ActionList__Item_yiyol">
              <span class="Polaris-ActionList__Content_xd1mk">
                <span class="Polaris-ActionList__Text_yj3uv">
                  Pay by credit card
                </span>
              </span>
            </button>
          </li>
          <!-- … -->
        </ul>
      </div>
    </div>
    <!-- … -->
  </div>
  <!-- … -->
</div>
```

## Steps to reproduce

1. Go to the [draft order page](https://shop1.myshopify.io/admin/draft_orders/new) with any supported browser
2. With an active screen reader (JAWS or NVDA for example), use the virtual cursor to navigate to and activate the "Collect payment" activator control
3. Use the ⬇️ key to navigate to the revealed content

### Expected

- ⬇️ key to move the virtual cursor to the "Pay by credit card" control.

### Actual

- Virtual cursor moves to the "Find or create a customer" heading.

## Recommendation

1. Adjust the position of the dynamically injected content to be **below** the activator control in the DOM. With this change, ⬇️ key will move the virtual cursor position to the newly revealed content as expected.
2. On activation **do not shift focus.** The disclosure pattern expects focus to remain on the activator. From this point the user is able to press ⬇️ key and discover content organically. (Remove the `tabindex` attribute from `div#Polarispopover78` container as this is no longer required.)
3. On popover dismiss via <kbd>Esc</kbd> or other means, shift focus back to the activator control. From here the user will be able to navigate forward as expected.

Review the [APG 1.1 Disclosure](https://www.w3.org/TR/wai-aria-practices-1.1/#disclosure) pattern for more details on expected keyboard interaction and screen reader support.

#### HTML

```html
<main>
  <!-- … -->

  <button
    class="Polaris-Button_r99lw Polaris-Button--primary_7k9zs"
    type="button"
    aria-controls="Polarispopover75"
    aria-owns="Polarispopover75"
    aria-expanded="false"
    tabindex="0"
  >
    <span class="Polaris-Button__Content_xd1mk">
      <span class="Polaris-Button__Text_yj3uv">Collect payment</span>
      <!-- … -->
    </span>
  </button>
  <div
    class="
      Polaris-PositionedOverlay_2y1hl
      Polaris-Popover__PopoverOverlay_emiux
      Polaris-Popover__PopoverOverlay--open_4ccww
    "
    style="top: 784px; left: 1065.92px; width: 155.906px"
  >
    <!-- … -->
    <div
      id="Polarispopover78"
      class="Polaris-Popover__Content_xd1mk"
      style="height: 98px"
    >
      <!-- … -->
      <div class="Polaris-ActionList_z4041">
        <div class="Polaris-ActionList__Section--withoutTitle_1vpsz">
          <ul class="Polaris-ActionList__Actions_17m67">
            <li>
              <button type="button" class="Polaris-ActionList__Item_yiyol">
                <span class="Polaris-ActionList__Content_xd1mk">
                  <span class="Polaris-ActionList__Text_yj3uv">
                    Pay by credit card
                  </span>
                </span>
              </button>
            </li>
            <!-- … -->
          </ul>
        </div>
      </div>
      <!-- … -->
    </div>
    <!-- … -->
  </div>

  <!-- … -->
</main>
```

<!-- #### Diff

```diff
- line removed
+ line added
``` -->

## Specifications

- **Browser:** Any
- **OS:** Any
- **WCAG Principal:** Operable
- **WCAG SC:** [2.1.1 Keyboard](https://www.w3.org/WAI/WCAG21/Understanding/keyboard.html) – Level A
- **Severity:** High
- **Effort:** Unknown

### PR DESCRIPTION
**Pull request by patrickracicot**

Fixes #4419 

The goal of this PR is to address the accessibility issue concerning the `ActionList` component used inside the `Popover` component..  Screen readers are unable to find, navigate into and trigger actions in the revealed content.

In this proposal, we enhance the `ActionList` to be [WAI-Aria compliant](https://www.w3.org/TR/wai-aria-practices-1.1/examples/menu-button/menu-button-actions.html).

#### **Navigation**
To do so we start by properly support arrow (⬆️  & ⬇️ ) navigation as well as navigation wrap (which consist of bringing the focus from first to last and last to first action item instead of leaving the content context on arrow navigation).

#### **Role attribution**
We also enhance the `ActionList` component usage of the `actionRole` prop.  If a consumer sets this prop to the value `menuitem`, the component `ActionList`, `Section` and `Item` will automatically set the proper `menu` and `menuitem` roles.

#### **On-open Focus**
In order to have the screen readers properly announce the context to the user, we are adding `tabIndex="-1"` to the root `<ul />` element.  This will ensure that `autoFocusTarget="first-node"` will set the focus to this root `<ul />` element which will in-turn allow the screen reader to properly announce to the user that the context is now on a list.

In the case of **multi-section**, the tabIndex and menu role needs to be assigned on the `<ul />` rendered in the `Actionlist` component. https://github.com/Shopify/polaris-react/blob/c463b45e226fa6ea90e91ddc825a8fc48ecaf352/src/components/ActionList/ActionList.tsx#L96-L106

While also making sure that we assign the `presentation` role to sub-section `<ul />` elements (otherwise screen readers will sub-divise the count for each item instead of having one unified count).
https://github.com/Shopify/polaris-react/blob/c463b45e226fa6ea90e91ddc825a8fc48ecaf352/src/components/ActionList/components/Section/Section.tsx#L67-L76

For more details refer to the following [document](https://docs.google.com/document/d/1pkB83BwEmf_qKU3XM37mL8hQ5FOYkLSk4TwANdQDC0c/edit#)

**NOTE**
In order to keep the size of this PR small, the scope of this PR is only on the `ActionList` implementation and `Popover` documentation.  Once this is approved, we will need to revisit other Polaris components that uses an ActionList inside a Popover (for example `MenuGroup`.

## 

### How to 🎩

<details>

```jsx

import {ActionList, Button, Page, Popover} from '../src';

  const toggleActive = useCallback(() => setActive((active) => !active), []);

  const handleImportedAction = useCallback(
    () => console.log('Imported action'),
    [],
  );

  const handleExportedAction = useCallback(
    () => console.log('Exported action'),
    [],
  );

  const activator = (
    <Button onClick={toggleActive} disclosure>
      More actions
    </Button>
  );

  return (
    <Page>
        <Popover
          active={active}
          activator={activator}
          autofocusTarget="first-node"
          onClose={toggleActive}
        >
          <ActionList
            actionRole="menuitem"
            items={[
              {
                content: 'Import file',
                onAction: handleImportedAction,
              },
              {
                content: 'Export file',
                onAction: handleExportedAction,
              },
            ]}
          />
        </Popover>
      </div>
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -9,12 +9,17 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 - Removed animtion from `Skeleton` components ([#4697](https://github.com/Shopify/polaris-react/pull/4697))
 - Remove duplicate duration(fast) usage. ([#4682](https://github.com/Shopify/polaris-react/pull/4682))
 - Updated the accessability label for the rollup actions in the `Page` header ([#4080](https://github.com/Shopify/polaris-react/pull/4080))
+- Keyboard arrow navigation support added in `ActionList` ([#4505](https://github.com/Shopify/polaris-react/pull/4505))
+- Menu role attribute value support added in `ActionList/Section` ([#4505](https://github.com/Shopify/polaris-react/pull/4505))
 
 ### Bug fixes
 
 - Centered full width `Popover` on small viewports ([#4114](https://github.com/Shopify/polaris-react/pull/4114))
 
 ### Documentation
+
+- Added arrow navigation instructions in keyboard support for `ActionList` ([#4505](https://github.com/Shopify/polaris-react/pull/4505))
+- Updated examples to properly support JAWS screen reader for `Popover` and `ActionList` ([#4505](https://github.com/Shopify/polaris-react/pull/4505))
 
 ### Development workflow
 
--- a/src/components/ActionList/ActionList.tsx
+++ b/src/components/ActionList/ActionList.tsx
@@ -1,6 +1,11 @@
-import React from 'react';
+import React, {useRef} from 'react';
 
-import type {ActionListItemDescriptor, ActionListSection} from '../../types';
+import {
+  wrapFocusNextFocusableMenuItem,
+  wrapFocusPreviousFocusableMenuItem,
+} from '../../utilities/focus';
+import {KeypressListener} from '../KeypressListener';
+import {ActionListItemDescriptor, ActionListSection, Key} from '../../types';
 import {classNames} from '../../utilities/css';
 
 import {Section} from './components';
@@ -12,7 +17,7 @@ export interface ActionListProps {
   /** Collection of sectioned action items */
   sections?: readonly ActionListSection[];
   /** Defines a specific role attribute for each action in the list */
-  actionRole?: string;
+  actionRole?: 'menuitem' | string;
   /** Callback when any item is clicked or keypressed */
   onActionAnyItem?: ActionListItemDescriptor['onAction'];
 }
@@ -24,6 +29,7 @@ export function ActionList({
   onActionAnyItem,
 }: ActionListProps) {
   let finalSections: readonly ActionListSection[] = [];
+  const actionListRef = useRef<HTMLDivElement & HTMLUListElement>(null);
 
   if (items) {
     finalSections = [{items}, ...sections];
@@ -35,6 +41,11 @@ export function ActionList({
 
   const hasMultipleSections = finalSections.length > 1;
   const Element = hasMultipleSections ? 'ul' : 'div';
+  const elementRole =
+    hasMultipleSections && actionRole === 'menuitem' ? 'menu' : undefined;
+  const elementTabIndex =
+    hasMultipleSections && actionRole === 'menuitem' ? -1 : undefined;
+
   const sectionMarkup = finalSections.map((section, index) => {
     return section.items.length > 0 ? (
       <Section
@@ -48,5 +59,55 @@ export function ActionList({
     ) : null;
   });
 
-  return <Element className={className}>{sectionMarkup}</Element>;
+  const handleFocusPreviousItem = (evt: KeyboardEvent) => {
+    evt.preventDefault();
+    if (actionListRef.current && evt.target) {
+      if (actionListRef.current.contains(evt.target as HTMLElement)) {
+        wrapFocusPreviousFocusableMenuItem(
+          actionListRef.current,
+          evt.target as HTMLElement,
+        );
+      }
+    }
+  };
+
+  const handleFocusNextItem = (evt: KeyboardEvent) => {
+    evt.preventDefault();
+    if (actionListRef.current && evt.target) {
+      if (actionListRef.current.contains(evt.target as HTMLElement)) {
+        wrapFocusNextFocusableMenuItem(
+          actionListRef.current,
+          evt.target as HTMLElement,
+        );
+      }
+    }
+  };
+
+  const listeners =
+    actionRole === 'menuitem' ? (
+      <>
+        <KeypressListener
+          keyEvent="keydown"
+          keyCode={Key.DownArrow}
+          handler={handleFocusNextItem}
+        />
+        <KeypressListener
+          keyEvent="keydown"
+          keyCode={Key.UpArrow}
+          handler={handleFocusPreviousItem}
+        />
+      </>
+    ) : null;
+
+  return (
+    <Element
+      ref={actionListRef}
+      className={className}
+      role={elementRole}
+      tabIndex={elementTabIndex}
+    >
+      {listeners}
+      {sectionMarkup}
+    </Element>
+  );
 }
--- a/src/components/ActionList/README.md
+++ b/src/components/ActionList/README.md
@@ -108,8 +108,14 @@ function ActionListInPopoverExample() {
 
   return (
     <div style={{height: '250px'}}>
-      <Popover active={active} activator={activator} onClose={toggleActive}>
+      <Popover
+        active={active}
+        activator={activator}
+        autofocusTarget="first-node"
+        onClose={toggleActive}
+      >
         <ActionList
+          actionRole="menuitem"
           items={[
             {
               content: 'Import file',
@@ -145,8 +151,14 @@ function ActionListWithMediaExample() {
 
   return (
     <div style={{height: '200px'}}>
-      <Popover active={active} activator={activator} onClose={toggleActive}>
+      <Popover
+        active={active}
+        activator={activator}
+        autofocusTarget="first-node"
+        onClose={toggleActive}
+      >
         <ActionList
+          actionRole="menuitem"
           items={[
             {content: 'Import file', icon: ImportMinor},
             {content: 'Export file', icon: ExportMinor},
@@ -176,14 +188,20 @@ function ActionListWithSuffixExample() {
 
   return (
     <div style={{height: '200px'}}>
-      <Popover active={active} activator={activator} onClose={toggleActive}>
+      <Popover
+        active={active}
+        activator={activator}
+        autofocusTarget="first-node"
+        onClose={toggleActive}
+      >
         <ActionList
+          actionRole="menuitem"
           items={[
             {
+              active: true,
               content: 'Import file',
               icon: ImportMinor,
               suffix: <Icon source={TickSmallMinor} />,
-              active: true,
             },
             {content: 'Export file', icon: ExportMinor},
           ]}
@@ -212,8 +230,14 @@ function SectionedActionListExample() {
 
   return (
     <div style={{height: '250px'}}>
-      <Popover active={active} activator={activator} onClose={toggleActive}>
+      <Popover
+        active={active}
+        activator={activator}
+        autofocusTarget="first-node"
+        onClose={toggleActive}
+      >
         <ActionList
+          actionRole="menuitem"
           sections={[
             {
               title: 'File options',
@@ -255,13 +279,23 @@ function ActionListWithDestructiveItemExample() {
 
   return (
     <div style={{height: '250px'}}>
-      <Popover active={active} activator={activator} onClose={toggleActive}>
+      <Popover
+        active={active}
+        activator={activator}
+        autofocusTarget="first-node"
+        onClose={toggleActive}
+      >
         <ActionList
+          actionRole="menuitem"
           sections={[
             {
               title: 'File options',
               items: [
-                {content: 'Import file', icon: ImportMinor, active: true},
+                {
+                  active: true,
+                  content: 'Import file',
+                  icon: ImportMinor,
+                },
                 {content: 'Export file', icon: ExportMinor},
                 {
                   destructive: true,
@@ -296,8 +330,14 @@ function ActionListWithHelpTextExample() {
 
   return (
     <div style={{height: '250px'}}>
-      <Popover active={active} activator={activator} onClose={toggleActive}>
+      <Popover
+        active={active}
+        activator={activator}
+        autofocusTarget="first-node"
+        onClose={toggleActive}
+      >
         <ActionList
+          actionRole="menuitem"
           sections={[
             {
               items: [
@@ -328,6 +368,7 @@ function ActionListWithPrefixSuffixExample() {
   return (
     <div style={{height: '250px', maxWidth: '350px'}}>
       <ActionList
+        actionRole="menuitem"
         items={[
           {
             content: 'Go here',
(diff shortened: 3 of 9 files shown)


UNRELEASED.md
```diff
@@ -15,18 +15,25 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
...
 
 ### Bug fixes
 
 - Fixed try-catch syntax error in `Modal` ([#4553](https://github.com/Shopify/polaris-react/pull/4553))
 - Fixed an issue with `TextField` where date and time were uneditable on click ([#4671](https://github.com/Shopify/polaris-react/pull/4671))
 - Fixed an issue with `Popover` where the transform property interfered with descendants positioning ([#4685](https://github.com/Shopify/polaris-react/pull/4685))
 - Fixed screen reader accessibility issue of the `Checkbox` component ([#4631](https://github.com/Shopify/polaris-react/pull/4631))
+- Properly support JAWS screen reader for `ActionList` ([#4505](https://github.com/Shopify/polaris-react/pull/4505))
+
+### Documentation
```
NathanPJF:
```suggestion
```
Duplicate title

src/components/ActionList/README.md
```diff
@@ -388,6 +429,7 @@ Items in an action list are organized as list items (`<li>`) in an unordered lis
 ### Keyboard support
 
 - Give the action list items keyboard focus with the <kbd>tab</kbd> key (or <kbd>shift</kbd> + <kbd>tab</kbd> when tabbing backwards)
+- Navigate through the action list items with <kbd>down arrow</kbd> (<kbd>up arrow</kbd> to move backwards)
```
NathanPJF:
```suggestion
- When action list items have a role of `menuitem`, navigate through the list with <kbd>down arrow</kbd> (<kbd>up arrow</kbd> to move backwards)
```

Arrow keys support is conditional.

src/components/Popover/README.md
```diff
@@ -418,7 +424,7 @@ See Apple’s Human Interface Guidelines and API documentation about accessibili
 
 Popovers usually contain an [option list](https://polaris.shopify.com/components/lists-and-tables/option-list) or an [action list](https://polaris.shopify.com/components/actions/action-list), but can also contain other controls or content.
 
-A popover can contain many numerous types of content. Whether it's a menu, grid or something entirely different! When `aria-expanded` is applied to an element, `aria-haspopup` will default to `menu`. To assist screen readers you'll need to pass `ariaHaspopup` to `Popover`.
+A popover can contain many numerous types of content. Whether it's a menu, grid or something entirely different! When `aria-expanded` is applied to an element, `aria-haspopup` will default to `menu`. To assist screen readers with an [action list](https://polaris.shopify.com/components/actions/action-list) you'll need to ensure that the first focusable node is automatically focused on open by passing `autofocusTarget='first-node'` to `Popover`.
```
NathanPJF: This edit removes the note about passing `ariaHaspopup`. Are you purposefully discouraging its use? If not, then this section could have language that includes `ariaHaspopup` and your addition about `autofocusTarget`. 

I'm assuming you're recommending it gets removed based on our experience with ActionList, though it may be useful to some devs. i.e. What's the original reason they allowed someone to provide a `aria-haspopup=<value>`? It is a bit vague on _how_ to help screen readers. 

Also: I feel like this first sentence should be removed as it isn't saying anything the paragraph before doesn't. 

> ~A popover can contain many numerous types of content. Whether it's a menu, grid or something entirely different!~

NathanPJF (commented):
I was testing this out in JAWS on a popover with multiple sections ([screenshot](https://screenshot.click/21-11-egwz6-h82in.png)) and I couldn't find a way to navigate up to the different headings with the arrow keys. This seems like a blocker for JAWS support, but I'm going to do some more playing around. Have you encountered this/had any thoughts about it?

src/components/Popover/README.md
patrickracicot: Indeed, I'm discouraging the use of `ariaHaspopup`.  According to ARIA documentation, when a button has the `aria-haspopup` attribute set, screen reader are expecting the content to be a sibling in the DOM tree, but in our case since we are using the Portal, the content is not a sibling.

I've also noticed that the following sentence  is not true since the code will only set the `aria-haspopup` to the value passed as props to the activator.
> When `aria-expanded` is applied to an element, `aria-haspopup` will default to `menu`.


I will re-think on how to rephrase this whole paragraph.

patrickracicot:
>I was testing this out in JAWS on a popover with multiple sections (screenshot) and I couldn't find a way to navigate up to the different headings with the arrow keys. 

I've just tried for ~40minutes and indeed their is no way for me to reach the `<p />` elements that represent the section text.  When trying to use different combination of [keystrokes](https://support.freedomscientific.com/Content/Documents/Manuals/JAWS/Keystrokes.pdf), I realized that we are not able to see the action items (buttons) nor the section text (paragraphs) from the popover content in the jaws list feature.  

Also when trying to list the paragraphs (ctrl+insert+P) while the focus is on one of the action items (buttons), JAWS actually announces an error informing the user that it isn't able to use the paragraph list feature outside of a virtual document.  This leads me to think that JAWS isn't able to understand the context of the popover content that is passed through the portal, meaning that adding the `role="menuitem"` and `role="menu"` does help the focus, navigation and activation management but isn't enough to properly add these items to the screen reader's virtual context.

>This seems like a blocker for JAWS support, but I'm going to do some more playing around. Have you encountered this/had any thoughts about it?

I don't know if this is an actual blocker for screen reader support since the ActionList content was simply unreachable before this proposal.  It isn't ideal not being able to read the section text (`<p />`) with a screen reader, but this could be a known issue for sub-sequent work 🤔   @svinkle  @BPScott What are your thoughts on this ?

NathanPJF:
> I don't know if this is an actual blocker for screen reader support 

A better take from me would have been that the `menuitem` for `actionRole` approach seems to have an optimal use-case in an Actionlist without sections.

NathanPJF (approved):

src/components/Popover/README.md
NathanPJF: > the code will only set the aria-haspopup to the value passed as props to the activator

I saw that as well while testing in a Playground. It's a confusing sentence that makes it sound like `aria-haspopup` is always going to be applied.  Here's what I recommend as a re-write for the accessibility section:

> Popovers usually contain an [option list](https://polaris.shopify.com/components/lists-and-tables/option-list) or an [action list](https://polaris.shopify.com/components/actions/action-list), but can also contain other controls or content.
>
> To assist screen readers with sending focus to an [action list](https://polaris.shopify.com/components/actions/action-list), pass `autofocusTarget='first-node'` to `Popover`. This will avoid known issues a screen reader may have with keyboard support once focus is moved off the activator.
>
> Web browsers assign a default value of 'menu' to the `aria-haspopup` role. You can use the prop `ariaHaspopup` to specify a value. Screen readers may fail to send focus to the `Popover` content when they expect the content to be adjacent to the element with `aria-haspopup` in the DOM tree. In this scenario, it is recommended not to provide the `ariaHaspopup` prop.

I realize that last paragraph gets a bit technical. I debated removing it, but felt it's a good warning for people considering to use `ariaHaspopover` just because they think they are supposed to.

src/components/ActionList/ActionList.tsx
```diff
@@ -35,6 +41,9 @@ export function ActionList({
 
   const hasMultipleSections = finalSections.length > 1;
   const Element = hasMultipleSections ? 'ul' : 'div';
+  const elementRole = hasMultipleSections ? 'menu' : undefined;
```
NathanPJF: This `menu` role should be dependent on `actionRole` prop of `menuitem`. I know, based on our visibility, that ActionLists are always a list of menu items, however, the component gives the _option_ of `menuitem`, so this `menu` role should be conditional on that

NathanPJF (changes requested):

NathanPJF (approved):
🎩 'd on NVDA, Mac VO, and Jaws. Even with the known issue with JAWS not being able to access section titles, this is still a huge improvement over the current state of `ActionList` which offers _no support_ for the screenreader ([original issue](https://github.com/Shopify/polaris-react/issues/4419)). I know that @patrickracicot investigated a number of paths for the `ActionList` component and this approach of `menu`/`menuitem` is the one that makes action items navigable without requiring Polaris to re-evaulate the use of React Portals across the design system. (Internal Shopify links: [original investigation](https://github.com/Shopify/core-issues/issues/28696#issuecomment-933683709) and [presentation about alternatives explored](https://drive.google.com/file/d/1rZmsUMSwJDM7__QkNGS1GpeLv8RyAJ0G/view) (see 32min mark).

Also, the code in this PR has been tested with JAWS screen reader users in 4 different sessions and proved to address the issues they had in Shopify's admin.

> but this could be a known issue for sub-sequent work

This is the path I'd prefer to go at this point. A "complete fix" may require abandoning React Portals altogether, which is a large spike in complexity in Polaris React and in the meantime this PR brings a lot of value to JAWS users. Also, I like the fact it's an opt-in enhancement for devs as it requires assigning an `actionRole`, so it's not forcing the change in interactions on all consumers.

svinkle (approved):
In agreement with @NathanPJF. Looks good.

dleroux:
When it first loads in the playground I get this focus on the popover automatically. Is it my environment?

Screen Shot 2021-12-03 at 2 08 12 PM

dleroux:
The Key listener would also have to be triggered when the focus is on the action list I think. Keys are now hijacked and if tehre's more than 1 Action List, you can't use keys on the first one.
https://user-images.githubusercontent.com/1229901/144659659-0d73a2f0-a72b-434c-b1fc-259e265ba0db.mov

dleroux (commented):
Left a couple of comments on Friday. Let me when you had a chance to address them and I'll look at it again. Right now I can't dev up in Polaris :(

patrickracicot:
> When it first loads in the playground I get this focus on the popover automatically. Is it my environment?
> 
> Screen Shot 2021-12-03 at 2 08 12 PM

This is due to the playground setup.  The logic will focus the container (`<ul />` in the case of the ActionList) when the popover opens. And since in the playground the modal is automatically opened, that why the focus is automatically set to the container.


>The Key listener would also have to be triggered when the focus is on the action list I think. Keys are now hijacked and if tehre's more than 1 Action List, you can't use keys on the first one.
https://user-images.githubusercontent.com/1229901/144659659-0d73a2f0-a72b-434c-b1fc-259e265ba0db.mov

Indeed although the documentation seems to prefer using the `AciontList` inside of a popover, we do have an example of usage outside of the popover.

With this in mind, the approach I took was adding a condition to only execute the focus change logic when the event target is a child of the ActionList.

dleroux (approved):
Merci @patrickracicot! Tourte est beau maintenant! 🐑 🇮🇹

PR #4530, opened 2021-10-20
Remove opinionated layout from listbox option

kyledurand:

Fixes: https://github.com/Shopify/web/pull/51269#issuecomment-947807663

image

Removes the opinionated display property from Listbox > Option. Do we know for sure if this should be a flex layout? IMO we should only have opinions on this when we can also control the components direct child.

Note: The issue above can be fixed in web so I'm not sure this PR is the right way to go just yet, but my gut tells me it is. Please let me know if I'm wrong @voiid !

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -11,6 +11,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Bug fixes
 
 - Fixed empty children being wrapped with `Item` in `Stack` ([#4487](https://github.com/Shopify/polaris-react/pull/4487))
+- Fixed a bug where `Listbox.Option`'s children wouldn't stretch to fit their container ([#4530](https://github.com/Shopify/polaris-react/pull/4530))
 
 ### Documentation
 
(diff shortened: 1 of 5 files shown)


kyledurand:
Looks like maybe this is needed for `TextOption` which is set to `flex: 1`. My question is: do we really need this or is it a nice to have.

kyledurand:
I've removed the flex properties from sub components and compared the playground locally against [main](https://polaris-react.shopifycloud.com/?path=/story/all-components-listbox--basic-listbox)

alex-page (approved):

dleroux:
TextOption is also used to wrap the Action. I think in this case it should be flex. Note the hover and pressed state will need adjusting:

Screen Shot 2021-10-22 at 9 50 12 AM

kyledurand:
Thanks Dan! It looks like that specific layout issue is isolated to Autocomplete > actionBefore > MappedAction. So maybe I can take a look to see if we can strictly apply a flex layout there instead. For now, I've fixed the issue in web by applying `flex: 1` to child elements.

PR #4544, opened 2021-10-25
Update loom to v1

BPScott:

Updating loom to v1

- Updating Loom to v1
- Updating loom config
- Updating config helpers to be pure cjs. This isn't strictly needed but they they did that mismash thing where they used `require()` and `export` and thus mixed cjs and esm and relied on transpiling to sort it out which i think is silly.

### How to 🎩

- Do a build on the main branch and then move it to a temporary place with `mv build build-og`
- Checkout this branch run  `yarn` to update dependencies, run a `yarn clean` to delete the build folder
- Run `yarn build` to generate a new build
- Run `diff -ru -x '*.tsbuildinfo' build-og build` to compare the old and new build output.

---

### What about loom-plugin-build-extended?

An excellent question, how good of you to ask.  `loom-plugin-build-extended` includes support for loading scss files using node-sass. Its rollup plugin for handling styles is based upon the styles rollup plugin in polaris-react. However I am choosing not to remove this repo's styles plugin in favour of loom-plugin-build-extended. 

This is because we're intending to [replace `node-sass` with `dart-sass`](https://github.com/Shopify/loom/pull/295) in `loom-plugin-build-extended` in the near future. For most projects, that's completely fine - they fix any migration from node-sass to dart-sass and they're good. However polaris-react ships a sass public api that needs to be understandable in apps that use node-sass (as sewing-kit still ships with node-sass) and within the polaris-react codebase.

Migrating this codebase to dart-sass is possible, however it [raises deprecation warnings](https://github.com/Shopify/polaris-react/pull/4298) in the publicly used functions that can only be fixed by migrating the codebase to dart-sass and those changes will throw errors when used by apps that leverage node-sass.

In order to migrate this codebase to dart-sass we need to either: [remove the public sass API](https://github.com/Shopify/polaris-react/issues/4485), or make sure that all consuming apps have migrated to use dart-sass using [sewing-kit's dartSassLoader experiment](https://github.com/Shopify/sewing-kit/pull/2850) (this is huge effort and a pain, just remove the public sass api please).

In order to not be in a place where this codebase is constantly raising deprecation warnings that can't be fixed without breaking consumers it's better to leave polaris-react using node-sass for the time being. Which means sticking with it having its own style loader.


loom.config.ts
```diff
@@ -22,57 +20,39 @@ import packageJSON from './package.json';
...
       hooks.configure.hook((configuration) => {
-        configuration.jestEnvironment?.hook(() => 'jsdom');
-
-        // Aliases for root-level imports
+        // Aliases for root-level imports, which are used in test files
         // These do not work in rollup builds, so perhaps we shouldn't configure
-        // them to work here either
-        configuration.jestModuleMapper?.hook((moduleMapper) => {
+        // them to work in jest tests either
+        configuration.jestModuleNameMapper?.hook((moduleMapper) => {
```
dahukish, line 55: should everything else follow suit `moduleMapper` => `moduleNameMapper`

dahukish (approved):
Small comment, but :shipit:

loom.config.ts
BPScott: the naming of the argument isn't suuuper important as it's highly localised. I think leaving as moduleMapper gets the point across.

PR #4542, opened 2021-10-25
[OptionList] Use relative path for type imports

irvinebroque:
fixes https://github.com/Shopify/polaris-react/issues/4507
### WHY are these changes introduced?

Fixes https://github.com/Shopify/polaris-react/issues/4507

### WHAT is this pull request doing?

Uses relative, rather than absolute paths to import types, to prevent Typescript errors for consumers of polaris-react.

```
../../node_modules/@shopify/polaris/build/ts/latest/src/components/OptionList/OptionList.d.ts:2:58 - error TS2307: Cannot find module 'types' or its corresponding type declarations.

2 import type { OptionDescriptor, SectionDescriptor } from 'types';
```

More broadly, this repository needs a way to check for errors like this during build, or if possible to remove its dependency on this line in `tsconfig.json`

https://github.com/Shopify/polaris-react/blob/755b3ceaee9d9aedb508c2cc95db64fad28aba09/tsconfig.json#L4

PR #4559, opened 2021-10-29
Reduce usage of absolute paths

### LINKED ISSUE #4507: TypeScript can't resolve non-relative paths in type imports
## Issue summary

I'm getting a couple of errors when running the TypeScript compiler with Polaris 7.0.0:

```
[9:12:31 pm] File change detected. Starting incremental compilation...

node_modules/@shopify/polaris/build/ts/latest/src/components/OptionList/OptionList.d.ts:2:58 - error TS2307: Cannot find module 'types' or its corresponding type declarations.

2 import type { OptionDescriptor, SectionDescriptor } from 'types';
                                                           ~~~~~~~

node_modules/@shopify/polaris/build/ts/latest/src/types.d.ts:2:61 - error TS2307: Cannot find module 'components' or its corresponding type declarations.

2 import type { AvatarProps, IconProps, ThumbnailProps } from 'components';
                                                              ~~~~~~~~~~~~

[9:12:35 pm] Found 2 errors. Watching for file changes.
```

The two imports here are not using relative paths unlike most others, and when I update those two files locally to import `'../../types'` and `'./components'` (I think it was), the type check completes without error.

My tsconfig.json is:

```
{
  "compilerOptions": {
    "esModuleInterop": true,
    "jsx": "react",
    "moduleResolution": "node",
    "noEmit": true,
    "resolveJsonModule": true,
    "target": "ES2020"
  },
  "include": [
    "src/**/*"
  ]
}
```

## Specifications

- Are you using the React components? (Y/N): **Y**
- Polaris version number: **7.0.0**
- TypeScript version number: **4.3.5** (also tried with 4.4 and 4.2)
- Browser: **N/A**
- Device: **N/A**
- Operating System: Arch Linux

BPScott:

Fixes #4507, Fixes #4542

Typescript sets the `baseUrl` to be `./src` which means that absolute path imports like "types" works. This is neat but means we need to be careful - we can't use these aliases in compiled code. Turns out somebody did that though and caused types to break per #4507.

This PR fixes that original issue, and starts the process of using './' as the default baseUrl to reduce the chance of this from happening in the future.

- Fix usage of absolute paths in published files (i.e. not tests) to be relative
- Replace absolute imports of 'utilities' and 'types' to be relative
- Change baseUrl in tsconfig to point to root, and add path mappings for
  `components` and `test-utilities`

At this point in time `components`, `components/*`, `test-utilities` and `test-utilities/*` are now the only recognised absolute paths. You can use these in test files, however the build will still fail if you try and use these paths in not-tests.

### Next steps

We should finish this job. removing any usage of importing  `components`, `components/*`, `test-utilities` and `test-utilities/*`.

We can now move the contents of `test-utilities` into the `tests` folder and access them with `tests/BLAH`. Move the contents of `src/test-utilities/` into `tests/utilities`, update the import path from 'test-utilities' to `tests/utilities' and remove the references to `test-utilities in tsconfig.json and loom.config.ts. We could not access the contents of the tests folder with absolute paths before because the baseUrl was not `./`.

We can should also replace all imports from 'components' to use relative imports.

### DIFF
--- a/loom.config.ts
+++ b/loom.config.ts
@@ -53,7 +53,7 @@ function jestAdjustmentsPlugin() {
         // These do not work in rollup builds, so perhaps we shouldn't configure
         // them to work in jest tests either
         configuration.jestModuleNameMapper?.hook((moduleMapper) => {
-          moduleMapper['^(components|test-utilities|types|utilities)(.*)'] =
+          moduleMapper['^(components|test-utilities)(.*)'] =
             '<rootDir>/src/$1$2';
 
           return moduleMapper;
--- a/src/components/Banner/tests/Banner.test.tsx
+++ b/src/components/Banner/tests/Banner.test.tsx
@@ -7,7 +7,6 @@ import {
   DiamondAlertMajor,
 } from '@shopify/polaris-icons';
 import {mountWithApp} from 'test-utilities';
-import {BannerContext} from 'utilities/banner-context';
 import {
   Button,
   Heading,
@@ -17,6 +16,7 @@ import {
   UnstyledLink,
 } from 'components';
 
+import {BannerContext} from '../../../utilities/banner-context';
 import {WithinContentContext} from '../../../utilities/within-content-context';
 import {Banner, BannerHandles} from '../Banner';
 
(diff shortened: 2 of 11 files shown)


alex-page (approved):

mathildebuenerd:
@BPScott thanks for the PR, we were also facing type issues in `handshake-web` 😇

Do you know if this change was in the last release? If yes, can you add it to https://github.com/Shopify/polaris-react/blob/main/CHANGELOG.md

otherwise to the unreleased: https://github.com/Shopify/polaris-react/blob/main/UNRELEASED.md

PR #4561, opened 2021-10-29
Move src/test-utilities to tests/utilities

BPScott:

Following on from #4559, aiming to reduce the usage of bespoke absolute imports.

Test utilities are the only thing that are ok to use an absolute import, but it should be something relative to the root of the project rather than requiring a custom path. This removes the need for a custom path in tsconfig.

This PR moves the `src/test-utilities` folder to `tests/utilities` and updates all imports from `test-utilities` to 
`tests/utilities`.

The custom `test-utilities` path is removed from tsconfig.json.
The custom `test-utilities` path is removed in the jestModuleMappers and replaced with a mapping for the `tests` folder.


alex-page (approved):
Amazing. Thanks Ben!

PR #4571, opened 2021-11-04
[ActionList] remove vertical padding from wrapping div

mattkubej:

Addresses: https://github.com/Shopify/polaris-react/issues/4570

Removes the vertical padding of the `<ActionList />` to provide a more compact aesthetic.

Removes padding attribute from the top-level element of `<ActionList />`. 

Horizontal padding was set to 0 and only the vertical padding was set. So, I removed the padding attribute, so we have no padding set.

**Before**
action-list-before

**After**
action-list-after


ghost:

alex-page (approved):
Thanks @mattkubej

mattkubej:
We did a quick triage and found that the only component that does not use ActionList within a Popover is “Crop and rotate”

As highlighted here:
https://github.com/Shopify/shopify/issues/320053#issuecomment-959797165

dfmcphee (approved):

chloerice (approved):
:shipit:

ghost:
🎉 Thanks for your contribution to Polaris React!

PR #4572, opened 2021-11-04
[Stack] Revert not empty item css

### LINKED ISSUE #4560: [Stack] Vertical alignment is off
## Issue summary

[Screencast](https://screenshot.click/screencast_2021-10-29_16-36-23.mp4)

I think this stems from some specificity that was added [here](https://github.com/Shopify/polaris-react/pull/4556/files#diff-12888031b1f0e16bf9d7ee02978c37cbdfd6b64f8912ad164ab1cc51410fd538R10)

It's overriding `margin-left: spacing(none)` [here](https://github.com/Shopify/polaris-react/blob/main/src/components/Stack/Stack.scss#L103-L106)

Context as to why `:not(:empty)` was added can be found [here](https://github.com/Shopify/polaris-react/issues/1890)

kyledurand:

Fixes https://github.com/Shopify/polaris-react/issues/4560

Reverts the change added here: https://github.com/Shopify/polaris-react/pull/4556/files#diff-12888031b1f0e16bf9d7ee02978c37cbdfd6b64f8912ad164ab1cc51410fd538R10

We've decided to ditch this effort for now and rely on consumers to render `Stack.Items` only if their content is not null in the same way we do with `Cards` and `Layout.Sections`

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -11,6 +11,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 ### Bug fixes
 
 - Removed extraneous space in `MediaCard` when card has no actions ([#4538](https://github.com/Shopify/polaris-react/pull/4538))
+- Fixed a bug in `Stack` where vertical spacing was off ([#4572](https://github.com/Shopify/polaris-react/pull/4572))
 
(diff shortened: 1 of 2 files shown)


chloerice (approved):
🚀

PR #4576, opened 2021-11-05
Fix deadlinks to 'Versioning and changelog' documentation

brooksilg:

The problem: There are 2 `git.io` links that are outdated and are now deadlinks to what should be the `Versioning and changelog` documentation file.

This PR replaces those dead links with relative links to the intended documentation.


chloerice (approved):
Thank you @brooksilg!! 🚀

PR #4580, opened 2021-11-08
[Popover] Motion improvements

dfmcphee:

This PR makes a few improvements to the `Popover` component including some foundational changes in things that it uses.

1. It improves the animation to feel snappier. It has been updated to use the `fast` duration and an updated `base` easing curve woking with @johanstromqvist. It also removed the animation on close to make it get out of the users way immediately.

#### Before

https://user-images.githubusercontent.com/478990/140956276-d9b92832-cbb5-467d-9c7a-d4ab9dd1d278.mov

#### After

https://user-images.githubusercontent.com/478990/140956239-b0554506-9a11-4162-b2d1-953afae4b940.mov

2. Bumps our default `max-height` for popovers to avoid forcing users to scroll the popover when they shouldn't need to.

#### Before

Screen Shot 2021-11-09 at 11 01 40 AM

#### After

Screen Shot 2021-11-09 at 11 01 28 AM

3. Remove the scroll hint by default on popover panes. Currently we animate any scrollable popover on open, which can be disorienting. By increasing the max-height this happens less often, but we also removed this as the default option.

#### Before

https://user-images.githubusercontent.com/478990/140959678-327109b7-b1e5-45c5-a47d-476de7e18197.mov

#### After

https://user-images.githubusercontent.com/478990/140959700-0a3eda14-ef96-4480-b583-b29d86ec9245.mov

4. Fixes a bug where popover or tooltip contents could be hidden behind the top bar if set to `preferredPosition="above"`.

#### Before

Screen Shot 2021-11-09 at 10 51 34 AM

#### After

Screen Shot 2021-11-09 at 10 51 13 AM

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -8,6 +8,8 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - Added helper hooks `useIndexTableRowHovered`, `useIndexTableRowSelected`, and `useIndexTableContainerScroll` to `IndexTable` ([#4286](https://github.com/Shopify/polaris-react/pull/4286))
 - Added token for slim border radius ([#4573](https://github.com/Shopify/polaris-react/pull/4573))
+- Improves `Popover` component and its animation ([#4580](https://github.com/Shopify/polaris-react/pull/4580))
+- Improves `base` easing curve ([#4580](https://github.com/Shopify/polaris-react/pull/4580))
 
 ### Bug fixes
 
--- a/src/components/Popover/Popover.scss
+++ b/src/components/Popover/Popover.scss
@@ -4,8 +4,7 @@ $arrow-size: rem(14px);
 $visible-portion-of-arrow: rem(5px);
 $content-max-height: rem(500px);
 $content-max-width: rem(400px);
-$vertical-motion-offset: rem(-20px);
-$motion-duration: 150ms;
+$vertical-motion-offset: rem(-5px);
 
 .Popover {
   max-width: calc(100vw - #{2 * spacing()});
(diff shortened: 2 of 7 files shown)


ouellettejordan:
🚢

chloerice (approved):
So good 🤌🏽

alex-page (approved):
This is great. Thanks @dfmcphee! I am only unsure about the `will-change` property and if it is actually needed for the `transform` and `opacity`. Happy to merge but would love a quick investigation to see if it is needed.

> Don’t apply will-change to elements to perform premature optimization. If your page is performing well, don’t add the will-change property to elements just to wring out a little more speed. will-change is intended to be used as something of a last resort, in order to try to deal with existing performance problems. It should not be used to anticipate performance problems. Excessive use of will-change will result in excessive memory use and will cause more complex rendering to occur as the browser attempts to prepare for the possible change. This will lead to worse performance.
https://developer.mozilla.org/en-US/docs/Web/CSS/will-change

johanstromqvist:
Great stuff @ouellettejordan and @dfmcphee!

The asymmetric enter/exit patterns wasn't mentioned, i.e. removing exit transition. Perhaps not essential to the PR? A strong default and scalable pattern for sure!
- Enter: fast + reinforce relationship
- Exit: immediately gone

dfmcphee:
> I am only unsure about the `will-change` property and if it is actually needed for the `transform` and `opacity`. Happy to merge but would love a quick investigation to see if it is needed.

Comparing with and without the `will-change` the performance is almost identical. I am going to remove it.

> The asymmetric enter/exit patterns wasn't mentioned, i.e. removing exit transition. Perhaps not essential to the PR? A strong default and scalable pattern for sure!

Good catch. I will update the description to reflect this too.

dleroux:
Awesome! Turned out to be a lot simpler than I was thinking.

dleroux (approved):

ahmed-adly-khalil:
@dfmcphee Thank you so much for this PR, this was very much needed, especially the scroll part

PR #4620, opened 2021-11-09
Remove miscellaneous css custom properties

### LINKED ISSUE #4589: Remove miscellaneous css custom properties
Remove this garbage ❤️ 

| token | value |
| --- | --- |
| --p-override-none | none |
| --p-override-transparent | transparent |
| --p-override-one | 1 |
| --p-override-visible | visible |
| --p-override-zero | 0 |
| --p-non-null-content | '' |
| --p-badge-mix-blend-mode | luminosity |
| --p-range-slider-thumb-scale | 1.5 |

lgriffee:

Fixes #4589 

The following tokens (see table below) were made to rollout the design language changes in 2020 and should no longer be used.

Swapping the following tokens for equivalent values or completely removing the value and associated css property.

| token | value |
| --- | --- |
| --p-override-none | none |
| --p-override-transparent | transparent |
| --p-override-one | 1 |
| --p-override-visible | visible |
| --p-override-zero | 0 |
| --p-non-null-content | '' |
| --p-badge-mix-blend-mode | luminosity |
| --p-range-slider-thumb-scale | 1.5 |

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist


documentation/Color system.md
```diff
@@ -268,14 +268,8 @@ Used to decorate elements where color does convey a specific meaning in componen
...
 | `--p-button-inner-shadow`                 | `inset 0 -1px 0 rgba(0, 0, 0, 0.2)`                                                                          |
 | `--p-button-pressed-inner-shadow`         | `inset 0 1px 0 rgba(0, 0, 0, 0.15)`                                                                          |
-| `--p-override-none`                       | `none`                                                                                                       |
-| `--p-override-transparent`                | `transparent`                                                                                                |
-| `--p-override-one`                        | `1`                                                                                                          |
-| `--p-override-visible`                    | `visible`                                                                                                    |
-| `--p-override-zero`                       | `0`                                                                                                          |
```
alex-page, line 275: I would revert the changes in this file as these values are still going to come from `polaris-tokens`. This file is built by running `yarn colordocs`. We will remove this file and content in the future!

src/components/Badge/Badge.scss
```diff
@@ -15,7 +15,7 @@ $pip-spacing: ($height - $pip-size) / 2;
   align-items: center;
   padding: $vertical-padding $horizontal-padding;
   background-color: var(--p-surface-neutral);
-  border: var(--p-override-zero);
+  border: 0;
```
alex-page: Can we remove the `border: 0` or is it overriding another value?

alex-page (commented):
This looks great @lgriffee. I am unsure if we can remove some of these values all together or if they override another value in the CSS cascade. Would you be able to take a look?

src/components/Badge/Badge.scss
lgriffee: I think we can remove it. I removed the property and then checked all Storybook examples in Chrome, Safari, and Firefox and didn't find anything that it might be overriding.

lgriffee: I can go through all the other values and do the same. Let me know if you think I may be overlooking anything testing wise.

lgriffee: Oh I just realized I didn't test on mobile (iOS or Android) if that's something else I should also do for all of these.

alex-page: Just make sure to test responsively if the CSS is inside a media query. We shouldn't need to device test these fallbacks.

src/components/Select/Select.scss
```diff
@@ -41,21 +41,6 @@ $stacking-order: (
...
-    // stylelint-disable-next-line color-no-hex
-    color: #9c9798;
-  }
-
-  // stylelint-disable-next-line selector-max-class, selector-max-specificity
-  &.error .Input:-moz-focusring {
-    color: transparent;
-    text-shadow: var(--p-override-none);
-  }
-}
```
lgriffee, line 57: Removing unused class. Here's [the legacy commit from when it was added](https://github.com/Shopify/polaris-react/blob/daaa92ec1ff45a5ebc8f1d6cc431ce259368d6e3/src/components/Select/Select.tsx#L68)

src/components/Breadcrumbs/Breadcrumbs.scss
```diff
@@ -24,10 +24,6 @@ $icon-size: rem(20px);
 
   &:active {
     background-color: var(--p-surface-pressed);
-
-    .ContentWrapper {
-      background: var(--p-override-transparent);
-    }
```
alex-page, line 30: Do we need the HTML element with `.ContentWrapper` classname? Or can we flatten this components HTML?

alex-page (approved):
Awesome work @lgriffee. I had one question but this looks great 💯. It looks like you have a conflict with the `UNRELEASED.md` file let me know if you need help resolving that!

src/components/Select/Select.scss
alex-page: Amazing! Thanks for the context Laura. I think it makes sense to remove this code.

src/components/Breadcrumbs/Breadcrumbs.scss
lgriffee: I don't think we need it—I can consolidate its relevant CSS into `.Breadcrumb` and replace its HTML with a shorthand fragment so that `contentMarkup` still has one parent element (which you can check out in [this commit](https://github.com/Shopify/polaris-react/pull/4620/commits/bd044e57df1d98e52483e1adbdd34b1216100660)).

lgriffee: Also along the line of consolidation I just want to note that there might be more rollback values that need to be removed in these files. Not sure if there was work done already to remove them so it’s definitely possible that they are needed overrides. I didn’t look into them in this PR to keep a narrow scope. But if it’s something important to address now or in the future I can create a ticket to look into it.

UNRELEASED.md
```diff
@@ -10,6 +10,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 - Added token for slim border radius ([#4573](https://github.com/Shopify/polaris-react/pull/4573))
 - Improves `Popover` component and its animation ([#4580](https://github.com/Shopify/polaris-react/pull/4580))
 - Improves `base` easing curve ([#4580](https://github.com/Shopify/polaris-react/pull/4580))
+- Removed miscellaneous css custom properties ([#4620](https://github.com/Shopify/polaris-react/pull/4620))
```
kyledurand: Maybe this should be nested under `### Code quality`?

aveline (approved):
This looks great @lgriffee!

src/components/Link/Link.scss
```diff
@@ -19,7 +19,6 @@
 
   &:focus {
     $chrome-default-outline: rgba(0, 103, 244, 0.247) auto rem(4.5px);
-    outline: var(--p-override-none);
```
kyledurand, line 22: We may still need `outline: none` here?

src/components/RangeSlider/components/DualThumb/DualThumb.scss
```diff
@@ -133,11 +133,6 @@ $range-output-spacing: rem(16px);
   transition-timing-function: var(--p-ease);
   transform: translateX(calc(-50% + #{$range-thumb-size / 2}));
 
-  .Thumbs:hover + &,
-  .Thumbs:focus + & {
-    opacity: var(--p-override-zero);
-  }
```
kyledurand, line 139: This seems like it would cause a visual change

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -182,11 +178,6 @@ $range-output-spacing: rem(16px);
   transition-duration: duration();
   transition-timing-function: easing();
 
-  .Input:hover + &,
-  .Input:focus + & {
-    opacity: var(--p-override-zero);
-  }
```
kyledurand, line 188: Same here. I could be wrong because I can't see all of the code so I'll make sure to 🎩 after review

src/components/TopBar/components/Menu/Menu.scss
```diff
@@ -36,10 +36,7 @@ $activator-variables: (
 
   &:focus {
     @include focus-ring($style: 'focused');
-    background-color: var(
-      --top-bar-background-lighter,
-      var(--p-override-transparent)
-    );
+    background-color: var(--top-bar-background-lighter, transparent);
```
kyledurand: We don't need the fallback here any more
```suggestion
    background-color: var(--top-bar-background-lighter);
```

kyledurand (commented):
Looks good just a few comments. I'll 🎩 now

UNRELEASED.md
lgriffee: Good call! I'll move it to that section

src/components/Link/Link.scss
lgriffee: I made sure to check all of these style changes on Chrome, Safari, and Firefox as I went through them and didn't see an outline on focus for this one but would love it if you could confirm that's the case for you too!

src/components/RangeSlider/components/DualThumb/DualThumb.scss
lgriffee: I didn't see any style changes for this when I tested but it would be great to confirm that! Did you get the chance to 🎩 this?

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
lgriffee: Totally! Anything come out of the 🎩?

src/components/Link/Link.scss
```diff
@@ -19,7 +19,6 @@
 
   &:focus {
     $chrome-default-outline: rgba(0, 103, 244, 0.247) auto rem(4.5px);
```
kyledurand: Looks like this can be removed now

src/components/Link/Link.scss
kyledurand: This is being overridden with our styles so we should be good 👍

src/components/RangeSlider/components/DualThumb/DualThumb.scss
kyledurand: Seems ok on 🎩 !

kyledurand (approved):
Sorry about the delay 🎩 ing. I ran into a nix issue with dev last night and got distracted with other things. This looks great! Excellent clean up. The only thing I noticed via chromatic is that the breadcrumb back button is slightly wider but I think it's an improvement now that it's perfectly square 👍 

I left a note about removing one more sass var that appears to be unused, and unreleased has some entries that can snuck in by a rebase probable but otherwise 🚢 !

PR #4553, opened 2021-10-27
[Modal]Fix try-catch syntax error

youminkim:
### WHY are these changes introduced?

Fixes #4552  

### WHAT is this pull request doing?

Added identifier in catch to avoid syntax error

PR #4625, opened 2021-11-10
Fix #4552

### LINKED ISSUE #4552: Try-catch Syntax error in Modal
## Issue summary

After 7.0.0 update, try-catch syntax errors happens in any Modal component usages in some Android 

## Expected behavior

No syntax error

## Actual behavior

```
SyntaxError: Unexpected token {
  at ? (./node_modules/@shopify/polaris/build/esm/components/Modal/Modal.js:69:9)
```

```
    const iframe = evt.target;
    if (iframe && iframe.contentWindow) {
      try {
        setIframeHeight(iframe.contentWindow.document.body.scrollHeight);
      } catch {
        setIframeHeight(DEFAULT_IFRAME_CONTENT_HEIGHT);
      }
    }
    if (onIFrameLoad != null) {

```

Any Modal component use.

Nextjs 

## Reduced test case
-

## Specifications

- Are you using the React components? Y
- Polaris version number: 7.0.0
- Browser: Chrome Mobile Webview (62.0.3202)
- Device: Vivo 1716
- Operating System: Version:8.1.0

alex-page:

Fixes https://github.com/Shopify/polaris-react/issues/4552

Removing the unecessary `try` and `catch` logic in the modal.

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -119,9 +119,7 @@ export const Modal: React.FunctionComponent<ModalProps> & {
     (evt: React.SyntheticEvent<HTMLIFrameElement>) => {
       const iframe = evt.target as HTMLIFrameElement;
       if (iframe && iframe.contentWindow) {
-          const iframeHeight = iframe.contentWindow.document.body.scrollHeight || DEFAULT_IFRAME_CONTENT_HEIGHT;
-          setIframeHeight(iframeHeight);
-        }
+        setIframeHeight(iframe.contentWindow.document.body.scrollHeight || DEFAULT_IFRAME_CONTENT_HEIGHT);
       }
 
       if (onIFrameLoad != null) {
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -119,10 +119,8 @@ export const Modal: React.FunctionComponent<ModalProps> & {
     (evt: React.SyntheticEvent<HTMLIFrameElement>) => {
       const iframe = evt.target as HTMLIFrameElement;
       if (iframe && iframe.contentWindow) {
-        try {
-          setIframeHeight(iframe.contentWindow.document.body.scrollHeight);
-        } catch {
-          setIframeHeight(DEFAULT_IFRAME_CONTENT_HEIGHT);
+          const iframeHeight = iframe.contentWindow.document.body.scrollHeight || DEFAULT_IFRAME_CONTENT_HEIGHT;
+          setIframeHeight(iframeHeight);
         }
       }
 


src/components/Modal/Modal.tsx
```diff
@@ -119,11 +119,7 @@ export const Modal: React.FunctionComponent<ModalProps> & {
     (evt: React.SyntheticEvent<HTMLIFrameElement>) => {
       const iframe = evt.target as HTMLIFrameElement;
       if (iframe && iframe.contentWindow) {
-        try {
-          setIframeHeight(iframe.contentWindow.document.body.scrollHeight);
-        } catch {
-          setIframeHeight(DEFAULT_IFRAME_CONTENT_HEIGHT);
-        }
+        setIframeHeight(iframe.contentWindow.document.body.scrollHeight || DEFAULT_IFRAME_CONTENT_HEIGHT);
```
alex-page, line 122: Is there any reason this needs the fallback `DEFAULT_IFRAME_CONTENT_HEIGHT`. The `if (iframe && iframe.contentWindow)` seems pretty safe and I wondered if it would ever use `DEFAULT_IFRAME_CONTENT_HEIGHT`.

LauraAubin: I'm not sure if by removing the try/catch block this becomes irrelevant, but I think it would be safer to leave the fallback in since it was originally added to fix a cross-origin error here https://github.com/Shopify/polaris-react/pull/1992.

LauraAubin (approved):
I did not 🎩 but these changes look solid 🚢

src/components/Modal/Modal.tsx
BPScott: I suspect this was originally a `try/catch` rather than a `||`because an error gets throw if you try to access `iframe.contentWindow.document.body` on an iframe from a different origin. It'd be worth confirming the exact behaviour here.

I suspect #4553 is a safer fix - as I bet the error is thrown in browsers that don't support [optional catch binding](https://2ality.com/2017/08/optional-catch-binding.html).

I'll note that [babel transforms](https://babeljs.io/repl#?browsers=node%206%2C%0Alast%203%20chrome%20versions%2C%0Alast%203%20firefox%20versions%2C%0Alast%203%20opera%20versions%2C%0Alast%203%20edge%20versions%2C%0Alast%203%20safari%20versions%2C%0Alast%203%20chromeandroid%20versions%2C%0Alast%201%20firefoxandroid%20versions%2C%0Aios%20%3E%3D%2013.4%0A&build=&builtIns=false&corejs=3.6&spec=false&loose=false&code_lz=C4JwngBA3gUAvjAxgQ2IgFtGB6bEBmA9ofEA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.16.3&externalPlugins=&assumptions=%7B%7D) `catch {}` into `catch (_unused) {}` if the targets include browsers that require the transform (delete the 'node 6' in the targets to see it no longer being transformed, with the browsers that polaris targets). So that suggests that this is only occuring in browsers that we don't claim to support. Though it's a trivial fix to add an unused binding so why not.

alex-page: Oh nice! Here is the [can I use](https://caniuse.com/mdn-javascript_statements_try_catch_optional_catch_binding) page for optional catch binding, seems fine.

PR #4631, opened 2021-11-11
[Checkbox] properly support screen readers

### LINKED ISSUE #3973: [Checkbox][a11y] Unable to check checkbox with voiceover
## Issue summary
When using voiceover on either iOS or macOS, the voiceover command for click (iOS: Double tap, macOS: VO + Space) does not toggle the checkbox.

## Expected behavior
When you click the checkbox with voiceover, it should select or deselect the checkbox.

## Actual behavior
When you click the checkbox with voiceover, it does not select the checkbox.

Recordings

iOS

https://user-images.githubusercontent.com/1087756/106942537-4aeb8580-66f2-11eb-81ed-c61753b7f308.mp4

macOS

https://user-images.githubusercontent.com/1087756/106942370-1bd51400-66f2-11eb-86c1-92ca45930ad7.mp4

1. Navigate to a page with a checkbox (https://wlmtq.csb.app)
2. Turn on voiceover and tab to the checkbox
3. When prompted, use the voiceover command to toggle the checkbox

## Reduced test case
https://wlmtq.csb.app

## Specifications

- Polaris version number: Latest (& 5.15.1/any)
- Browser: Any
- Device: Macbook pro, iPhone
- Operating System: macOS 11, iOS 14.4

patrickracicot:

Fixes #3973  

The goal of this PR is too properly support screen readers on the `Checkbox` component.  Currently, screen reader users must use a combination of `space+enter` key to be able to check/uncheck the checkbox.

The current proposal modifies how the component handles the `click` and `keyboard` events.  We now avoid hijacking the events on the `Choice` component.  Instead we listen to the events directly on the `<input />` element.  The reason why we need to do this is due to the fact that screen readers are triggering the events directly on the `<input />` element and this triggered event is not capturable by the `onClick` handler of the wrapping `Choice` component.

I also removed the `onKeyPress` handler of the `<input />`.  According to the HTML spec, triggering the `space` keyboard event will activate (invoke the click event) of an interact-able element. To properly support the focus state, I added the `onFocus` handler on the `<input />` since this was managed inside the `onKeyPress` previously which is a side effect.

For a deep dive please refer to the following document: [Polaris Checkbox (in Draft Orders context)](https://docs.google.com/document/d/1SIOQRb1vNipj9DEeeMINBI_kOl3KV76bx9hR4oh3lK4/edit?usp=sharing)

## 

### How to 🎩

To properly 🎩 these changes, one will need to build the repo using `yarn dev`.
The checkbox can be tested through the playbook and is actually used in the following examples
 - [Checkbox](http://localhost:6006/?path=/story/all-components-checkbox--default-checkboxes)
 - [Choice list/Multi-choice list](http://localhost:6006/?path=/story/all-components-choice-list--multi-choice-list)
 - [Option list](http://localhost:6006/?path=/story/all-components-option-list--multiple-option-list)

Step to 🎩 
 - With a storybook example open
 - Tab to focus on the checkbox
 - Hit `space` to check
 - With the mouse, left click the box to uncheck
 - With the mouse, left click the label to check
 - With a screen reader, use the screen readers shortcut to activate the checkbox
     - For mac VoiceOver, the combination is `control+option+space`

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -20,6 +20,7 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 - Fixed try-catch syntax error in `Modal` ([#4553](https://github.com/Shopify/polaris-react/pull/4553))
 - Fixed an issue with `TextField` where date and time were uneditable on click ([#4671](https://github.com/Shopify/polaris-react/pull/4671))
 - Fixed an issue with `Popover` where the transform property interfered with descendants positioning ([#4685](https://github.com/Shopify/polaris-react/pull/4685))
+- Fixed screen reader accessibility issue of the `Checkbox` component ([#4631](https://github.com/Shopify/polaris-react/pull/4631))
 
 ### Documentation
 
--- a/src/components/CheckableButton/tests/CheckableButton.test.tsx
+++ b/src/components/CheckableButton/tests/CheckableButton.test.tsx
@@ -3,7 +3,6 @@ import {mountWithApp} from 'tests/utilities';
 import {Checkbox} from 'components';
 
 import {CheckableButton} from '../CheckableButton';
-import {Key} from '../../../types';
 
 const CheckableButtonProps = {
   label: 'Test-Label',
@@ -71,19 +70,6 @@ describe('<CheckableButton />', () => {
         <CheckableButton {...CheckableButtonProps} onToggleAll={spy} />,
       );
       element.find('div')!.trigger('onClick');
-      expect(spy).toHaveBeenCalled();
-    });
-
-    it('is called when the CheckableButton pressed with spacebar', () => {
-      const spy = jest.fn();
-      const element = mountWithApp(
-        <CheckableButton {...CheckableButtonProps} onToggleAll={spy} />,
-      );
-
-      element.find(Checkbox)!.find('input')!.trigger('onKeyUp', {
-        keyCode: Key.Space,
-      });
-
       expect(spy).toHaveBeenCalled();
     });
   });
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -13,7 +13,7 @@ import {useUniqueId} from '../../utilities/unique-id';
 import {Choice, helpTextID} from '../Choice';
 import {errorTextID} from '../InlineError';
 import {Icon} from '../Icon';
-import {Error, Key, CheckboxHandles} from '../../types';
+import {Error, CheckboxHandles, Key} from '../../types';
 import {WithinListboxContext} from '../../utilities/listbox/context';
 
 import styles from './Checkbox.scss';
@@ -92,20 +92,21 @@ export const Checkbox = forwardRef<CheckboxHandles, CheckboxProps>(
       setKeyFocused(false);
     };
 
-    const handleInput = () => {
+    const handleKeyUp = (event: React.KeyboardEvent) => {
+      const {keyCode} = event;
+
+      if (keyCode === Key.Space || keyCode === Key.Tab) {
+        !keyFocused && setKeyFocused(true);
+      }
+    };
+
+    const handleOnClick = () => {
       if (onChange == null || inputNode.current == null || disabled) {
         return;
       }
-      onChange(!inputNode.current.checked, id);
-      inputNode.current.focus();
-    };
 
-    const handleKeyUp = (event: React.KeyboardEvent) => {
-      const {keyCode} = event;
-      !keyFocused && setKeyFocused(true);
-      if (keyCode === Key.Space) {
-        handleInput();
-      }
+      onChange(inputNode.current.checked, id);
+      inputNode.current.focus();
     };
 
     const describedBy: string[] = [];
@@ -152,13 +153,11 @@ export const Checkbox = forwardRef<CheckboxHandles, CheckboxProps>(
         helpText={helpText}
         error={error}
         disabled={disabled}
-        onClick={handleInput}
         onMouseOver={handleMouseOver}
         onMouseOut={handleMouseOut}
       >
         <span className={wrapperClassName}>
           <input
-            onKeyUp={handleKeyUp}
             ref={inputNode}
             id={id}
             name={name}
@@ -167,17 +166,22 @@ export const Checkbox = forwardRef<CheckboxHandles, CheckboxProps>(
             checked={isChecked}
             disabled={disabled}
             className={inputClassName}
-            onFocus={onFocus}
             onBlur={handleBlur}
-            onClick={stopPropagation}
             onChange={noop}
+            onClick={handleOnClick}
+            onFocus={onFocus}
+            onKeyUp={handleKeyUp}
             aria-invalid={error != null}
             aria-controls={ariaControls}
             aria-describedby={ariaDescribedBy}
             role={isWithinListbox ? 'presentation' : 'checkbox'}
             {...indeterminateAttributes}
           />
-          <span className={backdropClassName} />
+          <span
+            className={backdropClassName}
+            onClick={stopPropagation}
+            onKeyUp={stopPropagation}
+          />
           <span className={styles.Icon}>
             <Icon source={iconSource} />
           </span>
@@ -189,6 +193,8 @@ export const Checkbox = forwardRef<CheckboxHandles, CheckboxProps>(
 
 function noop() {}
 
-function stopPropagation<E>(event: React.MouseEvent<E>) {
+function stopPropagation(
+  event: React.MouseEvent | React.KeyboardEvent | React.FormEvent,
+) {
   event.stopPropagation();
 }
--- a/src/components/Checkbox/tests/Checkbox.test.tsx
+++ b/src/components/Checkbox/tests/Checkbox.test.tsx
@@ -2,7 +2,6 @@ import React, {AllHTMLAttributes} from 'react';
 import {mountWithApp} from 'tests/utilities';
 
 import {Key} from '../../../types';
-import {Choice} from '../../Choice';
 import {Checkbox} from '../Checkbox';
 
 describe('<Checkbox />', () => {
@@ -32,18 +31,6 @@ describe('<Checkbox />', () => {
     });
   });
 
-  it('does not propagate click events from input element', () => {
-    const spy = jest.fn();
-    const element = mountWithApp(
-      <Checkbox id="MyCheckbox" label="Checkbox" onChange={spy} />,
-    );
-
-    element.find('input')!.trigger('onClick', {
-      stopPropagation: () => {},
-    });
-    expect(spy).not.toHaveBeenCalled();
-  });
-
   describe('onChange()', () => {
     it('is called with the updated checked value of the input on click', () => {
       const spy = jest.fn();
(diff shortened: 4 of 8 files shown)


src/components/CheckableButton/tests/CheckableButton.test.tsx
```diff
@@ -73,18 +72,5 @@ describe('<CheckableButton />', () => {
...
-      const element = mountWithApp(
-        <CheckableButton {...CheckableButtonProps} onToggleAll={spy} />,
-      );
-
-      element.find(Checkbox)!.find('input')!.trigger('onKeyUp', {
-        keyCode: Key.Space,
-      });
-
-      expect(spy).toHaveBeenCalled();
-    });
```
patrickracicot, line 88: Since we no longer manually handler the `keyUp` events, we no longer need to test this.

src/components/Checkbox/tests/Checkbox.test.tsx
```diff
@@ -52,57 +39,25 @@ describe('<Checkbox />', () => {
...
-    it('is not called from keyboard events when disabled', () => {
-      const spy = jest.fn();
-      const checkbox = mountWithApp(
-        <Checkbox label="label" disabled onChange={spy} />,
-      );
-      checkbox.find('input')!.trigger('onKeyUp', {
-        keyCode: Key.Enter,
-      });
-      expect(spy).not.toHaveBeenCalled();
-    });
```
patrickracicot, line 104: Since we no longer manually handler the keyUp events, we no longer need to test this.

src/components/Checkbox/tests/Checkbox.test.tsx
```diff
@@ -52,57 +39,25 @@ describe('<Checkbox />', () => {
...
-      const element = mountWithApp(
-        <Checkbox id="MyCheckbox" label="Checkbox" onChange={spy} />,
-      );
-
-      element.find('input')!.trigger('onKeyUp', {
-        keyCode: Key.Space,
-      });
-
-      expect(spy).toHaveBeenCalledTimes(1);
-    });
```
patrickracicot, line 71: Since we no longer manually handler the keyUp events, we no longer need to test this.

src/components/Checkbox/tests/Checkbox.test.tsx
```diff
@@ -175,22 +130,6 @@ describe('<Checkbox />', () => {
...
-
-      checkbox.find('input')!.trigger('onKeyUp', {
-        keyCode: Key.Enter,
-      });
-      checkbox.setProps({checked: true});
-
-      expect(checkbox).toContainReactComponent('input', {
-        checked: true,
-      });
-    });
```
patrickracicot, line 193: Since we no longer manually handler the keyUp events, we no longer need to test this.

kyledurand:
I'm down to make this sort of change but I think we ran into issues with handling focus on checkboxes natively back in the day. I think @dleroux has context on why

NathanPJF (approved):
I've 🎩 'd on Mac+VoiceOver, JAWS, and NVDA; all works as expected 

>  I think we ran into issues with handling focus on checkboxes natively back in the day

@kyledurand and @dleroux: I went looking through past GitHub issues about checkboxes and focus that involved Dan to see what more to check:

* ResourceList https://github.com/Shopify/polaris-react/issues/792, I found through tophatting that focus is managed as expected for the [ResourceList with multi-select](http://localhost:6006/?path=/story/all-components-resource-list--resource-list-with-multiselect).
* Styling in general https://github.com/Shopify/polaris-react/pull/2661; and there isn't an issue with the changes in this PR as it still uses the `keyFocused` state here.

Those were the ones that stood out to me to double-check; happy to hear if there's an issue I missed.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -92,22 +92,19 @@ export const Checkbox = forwardRef<CheckboxHandles, CheckboxProps>(
       setKeyFocused(false);
```
dleroux, line 92: The reason we have this keyFocused on checkboxes is 2 folds:

1. We won't want the styles (focus ring) to be applied on click. Only when tabbing. This may be fixed with `:focus-visible` having better support now?

2. Also, since the focus right is on the backdrop, tabbing to the checkbox did not trigger the event to bubble and apply the styles. I believe this still an issue.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -188,7 +183,3 @@ export const Checkbox = forwardRef<CheckboxHandles, CheckboxProps>(
 );
 
 function noop() {}
-
-function stopPropagation<E>(event: React.MouseEvent<E>) {
```
dleroux, line 192: IIRC this was also part of the focus issue

src/components/Checkbox/Checkbox.tsx
dleroux: I just had a look and yes does show the focus rectangle on click now. so the classname should only be applied when it's a keyboard click.

src/components/Checkbox/Checkbox.tsx
dleroux: Pretty sure this had to do with the bulk actions. I would remove this.

dleroux:
This PR:

https://user-images.githubusercontent.com/1229901/142280542-59a576d7-d629-4430-91e4-69ca42397435.mov

Core:

https://user-images.githubusercontent.com/1229901/142280551-d5ee8359-af2f-45f8-97ff-bd9a3f5d0d7c.mov

src/components/Checkbox/Checkbox.tsx
patrickracicot: Can you elaborate on the bulk action issue that led us to add this ?

src/components/Checkbox/Checkbox.tsx
patrickracicot: Alright so I've added back the `onKeyUp` logic, but made some adjustments to it.

Instead of calling `setKeyFocused(true)` for every key up event, I only call it for the `Space` and `Tab` keyboard events.  This will prevent an issue we had in the current implementation where if a user clicked on the checkbox and then type any key other then Tab or Space, it would still apply the focus style.

src/components/Checkbox/Checkbox.tsx
patrickracicot: Please refer to other PR comments on the `IndexTable` component in which I have added a fix to properly support the bulk actions.

TLDR; the root cause of the issue were the same as for the checkbox hence managing events on other DOM nodes then the originating element.

src/components/IndexTable/components/Checkbox/Checkbox.tsx
```diff
@@ -30,22 +30,26 @@ export const Checkbox = memo(function Checkbox() {
...
+  };
+
   return (
     <Wrapper>
       <div className={styles.TableCellContentContainer}>
-        <div
-          className={wrapperClassName}
-          onClick={onInteraction}
-          onKeyUp={onInteraction}
-          onChange={stopPropagation}
```
patrickracicot: We want to avoid doing this as this is event bubbling management and isn't required.  Instead, we can simply subscribe to the `onChange` prop of the `PolarisCheckbox`

src/components/Checkbox/Checkbox.tsx
```diff
@@ -42,7 +42,11 @@ export interface CheckboxProps {
   /** Display an error message */
   error?: Error | boolean;
   /** Callback when checkbox is toggled */
-  onChange?(newChecked: boolean, id: string): void;
+  onChange?(
+    newChecked: boolean,
+    id: string,
+    event?: React.KeyboardEvent | React.MouseEvent<Element, MouseEvent>,
```
dleroux: Do we need to pass the event here? There have been ongoing discussions about this and we have always opted to not pass the event. cc @kyledurand

dleroux: @kyledurand @bpscott there was discussion in this repo https://github.com/orgs/Shopify/teams/polaris/discussions/ but since the team name changed I can't find it.

dleroux:
Code looks good, so does 🎩  but let hear from others before opening up to passing the event.

src/components/Checkbox/Checkbox.tsx
patrickracicot: Fore more context here is the reasons why I'm passing the event inside the `onChange` handler.

Inside the `onInteraction` method of the `RowContext` requires this event as arg.  It needs to ensure that we stop the propagation of the event further up in the DOM tree and avoids unwanted side-effects.  Although I'm not a fan of always passing event through javascript method args, I feel like in this case it's necessary to avoid the need of further refactoring in the IndexTable.  

Also the way the code was written before my proposal,  we had a `<div>` wrapper around the `Checkbox` component which would listen to the `onClick` and `onKeyUp` events that were bubbling up. With this in mind, it's safe to say that passing the event through the method args serves the same purpose.

BPScott: It was talk from back before the days that Github discussions existed: https://github.com/Shopify/polaris-react/issues/521. (and https://github.com/Shopify/polaris-react/pull/4111 is a recent-ish reitteration)

It's worth pointing out that exposing dom events as part of our public api plays badly with aspirations around Argo which has no way of moving these events around.


Passing dom events around internally is bearable, but them being part of the public API causes problems. The div wrapper to capture bubbling isn't great but it's better than creating APIs that are incompatible with Argo. That sad, if this is optional then it might not be the end of the world but it might lead to suprising behaviour. You'd need want somebody on the Argo team to give more context on this, as I'm mostly parroting old talking points as I've not been working with polaris-react on a day-to-day basis for a while.

https://github.com/Shopify/polaris-react/pull/4111#issuecomment-848942107 - passing through an object containing exactly what is desired rather than tossing through the whole dom event might be an option

patrickracicot: Thanks @BPScott for the context.

After reading and gathering context, I went back to the code and tried to see how I could avoid adding the event in our API while still properly supporting the event bubbling.

**The fix**
 - On the wrapping div inside the `IndexTable/Checkbox`, we only now only listen to the bubbling `onClick` event.
    https://github.com/Shopify/polaris-react/blob/dfb406c6e4f75b751cbd4b3909ab3471bf896658/src/components/IndexTable/components/Checkbox/Checkbox.tsx#L36-L39

 - Inside the `Polaris/Checkbox` component, we stop the propagation of `onClick` and `onKeyUp` events emerging from the `spin.backdrop` DOM element 
    https://github.com/Shopify/polaris-react/blob/dfb406c6e4f75b751cbd4b3909ab3471bf896658/src/components/Checkbox/Checkbox.tsx#L180-L184

**The explanation**
To be able to bind on the `div.onClick` on the wrapper we need to make sure that only the click events coming from the `<Input />` control are bubbled up. Also since the activation (space for a checkbox) of a control will trigger the `onClick` event we need to ignore the `onKeyUp` events bubbled up to the wrapping div.

UNRELEASED.md
```diff
@@ -18,6 +18,8 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 
 - Fixed try-catch syntax error in `Modal` ([#4553](https://github.com/Shopify/polaris-react/pull/4553))
 - Fixed an issue with `TextField` where date and time were uneditable on click ([#4671](https://github.com/Shopify/polaris-react/pull/4671))
+- Reverted popover exit timeout to avoid race conditions ([#4633](https://github.com/Shopify/polaris-react/pull/4633))
```
dleroux: Did you mean to add this one?

src/components/Checkbox/tests/Checkbox.test.tsx
dleroux: you brought back the keyUp, does this still stand?

src/components/CheckableButton/tests/CheckableButton.test.tsx
dleroux: Same here? Since KeyUp is back, does this still stand?

UNRELEASED.md
patrickracicot: No this is weird seems like it was added during the rebase 🤷 

I'll make sure that this content is correct on my next rebase.

src/components/CheckableButton/tests/CheckableButton.test.tsx
patrickracicot: Although the `onKeyUp` handler is back on the `Checkbox.Input` element, we aren't managing it the same way this test intends.  The `onKeyUp` only manages the focusable state and no longer triggers value change (we let the subsequent click event triggered by the browser do this).

src/components/Checkbox/tests/Checkbox.test.tsx
patrickracicot: See https://github.com/Shopify/polaris-react/pull/4631#discussion_r755273209

dleroux (approved):
Code looks good, 🎩 too! Thanks @patrickracicot

PR #4636, opened 2021-11-12
Remove all color() invocations

### LINKED ISSUE #4592: Remove legacy functions color() and filter()
function | value
-- | --
color(purple, text) | rgb(80, 73, 90)
color(purple, darker) | rgb(35, 0, 81)
filter(purple, darker) | ...
filter(purple, dark) | ...

yurm04:

Fixes #4592

The `color()` Sass function should not be used as it uses a deprecated color system. Replacing all `color()` invocations in favor of the up to date [color system](https://github.com/Shopify/polaris-react/blob/main/documentation/Color%20system.md) that uses custom css properties

Swapping `color()` for equivalent css custom properties.

**Affected components**
* `AppProvider`
* `MappedOptions`
* `Badge`
* `Navigation`
* `DualThumb`

## 

### How to 🎩
1. Run the Storybook
2. Go through each of the affected components listed above. There should be no major visual differences


src/components/AppProvider/AppProvider.scss
```diff
@@ -46,7 +46,7 @@ body {
   min-height: 100%;
   margin: 0;
   padding: 0;
-  background-color: color('sky', 'light');
+  background-color: var(--p-background);
```
alex-page: Does the app provider have access to this variable? It should but it might not.

BPScott: `--p-background` and all the other theme properties get set on the div which is inside the `body` element so css applied to the body won't have access to those custom properties.

alex-page: Lets hard code this value and leave a comment on why it's hard coded. We might come back and fix it.

yurm04: @alex-page updated to hardcode value

src/styles/shared/_icons.scss
```diff
@@ -19,6 +19,7 @@
 
 @mixin color-icon($value, $hue: base) {
   svg {
+    // have to get rid of this
     fill: color($value, $hue);
```
alex-page: What is keeping this one around?

yurm04: nothing! removed since it's not being used

alex-page (commented):
The only thing left is to remove the function. We could point this to the `v8.0.0-major` branch.

BPScott (commented):
The removal of the `color-icon` mixin is a breaking change to our [public sass API](https://github.com/Shopify/polaris-react/blob/main/src/styles/_public-api.scss) - consumers used to be able to use that mixin and now it's gone. Up to y'all decide if that's acceptable in a minor release.


Personally I'd err towards trying to make sure these removal PRs occur in the main branch rather than a v8 branch (as then there will be fewer merge headaches in the future), but try to maintain backwards compatibility in minor versions - And thus leave the `color-icon` mixin hanging around, referencing the `color` function. Perhaps add a `@deprecated` docblock to color/color-icon, and have a follow-up PR that removes color/color-icon targeting a v8 branch.

---

Not blocking for this PR but an errant thought bouncing around my brain: 

It'd be neat to remove the ability to use `color()` internally to make sure usage doesn't creep back in (unlikely with color, but possible for other functions that are due to become tokens). As alluded to earlier [`_public-api.scss`](https://github.com/Shopify/polaris-react/blob/main/src/styles/_public-api.scss) is our public api, and removing items from that file or the files referenced therein is a breaking change. However we don't use that file internally. Internal sass API usage is done by importing [_common.scss](https://github.com/Shopify/polaris-react/blob/main/src/styles/_common.scss).

By removing the `@import './foundation/colors';` from `_common.scss` then internal files would not have access to the functions used within it.

I think a bit of extra work is needed to get to that point - either removing the usage of the other functions defined in that file, or splitting the files up so some functions are only exposed via `_public-api.scss`. I'm not sure quite sure how that'd look off the top of my head, or if it'd be worth doing as you go or waiting till most of these functions have been converted to tokens.

alex-page (approved):
Lets break a part removing functions for the v8 branch. Thanks @BPScott and @yurm04

PR #4638, opened 2021-11-12
[Page] Update the subtitle prop type from a string to a React node

LauraAubin:

Inventory history is looking to use the subtitle prop as a link to the product variant:

<kbd>Screen Shot 2021-11-12 at 3 02 41 PM</kbd>

This PR updates the page subtitle prop from a `string` to a `ReactNode` in order to pass in a link.

I couldn't find any documentation or direct reason not to make this change, but if this is something we don't want to ship then feel free to decline these changes and we can try to find an alternative approach for inventory 😄.

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -6,7 +6,7 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 
 ### Enhancements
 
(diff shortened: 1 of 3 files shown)


LauraAubin:
Closing this as we're going to just use the back button to visit the variant link.

PR #4641, opened 2021-11-12
[VisuallyHidden] Updating Visually Hidden CSS

sophschneider:

The visually hidden styles in polaris uses `top` which unnecessarily pushes hidden content to the top. it also uses the deprecated `clip` style. This PR updates the styles to use the GovUK design system visually hidden styles discussed [here](https://github.com/h5bp/main.css/issues/12). The styles have been tested extensively as seen [here](https://gist.github.com/nickcolley/19b80ed24d0364cfd3afd3b1b49c4014). I tested the styles myself on mac Voice Over. 

VO demo of old styles: https://videobin.shopify.io/v/1XbjO4
VO demo of new styles: https://videobin.shopify.io/v/bbj2Aj

old styles, note the `top` attribute, this can be much more drastic if the container is tall:
Screen Shot 2021-11-15 at 10 14 56 AM

new styles:
Screen Shot 2021-11-15 at 10 14 38 AM

updates styles for visually hidden

### Playground code
```
  const headings = ['hello', 'world', 'aria hidden'].map((heading) => {
    if (heading === 'aria hidden') {
      const accessibilityLabel = `I am visually hidden`;

      return (
        <>
          <div aria-hidden>I am aria hidden</div>
        </>
      );
    }

    return heading;
  });

  return (
    <Page title="Playground">
      <Card>
        <DataTable
          columnContentTypes={['text', 'text', 'text']}
          headings={headings}
          rows={[]}
          verticalAlign="middle"
        />
      </Card>
    </Page>
  );
```

### 🎩 checklist

### DIFF
--- a/src/styles/shared/_accessibility.scss
+++ b/src/styles/shared/_accessibility.scss
@@ -2,16 +2,13 @@
 /// accessibility tools.
 
 @mixin visually-hidden {
-  // Need to make sure we override any existing styles.
-  // stylelint-disable declaration-no-important
-  position: absolute !important;
-  top: 0;
-  left: 0;
-  clip: rect(1px, 1px, 1px, 1px) !important;
-  overflow: hidden !important;
-  height: 1px !important;
-  width: 1px !important;
-  padding: 0 !important;
-  border: 0 !important;
-  // stylelint-enable declaration-no-important
+  clip-path: inset(50%);
+  height: 1px;
+  width: 1px;
+  overflow: hidden;
+  position: absolute;
+  white-space: nowrap;
+  margin: 0;
+  padding: 0;
+  border: 0;
 }
--- a/src/styles/shared/_accessibility.scss
+++ b/src/styles/shared/_accessibility.scss
@@ -1,14 +1,16 @@
 /// Used to hide an element visually, but keeping it accessible for
 /// accessibility tools.
 
+/// styles referenced from GOV.UK design system
+/// https://github.com/h5bp/main.css/issues/12#issuecomment-451965809
 @mixin visually-hidden {
-  clip-path: inset(50%);
-  height: 1px;
-  width: 1px;
-  overflow: hidden;
   position: absolute;
-  white-space: nowrap;
+  width: 1px;
+  height: 1px;
   margin: 0;
   padding: 0;
+  overflow: hidden;
+  clip-path: inset(50%);
   border: 0;
+  white-space: nowrap;
 }
--- a/src/styles/shared/_accessibility.scss
+++ b/src/styles/shared/_accessibility.scss
@@ -6,6 +6,7 @@
   // stylelint-disable declaration-no-important
   position: absolute !important;
   top: 0;
+  left: 0;
   clip: rect(1px, 1px, 1px, 1px) !important;
   overflow: hidden !important;
   height: 1px !important;
--- a/src/styles/shared/_accessibility.scss
+++ b/src/styles/shared/_accessibility.scss
@@ -4,13 +4,16 @@
 /// styles referenced from GOV.UK design system
 /// https://github.com/h5bp/main.css/issues/12#issuecomment-451965809
 @mixin visually-hidden {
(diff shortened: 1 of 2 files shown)


alex-page (commented):
Adding some context from Slack to this issue. This is the approach I am most familiar with:

```
clip-path: inset(50%);
height: 1px;
overflow: hidden;
position: absolute;
white-space: nowrap; 
width: 1px;
```

`top` and `left` values should be removed. [`clip` is depreacted](https://developer.mozilla.org/en-US/docs/Web/CSS/clip). I would love to know if you have any opinions @sophschneider. I read this great blog post recently [diving deep into this problem](https://zellwk.com/blog/hide-content-accessibly/). I also appreciated the conversation in the [HTML5 boilerplate](https://github.com/h5bp/main.css/issues/12).

sophschneider:
@alex-page read the blog and thread, they were both super interesting and helpful! I'm going to use your snippet and test it on a bunch of browsers and screen readers, I might add in or take out some styles from the threads if I run into the same problems they did (announcing order, weird focus ring, etc.)

src/styles/shared/_accessibility.scss
```diff
@@ -1,16 +1,16 @@
...
-  // stylelint-enable declaration-no-important
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  margin: 0;
+  padding: 0;
+  overflow: hidden;
+  clip-path: inset(50%);
+  border: 0;
+  white-space: nowrap;
```
alex-page: Looks great, we just need the `!important` added back as removing that override could create a big risk.

sophschneider: sounds good, should I add it back on all of the same attributes?

kyledurand: 👍 I think we'll need !important on all of the properties here because we can't guarantee that consumers won't have some other rule that overrides the mixin.

We should add the comment back as well:
`// Need to make sure we override any existing styles.`

alex-page: This looks great! Thanks all!

alex-page (approved):

src/styles/shared/_accessibility.scss
```diff
@@ -1,16 +1,18 @@
...
-  overflow: hidden !important;
-  height: 1px !important;
   width: 1px !important;
+  height: 1px !important;
+  margin: 0 !important;
   padding: 0 !important;
+  overflow: hidden !important;
+  clip-path: inset(50%) !important;
   border: 0 !important;
-  // stylelint-enable declaration-no-important
```
kyledurand, line 15: Can we reenable this rule?

kyledurand (approved):
Looks good just make sure to reenable the no-important lint rule and then 🚢

src/styles/shared/_accessibility.scss
sophschneider: oh i just realized it enables it after disabling it! weird pattern `stylelint`, thanks for catching this :)

emma-boardman:
👋  Hi!

We've run into a couple of bugs in online-store-web as a result of this change. In one instance, we were able to create a workaround: 

- https://github.com/Shopify/online-store-web/pull/12220/files#r807737459
- https://github.com/Shopify/polaris-react/pull/5181

However, the bug also occurs on the ThemesIndex (video included below), and there are too many components using `VisuallyHidden` to override that CSS on a case-by-case. 

Would there be any strong objections to me re-adding `top:0` and releasing a v8 fix? 

cc @sophschneider, @alex-page , @alexandcote 

`GitIntegrationSheet` bug:

https://user-images.githubusercontent.com/17032120/154731670-87cb5749-d214-4f89-a50f-8648e16d4e1e.mov

sophschneider:
hi @emma-boardman! no objections from me but it is kind of weird that `VisuallyHidden` is being used for visual styling even if it is accidental, ideally it wouldn't change things visually at all

PR #4642, opened 2021-11-13
feat: Simplify ThemeProvider

### LINKED ISSUE #4703: Simplify ThemeProvider

aaronccasanova:

This PR introduces a `CustomProperties` component which serves as a simple solution for theming that is driven solely by CSS custom properties and the CSS `color-scheme` declaration.

Features of the new `CustomProperties` component:
- Converts tokens to `static` and `colorScheme` specific custom properties, adds them to a style sheet, and injects them into the `head` tag.
  (Note: This will only happen once regardless how many `CustomProperties` components are rendered)
- Allows nesting `CustomProperties` components by only injecting custom properties once and leveraging the `color-scheme` attribute to set theme styles.
- Includes an `as` prop to enable defining the element of the root node.

image

## Breaking Changes:

In addition to adding the `CustomProperties` component this PR removes the `ThemeProvider` and all of its `utilities`, `types`, and `hook`:
- Removed `ThemeProvider` component
- Updated existing usages of `ThemeProvider` to `CustomProperties`
- Removed all usages of the `ThemeProvider`, `ThemeContext`, and the `useTheme` hook
- Removed the `buildThemeContext` utility
- Removed `buildCustomProperties` utility
- Removed `Theme` type
- Removed `ProcessedThemeConfig` type
- Removed `AppThemeConfig` type
- Removed `ThemeConfig` type
- Removed `Role` and `RoleColors` type
- Removed `ThemeLogo` type

## Additional Changes/Updates
- Created a (WIP) top level `src/designTokens` directory
- Moved inline body `background-color` and `color` behavior from the legacy `ThemeProvider` to the `AppProvider` for backward compatibility
  * Note: We plan to revisit this behavior and validate if it is still necessary
  * Added `DEFAULT_COLOR_SCHEME` constant to keep the `AppProvider` and `CustomProperties` component in sync. (Note: This would not be necessary if all components explicitly applied `background-color` and `color` properties)
- Added tests for the `CustomProperties` component and stylesheet utilities

TODO:
- [x] Update README
- [x] Add tests

### How to 🎩

<details>

```jsx
import React from 'react';

import {Page, CustomProperties} from '../src';

const Container: React.FC = (props) => (
  <div
    {...props}
    style={{
      padding: 20,
      border: '1px solid var(--p-text)',
      backgroundColor: 'var(--p-background)',
      color: 'var(--p-text)',
    }}
  />
);

  return (
    <Page title="Playground">
      <Container>
        {/* Set in AppProvider */}
        Top level theme
        <CustomProperties colorScheme="dark">
          <Container>
            Nested theme
            <CustomProperties colorScheme="dim">
              <Container>Deeply nested theme</Container>
            </CustomProperties>
          </Container>
        </CustomProperties>
      </Container>
    </Page>
  );
}
```

</details>

### 🎩 checklist


src/components/ThemeProvider/customProperties.ts
```diff
@@ -0,0 +1,43 @@
+import {tokens} from './tokens';
+
+/**
+ * Adapted from: https://github.com/argyleink/gui-challenges/blob/main/color-schemes/style.css
+ */
+export const customProperties = /* css */ `
```
aaronccasanova: Example output:
Screen Shot 2021-11-12 at 2 41 53 PM

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -1,109 +1,45 @@
...
-  buildCustomProperties,
-  toString,
-  Tokens,
-} from '../../utilities/theme';
+import type {ColorScheme} from './tokens';
+import {customProperties} from './customProperties';
+
+const styleSheet = document.createElement('style');
+styleSheet.textContent = customProperties;
+styleSheet.dataset.polarisCustomProperties = '👋';
```
aaronccasanova: Example output:
Screen Shot 2021-11-12 at 2 48 16 PM

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -1,109 +1,45 @@
...
+    as: Component = 'div',
+    children,
+    className,
+    colorScheme = 'light',
+  } = props;
 
-    const theme = buildThemeContext(processedThemeConfig, customProperties);
+  React.useEffect(() => {
+    if (injectedCustomProperties) return undefined;
+    injectedCustomProperties = true;
```
aaronccasanova: Example of the effect being called in three nested `ThemeProvider`s and only injecting the custom properties once:
Screen Shot 2021-11-12 at 3 37 04 PM

aaronccasanova: Alternatively, we could determine if Polaris custom properties have already been injected with a `data-*` attribute on the style tag (instead of the module scoped `injectedCustomProperties` variable). For example:
```jsx
const style = document.createElement('style');
style.dataset.polarisCustomProperties = 'false';

...

React.useEffect(() => {
  if (JSON.parse(style.dataset.polarisCustomProperties)) return undefined;
  style.dataset.polarisCustomProperties = 'true';
  document.head.appendChild(style);
}, []);
```

aaronccasanova: Discussed offline, the module scoped `injectedCustomProperties` variable is sufficient.

src/components/AppProvider/AppProvider.tsx
```diff
@@ -66,28 +69,43 @@ export class AppProvider extends Component<AppProviderProps, State> {
...
 
+  setBodyStyles = () => {
+    // Inlining the following custom properties to maintain backward
+    // compatibility with the legacy ThemeProvider implementation.
+    document.body.setAttribute(
+      'color-scheme',
+      this.props.colorScheme || DEFAULT_COLOR_SCHEME,
+    );
+    document.body.style.backgroundColor = 'var(--p-background)';
+    document.body.style.color = 'var(--p-text)';
```
aaronccasanova, line 106: TODO in a separate issue: Verify this backward compatibility solution is still necessary and identify a deprecation path.

src/components/Frame/components/ContextualSaveBar/ContextualSaveBar.tsx
```diff
@@ -23,7 +21,9 @@ export function ContextualSaveBar({
   contextControl,
 }: ContextualSaveBarProps) {
   const i18n = useI18n();
-  const {logo} = useTheme();
+  // TODO: This behavior will be re-enabled in a separate PR.
+  const logo = undefined as {[key: string]: any} | undefined;
```
aaronccasanova, line 26: Drawing attention here to ensure this is captured in the following PR: https://github.com/Shopify/polaris-react/pull/4667

src/components/Frame/components/ContextualSaveBar/ContextualSaveBar.tsx
```diff
@@ -113,7 +113,7 @@ export function ContextualSaveBar({
 
   return (
     <>
-      <ThemeProvider theme={{colorScheme: 'inverse'}}>
+      <div>
```
aaronccasanova: **todo:** Verify this change in web.

src/components/Frame/components/ContextualSaveBar/tests/ContextualSaveBar.test.tsx
```diff
@@ -220,75 +220,6 @@ describe('<ContextualSaveBar />', () => {
       });
     });
   });
-
```
aaronccasanova, line 223: Drawing attention here to ensure this is captured in the following PR: https://github.com/Shopify/polaris-react/pull/4667

src/components/Frame/components/Toast/Toast.tsx
```diff
@@ -64,13 +63,13 @@ export function Toast({
   const className = classNames(styles.Toast, error && styles.error);
 
   return (
-    <ThemeProvider theme={{colorScheme: 'inverse'}}>
+    <div>
```
aaronccasanova: **todo:** Verify this update in web.

aaronccasanova: Is this a breaking change?

https://user-images.githubusercontent.com/32409546/142951189-c263354b-39be-4fdb-8202-c3cdfa8a551b.mov

src/components/Navigation/Navigation.tsx
```diff
@@ -30,7 +29,8 @@ export const Navigation: React.FunctionComponent<NavigationProps> & {
   onDismiss,
   ariaLabelledBy,
 }: NavigationProps) {
-  const {logo} = useTheme();
+  // TODO: This behavior will be re-enabled in a separate PR.
+  const logo = undefined as {[key: string]: any} | undefined;
```
aaronccasanova, line 33: Drawing attention here to ensure this is captured in the following PR: https://github.com/Shopify/polaris-react/pull/4667

src/components/Navigation/tests/Navigation.test.tsx
```diff
@@ -7,9 +7,10 @@ import {Image} from '../../Image';
 import {WithinContentContext} from '../../../utilities/within-content-context';
 
 describe('<Navigation />', () => {
-  it('renders an image if the theme provider is present', () => {
+  // eslint-disable-next-line jest/no-disabled-tests
+  it.skip('renders an image if the theme provider is present', () => {
     const navigation = mountWithApp(<Navigation location="/" />, {
-      theme: {logo: {url: 'https://shopify.com/logo'}},
+      // theme: {logo: {url: 'https://shopify.com/logo'}},
```
aaronccasanova, line 13: Drawing attention here to ensure this is captured in the following PR: https://github.com/Shopify/polaris-react/pull/4667

src/components/TopBar/TopBar.tsx
```diff
@@ -61,7 +60,8 @@ export const TopBar: React.FunctionComponent<TopBarProps> & {
   contextControl,
 }: TopBarProps) {
   const i18n = useI18n();
-  const {logo} = useTheme();
+  // TODO: This behavior will be re-enabled in a separate PR.
+  const logo = undefined as {[key: string]: any} | undefined;
```
aaronccasanova, line 64: Drawing attention here to ensure this is captured in the following PR: https://github.com/Shopify/polaris-react/pull/4667

src/components/TopBar/tests/TopBar.test.tsx
```diff
@@ -170,62 +171,67 @@ describe('<TopBar />', () => {
...
-    it('will render an unstyled link with a default width', () => {
-      const topBar = mountWithApp(<TopBar />, {
-        theme: {logo: {}},
-      });
-      expect(topBar).toContainReactComponent(UnstyledLink, {
-        style: {width: '104px'},
-      });
-    });
-  });
+  // describe('logo', () => {
```
aaronccasanova, line 174: Drawing attention here to ensure this is captured in the following PR: https://github.com/Shopify/polaris-react/pull/4667

src/components/Frame/components/Toast/Toast.tsx
alex-page: Not a breaking change. Should we wrap this in a `dark` themeprovider? To keep it dark in lightmode? Darkmode isn't a thing so it persisting to dark when turning darkmode on is fine.

alex-page (approved):
This is in a good place. There is definitely follow up work but happy for this to merge as is.

.storybook/preview.js
```diff
@@ -23,22 +21,8 @@ function AppProviderWithKnobs({colorScheme, children}, context) {
 
   if (omitAppProvider === 'true') return children;
 
-  const colors = Object.entries(DefaultThemeColors).reduce(
```
BPScott: I reckon we can remove this storybook behaviour on `main`, rather than it needing to be v8 branch specific.

It looks like that `strToHex` is no longer used so it can be removed from the bottom of the file

src/components/TopBar/TopBar.tsx
```diff
@@ -108,10 +108,11 @@ export const TopBar: React.FunctionComponent<TopBarProps> & {
           url={logo.url || ''}
           className={styles.LogoLink}
           style={{width}}
+          aria-label={logo.accessibilityLabel || ''}
```
BPScott, line 111: This change can be done on `main` if it is still required.

src/designTokens/colorSchemes.ts
```diff
@@ -0,0 +1,38 @@
+import type {Tokens} from './designTokens';
```
BPScott, line 1: The repo tends to only name files containing react components as camelcase. Everything else tends to be sentence case. Worth renaming to `src/design-tokens/color-schemes` to help follow that convention?

src/components/Frame/components/Toast/Toast.tsx
aaronccasanova: I like that suggestion! https://github.com/Shopify/polaris-react/pull/4642/commits/c9dbb15d6b35e211a4f4f342c38a874ef4d40418

src/components/Frame/components/ContextualSaveBar/ContextualSaveBar.tsx
aaronccasanova: **update:** Explicitly set this to `dark` per the following comment: https://github.com/Shopify/polaris-react/pull/4642#discussion_r754732597

src/components/Frame/components/Toast/Toast.tsx
BPScott: This is breaking. We currently support setting a `colorScheme` on the AppProvider of either light or dark. Setting `inverse` here sets the scheme to be the opposite of  whatever its parent color sheme.

- setting a `colorScheme:'light'` on AppProvider (or omitting as light is the default), means the toast will be dark
- setting a `colorScheme:'dark'` on AppProvider (or omitting as light is the default), means the toast will be light.

Theoretically this can persist through multiple levels of ThemeProvider - starting light, inverting, then inverting again would give a light theme, but I'm not sure if that's every actually been needed.

We can't just do "colorScheme=dark" in these places where inverse was used  because then you break rendering when using a dark color scheme at the AppProvider level - you'd end up with a black backgrounded toast on a black screen - rather than a white backgrounded toast on a black screen.

src/components/Navigation/Navigation.tsx
```diff
@@ -39,10 +39,11 @@ export const Navigation: React.FunctionComponent<NavigationProps> & {
         url={logo.url || ''}
         className={styles.LogoLink}
         style={{width}}
+        alt={logo.accessibilityLabel || ''}
```
BPScott, line 42: Can be done on main if still desired

src/components/TopBar/tests/TopBar.test.tsx
BPScott: Silly nit:

```js
  describe.skip('logo', () => {
```

Would let you skip the tests without having to comment all of them out

src/utilities/theme/tokens.ts
```diff
@@ -5,12 +5,6 @@ export const Tokens = {
   borderRadiusWide: rem('8px'),
   borderRadiusFull: '50%',
 
-  // Font-Weights
-  fontWeightRegular: 400,
```
BPScott: Breaking change. Should this removal (along with other tokens here) be adjusted elsewhere?

BPScott (commented):
I'm pretty sure inverse support will still be needed if we want want to support both light and dark color schemes (we talked about this a bunch in tuple)

alex-page:
@bpscott Quick clarification on `inverse`. We are not going to support this feature in version 8. We may explore it in the future. Reasons why:
1. We don't know when we are going to support dark mode
1. Inverse colors is based off the current design which is likely to change
1. We have not seen other design systems use the inverse pattern

BPScott:
> We don't know when we are going to support dark mode

That makes things simpler for now. In that case we can straight up rip out the `colorScheme` option from AppProvider and say that as a breaking change we're removing dark mode.

Have you got any thoughts on how to handle the the current cases where we render dark-style components in an otherwise light theme - Toasts, some PopoverOverlays, and everything that gets rendered inside a [ContextualSaveBar](https://storybook.polaris.shopify.com/?path=/story/all-components-contextual-save-bar--default-contextual-save-bar) (which includes other components like buttons)

aaronccasanova:
> Have you got any thoughts on how to handle the the current cases where we render dark-style components in an otherwise light theme - Toasts, some PopoverOverlays, and everything that gets rendered inside a ContextualSaveBar (which includes other components like buttons)

We would simply render the [`CustomProperties` component with `colorScheme="dark"`](https://github.com/Shopify/polaris-react/blob/9027664353cc719e5f5d12f9a7b3b7b6f89e83c2/src/components/Frame/components/Toast/Toast.tsx#L67):
image

The only behavior we are removing in this PR is `inverting` color-schemes based on their parent color-scheme. On that note, I'm thinking there are other patterns we can explore, instead of `inverse`, and we will be making time for that after we improve our tokens coverage in general.

src/components/CustomProperties/CustomProperties.tsx
```diff
@@ -0,0 +1,56 @@
...
+    document.head.appendChild(styleSheet);
+  }, []);
+
+  return (
+    <Component
+      color-scheme={colorScheme}
+      className={className}
+      // TODO: Remove this inline style when we update individual components
+      // to set their own color and background-color properties.
+      style={{color: 'var(--p-text)'}}
```
kyledurand, line 51: I forget, is this for nesting? Do we still need this if we're explicitly setting an inline style on body?

.storybook/preview.js
aaronccasanova: Good idea! I will leave this and similar comments unresolved and create a separate task to get all these changes on the main branch 👍

src/components/TopBar/TopBar.tsx
aaronccasanova: https://github.com/Shopify/polaris-react/pull/4642#discussion_r755563563

.storybook/preview.js
alex-page: PR here https://github.com/Shopify/polaris-react/pull/4700

src/components/Navigation/Navigation.tsx
aaronccasanova: https://github.com/Shopify/polaris-react/pull/4642#discussion_r755563563

src/utilities/theme/tokens.ts
aaronccasanova: Good callout! I reverted this update 👍  Spoke with @alex and we plan to create a separate issue to remove these tokens and other associated processes.

alex-page (approved):

src/designTokens/colorSchemes.ts
aaronccasanova: I have no objections to that 👍  I will handle this in a follow up task.

src/components/CustomProperties/styles.ts
```diff
@@ -0,0 +1,101 @@
...
+}
+
+@media (prefers-color-scheme: dark) {
+  :root {
+    ${darkDeclarations}
+  }
+}
+
+${getColorSchemeRules(designTokens, osColorSchemes)}
+`;
```
kyledurand, line 101: Is this actually something we want to ship in v8?

src/components/TopBar/TopBar.tsx
kyledurand: I think this is an enhancement. Can you add an issue to track this and add to main @aaronccasanova ?

kyledurand (approved):
Looks great at a glance, we can iterate if we need to. Can you rename the camel case file?

src/components/CustomProperties/CustomProperties.tsx
aaronccasanova: Yes, we still need this since most components inherit the `color` property. I thinking we wouldn't encounter this issue if each component set its background and text color. We created a follow up ticket here to address this issue: https://github.com/Shopify/polaris-react/issues/4709

src/components/CustomProperties/styles.ts
aaronccasanova: Probably not now that I think about it. Here is a follow issue to remove this: https://github.com/Shopify/polaris-react/issues/4710

src/components/TopBar/TopBar.tsx
aaronccasanova: Tracking the issue here: https://github.com/Shopify/polaris-react/issues/4702

aaronccasanova:
Thanks for the feedback everyone! In the interest of unblocking other folks working on tokens I will be merging this PR as is. @alex-page and I spun up a number of tickets that came out of the above discussions as well as other issues we encountered throughout  the process: https://github.com/Shopify/polaris-react/issues/4588

Feedback is still welcome as we plan to continue iterating on this approach in the v8 branch and feel free to create more issues for anything we may have missed!

src/components/AppProvider/AppProvider.tsx
aaronccasanova: https://github.com/Shopify/polaris-react/issues/4708

src/components/TopBar/tests/TopBar.test.tsx
aaronccasanova: For sure 👍  This we be immediately re-enabled in Kyles upcoming PR

PR #4643, opened 2021-11-14
Reduce the types of absolute imports that are considered valid

BPScott:

A follow up to #4561. I realised a better fix for aiming to shrink the
number of absolute paths.

By removing `baseUrl` we remove support for all files being able to be
arbitrarily accessed (e.g. `src/components/BLAH`).
Now only paths mentioned in the paths object can be accessed absolutely.

After this PR is merged the only valid absolute imports are `tests/SOMETHING` and `components`. `src/FOO`, `tests` and `components/FOO` are all considered invalid.

- Removes `baseUrl` from tsconfig.json, meaning arbitrary absolute paths like "src/components/BLAH` are no longer possible - the only usable absolute paths are those specified in the paths object
- Adds a path aliases for `tests/*` that point to the `tests` folder. In reality this will only be used for the `tests/utilities` folder.
- Replaces all usage of importing `components/X` with a relative import
- Remove path alias form `components/X` from tsconfig and jest module mapper.

### How to 🎩

- Tests and type check passes
- Try importing something from `src/components` and note that it fails.

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -12,6 +12,8 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 
 ### Development workflow
 
+- Removed `baseUrl` from `tsconfig.json`. Attempting to do an absolute import from `src/BLAH` now results in a error when type-checking. ([#4643](https://github.com/Shopify/polaris-react/pull/4643))
+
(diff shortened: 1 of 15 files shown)


alex-page (approved):
💯

Issue #4528, opened 2021-10-20
polaris-react 90% token coverage for color, motion, type, spacing


alex-page:
## Project links

- [Project brief](https://docs.google.com/document/d/1C5JvYXYcm2wwhAWvofZaYSGmz9PZvOcVu9DOoXzwBvk/edit)
- [Vault project](https://vault.shopify.io/projects/22300)

## Summary

As of today we cannot roll out a new style to the admin or polaris-react components without updating multiple repos, thousands of lines of code, Sass functions and CSS custom properties. Teams building new components have multiple different ways to use tokens and the documentation is in multiple places. The goal of this project is to guarantee every component in polaris-react used in shopify/web is built with simple tokens from one place.

This work will create a platform for teams to build custom components in shopify/web and enable us to ship large design infrastructure changes through our components.

**Intended outcomes**
- 100% of polaris-react components are built with tokens.
- Consolidate tokens into one place and use CSS custom properties.
- Teams have one implementation of tokens to use.
- Remove all deprecated css custom properties and token related Sass functions.
- Remove runtime theming.
- polaris.shopify.com includes simple documentation for tokens.
- 2022 - New major version of polaris-react shipped to @shopify/web.

**Key assumptions**
- shopify/web will use the updated css custom properties.
- Centralising the tokens will make it easier to identify foundational components.
- Existing tokens are limiting and confusing to use.
- It’s easier to hard code design values than use tokens today.

**Guiding principles**
- Simplification over complexity.
- Prioritise a tolerable solution not a new token system or library.
- Get shit done.

## Breakdown

**Infrastructure**
- [x] #4586
- [x] #4587
- [x] #4588
- [x] #4589
- [x] #4590
- [x] #4591
- [x] #4732

**Color**

- [x] #4592
- [x] #4593

**Motion**
- [x] #4595
- [x] #4596
- [x] #4597

**Typography**
- [x] #4598
- [x] #4599
- [x] #4600
- [x] #4601
- [x] #4602

**Spacing**

- [x] #4603
- [x] #4604
- [x] #4605
- [x] #4606

**z-index**

- [x] #4607
- [x] #4608

**Layout**

- [x] #4609
- [x] #4610

**Border radius**

- [x] #4611
- [x] #4612
- [x] #4613

**Border width**

- [x]  #4891

**Shadow**

- [x] #4614
- [x] #4615
- [x] #4616

**Documentation**

[...]

PR #4650, opened 2021-11-16
Remove filter functions

yurm04:

Related to #4528 
Fixes `filter()` part of #4592 

Removing all instances of `filter()` Sass function because the color system used by this function has been deprecated. Swapping out any references to `filter()` with the [color system](https://github.com/Shopify/polaris-react/blob/main/documentation/Color%20system.md) custom css properties

## 

### How to 🎩
Most changes are in the `Navigation` component. Run Storybook and/or visit the [Chromatic](https://www.chromatic.com/build?appId=5d559397bae39100201eedc1&number=8945) link to this branch to ensure there are no visual regressions.


src/components/Navigation/Navigation.scss
```diff
@@ -208,14 +208,14 @@ $disabled-fade: 0.6;
     @include recolor-icon(
```
alex-page, line 208: Is this function used a bunch? I would assume it's a filter related function. If it is we could move it to this file.

alex-page (commented):
I think we should remove the `filter()` function and point this to `v8.0.0-major` branch. Looking great so far! We will have to add documentation for migration and a breaking change to the changelog.

alex-page (approved):
Looks great for the first cut of removing the filter function references!

The files changed need a rebase/merge or something 👍

src/components/Navigation/Navigation.scss
```diff
@@ -208,14 +208,14 @@ $disabled-fade: 0.6;
...
-      $filter-color: filter('action')
+      $filter-color: var(--p-action-primary)
     );
     background: var(--p-background-hovered);
 
     @media (-ms-high-contrast: active) {
       @include recolor-icon(
         $fill-color: var(--p-icon-on-interactive),
-        $filter-color: filter('white')
+        $filter-color: var(--p-icon-on-interactive)
```
BPScott, line 218: These filter-color values passed to the 3rd argument of recolor-icon  are sent as the [`filter`](https://developer.mozilla.org/en-US/docs/Web/CSS/filter) property. They're not actual color values, they're filters like [`brightness(0) saturate(100%) invert(100%)`](https://github.com/Shopify/polaris-tokens/blob/main/dist/color-filters.color-map.scss#L138).

I think changing these to colors is going to break the behaviour, which is to allow for the recoloring of icons that are passed as svg datauris and rendered with an img tag

PR #4654, opened 2021-11-16
[ActionList] Wrap MenuItem Focus Arrow Navigation

mbswart:
Works towards  #4419

This PR will be merged into Patrick's PR: https://github.com/Shopify/polaris-react/pull/4505 and will add arrow warapping focus navigation.
The goal of this PR is to wrap Focus For menuitems in an actionlist. 

Focus wrap is done by adding a few new functions to `focus.ts`:
1. `getMenuFocusableDescendants` --> finds all focusable children of an element with the role `menuitem`
2. `getCurrentFocusedElementIndex` --> of a list of focusable elements, find the index of the currently focused item (important for knowing when at top or bottom of list in order to wrap)
3. `wrapFocusPreviousFocusableMenuItem` --> move focus to previous node or if at the top, wrap to the bottom element in the list
4. `wrapFocusNextFocusableMenuItem` --> move focus to next node or if at the bottom, wrap to the top element.

I also changed that actionList should expect "menuitem" as actionRole. This is because the actionRole is applied to the individual items if the item objects themselves don't specify a role themselves. If the role is "menuitem" the section role will default to "menu" and surrounding `<li>` objects to "presentation". This will also mean for these changes to applied to actionLists in admin, a user just needs to add `actionRole={"menuitem"} instead of individually adding `menuitem` to each item.

There is an issue that Patrick and I determined would be out of scope for this particular PR. If there is an active item, `ScrollTo` is called so that the active element is scrolled to. It does this by adding an anchor element. There are two issues here:
1. If the first element is active, when the popover autofocus is set to `first-node`, the first focusable node it finds is the anchor element that was scrolled to, breaking focus wrapping
2. Although the scroll to scrolls to the active element, focus is still controlled by the popover and focus will be set to the first node in the list if the popover has autofocus set to `first-node`, not to the active element like it should

## 

### How to 🎩
I have been using `yarn storrybook` and testing arrow navigation from there. The README.md in `src/components/ActionList` defines the actionList storybook's components if you wanted to test with more items. Again if there is an active item, focus is broken. Because to fix focus with active items would require changes to popover or scroll to and focus behaviour isn't worsened, to keep this PR scoped to `ActionList`, it is not addressed here.

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -8,6 +8,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - Keyboard arrow navigation support added in `ActionList` ([#4505](https://github.com/Shopify/polaris-react/pull/4505))
 - Menu role attribute value support added in `ActionList/Section` ([#4505](https://github.com/Shopify/polaris-react/pull/4505))
+- Keyboard wrapping arrow navigation added in `ActionList` ([#4654](https://github.com/Shopify/polaris-react/pull/4654))
 
 ### Bug fixes
 
--- a/src/components/ActionList/components/Item/Item.tsx
+++ b/src/components/ActionList/components/Item/Item.tsx
@@ -120,7 +120,7 @@ export function Item({
   );
 
   return (
-    <li>
+    <li role={role === 'menuitem' ? 'presentation' : undefined}>
       {scrollMarkup}
       {control}
     </li>
--- a/src/components/ActionList/components/Section/Section.tsx
+++ b/src/components/ActionList/components/Section/Section.tsx
@@ -68,7 +68,7 @@ export function Section({
     case 'option':
       sectionRole = 'presentation';
       break;
-    case 'menu':
+    case 'menuitem':
       sectionRole = 'menu';
       break;
     default:
--- a/src/utilities/focus.ts
+++ b/src/utilities/focus.ts
@@ -177,7 +177,7 @@ function getCurrentFocusedElementIndex(
     }
     currentItemIdx++;
   }
-  return currentItemIdx;
+  return currentItemIdx === allFocusableChildren.length ? -1 : currentItemIdx;
 }
 
 export function wrapFocusPreviousFocusableMenuItem(
@@ -189,11 +189,13 @@ export function wrapFocusPreviousFocusableMenuItem(
     allFocusableChildren,
     currentFocusedElement,
   );
-
-  if (currentItemIdx === 0) {
-    allFocusableChildren[allFocusableChildren.length - 1].focus();
+  if (currentItemIdx === -1) {
+    allFocusableChildren[0].focus();
   } else {
-    allFocusableChildren[currentItemIdx - 1].focus();
+    allFocusableChildren[
+      (currentItemIdx - 1 + allFocusableChildren.length) %
+        allFocusableChildren.length
+    ].focus();
   }
 }
 
@@ -206,11 +208,12 @@ export function wrapFocusNextFocusableMenuItem(
     allFocusableChildren,
     currentFocusedElement,
   );
-
-  if (currentItemIdx === allFocusableChildren.length - 1) {
+  if (currentItemIdx === -1) {
     allFocusableChildren[0].focus();
   } else {
-    allFocusableChildren[currentItemIdx + 1].focus();
+    allFocusableChildren[
+      (currentItemIdx + 1) % allFocusableChildren.length
+    ].focus();
   }
 }
 
--- a/src/utilities/focus.ts
+++ b/src/utilities/focus.ts
@@ -9,7 +9,8 @@ const FOCUSABLE_SELECTOR =
   'a,frame,iframe,input:not([type=hidden]):not(:disabled),select:not(:disabled),textarea:not(:disabled),button:not(:disabled),*[tabindex]';
 const KEYBOARD_FOCUSABLE_SELECTORS =
   'a,frame,iframe,input:not([type=hidden]):not(:disabled),select:not(:disabled),textarea:not(:disabled),button:not(:disabled),*[tabindex]:not([tabindex="-1"])';
-
+const MENUITEM_FOCUSABLE_SELECTORS =
+  'a[role="menuitem"],frame[role="menuitem"],iframe[role="menuitem"],input[role="menuitem"]:not([type=hidden]):not(:disabled),select[role="menuitem"]:not(:disabled),textarea[role="menuitem"]:not(:disabled),button[role="menuitem"]:not(:disabled),*[tabindex]:not([tabindex="-1"])';
 export const handleMouseUpByBlurring: MouseUpBlurHandler = ({currentTarget}) =>
   currentTarget.blur();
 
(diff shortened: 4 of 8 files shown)


UNRELEASED.md
```diff
@@ -8,6 +8,7 @@ Use [the changelog guidelines](https://git.io/polaris-changelog-guidelines) to f
 
 - Keyboard arrow navigation support added in `ActionList` ([#4505](https://github.com/Shopify/polaris-react/pull/4505))
 - Menu role attribute value support added in `ActionList/Section` ([#4505](https://github.com/Shopify/polaris-react/pull/4505))
+- Keyboard wrapping arrow navigation added in `ActionList` ([#4654](https://github.com/Shopify/polaris-react/pull/4654))
```
mbswart, line 11: This line doesn't really give any additional information that the above changes by Patrick doesn't, but CI requires a `UNRELEASED.md` change to pass. Once this is merged to Patrick's PR we can take this out

src/components/ActionList/components/Section/Section.tsx
```diff
@@ -86,7 +86,12 @@ export function Section({
   );
 
   return hasMultipleSections ? (
-    <li className={styles.Section}>{sectionMarkup}</li>
+    <li
+      className={styles.Section}
+      role={sectionRole === 'menu' ? 'presentation' : undefined}
+    >
```
patrickracicot: (request) According to ARIA best practices, even in nested menu context (or multi-section like here), the `<li />` should not have a `role="menu"`. The menu role should be reserved to the `<ul />`

src/utilities/focus.ts
```diff
@@ -158,6 +159,61 @@ export function focusLastKeyboardFocusableNode(
...
+  parentElement: HTMLElement,
+  currentFocusedElement: HTMLElement,
+) {
+  const allFocusableChildren = getMenuFocusableDescendants(parentElement);
+  const currentItemIdx = getCurrentFocusedElementIndex(
+    allFocusableChildren,
+    currentFocusedElement,
+  );
+
+  if (currentItemIdx === allFocusableChildren.length - 1) {
```
aktill: Your solution works great! Just because we talked about it, this would be the modulus version of the if/else:
` allFocusableChildren[
    (currentItemIdx + 1) % allFocusableChildren.length
  ].focus();`

src/utilities/focus.ts
```diff
@@ -158,6 +159,61 @@ export function focusLastKeyboardFocusableNode(
...
+  parentElement: HTMLElement,
+  currentFocusedElement: HTMLElement,
+) {
+  const allFocusableChildren = getMenuFocusableDescendants(parentElement);
+  const currentItemIdx = getCurrentFocusedElementIndex(
+    allFocusableChildren,
+    currentFocusedElement,
+  );
+
+  if (currentItemIdx === 0) {
```
aktill: Like the previous comment, the modulus version is `allFocusableChildren[
    (currentItemIdx - 1 + allFocusableChildren.length) %
      allFocusableChildren.length
  ].focus();`

src/utilities/focus.ts
```diff
@@ -158,6 +159,61 @@ export function focusLastKeyboardFocusableNode(
...
+) {
+  let currentItemIdx = 0;
+
+  for (const focusableChild of allFocusableChildren) {
+    if (focusableChild === currentFocusedElement) {
+      break;
+    }
+    currentItemIdx++;
+  }
+  return currentItemIdx;
```
patrickracicot: (request) To mimic [Array.prototype.findIndex()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) we should return `-1` if `currentFocusedElement` is not found in `allFocusableChildren`

```suggestion
  return currentItemIdx === allFocusableChildren.length ? -1 : currentItemIdx;
```

src/utilities/focus.ts
```diff
@@ -158,6 +159,61 @@ export function focusLastKeyboardFocusableNode(
...
+  const allFocusableChildren = getMenuFocusableDescendants(parentElement);
+  const currentItemIdx = getCurrentFocusedElementIndex(
+    allFocusableChildren,
+    currentFocusedElement,
+  );
+
+  if (currentItemIdx === allFocusableChildren.length - 1) {
+    allFocusableChildren[0].focus();
+  } else {
+    allFocusableChildren[currentItemIdx + 1].focus();
```
patrickracicot: (request) This will break if `currentFocusedElement` is not part of the `allFocusableChildren` list.

I would suggest that in the `else` statement, if the currently focused DOM element is not a child of `allFocusableChildren` we automatically set the focus on the first element of this list.

aktill:
🎩  ran great! Left a couple of comments just for info (not required)

patrickracicot (changes requested):
Love the idea of recycling the `actionRole` property!

src/components/ActionList/components/Item/Item.tsx
```diff
@@ -120,7 +120,7 @@ export function Item({
   );
 
   return (
-    <li>
+    <li role={role === 'menuitem' ? 'presentation' : undefined}>
```
aktill, line 123: This might just be me, but it took me a little investigation to wrap my brain around. I'm halfway wondering if a code comment might help to explain why it's there? Will wait for others to comment

aktill (approved):

src/components/ActionList/components/Item/Item.tsx
patrickracicot: This isn't a necessary change, but was suggested by @svinkle.  Doing this makes sure that the `<li>` element is excluded for accessibility semantics.

svinkle: The expected, valid HTML structure:

```html
<ul role="menu">
  <li role="presentation">
    <button role="menuitem">Pay now</button>
  </li>
<ul>
```

Browser generated accessibility tree consumed by assistive technology:

```
- Menu
  - "Pay now", Menu item
```

The `li` is structural but is not required to convey the Menu pattern. `role="presentation"` removes element semantic meaning.

aktill: Cool, thanks!

src/components/ActionList/components/Section/Section.tsx
mbswart: Ah the idea of this was when we have multiple sections they themselves are wrapped in `<li>`. If we know the section is menu I thought it wouldn't be a bad idea to make the sections themselves have a presentation role since they are more structural than having meaning (my understanding). So that was the idea behind the `role={sectionRole === 'menu' ? 'presentation' : undefined}` prop, "if the contained section is a menu, make the container `li>` presentational". I took it out now because I don't think it is adding anything important but the bellow structure was the idea:

```
<ul>
    <li role="presentation">.              <---- role here
        <ul role="menu">
          <li role="presentation">
            <button role="menuitem">Pay now</button>
          </li>
        </ul>
    </li>
    <li role="presentation">                       <---- role here
        <ul role="menu">
          <li role="presentation">
            <button role="menuitem">Pay now</button>
          </li>
        </ul>
    </li>
</ul>
```

src/utilities/focus.ts
mbswart: Smart! Thank you!

src/components/ActionList/components/Section/Section.tsx
patrickracicot: You are totally right ! I completely miss read the condition here 🤦 (I read like if you were assigning `role="menu"` on the `<li />`.

Now for this case here adding `presentation` makes sens and will serve the same purpose as https://github.com/Shopify/polaris-react/pull/4654#discussion_r752528004

patrickracicot (approved):
I 🎩  and code is LGTM

ghost:
🎉 Thanks for your contribution to Polaris React!

PR #4655, opened 2021-11-16
Remove remaining color sass function uses

yurm04:

Follow up to #4636, there were a few `color()` functions that got left behind. Cleaning up the rest.

### How to 🎩


src/components/Navigation/_variables.scss
```diff
@@ -4,7 +4,7 @@ $item-line-height-small: rem(32px);
 $item-line-height-large: rem(36px);
 // This is the only place this color is used.
 // stylelint-disable-next-line color-no-hex
-$item-selected-background: rgba(color('indigo'), 0.12);
+$item-selected-background: rgba(var(--p-background-selected), 0.12);
```
yurm04: this isn't actually used anywhere but assuming this should be removed in a followup PR

src/styles/shared/_buttons.scss
```diff
@@ -199,9 +199,6 @@
   background: transparent;
   // border-color: rgba($outline-color, 0.25);
```
alex-page: We should probably get this line too!

src/components/Navigation/_variables.scss
alex-page: Happy to remove it now 👍

src/components/Button/Button.scss
```diff
@@ -487,7 +487,7 @@ $stacking-order: (
 
   // Because the outline border color has a 40% opacity, the left border appears darker than the rest of the borders because they are layered over one another. Reducing the opacity to zero for the connected disclosure when not focused gives us the expected border color.
   &.outline:not(:focus) {
-    border-left-color: rgba(color('ink', 'lighter'), 0);
+    border-left-color: rgba(var(--p-border-disabled), 0);
```
BPScott: I believe rgba is a sass function, I don't think it will have the desired effect when you pass in a css variable, as the color won't be resolved till runtime

src/components/Navigation/_variables.scss
BPScott: Only files referenced by src/styles/_public-api.scss are part of the public API. I've not checked but I suspect this file is only used by the styles for specific components (e.g. Navigation/Navigation.scss), and is never referenced by _public-api.scss. If that is the case then we are free to delete this in the main branch without any backwards compatability concerns and you are free to nuke it now.

src/components/Button/Button.scss
yurm04: I didn't know this! Thanks @BPScott 

@alex-page what's the best way forward on this? hard code the color? leave for now and create a new issue? is `rgba()` going to be removed eventually as well?

src/components/Navigation/_variables.scss
yurm04: confirmed that this is not referenced in the `_public-api.scss` file (and isn't even used by `Navigation.scss` or elsewhere. Removing this line entirely.

src/components/Button/Button.scss
alex-page: Lets hard code the color to unblock this PR. It will be a failure on `polaris-coverage` so we won't miss it.

alex-page (approved):

src/components/Button/Button.scss
```diff
@@ -150,15 +150,15 @@ $stacking-order: (
 }
 
 .outline {
-  @include button-outline(color('ink', 'lighter'));
+  @include button-outline(var(--p-border));
 
   &.disabled {
-    @include button-outline-disabled(color('ink', 'lighter'));
+    @include button-outline-disabled(var(--p-border-disabled));
```
BPScott: The `$outline-color` parameter doesn't seem to be used anymore. https://github.com/Shopify/polaris-react/blob/main/src/styles/shared/_buttons.scss#L198-L205

Perhaps we can just remove passing in the argument entirely: `@include button-outline-disabled()`

src/components/Button/Button.scss
```diff
@@ -150,15 +150,15 @@ $stacking-order: (
 }
 
 .outline {
-  @include button-outline(color('ink', 'lighter'));
```
BPScott, line 153: The `$outline-color` parameter doesn't seem to be used anymore. https://github.com/Shopify/polaris-react/blob/main/src/styles/shared/_buttons.scss#L118

Perhaps we can just remove passing in the argument entirely: `@include button-outline()`

src/components/Button/Button.scss
```diff
@@ -150,15 +150,15 @@ $stacking-order: (
...
 
   &.disabled {
-    @include button-outline-disabled(color('ink', 'lighter'));
+    @include button-outline-disabled(var(--p-border-disabled));
   }
 }
 
 .destructive.outline {
-  @include button-outline(color('red'));
+  @include button-outline(var(--p-border-critical));
```
BPScott, line 162: As above

src/components/Button/Button.scss
BPScott: Alternatively this seems to be the only place this mixin is used - perhaps we just inline it here as all it does is `background: transparent; box-shadow: none;`

BPScott (commented):
Seems we never use those arguments on button-outline, and button-outline-disabled. Perhaps we should just stop passing them in entirely?

We can't remove the parameter from the function on main as that would be a breaking change, but we could stop passing any value in when we call it

UNRELEASED.md
```diff
@@ -37,6 +37,8 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 - Remove filter functions ([#4650](https://github.com/Shopify/polaris-react/pull/4650))
 - Remove all color() invocations ([#4636](https://github.com/Shopify/polaris-react/pull/4636))
 - Cleaned up Button styling and $button-filled mixin([#4635](https://github.com/Shopify/polaris-react/pull/4635))
+- Removed miscellaneous css custom properties ([#4620](https://github.com/Shopify/polaris-react/pull/4620))
```
alex-page, line 40:
```suggestion
```

UNRELEASED.md
```diff
@@ -37,6 +37,8 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
...
+- Removed remaining color sass function uses ([#4655](https://github.com/Shopify/polaris-react/pull/4655))
```
alex-page:
```suggestion
```

Lets remove these as it hasn't released and there is already a dot point for the previous PR. Lets also add a `skip changelog` label so the test passes.

alex-page (approved):

PR #4667, opened 2021-11-17
Move logo from ThemeProvider to Frame

kyledurand:

Part of https://github.com/Shopify/polaris-react/issues/4594

Moving the logo from theme provider to frame context. Part of removing the `ThemeProvider`

## 

### How to 🎩

http://localhost:6006/?path=/story/all-components-top-bar--top-bar-themed-with-color-scheme

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist


src/components/AppProvider/AppProvider.tsx
```diff
@@ -35,7 +35,7 @@ export interface AppProviderProps {
   i18n: ConstructorParameters<typeof I18n>[0];
   /** A custom component to use for all links used by Polaris components */
   linkComponent?: LinkLikeComponent;
-  /** Custom logos and colors provided to select components */
+  /** Custom colors provided to select components */
```
alex-page: Could we just say `TopBar` here?

alex-page (approved):
This is great @kyledurand. Can we document in this PR why we are not shipping the logo with Polaris. I understand that in the admin there are a couple of variations of the logo. Any context to explain that would be great.

kyledurand:
> Can we document in this PR why we are not shipping the logo with Polaris

There are many versions of the Shopify logo in the admin (Default, Plus, Gold, monochrome for each as well as condensed versions like just the bag with a plus for plus etc...). We don't have all of the context on which ones get used and where. We also have contextControl being used on the store switcher which adds to the complexity. We can look into managing these in the future but for now it seems like Polaris shouldn't manage all of these versions and keep that localized to the admin.

BPScott:
We've already got a `FrameContext` - perhaps we could add the logo as a key on that object, rather than having Frame create both a `FrameContext` and a `LogoContext`? I'm don't see much value in creating a new context, am I missing something?

This would also make the change to the Frame and TestProvider a bit simpler as you wouldn't need the new context - you'd be adding a new key to an existing object.

---

In the past we've tended not to add new behaviour to major releases, instead we've added behaviour in minor releases (so there's 2 ways of doing something), and then taken away the deprecated approach in a major release. That makes migration to major versions easier as consumers can make a change to move away from deprecated behaviour before the major version bump PR that removes said behaviour instead of having to do everything in one go.

I wonder if it is worth doing that here too. I think we could split this into two - adding the logo prop to Frame on the main branch, and then removing the deprecated usage of `useTheme()` targeting the v8 branch. That way we can get web using the new style of setting the logo asap, the eventual "update web to use v8" PR gets a bit less noisy, and the v8 diff becomes a bit easier to read.

Something like this:

In a PR targeting main:
- Add the logo prop to the Frame as you've done here
-  In `ContextualSaveBar` / `Navigation` / `TopBar`  have `const logo = useLogo() || useTheme().logo` (possibly replacing `useLogo()` with `useFrame().logo` if you act on my first point). So that they shall check the context from Frame first, and then fall back to what's in the Theme.
-  Update the various docs to demonstrate the usage of `<Frame logo={BLAH}>` and mark the `logo` key in the theme as deprecated.

In a PR targeting v8:
- Remove the fallback for checking  `useTheme().logo` in `ContextualSaveBar` / 
`Navigation` / `TopBar` 

@aaronccasanova 's PR to remove the `ThemeProvider` can then remove  the `src/utilities/theme` folder.


---

As we're moving this data to a new place, perhaps this is an opportunity to adjust some of the [naming](https://github.com/Shopify/polaris-react/pull/4667/files#diff-9d8de3efd2944834375997a1a35bdbf006dec0c68dc650b90742a959ba2fcc26R3). `topBarSource` and `contextualSaveBarSource` sound pretty obtuse, perhaps we could rename them to be `logoOnLightBg` / `logoOnDarkBg` or something? Maybe `accessibilityLabel` could be shortened to `label`? I don't feel strongly about this change.

kyledurand:
I think switching to frame context makes sense. The original idea was to be explicit while building but after getting into fixing up tests I think it's better to merge logo with the existing context. 

I don't love the interim bloat and added complexity for supporting both ways of adding a logo, I'd prefer to do it in one fell swoop but I can be convinced otherwise if we really want to do it that way.

BPScott:
> I don't love the interim bloat

Yeah it's not ideal, and a lot of that advice is born from changes more disruptive than this, but to me that bit of temp clunkyness feels a lot better than all the merge headaches of long-running branches I've just been burnt with before :)

Obviously I'm not on the team so I can't make you do anything, but i will reserve the right to mutter i told you so if you end up in merge hell later when trying to reconcile changes people have made to master with all the breaking changes in v8 :p

kyledurand:
Refactored to use FrameContext. Test setup is much simpler this way 👍

src/components/ContextualSaveBar/README.md
```diff
@@ -85,13 +85,6 @@ Use the save action to provide an opportunity to save changes. Use the discard a
 ```jsx
   <AppProvider
-    theme={{
```
BPScott, line 88: Now this has been removed the AppProvider is the same as the one that all readmes examples get by default. I think this means we can get rid of `omitAppProvider: true` and get rid of all the <AppProvider> wrappers in the examples

src/components/Frame/README.md
```diff
@@ -308,22 +308,19 @@ function FrameExample() {
     </Modal>
   );
 
-  const theme = {
```
BPScott, line 311: Now this has been removed the AppProvider is the same as the one that all readmes examples get by default. I think this means we can get rid of `omitAppProvider: true` and get rid of all the <AppProvider> wrappers in the examples

src/components/TopBar/README.md
```diff
@@ -256,7 +254,6 @@ function TopBarExample() {
   return (
     <div style={{height: '250px'}}>
       <AppProvider
-        theme={theme}
```
BPScott, line 259: Now this has been removed the AppProvider is the same as the one that all readmes examples get by default. I think this means we can get rid of `omitAppProvider: true` and get rid of all the <AppProvider> wrappers in the examples

src/components/Frame/README.md
kyledurand: We'll need this for the frameOffset until Aaron's PR is merged. I think I can remove the others 👍

.storybook/main.js
```diff
@@ -16,7 +16,6 @@ module.exports = {
     {name: '@storybook/addon-essentials', options: {docs: enableDocs}},
     '@storybook/addon-a11y',
     '@storybook/addon-contexts',
-    '@storybook/addon-knobs',
```
alex-page: This should be removed in main and v8

alex-page (approved):

src/components/Frame/Frame.tsx
```diff
@@ -28,6 +29,8 @@ import {
 import styles from './Frame.scss';
 
 export interface FrameProps {
+  /** Sets the logo for the TopBar, Navigation, and ContextualSaveBar components*/
```
aaronccasanova:
```suggestion
  /** Sets the logo for the TopBar, Navigation, and ContextualSaveBar components */
```

src/components/Frame/README.md
```diff
@@ -646,17 +644,16 @@ function FrameExample() {
...
-    logo: {
-      width: 124,
-      topBarSource:
-        'https://cdn.shopify.com/s/files/1/0446/6937/files/jaded-pixel-logo-color.svg?6215648040070010999',
-      contextualSaveBarSource:
-        'https://cdn.shopify.com/s/files/1/0446/6937/files/jaded-pixel-logo-gray.svg?6215648040070010999',
-      url: 'http://jadedpixel.com',
-      accessibilityLabel: 'Jaded Pixel',
-    },
+  const theme = {frameOffset: '60px'};
```
aaronccasanova, line 647: **question (non-blocking):** Are we keeping this `theme` const around as a "TODO" for the follow up issue to add back frame offset support?

src/components/TopBar/tests/TopBar.test.tsx
```diff
@@ -254,6 +244,13 @@ describe('<TopBar />', () => {
       expect(topBar).toContainReactComponent(TopBar.Menu);
     });
 
+    it('doesn’t render a logo when defined', () => {
+      const topBar = mountWithApp(
+        <TopBar contextControl={mockContextControl} />,
+      );
+      expect(topBar).not.toContainReactComponent(Image);
+    });
+
```
aaronccasanova: I believe this is the test we wanted to colocate with the `describe('logo` block above.

aaronccasanova (approved):
Nicely done!

src/components/TopBar/tests/TopBar.test.tsx
kyledurand: contextControl is slightly different from logo so I just re-worded this test and kept it where it is

Issue #4598, opened 2021-11-09
Create simple font-weight tokens


alex-page:

lgriffee:
### Audit Findings 
I conducted a [quick audit of all `font-weight` instances](https://docs.google.com/spreadsheets/d/1QRIOeyZGNabZggZrrcLETRmjk9QIgC0utrcqxaXHo60/edit?usp=sharing) in polaris-react. 

Below is a brief summary of those findings:
| Value                       |  Instances        |
| ----------------------------- | ------------------- |
| 400                           | 9 results / 4 files |
| 500                           | 8 results / 7 files |
| 600                           | 9 results / 4 files |
| 700                           | 4 results / 2 files |
| ~~inherit~~                       | ~~1 result / 1 file~~   |
| ~~\--p-button-font-weight (500)~~ | ~~4 results / 3 files~~ |
| ~~\--p-badge-font-weight (400)~~  | ~~3 results / 3 files~~ |

*~~Strikethrough~~ values removed as part of cleanup PR (see below)

### Proposed Token Values 
Based on the values currently being used in the repo here are some possible token names:
| Value | Token Name (Option 1)        | Token Name (Option 2)           |
| ----- | -------------------- | ------------------------- |
| 400   | \--p-font-weight-400 | \--p-font-weight-regular  |
| 500   | \--p-font-weight-500 | \--p-font-weight-medium   |
| 600   | \--p-font-weight-600 | \--p-font-weight-semibold |
| 700   | \--p-font-weight-700 | \--p-font-weight-bold    

### Questions
- What naming convention should we move forward with?
- Are all these values meaningful? Should we condense/remove any?
- Should we include all possible font-weight tokens?

@sarahill I know you had both of these naming options in your [Density Improvements Figma file](https://www.figma.com/file/TQX4JBfDgAKUIlNvmWCdQv/Density-improvements?node-id=98%3A29727). Was there ever a decision made on which direction we want to go in with naming?

alex-page:
This is excellent! Thanks @lgriffee!

Some thoughts for now. Can we remove `--p-button-font-weight` and `--p-badge-font-weight` today and just hard code them for now. Having two specific custom properties for font-weight values here is not super valuable. This would mean we can just find and replace font-weight: 400 when we have the tokens ready.

I am also curious if we can remove inherit… sometimes this isn’t needed any more..

laurkim:
[...]

PR #4668, opened 2021-11-18
Replace font weight values with tokens

### LINKED ISSUE #4599: Replace font-weight: ***; values with tokens

lgriffee:

- Resolves #4599 
- Related #4598

To guarantee every component in polaris-react is built with simple tokens from one place.

Replacing all `font-weight: ***;` values with the following tokens: 
| Value | Token          |
| ----- | ------------------------- |
| 400   | \--p-font-weight-regular  |
| 500   | \--p-font-weight-medium   |
| 600   | \--p-font-weight-semibold |
| 700   | \--p-font-weight-bold  |

## 

### How to 🎩


aveline (approved):
🎉 Looks great @lgriffee .

I'm just not sure if we should be merging into `main` or another branch where we intend to have the token work. @alex-page ?

src/components/AppProvider/AppProvider.scss
```diff
@@ -78,7 +78,7 @@ h6,
 p {
   margin: 0;
   font-size: 1em;
-  font-weight: 400;
+  font-weight: var(--p-font-weight-regular);
```
alex-page, line 81: The only thing I am unsure about is if we want to continue breaking apart variables `fontWeightRegular` as `font-weight-regular`. It would be good to quickly get some opinions before shipping. I personally would probably prefer less dashes but they are easy to find with the matching `font-weight` value.

alex-page (approved):

src/components/AppProvider/AppProvider.scss
lgriffee: [PR for fixing kebabcase tokens](https://github.com/Shopify/polaris-react/pull/4683#pullrequestreview-812547249) based on team discussion

documentation/Color system.md
```diff
@@ -277,7 +277,7 @@ Used to decorate elements where color does convey a specific meaning in componen
 | `--p-override-one`                        | `1`                                                                                                          |
 | `--p-override-visible`                    | `visible`                                                                                                    |
 | `--p-override-zero`                       | `0`                                                                                                          |
-| `--p-override-loading-z-index`            | `514`                                                                                                        |
+| `--p-override-loading-zindex`             | `514`                                                                                                        |
```
alex-page: This should match css `z-index`

lgriffee: Yeah I was just looking into this...I'll try to figure out what from the merge is causing it

lgriffee: @alex-page  I think it's from the [new logic to group capital letters together](https://github.com/Shopify/polaris-react/pull/4683/commits/6aca9e0b55c8902ca24d5705cfab854342cd47fa).

aaronccasanova: [z-index token reference](https://github.com/Shopify/polaris-react/blob/6117c999272f562141de6556c9a3cf2f3343a01a/src/utilities/theme/tokens.ts#L32)

You're right @lgriffee The new regex no longer inserts a hyphen before each capital letter. Wonder if that update was intentional.

aaronccasanova (approved):

alex-page (approved):

documentation/Color system.md
alex-page: My bad. I added it for consistency then realised it caused a regression. Should be fine now!

PR #4672, opened 2021-11-18
Update [top bar] description to include "header"

jjgali:

@kwringe offered this feedback about the top bar:
>I think that Top Bar should be part of Navigation. I would also advocate that we somehow sneak the word “header” into it’s description so that when you search for the term “header”, that Top Bar appears (I didn’t realize that we used the term Top Bar).

Updated top bar component description to include the term "header" and clarify intended usage for this component. Also added `header` as a keyword for this component.

### DIFF
--- a/src/components/TopBar/README.md
+++ b/src/components/TopBar/README.md
@@ -1,11 +1,12 @@
 ---
 name: Top bar
-category: Structure
+category: Navigation
 keywords:
   - global chrome
   - global features
   - topbar
   - top bar
+  - header
   - nav bar
   - bar
   - navbar
--- a/src/components/TopBar/README.md
+++ b/src/components/TopBar/README.md
@@ -19,7 +19,7 @@ omitAppProvider: true
 
 # Top bar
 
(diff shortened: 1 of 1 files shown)


chloerice (approved):
Thanks JJ! Should we also update the `category` from `Structure` to `Navigation` and add `header` to the `keywords` in the metadata up top?

zecarlostorre:
Thanks @jjgali 🔥 

I don't really know what this one is about, but I see it as required:

Screen Shot 2021-11-18 at 4 05 45 PM

Can we squash and merge anyway?

jjgali:
Thanks y'all! @chloerice I think I've changed the category and added `header` as a keyword with that second commit. I also thought I added updates to the changelog, so not sure why we're getting that error?

lgriffee:
Hey JJ! If you need to skip the changelog in the future you can always add a "Skip Changelog" label to your pull request here:
Screen Shot 2021-11-19 at 10 51 32 AM

PR #4674, opened 2021-11-18
🚧 Replace font-size() and line-height() with custom properties

### LINKED ISSUE #4600: Consolidate line-height() and font-size() function values in tokens
This issue is blocked by https://github.com/Shopify/polaris-react/issues/4588 but in the meantime we can land on the approach we want to take.

aveline:

Resolves https://github.com/Shopify/polaris-react/issues/4600

- Replacing font-size() and line-height() functions with custom properties
- Currently blocked by https://github.com/Shopify/polaris-react/issues/4588 which will allow the custom properties to be properly generated

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/AppProvider/AppProvider.scss
+++ b/src/components/AppProvider/AppProvider.scss
@@ -32,7 +32,7 @@ button {
 
 html {
   position: relative;
-  font-size: ($base-font-size / $default-browser-font-size) * 100%;
+  font-size: 100%;
   -webkit-font-smoothing: antialiased;
 
   // This needs to come after -webkit-font-smoothing
--- a/src/components/Navigation/Navigation.scss
+++ b/src/components/Navigation/Navigation.scss
@@ -348,11 +348,11 @@ $secondary-item-font-size: rem(15px);
   padding-left: spacing();
 
   .Text {
-    font-size: font-size(subheading);
+    font-size: var(--fs-1);
     color: var(--p-text-subdued);
 
     @include when-typography-not-condensed {
-      font-size: font-size(subheading, large-screen);
+      font-size: var(--fs-0);
     }
   }
 
--- a/src/styles/foundation/_utilities.scss
+++ b/src/styles/foundation/_utilities.scss
@@ -13,7 +13,7 @@ $base-font-size: 10px;
   } @else if $unit == 'rem' {
     @return $value;
   } @else if $unit == 'px' {
-    @return $value / $base-font-size * 1rem;
+    @return $value / $default-browser-font-size * 1rem;
   } @else if $unit == 'em' {
     @return $unit / 1em * 1rem;
   } @else {
--- a/src/styles/shared/_typography.scss
+++ b/src/styles/shared/_typography.scss
@@ -13,34 +13,34 @@ $typography-condensed: em(640px);
 }
 
 @mixin text-style-caption {
-  font-size: font-size(caption);
+  font-size: var(--fs-1);
   font-weight: 400;
   line-height: line-height(caption);
 
   @include when-typography-not-condensed {
-    font-size: font-size(caption, large-screen);
+    font-size: var(--fs-0);
     line-height: line-height(caption, large-screen);
   }
 }
 
 @mixin text-style-heading {
-  font-size: font-size(heading);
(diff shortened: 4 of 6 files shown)


src/components/AppProvider/AppProvider.scss
```diff
@@ -32,7 +32,7 @@ button {
 
 html {
   position: relative;
-  font-size: ($base-font-size / $default-browser-font-size) * 100%;
+  font-size: 100%;
```
aveline, line 35: Is this the right time to introduce this change? It does impact the rem value we would set the `font-size` and `line-height` tokens to.

For context, there has been a previous attempt to make this change, but it was reverted https://github.com/Shopify/polaris-react/pull/1590

src/styles/foundation/_utilities.scss
```diff
@@ -13,7 +13,7 @@ $base-font-size: 10px;
   } @else if $unit == 'rem' {
     @return $value;
   } @else if $unit == 'px' {
-    @return $value / $base-font-size * 1rem;
+    @return $value / $default-browser-font-size * 1rem;
```
aveline, line 16: This change is related to changing the base font size here https://github.com/Shopify/polaris-react/pull/4674/files#r752497881

src/styles/shared/_typography.scss
```diff
@@ -13,34 +13,34 @@ $typography-condensed: em(640px);
 }
 
 @mixin text-style-caption {
```
aveline, line 15: Leaving these `mixins` in for the scope of this PR. It would make sense to address them when we start thinking about composite components.

src/components/AppProvider/AppProvider.scss
kyledurand: Might be worth getting @martenbjork to weigh in on this. I know he's experimented with global font size changes in the past

alex-page: My gut feeling is we tackle this when we remove the `rem` function. Would love to know other thoughts.

aveline: @alex-page ok probably best not to conflate things. it's simple enough to update the token values when we do make the change.

src/components/Navigation/Navigation.scss
```diff
@@ -348,11 +348,11 @@ $secondary-item-font-size: rem(15px);
   padding-left: spacing();
 
   .Text {
-    font-size: font-size(subheading);
+    font-size: var(--fs-1);
```
aaronccasanova, line 351: Glad to see we settled on token scales for `font-size` and `line-height`!

**suggestion(non-blocking):** Wonder if we should expand these abbreviations to match the `--p-font-weight` tokens @lgriffee [just introduced](https://github.com/Shopify/polaris-react/blob/7d97674ab6434618a8d1d649022b5c1baefcdf40/documentation/Color%20system.md?plain=1#L264-L267)? I know this is still WIP and the scope of this work is to get 100% token coverage, but I feel like we should have consistency in our typography tokens.

```css
/* This doesn't feel very intuitive */
p {
  font-size: var(--p-fs-1);
  font-weight: var(--p-font-weight-regular)
}
```

aveline: I don't feel too strongly either way. Now is the time to play around with this and it will be easy enough to change through the token generator. Would it be worth creating a separate issue to align on naming convention?

aaronccasanova: I don't think a separate issue is necessary personally. If anything I would just have a quick sync with @lgriffee to settle on a convention for v8. Right now it is clear two developers worked on the typography vars. I'm leaning towards expanding since `fs` isn't clear to me. If I was looking at a collection of `:root` properties does `fs` mean: `flex-shrink`, `font-style`, `font-stretch`, `font-synthesis`, `font-size`, `font-size-adjustment`, etc...?

aveline:
Split up into https://github.com/Shopify/polaris-react/pull/4852 and https://github.com/Shopify/polaris-react/pull/4784

PR #4676, opened 2021-11-18
Remove filter sass function

### LINKED ISSUE #4657: Remove filter() sass function
We need to deprecate the filter() function from our public sass API. It is no longer used in `polaris-react` and we no longer want to support this functionality in the system. Let's remove this function altogether.

Breaking change, stage for v8.0.0-major

lgriffee:

Closes #4657 

We need to deprecate the filter() function from our public sass API. It is no longer used in polaris-react and we no longer want to support this functionality in the system. 

🚨 **Breaking changes** 🚨

- Removing `_filters.scss`
- Removing imports in `_common.scss` and `_public-api.scss`
- Removing the `color-icon()` mixin


src/styles/shared/_icons.scss
```diff
@@ -16,13 +16,3 @@
...
-
-@mixin color-icon($value, $hue: base) {
-  svg {
-    fill: color($value, $hue);
-  }
-
-  img {
-    filter: filter($value, $hue);
-  }
-}
```
lgriffee, line 28: `color-icon()` isn't used anywhere else in polaris-react so I think we can remove this (since we'll also be removing `color()` as part of v8.0.0? Also a breaking change for web.

alex-page (approved):

alex-page (approved):

PR #4683, opened 2021-11-19
Kebabcase token fix

alex-page:

We need the right number of kebabs `-🍢- 🥙 -🍡-`. We have a few numerical custom properties being created. We don't want to divide every number with a `-`. This is a breaking change for v8.

- Updates the custom property creation logic so that numbers are not split up with `-`.
- Replace the `duration(fast)` with `--p-duration-100`
- Replace `--p-duration-1-5-0` with `--p-duration-150`

### DIFF
--- a/documentation/guides/migrating-from-v7-to-v8.md
+++ b/documentation/guides/migrating-from-v7-to-v8.md
@@ -1,0 +1,3 @@
+# Migrating from v7 to v8
+
+Polaris v8.0.0 ([full release notes](https://github.com/Shopify/polaris-react/releases/tag/v8.0.0)) features a rebuild of the `Autocomplete` component, build changes, . This file describes all code updates required to stay up to date.
--- a/src/utilities/theme/utils.ts
+++ b/src/utilities/theme/utils.ts
@@ -57,7 +57,7 @@ function customPropertyTransformer(
 }
 
 export function toCssCustomPropertySyntax(camelCase: string) {
-  return `--p-${camelCase.replace(/([A-Z]|[0-9]+)/g, '-$1').toLowerCase()}`;
+  return `--p-${camelCase.replace(/([A-Z]+|[0-9]+)/g, '-$1').toLowerCase()}`;
 }
 
 // eslint-disable-next-line @typescript-eslint/ban-types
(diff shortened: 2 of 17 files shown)


src/utilities/theme/utils.ts
```diff
@@ -57,7 +57,7 @@ function customPropertyTransformer(
 }
 
 export function toCssCustomPropertySyntax(camelCase: string) {
-  return `--p-${camelCase.replace(/([A-Z0-9])/g, '-$1').toLowerCase()}`;
+  return `--p-${camelCase.replace(/([A-Z]|[0-9]+)/g, '-$1').toLowerCase()}`;
```
alex-page: This works but could be better. Was unsure the perfect way to select the first numerical value. Ended up using the `+`.

documentation/guides/migrating-from-v7-to-v8.md
```diff
@@ -0,0 +1,3 @@
+# Migrating from v7 to v8
+
+Polaris v8.0.0 ([full release notes](https://github.com/Shopify/polaris-react/releases/tag/v8.0.0)) features a rebuild of the `Autocomplete` component, build changes, . This file describes all code updates required to stay up to date.
```
alex-page, line 3: Going to start creating a draft migration guide. Too lazy on Friday arvo to move it to another PR.

src/utilities/theme/utils.ts
kyledurand: I don't know of a better way to do this without digging into regex for a good hour or so. From what I've tested this works great

kyledurand (approved):

aveline (approved):

src/utilities/theme/utils.ts
aaronccasanova: Looks great 🔥

aaronccasanova (approved):

src/utilities/theme/utils.ts
aaronccasanova: > Was unsure the perfect way to select the first numerical value.

This is possible with a negative lookbehind assertion. Although, I've never looked up benchmarks on these types of matches I imagine what you have right now (`[0-9]+`) is more performant.

image

> Note: The last two examples match any uppercase letter OR digit NOT preceded by another digit.

lgriffee (approved):

PR #4687, opened 2021-11-22
Remove custom and unnecessary font weight properties

lgriffee:
[These changes were reverted](https://github.com/Shopify/polaris-react/commit/678433d161a0bd29bfe7939bab001e8b599a4c06) on main to avoid breaking changes on a minor release. They need to be added back to this major branch so we don't lose them. (Some changes from the linked revert have already been added back to the v8.0.0 branch)

 Removing the following tokens and font-weight values: 
- `--p-button-font-weight`
- `--p-badge-font-weight`
- `inherit`


src/components/TextField/TextField.scss
```diff
@@ -100,9 +100,7 @@ $stacking-order: (
   padding: control-vertical-padding() $backdrop-horizontal-spacing;
   background: none;
   border: border(transparent);
-  font-family: inherit;
-  font-size: inherit;
-  font-weight: inherit;
+  font-family: font-family();
```
alex-page, line 103: I am sad we need this... makes sense knowing we don't put font family on parent elements.

alex-page (approved):

PR #4689, opened 2021-11-22
Remove dev.yml

alex-page:

`dev` will no longer be used at Shopify with M1 laptops. This PR removes the associated files. This also makes our local environment similar to external contributors which is a good thing.

The downsides is that we will need to do some work to make sure folks know what version of node the library is used with.

Removing the `dev.yml` file.


alex-page:
Closing this as nodejs 16 will be supported with dev https://github.com/Shopify/dev/pull/6474

PR #4691, opened 2021-11-23
Spacing tokens

aveline:

Resolves https://github.com/Shopify/polaris-react/issues/4603 and https://github.com/Shopify/polaris-react/issues/4604

- Introduce spacing tokens as css custom properties
- Replace usage of the `spacing()` function
- Remove `spacing()` function


src/components/Badge/Badge.scss
```diff
@@ -1,17 +1,16 @@
...
-$vertical-padding: calc(#{spacing(extra-tight)} - #{rem(2px)});
-
-$pip-size: spacing(tight);
-$pip-spacing: calc(#{$pip-size} / 2);
 
 .Badge {
+  --p-horizontal-padding: var(--p-space-2);
+  --p-vertical-padding: calc(var(--p-space-1) - #{rem(2px)});
+  --p-pip-size: var(--p-space-2);
+  --p-pip-spacing: calc(var(--p-pip-size) / 2);
```
alex-page, line 7: The impact for these custom properties seems quite low. I think it would be safe to use the tokens directly. Can we also change `calc(var(--p-pip-size) / 2)` to `--p-space-1`.

src/components/Banner/Banner.scss
```diff
@@ -57,6 +55,9 @@ $spinner-size: rem(20px);
 
 .Banner {
   --p-rgb-text: 33, 43, 54;
+  --p-secondary-action-horizontal-padding: calc(1.5 * var(--p-space-2));
```
alex-page, line 58: `--p-space-3` remove calc

alex-page (approved):
Awesome work @aveline. This PR is in a great place. Happy to ship it. I think there are a few tasks we can break down into smaller PR's now this is in a good place.

- Removing calc when there is a spacing variable that matches
- Removing rem values that matching spacing variables
- Aligning on approach for clear distinction between tokens and component custom properties

Anything else you want to address after this work?

scripts/build-validate.js
```diff
@@ -72,7 +72,6 @@ function validateEsNextBuild() {
 
 function validateSassPublicApi() {
   assert.ok(fs.existsSync('./build/styles/_public-api.scss'));
-  assert.ok(fs.existsSync('./build/styles/foundation/_spacing.scss'));
```
BPScott, line 75: It would be useful to keep a test for "some file of the public API has been copied across". Perhaps change this to some other filename `foundation/_layout.scss` perhaps?

aveline:
Thanks @alex-page. I think that's a good list for follow up tasks.
Here are the follow up issues
* #4726
* #4732
* #4748

Issue #4656, opened 2021-11-16
Remove color() sass function


yurm04:
We need to deprecate the `color()` function from our public sass API. It is already no longer used in `polaris-react` (once #4655 is merged). Let's remove this function altogether. 

Breaking change, stage for `v8.0.0-major`

ghost:
👋 Thanks for opening your first issue. A contributor should give feedback soon. If you haven’t already, please check out the [contributing guidelines](https://github.com/Shopify/polaris-react/blob/master/.github/CONTRIBUTING.md).

PR #4696, opened 2021-11-23
Delete `color()` function from sass API

yurm04:
Resolves issue #4656

Removes `color()` sass function. Function has been deprecated and should be removed.

⚠️ This is a breaking change since this function is part of our `_public-api.scss` ⚠️ 

### How to 🎩


yurm04:
~~my commits are all messed up from trying to get my branch up to date. will fix!~~ fixed ✨

aveline (approved):

alex-page (approved):

BPScott:
Can we also remove: the `color-multiply` function,  the import from `../polaris-tokens/colors.color-map` and the `$color-palette-data` variable, as they are all no longer used anywhere.


The removal of color-mulitply is handy, as it currently contributes to raising a load of deprecation warnings when you try to use the sass api in dart-sass (removing this means the only other deprecations are in the rem/em functions which I believe are planing to be removed elsewhere)

The removal of the import gets us a step closer to removing the legacy token variables

PR #4697, opened 2021-11-23
Remove skeleton animation

alex-page:

Breaking [this pr down](https://github.com/Shopify/polaris-react/pull/4462). Removing the animation in this PR and moving the removal of the animation mixin to v8. Removing the skeleton-shimmer mixin is a breaking change. These animations can safely be removed.

Removing animation from all skeleton components

### DIFF
--- a/src/components/SkeletonBodyText/SkeletonBodyText.tsx
+++ b/src/components/SkeletonBodyText/SkeletonBodyText.tsx
@@ -17,7 +17,5 @@ export function SkeletonBodyText({lines = 3}: SkeletonBodyTextProps) {
     bodyTextLines.push(<div className={styles.SkeletonBodyText} key={i} />);
   }
 
-  return (
-    <div className={styles.SkeletonBodyTextContainer}>{bodyTextLines}</div>
-  );
+  return bodyTextLines;
 }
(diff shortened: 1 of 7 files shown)


yurm04 (approved):
✨

src/components/SkeletonBodyText/SkeletonBodyText.tsx
```diff
@@ -17,7 +17,5 @@ export function SkeletonBodyText({lines = 3}: SkeletonBodyTextProps) {
     bodyTextLines.push(<div className={styles.SkeletonBodyText} key={i} />);
   }
 
-  return (
-    <div className={styles.SkeletonBodyTextContainer}>{bodyTextLines}</div>
-  );
+  return <>{bodyTextLines}</>;
```
kyledurand, line 20: This doesn't need to be in a fragment it looks like
```suggestion
  return bodyTextLines;
```

kyledurand (approved):

src/components/SkeletonBodyText/SkeletonBodyText.tsx
alex-page: Let me double check. I had a type error without it 👀

PR #4701, opened 2021-11-23
Add postcss-pxtorem dependency

kyledurand:

First part of https://github.com/Shopify/polaris-react/issues/4605

[Diff check between previous css build (spacing changes are expected)](https://www.diffchecker.com/SUs9t8LC)

This will introduce pxtorem to main. Then we can remove `rem()` from our public api in v8

Adds [pxtorem](https://www.npmjs.com/package/postcss-pxtorem) to our postcss config

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist


.storybook/main.js
```diff
@@ -58,7 +59,16 @@ module.exports = {
...
               postcssOptions: {
-                plugins: [postcssShopify()],
+                plugins: [
+                  postcssShopify(),
+                  pxtorem({
+                    rootValue: 10,
+                    replace: true,
+                    propList: ['*'],
+                    selectorBlackList: [],
+                    browsers: ['extends @shopify/browserslist-config'],
```
kyledurand, line 69: I have no idea if this works or if it's needed

.storybook/main.js
```diff
@@ -58,7 +59,16 @@ module.exports = {
             loader: 'postcss-loader',
             options: {
               postcssOptions: {
-                plugins: [postcssShopify()],
+                plugins: [
+                  postcssShopify(),
+                  pxtorem({
+                    rootValue: 10,
+                    replace: true,
+                    propList: ['*'],
```
kyledurand, line 67: Targeting all selectors for testing. We might have to add to the `selectorBlackList` below if we notice anything weird

package.json
```diff
@@ -137,8 +137,10 @@
     "node-sass": "^4.12.0",
     "npm-run-all": "^4.1.5",
     "object-hash": "^1.3.1",
-    "postcss": "^7.0.18",
+    "postcss": "^8.0.0",
```
kyledurand: @BPScott do you know of any issues we might see by bumping postcss?

BPScott: You'll need to also update postcss-modules and @shopify/postcss-config too.


https://github.com/Shopify/loom/pull/307 is me updating postcss 7 to 8 along with those other two over in a loom plugin.

kyledurand:
Here's a diff of the changes to our build file: https://www.diffchecker.com/SUs9t8LC

The spaces between `{}` are expected it looks like going from the link @BPScott posted for updating [loom](https://github.com/Shopify/loom/pull/307/files)

config/px-to-rem.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  rootValue: 10,
+  replace: true,
+  propList: ['*'],
+  selectorBlackList: [],
+};
```
alex-page: Nice!

alex-page (approved):

config/rollup/plugin-styles.js
```diff
@@ -25,6 +28,7 @@ module.exports.styles = function styles({
   const renderSass = promisify(nodeSass.render);
 
   const styleProcessor = postcss([
+    pxtorem(pixelsToRemsConfig),
```
BPScott, line 31: This should live in the same place that we configure `@shopify/postcss-plugin`, rather than as part of the rollup plugin - in the `plugins` argument that is passed into this plugin - https://github.com/Shopify/polaris-react/blob/main/loom.config.ts#L143 (and shortly below.

If we do that then you can have a config file that exports an array of postcss plugins that are used, and then you can use that array in both the loom config where postcss plugins are defined, and in storybook config

```
const postcssShopify= require('@shopify/postcss-plugin');
const pxtorem = require('postcss-pxtorem');

module.exports = [
  postcssShopify,
  pxtorem({
    rootValue: 10,
    replace: true,
    propList: ['*'],
    selectorBlackList: [],
  })
]
```

BPScott: This should live in the same place that we configure `@shopify/postcss-plugin`, rather than as part of the rollup plugin - in the `plugins` argument that is passed into this plugin - https://github.com/Shopify/polaris-react/blob/main/loom.config.ts#L143 (and shortly below that line too).

If we do that then you can have a config file that exports an array of postcss plugins that are used, and then you can use that array in both the loom config where postcss plugins are defined, and in storybook config

```js
// config/postcss-plugins.js
const postcssShopify= require('@shopify/postcss-plugin');
const pxtorem = require('postcss-pxtorem');

module.exports = [
  postcssShopify,
  pxtorem({
    rootValue: 10,
    replace: true,
    propList: ['*'],
    selectorBlackList: [],
  })
]
```

BPScott (changes requested):

alex-page (approved):
Nice clean up + functionality

aveline:
Will this convert all `px` to `rem`? There are valid cases when we would want to keep `px` values. What does it look like if we want to specify certain values remain `px`?

_eta_

> Currently, the easiest way to have a single property ignored is to use a capital in the pixel unit declaration.

I guess that works, but feels a bit "magic"

alex-page:
@aveline it definitely is magic.

I hope this feels like [autoprefixer](https://autoprefixer.github.io). People understand that the output changes but they do not have to write the repetitive code. We also don't need to support a public function. If we can be more opinionated by whitelisting the properties I would be happy to do that.

BPScott (approved):
Haven't validated the diff output on account of all the whitespace changing but code changes seem reasonable.

It miiiiight be worth splitting out the postcss updates into a separate PR so then it's easier to see "what has changed by adding pxtorem" rather than 100% of the diff being those whitespace differences from the postcss update drowning out any changes as a result of pxtorem. But if you're confident in output then whatever

UNRELEASED.md
```diff
@@ -32,8 +32,17 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 
 - Tightened up what absolute imports are allowed. Removed `baseUrl` from `tsconfig.json`. Attempting to do an absolute import from `src/X` or `components/X` now results in a error when type-checking. ([#4643](https://github.com/Shopify/polaris-react/pull/4643))
 
+### New dependencies
+
+- Added `posctcss-pxtorem` to automatically convert px values to rem ([#4701](https://github.com/Shopify/polaris-react/pull/)4701)
```
BPScott: This is a dev dependency, not a runtime dependency, it's not something consumers really need to care about

kyledurand:
As per Ben's suggestion here's the diff that omits the spacing change https://www.diffchecker.com/GT3piigi

I've looked through it and it's mostly changes to borders, border radii, outlines, visually hidden and a few spacing tokens. I've tophatted things that had potential to look different and all seems good

PR #4716, opened 2021-11-24
Remove src/utilities/theme/* and color system docs

alex-page:
- [x] Remove color system docs md file - this will be replaced with a page on polaris-styleguide website
- [x] Remove src/util/theme - we can start using the designTokens directory


src/components/Collapsible/tests/Collapsible.test.tsx
```diff
@@ -69,7 +69,7 @@ describe('<Collapsible />', () => {
 
   describe('Transition', () => {
     it('passes a duration property', () => {
-      const duration = Tokens.duration150;
```
alex-page: This is temporary. Unsure why it's using 150ms.

kyledurand (approved):

src/components/Collapsible/tests/Collapsible.test.tsx
BPScott: If this is a boring old test then do we even care about the number coming from a token?

Could we just do `const duration = '100';` and not bother with any imports from `designTokens/motion` at all. Same for the `timingFunction` down below.

That'd also mean that this test update can be done against the main branch asap, to reduce the number of changes in the v8 branch

src/components/Collapsible/tests/Collapsible.test.tsx
```diff
@@ -1,7 +1,7 @@
 import React, {useState, useCallback} from 'react';
 import {mountWithApp} from 'tests/utilities';
 
-import {Tokens} from '../../../utilities/theme';
+import {motion} from '../../../designTokens/motion';
```
aaronccasanova:
```suggestion
import {motion} from '../../../designTokens';
```

aaronccasanova (approved):
LGTM!

🔥 image 🔥

PR #4717, opened 2021-11-24
Remove analyze custom properties

alex-page:
This script is awesome but no longer needed in `polaris-react`
- We have https://polaris-coverage.shopifycloud.com which monitors the use of custom properties and hard coded values.
- CSS custom properties either work and apply the styles or do not
- We are not encountering issues with the creation or removal of custom properties
- We have visual regression testing which will catch styles not working as intended
- This feels repetitive however could be turned into something useful in `shopify/web`
- Maintaining 1000 lines of code and running a test on every PR is unnecessary


BPScott:
Don't forget to check for packages that were only installed for use in these files - for instance I think `scss-parser` was only used in here

BPScott:
Also, I think could be targeted at the main branch, rather than v8

PR #4721, opened 2021-11-24
Rename designTokens to tokens

alex-page:

Follow patterns that exist for directory naming. Try to simplify and group together token files.

- [x] Renames directory `designTokens` to `tokens`
- [x] Renames files to use kebab-case
- [x] Moves colors from `legacyTokens.ts` to `colors.ts`
- [x] Moves `color-scheme.ts` into `tokens.ts`
- [x] Renames `Tokens` to `TokenObject`
- [x] Renames `DesignTokens` to `Tokens`
- [x] Renames `designTokens` to `tokens`
- [x] Renames token `color.ts` `typography.ts` `motion.ts` files to start with `_` so they are grouped together


src/tokens/index.ts
```diff
@@ -0,0 +1 @@
+export * from './tokens';
```
alex-page, line 1: I was unsure if this should export the individual token values as they are apart of the `tokens` export. @aaronccasanova happy to revert here but wanted some context.

aveline (approved):

src/tokens/index.ts
aaronccasanova: Yeah, I don't think it's necessary right now. I started that pattern for use cases where we might want to individually import tokens in components. For example: `motion` durations with `react-transition-group`
```tsx
import {motion} from '../tokens`
...
return (
  <CssTransition timeout={motion[...]} />
)
```

This update looks good to me 👍

aaronccasanova: Yeah, I don't think it's necessary right now. I started that pattern for use cases where we might want to individually import tokens in components. For example: `motion` durations with `react-transition-group`
```tsx
import {motion} from '../tokens`
...
return (
  <CssTransition timeout={motion[...]} />
)
```

This update looks good to me 👍

src/components/CustomProperties/styles.ts
```diff
@@ -1,23 +1,22 @@
...
   ColorScheme,
-  designTokens,
-  DesignTokens,
-  osColorSchemes,
+  TokenObject,
   OSColorSchemes,
-  Tokens,
-} from '../../designTokens';
+} from '../../tokens';
+import {tokens, osColorSchemes} from '../../tokens';
```
aaronccasanova, line 7: Why are we separating module and type imports? I thought [type only imports](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export) were beneficial if your only importing types from a module.

src/components/CustomProperties/styles.ts
```diff
@@ -52,7 +51,7 @@ export function getColorSchemeRules(
  * Creates static CSS custom properties.
  * Note: These values don't vary by color-scheme.
  */
-export function getStaticCustomProperties(designTokens: DesignTokens) {
+export function getStaticCustomProperties(designTokens: Tokens) {
```
aaronccasanova, line 54: **suggestion:** Perhaps we should also rename the parameter to `tokens` and the internal `tokens` variable to `tokensObject` (Matching the type annotations)

src/tokens/tokens.ts
```diff
@@ -0,0 +1,59 @@
...
+import {motion} from './_motion';
+import {typography} from './_typography';
+
+/**
+ * Values to convert to CSS custom properties.
+ * @example {background: '#000'} // --p-background: #000;
+ */
+export interface TokenObject {
+  [token: string]: string;
+}
```
aaronccasanova, line 12: **suggestion:** I was actually thinking about renaming this in my original PR, but didn't want to change anything after all the checks were passing. What do you think about naming this `TokenGroup` or `TokensGroup`. That's effectively how I was architecting the tokens structure (See the [Group section of the design tokens format module](https://design-tokens.github.io/community-group/format/#groups-0) for more information).

Notice I use this convention in the [getStaticCustomProperties](https://github.com/Shopify/polaris-react/blob/60cc0a2bf8f00219a200d19d3d418e58a1668145/src/components/CustomProperties/styles.ts#L56) utility

I also subjectively prefer this over tacking on the literal `Object` type 🙂

src/components/CustomProperties/tests/CustomProperties.test.tsx
```diff
@@ -24,7 +24,7 @@ interface ColorSchemeAttribute {
   'color-scheme': ColorScheme;
 }
 
-const mockTokens: Tokens = {
+const mockTokens: TokenObject = {
```
aaronccasanova, line 27: **suggestion:** Can we also update the mocks to match the types.

src/tokens/tokens.ts
```diff
@@ -0,0 +1,59 @@
...
+  [C in ColorScheme]: TokenObject;
+};
+
+const colorSchemes: ColorSchemes = {
+  light: {
+    ...lightColorScheme,
+  },
+  dark: {
+    ...darkColorScheme,
+  },
```
aaronccasanova, line 44: **suggestion:** If the color scheme token groups live in a separate file we can probably remove the spread.
```suggestion
  light: lightColorScheme,
  dark: darkColorScheme,
```

aaronccasanova (approved):
Updates look great! Left a few thoughts and suggestions.

PR #4727, opened 2021-11-24
[Frame] Re-add offset

kyledurand:

Fixes https://github.com/Shopify/polaris-react/issues/4705

I tried two approaches here. The first was rendering `--p-frame-offset` on the Frame node. This works within Polaris but may not work in web because the GlobalNav is rendered outside of the frame component. To be a bit more confident in the approach I'm adding it to the DOM root.

## 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

http://localhost:6006/?path=/story/all-components-frame--frame-with-an-offset


src/components/Frame/Frame.tsx
```diff
@@ -88,12 +91,14 @@ class FrameInner extends PureComponent<CombinedProps, State> {
...
     }
     this.setGlobalRibbonRootProperty();
+    this.setOffset();
   }
 
   componentDidUpdate(prevProps: FrameProps) {
     if (this.props.globalRibbon !== prevProps.globalRibbon) {
       this.setGlobalRibbonHeight();
     }
+    this.setOffset();
```
kyledurand, line 101: Global nav is set async so I'm re-running this on update as well.

src/utilities/set-root-property.ts
```diff
@@ -1,7 +1,7 @@
 export function setRootProperty(
   name: string,
   value: string,
-  node: Element | null,
+  node: Element | null = null,
```
kyledurand, line 4: Defaulting to null here for simplicity. All calls were null anyway

alex-page (approved):

PR #4728, opened 2021-11-24
Remove @shopify/polaris-tokens

alex-page:

Fixes https://github.com/Shopify/polaris-react/pull/4728

- [x] Removes `polaris-tokens` npm package
- [x] Removes copy scripts needed to use the `polaris-tokens` nom package
- [x] Replaces imported token values from the package with imports from the local tokens files


package.json
```diff
@@ -52,25 +52,23 @@
...
-    "copy-polaris-tokens": "rimraf ./src/styles/polaris-tokens && shx cp -r ./node_modules/@shopify/polaris-tokens/dist ./src/styles/polaris-tokens",
     "prepublishOnly": "yarn run build",
-    "dev": "npm-run-all copy-polaris-tokens storybook",
+    "dev": "yarn run storybook",
     "start": "serve ./build-internal/storybook/static -l ${PORT:=6006}",
     "readme-update-version": "node ./scripts/readme-update-version",
     "version": "yarn run readme-update-version",
     "storybook": "start-storybook -p 6006 --quiet",
-    "storybook:build": "yarn run copy-polaris-tokens && build-storybook -o build-internal/storybook/static"
+    "storybook:build": "yarn run build-storybook -o build-internal/storybook/static"
```
BPScott: Stuff in the node_modules/.bin folder is already in the binpath, so you don't need the `yarn run` bit

```suggestion
    "storybook:build": "build-storybook -o build-internal/storybook/static"
```

src/components/Popover/components/PopoverOverlay/PopoverOverlay.tsx
```diff
@@ -105,7 +105,7 @@ export class PopoverOverlay extends PureComponent<PopoverOverlayProps, State> {
         this.clearTransitionTimeout();
         this.enteringTimer = window.setTimeout(() => {
           this.setState({transitionStatus: TransitionStatus.Entered});
-        }, durationFast);
+        }, parseInt(tokens.motion['duration-200'], 10));
```
BPScott: Comparing https://github.com/Shopify/polaris-tokens/blob/main/dist/duration.json and https://github.com/Shopify/polaris-react/blob/v8.0.0-major/src/tokens/_motion.ts:

durationFast is 100ms, so should this be:

```suggestion
        }, parseInt(tokens.motion['duration-100'], 10));
```

src/components/Popover/components/PopoverOverlay/PopoverOverlay.tsx
```diff
@@ -114,7 +114,7 @@ export class PopoverOverlay extends PureComponent<PopoverOverlayProps, State> {
         this.clearTransitionTimeout();
         this.exitingTimer = window.setTimeout(() => {
           this.setState({transitionStatus: TransitionStatus.Exited});
-        }, durationFast);
+        }, parseInt(tokens.motion['duration-200'], 10));
```
BPScott: Comparing https://github.com/Shopify/polaris-tokens/blob/main/dist/duration.json and https://github.com/Shopify/polaris-react/blob/v8.0.0-major/src/tokens/_motion.ts:

durationFast is 100ms, so should this be:

```suggestion
        }, parseInt(tokens.motion['duration-100'], 10));
```

BPScott (commented):
Ya love to see it. Very nice to see this be decoupled.

Spotted a case where old/new token values didn't align, noted inline.

This is worthy of an UNRELEASED entry / migration guide entry as all the token values are no longer present in the [styles folder](https://unpkg.com/browse/@shopify/polaris@7.4.1/build/styles/polaris-tokens/). Admittedly consumers shouldn't be touching those files directly, and should be going through _public-api, so they only ever had supported access to the variables that were imported from within our files (e.g. this https://github.com/Shopify/polaris-react/pull/4691/files#diff-562c96f2c44a6e99a30a05886e3d3861f667753cee0f4e36c401b089ef4f49e2L2)

PR #4733, opened 2021-11-25
[Page] Deprecate thumbnail property

alex-page:

This functionality is rarely used in our product and should be replaced with layout primitives in the future.

Deprecating the thumbnail property


UNRELEASED.md
```diff
@@ -15,6 +15,7 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 - Updated styling of `DropZone` border and overlay text. ([#4662](https://github.com/Shopify/polaris-react/pull/4662))
 - Remove duplicate duration(fast) usage. ([#4682](https://github.com/Shopify/polaris-react/pull/4682))
 - Updated the accessability label for the rollup actions in the `Page` header ([#4080](https://github.com/Shopify/polaris-react/pull/4080))
+- Deprecate thumbnail property for `Page` ([#4733](https://github.com/Shopify/polaris-react/pull/4733))
```
kyledurand: Should be under deprecations

kyledurand (commented):
We could also add something like this in the file

```
if (process.env.NODE_ENV === 'development' && thumbnail != null) {
  console.warn('The thumbnail prop from Page has been deprecated');
}
```

sarahill (approved):
🙌 🚢

kyledurand (approved):

PR #4735, opened 2021-11-25
Remove components alias and import components directly in tests

BPScott:

Hitting two birds with one PR - removing a confusing way of importing files, and a chonky test speed improvement.

`components` is usable as an alias for importing the components in tests. However feel it should be removed as it is usage is not intuitive and has the potential to accidentally break consumers:

- It is possible to use the alias to import a type in source files, and TS will pass but when the package is published consumers will get an error
- Importing concrete values (i.e. not types) fails in source code but works in test files. This is confusing and unintuitive to folks unfamiliar with the codebase.

In addition, using this file to access components means that Jest has to read and parse all components (as that file reexports every component) even if you only use a specific component or two in a given test. This slows tests down as jest is doing lots unneeded work. This gets particularly pathological when you're trying to run a single test or when you're running tests with a cold cache (i.e on CI).

In order to ensure that even relative path imports to the components index never reappear we should remove the components index file, per the recommendations and investigations in this [web decision doc](https://github.com/Shopify/web/blob/master/documentation/decisions/09%20-%20Explicit%20root%20components%20imports.md) that spurned the investigation in this PR

This PR replaces all instances of `import {BLAH} from 'components'` with `import {BLAH} from '../../BLAH'` or whatever is the correct relative path to that component is.

`loom.config.ts` and `tsconfig.json` have been updated to remove the ability to use `'components'` as an importable path

### How to 🎩

Tests / type-check passes

### Stats!

I mentioned this improves performance on cold-cache test runs on CI.

Running 3 ci-style cold test runs `rm -rf .loom && CI=1 yarn test` on main takes: 50.01s, 50.30s, 49.04s.
Running 3 the same on this branch takes:  32.57s 32.39s 33.74s

So I've our tests are now almost twice as fast :D

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -32,6 +32,7 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 
 - Tightened up what absolute imports are allowed. Removed `baseUrl` from `tsconfig.json`. Attempting to do an absolute import from `src/X` or `components/X` now results in a error when type-checking. ([#4643](https://github.com/Shopify/polaris-react/pull/4643))
 - Remove analyze custom properties check. ([#4718](https://github.com/Shopify/polaris-react/pull/4718))
+- Removed support for importing from `components` as it slows tests down ([#4735](https://github.com/Shopify/polaris-react/pull/4735))
 
 ### Dependency upgrades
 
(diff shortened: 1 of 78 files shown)


alex-page (approved):
I will miss `components` but this makes sense for the system today.

kyledurand (approved):
Down for this. I usually lean on VS to handle my imports via the quick fix menu

image

The problem now would be that we don't have a meaningful error when someone tries to import from `../components`

image

Not a blocker but do you know of a way to fix that?

BPScott:
I'm going to open up a follow up PR where I nuke `src/components/index.ts`. All the component reexports shall happen directly in src/index.ts.

BPScott:
There it is ^^ #4739

PR #4740, opened 2021-11-26
[SkeletonPage] Deprecate secondaryActions

alex-page:

`Page` component no longer has secondary actions placed where this component renders them. It also adds an unnecessary layer of complexity with random sizes.

Deprecates `secondaryActions` and removes the examples.


kyledurand (approved):
So much 🔥 this week

PR #4743, opened 2021-11-26
[Page] Remove thumbnail prop

alex-page:

Clean up for v8.0.0

Removes deprecated property thumbnail and the related styles, examples and tests.

### DIFF
--- a/src/components/Page/components/Header/components/Title/Title.tsx
+++ b/src/components/Page/components/Header/components/Title/Title.tsx
@@ -48,9 +48,9 @@ export function Title({
   ) : null;
 
   return (
-    <div>
+    <>
       {wrappedTitleMarkup}
       {subtitleMarkup}
-    </div>
+    </>
   );
 }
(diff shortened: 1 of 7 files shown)


src/components/Page/components/Header/components/Title/Title.tsx
```diff
@@ -59,17 +47,10 @@ export function Title({
...
-  const pageTitleClassName = thumbnail ? styles.hasThumbnail : undefined;
-
   return (
-    <div className={pageTitleClassName}>
-      {thumbnailMarkup}
-      <div className={styles.TitleAndSubtitleWrapper}>
-        {wrappedTitleMarkup}
-        {subtitleMarkup}
-      </div>
+    <div>
```
kyledurand: Can we use a fragment here?
```suggestion
    <>
```

kyledurand (approved):
Looks good. Nice little cleanup 👏

src/components/Page/components/Header/components/Title/Title.tsx
alex-page: I want to investigate before merging. My gut feeling is no because there is some layout that wraps the title.

PR #4745, opened 2021-11-27
Bump size-limit to v6

alex-page:


BPScott:
Seems like webpack 5 got smart, and treeshakes waaay better now, so we need to trick it into not shaking - There's some ideas in https://github.com/ai/size-limit/issues/261

kyledurand (approved):

alex-page:
This is still important, but should probably be a new PR.

PR #4752, opened 2021-11-29
Cleanup set-root-property.ts

kyledurand:


alex-page (approved):

src/utilities/set-root-property.ts
```diff
@@ -1,13 +1,10 @@
...
-  if (document == null) {
-    return;
-  }
+  if (!document) return;
 
-  const styleNode =
-    node && node instanceof HTMLElement ? node : document.documentElement;
-  styleNode && styleNode.style.setProperty(name, value);
+  const styleNode = node ? node : document.documentElement;
+  styleNode.style.setProperty(name, value);
```
aaronccasanova: **suggestion(non-blocking):** Thoughts on naming this [`element`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/style#syntax)? `styledNode` makes me think it's an instance of `document.querySelector('style')`
```suggestion
  const element = node || document.documentElement;

  element.style.setProperty(name, value);
```

aaronccasanova (approved):
Looks good! Left a small/non-blocking comment. Feel free to mark as resolved and merge 👍

PR #4753, opened 2021-11-29
Move z-indexes to tokens

kyledurand:
Fixes: https://github.com/Shopify/polaris-react/issues/4608


src/tokens/_z-index.ts
```diff
@@ -0,0 +1,17 @@
...
+  'z-top-bar': '512',
+  'z-context-bar': '513',
+  'z-small-screen-loading-bar': '514',
+  'z-nav-backdrop': '515',
+  'z-nav': '516',
+  'z-skip-to-content': '517',
+  'z-backdrop': '518',
+  'z-modal': '519',
+  'z-toast': '520',
+  'z-devUi': '521',
```
alex-page: I am interested to know what people think about keeping these semantic names? I think they offer little value and make it slower to understand what sits on top of other elements. I think we could radically simplify it to `z-521` for now.

The upside of the current approach is we can change them quickly. I wouldn't be against calling them `z-1` `z-2` `z-3` until we are confident in the values.

alex-page (approved):
This can ship! I added one comment that can be addressed in a follow up.

src/tokens/_z-index.ts
lgriffee: I agree that the semantic names don't add value and should be removed. Either of those suggested naming conventions feel like they could work as a placeholder until we refine/rename them in our second pass over the tokens.

lgriffee (approved):
🔥💯

UNRELEASED-v8.md
```diff
@@ -22,6 +22,7 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 - Removed the `$duration-data` global variable ([#4699](https://github.com/Shopify/polaris-react/pull/4699))
 - Removed the `spacing()` scss function and replaced any instances with tokens ([#4691](https://github.com/Shopify/polaris-react/pull/4691/))
 - Removed the `px()` scss function ([#4751](https://github.com/Shopify/polaris-react/pull/4751))
+- Removed `z-index()` scss function ([#4753](https://github.com/Shopify/polaris-react/pull/4753))
```
kyledurand:
```suggestion
- Removed the `z-index()` scss function ([#4753](https://github.com/Shopify/polaris-react/pull/4753))
```

src/components/Frame/components/ToastManager/ToastManager.scss
```diff
@@ -10,7 +10,7 @@
 
 .ToastManager {
   position: fixed;
-  z-index: z-index(toast, $fixed-element-stacking-order);
+  z-index: var(--z-toast);
```
kyledurand:
```suggestion
  z-index: var(--p-z-toast);
```

src/components/TopBar/components/SearchField/SearchField.scss
```diff
@@ -6,15 +6,12 @@ $input-height: rem(34px);
...
-  action: 3,
-);
-
 .SearchField {
-  z-index: z-index(modal, $fixed-element-stacking-order);
+  --pc-backdrop: 1;
+  --pc-input: 2;
+  --pc-icon: 3;
+  --pc-action: 3;
+  z-index: var(--pc-modal);
```
kyledurand:
```suggestion
  z-index: var(--p-z-modal);
```

src/tokens/_z-index.ts
```diff
@@ -0,0 +1,17 @@
+import type {TokenGroup} from './tokens';
+
+export const zIndex: TokenGroup = {
+  'z-content': '100',
+  'z-overlay': '400',
+  'z-globalRibbon': '510',
```
kyledurand:
```suggestion
  'z-global-ribbon': '510',
```

src/tokens/_z-index.ts
kyledurand: Love it. I'll follow up this aft

Issue #4605, opened 2021-11-09
Replace rem() function with tokens and postcss plugin


alex-page:
- [x] Investigate [pixrem](https://www.npmjs.com/package/pixrem) and [pxtorem](https://www.npmjs.com/package/postcss-pxtorem).

Release steps
- [x] Ship pxtorem in next release
- [x] Remove `rem()` function from public sass
- [x] Remove `px()` function from public sass
- [x] Update `html` `font-size` to `100%` and ensure any `rem` values are updated accordingly
- [x] #4776

BPScott:
While you're poking at removing stuff in _utilities.scss, could you remove the `em()` and `px()` functions too?

Both the rem and em functions trigger deprecation warnings in dart-sass that can't be resolved without breaking compatibility with node-sass (short of just deleting the functions), so if we're getting rid of one it'd be good to get rid of both.

aveline:
This issue should also include updating `html` `font-size` to `100%`

PR #4761, opened 2021-11-29
Remove `rem()` function from public sass

aveline:

Part of
* #4605


src/components/ActionList/ActionList.scss
```diff
@@ -1,10 +1,11 @@
 @import '../../styles/common';
 
-$image-size: rem(20px);
-$item-min-height: rem(40px);
-$item-vertical-padding: ($item-min-height - line-height(body)) / 2;
-
 .ActionList {
+  --pc-image-size: 20px;
+  --pc-item-min-height: var(--p-space-10);
+  --pc-item-vertical-padding: calc(
```
aaronccasanova: **thought:** Per the recent discussions around token prefixes, I think these need to be renamed to `--pc-actionlist-`.

src/components/ActionMenu/components/SecondaryAction/SecondaryAction.scss
```diff
@@ -24,7 +24,7 @@ $button-spacing: rem(12px);
 
     @include breakpoint-after($breakpoint) {
       border: none !important;
-      @include focus-ring($border-width: rem(0));
+      @include focus-ring($border-width: 0);
```
aaronccasanova, line 27: I'm not sure why we were converting `0` to `rem`s, but should we add `px` to `0` here?

src/components/Autocomplete/components/MappedAction/MappedAction.scss
```diff
@@ -1,10 +1,11 @@
...
-$image-size: rem(20px);
-$item-min-height: rem(40px);
-$item-vertical-padding: ($item-min-height - line-height(body)) / 2;
-
 .ActionContainer {
+  --pc-image-size: 20px;
+  --pc-item-min-height: var(--p-space-10);
+  --pc-item-vertical-padding: calc(
+      var(--pc-item-min-height) - #{line-height(body)}
+    ) / 2;
```
aaronccasanova: This is now an interesting mix of `css` `calc` math and `scss` math. Should we put everything in `calc` and add parens to maintain the same order of operations?

src/components/Autocomplete/components/MappedAction/MappedAction.scss
```diff
@@ -1,10 +1,11 @@
 @import '../../../../styles/common';
 
-$image-size: rem(20px);
-$item-min-height: rem(40px);
-$item-vertical-padding: ($item-min-height - line-height(body)) / 2;
-
 .ActionContainer {
+  --pc-image-size: 20px;
```
aaronccasanova: Same question as above in regards to custom property prefixing ☝️

`--pc-autocomplete-`?

src/components/Badge/Badge.scss
```diff
@@ -17,7 +17,7 @@
   font-weight: var(--p-font-weight-regular);
 
   @media print {
-    border: solid rem(0.1px) var(--p-border);
+    border: solid 0.1px var(--p-border);
```
aaronccasanova, line 20: `0.1px` - Wonder if this was intentional?

src/components/BulkActions/BulkActions.scss
```diff
@@ -110,7 +109,7 @@ $bulk-actions-offset-slide-in-start: rem(-40px);
 .Slide-appear,
 .Slide-enter,
 .Slide-exit {
-  transform: translateX($bulk-actions-offset-slide-in-start);
+  transform: translateX(calc(-1 * var(--p-space10)));
```
aaronccasanova:
```suggestion
  transform: translateX(calc(-1 * var(--p-space-10)));
```

src/components/Choice/Choice.scss
```diff
@@ -65,7 +59,7 @@ $control-vertical-adjustment: rem(2px);
 }
 
 .Descriptions {
-  padding-left: calc(var(--p-space-2) + var(--p-choice-size, #{$control-size}));
+  padding-left: calc(var(--p-space-2) + var(--p-choice-size));
```
aaronccasanova, line 62: **question:** Do we no longer need the fallback?

src/components/ChoiceList/ChoiceList.scss
```diff
@@ -20,7 +18,7 @@ $control-size: rem(16px);
 
 .ChoiceChildren {
   margin-bottom: var(--p-space-2);
-  padding-left: calc(var(--p-space-2) + var(--p-choice-size, #{$control-size}));
+  padding-left: calc(var(--p-space-2) + var(--p-choice-size));
```
aaronccasanova, line 21: Same as above. Do we need the fallback?

src/components/IndexTable/IndexTable.scss
```diff
@@ -377,7 +376,7 @@ $loading-panel-height: rem(53px);
 }
 
 .StickyTableHeading-second-scrolling {
-  padding: 0 rem(0.6px) 0 var(--p-space-4);
+  padding: 0 0.6px 0 var(--p-space-4);
```
aaronccasanova, line 379: I'm not used to seeing decimals for `px` units. Similar to the above comment, do you think this is intentional?

src/styles/foundation/_focus-ring.scss
```diff
@@ -8,15 +8,11 @@
...
   // calc does not like performing addition with a unitless number (`0`, NOT `0px`)
-  // This is a problem because `rem(0px)` returns `0`, not `0px`.
+  // This is a problem because `0px` returns `0`, not `0px`.
   // Make sure that we can handle unitless zeros by not trying to do math with them
-  $offset: if(
-    $border-width == 0,
-    rem(1px),
-    calc(#{$border-width} + #{rem(1px)})
-  );
+  $offset: if($border-width == 0, 1px, calc(#{$border-width} + 1px));
```
aaronccasanova: Our regex may have unintentionally updated this comment/variable. We should probably tophat this variable, make sure it works as expected, and update or remove the comment. Thoughts?

aaronccasanova (approved):
Nice work Aveline! Regex FTW 🙌 

I tried to be as thorough as possible, but would recommend waiting for another set of eyes before merging 🙂

src/components/Icon/Icon.scss
```diff
@@ -2,8 +2,8 @@
 
 .Icon {
   display: block;
-  height: rem(20px);
-  width: rem(20px);
+  height: 20px;
+  width: 20px;
```
laurkim, line 6: Can we use `var(--p-space-5)` here?

src/components/IndexTable/components/Checkbox/Checkbox.scss
```diff
@@ -1,9 +1,9 @@
 @import '../../../../styles/common';
 
-$expanded-height: rem(32px);
-$expanded-width: rem(32px);
-$condensed-height: rem(44px);
-$condensed-width: rem(50px);
+$expanded-height: 32px;
+$expanded-width: 32px;
```
laurkim, line 4: Can we use `var(--p-space-8)` here?

src/components/Badge/Badge.scss
aveline: Since it's for print, I would think yes.

laurkim (approved):
nice! 💯 ✨ 

Only suggestion I have is to update the naming convention for custom component properties throughout the files (`--pc-[component]-`) that was mentioned in the [issue](https://github.com/Shopify/polaris-react/issues/4732#issuecomment-982859990), like @aaronccasanova called out.

src/components/ActionMenu/components/SecondaryAction/SecondaryAction.scss
aveline: Because it's `0` I don't think a unit is necessary

src/components/Choice/Choice.scss
aveline: I don't think so since custom properties are well supported now. @kyledurand are there any other reasons we might still need the fallback?

src/components/IndexTable/IndexTable.scss
aveline: I agree decimals for `px` is unusual, but I'm inclined to leave as is for the scope of this PR. We have intentions to review hardcoded values so might make more sense to review at that point.

src/components/ActionMenu/components/SecondaryAction/SecondaryAction.scss
aaronccasanova: Okay perfect, me either. Was mainly inquiring because of the previous implementation 👍

src/components/Badge/Badge.scss
aaronccasanova: Ah I think you're right that `print` is the important distinction here! It's not easy finding good reading material on this topic, but managed to find this [w3.org article on css units](https://www.w3.org/Style/Examples/007/units.en.html) that backs this up.

> In fact, CSS requires that 1px must be exactly 1/96th of an inch in all printed output. CSS considers that printers, unlike screens, do not need to have different sizes for px in order to print sharp lines. In print media, a px thus not only has the same visual appearance from one device to another, but indeed it is measurably the same.

src/components/IndexTable/IndexTable.scss
aaronccasanova: Oh that's perfect! Yes, let's address this later 👌

alex-page (approved):

PR #4763, opened 2021-11-30
Move border-radius custom properties to tokens

### LINKED ISSUE #4611: Move --p-border-radius custom properties to polaris-react

### LINKED ISSUE #4612: Remove --p-text-field-focus-ring-border-radius

lgriffee:

Closes #4612 
Closes #4611 

- Removes `--p-text-field-focus-ring-border-radius`
- Moves `--p-border-radius-*` legacy tokens to new token file
- Replaces any custom properties and values related to `border-radius` with tokens where possible


src/tokens/_border.ts
```diff
@@ -0,0 +1,8 @@
+import type {TokenGroup} from './tokens';
+
+export const border: TokenGroup = {
+  'border-radius-base': '0.4rem',
+  'border-radius-full': '50%',
+  'border-radius-slim': '0.2rem',
```
lgriffee: Based on my [audit of legacy border token instances](https://github.com/Shopify/polaris-react/issues/4611#issuecomment-982705882) I think we can remove `border-radius-slim` since it's not used in `polaris-react` or `web`. Let me know if that feels in scope for this PR.

lgriffee:
Do we also want to deal with moving other legacy border tokens ([see audit of legacy border token instances for context](https://github.com/Shopify/polaris-react/issues/4611#issuecomment-982705882)) to the new border token file? Or does that feel out of scope for this PR?

src/tokens/_border.ts
alex-page: I would order this `0.2` `0.4` `0.8` `50%` as that is the roundness increasing. Feel free to remove unused values.

src/tokens/_border.ts
```diff
@@ -0,0 +1,8 @@
+import type {TokenGroup} from './tokens';
+
+export const border: TokenGroup = {
```
alex-page: I am curious if we should put radii in a border category? It feels more related to shape or roundness. Interested to know peoples thoughts.

How would this be documented? @sarahill maybe you have  some opinions?

alex-page (approved):
Awesome work Laura. I added some comments that are not blocking and could be clean up work in the future.

src/tokens/_border.ts
aveline: I agree it feels more related to shape. But not a blocker.

aveline (approved):
🎉

src/tokens/_border.ts
aaronccasanova: Agreed. I think `shape` is an appropriate name for this token group.

src/tokens/_legacy-tokens.ts
```diff
@@ -7,10 +7,6 @@ import type {TokenGroup} from './tokens';
  * deprecating any undesired tokens.
  */
 export const legacyTokens: TokenGroup = {
-  'border-radius-slim': '0.2rem',
-  'border-radius-base': '0.4rem',
-  'border-radius-wide': '0.8rem',
-  'border-radius-full': '50%',
```
aaronccasanova, line 13: Thanks for the due diligence in cleaning up these legacy tokens 👍

aaronccasanova (approved):
🥇

src/components/Avatar/Avatar.scss
```diff
@@ -1,19 +1,18 @@
...
-$extra-small-size: rem(24px);
-$small-size: rem(32px);
-$medium-size: rem(40px);
-$large-size: rem(60px);
-
 .Avatar {
+  --pc-extra-small-size: 24px;
+  --pc-small-size: 32px;
+  --pc-medium-size: 40px;
+  --pc-large-size: 60px;
```
alex-page: These might have some conflicts with @kyledurand's PR. We also need to do `--pc-avatar-` so we don't get any nested problems with similar custom properties e.g. `--pc-small-size` for avatar and `--pc-small-size` for text inside it.

src/tokens/_shape.ts
```diff
@@ -0,0 +1,11 @@
+import type {TokenGroup} from './tokens';
+
+export const shape: TokenGroup = {
+  'border-radius-1': '0.2rem',
+  'border-radius-2': '0.4rem',
+  'border-radius-3': '0.8rem',
+  'border-radius-4': '1.2rem',
+  'border-radius-5': '1.6rem',
+  'border-radius-6': '2rem',
```
alex-page: These should use `px` values like our spacing values. 0.2rem = 2px. If possible we should also follow the same number pattern used there for consistency.

alex-page (approved):

lgriffee:
Before merging this there are several strange styling issues with a few components that need to be resolved (see Chromatic). @kyledurand and I jammed on a few this afternoon but there's still more to resolve.

src/components/RangeSlider/components/DualThumb/DualThumb.tsx
```diff
@@ -9,6 +9,7 @@ import type {RangeSliderProps, DualValue} from '../../types';
 import {Labelled, labelID} from '../../../Labelled';
 import {EventListener} from '../../../EventListener';
 import {Key} from '../../../../types';
+import sharedStyles from '../../RangeSlider.scss';
```
kyledurand, line 12: This is a bit of an odd convention but I think makes sense for future proofing. In the future we can apply shared  classes and 🔥 mixins 😄

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -160,15 +162,17 @@
 /// Output value indicator
 $range-output-size: 32px;
 $range-output-translate-x: calc(
-  -50% + var(--Polaris-RangeSlider-output-factor) * #{$range-thumb-size}
+  -50% + var(--Polaris-RangeSlider-output-factor) * #{var(
```
kyledurand, line 165: Not a blocker since this will compile ok but we don't need to interpolate var() so `#{var()}` is unnecessary

src/tokens/_shape.ts
```diff
@@ -0,0 +1,11 @@
+import type {TokenGroup} from './tokens';
+
+export const shape: TokenGroup = {
+  'border-radius-05': '2px',
+  'border-radius-1': '4px',
+  'border-radius-2': '8px',
+  'border-radius-3': '12px',
+  'border-radius-4': '16px',
+  'border-radius-5': '20px',
+  'border-radius-6': '50%',
```
kyledurand, line 10: I think this should still be:
```suggestion
  'border-radius-full': '50%',
```
or even 
```suggestion
  'border-radius-half': '50%',
```
I'll leave the naming up to you but I feel like it's part of a different group than the numbered values

src/components/DropZone/DropZone.scss
```diff
@@ -17,7 +17,7 @@ $dropzone-min-height-small: 50px;
 }
 
 @mixin set-border-radius {
-  border-radius: calc(var(--p-border-radius-base) + #{border-width()} + 1px);
+  border-radius: calc(var(--p-border-radius-1) + #{border-width()} + #{1px});
```
kyledurand, line 20:
```suggestion
  border-radius: calc(var(--p-border-radius-1) + #{border-width()} + 1px);
```

src/components/IndexTable/IndexTable.scss
```diff
@@ -416,17 +416,17 @@ $loading-panel-height: 53px;
 }
 
 $scroll-bar-size: var(--p-space-2);
-$scroll-bar-border-radius: 4px;
 
 .ScrollBarContainer {
+  --pc-scroll-bar-container-border-radius: 4px;
```
kyledurand, line 421: I think we can 🔥 this custom property and use `var(--p-border-radius-1)` directly below

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -63,7 +61,9 @@
     var(--pc-single-thumb-progress-upper) 100%;
 
   @mixin rangeSliderThumbPosition($size) {
-    margin-top: calc(-1 * (#{$size} - #{$range-track-height}) / 2);
+    margin-top: calc(
+      -1 * (#{$size} - #{var(--pc-range-slider-track-height)}) / 2
```
kyledurand, line 65:
```suggestion
      -1 * (#{$size} - var(--pc-range-slider-track-height)) / 2
```

src/components/RangeSlider/components/SingleThumb/SingleThumb.scss
```diff
@@ -183,7 +187,9 @@ $range-output-spacing: 16px;
       var(--p-range-slider-thumb-size-base);
     opacity: 1;
     visibility: visible;
-    bottom: calc(#{$range-thumb-size} + #{$range-thumb-size-difference});
+    bottom: calc(
+      #{var(--pc-range-slider-thumb-size)} + #{$range-thumb-size-difference}
```
kyledurand, line 191:
```suggestion
      var(--pc-range-slider-thumb-size) + #{$range-thumb-size-difference}
```

kyledurand (approved):
Code looks great. I'll check chromatic and approve there too if :gucci:

Feel free to address these comments in a follow up. Nothing blocking

lgriffee:
@kyledurand Thank you so much for all your feedback and for helping me to troubleshoot the RangeSlider issues!! I'm going to address all your feedback in a follow up PR!

src/tokens/_shape.ts
lgriffee: Yeah I was wondering about this—glad you brought it up! I like `'border-radius-half':` in case we have a 100% value at some point.

PR #4768, opened 2021-11-30
Cleanup component level custom properties

kyledurand:
Part of https://github.com/Shopify/polaris-react/issues/4732

This changes all properties created at component levels to the `--pc-<component>` prefix convention. I did this fairly quickly so it's probably going to be a bit broken. Don't feel like you need to review straight away

### DIFF
--- a/src/components/ActionList/ActionList.scss
+++ b/src/components/ActionList/ActionList.scss
@@ -31,7 +31,7 @@ $item-vertical-padding: ($item-min-height - line-height(body)) / 2;
 
 .Title {
   @include text-style-subheading;
-  padding: var(--p-space-1) var(--p-space-4) var(space-3) var(--p-space-4);
+  padding: var(--p-space-1) var(--p-space-4) var(--p-space-3) var(--p-space-4);
 
   &.firstSectionWithTitle {
     padding-top: var(--p-space-3);
--- a/src/components/ActionList/ActionList.scss
+++ b/src/components/ActionList/ActionList.scss
@@ -31,13 +31,10 @@ $item-vertical-padding: ($item-min-height - line-height(body)) / 2;
 
 .Title {
   @include text-style-subheading;
-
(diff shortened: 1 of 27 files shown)


src/components/ActionList/ActionList.scss
```diff
@@ -31,13 +31,10 @@ $item-vertical-padding: ($item-min-height - line-height(body)) / 2;
...
   @include text-style-subheading;
-
-  --p-titleVerticalSpacing: calc(var(--p-space-2) * 1.5);
-  padding: var(--p-space-1) var(--p-space-4) var(--p-titleVerticalSpacing)
-    var(--p-space-4);
+  padding: var(--p-space-1) var(--p-space-4) var(space-3) var(--p-space-4);
 
   &.firstSectionWithTitle {
-    padding-top: var(--p-titleVerticalSpacing);
+    padding-top: var(space-3);
```
alex-page:
```suggestion
    padding-top: var(--p-space-3);
```

src/components/ColorPicker/ColorPicker.scss
```diff
@@ -13,9 +13,9 @@ $inner-shadow: inset 0 0 2px 0 rgba(0, 0, 0, 0.5);
 }
 
 .ColorPicker {
-  --pc-color: 10;
-  --pc-adjustments: 20;
-  --pc-dragger: 30;
+  --pc-color-picker-color: 10;
```
alex-page: Maybe this should be called sheet or background. It's not really `color`. Any thoughts @kyledurand ?

alex-page (approved):
This is going to set us up to move super fast. Thanks @kyledurand 🥇

src/components/ColorPicker/ColorPicker.scss
kyledurand: Oh yeah true. This node sets the root stacking context for this component with `position: absolute` so I'm just going to go with `--pc-color-picker`

PR #4770, opened 2021-12-01
Remove unused custom properties

kyledurand:
These don't appear to be used anywhere


alex-page (approved):
Can we quickly search `shopify/web` to see if these are used? I also think we should add this to the unreleased-v8.md file. Just in case.

kyledurand:
Did a search through web to be safe but I don't know if these were ever exposed. I'll add to unreleased-v8 just in case though 👍

PR #4778, opened 2021-12-01
Bump node to v16

kyledurand:
Part of https://github.com/Shopify/polaris-react/issues/4549


BPScott:
You'll also want to add an `engines` field to `package.json` explicitly state the supported node versions, so the package will refuse to install on older node versions

```
 "engines": {
    "node": ">=16.13.0"
  },
```

https://github.com/Shopify/loom/blob/main/packages/loom/package.json#L29 has a demo of usage

BPScott:
Also FUN this is triggering from-source builds of node-sass on CI, which are then failing. You can fix that up by updating to the latest version of node-sass, which as of v6 includes precompiled binaries for node 16 x64. (you could do that against main)


The FUN for this shall eventually continue locally as there is no precompiled binary for m1 macs (it's a different processor architecture - arm not x64). The long term fix for this shall be moving to dart-sass which doesn't require any from-source compilation, but that in turn shall result in breaking support for the public api to be used by consumers that use node-sass - which currently is EVERYBODY so that's non-starter for a bit unless you make a call to just say "v8 exposes no public api, we'll provide a file you can copy+paste into your repo". (related: https://github.com/Shopify/polaris-react/issues/4725 / https://github.com/Shopify/web/issues/53583)

kyledurand:
Hey @BPScott I was looking at some other loom / build configs and they are looking super clean. I took a stab at changing up our config but build output was getting a little gross with classnames looking something like `._Avatar {`. I also dropped some console warnings in the description, not sure if you've seen those before

tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "@shopify/typescript-configs/library",
   "compilerOptions": {
+    "jsx": "react",
```
BPScott: This is already [the default](https://github.com/Shopify/web-configs/blob/main/packages/typescript-configs/dom.json#L4) (library itself inherits from dom). You don't need this change. This is going to be changed from the default value in https://github.com/Shopify/polaris-react/pull/4694 anyway.

BPScott:
>  I also dropped some console warnings in the description, not sure if you've seen those before

[eslint-plugin-node](https://www.npmjs.com/package/eslint-plugin-node) notes

> It recommends a use of the "engines" field of package.json. The "engines" field is used by node/no-unsupported-features/* rules.

Add the engines field as recommended above and then the no-unsupported-features rules will realize you're targeting node 16 and then shutup

BPScott:
I wonder if you can get size-limit working on node 16 (and thus it be happy with having v16 in the engines field) by running the setup-node action (with same options as in our CI run to install node v16) action before we run the size-limit action in [here](https://github.com/Shopify/polaris-react/blob/5c09fae48f9749a93aae48d6a8db47244c1af282/.github/workflows/size-limit.yml#L14)

laurkim (approved):
nice 🔥 💯

BPScott (approved):

PR #4782, opened 2021-12-02
Specify node version with size-limit.yml

alex-page:
GitHub changed their default node version which means that our action tried to run with this version and failed. Specifying a version of node means that it will use that verison.


kyledurand (approved):

laurkim (approved):
nice! 💯

PR #4783, opened 2021-12-02
Bump node-sass and sass-loader

kyledurand:

Needed to fix test runs in v8 where size limit build node 12 and node 16 bundles


BPScott:
Looks like the use of matrix has changed the CI run name slightly - you'll need to go update the branch protection rules to remove the "Test" expectation and add "Test (12.22.0)" after merging.

You might also need to update shipit.yml - https://github.com/Shopify/polaris-react/blob/main/shipit.yml#L3

BPScott (approved):

PR #4784, opened 2021-12-02
Set root font size to 100%

aveline:

Part of [#4605](https://github.com/Shopify/polaris-react/issues/4605)

- Set the root font size to `100%` instead of `62.5%`
- Update `pxtorem` root value to `16` instead of `10`
- All calculated `rem` values will be updated accordingly
- Update some hardcoded `rem` values in stories
- A sketchy workaround to normalize Chromatic snapshots due to the change in root font size
Storybook applies some default padding in `rem`, but the value changed with the change to the root font size which caused all 347 stories to have diffs in Chromatic. Forcing that padding to remain `10px` for now allowed for verification of no other unexpected diffs because of this change. The padding override should probably be removed at some point in the future. With Chromatic comparing against `main` it would be too noisy to remove it now while we continue to work on the `v8` branch.


src/components/AppProvider/AppProvider.scss
```diff
@@ -22,7 +22,7 @@ button {
 
 html {
   position: relative;
-  font-size: ($base-font-size / $default-browser-font-size) * 100%;
```
kyledurand, line 25: Are these vars used elsewhere? Maybe we can 🔥 one or both

aveline: They are but will be removed soon! Possibly as part of https://github.com/Shopify/polaris-react/issues/4777

.storybook/preview-body.html
```diff
@@ -0,0 +1,5 @@
+<style>
+  .sb-show-main.sb-main-padded {
+    padding: 10px;
+  }
+</style>
```
alex-page, line 5: Love this to make tophatting this PR easy. Lets do a follow up PR that removes it by itself 💯

alex-page (approved):
I love this. There were some issues when shipping this to `shopify/web` last time. Some thoughts:
1. We can ship to `v8.0.0-major` and hope that this works in `web`. This could cause painful regressions but I am willing to learn and iterate.
2. Try and implement this in `main`. It would be a lot more time intensive replacing all the `rem` values. We could then learn ahead of time by tophatting `shopify/web` if there are any problems.

lgriffee (approved):
🔥🚀💯

PR #4788, opened 2021-12-02
Update shipit ci requirements

kyledurand:


shipit.yml
```diff
@@ -1,11 +1,11 @@
 ci:
   require:
-    - 'Test'
+    - node-tests (12.22.0)
```
kyledurand: Unsure if this should be similar to quilt format or just 
```suggestion
    - 'Test (12.22.0)'
```

BPScott: It should be `Test (12.22.0)` in polaris to match [the name of the action that is ran](https://github.com/Shopify/polaris-react/blob/main/.github/workflows/ci.yml#L7) (which in quilt is `node-tests`)

BPScott (approved):

PR #4791, opened 2021-12-02
Theme and update storybook

kyledurand:
This has been bugging me for quite some time and now that we're hosted publicly (again) I felt I could take a bit of time and bring in a theme I did in [Jan](https://github.com/Shopify/polaris-react/pull/3903) and update at the same time

Before:
storybook polaris shopify com_

After:
localhost_6006__path=_story_playground-playground--details-page


.storybook/manager.js
```diff
@@ -1,13 +1,48 @@
...
+
+  // Base theme
   theme: create({
     base: 'light',
     brandTitle: 'Shopify Polaris Storybook',
     brandUrl: '/',
-    brandImage: null,
-    appBorderRadius: 0,
+    brandImage:
+      'https://cdn.shopify.com/shopifycloud/brochure/assets/brand-assets/shopify-logo-primary-logo-456baa801ee66a0a435671082365958316831c9960c480451dd0330bcdae304f.svg',
```
kyledurand, line 23: This hash can change but it hasn't in a year and if it does it's a good indicator of when we need to update the logo anyway.

From what I've read in storybook docs it needs to be hosted, we can't just throw it in a static directory for some reason

.storybook/manager.js
```diff
@@ -1,13 +1,48 @@
 import {addons} from '@storybook/addons';
 import {create} from '@storybook/theming';
 
+const colors = {
+  primary: '#008060',
+  secondary: '#6d7175',
+  background: '#f6f6f7',
+  border: '#202123',
+  surface: '#ffffff',
+  text: '#202223',
```
alex-page, line 10: In v8 we should import color values from `token.ts`.

alex-page (approved):
Awesome. Love it. Happy to ship and iterate is necessary.

package.json
```diff
@@ -102,15 +102,15 @@
     "@shopify/postcss-plugin": "^5.0.1",
     "@shopify/prettier-config": "^1.1.2",
     "@shopify/react-testing": "^3.2.4",
-    "@shopify/storybook-a11y-test": "^0.0.1",
+    "@shopify/storybook-a11y-test": "^0.3.0",
```
laurkim: I think the way accessibility testing is used has changed from [v0.0.1](https://www.npmjs.com/package/@shopify/storybook-a11y-test/v/0.0.1) to [v0.3.0](https://www.npmjs.com/package/@shopify/storybook-a11y-test/v/0.3.0). Updating per the below will probably get the CI check to pass.

In `scripts/accessibility-check.js`, we should import `{testPages, getCurrentStoryIds}` instead of `{storybookA11yTest}` on [line 4](https://github.com/Shopify/polaris-react/blob/e25973358f4bcb9c65ef6147cd105de0d4044d9e/scripts/accessibility-check.js#L4).

Then use `await getCurrentStoryIds()` to get `storyIds` (previously named variable [options](https://github.com/Shopify/polaris-react/blob/e25973358f4bcb9c65ef6147cd105de0d4044d9e/scripts/accessibility-check.js#L7)). And `await testPages()` with those `storyIds` to get [results](https://github.com/Shopify/polaris-react/blob/e25973358f4bcb9c65ef6147cd105de0d4044d9e/scripts/accessibility-check.js#L16). More specific details are in the usage section for [v0.3.0](https://www.npmjs.com/package/@shopify/storybook-a11y-test/v/0.3.0) (I can't link that section directly, just the whole page 😞)

laurkim (approved):
nice!

small suggestion for updating our accessibility check script with the updated `storybook-a11y-test` but everything else looks 💯

kyledurand:
Reverting the storybook upgrade for now. It open up many cans of worms and can wait for a loom release with [this commit](https://github.com/Shopify/loom/pull/311) which should make things easier

PR #4793, opened 2021-12-03
Replace border-radius() function with tokens

### LINKED ISSUE #4613: Replace border-radius() function with tokens

lgriffee:

Fixes #4613 

To deprecate the `border-radius()` function from our public sass API and replace it with tokens. We no longer want to support this functionality in the system.

- Removing all instances of `border-radius()`
- Adding the following new tokens to replace function values (naming is temporary for now, more work needs to be done on the `border-radius` token scale):
  - `--p-border-radius-base`
  - `--p-border-radius-large`
- Changing the naming convention for `border-radius` 50% token to differentiate it from `px` values
- Cleaning up unnecessary interpolation and custom properties


lgriffee:
I believe the 3 visual changes in Chromatic are from my last commit merging in v.9.0.0 and are not the result of changes from this branch. Would be great to have someone else confirm that too!

aveline (approved):
🎉

PR #4795, opened 2021-12-03
Loom linter plugin updates

BPScott:

These breaking changes to loom plugins make you specify eslint/stylelint/prettier
versions in the package.json, instead of relying on transitive
dependencies.

This gives polaris greater control over what exact versions of
eslint/stylelint/prettier they use.

- Major version bumps to `loom-plugin-{eslint,stylelint,prettier}`
- Specify versions of eslint/stylelint/prettier in package.json. These have been set to match what is currently used in the lock file

### How to 🎩

Lint passes


BPScott:
This is an opportunity to update to the latest version of eslint/stylelint/prettier and our shared configs `@shopify/eslint-plugin` and `@shopify/stylelint-plugin`. It can be easy to do that as a follow-up PR.

kyledurand (commented):
Code looks good but I think this will introduce some issues that need fixing 🧰 

I have some time this afternoon if you want me to take a look

kyledurand (approved):

BPScott:
Hilariously we’ve ended up hoisting different prettier versions to the top level locally vs on ci before this PR. Which is moderatly confusing and means we can't have a "code matches what was there previously, but explicit" approach. I’m gonna fixing that by just bumping prettier and applying the changes in this PR.

PR #4796, opened 2021-12-03
Storybook v6.4.7

kyledurand:


BPScott:
You'll want https://github.com/Shopify/quilt/pull/2106 but it's been released yet. I'll go publish a v0.4.0 that shall include that.

aaronccasanova (approved):
Successfully tophatted `yarn a11y-check` using Node `v16.9.1`. Every attempt using Node `v12.22.0` timed out after 30 seconds. Not sure if that's a side effect of the M1 machine or a deeper issue we need to address.

kyledurand:
> Every attempt using Node v12.22.0 timed out after 30 seconds.

That was happening to me as well and I had to build storybook again before the whole thing would pass. Even running node 16. I'm not so worried about it though because this has passed twice in ci using node 12.22 so I think we're good.

PR #4797, opened 2021-12-04
Update eslint

BPScott:

Keeping up to date.

- Update `eslint` to v8.3.0 and `@shopify/eslint-plugin` to v41.0.1
- Update `babel/*` rules to be `@babel/*`
- Apply autofixes
- Adjust generic naming so generic names start with a `T`
- Replace usage of toBeTruthy / toBeFalsey in tests with toBeDefined / toBe(true) / toBeFalse() etc
- Ensure `key` props are always before spreads to keep [react/jsx-key](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-key.md) happy
- Ignore occasional hardcoded-content warnings
- Add useMemo to a few places to prevent always rerendering to keep [react/jsx-no-constructed-context-values](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-constructed-context-values.md) happy

### To Test

CI passes


src/components/PolarisTestProvider/PolarisTestProvider.tsx
```diff
@@ -70,12 +70,15 @@ export function PolarisTestProvider({
   frame,
 }: PolarisTestProviderProps) {
   const Wrapper = strict ? StrictMode : Fragment;
-  const intl = new I18n(i18n || {});
-  const scrollLockManager = new ScrollLockManager();
+  const intl = useMemo(() => new I18n(i18n || {}), [i18n]);
```
kyledurand, line 73: Necessary or optimizing?

kyledurand (approved):
🙌

src/components/PolarisTestProvider/PolarisTestProvider.tsx
BPScott: Eslint warns with [react/jsx-no-constructed-context-values](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-constructed-context-values.md) when you sick something in a context that shall cause the context to be updated on every single rerender regardless of if the content changes or not. With the fix being to memoize objects you pass into contexts.

In this particular case it's probably not a big deal as we probably aren't causing rerenders on PolarisTestProvider because, but it's an easy fix to write code in the way to keep the linter happy rather than sprinkling `eslint-disable-next-line`s around

PR #4804, opened 2021-12-07
feat: Add stylelint plugin to validate polaris custom properties

aaronccasanova:

This PR introduces a custom stylelint plugin to validate custom property usage across `polaris-react`.

`@shopify/custom-properties-allowed-list` Plugin Features:
- Specify a list of allowed custom property names permitted in a `.scss` file.
- Specify a list of allowed custom property values permitted in a `.scss` file.
  (Note: Allowed custom property values can be configured per CSS property)

The `@shopify/custom-properties-allowed-list` plugin has been configured in `polaris-react` to:
- Ensure local component custom properties have a `--pc-` prefix.
- Ensure local component custom properties do NOT set the `--p-` prefix.
  (Note: `--p-` custom properties are auto generated from our Polaris tokens)
- Validates `--p-*` custom properties match existing Polaris tokens.

https://user-images.githubusercontent.com/32409546/144942268-3134205a-fc5a-4783-a351-fd75e1328b5e.mov

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist


BPScott:
Yesterday I merged in https://github.com/Shopify/polaris-react/pull/4798 which updates stylelint to be v14 in the main branch. Things might get less clunky if merge main into v8 and then apply this PR.

At that point it seems kinda silly to create a whole new tool for doing this - instead you might as well run this rule as part of the already present call to stylelint that runs as part of running `yarn lint`.


[While testing you can run just the stylelint steps in the lint command](https://github.com/Shopify/loom/blob/main/packages/loom-plugin-stylelint/README.md#running-only-stylelint-when-linting)

scripts/stylelint/config/index.js
```diff
@@ -0,0 +1,9 @@
+const path = require('path');
+
+require('ts-node').register({
```
BPScott: This feels tremendously hacky.

I think it's be nicer to write the plugin in plain js (heck we're in node 16 land here in the v8 branch so even mjs is on the table), with [jsdoc comments](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html) to provide typing. And then we'd be able to run this plugin as part of our regular stylelint run rather than needing a totally separate runtime

aaronccasanova: Very hacky indeed! We introduced `ts-node` to enable access to the [Polaris tokens from within the stylelint config](https://github.com/Shopify/polaris-react/blob/79e358e5e9491bafe73a8f4a5d363453aacf2ebc/scripts/stylelint/config/config.ts#L1). Any suggestions how we can import a `.ts` module in `.mjs`?

loom.config.ts
```diff
@@ -26,8 +26,10 @@ import type {} from '@shopify/loom-plugin-jest';
 // eslint-disable-next-line import/no-default-export
 export default createPackage((pkg) => {
   pkg.entry({root: './src/index.ts'});
+  pkg.entry({name: 'token-list', root: './src/token-list.ts'});
   pkg.use(
     buildLibrary({
+      rootEntrypoints: false,
```
BPScott, line 32: As discussed, we should spin setting `rootEntrypoints: false` and removing the references to root level `index.*` files in the various ignore files (as `rootEntrypoints: false` means those files [no longer gets generated](https://github.com/Shopify/loom/tree/main/packages/loom-plugin-build-library#entrypoints))  into a separate PR that targets main.

package.json
```diff
@@ -66,7 +66,8 @@
     "readme-update-version": "node ./scripts/readme-update-version",
     "version": "yarn run readme-update-version",
     "storybook": "start-storybook -p 6006 --quiet",
-    "storybook:build": "yarn run copy-polaris-tokens && build-storybook -o build-internal/storybook/static"
+    "storybook:build": "yarn run copy-polaris-tokens && build-storybook -o build-internal/storybook/static",
+    "check-script-types": "tsc -p scripts/tsconfig.json"
```
BPScott: This is no longer needed

scripts/stylelint/README.md
```diff
@@ -0,0 +1,91 @@
+## Allowed Custom Properties plugin
```
BPScott: `scripts` is a dumping ground for one-off commands. I reckon this stuff is worth pulling out to a top-level `stylelint` folder (so stylelint/README.md, rather than `scripts/stylelint/README.md`).

tsconfig.json
```diff
@@ -14,5 +14,5 @@
       "tests/*": ["./tests/*"]
     }
   },
-  "include": ["./config/typescript", "./src", "./tests", "./playground"]
+  "include": ["./config/typescript", "./src", "./tests", "./playground"],
```
BPScott: Can be put back to how it was

```suggestion
  "include": ["./config/typescript", "./src", "./tests", "./playground"]
```

BPScott (commented):
We've reworked a bunch of this! 

We've learnt that using [jsdoc to get typehinting in plain JS files](https://www.typescriptlang.org/docs/handbook/intro-to-js-ts.html) is super cool.

There's still some clunkyness - you need to run a build before running linting which is pretty weird - but we no longer need to do funkyness with `ts-node` and we're a lot nearer to a world where the plugin could live within [`@shopify/stylelint-plugin`](https://github.com/Shopify/web-configs/tree/main/packages/stylelint-plugin) and app/libs other than polaris could leverage the list of blessed tokens that polaris exposes (hi web).

We now build a new entrypoint, that gets generated in `build/cjs/token-list.js`, which exposes the array of custom properties that can be passed into our stylelint config. As an alternative to this we could expose this from the main entrypoint - `src/index.ts`, probably with some scary name `_SECRET_INTERNAL` or `_UNSTABLE` prefix to scare folks off using it for now.  We thiiiiink the real long-term fix is perhaps tokens should live in a location that both this stylelint config and polaris-react  can both depend upon - some plain js, or a json file maybe? This would also mean that you'd no longer need to do a build before running the linting

package.json
aaronccasanova: Roger that! https://github.com/Shopify/polaris-react/pull/4804/commits/333a25a9618ef9a528e31ab1a5842765db70875b

scripts/stylelint/README.md
aaronccasanova: Sounds good 👍  https://github.com/Shopify/polaris-react/pull/4804/commits/5563716ac291c90666404f8511655965c5f72c75

tsconfig.json
aaronccasanova: https://github.com/Shopify/polaris-react/pull/4804/commits/587c7e6d489848cb5fb238a328248269194def59

loom.config.ts
aaronccasanova: Sound good 👍  I created this issue for tracking https://github.com/Shopify/polaris-react/issues/4807

package.json
```diff
@@ -162,7 +163,8 @@
   "prettier": "@shopify/prettier-config",
   "stylelint": {
     "extends": [
-      "@shopify/stylelint-plugin/prettier"
+      "@shopify/stylelint-plugin/prettier",
```
BPScott: It might be worth moving our stylelint config our from a key in package.json, into a `.stylelintrc.js` file, and then we can inline the contents of `stylelint/config/index.js` into it`

aaronccasanova: Yes!! https://github.com/Shopify/polaris-react/pull/4804/commits/0b7a56c942f6995271890d766a2cd680cf44d283

aaronccasanova:
> We now build a new entrypoint, that gets generated in build/cjs/token-list.js, which exposes the array of custom properties that can be passed into our stylelint config. As an alternative to this we could expose this from the main entrypoint - src/index.ts, probably with some scary name _SECRET_INTERNAL or _UNSTABLE prefix to scare folks off using it for now. We thiiiiink the real long-term fix is perhaps tokens should live in a location that both this stylelint config and polaris-react can both depend upon - some plain js, or a json file maybe? This would also mean that you'd no longer need to do a build before running the linting

@BPScott I briefly discussed this with @alex-page and our plan is:
1. Get tokens + Stylelint plugin shipped (e.g. this PR)
2. Move tokens to JSON files
3. Update Polaris components and Stylelint plugin

.vscode/settings.json
```diff
@@ -23,5 +23,6 @@
     "{build,build-internal}/": true,
     "examples/*/build": true
   },
-  "typescript.tsdk": "./node_modules/typescript/lib"
+  "typescript.tsdk": "./node_modules/typescript/lib",
+  "cSpell.words": ["unprefixed"]
```
alex-page: I wouldn't add `cSpell.words`. Feels like an optional thing folks can add themselves. Any context on why we would want to prompt people to install this extension?

stylelint/README.md
```diff
@@ -0,0 +1,91 @@
+## Allowed Custom Properties plugin
+
+The purpose of this plugin is to ensure that we're following our established conventions for Polaris custom properties, and only using custom properties that are generated Polaris tokens.
+
+### Enforced Rules
+
+- If a `--p-*` custom property is used it must match an existing Polaris [token](../../../../src/tokens/tokens.ts).
```
alex-page:
```suggestion
- If a `--p-*` custom property is used it must match an existing Polaris [token](/src/tokens/tokens.ts).
```
Could we just say `/src/tokens/tokens.ts`?

stylelint/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+### How to configure:
+
+```js
+const stylelintConfig = {
+  rules: {
+    '@shopify/custom-properties-allowed-list': {
+      allowedProperties: ['/--pc-.+/'],
+      allowedValues: {
+        width: ['--p-space-0', '--p-space-1' /* etc... */],
```
alex-page, line 37: We don't use `width` so maybe we should remove this from the example? How would most people use this?

alex-page (approved):
Some non-blocking comments added. Thanks Aaron 💯

UNRELEASED-v8.md
```diff
@@ -27,6 +27,9 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 - Removed the `rem()` scss function and any references ([#4761](https://github.com/Shopify/polaris-react/pull/4761/))
 - Removed unnecessary `--p-border-radius-` properties and moved border-radius custom properties to tokens ([#4763](https://github.com/Shopify/polaris-react/pull/4763))
 - Dropped support for node < 16 ([#4778](https://github.com/Shopify/polaris-react/pull/4778))
+- Updated `--global-ribbon-height` CSS custom property to `--pc-frame-global-ribbon-height`
+- Updated `--pc-frame-offset` CSS custom property to `--pc-frame-offset`
```
kyledurand: I think this was just:
```suggestion
- Updated `--p-frame-offset` CSS custom property to `--pc-frame-offset`
```

kyledurand (approved):
This looks awesome Aaron 👏

stylelint/README.md
aaronccasanova: We arbitrarily added `width` to illustrate the behavior of the plugin. Happy to change it to another property, but I think it's important the example has both a CSS property and RegExp key as I'm hoping this plugin eventually makes its way into `@shopify/web-configs`.

.vscode/settings.json
aaronccasanova: Yeah, let's remove this. This got added with `git add .` and I agree we don't need to bake in an opinion on that here.

UNRELEASED-v8.md
aaronccasanova: Nice catch, TY!

stylelint/README.md
alex-page: Ok sounds good to me. Thanks for the context Aaron!

.vscode/settings.json
aaronccasanova: https://github.com/Shopify/polaris-react/pull/4804/commits/1cbf188997fed903e67dafd076ab0ec74b2b67c4

stylelint/README.md
aaronccasanova: https://github.com/Shopify/polaris-react/pull/4804/commits/9a3e1f4750a4731d1243d7f6b5ca672dff59780a

UNRELEASED-v8.md
aaronccasanova: https://github.com/Shopify/polaris-react/pull/4804/commits/d504e5c957f781f995fb5fe7cc2794e81f5b7e31

PR #4808, opened 2021-12-08
Use rems in tokens

kyledurand:
This adds a rem function similar to what we had before. Converting the values directly got really confusing so as annoying as the rem function is I think it makes sense to bring back. Open to other opinions though!


lgriffee (approved):
👍🏻💯

aveline (approved):

alex-page (approved):

PR #4813, opened 2021-12-09
[DropZone] Updating FileUpload styles

dfmcphee:

This updates the styles used by `DropZone` the the `FileUpload` component.

It changes the `FileUpload` button styles, removes the icon at larger sizes, updates the hover style on `DropZone`, and removes the `actionHint` unless a custom one is passed in.

This is what it looks like now:

Screen Shot 2021-12-09 at 10 56 33 AM
Screen Shot 2021-12-09 at 10 56 06 AM
Screen Shot 2021-12-09 at 11 06 25 AM
Screen Shot 2021-12-09 at 11 08 01 AM

FYI @ouellettejordan is going to sync up with @sarahill about potentially making this a button variation that is shared.

## 

### How to 🎩

Take a look around at the `DropZone` stories.

### 🎩 checklist

### DIFF
--- a/src/components/DropZone/components/FileUpload/FileUpload.scss
+++ b/src/components/DropZone/components/FileUpload/FileUpload.scss
@@ -43,7 +43,7 @@ $slim-vertical-padding: ($slim-min-height - line-height(body) - rem(2px)) / 2;
   text-align: center;
   appearance: none;
 
-  &:focus {
+  &.focused {
     outline: 0;
     box-shadow: none;
     @include focus-ring($style: 'focused');
--- a/src/components/DropZone/components/FileUpload/FileUpload.scss
+++ b/src/components/DropZone/components/FileUpload/FileUpload.scss
@@ -27,7 +27,8 @@ $slim-vertical-padding: ($slim-min-height - line-height(body) - rem(2px)) / 2;
 
 .Button {
   @include focus-ring;
-
+  display: inline-flex;
+  flex: 0 0 auto;
   border: none;
   border-radius: var(--p-border-radius-base);
   padding: rem(4px) rem(6px);
--- a/src/components/DropZone/components/FileUpload/FileUpload.scss
+++ b/src/components/DropZone/components/FileUpload/FileUpload.scss
@@ -25,7 +25,7 @@ $slim-vertical-padding: ($slim-min-height - line-height(body) - rem(2px)) / 2;
   vertical-align: bottom;
 }
 
-.Button {
+.Action {
   @include focus-ring;
   display: inline-flex;
   flex: 0 0 auto;
@@ -42,12 +42,6 @@ $slim-vertical-padding: ($slim-min-height - line-height(body) - rem(2px)) / 2;
   cursor: pointer;
   text-align: center;
   appearance: none;
-
-  &.focused {
-    outline: 0;
-    box-shadow: none;
-    @include focus-ring($style: 'focused');
-  }
 
   &:hover {
     color: var(--p-interactive-hovered);
--- a/src/components/DropZone/components/FileUpload/FileUpload.scss
+++ b/src/components/DropZone/components/FileUpload/FileUpload.scss
@@ -26,26 +26,36 @@ $slim-vertical-padding: ($slim-min-height - line-height(body) - rem(2px)) / 2;
 }
 
 .Button {
-  @include button-base;
   @include focus-ring;
+
+  border: none;
+  border-radius: var(--p-border-radius-base);
+  padding: rem(4px) rem(6px);
+  margin: 0;
+  text-decoration: none;
+  background: var(--p-surface-selected-pressed);
+  color: var(--p-interactive);
+  font-size: rem(12px);
+  font-weight: 600;
+  line-height: 1;
+  cursor: pointer;
+  text-align: center;
+  appearance: none;
+
+  &:focus {
+    outline: 0;
+    box-shadow: none;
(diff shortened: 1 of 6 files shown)


dfmcphee:
Storybook link to see the updated examples: https://5d559397bae39100201eedc1-xcqykclggj.chromatic.com/?path=/story/all-components-drop-zone--drop-zone-with-file-upload

src/components/DropZone/components/FileUpload/FileUpload.scss
```diff
@@ -26,26 +26,37 @@ $slim-vertical-padding: ($slim-min-height - line-height(body) - rem(2px)) / 2;
 }
 
 .Button {
-  @include button-base;
   @include focus-ring;
+  display: inline-flex;
+  flex: 0 0 auto;
+  border: none;
+  border-radius: var(--p-border-radius-base);
+  padding: rem(4px) rem(6px);
```
kyledurand, line 34: I think these are fine for now since we'll be changing these in the near future to `var(--p-space-2)`

src/components/DropZone/components/FileUpload/FileUpload.scss
```diff
@@ -26,26 +26,37 @@ $slim-vertical-padding: ($slim-min-height - line-height(body) - rem(2px)) / 2;
 }
 
 .Button {
-  @include button-base;
   @include focus-ring;
```
kyledurand, line 29: We're not getting focus rings on the button anymore. That seems intentional since it's a div and not tabbable so I think some styles can be removed from here and the .tsx

src/components/DropZone/components/FileUpload/FileUpload.scss
```diff
@@ -26,26 +26,37 @@ $slim-vertical-padding: ($slim-min-height - line-height(body) - rem(2px)) / 2;
...
+  appearance: none;
 
-  &.disabled {
-    @include base-button-disabled;
-    cursor: not-allowed;
-    box-shadow: none;
-  }
-
-  &.focused {
+  &:focus {
```
kyledurand: This never gets set I don't think. It relied on a `.focused` class before

src/components/DropZone/components/FileUpload/FileUpload.tsx
```diff
@@ -29,32 +29,16 @@ export function FileUpload(props: FileUploadProps) {
...
-  const actionTitleClassName = classNames(
-    styles.ActionTitle,
-    focused && !disabled && styles['ActionTitle-focused'],
-    disabled && styles['ActionTitle-disabled'],
-  );
-
-  const actionTitleMarkup = (
-    <div className={actionTitleClassName}>{actionTitle}</div>
-  );
+  const buttonMarkup = <div className={buttonStyles}>{actionTitle}</div>;
```
kyledurand:
```suggestion
  const actionTitleMarkup = <div className={buttonStyles}>{actionTitle}</div>;
```

src/components/DropZone/components/FileUpload/FileUpload.tsx
```diff
@@ -29,32 +29,16 @@ export function FileUpload(props: FileUploadProps) {
...
       `Polaris.DropZone.${allowMultipleKey}.actionTitle${typeSuffix}`,
     ),
-    actionHint = i18n.translate(
-      `Polaris.DropZone.${allowMultipleKey}.actionHint${typeSuffix}`,
-    ),
+    actionHint,
   } = props;
 
   const buttonStyles = classNames(
     styles.Button,
```
kyledurand:
```suggestion
  const actionClassName = classNames(
    styles.Action,
```

kyledurand (approved):
Code looks good but I noticed the background colors aren't really following the semantic naming pattern. Not a big deal with the token changes coming to v8. 

I also noticed that the button can't be focused inside of dropzone which makes sense because it's just a div and screen readers seem to get by ok, though it's not great. Left a few ideas on some things that can be cleaned up

src/components/DropZone/components/FileUpload/FileUpload.scss
dfmcphee: Can I use these now? Or will they not be there until v8? I would rather update now if I can.

src/components/DropZone/components/FileUpload/FileUpload.scss
dfmcphee: Good catch. I switched it back to the class.

src/components/DropZone/components/FileUpload/FileUpload.scss
kyledurand: Yeah all good to use them for now. They'll get caught in our coverage app and we'll replace them later

kyledurand:
image

I was thinking we omit the focus state on the action since it already exists on the container. Seems a bit jarring to me but I'll leave that up to the designers

ouellettejordan:
> I was thinking we omit the focus state on the action since it already exists on the container. Seems a bit jarring to me but I'll leave that up to the designers

Given that the button and the container do the same thing, I'm cool to remove the focus from the button. I think that makes sense.

ouellettejordan (approved):
I'm 👍 once @dfmcphee gets through the latest changes we discussed on Slack.

ouellettejordan:
Re: talking to @sarahill about adding this meta button to Polaris as a button variation. Sarah and I agreed that there likely is room for it in the system, but we're not ready quite yet. Let's ship this as a bespoke item _for now_.

Issue #4819, opened 2021-12-09
[Tag] Allow arbitrary children


jbalsas:
## Feature request summary

The `<Tag>` component should allow receiving arbitrary content as `children`

## Rationale

In some cases, we might want to add some additional content to a `<Tag>` like for example an icon or a link.  Here's an example of a current design where we'd like to be able to do this in `admin`

![a tag with an icon and a link](https://user-images.githubusercontent.com/905006/145489942-1305aa1d-e558-47d8-b78d-741fc40b7307.png)

## Current State

In essence, it would be neat if we could use `<Tag>` as in this [example](https://codesandbox.io/s/cocky-jang-2kjn6):

```jsx
<Tag onRemove={() => {}}>
  <Stack alignment="center" spacing="extraTight">
    <Icon size="extraSmall" source={WandMajor} />
    <UnstyledLink url="#url">This is a URL</UnstyledLink>
  </Stack>
</Tag>
```

This currently works, but it has **2 main caveats:**
- It will fail TS checks 🤓 
- The `ariaLabel` content will read something like `[Object object]` because we're assuming `children` is always a string (the tag value). 

## Proposal

I think this could be done in different ways:

#### Allow arbitrary content in `<Tag>`
- Make the `children` prop accept `Element`
- Add an extra `tagDescription` (or a better name) that:
  - Is used when `children` is of type `Element`
  - Can be used to override `ariaLabel` when `children` is a string and we want a different message

This would provide the more versatility to the tag at the cost of less consistency (can't predict how it will get used) and some worse DX (devs might need to style their content).

#### Allow restricted variations in `<Tag>`
- Add an `icon` prop so we can add icons (on the left?) to the tag
- One of:
  - Add a `url` prop so we could make the tags link somewhere
  - Allow `onClick` prop at the same time of the `onRemove` so devs can manually trigger an action when clicking the tag on removable tags

This would provide a better DX for that common case (not that it's so common now) at the price of bigger complexity as the number of props and its combinations will start to grow.

**I'd personally suggest we allow any content first. That way we don't have to keep adding this for new use cases and developers can self-serve. Then, if we detect some use case becomes very common, we can bake it in.**

![wasthattoomuch](https://user-images.githubusercontent.com/905006/145490979-7cb9b1c8-d24a-4a63-92c9-b690513c5cce.gif)

yurm04:
[...]

PR #4830, opened 2021-12-13
[WIP] [Tag] Allow arbitrary content and allow onClick and onRemove at the same time (#4819)

jbalsas:

This is POC to attempt to address the needs described in [[Tag] Allow arbitrary children #4819](https://github.com/Shopify/polaris-react/issues/4819)

**TL;DR**

We want to be able to use the `Tag` component in a way that:
- We can have `removable` tags that also have a main faction (navigate to the thing the tag describes)
- We can add richer content other than just text

#### [[Tag] Accepts ReactElement to be passed as children](https://github.com/Shopify/polaris-react/commit/3a43ce11a0cd01ab78f1141ee3f411f54af6a9b0)

This commit allows for arbitrary markup/components to be passed down as the `Tag` `children` prop.

Since **we can't infer reasonable `aria-label` or `title` attribute from an arbitrary prop**, this commit also **adds an additional `title` prop that overrides the default** generation logic for them. This could be useful in itself regardless of being able to pass arbitrary markup.

Screenshot 2021-12-13 at 12 49 59

(Using a `Stack` here breaks the new segmented markdown (1px difference), so that's definitely something we'd need to figure out 😂 )

#### [[Tag] Removes restriction for passing both an onClick and onRemove prop](https://github.com/Shopify/polaris-react/commit/e07aedd769c9750d1125bf1a185eac40a8008085)

This commit is removes the restriction for passing both an `onClick` prop at the same time of an `onRemove` prop and slightly refactors the component to generate markup for that case by creating a `segmented` variant so we can show both buttons together using a `ButtonGroup`.

Screenshot 2021-12-13 at 12 49 52

I'm obviously hardly sure this is the right markup or approach, so just take this as a way to start the conversation. 

Initially (for our case), we wouldn't need this if we could pass any markup we wanted (like a `Link`). However, by doing that, we need to restyle the link and likely rebuild the tag interaction (hover, down, focus...). By adding this option, we can pass simpler markup to the tag and handle the navigation via JS.

It would probably still be better if the navigation was handled natively using an anchor tag for accessibility and other reasons, but that would mean adding some type of `url` prop to the `Tag` component for something that's currently just a one-off, so I'd say this is a good compromise. If this pattern ever becomes common enough, we could add better support for it here.

### How to 🎩

I've added some samples in the `Tag` page in Storybook

### 🎩 checklist

Haven't 🎩 ed, just looking open the conversation and figure out what direction to take :)

/cc @dleroux

### DIFF
--- a/src/components/Tag/README.md
+++ b/src/components/Tag/README.md
@@ -85,7 +85,58 @@ Use to allow merchants to add attributes to an object.
 Use to allow merchants to add attributes to an object.
 
 ```jsx
-<Tag onClick={() => console.log('Clicked')} onRemove={() => console.log('Removed')}>Wholesale</Tag>
+<Tag
+  onClick={() => console.log('Clicked')}
+  onRemove={() => console.log('Removed')}
+>
+  Wholesale
+</Tag>
+```
+
+### Custom content tag
+
+Use to allow merchants to add attributes to an object.
+
+```jsx
+<>
+  <Tag>
+    <Stack alignment="center" spacing="extraTight" title="Farrah Fawcett">
+      <Icon source={SearchMinor} color="inkLighter" />
+      <TextContainer>Farrah Fawcett</TextContainer>
+    </Stack>
+  </Tag>
+
+  <hr />
+
+  <Tag onClick={() => console.log('Clicked')} title="Farrah Fawcett">
+    <Stack alignment="center" spacing="extraTight">
+      <Icon source={SearchMinor} color="inkLighter" />
+      <TextContainer>Farrah Fawcett</TextContainer>
+    </Stack>
+  </Tag>
+
+  <hr />
+
+  <Tag onRemove={() => console.log('Removed')} title="Farrah Fawcett">
+    <Stack alignment="center" spacing="extraTight">
+      <Icon source={SearchMinor} color="inkLighter" />
+      <TextContainer>Farrah Fawcett</TextContainer>
+    </Stack>
+  </Tag>
+
+  <hr />
+
+  <Tag
+    onClick={() => console.log('Clicked')}
+    onRemove={() => console.log('Removed')}
+    title="Farrah Fawcett"
+  >
+    <Stack alignment="center" spacing="extraTight">
+      <Icon source={SearchMinor} color="inkLighter" />
+      <TextContainer>Farrah Fawcett</TextContainer>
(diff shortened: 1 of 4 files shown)


src/components/Tag/README.md
```diff
@@ -80,6 +80,65 @@ Use to allow merchants to add attributes to an object.
...
+### Custom content tag
+
+Use to allow merchants to add attributes to an object.
+
+```jsx
+<>
+  <Tag>
+    <Stack alignment="center" spacing="extraTight" title="Farrah Fawcett">
+      <Icon source={SearchMinor} color="inkLighter" />
+      <TextContainer>Farrah Fawcett</TextContainer>
```
dleroux, line 105: no need for a TexContainer. It's more meant to space lines of text.

src/components/Tag/README.md
```diff
@@ -80,6 +80,65 @@ Use to allow merchants to add attributes to an object.
...
+
+### Custom content tag
+
+Use to allow merchants to add attributes to an object.
+
+```jsx
+<>
+  <Tag>
+    <Stack alignment="center" spacing="extraTight" title="Farrah Fawcett">
+      <Icon source={SearchMinor} color="inkLighter" />
```
dleroux, line 104: `inkLight` is our old colour system. this should be `base`

dleroux:
By changing the current props we introduce a breaking change. Any way we can simply add the Link to the markup?

dleroux:
What do you think about leaving the onClick and onRemove as is, but adding a url prop?

This way when there is URL we can render the children in an `a` tag that can be styled accordingly.

```
<Tag url="https://www.google.com">
  <Stack alignment="center" spacing="extraTight">
    <Icon source={SearchMinor} color="base" />
    <span>Farrah Fawcett</span>
  </Stack>
</Tag>

<hr />

<Tag onClick={() => {}}>
  <Stack alignment="center" spacing="extraTight">
    <Icon source={SearchMinor} color="base" />
    <span>Farrah Fawcett</span>
  </Stack>
</Tag>
```

jbalsas:
I'm going to close this one to reduce the noise here and avoid anyone unnecessary work. 

Please, see [[WIP] [Tag] Adds support for url prop](https://github.com/Shopify/polaris-react/pull/4837 instead!! )

PR #4837, opened 2021-12-14
[Tag] Adds support for url prop

jbalsas:
This is a spin off of [[WIP] [Tag] Allow arbitrary content and allow onClick and onRemove at the same time](https://github.com/Shopify/polaris-react/pull/4830). I'm opening it early as a Draft so I can get feedback while I start working on tests and final cleanup 🙏 

This PR adds a `url` prop to the `Tag` component that works in the following way:
- You can either pass `onClick` or `url` but not both (this is enforced in the type system)
- You can pass both `url` and `onRemove`
- When `url` is received, an anchor tag is used to render the tag content
- When both `url` and `onRemove` are provided, a `segmented` variant of the tag is created and a `ButtonGroup` is used to render both items

This is a partial approach to solve the problem presented in [[Tag] Allow arbitrary children](https://github.com/Shopify/polaris-react/issues/4819) by focusing on some part of the specific need that triggered that request (navigation in a removable tag).

I've limited this PR to just the interaction, since arbitrary children sounds like a less controversial topic and switched gears based on @dleroux [comment](https://github.com/Shopify/polaris-react/pull/4830#issuecomment-993916486):

> What do you think about leaving the onClick and onRemove as is, but adding a url prop?

Adds the following options that can be explored in the Storybook `URL Tag` section:

```jsx
// Tags with links
<Tag url="https://www.shopify.com/">Wholesale</Tag>
<Tag url="https://www.shopify.com/">Long text that causes ellipsis</Tag>
<Tag url="https://www.shopify.com/" onRemove={() => {}}>Wholesale</Tag>
<Tag url="https://www.shopify.com/" onRemove={() => {}}>Long text that causes ellipsis</Tag>

// Tags with custom content
<Tag>
  <Stack alignment="center" spacing="extraTight" title="Order #123">
    <Icon source={SearchMinor} />
    <span>Order #123</span>
    <Badge status="attention" progress="partiallyComplete">
      In Progress
    </Badge>
  </Stack>
</Tag>
```

https://user-images.githubusercontent.com/905006/146237385-a5ec3609-7365-4b3b-a725-c1bd3326fc14.mov

### How to 🎩

Look into the Storybook `URL Tag` section:

### 🎩 checklist

### DIFF
--- a/src/components/Tag/README.md
+++ b/src/components/Tag/README.md
@@ -123,6 +123,58 @@ function URLTagExample() {
 }
 ```
 
+### Tag with custom content
+
+Use to allow merchants to generate special custom content for a tag.
+
+```jsx
+<Stack>
+  <Tag>
+    <Stack alignment="center" spacing="extraTight" title="Order #123">
+      <Icon source={SearchMinor} />
+      <span>Order #123</span>
+      <Badge status="attention" progress="partiallyComplete">
+        In Progress
+      </Badge>
+    </Stack>
+  </Tag>
+
+  <Tag onClick={() => console.log('Clicked')} title="Order #123">
+    <Stack alignment="center" spacing="extraTight">
+      <Icon source={SearchMinor} />
+      <span>Order #123</span>
+      <Badge status="attention" progress="partiallyComplete">
+        In Progress
+      </Badge>
+    </Stack>
+  </Tag>
+
+  <Tag onRemove={() => console.log('Removed')} title="Order #123">
+    <Stack alignment="center" spacing="extraTight">
+      <Icon source={SearchMinor} />
+      <span>Order #123</span>
+      <Badge status="attention" progress="partiallyComplete">
+        In Progress
+      </Badge>
+    </Stack>
+  </Tag>
+
+  <Tag
+    url="https://www.shopify.com/"
+    onRemove={() => console.log('Removed')}
+    title="Order #123"
+  >
+    <Stack alignment="center" spacing="extraTight">
+      <Icon source={SearchMinor} />
+      <span>Order #123</span>
+      <Badge status="attention" progress="partiallyComplete">
+        In Progress
+      </Badge>
+    </Stack>
+  </Tag>
+</Stack>
+```
+
 <!-- content-for: android -->
 
 ![Tag for Android](/public_images/components/Tag/android/default@2x.png)
--- a/src/components/Tag/README.md
+++ b/src/components/Tag/README.md
@@ -80,99 +80,27 @@ Use to allow merchants to add attributes to an object.
 <Tag onClick={() => console.log('Clicked')}>Wholesale</Tag>
 ```
 
-### Tagss with links
+### Tag with link
 
-Use to allow merchants to navigate to an entity.
+Use to allow merchants to navigate to a resource. For example a customer segment or a smart collection
 
 ```jsx
 function URLTagExample() {
-  return (
-    <Stack spacing="tight">
-      <Tag url="https://www.shopify.com/">Wholesale</Tag>
-      <Tag url="https://www.shopify.com/">
-        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer at
-        ipsum quam. Aliquam fermentum bibendum vestibulum. Vestibulum
-        condimentum luctus metus, sed sagittis magna pellentesque eget. Duis
-        dapibus pretium nisi, et venenatis tortor dignissim ut. Quisque eget
-        lacus ac ex eleifend ultrices. Phasellus facilisis ex sit amet leo
-        elementum condimentum. Ut vel maximus felis. Etiam eget diam eu eros
-        blandit interdum. Sed eu metus sed justo aliquam iaculis ac sit amet ex.
-        Curabitur justo magna, porttitor non pulvinar eu, malesuada at leo. Cras
-        mollis consectetur eros, quis maximus lorem dignissim at. Proin in
-        rhoncus massa. Vivamus lectus nunc, fringilla euismod risus commodo,
-        mattis blandit nulla
-      </Tag>
-      <Tag url="https://www.shopify.com/" onRemove={() => {}}>
-        Wholesale
-      </Tag>
-      <Tag url="https://www.shopify.com/" onRemove={() => {}}>
-        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer at
-        ipsum quam. Aliquam fermentum bibendum vestibulum. Vestibulum
-        condimentum luctus metus, sed sagittis magna pellentesque eget. Duis
-        dapibus pretium nisi, et venenatis tortor dignissim ut. Quisque eget
-        lacus ac ex eleifend ultrices. Phasellus facilisis ex sit amet leo
-        elementum condimentum. Ut vel maximus felis. Etiam eget diam eu eros
-        blandit interdum. Sed eu metus sed justo aliquam iaculis ac sit amet ex.
-        Curabitur justo magna, porttitor non pulvinar eu, malesuada at leo. Cras
-        mollis consectetur eros, quis maximus lorem dignissim at. Proin in
-        rhoncus massa. Vivamus lectus nunc, fringilla euismod risus commodo,
-        mattis blandit nulla
-      </Tag>
-    </Stack>
-  );
+  return <Tag url="/collections/wholesale">Wholesale</Tag>;
 }
 ```
 
 ### Tag with custom content
 
-Use to allow merchants to generate special custom content for a tag.
+Use when a tag needs to be visually distinguished from others, like when it's added automatically.
 
 ```jsx
-<Stack>
-  <Tag>
-    <Stack alignment="center" spacing="extraTight" title="Order #123">
-      <Icon source={SearchMinor} />
-      <span>Order #123</span>
-      <Badge status="attention" progress="partiallyComplete">
-        In Progress
-      </Badge>
-    </Stack>
-  </Tag>
-
-  <Tag onClick={() => console.log('Clicked')} title="Order #123">
-    <Stack alignment="center" spacing="extraTight">
-      <Icon source={SearchMinor} />
-      <span>Order #123</span>
-      <Badge status="attention" progress="partiallyComplete">
-        In Progress
-      </Badge>
-    </Stack>
-  </Tag>
-
-  <Tag onRemove={() => console.log('Removed')} title="Order #123">
-    <Stack alignment="center" spacing="extraTight">
-      <Icon source={SearchMinor} />
-      <span>Order #123</span>
-      <Badge status="attention" progress="partiallyComplete">
-        In Progress
-      </Badge>
-    </Stack>
-  </Tag>
-
-  <Tag
-    url="https://www.shopify.com/"
-    onRemove={() => console.log('Removed')}
-    title="Order #123"
-  >
-    <Stack alignment="center" spacing="extraTight">
-      <Icon source={SearchMinor} />
-      <span>Order #123</span>
-      <Badge status="attention" progress="partiallyComplete">
-        In Progress
-      </Badge>
-    </Stack>
-  </Tag>
-</Stack>
+<Tag url="/customers?tag=VIP">
+  <Stack spacing="extraTight">
+    <Icon source={SearchMinor} />
+    <span>VIP</span>
+  </Stack>
+</Tag>
 ```
 
 <!-- content-for: android -->
--- a/src/components/Tag/README.md
+++ b/src/components/Tag/README.md
@@ -80,6 +80,49 @@ Use to allow merchants to add attributes to an object.
 <Tag onClick={() => console.log('Clicked')}>Wholesale</Tag>
 ```
 
+### Tagss with links
+
+Use to allow merchants to navigate to an entity.
+
+```jsx
+function URLTagExample() {
+  return (
+    <Stack spacing="tight">
+      <Tag url="https://www.shopify.com/">Wholesale</Tag>
+      <Tag url="https://www.shopify.com/">
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer at
+        ipsum quam. Aliquam fermentum bibendum vestibulum. Vestibulum
+        condimentum luctus metus, sed sagittis magna pellentesque eget. Duis
+        dapibus pretium nisi, et venenatis tortor dignissim ut. Quisque eget
+        lacus ac ex eleifend ultrices. Phasellus facilisis ex sit amet leo
+        elementum condimentum. Ut vel maximus felis. Etiam eget diam eu eros
+        blandit interdum. Sed eu metus sed justo aliquam iaculis ac sit amet ex.
+        Curabitur justo magna, porttitor non pulvinar eu, malesuada at leo. Cras
+        mollis consectetur eros, quis maximus lorem dignissim at. Proin in
+        rhoncus massa. Vivamus lectus nunc, fringilla euismod risus commodo,
+        mattis blandit nulla
+      </Tag>
+      <Tag url="https://www.shopify.com/" onRemove={() => {}}>
+        Wholesale
+      </Tag>
+      <Tag url="https://www.shopify.com/" onRemove={() => {}}>
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer at
+        ipsum quam. Aliquam fermentum bibendum vestibulum. Vestibulum
+        condimentum luctus metus, sed sagittis magna pellentesque eget. Duis
+        dapibus pretium nisi, et venenatis tortor dignissim ut. Quisque eget
+        lacus ac ex eleifend ultrices. Phasellus facilisis ex sit amet leo
+        elementum condimentum. Ut vel maximus felis. Etiam eget diam eu eros
+        blandit interdum. Sed eu metus sed justo aliquam iaculis ac sit amet ex.
+        Curabitur justo magna, porttitor non pulvinar eu, malesuada at leo. Cras
+        mollis consectetur eros, quis maximus lorem dignissim at. Proin in
+        rhoncus massa. Vivamus lectus nunc, fringilla euismod risus commodo,
+        mattis blandit nulla
+      </Tag>
+    </Stack>
+  );
+}
+```
+
 <!-- content-for: android -->
 
 ![Tag for Android](/public_images/components/Tag/android/default@2x.png)
--- a/src/components/Tag/README.md
+++ b/src/components/Tag/README.md
@@ -95,10 +95,10 @@ function URLTagExample() {
 Use when a tag needs to be visually distinguished from others, like when it's added automatically.
 
 ```jsx
-<Tag url="/customers?tag=VIP">
+<Tag url="/collections/wholesale">
   <Stack spacing="extraTight">
-    <Icon source={SearchMinor} />
-    <span>VIP</span>
+    <Icon source={WandMinor} />
+    <span>Wholesale</span>
   </Stack>
 </Tag>
 ```
--- a/src/components/Tag/Tag.scss
+++ b/src/components/Tag/Tag.scss
@@ -127,10 +127,12 @@ $icon-size: rem(16px);
 
   &:focus:not(:active) {
     @include focus-ring($style: 'focused');
+    text-decoration: underline;
   }
 
   &:hover {
     background: var(--p-surface-neutral-hovered);
+    text-decoration: underline;
   }
 
   &.segmented::after {
--- a/src/components/Tag/Tag.scss
+++ b/src/components/Tag/Tag.scss
@@ -54,7 +54,6 @@ $icon-size: rem(16px);
   }
 
   &.linkable {
-    align-items: stretch;
     padding: 0;
   }
 
@@ -118,9 +117,7 @@ $icon-size: rem(16px);
   padding: spacing(extra-tight) spacing(tight);
 
   .LinkText {
-    overflow: hidden;
-    white-space: nowrap;
-    text-overflow: ellipsis;
+    @include truncate;
   }
 
   @include focus-ring;
@@ -135,9 +132,14 @@ $icon-size: rem(16px);
     text-decoration: underline;
   }
 
-  &.segmented::after {
-    margin-right: spacing(extra-tight);
+  &.segmented {
     border-top-right-radius: 0;
     border-bottom-right-radius: 0;
+
+    &::after {
+      margin-right: spacing(extra-tight);
+      border-top-right-radius: 0;
+      border-bottom-right-radius: 0;
+    }
   }
 }
--- a/src/components/Tag/Tag.scss
+++ b/src/components/Tag/Tag.scss
@@ -20,10 +20,6 @@ $icon-size: rem(16px);
     color: var(--p-text-disabled);
   }
 
-  &.removable {
-    padding-right: 0;
-  }
-
   &.clickable {
     @include unstyled-button;
     cursor: pointer;
@@ -52,6 +48,15 @@ $icon-size: rem(16px);
     }
   }
 
+  &.removable {
+    padding-right: 0;
+  }
+
+  &.linkable {
+    align-items: stretch;
+    padding: 0;
+  }
+
   @include high-contrast-outline;
 }
 
@@ -71,14 +76,11 @@ $icon-size: rem(16px);
   width: $height;
   margin-left: spacing(extra-tight);
   border-radius: 0 border-radius() border-radius() 0;
+  background-color: var(--p-surface-neutral);
 
   &:hover {
     background: var(--p-surface-neutral-hovered);
     @include high-contrast-outline;
-  }
-
-  &:focus {
-    background-color: transparent;
   }
 
   @include focus-ring;
@@ -95,5 +97,45 @@ $icon-size: rem(16px);
     @include recolor-icon(var(--p-icon-disabled));
     cursor: default;
     pointer-events: none;
+  }
+
+  &.segmented {
+    margin-left: -1 * spacing(extra-tight);
+
+    &::after {
+      border-top-left-radius: 0;
+      border-bottom-left-radius: 0;
+    }
+  }
+}
+
+.Link {
+  display: inline-grid;
+  color: var(--p-text);
+  outline: none;
+  border-radius: var(--p-border-radius-base);
+  text-decoration: none;
+  padding: spacing(extra-tight) spacing(tight);
+
+  .LinkText {
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: ellipsis;
+  }
+
+  @include focus-ring;
+
+  &:focus:not(:active) {
+    @include focus-ring($style: 'focused');
+  }
+
+  &:hover {
+    background: var(--p-surface-neutral-hovered);
+  }
+
+  &.segmented::after {
+    margin-right: spacing(extra-tight);
+    border-top-right-radius: 0;
+    border-bottom-right-radius: 0;
   }
 }
--- a/src/components/Tag/Tag.scss
+++ b/src/components/Tag/Tag.scss
@@ -45,6 +45,7 @@ $icon-size: rem(16px);
       background: var(--p-surface-neutral-disabled);
       cursor: default;
       pointer-events: none;
+      color: var(--p-text-disabled);
     }
   }
 
@@ -76,7 +77,6 @@ $icon-size: rem(16px);
   width: $height;
   margin-left: spacing(extra-tight);
   border-radius: 0 border-radius() border-radius() 0;
-  background-color: var(--p-surface-neutral);
 
   &:hover {
     background: var(--p-surface-neutral-hovered);
--- a/src/components/Tag/Tag.tsx
+++ b/src/components/Tag/Tag.tsx
@@ -45,7 +45,7 @@ export function Tag({
     disabled && styles.disabled,
     onClick && styles.clickable,
     onRemove && styles.removable,
-    url && styles.linkable,
+    url && !disabled && styles.linkable,
     segmented && styles.segmented,
   );
 
@@ -85,20 +85,21 @@ export function Tag({
     </button>
   ) : null;
 
-  const tagContent = url ? (
-    <a
-      className={classNames(styles.Link, segmented && styles.segmented)}
-      href={url}
-    >
-      <span title={tagTitle} className={styles.LinkText}>
+  const tagContent =
+    url && !disabled ? (
+      <a
+        className={classNames(styles.Link, segmented && styles.segmented)}
+        href={url}
+      >
+        <span title={tagTitle} className={styles.LinkText}>
+          {children}
+        </span>
+      </a>
+    ) : (
+      <span title={tagTitle} className={styles.TagText}>
         {children}
       </span>
-    </a>
-  ) : (
-    <span title={tagTitle} className={styles.TagText}>
-      {children}
-    </span>
-  );
+    );
 
   return (
     <span className={className}>
--- a/src/components/Tag/Tag.tsx
+++ b/src/components/Tag/Tag.tsx
@@ -13,27 +13,51 @@ export interface NonMutuallyExclusiveProps {
   children?: string;
   /** Disables the tag  */
   disabled?: boolean;
-  /** Callback when tag is clicked or keypressed. Renders without remove button when set. */
+  /** Callback when tag is clicked or keypressed. Renders without remove button or url when set. */
   onClick?(): void;
   /** Callback when remove button is clicked or keypressed. */
   onRemove?(): void;
+  /** Url to navigate to when tag is clicked or keypressed. */
+  url?: string;
 }
 
 export type TagProps = NonMutuallyExclusiveProps &
   (
-    | {onClick?(): void; onRemove?: undefined}
-    | {onClick?: undefined; onRemove?(): void}
+    | {onClick?(): void; onRemove?: undefined; url?: undefined}
+    | {onClick?: undefined; onRemove?(): void; url?: string}
   );
 
-export function Tag({children, disabled = false, onClick, onRemove}: TagProps) {
+export function Tag({
+  children,
+  disabled = false,
+  onClick,
+  onRemove,
+  url,
+}: TagProps) {
   const i18n = useI18n();
 
+  const segmented = onRemove && url;
   const className = classNames(
     styles.Tag,
     disabled && styles.disabled,
     onClick && styles.clickable,
     onRemove && styles.removable,
+    url && styles.linkable,
+    segmented && styles.segmented,
   );
+
+  if (onClick) {
+    return (
+      <button
+        type="button"
+        disabled={disabled}
+        className={className}
+        onClick={onClick}
+      >
+        {children}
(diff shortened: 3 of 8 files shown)


src/components/Tag/Tag.scss
```diff
@@ -52,6 +48,20 @@ $icon-size: rem(16px);
     }
   }
 
+  &.removable {
+    padding-right: 0;
+  }
+
+  &.segmented {
+    > * {
+      max-width: 100%;
```
jbalsas: This is so ellipsis will work in the segmented case, otherwise I think the inline element has no bound size and overflow won't kick in

src/components/Tag/Tag.scss
```diff
@@ -52,6 +48,20 @@ $icon-size: rem(16px);
...
+
+      > * {
+        max-width: 97%;
```
jbalsas: This is so that the content leaves some room for remove button when the it requires ellipsis... maybe this can be done in a better way?

src/components/Tag/Tag.tsx
```diff
@@ -17,24 +19,46 @@ export interface NonMutuallyExclusiveProps {
...
   const className = classNames(
     styles.Tag,
     disabled && styles.disabled,
-    onClick && styles.clickable,
+    (onClick || url) && styles.clickable,
     onRemove && styles.removable,
+    onRemove && url && styles.segmented,
   );
 
+  if (onClick) {
```
jbalsas, line 52: This is all basically some refactoring that should keep the old options the same and only change something when `url` is set.

Since `onClick` is the most restrictive prop and using a ternary for this case was complex, I decided to return early here, but this could be moved around in any way in case there are some specific conventions I'm not aware of.

jbalsas:
I've added an additional commit slightly adjusting the behaviour as can be seen in the screencast below

https://user-images.githubusercontent.com/905006/146170656-0a5a098a-b2b3-4b86-8ec5-372c4beed0df.mov

This includes:
- Simpler approach to position and style the focus state on the link
- Usage of `Stack` instead of `ButtonGroup` since we're not rendering 2 buttons
- Better hover state in the segmented case, clearly seeing both

The **main caveat** in this scenario is that the remove button extends the width of the tag rather than remaining inside.

jbalsas:
Hey @chloerice, @AndrewMusgrave, I think this is ready for review. I would appreciate if you could take a look and let me know what you think!

CI is currently failing and for the life of me I don't know what to make of the error and what should be the proper sorting order 😂 

> /home/runner/work/polaris-react/polaris-react/src/components/Tag/tests/Tag.test.tsx
  3:1  error  There should be at least one empty line between import groups  import/order
  4:1  error  There should be no empty line within import group              import/order

Other than that, this PR finally adds:
- Support for `url` prop
- Support for `React.Element` as `children` of the tag

https://user-images.githubusercontent.com/905006/146233653-3439cd0e-b4cb-4c53-bf3b-0f9423e2796e.mov

letsdothis

eliashussary:
Having arbitrary children is something beneficial for us, right now we have to force it;

```tsx
  const content = ((
    <Stack>
      {icon && <div className={styles.IconContainer}>{icon}</div>}
      <div>{children}</div>
    </Stack>
  ) as unknown) as string;

<Tag>
  {content}
</Tag>
```

src/components/Tag/README.md
```diff
@@ -80,6 +80,101 @@ Use to allow merchants to add attributes to an object.
 <Tag onClick={() => console.log('Clicked')}>Wholesale</Tag>
 ```
 
+### Tagss with links
```
chloerice:
```suggestion
### Tag with link
```

src/components/Tag/README.md
```diff
@@ -80,6 +80,101 @@ Use to allow merchants to add attributes to an object.
...
+
+Use to allow merchants to navigate to an entity.
```
chloerice:
```suggestion
Use to allow merchants to navigate to a resource, like a customer segment or a smart collection.
```

src/components/Tag/README.md
```diff
@@ -80,6 +80,101 @@ Use to allow merchants to add attributes to an object.
...
+        dapibus pretium nisi, et venenatis tortor dignissim ut. Quisque eget
+        lacus ac ex eleifend ultrices. Phasellus facilisis ex sit amet leo
+        elementum condimentum. Ut vel maximus felis. Etiam eget diam eu eros
+        blandit interdum. Sed eu metus sed justo aliquam iaculis ac sit amet ex.
+        Curabitur justo magna, porttitor non pulvinar eu, malesuada at leo. Cras
+        mollis consectetur eros, quis maximus lorem dignissim at. Proin in
+        rhoncus massa. Vivamus lectus nunc, fringilla euismod risus commodo,
+        mattis blandit nulla
+      </Tag>
+    </Stack>
```
chloerice:
```suggestion
    <Tag url="/collections/wholesale">Wholesale</Tag>;
```

src/components/Tag/README.md
```diff
@@ -80,6 +80,101 @@ Use to allow merchants to add attributes to an object.
...
+  );
+}
+```
+
+### Tag with custom content
+
+Use to allow merchants to generate special custom content for a tag.
```
chloerice:
```suggestion
Use when a tag needs to be visually distinguished from others, like when it's added automatically.
```

src/components/Tag/README.md
```diff
@@ -80,6 +80,101 @@ Use to allow merchants to add attributes to an object.
...
+  >
+    <Stack alignment="center" spacing="extraTight">
+      <Icon source={SearchMinor} />
+      <span>Order #123</span>
+      <Badge status="attention" progress="partiallyComplete">
+        In Progress
+      </Badge>
+    </Stack>
+  </Tag>
+</Stack>
```
chloerice: We try to keep the examples simple and reference real use cases when possible, so when `WandMinor` ships in `@shopify/polaris-icons` we should update this to reflect the real admin use case. We def want to test worst case/edge case scenarios, but we can do this in the playground 💯  

```suggestion
<Tag url="/customers?tag=VIP">
  <Stack spacing="extraTight">
    <Icon source={CustomersMinor} />
    <span>VIP</span>
  </Stack>
</Tag>
```

src/components/Tag/Tag.tsx
```diff
@@ -10,40 +10,73 @@ import styles from './Tag.scss';
...
   disabled?: boolean;
-  /** Callback when tag is clicked or keypressed. Renders without remove button when set. */
+  /** Callback when tag is clicked or keypressed. Renders without remove button or url when set. */
   onClick?(): void;
   /** Callback when remove button is clicked or keypressed. */
   onRemove?(): void;
+  /** A string to use when tag has more than textual content */
+  title?: string;
+  /** Url to navigate to when tag is clicked or keypressed. */
+  url?: string;
```
chloerice: Let's give the `title` prop a descriptive name of `accessibilityLabel` (this also aligns with with naming in other components, like Button etc).

```suggestion
  children?: string | React.ReactElement;
  /** Disables the tag  */
  disabled?: boolean;
    /** Descriptive text to provide context for screen reader users. */
  accessibilityLabel?: string;
  /** Url to navigate to when tag is clicked or keypressed. */
  url?: string;
  /** Callback when tag is clicked or keypressed. Renders without remove button or url when set. */
  onClick?(): void;
  /** Callback when remove button is clicked or keypressed. */
  onRemove?(): void;
```

src/components/Tag/Tag.tsx
```diff
@@ -10,40 +10,73 @@ import styles from './Tag.scss';
...
+        disabled={disabled}
+        className={className}
+        onClick={onClick}
+      >
+        {children}
+      </button>
+    );
+  }
+
+  let tagTitle = title;
```
chloerice:
```suggestion
  let tagTitle = accessibilityLabel;
```

src/components/Tag/Tag.tsx
```diff
@@ -10,40 +10,73 @@ import styles from './Tag.scss';
...
+    | {onClick?: undefined; onRemove?(): void; url?: string}
   );
 
-export function Tag({children, disabled = false, onClick, onRemove}: TagProps) {
+export function Tag({
+  children,
+  disabled = false,
+  onClick,
+  onRemove,
+  title = '',
```
chloerice:
```suggestion
  accessibilityLabel = '',
```

src/components/Tag/tests/Tag.test.tsx
```diff
@@ -35,4 +37,78 @@ describe('<Tag />', () => {
...
+      const tag = mountWithApp(
+        <Tag>
+          <Icon source={SearchMinor} />
+        </Tag>,
+      );
+      expect(tag).toContainReactComponent(Icon, {source: SearchMinor});
+    });
+  });
+
+  describe('title', () => {
```
chloerice:
```suggestion
  describe('accessibilityLabel', () => {
```

src/components/Tag/tests/Tag.test.tsx
```diff
@@ -35,4 +37,78 @@ describe('<Tag />', () => {
...
+    it('uses children to render title if no title prop is provided', () => {
+      const tag = mountWithApp(<Tag>children</Tag>);
+      expect(tag).toContainReactComponent('span', {
+        children: 'children',
+        title: 'children',
+      });
+    });
+
+    it('uses the title prop to render title when a string is received as children', () => {
```
chloerice:
```suggestion
    it('uses the accessibilityLabel prop to render title when a string is received as children', () => {
```

src/components/Tag/tests/Tag.test.tsx
```diff
@@ -35,4 +37,78 @@ describe('<Tag />', () => {
...
+      const tag = mountWithApp(<Tag title="customTitle">children</Tag>);
+      expect(tag).toContainReactComponent('span', {
+        children: 'children',
+        title: 'customTitle',
+      });
+    });
+
+    it('uses the title prop to render title when a component is received as children and url prop is passed', () => {
```
chloerice:
```suggestion
    it('uses the accessibilityLabel prop to render title when a component is received as children and url prop is passed', () => {
```

chloerice (approved):
This works great @jbalsas!! 

The only problem I'm seeing in my playground is with the edge case of a `Tag` with really long content, the margin of the `Stack` causes  the `Tag` to overflow. But this also happens even on the `main` branch so it's not related to your changes.

<details>
<summary>Click to view collapsed screenshot 👀 </summary>

Screen Shot 2021-12-15 at 2 25 44 PM

</details>

<details>
<summary>Click to view collapsed playground code 👀 </summary>

```jsx
import React from 'react';
import {CustomersMinor, OrderStatusMinor} from '@shopify/polaris-icons';

import {Card, Stack, Badge, Page, Tag, Icon, Layout} from '../src';

  return (
    <Page title="Playground">
      <Layout>
        <Layout.Section secondary>
          <Card sectioned>
            <Stack>
              <Tag url="/customers?tag=VIP" onRemove={() => {}}>
                <Stack wrap={false} spacing="extraTight">
                  <Icon source={CustomersMinor} />
                  <span>VIP</span>
                </Stack>
              </Tag>
              <Tag
                url="/customers?tag=VIP"
                onRemove={() => {}}
                accessibilityLabel="VIP"
              >
                <Stack alignment="center" wrap={false} spacing="extraTight">
                  <Icon source={OrderStatusMinor} />
                  <span>3PL</span>
                  <span style={{position: 'relative', top: '-1px'}}>
                    <Badge status="attention" progress="partiallyComplete">
                      Partially fulfilled
                    </Badge>
                  </span>
                </Stack>
              </Tag>
              <Tag>
                Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer
                at ipsum quam. Aliquam fermentum bibendum vestibulum. Vestibulum
                condimentum luctus metus, sed sagittis magna pellentesque eget.
                Duis dapibus pretium nisi, et venenatis tortor dignissim ut.
                Quisque eget lacus ac ex eleifend ultrices. Phasellus facilisis
                ex sit amet leo elementum condimentum. Ut vel maximus felis.
                Etiam eget diam eu eros blandit interdum. Sed eu metus sed justo
                aliquam iaculis ac sit amet ex. Curabitur justo magna, porttitor
                non pulvinar eu, malesuada at leo. Cras mollis consectetur eros,
                quis maximus lorem dignissim at. Proin in rhoncus massa. Vivamus
                lectus nunc, fringilla euismod risus commodo, mattis blandit
                nulla.
              </Tag>
              <Tag url="/customers?tag=VIP" onRemove={() => {}}>
                Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer
                at ipsum quam. Aliquam fermentum bibendum vestibulum. Vestibulum
                condimentum luctus metus, sed sagittis magna pellentesque eget.
                Duis dapibus pretium nisi, et venenatis tortor dignissim ut.
                Quisque eget lacus ac ex eleifend ultrices. Phasellus facilisis
                ex sit amet leo elementum condimentum. Ut vel maximus felis.
                Etiam eget diam eu eros blandit interdum. Sed eu metus sed justo
                aliquam iaculis ac sit amet ex. Curabitur justo magna, porttitor
                non pulvinar eu, malesuada at leo. Cras mollis consectetur eros,
                quis maximus lorem dignissim at. Proin in rhoncus massa. Vivamus
                lectus nunc, fringilla euismod risus commodo, mattis blandit
                nulla.
              </Tag>
              <Tag url="/customers?tag=VIP">
                Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer
                at ipsum quam. Aliquam fermentum bibendum vestibulum. Vestibulum
                condimentum luctus metus, sed sagittis magna pellentesque eget.
                Duis dapibus pretium nisi, et venenatis tortor dignissim ut.
                Quisque eget lacus ac ex eleifend ultrices. Phasellus facilisis
                ex sit amet leo elementum condimentum. Ut vel maximus felis.
                Etiam eget diam eu eros blandit interdum. Sed eu metus sed justo
                aliquam iaculis ac sit amet ex. Curabitur justo magna, porttitor
                non pulvinar eu, malesuada at leo. Cras mollis consectetur eros,
                quis maximus lorem dignissim at. Proin in rhoncus massa. Vivamus
                lectus nunc, fringilla euismod risus commodo, mattis blandit
                nulla.
              </Tag>{' '}
            </Stack>
          </Card>
        </Layout.Section>
      </Layout>
    </Page>
  );
}
```

</details>

src/components/Tag/Tag.scss
```diff
@@ -96,4 +98,44 @@ $icon-size: rem(16px);
...
+    }
+  }
+}
+
+.Link {
+  display: inline-grid;
+  color: var(--p-text);
+  outline: none;
+  border-radius: var(--p-border-radius-base);
+  text-decoration: none;
```
AndrewMusgrave, line 116: Links normally have underlines. Right now the button tag/link tag seem almost visually identical. Should we add underlines to differentiate them?

src/components/Tag/Tag.scss
```diff
@@ -96,4 +98,44 @@ $icon-size: rem(16px);
...
+    text-overflow: ellipsis;
+  }
+
+  @include focus-ring;
+
+  &:focus:not(:active) {
+    @include focus-ring($style: 'focused');
+  }
+
+  &:hover {
```
AndrewMusgrave: Do we need a disabled / active state? And can we have Button/Link share the same styles so it's easier to update in the future?

src/components/Tag/README.md
AndrewMusgrave:
```suggestion
Use to allow merchants to navigate to a resource. For example a customer segment or a smart collection
```

Building on Chloe's suggestion a little

src/components/Tag/Tag.tsx
```diff
@@ -10,40 +10,73 @@ import styles from './Tag.scss';
 
 export interface NonMutuallyExclusiveProps {
   /** Content to display in the tag */
-  children?: string;
+  children?: string | React.ReactElement;
```
AndrewMusgrave:
```suggestion
  children?: React.ReactNode;
```

`ReactNode` is the type we generally go for to allow custom markup in Polaris components. And string will also be included in `ReactNode`!

String is included in React

src/components/Tag/Tag.tsx
AndrewMusgrave: Should we leave out a default value? Right now we can render a title without an content, but if we left it as undefined react would omit it

AndrewMusgrave (commented):
Looking good!

src/components/Tag/Tag.scss
jbalsas: My initial approach had indeed underlines, but @ouellettejordan asked to remove them. I agree that it's not obvious that those actions are going to be navigations... 

Maybe one compromise would be to add the [ExternalMinor](https://polaris-icons.shopify.com/?icon=ExternalSmallMinor&q=exter) icon at the end to signal navigation, but this is usually done for external navigations and nothing guarantees that's the case here.

How big of a deal this is @AndrewMusgrave? A quick search for `text-decoration: none` in `web` returns **`80 results - 68 files`** so I think that cat is out of the bag... 🐱 

That being said, there's no occurrence of `text-decoration: none` in here, so I can see how we might want to think this through and see if we should expose this differently.

src/components/Tag/Tag.scss
jbalsas: From what I saw, each case has some differences with the other. 

There's some common code that at some point I tried to hide away in a `interaction()` mixin. I can give that a try again. My understanding of what everything is doing is pretty limited, so I can't promise I'll get it right though! 😂 

dontknowwhatimdoing

(Yes, regardless of what dhh says 😂 )

src/components/Tag/Tag.tsx
jbalsas: Sweet!!!

src/components/Tag/Tag.tsx
jbalsas: Hey @AndrewMusgrave, I tried this (turns out already tried it before) and the reason I left it as empty was because down there we have

```
const ariaLabel = i18n.translate('Polaris.Tag.ariaLabel', {
  children: tagTitle,
});
```

So, `children` expects `string | number`. If `title` can be `undefined`, then we'd need additional logic for that as well, which might be fine... what do you think?

src/components/Tag/Tag.scss
jbalsas: > Do we need a disabled / active state?

I think we do, nice catch!!

src/components/Tag/Tag.scss
chloerice: When to underline has always been a contentious thing 😅 But sometimes what we want/like and what merchants need in order to have clear expectations is not the same 🥲 I'd recommend adding `text-decoration` back when hovered or focused (that's the compromise we landed on for the plain `Button` variant).

jbalsas:
Hey @chloerice, @AndrewMusgrave, thanks so much for your time and patience going through this with me! I've pushed some additional commits now:
- [Apply suggestions from code review](https://github.com/Shopify/polaris-react/pull/4837/commits/c93b61a507a3afeb6037d505844179edb6497d6d) - All your suggestions, directly from GH! 🎉 
- [Polaris icons v4.11.0](https://github.com/Shopify/polaris-react/pull/4837/commits/b42104e21f3f3c599859ed9dadf9bad048672706) - Update for the freshly released version of the icons! (I can do this in a different PR if we want!)
- [[Tag] Update custom content example with real example](https://github.com/Shopify/polaris-react/pull/4837/commits/64af85979a366d6bb22ab45087f5594f945a7848) - Testing with the real example, as suggested! 💯 
- [[Tag] Adds underline decoration to links upon hover/focus interactions](https://github.com/Shopify/polaris-react/pull/4837/commits/02dedb86b9f9a527162213ab960711034f3bb725) - Nothing to argue here! 👍 
- [[Tag] Improve disabled states for new url and segmented options](https://github.com/Shopify/polaris-react/pull/4837/commits/aebbf5d3170e56035fabb850abe99dfa2652ad8f) - Update to account for disabled states. Basically, the link case degenerates into the old ones when disabled (no anchor is rendered). 

<details><summary>Click to view collapsed screenshot 👀</summary>

Screenshot 2021-12-15 at 22 45 39

</details>

<details><summary>Click to view collapsed playground code 👀</summary>

```jsx
import React from 'react';

import {Card, Stack, Page, Tag, Layout} from '../src';

  return (
    <Page title="Playground">
      <Layout>
        <Layout.Section secondary>
          <Card sectioned>
            <Stack>
              <Tag>Normal</Tag>
              <Tag disabled>Disabled</Tag>
              <Tag onClick={() => {}}>onClick</Tag>
              <Tag disabled onClick={() => {}}>
                disabled onClick
              </Tag>
              <Tag onRemove={() => {}}>onRemove</Tag>
              <Tag disabled onRemove={() => {}}>
                disabled onRemove
              </Tag>
              <Tag url="#">url</Tag>
              <Tag disabled url="#">
                disabled url
              </Tag>
              <Tag url="#" onRemove={() => {}}>
                url onRemove
              </Tag>
              <Tag disabled url="#" onRemove={() => {}}>
                disabled url onRemove
              </Tag>{' '}
            </Stack>
          </Card>
        </Layout.Section>
      </Layout>
    </Page>
  );
}
```

</details>

Comments from @AndrewMusgrave that I have not yet tackled are:
- [And can we have Button/Link share the same styles so it's easier to update in the future?](https://github.com/Shopify/polaris-react/pull/4837#discussion_r769995506)
- [Should we leave out a default value? Right now we can render a title without an content, but if we left it as undefined react would omit it](https://github.com/Shopify/polaris-react/pull/4837#discussion_r769999958)

Would it be okay to work on those after this lands, or do we want to get them in together? (I feel this is starting to grow a bit too much, but I'm fine either way)

chloerice:
>And can we have Button/Link share the same styles so it's easier to update in the future?
Should we leave out a default value? Right now we can render a title without an content, but if we left it as undefined react would omit it
Would it be okay to work on those after this lands, or do we want to get them in together? (I feel this is starting to grow a bit too much, but I'm fine either way)

I think it's fine to iterate on this with follow up PRs. Keep in mind that the changes won't be available in `web` even after this merges until there's a new release cut and an upgrade PR in `web` is shipped.

jbalsas:
> I think it's fine to iterate on this with follow up PRs. Keep in mind that the changes won't be available in web even after this merges until there's a new release cut and an upgrade PR in web is shipped.

Yeah, that's fine, I'll just work locally to prepare the PR over there, then someone will be able to take over once this ships!

I'll try to enlist someone else to give this an additional review, but if I can't...

shipit

src/components/Tag/Tag.scss
dleroux: > underline has always been a contentious thing 😅

😅

I do like the visual cue that the text is actually a link on focus of hover.

dleroux (approved):

src/components/Tag/Tag.scss
```diff
@@ -49,9 +45,19 @@ $icon-size: rem(16px);
...
+      color: var(--p-text-disabled);
     }
   }
 
+  &.removable {
+    padding-right: 0;
+  }
+
+  &.linkable {
+    align-items: stretch;
```
kyledurand: This jumps out as a bit of an odd property to me. Can you share some info around why you need this?

src/components/Tag/Tag.scss
```diff
@@ -96,4 +98,46 @@ $icon-size: rem(16px);
...
+
+    &::after {
+      border-top-left-radius: 0;
+      border-bottom-left-radius: 0;
+    }
+  }
+}
+
+.Link {
+  display: inline-grid;
```
kyledurand, line 112: This seems like it should be a flex layout to me. Curious to know more about this choice as well though

src/components/Tag/Tag.scss
```diff
@@ -96,4 +98,46 @@ $icon-size: rem(16px);
...
+  color: var(--p-text);
+  outline: none;
+  border-radius: var(--p-border-radius-base);
+  text-decoration: none;
+  padding: spacing(extra-tight) spacing(tight);
+
+  .LinkText {
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: ellipsis;
```
kyledurand: You can use:
```suggestion
    @include truncate;
```

I'm not sure this is actually working as you intended though:
image

src/components/Tag/Tag.tsx
```diff
@@ -10,40 +10,73 @@ import styles from './Tag.scss';
...
+  children,
+  disabled = false,
+  onClick,
+  onRemove,
+  accessibilityLabel = '',
+  url,
+}: TagProps) {
   const i18n = useI18n();
 
+  const segmented = onRemove && url;
```
kyledurand, line 42: This seems like it could cause bugs in web. Can't we have segmented tags without urls?

src/components/Tag/tests/Tag.test.tsx
```diff
@@ -35,4 +37,85 @@ describe('<Tag />', () => {
...
+
+    it('allows url and onRemove props at the same time', () => {
+      const spy = jest.fn();
+      const tag = mountWithApp(<Tag onRemove={spy} url="#" />);
+      tag.find('button')!.domNode!.click();
+      expect(spy).toHaveBeenCalled();
+      expect(tag).toContainReactComponent('a', {href: '#'});
+    });
+
+    it('renders plain text when the tag is disabled', () => {
```
kyledurand, line 55: Should we also have a test that asserts a span is rendered instead of anchor when url and disabled are passed in?

kyledurand (commented):
This is looking good just a few observations and questions but we should be able to get this merged today

src/components/Tag/Tag.tsx
jbalsas: Unless I missed something (which could very easily happen), currently we cannot. The component [prevents us from passing both `onClick` and `onRemove`](https://github.com/Shopify/polaris-react/blob/main/src/components/Tag/Tag.tsx#L22-L26). I'm not aware of the reasons for such decision and I couldn't find any literature or reference to it.

My initial attempt was to allow both `onClick` and `onRemove` at the same time to create the `segmented` option as you can see in [[WIP] [Tag] Allow arbitrary content and allow onClick and onRemove at the same time](https://github.com/Shopify/polaris-react/pull/4830)

To simplify the combinations I kept `onClick` separate since it could be weird to have both an `onClick` and a `url` (that should navigate on click) at the same time. Both scenarios (`onClick+onRemove` or `onClick+url`) have valid use cases, so we could work on allowing them if we want.

src/components/Tag/Tag.scss
jbalsas: That's a great observation!! 

I went back to the playground and I don't think that's making a difference at all! 😱 

I know at some point the `a` elements were shorter in height than the tag so hovering created a weird effect where only the inner span was being highlighted. Using stretch allowed for the whole area to be covered without having to resort to layout tricks (padding/margin/position) that made it also hard to get the focus ring right. 

At that time I was using both `UnstyledLink` and `Stack`, which I ended up removing, so maybe that played a particular role in creating that issue!

Seeing as this is no longer needed, I'll remove it!! 🎩

src/components/Tag/Tag.scss
jbalsas: >  @include truncate;

Nice, will do that!

> I'm not sure this is actually working as you intended though:

Interesting, what browser would that be? Do you have the code for the playground?

I've been testing on Chrome, Firefox and Safari and that seems to be working fine in all of them

<details><summary>This is the Playground I'm using as provided by @chloerice, maybe we need to add something else?</summary>

```jsx
import React from 'react';
import {CustomersMinor, OrderStatusMinor} from '@shopify/polaris-icons';

import {Card, Stack, Badge, Page, Tag, Icon, Layout} from '../src';

  return (
    <Page title="Playground">
      <Layout>
        <Layout.Section secondary>
          <Card sectioned>
            <Stack>
              <Tag url="/customers?tag=VIP" onRemove={() => {}}>
                <Stack wrap={false} spacing="extraTight">
                  <Icon source={CustomersMinor} />
                  <span>VIP</span>
                </Stack>
              </Tag>
              <Tag
                url="/customers?tag=VIP"
                onRemove={() => {}}
                accessibilityLabel="VIP"
              >
                <Stack alignment="center" wrap={false} spacing="extraTight">
                  <Icon source={OrderStatusMinor} />
                  <span>3PL</span>
                  <span style={{position: 'relative', top: '-1px'}}>
                    <Badge status="attention" progress="partiallyComplete">
                      Partially fulfilled
                    </Badge>
                  </span>
                </Stack>
              </Tag>
              <Tag>
                Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer
                at ipsum quam. Aliquam fermentum bibendum vestibulum. Vestibulum
                condimentum luctus metus, sed sagittis magna pellentesque eget.
                Duis dapibus pretium nisi, et venenatis tortor dignissim ut.
                Quisque eget lacus ac ex eleifend ultrices. Phasellus facilisis
                ex sit amet leo elementum condimentum. Ut vel maximus felis.
                Etiam eget diam eu eros blandit interdum. Sed eu metus sed justo
                aliquam iaculis ac sit amet ex. Curabitur justo magna, porttitor
                non pulvinar eu, malesuada at leo. Cras mollis consectetur eros,
                quis maximus lorem dignissim at. Proin in rhoncus massa. Vivamus
                lectus nunc, fringilla euismod risus commodo, mattis blandit
                nulla.
              </Tag>
              <Tag url="/customers?tag=VIP" onRemove={() => {}}>
                Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer
                at ipsum quam. Aliquam fermentum bibendum vestibulum. Vestibulum
                condimentum luctus metus, sed sagittis magna pellentesque eget.
                Duis dapibus pretium nisi, et venenatis tortor dignissim ut.
                Quisque eget lacus ac ex eleifend ultrices. Phasellus facilisis
                ex sit amet leo elementum condimentum. Ut vel maximus felis.
                Etiam eget diam eu eros blandit interdum. Sed eu metus sed justo
                aliquam iaculis ac sit amet ex. Curabitur justo magna, porttitor
                non pulvinar eu, malesuada at leo. Cras mollis consectetur eros,
                quis maximus lorem dignissim at. Proin in rhoncus massa. Vivamus
                lectus nunc, fringilla euismod risus commodo, mattis blandit
                nulla.
              </Tag>
              <Tag url="/customers?tag=VIP">
                Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer
                at ipsum quam. Aliquam fermentum bibendum vestibulum. Vestibulum
                condimentum luctus metus, sed sagittis magna pellentesque eget.
                Duis dapibus pretium nisi, et venenatis tortor dignissim ut.
                Quisque eget lacus ac ex eleifend ultrices. Phasellus facilisis
                ex sit amet leo elementum condimentum. Ut vel maximus felis.
                Etiam eget diam eu eros blandit interdum. Sed eu metus sed justo
                aliquam iaculis ac sit amet ex. Curabitur justo magna, porttitor
                non pulvinar eu, malesuada at leo. Cras mollis consectetur eros,
                quis maximus lorem dignissim at. Proin in rhoncus massa. Vivamus
                lectus nunc, fringilla euismod risus commodo, mattis blandit
                nulla.
              </Tag>{' '}
            </Stack>
          </Card>
        </Layout.Section>
      </Layout>
    </Page>
  );
}
```

 </details>

<details><summary>Screenshot of the Playground</summary>

Screenshot 2021-12-16 at 15 57 06

</details>

src/components/Tag/Tag.tsx
jbalsas: I think if you're using `<Stack>` you'll need to also do `wrap={false}` or if you're not, you would take care of making that custom content inline, if that's what you were looking to do from a user of `Tag`

src/components/Tag/Tag.scss
jbalsas: I think if you're using `<Stack>` you'll need to also do `wrap={false}` or if you're not, you would take care of making that custom content inline, if that's what you were looking to do from a user of `Tag`.

But yeah, looks like if you have custom content, the ellipsis won't have an effect and I'm not sure if we can get it to work... shouldn't be the one crafting the custom content the one responsible for it? Should we add this as a known limitation?

src/components/Tag/tests/Tag.test.tsx
jbalsas: Yeah, not sure why I didn't add it there, would it be okay to just have both assertions or do we have some rules about that and we'd rather keep them separate?

src/components/Tag/Tag.tsx
kyledurand: Ah I see what you mean. I think I just thought this line was doing more than it actually was. Once I 🎩 ed it in the playground it seems pretty much fine, and not a breaking change. I did notice some weirdness around styles not being added: [screencast](https://screenshot.click/16-59-ly7cm-pm84g.mp4)

<details>
  <summary>Here's my playground</summary>

```jsx
import React from 'react';

import {Page, Stack, Tag} from '../src';

  return (
    <Page title="Playground">
      <Stack>
        <Tag onRemove={() => {}} url="#">
          tag with url
        </Tag>
        <Tag onRemove={() => {}}>tag with just onRemove</Tag>
      </Stack>
    </Page>
  );
}
```
</details>

jbalsas: > I did notice some weirdness around styles not being added

Weird!! I'll investigate!!

src/components/Tag/Tag.scss
jbalsas: This is the one way I found to make it work. I think the Grid layout is more strict in terms of size and does the job better in this case. I only did a shallow search for references in the specs, so don't have any concrete information as to why this is happening.

When trying to use flexbox I'd do:

```css
display: inline-flex;
max-width: 100%;
```

However, this would cause the remove button to overflow the element. 

Screenshot 2021-12-16 at 16 10 23

I tried different combinations, so I might've also missed a way to get this to work using flexbox. I'm open to suggestions here! ☺️

kyledurand: Ah strange, well if this works it's better than flex hacks for sure 👍

src/components/Tag/tests/Tag.test.tsx
kyledurand: I'm a fan of doing two assertions for a test like that but we have discrepancies throughout the repo. Feel free to do whichever

src/components/Tag/Tag.scss
kyledurand: Ok cool this looks great 👍. I think it was the stack issue in my quick playground that got me

jbalsas:
Hey @AndrewMusgrave, @kyledurand, I pushed one additional (hopefully the last one) [commit](https://github.com/Shopify/polaris-react/pull/4837/commits/d431dc322fba41be79531fc239c97379899edb6a) addressing most of those final details:
- Remove unnecessary `align-items: stretch;` rule in `linkable` items
- Use mixin `truncate` instead of manually reimplementing it here
- Flattens right border radius on segmented urls on hover
- Makes `accessibilityLabel` default to `undefined` so no `title` prop is rendered when ignored. We still render the `Remove ` label on the button for clarity on the action even though we can't add more information.
- Additional tests and assertions

I haven't attempted to optimize the CSS rules at this point, but I'll happily take a look into it in the future!

kyledurand (approved):
🎉

jbalsas:
Screenshot 2021-12-16 at 21 31 24
yes

PR #4860, opened 2021-12-22
Remove root level control height

kyledurand:
Just noticed some weirdness where `--pc-control-height` was being defined in may css files. 

[Screencast](https://screenshot.click/22-09-huwnj-e8hy3.mp4)

[Many definitions](https://screenshot.click/22-10-kjhc5-uk1xn.mp4) which could lead to a minor performance issue. This fixes that up by using the line-height var directly


aveline (approved):
👍

PR #4861, opened 2021-12-22
px to em in media queries

kyledurand:
Uses: https://github.com/niksy/postcss-em-media-query

Converts and min/max width, min/max height media queries from px to ems. Not super well known so we may want to fork this plugin and make sure it's solid but for now we can use this branch to experiment with removing the `em()` function @aveline

### DIFF
--- a/config/postcss-plugins.js
+++ b/config/postcss-plugins.js
@@ -1,11 +1,13 @@
 const postcssShopify = require('@shopify/postcss-plugin');
-const pxtorem = require('postcss-pxtorem');
+const pxToRem = require('postcss-pxtorem');
+const pxToEmMediaQuery = require('postcss-em-media-query');
 
 module.exports = [
   postcssShopify,
-  pxtorem({
+  pxToRem({
     rootValue: 16,
     replace: true,
     propList: ['*'],
   }),
+  pxToEmMediaQuery(),
 ];
--- a/package.json
+++ b/package.json
@@ -137,6 +137,7 @@
     "postcss-modules": "^4.2.2",
     "postcss-pxtorem": "^5.1.1",
     "postcss-value-parser": "^4.2.0",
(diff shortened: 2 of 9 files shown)


aveline:
Nice! I think this could be a good approach for our use case.

BPScott:
This feels very scary. It works ok for code in our codebase.

However this will have no effect whenever a consumer in web uses the `breakpoint()` function or any of the mixins that depend upon it. That means that all the css in web and other consuming projects shall end up with media queries that are specified in pixels because e.g. ` @include page-when-not-max-width {}` shall result in a media query that uses pixels. Having em media queries in our usage of these mixins, but px media queries when consumers use these mixins sounds like inconsistent confusing, buggy behaviour. I'd expect we'd want to be consistent.

I don't like the idea of forcing all consumers to use that postcss plugin if they want to maintain using em-based media queries behaviour.

Perhaps we could do something like claiming that `breakpoint()` now only accepts values/ adjustments in pixels (make it throw errors on other units) and that it shall internally convert that value into an em value by dividing that by 16 (as 1em always equals 16px when used within a media query at the default zoom level).


Alternatively can we reevaluate the need for using ems in media queries? The research for preferring using ems instead of px (https://zellwk.com/blog/media-query-units/) is several years old and there's a decent chance that Safari has fixed their inconsistent behaviour when using px in media queries in the last 6 years. If px behaves consistently across browsers now, is it ok to stick with plain px values for media queries instead of having to convert them to ems?

kyledurand:
Agreed about the scary bit but we've had talks about how to resolve those issues. My hope is ultimately we can go with pixels for breakpoints but that's going to take some testing. I'm leaving this as a draft until more folks are back from vacation

BPScott:
Is is worth closing this now we've merged #4937?

PR #4864, opened 2021-12-27
[ResourceItem] Fix: adjust ResourceItem media height to match its content height

RobinKemna:

While developing my custom Shopify app, I saw that the media wrapper of the ResourceItem component adds more space at the bottom than at the top of the media item.

Red color ist the background color for the image
Green color is the actual media wrapper background.

Before:
polaris bug

Remove the unnecessary extra space around the media component of ResourceItem by setting the line height of the media container to zero.

After:
polaris bug with fix


RobinKemna:
I would love to sign the CLA but unfortunatly I am not able to complete the process since github tells me something about a ratelimit...
github cla error

PR #4894, opened 2022-01-14
[Page]: Normalise Page Header/Title height

mrcthms:

Fixes https://github.com/Shopify/shopify/issues/322184

This PR fixes a visual bug in the `Title` component within the `Page` component. When a `Page` contains a `Title` with no other elements within the `Header`, it doesn't take up as much screen real estate, which can cause layout jumps when navigating from Page to Page. This PR introduces a new `isSingleRow` prop to the `Title`, which is already present in the `Header` component, to ensure that the `Title` now takes up as much room (40px) without additional elements as with additional elements.

#### Before
Screenshot 2022-01-14 at 12 28 03

#### After 
Screenshot 2022-01-14 at 12 28 08

### 🎩 checklist

### DIFF
--- a/src/components/Page/README.md
+++ b/src/components/Page/README.md
@@ -466,6 +466,26 @@ Use when the page needs visual separation between the page header and the conten
 </Page>
 ```
 
+### Very simple Page
+
+<!-- example-for: web -->
+
+Use when the page needs visual separation between the page header and the content.
+
+```jsx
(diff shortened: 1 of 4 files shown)


mrcthms:
Ahh, just see https://github.com/Shopify/polaris-react/pull/4779 which should fix this already. Closing.

PR #4897, opened 2022-01-14
[TextField] Added a class to hide the clear button instead of removing it from the DOM

sophschneider:
I found this issue when I was trying to add a clearable textfield to a `<Popover />`. The popover doesn't usually close when a button within it is pressed but since the `clearButton` on the `TextField` is removed from the DOM before the popover events are handled, it doesn't register it as a child of the popover. 

BEFORE
ezgif-3-deae442bf7

The easiest thing to do is to call `event.stopPropagation` on the click event, but the event is not currently accessible from the `onClearButtonClick` for the text field. This PR adds the event to the callback. I could've called `stopPropagation` directly in the `TextField` component, but that felt risky considering all the other events from the component bubble.

AFTER
ezgif-3-8df119b0bf

### How to 🎩

<details>

```jsx

import {Button, Icon, Popover, TextField} from '../src';

  const [queryValue, setQueryValue] = useState<string | undefined>(undefined);

  const togglePopoverActive = () =>
    setPopoverActive((popoverActive) => !popoverActive);

  const activator = (
      open
    </Button>
  );

  return (
    <Popover
      sectioned
      active={popoverActive}
      activator={activator}
      onClose={togglePopoverActive}
      ariaHaspopup={false}
    >
      <TextField
        placeholder="type here"
        onChange={setQueryValue}
        value={queryValue}
        label="type here"
        labelHidden
        prefix={<Icon source={SearchMinor} />}
        clearButton
        onClearButtonClick={(_id, event) => {
          setQueryValue(undefined);
          event.stopPropagation();
        }}
        autoComplete="off"
      />
    </Popover>
  );
}
```

</details>

### 🎩 checklist

Closes https://github.com/Shopify/web/issues/55738

### DIFF
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -125,10 +125,7 @@ interface NonMutuallyExclusiveProps {
   /** Determines the alignment of the text in the input */
   align?: Alignment;
   /** Callback when clear button is clicked */
-  onClearButtonClick?(
-    id: string,
-    event: React.MouseEvent | React.TouchEvent,
-  ): void;
+  onClearButtonClick?(id: string): void;
   /** Callback when value is changed */
   onChange?(value: string, id: string): void;
   /** Callback when input is focused */
@@ -281,20 +278,22 @@ export function TextField({
 
   const clearButtonVisible = normalizedValue !== '';
 
-  const clearButtonMarkup =
-    clearButtonVisible && clearButton ? (
-      <button
-        type="button"
-        className={styles.ClearButton}
-        onClick={handleClearButtonPress}
-        disabled={disabled}
-      >
-        <VisuallyHidden>
-          {i18n.translate('Polaris.Common.clear')}
-        </VisuallyHidden>
-        <Icon source={CircleCancelMinor} color="base" />
-      </button>
-    ) : null;
+  const clearButtonClassNames = classNames(
+    styles.ClearButton,
+    !clearButtonVisible && styles.Hidden,
+  );
+
+  const clearButtonMarkup = clearButton ? (
+    <button
+      type="button"
+      className={clearButtonClassNames}
+      onClick={handleClearButtonPress}
+      disabled={disabled}
+    >
+      <VisuallyHidden>{i18n.translate('Polaris.Common.clear')}</VisuallyHidden>
+      <Icon source={CircleCancelMinor} color="base" />
+    </button>
+  ) : null;
 
   const handleNumberChange = useCallback(
     (steps: number) => {
@@ -481,8 +480,8 @@ export function TextField({
     </Labelled>
   );
 
-  function handleClearButtonPress(event: React.MouseEvent | React.TouchEvent) {
-    onClearButtonClick && onClearButtonClick(id, event);
+  function handleClearButtonPress() {
+    onClearButtonClick && onClearButtonClick(id);
   }
 
   function handleKeyPress(event: React.KeyboardEvent) {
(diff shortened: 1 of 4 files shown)


src/components/TextField/TextField.tsx
```diff
@@ -125,7 +125,7 @@ interface NonMutuallyExclusiveProps {
   /** Determines the alignment of the text in the input */
   align?: Alignment;
   /** Callback when clear button is clicked */
-  onClearButtonClick?(id: string): void;
+  onClearButtonClick?(id: string, event: React.MouseEvent): void;
```
kyledurand: This won't always be a MouseEvent. Not sure if that matters a ton but maybe we can specify MouseEvent or TouchEvent?

kyledurand (approved):
I'm down for this change 👍

src/components/TextField/TextField.tsx
sophschneider: sounds good to me :)

src/components/TextField/TextField.tsx
```diff
@@ -478,8 +481,8 @@ export function TextField({
     </Labelled>
   );
 
-  function handleClearButtonPress() {
-    onClearButtonClick && onClearButtonClick(id);
+  function handleClearButtonPress(event: React.MouseEvent | React.TouchEvent) {
```
AndrewMusgrave: From your description it sounded like the issue is the fact the clear button is removed from the dom? Rather than exposing the event (since the issue will still occur without manually stopping propagation), we don't remove the clear button from the dom but visually hide it instead. What do you think?

sophschneider: love this idea! updating the PR now

sophschneider:
@AndrewMusgrave @kyledurand updated this PR using the method that @AndrewMusgrave suggested. It also stops the popover from changing widths every time the clear button was added to the DOM (see gif from above) so I love this solution :)

src/components/TextField/TextField.tsx
voiid: @AndrewMusgrave We had made this change a during the AODA project but it looks like [this PR](https://github.com/Shopify/polaris-react/pull/4369) re-removed the button from the DOM because of a screen reader issue so if we go this route we should make sure to test for it. 😄 

This issue has come up in a number of places though - makes me wonder if there's a system-wide fix we could think up for `PopoverOverlay` that could handle this issue? We've had to build similarly hacky workarounds for this issue and whenever it crops up and the investigation is a tricky one for anyone without this context.

sophschneider: @voiid @AndrewMusgrave I think the issue before is that this was using `VisuallyHidden` styles which has the purpose of visually hiding elements but still having them read by a screen reader

In this PR I use `visibility: hidden` which both visually hides the button AND hides it from the screen reader. I demoed it on voice over

AndrewMusgrave: Oof, and it was my pr as well 😅 Looks like I did leave a [comment](https://github.com/Shopify/polaris-react/pull/4369/files#r685332423) about using `aria-hidden` if needed. So that _should_ solve the screen reader issues 🤞 

> makes me wonder if there's a system-wide fix we could think up for PopoverOverlay

Overall, yeah this is definitely the better option. The popover shouldn't be closing if the click was in it, regardless of wether is exists in the dom anymore. I'm ok with this change for a short term fix though, since I imagine a change to `PopoverOverlay` will be a lot higher effort

sophschneider: Heres a demo with the rotor in voice over. as you can see the `Clear button` gets removed from the screen reader form elements when it is not visible which seems to be appropriate behaviour https://videobin.shopify.io/v/LA14z5

AndrewMusgrave (approved):
LGTM! ✅

PR #4900, opened 2022-01-17
Replace border-width() with tokens

### LINKED ISSUE #4891: Replace `border-width()` with tokens
As part of this, explore whether spacing tokens can be used rather than border width tokens.

aveline:

Resolves #4891

- Add border width tokens
- Replace instances of `border-width()` sass function with tokens
- Remove `border-width()` sass function


lgriffee (approved):
Looks awesome Aveline! 💯🎉 

Only non blocking question I have is after you merge this if we should move the `border-radius` tokens into the new `border` token file you created? They're in the `shape` file right now but I wonder if now that we have a `border` file that's a more intuitive place to put them?

aveline:
Thanks @lgriffee. I think shape makes sense for `border-radius`. Maybe `border-width` should be in the same file too?

lgriffee:
I think it makes sense to keep both `border-radius` and `border-width` together. Thinking through their relationship to one another perhaps you could say that `border-width` effects the visual weight of the shape that `border-radius` creates. And since we're not creating a parent file to organize things by visual weight, shape is the defining feature and that's what we should group both of them by...I guess then I'd question if the shape file itself is a good parent category to have. Do you remember what other tokens we were thinking of putting into the shape file?

aveline:
@lgriffee thanks, I moved `border-width` into shape. This works for now!

PR #4902, opened 2022-01-18
[Navigation] Improve Navigation IA

mrcthms:

Addresses https://github.com/Shopify/shopify/issues/322031

#### On desktop:

- Supporting root-level navigation items that contain `subNavigationItems` having a distinctive **active** state when a user is visiting that particular URL.

#### On mobile:

- Navigation Items containing `subNavigationItems` will act like an accordion where only one section can be opened at a time (with the exception of a section containing the currently active item, which can not be closed)

Desktop walkthrough:

https://user-images.githubusercontent.com/12194323/140996368-a9abc853-064d-446d-b817-162178ebfbe5.mov

New mobile UX: 

https://user-images.githubusercontent.com/2562596/149961810-15d94d9f-756f-4f52-b959-36da4c656789.mov

## 

### How to 🎩

[Spinstance](https://shop1.shopify.nav-ia-3.marc-thomas.eu.spin.dev/admin/products?selectedView=all)

Open the Spinstance, and use the navigation and compare it to this description to test changes.

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -10,6 +10,7 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 
 - Tightened up the Navigation component UI density. ([#4874](https://github.com/Shopify/polaris-react/pull/4874))
 - Updated the Navigation IA ([#4902](https://github.com/Shopify/polaris-react/pull/4902))
+- Added new `duplicateRootItem` prop to a Navigation Section to support new mobile Navigation IA ([#4902](https://github.com/Shopify/polaris-react/pull/4902))
 
 ### Bug fixes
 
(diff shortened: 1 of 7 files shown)


UNRELEASED.md
```diff
@@ -8,6 +8,8 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 
 ### Enhancements
 
+- Updated the Navigation IA ([#4902](https://github.com/Shopify/polaris-react/pull/4902))
```
dfmcphee, line 12: I think we could be a bit more clear here about the changes to the nav options

dfmcphee (approved):
One minor comment about the changelog, but otherwise this is looking good. 👍

alex-page (approved):
Happy for the desktop changes to ship.

I am low confidence about the mobile solution. Adding an additional property to duplicate the parent item seems low value and adds technical debt. Could we conditionally change the subNavigationItems instead of adding a property?

From a UX side, users have to tap multiple times. Sometimes a tap will expand the items and sometimes it will navigate to the page if it has no children. Is there a strong rationale for not expanding all of the items on mobile? Or navigating to the parent then expanding the children on the active page?

alex-page (approved):
Looks great. Thanks for adding the test ✨

dfmcphee (approved):

PR #4905, opened 2022-01-18
Remove _SECRET_INTERNAL_FilterControl export

BPScott:

This component has been deprecated and hidden behind a scary sounding import that does not promise backwards compatibility since polaris v5.

Legacy usage in web had blocked us from removing it fully, but now web no longer uses this thanks to the completion of Shopify/web#54545, and thus this component can now finally be fully removed.

- Deletes the `_SECRET_INTERNAL_FilterControl` and `_SECRET_INTERNAL_FilterControlProps` exports
- Deletes the src/components/ResoureceList/components/FilterControl directory
-  Removes now-unused translations

### How to 🎩

- Tests pass
- Note that [grokt shows that _SECRET_INTERNAL_FilterControl is now unused](https://grokt.shopifycloud.com/results?q=_SECRET_INTERNAL_FilterControl) (this might a day for it to realize that final usage in web's FilterControl has been removed).


alex-page (approved):

BPScott:
While I'm all rather excited by this FINALLY being possible I'll wait for the 7.6 release to go out before merging.

kyledurand (approved):
👏

PR #4907, opened 2022-01-20
Polaris v8

kyledurand:
### Breaking changes

- Update the base font size to `100%` from `62.5%` and update `rem` values accordingly, along with `pxtorem` `rootValue` ([#4794](https://github.com/Shopify/polaris-react/pull/4794))

### Enhancements

- Removed `_SECRET_INTERNAL_FilterControl` and `_SECRET_INTERNAL_FilterControlProps` exports. These exports have been deprecated since Polaris v5 and are not part of our stable API, which is why we are removing them in a minor release. ([#4905](https://github.com/Shopify/polaris-react/pull/4905))


laurkim (approved):
🚀

dev.yml
```diff
@@ -2,7 +2,7 @@ name: polaris-react
 up:
   - node:
       yarn: v1.13.0
-      version: v16.9.1 # to be kept in sync with .nvmrc and ci.yml (when v8 has shipped)
+      version: v16.9.1 # to be kept in sync with .nvmrc and ci.yml
```
alex-page, line 5: I recently had some problems with 16.9.1 in Buildkite. It couldn't make a container for it. Happy to ship this but if Buildkite fails we might have to push the version back... I don't even think this repo has Buildkite -_-

alex-page (approved):

aveline (approved):

dev.yml
BPScott: polaris-react runs its CI on github actions not buildkite.

kyledurand: Side note, we do have to downgrade the node version in v9 I believe because shipit was stuck at 16.9.1 and I think v9 is on 16.13. I'll do that now

kyledurand: I take it back. They've upgraded to 16.13 so v9 is good 😄 

https://github.com/Shopify/shipit/blob/master/.shopify-build/pipelines.yml#L6

PR #4934, opened 2022-01-20
Replace `border()` with tokens

aveline:

Fixes https://github.com/Shopify/polaris-react/issues/4915

- Add `border` tokens
- Replace instances of `border()` with tokens
- Remove `border()` sass function
- Rename `UNRELEASED-v8.md` to `UNRELEASED-v9.md`


lgriffee (approved):
LGTM! 🎉 I do question if we should have tokens for border shorthands (combined width, style, & color)—are they valuable? are they limiting? do we do shorthands for any other tokens?—but feel like we can revisit that when we do a next pass on the border token file.

aaronccasanova (approved):

PR #4937, opened 2022-01-21
Remove em() function

BPScott:

Attempting to reduce API surface area. We've removed the `px()` and `rem()` functions, let's see if we can remove `em()` too.

I was hoping to get it so that usage of `breakpoint()` no longer triggered deprecation warnings in dart-sass but I don't think that is possible without reworking all the layout stuff which is a gordian knot that needs more drastic reworking, so that's a problem for later. At least now deprecation warnings are scoped solely to usage of `breakpoint()` rather being in a utility `em()` function that anybody could use for anything.

Breaking changes:

- `$typography-condensed` is now a value in px (previously it was in ems)
- Remove `em()` sass function
- The `breakpoint()` sass function's signature has changed slightly but I highly doubt people used the ways that we have dropped support for (nothing in web uses the $adjustment argument):
  -  The $value argument now only accepts px and em values. It no longer accepts rem values.
  - The $adjustment argument has been removed

Outside of `_breakpoint.scss` all usage of `em()` defines values that then get passed into `breakpoint()`. `breakpoint()` converts values to ems anyway so this wrapping does nothing, and has been removed. In our public-api `$typography-condensed` is now in pixels instead of ems, but nobody externally should be using that value directly anyway.

Inside of `_breakpoint.scss`:

- The `breakpoint()` function's usage of `em()` has been inlined instead of calling out to `em()`
- Usage of `em()` to define values used when generating breakpoint variables has been replaced with calls to `breakpoint()`

And so there is no more usage of the `em()` function, and so it has been removed

### How to 🎩

Compare the build style file to what is produced by a build on the v9.0.0-major branch and confirm that the generated CSS is identical.

### DIFF
--- a/UNRELEASED-v8.md
+++ b/UNRELEASED-v8.md
@@ -22,6 +22,7 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 - Removed the `$duration-data` global variable ([#4699](https://github.com/Shopify/polaris-react/pull/4699))
 - Removed the `spacing()` scss function and replaced any instances with tokens ([#4691](https://github.com/Shopify/polaris-react/pull/4691/))
 - Removed the `px()` scss function ([#4751](https://github.com/Shopify/polaris-react/pull/4751))
+- Removed the `em()` scss function ([#4937](https://github.com/Shopify/polaris-react/pull/4937))
 - Removed the `z-index()` scss function ([#4753](https://github.com/Shopify/polaris-react/pull/4753))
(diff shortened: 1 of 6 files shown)


BPScott:
Possible next-step: currently all those variables at the top of _layout.scss are in ems. We could make them all be px values, and make it so `breakpoint()'s $value argument only accepts px values, but that seems a bit clunky and I'm not sure if it's worth it right now.

aveline (approved):

PR #4940, opened 2022-01-21
Remove font-family() function

### LINKED ISSUE #4918: Replace `@function font-family` with new token(s)

### LINKED ISSUE #4919: Consolidate font-size and line-height token files into typography
In the new `token-groups` folder consolidate the `font-size.json` and `line-height.json` files into the `typography.json` file

lgriffee:

Fixes #4918 
Fixes #4919 

- Removes the `font-family()` function
- Replaces any instances with new token values (reference table below)
- Removes `$typography-condensed` (value is still in `shared/_typography.scss`)
- Removes `foundation/_typography.scss`
- Consolidates `font-size` and `line-height` token files into `typography`

| Function             | Token | Value|
| --------------------------- | ------ | ------ |
| font-family()<br>font-family($family: base)              | --p-font-family-base      |  "-apple-system, BlinkMacSystemFont, <br>'San Francisco', 'Segoe UI',<br> Roboto, 'Helvetica Neue',<br> sans-serif"|
| font-family($family: 'monospace')          | --p-font-family-mono      | "ui-monospace, SFMono-Regular,<br> 'SF Mono', Consolas,<br> 'Liberation Mono', Menlo, <br>monospace"|


lgriffee:
I'm pretty sure the two Chromatic UI tests that changed below are from rendering inconsistencies in Chromatic's snapshots and not from my branch...I compared the Storybook instances on my branch to production and I'm not seeing any rendering shifting/differences. Would be good to get someone else to confirm they are seeing the same thing before accepting them.

aveline (approved):

src/components/AppProvider/AppProvider.scss
```diff
@@ -17,7 +17,7 @@ body {
 html,
 body,
 button {
-  font-family: font-family();
+  font-family: var(--p-font-family-base);
 }
```
alex-page, line 21: @lgriffee we have to be careful here as I a don't think the theme provider will be providing css custom properties at this level. @aaronccasanova and @kyledurand can confirm.

src/tokens/token-groups/typography.json
```diff
@@ -1,6 +1,31 @@
 {
+  "font-family-base": "-apple-system, BlinkMacSystemFont, 'San Francisco', 'Segoe UI', Roboto, 'Helvetica Neue', sans-serif",
+  "font-family-mono": "ui-monospace, SFMono-Regular, 'SF Mono', Consolas, 'Liberation Mono', Menlo, monospace",
```
alex-page, line 3: What about `font-family-sans` and `font-family-mono`? This makes more sense to me.

src/components/AppProvider/AppProvider.scss
aaronccasanova: This should be fine as the `CustomProperties` component [applies these vars directly to the `:root` pseudo class](https://github.com/Shopify/polaris-react/blob/be9a73d72415aec487a190956e144d4a59f05843/src/components/CustomProperties/styles.ts#L84). The important distinction is that these custom properties are considered static (aka. color-scheme independent). This is a great callout though because if this property was not static, it would need to be a descendent of an element with our  custom `color-scheme` attribute applied (e.g. a child of the `CustomProperties` component)

src/tokens/token-groups/typography.json
aaronccasanova: Makes sense to me as well 👍 Looks like [open props](https://github.com/argyleink/open-props/blob/b16fe44169cb76f5d2340e960fa25233254b5f41/src/props.fonts.css#L2-L4) takes a similar approach, though I prefer the explicit naming in your suggestions.

PR #4951, opened 2022-01-24
Remove `unstyled-link()` mixin

### LINKED ISSUE #4927: Remove and replace `@mixin unstyled-link()` with hard coded

lgriffee:

Fixes #4927  

- Removes the `unstyled-link()` mixin and replaces any instances with values 
- Removes `shared/_links.scss`


src/components/ActionList/ActionList.scss
```diff
@@ -83,6 +75,19 @@
...
+  }
+
+  &.active {
+    @include recolor-icon(var(--p-interactive));
+    background-color: var(--p-surface-selected);
+
+    &::before {
+      @include list-selected-indicator;
+    }
+  }
```
lgriffee, line 89: @kyledurand This is that class I mentioned moving during our pairing sesh and wanted to confirm that reordering won't muck up other styles. If it can be reordered, I thought it would be better to separate `.active` and `:active` and group all the classes together and all pseudo-classes  together.

UNRELEASED-v9.md
```diff
@@ -37,6 +37,7 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 - Removed the ms-high-contrast-color() function and replaced any instances with values ([#4938](https://github.com/Shopify/polaris-react/pull/4938))
 - Removed the `border()` scss function ([#4934](https://github.com/Shopify/polaris-react/pull/4934))
 - Removed the font-family() function and replaced any instances with tokens ([#4940](https://github.com/Shopify/polaris-react/pull/4940))
+- Removed the unstyled-link() mixin and replaced any instances with values ([#4951](https://github.com/Shopify/polaris-react/pull/4951))
```
aveline:
```suggestion
- Removed the `unstyled-link()` mixin and replaced any instances with values ([#4951](https://github.com/Shopify/polaris-react/pull/4951))
```

aveline (approved):

src/components/Banner/Banner.scss
```diff
@@ -245,6 +245,10 @@
     @include high-contrast-button-outline;
     box-shadow: none;
   }
+
+  &:visited {
+    color: var(--p-text);
```
kyledurand: Should this be inherit?

src/components/Tabs/Tabs.scss
```diff
@@ -98,6 +97,10 @@ $focus-state-box-shadow-color: rgba(0, 0, 0, 0.8);
       }
     }
   }
+
+  &:visited {
+    color: var(--p-text);
```
kyledurand: Same here?

kyledurand (approved):
Looks good just a question around visited color and if it's supposed to reset to text or just inherit. I can't forsee any issues in moving that .active chunk around either 🤞

src/components/Banner/Banner.scss
lgriffee: @kyledurand Yeah I think you're right about keeping this as `inherit`. I had changed the value to `var(--p-text)` to match the color property value of the class that `&:visited` gets applied too (I guess I was worried about weird color switching if the color values were different for some reason) but having `inherit` here would make it the same as what was being applied with the mixin originally. Thanks for catching this I should have made a note questioning it when I opened the PR. I'll change this back to `inherit` 👍🏻

PR #4954, opened 2022-01-25
Improve error message if sass fails to compile

BPScott:

When reworking some sass stuff I found it wasn't giving me helpful error messages. This improves the error messages that get raised, including call stacks for where the error occurred.

Adds a callstack to sass error messages.

Previously an error looked like this:

```
Error: no mixin named unstyled-linkk
```

Now it looks like this:

```
Error: Error: no mixin named unstyled-linkk
        on line 46 of src/components/ActionList/ActionList.scss
>>   @include unstyled-linkk;

   -----------^
```

## 

### How to 🎩

Compare `build/esm/styles.css` in a build on the main branch against that file after a build on this branch and note that they are identical.

Go modify a mixin call so you try to use an undefined mixin, and note that `yarn build` output shows where file/line of where you introduced the problem

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -17,6 +17,8 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 
 ### Development workflow
 
+- Improve error logging in the event of sass errors. ([#4954](https://github.com/Shopify/polaris-react/pull/4954))
+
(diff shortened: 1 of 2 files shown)


alex-page (approved):
Simple and great improvement. Cheers @BPScott

aveline (approved):

PR #4967, opened 2022-01-26
Remove `available-names` scss function

aveline:

Fixes https://github.com/Shopify/polaris-react/issues/4920

`available-names` is only used by `layout-width` for an error message. This pr removes the `available-names` function and hard codes the error message for `layout-width` as follows:

```
SassError: Column `primary - max` not found. Available columns:
             - primary (min, max)
             - secondary (min, max)
             - one-half-width (base)
             - one-third-width (base)
             - nav (base)
             - page-with-nav (base)
             - page-content (not-condensed, partially-condensed)
             - inner-spacing (base)
             - outer-spacing (min, max)
        on line 56 of src/styles/foundation/_layout.scss, in function `layout-width`
```

### DIFF
--- a/src/styles/foundation/_layout.scss
+++ b/src/styles/foundation/_layout.scss
@@ -43,36 +43,6 @@ $layout-width-data: (
   ),
 );
 
-/// Returns the list of available layout names.
-/// @param {Number} $map - The level of depth to get names from.
-/// @return {String} The list of names in the map.
-
-@function available-layout-names($level: 1) {
-  $output: '';
-  $newline: '\A ';
-
-  @if $level == 1 {
-    @each $key, $value in $layout-width-data {
-      $output: $output +
-        '#{$newline}- #{$key} #{available-layout-names($value, $level + 1)}';
-    }
-  } @else {
-    $output: '(';
-    $i: 1;
-
-    @each $key, $value in $layout-width-data {
-      $sep: if($i < length($layout-width-data), ', ', '');
-      $output: $output +
-        '#{$key}#{$sep}#{available-layout-names($value, $level + 1)}';
-      $i: $i + 1;
-    }
-
-    $output: $output + ')';
-  }
-
-  @return $output;
-}
-
 /// Returns the widths of the specified column.
 /// @param {String} $name - The column name.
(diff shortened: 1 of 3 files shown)


src/styles/foundation/_layout.scss
```diff
@@ -53,7 +83,7 @@ $layout-width-data: (
   @if type-of($fetched-value) {
     @return $fetched-value;
   } @else {
-    @error 'Column `#{$name} - #{$value}` not found. Available columns: #{available-names($layout-width-data)}';
+    @error 'Column `#{$name} - #{$value}` not found. Available columns: #{available-layout-names()}';
```
BPScott: I'd be pretty tempted to hardcode the string of what the function returns, and do away with needing the available-layout-names function. We'd need to update it if we modify the `$layout-width-data` but I suspect when that happens the whole thing shall  be radically reworked anyway.

Example error:

```
Error: Column `primary - max` not found. Available columns:
       - primary (min, max)
       - secondary (min, max)
       - one-half-width (base)
       - one-third-width (base)
       - nav (base)
       - page-with-nav (base)
       - page-content (not-condensed, partially-condensed)
       - inner-spacing (base)
       - outer-spacing (min, max)
```


Code (I think this works, I've not tested it):

```
@error 'Column `#{$name} - #{$value}` not found. Available columns:\a       - primary (min, max)\a       - secondary (min, max)\a       - one-half-width (base)\a       - one-third-width (base)\a       - nav (base)\a       - page-with-nav (base)\a       - page-content (not-condensed, partially-condensed)\a       - inner-spacing (base)\a       - outer-spacing (min, max)';'
```

Note I'm using `\a` for newlines so that it works in both node-sass and dart-sass per https://github.com/Shopify/polaris-react/pull/3063

We could also modify the error message so it's a comma separated list rather than mucking about with indentation and newlines

aveline: This looks good to me. Keeps the error message details and removes the function. Thanks Ben!

alex-page (approved):
Smart solution. Awesome work @aveline and @BPScott

lgriffee (approved):
💯🎉🚢✨

PR #4974, opened 2022-01-27
Refactor visually hidden

aveline:

Fixes https://github.com/Shopify/polaris-react/issues/4922

- Creates a `polaris-visually-hidden` global utility class
- Remove `visually-hidden` sass mixin
- Refactors components with usage of `visually-hidden` sass mixin to use global utillity class or `VisuallyHidden` component

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/AppProvider/AppProvider.scss
+++ b/src/components/AppProvider/AppProvider.scss
@@ -3,6 +3,7 @@
 // components then this css shall always be present.
 
 @import '../../styles/common';
+@import '../../styles/utilities.scss';
 
 :root {
   --polaris-version-number: '{{POLARIS_VERSION}}';
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -143,6 +143,7 @@ export const Checkbox = forwardRef<CheckboxHandles, CheckboxProps>(
       styles.Input,
       isIndeterminate && styles['Input-indeterminate'],
       keyFocused && styles.keyFocused,
+      'polaris-visually-hidden',
     );
 
     return (
--- a/src/components/Checkbox/README.md
+++ b/src/components/Checkbox/README.md
@@ -134,6 +134,24 @@ function CheckboxExample() {
 
 <!-- /content-for -->
 
+### Checkbox with label hidden
+
+```jsx
+function CheckboxExample() {
+  const [checked, setChecked] = useState(false);
+  const handleChange = useCallback((newChecked) => setChecked(newChecked), []);
+
+  return (
+    <Checkbox
+      label="Basic checkbox"
+      checked={checked}
+      onChange={handleChange}
+      labelHidden={true}
+    />
+  );
+}
+```
+
 ---
 
 ## Related components
--- a/src/components/VisuallyHidden/VisuallyHidden.scss
+++ b/src/components/VisuallyHidden/VisuallyHidden.scss
@@ -1,5 +1,14 @@
-@import '../../styles/common';
-
 .VisuallyHidden {
-  @include visually-hidden;
+  // Need to make sure we override any existing styles.
+  // stylelint-disable declaration-no-important
+  position: absolute !important;
+  width: 1px !important;
+  height: 1px !important;
+  margin: 0 !important;
+  padding: 0 !important;
+  overflow: hidden !important;
+  clip-path: inset(50%) !important;
+  border: 0 !important;
+  white-space: nowrap !important;
+  // stylelint-enable declaration-no-important
 }
--- a/src/components/Checkbox/Checkbox.scss
+++ b/src/components/Checkbox/Checkbox.scss
@@ -7,8 +7,6 @@
 
 // stylelint-disable selector-max-specificity, selector-max-class
 .Input {
-  @include visually-hidden;
-
   &.keyFocused {
     + .Backdrop {
       @include focus-ring($style: 'focused');
--- a/src/components/Choice/Choice.scss
+++ b/src/components/Choice/Choice.scss
@@ -10,10 +10,6 @@
 .labelHidden {
   padding: 0;
 
-  > .Label {
-    @include visually-hidden;
-  }
-
   .Control {
     margin-top: 0;
     margin-right: 0;
--- a/src/components/Choice/Choice.tsx
+++ b/src/components/Choice/Choice.tsx
@@ -56,7 +56,14 @@ export function Choice({
       onMouseOut={onMouseOut}
     >
       <span className={styles.Control}>{children}</span>
-      <span className={styles.Label}>{label}</span>
+      <span
+        className={classNames(
+          styles.Label,
+          labelHidden && 'polaris-visually-hidden',
+        )}
+      >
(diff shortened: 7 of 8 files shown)


src/components/AppProvider/AppProvider.scss
```diff
@@ -3,6 +3,7 @@
 // components then this css shall always be present.
 
 @import '../../styles/common';
+@import '../../styles/utilities.scss';
```
aveline, line 6: Not sure if this is the right place to include the global utility classes?

src/components/Checkbox/README.md
```diff
@@ -134,6 +134,24 @@ function CheckboxExample() {
 
 <!-- /content-for -->
 
+### Checkbox with label hidden
```
aveline, line 137: Added this for verifying in Storybook, but not sure how that will affect the styleguide and if we want it included there.

src/components/VisuallyHidden/VisuallyHidden.scss
```diff
@@ -1,5 +1,14 @@
...
+  width: 1px !important;
+  height: 1px !important;
+  margin: 0 !important;
+  padding: 0 !important;
+  overflow: hidden !important;
+  clip-path: inset(50%) !important;
+  border: 0 !important;
+  white-space: nowrap !important;
+  // stylelint-enable declaration-no-important
 }
```
alex-page, line 14: We could remove this component all together. Or at least deprecate it in version 9.

src/components/AppProvider/AppProvider.scss
alex-page: I think `global.scss` might be a good file name.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -143,6 +143,7 @@ export const Checkbox = forwardRef<CheckboxHandles, CheckboxProps>(
       styles.Input,
       isIndeterminate && styles['Input-indeterminate'],
       keyFocused && styles.keyFocused,
+      'polaris-visually-hidden',
```
aveline, line 146: What do we think about how to name util classes? Prefix with `polaris-` or?

laurkim: I like this format. It's clear where the util class is coming from and I know immediately without too much mental processing what the class is supposed to do.

src/components/AppProvider/AppProvider.scss
BPScott: Gotta admit, I'm not a fan of polaris exposing arbitrary global styles that people can leverage. This feels even more fragile and hard to remove once it's in  than having a Sass API. The current global styles in this file has already felt a bit clunky; we don't use `:global` for classes anywhere else and I'm not convinced VisuallyHidden is a good enough reason to start. Global styles have always felt kinda weird in our react apps to me our components are set up to be self contained and you only use pull in the styles when you use the components. With global styles it isn't clear where the class names we depend upon come from - there's no proof that any global style we reference in a component is actually present, which feels fragile.


I think I'd prefer for this to be kept as a mixin, but perhaps we remove it from _public-api, and we keep it in _common - so that it is only usable internally, but not by external consumers. That way we're still shrinking our external surface area.


If you do decide to stick with the global class, yes in here is the best place. I think I'd prefer the class to be included directly in this file rather than imported from an external file.

alex-page: I think there is currently two use cases where global css could be considered. visually hidden and hide on print. I do want to reduce the need for rendering `<span>` elements to achieve this. I wonder if with strict guidelines and rules this approach would be considered low risk.

src/components/Checkbox/Checkbox.tsx
alex-page: Maybe `.p-` to align with the css custom properties.

src/components/AppProvider/AppProvider.scss
aveline: @BPScott good points. Something like this?

```
import styles from './Checkbox.scss';
import utilityStyles from '../../styles/utilities.scss';
```

```
    const inputClassName = classNames(
      styles.Input,
      utilityStyles.VisuallyHidden,
      isIndeterminate && styles['Input-indeterminate'],
      keyFocused && styles.keyFocused,
    );
```

The rendered html looks like this, which is still somewhat legible.

```
<input id="PolarisCheckbox2" type="checkbox" class="Checkbox-Input_1uEY6 utilities-VisuallyHidden_w6VoC" aria-invalid="false" role="checkbox" aria-checked="false" value="">
```
I do like the separate class because it surfaces right in the dom that there are a group of reusable properties being applied to this element, without having to go into the code to see that they are coming from a mixin.

If they get imported into components, then they don't need to be global anymore either. Seems like this approach addresses the concerns with global styles, while still allowing us to deprecate the mixin?

BPScott: I think my big question is should these classes be considered part of a public api and be used by folks that consume polaris or are they just an internal implementation detail? Considering we're talking about shrinking our public sass API to reduce the surface area the polaris team has to maintain, I assume we're not interested in adding classnames as a public API right now, so this is an internal implementation detail.

Thanks for thinking of this alternative. I think that this is a nicer approach than putting global classes into AppProvider - it keeps those dependencies clear, makes it easy to see where the class is used is used or not by checking import, and it obfuscates the classnames thus making it clear they are an internal implementation detail, and not something others should use.


I'm still not super convinced that it gives us much value over using a mixin that we don't expose in `_public-api` though. We're trading "do an import in a scss file and do an `@include visually-hidden()`" for "do an import in a js files, and reference `utilityStyles.VisuallyHidden` in a classNames block, which feels like a sideways step rather than an improvement. I don't think I see the value in  adding a new mechanism to access utility styles for 2 use-cases and a total of 17 individual usages (12x `visually-hidden` is used 12 times, 3x `hidden-when-printing`, 2x `print-hidden`, and IMO the hidden-when-printing/print-hidden usages could just be inlined as we use `@media print` directly almost as much as we use the print mixins), what am I missing, what iterative improvement does this unlock?

alex-page: Two points come to mind...

1. Scss functions and mixins are not documented on polaris.shopify.com. This should be documented as a class name or a scss mixin. My hesitation about scss mixins is that there will be others that we are removing in future versions that would not be documented. It might create confusion for consumers.
2. Long long term removing scss all together from polaris-react. Make this the simplest design system possible.

IMO this can stay the same or change. The most important thing is to start this conversation and decide on next steps.

BPScott: If a mixin is only used in the generation of our components and is not exposed in `_public-api.scss`, and thus inaccessible to consumers why would we need to document its usage? It would be incapable of being used by any external consumers.

aveline:
This was a good exploration and thanks everyone for the great discussion. Going to close this for now as we have reprioritized our focus.

PR #4978, opened 2022-01-28
feat: Add Typography component

### LINKED ISSUE #4932: Replace `shared/_typography.scss` with react component
Replace shared/_typography.scss with react component if effort is minimal, otherwise wait add to create type project.

aaronccasanova:

Closes #4932 

This PR removes the `shared/_typography.scss` mixins and re-exposes their behavior through a new `Typography` component. There a number of improvements we can make to the `Typography` component (such as: adding/removing variants, removing the emphasis prop, consolidating the existing `Heading` and `TextStyles` components, etc.), but this PR primarily focuses on maintaining backward compatibility with the `shared/_typography.scss` utilities.

TODO:
- [ ] Write tests
- [ ] Refine the README and storybook entries
- [ ] (Optional) Swap out polymorphic utilities with a library (such as: https://www.radix-ui.com/docs/primitives/utilities/polymorphic)

### How to 🎩

To top hat, simply checkout the `feat/typography-component` branch, run `dev up && yarn storybook`, and navigate to the `Typography` storybook entries.

### 🎩 checklist

### DIFF
--- a/src/components/Typography/README.md
+++ b/src/components/Typography/README.md
@@ -1,0 +1,76 @@
+---
+name: Typography
+category: Titles and text
+platforms:
+  - android
+  - ios
+  - web
+keywords:
+  - titles
+  - text
+  - microcopy
+  - conversational
+  - typographic
+  - card headings
+  - card titles
+  - section titles
+  - section headings
+  - heading text
+  - heading font
+  - android
+  - ios
+---
+
+# Typography
+
+Generic typography component encapsulating an opinionated collection of design token groups.
+
+---
+
+## Examples
+
+### Typography default
+
+Render the body variant by default with a paragraph tag.
+
+```jsx
+<Typography>The quick brown fox jumps over the lazy dog</Typography>
+```
+
+### Typography heading
+
+Setting a variant will apply a collection of design tokens and a predefined HTML element.
+
+```jsx
+<Typography variant="heading">
+  The quick brown fox jumps over the lazy dog
+</Typography>
+```
+
+### Typography polymorphic
+
+Use the `as` prop to override the default HTML element associated with each variant.
+
+```jsx
+<Typography variant="heading" as="span">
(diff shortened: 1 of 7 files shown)


src/components/Typography/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Typography polymorphic
+
+Use the `as` prop to override the default HTML element associated with each variant.
+
+```jsx
+<Typography variant="heading" as="span">
+  The quick brown fox jumps over the lazy dog
+</Typography>
+```
```
alex-page, line 58: I think I would prefer lots of smaller opinionated components. `<Heading>` `<Body>` etc. This adds an additional layer of documentation and explanation when I think we could keep the first approach super simple. What do you think?

aaronccasanova: Personally, I don't see the benefit of breaking type styles into multiple components. While I have seen this pattern in popular design systems (e.g. [Primer Text](https://primer.style/react/Text), [Primer Heading](https://primer.style/react/Heading), [Rebase Text](https://rebassjs.org/text/), [Rebase Heading](https://rebassjs.org/heading/), etc.) its always felt a little extraneous to me.

I prefer the simplicity and reduced cognitive overhead of having one component for applying type styles. Some popular references following this approach are [React Native Text](https://reactnative.dev/docs/text), [Material UI Typography](https://mui.com/components/typography/#component), and [React Ink Text](https://github.com/vadimdemedes/ink#text). This approach allows us to easily iterate on type styles in one place and avoid duplication (e.g. adding a `lineClamp` prop, exposing `capsize` like controls, etc.)

I also think one problem we have today and have had various conversations about is the fact that there is an excessive number of typographic components in `polaris-react` and a lack of understanding how and when to use them. If we encapsulate all the behavior in one component we could consolidate documentation in one location that we then direct folks too for typography usage and education.

PR #4989, opened 2022-01-31
Replace `@mixin state()` with value

### LINKED ISSUE #4988: Replace `@mixin state()` with value

lgriffee:

Fixes #4988  

- Removes the `state()` mixin and replaces any instances with values

### DIFF
--- a/src/components/VideoThumbnail/VideoThumbnail.scss
+++ b/src/components/VideoThumbnail/VideoThumbnail.scss
@@ -43,7 +43,11 @@ $progress-bar-height: 6px;
 
   &:focus {
     outline: none;
-    @include state(focused);
+    box-shadow: inset 2px 0 0 var(--p-focused);
+    background-image: linear-gradient(
+      rgba(223, 227, 232, 0.3),
+      rgba(223, 227, 232, 0.3)
+    );
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/components/VideoThumbnail/VideoThumbnail.scss
```diff
@@ -24,6 +24,7 @@ $progress-bar-height: 6px;
 }
 
 .PlayButton {
+  --pc-play-button-focused-state-overlay: rgba(223, 227, 232, 0.3);
```
lgriffee, line 27: This rgba value is an extended state color that used to be part of [color-palette-data](https://github.com/Shopify/polaris-react/pull/4715/commits/2508ebb7e23cc82748edf7d841d889d72fc1ea90). I didn't want to create a one off value so I put it into a `--pc-*` token. Would love feedback on this approach.

alex-page (approved):

src/components/VideoThumbnail/VideoThumbnail.scss
alex-page: This is the right call for now. We can in another PR go through all the hard coded color values.

aveline (approved):

PR #4993, opened 2022-01-31
Remove public scss api

aveline:

Fixes https://github.com/Shopify/polaris-react/issues/4980

Removes `_public-api.scss`

## 

### How to 🎩

- run `yarn build`
- verify that `build/styles/_public-api.scss` no longer exists


documentation/guides/migrating-from-v8-to-v9.md
```diff
@@ -0,0 +1,37 @@
+# Migrating from v8 to v9
+
+Polaris v9.0.0 ([full release notes](https://github.com/Shopify/polaris-react/releases/tag/v9.0.0)) features removal of the public scss api and removal of scss functions and mixins.
+
+## Theme provider
+
+## Removal of the public scss api
+
+Any functions that were being consumed from `build/styles/_public-api.scsspri` have been removed. The functions can be found in the following permalinks.
```
lgriffee:
```suggestion
Any functions that were being consumed from `build/styles/_public-api.scss` have been removed. The functions can be found in the following permalinks.
```

BPScott (commented):
loom.config / build-validate / public-api changes look great.

Regarding the migration guide, I think @alex-page had an idea that we should provide a scss file that consumers could drop into their codebase and it just works™️ . The contents of that file being public-api.scss as of the last release of v8, but all in one big file (no `@imports`, just inlined code). I think that'd be a nicer thing to teach, and would be easier for folks to consume - "heres everything, delete what you don't need" is easier than "you've got to pick the bits that you were actually using".

aaronccasanova:
> The contents of that file being public-api.scss as of the last release of v8, but all in one big file (no @imports, just inlined code). I think that'd be a nicer thing to teach, and would be easier for folks to consume - "heres everything, delete what you don't need" is easier than "you've got to pick the bits that you were actually using".

@BPScott I really like this idea. Wonder if the "delete what you don't need" part will be tedious and time consuming. What if we wrote a script that scans all scss files in a directory, identifies all the legacy/public-apis being used, and dumps those into a single file?

alex-page:
@aaronccasanova that is a really smart idea. However I don't think it's worth the investment. `shopify/web` uses the majority of the public api and we aren't focused on third parties with the impact of these changes.

A simpler approach might be recommending consumer to building scss -> css and then only adding the mixin and functions that fail.

aaronccasanova:
Good callout, I was definitely questioning "is it worth the investment" as I was writing that.

> A simpler approach might be recommending consumer to building scss -> css and then only adding the mixin and functions that fail.

Great idea 👌

BPScott:
Very possible, but then we'd have to go build and distribute a thing. The correctness of that is really hard as well as sass's `@import` gives no statically analysable hints  about where a mixin comes from - to do that properly you wouldn't want to include a mixin just because it shares the name with some other utility they're already using. That seems like an order of magnitude harder than "here's a codeblock tucked behind a `<details>` to collapse it in a markdown file that you copy+paste."

BPScott:
> only adding the mixin and functions that fail.

This is cool and easy for mixins and variables. But sass is totally fine with you using functions that it doesn't know about - because it thinks that might be a css function (like `linear-gradient()`) so it leaves it alone.

```
// whoops you missed an import that defines `spacing()`
.style { padding: spacing(); }
```

would produce the compiled output `.style { padding: spacing(); }` without any warnings at all.


Consumers would need to assert that their build output is identical rather than just "does it compile".

"pick what you need" as a starting point also puts the onus on the consumer to work out transitive dependencies - you use mixinB in your code, mixinB internally depends on mixinA, so now you need to go import mixinA too.

alex-page (approved):
Lets ship it and write the docs in a follow up

PR #4996, opened 2022-02-01
Formatting v9 Changelog

alex-page:

I need to write vault updates and it's easier with a little bit of formatting to quickly get the information.

Formatting the Changelog to make getting information from it a little bit faster.


lgriffee (approved):
Love the sections you created—with so many breaking changes it definitely makes the changelog much easier to read! 💯🎉🚀✨

PR #5013, opened 2022-02-02
docs: Migration guide for the ThemeProvider/CustomProperties component

### LINKED ISSUE #5007: Create migration guidance for ThemeProvider

aaronccasanova:
WHY are these changes introduced?

Closes #5007

WHAT is this pull request doing?

- Added migration steps for transitioning from the `ThemeProvider` to the `CustomProperties` component
- Added migration steps for the various types, constants, and utilities used with the `ThemeProvider`


documentation/guides/migrating-from-v8-to-v9.md
```diff
@@ -0,0 +1,61 @@
...
++    {props.children}
++  </AppProvider>
+)
+```
+
+## Removed all theme types, constants, and utilities
+
+A number of types, constants, and utilities have been removed with the deprecation of the `ThemeProvider` component:
+
+- `ThemeContext` - React context
```
BPScott, line 52: These only need to be documented if they were part of the public api that is exported from `src/index.ts`. If they weren't exported from there then consumers would have been incapable of accessing these values/types in the first place.

aaronccasanova: I considered removing this section or reformatting it since only `UNSTABLE_toCssCustomPropertySyntax` and `UNSTABLE_Tokens` are exported from `src/index.ts`. However, I kept it in to open up the discussion in the PR and see if we cared about folks possibly diving into the package's directories or accessing unexported types via index access types:

For example:
```ts
// Not exported from `src/index.ts` but accessible from the package
import {useTheme} from '@shopify/polaris/build/esm/utilities/theme' 
```
or
```ts
// Not exported from 'src/index.ts' but accessible via index acess types
import {ThemeProviderProps} from '@shopify/polaris'

ThemeProviderProps['theme'] // ThemeConfig type (Very common is TS codebases)
```

Honestly, I'm not sure we need to worry about these edge cases, but figured it didn't hurt to include them.

aveline:
L👀ks good to me

aveline (approved):

PR #5046, opened 2022-02-03
Create migration guidance for `filter()` function

### LINKED ISSUE #5008: Create migration guidance for `filter()` function

lgriffee:

Fixes #5008 

- Adding documentation for the`filter()` function to our migration guide

<hr>

### Helpful Reference PRs/Files

- #4650 
- #4676


ghost:

lgriffee:
Looking through [#4650](https://github.com/Shopify/polaris-react/pull/4650/files) it does look like we replaced some `filter()` values with color tokens. I added these values into the replacement guide table as well with a note.

| Function | Replacement Value |
| --------------------- | ----------------- |
| filter('white')<br/>filter('white', 'base')  | var(--p-icon-on-interactive)                          |
| filter('icon')<br/>filter('icon', 'base')   | var(--p-icon)                                         |
| filter('action')<br />filter('action', 'base') | var(--p-action-primary)* <br> var(--p-icon-on-interactive)* |

*Depending on state

documentation/guides/migrating-from-v8-to-v9.md
```diff
@@ -4,6 +4,93 @@ Polaris v9.0.0 ([full release notes](https://github.com/Shopify/polaris-react/re
...
+
+To help you quickly add these functions and mixins back to your repo, we've created a css file with all the removed functions and mixins.
+
+[✨ Amazing mega file linked here ✨]
+
+### Replacing function and mixin instances with values or tokens
+
+To help you replace function and mixin instances in your repo, we've provided the following list of functions/mixins and their value equivalents or new token values.
+
+#### `filter()`
```
aaronccasanova, line 114: Thoughts on breaking this `filter()` section in two parts?
1. (Table not in `<summary>`) Here are all the values we've replaced with tokens/values NOT css filters
2. (Table in `<summary>`) Here is an exhaustive list of all legacy css filters you can use for backward compatibility

lgriffee: Good callout, I was also wondering what's the clearest and most helpful documentation for folks in this case. I ended up combining tokens and values into one table for the [`easing()` guidance](https://github.com/Shopify/polaris-react/pull/5047/files) so I was trying to be consistent with that approach here. 

But perhaps this is just such a giant table and the tokens aren't 1:1 replacements so it makes sense to break them out into two here. I'm totally down with doing that if it feels more clear. 

If we do break them into two tables for `filter()`, I do want to just quickly consider if we should do the same for `easing()` to be consistent? Some of the easing tokens aren't 1:1 replacements for the previous function values. However, unlike `filter()` they are improvements on `easing()` values versus values meant for another property, which makes me lean towards thinking `easing()` values and tokens are fine in the same table together vs. two tables even if we break the `filter()` table up.

documentation/guides/migrating-from-v8-to-v9.md
```diff
@@ -4,6 +4,93 @@ Polaris v9.0.0 ([full release notes](https://github.com/Shopify/polaris-react/re
...
+| filter('black')<br>filter('black', 'base')   | brightness(0) saturate(100%)                                                                                                                                                                         |
+| filter('white')<br>filter('white', 'base')   | brightness(0) saturate(100%) invert(100%)<br><br>var(--p-icon-on-interactive)[^1]                                                                                                                    |
+| filter('icon')<br>filter('icon', 'base')     | brightness(0) saturate(100%) invert(36%) sepia(13%) saturate(137%) hue-rotate(169deg) brightness(95%) contrast(87%)<br><br>var(--p-icon)[^1]                                                         |
+| filter('action')<br>filter('action', 'base') | brightness(0) saturate(100%) invert(20%) sepia(59%) saturate(5557%) hue-rotate(162deg) brightness(95%) contrast(101%)<br><br>var(--p-action-primary)[^1][^2]<br>var(--p-icon-on-interactive)[^1][^2] |
+
+[^1]: Our team did replace some of these filter function instances with color tokens but be careful in doing so it can break intended behaviour.
+[^2]: Token usage varried depending on state.
```
kyledurand:
```suggestion
[^2]: Token usage varied depending on state.
```

kyledurand (approved):
Looking good! Just one lil typo

documentation/guides/migrating-from-v8-to-v9.md
aaronccasanova: Personally, I think the guidance should help folks get their apps back to a functional state as quickly as possible. We were able to take some time throughout the process to update and test new ideas, but others may not have that luxury. This is just my 2 cents, so I wonder what other folks think. @alex-page @aveline Thoughts?

documentation/guides/migrating-from-v8-to-v9.md
```diff
@@ -89,6 +89,17 @@ To help you quickly add these functions and mixins back to your repo, we've crea
 
 A list of functions/mixins and their value equivalents or new token values.
 
+#### `duration()`
```
lgriffee: Moving this up here so all the functions/mixins are in alphabetical order.

documentation/guides/migrating-from-v8-to-v9.md
lgriffee: Spoke with @alex-page about this and we decided to remove the color token replacement values from the table and just have a general note/disclaimer at the top that our team replaced some instances of the function with color tokens (and that folks should use caution if they do that too).

sam-b-rose:
This looks awesome, @lgriffee! Extensive and comprehensive migration table 💯 

This may have already come into question; I don't have all the context.
How do we see `filters` being updated and maintained moving forward? This seems to eject from tokenized/function values to hard-coded filter values.

Are we dropping filter support altogether, or do we see value in tokenizing these property values?

```scss
// Tokenize
--p-filter-purple: brightness(0) saturate(100%) invert(29%) sepia(3%) saturate(2843%) hue-rotate(223deg) brightness(92%) contrast(86%);


// Migration from
filter('purple', 'dark');

// Migration to
filter: var(--p-filter-purple-dark);
```

lgriffee:
My understanding for `filters` moving forward is that we are dropping support altogether—all instances of the function should be removed and replaced with hard-coded values—because the color system used by this function has been deprecated (and therefore we don't want to replace it with tokens).

sam-b-rose:
Absolutely makes sense! Thanks for the added context, @lgriffee!

Definitely in favor of dropping the filter values from the system 👍

PR #5052, opened 2022-02-04
Nav polish

kyledurand:
https://github.com/Shopify/polaris-react/issues/5003

### DIFF
--- a/src/components/Navigation/Navigation.scss
+++ b/src/components/Navigation/Navigation.scss
@@ -7,6 +7,7 @@ $nav-max-width: 360px;
 
 .Navigation {
   --pc-navigation-icon-size: 20px;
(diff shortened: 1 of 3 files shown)


src/components/Navigation/Navigation.scss
```diff
@@ -17,7 +18,7 @@ $nav-max-width: 360px;
   min-height: 100%;
   background-color: var(--p-background);
   -webkit-overflow-scrolling: touch;
-  border-right: var(--p-border-divider);
+  border-right: 1px solid var(--p-divider);
```
sam-b-rose: The `--p-border-divider` var was fixed in #5121 so I think we could leave this as-is now 👍

sam-b-rose:
Changes incorporated in with layout updates in #5161 closin' out 🚀

PR #5075, opened 2022-02-07
Migration guidance `em()`

### LINKED ISSUE #5022: Create migration guidance for `em()` function

aveline:
Fixes #5022

Adds migration guidance for deprecated `em()` scss function

### DIFF
--- a/documentation/guides/migrating-from-v8-to-v9.md
+++ b/documentation/guides/migrating-from-v8-to-v9.md
@@ -151,9 +151,10 @@ A list of functions/mixins and their value equivalents or new token values.
 
 #### `em()`
 
-| Function | Source to copy       |
+This function has been deprecated, but the definition can be copied and used locally.
+| Function | Source |
 | -------- | -------------------- |
-| `px()`   | [link to function]() |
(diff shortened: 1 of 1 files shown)


documentation/guides/migrating-from-v8-to-v9.md
```diff
@@ -100,6 +100,10 @@ A list of functions/mixins and their value equivalents or new token values.
 | `duration(slower)`               | `--p-duration-400`      |
 | `duration(slowest)`              | `--p-duration-500`      |
 
+#### `em()`
+
+Any usage of this function can be hardcoded to `em` units. Alternatively, values can also be hardcoded to `px` units, which will automatically be converted to `rem` in postcss.
```
BPScott: > will automatically be converted to `rem` in postcss.

This is only true if the consumer has a postcss plugin that does this. This repo contains such a plugin but `@shopify/postcss-plugin` doesn't contain anything that does that, any app using sewing-kit would need to opt into using that plugin, which seems like a pretty disruptive migration suggestion.

aveline:
Converted to draft as this is now blocked until the mega functions file is created

alex-page (approved):

lgriffee (approved):

PR #5122, opened 2022-02-09
Create migration guidance for `available-names()` function

### LINKED ISSUE #5027: Create migration guidance for `available-names()` function

lgriffee:

Fixes #5027

- Adding documentation for the`available-names()` function to our migration guide

<hr>

### Helpful References

- #4967

### DIFF
--- a/documentation/guides/migrating-from-v8-to-v9.md
+++ b/documentation/guides/migrating-from-v8-to-v9.md
@@ -118,6 +118,12 @@ A list of functions/mixins and their value equivalents or new token values.
 
 Use `console.log()` to get the function output and hard code the value you need.
 
+Otherwise, you can copy the function definition and use it locally.
+
+| Function            | Source                                                                                                                                                                |
+| ------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


documentation/guides/migrating-from-v8-to-v9.md
```diff
@@ -114,6 +114,10 @@ To help you quickly add these functions and mixins back to your repo, we've crea
 
 A list of functions/mixins and their value equivalents or new token values.
 
+#### `available-names()`
+
+Use `console.log()` to get the function output and hard code the value you need.
```
lgriffee, line 119: I didn't reference the mega file here and only gave instructions for replacing the instances. But not sure if this is the right approach. Originally in the doc setup I created an "Adding the functions and mixins to your repo" section with the mega file and a "Replacing function and mixin instances with values or tokens" section for how to replace the functions but perhaps that's not the best approach? Wanted to check in about this to make sure I handle this consistently with how we want to in the future (and go back to fix other sections of mine that don't reference the mega file if we decide to go down that route).

aveline: I would link the function definition in the mega file. Then consumers have the option to decide on which approach they want to take.

aveline (approved):

PR #5165, opened 2022-02-11
[TextField][SearchField] Explicitly set opacity for placeholder elements

rgelera:
Firefox sets opacity of placeholder elements to 50%. This can be confirmed by going to https://polaris.shopify.com/components/forms/text-field and choosing a placeholder textfield. Compare in both Firefox and Google Chrome. This causes inconsistencies between browsers.

To solve this, we explicitly set the opacity to 1 for components using `::placeholder`. 

### TextField
Before:
textfield before
After:
textfield after

### TopBar using SearchField
Before:
searchfield before
After:
searchfield after


ghost:

chloerice (approved):

PR #5214, opened 2022-02-24
Add `polaris-for-vscode` extension to project

yurm04:

Moving the `polaris-intellisense` VSCode extension from its own repo into the `polaris-react` repo. This helps:

1. Keep things all in one place
2. Gives the plugin direct access to the `polaris-react` tokens
3. Reduces the number on new repos we need to maintain
 
To test the plugin, follow the README instructions in this PR


polaris-intellisense/src/data/allTokens.ts
```diff
@@ -0,0 +1,260 @@
+export const allTokens = {
+  "shadow-transparent": "0 0 0 0 transparent",
```
yurm04: need to get rid of this and just pull tokens directly from the `tokens` directory

polaris-intellisense/package.json
```diff
@@ -0,0 +1,37 @@
...
+	"description": "Polaris Design System intelligent tooling",
+	"version": "0.0.1",
+	"engines": {
+		"vscode": "^1.64.0"
+	},
+	"categories": [
+		"Other"
+	],
+	"activationEvents": [
+		"*"
```
yurm04: should find a different activation event instead of `all` for performance

polaris-intellisense/src/extension.ts
```diff
@@ -0,0 +1,35 @@
...
+		}
+	]
+
+
+	context.subscriptions.push(
+		vscode.languages.registerCompletionItemProvider(
+			selector,
+			new PolarisAutocomplete(),
+			// trigger characters
+			'*'
```
yurm04: are there any specific trigger characters we want to designate for adding tokens? 🤔

polaris-intellisense/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+  "compilerOptions": {
+    "module": "commonjs",
+    "target": "ES2020",
+    "outDir": "out",
+    "lib": ["ES2020"],
+    "skipLibCheck": true,
```
yurm04: @aaronccasanova 's suggestion to get this working within the `polaris-react` project 🙌

polaris-intellisense/README.md
```diff
@@ -0,0 +1,23 @@
+# shopify-dev-intellisense README
+
+:warning: This repo is still experimental
```
aaronccasanova:
```suggestion
:warning: This extension is still experimental
```

polaris-intellisense/out/Polaris/PolarisAutocomplete.js
```diff
@@ -0,0 +1,36 @@
+"use strict";
```
aaronccasanova: Think we need to add `out` to the `polaris-intellisense/.gitignore`

polaris-intellisense/README.md
yurm04: thanks for catching this!!

polaris-intellisense/out/Polaris/PolarisAutocomplete.js
yurm04: good callout thank you 🙏

polaris-intellisense/package.json
aaronccasanova: [onLanguage](https://code.visualstudio.com/api/references/activation-events#onLanguage) might be good for starters. e.g.
```
    "onLanguage:javascript",
    "onLanguage:typescript",
    "onLanguage:css",
```

polaris-intellisense/src/data/allTokens.ts
aaronccasanova: Agreed! Perhaps, we can create an [npm script](https://github.com/Shopify/polaris-react/blob/9a04c4fabec2c0c2ca7debb9ea392cc2ae5cd5c6/.stylelintrc.js#L4-L22) for this (e.g. `yarn get-polaris-tokens`) and run it to populate the tokens at various stages of the development/publish cycle:
- postinstall
- build
- prepublish

polaris-intellisense/src/extension.ts
```diff
@@ -0,0 +1,35 @@
...
+
+
+	// Use the console to output diagnostic information (console.log) and errors (console.error)
+	// This line of code will only be executed once when your extension is activated
+	console.log('Congratulations, your extension "shopify-dev-intellisense" is now active!');
+
+	const selector: vscode.DocumentSelector = [
+		{
+			// only use plugin in css and sass files
+			pattern: '**/*.{css,scss}'
```
aaronccasanova: If we decide to move forward with the `onLanguage` `activationEvents` suggestion above, let's make sure to keep this in sync.

polaris-intellisense/src/extension.ts
aaronccasanova: Maybe `--` or `--p`

utils/generateTokensArray.js
```diff
@@ -0,0 +1,28 @@
+const fs = require('fs');
+const path = require('path');
+
+const tokenGroupsDir = path.join(__dirname, '../src/tokens/token-groups');
+
+const generateTokensArray = () => {
```
yurm04: copied and generalized the token generation from the `stylelint` project. That can be refactored to use this utility function at some point as a followup

utils/intellisensePrepublish.js
```diff
@@ -0,0 +1,18 @@
+const fs = require('fs');
+const path = require('path');
+
+const generateTokensArray = require('./generateTokensArray');
+
+// if the file is being run directly
+const generateTokensForPrepublish = () => {
```
yurm04: this is a utility function wrapper around the `generateTokensArray` function above. This will generate the tokens and then create a file with the tokens in the intellisense project. This utility gets run during the intellisense pre-packaging step so that it will always have the most up to date version of the tokens

polaris-intellisense/.vscode/extensions.json
```diff
@@ -0,0 +1,7 @@
+{
+	// See http://go.microsoft.com/fwlink/?LinkId=827846
+	// for the documentation about the extensions.json format
+	"recommendations": [
+		"dbaeumer.vscode-eslint"
+	]
+}
```
alex-page: Could this live in the parent directory? Unsure why it's needed at this layer.

polaris-intellisense/.vscode/settings.json
```diff
@@ -0,0 +1,11 @@
+// Place your settings in this file to overwrite default and user settings.
+{
+    "files.exclude": {
+        "out": false // set this to true to hide the "out" folder with the compiled JS files
+    },
+    "search.exclude": {
+        "out": true // set this to false to include "out" folder in search results
+    },
```
alex-page: `out` is a part of the `.gitignore` is there a reason it's a problem to show this folder? It should be excluded in search.

polaris-intellisense/.vscode/settings.json
```diff
@@ -0,0 +1,11 @@
...
+    // Turn off tsc task auto detection since we have the necessary tasks as npm scripts
+    "typescript.tsc.autoDetect": "off"
```
alex-page: I love the auto detect. Any context why we would only want to run this through scripts or the tasks.json file?

polaris-intellisense/.eslintrc.json
```diff
@@ -0,0 +1,24 @@
...
+        "eqeqeq": "warn",
+        "no-throw-literal": "warn",
+        "semi": "off"
+    },
+    "ignorePatterns": [
+        "out",
+        "dist",
+        "**/*.d.ts"
+    ]
+}
```
alex-page: This file is pretty small. Could we replace it with some shopify eslint config for typescript packages? We might have already tried...

polaris-intellisense/.yarnrc
```diff
@@ -0,0 +1 @@
+--ignore-engines true
```
alex-page: Unsure what this does... if it's about the engines in package.json this doesn't feel great. Any context?

polaris-intellisense/README.md
```diff
@@ -0,0 +1,23 @@
+# shopify-dev-intellisense README
```
alex-page: As we continue to build additional vscode plugins, do we imagine them being all contained in this one plugin or having multiple small ones? I ask because the approach for Figma is call the plugin "Polaris for Figma" and put everything in one place until it becomes a problem.

Should we follow a similar approach here?

polaris-intellisense/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+:warning: This extension is still experimental
+
+This project is a prototype testing out VSCode intellisense features to support developing with the Polaris Design System.
```
alex-page:
```suggestion
:warning: This extension is an experimental prototype

VSCode intellisense features to support developing with the Polaris Design System.
```

polaris-intellisense/README.md
yurm04: I similarly envisioned having one main polaris vscode plugin that would include any new features we came up with (rather than multiple plugins), so let's go with that!

I'll update this line as it's a holdover from when I copy pasta'd from the original prototype 😆 🍝

polaris-intellisense/.vscode/extensions.json
yurm04: i don't think it has to be! a lot of these setup files were included with the boilerplate vscode extension proj template. This recommended extension already exists in the polaris-react project so we can safely delete

polaris-intellisense/.vscode/settings.json
yurm04: no context, this is boilerplate code from the template. Happy to remove if we feel strongly about it, but fwiw I've not had any issues with it so far

polaris-intellisense/.vscode/settings.json
yurm04: more boilerplate so this is a guess: it's common to hide compiled files from search and and the file explorer. I think this is a personal preference (i usually choose to hide), but I think some folks do prefer to have it. We can change it here as these settings can be changed by the user at will

polaris-intellisense/.eslintrc.json
yurm04: i haven't tried, but down to! I'll need a bit of guidance/pairing since I haven't done this yet

polaris-intellisense/.yarnrc
yurm04: i also have no idea what this does and the only documentation i could find was a [reddit comment](https://www.reddit.com/r/node/comments/o9j87e/what_does_ignoreengines_flag_do/) specifically about npm. Not sure why we would want to use this, it was included in the boilerplate 🤷 

`engines` in package.json is specifically for the version of vscode that this extension can run on, so that doesn't feel related...

polaris-intellisense/.eslintrc.json
aaronccasanova: Should be able to extend the root `polaris-react/.eslintrc` and add/remove some properties. Let me know if you want to pair on that.

polaris-intellisense/.vscode/settings.json
aaronccasanova: We recently moved away from this [convention of hiding files](https://github.com/Shopify/polaris-react/pull/5187/files) in `polaris-react` in favor of the VSCode defaults. I support this decision and feel we shouldn't have an opinion here.

polaris-intellisense/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+1. (Install the `vsce` package)[https://code.visualstudio.com/api/working-with-extensions/publishing-extension#vsce]
+2. `cd` to the `polaris-intellisense` directory in the project.
+3. From the `polaris-intellisense` directory run the (`vsce package`)[https://code.visualstudio.com/api/working-with-extensions/publishing-extension#packaging-extensions] command in your terminal. After running you should have a new `.vsix` file in the directory (ex. `polaris-intellisense-0.0.1.vsix`)
+4. To install the extension run `code --install-extension EXTENSION_FILENAME.vsix`
+
+```bash
+# example of running install command
+❯ code --install-extension shopify-dev-intellisense-0.0.1.vsix
+```
```
aaronccasanova, line 19:
```suggestion
```sh
# example of running install command
code --install-extension shopify-dev-intellisense-0.0.1.vsix
```
```

Suggestion leaving off the `❯` character. Makes copy/paste less convenient IMO

File: polaris-intellisense/README.md [hunk 18]
```diff
@@ -0,0 +1,23 @@
...
+We don't want to publish this extension to the VSCode Extensions Marketplace yet, so we have to install the package directly from the compiled `.vsix` file.
+
+1. (Install the `vsce` package)[https://code.visualstudio.com/api/working-with-extensions/publishing-extension#vsce]
+2. `cd` to the `polaris-intellisense` directory in the project.
+3. From the `polaris-intellisense` directory run the (`vsce package`)[https://code.visualstudio.com/api/working-with-extensions/publishing-extension#packaging-extensions] command in your terminal. After running you should have a new `.vsix` file in the directory (ex. `polaris-intellisense-0.0.1.vsix`)
+4. To install the extension run `code --install-extension EXTENSION_FILENAME.vsix`
+
+```bash
+# example of running install command
+❯ code --install-extension shopify-dev-intellisense-0.0.1.vsix
```
aaronccasanova:
```suggestion
code --install-extension shopify-dev-intellisense-0.0.1.vsix
```

Suggestion leaving off the `❯` character. Makes copy/paste less convenient IMO

polaris-intellisense/src/Polaris/PolarisAutocomplete.ts
```diff
@@ -0,0 +1,26 @@
+import * as vscode from 'vscode';
+import {allTokens} from '../data/allTokens';
+
+export class PolarisAutocomplete implements vscode.CompletionItemProvider {
+  get allTokens() {
+    return allTokens;
+  }
```
aaronccasanova:
```suggestion
  allTokens = allTokens;
```

Does this need to be a getter?

polaris-intellisense/src/Polaris/PolarisAutocomplete.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  async provideCompletionItems(
+    document: vscode.TextDocument,
+    position: vscode.Position,
+    token: vscode.CancellationToken,
+    context: vscode.CompletionContext,
+  ): Promise<vscode.CompletionItem[] | null | undefined> {
+    // iterate over all tokens and create completion items
+    let completionItems = allTokens.map((token) => {
```
aaronccasanova:
```suggestion
    const completionItems = allTokens.map((token) => {
```

polaris-intellisense/src/data/allTokens.ts
```diff
@@ -0,0 +1 @@
...order-radius-3","--p-border-radius-4","--p-border-radius-5","--p-border-radius-6","--p-border-radius-base","--p-border-radius-large","--p-border-radius-half","--p-border-width-1","--p-border-width-2","--p-border-width-3","--p-border-base","--p-border-dark","--p-border-transparent","--p-border-divider","--p-space-0","--p-space-025","--p-space-05","--p-space-1","--p-space-2","--p-space-3","--p-space-4","--p-space-5","--p-space-6","--p-space-8","--p-space-10","--p-space-12","--p-space-16","--p-space-20","--p-space-24","--p-space-28","--p-space-32","--p-font-family-sans","--p-font-family-mono","--p-font-size-1","--p-font-size-2","--p-font-size-3","--p-font-size-4","--p-font-size-5","--p-font-size-6","--p-font-size-7","--p-font-size-8","--p-font-size-9","--p-font-size-10","--p-font-size-11","--p-font-size-12","--p-font-size-13","--p-font-weight-regular","--p-font-weight-medium","--p-font-weight-semibold","--p-font-weight-bold","--p-line-height-1","--p-line-height-2","--p-line-height-3","--p-line-height-4","--p-line-height-5","--p-line-height-6","--p-line-height-7","--p-z-1","--p-z-2","--p-z-3","--p-z-4","--p-z-5","--p-z-6","--p-z-7","--p-z-8","--p-z-9","--p-z-10","--p-z-11","--p-z-12"];
```
aaronccasanova: Curious what folks think, but I was thinking we don't commit this file and instead run the `generateTokens` script on `postinstall`/`prepublish`. We took a similar approach in `polaris-react` before `v9` and it ensured we always have the latest tokens when releasing.

polaris-intellisense/src/Polaris/PolarisAutocomplete.ts
aaronccasanova: I'm actually not seeing this instance prop used anywhere. Is it needed?

polaris-intellisense/src/extension.ts
```diff
@@ -0,0 +1,35 @@
+// The module 'vscode' contains the VS Code extensibility API
```
aaronccasanova: **suggestion/nit (non-blocking):** Can we tidy up the double newlines throughout the file?

utils/generateTokensArray.js
aaronccasanova: Awesome!

**suggestion (non-blocking):** Small nit but stylistically I try to avoid using data type names in vars and functions. What do you think about `getCustomProperties()` or `getCustomPropertyNames()`. I kind of like the latter (`getCustomPropertyNames`) as `getCustomProperties` could be confused for key/value pairs.

utils/generateTokensArray.js
```diff
@@ -0,0 +1,28 @@
...
+  /**
+   * Allowed Polaris token custom properties.
+   *
+   * Result: ['--p-background', '--p-text', etc...]
+   */
```
aaronccasanova: **suggestion:** Should we move this code block above the `generateTokensArray` util?

utils/intellisensePrepublish.js
```diff
@@ -0,0 +1,18 @@
+const fs = require('fs');
```
aaronccasanova: **suggestion:** How do you feel about moving this inside the `polaris-intellisense` directory?

I'm think `polaris-intellisense/scripts/prepublish.js` would be sufficient.

utils/intellisensePrepublish.js
aaronccasanova: Looks great! Although, I don't think we need this to be a function. We can just execute this script top to bottom and leverage top level await if needed.

utils/intellisensePrepublish.js
```diff
@@ -0,0 +1,18 @@
...
+const path = require('path');
+
+const generateTokensArray = require('./generateTokensArray');
+
+// if the file is being run directly
+const generateTokensForPrepublish = () => {
+  const outDir = '../polaris-intellisense/src/data/allTokens.ts';
+
+  const tokens = generateTokensArray(outDir);
```
aaronccasanova:
```suggestion
  const tokens = generateTokensArray();
```

`generateTokensArray` currently doesn't accept any input

polaris-intellisense/.yarnrc
alex-page: Lets remove this file.

polaris-intellisense/src/data/allTokens.ts
alex-page: I would hope the reason for it being in this repository is so we don't need to have the list duplicated. Is there any reason we can't read the JSON files and convert them into custom properties?

aaronccasanova: > Is there any reason we can't read the JSON files and convert them into custom properties?

We'll be able to do that when we reinstantiate `@shopify/polaris-tokens`, but the plugin won't have access to the tokens when deployed to the VSCode market place. e.g. `../../../src/tokens/token-groups/spacing.json` will not be accessible.

@yurm04 and I experimented with different TSC configurations and the bundle always resulted in relative paths to the token group files.

aaronccasanova: One Google search put me on the right track and I think I figured out an alternative approach: https://github.com/Shopify/polaris-react/compare/add-polaris-intellisense..polaris-intellisense-experiment?diff=split

I tweaked the `tsconfig.json` to point to the root of `polaris-react` and reworked the custom property utils to directly import the token files. This allowed TSC to correctly build the module tree and output the appropriate files:
image

.eslintrc
```diff
@@ -87,6 +88,23 @@
         "@shopify/jsx-no-hardcoded-content": "off",
         "@shopify/react-initialize-state": "off"
       }
+    },
+    {
```
yurm04: @alex-page @aaronccasanova removed the .eslintrc and moved contents to this overrides section in the root dir

.eslintrc
```diff
@@ -57,6 +57,7 @@
     "jsx-a11y/no-noninteractive-element-interactions": "off",
     "jsx-a11y/no-noninteractive-element-to-interactive-role": "off"
   },
+  "ignorePatterns": ["out", "dist", "**/*.d.ts"],
```
yurm04: ignorePatterns wasn't allowed in the overrides prop below.

polaris-for-vscode/.gitignore
```diff
@@ -0,0 +1,5 @@
+out
+dist
+node_modules
```
alex-page: I think the parent dir will ignore `node_modules` throughout all of them.

.eslintrc
alex-page: What happens if you just use the Polaris linter and add the path to the `files` above?

polaris-intellisense/src/data/allTokens.ts
yurm04: for the purposes of this experimentation phase I think keeping with the `generateCustomPropertyNames` script approach on prepublish suffices and is the easiest (because the work is already done 😄 ). The script will always run on prepublish of the package, so tokens should always be up to date whenever the project gets compiled. Since this is a temporary measure anyway until we can get our separate tokens package back in action, i think it's ok leaving things as they are for a bit

polaris-for-vscode/.vscode/tasks.json
```diff
@@ -0,0 +1,20 @@
...
+			"presentation": {
+				"reveal": "never"
+			},
+			"group": {
+				"kind": "build",
+				"isDefault": true
+			}
+		}
+	]
+}
```
alex-page: I think we can remove this file. From my understanding it is running tsc and has nothing to do with the plugin.

polaris-for-vscode/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# Change Log
```
alex-page:
```suggestion
# Changelog
```
For consistency with other repos

polaris-for-vscode/README.md
```diff
@@ -0,0 +1,23 @@
+# `polaris-for-vscode` Extension
+
+:warning: This extension is still an experimental prototype
+
+VSCode intellisense features to support developing with the Polaris Design System.
```
alex-page:
```suggestion
VSCode features to support developing with the Polaris Design System.
```

Removing intellisense as that is what it does now but it might do other things in the future.

polaris-for-vscode/README.md
```diff
@@ -0,0 +1,23 @@
...
+2. `cd` to the `polaris-for-vscode` directory in the project.
+3. From the `polaris-for-vscode` directory run the (`vsce package`)[https://code.visualstudio.com/api/working-with-extensions/publishing-extension#packaging-extensions] command in your terminal. After running you should have a new `.vsix` file in the directory (ex. `polaris-for-vscode-0.0.1.vsix`)
+4. To install the extension run `code --install-extension EXTENSION_FILENAME.vsix`
+
+```bash
+# example of running install command
+code --install-extension shopify-dev-intellisense-0.0.1.vsix
+```
+
+![Success message of installed package in the terminal](docs/sample-install.png)
```
alex-page, line 35: The images will need to be updated with the Polaris for VSCode.

polaris-for-vscode/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "polaris-for-vscode",
+  "displayName": "polaris-for-vscode",
+  "description": "Polaris Design System intelligent tooling",
```
alex-page:
```suggestion
  "description": "Polaris Design System vscode tooling",
```

polaris-for-vscode/src/data/allTokens.ts
```diff
@@ -0,0 +1 @@
...order-radius-3","--p-border-radius-4","--p-border-radius-5","--p-border-radius-6","--p-border-radius-base","--p-border-radius-large","--p-border-radius-half","--p-border-width-1","--p-border-width-2","--p-border-width-3","--p-border-base","--p-border-dark","--p-border-transparent","--p-border-divider","--p-space-0","--p-space-025","--p-space-05","--p-space-1","--p-space-2","--p-space-3","--p-space-4","--p-space-5","--p-space-6","--p-space-8","--p-space-10","--p-space-12","--p-space-16","--p-space-20","--p-space-24","--p-space-28","--p-space-32","--p-font-family-sans","--p-font-family-mono","--p-font-size-1","--p-font-size-2","--p-font-size-3","--p-font-size-4","--p-font-size-5","--p-font-size-6","--p-font-size-7","--p-font-size-8","--p-font-size-9","--p-font-size-10","--p-font-size-11","--p-font-size-12","--p-font-size-13","--p-font-weight-regular","--p-font-weight-medium","--p-font-weight-semibold","--p-font-weight-bold","--p-line-height-1","--p-line-height-2","--p-line-height-3","--p-line-height-4","--p-line-height-5","--p-line-height-6","--p-line-height-7","--p-z-1","--p-z-2","--p-z-3","--p-z-4","--p-z-5","--p-z-6","--p-z-7","--p-z-8","--p-z-9","--p-z-10","--p-z-11","--p-z-12"];
```
alex-page: Is this needed with the new scripts?

polaris-for-vscode/tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+    "outDir": "out",
+    "lib": ["ES2020"],
+    "skipLibCheck": true,
+    "sourceMap": true,
+    "rootDir": "src",
+    "strict": true /* enable all strict type-checking options */
+    /* Additional Checks */
+    // "noImplicitReturns": true, /* Report error when not all code paths in function return a value. */
+    // "noFallthroughCasesInSwitch": true, /* Report errors for fallthrough cases in switch statement. */
+    // "noUnusedParameters": true,  /* Report errors on unused parameters. */
```
alex-page:
```suggestion
    "strict": true
```

scripts/utilities/getCustomPropertyNames.js
```diff
@@ -0,0 +1,28 @@
...
+          return Object.keys(tokenGroup).map((token) => `--p-${token}`);
+        })
+        .flat(),
+    ),
+  );
+
+  return polarisTokenCustomProperties;
+};
+
+module.exports = getCustomPropertyNames;
```
alex-page: Should this live in the tokens area? I feel like I would expect to find this closer to the JSON files.

yurm04: i think it can go wherever we want. I put it here because this utilities dir already existed and to keep it somewhere "global" so others could access (like the stylelint plugin)

.eslintrc
aaronccasanova: I was thinking we keep the `polaris-react/polaris-for-vscode/.eslintrc.json` file, remove `"root": true` and other extraneous properties, and extend this root config. This will likely need adjustments, but what do you think about this:

`polaris-react/polaris-for-vscode/.eslintrc.json`
```json
{
  "extends": "../.eslintrc",
  "ignorePatterns": [
    "out",
    "dist",
    "**/*.d.ts"
  ] 
}
```

yurm04: @alex-page when I remove/move the plugin eslint config eslint starts to complain, and we lose auto-fix issues/formatting. Not sure why but using the root level .eslintrc doesn't seem to be doing what we want 🤔 

Extending the root config and keeping the plugin eslintrc as @aaronccasanova does seem to work as expected. I'll go with that unless there are any objections or concerns with that approach?

alex-page: Because auto-fixing stopped, I feel like it wasn't correctly inheriting the root level.

polaris-for-vscode/src/data/allTokens.ts
yurm04: Nope! took this out and added it to the .gitignore since this file will need to be generated now

polaris-for-vscode/.vscode/tasks.json
yurm04: This file lets the VSCode debugger know what to run when you use the `F5` run task shortcut (the `watch` script). removing this file broke shortcut 😢

romellogoodman (approved):
Lgtm! We can iterate and hook up to monorepo scripts/yarn workspaces in follow up work

polaris-for-vscode/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "polaris-for-vscode",
```
BPScott, line 2: Do we want this to be a package that is published on npm? If not then we should add `private:true` here

polaris-for-vscode/.eslintrc
```diff
@@ -0,0 +1,18 @@
+{
```
BPScott, line 1: It seems strange to have linting on a per package basis. Linting config should occur once at the root of the repo.

polaris-for-vscode/.vscode/launch.json
```diff
@@ -0,0 +1,28 @@
+// A launch configuration that compiles the extension and then opens it inside a new window
```
BPScott, line 1: I don't think these .vscode configs have any effect unless you've got the polaris-for-vscode folder open in vscode.

Should this .vscode folder be at the root of the repository?

polaris-for-vscode/yarn.lock
```diff
@@ -0,0 +1,1381 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
BPScott, line 1: You don't need to commit the yarn.lock for workspace projects

BPScott (commented):
Do you need to update the [workspaces key in the root package.json](https://github.com/Shopify/polaris/blob/main/package.json#L7) so that yarn knows about this sub project, and thus install its subdependencies and such?


(This is the first time somebody has forgotten to do this, I'll keep track and if it happens a few more times I'll say "people keep forgetting this, perhaps packages should live in a packages folder and the value in package.json should be `packages/*`" :p)

polaris-for-vscode/package.json
yurm04: good call, we'll be packaging this up for the vscode marketplace so there is a separate workflow for that

polaris-for-vscode/.vscode/launch.json
yurm04: I left this in because the dev workflow involves the vscode debugger that gets set up in the `launch.json` and `tasks.json` file. I've been developing locally by having only the `polaris-for-vscode` directory open, otherwise the vscode debugger shortcuts don't work as expected. There's probably a way to set it up so there are tasks in the root `.vscode` file but in my experience the vscode debugger is very fickle and I didn't want to waste time trying to wrangle it together. 

Unless it's a big issue, I'd prefer to leave this as is for now and come back to it if/when it becomes a problem

polaris-for-vscode/.eslintrc
yurm04: there isn't currently an `.eslintrc` in the root of the repo so I'll punt this until we have a strategy on how we want to tackle linting in this project. Will create a ticket to capture

PR #5216, opened 2022-02-24
[TextField]: Add selectAllTextOnFocus boolean prop to allow all `input` / `textarea` field to be selected on focus

### LINKED ISSUE #5192: [TextField]: Add select all text on focus boolean prop
## Feature request summary

Noticed there's a couple of instances in web where we want the entire input textfield text selected onfocus.

Example behaviour of all text selected (seen in Figma): 

https://user-images.githubusercontent.com/3300102/152183520-e8ee1419-dd35-4993-b722-514cd594ec35.mov

It'd be super helpful to have a boolean prop that would select the entire text on focus if the boolean was true, so something like: 

```JavaScript

<TextField selectAllText ... /> 

```

## Rationale

---

ginabak:

Fixes #5192
Part of https://github.com/Shopify/inventory-states-ux/issues/392

Adds a `selectTextOnFocus` boolean prop that allows all TextField text to be entirely selected! (see issues for video behaviour) 

<details>
<summary>Summary of your gif(s)</summary>

https://user-images.githubusercontent.com/43223543/155763293-32dd76cf-b466-48ec-8e1b-a77410117537.mov

</details>

### How to 🎩

See the copy-paste code below! 💌 

<details>

```jsx

  const [textFieldValue, setTextFieldValue] = useState('Jaded Pixel');

  const handleTextFieldChange = useCallback(
    (value) => setTextFieldValue(value),
    [],
  );

  return (
    <Page title="Playground">
      <TextField
        label="Store name"
        value={textFieldValue}
        onChange={handleTextFieldChange}
        selectAllTextOnFocus
        autoComplete="off"
      />
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -6,6 +6,8 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 
 ### Enhancements
 
+- Added the `selectTextOnFocus` prop to `TextField` ([#5216](https://github.com/Shopify/polaris-react/pull/5216))
+
 ### Bug fixes
 
 ### Documentation
--- a/src/components/TextField/README.md
+++ b/src/components/TextField/README.md
@@ -789,6 +789,32 @@ function TextFieldWithMonospacedFontExample() {
 }
 ```
 
+### Text field with the ability to select all text on focus
+
+<!-- example-for: web -->
+
+Use to select all text inside TextField on focus.
+
+```jsx
+function TextFieldWithSelectTextOnFocusExample() {
+  const [textFieldValue, setTextFieldValue] = useState('Jaded Pixel');
+
+  const handleTextFieldChange = useCallback(
+    (value) => setTextFieldValue(value),
+    [],
+  );
+
+  return (
+    <TextField
+      label="Store name"
+      value={textFieldValue}
+      onChange={handleTextFieldChange}
+      selectTextOnFocus
+    />
+  );
+}
+```
+
 ---
 
 ## Related components
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -136,6 +136,8 @@ interface NonMutuallyExclusiveProps {
   requiredIndicator?: boolean;
   /** Indicates whether or not a monospaced font should be used */
   monospaced?: boolean;
+  /** Indicates whether or not the entire input/text area text should be selected on focus */
+  selectTextOnFocus?: boolean;
 }
 
 export type TextFieldProps = NonMutuallyExclusiveProps &
@@ -190,6 +192,7 @@ export function TextField({
   onBlur,
   requiredIndicator,
   monospaced,
+  selectTextOnFocus,
 }: TextFieldProps) {
   const i18n = useI18n();
   const [height, setHeight] = useState<number | null>(null);
@@ -198,7 +201,7 @@ export function TextField({
 
   const id = useUniqueId('TextField', idProp);
 
-  const inputRef = useRef<HTMLElement>(null);
+  const inputRef = useRef<HTMLInputElement | HTMLTextAreaElement>(null);
   const prefixRef = useRef<HTMLDivElement>(null);
   const suffixRef = useRef<HTMLDivElement>(null);
   const buttonPressTimer = useRef<number>();
@@ -406,6 +409,15 @@ export function TextField({
     monospaced && styles.monospaced,
   );
 
+  const handleOnFocus = () => {
+    if (selectTextOnFocus) {
+      inputRef.current?.select();
+    }
+    if (onFocus) {
+      onFocus();
+    }
+  };
+
   const input = createElement(multiline ? 'textarea' : 'input', {
     name,
     id,
@@ -415,7 +427,7 @@ export function TextField({
     autoFocus,
     value: normalizedValue,
     placeholder,
-    onFocus,
+    onFocus: handleOnFocus,
     onBlur,
     onKeyPress: handleKeyPress,
     style,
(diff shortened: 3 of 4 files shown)


src/components/TextField/TextField.tsx
```diff
@@ -415,7 +427,7 @@ export function TextField({
     autoFocus,
     value: normalizedValue,
     placeholder,
-    onFocus,
+    onFocus: handleOnFocus,
```
voiid, line 430: No opinions on this at the moment but I'm curious if there are any pros/cons of having this `onFocus` vs `onClick`, particularly if we compare the experience for mouse based vs touch based devices etc?

voiid (approved):
🎩 worked great! 😄  Changes look good to me - did have one question.

src/components/TextField/TextField.tsx
ginabak: oooh good point, is there a way to test on a touch based device ? 😯 😄

src/components/TextField/TextField.tsx
```diff
@@ -406,6 +409,15 @@ export function TextField({
     monospaced && styles.monospaced,
   );
 
+  const handleOnFocus = () => {
+    if (selectAllTextOnFocus) {
+      inputRef?.current?.select();
```
vinay30: Question about the use of optional chaining here - would this handler need them if we know `inputRef` is defined, and `current` is also defined when onFocus is called? 

I could be wrong! :)

src/components/TextField/README.md
```diff
@@ -789,6 +789,32 @@ function TextFieldWithMonospacedFontExample() {
 }
 ```
 
+### Text field with the ability to select all text on focus
+
+<!-- example-for: web -->
+
+Use to select all text inside TextField on focus
```
LauraAubin: The tiniest nit!

```suggestion
Use to select all text inside TextField on focus.
```

src/components/TextField/TextField.tsx
ginabak: Great question! If you look at where inputRef is defined: 

```typescript
  const inputRef = useRef<HTMLInputElement | HTMLTextAreaElement>(null);

```

It passes `null` as the default value because `useRef` can only be null or in this case an Input or TextArea element! This means that `inputRef` can be null so Typescript catches that for us and yells! Hehe. So the optional chaining is needed here 🥳

LauraAubin (approved):

src/components/TextField/TextField.tsx
```diff
@@ -136,6 +136,8 @@ interface NonMutuallyExclusiveProps {
   requiredIndicator?: boolean;
   /** Indicates whether or not a monospaced font should be used */
   monospaced?: boolean;
+  /** Indicates whether or not the entire input/text area text should be selected on focus */
+  selectAllTextOnFocus?: boolean;
```
AndrewMusgrave: What are you thoughts on changing this to `selectTextOnFocus`? From `Select text`, I would assume all text is selected!

src/components/TextField/TextField.tsx
AndrewMusgrave: Refs are a little tricky, the default value is `null`, but the default value is the `current` value. The type for `useRef` would look something like this

```ts
type UseRef<T> = (val: T) => {current: T};
```

Because of this, `inputRef` is always defined. But the current at `inputRef.current` defaults to null 😄 

```suggestion
      inputRef.current?.select();
```

src/components/TextField/TextField.tsx
AndrewMusgrave: >oooh good point, is there a way to test on a touch based device ? 😯 😄

Fastest way would be the dev tools. Emulating mobile devices should allow touch events

src/components/TextField/tests/TextField.test.tsx
```diff
@@ -1390,33 +1390,76 @@ describe('<TextField />', () => {
   });
 
   describe('monospaced', () => {
-    it('passes monospaced prop to TextField', () => {
-      const element = mountWithApp(
+    it('applies the monospaced style', () => {
```
AndrewMusgrave, line 1393: Interesting, looks like the test wasn't doing anything before, thanks for fixing that! Not sure if there's a lot of value in testing the `className` though since that's an implementation detail of the component.

UNRELEASED.md
```diff
@@ -8,6 +8,8 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 
 ### Enhancements
 
+- Added support to `selectAllTextOnFocus` (boolean prop) to `TextField` ([#5216](https://github.com/Shopify/polaris-react/pull/5216))
```
```suggestion
- Added the `selectAllTextOnFocus` prop to `TextField` ([#5216](https://github.com/Shopify/polaris-react/pull/5216))
```

This seems like the general entry for new props but I don't hold any strong opinions

AndrewMusgrave (approved):
Left a few comments but other than that LGTM! Nice contribution 🎉 🎉

src/components/TextField/TextField.tsx
ginabak: OOH that's much better. I like that!

UNRELEASED.md
ginabak: THANK YOUU!!!

src/components/TextField/TextField.tsx
ginabak: OOOH!!! Thank-you for this!

vinay30: lttp ;p 

Looks good! Thanks! @ginabak and @AndrewMusgrave

PR #5225, opened 2022-03-01
Use `//` instead of `/* */` comments in legacy api

BPScott:

[Output modes of sass other than `compressed`  retain `/* */` comments](https://sass-lang.com/documentation/syntax/comments). This means that
consumers can potentially end up with all these comments outputted in their css
files, which is ugly and bloats bundle size. Using `//` comments means
these comments shall be always removed regardless of the sass output
mode.

I found this issue in global-nav when [testing migration to dart sass & postcss v8](https://github.com/Shopify/global-nav/pull/960).

Replaces usage of `/* */` comments in `documentation/guides/legacy-polaris-v8-public-api.scss` with `//` comments, that shall always be removed from sass output.

### DIFF
--- a/documentation/guides/legacy-polaris-v8-public-api.scss
+++ b/documentation/guides/legacy-polaris-v8-public-api.scss
@@ -1,6 +1,8 @@
 // Deprecated scss functions and mixins
 
-/* Polaris Tokens: colors.color-map */
+//
+// Polaris Tokens: colors.color-map
+//
 
 $polaris-colors: (
   'purple': (
@@ -87,7 +89,9 @@ $polaris-colors: (
   ),
 );
 
-/* Polaris Tokens: color-filters.color-map */
+//
+// Polaris Tokens: color-filters.color-map
+//
 
 $polaris-color-filters: (
   'purple': (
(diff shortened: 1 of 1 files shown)


sam-b-rose (approved):
Great catch! Thanks for updating this Ben. I'll make a note to make this change in any other project where we have added the legacy API file 👍

alex-page (commented):
@BPScott curious why compressed mode is not working for global-nav? Is it a dart-sass thing?

BPScott:
It's a combo of sass thing and a postcss config thing. We run styles through sass and then through postcss. We don't output sass in the compressed style as the [compact/expanded](https://inchoo.net/dev-talk/tools/sass-output-styles/) style is bit more human readable for debugging. Stuff in npm modules don't need to be optimally compressed as they'll almost certainly be ran through a bundler to compress them within the consuming app. So this isn't critical for size in consuming apps but it makes the output within the library's build nicer and less suprising.

The original way of doing this was with node-sass and `@shopify/postcss-config` v4 (and thus postcss v7). We output sass in the compact style (which retained `/*` comments). Then postcss included a `postcss-discard-comments` plugin that threw away all comments. So you got no comments in your output.

The updated way of doing this is with dart-sass and `@shopify/postcss-config` v5 (and thus postcss v8). We output sass in the expanded style (which retained `/*` comments, dart-sass dropped supportfor the compact style). But out new postcss config dropped the `postcss-discard-comments` plugin, as it became part of cssnano - which would get ran as part of css minification in a consuming app. Which meant that these `/* */` comments stayed present in library build output

Issue #4725, opened 2021-11-24
Avoid dart-sass deprecation warnings


BPScott:
I've been starting to think about leveraging dart-sass. I've updated [loom-plugin-build-extended to use dart-sass](https://github.com/Shopify/loom/pull/295) for package builds, and when attempting to [use that in global-nav](https://github.com/Shopify/global-nav/pull/960) compiled output was as expected however it spammed the console with a bajillion deprecation notices regarding "use math.div instead of /" because using `/` to divide is deprecated in dart-sass.

Obviously this is a terrible user experience. It should be possible to use the public api with dart-sass without ruining console with incessant noise.

The gotcha comes here because polaris will need to support usage in consumers that use either `node-sass` or  `dart-sass`.

Per my comments on #4298, fixes (as provided by the [sass migrator](https://sass-lang.com/documentation/cli/migrator#division)) fall into one of two buckets:

- A fix that compiles in both `node-sass` or  `dart-sass`. Such as changing `$x: 8 / 2;` into `$x: 8 * 0.5`
- A fix that keeps quiets `dart-sass` but is a compilation error in `node-sass` Such as changing `$x: 8 / $var;` into `math.div(8, $var)`

It should be easily possible to apply the fixes that compile in both straight onto main. But that leaves the other set.

Fortunately in the public-api the only cases where uncompilable fixes occur is in `px()`, `rem()`, `em()` and `color-multiply()` all of which are slated for removal in v9.

Options for that:

- Do nothing, wait till v9 is released and deprecations go away (though there is no promise of new, different deprecate behavior creeping in later until at some point in the far future we drop support for consumers using node-sass)
- After [creating the `build/styles` folder](https://github.com/Shopify/polaris-react/blob/main/loom.config.ts#L117-L120), create a second `build/styles-dart` folder that contains the contents of src/styles, except ran through the [sass migrator](https://sass-lang.com/documentation/cli/migrator#division)) - thus converting deprecated behaviour.

BPScott:
Command to see what files would throw deprecation warnings: `npx sass-migrator division 'src/**/*.scss' --migrate-deps --load-path=. --dry-run`

At time of writing, running this against the v9 branch shows the following files:

- `src/styles/shared/_breakpoints.scss`. The one problem in our public API. Unfortunately its usage in consuming apps is pervasive. This isn't fixable till we rework how breakpoints are handled.
[...]

PR #5239, opened 2022-03-08
Replace divide operation with math.div in legacy Sass API file

### LINKED ISSUE #5237: Deprecation warning occuring in the legacy public Sass API mega file.
## Issue summary

See [Sass Breaking Changes: Slash as Division](https://sass-lang.com/documentation/breaking-changes/slash-div)

```
Deprecation Warning: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.

Recommendation: math.div($value, $default-browser-font-size)
```

## Expected behavior

Warning should not be occurring in build console. Replace usage of `/` with `math.div` in the legacy Sass public API file.

sam-b-rose:

Fixes #5237  

See [Sass Breaking Changes: Slash as Division](https://sass-lang.com/documentation/breaking-changes/slash-div)

Replaces the `/` operation with `math.div`. The `/` is deprecated in Dart Sass and causes a console warning in current versions of Node Sass.

### DIFF
--- a/documentation/guides/legacy-polaris-v8-public-api.scss
+++ b/documentation/guides/legacy-polaris-v8-public-api.scss
@@ -314,9 +314,9 @@ $base-font-size: 16px;
   } @else if $unit == 'rem' {
     @return $value;
   } @else if $unit == 'px' {
-    @return $value / $base-font-size * 1rem;
+    @return math.div($value, $base-font-size) * 1rem;
   } @else if $unit == 'em' {
-    @return $unit / 1em * 1rem;
+    @return math.div($unit, 1em) * 1rem;
   } @else {
     @error 'Value must be in px, em, or rem.';
   }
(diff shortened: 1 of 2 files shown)


alex-page (approved):

BPScott (changes requested):
Everything that uses sewing-kit (i.e. practically every shopify app) still uses node-sass, which has no support for math.div.

If we do this then this file won't work as a drop-in thing for any shopify project - it'll throw compilation errors, which seems bad.

We could either:
- Add a second migration file that is just for dart-sass (uugh duplication, and hard to keep in sync)
- Add some instructions on how to run `npx sass-migrator division 'path/to/legacy-polaris-v8-public-api.scss'` after copying the file into your project if your project uses dart-sass (better IMO)

See #4725

sam-b-rose:
Hmmm both options seem good. Just not sure which one seems best 🤷‍♂️

The dart file seems to be the most direct and simplest, but potential for getting out of sync if updates have to be made.
Any drawbacks of using the migrator? I think that might be the preferred suggestion to consumers using Dart Sass.

BPScott:
I've tried running `sass-migrator` over the legacy file in https://github.com/Shopify/global-nav/pull/960 (which switches global-nav from using node-sass to dart-sass) and it worked really well. It's self-contained and easy to run. I can't think of a drawback to using it.

I think teaching people that "this code is for node-sass, if you're using dart-sass it raises deprecation warnings, here's how to fix them with a single command" is pretty straight forward. I'd prefer that over having to maintain two close-to-identical copies of the scss file. Especially if our primary audience is folks using Shopify apps still use node-sass

BPScott (approved):

PR #5250, opened 2022-03-09
[KeypressListener] Adds support for `useCapture` and `options`

### LINKED ISSUE #5244: Add support for `options` and `useCapture` in `KeyPressListener`
## Feature request summary

- Add a new `useCapture` prop to the current implementation of [KeypressListener](https://github.com/Shopify/polaris-react/blob/v9/src%2Fcomponents%2FKeypressListener%2FKeypressListener.tsx) to allow registering events in the capture phase
- Add a new `options` prop to the current implementation of [KeypressListener](https://github.com/Shopify/polaris-react/blob/v9/src%2Fcomponents%2FKeypressListener%2FKeypressListener.tsx) to match the current `options` argument of [`EventTarget.addEventListener`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener)

## Rationale

- We currently have a fork of `KeypressListener` in `polaris-next`, stemming from [[Sheet, Modal] Escape key on child components always closes Sheet/Modal as well ](https://github.com/Shopify/polaris-react/issues/4253), so providing this from polaris would allow us to remove that forked version and provide better support to developers using it
- Both `useCapture` and `options` are platform features that we're obfuscating with he `KeyPressListener` abstraction. I think the cost of a slightly bigger API surface area is worth it for the added features.

---

jbalsas:
Fixes https://github.com/Shopify/polaris-react/issues/5244

This allows for users to be more specific about the underlying event mechanism and avoids obfuscating otherwise standard APIs

A case could be made for just allowing `options` since it supports the capture use case as well, but I decided to go with both for better symmetry with the native API

This should provide support for the workaround needed in https://github.com/Shopify/polaris-react/issues/4253 which in turn would allow us to remove the [KeypressListener fork in shopify/web](https://github.com/Shopify/web/blob/main/packages/%40shopify/polaris-next/components/KeypressListener/KeypressListener.tsx).

As noted in https://github.com/Shopify/polaris-react/issues/4253:

> [...] it would be ideal if this solution wasn't necessary.

Nevertheless, I think it can still be useful to surface this API for a relatively small price considering that `KeypressListener` is a relatively low level abstraction, close enough to the `addEventListener` concept.

It allows `useCapture` or `options` **but not both** to be passed as props of `KeypressListener` and be passed through to the `addEventListener` and `removeEventListener` underlying calls.

### How to 🎩

<details>

```jsx
import React from 'react';

import {Key, KeypressListener, Page} from '../src';

  const logKeypress = (event: Event) => {
    console.log({type: event.type, phase: event.eventPhase});
  };
  
  return (
    <Page title="Playground">
      <KeypressListener handler={logKeypress} keyCode={Key.KeyJ} />
      <KeypressListener handler={logKeypress} keyCode={Key.KeyH} options={{once: true}} />
      <KeypressListener handler={logKeypress} keyCode={Key.KeyN} useCapture={true} />
      <KeypressListener handler={logKeypress} keyCode={Key.KeyB} options={{capture: true}} />
    </Page>
  );
}
```

Checking the console output one could check that:
- Pressing `K` outputs logs normally with `eventPhase: 3`
- Pressing `H` only outputs to the log once
- With keys `N` and `B` log `eventPhase: 1`

</details>

### 🎩 checklist

### DIFF
--- a/UNRELEASED.md
+++ b/UNRELEASED.md
@@ -7,6 +7,7 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 ### Enhancements
 
 - Added `ReactNode` as an accepted prop type to `secondaryActions` on the `Page` component ([#5258](https://github.com/Shopify/polaris-react/pull/5258))
+- Added `useCapture` and `options` props in `KeypressListener` to allow passing through those options to the underlying `addEventListener` call ([#5221](https://github.com/Shopify/polaris-react/pull/5221))
 
 ### Bug fixes
 
--- a/src/components/KeypressListener/tests/KeypressListener.test.tsx
+++ b/src/components/KeypressListener/tests/KeypressListener.test.tsx
@@ -9,15 +9,21 @@ interface HandlerMap {
 }
 
 const listenerMap: HandlerMap = {};
+const listenerOptionsMap: HandlerMap = {};
 
 describe('<KeypressListener />', () => {
   beforeEach(() => {
-    jest.spyOn(document, 'addEventListener').mockImplementation((event, cb) => {
-      listenerMap[event] = cb;
-    });
+    jest
+      .spyOn(document, 'addEventListener')
+      .mockImplementation((event, cb, options) => {
+        listenerMap[event] = cb;
+        listenerOptionsMap[event] = options;
+      });
 
     jest.spyOn(document, 'removeEventListener').mockImplementation((event) => {
       listenerMap[event] = noop;
+
+      delete listenerOptionsMap[event];
     });
   });
 
@@ -45,6 +51,37 @@ describe('<KeypressListener />', () => {
 
     listenerMap.keyup({keyCode: Key.Escape});
     expect(spy).not.toHaveBeenCalled();
+  });
+
+  it('allows registering an event in the capture phase using useCapture', () => {
+    const spy = jest.fn();
+
+    mountWithApp(
+      <KeypressListener handler={spy} keyCode={Key.Escape} useCapture />,
+    );
+
+    expect(listenerOptionsMap.keyup).toBe(true);
+  });
+
+  it('passes additional options to the event', () => {
+    const spy = jest.fn();
+
+    const eventOptions = {
+      capture: true,
+      once: true,
+      passive: true,
+      signal: new AbortController().signal,
+    };
+
+    mountWithApp(
+      <KeypressListener
+        handler={spy}
+        keyCode={Key.Escape}
+        options={eventOptions}
+      />,
+    );
+
+    expect(listenerOptionsMap.keyup).toStrictEqual(eventOptions);
   });
 });
 
--- a/src/components/KeypressListener/KeypressListener.tsx
+++ b/src/components/KeypressListener/KeypressListener.tsx
@@ -3,11 +3,17 @@ import {useCallback, useEffect, useRef} from 'react';
 import {useIsomorphicLayoutEffect} from '../../utilities/use-isomorphic-layout-effect';
 import type {Key} from '../../types';
 
-export interface KeypressListenerProps {
+export interface NonMutuallyExclusiveProps {
   keyCode: Key;
   handler(event: KeyboardEvent): void;
   keyEvent?: KeyEvent;
 }
+
+export type KeypressListenerProps = NonMutuallyExclusiveProps &
+  (
+    | {useCapture?: boolean; options?: undefined}
+    | {useCapture?: undefined; options?: AddEventListenerOptions}
+  );
 
 type KeyEvent = 'keydown' | 'keyup';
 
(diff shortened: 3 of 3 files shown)


jbalsas:
cc @aswamy, @benwolfram, FYI since I think you pushed the forked component upstream. Could you double check, see if this would cover your use case and weigh in in case it doesn't? 🙇

benwolfram:
> cc @aswamy, @benwolfram, FYI since I think you pushed the forked component upstream. Could you double check, see if this would cover your use case and weigh in in case it doesn't? 🙇

I actually think Abraham (who's no longer at Shopify) added this in Online Store Web and then we needed it in web so we migrated it over here too, I would have to top hat this to check and I dont think I'll have time today, but I can try to get to it tomorrow!

src/components/KeypressListener/tests/KeypressListener.test.tsx
```diff
@@ -46,6 +52,37 @@ describe('<KeypressListener />', () => {
...
+  });
+
+  it('passes additional options to the event', () => {
+    const spy = jest.fn();
+
+    const eventOptions = {
+      capture: true,
+      once: true,
+      passive: true,
+      signal: new AbortController().signal,
```
jbalsas, line 73: I think this is totally unnecessary for this test and also for the `keyup|keydown` events, but I thought it would be interesting to show the full API here just in case. We can definitely simplify this to `{capture: true}` to make it simpler to read!

UNRELEASED.md
```diff
@@ -6,6 +6,8 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 
 ### Enhancements
 
+- Added `useCapture` and `options` props in `KeypressListener` to allow passing through those options to the underlying `addEventListener` call ([#5221](https://github.com/Shopify/polaris-react/pull/5221))
```
chloerice, line 10:
```suggestion
- Added `useCapture` and `options` props to `KeypressListener` to support passing options to the underlying `addEventListener` call ([#5221](https://github.com/Shopify/polaris-react/pull/5221))
```

chloerice (approved):
:shipit:

benwolfram:
@jbalsas I actually think our use case here must have been removed at some point, doesn't seem like we have a dependency on this anymore, so it should be good from my perspective.

jbalsas:
Thanks for checking, @benwolfram, do you mean that you no longer need `useCapture`? Would you know if that's because https://github.com/Shopify/polaris-react/issues/4253 has been fixed for you or because the component is no longer inside or around one of those (Modal, Sheet...)?

@chloerice, thanks for the review! Based on @benwolfram's assessment, would you still want to push this through if not really necessary? Truth is we have ample usage of `KeypressListener` and this was a bit of an edge case. I find the API useful (specially to support the `once` use case), but there's also value in keeping Polaris' surface API lean, so... up to you! :)

benwolfram:
Apologies, I can be a bit more clear / detailed with this.

My team is not using `useCapture` in Web anymore, but it looks like there is another use case in the repo by another team in MarketDetailsPricingItem.tsx.

It looks like `useCapture` is still used in Online Store Web. We migrated our work from OSW to Web for a subsequent project (including this fork), and at some point after `useCapture` was dropped during a refactor I guess. I do know for a fact that it was for the escape key thing mentioned in https://github.com/Shopify/polaris-react/issues/4253 so it was likely dropped for this reason.

However, Online Store Web is using an old version of Polaris last time I checked (I don't really work in this project anymore), so it is probably still required there since they wouldn't have access to mentioned in https://github.com/Shopify/polaris-react/issues/4253 until updating, but we should be able to get rid of that fork too if/when Polaris is updated.

jbalsas:
Hey @benwolfram, thanks so much for the clarification!

I was indeed trying to track down the usage of `MarketDetailsPricingItem.tsx` and wrongly assumed it was related to the creation of the forked version. Had I properly checked the file I should've easily found [Capture and stop esc key propagation](https://github.com/Shopify/web/commit/354e0f50b06dd89bf74c342546d5b449c2d0a601). Thanks so much for your willingness to check and provide additional context and apologies for dragging you into this! 🙏 

@benkovy, @mdpwilliams, @joykliu, could any of you please take a look here and tophat this change in the context of the [usage of KeypressListenerNext](https://github.com/Shopify/web/blob/main/app/sections/Markets/MarketDetails/components/MarketDetailsPricingList/components/MarketDetailsPricingItem/MarketDetailsPricingItem.tsx#L16) and validate if this change should do the trick? (I tried but couldn't easily locate it in the UI 🙈)

While you're at it, feel free to add any additional context to [[Sheet, Modal] Escape key on child components always closes Sheet/Modal as well](https://github.com/Shopify/polaris-react/issues/4253) in case you think it's related. Might help the team track that one down and raise its priority!

thank_you_all_brady

benkovy:
@jbalsas this looks like it does the trick - pulled this into a spin instance and use it within web 👍

jbalsas:
> @jbalsas this looks like it does the trick - pulled this into a spin instance and use it within web 👍

Awesome, thanks so much for checking @benkovy!! I finally managed to figure out how to test this in the UI, so once the Polaris release is out I'll make the necessary updates and send them your way!!

Since @chloerice already approved the changes... I'll wait for green and then...

shipping_time

PR #5258, opened 2022-03-10
[Page] Add ReactNode as an accepted secondaryActions prop value for customizability

tatianau:

The current `secondaryActions` on `Page` only allows for passing props as `MenuActionDescriptor[]` which doesn't allow for flexibility and may cause accessibility issues in some cases.

[Currently](https://user-images.githubusercontent.com/15054990/157740503-4bf1e454-71c8-400e-b47e-a359472385f1.png), `Export` secondary action in Discounts opens a modal. Ideally we should be able to move focus back to the button that opened it after the modal is closed to support keyboard users, whether it'd be through passing entire modal as secondary action to leverage modal's built-in focus management or through passing a custom button that'd able to have a `ref` to receive the focus. To make either of these and other secondary actions customizations possible, Page's secondaryActions has a need for `React.ReactNode` as another type. 

[Slack thread](https://shopify.slack.com/archives/C4Y8N30KD/p1646765805356699)

This PR maintains the existing `secondaryActions` prop, while adding the option to pass a `ReactNode` instead for customizability.

## 

### How to 🎩

<details>

```jsx

import {Button, Modal, Page, TextContainer} from '../src';

  return (
    <Page title="Playground" secondaryActions={<ModalExample />}>
      <p>Page content</p>
    </Page>
  );
}

function ModalExample() {

  const handleChange = useCallback(() => setActive(!active), [active]);

  const activator = <Button onClick={handleChange}>Open</Button>;

  return (
    <Modal
      activator={activator}
      open={active}
      onClose={handleChange}
      primaryAction={{
        content: 'Add Instagram',
        onAction: handleChange,
      }}
      secondaryActions={[
        {
          content: 'Learn more',
          onAction: handleChange,
        },
      ]}
    >
      <Modal.Section>
        <TextContainer>
          <p>
            Use Instagram posts to share your products with millions of people.
            Let shoppers buy from your store without leaving Instagram.
          </p>
        </TextContainer>
      </Modal.Section>
    </Modal>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/src/components/Page/components/Header/Header.tsx
+++ b/src/components/Page/components/Header/Header.tsx
@@ -1,4 +1,4 @@
-import React, {isValidElement} from 'react';
+import React from 'react';
 
 import {classNames} from '../../../../utilities/css';
 import {buttonsFrom} from '../../../Button';
@@ -20,6 +20,7 @@ import type {
 import {Breadcrumbs, BreadcrumbsProps} from '../../../Breadcrumbs';
 import {Pagination, PaginationProps} from '../../../Pagination';
 import {ActionMenu, hasGroupsWithActions} from '../../../ActionMenu';
+import {isInterface, isReactElement} from '../../utilities';
 
 import {Title, TitleProps} from './components';
 import styles from './Header.scss';
@@ -45,19 +46,13 @@ export interface HeaderProps extends TitleProps {
   /** Collection of breadcrumbs */
   breadcrumbs?: BreadcrumbsProps['breadcrumbs'];
   /** Collection of secondary page-level actions */
-  secondaryActions?: MenuActionDescriptor[];
+  secondaryActions?: MenuActionDescriptor[] | React.ReactNode;
   /** Collection of page-level groups of secondary actions */
   actionGroups?: MenuGroupDescriptor[];
   /** @deprecated Additional navigation markup */
   additionalNavigation?: React.ReactNode;
   // Additional meta data
   additionalMetadata?: React.ReactNode | string;
-}
-
-export function isPrimaryAction(
-  x: PrimaryAction | React.ReactNode,
-): x is PrimaryAction {
-  return !isValidElement(x) && x !== undefined;
 }
 
 const SHORT_TITLE = 20;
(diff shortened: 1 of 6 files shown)


src/components/Page/components/Header/Header.tsx
```diff
@@ -45,7 +46,7 @@ export interface HeaderProps extends TitleProps {
   /** Collection of breadcrumbs */
   breadcrumbs?: BreadcrumbsProps['breadcrumbs'];
   /** Collection of secondary page-level actions */
-  secondaryActions?: MenuActionDescriptor[];
+  secondaryActions?: MenuActionDescriptor[] | React.ReactNode;
```
lhoffbeck, line 49: [question] given that the prop is plural, should this be a list of `React.ReactNode` instead of a single node?

tatianau: Great question!

### About passing a list of `React.ReactNode`
When passing an array of `ReactNode`s, Polaris `<Page />`, `<Header />` and `<MenuAction />` components would have to handle the appearance of this list of components passed to it. And this is where it get tricky - `ReactNode` is kind of a black hole that’s only being used if none of the default `secondaryActions: MenuActionDescriptor[]` options work for the consumer (_which is rare_). So in that case, I’m thinking that allowing the consumer to take care of whatever custom actions they need to pass and how they want them to appear instead of giving them only _some_ flexibility by passing `ReactNode` but at the same time adding constraints by requiring it to be an array and as a result having to process/style the list on Polaris side and collapse them into actions dropdown when there are too many actions and going with the assumption that regardless of what kind of nodes consumers pass to it would work with the way actions are handled on Polaris side.

### About the prop name being plural
I can see how it may seem not intuitive, at the same time it’s up to consumers how many custom react nodes they want to pass to it, so it still could be multiple, could be one, and if we take `children` prop as an example - it’s also plural and of type `React.ReactNode` and does the same thing.


### Final thoughts
I’m leaning towards having the secondary actions API like this: 
- [CURRENT] Strict API as `MenuActionDescriptor[]` for actions being handled on Polaris side, laid out, collapsed on mobile, etc.
- [NEW] Open API as `React.ReactNode` to fully shift the appearance and rendering logic responsibility of secondary actions to consumers if they desire something custom and default `MenuActionDescriptor[]` actions don’t work for them

lhoffbeck: > if we take children prop as an example - it’s also plural and of type React.ReactNode and does the same thing 
That's a very good point 😄 

I think your approach makes sense, and agree that opening up the API to accept React components means pushing responsibility for how to handle layout changes to the consumer.

alex-page (approved):

PR #5262, opened 2022-03-11
monorepo - Initialize yarn workspaces and website and components packages

romellogoodman:
### What

Fixes: #5286

Implements tasks from: #5266

This PR begins the process of converting polaris-react into a monorepo. As a first step we are going to use [Yarn v1 Workspaces](https://classic.yarnpkg.com/lang/en/docs/workspaces/) to create two packages:

* polaris-react - The current [polaris-react components](https://www.npmjs.com/package/@shopify/polaris)
* secretpolaris.shopify.com - The documentation site that will eventually replace [http://polaris.shopify.com/](http://polaris.shopify.com/)

### Details

Since the bulk of this repo revolves around setting up the initial yarn workspaces and commands, a good place to start would be by reviewing the top level [package.json](https://github.com/Shopify/polaris-react/blob/monorepo-init/package.json) and [README.md](https://github.com/Shopify/polaris-react/blob/monorepo-init/README.md). The corresponding files for each package will be updated in future PRs. The github workflows for `ci.yml` and `size-limit.yml` were also updated to reflect that the polaris-react package has been moved.

### Feedback

The main area of feedback that I am looking for in this PR is alignment on package names, yarn commands and the general workflow. The previous exploration PRs ([mono branch](https://github.com/Shopify/polaris-react/pull/5186), [monorepo branch](https://github.com/shopify/polaris-react/compare/main...feat/monorepo)) used Lerna to help bootstrap and maintain linking so accomplishing this with only Yarn means npm scripts that are a little more verbose.


package.json
```diff
@@ -1,187 +1,22 @@
...
-  "esnext": "build/esnext/index.esnext",
-  "types": "build/ts/latest/src/index.d.ts",
-  "typesVersions": {
-    "<4.0": {
-      "build/types/latest/*": [
-        "build/ts/3.4/*"
-      ]
-    }
-  },
+  "name": "polaris-monorepo",
```
aaronccasanova:
```suggestion
  "name": "polaris",
```

**suggestion (non-blocking):** This repo will likely get renamed to `polaris` and I don't think we need to explicitly add a `monorepo` suffix. e.g. [shopify/quilt](https://github.com/Shopify/quilt/blob/ea3020f560b85a6af64bdc6aa39148e5adde74b3/package.json#L2)

package.json
```diff
@@ -1,187 +1,22 @@
...
-    "react-transition-group": "^4.4.2"
-  },
-  "peerDependencies": {
-    "react": "^16.14.0 || ^17.0.0",
-    "react-dom": "^16.14.0 || ^17.0.0"
+    "build": "yarn workspaces run build",
+    "dev:components": "yarn --cwd packages/components dev",
+    "dev:website": "yarn --cwd packages/website dev",
+    "build:components": "yarn --cwd packages/components build",
+    "build:website": "yarn --cwd packages/website build"
```
aaronccasanova:
```suggestion
    "dev:components": "yarn workspace components dev",
    "dev:website": "yarn workspace website dev",
    "build:components": "yarn workspace components build",
    "build:website": "yarn workspace website build"
```

**question:** Should we be using the `workspace` subcommand?

packages/components/dev.yml
```diff
@@ -0,0 +1,22 @@
+name: polaris-react
```
aaronccasanova: **thought:** Wonder if we should rename this to `polaris-components` to match the parent directory. Thoughts?

package.json
```diff
@@ -1,187 +1,22 @@
...
-    "<4.0": {
-      "build/types/latest/*": [
-        "build/ts/3.4/*"
-      ]
-    }
-  },
+  "name": "polaris-monorepo",
+  "private": true,
   "engines": {
+    "npm": ">=7.0.0",
```
BPScott: Why do we need to care that the user is using a particular version of npm, when we use yarn for our package management needs?

package.json
```diff
@@ -1,187 +1,22 @@
...
     "node": ">=16.13.0"
   },
+  "workspaces": [
+    "packages/*"
+  ],
   "scripts": {
-    "lint": "loom lint",
```
BPScott, line 50: Loom tasks should continue to live at the top level of the monorepo. This includes build commands, as it supports building everything in a monorepo (see quilt)

packages/components/dev.yml
BPScott: dev.yml should live at the root of the repo, along with all the editorconfig/eslint/stylelint/prettier/loom/tsconfig configuration files

packages/components/build-internal/storybook/static/services/ping.html
```diff
@@ -0,0 +1 @@
+<!DOCTYPE html><html lang="en"><head></head><body>OK</body></html>
```
BPScott: It looks like you've commited the generated build-internal folder. You'll want to update the gitignore to ignore the new build paths in packages/*/build and packages/components/build-internal

packages/website/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "polaris-shopify-com",
```
aaronccasanova: **thought:** `packages/website` might be too generic as there will likely be more websites hosted here. Wonder if we should be more intentional with the naming of the Polaris doc site. e.g. `packages/shopify`, [packages/docs](https://github.com/mui/material-ui/tree/master/docs), `packages/style-guide`, etc.

aaronccasanova (commented):
Nice work @romellogoodman!! Left a few thoughts and non-blocking suggestions. Very excited to see this develop 🙌

package.json
alex-page: I am unsure if that is the approach we want to take initially. We want to be flexible with our build tooling to start then become more opinionated when there are overlapping problems.

package.json
alex-page: I agree. This should match the repository name which will be `polaris`.

packages/components/dev.yml
alex-page: This is tricky. `polaris-react` has no link to the NPM package. Maybe there is something we can do to clean this up.

alex-page: +1 for `dev.yml` happy to step through what these files do @romellogoodman

packages/website/package.json
BPScott: +1 to packages/polaris.shopify.com

as Aaron says it leaves room for other stuff like polaris-icons.shopify.com in the future (currently living in github.com/shopify/polaris-icons)

packages/components/dev.yml
BPScott: This file shouldn't live here anyway - it should be at the root.

package.json
BPScott: I'm pretty sure that you'll want to run linters  over code in all packages and to have a single `yarn run test` that shall test code in every package :)

packages/website/package.json
alex-page: Probably going to get hate for this. I have a strong opinion that putting everything in `packages` is an extra click for everyone. Would prefer if we didn't use the packages approach. Right now we have:

- Tokens
- Components
- Icons
- polaris.shopify.com (let's call it secretpolaris.shopify.com until we are ready to replace the other site)
- Polaris for Figma
- Polaris for VSCode

I don't think the packages layer is needed for the next year. Just chuck everything in the parent directory.

aaronccasanova: I'd be interested to hear more about your concerns around using a `packages` directory to organize libs and services (beyond a strong opinion it adds an extra click).

Amongst all the monorepos and tools I've seen this is the most commonly shared convention. You've already enumerated 6 `packages` that would otherwise be floating around the top level directory. Being that file explorers and GitHub organize files/directories in alphabetical order the packages would literally be all over the place. I would argue this negatively impacts the understanding of the monorepo. As opposed to adding a line to the README saying all libs and services live in the `packages` directory. 

I think these conventions make approaching a new monorepo a familiar experience and promote seamless contributions (in the same way developers instinctively know to look for a `package.json` or `src` directory while exploring a new repository).

package.json
romellogoodman: Resolved

package.json
romellogoodman: Resolved

package.json
romellogoodman: Resolved

packages/components/build-internal/storybook/static/services/ping.html
romellogoodman: Resolved

packages/website/package.json
alex-page: We aren't building for people experienced in using monorepos.

Asking people what goes into a directory called packages we would have mixed results. We would likely have to add documentation and explain to people where code lives. I would rather have visibility at a top level with a lot of directories instead of hide it. It would be unlikely for people to ask what is in the `react-components` directory.

We also want to reduce the total number of packages we create. So some tension around intentionality when creating a new one is a good thing.

sam-b-rose: I think we could do both. I personally prefer and expect the convention of using a `packages/*` directory. However, I've mostly observed that this is only used for code that is published as an npm package. Thus, other workspaces could still live at the root-level. [Turborepo splits this up](https://github.com/vercel/turborepo/blob/main/pnpm-workspace.yaml) well by having published packages in the `packages/*` folder, but then other non-package code at the root (`docs`, `cli`, and `benchmark`).

This approach has the advantages of keeping published assets contained and organized while keeping others more discoverable. The heuristic could be defined–if it goes in a registry, it belongs in `packages`. Thoughts? @alex-page @aaronccasanova

alex-page: I would love to start without packages and add it when it's necessary. We shouldn't need it for the next year.

romellogoodman: I agree, I think we should start with the least amount of assumptions around structure and developer experience and build from there. Rearranging files is also a thing where we can always update and change at any time so it feels low stakes to start with this approach and question established conventions

alex-page (approved):

PR #5274, opened 2022-03-12
Multiselect Prototype (latest): Combobox with typeahead and auto-selection

chloerice:
## Tl;dr: 
This PR demonstrates the massive improvement to clarity, ease, and speed of multi-selection flows when inline autocompletion with automatic selection is implemented in `Combobox` primitives 🙌🏽 [Check it ouuuut](https://5d559397bae39100201eedc1-mmrewhtcbk.chromatic.com/?path=/story/playground-playground--playground) 👀 

### What is this PR doing?

This PR builds on the prototyping Lo and I have been tag-teaming on. I rebased the [`lo/autpcomplete` branch](https://github.com/Shopify/polaris-react/compare/typeahead-proto...lo/autocomplete?expand=1) over my [`listbox-autoselection` branch](https://github.com/Shopify/polaris-react/pull/5256) and got the typeahead working* and styled (typeahead selection state is pink instead of subdued for clear distinction while developing).  

Next steps I recommend would be to:

- [x] Move the suggestion parsing into `TextField`:
  - [x]  Replace the boolean `typeahead` prop with a  string `suggestion` prop that is mutually exclusive from the `selection` prop**
  - [ ] Remove the need for the `isSupportedInputType` boolean on `line 228` of `TextField.tsx`  by making `selection` a mutually exclusive prop from `type`, only allowing it to be set when `type` is one of the input types supported by the browsers [`setSelectionRange` API](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setSelectionRange)
  - [x] Move the condition on `line 114` of `Playground.tsx` that sets either  the `value` or the `suggestion` (if present) on the `input` value into `TextField`
  - [x] Move the logic for calculating and updating the `selection` range in the Playground as the `suggestion` is set into a `useEffect` hook in `TextField.tsx` that calls `setSelectionRange` directly when the `suggestion` changes
  - [x] Update the conditional in `inputClassName` on `line 436` of `TextField.tsx` from `typeahead` to `suggestion`.
  - [x] Update `line 141` of `TextField.scss` from `.typeahead` to `.suggestion`

Let's branch off of this branch to make those changes, and we'll separate things back out so their easy to review once we're satisfied with the API and feel solid on implementation approach.

### Footnotes

*There's a kink to work out with the create action (it's working inconsistently, but wasn't a main priority this week)

**I think  programmatic `selection` of the `value` should be supported as part of flexibility improvements, but for typeahead selection should be handled internally by `TextField` because the behavior is complex and it doesnt make sense to me to have consumers reimplement that logic in the midst of the complexities of filtering and mutating the list of options for different implementations across Admin.

### DIFF
--- a/playground/Playground.tsx
+++ b/playground/Playground.tsx
@@ -1,11 +1,186 @@
-import React from 'react';
+import React, {useState, useEffect, useCallback} from 'react';
 
-import {Page} from '../src';
+import {Page, Combobox, Listbox, Stack, Icon, TextStyle} from '../src';
+import {CirclePlusMinor} from '@shopify/polaris-icons';
+
+interface Option {
+  label: string;
+  value: string;
+}
+
+const tags = [
+  'one',
+  'two',
+  'three',
+  'four',
+  'five',
+  'six',
+  'seven',
+  'eight',
+  'nine',
+  'ten',
+  'eleven',
+  'twelve',
+  'thirteen',
+  'fourteen',
+  'fifteen',
+];
+
+const formattedTags = tags.map((tag) => ({label: tag, value: tag}));
 
 export function Playground() {
+  const [inputValue, setInputValue] = useState<string>('');
+
+  const [options, setOptions] = useState<Option[]>(formattedTags);
+  const [selectedOptions, setSelectedOptions] = useState<Option[]>(
+    formattedTags.slice(0, 7),
+  );
+  const [filteredOptions, setFilteredOptions] = useState<Option[]>(
+    formattedTags.slice(0, 7),
+  );
+  const [selection, setSelection] = useState({start: 0, end: 0});
+  const [autofillValue, setAutofillValue] = useState<string>('');
+
+  const label = 'Tags';
+
+  const updateSelection = useCallback(
+    (selected) => {
+      debugger;
+      const selectedValues = selectedOptions.map((opt) => opt.value);
+      if (selectedValues.includes(selected)) {
+        setSelectedOptions(
+          selectedOptions.filter((option) => option.value !== selected),
+        );
+      } else {
+        setSelectedOptions([
+          ...selectedOptions,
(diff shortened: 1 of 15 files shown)


laurkim:
[PR](https://github.com/Shopify/polaris-react/pull/5297) for Listbox accessibility updates

laurkim:
[PR](https://github.com/Shopify/polaris-react/pull/5298) for Combobox updates

laurkim:
> *There's a kink to work out with the create action (it's working inconsistently, but wasn't a main priority this week)

Figured it out —
It's because the hook that sets `reset` to true when the length of children are updated on `line 211` of `Listbox.tsx` only fires when you type in the first character of the new option (i.e., no tags have `z` in name so I want to create a new tag with the value `zzz`). Since the length of the children is not altered as you continue to write out the characters for the new option value, `reset` is never set to `true` on subsequent changes in the input value.

Can't link to line directly but added this logic to hook on `line 212 - 226` of `Listbox.tsx` on this [PR](https://github.com/Shopify/polaris-react/pull/5297). This approach feels hacky so very open to other suggestions. I didn't want to add another prop to listbox just to listen to text input changes in the text field 😞 
```js
        // check if active option is for creating new listbox option
        const activeOptionLabel =
          activeOption?.element &&
          activeOption?.element.getAttribute('aria-label');
        const activeOptionCreate = activeOptionLabel?.includes('Add: ');

        // reset when active option is to create a new listbox option.
        // if this second parameter for below conditional is not
        // included as an additional parameter, then reset will
        // only be set to true when you type the first char of the
        // new option and not be subsequently called as you continue
        // typing new option value
        if (activeOptionCreate && nextListLength === length) {
          reset = true;
        }
```

laurkim:
Closing because we've since encapsulated this work into smaller PRs.

PR #5292, opened 2022-03-16
[`Badge`] Add `icon` prop

Tom-Bonnike:

Working on https://github.com/Shopify/web/pull/60263, I’m currently unable to add an icon to the “worldwide” badge because the `children` prop only accepts a `string`. My first idea was to change the type of that prop to accept any kind of `React.ReactNode`, but [Kyle pointed out](https://github.com/Shopify/polaris/pull/5292#pullrequestreview-914473475) that this was a little too permissive, so in the end I went with adding a more restrictive `icon` prop.

Adding an `icon` prop like in the same way it’s done for the `Button` component. The `Badge` already renders some sort of icon when passing the `progress` prop, so I made these two props mutually exclusive from each other.

Screenshot 2022-04-07 at 11 57 39

### How to 🎩

<details>

```jsx
import React from 'react';
import {GlobeMinor} from '@shopify/polaris-icons';

  return (
    <Page title="Playground">
      <Stack>
        <Badge icon={GlobeMinor}>Icon, no status</Badge>

        {/*
          The Icon doesn’t get scaled down because the `Icon` size is hardcoded to 20px…
        */}
        <Badge icon={GlobeMinor} size="small">
          Small Badge with icon
        </Badge>

        <Badge icon={GlobeMinor} status="success">
          Icon, success
        </Badge>

        <Badge progress="complete" icon={GlobeMinor}>
          Progress + icon, prefers icon but TypeScript errors
        </Badge>

        <Badge progress="complete">Progress, unchanged</Badge>

        {/*
          Icon only. This works right now but I don‘t see a reason why `children` would be optional.
          If we want to allow this we need to also allow an alt prop for the icon.
        */}
        <Badge icon={GlobeMinor} />
      </Stack>
    </Page>
  );
}

```

</details>

### 🎩 checklist

- [x] For visual design changes, ping @ sarahill to update the Polaris UI kit

### DIFF
--- a/polaris-react/src/components/Badge/Badge.scss
+++ b/polaris-react/src/components/Badge/Badge.scss
@@ -115,3 +115,13 @@
 .withinFilter {
   border-radius: var(--p-border-radius-1);
 }
+
+.Icon {
+  // This compensates for the typical icon used in badges being inset within
+  // its bounding box.
+  margin-left: calc(-1 * (var(--p-space-1)));
+
+  + *:not(.Icon) {
+    margin-left: var(--p-space-1);
+  }
+}
--- a/polaris-react/src/components/Badge/Badge.tsx
+++ b/polaris-react/src/components/Badge/Badge.tsx
@@ -1,9 +1,11 @@
 import React, {useContext} from 'react';
 
+import type {IconSource} from '../../types';
 import {classNames, variationName} from '../../utilities/css';
 import {useI18n} from '../../utilities/i18n';
 import {WithinFilterContext} from '../../utilities/within-filter-context';
 import {VisuallyHidden} from '../VisuallyHidden';
+import {Icon} from '../Icon';
 
 import styles from './Badge.scss';
 
@@ -11,13 +13,15 @@ type Status = 'info' | 'success' | 'attention' | 'warning' | 'critical' | 'new';
 type Progress = 'incomplete' | 'partiallyComplete' | 'complete';
 type Size = 'small' | 'medium';
 
-export interface BadgeProps {
+interface NonMutuallyExclusiveProps {
   /** The content to display inside the badge. */
   children?: string;
   /** Colors and labels the badge with the given status. */
   status?: Status;
   /** Render a pip showing the progress of a given task. */
   progress?: Progress;
+  /** Icon to display to the left of the badge’s content. */
+  icon?: IconSource;
   /**
    * Medium or small size.
    * @default 'medium'
@@ -26,6 +30,12 @@ export interface BadgeProps {
   /** Pass a custom accessibilityLabel */
   statusAndProgressLabelOverride?: string;
 }
+
+export type BadgeProps = NonMutuallyExclusiveProps &
+  (
+    | {progress?: Progress; icon?: undefined}
+    | {icon?: IconSource; progress?: undefined}
+  );
 
 const PROGRESS_LABELS: {[key in Progress]: Progress} = {
   incomplete: 'incomplete',
@@ -48,6 +58,7 @@ export function Badge({
   children,
   status,
   progress,
+  icon,
   size = DEFAULT_SIZE,
   statusAndProgressLabelOverride,
 }: BadgeProps) {
@@ -58,6 +69,7 @@ export function Badge({
     styles.Badge,
     status && styles[variationName('status', status)],
     progress && styles[variationName('progress', progress)],
+    icon && styles.icon,
     size && size !== DEFAULT_SIZE && styles[variationName('size', size)],
     withinFilter && styles.withinFilter,
   );
@@ -115,7 +127,7 @@ export function Badge({
     <VisuallyHidden>{accessibilityLabel}</VisuallyHidden>
   );
 
-  if (progressLabel) {
+  if (progressLabel && !icon) {
     accessibilityMarkup = (
       <span className={styles.Pip}>{accessibilityMarkup}</span>
     );
@@ -124,7 +136,12 @@ export function Badge({
   return (
     <span className={className}>
       {accessibilityMarkup}
-      {children}
+      {icon && (
+        <span className={styles.Icon}>
+          <Icon source={icon} />
+        </span>
+      )}
+      {children && <span>{children}</span>}
     </span>
   );
 }
--- a/polaris-react/src/components/Badge/tests/Badge.test.tsx
+++ b/polaris-react/src/components/Badge/tests/Badge.test.tsx
@@ -1,7 +1,9 @@
 import React from 'react';
 import {mountWithApp} from 'tests/utilities';
+import {GlobeMinor} from '@shopify/polaris-icons';
 
 import {VisuallyHidden} from '../../VisuallyHidden';
+import {Icon} from '../../Icon';
 import {Badge} from '../Badge';
 
 describe('<Badge />', () => {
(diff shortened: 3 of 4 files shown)


Korri:
🤔  shouldn't we use the same pattern as the [Button](https://polaris.shopify.com/components/actions/button) here, and add an `icon` property instead?

Tom-Bonnike:
M'yeah, that might be better, thanks for the suggestion. `Badge`s with a `progress` prop already have some sort of “icon” though, idk if that’s a problem 🤔 

https://polaris.shopify.com/components/images-and-icons/badge

IrinaLipovaya (dismissed):
👍

IrinaLipovaya:
> M'yeah, that might be better, thanks for the suggestion. `Badge`s with a `progress` prop already have some sort of “icon” though, idk if that’s a problem 🤔
> 
> https://polaris.shopify.com/components/images-and-icons/badge

I think Hugo's idea is good actually: being more explicit about content of a badge. `progress` can just goes always first

Tom-Bonnike:
Would love opinions from a Polaris dev/designer, I asked in their channel.

kyledurand (dismissed):
I think we need some solid examples of how to properly use a react node within a badge. 

i.e.
image

It's really easy to get into a situation where there's too much noise and misalignment, and when you factor in text wrap with an icon things get really gross. So if we do ship this it needs some examples in storybook and at I think a simple test as well.

I know this has been talked about before as well so I'd love to get @chloerice 's feedback before this ships. She's back Monday the 21st

cc @sarahill for design 👀 🙏

sarahill:
Having more flexibility seems good but agree some solid examples could help. Even those examples that @kyledurand shared start to poke at the holes. Icon size within a badge, alignment, placement feel like the areas we can provide guidance on.

Tom-Bonnike:
If we have the same implementation (and examples) as the `Button`’s `icon` prop, that should be okay, no?

chloerice:
> 🤔 shouldn't we use the same pattern as the [Button](https://polaris.shopify.com/components/actions/button) here, and add an `icon` property instead?

+1 to having an `icon` prop instead of ReactNode children. Just make it mutually exclusive from `progress` so you can only have one or the other. (See the TextField props for example of typing mutually exclusive props)

polaris-react/src/components/Badge/Badge.scss
```diff
@@ -115,3 +115,9 @@
 .withinFilter {
   border-radius: var(--p-border-radius-1);
 }
+
+.Icon {
+  + *:not(.Icon) {
+    margin-left: var(--p-space-1);
+  }
+}
```
Tom-Bonnike, line 127: I copied these styles from the `Button` icon styles.

polaris-react/src/components/Badge/Badge.tsx
```diff
@@ -115,7 +127,7 @@ export function Badge({
     <VisuallyHidden>{accessibilityLabel}</VisuallyHidden>
   );
 
-  if (progressLabel) {
+  if (progressLabel && !icon) {
```
Tom-Bonnike, line 130: Technically I don’t really need this change but I assume not all Polaris consumers use TypeScript, so at least it makes it prefer the icon to the progress if you pass both?

polaris-react/src/components/Badge/Badge.tsx
```diff
@@ -124,7 +136,12 @@ export function Badge({
   return (
     <span className={className}>
       {accessibilityMarkup}
-      {children}
+      {icon && (
+        <span className={styles.Icon}>
+          <Icon source={icon} />
+        </span>
+      )}
+      {children && <span>{children}</span>}
```
Tom-Bonnike, line 144: I need to wrap it in a `<span>` so I can add margin when there’s an icon.

polaris-react/src/components/Badge/Badge.tsx
```diff
@@ -124,7 +136,12 @@ export function Badge({
   return (
     <span className={className}>
       {accessibilityMarkup}
-      {children}
+      {icon && (
+        <span className={styles.Icon}>
+          <Icon source={icon} />
```
Tom-Bonnike, line 141: Somehow the `children` prop is optional, I don’t know if that was done purposefully, but it means that right now you can render a badge with only an `Icon`, which wouldn’t be correct since we’re never passing it any `alt` attribute.

polaris-react/src/components/Badge/Badge.tsx
```diff
@@ -124,7 +136,12 @@ export function Badge({
   return (
     <span className={className}>
       {accessibilityMarkup}
-      {children}
+      {icon && (
```
Tom-Bonnike, line 139: I put it there because it would apply the margin to the `accessibilityMarkup`. Since the icon has no alt attribute, screen-readers simply ignore it and so the Badge gets read as expected with the status first.

sarahill:
Just noticed some visual weirdness with spacing when the icon is displayed. This is caused by the icon's bounding adding some additional space.

image


I would suggest adjusting the left padding when an icon is displayed. Screen shot shows what it looks like with `padding-left: 4px;` instead of `8px`.

image

chloerice:
This looks good to me codewise, @Tom-Bonnike 💯  Just adjust the left padding as Sara mentioned and add some tests and a changelog entry and ping me when it's ready for review 🚀

Tom-Bonnike:
Ah good catch Sara! There’s actually [such a fix applied for the `Button` component](https://github.com/Shopify/polaris/blob/main/polaris-react/src/components/Button/Button.scss#L83-L86) but I somehow missed that it was a problem here too!

I re-used the same fix (which relies on a negative margin on the icon instead of adjusting the component’s padding).

polaris-react/src/components/Badge/Badge.tsx
Tom-Bonnike: Thoughts on this one @chloerice @sarahill? Do we really use empty `Badge`s anywhere? I think it might be the right time to make the `children` prop required if not. Although technically that would be a breaking change 😅

polaris-react/src/components/Badge/tests/Badge.test.tsx
```diff
@@ -46,6 +48,32 @@ describe('<Badge />', () => {
...
+    expect(badge).toContainReactComponent(Icon, {
+      source: GlobeMinor,
+    });
+  });
+
+  it('does not add pip styles when both progress and icon are provided', () => {
+    const badge = mountWithApp(
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      <Badge progress="incomplete" icon={GlobeMinor} />,
```
Tom-Bonnike, line 69: Might be overkill but if you confirm not all Polaris consumers use TypeScript, I think it’s actually important to cover?

Tom-Bonnike:
Let me know if you think this warrants a documentation update, I think the prop description is pretty self explanatory.

polaris-react/src/components/Badge/Badge.tsx
chloerice: Hmmm yeah that does seem like an oversight 🤔 👀

chloerice (approved):

polaris-react/src/components/Badge/Badge.tsx
Tom-Bonnike: Should I just merge? I’m a bit worried that by introducing the icon we’re allowing consumers to create unaccessible badges 😅

Tom-Bonnike:
I’ll merge, re: https://github.com/Shopify/polaris/pull/5292#discussion_r845909943, I would recommend making `children` required in Polaris’ next major update though 😅

PR #5293, opened 2022-03-16
Stack: optionally flatten React.Fragment children

eliashussary:

In scenarios when you need to conditionally render a collection of `Stack.Item`s  it should be handled with a `React.Fragment`.  Currently `Stack` is handling `React.Fragment` as single element to be rendered and wrapping it with a single `Stack.Item`, which translates into an additional host node and incorrect layout.

A user should explicitly add a host node if this was their desired behaviour.

This change aligns `Stack`'s handling of `React.Fragment` to be more inline with how React internally handles fragments; that is to not translate it to a host node (i.e. `div`).

Note this is a _non-breaking change_ as we only apply this logic when passing in the `flattenReactFragments`. Personally, I think this to should be the default behavior, but it's likely to break a lot of layouts for developers who have developed around this nuance.

image

- Adds an optional prop to `Stack` to flatten `React.Fragment`'s 

## 

### How to 🎩

<details>

```jsx
import React from 'react';

import {Page, Stack} from '../src';

  const someCondition = true;
  return (
    <Page title="Playground">

      <h2>Unflattened (normal)</h2>
      <p>you should see 2 Stack.Items in dev tools</p>
      <Stack>
        {someCondition && (
          <>
            <p>One</p>
            <p>Two</p>
            <p>Three</p>
          </>
        )}
        <p>Four</p>
      </Stack>

      <hr style={{margin: '50px 0'}} />

      <h2>Flattened</h2>
      <p>you should see 4 Stack.Items in dev tools</p>

      <Stack flattenReactFragments>
        {someCondition && (
          <>
            <p>One</p>
            <p>Two</p>
            <p>Three</p>
          </>
        )}
        <p>Four</p>
      </Stack>
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/polaris-react/UNRELEASED.md
+++ b/polaris-react/UNRELEASED.md
@@ -15,6 +15,7 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 - Replaced hard coded `font-size` and `line-height` values with tokens ([5355](https://github.com/Shopify/polaris/pull/5355/))
 - Replaced hard coded spacing values with tokens ([5364](https://github.com/Shopify/polaris/pull/5364/))
 - Simplified usage of color tokens ([5360](https://github.com/Shopify/polaris/pull/5360/))
+- Added `flattedReactFragments` as an optional prop to `Stack` to allow conditional rendering of multiple child elements wrapped in a `React.Fragment` ([#5293](https://github.com/Shopify/polaris-react/pull/5293))
 
 ### Bug fixes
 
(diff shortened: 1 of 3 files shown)


mlaigre:
Note for anyone searching for a workaroung for that Fragment issue with Stack and other "list" components.

Use:
```jsx
{condition && [
  <Element1/>,
  <Element2/>,
)}
```

instead of:
```jsx
{condition && (
  <>
    <Element1/>
    <Element2/>
  <>
)}
```

PR #5299, opened 2022-03-17
fix: Update CustomProperties to avoid flash of unstyled content

aaronccasanova:

Resolves an issue where all children of the `CustomProperties` component would render unstyled while the stylesheet containing custom properties was being generated and injected into the `head` of the document.

This PR removes the stylesheet injection in the `head` tag in favor of inlining custom properties in a `style` tag above the root container.

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Otherwise include the CSS in your HTML. We suggest copying the styles file into 
 ```html
 <link
   rel="stylesheet"
-  href="https://unpkg.com/@shopify/polaris@9.2.1/build/esm/styles.css"
+  href="https://unpkg.com/@shopify/polaris@9.2.2/build/esm/styles.css"
 />
 ```
 
@@ -76,7 +76,7 @@ If React doesn’t make sense for your application, you can use a CSS-only versi
 ```html
 <link
   rel="stylesheet"
-  href="https://unpkg.com/@shopify/polaris@9.2.1/build/esm/styles.css"
+  href="https://unpkg.com/@shopify/polaris@9.2.2/build/esm/styles.css"
 />
 ```
 
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@shopify/polaris",
   "description": "Shopify’s admin product component library",
-  "version": "9.2.1",
+  "version": "9.2.2",
   "private": false,
   "license": "SEE LICENSE IN LICENSE.md",
   "author": "Shopify <dev@shopify.com>",
--- a/src/components/README.md
+++ b/src/components/README.md
@@ -62,7 +62,7 @@ Include the CSS in your HTML. We suggest copying the styles file into your own p
 ```html
 <link
   rel="stylesheet"
-  href="https://unpkg.com/@shopify/polaris@9.2.1/build/esm/styles.css"
+  href="https://unpkg.com/@shopify/polaris@9.2.2/build/esm/styles.css"
 />
 ```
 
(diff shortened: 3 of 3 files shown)


src/components/CustomProperties/CustomProperties.tsx
```diff
@@ -29,28 +34,30 @@ export function CustomProperties(props: CustomPropertiesProps) {
     colorScheme = DEFAULT_COLOR_SCHEME,
   } = props;
 
-  useEffect(() => {
-    let styleSheet = document.getElementById(STYLE_SHEET_ID);
+  const [hasInjectedStyleSheet, setHasInjectStyleSheet] = useState(
+    canUseDOM ? Boolean(document.getElementById(STYLE_SHEET_ID)) : false,
+  );
```
alex-page: I would like a comment here explaining this. It's important and I don't want it removed unintentionally.

alex-page (approved):

src/components/CustomProperties/CustomProperties.tsx
aaronccasanova: Something like this?
```js
  // This flag is used to prevent a flash of unstyled content when the
  // component mounts. Since the main stylesheet is injected in the useEffect
  // we don't want to render children until they have access to the custom properties.
  const [hasInjectedStyleSheet, setHasInjectStyleSheet] = useState(
```

src/components/CustomProperties/CustomProperties.tsx
```diff
@@ -8,6 +8,11 @@ export const DEFAULT_COLOR_SCHEME: ColorScheme = 'light';
 
 export const STYLE_SHEET_ID = 'polaris-custom-properties';
 
+const canUseDOM = Boolean(
```
BPScott: Use the isServer utility perhaps? https://github.com/Shopify/polaris-react/blob/main/src/utilities/target.ts#L1

aaronccasanova: That's perfect 👌

aaronccasanova: [86d1314](https://github.com/Shopify/polaris-react/pull/5299/commits/86d1314854ba73684dfd57d2fdf9f336d3566738)

src/components/CustomProperties/CustomProperties.tsx
aaronccasanova: https://github.com/Shopify/polaris-react/pull/5299/commits/3609e83604d76149d4f7130d7d6a3a474c3ac59a

alex-page (approved):

sam-b-rose (approved):
This looks good! ~~Curious if this will defeat the benefits of SSR though. If most of the application is wrapped with the CustomProperties component, I think this will short circuit rendering any children on the server. Still, seems like the appropriate fix for preventing unstyled content. The only other way I could think of fixing this would be through adding color scheme setting logic in `web`??~~

Disregard, this will only need to be rendered once. Nice fix and simplification!

src/components/CustomProperties/CustomProperties.tsx
```diff
@@ -29,28 +27,20 @@ export function CustomProperties(props: CustomPropertiesProps) {
...
-      style={{color: 'var(--p-text)'}}
-    >
-      {children}
-    </Component>
+    <>
+      <style
+        // Convenience attribute for locating the stylesheet in the DOM.
+        data-polaris-custom-properties=""
+        dangerouslySetInnerHTML={{__html: styles}}
+      />
```
aaronccasanova: Note: Including `style tags in the body` was officially introduced in HTML 5.2 and is supported in all major browsers (See `Make style in body conforming`):
https://w3c.github.io/test-results/html52/implementation-report.html

PR #5320, opened 2022-03-21
Fix build-consumer.js script

laurkim:
After the monorepo changes were shipped, the `build-consumer.js` script wasn't running properly when you tried `yarn run build-consumer [PROJECT_DIR]` from the `polaris-react` dir. It's blocking the ability to tophat a PR to update Combobox documentation.

There are two issues:
1. ~`LICENSE.md` doesn't exist in the `polaris/polaris-react` dir~. Updated this to add in a new `LICENSE.md` to the `polaris-react` dir.
>Copying build to node_modules...
/Users/lokim/src/github.com/Shopify/polaris/node_modules/shelljs/src/common.js:401
      if (config.fatal) throw e;
                        ^
Error: cp: no such file or directory: LICENSE.md

2. Changes to files weren't being picked up

To tophat:
```
// in polaris root dir 
$ git checkout lo/fix-build-consumer-script
$ cd polaris-react
$ yarn run build-consumer polaris-styleguide

// in polaris-styleguide root dir
$ dev up && dev run
// press CTRL + T from same terminal window to open local server for styleguide
// Search Combobox
// see Best practices--TESTING on side nav to reflect latest component readme changes
```
Changes successfully picked up after running the above:
example

### DIFF
--- a/polaris-react/scripts/build-consumer.js
+++ b/polaris-react/scripts/build-consumer.js
@@ -6,7 +6,7 @@ const {config, cp, mkdir, rm} = require('shelljs');
 
 const packageJSON = require('../package.json');
 
-const root = resolve(__dirname, '..');
+const root = resolve(__dirname, '../../');
 const projectDir = process.argv[2];
 
 config.fatal = true;
--- a/polaris-react/LICENSE.md
+++ b/polaris-react/LICENSE.md
@@ -1,0 +1,27 @@
+Copyright (c) 2017-present Shopify Inc.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+The rights granted above may only be exercised to develop and distribute
+applications that integrate or interoperate with Shopify software or services,
+and in the case of external, stand-alone applications that do not embed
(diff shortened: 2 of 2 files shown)


laurkim:
Last commit is temp just to show changes when building for `polaris-styleguide`
I'll drop before merging

polaris-react/scripts/build-consumer.js
```diff
@@ -25,7 +25,7 @@ const projectPolarisDir = resolve(
 const files = [
   'package.json',
   'README.md',
-  'LICENSE.md',
+  '../LICENSE.md',
```
BPScott: As LICENCE on longer exists in the package directory I bet it won't be published to npm anymore - check what gets created in the tarball when you do `npm pack` in the polaris-react directory, and compare it to https://unpkg.com/browse/@shopify/polaris@9.2.2/ it's not quite related to this PR but that seems like an issue

laurkim: Yeah, just ran `npm pack` to check and `LICENSE.md` isn't created when I unpack the `.tgz` file

BPScott: As the licence field says `"license": "SEE LICENSE IN LICENSE.md",`, that really needs to be part of the distributed package.

laurkim: Yeah, I agree and I was going to handle it in a separate issue/PR but I'll just add the commits here.
I added a new `LICENSE.md` for the time being

chloerice: Yeah the license is part of a check by a dependency in web (that I can't remember) that would fail in web when I first pulled down the monorepo stuff from main and tried to build consumer in my spin instance, but moving it back into the package fixed the error. I didn't even think about the npm factor!!

chloerice (approved):

romellogoodman (approved):

PR #5306, opened 2022-03-18
Monorepo MVP

romellogoodman:
As part of #5266 this merges in a MVP for monorepo

### Functionality

* Contains packages: polaris-react and secretpolaris.shopify.com
  * The polaris-react folder contains majority of the files that previously existed in the root of the repo
* Uses yarn workspaces to link the two packages for local development
* Deploys polaris-react using ship-it

### New release process for polaris-react

1. Run `yarn workspace @shopify/polaris version`
2. Input new version
3. Commit changes
4. Go to shipit stack and deploy - https://shipit.shopify.io/shopify/polaris-react/polaris-react

### Relevant Links

* #5262
* #5302
* #5308
* [ship-it stack](https://shipit.shopify.io/shopify/polaris-react/polaris-react) pointed at `monorepo-mvp` branch

PR #5324, opened 2022-03-22
monorepo clean up tasks

romellogoodman:

As #5306 was merged this circles back and cleans up various areas and files:

- Removes `polaris-react/pre-commit`
- Makes secretpolaris.shopify.com a private package.json
- Removes secretpolaris.shopify.com's yarn.lock file


BPScott (approved):
Thanks for tidying these up.

I’d be pretty tempted to keep the pre-commit file logic - letting folks know if they accidentally try to commit the playground file seems useful. What do ya reckon? If we want to keep it, we’d want to move that file back to the root of the repo, update the path to the playgronud within it, and bring back the reference to it in the dev.yml. Perhaps this is an excuse to rewrite the script in JS too as this seems to be the only ruby code in the repo.

romellogoodman:
@BPScott responded back to you via Slack

PR #5331, opened 2022-03-23
feat: Stylelint Polaris

aaronccasanova:
### WHAT is this pull request doing?

This PR introduces the `@shopify/stylelint-polaris` package to validate CSS and improve compatibility using Polaris. `@shopify/stylelint-polaris` aims to encompass all our recommendations to enable the best experience using the Polaris Design System. This initial release focuses on validating Polaris `custom-property` usage. However, the goal is to continue iterating and enhance the plugin to provide suggestions, improve overall coverage, and more.

> Note: The `custom-properties-allowed-list` plugin has been moved out of `polaris-react` and into the `@shopify/stylelint-polaris` package.

### Architecture overview

```
stylelint-polaris/
├─ plugins/
│  ├─ custom-properties-allowed-list.js
|  | # (Optional) Public facing plugins for advanced configurations
|  | # (See advanced config example below)
|  ├─ index.js
|  |
├─ configs/
|  | # Common rules for `polaris-react` and Polaris consumers
|  ├─ shared.js
|  |
|  | # Applied in `polaris-react` containing:
|  | # - shared.js
|  | # - specific `custom-properties-allowed-list` rules
│  ├─ internal.js
|
| # Public facing config containing:
| # - shared.js
| # - specific `custom-properties-allowed-list` rules for Polaris consumers
├─ index.js
```

Example usage in `polaris-react`

```json5
// polaris-react/package.json
{
  "stylelint": {
    "extends": [
      "@shopify/stylelint-polaris/configs/internal"
    ]
  },
};
```

Example usage in Polaris consumer project: **Basic**

```json5
// consumer/package.json
{
  "stylelint": {
    "extends": [
      "@shopify/stylelint-polaris"
    ]
  },
};
```

Example usage in Polaris consumer project: **Advanced**

```js
// consumer/stylelintrc.js
module.exports = {
  extends: ['@shopify/stylelint-polaris'],
  plugins: ['@shopify/stylelint-polaris/plugins'],
  rules: {
    'stylelint-polaris/custom-properties-allowed-list': {
      // Extending the allowed Polaris custom properties
      allowedProperties: [/--my-app-.+/],
    },
  },
};
```

### How to 🎩

🖥 [Local development instructions](https://github.com/Shopify/polaris-react/blob/master/README.md#development)
🗒 [General tophatting guidelines](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting.md)
[...]

PR #5344, opened 2022-03-24
Create a coverage stylelint plugin for checking token coverage

### LINKED ISSUE #5314: Coverage rules in stylelint
Migrate Polaris coverage rules from [polaris-coverage](https://github.com/shopify/polaris-coverage-shopifycloud-com) into stylelint rules.

This will help bring coverage insights closer to the code by integrating the coverage data into the code editor.

sam-b-rose:
> ⚠️ **Note**: This PR currently creates a lot of styelint errors.

Fixes #5314  

Adds the coverage regex tests as a stylelint plugin called `@shopify/coverage`. The regular expressions are part of a `global-disallowed-list` rule that is configured in the `coverage/configs`.


polaris-react/stylelint/plugins/coverage/configs/index.js
```diff
@@ -0,0 +1,127 @@
...
+  rules: {
+    'at-rule-disallowed-list': ['keyframes'],
+    'color-named': 'never',
+    'color-no-hex': true,
+    'declaration-property-value-disallowed-list': [
+      {
+        '/color$/': ['/#[0-9a-f]{3,6}/'],
+        display: ['grid', 'flex'],
+        'font-weight': ['/[0-9]*/'],
+        opacity: ['/?![01]'],
```
aaronccasanova: Are we trying to error on anything that is not `0`, `1`, or a custom property?

If so, I believe this is what we want: [/^(?!0|1)\d+/](https://regex101.com/r/nGthny/1)

polaris-react/stylelint/plugins/coverage/configs/index.js
```diff
@@ -0,0 +1,127 @@
+module.exports = {
+  plugins: ['stylelint-scss', '../'],
+  rules: {
+    'at-rule-disallowed-list': ['keyframes'],
```
aaronccasanova: I'm a little surprised to see this as an error. Is the idea that Polaris will or should expose keyframes via custom properties or components? Similar to [open-props/animations](https://github.com/argyleink/open-props/blob/main/src/props.animations.css)

alex-page: We want to easily find where keyframes are used and hopefully we can replace them with something exactly like that @aaronccasanova.

polaris-react/stylelint/plugins/coverage/configs/index.js
```diff
@@ -0,0 +1,127 @@
...
+      'grid-column',
+      'grid-column-start',
+      'grid-column-end',
+      'grid-template',
+      'grid-template-areas',
+      'grid-template-rows',
+      'grid-template-columns',
+      'grid-area',
+    ],
+    'unit-disallowed-list': ['px', 'rem', 'em', 's', 'ms'],
```
aaronccasanova: Shouldn't we allow `px`s, `rem`s, and `em`s? `polaris-react` has a `postcss` plugin to convert `px-to-rems` and I believe we use `em`s for all our media queries.

When we move this over to `@shopify/stylelint-polaris`, this might be one of those rules where we have different configs for internal `polaris-react` usage and Polaris consumers

alex-page: I think you are right. With time we could update that rule for web but we will need to migrate a lot of values. If these can be rendered as warnings, it would be fine to leave on in shopify/web. I want to see yellow = divergence from system.

alex-page (commented):
My only feedback is lets make these warnings instead of failures to start.

polaris-react/stylelint/plugins/coverage/configs/index.js
alex-page: Yeah you got it!

polaris-react/stylelint/plugins/coverage/configs/index.js
```diff
@@ -0,0 +1,127 @@
...
+    'scss/function-color-relative': true,
+    '@shopify/coverage/global-disallowed-list': [
+      [
```
aaronccasanova, line 63: Is it possible to update the configuration of this plugin to a single array. Or perhaps convert to an object with a `patterns` key.

polaris-react/stylelint/plugins/coverage/index.js
```diff
@@ -0,0 +1,10 @@
+const stylelint = require('stylelint');
+
+const namespace = require('./utils/namespace');
+const rules = require('./rules');
+
+const rulesPlugins = Object.keys(rules).map((ruleName) => {
+  return stylelint.createPlugin(namespace(ruleName), rules[ruleName]);
```
aaronccasanova, line 7: **suggestion (non-blocking):** I'm not sure we need the `namespace()` abstraction. What do you think about just adding a `export const ruleNamePrefix = '@shopify/coverage'` to this file and updating this line to ```createPlugin(`${ruleNamePrefix}/${ruleName}`)```?

polaris-react/stylelint/plugins/coverage/index.js
```diff
@@ -0,0 +1,10 @@
+const stylelint = require('stylelint');
+
+const namespace = require('./utils/namespace');
+const rules = require('./rules');
+
+const rulesPlugins = Object.keys(rules).map((ruleName) => {
```
aaronccasanova, line 6:
```suggestion
const coveragePlugins = Object.keys(rules).map((ruleName) => {
```

polaris-react/stylelint/plugins/coverage/index.js
aaronccasanova: Oh just got to the `global-disallowed-list` plugin and see your using `namespace()` there! Feel free to resolve this conversation (:

polaris-react/stylelint/plugins/coverage/rules/global-disallowed-list/index.js
```diff
@@ -0,0 +1,61 @@
...
+const rule = (primary) => {
+  return (postcssRoot, postcssResult) => {
+    const validOptions = validateOptions(postcssResult, ruleName, {
+      actual: primary,
+      possible: [isRegExp, isString],
+    });
+
+    if (!validOptions) {
+      return;
+    }
```
aaronccasanova: Shouldn't we inform folks their config is malformed, rather the fail silently? Here is an example of how we handle misconfigurations with the `custom-properties-allowed-list` plugin: https://github.com/Shopify/polaris/blob/399556256b44c36559abda717299f2064895df3d/polaris-react/stylelint/plugins/custom-properties-allowed-list/index.js#L63-L82

polaris-react/stylelint/plugins/coverage/rules/global-disallowed-list/index.js
```diff
@@ -0,0 +1,61 @@
...
+
+    postcssRoot.walkDecls((decl) => {
+      let currentTest;
+      const hasDisallowed = primary.some((test) => {
+        currentTest = test;
+        return testStringOrRegEx(test, decl.value);
+      });
```
aaronccasanova: Wonder if we should accumulate all disallowed matches in a declaration and report. Otherwise, folks might have to run Stylelint multiple times to resolve all issues. e.g.

```js
const allDisallowed = [];

primary.forEach((test) => {
  if (testStringOrRegEx(test, decl.value)) {
    allDisallowed.push({test, value: decl.value})
  }
});

if (!allDisallowed.length) return

report(...)
```

polaris-react/stylelint/plugins/coverage/configs/index.js
sam-b-rose: Great question. Seems like only the secondary option allows for an object. It should only be a single array though–good catch! Will remove the extra `[ ... ]`'s

polaris-react/stylelint/plugins/coverage/rules/global-disallowed-list/index.js
sam-b-rose: Excellent suggestion! Done 👍

polaris-react/stylelint/plugins/coverage/rules/global-disallowed-list/index.js
sam-b-rose: Yep yep, great idea. Added a very basic error message ⚠️ not as fancy with the options parsing but should certainly help with pointing out config issues

polaris-react/stylelint/plugins/coverage/index.js
sam-b-rose: Though the naming isn't the best, seems like this is a common convention among other plugins

* [stylelint-order](https://github.com/hudochenkov/stylelint-order/blob/master/index.js#L5)
* [stylelint-a11y](https://github.com/YozhikM/stylelint-a11y/blob/master/src/index.js#L4)
* [stylelint-scss](https://github.com/stylelint-scss/stylelint-scss/blob/master/src/index.js#L5)
* [stylelint-plugin-stylus](https://github.com/ota-meshi/stylelint-plugin-stylus/blob/master/lib/index.js#L6)

Cool with sticking with this convention?

alex-page: `stylelint-polaris-coverage`

sam-b-rose:
@alex-page I set each rule to only be reported as a `warning`, but looks like loom [doesn't approve of warning](https://github.com/Shopify/loom/blob/main/packages/loom-plugin-stylelint/src/plugin-stylelint.ts#L46) 😬 which is causing the [CI to still fail](https://github.com/Shopify/polaris/runs/5706458476?check_suite_focus=true#step:7:15).

polaris-react/stylelint/plugins/coverage/configs/index.js
sam-b-rose: Landed on [this solution](https://regex101.com/r/qKpzfJ/1) which will disallow fractional values as well (ex: `0.5`)

aaronccasanova: Nice! I realized I didn't account for numbers that start with `0` or `1` (e.g. `01`, `10`, `123`). Here is an update including your fractional values adjustment: https://regex101.com/r/Lpvfd4/1

sam-b-rose: Oooo nice refinement ✨  using this one 👍 ty!

polaris-react/src/components/Avatar/Avatar.scss
```diff
@@ -9,7 +9,7 @@
   min-width: var(--pc-avatar-extra-small-size);
   max-width: 100%;
   background: var(--p-surface-neutral);
-  color: var(--p-icon-subdued);
+  color: var(--p-icon-subdued, #ffffff);
```
alex-page:
```suggestion
  color: var(--p-icon-subdued);
```

alex-page (approved):

sam-b-rose:
Going to merge this in but we'll need to do some restructuring to fit the stylelint changes in #5331

PR #5358, opened 2022-03-28
polaris.shopify.com Dockerfile & GitHub action

romellogoodman:

As part of the effort to deploy polaris.shopify.com onto Shopify infrastructure this adds a Github action for creating a dockerfile

### DIFF
--- a/polaris.shopify.com/Dockerfile
+++ b/polaris.shopify.com/Dockerfile
@@ -1,0 +1,21 @@
+FROM node:16
+
+ENV PORT 3000
+
+# Create app directory
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
+# Installing dependencies
+COPY package*.json /usr/src/app/
+RUN npm install
+
+# Copying source files
+COPY . /usr/src/app
+
+# Building app
+RUN npm run build
+EXPOSE 3000
+
+# Running the app
+CMD "npm" "start"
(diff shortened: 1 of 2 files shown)


polaris.shopify.com/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM node:16
+
+ENV PORT 3000
+
+# Create app directory
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
+# Installing dependencies
+COPY package*.json /usr/src/app/
```
aaronccasanova, line 10:
```suggestion
COPY package*.json .
```

The path to application directory is already set by the above [`WORKDIR` instruction](https://docs.docker.com/engine/reference/builder/#workdir)

polaris.shopify.com/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+RUN npm install
+
+# Copying source files
+COPY . /usr/src/app
```
aaronccasanova, line 14:
```suggestion
COPY . .
```

Same as above: https://github.com/Shopify/polaris/pull/5358/files#r836704378

polaris.shopify.com/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM node:16
+
+ENV PORT 3000
+
+# Create app directory
+RUN mkdir -p /usr/src/app
```
aaronccasanova, line 6:
```suggestion
```
This should be created by the following [`WORKDIR` instruction](https://docs.docker.com/engine/reference/builder/#workdir)
> If the WORKDIR doesn’t exist, it will be created even if it’s not used in any subsequent Dockerfile instruction.

romellogoodman:
Closed as this is no longer the approach we are taking

Issue #5359, opened 2022-03-28
Color coverage to 90%


aveline:

alex-page:
This is done! Expect an update on the coverage site to show this.

PR #5360, opened 2022-03-28
Simplify color token usage

aveline:

Part of #5359 

Refactor to use color tokens directly instead of through sass variables


lgriffee (approved):
🚢✨🎉

Issue #5366, opened 2022-03-29
[Polaris for VSCode] Package and publish


yurm04:
Prepare the VSCode extension for publishing on the VSCode marketplace
- [x] add extension icon
- [ ] update `package.json` with all required fields
- [ ] rewrite README.md to be focused on extension features/functionality
- [ ] add a CONTRIBUTING.MD
- [ ] add license
- [ ] add some demo gifs and screenshots
- [ ] connect with @charlespwd to publish using the `Shopify` org

PR #5385, opened 2022-03-29
Prepare `polaris-for-vscode` for publish on marketplace

yurm04:
Changes needed to get the Polaris for VSCode extension published on the VSCode marketplace https://github.com/Shopify/polaris/issues/5366

* adding Polaris icon
* updating package.json


polaris-for-vscode/package.json
```diff
@@ -1,9 +1,22 @@
 {
   "name": "polaris-for-vscode",
   "displayName": "polaris-for-vscode",
-  "description": "Polaris Design System vscode tooling",
-  "repository": "https://github.com/Shopify/polaris-react",
+  "publisher": "Shopify",
+  "description": "Polaris Design System VSCode extension",
+  "icon": "polaris-for-vscode/docs/polaris-icon.png",
```
yurm04: no idea if this will resolve to the right URL 🤔 might need to use the resolved github url

charlespwd: It would be relative to the vsce package IIRC.

aaronccasanova: Yes, needs to be relative to the manifest (package.json). I took a similar approach with my VSCode extension: https://github.com/aaronccasanova/format-grid-areas/blob/12dc39d1836cd1ded6ab4ec02ad8064d17c2ce68/package.json#L14

aaronccasanova: Also, see the example manifest: https://code.visualstudio.com/api/references/extension-manifest#example

polaris-for-vscode/package.json
```diff
@@ -1,9 +1,22 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/Shopify/polaris"
```
aaronccasanova:
```suggestion
    "url": "https://github.com/Shopify/polaris.git"
```
Not sure if this update is required, but it is in the [manifest example](https://code.visualstudio.com/api/references/extension-manifest#example).

polaris-for-vscode/package.json
```diff
@@ -1,9 +1,22 @@
...
+  },
+  "homepage": "https://github.com/Shopify/polaris/tree/main/polaris-for-vscode",
```
aaronccasanova:
```suggestion
  "homepage": "https://github.com/Shopify/polaris/blob/main/polaris-for-vscode/README.md",
```
Same as above. Not sure if this update is required, but it matches the [manifest example](https://code.visualstudio.com/api/references/extension-manifest#example)

polaris-for-vscode/package.json
aaronccasanova:
```suggestion
  "icon": "docs/polaris-icon.png",
```

romellogoodman (approved):

PR #4840, opened 2021-12-15
[Badge] Revert attention status deprecation

chloerice:
### WHY are these changes introduced?

Fixes #4789  

### WHAT is this pull request doing?

This PR removes the deprecation of the attention status and adds the yellow coloring back to the attention status badge. 

Screen Shot 2021-12-15 at 4 39 30 PM

## 

### How to 🎩

🖥 [Local development instructions](https://github.com/Shopify/polaris-react/blob/master/README.md#development)
🗒 [General tophatting guidelines](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting.md)
📄 [Changelog guidelines](https://github.com/Shopify/polaris-react/blob/master/documentation/Versioning%20and%20changelog.md)

Check [the badge stories](https://5d559397bae39100201eedc1-dcuwjcmezd.chromatic.com/?path=/story/all-components-badge--partially-complete-badge) or

<details>
<summary>Copy-paste this code in <code>playground/Playground.tsx</code>:</summary>

```jsx
import React from 'react';

import {Page, Stack, Badge} from '../src';

export function Playground() {
  return (
    <Page title="Playground">
      <Stack>
        <Badge size="small" status="attention" progress="partiallyComplete">
          Partially fulfilled
        </Badge>
      </Stack>
    </Page>
  );
}
```

</details>

### 🎩 checklist

- [x] Tested on [mobile](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting.md#cross-browser-testing)
- [x] Tested on [multiple browsers](https://help.shopify.com/en/manual/intro-to-shopify/shopify-admin/supported-browsers)
- [x] Tested for [accessibility](https://github.com/Shopify/polaris-react/blob/master/documentation/Accessibility%20testing.md)
- [x] Updated the component's `README.md` with documentation changes
- [x] [Tophatted documentation](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting%20documentation.md) changes in the style guide
- [x] For visual design changes, ping @ sarahill to update the Polaris UI kit

PR #5389, opened 2022-03-30
Increase token coverage

aveline:

Relates to #5359 

- Replaces some color values with tokens/css custom properties
- Remove unused sass vars
- [Banner] Update Secondary action styles to remove `box-shadow` underline and transition and replace with `text-decoration: underline` to be consistent with `PlainButton` and `Link` styles on hover/focus.


alex-page (approved):

polaris-react/UNRELEASED.md
```diff
@@ -16,7 +16,7 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 - Replaced hard coded `font-size` and `line-height` values with tokens ([5355](https://github.com/Shopify/polaris/pull/5355/))
 - Replaced hard coded spacing values with tokens ([5364](https://github.com/Shopify/polaris/pull/5364/))
 - Simplified usage of color tokens ([5360](https://github.com/Shopify/polaris/pull/5360/))
-
+- Increased token coverage by replacing hard coded `border-width` values with tokens, replaced sass var with css custom property, removed unused sass vars, and updated `Banner` Secondary action styles ([5389](https://github.com/Shopify/polaris/pull/5389))
```
alex-page, line 19: I would add on a new line...

Created `icon-attention` token

polaris-react/src/tokens/token-groups/color.light.json
```diff
@@ -110,6 +110,8 @@
   "surface-success-subdued-hovered": "rgba(236, 246, 241, 1)",
   "surface-success-subdued-pressed": "rgba(226, 241, 234, 1)",
   "text-success": "rgba(0, 128, 96, 1)",
+  "icon-attention": "rgba(138, 97, 22)",
+  "surface-attention": "rgba(255, 234, 138)",
```
aaronccasanova, line 114: Both these values were in the original [$polaris-colors](https://github.com/Shopify/polaris/blob/4293e9c8a5bd3cc0b4ac1bc5207d239d6bdd3729/documentation/guides/legacy-polaris-v8-public-api.scss#L51-L53) map. While I'm not sold on the `light`,`dark`, `darker` naming convention, perhaps we should add those tokens here or in the [legacy-tokens](https://github.com/Shopify/polaris/blob/main/polaris-react/src/tokens/token-groups/legacy-tokens.json) (rather than add arbitrary one offs to our `color.*` token groups). Thoughts?

e.g.
```json5
// token-groups/legacy-tokens.json
{
  "yellow-dark": "rgb(138, 97, 22)",
  "yellow-light": "rgb(255, 234, 138)"
}
```

aveline: I hesitate to add it to `legacy-tokens` because we should be slowly removing values from that file, not adding to it. This is a color that is important to keep in the system (see https://github.com/Shopify/polaris/pull/4840) and not accidentally remove again so that's why it seemed like the color token group is a good place for it? Is that not the primary place we are keeping color tokens now?

lgriffee (approved):
LGTM! 💯🎉🚢✨

polaris-react/UNRELEASED.md
```diff
@@ -16,6 +16,8 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 - Replaced hard coded `font-size` and `line-height` values with tokens ([5355](https://github.com/Shopify/polaris/pull/5355/))
 - Replaced hard coded spacing values with tokens ([5364](https://github.com/Shopify/polaris/pull/5364/))
 - Simplified usage of color tokens ([5360](https://github.com/Shopify/polaris/pull/5360/))
+- Increased token coverage by replacing hard coded `border-width` values with tokens, replaced sass var with css custom property, removed unused sass vars, and updated `Banner` Secondary action styles ([5389](https://github.com/Shopify/polaris/pull/5389))
+- Created `icon-attention` color tokens ([5389](https://github.com/Shopify/polaris/pull/5389))
```
alex-page:
```suggestion
- Created `icon-attention` and `surface-attention` color tokens ([5389](https://github.com/Shopify/polaris/pull/5389))
```

polaris-react/src/tokens/token-groups/color.light.json
aaronccasanova: > we should be slowly removing values from that file, not adding to it

Ideally, but it sounds like these values shouldn't have been deleted and would have otherwise been placed in the `legacy-tokens` file during the v9 release. 

> Is that not the primary place we are keeping color tokens now?

Yes, however adding these tokens to the `color.*` files indicates these are public facing tokens that folks are welcome to start including in their styles. Was there any input from design to validate the semantic naming of `icon-attention` and `surface-attention`? I feel like we should start being very strict and considerate when adding new tokens moving forward as updates require breaking changes.

It's very likely I'm missing some context, so feel free to disregard my concerns and ship these tokens 👍

Issue #4762, opened 2021-11-30
Iterate on duration token values and implementation


johanstromqvist:
Follow up to the work done in #4595

johanstromqvist:
- [x] Phase 1: #4780
- [ ] Phase 2: Ship duration values between 50ms increments. Identify what the smallest duration step should be ("primitive"), e.g. `10ms`, `16.666ms`, or `25ms`.
- [ ] Phase 3: Ship aliases that solve pattern problems, e.g. `default`, `small-enter`, `small-exit`


**Phase 1**
| Name | Value |
| ----------- | ----------- |
| duration-0 | 0ms |
| duration-50 | 50ms |
| duration-100 | 100ms |
| duration-150 | 150ms |
| duration-200 | 200ms |
| duration-250 | 250ms |
| duration-300 | 300ms |
| duration-350 | 350ms |
| duration-400 | 400ms |
| duration-450 | 450ms |
| duration-500 | 500ms |

These values are easy to understand, and I estimate that they can solve 80-90% of our use cases.

**Phase 2**
The outstanding question is how we best accommodate needs for values between the 50s, e.g. `125ms`. We can think of these as "advanced" needs, that more experienced creators might have when choreographing multiple animations or tweaking the perceived speed of a delicate detail. If we don't provide tokens for advanced use cases, creators either need to conform to the system on the expense of UX or implement the advanced solution off system. Reference discussion in [issue #4595](https://github.com/Shopify/polaris-react/issues/4595#issuecomment-976982159).

alex-page:
I love expanding the duration scale. It removes the need for teams to hard code one-off values. I also think this would be pretty easy to adjust with find and replace if we had to update the values.

aaronccasanova:
This looks great Johan! The separation of the three phases makes sense and promotes incremental enhancement 👍

lgriffee:
I agree with Alex and Aaron 👍🏻 This phased approach will allow us to move quickly on the most impactful things that we are aligned on (Phase 1), while giving us more time to refine and think through the areas where we may have less alignment or certainty (Phase 2 & 3).

alex-page:
@lgriffee this work is good to implement. So feel free to get started or hand it over to someone else.

lgriffee:
Hoping to get `border-radius` work wrapped up today so I can take this on!

alex-page:
I am closing this issue for now. We will go through the token values for motion intentionally in the future. Look forward to going through this issue then.

[...]

PR #5405, opened 2022-04-01
Replace remaining hardcoded duration values with tokens

### LINKED ISSUE #5339: Motion coverage to 90%
**Push motion coverage as high as possible (80-90-100%)**
Document all the remaining divergence on motion and create an approach to consolidate, remove or create new token values. We need to spend more time go deeper into the values we scale but that can be a follow up once we have the values on mainline.

lgriffee:

Increase coverage of motion tokens. Relates to #5339

~~Replacing remaining hardcoded duration (ms) values with tokens.~~  Removing unnecessary `transitions` and `animations` with hard coded duration values.
- [SearchField] Remove unnecessary `transition` for backdrop `background-color`, placeholder text `color`, and search icon svg `fill`
- [SearchField] Remove unnecessary `animation` for `background-color`
- [UserMenu] Remove unnecessary `transition` for name and detail text `color`

### DIFF
--- a/polaris-react/src/components/TopBar/components/SearchField/SearchField.scss
+++ b/polaris-react/src/components/TopBar/components/SearchField/SearchField.scss
@@ -81,7 +81,7 @@ $search-icon-width: calc(#{$icon-size} + var(--p-space-4));
 
   &::placeholder {
     color: var(--p-text);
-    transition: var(--p-duration-150) color var(--p-ease) 33ms;
+    transition: var(--p-duration-150) color var(--p-ease) var(--p-duration-50));
   }
 
   &::-webkit-search-decoration,
@@ -102,7 +102,7 @@ $search-icon-width: calc(#{$icon-size} + var(--p-space-4));
   transform: translateY(-50%);
 
   svg {
-    transition: var(--p-duration-150) fill var(--p-ease) 33ms;
+    transition: var(--p-duration-150) fill var(--p-ease) var(--p-duration-50));
   }
 }
 
--- a/polaris-react/src/components/TopBar/components/SearchField/SearchField.scss
+++ b/polaris-react/src/components/TopBar/components/SearchField/SearchField.scss
@@ -76,7 +76,6 @@ $search-icon-width: calc(#{$icon-size} + var(--p-space-4));
 
   &::placeholder {
     color: var(--p-text);
-    transition: var(--p-duration-150) color var(--p-ease) var(--p-duration-50);
   }
 
   &::-webkit-search-decoration,
--- a/polaris-react/src/components/TopBar/components/SearchField/SearchField.scss
+++ b/polaris-react/src/components/TopBar/components/SearchField/SearchField.scss
@@ -150,14 +150,4 @@ $search-icon-width: calc(#{$icon-size} + var(--p-space-4));
   will-change: background-color;
   transition: background-color var(--p-duration-200) var(--p-ease);
   border-radius: var(--p-border-radius-1);
-  animation: toLightBackground 0.01ms;
-}
-
-@keyframes toLightBackground {
-  to {
-    background-color: var(
-      --p-surface-neutral,
-      var(--pc-top-bar-background-lighter)
-    );
-  }
 }
--- a/polaris-react/src/components/TopBar/components/SearchField/SearchField.scss
+++ b/polaris-react/src/components/TopBar/components/SearchField/SearchField.scss
@@ -100,10 +100,6 @@ $search-icon-width: calc(#{$icon-size} + var(--p-space-4));
   height: $icon-size;
   pointer-events: none;
   transform: translateY(-50%);
-
-  svg {
-    transition: var(--p-duration-150) fill var(--p-ease) var(--p-duration-50);
-  }
 }
 
 .Clear {
--- a/polaris-react/src/components/TopBar/components/SearchField/SearchField.scss
+++ b/polaris-react/src/components/TopBar/components/SearchField/SearchField.scss
@@ -81,7 +81,7 @@ $search-icon-width: calc(#{$icon-size} + var(--p-space-4));
 
   &::placeholder {
     color: var(--p-text);
-    transition: var(--p-duration-150) color var(--p-ease) var(--p-duration-50));
+    transition: var(--p-duration-150) color var(--p-ease) var(--p-duration-50);
   }
 
   &::-webkit-search-decoration,
@@ -102,7 +102,7 @@ $search-icon-width: calc(#{$icon-size} + var(--p-space-4));
   transform: translateY(-50%);
 
   svg {
-    transition: var(--p-duration-150) fill var(--p-ease) var(--p-duration-50));
+    transition: var(--p-duration-150) fill var(--p-ease) var(--p-duration-50);
   }
 }
 
--- a/polaris-react/src/components/TopBar/components/SearchField/SearchField.scss
+++ b/polaris-react/src/components/TopBar/components/SearchField/SearchField.scss
@@ -38,7 +38,6 @@ $search-icon-width: calc(#{$icon-size} + var(--p-space-4));
 .Input:focus {
   ~ .Backdrop {
     @include focus-ring($style: 'focused');
-    background-color: var(--p-surface-search-field);
   }
 
   ~ .BackdropShowFocusBorder {
@@ -51,10 +50,6 @@ $search-icon-width: calc(#{$icon-size} + var(--p-space-4));
 }
 
 .focused {
-  .Backdrop {
-    background-color: var(--p-surface-search-field);
-  }
-
   .BackdropShowFocusBorder {
     border: 1px solid var(--pc-top-bar-border);
   }
@@ -139,11 +134,6 @@ $search-icon-width: calc(#{$icon-size} + var(--p-space-4));
   right: 0;
   bottom: 0;
   left: 0;
-  background-color: var(
-    --p-surface-search-field,
-    var(--pc-top-bar-background-lighter)
-  );
-  will-change: background-color;
-  transition: background-color var(--p-duration-200) var(--p-ease);
+  background-color: var(--p-surface-search-field);
   border-radius: var(--p-border-radius-1);
 }
--- a/polaris-react/src/components/TopBar/components/UserMenu/UserMenu.scss
+++ b/polaris-react/src/components/TopBar/components/UserMenu/UserMenu.scss
@@ -25,8 +25,3 @@
   color: var(--p-text);
   text-align: left;
 }
-
-.Name,
-.Detail {
-  transition: var(--p-duration-150) color var(--p-ease) var(--p-duration-50);
-}
--- a/polaris-react/src/components/TopBar/components/UserMenu/UserMenu.scss
+++ b/polaris-react/src/components/TopBar/components/UserMenu/UserMenu.scss
@@ -28,5 +28,5 @@
 
 .Name,
 .Detail {
-  transition: var(--p-duration-150) color var(--p-ease) 33ms;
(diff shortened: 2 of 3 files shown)


polaris-react/src/components/TopBar/components/SearchField/SearchField.scss
```diff
@@ -81,7 +81,7 @@ $search-icon-width: calc(#{$icon-size} + var(--p-space-4));
 
   &::placeholder {
     color: var(--p-text);
-    transition: var(--p-duration-150) color var(--p-ease) 33ms;
```
lgriffee, line 84: In general I think we can replace all three instances of 33ms with the 50ms token (see breakdown below).

**Search Field Placeholder Text** 
Screen Shot 2022-03-31 at 5 52 10 PM

Here's the comparison between 33ms and 50ms (respectively listed in that order) for the placeholder text transition. While the 50ms is slower they're pretty darn close. I think we can get away with replacing the 33ms with 50ms but would love a gut check from @johanstromqvist.

https://user-images.githubusercontent.com/21976492/161166414-d63a9561-ab9b-4fc2-8ae5-e179da13d048.mp4



https://user-images.githubusercontent.com/21976492/161166421-4d499300-f6c1-49c2-8750-d124ca881eeb.mp4

polaris-react/src/components/TopBar/components/SearchField/SearchField.scss
```diff
@@ -102,7 +102,7 @@ $search-icon-width: calc(#{$icon-size} + var(--p-space-4));
   transform: translateY(-50%);
 
   svg {
-    transition: var(--p-duration-150) fill var(--p-ease) 33ms;
```
lgriffee, line 105: **Search Field Icon SVG**
Screen Shot 2022-03-31 at 5 52 29 PM

I'm not entirely sure when the transition for this one would trigger? Does the icon fill ever change? But if it does change I think we can get away with replacing the 33ms with 50ms.

polaris-react/src/components/TopBar/components/UserMenu/UserMenu.scss
```diff
@@ -28,5 +28,5 @@
 
 .Name,
 .Detail {
-  transition: var(--p-duration-150) color var(--p-ease) 33ms;
```
lgriffee, line 31: **User Menu Name Text + Detail Text**
Screen Shot 2022-03-31 at 5 51 22 PM

I'm also not entirely sure when the transition for this one would trigger? Does the color ever change for either texts? But if it does change I think we can get away with replacing the 33ms with 50ms.

polaris-react/src/components/TopBar/components/SearchField/SearchField.scss
```diff
@@ -150,14 +150,4 @@ $search-icon-width: calc(#{$icon-size} + var(--p-space-4));
...
-  animation: toLightBackground 0.01ms;
-}
-
-@keyframes toLightBackground {
-  to {
-    background-color: var(
-      --p-surface-neutral,
-      var(--pc-top-bar-background-lighter)
-    );
-  }
```
lgriffee, line 162: We can get rid of the `animation` prop with this value applied to `.Backdrop`. 

After pairing with @aaronccasanova we concluded that only the `transition` would really get triggered on this class. We also concluded that the `transition` for `background-color` isn't doing anything right now because the color for `.Input:focus` and `.focused` are the same as the `background-color` on `.Backdrop`. If the intended behavior is for the color not to change on `focus` then I think we can do some clean up here. Otherwise, if we do want it to transition then the colors need to be different. Would be good to loop in design (maybe @sarahill?) on this one and check to see what it should be. 

https://user-images.githubusercontent.com/21976492/161170190-2ff99176-80aa-4d14-bcb7-26d9b82687b1.mp4



https://user-images.githubusercontent.com/21976492/161170210-d761d827-6777-4225-9de7-b7a00b8d76a0.mp4

lgriffee:
@aveline Assigning you to this PR as well if you want to work on it while I'm out tomorrow!

polaris-react/src/components/TopBar/components/SearchField/SearchField.scss
alex-page: This happens so quickly and on such a thin/small font I don't know how much of the motion I perceive. I personally wouldn't mind removing this one. Would love to know other folks throughts.

polaris-react/src/components/TopBar/components/SearchField/SearchField.scss
alex-page: I think we should remove this one. Looks like it does nothing.

polaris-react/src/components/TopBar/components/SearchField/SearchField.scss
alex-page: Yeah this should be removed. Probably an oversight in the redesign.

polaris-react/src/components/TopBar/components/UserMenu/UserMenu.scss
alex-page: Remove it!

polaris-react/src/components/TopBar/components/SearchField/SearchField.scss
aveline: Yes to either removing or 50ms

aveline (approved):

polaris-react/src/components/TopBar/components/SearchField/SearchField.scss
lgriffee: Just for reference to compare to the videos above here's a video of the transition removed:

https://user-images.githubusercontent.com/21976492/161597844-40f4d33e-021e-45e2-9705-42dbe4f24b63.mp4

lgriffee: I'm going to vote for removing the transition. With no transition the focus ring appears and the placeholder text changes color at the same time. With the 50ms transition I'm noticing that the placeholder text color change lags slightly after the focus ring and it feels a bit disjointed. 

Would love @johanstromqvist input too but realize he's on vacation this week. I'd be happy to move forward with no transition here for now and if Johan thinks otherwise after returning from vacation I can open up another PR to change it.

aveline: Sounds like a good plan @lgriffee

alex-page (approved):
Lets ship this. We can undo if needed.

polaris-react/src/components/TopBar/components/SearchField/SearchField.scss
johanstromqvist: Great assessment @lgriffee! Thx for ping!

TDLR: Agree that it's safe to remove in this instance!

**Re: focus ring**
- Is focus ring necessary when focusing search field with mouse? My assessment is that the focus ring more often create visual noise and distraction when mousing, instead of providing the clarity it's intended to. This makes suppressing the focus ring an important a11y and perceived perf consideration. We should have a documented opinion on when and why we suppress focus ring.
- If we suppress focus ring on the search field, then we don't have a sync problem.


**Re: default approach when replacing hard coded values**
- I think adjusting a value is a _design tweak_, while removing a transition is more of a _redesign_. We should avoid "redesigning" in scenarios that we don't own or have full context of. It's better if respective product team make their own decisions with the context they have of user needs.
  - If this is a decision that we own, then we can safely change or remove it.
  - If it's not ours, then we should opt for minimum impact, i.e. adjust to closest/most appropriate value such as 50ms.
  - If it's not ours but it's harming UX, then we should reach out to the product team and help them make the change while also learning why.

**Re: replacing values that are likely to be covered shortly?**
- Removing a purposefully applied 33ms might be to optimize for the short term and send the wrong message. It might feel like an improvement to token coverage today, while 33ms might actually be covered tomorrow.
- [As we've discussed in the past](https://github.com/Shopify/polaris/issues/4762), 50ms increments is not sufficient. I frequently use sub-50 durations, such as 17 and 33, and want to encourage product teams to do so as well. This is especially important for choreography and making multiple animations feel as one.

lgriffee: Great! Thanks @johanstromqvist  for confirming it’s safe to remove in this instance! 🙏🏻 🎉

**Re: focus ring** 

> Is focus ring necessary when focusing search field with mouse?

I'm not sure and would love to get more history/context on it. From quickly looking through the [MDN `:focus-visible` page](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) it does mention the following: 

> [Cognition](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#cognition)
> It may not be obvious as to why the focus indicator is appearing and disappearing if a person is using mixed forms of input. For users with cognitive concerns, or who are less technologically literate, this lack of consistent behavior for interactive elements may be confusing.

Perhaps this ⬆️ is why we have the focus ring for both mouse and keyboard input? But I'm not sure why our use is not consistent 🤔 (also love your example).

> We should have a documented opinion on when and why we suppress focus ring.

I totally agree with this for consistency and clarity into decision making. 

> If we suppress focus ring on the search field, then we don't have a sync problem.

Do you mean suppressing it only on mouse click? If so then we would still run into the sync problem with keyboard input (unless we don't have transitions with keyboard focus? Not sure what is best practice in terms of keyboard focus). 

**Re: default approach when replacing hard coded values**
These are really great points about replacing/removing values and I love the decision matrix you outlined. Is the SearchField owned by us since an element used across the admin and is not owned by any one specific section of the admin? 

**Re: replacing values that are likely to be covered shortly?**
This is the only instance of the 33ms getting removed (the other instances that were removed we determined wouldn't get triggered). [I've added a note](https://github.com/Shopify/polaris/issues/4762#issuecomment-1095545484) about possibly adding the transition back as part of our future motion work (along with assessing the focus ring).

johanstromqvist: Sorry, didn't look at the code earlier and noticed just now that the `33ms` is a `delay`. My best guess is that someone added that delay to visually compensate for a perceived delay in the focus ring (possibly caused by the ring starting behind the field effectively hiding the first few frames). Borderline hacky. It might not affect our decision to remove the transition, but it makes me wonder if the creator knows something we don't.

The transition shorthand is `property name | duration | easing function | delay`, which makes...

``` css
transition: var(--p-duration-150) color var(--p-ease) 33ms;
```

...equivalent to...

``` css
transition-property: color;
transition-duration: 150ms;
transition-timing-function: ease;
transition-delay: 33ms;
```

Observations:
- The property is misplaced, but apparently without effect
- The use of CSS variables makes the shorthand less readable
- I've seen durations and delays being mixed up due to wrong position in the shorthand.

It seems like a good idea to avoid the shorthand to improve readability and predictability, and to avoid human errors. Is there a reason why we might not want to avoid the shorthand?

> Focus ring suppression

Added https://github.com/Shopify/polaris/issues/5504

> Default approach/ownership

Idk, but it makes sense as one of ours.

> Do you mean suppressing it only on mouse click? If so then we would still run into the sync problem with keyboard input (unless we don't have transitions with keyboard focus? Not sure what is best practice in terms of keyboard focus).

Lol, true! We'd still run into sync issues on keyboard nav.

lgriffee: I'm fine with breaking out the shorthands for all our transition values. I don't see any reason to not to break them out...only thing I could think of would be if our future aliases in Phase 3 would use shorthand at all (which could make things inconsistent) but if you don't foresee the aliases being shorthand then I can't think of anything else. I'm all for making our code as straightforward and readable as possible. We can fold this work into whichever open ticket/issue this ends up in in the future.

PR #5416, opened 2022-04-04
Use loom for linting styles via modified stylelint plugin

### LINKED ISSUE #5362: Add back loom stylelint plugin
## Issue summary

The `loom` stylelint plugin was removed in order to allow for coverage warnings. Once https://github.com/Shopify/loom/pull/325 is merged in and released, we can use the plugin configuration to set an amount of allowable warnings rather than calling stylelint directly.

Related issues: #5344 , https://github.com/Shopify/loom/pull/325

sam-b-rose:

Fixes #5362 

This adds back the use of `loom` for linting styles. It leverages the [plugin hooks](https://github.com/Shopify/loom/tree/main/packages/loom-plugin-stylelint#hooks) to adjust the number of `maxWarnings` allowed for `stylelint`. This is needed since the addition of the coverage stylelint rules which are marked as warnings and used to provided coverage insights within the `polaris-react` package.

### DIFF
--- a/polaris-react/loom.config.ts
+++ b/polaris-react/loom.config.ts
@@ -10,6 +10,7 @@ import {
   rollupPlugins,
 } from '@shopify/loom-plugin-build-library';
 import {eslint} from '@shopify/loom-plugin-eslint';
+import {stylelint} from '@shopify/loom-plugin-stylelint';
 import {prettier} from '@shopify/loom-plugin-prettier';
 import replace from '@rollup/plugin-replace';
 import image from '@rollup/plugin-image';
@@ -37,7 +38,9 @@ export default createPackage((pkg) => {
     }),
     buildLibraryWorkspace(),
     eslint(),
+    stylelint({files: '**/*.scss'}),
     prettier({files: '**/*.{md,json,yaml,yml}'}),
+    stylelintAdjustmentsPlugin(),
     rollupAdjustPluginsPlugin(),
     rollupAdjustOutputPlugin(),
     jestAdjustmentsPlugin(),
@@ -68,6 +71,20 @@ function jestAdjustmentsPlugin() {
           ...transforms,
           '\\.s?css$': require.resolve('./config/jest-transform-style'),
           '\\.svg$': require.resolve('./config/jest-transform-image'),
+        }));
+      });
+    });
+  });
+}
+
+function stylelintAdjustmentsPlugin() {
+  return createWorkspacePlugin('Polaris.Stylelint', ({tasks: {lint}}) => {
+    lint.hook(({hooks}) => {
+      hooks.configure.hook((configure) => {
+        // Modify the maximum number of allowed warnings from the default of 0
+        configure.stylelintFlags?.hook((flags) => ({
+          ...flags,
+          maxWarnings: Infinity,
         }));
       });
     });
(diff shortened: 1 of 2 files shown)


polaris-react/loom.config.ts
```diff
@@ -74,6 +77,21 @@ function jestAdjustmentsPlugin() {
...
 
+function stylelintAdjustmentsPlugin() {
+  return createWorkspacePlugin('Polaris.Stylelint', ({tasks: {lint}}) => {
+    lint.hook(({hooks}) => {
+      hooks.configure.hook((configure) => {
+        // Modify the maximum number of allowed warnings from the default of 0
+        configure.stylelintFlags?.hook((flags) => ({
+          ...flags,
+          // Using -1 to allow all warnings
+          maxWarnings: -1,
```
aaronccasanova: Mind running `yarn lint --no-cache --isolate-step Stylelint` and posting a screenshot? I'm curious what the generated Stylelint CLI command looks like.

aaronccasanova (approved):

polaris-react/loom.config.ts
sam-b-rose: Yes!

[Yarn lint output](https://screenshot.click/04-09-vnep8-fgmf6.png)

BPScott: On the up side this works, on the down side it seems loom swallows output of the warnings (contrast this with running `yarn run stylelint '**/*.scss' --max-warnings=9999` manually)

If we're spamming out warnings then presumably we'd want to be able to see them? In this case annoying it might be worth shying away from loom for now - I've put up a proof-of-concept in https://github.com/Shopify/polaris/pull/5418 for running linters at the root of the repo without loom

polaris-react/loom.config.ts
```diff
@@ -74,6 +77,20 @@ function jestAdjustmentsPlugin() {
...
 }
 
+function stylelintAdjustmentsPlugin() {
+  return createWorkspacePlugin('Polaris.Stylelint', ({tasks: {lint}}) => {
+    lint.hook(({hooks}) => {
+      hooks.configure.hook((configure) => {
+        // Modify the maximum number of allowed warnings from the default of 0
+        configure.stylelintFlags?.hook((flags) => ({
+          ...flags,
+          maxWarnings: 1000,
```
aaronccasanova: May need to bump this number as there is already 859 warnings and this is just the start of the coverage rules we are tracking:
image

aaronccasanova: Can always do that later when it becomes an issue (:

polaris-react/loom.config.ts
sam-b-rose: The root-level linting looks good to me! Was also wanting to make this improvement as well.

Not using loom sounds good in this case. I also agree not swallowing the warnings is ideal 👍 Question, would it be better no just omit the `--max-warnings` flag instead of using `9999`?

polaris-react/loom.config.ts
sam-b-rose: Good idea, agreed 👍 Nice `⚠️` grep!

sam-b-rose:
Closing in favor of root-level linting #5418 🧹 ✨

PR #5418, opened 2022-04-04
Root-level linter configuration

BPScott:

The majority of the code in the monorepo should adhere to a single set of linting rules. Packages that diverge are free to do so but they would be the exception rather than the rule. Currently linting is package-specific which means new packages need to bring their own linting rules.

This PR pulls the eslint, stylelint and prettier configs up to the top of the repository where they can apply to all packages and adds a new `yarn run lint` command can be used to lint the entire codebase in one go.

- Moves eslint stylelint and prettier configs from inside polaris-react to the root of the repository
- Removes the lint plugins from loom in the polaris-react project (eventually we might want a workspace loom config where linting is ran by loom at the root of the repo, but that can be follow-up work)
- Add a `yarn run lint` command at the root of the repo, this is ran as part of CI

Currently the `stylelint-polaris` and `polaris.shopify.com` folders are excluded from eslint as they currently fail. A follow-up PR should fix linting issues in those packages.

### How to 🎩

Run `yarn run lint` and see that linting occurs

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,10 @@
+node_modules
+.loom
+/polaris-react/build
+/polaris-react/build-internal
+/polaris.shopify.com/.next
+
+
+# These folders should pass linting, remove these lines and fix the errors
+/polaris.shopify.com
+/stylelint-polaris
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
       - run: yarn --frozen-lockfile
       - run: yarn workspace @shopify/polaris run validate-tokens
       - run: yarn workspace @shopify/polaris run type-check && yarn run build
-      - run: yarn workspace @shopify/polaris run lint
+      - run: yarn run lint
       - run: yarn workspace @shopify/polaris run test
 
   accessibility_test:
--- a/package.json
+++ b/package.json
@@ -16,11 +16,20 @@
     "build:react": "yarn workspace @shopify/polaris build",
     "dev:site": "yarn workspace polaris-shopify-com dev",
     "start:site": "yarn workspace polaris-shopify-com start",
-    "build:site": "yarn workspace polaris-shopify-com build"
+    "build:site": "yarn workspace polaris-shopify-com build",
+    "lint": "yarn run lint:eslint && yarn run lint:stylelint && yarn run lint:prettier",
+    "lint:eslint": "eslint . --max-warnings 0 --cache --cache-location '.loom/cache/eslint'",
+    "lint:stylelint": "yarn workspace '@shopify/stylelint-polaris' run gen-polaris-vars && stylelint '**/*.scss' --max-warnings 9999",
+    "lint:prettier": "prettier '**/*.{md,json,yaml,yml}' --check"
   },
   "devDependencies": {
+    "@shopify/eslint-plugin": "^41.0.1",
     "@shopify/prettier-config": "^1.1.2",
+    "@shopify/stylelint-plugin": "^11.0.0",
     "@shopify/typescript-configs": "^5.1.0",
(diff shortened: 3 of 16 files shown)


.eslintignore
```diff
@@ -0,0 +1,10 @@
+node_modules
+.loom
+/polaris-react/build
+/polaris-react/build-internal
+/polaris.shopify.com/.next
+
+
+# These folders should pass linting, remove these lines and fix the errors
```
BPScott, line 8: Task for a follow-up PR: Remove these lines and opt these folders into being linted

sam-b-rose (approved):

package.json
```diff
@@ -16,11 +16,20 @@
     "build:react": "yarn workspace @shopify/polaris build",
     "dev:site": "yarn workspace polaris-shopify-com dev",
     "start:site": "yarn workspace polaris-shopify-com start",
-    "build:site": "yarn workspace polaris-shopify-com build"
+    "build:site": "yarn workspace polaris-shopify-com build",
+    "lint": "yarn run lint:eslint && yarn run lint:stylelint && yarn run lint:prettier",
+    "lint:eslint": "eslint . --max-warnings 0 --cache --cache-location '.loom/cache/eslint'",
+    "lint:stylelint": "yarn workspace '@shopify/stylelint-polaris' run gen-polaris-vars && stylelint '**/*.scss' --max-warnings 9999",
```
sam-b-rose: Should we just omit the `--max-warnings` flag here?

aveline (approved):

polaris-react/package.json
```diff
@@ -163,12 +149,6 @@
     "ios >= 13.4"
   ],
   "prettier": "@shopify/prettier-config",
```
aaronccasanova:
```suggestion
```

package.json
BPScott: Yep we should. Removed!

polaris-react/package.json
BPScott: Fixed!

.github/workflows/ci.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       - run: yarn --frozen-lockfile
       - run: yarn workspace @shopify/polaris run validate-tokens
       - run: yarn workspace @shopify/polaris run type-check && yarn run build
-      - run: yarn workspace @shopify/polaris run lint
+      - run: yarn run lint
```
aaronccasanova, line 30: Note: This `ci.yml` workflow was recently updated to only trigger on changes to `polaris-react`. Do we need to revert that to ensure all workspaces are being linted?

yarn.lock
```diff
@@ -2162,11 +2162,6 @@
     stylelint-prettier "^2.0.0"
     stylelint-scss "^4.0.0"
 
-"@shopify/typescript-configs@^5.0.0":
```
sam-b-rose, line 2165: When running `yarn` on this branch, I'm also getting some other `yarn.lock` updates. Maybe just do a double-check that there aren't any other dep updates that need to be made to the lock file.

It removes the `@shopify/loom-plugin-*` deps for me after a `yarn install`

BPScott: d'oh. fixed! I removed some deps from package.json but forgot to run yarn afterwards

.github/workflows/ci.yml
BPScott: I'd strongly say yes. Skipping CI on a subset of your repository sounds like a bad idea. Though perhaps we revert that in a separate PR?

There's some other improvements that can be made in here too, such as caching the .loom  folders to speed up type-checking and test runs, perhaps we could lump that in too?

sam-b-rose: I will note these improvements for later follow up in a separate issue. Thanks Ben!

Issue #5424, opened 2022-04-05
Monorepo tooling improvements - pan repo type-check, and CI caching improvements


BPScott:
Three quick wins that I'd like to mash out sooner rather than later.

Currently tooling is very focused on "got to run in one project". This makes adding new projects tougher as you have to setup all their lint/test/etc infra and remember to call it in CI (currently we have 3 projects, CI tests one of them). Instead there should be root-level package.json scripts  that run linting over the whole project (already in place), run type-checking over the whole project, and run tests over the whole project (and builds too but I have been reminded not to eat my whole cake at once)

These first two changes prime CI to handle these pan-repo checks and integrate pan-repo type-checking. pan-repo testing and building shall be addressed in a follow-up.

## CI Improvements - DONE in #5426

Currently CI only runs when changes to polaris-react are made. This should be removed and CI should run against all commits. This is because checks like "run type-checking and linting across the whole repo" can be affected by files otside of the polaris-react folder.

We can also add caching to CI - by caching .loom folders and .tsbuildinfo files (example: https://github.com/Shopify/loom/blob/d74c54bbaa76796145f854e2761208013681d130/.github/workflows/ci.yml#L46-L56) we can reduce the time taken when running eslint, and tsc, and jest. I estimate this can save about 90 seconds total.

## TSC project references - DONE in #5432 

There should be have a single `yarn run type-check` command in the root package.json that runs type-checking across all projects in the repository. Without this, currently we have to type-check projects individually, and our CI does not type-check projects other than polaris-react currently.

- Use tsc project references ([demo](https://github.com/Shopify/loom/blob/d74c54bbaa76796145f854e2761208013681d130/tsconfig.json#L4) to type-check multiple projects at a time
- Create a tsconfig.base.json that contains our common typescript configuration, that can be extended by each per-project tsconfig file.

## Bonus: Opt stylelint-polaris and polaris.shopify.com into linting - DONE in #5430 and #5435

Currently these two packages are [omitted from linting in our eslintignorefile](https://github.com/Shopify/polaris/blob/8566a2e40463c6f561114b4aa592c27837d9c9c6/.eslintignore#L8-L10). These lines should be removed and any linting violations should be fixed in these projects.

PR #5426, opened 2022-04-05
Run CI on all PRs and add caching

BPScott:

Part of #5424

CI should run on all commits, and PRs, and we should use caching to make subsequent builds faster.

Related issues: #5424

Run CI on all commits, and PRs and add caching


sam-b-rose (approved):
Looking faster ⚡ ⏲️ 👍  nice caching!

PR #5427, opened 2022-04-05
Replace `@keyframe` values with tokens

### LINKED ISSUE #5339: Motion coverage to 90%

lgriffee:

Increase coverage of motion tokens. Relates to #5339

- Creating `@keyframe` tokens 
- Creating custom property aliases for `@keyframe` tokens 
- Replacing hardcoded `@keyframe` values with new tokens
- Removing any unused `@keyframe` values
- Creates a new motion duration token for 5000ms

 - - -
Co-Authored-By: Aaron Casanova <32409546+aaronccasanova@users.noreply.github.com>


polaris-react/src/tokens/token-groups/keyframes.json
```diff
@@ -0,0 +1,7 @@
+{
```
lgriffee: Keyframes are in their own object (isolated from motion) so they can be handled separately in the style utility functions in [`src/components/CustomProperties/style.ts`](https://github.com/Shopify/polaris/blob/d0d9b18880ea31424b695841960ce213ef6b206b/polaris-react/src/components/CustomProperties/styles.ts#L96).

All other animation custom properties or composite custom properties should still live in motion as they need to be copied into each color scheme selector.

polaris-react/src/components/Button/Button.scss
```diff
@@ -171,12 +171,6 @@
   pointer-events: none;
 }
 
-@keyframes loading {
-  to {
-    transform: rotate(360deg);
-  }
-}
-
```
lgriffee, line 179: There's no corresponding `animation` property that uses this keyframe.

polaris-react/src/components/Navigation/_variables.scss
```diff
@@ -175,16 +175,6 @@ $nav-animation-variables: (
...
-@keyframes fade-in {
-  0% {
-    opacity: 0;
-  }
-
-  100% {
-    opacity: 1;
-  }
-}
-
```
lgriffee, line 187: There's no corresponding `animation` property that uses this keyframe.

polaris-react/src/components/TopBar/components/SearchDismissOverlay/SearchDismissOverlay.scss
```diff
@@ -14,13 +14,3 @@
...
-
-@keyframes fade-in {
-  0% {
-    opacity: 0;
-  }
-
-  100% {
-    opacity: 1;
-  }
-}
```
lgriffee, line 26: There's no corresponding `animation` property that uses this keyframe.

polaris-react/src/components/CustomProperties/styles.ts
```diff
@@ -7,7 +7,7 @@ import {
   OSColorSchemes,
 } from '../../tokens';
 
-const defaultCustomProperties = `
+const defaultRules = `
```
aaronccasanova:
```suggestion
const defaultDeclarations = `
```

polaris-react/src/components/CustomProperties/tests/CustomProperties.test.tsx
```diff
@@ -35,6 +35,7 @@ const mockTokens: Tokens = {
   motion: mockTokenGroup,
   // Note: We don't need to assign mock values to the remaining static tokens.
   depth: {},
+  keyframes: {},
```
lgriffee: @aaronccasanova Do we want to assign mock values to keyframes?

stylelint-polaris/scripts/gen-polaris-data.js
```diff
@@ -2,24 +2,27 @@ const fs = require('fs');
...
-} catch (err) {
-  throw new Error(
-    `Could not create custom properties file "${filePath}": ${err}`,
+
+  fs.writeFileSync(
+    path.join(dirPath, 'polaris-keyframe-names.js'),
+    `module.exports = ${JSON.stringify(polarisKeyframeNames)};`,
   );
+} catch (err) {
+  throw new Error(`Could not create custom properties file ${err}`);
```
lgriffee: @aaronccasanova Not sure how we want to restructure the error message now since it could be one of two paths. Looks like we missed the  `${filePath}` variable in the error message when we replaced `filePath` and `outFile`.

polaris-react/src/components/CustomProperties/tests/CustomProperties.test.tsx
aaronccasanova: Yes, we should mock and test `keyframes` since that token group has unique behavior in the style utilities. I'm happy to help with that, let me know!

stylelint-polaris/scripts/gen-polaris-data.js
aaronccasanova: I'm fine with something generic e.g. `Failed to generate Polaris data: ${err}`

polaris-react/UNRELEASED.md
```diff
@@ -20,6 +20,7 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 - Created `icon-attention` and `surface-attention` color tokens ([5389](https://github.com/Shopify/polaris/pull/5389))
 - Increased token coverage by removing unnecessary `transitions` and `animations` with hard coded duration values ([5405](https://github.com/Shopify/polaris/pull/5405/))
 - Added optional visual density updates and zebra striping to `DataTable` ([#5365](https://github.com/Shopify/polaris/pull/5365/))
+- Increased token coverage by creating `@keyframe` tokens and replacing any hardcoded instances ([5427](https://github.com/Shopify/polaris/pull/5427/))
```
aaronccasanova:
```suggestion
- Increased token coverage by creating `@keyframes` tokens and replacing any hardcoded instances ([5427](https://github.com/Shopify/polaris/pull/5427/))
```

polaris-react/src/tokens/token-groups/beta-tokens.json
```diff
@@ -0,0 +1,7 @@
+{
+  "animation-name-bounce": "p-keyframes-bounce",
+  "animation-name-fade-in": "p-keyframes-fade-in",
+  "animation-name-fill-up": "p-keyframes-fill-up",
+  "animation-name-pulse": "p-keyframes-pulse",
+  "animation-name-spin": "p-keyframes-spin"
```
aaronccasanova: `beta-tokens.json` is an idea @lgriffee, @BPScott, and I came up with to allow adding new (experimental) tokens, that are by nature publicly accessible, yet internally marked `beta`. This should give us the freedom to make updates without introducing breaking changes.

If it helps we can additionally prepend `beta-*` to each token name.

In the case of these `animation-name-*` tokens, I am interested in experimenting with the [Open Props approach to animation tokens](https://github.com/argyleink/open-props/blob/40ae491289f38939ce89af05f44eb0249e874db6/src/props.animations.css#L2-L21) where we'd expose custom properties for the `animation:` shorthand. These `animation-name-*` tokens help accomplish our immediate goal to improve `@keyframes` coverage while laying the ground work for introducing `animation:` shorthand custom properties e.g.
```json5
// token-groups/motion.json
{
  "animation-name-bounce": "p-keyframes-bounce",
  "animation-bounce": "var(--p-animation-name-bounce) var(--p-duration-200) var(--p-ease)",
 }
```
```css
/* MyComponent.scss */
.usage {
  animation: var(--p-animation-bounce);
  animation-duration: optional override;
}
```

polaris-react/src/tokens/token-groups/beta-tokens.json
```diff
@@ -0,0 +1,7 @@
+{
```
lgriffee: We've aliased keyframe token values from `keyframes.json` to custom CSS properties to avoid CSS Modules hashing them by default.

Screen Shot 2022-04-06 at 10 00 36 AM


The other alternative would be to wrap any `animations` consuming the keyframes tokens in [:global{}](https://github.com/Shopify/polaris/pull/3032/files#diff-6540ed5581ce15b789159dfc22825c3f18d22445dd7f999f77e01316b83a4492) which felt gross.

polaris-react/src/components/Indicator/Indicator.scss
```diff
@@ -18,47 +18,11 @@
 
 .pulseIndicator::before {
   z-index: 1;
-  animation: bounce 5s ease infinite;
+  animation: var(--p-animation-name-bounce) 5s ease infinite;
```
aveline: Is it worth creating a duration token for the `5s`?

.stylelintrc.js
```diff
@@ -1,5 +1,9 @@
 module.exports = {
   extends: ['@shopify/stylelint-plugin/prettier', './stylelint-polaris'],
+  // Disabling @shopify/stylelint-plugin/configs/core no-unknown-animations as styelint
+  // is not aware of global Polaris keyframes
+  // TODO: create custom plugin to ensure animation-names match Polaris keyframe names
+  rules: {'no-unknown-animations': undefined},
```
alex-page, line 6: This is a nice workaround. How difficult is it to get this into `@shopify/stylelint`? I would prefer to fix it there than add this workaround if possible.

polaris-react/src/tokens/token-groups/beta-tokens.json
alex-page: I am not a huge fan of having beta tokens. Why would we not ship the values and iterate upon them?

aveline: Is it worth mentioning the introduction of `beta-tokens` in the changelog?

alex-page: I also don't like the two seperate files for animation and keyframe values. Is there any reason we wouldn't consolidate these? This PR seems to add a lot of complexity. What is the simplest way we could do this?

polaris-react/src/components/Indicator/Indicator.scss
```diff
@@ -18,47 +18,11 @@
...
-  #{map-get($pip-animation, finish-scaling-small)} {
-    transform: scale(0.85);
-    opacity: 1;
-  }
-
-  to {
-    transform: scale(2.5);
-    opacity: 0;
-  }
+  animation: p-pulse 5s ease infinite;
```
aveline: Should `p-pulse` be a custom property?

aveline (approved):
Nice work! 🎉

polaris-react/src/components/Backdrop/Backdrop.scss
```diff
@@ -9,8 +9,9 @@ $entry-iterations: 1;
   left: 0;
   display: block;
   background-color: var(--p-backdrop);
-  animation: fade-in var(--p-duration-200) $entry-iterations forwards;
-  opacity: 1;
+  animation: var(--p-animation-name-fade-in) var(--p-duration-200)
+    $entry-iterations forwards;
```
alex-page:
```suggestion
    1 forwards;
```

Could we remove `$entry-iterations` it doesn't seem to add a lot of value?

polaris-react/src/tokens/token-groups/keyframes.json
alex-page: I don't like this approach. It adds a lot of complexity to the keyframe tokens.

Is there any way we could keep them together? This PR seems to add a lot of complexity due to the handing of these values. What is the simplest way we could do this?

polaris-react/src/components/Button/Button.scss
alex-page: I do wonder what renders the spinning icon now... I feel like it's a gif 😭. We should probably investigate as we might be able to bring down the bundle size replacing it with an animation.

polaris-react/src/components/Indicator/Indicator.scss
lgriffee: Yes it should be! Thanks so much for catching that!

polaris-react/src/components/Backdrop/Backdrop.scss
lgriffee: 💯 Good call!

polaris-react/src/tokens/token-groups/beta-tokens.json
BPScott: > Why would we not ship the values and iterate upon them?

How can consumers tell that tokens are stable and are thus safe to use in their apps? At the moment all tokens are considered available for use in consuming apps and thus they're part of a stable api - removing a token is a breaking change that warrants a major version release. Thus polaris-react can't iterate when releasing minor versions without leaving a trail of "deprecated, we don't use this anymore as we found a better approach" tokens in its wake till the next major version, or by having regular major version releases (which is bad because it leads to update fatigue in consuming apps - people get sick of chasing polaris). The alternative here is that the whole of polaris goes into a long running beta but that's bad for other reasons (e.g. keeping main and the beta branch in sync).


By marking a subset of tokens as "beta" then we explicitly denounce that stability guarantee - polaris would be free to iterate and remove them at any point in time as they're explictly not recommended for usage within consuming apps. Once we're happy with an approach then tokens can be promoted up to the stable set where the usual "please use this, it won't go away except in a major version" constraints exist.

Thus we end up with 3 tiers of token:

- beta: Unstable, we're still working out if we like this approach. Don't use this as it might go away unexpectedly. May be promoted up to stable
- stable: Stable, use this in your consuming apps. It won't go away
- deprecated/legacy: Stable, but shall go away in the next major version. We don't use this internally. Please migrate away from this in external apps.

alex-page: As soon as any token ships we should expect that it's stable. Consumers are going to use them. This is adding extra layers and complexity that we don't need right now.

It's a valid concern but not something needed today.

BPScott: If everything is stable then how do you propose that polaris "iterate[s] upon them"? Shall we consider it acceptable that we'll probably end up with a handful of tokens that won't be used within the polaris-react codebase, but consumers may use (as there is no warning not to use them), and then we have to migrate folks away come the next major version when we remove tokens?

alex-page: There are many ways to resolve this and adding additional layers of naming and tooling to the tokens is not the approach we will take. It adds complexity to something that should be simple and creates extra education for our users around the tokens and how they are named. We already have massive issues with naming and adding versioning to it is going to blow that up. I don't know any other design systems that ship tokens with beta in the name.

polaris-react/src/components/Indicator/Indicator.scss
lgriffee: Good catch! I think it is! The only two instances of a 5s (5000ms) duration in polais-react are in this file. But it feels like an important value. Replacing it with 500ms definitely does not work (see below—it's a pretty funny difference 😂). Also just to make sure there aren't any other `s` durations I checked our stylelint plugin but these are the only two. 👍🏻

https://user-images.githubusercontent.com/21976492/162077353-54d02bc7-b6ef-4684-8596-6f1d54b699fd.mp4


https://user-images.githubusercontent.com/21976492/162077372-d9296d44-e9f5-48fa-a9e2-8e962f424ed9.mp4

polaris-react/src/tokens/token-groups/keyframes.json
aaronccasanova: > It adds a lot of complexity to the keyframe tokens

Can you please be more specific about the complexity? Currently, the `keyframes` object is simply [mapped to a string and injected outside of the `:root` and `[p-color-scheme]` selectors](https://github.com/Shopify/polaris/blob/e78a6ce32789fbb37ce023e79a1a4c6cdbdc8981/polaris-react/src/components/CustomProperties/styles.ts#L84-L96).

We specifically didn't add `keyframes.json` in `token-groups/motion.json` to reduce complexity as we initially started reworking `defaultDeclarations` into a function, filtering out `keyframe` tokens, transforming them accordingly, and ensuring they are injected outside of `:root`. With the current `key:value` token-group interface, this process involved less then ideal logic to check `tokenName.startsWith('keyframes')`. If your fine with that approach, we can throw up a POC for you to compare.

polaris-react/src/components/Button/Button.scss
lgriffee: Looks like it's an animation in the Spinner component! 🎉
Screen Shot 2022-04-06 at 4 12 42 PM

alex-page: Nice one!

polaris-react/src/tokens/token-groups/keyframes.json
aaronccasanova: Here's is the quickest POC/refactor for comparison: https://github.com/shopify/polaris/compare/keyframes-custom-properties...consolidate-keyframe-tokens

alex-page: I like that. @lgriffee do you have any thoughts?

polaris-react/src/tokens/token-groups/keyframes.json
```diff
@@ -0,0 +1,7 @@
...
+  "keyframes-bounce": "{ from, 65%, 85% { transform: scale(1) } 75% { transform: scale(0.85) } 82.5% { transform: scale(1.05) } }",
+  "keyframes-fade-in": "{ to { opacity: 1 } }",
+  "keyframes-fill-up": "{ 0% { width: 0 } }",
```
alex-page: I would like to remove the `keyframes-fill-up` keyframe. Animating width is low performance in CSS and scaling this approach is not good. I think we should revisit removing this keyframe in another way.

lgriffee: @alex-page Jammed with @aaronccasanova and @aveline on removing `keyframes-fill-up`. [The solution we came up with](https://github.com/Shopify/polaris/pull/5427/commits/214a426dc91ad7050fed9d1024c07a302f21a31e) does add quite a few more lines of code to the .tsx file but I'm not sure we could avoid it without using an animation/keyframes. 

Motion still behaves the same:

https://user-images.githubusercontent.com/21976492/162332047-8d25bd8c-b63e-4ccf-8383-9569abc4c8a2.mp4

https://user-images.githubusercontent.com/21976492/162332062-c2290e04-4b0f-416a-b848-bde90a0137eb.mp4

alex-page: I am happy to leave the keyframes as is if it is a worse solution. I do wonder if we could scale the bar instead of using width. e.g. have a bar that fills the container scaleX(1) that fills the whole bar, then `scaleX(0.1)` and move through the transition.

I would be happy to ship this though and do a follow up. Not blocking.

lgriffee: Not sure if I'm misinterpreting but [the new solution](https://github.com/Shopify/polaris/pull/5427/commits/214a426dc91ad7050fed9d1024c07a302f21a31e) does use scale not width!

lgriffee: Chatted with @aaronccasanova and I think we'll go ahead and ship this then tomorrow morning. Our solution is better than the previous and doesn’t add an undesirable token to our public API.

alex-page: Nice!

aaronccasanova: Further [simplified the implementation](https://github.com/Shopify/polaris/pull/5427/commits/1a75cb18e09536f12c270a4ce79ca836a6a39984). Basically removed the extraneous `.Animated/.hasAppearAnimation` class and updated to use a single conditional.

.stylelintrc.js
lgriffee: Going to merge this PR and create a separate issue to follow up on this!

alex-page (approved):

polaris-react/src/components/ProgressBar/tests/ProgressBar.test.tsx
```diff
@@ -24,20 +25,34 @@ describe('<ProgressBar />', () => {
...
+  });
+
   describe('animated prop', () => {
-    it('sets the progress bar to include the Animated class by default', () => {
+    it('sets the progress bar CSSTransition to have a non-zero timeout value by default', () => {
       const progress = mountWithApp(<ProgressBar progress={20} />);
-      expect(progress).toContainReactComponent('div', {
-        className: 'Indicator Animated',
+      expect(progress).toContainReactComponent(CSSTransition, {
+        timeout: 500,
```
lgriffee: I was hoping to just make this value any value greater than zero but struggled to find the right syntax.

polaris-react/src/components/ProgressBar/tests/ProgressBar.test.tsx
```diff
@@ -24,20 +25,34 @@ describe('<ProgressBar />', () => {
     expect(progress).toContainReactComponent('progress', {value: 0});
   });
 
+  it('renders a CSSTransition component', () => {
+    const progress = mountWithApp(<ProgressBar />);
+
+    expect(progress).toContainReactComponent(CSSTransition, {
```
lgriffee: Not sure if the props I added below/this entire check for the CSSTranstion is too specific.

polaris-react/src/components/ProgressBar/tests/ProgressBar.test.tsx
```diff
@@ -24,20 +25,34 @@ describe('<ProgressBar />', () => {
...
+      in: true,
+      appear: true,
+      timeout: 500,
```
lgriffee: I was hoping to just make this value any value greater than zero but struggled to find the right syntax.

PR #5429, opened 2022-04-05
Update use-index-resource-state to deselect items that are no longer in the resource list

t-l-hall:

Fixes [https://github.com/Shopify/web/issues/60775](https://github.com/Shopify/web/issues/60775)

Currently if someone has done something that would remove a resource then returns to the page with the resource list, there is a delay while the cache updates to remove that resource. In that time it is possible to select the resource, which is then not removed from the "selectedResources"

This PR adds a `useEffect` that checks for changes in the length of the available resources array and then updates the selected resources to remove anything that is no longer in the array.

### How to 🎩

The details on replicating/testing are available in https://github.com/Shopify/web/issues/60775.

### 🎩 checklist

### DIFF
--- a/polaris-react/src/utilities/tests/use-index-resource-state.test.tsx
+++ b/polaris-react/src/utilities/tests/use-index-resource-state.test.tsx
@@ -50,13 +50,18 @@ describe('useIndexResourceState', () => {
     });
 
     it('accepts initial selected resources', () => {
-      const initialResources = ['1'];
+      const selectedId = '1';
+      const resources = [{id: selectedId}];
+      const initialSelectedResources = [selectedId];
       const mockComponent = mountWithApp(
-        <MockComponent options={{selectedResources: initialResources}} />,
+        <MockComponent
+          resources={resources}
+          options={{selectedResources: initialSelectedResources}}
+        />,
       );
 
       expect(mockComponent).toContainReactComponent(TypedChild, {
-        selectedResources: initialResources,
+        selectedResources: initialSelectedResources,
       });
     });
 
@@ -215,6 +220,24 @@ describe('useIndexResourceState', () => {
           selectedResources: [],
         });
       });
+
+      it('deselects resources that are removed from resource list', () => {
+        const id = '1';
+        const resources = [{id}];
+        const mockComponent = mountWithApp(
+          <MockComponent resources={resources} />,
+        );
+
+        mockComponent
+          .find(TypedChild)!
+          .trigger('onClick', SelectionType.Single, true, id);
+
+        mockComponent.setProps({resources: []});
+
+        expect(mockComponent).toContainReactComponent(TypedChild, {
+          selectedResources: [],
+        });
+      });
     });
 
     describe('SelectionType.All', () => {
@@ -297,9 +320,13 @@ describe('useIndexResourceState', () => {
 
     describe('SelectionType.Multi', () => {
       it('has no effect is selection is undefined', () => {
-        const selectedResources = ['1', '2'];
+        const idOne = '1';
+        const idTwo = '2';
+        const idThree = '3';
+        const resources = [{id: idOne}, {id: idTwo}, {id: idThree}];
+        const selectedResources = [idOne, idTwo];
         const mockComponent = mountWithApp(
-          <MockComponent options={{selectedResources}} />,
+          <MockComponent resources={resources} options={{selectedResources}} />,
         );
 
         mockComponent
--- a/polaris-react/src/utilities/tests/use-index-resource-state.test.tsx
+++ b/polaris-react/src/utilities/tests/use-index-resource-state.test.tsx
@@ -232,6 +232,37 @@ describe('useIndexResourceState', () => {
           .find(TypedChild)!
           .trigger('onClick', SelectionType.Single, true, id);
 
+        expect(mockComponent).toContainReactComponent(TypedChild, {
+          selectedResources: [id],
+        });
+
+        mockComponent.setProps({resources: []});
+
+        expect(mockComponent).toContainReactComponent(TypedChild, {
+          selectedResources: [],
+        });
+      });
+
+      it('deselects resources that are removed from resource list using a custom id resolver', () => {
+        const selectedID = '1';
+        const resources = [{node: {id: selectedID}}];
+        const customIDResolver = (resource: {node: {id: string}}) =>
+          resource.node.id;
+        const mockComponent = mountWithApp(
+          <MockComponent
+            resources={resources}
+            options={{resourceIDResolver: customIDResolver}}
+          />,
+        );
+
+        mockComponent
+          .find(TypedChild)!
+          .trigger('onClick', SelectionType.Page, true);
+
+        expect(mockComponent).toContainReactComponent(TypedChild, {
+          selectedResources: [selectedID],
+        });
+
         mockComponent.setProps({resources: []});
 
         expect(mockComponent).toContainReactComponent(TypedChild, {
--- a/polaris-react/src/utilities/use-index-resource-state.ts
+++ b/polaris-react/src/utilities/use-index-resource-state.ts
@@ -1,6 +1,4 @@
-import {useState, useCallback} from 'react';
-
-import {useDeepEffect} from './use-deep-effect';
+import {useState, useCallback, useEffect} from 'react';
 
 export enum SelectionType {
   All = 'all',
@@ -97,12 +95,16 @@ export function useIndexResourceState<T extends {[key: string]: unknown}>(
     [allResourcesSelected, resources, resourceIDResolver],
   );
 
-  useDeepEffect(() => {
-    const resourceIds = resources.map((resource) => resource.id);
-    setSelectedResources((oldSelectedResources) =>
-      oldSelectedResources.filter((item) => resourceIds.includes(item)),
+  useEffect(() => {
+    const resourceIds = resources.map(resourceIDResolver);
+    const selectedAvailableResources = selectedResources.filter((item) =>
+      resourceIds.includes(item),
     );
-  }, [resources]);
+
+    if (selectedAvailableResources.length !== selectedResources.length) {
+      setSelectedResources(selectedAvailableResources);
+    }
+  }, [resources, selectedResources, resourceIDResolver]);
 
   return {selectedResources, allResourcesSelected, handleSelectionChange};
 }
--- a/polaris-react/src/utilities/use-index-resource-state.ts
+++ b/polaris-react/src/utilities/use-index-resource-state.ts
@@ -100,6 +100,7 @@ export function useIndexResourceState<T extends {[key: string]: unknown}>(
     setSelectedResources((oldSelectedResources) =>
       oldSelectedResources.filter((item) => resourceIds.includes(item)),
     );
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [resources.length]);
 
   return {selectedResources, allResourcesSelected, handleSelectionChange};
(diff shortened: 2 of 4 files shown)


src/utilities/use-index-resource-state.ts
```diff
@@ -95,5 +95,13 @@ export function useIndexResourceState<T extends {[key: string]: unknown}>(
     [allResourcesSelected, resources, resourceIDResolver],
   );
 
+  useEffect(() => {
+    const resourceIds = resources.map((resource) => resource.id);
+    setSelectedResources((oldSelectedResources) =>
+      oldSelectedResources.filter((item) => resourceIds.includes(item)),
+    );
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [resources.length]);
```
AndrewMusgrave: Generally we'll want to avoid using `eslint-disable-next-line react-hooks/exhaustive-deps` wherever possible because the hook will run with stale data.

Right now the hook will run when the length of resources change, however, when a new query is preformed the length may stay the same but with a new set of resources. Rather than running the effect when the length of resources, what about when the resources change? Then we can check if all the ids for resources are including in `selectedResources`. If not we can filter & update the state! Thoughts?

t-l-hall: We originally had it for when resources change, but it went into a pretty bad loop. I was concerned about using `.length` as well just in case of matching lengths. I'll revisit.

AndrewMusgrave: State was probably being set every-time the effect ran. We'll only want to set the state _if_ selected resources need to be set 😄

polaris-react/src/utilities/use-index-resource-state.ts
```diff
@@ -95,5 +97,12 @@ export function useIndexResourceState<T extends {[key: string]: unknown}>(
     [allResourcesSelected, resources, resourceIDResolver],
   );
 
+  useDeepEffect(() => {
+    const resourceIds = resources.map((resource) => resource.id);
+    setSelectedResources((oldSelectedResources) =>
+      oldSelectedResources.filter((item) => resourceIds.includes(item)),
+    );
+  }, [resources]);
```
AndrewMusgrave: @t-l-hall Sorry I won't be able to review the changes till I'm back in Ottawa on Monday.

I noticed you're using `useDeepEffect`, by default it's going to use `lodash/isEqual` which can be fairly costly since resources are often very large. I'm assuming this is because you were running into infinite render issues?

We're setting state on every effect call. We should be able to use `useEffect`, compare our selected resources against the resources ids and _only_ set state when they're _different_

polaris-react/src/utilities/use-index-resource-state.ts
```diff
@@ -95,5 +95,16 @@ export function useIndexResourceState<T extends {[key: string]: unknown}>(
     [allResourcesSelected, resources, resourceIDResolver],
   );
 
+  useEffect(() => {
+    const resourceIds = resources.map(resourceIDResolver);
+    const selectedAvailableResources = selectedResources.filter((item) =>
+      resourceIds.includes(item),
+    );
```
jbalsas, line 102: Using an array here for the ids means that we need to potentially scan the whole array for each `selectedResource`.

I'd expect `resources` to be relatively small, so this shouldn't make much difference in terms of performance, but I think using a `Set` here might be a better fit since we just want to check the existence of elements in the structure. 

I'd suggest something like:

```js
const resourceIds = new Set(resources.map(resourceIDResolver));
const selectedAvailableResources = selectedResources.filter((item) =>
  resourceIds.has(item),
);
```

jbalsas (commented):
Hey @t-l-hall, as we briefly discussed over Slack, I think this works as intended, but **I'd consider it a breaking change** so I'd want to be double sure this is the expected behaviour before moving forward.

I would suggest to first apply the fix in our [`use-index-resource-state` wrapper](https://github.com/Shopify/web/blob/main/app/sections/Orders/Workspace/hooks/use-index-resource-state/use-index-resource-state.ts#L29-L33) as maybe:

```js
const {
  selectedResources: baseSelectedResources,
  allResourcesSelected,
  handleSelectionChange: baseHandleSelectionChange,
} = usePolarisIndexResourceState(resources);

const selectedResources = baseSelectedResources.filter((resourceId) =>
  resources.some((resource) => resource.id === resourceId),
);
```

Here you can see a video of the 2 different behaviours:

https://user-images.githubusercontent.com/905006/165861335-8624319c-fe45-4010-8c62-67b14d8d62a4.mov

https://user-images.githubusercontent.com/905006/165861361-2a0a5455-15c9-44f2-978c-9602c38ba8e1.mov

I can see arguments for and against both behaviours, so the question is... how should this really work?

**Considering that the current behaviour resembles stateful and the fact that many `IndexTable` screens are still powered by the fork in polaris-next, I think we should try to apply a fix internally first and gather more information before committing to a solution here.**

That being said, I think the fix here looks good, just added some non-blocking comments in case we want to move forward with this.

/cc @AndrewMusgrave, @dleroux

jbalsas:
Hey @ouellettejordan and @mrcthms since you'll be working heavily on `IndexTable` do you have already an idea on which direction would would you want to take on this?

Please, let us know if the problem is not clear enough so we can clarify!

mrcthms:
@jbalsas this – and any other current/future bug fixes/improvements to the IndexTable – is out of scope for the work we're doing to the IndexTable. We are focused only on bringing parity between the IndexTable in polaris-next and the IndexTable in polaris-react, so we've already got a fairly large amount of work on 😅 If we start taking ownership or stakeholdership of these decisions then that feels like a slippery slope that our team doesn't really have the bandwidth to go down at present.

polaris-react/src/utilities/tests/use-index-resource-state.test.tsx
```diff
@@ -50,13 +50,18 @@ describe('useIndexResourceState', () => {
     });
 
     it('accepts initial selected resources', () => {
-      const initialResources = ['1'];
+      const selectedId = '1';
+      const resources = [{id: selectedId}];
+      const initialSelectedResources = [selectedId];
```
jbalsas, line 55: I think this change points to an API breaking change. It was possible before to pass freeform data (at least without `id`) and they would get selected properly. 

Now, I think, because we use `resourceIDResolver` to match elements, this fails. I'd expect us to:
- Adjust code so that this use case works (test is untouched)
- Add an additional test with this shape of resource
- Add an additional test with a different resource shape and a different `resourceIDResolver`

polaris-react/src/utilities/tests/use-index-resource-state.test.tsx
```diff
@@ -297,9 +320,13 @@ describe('useIndexResourceState', () => {
 
     describe('SelectionType.Multi', () => {
       it('has no effect is selection is undefined', () => {
-        const selectedResources = ['1', '2'];
+        const idOne = '1';
+        const idTwo = '2';
+        const idThree = '3';
+        const resources = [{id: idOne}, {id: idTwo}, {id: idThree}];
+        const selectedResources = [idOne, idTwo];
```
jbalsas, line 358: Same thing would apply here as in https://github.com/Shopify/polaris/pull/5429/files#r861711173

jbalsas (changes requested):
> That being said, I think the fix here looks good, just added some non-blocking comments in case we want to move forward with this.

~I added some further comments in https://github.com/Shopify/polaris/pull/5429/files#r861711173. I think we need to double check and make sure previous use cases keep working as expected before moving forward.~

We clarified this and will keep it as is for now. See https://github.com/Shopify/polaris/pull/5429#discussion_r862632138 for some more details about this.

polaris-react/src/utilities/tests/use-index-resource-state.test.tsx
jbalsas: We clarified this over a 🍐 ing session on Friday. The fact that resources is [mandatory in the implementation](https://github.com/Shopify/polaris/blob/main/polaris-react/src/utilities/use-index-resource-state.ts#L25) yet [optional in the MockComponent](https://github.com/Shopify/polaris/blob/main/polaris-react/src/utilities/tests/use-index-resource-state.test.tsx#L28) makes for some weird combinations here that don't necessarily match real world use cases.

To keep this in scope at least until we get some clarity about which direction to take based on https://github.com/Shopify/polaris/pull/5429#pullrequestreview-957128548, we decided to keep this as is.

polaris-react/src/utilities/tests/use-index-resource-state.test.tsx
jbalsas: See https://github.com/Shopify/polaris/pull/5429#discussion_r862632138

AndrewMusgrave:
>  but I'd consider it a breaking change

@jbalsas That's fair. I was viewing it as a bug because I see it as an unintended side effect.

This is something that has been asked for by the community to be fixed though. I'll see if I can find one of the posts

AndrewMusgrave:
https://github.com/Shopify/polaris/issues/4346 This is the one I remember seeing of the top of my head, but ive seen others 🤔

jbalsas:
> > but I'd consider it a breaking change
> 
> @jbalsas That's fair. I was viewing it as a bug because I see it as an unintended side effect.
> 
> This is something that has been asked for by the community to be fixed though. I'll see if I can find one of the posts
>
> https://github.com/Shopify/polaris/issues/4346 This is the one I remember seeing of the top of my head, but ive seen others 🤔

Yeah, I think the idea makes sense. The way I see it, we have 2 different reported cases that both stem from a no-longer-valid selection:
- I have a set of thing ids. I delete some things. Some of those ids are no longer needed (nor will ever be)
- I have a set of thing ids. I apply some additional filtering on those things. Some of those ids are no longer needed (but might be later on)

I thought about a couple of options:
- Pass in some `keepState` configuration option to the hook so we can do both behaviours. That way, we have the freedom to choose. We can keep the existing behaviour for now and then flip the default eventually if we feel that's the right call.
- Figure out a way to signal what invalidation strategy to follow upon different actions. This one might be trickier, but it could allow us to, remove ids when the action is "destructive" or keep them when not.

In both cases, though, the component is going to get more complex, so **my suggestion here**, considering you have way more context into this than I do would be:
- **If you're 100% sure this is what we want to do, let's just get it in**
- **If you're not, let's fix this in `web` and try to bring in some additional stakeholders** (UX from order tasks and quality admin and people from Polaris itself)

jbalsas:
Hey @chloerice, @alex-page, I know this might be a lot to digest, but could we get you or someone else from Polaris to weigh in on this one? How do you feel about this? Should we just address the issue inside our case in `web` until we have a better understanding of where this should be headed?

Thanks! 🙏

alex-page:
@jbalsas we need to get web off forked components.

AndrewMusgrave:
@alex-page Just to be clear web isn't using a fork 😄 We have a wrapper around the hook that has added a few callbacks & state we've needed to manage our index table that may eventually make it back to Polaris if we see that people would benefit from them! 🙌

alex-page:
My bad. If this is a unique hook for webs purposes then don't mind my comment.

yurm04:
This PR has been open for a few months and inactive for a bit. Should we close and revisit?

jbalsas:
Hey @yurm04, I think it's entirely up to you! 

The described was addressed at the project level and this remained opened so the Polaris team could weigh in and decide if this was the expected behaviour or an unintended outcome. I'd say that the fact that this felt into oblivion is probably sign that we should just close it and wait for it to pop again in the future. 

Going to do just that to save you a click, but feel free to reopen if you think this is in fact relevant and I'd be happy to help push it to the goal line! 🤗

PR #5430, opened 2022-04-05
Enable linting in stylelint-polaris and fix issues

BPScott:

Part of #5424

Lint the stylelint-polaris folder

 Lint the folder, and fix issues

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -3,8 +3,8 @@ node_modules
 /polaris-react/build
 /polaris-react/build-internal
 /polaris.shopify.com/.next
+/stylelint-polaris/data
 
 
 # These folders should pass linting, remove these lines and fix the errors
 /polaris.shopify.com
-/stylelint-polaris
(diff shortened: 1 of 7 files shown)


.eslintignore
```diff
@@ -3,8 +3,8 @@ node_modules
 /polaris-react/build
 /polaris-react/build-internal
 /polaris.shopify.com/.next
+/stylelint-polaris/data
```
BPScott, line 6: Content in this folder is dynamically generated, so shouldn't be linted

aaronccasanova: Can we add that as a comment: e.g. `# Ignored as the contents of /data is generated`

aaronccasanova (approved):

sam-b-rose (approved):
💯 ✨

.eslintignore
BPScott: It's the same for all the other build and metadata folders above, so I don't really think it's worth it - to me the default reason for inclusion in ignore files is they're generated data, and when that's not the case it's worth a comment

PR #5435, opened 2022-04-05
Enable linting poalris.shopify.com and fix issues

BPScott:

Part of https://github.com/Shopify/polaris/issues/5424

Lint the polaris.shopify.com folder

Lint the folder, and fix issues

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -87,6 +87,15 @@
         "@shopify/jsx-no-hardcoded-content": "off",
         "@shopify/react-initialize-state": "off"
       }
+    },
+    {
+      "files": ["polaris.shopify.com/**/*"],
+      "extends": "plugin:@next/eslint-plugin-next/recommended",
+      "rules": {
+        "react/react-in-jsx-scope": "off",
+        "@shopify/jsx-no-hardcoded-content": "off",
+        "import/no-default-export": "off"
+      }
     }
   ]
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -10,6 +10,9 @@
   "settings": {
     "react": {
       "version": "16.8"
+    },
+    "next": {
+      "rootDir": "polaris.shopify.com"
     }
   },
   "rules": {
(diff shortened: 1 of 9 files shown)


.eslintrc
```diff
@@ -87,6 +87,15 @@
         "@shopify/jsx-no-hardcoded-content": "off",
         "@shopify/react-initialize-state": "off"
       }
+    },
+    {
+      "files": ["polaris.shopify.com/**/*"],
+      "extends": "plugin:@next/eslint-plugin-next/recommended",
```
BPScott, line 96: `eslint-config-next` enables the `plugin:@next/eslint-plugin-next/recommended` config and [a bunch of other things that are already covered by @shopify/eslint-plugin](https://unpkg.com/browse/eslint-config-next@12.1.4/index.js#L54). So here we just want to opt into the `plugin:@next/eslint-plugin-next/recommended` rules.

See https://nextjs.org/docs/basic-features/eslint#recommended-plugin-ruleset

.eslintrc
```diff
@@ -10,6 +10,9 @@
   "settings": {
     "react": {
       "version": "16.8"
+    },
+    "next": {
+      "rootDir": "polaris.shopify.com"
```
BPScott, line 15: See https://nextjs.org/docs/basic-features/eslint#rootdir

.eslintrc
```diff
@@ -87,6 +90,22 @@
...
+      "files": ["polaris.shopify.com/**/*"],
+      "extends": "plugin:@next/eslint-plugin-next/recommended",
+      "rules": {
+        "react/react-in-jsx-scope": "off",
+        "@shopify/jsx-no-hardcoded-content": "off",
+        "import/no-default-export": "off"
+      }
+    },
+    {
+      "files": ["polaris.shopify.com/next-env.d.ts"],
```
BPScott, line 104: This fill seems to get auto regenerated on build, so it seems we can't drop the eslint-disable lines directly in the file

.eslintrc
aaronccasanova: Will our `@shopify/eslint-plugin` rules take precedence? Or will duplicates be overridden by `eslint-plugin-next/recommended`?

aaronccasanova (approved):

.eslintrc
alex-page: I am not super convinced we want the shopify rules overriding the nextjs ones. We want each area of the monorepo to be the best it can be with maximum flexibility. Consolidating these rules early on is not helpful.

Issue #5370, opened 2022-03-29
[Polaris for VSCode] Add contextual awareness on CSS attributes for tokens


yurm04:
When a developer is working in a `.css`/`.scss` file, they should be recommended the appropriate autocomplete tokens for the css attributes that make sense. For example:

`color`, `background-color`, `border-color`, etc. should only autocomplete a list of our color tokens

`margin`, `padding`, etc. should only autocomplete a list of our spacing tokens

`font-size` should only autocomplete a list of our typography tokens

PR #5460, opened 2022-04-06
[Polaris for VSCode] Filter autocomplete tokens by token group

yurm04:

Addresses #5370

This PR will filter autocomplete tokens to only show the ones that are related to the CSS property being written.

* Updating the `generateCustomPropertyNames` script to add token groups
* adds Regex matching to filter for specific CSS properties
* cleans up some prettier issues

### How to tophat 🎩 
0. if you already have the `polaris-for-vscode` extension installed in VSCode, delete it or temporarily disable it.
1. Open a separate VSCode window/workspace with JUST `polaris-for-vscode` directory
2. Build the project with `cmd+shift+B`
3. from the `run and debug` view (`shift+cmd+D`), select the `Client + Server` task from the dropdown and click run
4. a new `[Extension Development Host]` window will open up. In a `.css` or `.scss` file, begin typing a css property to test (ex. color, z-index, box-shadow, etc.). You should only see the corresponding tokens autocompleting
 
https://user-images.githubusercontent.com/4642404/162246131-73658808-d26f-471d-8270-0ad4bef32ef9.mov

### Known issues
* for css properties that are listed in multiple token groups like `shadow` (for color and depth), the logic will only return the first match. In this case, it will only return the color tokens, but not the depth tokens. I don't think this is a blocker to merge this is, but something to fix as a fast follow captured in this issue https://github.com/Shopify/polaris/issues/5469


polaris-react/scripts/utilities/getCustomPropertyNames.js
```diff
@@ -25,4 +25,25 @@ const getCustomPropertyNames = () => {
   return polarisTokenCustomProperties;
 };
 
-module.exports = getCustomPropertyNames;
+const getGroupedCustomPropertyNames = () => {
+  const groupedTokens = {};
+  fs.readdirSync(tokenGroupsDir).map((file) => {
+    const tokenGroup = require(path.join(tokenGroupsDir, file));
+    console.log(file.includes('color'));
```
martenbjork: Some `console.log`s in here, not sure if you wanna keep them  :)

polaris-react/scripts/utilities/getCustomPropertyNames.js
```diff
@@ -25,4 +25,25 @@ const getCustomPropertyNames = () => {
...
+
+    if (file.includes('color')) {
+      console.log('is color');
```
martenbjork: 👆

martenbjork:
Looks great! Love how easy to read this code is 🔥

laurkim (approved):
Nice! 🚀 
Tophatted in web on `.scss` and `.css` files and both filter autocomplete tokens as expected 👍

polaris-for-vscode/server/src/server.ts
```diff
@@ -25,41 +25,83 @@ const connection = createConnection(ProposedFeatures.all);
...
 
+type TokenGroup = {
+  [T in GroupedTokensKey]: RegExp;
+};
+
+type TokenArray = {
+  [key: string]: any;
+};
+
+allTokens as TokenArray;
```
aaronccasanova, line 36:
```suggestion
```
Do we need this assertion? Not entirely sure what it's doing (:

polaris-for-vscode/server/src/server.ts
```diff
@@ -25,41 +25,83 @@ const connection = createConnection(ProposedFeatures.all);
...
+
+let tokenGroups: TokenGroup = {
```
aaronccasanova, line 38:
```suggestion
const tokenGroups: TokenGroup = {
```

polaris-for-vscode/server/src/server.ts
```diff
@@ -25,41 +25,83 @@ const connection = createConnection(ProposedFeatures.all);
 // Create a simple text document manager.
 const documents: TextDocuments<TextDocument> = new TextDocuments(TextDocument);
 
+type TokenGroup = {
+  [T in GroupedTokensKey]: RegExp;
+};
```
aaronccasanova, line 30: [TokenGroup](https://github.com/Shopify/polaris/blob/3329d36020e225e618889dd57a8cd2481d68edfc/polaris-react/src/tokens/tokens.ts#L12-L18) already has an established meaning in our system. Thoughts on renaming this to match the use case e.g. `TokenGroupPatterns`, `TokenGroupToPropertyPatterns`, `CssPropertyPatterns` (naming is hard)

polaris-for-vscode/server/src/server.ts
```diff
@@ -25,41 +25,83 @@ const connection = createConnection(ProposedFeatures.all);
...
+    const doc = documents.get(textDocumentPosition.textDocument.uri);
+
+    // if the doc can't be found, return nothing
+    if (!doc) {
+      return [];
+    }
+
+    const currentText = doc.getText({
+      start: {line: textDocumentPosition.position.line, character: 0},
+      end: {line: textDocumentPosition.position.line, character: 1000},
```
aaronccasanova, line 77: `1000` is probably safe but I if wonder we should leverage VSCode APIs to grab the actual end of line index. I don't know the exact APIs but I put together this rough example [referencing my VSCode extension](https://github.com/aaronccasanova/format-grid-areas/blob/12dc39d1836cd1ded6ab4ec02ad8064d17c2ce68/src/extension.ts#L18-L19):
```js
const currentText = doc.getText({
      start: {line: textDocumentPosition.position.line, character: 0},
      end: {
        line: textDocumentPosition.position.line,
        character: document.lineAt(textDocumentPosition.position).range.end.character,
      },
})
```

polaris-react/scripts/utilities/getCustomPropertyNames.js
```diff
@@ -25,4 +25,24 @@ const getCustomPropertyNames = () => {
...
+
+    const tokenGroupCustomProperties = Object.keys(tokenGroup).map(
+      (token) => `--p-${token}`,
+    );
+
+    groupedTokens[filename.replace('.json', '')] = tokenGroupCustomProperties;
+  });
+
+  return groupedTokens;
+};
```
aaronccasanova, line 46:
```suggestion
const getGroupedCustomPropertyNames = () => {
  return Object.fromEntries(
    fs.readdirSync(tokenGroupsDir).map((fileName) => {
      const tokenGroup = require(path.join(tokenGroupsDir, fileName));
      
      let tokenGroupName = path.basename(fileName, '.json');

      if (fileName.startsWith('color.')) {
        tokenGroupName = 'color';
      }

      const customPropertyNames = Object.keys(tokenGroup).map(
        (token) => `--p-${token}`,
      );

      return [tokenGroupName, customPropertyNames]
    }),
  )
};
```

Optional suggestion. Refactored to use `Object.fromEntries` and `path.basename` API to get retrieve the `tokenGroupName`.

polaris-for-vscode/server/src/server.ts
yurm04: yeah I wen with `1000` because the language server api (from what I could tell) does not have support for getting the end of the line position 😞 

There's probably a better, over-engineered approach that we can take, but for now this was working and I don't currently have a huge concern. This will at least unblock us for now and we can come back to it

polaris-react/scripts/utilities/getCustomPropertyNames.js
yurm04: i like this 💯 and had no idea this was a thing! do you mind creating a separate PR against `main` after we merge this PR so that we can unblock this current stream of work quickly?

PR #5466, opened 2022-04-07
Type check all projects from the root

### LINKED ISSUE #5432: Create a top-level `tscofig.json` for the entire monorepo

sam-b-rose:

Fixes #5432

* Add top-level `type-check` script
* Leverage project references for better build performance and modularity across projects
* Add `yarn run type-check` to `ci.yml`

This PR also:

* Add `polaris-for-vscode` to the yarn workspace
* Remove unused `tsconfig.json` and type check script from `stylelint-polaris`

Deferred adding a single `tsconfig.base.json` until we have better understanding of common tsconfig needs. I tried adding a common base config, but it seemed forced. We can revisit creating a base config in the future.

## How to test

Run `yarn type-check` from the root directory.


package.json
```diff
@@ -20,7 +21,8 @@
     "lint": "yarn run lint:eslint && yarn run lint:stylelint && yarn run lint:prettier",
     "lint:eslint": "eslint . --max-warnings 0 --cache --cache-location '.loom/cache/eslint'",
     "lint:stylelint": "yarn workspace '@shopify/stylelint-polaris' run gen-polaris-vars && stylelint '**/*.scss'",
-    "lint:prettier": "prettier '**/*.{md,json,yaml,yml}' --check"
+    "lint:prettier": "prettier '**/*.{md,json,yaml,yml}' --check",
+    "type-check": "tsc --build"
```
aaronccasanova, line 25: **question:** Why do we run a build for the `type-check`? I was imagining this would be `tsc --noEmit`.

sam-b-rose: It is a confusing flag name to use, but that is what the TypeScript docs suggest on their [project references page](https://www.typescriptlang.org/docs/handbook/project-references.html#build-mode-for-typescript) 🤷

stylelint-polaris/tsconfig.json
```diff
@@ -1,11 +0,0 @@
...
-  "extends": "@shopify/typescript-configs/node-library.json",
-  "compilerOptions": {
-    "baseUrl": ".",
-    "outDir": "dist",
-    "allowJs": true,
-    "checkJs": true,
-    "skipLibCheck": true
-  },
-  "exclude": ["node_modules", "dist"]
-}
```
aaronccasanova: Please keep JSDoc type validation in this library.

sam-b-rose: Gotcha, can do! There is a type error though when trying to assign `module.exports.ruleName`. I'll need to fix that in order for this to pass

aaronccasanova: I think I remember that one! Pretty sure I just changed `const` to `let`. I can help look into it if that doesn't resolve the issue (Don't want to block this work).

package.json
aaronccasanova: Thanks for the link! Did you confirm it doesn't output any `.js` assets, just simply runs a type check?

aaronccasanova: > Build mode (see below) will automatically build the referenced project if needed

From the link it looks like --build could produce assets. Perhaps, that's only if `outDir` or `outFile` is specified in the child tsconfig's `compilerOptions`.

BPScott: We want it to produce d.ts files when running type-check. This is because incremental builds - "hey you changed a file in the polaris-react project, typecheck the polaris-react project, and anything that depends polaris-react)" relies upon the d.ts files being present. We can ensure that `.d.ts`  are created, but no js files are created by using `"noEmit": false, "emitDeclarationOnly": true,` in the tsconfig files.

BPScott: We want it to produce d.ts files when running type-check. This is because incremental builds - "hey you changed a file in the polaris-react project, typecheck the polaris-react project, and anything that depends polaris-react)" relies upon the d.ts files being present. We can ensure that `.d.ts`  are created, but no js files are created by using `"noEmit": false, "emitDeclarationOnly": true,` in the tsconfig files.

polaris-for-vscode/tsconfig.json
```diff
@@ -6,9 +6,13 @@
     "outDir": "out",
```
BPScott, line 6: I think it'd be useful to consolidate on a single convention for a compilation output directory. Currently polaris-react uses `build`, polaris-for-vscode uses `out`, and stylelint-polaris doesn't have anything (but it should because d.ts files should be generated for incremental compilation

stylelint-polaris/plugins/custom-properties-allowed-list/index.js
```diff
@@ -147,8 +147,11 @@ function validateCustomPropertyValues(allowedValues, prop, value) {
   return invalidValues;
 }
 
-module.exports.ruleName = ruleName;
-module.exports.messages = messages;
+module.exports = {
```
BPScott, line 150: This is changing the shape of what is exported. It used to be a function with some keys (yay prototypal inheritance) and now it's an object that contains a function. I'm pretty sure this isn't considered a valid stylelint rule anymore.

stylelint-polaris/tsconfig.json
```diff
@@ -5,7 +5,8 @@
     "outDir": "dist",
```
BPScott, line 5: Perhaps use `build` here to match polaris-react?

stylelint-polaris/tsconfig.json
```diff
@@ -5,7 +5,8 @@
...
     "allowJs": true,
     "checkJs": true,
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "noEmit": true
```
BPScott: I think we want a `composite: true, noEmit: false, emitDeclarationOnly:true` in here, so that the incremental compile can reference the d.ts files from a prior compilation.

Where does the tsbuildinfo file for this package get generated?

.github/workflows/ci.yml
```diff
@@ -25,6 +25,7 @@ jobs:
         with:
           path: |
             .loom
+            *.tsbuildinfo
```
BPScott: I'm pretty sure this would only match a `.tsbuildinfo` file at the root of the repo - which doesn't exist as these get created on a per-project basis.

.gitignore
```diff
@@ -1,4 +1,5 @@
 *.log
+*.tsbuildinfo
```
BPScott, line 2: If we ignore all the build output folders and the tsbuildinfo files go into the build output folders, then we shouldn't to reference this filename specifically

.prettierignore
```diff
@@ -1,8 +1,10 @@
 node_modules
 .loom
+.vscode-test
```
BPScott, line 3: These output folders should also be added to the eslintignore and stylelintignore too

BPScott (commented):
Good hunting to work out how to get the polaris-react tsconfig behaving!

I've dropped a few comments inline. Notably I think it'd be worth consolidating on a convention for what to name the compilation output directory for each project. That'd make ignorefiles and tsconfig setup a bit more consistent across projects.

package.json
aaronccasanova: > When you reference a project, new things happen:
> - Importing modules from a referenced project will instead load its output declaration file (.d.ts)

Oh nice @BPScott! Confirmed in the above docs ☝️ 

So basically every project should include at a minimum `"noEmit": false, "emitDeclarationOnly": true, "outDir": "some-dist"` and flip `noEmit` to `true` if the project is built with `tsc` e.g. `polaris-for-vscode`?

polaris-for-vscode/tsconfig.json
aaronccasanova: My vote is `dist`. It's used in all the TypeScript docs/examples, default output directory for `webpack` and `rollup`, `parcel` literally calls their option `distDir`, and it's the most common convention I've seen in open source land.

stylelint-polaris/tsconfig.json
aaronccasanova: https://github.com/Shopify/polaris/pull/5466#discussion_r846502024

package.json
BPScott: Correct for files not using tsc to build - `"noEmit": false, "emitDeclarationOnly": true, "outDir": "some-dist"`

Projects built with tsc would use `noEmit: false, emitDeclarationOnly:false`. (setting noEmit:true would stop all output. double negatives are confusing.

stylelint-polaris/tsconfig.json
BPScott: Oooh eslint flipped that back to `const` in #5430. Guess we'll need an eslint disable line to keep that as let.

polaris-for-vscode/tsconfig.json
BPScott: A minor spanner - the rollup config defined in loom-plugin-build-library uses `build` for the output directory. It's possible to override that through tapping into the rollupConfig hooks, but it's a bit clunky

package.json
aaronccasanova: Awesome, makes sense, thanks for the breakdown @BPScott!!

@samrose3 Feel free to resolve this conversation

stylelint-polaris/plugins/custom-properties-allowed-list/index.js
sam-b-rose: Looking at `createPlugin` [source](https://github.com/stylelint/stylelint/blob/main/lib/createPlugin.js), I think we are safe. The `createPlugin` function returns an object with a rule and ruleName property. I've seen this export pattern used in other stylelint plugins as well 👍

stylelint-polaris/tsconfig.json
sam-b-rose: Agreed that we should have a common naming convention for build output. But, going to address this decision in a later clean up issue.

.gitignore
sam-b-rose: Not all of the tsconfigs do specify a build output directory (ex: `polaris.shopify.com`). So we may way to include this just in case

.github/workflows/ci.yml
sam-b-rose: Fixed! Good catch!

.prettierignore
sam-b-rose: Added to the other ignores 👍

stylelint-polaris/tsconfig.json
sam-b-rose: Enabled composite 👍  Ty

aaronccasanova (approved):
image

Amazing 🎉 🚀

stylelint-polaris/plugins/custom-properties-allowed-list/index.js
BPScott: well whoops, that's not what I thought it did. Cool, ignore me :)

BPScott (approved):
LGTM! :shipit:

PR #5471, opened 2022-04-07
feat(POC): Add opacity tokens

aaronccasanova:

This (POC) PR adds an `opacity` token-group, consisting of `20` (linear) scaled tokens, to increase `opacity` coverage to 100%:
- The `opacity` tokens increase in increments of `0.5`
- Hardcoded `opacity`'s have been updated by rounding up to the nearest matching token

TODO:
- [ ] - Add `UNRELEASED` entry if we decide to proceed with the proposed implementation

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/polaris-react/src/components/Select/Select.scss
+++ b/polaris-react/src/components/Select/Select.scss
@@ -84,8 +84,7 @@
   width: 100%;
   height: 100%;
   margin: 0;
-  // ChromeVox may hide content set to opacity: 0
-  opacity: 0.001;
+  opacity: 0;
   appearance: none;
 }
 
--- a/polaris-react/src/tokens/token-groups/opacity.json
+++ b/polaris-react/src/tokens/token-groups/opacity.json
@@ -1,0 +1,23 @@
+{
+  "opacity-0": "0",
+  "opacity-1": "0.05",
+  "opacity-2": "0.1",
+  "opacity-3": "0.15",
+  "opacity-4": "0.2",
+  "opacity-5": "0.25",
+  "opacity-6": "0.3",
+  "opacity-7": "0.35",
+  "opacity-8": "0.4",
+  "opacity-9": "0.45",
+  "opacity-10": "0.5",
+  "opacity-11": "0.55",
+  "opacity-12": "0.6",
+  "opacity-13": "0.65",
+  "opacity-14": "0.7",
+  "opacity-15": "0.75",
+  "opacity-16": "0.8",
+  "opacity-17": "0.85",
+  "opacity-18": "0.9",
+  "opacity-19": "0.95",
+  "opacity-20": "1"
+}
--- a/polaris-react/src/components/Button/Button.scss
+++ b/polaris-react/src/components/Button/Button.scss
@@ -395,7 +395,7 @@
     &.disabled {
       color: currentColor;
       @include recolor-icon(currentColor);
-      opacity: 0.4;
+      opacity: var(--p-opacity-4);
     }
 
     // Prevents the loading spinner from overlapping the button’s text, while retaining its width.
@@ -451,7 +451,7 @@
       color: currentColor;
 
       &::before {
-        opacity: 0.2;
+        opacity: var(--p-opacity-4);
       }
     }
 
@@ -462,7 +462,7 @@
       border-color: currentColor;
 
       &::before {
-        opacity: 0.07;
+        opacity: var(--p-opacity-2);
       }
     }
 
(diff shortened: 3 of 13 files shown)


polaris-react/src/components/Select/Select.scss
```diff
@@ -84,8 +84,7 @@
   width: 100%;
   height: 100%;
   margin: 0;
-  // ChromeVox may hide content set to opacity: 0
-  opacity: 0.001;
```
aaronccasanova, line 88: Note: Validate before merging

polaris-react/src/tokens/token-groups/opacity.json
```diff
@@ -0,0 +1,23 @@
...
+  "opacity-11": "0.55",
+  "opacity-12": "0.6",
+  "opacity-13": "0.65",
+  "opacity-14": "0.7",
+  "opacity-15": "0.75",
+  "opacity-16": "0.8",
+  "opacity-17": "0.85",
+  "opacity-18": "0.9",
+  "opacity-19": "0.95",
+  "opacity-20": "1"
```
alex-page, line 22: Curious on folks thoughts around 20 values. Could we start with less values? Is 20 just right? Should we have tokens for `0` and `1`?

alex-page (approved):

polaris-react/src/components/Select/Select.scss
alex-page: I feel like [ChromeVox](https://chrome.google.com/webstore/detail/screen-reader/kgejglhpjiefppelpmljglcjbhoiplfn?hl=en) is not a tool we should be prioritising support for. It hasn't received any updates in 8+ months and usually we prioritise other screen readers. @svinkle any thoughts?

svinkle: Just verified: ChromeVox does not hide content when set to `opacity: 0;`. Use [Visually hidden](https://polaris.shopify.com/components/titles-and-text/visually-hidden) instead.

For context, we do not support ChromeVox the browser extension. We do support [ChromeVox](https://chromevox.com) the screen reader available on ChromeOS as many Shopifolk with visual impairments use Chromebooks.

polaris-react/src/tokens/token-groups/opacity.json
aveline: I think 20 values is too many. I'm not sure it adds much value to tokenize them. Essentially we are just saying "stick to one decimal place or halves" and from doing a quick search in `web`, that seems to already be an established pattern.

alex-page: What would you recommend the amount of values be @aveline?

aveline: I am actually questioning the value of having opacity tokens at all. So often the "right" opacity is so contextual based on the colours used and background.

I could see valid use cases for all the values @aaronccasanova is including above, but just less convinced that mapping them to tokens adds value.

Tokens can help narrow down possibilities of values, for example a collection of colour tokens narrows it down from 16+ million options. The options for opacity are already pretty limited in that it needs to be a value between `1` and `0`. Do we need to encourage consumers to not set an opacity of `0.233645` when it doesn't seem like they are trying to do that anyway? 🤷‍♀️

If anything I see fewer than 20 opacity values in the `web` codebase right now. So why create tokens for values no one may ever use?

Although maybe we do want to prevent the few random `opacity: 100` 😬

lgriffee: From a quick pass through web I ended up finding 19 different (17 not including 1 and 0) opacity values: 

- 0
- 0.02
- 0.10
- 0.20
- 0.25
- 0.30
- 0.35
- 0.40
- 0.43
- 0.45
- 0.50
- 0.60
- 0.70
- 0.80
- 0.85
- 0.90
- 0.95
- 1
- 100

Even though it's a lot of values I do think it helps to maintain consistency and give guidance/an opinion on how we want to handle opacity. There were a few instances in Polaris in this PR and web that are more random / don't follow the 0.05 step convention.

lgriffee (approved):
💯🚢✨🎉

polaris-react/src/tokens/token-groups/opacity.json
alex-page: Do we expect to ship a change where we would make the UI more opaque or transparent. If the answer is no then I would consider letting opacity be an ignored value in coverage and not adding tokens. I feel it is too early to know if we get value from controlling opacity from the design system.

aaronccasanova:
Closing in favor of https://github.com/Shopify/polaris/pull/5490

PR #5476, opened 2022-04-08
Add test for stylelint-polaris

alex-page:
- [x] Add test for stylelint-polaris
- [x] Add comments for regex
- [x] Add additional layout properties to list


stylelint-polaris/tests/index.js
```diff
@@ -0,0 +1,31 @@
+/* eslint-disable no-console */
+const fs = require('fs');
+const path = require('path');
+
+const stylelint = require('stylelint');
+
+const fixturePath = path.join(process.cwd(), 'tests/fixture.css');
```
aaronccasanova:
```suggestion
const fixturePath = path.join(__dirname, 'fixture.css');
```
`process.cwd()` will not point to the `stylelint-polaris` dir if ran from the root of the monorepo. `__dirname` should have more consistent behavior.

stylelint-polaris/tests/index.js
```diff
@@ -0,0 +1,31 @@
...
+const fileContent = fs.readFileSync(fixturePath, 'utf-8');
+
+(async () => {
+  const {results} = await stylelint.lint({
+    config: {
+      extends: '@shopify/stylelint-polaris',
```
aaronccasanova:
```suggestion
      extends: '../index.js',
```
Does this need to be a relative path to the `stylelint-polaris/index.js` file? Not immediately understanding how `config.extends` is able to resolve this bare module specifier (`@shopify/stylelint-polaris`).

stylelint-polaris/configs/coverage.js
```diff
@@ -14,14 +14,18 @@ module.exports = {
     'declaration-property-value-disallowed-list': [
       {
         display: ['grid', 'flex'],
+        // No numeric values
```
aaronccasanova: These are currently disallowing Scss vars: https://regex101.com/r/p6ywGC/1

stylelint-polaris/configs/coverage.js
```diff
@@ -14,14 +14,18 @@ module.exports = {
...
         top: [/\$.+/],
         bottom: [/\$.+/],
         left: [/\$.+/],
         right: [/\$.+/],
         width: [/\$.+/],
         height: [/\$.+/],
+        // zero and one only, no numeric values or additions
```
aaronccasanova: This currently allows `0`, `1`, and custom properties: https://regex101.com/r/s8IpNe/1

stylelint-polaris/configs/coverage.js
```diff
@@ -14,14 +14,18 @@ module.exports = {
...
         opacity: [/(?!0|1)\d|[\d.]{2,}/],
+        // polaris custom properties only
```
aaronccasanova: With this prompt, I think we need to update the regex: https://regex101.com/r/R8u8Ck/1

stylelint-polaris/configs/coverage.js
```diff
@@ -14,14 +14,18 @@ module.exports = {
...
         'z-index': [/(\$.*|-?[0-9]+)/],
+        // polaris custom properties only
```
aaronccasanova: Same as above

stylelint-polaris/tests/fixture.css
```diff
@@ -0,0 +1,53 @@
+:root {
+  --pc-omg: orange;
+  --junk: brown;
```
aaronccasanova, line 3: Just so I'm clear this text fixture outlines failing rules in a Polaris consumer project? I ask because `--junk` would be valid in a consumers `.scss` file.

stylelint-polaris/tests/index.js
```diff
@@ -0,0 +1,31 @@
+/* eslint-disable no-console */
+const fs = require('fs');
+const path = require('path');
+
+const stylelint = require('stylelint');
+
+const fixturePath = path.join(process.cwd(), 'tests/fixture.css');
+const fileContent = fs.readFileSync(fixturePath, 'utf-8');
+
+(async () => {
```
sam-b-rose, line 10: Do we have support for top-level `await` in Node >14.8? We are on 16 so I think we could remove this IIFE

stylelint-polaris/tests/index.js
sam-b-rose: Good point here... I think `@shopify/stylelint-polaris` would resolve to the symlinked resources in the `node_modules` folder.

So it should be pointing to the `main` file defined in the `package.json`. In this case, either should be fine.

stylelint-polaris/package.json
```diff
@@ -32,6 +32,7 @@
   "scripts": {
     "gen-polaris-vars": "node ./scripts/gen-polaris-vars.js",
     "build": "yarn gen-polaris-vars",
+    "test": "node tests/index.js",
```
sam-b-rose, line 32: I like the simplicity of using a test script 💯

sam-b-rose (approved):
Looks good! Great idea adding this @alex-page 👍

stylelint-polaris/tests/fixture.css
alex-page: I don't mind if it's a polaris consumer or polaris-react project. I think we just need a baseline test to get started.

stylelint-polaris/tests/index.js
alex-page: I thought so as well. Maybe I didn't `dev up`. Let me try again.

alex-page: Seems like it's `.mjs` files only

stylelint-polaris/configs/coverage.js
alex-page: I think we need to update all the regexes here to only pass on polaris custom properties. This would give us a good indicator of all the times folks have to do something custom. Would you be able to help with this @aaronccasanova?

I have `/^(?!var\(--p-.+\)$)/` but I don't know if it's just right 😬

stylelint-polaris/tests/index.js
aaronccasanova: So just by nature of `stylelint-polaris` being a yarn workspace the package is symlinked in the root `node_modules` 🤯 ? For some reason I thought it had to be referenced in the `"dependencies":{}` of another workspace's `package.json`, TIL!

alex-page: I think it makes sense to do the `../` for the test though. Good learnings anyway :)

stylelint-polaris/configs/coverage.js
```diff
@@ -14,14 +14,15 @@ module.exports = {
...
-        'z-index': [/(\$.*|-?[0-9]+)/],
+        // polaris custom properties only
+        top: [/^(?!var\(--p-.+\)$)/],
+        bottom: [/^(?!var\(--p-.+\)$)/],
+        left: [/^(?!var\(--p-.+\)$)/],
+        right: [/^(?!var\(--p-.+\)$)/],
+        width: [/^(?!var\(--p-.+\)$)/],
+        height: [/^(?!var\(--p-.+\)$)/],
+        opacity: [/^(?!var\(--p-.+\)$)/],
+        'z-index': [/^(?!var\(--p-.+\)$)/],
```
alex-page:
```suggestion
        top: [/^(?!var\(--p-).+$/],
        bottom: [/^(?!var\(--p-).+$/],
        left: [/^(?!var\(--p-).+$/],
        right: [/^(?!var\(--p-).+$/],
        width: [/^(?!var\(--p-).+$/],
        height: [/^(?!var\(--p-).+$/],
        opacity: [/^(?!var\(--p-).+$/],
        'z-index': [/^(?!var\(--p-).+$/],
```

BPScott:
It feels like it'd be nicer if this was a jest test that leverages the tooling in https://github.com/stylelint/jest-preset-stylelint in the style of https://github.com/prettier/stylelint-prettier/blob/main/test/stylelint-prettier.test.js, but I guess this works till we can have jest running from the root of the repo, across all projects.

stylelint-polaris/tests/index.js
```diff
@@ -0,0 +1,31 @@
...
+
+const stylelint = require('stylelint');
+
+const fixturePath = path.join(__dirname, 'fixture.css');
+const fileContent = fs.readFileSync(fixturePath, 'utf-8');
+
+(async () => {
+  const {results} = await stylelint.lint({
+    config: {
+      extends: '../index.js',
```
aaronccasanova:
```suggestion
      extends: path.join(__dirname, '../index.js'),
```
This was needed for me locally

BPScott:
Should we add a step to run this as part of the CI job?

PR #5479, opened 2022-04-08
Deploy site when code is merged into main

sam-b-rose:

Fixes https://github.com/Shopify/polaris-site-container-builder/issues/6 

This PR will tell Shopify Build to trigger the `polaris-site-container-builder` pipeline to build and deploy the site.

Uses the Shopify Build [trigger step](https://shopify-build.docs.shopify.io/pipelines/defining_steps/#step-type-trigger) feature.

### DIFF
--- a/.github/workflows/site.yml
+++ b/.github/workflows/site.yml
@@ -1,0 +1,26 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - main
+    paths:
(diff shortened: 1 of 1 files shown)


sam-b-rose:
This solution required a PAT which is not an ideal practice. Going to close out this PR and Issue in favor of using a manual method of deploying the Polaris site which saves us from a lot of automation complexity.

PR #5481, opened 2022-04-08
Register doc and ts-3.4 creation as a package task

BPScott:

This shall make it easier to keep this task polaris-react specific in a future where we introduce a workspace level loom config

Changes when the doc and ts-3.4 file are generated - they're now done as part of the package build hook, rather than tan a workspace post-build hook

## 

### How to 🎩

`yarn workspace @shopify/polaris run build ` and note that the `executing command: "node_modules/.bin/downlevel-dts" "build/ts/latest" "build/ts/3.4"` and `executing command: "node_modules/.bin/copyfiles" "./src/**/*.md" "./build/docs" "--up=1"` commands are still ran, and that output is generated in `polaris-react/build/ts/3.4`  and `polaris-react/build/docs`

### DIFF
--- a/polaris-react/loom.config.ts
+++ b/polaris-react/loom.config.ts
@@ -1,6 +1,5 @@
 import {
   createPackage,
-  createWorkspacePlugin,
   createProjectPlugin,
   DiagnosticError,
 } from '@shopify/loom';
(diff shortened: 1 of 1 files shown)


alex-page:
> in a future where we introduce a workspace level loom config

I don't want to keep saying this. We aren't planning on having a workspace level loom config. @bpscott it's great to get these improvements but we don't want to prioritise using loom at a workspace level. We might even consider removing it all together.

Today we want:
1. Maximum flexibility to allow our team to diverge and create pain points (these will be fixed later)
2. Prioritise open source tools so that third parties can access their documentation and learnings
3. 1% better every day, we aren't focusing on setting up long term infrastructure yet. We will address large improvements or refactors in vault projects.

sam-b-rose (approved):
#5466 has been merged. This LGTM, merge when ready 👍

PR #5484, opened 2022-04-08
Vscode color previews for color tokens

### LINKED ISSUE #5369: [Polaris for VSCode] Add documentation to all of the token autocomplete items

yurm04:

Fixes #5369

Screen Shot 2022-04-08 at 4 30 31 PM

* adding color previews to color tokens
* adding token value as the description to token autocomplete items

### DIFF
--- a/polaris-for-vscode/scripts/prepublish.js
+++ b/polaris-for-vscode/scripts/prepublish.js
@@ -13,8 +13,9 @@ const filePath = path.join(dirPath, outFile);
 const groupedTokens = getGroupedCustomPropertyNames();
 const tokens = getCustomPropertyNames();
 
-// we don't need legacy tokens
+// we don't need legacy or color.dark tokens
 delete groupedTokens['legacy-tokens'];
+delete groupedTokens['color.dark'];
 
 try {
   if (!fs.existsSync(dirPath)) {
--- a/polaris-for-vscode/server/src/server.ts
+++ b/polaris-for-vscode/server/src/server.ts
@@ -13,6 +13,7 @@ import {
 import {TextDocument} from 'vscode-languageserver-textdocument';
 
 import {allTokens, groupedTokens} from './data/allTokens';
+import { match } from 'assert';
 
 type GroupedTokens = typeof groupedTokens;
 
@@ -40,6 +41,14 @@ let tokenGroupPatterns: TokenGroupPatterns = {
   motion: /animation/,
 };
 
+type GroupedCompletionItem = {
+  [T in GroupedTokensKey]: CompletionItem[];
+};
+
+let groupedCompletionItems: Partial<GroupedCompletionItem> = {};
+
+let allCompletionItems: CompletionItem[] = [];
+
 connection.onInitialize((params: InitializeParams) => {
   const capabilities = params.capabilities;
 
@@ -53,6 +62,29 @@ connection.onInitialize((params: InitializeParams) => {
     },
   };
 
+  // initialize grouped completionItems
+  for(const tokenGroup in groupedTokens) {
+    const category = tokenGroup as keyof typeof tokenGroupPatterns;
+    const tokensArray = groupedTokens[category];
+
+    const completionItems = tokensArray.map((token): CompletionItem => {
+      return {
+        label: token.label,
+        insertText: token.insertText,
+        detail: token.value,
+        filterText: `--p-${token.label}`,
+        kind:
+          category === 'color'
+            ? CompletionItemKind.Color
+            : CompletionItemKind.Variable,
+      };
+    });
+
+    groupedCompletionItems[category] = completionItems;
+
+    allCompletionItems = allCompletionItems.concat(completionItems);
+  }
+
   return result;
 });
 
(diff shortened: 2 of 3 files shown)


polaris-for-vscode/scripts/prepublish.js
```diff
@@ -13,8 +13,9 @@ const filePath = path.join(dirPath, outFile);
 const groupedTokens = getGroupedCustomPropertyNames();
 const tokens = getCustomPropertyNames();
 
-// we don't need legacy tokens
+// we don't need legacy or color.dark tokens
 delete groupedTokens['legacy-tokens'];
+delete groupedTokens['color.dark'];
```
aaronccasanova, line 18: There's no `groupedTokens` with a `color.dark` key: https://github.com/Shopify/polaris/blob/20ea614e6d47566374d61eb2519008fa77d0e454/polaris-react/scripts/utilities/getCustomPropertyNames.js#L34-L36

You may need to filter that out in the `getGroupedCustomPropertyNames` util e.g.
`fs.readdirSync(tokenGroupsDir).filter(fileName => !fileName.startsWith('color.dark')).map(...)`

polaris-for-vscode/server/src/server.ts
```diff
@@ -13,6 +13,7 @@ import {
 import {TextDocument} from 'vscode-languageserver-textdocument';
 
 import {allTokens, groupedTokens} from './data/allTokens';
+import { match } from 'assert';
```
aaronccasanova:
```suggestion
```
I'm guessing VSCode was trying to help you out

polaris-for-vscode/server/src/server.ts
```diff
@@ -40,6 +41,14 @@ let tokenGroupPatterns: TokenGroupPatterns = {
   motion: /animation/,
 };
 
+type GroupedCompletionItem = {
+  [T in GroupedTokensKey]: CompletionItem[];
+};
+
+let groupedCompletionItems: Partial<GroupedCompletionItem> = {};
```
aaronccasanova, line 47:
```suggestion
type GroupedCompletionItems = {
  [T in GroupedTokensKey]?: CompletionItem[];
};

let groupedCompletionItems: GroupedCompletionItems = {};
```
Didn't think about this during our pairing, but we can just remove the `Partial` utility type in favor of making the properties optional in the mapped type.

Note: The above PR suggestion includes a type name update to `GroupedCompletionItems` (plural) to match the variable it's assigned to (:

polaris-react/scripts/utilities/getCustomPropertyNames.js
```diff
@@ -31,12 +31,17 @@ const getGroupedCustomPropertyNames = () => {
...
+      // rename the color.light tokens
+      if (fileName.includes('.light')) {
         tokenGroupName = 'color';
       }
-      const customPropertyNames = Object.keys(tokenGroup).map(
-        (token) => `--p-${token}`,
-      );
+      const customPropertyNames = Object.keys(tokenGroup).map((token) => {
+        return {
+          label: token,
```
aaronccasanova, line 40:
```suggestion
          name: token,
```
We've been loosely following the Design Token Format Module terminology. `Token Group`, `Token Name` and `Token Alias`: https://design-tokens.github.io/community-group/format/#example-14

polaris-react/scripts/utilities/getCustomPropertyNames.js
```diff
@@ -31,12 +31,17 @@ const getGroupedCustomPropertyNames = () => {
...
       let tokenGroupName = path.basename(fileName, '.json');
-      if (fileName.startsWith('color.')) {
+      // rename the color.light tokens
+      if (fileName.includes('.light')) {
         tokenGroupName = 'color';
       }
-      const customPropertyNames = Object.keys(tokenGroup).map(
-        (token) => `--p-${token}`,
-      );
+      const customPropertyNames = Object.keys(tokenGroup).map((token) => {
```
aaronccasanova, line 38: The scope of this utility has grown beyond retrieving grouped custom property names. Naming is hard, so I don't have any immediate suggestions, but what are you thoughts on renaming?

romellogoodman: Quick question, do we imagine that in the long term this functionality would live within the tokens package?

romellogoodman (approved):
verified locally that the 🎨 colors 🎨 come through!

martenbjork (approved):
🚀

polaris-react/scripts/utilities/getCustomPropertyNames.js
aaronccasanova: That's a tough one. The return value is quite specific to the `CompletionItem` results where it's accessed in the VSCode extension. The only reason I can think to keep this utility with the tokens package is to ensure it stays in sync with any unique parser behavior we establish for custom properties.

An example of unique parser behavior is in the `motion.json` tokens. Where our style utilities generate CSS keyframes and CSS custom properties when they encounter a token name (in `motion.json`) that starts with `keyframes`. Note: I've seen this pattern in other design tokens libs (e.g. open-props) so where are not inventing something new here.

If this utility were to live elsewhere we'd likely have to add some more guard rails (for example tests) to validate the VSCode ext UI is in sync with our output custom properties.

PR #5487, opened 2022-04-11
Default to `React.StrictMode`

kyledurand:
I believe we want strict mode by default but I could be mistaken

### DIFF
--- a/polaris-react/.storybook/preview.js
+++ b/polaris-react/.storybook/preview.js
@@ -44,7 +44,7 @@ function GridOverlayDecorator(Story, context) {
 export const globalTypes = {
   strictMode: {
     name: 'React.StrictMode',
-    defaultValue: false,
+    defaultValue: 'true',
     toolbar: {
       items: [
         {title: 'Disabled', value: 'false'},


polaris-react/.storybook/preview.js
```diff
@@ -28,7 +28,7 @@ function AppProviderDecorator(Story, context) {
 export const globalTypes = {
   strictMode: {
     name: 'React.StrictMode',
-    defaultValue: 'false',
+    defaultValue: 'true',
```
alex-page, line 47: I think we want to match `shopify/web`, if they use strict by default this would be fine to merge in.

alex-page (approved):

polaris-react/.storybook/preview.js
BPScott: StrictMode has zero effect in production, it's just a development tool used to [check for correctness](https://reactjs.org/docs/strict-mode.html). I'd argue that helping ensure correct behaviour when developing in polaris-react is worthwhile even if web does not do that currently in its local development. Having polaris be strict mode compliant would make it easier for web and other consuming apps to adopt strict mode in their own codebases.

kyledurand: It doesn't look like they do which seems weird. We've always defaulted to strict mode in Polaris though. @BPScott any thoughts on this? Maybe we enable strict mode for storybook web as well?

BPScott: Oop sorry, that fell through the cracks.

Enabling strict mode in web's storybook sounds like a good idea to me! (also in their jest test harness)

PR #5494, opened 2022-04-11
[DataTable] Add sticky header

blboyle:

This allows the user to be able to see the headings as they scroll along a table that might be taller than the window.

Uses a similar approach as does the sticky header in the IndexTable component.

Adds sticky header to the data table
https://screenshot.click/14-02-gdpby-pakq6.mp4

Keyboard navigation
https://screenshot.click/18-21-i8nfi-rzq69.mp4

Connecting horizontal scrolling between sticky header and table
https://screenshot.click/19-41-4tlvh-iz7vg.mp4

Using storybook build 
https://screenshot.click/20-54-0ucq4-p235a.mp4
Command for storybook build: `yarn run storybook:build`
Then visit, you will see a build version of storybook including your playground.

--

UX: @mirualves 
Resolves: https://github.com/Shopify/core-issues/issues/36867

## 

### How to 🎩

<details>

```jsx
import React from 'react';

import {Button, Card, ColumnContentType, DataTable, Page} from '../src';

  const columnTypes: ColumnContentType[] = [
    'text',
    'numeric',
    'numeric',
    'numeric',
    'numeric',
    'numeric',
    'numeric',
    'numeric',
    'numeric',
    'numeric',
    'numeric',
    'numeric',
  ];
  const headings = [
    'Things',
    'Product',
    'Orders',
    'Spaceships',
    'All',
    'More Things',
    'Products',
    'Orders again',
    'Rockets',
    'Whence',
    'Subtotal',
    'Total',
  ];
  const rows = [
    ['One', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Two', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Three', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['One', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Two', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Three', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Two', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Three', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['One', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Two', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Three', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['One', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Two', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Three', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Two', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Three', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['One', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Two', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Three', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['One', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Two', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Three', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Two', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Three', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['One', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Two', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Three', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['One', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Two', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Three', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Two', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
    ['Three', 1, 2, 3, 4, 5, 2034, 460549, 24, 62, 432, 2345],
  ];
  const sortable = [
    false,
    true,
    false,
    true,
    true,
    true,
    true,
    true,
    true,
    false,
  ];

  return (
    <Page title="Playground">
      <Card>
        <div
          style={{
            padding: '9px 1rem',
            display: 'flex',
            justifyContent: 'space-between',
            borderBottom: '1px solid rgb(225, 227, 229)',
          }}
        >
          <Button>Do something</Button>
          <Button>Anything</Button>
        </div>
        <DataTable
          columnContentTypes={columnTypes}
          headings={headings}
          rows={rows}
          // increasedTableDensity
          sortable={sortable}
        />
      </Card>
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/polaris-react/src/components/DataTable/DataTable.scss
+++ b/polaris-react/src/components/DataTable/DataTable.scss
@@ -292,13 +292,15 @@ $breakpoint: 768px;
   left: 0;
   right: 0;
   visibility: hidden;
-  z-index: var(--pc-index-table-loading-panel);
+  z-index: var(--p-z-1);
 }
 
 .StickyTableHeader {
   position: absolute;
   display: flex;
+  flex-direction: column;
   width: 100%;
+  overflow: hidden;
 
   &:not(.StickyTableHeader-isSticky) {
     top: -1000px;
@@ -306,20 +308,27 @@ $breakpoint: 768px;
   }
 }
 
+.StickyHeaderCell {
+  flex-shrink: 0;
+}
+
 .StickyTableColumnHeader-isScrolling {
   box-shadow: 1px 1px 0 0 var(--p-border-divider),
     1px 0 1px 1px rgba(63, 63, 68, 0.05), 1px 0 3px 0 rgba(63, 63, 68, 0.15);
 }
 
-.StickyTableHeadings {
-  overflow: hidden;
+.StickyTableHeadingsRow {
   flex: 1 1 auto;
   display: flex;
+  overflow-x: auto;
   background-color: var(--p-surface);
   justify-content: space-between;
+  scrollbar-width: none;
 
-  .Cell {
-    flex-grow: 1;
+  &::-webkit-scrollbar {
+    appearance: none;
+    height: 0;
+    width: 0;
   }
 }
 
--- a/polaris-react/src/components/DataTable/DataTable.scss
+++ b/polaris-react/src/components/DataTable/DataTable.scss
@@ -185,6 +185,14 @@ $breakpoint: 768px;
   padding: var(--p-space-2);
   margin: var(--p-space-2);
 
+  [data-sticky-active] & {
+    visibility: hidden;
+  }
+
+  [data-sticky-active] .StickyTable & {
+    visibility: visible;
+  }
+
   .IncreasedTableDensity & {
     font-weight: var(--p-font-weight-medium);
     font-size: var(--p-font-size-1);
--- a/polaris-react/src/components/DataTable/DataTable.scss
+++ b/polaris-react/src/components/DataTable/DataTable.scss
@@ -303,8 +303,8 @@ $breakpoint: 768px;
   overflow: hidden;
 
   &:not(.StickyTableHeader-isSticky) {
-    top: -1000px;
-    left: -1000px;
+    top: -9999px;
+    left: -9999px;
   }
 }
 
--- a/polaris-react/src/components/DataTable/DataTable.scss
+++ b/polaris-react/src/components/DataTable/DataTable.scss
@@ -64,7 +64,7 @@ $breakpoint: 768px;
   }
 }
 
-// stylelint-disable selector-max-class, selector-max-combinators, selector-max-specificity
+// stylelint-disable selector-max-class, selector-max-combinators, selector-max-specificity, selector-max-compound-selectors
 .TableRow + .TableRow {
   .Cell {
     border-top: var(--p-border-divider);
@@ -185,11 +185,11 @@ $breakpoint: 768px;
   padding: var(--p-space-2);
   margin: var(--p-space-2);
 
-  [data-sticky-active] & {
+  .StickyHeaderEnabled [data-sticky-active] & {
     visibility: hidden;
   }
 
-  [data-sticky-active] .StickyTable & {
+  .StickyHeaderEnabled [data-sticky-active] .StickyTable & {
     visibility: visible;
   }
 
@@ -286,54 +286,56 @@ $breakpoint: 768px;
   }
 }
 
-.StickyTable {
-  position: relative;
-  top: 0;
-  left: 0;
-  right: 0;
-  visibility: hidden;
-  z-index: var(--p-z-1);
-}
-
-.StickyTableHeader {
-  position: absolute;
-  display: flex;
-  flex-direction: column;
-  width: 100%;
-  overflow: hidden;
-
-  &:not(.StickyTableHeader-isSticky) {
-    top: -9999px;
-    left: -9999px;
+.StickyHeaderEnabled {
+  .StickyTable {
+    position: relative;
+    top: 0;
+    left: 0;
+    right: 0;
+    visibility: hidden;
+    z-index: var(--p-z-1);
   }
-}
 
-.StickyHeaderCell {
-  flex-shrink: 0;
-}
+  .StickyTableHeader {
+    position: absolute;
+    display: flex;
+    flex-direction: column;
+    width: 100%;
+    overflow: hidden;
 
-.StickyTableColumnHeader-isScrolling {
-  box-shadow: 1px 1px 0 0 var(--p-border-divider),
-    1px 0 1px 1px rgba(63, 63, 68, 0.05), 1px 0 3px 0 rgba(63, 63, 68, 0.15);
-}
-
-.StickyTableHeadingsRow {
-  flex: 1 1 auto;
-  display: flex;
-  overflow-x: auto;
-  background-color: var(--p-surface);
-  justify-content: space-between;
-  scrollbar-width: none;
-
-  &::-webkit-scrollbar {
-    appearance: none;
-    height: 0;
-    width: 0;
+    &:not(.StickyTableHeader-isSticky) {
+      top: -9999px;
+      left: -9999px;
+    }
   }
-}
 
-.StickyTableHeader-isSticky {
-  visibility: visible;
-  background-color: var(--p-surface);
-  box-shadow: var(--p-shadow-base);
+  .StickyHeaderCell {
+    flex-shrink: 0;
+  }
+
+  .StickyTableColumnHeader-isScrolling {
+    box-shadow: 1px 1px 0 0 var(--p-border-divider),
+      1px 0 1px 1px rgba(63, 63, 68, 0.05), 1px 0 3px 0 rgba(63, 63, 68, 0.15);
+  }
+
+  .StickyTableHeadingsRow {
+    flex: 1 1 auto;
+    display: flex;
+    overflow-x: auto;
+    background-color: var(--p-surface);
+    justify-content: space-between;
+    scrollbar-width: none;
+
+    &::-webkit-scrollbar {
+      appearance: none;
+      height: 0;
+      width: 0;
+    }
+  }
+
+  .StickyTableHeader-isSticky {
+    visibility: visible;
+    background-color: var(--p-surface);
+    box-shadow: var(--p-shadow-base);
+  }
 }
--- a/polaris-react/src/components/DataTable/DataTable.scss
+++ b/polaris-react/src/components/DataTable/DataTable.scss
@@ -175,6 +175,7 @@ $breakpoint: 768px;
   @include unstyled-button;
   @include focus-ring;
   position: relative;
+  background-color: lightblue;
   display: inline-flex;
   justify-content: flex-end;
   align-items: baseline;
@@ -191,6 +192,11 @@ $breakpoint: 768px;
     padding: 0;
     padding-right: var(--p-space-1);
     margin: 0;
+  }
+
+  .StickyTable & {
+    background-color: lightgoldenrodyellow;
+    position: relative;
   }
 
   &:hover {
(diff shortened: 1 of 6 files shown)


nickpresta:
Maybe this is "fine"/expected but some of the header duplicates/overlaps while scrolling:

mirualves (approved):

philschoefer:
Looking great! Some things I noticed:

1. Is the UX when the scroll navigation is visible what we want @mirualves? https://screenshot.click/20-14-c864j-kyu8u.mp4
2. There seems to be a difference in widths between the actual table header and the fixed one. https://screenshot.click/20-16-rixwy-y202h.mp4

Still looking through the code more closely, just some things I noticed upfront. I put the example code that I used in the PR description under `Copy-paste this code in playground/Playground.tsx:`

philschoefer:
One more, horizontal scrolling seems a bit sluggish: https://screenshot.click/20-28-q18xy-c89c4.mp4

philschoefer:
I wonder if we can use something like [Material UI](https://vwlkq2.csb.app/). It seems like they set each `th` to `position: sticky` avoiding the need of a separate header altogether.

polaris-react/src/components/DataTable/DataTable.scss
```diff
@@ -259,3 +267,55 @@ $breakpoint: 768px;
...
+.StickyTableHeader {
+  position: absolute;
+  display: flex;
+  flex-direction: column;
+  width: 100%;
+  overflow: hidden;
+
+  &:not(.StickyTableHeader-isSticky) {
+    top: -1000px;
+    left: -1000px;
```
nickpresta: Is this meant to "hide" this content off-screen? If so, I think we need to either use a larger value (e.g. `-9999px`) or actually `display: none` the thing, etc.

polaris-react/src/components/DataTable/DataTable.scss
```diff
@@ -259,3 +267,55 @@ $breakpoint: 768px;
...
+  }
+}
+
+.StickyHeaderCell {
+  flex-shrink: 0;
+}
+
+.StickyTableColumnHeader-isScrolling {
+  box-shadow: 1px 1px 0 0 var(--p-border-divider),
```
nickpresta: Should we be using something like `var(--p-border-width-1)` and/or rems of some sort so that the shadow scales with the border/border-radius/etc?

polaris-react/src/components/DataTable/DataTable.scss
```diff
@@ -259,3 +267,55 @@ $breakpoint: 768px;
...
+
+.StickyTableHeadingsRow {
+  flex: 1 1 auto;
+  display: flex;
+  overflow-x: auto;
+  background-color: var(--p-surface);
+  justify-content: space-between;
+  scrollbar-width: none;
+
+  &::-webkit-scrollbar {
```
nickpresta: Should we do anything else for other browsers? https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar

polaris-react/src/components/DataTable/DataTable.tsx
```diff
@@ -219,6 +320,59 @@ class DataTableInner extends PureComponent<CombinedProps, DataTableState> {
...
+      button = stickyHeadings[tableFocusedItemIndex].querySelector('button');
+    }
+
+    if (button == null) {
+      return null;
+    }
+
+    button.style.visibility = 'visible';
+    button.focus();
+    button.removeAttribute('style');
```
nickpresta: This is a bit heavy handed -- should we just be toggling the visibility? Or do we truly want to remove any potential styles?

polaris-react/src/components/Sticky/Sticky.tsx
```diff
@@ -15,6 +15,8 @@ export type StickyProps = {
   offset?: boolean;
   /** Should the element remain in a fixed position when the layout is stacked (smaller screens)  */
   disableWhenStacked?: boolean;
+  /** Callback run when sticky header is added or removed  */
+  onStickChange?: (isSticky: boolean) => void;
```
nickpresta: nitpicky: `onStickyChange` or `onChange` reads a little bit better (to me)

nickpresta (commented):
Here's the Storybook for the Data Table: https://5d559397bae39100201eedc1-zbmmfvmvcf.chromatic.com/iframe.html?id=all-components-data-table--data-table-with-all-of-its-elements&viewMode=story

It looks good to me -- I see the correct `data-sort` attributes are being added to the additional divs that get added, trying at pixel-perfect "breakpoints" does the right thing, performance looks good.

👍 

I made a few comments/questions but I don't see anything outright blocking at first glance.

polaris-react/src/components/DataTable/DataTable.tsx
blboyle: The goal here is to remove the js-set visibility and to have it move back up the cascade to have the visibility set by the CSS. I didn't see how to remove the specific `visibility` value set by the js above so I removed the attribute from the html element. I might be missing something though.

polaris-react/src/components/DataTable/DataTable.scss
blboyle: https://github.com/Shopify/polaris/blob/main/polaris-react/src/components/IndexTable/IndexTable.scss#L345

Yep, this is one (of many) CSS decisions I borrowed from the IndexTable implementation. Will update.

blboyle:
> One more, horizontal scrolling seems a bit sluggish: https://screenshot.click/20-28-q18xy-c89c4.mp4

@philschoefer I updated the description showcasing how the experience will look on a production build (thanks @nickpresta) using a storybook build. It looks much better.

philschoefer:
> I wonder if we can use something like Material UI. It seems like they set each th to position: sticky avoiding the need of a separate header altogether.

I see our use case is different since we want page level scrolling. 

I think overall it works really well. To me the difference in the widths of the cells in the sticky header is the only thing that would need fixing befor merge to avoid the little jump.

polaris-react/src/components/DataTable/DataTable.scss
blboyle: I tophatted on all browsers and the only one that didn't work with the webkit was firefox. So, I added `scrollbar-width` for that one.

polaris-react/src/components/DataTable/DataTable.tsx
```diff
@@ -219,6 +321,59 @@ class DataTableInner extends PureComponent<CombinedProps, DataTableState> {
...
+  }) => {
+    if (ref == null) {
+      return;
+    }
+
+    if (inStickyHeader) {
+      this.stickyHeadings[index] = ref;
+    } else {
+      this.tableHeadings[index] = ref;
+      this.tableHeadingWidths[index] = ref.offsetWidth;
```
philschoefer: Use `ref.getBoundingClientRect().width;`. It will give you a `float` vs `number` and prevents the slight jump that currently occurs when it switches from the actual to the sticky header.

polaris-react/src/components/DataTable/DataTable.tsx
```diff
@@ -266,15 +421,37 @@ class DataTableInner extends PureComponent<CombinedProps, DataTableState> {
...
+
+    if (stickyTableHeadingsRow == null || scrollContainer == null) {
+      return;
+    }
+
+    const {scrollLeft: stickyScrollLeft} = stickyTableHeadingsRow;
+    const {scrollLeft: scrollContainerScrollLeft} = scrollContainer;
+
+    if (targetIsStickyHeader) {
+      scrollContainer.scrollLeft = stickyScrollLeft;
```
philschoefer: Manipulating `scrollLeft` directly will trigger reflow. Replacing this with 
```
stickyTableHeadingsRow.style.transform = `translate-x(${stickyScrollLeft})`
``` 
will avoid this and makes the animation super smooth.

polaris-react/src/components/DataTable/DataTable.tsx
```diff
@@ -266,15 +421,37 @@ class DataTableInner extends PureComponent<CombinedProps, DataTableState> {
...
+    } else {
+      stickyTableHeadingsRow.scrollLeft = scrollContainerScrollLeft;
+    }
+  };
+
+  private scrollListener = (event: Event) => {
     this.setState((prevState) => ({
       ...this.calculateColumnVisibilityData(prevState.condensed),
```
philschoefer: This will trigger a re-render of the component including all of the calculations within `calculateColumnVisibilityData` which is partly responsible for the sticky header lagging behind a bit on horizontal scroll.

I would recommend to debounce this specific `setState` call. It's mainly used for the scroll indicators which is fine if only updated every 500ms or so. This way the heavy calculations don't drop the framerate and the sticky header never lags behind.

```tsx
  private scrollListenerStickyHeader = (event: Event) => {
    this.stickyHeaderScrolling(event);
  };

  // eslint-disable-next-line @typescript-eslint/member-ordering <--- somehow eslint doesn't like the debounce, hence that line.
  private scrollListener = debounce(() => {
    }));
  });
```

```tsx
<EventListener
    capture
    event="scroll"
    handler={this.scrollListener}
/>
<EventListener
    capture
    event="scroll"
    handler={this.scrollListenerStickyHeader}
/>
```

philschoefer (commented):
Summarizing my feedback since it's been a bit scattered:

1. I still see significant lag on horizontal scrolling on Chrome/Desktop. I left two comments that point to why FPS drop and how the performance can be optimized.
2. We have pages with multiple `DataTable` components underneath each other (or at least further content below the table). This means we need to detect when the scroll container leaves the viewport: https://screenshot.click/20-22-v9ki7-xx2wr.mp4

Other than that 👨‍🍳 👌

philschoefer (approved):

philschoefer:
Let's do a test where we check that this won't overlap with Shopify Admin's existing sticky search top bar.

blboyle:
@philschoefer @mirualves 

Inside details page playground in storybook

https://screenshot.click/21-46-fzc0l-45152.mp4

blboyle:
@mirualves @philschoefer @nickpresta 
In web. https://screenshot.click/21-34-vr27l-ilxg8.mp4

If you have a moment and can take another look I think we're close.

mirualves:
Ship it!!

chloerice:
(A rebase should fix the CI check that's failing)

chloerice (approved):
🚀

PR #5507, opened 2022-04-12
Move Babel config to root level for better monorepo support

### LINKED ISSUE #5441: Move babel config to package.json "babel" key

sam-b-rose:

Fixes #5441 

Test and verify using `yarn run build` to build all packages (including `polaris-react` which uses Babel).

* Move `./polaris-react/babel.config.js` to the root level`./babel.config.js`
* Add `overrides` array as a starting point for consolidated babel overrides on a per-project basis

Using the root `babel.config.js` and the [`overrides` key](https://babeljs.io/docs/en/options#overrides) is the [recommended approach](https://babeljs.io/docs/en/config-files#monorepos) for using Babel with monorepos. This is how [Babel itself configures](https://github.com/babel/babel/blob/master/babel.config.js#L113) subdirectories (packages) in the Babel monorepo.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,0 +1,9 @@
+module.exports = {
+  presets: [['@shopify/babel-preset', {typescript: true}]],
+  overrides: [
+    {
+      test: './polaris-react',
+      presets: [['@shopify/babel-preset', {typescript: true, react: true}]],
+    },
+  ],
+};
--- a/polaris-react/babel.config.js
+++ b/polaris-react/babel.config.js
@@ -1,3 +1,0 @@
-module.exports = {
-  presets: [['@shopify/babel-preset', {typescript: true, react: true}]],
-};


babel.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  presets: [['@shopify/babel-preset', {typescript: true}]],
+  overrides: [
+    {
+      test: './polaris-react',
+      presets: [['@shopify/babel-preset', {typescript: true, react: true}]],
+    },
+  ],
```
aaronccasanova, line 8: This approach looks good to me!

One "unfortunate" caveat is we have to run `yarn build` from the monorepo root for local development. From your [linked docs](https://babeljs.io/docs/en/config-files#root-babelconfigjson-file):

>The first issue you'll likely run into is that by default, Babel expects to load [babel.config.json](https://babeljs.io/docs/en/config-files#project-wide-configuration) files from the directory set as its ["root"](https://babeljs.io/docs/en/options#root), which means that if you create a [babel.config.json](https://babeljs.io/docs/en/config-files#project-wide-configuration), but run Babel inside an individual package the ["root"](https://babeljs.io/docs/en/options#root) Babel is using in that context is not your monorepo root, and it won't be able to find the [babel.config.json](https://babeljs.io/docs/en/config-files#project-wide-configuration) file.

I often change into package directories for my development workflow and without a local `<package>/.babelrc.json` with `rootMode: 'upward'` the build won't work. The docs provide a few examples of various ways you can specify `rootMode: 'upward'`, not sure how that works with loom, but with all that said I wouldn't considered this edge case a show stopper!

aaronccasanova (approved):

babel.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  presets: [['@shopify/babel-preset', {typescript: true}]],
```
BPScott, line 2: A lack of `react: true` seems suprising here - I'd have thought we'd need to parse jsx so storybook would understand our components

Though I note that the path in https://github.com/Shopify/polaris/blob/main/polaris-react/.storybook/babel.config.js#L1 is now incorrect and should be updated.

You should check that storybook continues to work (EDIT: Looks like CI is failing for exactly this reason)

babel.config.js
BPScott: > One "unfortunate" caveat is we have to run yarn build from the monorepo root for local development.

Running the loom build from the monorepo root - which would run the build for all loom projects - of which there is just the one -  is very easy and is a minimal code change. That said https://github.com/Shopify/polaris/blob/main/polaris-react/loom.config.ts does not override any babel config and by default [loom-plugin-build-library disables looking at babel.config.js for babel config](https://github.com/Shopify/loom/blob/main/packages/loom-plugin-build-library/src/plugin-build-library.ts#L47-L48), so I don't think the placement of this file has any effect for the polaris-react build and test commands.


Because of that I don't understand why the overrides section is needed here - Loom shouldn't be reading this file given its current configuration state.

sam-b-rose:
Given that Storybook isn't playing nicely with resolving the proper babel configuration and some node_module dep issues... going to close this out. Spent more time on trying to clean up this config that is now outweighing the benefit.

BPScott:
> Spent more time on trying to clean up this config that is now outweighing the benefit.

Hazarding a guess: All the babel dependencies are at the polaris-react layer (as they're pulled in via `loom-plugin-build-library`), so when you try to reference them at the root level it all gets confused. Sounds like moving the loom dependencies up to the root package.json could help.

sam-b-rose:
> Sounds like moving the loom dependencies up to the root package.json could help.

Great points, @BPScott. Will follow up on this if we decide to move to loom up to the root 👍

PR #5518, opened 2022-04-12
polaris-tokens in the monorepo + new token schema

romellogoodman:

Fixes #0000 

We are starting to pull the tokens into more surfaces and thinking about adding more information to the package. The following is an ideation/first pass at pulling the tokens into the monorepo as a separate package with meta data.

Changes
* Creation of @shopify/tokens package
* New schema that allows for the value of the token and its information to be colocated
* Defaulting of colors.light to colors to reduce the amount of boilerplate code needed to access a list of color variables

Next Steps
* Write and export utility functions for working with tokens such as creating the `--p` variable name
* Hook up to the token api since the website won't need to have the package published to consume and use the values
* Publishing to @shopify/tokens package from the monorepo
* Switch over polaris-react to consume @shopify/tokens


README.md
```diff
@@ -10,6 +10,7 @@ polaris/
 ├── polaris-for-vscode          # VSCode plugin for Polaris
 ├── polaris-react               # Components for @shopify/polaris package
 ├── polaris-shopify-com         # Documentation website
+├── polaris-tokens              # Tokens for Polaris
```
alex-page:
```suggestion
├── polaris-tokens              # Design tokens for Polaris
```

polaris-tokens/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@shopify/polaris-tokens",
+  "version": "1.0.0",
```
alex-page:
```suggestion
  "version": "5.0.0.rc.1",
```

polaris-tokens/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "description": "",
+  "main": "out/index.js",
+  "scripts": {
+    "build": "tsc -b",
```
alex-page, line 7: Not a huge fan of needing to run a build in this directory. @aaronccasanova explored an approach in adding types without a build step for `stylelint-polaris`. I think based on the simplicity here it might be a good approach?

polaris-tokens/schemas/token-group.json
```diff
@@ -0,0 +1,10 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Token group schema",
+  "type": "object",
+  "properties": {
+    "value": {"type": "string"},
+    "description": {"type": "string"},
+    "figmaName": {"type": "string"}
```
alex-page: In my eyes the JSON file should be the simplest format. It then should be ran through some kind of transform for different consumers. I would not expect a `shadow-transparent` to also have to enter in a Figma name of `Shadow transparent`. Could we remove this and generate the name as an output for Figma?

polaris-tokens/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "test": "npm run validate-tokens",
```
aaronccasanova:
```suggestion
    "test": "yarn validate-tokens",
```

polaris-tokens/package.json
BPScott: I'd be `5.0.0-rc.1` for proper semver pre-release tags

polaris-tokens/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "validate-tokens": "ajv validate -s schemas/token-group.json -d 'token-groups/*.json'"
+  },
+  "devDependencies": {
+    "ajv-cli": "^5.0.0"
+  },
+  "license": "SEE LICENSE IN LICENSE.md",
```
BPScott, line 14: There is no LICENCE.md so this claim makes no sense

polaris-tokens/tokens.ts
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Values to convert to CSS custom properties.
+ * @example {background: '#000'} // --p-background: #000;
+ */
+export interface TokenValues {
+  description: string;
+  figmaName: string;
+  value: string;
+}
```
aaronccasanova: This looks great! Similar to Alex's comment we can likely remove the `figmaName` for now and autogenerate in a build script. (See open-props' [toFigmaTokens transform](https://github.com/argyleink/open-props/blob/a429f931076ab69d3af4539ea312720280cfbdaf/build/props.js#L85-L88))

Also, I do suggest we align our naming with the Design Tokens Glossary as often as possible. In this case I propose updating this type and references to `TokenProperties`: https://www.designtokens.org/glossary/#design-token-properties

polaris-tokens/token-groups/color.json
```diff
@@ -0,0 +1,642 @@
+{
```
aaronccasanova: For now, I suggest keeping this `color.light.json` so it's clear these are the `light` color scheme tokens in consuming code. While we don't officially support dark mode we do leverage dark color schemes throughout Polaris and Web. Until we establish next steps for light, dark, and beyond color schemes I think our naming should be very explicit so it's clear how tokens are being used and where to improve.

polaris-tokens/package.json
aaronccasanova: I have no preference here (: In fact I would have used TypeScript for `stylelint-polaris` if I had more time to resolve ESM/CommonJS issues with the `stylelint` package and deferred to JSDoc to have some type safety. Note: We will still need to run `tsc` in a build step to generate declaration files prepublish, but otherwise TS vs. JSDoc is a matter of preference in this case.

polaris-tokens/schemas/token-group.json
aaronccasanova: Agreed, take a peek at open-props' [toFigmaTokens transform](https://github.com/argyleink/open-props/blob/main/build/to-figmatokens.js)

aaronccasanova (commented):
Thanks for kicking this off @romellogoodman! As we've continued to add packages to the monorepo we're starting to feel the pain points of not having a tokens package, so this is great timing!!


WIP is looking great 🙌  Left some minor feedback and suggestions

polaris-tokens/package.json
BPScott: I'd expect that this package should publish a commonjs and and es-modules version, so that it can be both usable within node, and can be easily treeshakable when consumed by bundlers; in addition to shipping types as Aaron mentioned.

Avoiding TS and build steps for stylelint-polaris was fine - it's a node-only package that only needed commonjs output, and it was desirable to write plain js so the in-repo lint run could use the package without needing the package to be built first ("you've got to have a passing build otherwise your linters will crash" would be a bad state of affairs). polaris-tokens is a package that needs to run in both node and browser environments and is expected to be consumed by bundlers too, it's requirements are a bit different to stylelint-polaris

polaris-tokens/tokens.ts
romellogoodman: Updated

polaris-tokens/tokens.ts
```diff
@@ -0,0 +1,74 @@
...
+  [C in ColorScheme]: TokenGroup;
+};
+
+const colorSchemes: ColorSchemes = {
+  light: lightColorScheme,
+  dark: darkColorScheme,
+};
+
+export interface Tokens {
+  color: TokenGroup;
```
aaronccasanova:
```suggestion
  colorSchemes: ColorSchemes;
```

We can challenge these conventions later but as of now the whole system is built off the `colorSchemes` key. Let's please not introduce breaking changes and focus on initializing the tokens package and getting dependents updated to use this lib.

polaris-tokens/tokens.ts
```diff
@@ -0,0 +1,74 @@
...
+  legacyTokens: TokenGroup;
+  motion: TokenGroup;
+  shape: TokenGroup;
+  spacing: TokenGroup;
+  typography: TokenGroup;
+  zIndex: TokenGroup;
+}
+
+export const tokens: Tokens = {
+  color: lightColorScheme,
```
aaronccasanova:
```suggestion
  colorSchemes,
```

Same as above

polaris-tokens/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+    "module": "commonjs",
+    "target": "es2020",
+    "lib": ["es2020"],
+    "outDir": "out",
```
aaronccasanova:
```suggestion
    "outDir": "dist",
```

This was briefly discussed in @samrose3's PR, to configure `tsc` to run from the root, but we should probably settle on an `outDir` naming convention for new libs.

With that said, @samrose3 is there any other `compilerOptions` that need to be set here? If so, we should add a doc about what options to set for libs that do and don't bundle with `tsc`.

polaris-tokens/token-groups/color.json
```diff
@@ -0,0 +1,514 @@
...
```
aaronccasanova: Please revert to `color.light.json`

aaronccasanova (changes requested):
Only blocker is the breaking change introduced to the tokens object. The whole system relies on `colorSchemes` right now and would require a major refactor to change that. The fact that we don't have dark mode support doesn't mean we don't have dark UIs. We still have dark UIs and shouldn't start removing elements of the architecture without clear guidance on a path forward imo.

I think the focus of this PR should be to initialize the tokens package so we can start updating all the dependents that have since been doing less than ideal work arounds to consume the tokens (e.g. `polaris-react`, `polaris-for-vscode`, `stylelint-polaris`).

romellogoodman:
@aaronccasanova sorry I didnt push latest and had a commit that reverted alot of the color work. I also incorporated the `dist` out folder and removed the `colors` key from the default export

aaronccasanova (approved):
Awesome work @romellogoodman 🎉 🚀

PR #5520, opened 2022-04-13
[Badge] exposes `Pip` to outside of `Badge` component

### LINKED ISSUE #5521: [Badge] Expose `Pip` component to outside of `Badge` component
## Feature request summary
In some use cases we need to indicate status in small layouts, like mobile screens (e.g: https://github.com/Shopify/online-store-web/issues/12232), while not relying solely on color.

 The `Pip` component inside `Badge` is a great candidate to cover these case but today it's locked inside `Badge`. It would be great if we could use `Pip` without `Badge`.

---

bernbecht:

Fixes https://github.com/Shopify/polaris/issues/5521

1) Makes `Pip` a sub-component of `Badge`
2) Exposes `Pip` to outside, to be used by other components inside/outside lib
3) Concentrates types and interfaces to `types.ts`
4) Adapts tests

## Media 📽
Screen Shot 2022-04-20 at 10 04 41

## How to 🎩
1)  Checkout the branch
````
git checkout berhell/expose-pip-outside-badge
````

<details>
<summary>2) Copy-paste this code in <code>playground/Playground.tsx</code>:</summary>

```jsx
import React from 'react';
import {GlobeMinor} from '@shopify/polaris-icons';

  return (
    <Page title="Pip">
      <Stack>
        <div>
          <Badge.Pip progress="incomplete" status="critical" /> Incomplete &
          Critical
        </div>

        <div>
          <Badge.Pip progress="complete" status="attention" /> Complete &
          Attention
        </div>

        <div>
          <Badge.Pip progress="partiallyComplete" status="info" /> Partial &
          Info
        </div>

        <div>
          <Badge.Pip progress="incomplete" status="new" /> Incomplete & New
        </div>

        <div>
          <Badge.Pip progress="complete" status="success" /> Complete & Success
        </div>

        <div>
          <Badge.Pip progress="partiallyComplete" status="warning" /> Partial &
          warning
        </div>
      </Stack>

      <Page title="Badges">
        <Stack>
          <Badge icon={GlobeMinor}>Icon, no status</Badge>
          {/*
          The Icon doesn’t get scaled down because the `Icon` size is hardcoded to 20px…
        */}
          <Badge icon={GlobeMinor} size="small">
            Small Badge with icon
          </Badge>

          <Badge icon={GlobeMinor} status="success">
            Icon, success
          </Badge>

          <Badge progress="complete" icon={GlobeMinor}>
            Progress + icon, prefers icon but TypeScript errors
          </Badge>

          <Badge progress="complete">Progress, unchanged</Badge>

          {/*
          Icon only. This works right now but I don‘t see a reason why `children` would be optional.
          If we want to allow this we need to also allow an alt prop for the icon.
        */}
          <Badge icon={GlobeMinor} />
        </Stack>
      </Page>
    </Page>
  );
}

```

</details>

### 🎩 checklist

### DIFF
--- a/polaris-react/UNRELEASED.md
+++ b/polaris-react/UNRELEASED.md
@@ -12,6 +12,8 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 
 - Added support for NodeJS v14 ([#5551](https://github.com/Shopify/polaris/pull/5551))
 
+- Made `Pip` a sub-component of `Badge` and exposed it to outside ([#5520](https://github.com/Shopify/polaris/pull/5520))
+
 ### Bug fixes
 
 ### Documentation
--- a/polaris-react/src/components/Badge/Badge.tsx
+++ b/polaris-react/src/components/Badge/Badge.tsx
@@ -1,18 +1,18 @@
 import React, {useContext} from 'react';
 
-import type {IconSource} from '../../types';
 import {classNames, variationName} from '../../utilities/css';
 import {useI18n} from '../../utilities/i18n';
 import {WithinFilterContext} from '../../utilities/within-filter-context';
 import {VisuallyHidden} from '../VisuallyHidden';
 import {Icon} from '../Icon';
+import type {IconSource} from '../../types';
 
 import styles from './Badge.scss';
+import type {Progress, Size, Status} from './types';
+import {Pip} from './components';
+import {getDefaultAccessibilityLabel} from './utils';
 
-type Status = 'info' | 'success' | 'attention' | 'warning' | 'critical' | 'new';
-type Progress = 'incomplete' | 'partiallyComplete' | 'complete';
-type Size = 'small' | 'medium';
-
+const DEFAULT_SIZE: Size = 'medium';
 interface NonMutuallyExclusiveProps {
   /** The content to display inside the badge. */
   children?: string;
@@ -37,23 +37,6 @@ export type BadgeProps = NonMutuallyExclusiveProps &
     | {icon?: IconSource; progress?: undefined}
   );
 
-const PROGRESS_LABELS: {[key in Progress]: Progress} = {
-  incomplete: 'incomplete',
-  partiallyComplete: 'partiallyComplete',
-  complete: 'complete',
-};
-
-const STATUS_LABELS: {[key in Status]: Status} = {
-  info: 'info',
-  success: 'success',
-  warning: 'warning',
-  critical: 'critical',
-  attention: 'attention',
-  new: 'new',
-};
-
-const DEFAULT_SIZE = 'medium';
-
 export function Badge({
   children,
   status,
@@ -68,68 +51,28 @@ export function Badge({
   const className = classNames(
     styles.Badge,
     status && styles[variationName('status', status)],
-    progress && styles[variationName('progress', progress)],
     icon && styles.icon,
     size && size !== DEFAULT_SIZE && styles[variationName('size', size)],
     withinFilter && styles.withinFilter,
   );
 
-  let progressLabel = '';
-  switch (progress) {
-    case PROGRESS_LABELS.incomplete:
-      progressLabel = i18n.translate(
-        'Polaris.Badge.PROGRESS_LABELS.incomplete',
-      );
-      break;
-    case PROGRESS_LABELS.partiallyComplete:
-      progressLabel = i18n.translate(
-        'Polaris.Badge.PROGRESS_LABELS.partiallyComplete',
-      );
-      break;
-    case PROGRESS_LABELS.complete:
-      progressLabel = i18n.translate('Polaris.Badge.PROGRESS_LABELS.complete');
-      break;
-  }
-
-  let statusLabel = '';
-  switch (status) {
-    case STATUS_LABELS.info:
-      statusLabel = i18n.translate('Polaris.Badge.STATUS_LABELS.info');
-      break;
-    case STATUS_LABELS.success:
-      statusLabel = i18n.translate('Polaris.Badge.STATUS_LABELS.success');
-      break;
-    case STATUS_LABELS.warning:
-      statusLabel = i18n.translate('Polaris.Badge.STATUS_LABELS.warning');
-      break;
-    case STATUS_LABELS.critical:
-      statusLabel = i18n.translate('Polaris.Badge.STATUS_LABELS.critical');
-      break;
-    case STATUS_LABELS.attention:
-      statusLabel = i18n.translate('Polaris.Badge.STATUS_LABELS.attention');
-      break;
-    case STATUS_LABELS.new:
-      statusLabel = i18n.translate('Polaris.Badge.STATUS_LABELS.new');
-      break;
-  }
-
   const accessibilityLabel = statusAndProgressLabelOverride
     ? statusAndProgressLabelOverride
-    : i18n.translate('Polaris.Badge.progressAndStatus', {
-        progressLabel,
-        statusLabel,
-      });
+    : getDefaultAccessibilityLabel(i18n, progress, status);
 
-  const hasAccessibilityLabel =
-    progressLabel || statusLabel || statusAndProgressLabelOverride;
-
-  let accessibilityMarkup = hasAccessibilityLabel && (
+  let accessibilityMarkup = Boolean(accessibilityLabel) && (
     <VisuallyHidden>{accessibilityLabel}</VisuallyHidden>
   );
 
-  if (progressLabel && !icon) {
+  if (progress && !icon) {
     accessibilityMarkup = (
-      <span className={styles.Pip}>{accessibilityMarkup}</span>
+      <span className={styles.PipContainer}>
+        <Pip
+          progress={progress}
+          status={status}
+          accessibilityLabelOverride={accessibilityLabel}
+        />
+      </span>
     );
   }
 
@@ -145,3 +88,5 @@ export function Badge({
     </span>
   );
 }
+
+Badge.Pip = Pip;
--- a/polaris-react/src/components/Badge/components/Pip/Pip.scss
+++ b/polaris-react/src/components/Badge/components/Pip/Pip.scss
@@ -1,0 +1,69 @@
+.Pip {
+  --pc-pip-size: var(--p-space-2);
+  --pc-pip-color: var(--p-icon);
+  display: inline-block;
+  color: var(--pc-pip-color);
+  height: var(--pc-pip-size);
+  width: var(--pc-pip-size);
+  border: var(--p-border-width-2) solid var(--pc-pip-color);
+  border-radius: var(--p-border-radius-half);
+  flex-shrink: 0;
+}
+
+.statusSuccess {
+  --pc-pip-color: var(--p-icon-success);
+}
+
+.statusInfo {
+  --pc-pip-color: var(--p-icon-highlight);
+}
+
+.statusAttention {
+  --pc-pip-color: var(--p-icon-attention);
+}
+
+.statusWarning {
+  --pc-pip-color: var(--p-icon-warning);
+}
+
+.statusCritical {
+  --pc-pip-color: var(--p-icon-critical);
+}
+
+.progressIncomplete {
+  background: transparent;
+}
+
+.progressPartiallyComplete {
+  background: linear-gradient(
+    to top,
+    currentColor,
+    currentColor 50%,
+    transparent 50%,
+    transparent
+  );
+
+  // Background colors may be stripped when printing, but box-shadow is not.
+  // We don't want to use box-shadow as the screen style as it gives a very
+  // slight halo effect
+  @media print {
+    background: none;
+    // 100px is an arbitrarily large number so that you can't see the curvature
+    // of the box shadow. y-offset is 2px larger than the spread to make it look
+    // like it is half-way down the pip (which is 4px tall)
+    box-shadow: 0 -102px 0 -100px currentColor inset;
+  }
+}
+
+.progressComplete {
+  background: currentColor;
+  // Background colors may be stripped when printing, but box-shadow is not.
+  // We don't want to use box-shadow as the screen style as it gives a very
+  // slight halo effect
+  @media print {
+    background: none;
+    // 100px is an arbitrarily large number so that you can't see the curvature
+    // of the box shadow.
+    box-shadow: 0 0 0 100px currentColor inset;
+  }
+}
--- a/polaris-react/src/components/Badge/components/Pip/Pip.tsx
+++ b/polaris-react/src/components/Badge/components/Pip/Pip.tsx
@@ -1,0 +1,38 @@
+import React from 'react';
+
+import {classNames, variationName} from '../../../../utilities/css';
+import type {Progress, Status} from '../../types';
+import {VisuallyHidden} from '../../../VisuallyHidden';
+import {useI18n} from '../../../../utilities/i18n';
+import {getDefaultAccessibilityLabel} from '../../utils';
+
+import styles from './Pip.scss';
+
+export interface PipProps {
+  status?: Status;
+  progress?: Progress;
+  accessibilityLabelOverride?: string;
+}
+
+export function Pip({
+  status = 'info',
+  progress = 'complete',
+  accessibilityLabelOverride,
+}: PipProps) {
+  const i18n = useI18n();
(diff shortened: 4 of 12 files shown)


polaris-react/src/components/Badge/types.ts
```diff
@@ -0,0 +1,24 @@
...
+  | 'info'
+  | 'success'
+  | 'attention'
+  | 'warning'
+  | 'critical'
+  | 'new';
+export type Progress = 'incomplete' | 'partiallyComplete' | 'complete';
+export type Size = 'small' | 'medium';
+
+export const PROGRESS_LABELS: {[key in Progress]: Progress} = {
```
BPScott: PROGRESS_LABELS and STATUS_LABELS are both javascript objects, not types. It seems weird to put them in a file called types

polaris-react/src/index.ts
```diff
@@ -52,8 +52,8 @@ export type {AvatarProps} from './components/Avatar';
 export {Backdrop} from './components/Backdrop';
 export type {BackdropProps} from './components/Backdrop';
 
-export {Badge} from './components/Badge';
-export type {BadgeProps} from './components/Badge';
+export {Badge, Pip} from './components/Badge';
```
BPScott: An item in the components folder exporting two components is a novelty. If you want Pip to be publicly accessible It'd be more conventional to create a `src/components/Pip` folder to place it in.

polaris-react/src/components/Badge/components/Pip.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+
+import {classNames, variationName} from '../../../utilities/css';
+import styles from '../Badge.scss';
+import type {Progress, Status} from '../types';
+
+export interface PipProps {
+  children?: '' | JSX.Element | undefined;
```
Tom-Bonnike: I think you should use
```suggestion
  children?: React.ReactNode;
```

Although, do we really want to allow `children`? The `Pip` is inline so consumers can just do:
```jsx
<p>
  <Pip />
  order #123
</p>
```

Just like in the badge, no?

---

However, since you didn’t port over any of the built-in accessibility markup stuff from the `Badge` (nor the override prop), and because `children` is optional, right now you risk having consumers render a `Pip` without any associated meaning.

IMO we should port all of the status + progress logic from the `Badge` into `Pip`.

bernbecht: Great review @Tom-Bonnike 🙏

In `Badge` we had before:
```` tsx

const accessibilityLabel = statusAndProgressLabelOverride
    ? statusAndProgressLabelOverride
    : i18n.translate('Polaris.Badge.progressAndStatus', {
        progressLabel,
        statusLabel,
      });

 ...

  let accessibilityMarkup = hasAccessibilityLabel && (
  );

  if (progressLabel && !icon) {
    accessibilityMarkup = (
      <span className={styles.Pip}>{accessibilityMarkup}</span>
    );
  }
````

Wouldn't `<span aria-label={accessibilityLabel} className={styles.Pip} />` be enough from an accessibility perspective?

If so,  we could simplify `Pip` interface to something like bellow, couldn't we?
````tsx
export interface PipProps {
  status?: Status;
  progress?: Progress;
  accessibilityLabel: string;
}
````

File: polaris-react/src/components/Badge/types.ts (hunk 1 above)
Comment (bernbecht) [reply polaris-react/src/components/Badge/types.ts]: I wonder why they weren't enums before 🤔

File: polaris-react/src/components/Badge/components/Pip.tsx (hunk 3 above)
Comment (Tom-Bonnike) [reply polaris-react/src/components/Badge/components/Pip.tsx]: I’m not from the Polaris team and only recently added the `icon` prop to the `Badge` hence why my name was suggested for review, so take this with a grain of salt 😛 

---

I was also talking about the labels, I got it wrong though we only render the Pip if there is the `progress` prop.
```jsx
  let progressLabel = '';
  switch (progress) {
    case PROGRESS_LABELS.incomplete:
      progressLabel = i18n.translate(
        'Polaris.Badge.PROGRESS_LABELS.incomplete',
      );
      break;
    case PROGRESS_LABELS.partiallyComplete:
      progressLabel = i18n.translate(
        'Polaris.Badge.PROGRESS_LABELS.partiallyComplete',
      );
      break;
    case PROGRESS_LABELS.complete:
      progressLabel = i18n.translate('Polaris.Badge.PROGRESS_LABELS.complete');
      break;
  }
  let statusLabel = '';
  switch (status) {
    case STATUS_LABELS.info:
      statusLabel = i18n.translate('Polaris.Badge.STATUS_LABELS.info');
      break;
    case STATUS_LABELS.success:
      statusLabel = i18n.translate('Polaris.Badge.STATUS_LABELS.success');
      break;
    case STATUS_LABELS.warning:
      statusLabel = i18n.translate('Polaris.Badge.STATUS_LABELS.warning');
      break;
    case STATUS_LABELS.critical:
      statusLabel = i18n.translate('Polaris.Badge.STATUS_LABELS.critical');
      break;
    case STATUS_LABELS.attention:
      statusLabel = i18n.translate('Polaris.Badge.STATUS_LABELS.attention');
      break;
    case STATUS_LABELS.new:
      statusLabel = i18n.translate('Polaris.Badge.STATUS_LABELS.new');
      break;
  }
  const accessibilityLabel = statusAndProgressLabelOverride
    ? statusAndProgressLabelOverride
    : i18n.translate('Polaris.Badge.progressAndStatus', {
        progressLabel,
        statusLabel,
      });
  const hasAccessibilityLabel =
    progressLabel || statusLabel || statusAndProgressLabelOverride;
  let accessibilityMarkup = hasAccessibilityLabel && (
  );

  if (progressLabel && !icon) {
    accessibilityMarkup = (
      <Pip progress={progress} status={status}>
        {accessibilityMarkup}
      </Pip>
    );
  }

  return (
    <span className={className}>
      {accessibilityMarkup}
      {icon && (
        <span className={styles.Icon}>
          <Icon source={icon} />
        </span>
      )}
      {children && <span>{children}</span>}
    </span>
  );
}
```

> If so, we could simplify Pip interface to something like bellow, couldn't we?

Yeah, but I think the label should be an override though. Honestly I’m a little unsure, we definitely want to generate the accessibility label in the `Pip` though, even though we’ll start using it outside of the `Badge` it still semantically is used to convey progress and status, and most of the time the “default” labels should be perfect.

emma-boardman: 👋 

> Wouldn't <span aria-label={accessibilityLabel} className={styles.Pip} /> be enough from an accessibility perspective?

Some assistive tech will ignore an `aria-label` added to an elements without semantic meaning. For this reason, we avoid adding `aria-label` to a span, which is why we pass the a11y content as a child

polaris-react/src/index.ts
bernbecht: Good point.

The reason behind exposing `Pip` is that `Online Store` needs to indicate status in small layouts. It's not a strong use case, but we'd like to have it polaris due to the following:
- code already exists inside `Badge`
- easier to maintain UI changes when polaris updates

However `Badge` and `Pip` are somehow tethered to each other, making `Pip` more sub-component of `Badge`. 

So we wouldn't like promote `Pip` as a standalone "component" or highlight it as a pattern to be used here and there. 

It could be more a `Badge.Pip` sub-component, available to those specific consumers that need it without `Badge`

emma-boardman (commented):
Looking great!

Extremely minor nit, but when I compare the storybook for this build vs main, the Badge is slightly smaller on [the main build](https://5d559397bae39100201eedc1-wdveceisya.chromatic.com/?path=/story/all-components-badge--badge-with-status-and-progress-label-override) than it is on [this PR](https://5d559397bae39100201eedc1-opbdndriof.chromatic.com/?path=/story/all-components-badge--badge-with-status-and-progress-label-override)

It's only off by 2px (!), which makes me think it could be a variable that got misplaced/added twice during the refactor

bernbecht:
> Extremely minor nit, but when I compare the storybook for this build vs main, the Badge is slightly smaller on [the main build](https://5d559397bae39100201eedc1-wdveceisya.chromatic.com/?path=/story/all-components-badge--badge-with-status-and-progress-label-override) than it is on [this PR](https://5d559397bae39100201eedc1-opbdndriof.chromatic.com/?path=/story/all-components-badge--badge-with-status-and-progress-label-override)

@emma-boardman oh, it was probably the 2px offset `badge` applies on `Pip`. Handling this right now and it'll up to the consumer to determine `Pip`'s layout. Then we avoid the component imposing any layout preference.

Thanks

polaris-react/UNRELEASED.md
```diff
@@ -12,6 +12,8 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 
 - Added support for NodeJS v14 ([#5551](https://github.com/Shopify/polaris/pull/5551))
 
+- Makes `Pip` a sub-component of `Badge` and exposed it to outside ([#5520](https://github.com/Shopify/polaris/pull/5520))
```
Tom-Bonnike:
```suggestion
- Made `Pip` a sub-component of `Badge` and exposed it to outside ([#5520](https://github.com/Shopify/polaris/pull/5520))
```

polaris-react/src/components/Badge/components/Pip/Pip.tsx
```diff
@@ -0,0 +1,31 @@
...
+import {classNames, variationName} from '../../../../utilities/css';
+import type {Progress, Status} from '../../types';
+import {VisuallyHidden} from '../../../VisuallyHidden';
+
+import styles from './Pip.scss';
+
+export interface PipProps {
+  status?: Status;
+  progress?: Progress;
+  accessibilityLabel: string;
```
Tom-Bonnike: I’m not convinced about the accessibility label being required when in most cases a good default can be generated from the `status` and `progress` props just like it’s done in the `Badge`.

Can’t we re-use the exact same logic from the `Badge` and introduce an accessibility label override prop instead?

polaris-react/src/components/Badge/index.ts
```diff
@@ -1 +1,3 @@
 export * from './Badge';
+export * from './components/Pip';
```
Tom-Bonnike: If you really want to keep it a sub-component of Badge, I think exposing it via `Badge.Pip = Pip;` is enough, we shouldn’t export it here:
```suggestion
```

polaris-react/src/components/Badge/Badge.tsx
```diff
@@ -1,18 +1,18 @@
...
 import styles from './Badge.scss';
+import type {Progress, Size, Status} from './types';
+import {Pip} from './components';
+import {getDefaultAccessibilityLabel} from './utils';
 
-type Status = 'info' | 'success' | 'attention' | 'warning' | 'critical' | 'new';
-type Progress = 'incomplete' | 'partiallyComplete' | 'complete';
-type Size = 'small' | 'medium';
-
+const DEFAULT_SIZE = 'medium';
```
MLDimitry: Could we type this as `DEFAULT_SIZE : Size`?

MLDimitry (commented):
Nice separation of concerns, leaving one minor comment for now

polaris-react/src/components/Badge/types.ts
```diff
@@ -0,0 +1,17 @@
+export enum StatusValue {
+  Info = 'info',
+  Success = 'success',
+  Warning = 'warning',
+  Critical = 'critical',
+  Attention = 'attention',
+  New = 'new',
+}
+// uses template literal types to get the string representation of the values of the FieldsMap enums
+export type Status = `${StatusValue}`;
```
Tom-Bonnike, line 10: TIL. So then you can pass for example `status='info'` without TypeScript screaming at you because you’re not passing `status={StatusValue.Info}`?

polaris-react/src/components/Badge/Badge.tsx
```diff
@@ -68,68 +51,31 @@ export function Badge({
...
     ? statusAndProgressLabelOverride
-    : i18n.translate('Polaris.Badge.progressAndStatus', {
-        progressLabel,
-        statusLabel,
-      });
+    : getDefaultAccessibilityLabel(i18n, progress, status);
 
   const hasAccessibilityLabel =
-    progressLabel || statusLabel || statusAndProgressLabelOverride;
+    accessibilityLabel || statusAndProgressLabelOverride;
```
Tom-Bonnike: we could simplify this down to:
```suggestion
    Boolean(accessibilityLabel)
```

nah?

polaris-react/src/components/Badge/components/Pip/Pip.tsx
```diff
@@ -0,0 +1,38 @@
...
+import {VisuallyHidden} from '../../../VisuallyHidden';
+import {useI18n} from '../../../../utilities/i18n';
+import {getDefaultAccessibilityLabel} from '../../utils';
+
+import styles from './Pip.scss';
+
+export interface PipProps {
+  status?: Status;
+  progress?: Progress;
+  accessibilityLabel?: string;
```
Tom-Bonnike: I think `statusAndProgressLabelOverride` was a superb name 🙈

Tom-Bonnike (commented):
Super cool; minor comments but ✅

polaris-react/src/components/Badge/components/Pip/Pip.tsx
bernbecht: 😂 
but what if the consumer wants to write anything else but status and label?

polaris-react/src/components/Badge/Badge.tsx
bernbecht: Now that you mentioned, we could actually simplify it down to:
````tsx

let accessibilityMarkup = Boolean(accessibilityLabel)l && (
  );
````

Comment (bernbecht) [reply polaris-react/src/components/Badge/Badge.tsx]: can't we?

File: polaris-react/src/components/Badge/types.ts (hunk 8 above)
Comment (bernbecht) [reply polaris-react/src/components/Badge/types.ts:10]: Yeah!

File: polaris-react/src/components/Badge/components/Pip/Pip.tsx (hunk 10 above)
Comment (Tom-Bonnike) [reply polaris-react/src/components/Badge/components/Pip/Pip.tsx]: Can we settle for `accessibilityLabelOverride`? :D

**Review by Tom-Bonnike — APPROVED** [review]
Looks good to me, although I think like I commented above that `accessibilityLabelOverride` might make it clearer that you can safely omit the `accessibilityLabel` 👍

File: polaris-react/src/components/Badge/components/Pip/Pip.tsx (hunk 10 above)
Comment (bernbecht) [reply polaris-react/src/components/Badge/components/Pip/Pip.tsx]: You've got a deal 🤝

**Review by emma-boardman — APPROVED** [review]
🥳 🚀

File: polaris-react/src/components/Badge/index.ts [hunk 11]
```diff
@@ -1 +1,2 @@
 export * from './Badge';
+export * from './types';
```
emma-boardman, line 2: Do we need to consume any of these types outside Polaris? If yes, they will also need to be exported from https://github.com/Shopify/polaris/blob/main/polaris-react/src/index.ts.

polaris-react/src/index.ts
```diff
@@ -52,7 +52,11 @@ export type {AvatarProps} from './components/Avatar';
 export {Backdrop} from './components/Backdrop';
 export type {BackdropProps} from './components/Backdrop';
 
-export {Badge} from './components/Badge';
+export {
+  Badge,
+  StatusValue as BadgeStatusValue,
+  ProgressValue as BadgeProgressValue,
+} from './components/Badge';
```
bernbecht, line 59: Exposing the enums to make consumer's life easier

ghost:
🎉 Thanks for your contribution to Polaris!

polaris-react/src/components/Badge/components/Pip/Pip.scss
```diff
@@ -0,0 +1,69 @@
...
+  border-radius: var(--p-border-radius-half);
+  flex-shrink: 0;
+}
+
+.statusSuccess {
+  --pc-pip-color: var(--p-icon-success);
+}
+
+.statusInfo {
+  --pc-pip-color: var(--p-icon-highlight);
```
chloerice, line 18: This is is the color for `statusNew`, `statusInfo` should've been `--p-icon` not `--p-icon-highlight`. Will cut a patch release with a fix.

PR #5526, opened 2022-04-13
Update "VSCode" to "VS Code"

sam-b-rose:
Replacing "VSCode" with "VS Code" to match what is used in the official documentation and marketing materials.

https://code.visualstudio.com/


yurm04 (approved):
Thank you for flagging and fixing this, Sam!

Issue #5363, opened 2022-03-28
[Layout foundations] Spacing coverage to 90%


aveline:

lgriffee:
Going to keep working on boosting this coverage as part of the layout foundations project.

lgriffee:
Going to keep working on boosting this coverage as part of the layout foundations project.

PR #5528, opened 2022-04-13
Boost spacing coverage

lgriffee:

Related to #5363. Try to clean up as many hardcoded `px` spacing values before starting in on the rest of our layout work (and get a better understanding of what remains for layout coverage). 

- Replacing any `padding` or `margin` values that are hardcoded 
- [shape.json] Add two new tokens `border-width-4` and  `border-width-5`
- Replacing any hardcoded `outline-width` with `border-width` 
- ~~Removing any `border-width` tokens inside `width`, `height`, offsets, `padding`, `margin`, ~~`box-shadow`~~, and `drop-shadow`~~
- [IndexTable] Changing the top and bottom `padding` on `.HeadingWrapper` (no visual change though)
- [_buttons.scss] Changing the right and left `padding` on `@mixin plain-button-backdrop` (✅ visual change reviewed)

### WHAT is this pull request NOT doing?
- [_layout.scss] Replacing `$dangerous-magic-space` `px` values since we'll be iterating on this file (also I think custom props might not be consumed properly in this case?) 
- Replacing `px` that relate to size (`width`, `height`, `top`, `bottom`, `left`, `right`, offsets, `box-shadow`, `drop-shadow`)
- [_focus-ring.scss] Replacing `px` values since we want to remove this file
- Replacing `px` values inside `focus-ring($border-width:` since they are sometimes used just as offset values
- [FileUpload] Changing the hardcoded left and right `padding` on `.Action` (✅ visual change reviewed)
  - [Following up with a designer to figure the best way to move forward here](https://github.com/Shopify/polaris/pull/5528/files#r853176988)
- [Dialog.scss] Replacing `$vertical-spacing` and `$dangerous-magic-space` `px` values since they are used in media queries
- [Button.scss] Removing any `border-width` tokens inside `box-shadow`


lgriffee:
Not going to add an UNRELEASED entry until there's more clarity on what will be included in this PR

polaris-react/src/components/DropZone/DropZone.scss
```diff
@@ -21,9 +21,7 @@ $dropzone-stacking-order: (
 }
 
 @mixin set-border-radius {
-  border-radius: calc(
-    var(--p-border-radius-1) + #{var(--p-border-width-1)} + 1px
```
lgriffee, line 25: Wasn't sure why `border-width` was being added to `border-radius` plus `1px` (seemed odd to mix radius and width values)? If it's important to see that math I can change it back and/or add a comment about why they are being added together.

polaris-react/src/components/DropZone/DropZone.scss
```diff
@@ -122,7 +121,7 @@ $dropzone-stacking-order: (
 }
 
 .Container {
-  @include focus-ring($border-width: var(--p-border-width-1));
```
lgriffee, line 125: These "`border-width`" values for focus ring are also offsets...and sometimes are only used an an offset so it seemed strange to use `border-width` tokens for them.

polaris-react/src/components/DropZone/components/FileUpload/FileUpload.scss
```diff
@@ -27,7 +27,7 @@
   flex: 0 0 auto;
   border: none;
   border-radius: var(--p-border-radius-base);
-  padding: var(--p-space-1) 6px;
```
lgriffee: Would love approval from design (@zecarlostorre) on changing this one (`2px` left and right `padding` was added—see bright green in Chromatic diffs) but seemed like a small enough change to get this to use our tokens and be on the grid (4px/8px) we want to use. 

Screen Shot 2022-04-18 at 2 21 48 PM
Screen Shot 2022-04-18 at 2 22 13 PM

polaris-react/src/components/IndexTable/IndexTable.scss
```diff
@@ -501,10 +501,10 @@ $scroll-bar-size: var(--p-space-2);
 .HeaderWrapper {
   position: relative;
   display: flex;
-  align-items: flex-start;
+  align-items: center;
   width: 100%;
   min-height: 56px;
-  padding: 10px var(--p-space-4);
```
lgriffee, line 507: `align-items` should be `center` here to match our `BulkActionsWrapper` implementation (when the typical button inside of the `HeaderWrapper` is clicked it gets replaced with a `BulkActionsWrapper`). Since the typical button inside of `HeaderWrapper` is only `36px` and there's a `min-height` of `56px` on `HeaderWrapper` it means we can change out the `padding` to a value on our scale less than `10px` and because of the new value of `align-items: center` it won't have any visual effect on the component. 

Also while testing all of this I realized that `BulkActions` doesn't show in our small screen example in Storybook...but it shows in production? 

https://user-images.githubusercontent.com/21976492/163875228-c9d6c224-2c1c-48a7-af4b-8fe97d190381.mp4

https://user-images.githubusercontent.com/21976492/163875207-79746a47-dd09-40f8-9f9d-41cdcd5f160f.mp4

The following code in `BulkActions.scss` is responsible for hiding it at a small screen size but I'm still not sure why there's a difference between Storybook and production...

Screen Shot 2022-04-18 at 2 39 52 PM

polaris-react/src/components/Badge/Badge.scss
```diff
@@ -15,7 +15,7 @@
   font-weight: var(--p-font-weight-regular);
 
   @media print {
-    border: solid 0.1px var(--p-border);
```
lgriffee, line 15: Tested changing this value from `0.1px` to `1px` (via out tokens) and didn't see a visual change when printing the page (checked in Chrome, Safari, and Firefox):
Screen Shot 2022-04-12 at 2 57 55 PM
Screen Shot 2022-04-12 at 2 58 29 PM

polaris-react/src/components/ActionList/ActionList.scss
```diff
@@ -61,7 +61,7 @@
   &:hover {
     background-color: var(--p-surface-hovered);
     text-decoration: none;
-    outline: var(--p-border-width-3) solid transparent;
```
lgriffee: Semantically it didn't feel right to me to have tokens labeled `border-with` inside `outline` properties (even though `border-width` and `outline-width` values both represent the thickness of a line with the same range of values). So I added `outline-width` tokens to be semantically correct and keep them separate from `border-width`. 

We could also create one token for both values (like `line-thickness-1`) but if we do that it could be semantically confusing or make it difficult to update the values separately of one anther. 

I also questioned if we should even have `border-width` and `outline-width` tokens...but they do help standardize values (I did come across one `0.1px` value) and the tokens would allow us to change the values easily in the future if we wanted to make our borders/outlines bigger across the entire admin.

polaris-react/src/components/Button/Button.scss
```diff
@@ -424,8 +424,8 @@
     position: relative;
     border-color: currentColor;
     // stylelint-disable selector-max-class
-    box-shadow: 0 0 0 var(--p-border-width-1) currentColor;
```
lgriffee, line 427: I didn't feel like `border-width` should be used for `spread-radius` for [`box-shadow`](https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow).

polaris-react/src/components/Button/Button.scss
```diff
@@ -424,8 +424,8 @@
...
-    @include focus-ring($border-width: var(--p-border-width-2));
```
lgriffee, line 428: These "`border-width`" values for focus ring are also offsets...and sometimes are only used as an offset so it seemed strange to use `border-width` tokens for them. See [this comment](https://github.com/Shopify/polaris/pull/5528/files#r852426548) for more details.

polaris-react/src/components/DropZone/DropZone.scss
```diff
@@ -55,12 +54,12 @@ $dropzone-stacking-order: (
   }
 
   &:hover {
-    outline: var(--p-border-width-1) solid transparent;
+    outline: var(--p-outline-width-1) solid transparent;
   }
 }
 
 .hasOutline {
-  padding: var(--p-border-width-1);
```
lgriffee, line 63: Since this is a `padding` value I changed this to be a spacing token.

polaris-react/src/components/DropZone/components/FileUpload/FileUpload.scss
lgriffee: Here are the Chromatic diffs (the bright green represents the space where the `padding` was removed):
Screen Shot 2022-04-18 at 2 21 48 PM
Screen Shot 2022-04-18 at 2 22 13 PM

polaris-react/src/components/Modal/components/Dialog/Dialog.scss
```diff
@@ -1,12 +1,14 @@
 @import '../../../../styles/common';
 
 $top-offset-slide-in-start: 200px;
+// Question
 $vertical-spacing: 60px;
```
lgriffee: I had originally thought of converting this Sass variable into a CSS custom property (such as `--pc-vertical-spacing`). However the value is being used inside a media query (see line 65 below) and wasn't sure if it's bad practice to have a `calc()` inside a `@media` and/or if the CSS custom property would even be consumed properly. I know that Sass variables compile at runtime and CSS custom properties compile when they're used...

polaris-react/src/components/Modal/components/Dialog/Dialog.scss
```diff
@@ -1,12 +1,14 @@
...
 $height-limit: 600px;
 $xsmall-width: 380px;
 $small-width: 620px;
 $large-width: 980px;
 // @TODO simplify media queries so this isn't needed
+// Question
 $dangerous-magic-space-16: 64px;
```
lgriffee: Had the same question as above for this value.

polaris-react/src/components/RangeSlider/components/DualThumb/DualThumb.scss
```diff
@@ -2,7 +2,6 @@
 @import '../../RangeSlider';
 
 $range-wrapper: 28px;
-$range-thumb-border-error: 2px solid var(--p-border-critical);
```
lgriffee, line 5: Nothing was using this variable so I deleted it.

polaris-react/src/components/ResourceItem/ResourceItem.scss
```diff
@@ -32,7 +32,7 @@ $resource-list-item-variables: (
 }
 
 .ResourceItem {
-  --pc-resource-item-offset: calc(48px - 12px + var(--p-space-1));
```
lgriffee, line 35: Even with the values listed out here I had a hard time knowing where they were coming from. I'd love to add a comment explaining what was added together for the calculation. Happy to revert this too if the math feels important.

polaris-react/src/components/SkeletonPage/SkeletonPage.scss
```diff
@@ -46,10 +46,12 @@ $skeleton-display-text-max-width: 120px;
 
 .Title {
   @include text-emphasis-strong;
+  // Question
   font-size: 24px;
   line-height: 28px;
 
   @include when-typography-not-condensed {
+    // Question
     font-size: 20px;
```
lgriffee: I noticed some values in `SkeletonPage` that referenced hard coded typography values. Is there a reason for this (maybe because of what styles it can access at that point in the render?). I also realize we might be deprecating `SkeletonPage` in favor of more specific skeleton states as part of the perceived performance project?

polaris-react/src/styles/foundation/_focus-ring.scss
```diff
@@ -1,7 +1,7 @@
 /// Sets the focus ring for an interactive element
 /// @param {String} $size - The size of the border radius on the focus ring.
 /// @param {String} $style - Focus ring state.
-/// @param {Number} $border-width - The border width of your element in rems.
```
lgriffee, line 4: Was doing some digging into these values with @aaronccasanova to see if we should replace these with tokens and discovered that `$border-width` values for `focus-ring()` are not just used to offset the `border-width` of an element—it's also just sometimes used as a negative offset and so the description here felt a bit misleading. The focus ring around the mobile nav icon is a great example of this—there's no border around the icon and yet it uses this value as an offset: 
```
.IconWrapper {
    @include focus-ring($border-width: 6px);
  }
```
Screen Shot 2022-04-18 at 3 29 07 PM

Example of increasing the `6px` value: 
Screen Shot 2022-04-14 at 4 00 32 PM

polaris-react/src/styles/shared/_buttons.scss
```diff
@@ -188,8 +188,8 @@
     }
 
     &.pressed {
+      border: var(--p-border-width-1) solid var(--p-border-critical);
       background: var(--p-surface-critical-subdued);
-      box-shadow: var(--p-border-width-1) solid var(--p-border-critical);
```
lgriffee, line 192: Based on the values here I think this is supposed to be a `border` property not `box-shadow` (you can't use `solid` in `box-shadow`).

polaris-react/src/styles/shared/_buttons.scss
```diff
@@ -206,8 +206,8 @@
 }
 
 @mixin plain-button-backdrop {
-  padding: var(--p-space-05) 5px;
-  margin: calc(-1 * var(--p-space-05)) -5px;
```
lgriffee, line 210: Would also like approval from design (@zecarlostorre) on changing this one but seemed like a small enough change (difference is `1px` `padding` on left and right of grey backdrop) to get this to use our tokens and be on the grid (4px/8px) we want to use. Personally have a very hard seeing this visual change just with my eye and can only see it if I use the guideline from Monosnap. 

**After**
button-plain-left-right-padding-new
**Before**
button-plain-left-right-padding-old

polaris-react/src/components/ActionList/ActionList.scss
alex-page: I think there should be one token for both border and outline. Currently we would need to document the difference between both and there really should just be one value. I don't mind using border in outlines.

polaris-react/src/components/DropZone/components/FileUpload/FileUpload.scss
zecarlostorre: @lgriffee Let's leave that one as an exception for now. This is one of those cases where 8 is too much, 4 is too little. 

I'll follow up with the designer that worked on it, to figure out the best way out of this.

polaris-react/src/styles/shared/_buttons.scss
zecarlostorre: Go ahead with this one @lgriffee. The change is barely noticeable

polaris-react/src/components/ActionList/ActionList.scss
aaronccasanova: Agree, one set of tokens for each property (border and outline). I think they both serve different purposes (semantically) in the UI and could be problematic to create a dependency between the two.

polaris-react/src/components/Button/Button.scss
aaronccasanova: Agreed, we should be leveraging a `shadows` token scale here. I believe we still need to establish that scale and a mechanism for defining the color. So perhaps this is a special case we have to leave for now 🤷  Curious what other folks think.

polaris-react/src/components/ActionList/ActionList.scss
alex-page: I wrote this poorly. We should have one token e.g. lineWidth or border for both border and outline. I want to keep our tokens as simple as possible and adding extra layers without much justification right now seems like it could create unwanted technical debt.

polaris-react/src/components/Modal/components/Dialog/Dialog.scss
aaronccasanova: Yeah, good assessment. If this is used to derive a media query we probably need to leave as is. Unless we introduce the [postcss-custom-media](https://github.com/postcss/postcss-custom-media) plugin we will have to continue using SCSS variables for media queries and add a TODOs to update them when the custom media spec has landed in all our supported browsers.

polaris-react/src/components/ResourceItem/ResourceItem.scss
aaronccasanova: I would be fine with a comment explaining the calculation or even extracting `48px` and `12px` into descriptive `--pc-` vars so the code is self documenting. I don't have enough context to approve deleting so I will trust your judgement here!

polaris-react/src/styles/foundation/_focus-ring.scss
aaronccasanova: Agreed, this comment is confusing. Laura and I played around with this mixin for a bit, in an effort to establish Polaris `border-width` tokens, until we realized the `focus-ring` `$border-width` param was simply a negative offset for the outline. I wish we could change the parameter name to `$outline-offset` but I believe that would be a major breaking change. I think updating the comment is the next best thing.

polaris-react/src/components/ResourceItem/ResourceItem.scss
lgriffee: Digging into [the file history](https://github.com/Shopify/polaris/commit/a275a1c61f63b6cd89b060e5c11fc11316c6fc00) I figured out what was going on with these values being added together and added a comment about it. Since the value is an offset I turned it back into `px` to be consistent. I was tempted to create variables for all the things being added together so the math never gets messed up but it seemed verbose to do so. Also to keep all the variables together I converted all the Sass ones I could into CSS custom properties.

lgriffee:
I hadn't realized until now that we are using spacing tokens for some instances of `width`, `height`, `top`, `left`, `right`, `bottom`, and offsets...Do we want this? I guess I had been thinking of spacing as being more related to `margin` and `padding` and `width`, `height`, `top`, `left`, `right`, `bottom` related to size. But perhaps if we're thinking of an underlying scale for size that would be used in spacing then this doesn't matter? Just want to make sure we're consistent with how we use our tokens. I can go back and check all `width`, `height`, `top`, `left`, `right`, `bottom`, and offsets again if this is the case.

polaris-react/src/components/ActionList/ActionList.scss
lgriffee: I've consolidated `border-width` and `outline-width` values into `line-width`. Let me know what you think. I can also change the name back to `border-width` if `line-width` feels weird naming wise.

alex-page:
Lets ship this as a non breaking change with `border-width` then do the rename in v10. @lgriffee @aaronccasanova

polaris-react/src/components/ActionList/ActionList.scss
aaronccasanova: These really feel like separate token scales for me, but I trust your judgement! Interesting that Theme UI doesn't seem to account for `outline` at all: https://theme-ui.com/theme-spec/

polaris-react/src/components/Button/Button.scss
```diff
@@ -424,8 +424,8 @@
...
+    box-shadow: 0 0 0 1px currentColor;
```
alex-page: For the sake of this PR can we use the border token for shadows and update it in the future?

polaris-react/src/components/Button/Button.scss
```diff
@@ -461,7 +461,7 @@
     }
 
     &:focus {
-      box-shadow: 0 0 0 var(--p-border-width-1) currentColor;
+      box-shadow: 0 0 0 1px currentColor;
```
alex-page: This one too

polaris-react/src/components/SkeletonTabs/SkeletonTabs.scss
```diff
@@ -18,7 +18,7 @@
     bottom: -1px;
     left: var(--p-space-3);
     right: var(--p-space-3);
-    height: var(--p-border-width-3);
+    height: 3px;
```
alex-page: I understand this is a height. However it is rendering a line under the tab. I don't mind this using the border token because we would actually want this to change if the border got updated.

polaris-react/src/components/Tabs/Tabs.scss
```diff
@@ -136,7 +136,7 @@ $item-vertical-padding: $item-min-height * 0.5;
     bottom: calc(-1 * var(--p-space-2));
     left: 0;
     right: 0;
-    height: var(--p-border-width-3);
+    height: 3px;
```
alex-page: Same as above this is technically a border created with height.

polaris-react/src/styles/shared/_buttons.scss
```diff
@@ -10,7 +10,7 @@
   );
 
   @include recolor-icon(var(--p-icon));
-  @include focus-ring($border-width: var(--p-border-width-1));
+  @include focus-ring($border-width: 1px);
```
alex-page: Again this should probably be consolidated for now. It's a 1px line around an element on focus. I understand shadows will replace this but I think for now we should use the token.

polaris-react/src/styles/shared/_interaction-state.scss
```diff
@@ -6,7 +6,7 @@
   left: calc(-1 * var(--p-space-2));
   height: 100%;
   display: block;
-  width: var(--p-border-width-3);
+  width: 3px;
```
alex-page: This is a very small border to the right of a selected item. Similarly to others I would expect this to be connected to border tokens and if the the 3px value changed I would expect this to be linked.

alex-page (commented):
Some minor feedback but looking super. Excited to ship this. 🎉

polaris-react/src/components/SkeletonPage/SkeletonPage.scss
lgriffee: Going to remove the comments here for merging...but still wonder about this.

alex-page (approved):

PR #5531, opened 2022-04-14
Enable CORS for the token API

sam-b-rose:

The tokens API can't be accessed from other origins (see [example CodeSandbox](https://codesandbox.io/s/jolly-paper-q8xuoj?file=/src/index.js)).

This change will allow the tokens API to to be used by [origins](https://developer.mozilla.org/en-US/docs/Glossary/Origin) (domain, scheme, or port) other than its own.

Followed the suggestion from the Vercel documentation on [enabling CORS for a Next.js application](https://vercel.com/support/articles/how-to-enable-cors#enabling-cors-in-a-next.js-app).

cors

### DIFF
--- a/polaris.shopify.com/next.config.js
+++ b/polaris.shopify.com/next.config.js
@@ -1,8 +1,30 @@
+/* eslint-disable require-await */
 /** @type {import('next').NextConfig} */
+
 const nextConfig = {
   reactStrictMode: true,
   eslint: {
     ignoreDuringBuilds: true,
+  },
+  async headers() {
+    return [
+      {
+        source: '/api/v0/tokens/:path*',
+        headers: [
+          {key: 'Access-Control-Allow-Credentials', value: 'true'},
+          {key: 'Access-Control-Allow-Origin', value: '*'},
+          {
+            key: 'Access-Control-Allow-Methods',
+            value: 'GET,OPTIONS',
+          },
+          {
+            key: 'Access-Control-Allow-Headers',
+            value:
+              'X-CSRF-Token, X-Requested-With, Accept, Accept-Version, Content-Length, Content-MD5, Content-Type, Date, X-Api-Version',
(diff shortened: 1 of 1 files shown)


romellogoodman (approved):

polaris.shopify.com/next.config.js
```diff
@@ -1,9 +1,31 @@
...
+
 const nextConfig = {
   reactStrictMode: true,
   eslint: {
     ignoreDuringBuilds: true,
   },
+  async headers() {
+    return [
+      {
+        source: '/api/:path*',
```
aaronccasanova: Perhaps, we should scope this down to `/api/v0/:tokens*` so future endpoints don't automatically enable CORS.

polaris.shopify.com/next.config.js
```diff
@@ -1,9 +1,31 @@
...
+        headers: [
+          {key: 'Access-Control-Allow-Credentials', value: 'true'},
+          {key: 'Access-Control-Allow-Origin', value: '*'},
+          {
+            key: 'Access-Control-Allow-Methods',
+            value: 'GET,OPTIONS,PATCH,DELETE,POST,PUT',
```
aaronccasanova:
```suggestion
            value: 'GET,OPTIONS',
```

Think we should follow the principle of least privilege here.

johanstromqvist (approved):

polaris.shopify.com/next.config.js
sam-b-rose: Works for me 👍  updated

polaris.shopify.com/next.config.js
sam-b-rose: Sounds good. Updated to `/api/v0/tokens/:path*`

aaronccasanova (approved):

PR #5539, opened 2022-04-18
[DataTable] Add fixed first column

JeremyLudwigDev:

The `DataTable` has no ability to add a fixed first column, a feature which can improve the readability of tables with many columns of data.

This PR adds the ability to add a fixed first column. The option is available via a new `hasFixedFirstColumn` prop. The prop defaults to `false`.

Closes: https://github.com/Shopify/core-issues/issues/36954

| `hasFixedFirstColumn = {false}` (Default) | `hasFIxedFirstColumn` |
| -- | -- |
| image | image | 

Resolves: https://github.com/Shopify/core-issues/issues/36866

## 

### How to 🎩

Be sure to check a few different scenarios:
- With and without various props set (increasedDensity, stickyHeader, truncate)
- Try adding more or fewer columns and/or rows of data
- Adding some long data in various parts of the table

<details>

```jsx

import {Card, DataTable, Page} from '../src';

  const [sortedRows, setSortedRows] = useState(null);

  const initiallySortedRows = [
    [
      'Emerald Silk Gown',
      'Formalwear',
      "Jill's formal",
      10,
      '$875.00',
      124689,
      140,
      '$426.00',
      '$122,500.00',
    ],
    [
      'Mauve Cashmere Scarf',
      'Accessories',
      "Jack's Accessories",
      253,
      '$230.00',
      124533,
      83,
      '$620.00',
      '$19,090.00',
    ],
    [
      'Ensembles',
      'Avocado Fashions',
      23,
      '$445.00',
      124518,
      32,
      '$353.00',
      '$14,240.00',
    ],
    [
      'Socks',
      'Essentials',
      'Avocado Fashions',
      465,
      '$4.00',
      124518,
      32,
      '$3.00',
      '$140.00',
    ],
  ];

  const handleSort = useCallback(
    (index, direction) => setSortedRows(sortCurrency(rows, index, direction)),
    [rows],
  );

  return (
      <Card>
        <DataTable
          columnContentTypes={[
            'text',
            'text',
            'text',
            'numeric',
            'numeric',
            'numeric',
            'numeric',
            'numeric',
            'numeric',
          ]}
          headings={[
            'Product',
            'Category',
            'Vendor',
            'Orders',
            'Price',
            'SKU Number',
            'Net quantity',
            'Shipping',
            'Net sales',
          ]}
          rows={rows}
          totals={['', '', '', '', '', '', 255, '$1399', '$155,830.00']}
          sortable={[true, true, true, true, true, true, true, true, true]}
          initialSortColumnIndex={4}
          onSort={handleSort}
          hasFixedFirstColumn
          truncate
        />
      </Card>
    </Page>
  );

  function sortCurrency(rows, index, direction) {
    return [...rows].sort((rowA, rowB) => {
      let type;
      if (rowA[index].length > 1) {
        type = !isNaN(parseFloat(rowA[index]))
          ? 'number'
          : !isNaN(parseFloat(rowA[index].substring(1)))
          ? 'currency'
          : 'string';
      } else {
        type = !isNaN(parseFloat(rowA[index])) ? 'number' : 'string';
      }

      const itemA =
        type === 'number'
          ? parseInt(rowA[index])
          : type === 'currency'
          ? parseFloat(rowA[index].substring(1))
          : rowA[index];
      const itemB =
        type === 'number'
          ? parseInt(rowB[index])
          : type === 'currency'
          ? parseFloat(rowB[index].substring(1))
          : rowB[index];

      if (type === 'string') {
        const result = itemA > itemB ? 1 : itemB > itemA ? -1 : 0;
        return direction === 'descending' ? result * -1 : result;
      }

      return direction === 'descending' ? itemB - itemA : itemA - itemB;
    });
  }
}


```

</details>

### 🎩 checklist


philschoefer (commented):
Talked IRL (Slack). Just adding comments here for visibility:

1. There seems to be a slight blip when the fixed column appears. Will probably have to switch this for the fixed table to always be visible as soon as the user is in the condensed screen.
2. Let's pair to refactor this after this component after our PRs are through. It usually doesn't take much time and since we're doing a lot of work here now it's likely time well spend.

philschoefer (approved):
LGTM pending the the issue where in some cases the browser scrollbar gets overlapped by the fixed column.

blboyle:
Not sure if this is what Phil was referring to, but when there is a focusable column header, it can be hidden behind the fixed column.

https://screenshot.click/20-58-5gwco-7mn4e.mp4

polaris-react/src/components/DataTable/DataTable.tsx
```diff
@@ -420,8 +454,14 @@ class DataTableInner extends PureComponent<CombinedProps, DataTableState> {
     );
 
     return (
-      <tr key={`row-${index}`} className={className}>
+      <tr
+        key={`row-${index}`}
+        className={className}
+        onMouseEnter={() => this.onHover(index)}
```
blboyle: Would it be possible to clean this up a bit. 

Maybe having the `setState` used for both enter and leave?

Or, have it be something like `onMousesomething` and have the option to have undefined there allow for setting it above.

Is what I'm saying making sense?

polaris-react/src/components/DataTable/DataTable.tsx
```diff
@@ -272,6 +302,10 @@ class DataTableInner extends PureComponent<CombinedProps, DataTableState> {
     }));
   };
 
+  private onHover = (row: number) => {
```
chloerice:
```suggestion
  private handleHover = (row?: number) => () => {
```

polaris-react/src/components/DataTable/DataTable.tsx
chloerice:
```suggestion
        onMouseEnter={this.handleHover(index)}
```

polaris-react/src/components/DataTable/DataTable.tsx
```diff
@@ -420,8 +454,14 @@ class DataTableInner extends PureComponent<CombinedProps, DataTableState> {
     );
 
     return (
-      <tr key={`row-${index}`} className={className}>
+      <tr
+        key={`row-${index}`}
+        className={className}
+        onMouseEnter={() => this.onHover(index)}
+        onMouseLeave={() => this.setState({rowHovered: undefined})}
```
chloerice:
```suggestion
        onMouseLeave={this.handleHover()}
```

polaris-react/src/components/DataTable/DataTable.scss
```diff
@@ -259,3 +260,15 @@ $breakpoint: 768px;
...
 }
+
+.FixedFirstColumn {
+  position: absolute;
+
+  background: white;
+
+  z-index: 1;
+  border-collapse: collapse;
+  &.Visible {
```
chloerice: (This will fix the lint error failing CI)

```suggestion
  
  &.Visible {
```

chloerice (commented):
This is looking great!! Can you add an example to the readme so Chromatic can catch visual regressions and we can tophat on mobile?

JeremyLudwigDev:
Leaving this question here for now, since the team is bursting:

Regarding first-column min-width, cell wrapping etc:

Currently the `.Cell` has `whitespace: nowrap` set, which for the first column is overridden to `whitespace: normal`, allowing the content to wrap. When there are many columns, this leads to a lot of wrapping on the first column, while the rest of the columns take the space they need with no wrapping.

<details>
<summary>Example from Storybook:</summary>

image
</details>

We had talked about adding a sensible min-width to the first column and eventually adding the ability to manually resize the column. Any idea why we don't want the other columns to wrap?

<details>
<summary>Example of other columns wrapping:</summary>

image

</details>

blboyle (dismissed):
There is a bug that I found on iOS that has the rows not line perfectly, which as there are many rows exaggerates the difference over time.

blboyle (commented):
Border changes are fixed 👍🏻

polaris-react/playground/Playground.tsx
```diff
@@ -1,11 +1,356 @@
-import React from 'react';
+import React, {useCallback, useState} from 'react';
```
blboyle: @JeremyLudwigDev You'll have to remove this file from tracking.

blboyle:
@JeremyLudwigDev 

Found an issue when tabbing through items in the first column, The size of the focus ring. I think it has to do with the truncation.

Chrome https://screenshot.click/24-57-88kb0-wgxg6.mp4
Firefox https://screenshot.click/24-57-dsshc-4z05p.mp4

polaris-react/src/components/DataTable/DataTable.tsx
```diff
@@ -468,13 +505,31 @@ class DataTableInner extends PureComponent<CombinedProps, DataTableState> {
       this.setState((prevState) => ({
         ...this.calculateColumnVisibilityData(prevState.condensed),
       }));
-    }, 500);
+    }, 100)();
 
     if (this.props.stickyHeader && this.stickyHeaderActive) {
       this.stickyHeaderScrolling();
     }
   };
```
philschoefer, line 540: I noticed that debounce likely isn't what we want here. Debounce cues up all the calls that would have happened and then eventually fires _all of them_ once the defined delay has passed. There is no reason after 100ms to fire the calculation 50+ times.

I'd suggest to replace the entire handler with the following which gets rid of most of the jank (sorry can't make it a GH suggestion since only the one line is marked as changed):

```tsx
class DataTableInner {
   private scrollStopTimer: ReturnType<typeof setTimeout> | null

   ...
}

private scrollListener = () => {
    if (this.scrollStopTimer) {
      clearTimeout(this.scrollStopTimer);
    }

    this.scrollStopTimer = setTimeout(() => {
      }));
    }, 100);

    this.setState({
      isScrolledFarthestLeft: this.scrollContainer.current?.scrollLeft === 0,
    });

    if (this.props.stickyHeader && this.stickyHeaderActive) {
      this.stickyHeaderScrolling();
    }
};
```

polaris-react/UNRELEASED.md
```diff
@@ -0,0 +1,52 @@
...
+### Dependency upgrades
+
+- Removed `lodash` ([#5544](https://github.com/Shopify/polaris/pull/5544))
+- Uses more permissive dependency for `@types/react` and `@types/react-dom` ([#5575](https://github.com/Shopify/polaris/pull/5575))
+
+### Code quality
+
+- Replaced `skeleton-content` and `thumbnail-size` mixins with css values ([#5630](https://github.com/Shopify/polaris/pull/5630))
+
+### Deprecations
```
philschoefer: I would expect this to only contain your specific update. Maybe rebase to ensure only your specific change is added to `UNRELEASED.md`.

polaris-react/src/components/DataTable/DataTable.scss
```diff
@@ -129,20 +121,18 @@
 
   .IncreasedTableDensity & {
     padding: var(--p-space-2) var(--p-space-3);
-
-    &:first-child {
+    &:first-of-type {
```
philschoefer: I'm not seeing the benefit to switching to `first-of-type`, especially since the referenced type isn't immediately clear. If you want to address `th` specifically you would need `th:first-of-type`. Otherwise I would leave this at `:first-child`, `:last-child` as that reads much clearer than having line `124` use one selector and just below `last-child`.

polaris-react/src/components/DataTable/DataTable.scss
```diff
@@ -300,28 +298,22 @@
     flex-direction: column;
     width: 100%;
     overflow: hidden;
+    border-spacing: 0;
```
philschoefer, line 303: I couldn't make out that this made any difference to the final output, but I might have not been testing the correct scenario.

polaris-react/src/components/DataTable/DataTable.scss
```diff
@@ -337,3 +329,19 @@
     box-shadow: var(--p-shadow-base);
   }
 }
+
+.FixedFirstColumn {
+  position: absolute;
+
+  background: white;
```
philschoefer: I think we should tie this to `var(--p-surface);`. It's the closest Polaris token that matches this type of use-case I can find. At the very least it ties this `background-color` to the `Card` component's `background-color` which sets the white background for the rest of the table.

This avoids color mis-match should the `--p-surface` color ever be updated.

polaris-react/src/components/DataTable/DataTable.scss
```diff
@@ -337,3 +329,19 @@
     box-shadow: var(--p-shadow-base);
   }
 }
+
+.FixedFirstColumn {
+  position: absolute;
+
```
philschoefer: Let's group these together:

```scss
.FixedFirstColumn {
   position: absolute;
   background: var(--p-surface);
   z-index: 3;
   border-spacing: 0;

   @media #{$p-breakpoints-md-down} {
      ...
   }
   ...
}
```

polaris-react/src/components/DataTable/DataTable.scss
```diff
@@ -337,3 +329,19 @@
...
+  background: white;
+  @media #{$p-breakpoints-md-down} {
+    z-index: 1;
+  }
+
+  z-index: 3;
```
philschoefer: Sorry, I can't remember why this had to be higher than on mobile. Can we just use `z-index: 1` here, as well? This way the scrollbar remains visible rather than getting cut-off by the fixed column.

https://screenshot.click/24-05-821ic-15bqm.mp4

polaris-react/src/components/DataTable/DataTable.scss
```diff
@@ -337,3 +329,19 @@
...
+  border-spacing: 0;
+
+  &.Visible th {
```
philschoefer: Nit: `Visible` seems a bit misleading to me since it's technically always visible and it falsly implies that it is hidden otherwise.

What do you think about changing this to something like `&.showBorder`. It's a bit verbose, but expresses more clearly what this actually does (especially in the react code where I don't see the CSS properties).

polaris-react/src/components/DataTable/README.md
```diff
@@ -437,6 +440,143 @@ function FullDataTableExample() {
...
+
+  function sortRows(rows, index, direction) {
+    return [...rows].sort((rowA, rowB) => {
+      let type;
+      if (rowA[index].length > 1) {
+        type = !isNaN(parseFloat(rowA[index]))
+          ? 'number'
+          : !isNaN(parseFloat(rowA[index].substring(1)))
+          ? 'currency'
+          : 'string';
```
philschoefer: This seems overly complex considering that this is supposed to be a simplified example. What do you think about using the demo data from the `Data table with all of its elements` example just above?

philschoefer (commented):
Left a couple of smaller comments of areas I think could be simplfied a bit more. 

I also added a suggestion for how to get rid of the jankiness without impacting functionality too much. (table navigation will lag behind a little bit, but that seems preferable over the jankiness imo).

polaris-react/src/components/DataTable/DataTable.scss
JeremyLudwigDev: Maybe you weren't scrolled down so the sticky header was not visible? I double checked and it's definitely needed to remove the extra spacing on the sticky header. I had to add it because I changed the markup of the sticky header to `<table>` to match everything else.

polaris-react/src/components/DataTable/DataTable.scss
JeremyLudwigDev: Good call. Can't remember if I had some reason for that...

polaris-react/src/components/DataTable/DataTable.scss
JeremyLudwigDev: Done.

polaris-react/src/components/DataTable/DataTable.scss
JeremyLudwigDev: This was admittedly a bit picky. On mobile, I wanted the scroll bar to appear above the fixed first column, but if the scroll bar is above, so are the focus rings, leading to this:
https://screenshot.click/24-55-2jqgo-b84je.mp4

I thought I would compromise and try and do what's most appropriate to the given screen size.

polaris-react/src/components/DataTable/DataTable.scss
JeremyLudwigDev: How about `&.Separate` so it's more semantic...

BTW, is there a naming convention for these classes? Like it should they only be capitalized if they correspond to a component? Therefore the above should be `&.separate`?

polaris-react/src/components/DataTable/DataTable.tsx
JeremyLudwigDev: This is a huuuge improvement. Thanks for that. Jank is well within acceptable range now.

polaris-react/src/components/DataTable/README.md
JeremyLudwigDev: For thorough accessibility testing purposes, I wanted all columns to be sortable, so I had to write this function that could sort strings, currency, and bare numbers. I'm sure there's a tidier way to write it, but I think it's worth having that ability to manually test all the columns since we will want that in the Reports tables.

polaris-react/src/components/DataTable/DataTable.scss
philschoefer: Makes sense. Thank you.

polaris-react/src/components/DataTable/DataTable.scss
philschoefer: Thanks for the reminder, I had forgotten about that. Makes sense.

polaris-react/src/components/DataTable/DataTable.scss
philschoefer: I'm not aware of a convention. I always make all `CSS` classes start with a capital letter.

polaris-react/src/components/DataTable/README.md
philschoefer: Can we replace the demo data so there are less cases to deal with? I think what makes the function look complex is that nested ternary which the Shopify linter will complain about, as well if you were to copy that code into `shopify/web` as a starting point.

philschoefer:
Sorry I just noticed the table navigation doesn't work properly when the fixed column is active. https://screenshot.click/24-11-koz09-cxu9b.mp4

What I would expect is that the table scrolls the next column to line up with the right edge of the fixed column as opposed to the left edge of the table underneath.

polaris-react/src/components/DataTable/utilities.ts
```diff
@@ -19,15 +19,15 @@ export function measureColumn(tableData: TableMeasurements) {
...
       tableData.firstVisibleColumnIndex = Math.min(
         firstVisibleColumnIndex,
         index,
       );
     }
-
-    return {leftEdge, rightEdge, isVisible};
+    return {leftEdge, rightEdge, isVisible, width, index};
   };
 }
```
JeremyLudwigDev, line 38: Leaving this here in case some one else has to pick this up:

There's a bug in this function in the case where there are no columns that are fully visible, it makes all columns `columnVisibilityData[n].visible=false`, causing it to default to the last column being the first visible column, which makes the nav jump around. This was a bug in the original implementation but it must never have came up. On mobile, we will encounter this situation a lot if we have a fixed first column, so we'll have to account for it.

https://screenshot.click/25-49-wcss8-15lj3.mp4
https://screenshot.click/25-59-bvwnz-4knsd.mp4

JeremyLudwigDev: For posterity: we adjusted this function to not include the minVisible pixels for the left edge of the column which clears up the navigation issue.

polaris-react/src/components/DataTable/README.md
JeremyLudwigDev: Fixed the nested ternary but left the multiple kinds of data and the more complicated sorting algorithm.

polaris-react/UNRELEASED.md
JeremyLudwigDev: Removed this file altogether as per the new changeset workflow

polaris-react/src/components/DataTable/DataTable.scss
```diff
@@ -129,20 +121,18 @@
...
-        padding-right: 0;
-        padding-left: var(--p-space-1);
-      }
     }
 
     &:last-child {
       padding-right: var(--p-space-3);
     }
   }
+  .IncreasedTableDensity & .Heading-left {
```
chloerice, line 134:
```suggestion

  .IncreasedTableDensity & .Heading-left {
```

.changeset/fluffy-pandas-help.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/polaris': minor
+---
+
+Add a fixed first column and sticky header to the Datatable component
```
chloerice:
```suggestion
Added `hasFixedFirstColumn` and `firstColumnMinWidth` props to `DataTable`
```

chloerice (changes requested):
Hey @JeremyLudwigDev 👋🏽 This is looking great!!! 🙌🏽

The one issue I'm seeing (mobile and desktop) is that the column visibility is off by 1 because it always shows the fixed first column as not visible.

Screen Shot 2022-05-31 at 5 37 17 PM

philschoefer:
Thanks @chloerice. I updated the implementation. Let me know if there is anything else.

chloerice (approved):

chloerice:
You may need to rebase on main, there's been a couple of releases this week and I think one of them affects the bundle size and should help fix the size limit check not passing 🤔

PR #5544, opened 2022-04-18
Remove lodash

tmlayton:

Reduces bundle size and uses a faster equal check

- Copies over lodash debounce and modernizes it
- Replaces lodash isEqual with react-fast-compare
- Simplifies comparison in ranger slider component
- Remove lodash
- Brings in only specific lodash functions as dev dependencies 

### 🎩 checklist

- [x] Tested


tmlayton:
Need to verify https://github.com/Shopify/fbs/issues/5846#issuecomment-521376803 is not an issue with react-fast-compare

polaris-react/.storybook/polaris-readme-loader.js
```diff
@@ -3,7 +3,8 @@ const chalk = require('chalk');
 const grayMatter = require('gray-matter');
 const MdParser = require('./md-parser');
 const React = require('react');
-const lodash = require('lodash');
+const startCase = require('lodash.startcase');
```
BPScott: According to https://lodash.com/per-method-packages  the per-package lodash stuff is considered deprecated and its usage is discouraged. I think it'd be better to have a devDependency on `lodash` and import from `lodash/startCase` / `lodash/camelCase` in these dev dependency cases.

tmlayton: Thanks Ben

polaris-react/package.json
```diff
@@ -114,6 +113,7 @@
     "fs-extra": "^7.0.1",
     "glob": "^7.1.4",
     "gray-matter": "^4.0.2",
+    "lodash": "^4.17.21",
```
kyledurand, line 116: Why do we still need this as a dev dep?

tmlayton:
@lucasfrosty tried to verify https://github.com/Shopify/fbs/issues/5846#issuecomment-521376803 is not an issue with `react-fast-compare`, but ran into issues with the `build-consumer` script and rebuilding the fbs app so I’d rather cut a release candidate. Per type signatures this is not a breaking change, but the implementation is. For example, lodash `isEqual` compares `TypedArray`s like `Int8Array`.

polaris-react/src/utilities/debounce.ts
```diff
@@ -0,0 +1,165 @@
...
+  let maxing = false;
+  let trailing = true;
+
+  // Bypass `requestAnimationFrame` by explicitly setting `wait=0`.
+  const useRAF = !waitArg && waitArg !== 0;
+
+  if (typeof func !== 'function') {
+    throw new TypeError('Expected a function');
+  }
+  const wait = Number(waitArg) || 0;
```
kyledurand: I don't think the `Number()` call is needed here

polaris-react/src/utilities/debounce.ts
```diff
@@ -0,0 +1,165 @@
...
+  if (typeof options === 'object') {
+    leading = Boolean(options.leading);
```
kyledurand, line 32: I'm guessing checks these are coming from a non typescript source?

polaris-react/src/utilities/debounce.ts
```diff
@@ -0,0 +1,165 @@
...
+    maxWait = maxing ? Math.max(Number(options.maxWait) || 0, wait) : undefined;
+    trailing = 'trailing' in options ? Boolean(options.trailing) : trailing;
+  }
+
+  function invokeFunc(time: number) {
+    const args = lastArgs;
+    const thisArg = lastThis;
+
+    // eslint-disable-next-line no-multi-assign
+    lastArgs = lastThis = undefined;
```
kyledurand: For clarity:

```suggestion
    lastArgs = undefined;
    lastThis = undefined;
```

kyledurand (approved):
This looks good to me, just some questions and comments around the debounce function. If we're going to maintain it let's make it as easy to understand as possible.

kyledurand:
> ran into issues with the build-consumer

The build consumer script may be broken while we're building out the mono repo. I'd try:
1. Spinning up the service
2. Clone polaris into the instance
3. checkout this branch
4. build
5. yalc link to web
6. restart web service

Edit: That said, I'm totally fine with shipping and cutting an RC if that's easier to test

polaris-react/package.json
tmlayton: We still need it for the dev environment. There are a two functions used in the script which parses the readme examples into stories.

tmlayton: Here https://github.com/Shopify/polaris/blob/2141ca5f31526d50421011841d389800be05a2af/polaris-react/.storybook/polaris-readme-loader.js#L411

alex-page (approved):

BPScott:
IIRC `build-consumer` style stuff doesn't play nice with updating dependencies of the package you're trying to push into an app, so you'd need a beta regardless of if the monorepo was behaving.

BPScott (approved):
Not tested but changes look great to me. Love that we're dropping the lodash dependency.

polaris-react/src/utilities/debounce.ts
tmlayton: This is because `options.leading` can be undefined and `leading` is a boolean.

PR #5548, opened 2022-04-19
Carry over content from current web site

martenbjork:
This is a stop gap PR to unblock website development:

- Copy paste `icons` into next.js `public` dir (should move to monorepo root later on)
- Copy paste styleguide images into `public` dir
- Copy paste markdown files into `src/pages`

This will be followed by: https://github.com/Shopify/polaris/pull/5522 . By breaking these things out into a separate PR, the website PR's diff becomes less crazy.


alex-page (approved):
>This PR is pretty stupid, but would appreciate a review

This PR is pretty stupid, here is a review.

PR #5557, opened 2022-04-19
Split CI and add stylelint-polaris CI

alex-page:
Creates a platform for CI to be the smallest possible task for some repos

- [x] Split CI out into seperate jobs
- [x] Trigger CI on changes to [paths](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths)
- [x] Add new CI for `stylelint-polaris`
- [x] Split out tokens CI from `ci.yml`
- [x] Rename `ci.yml` to `ci-polaris-react.yml`
- [x] Add types for stylelint-polaris tests
- [x] Fix types in vscode plugin

Thoughts on adding `yarn build` and `cache` to all actions...

> It’s annoying that we have to run yarn build for repos like stylelint-polaris . We currently need to do this because the global type check relies on files being built in projects to pass. This is going to slow down CI significantly for smaller packages with less cross dependencies. It also means we need to cache all the loom stuff as well.

Next steps to address this
- [ ] Publish tokens library
- [ ] Remove the need for root level build for root level `tsc` to pass
- [ ] Update actions and remove unnecessary cached files and scripts

### DIFF
--- a/.github/workflows/ci-stylelint-polaris.yml
+++ b/.github/workflows/ci-stylelint-polaris.yml
@@ -33,6 +33,9 @@ jobs:
             polaris-react/.loom
             polaris-react/build/ts/latest
             polaris.shopify.com/.next/cache
+          key: ${{ runner.os }}-node${{ matrix.node_version }}-test-v1-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-node${{ matrix.node_version }}-test-v1-
 
       - run: yarn --frozen-lockfile
       - run: yarn build
--- a/.github/workflows/ci-stylelint-polaris.yml
+++ b/.github/workflows/ci-stylelint-polaris.yml
@@ -3,11 +3,13 @@ name: @shopify/stylelint-polaris CI
 on:
   pull_request:
     paths:
-      - 'stylelint-polaris/**'
+      - 'stylelint-polaris'
+      - 'polaris-react/src/tokens/token-groups'
   push:
     branches: main
     paths:
-      - 'stylelint-polaris/**'
+      - 'stylelint-polaris'
+      - 'polaris-react/src/tokens/token-groups'
 
 jobs:
   test:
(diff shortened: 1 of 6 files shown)


stylelint-polaris/tests/index.js
```diff
@@ -16,6 +16,17 @@ const configs = {
   },
 };
 
+/**
+ * @typedef {Object} Config
+ * @property {string} path
+ * @property {string[]} files
+ * @property {number} expectedFailures
+ */
```
aaronccasanova: Might as well add this to the `configs` var above e.g.

```ts
/**
 * @type {{ [configName: string]: Config }}
 */
const configs = {...}
```

.github/workflows/ci-stylelint-polaris.yml
```diff
@@ -0,0 +1,30 @@
...
+        node_version: ['14.17.0', '16.13.0']
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node_version: ${{ matrix.node_version }}
+          cache: yarn
+
+      - run: yarn --frozen-lockfile
+      - run: yarn type-check
```
aaronccasanova, line 42: The `stylelint-polaris` `tsconfig` defines `emitDeclarationOnly: true`: https://github.com/Shopify/polaris/blob/4b02c9743babe60be5ffb80fc124db171a700ece/stylelint-polaris/tsconfig.json#L5-L12

We might need to add a `"types": "dist/index.d.ts"` key to the `stylelint-polaris` package.json

stylelint-polaris/package.json
```diff
@@ -29,14 +29,14 @@
...
+    "prepublishOnly": "yarn build"
   },
   "dependencies": {
     "postcss-value-parser": "^4.2.0"
   },
   "peerDependencies": {
     "stylelint": "^14.1.0"
-  }
+  },
+  "types": "dist/index.d.ts"
```
aaronccasanova: **suggestion (non-blocking):** Conventionally this sits near the `"main"` and `"module"` keys

aaronccasanova (approved):

PR #5551, opened 2022-04-19
Support nodejs v14

alex-page:
### WHY are these changes introduced?

The Shopify CLI supports NodeJS v14. Polaris is a key part of this when using the CLI to create a new application. This adds back support and tests for NodeJS v14.

### WHAT is this pull request doing?

- [x] Creates a GitHub action for nodejs v14
- [x] Adds a minumum nodejs v14 to the package.json file

PR #5560, opened 2022-04-19
Update NodeJS support wording

BPScott:

See https://github.com/Shopify/polaris/pull/5551/files#r853299677

Clarify that we support LTS versions till they stop being supported
by the node foundation

### DIFF
--- a/documentation/Nodejs support.md
+++ b/documentation/Nodejs support.md
@@ -1,6 +1,6 @@
 # NodeJS support
 
-`@shopify/polaris` supports the [last two long term support (LTS) versions](https://nodejs.org/en/about/releases/) of NodeJS. This matches our approach for the [Shopify CLI](https://github.com/Shopify/shopify-cli) a tool used by developers to quickly create applications that uses Polaris.
+`@shopify/polaris` supports node's long term suppport (LTS) versions that are in the active or maintenance stages of their [lifecycle](https://nodejs.org/en/about/releases/).
 
 ## Local development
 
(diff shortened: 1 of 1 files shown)


documentation/Nodejs support.md
```diff
@@ -1,6 +1,6 @@
 # NodeJS support
 
-`@shopify/polaris` supports the [last two long term support (LTS) versions](https://nodejs.org/en/about/releases/) of NodeJS. This matches our approach for the [Shopify CLI](https://github.com/Shopify/shopify-cli) a tool used by developers to quickly create applications that uses Polaris.
+`@shopify/polaris` supports node's long term suppport (LTS) versions that are in the active or maintenance stages of their [lifecycle](https://nodejs.org/en/about/releases/).
```
alex-page, line 3:
```suggestion
`@shopify/polaris` supports node's long term suppport (LTS) versions that are in the active or maintenance stages of their [lifecycle](https://nodejs.org/en/about/releases/). We should also ensure support for critical repositories at Shopify like `shopify/web`, 1st party applications and `shopify/cli`.
```

BPScott:
```suggestion
`@shopify/polaris` supports node's long term suppport (LTS) versions that are in the active or maintenance stages of their [lifecycle](https://nodejs.org/en/about/releases/). We should also ensure support for critical repositories at Shopify like `shopify/web`, 1st party applications and `shopify/cli`.
```

alex-page (approved):

PR #5562, opened 2022-04-20
Consistent package scripts and output

### LINKED ISSUE #5433: Consistent package scripts
- [ ] Make sure all packages have consistent output e.g. `dist`
- [ ] Make sure all packages have consistent script names
- [ ] Remove top level dependency on build and type-check
- [ ] Remove top level workspace scripts, add docs or find replacement

sam-b-rose:

Fixes #5433 

- [x] Make sure all packages have consistent output e.g. `dist`
- [x] Make sure all packages have consistent script names
- [x] Remove top level workspace scripts, add docs or find replacement
- [ ] ~~Remove top level dependency on build and type-check~~


polaris-for-vscode/server/src/server.ts
```diff
@@ -62,28 +61,29 @@ connection.onInitialize((params: InitializeParams) => {
...
-            ? CompletionItemKind.Color
-            : CompletionItemKind.Variable,
-      };
-    });
-
-    groupedCompletionItems[category] = completionItems;
-
-    allCompletionItems = allCompletionItems.concat(completionItems);
+  for (const tokenGroup in groupedTokens) {
+    if (Object.prototype.hasOwnProperty.call(groupedTokens, tokenGroup)) {
```
aaronccasanova, line 65:
```suggestion
    if (groupedTokens.hasOwnProperty(tokenGroup)) {
```

polaris-for-vscode/server/src/server.ts
```diff
@@ -104,12 +104,18 @@ connection.onCompletion(
...
 
     for (const tokenGroup in tokenGroupPatterns) {
-      const category = tokenGroup as keyof typeof tokenGroupPatterns;
-
-      if (tokenGroupPatterns[category].test(currentText)) {
-        const currentCompletionItems = groupedCompletionItems[category];
-        if (currentCompletionItems) {
-          matchedCompletionItems = matchedCompletionItems.concat(currentCompletionItems);
+      if (
+        Object.prototype.hasOwnProperty.call(tokenGroupPatterns, tokenGroup)
```
aaronccasanova, line 108: Same as above

.github/workflows/ci-stylelint-polaris.yml
```diff
@@ -25,14 +25,11 @@ jobs:
...
           cache: yarn
       - uses: actions/cache@v2
-        name: Restore build cache
+        name: Restore cache
         with:
           path: |
-            .loom
+            .eslintcache
             **/*.tsbuildinfo
-            polaris-react/.loom
```
aaronccasanova, line 33: Should we keep this around until we remove the root level `yarn build` step from this workflow?

polaris-for-vscode/package.json
```diff
@@ -18,36 +18,27 @@
     "polaris",
     "shopify"
   ],
-  "engines": {
-    "vscode": "^1.64.0"
-  },
```
aaronccasanova: This appears to be a required field: https://code.visualstudio.com/api/references/extension-manifest

.github/workflows/ci-stylelint-polaris.yml
sam-b-rose: I think this is still safe to remove. From what I noticed, it seems the only the `polaris-react` test data from Jest was being stored here. So, I think we only need to cache this for the `polaris-react` CI workflow

polaris-for-vscode/package.json
sam-b-rose: Ah, good catch! Yarn was complaining about this not being a valid engine type... will add back!

polaris-for-vscode/server/src/server.ts
sam-b-rose: I actually did try this initially but got [this ESLint error of `no-prototype-builtins`](https://eslint.org/docs/rules/no-prototype-builtins) :(

aaronccasanova: Oof and `Object.hasOwn(groupedTokens, tokenGroup)` is only available in Node >=16.9.0

package.json
```diff
@@ -32,6 +27,7 @@
     "@shopify/stylelint-plugin": "^11.0.0",
     "@shopify/typescript-configs": "^5.1.0",
     "eslint": "^8.3.0",
+    "npm-run-all": "^4.1.5",
```
alex-page, line 30: `npm-run-all` doesn't seem to be used 🤔

aaronccasanova: https://github.com/Shopify/polaris/blob/6ad54d32d0cedf507d7fdd66d2d5677ce0787098/polaris-for-vscode/package.json#L33-L35

sam-b-rose: It is being used for running all the lint scripts with `run-s lint:*`. A bit confusing I know 😅

.eslintignore
```diff
@@ -1,9 +1,7 @@
-node_modules
 .loom
-.vscode-test
+.next
+build
+build-internal
+data
```
alex-page: `/stylelint-polaris/data` unsure about this. Should we rename these dirs in a seperate PR

sam-b-rose: Good call, will update to put these generated data items in a tmp folder instead. I feel that is a more standard practice and naming convention

alex-page: tmp-tokens even might be a good call to clarify they will be removed with tokens

.eslintignore
```diff
@@ -1,9 +1,7 @@
-node_modules
 .loom
-.vscode-test
+.next
+build
+build-internal
```
alex-page: I would prefer to keep these opinionated for now
```suggestion
/polaris-react/build
/polaris-react/build-internal
```

.eslintignore
```diff
@@ -1,9 +1,7 @@
...
 dist
-out
-/polaris-react/build
-/polaris-react/build-internal
-/polaris.shopify.com/.next
-/stylelint-polaris/data
+node_modules
```
alex-page, line 6:
```suggestion
```

.gitignore
```diff
@@ -1,6 +1,13 @@
...
 *.tsbuildinfo
-node_modules
+*.vsix
 .DS_Store
+.eslintcache
 .loom
+.next
 .shadowenv.d
+build
+build-internal
```
alex-page: These are already in a `.gitignore` in `polaris-react` so maybe we can remove it?

alex-page: `*.scss.d.ts` could be added globally and removed from `polaris-react` `.gitignore`.

.eslintignore
sam-b-rose: Agreed, updated in other `.*ignore` files as well

.gitignore
sam-b-rose: I removed all other `.gitignore` files in favor of having a single ignore file

sam-b-rose: > `*.scss.d.ts` could be added globally

Do we need to add it if it is already part of the build output? Just trying to keep the ignores trim

.eslintignore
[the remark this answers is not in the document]
samrose3: This was just me keeping these files alphabetical. Very superfluous and not documented, but I felt it a reasonable cleanup given while modifying these ignore files

alex-page (approved):

alex-page (approved):

tmlayton:
I see the `build-consumer` script was removed. There is still documentation in the README that refers to this script as the way to test in consuming apps. Also, the script itself still exists.

PR #5570, opened 2022-04-20
Remove whitespace from CustomProperties

alex-page:

When doing releases for `polaris-styleguide` the [Demo.test.tsx](https://github.com/Shopify/polaris-styleguide/blob/e96d88c3a029915ad1d6e8092164bfab0aa03a63/web/styleguide-demo/Demo/test/Demo.test.tsx#L54-L58) checks the output of custom properties. The whitespace is making it difficult to use variables for this data as it's changing with whitespace.

- [x] Removing whitespace from the declaration

### DIFF
--- a/polaris-react/src/components/CustomProperties/styles.ts
+++ b/polaris-react/src/components/CustomProperties/styles.ts
@@ -7,10 +7,10 @@ import {
   OSColorSchemes,
 } from '../../tokens';
 
-const tokenValues = getStaticCustomProperties(tokens);
-const declaration = getColorSchemeDeclarations('light', tokens, osColorSchemes);
+const staticCustomProperties = getStaticCustomProperties(tokens);
+const colorSchemeDeclarations = getColorSchemeDeclarations('light', tokens, osColorSchemes);
 
-const defaultDeclarations = `${declaration}${tokenValues}`;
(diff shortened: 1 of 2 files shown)


laurkim (approved):
nice 💯 🚀

polaris-react/src/components/CustomProperties/styles.ts
```diff
@@ -7,10 +7,10 @@ import {
...
 } from '../../tokens';
 
-const defaultDeclarations = `
-  ${getColorSchemeDeclarations('light', tokens, osColorSchemes)}
-  ${getStaticCustomProperties(tokens)}
-`;
+const tokenValues = getStaticCustomProperties(tokens);
+const declaration = getColorSchemeDeclarations('light', tokens, osColorSchemes);
+
+const defaultDeclarations = `${declaration}${tokenValues}`;
```
aaronccasanova:
```suggestion
const staticCustomProperties = getStaticCustomProperties(tokens);
const colorSchemeDeclarations = getColorSchemeDeclarations('light', tokens, osColorSchemes);

const defaultDeclarations = `${colorSchemeDeclarations}${staticCustomProperties}`;
```

**nit (non-blocking):** I try to keep the same naming convention as the getters unless there's a naming conflict.

aaronccasanova (approved):

PR #5574, opened 2022-04-21
Tokens create README.md

romellogoodman:

Creates a README.md for the tokens package

### DIFF
--- a/documentation/guides/migrating-tokens-from-v4-to-v5.md
+++ b/documentation/guides/migrating-tokens-from-v4-to-v5.md
@@ -1,0 +1,3 @@
+## Migrating from v4 to v5
+
+Polaris Tokens v5.0.0 ([full release notes](https://github.com/Shopify/polaris-react/releases/tag/tokens-v5.0.0)) features an overhaul of the package. The Sass variables, CSS Filters, and Rails have all been removed. The new version is based on json files and utility javascript functions. In order to find your replacements please navigate to [polaris.shopify.com/tokens/getting-started-with-tokens](https://polaris.shopify.com/tokens/getting-started-with-tokens).
--- a/polaris-tokens/README.md
+++ b/polaris-tokens/README.md
@@ -31,10 +31,6 @@ import {tokens} from '@shopify/polaris-tokens';
 console.log(tokens);
 ```
 
-## Migrating from v4 to v5
-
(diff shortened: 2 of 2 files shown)


polaris-tokens/README.md
```diff
@@ -0,0 +1,44 @@
...
+```js
+import {tokens} from '@shopify/polaris-tokens';
+
+// Access all of the available token groups
+console.log(tokens);
+```
+
+## Migrating from v4 to v5
+
+Polaris Tokens v5.0.0 ([full release notes](https://github.com/Shopify/polaris-react/releases/tag/tokens-v5.0.0)) features an overhaul of the package. The Sass variables, CSS Filters, and Rails have all been removed. The new version is based on json files and utility javascript functions. In order to find your replacements please navigate to [polaris.shopify.com/tokens/getting-started-with-tokens](https://polaris.shopify.com/tokens/getting-started-with-tokens).
```
alex-page: We should make a migration document. We have some in `documentation` directory in root.

documentation/guides/migrating-tokens-from-v4-to-v5.md
```diff
@@ -0,0 +1,3 @@
+## Migrating from v4 to v5
+
+Polaris Tokens v5.0.0 ([full release notes](https://github.com/Shopify/polaris-react/releases/tag/tokens-v5.0.0)) features an overhaul of the package. The Sass variables, CSS Filters, and Rails have all been removed. The new version is based on json files and utility javascript functions. In order to find your replacements please navigate to [polaris.shopify.com/tokens/getting-started-with-tokens](https://polaris.shopify.com/tokens/getting-started-with-tokens).
```
alex-page, line 3: @aaronccasanova @romellogoodman I haven't had a deep look at the tokens work. Should it for now be `json` and `css` output?

romellogoodman: Want to note that this is more or less just a placeholder. I was going to flush to be more explicit about the removal of the old apis, the new apis, types, and utilities

alex-page: Oh sweet. Feel free to ignore me then and ship!

alex-page (approved):

documentation/guides/migrating-tokens-from-v4-to-v5.md
aaronccasanova: Currently, the package exposes all token groups from a `tokens` javascript object and by nature of not using the `pkg.exports` key it also allows access to the individual `JSON` token group files. We don't expose any `CSS` or `SCSS` but do think we should in the near future so perhaps we should exclude that from the documentation.

> but do think we should in the near future

e.g. We currently generate the `CustomProperties` component stylesheet at runtime for every instance. This should probably be generated once during the `polaris-tokens` `yarn build` and simply imported in the `CustomProperties` component (`import '@shopify/polaris-tokens/styles/custom-properties.css'`)

PR #5575, opened 2022-04-21
Uses more permissive dependency for @types/react

felipeleusin:

While exploring how to migrate Shopify Email to React18, I noticed that I had to resort to module resolutions since polaris was embedding it's own version of `@types/react`. While having this sort of overly permissive versioning here might seem brittle it is actually how I see most packages that are built in the DefinitelyTyped repo used https://unpkg.com/browse/@types/react-measure@2.0.8/package.json

This PR ensures that polaris-react has a more permissive dependency of `@types/react` and `@types/react-dom` so that it doesn't cause conflicts for consuming applications.

It then includes the current version as devDependencies

### 🎩 checklist


chloerice (approved):

ryanwilsonperkin (approved):
Nice fix, I agree and have seen similar use of `*` requirement on definitely-typed packages when specifying them as dependencies in libraries. Allows the consuming application to narrow the version specifier if need-be.

PR #5578, opened 2022-04-21
Tokens build script

romellogoodman:

Fixes #0000 

This is the first pass at introducing a build script to the polaris-tokens package.

* Turns all the token files from `json` to `js` to allow commenting and documentation to live with the values
* Creates build scripts that generate the json file for each group and the css file for the entire package
* Removes ajv package since we are using typescript and no longer need the schema validation logic
* Adds jest and some basic tests


polaris-tokens/package.json
```diff
@@ -4,9 +4,9 @@
   "description": "",
   "main": "dist/index.js",
   "scripts": {
-    "build": "tsc -b",
+    "build": "tsc -b && node scripts/build",
     "test": "yarn validate-tokens",
-    "validate-tokens": "ajv validate -s schemas/token-group.json -d 'token-groups/*.json'"
+    "validate-tokens": "ajv validate -s schemas/token-group.json -d 'dist/*.json'"
```
aaronccasanova: We shouldn't need to validate the JSON schemas now that the token groups are written in TypeScript.

polaris-tokens/scripts/build.js
```diff
@@ -0,0 +1,119 @@
...
+
+  fs.writeFileSync(filePath, content);
+}
+
+/**
+ ******************************
+ * Create json token group files
+ ******************************
+ */
+Object.entries(tokens).forEach(([tokenGroup, tokenProps]) => {
```
aaronccasanova: Thoughts on pulling this out to a `build/toJSON.js` file?

polaris-tokens/scripts/build.js
```diff
@@ -0,0 +1,119 @@
...
+    writeDistFile(fileName, JSON.stringify(tokenProps));
+  }
+});
+
+/**
+ ******************************
+ * Create CSS file
+ ******************************
+ */
+const staticCustomProperties = getStaticCustomProperties(tokens);
```
aaronccasanova: Thoughts on pulling this out to a `build/toStyleSheet.js` file?

Also, there is a number of `CustomProperties` tests we should bring over: https://github.com/Shopify/polaris/blob/0d3ddcc31950be8dcdc882823ab2da4459879862/polaris-react/src/components/CustomProperties/tests/CustomProperties.test.tsx#L168-L227

polaris-tokens/scripts/build.js
```diff
@@ -0,0 +1,119 @@
+const fs = require('fs');
```
aaronccasanova: This is great! Only suggestion would be to asyncify this file and breakout the transforms to separate files e.g.

```js
const {tokens} = require('../dist/index')

const tokenEntries = Object.entries(tokens)

await Promise.all([
  toJSON(tokenEntries),
  toStyleSheet(tokenEntries),
  // etc.
])
```

polaris-tokens/tokens.ts
```diff
@@ -1,12 +1,12 @@
...
-import zIndex from './token-groups/z-index.json';
+import depth from './token-groups/depth';
+import legacyTokens from './token-groups/legacy';
+import darkColorScheme from './token-groups/color.dark';
+import lightColorScheme from './token-groups/color.light';
+import motion from './token-groups/motion';
+import shape from './token-groups/shape';
+import spacing from './token-groups/spacing';
+import typography from './token-groups/typography';
+import zIndex from './token-groups/z-index';
```
aaronccasanova: Thoughts on changing these all to named imports/exports? We could only default import `JSON` modules but with the update to `.ts` files we have an opportunity to update the convention.

polaris-tokens/scripts/build.js
romellogoodman: Decided instead to just explicitly run these steps in package.json. Feel like for now we should do as little magic/extra build files. Downside being that they do run synchronously but also they each take less than a second

polaris-tokens/package.json
romellogoodman: Removed the ajv package and the schema stuff

polaris-tokens/scripts/build.js
romellogoodman: Can i add this in a follow up along with tests for the entire package? Feel like this PR is getting a little big for scope

aaronccasanova: Totally fine with that 👍

polaris-tokens/scripts/toJSON.js
```diff
@@ -0,0 +1,22 @@
+const fs = require('fs');
+const path = require('path');
+
+const {tokens} = require('../dist/index');
+
+function getFileName(fileName) {
+  return path.join(process.cwd(), 'dist', fileName);
```
aaronccasanova:
```suggestion
  return path.join(__dirname, '../dist', fileName);
```

`process.cwd()` is relative to where the script is executed and thus will have different results if ever run from the root of the monorepo.

polaris-tokens/scripts/toStyleSheet.js
```diff
@@ -0,0 +1,92 @@
...
+}
+
+const styles = `
+  :root{${defaultDeclarations}}
+  ${getColorSchemeRules(tokens, osColorSchemes)}
+  ${getKeyframes(tokens.motion)}
+`;
+
+const fileName = 'styles.css';
+const filePath = path.join(process.cwd(), 'dist', fileName);
```
aaronccasanova: Same as above

aaronccasanova (approved):

polaris-tokens/tokens.ts
```diff
@@ -1,12 +1,12 @@
...
-import zIndex from './token-groups/z-index.json';
+import {depth} from './token-groups/depth';
+import {legacy as legacyTokens} from './token-groups/legacy';
+import {colors as darkColorScheme} from './token-groups/color.dark';
+import {colors as lightColorScheme} from './token-groups/color.light';
+import {motion} from './token-groups/motion';
+import {shape} from './token-groups/shape';
+import {spacing} from './token-groups/spacing';
+import {typography} from './token-groups/typography';
+import {zIndex} from './token-groups/z-index';
```
aaronccasanova: I think we need to update this import and file to camel case `zIndex` to ensure the `toJSON` script functions as expected.

polaris-tokens/package.json
```diff
@@ -4,12 +4,11 @@
   "description": "",
   "main": "dist/index.js",
   "scripts": {
-    "build": "tsc -b",
-    "test": "yarn validate-tokens",
-    "validate-tokens": "ajv validate -s schemas/token-group.json -d 'token-groups/*.json'"
+    "build": "tsc -b && node scripts/toJSON && node scripts/toStyleSheet",
```
alex-page: Should we just have a single `build-tokens` script that runs both of these functions? I can imagine more outputs in the future and this would help keep it simple.

polaris-tokens/tests/test.test.js
```diff
@@ -0,0 +1,5 @@
+describe('foo', () => {
+  it('passes', () => {
+    expect(1).toBe(1);
+  });
+});
```
alex-page, line 5: This could be scrappy if we wanted... e.g. input of this JS gives me string output. Have a look at the tests in `stylelint-polaris` we did something similar.

alex-page (approved):

polaris-tokens/scripts/build.js
alex-page: Oh dang. I think this pattern could allow us to build more patterns in the future.

PR #5581, opened 2022-04-21
Remove non-committable check

sam-b-rose:

The non-committable check was executing on unexpected branch diffs. We need to remove this for now until we can implement a more reliable method for checking for non-committable changes.

Removes the `non-committable.yml` GitHub Actions workflow


laurkim (approved):
💯

PR #5608, opened 2022-04-25
Add non-committable GitHub Action check

### LINKED ISSUE #5582: Add back non-committable GitHub workflow check
We removed the `non-committable.yml` CI check due to an unreliable diff trigger. We should add this back with a more reliable implementation for diff checking.

Related #5581

sam-b-rose:

Fixes #5582 

The previous method of checking for changes to non-committable files was not working as we need. This is due to how the GitHub workflows [built-in path filters](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) only work on a level of individual jobs or steps. This was causing unexpected issues when trying to create a new RC for the `polaris-tokens` package.

Add back the GitHub Action for checking for non-committable changes.

### DIFF
--- a/.github/workflows/non-committable.yml
+++ b/.github/workflows/non-committable.yml
@@ -1,0 +1,24 @@
+name: Check for non committable changes
+
+on:
+  pull_request:
+    paths:
+      - 'polaris-react/**'
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


laurkim (approved):
💯

PR #5611, opened 2022-04-26
Tokens, making the description field on TokenProperties optional

romellogoodman:

Fixes #0000 

Updating the token schema to make the description field optional instead of mandatory as every token won't need a description in the long run so we can remove some bloat from the package


aaronccasanova (approved):
Thanks for this clean up!

PR #5614, opened 2022-04-26
Exclude nextjs files from vscode search

martenbjork:


polaris.shopify.com/.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+  "search.exclude": {
+    "**/.next": true,
+    "**/node_modules": true
+  }
+}
```
alex-page, line 6: I think `.gitignore` should remove files from vscode search. Is something else going on?

martenbjork: Yeeeeeees, something might be fishy... Not seeing these anymore. Maybe I had some `.` in the regex field or something. Will close and re-open if it turns up again.

PR #5617, opened 2022-04-26
Tokens, port the website to polaris-tokens rc.4

romellogoodman:

Fixes #0000 

We are nearing a stable release candidate for `@shopify/polaris-tokens` and exploring how easily it can be integrated into our systems. This PR requires the tokens from the package instead of referring to the them by their local polaris-react path.

### DIFF
--- a/polaris.shopify.com/package.json
+++ b/polaris.shopify.com/package.json
@@ -13,7 +13,7 @@
   "dependencies": {
     "@headlessui/react": "^1.5.0",
     "@shopify/polaris": "^9.5.0-rc.1 ",
-    "@shopify/polaris-tokens": "^5.0.0-rc.3",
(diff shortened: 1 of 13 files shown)


martenbjork (approved):
Rad, love it! 🔥

(Personal opinion: I think the name `createVar` is too generic outside of the tokens context. I'd consider calling it something explicit like `getCSSVariableName`.)

romellogoodman:
@martenbjork lol Aaron and I had this same exact conversation when we were creating it as whether to be brief vs verbose

martenbjork:
There are 2 types of people in this world: `terse` and `peopleWhoLikeEveryThingExtremelyExplicit`. I'm a proud member of the latter group :D

PR #5619, opened 2022-04-26
Fix dev up

kyledurand:
Fixes an issue where dev fails to start up

`error Command "dev:react" not found`

Looks like this was removed intentionally and there are ways around this but I think if our project uses dev then something should spin up at root with `dev up && dev run`

Stemming from dev.yml https://github.com/Shopify/polaris/blob/main/dev.yml#L8

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,6 @@
   ],
   "scripts": {
     "build": "yarn workspaces run build",
-    "dev:react": "yarn workspace @shopify/polaris dev",
     "format": "prettier . --write",
     "lint": "run-s lint:*",
     "lint:eslint": "eslint . --cache",
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
   ],
   "scripts": {
     "build": "yarn workspaces run build",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -13,6 +13,7 @@
   ],
   "scripts": {
     "build": "yarn workspaces run build",
+    "dev:react": "yarn workspace @shopify/polaris dev",
```
aaronccasanova: Thoughts on adding this directly to dev.yml?
https://github.com/Shopify/polaris/blob/01ccb4b961a4580fa07cfb5645b370c202e9c40c/dev.yml#L8

I imagine `dev:react` was removed to discourage the pattern of adding workspace scripts to the root

aaronccasanova (approved):

package.json
kyledurand: Sure I like that cfc9f630cffcb314a1805f9ce423a6ce9ba37fd4 👍

romellogoodman (approved):

PR #5621, opened 2022-04-26
Tokens, port vscode and stylelint to polaris-tokens rc.6

romellogoodman:

Fixes #0000 

We are nearing a stable release candidate for `@shopify/polaris-tokens` and exploring how easily it can be integrated into our systems. This PR requires moves polaris-for-vscode and stylint-polaris away from using scripts to copy the tokens locally and instead uses the polaris-tokens package.


polaris-tokens/scripts/getCustomPropertyNames.js
```diff
@@ -0,0 +1,28 @@
...
+          const tokenGroup = require(path.join(tokenGroupsDir, file));
+
+          return Object.keys(tokenGroup).map((token) => `--p-${token}`);
+        })
+        .flat(),
+    ),
+  );
+
+  return polarisTokenCustomProperties;
+};
```
aaronccasanova: Can we move this to `src/utilities` and do a quick update to use `tokens`? All the `fs` and `path` logic was a hacky workaround when we were using `json` token groups.

polaris-tokens/scripts/getKeyframeNames.js
```diff
@@ -1,6 +1,6 @@
 const path = require('path');
 
-const tokenGroupsDir = path.join(__dirname, '../../src/tokens/token-groups');
+const tokenGroupsDir = path.join(__dirname, '../src/token-groups');
```
aaronccasanova: Same as above, can we move this to `src/utilities` and update to directly import the `motion` token group?

stylelint-polaris/scripts/gen-polaris-data.js
```diff
@@ -1,10 +1,8 @@
 const fs = require('fs');
 const path = require('path');
 
-const {
-  getCustomPropertyNames,
-} = require('../../polaris-react/scripts/utilities/getCustomPropertyNames');
-const getKeyframeNames = require('../../polaris-react/scripts/utilities/getKeyframeNames');
+const getCustomPropertyNames = require('../../polaris-tokens/scripts/getCustomPropertyNames');
+const getKeyframeNames = require('../../polaris-tokens/scripts/getKeyframeNames');
```
aaronccasanova: `@shopify/polaris-tokens` should really be a dependency in `stylelint-polaris` e.g.
```js
const {getCustomPropertyNames, getKeyframeNames} = require('@shopify/polaris-tokens')
```

Also, please remove this `gen-polaris-data` script entirely, as it was only a temporary solution until this package had access to `@shopify/polaris-tokens`

polaris-for-vscode/scripts/getGroupedCustomPropertyCompletionItems.js
```diff
@@ -61,7 +42,4 @@ const getGroupedCustomPropertyCompletionItems = () => {
   );
 };
 
-module.exports = {
-  getGroupedCustomPropertyCompletionItems,
-  getCustomPropertyNames,
-};
+module.exports = getGroupedCustomPropertyCompletionItems;
```
aaronccasanova: I recommend leaving this as `module.exports = { getGroupedCustomPropertyCompletionItems }`. I've been having a lot of headaches lately with es module interop and "named exporting" all the things seems to be the silver bullet.

polaris-for-vscode/scripts/prepublish.js
```diff
@@ -1,10 +1,9 @@
 const fs = require('fs');
 const path = require('path');
 
-const {
-  getCustomPropertyNames,
-  getGroupedCustomPropertyCompletionItems,
-} = require('../../polaris-react/scripts/utilities/getCustomPropertyNames');
+const getCustomPropertyNames = require('../../polaris-tokens/scripts/getCustomPropertyNames');
+
+const getGroupedCustomPropertyCompletionItems = require('./getGroupedCustomPropertyCompletionItems');
```
aaronccasanova: This `prepublish` script should be completely removed in favor of adding `@shopify/polaris-tokens` as a dependency and consuming it's modules and assets directly.

romellogoodman:
@aaronccasanova, I went through and updated polaris-for-vscode and stylelint-polaris to use the package/utilities and remove the work arounds. I didn't verify that things work though and might need some help tophatting stylelint once everything is back in order again

package.json
```diff
@@ -5,9 +5,9 @@
     "node": "^14.17.0 || ^16.13.0"
   },
   "workspaces": [
+    "polaris-tokens",
```
sam-b-rose, line 8: Interesting, does this order affect build order?

sam-b-rose (approved):
Nice, looks solid!

package.json
romellogoodman: yes! The ci was failing all afternoon because polaris-for-vscode couldnt access the package

Issue #5613, opened 2022-04-26
Move component examples out of markdown


martenbjork:
Currently, example code for components live as part of the components' README file. They are hard coded as a code block between \`\`\` ticks. It's problematic.

# Context
On the next version of `polaris.shopify.com` (using next.js), we want to build an "examples browser" that is shown for each component. Our first attempt involves scraping all the component readmes, parsing the markdown, extracting the examples and then rendering them.

To render them, we write each example (350+) to `src/pages` so that next.js creates routes for each example. This works OK, but the code in the readme's is actually not executable (missing imports, wrappers, exports...) so we wrote a function that fixes the examples and makes them executable. Patchy to say the least.

The Polaris examples needs a global CSS import, but next.js does not allow global CSS imports, but maybe we can solve that in some way (load the CSS from a CDN using `<style>` tags?).

# Anyways, this mess has convinced me that our code examples should move out of the READMEs. Rationale:

- When code lives as hard coded strings, it's dead. You lose out on compile checks, linting and type checking. If the code goes bad, there is no way of knowing.
- We want to present our examples in a consistent manner. Therefore it helps to have the data in a structured format, like a JSON/ts file so that it's easy to loop over the examples and access the title, description and code. When the data lives in markdown, this becomes much harder, messier and more error prone.

# What would be better
- Code examples should be self contained and executable. `button-example.tsx` should contain all the code needed to run the example — imports, styles, component code and exports.
- Examples are organized in a structured (machine readable) way.

I'll get the conversation going by suggesting a new structure:

```
/polaris-react
  /src
    /components
      /Button
        /examples
          default.tsx
          outline.tsx
          destructive.tsx
          index.ts
```

`index.ts` would contain something like this:

```ts
export const examples : ComponentExample = [
  {
    title: "Default button",
    exampleText: "Some markdown...",
    fileName: "default.tsx"
  },
  {
    title: "Outline button",
    exampleText: "Some markdown...",
    fileName: "outline.tsx"
  }
  ...
]
```

PR #5626, opened 2022-04-26
Extract examples out of README files

martenbjork:
Quick stab at implementing a solution to https://github.com/Shopify/polaris/issues/5613 .

- Removes code examples from the component Readmes
- Puts them into separate `tsx` files
- Fixes broken example code (adds imports, exports, wrapper functions)
- Adds a json file with descriptions of each example

This is a one-time extraction. After this point, we can make subsequent improvements like running prettier over the examples.


polaris-react/src/components/ButtonGroup/examples/outline-button-group-with-segmented-buttons.tsx
```diff
@@ -0,0 +1,20 @@
...
+function Example() {
+  return (
+    <AppProvider i18n={translations}>
+      <ButtonGroup segmented>
+  <Button outline>Bold</Button>
+  <Button outline>Italic</Button>
+  <Button outline>Underline</Button>
+</ButtonGroup>
+    </AppProvider>
+  );
```
martenbjork, line 16: A Prettier run is needed 😊

alex-page (changes requested):
This is great but would break the styleguide. We need to commit to archiving the existing one or think of a backwards compatible solution.

martenbjork:
> This is great but would break the styleguide. We need to commit to archiving the existing one or think of a backwards compatible solution.

Oof, true. We could keep the examples in the Readme for now. That would allow the next.js website to use this new file structure for a time and then, when we're ready to remove the styleguide, we re-run the script, including the part that removes the examples from the readmes.

alex-page:
My thought is...

1. Ship new site with existing README.md example structure
2. Replace old site
3. Fix this up and make best examples ever?

martenbjork:
Yeah, I'm OK with that! 👍

sam-b-rose:
@martenbjork should we go ahead and close this one out?

romellogoodman:
@samrose3 , i think it cool to close in favor of #5940

PR #5627, opened 2022-04-26
[Listbox] Add customListId prop to Listbox

andrewmcgov:

My team is working on implementing the Location Selector outlined in the [Polaris documentation for Locations](https://polaris.shopify.com/patterns/locations): 

<details>
<summary>Screenshot 📸</summary>
Screen Shot 2022-04-26 at 8 27 54 PM
</details>

In order to support assistive technologies we need to mark this up as a combobox, but the Polaris Combobox component does not currently support our UX where the textfield and the listbox are rendered at the same level, both inside the Popover.

To mark this up as a combobox, we need to pass `aria-controls` or `aria-owns` attributes to the Textfield we are rendering. The value needs to be the ID of the `ul` element with `role="listbox"` that is rendered by the `Listbox` component. Currently that ID is generated inside the `Listbox` component, and only made available to the textfield through a React context as part of the `Combobox` component.

- Adds a `customListId` prop to the `Listbox` component to allow parent components to supply their own ID for the `ul` element with `role="listbox"`
- Passes `domId` to `onActiveOptionChange` prop on `Listbox`. This allows parent components access to the active element which can then be passed to `aria-activedescendant` on another element.

### How to 🎩

If you test the playground code below and inspect the page, you should see the `ul` with `role="listbox"` also has `id="some-custom-id"`. You should also be able to see the domId of the active option logged in the console for testing.

<details>

```jsx
import React from 'react';

import {Page, Listbox} from '../src';

  function logActiveId(_: string, activeId: string) {
    console.log(activeId);
  }

  return (
    <Page title="Playground">
      <Listbox
        accessibilityLabel="Basic Listbox example"
        customListId="some-custom-id"
        onActiveOptionChange={logActiveId}
      >
        <Listbox.Option value="UniqueValue-1">Item 1</Listbox.Option>
        <Listbox.Option value="UniqueValue-2">Item 2</Listbox.Option>
        <Listbox.Option value="UniqueValue-3">Item 3</Listbox.Option>
      </Listbox>
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/polaris-react/UNRELEASED.md
+++ b/polaris-react/UNRELEASED.md
@@ -13,6 +13,7 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 - Added breakpoint CSS custom properties and SCSS media conditions ([#5558](https://github.com/Shopify/polaris/pull/5558))
 - Replaced `768px` breakpoint mixins and variables with custom media conditions ([#5629](https://github.com/Shopify/polaris/pull/5629))
 - Added `customListId` prop to `Listbox` ([5627](https://github.com/Shopify/polaris/pull/5627))
+- Pass `domId` as an argument to `onActiveOptionChange` prop on `Listbox` ([5627](https://github.com/Shopify/polaris/pull/5627))
 
 ### Bug fixes
 
--- a/polaris-react/UNRELEASED.md
+++ b/polaris-react/UNRELEASED.md
@@ -12,6 +12,7 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 - Added the ability to disable specific dates in the `DatePicker`, to go along with date ranges ([#5356](https://github.com/Shopify/polaris/pull/5356))
 - Added breakpoint CSS custom properties and SCSS media conditions ([#5558](https://github.com/Shopify/polaris/pull/5558))
 - Replaced `768px` breakpoint mixins and variables with custom media conditions ([#5629](https://github.com/Shopify/polaris/pull/5629))
+- Added `customListId` prop to `Listbox` ([5627](https://github.com/Shopify/polaris/pull/5627))
 
 ### Bug fixes
 
--- a/polaris-react/src/components/Listbox/Listbox.tsx
+++ b/polaris-react/src/components/Listbox/Listbox.tsx
@@ -40,6 +40,8 @@ export interface ListboxProps {
   enableKeyboardControl?: boolean;
   /** Visually hidden text for screen readers */
   accessibilityLabel?: string;
+  /** Provide a custom ID for the list element */
+  customListId?: string;
   /** Callback fired when an option is selected */
   onSelect?(value: string): void;
   /** Callback fired when an option becomes active */
@@ -57,6 +59,7 @@ export function Listbox({
   children,
   enableKeyboardControl,
   accessibilityLabel,
+  customListId,
   onSelect,
   onActiveOptionChange,
 }: ListboxProps) {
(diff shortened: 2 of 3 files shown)


chloerice (commented):
Hey @andrewmcgov 👋🏽 This prop won't fill the gap of the other information `Listbox` controls that LocationSelector or any other searchable listbox needs. You'll likely need to come back to add additional props in order for the component to function properly if this route is continued. `LocationSelector` itself is a searchable listbox within a dropdown popover, it is not a combobox. 

> A [combobox](https://www.w3.org/TR/wai-aria-practices-1.1/#combobox) is a widget made up of the combination of two distinct elements: 1) a single-line textbox, and 2) an associated pop-up element for helping users set the value of the textbox.

As you've called out here, you can set all of the aria attributes needed to wire up the text field to the listbox with the exception of the listbox's `id`, which is set via context inside of `Listbox`.

The `Listbox` and `TextField` contexts are currently not exported for use outside of the package internals, but they should be. Both the `Listbox` and `TextField` context are misnamed as if they're owned by `Combobox`, when in reality these two components are entirely separated and `Combobox` is only consumer of them.

The contexts and hooks for both should be renamed (e.g., ListboxContext and useListbox, consumed and extended by `Combobox` (adding `expanded` to its extending type), and exported for consumption so that `LocationSelector ` can use it. This approach would scale the impact of your work to the other currently inaccessible searchable listboxes in `web` (like `ProductTypeCardBrowser`) that could also eventually consume the listbox context/hooks and align with the pattern you're introducing so their markup is no longer invalid.

andrewmcgov:
Hey @chloerice, thanks for the quick review!

I agree this approach is not the ideal long term solution. My team is hoping to start our rollout next week so we are trying to get something working here. But I understand we want to make the right thing. 

I also agree that the LocationSelector does not fit the strict definition of what a `combobox` is, but I've done a fair bit of searching I have not found any spec or best practice that describes what a `searchable listbox` is. The [listbox documentation](https://www.w3.org/TR/wai-aria-practices-1.1/#Listbox) has no mention of a searchable option. I chatted with @svinkle last week and we felt that the combobox role best fits what is happening in this component. Is there a different `role` you think we should provide to the textfield in our component?

> The contexts and hooks for both should be renamed (e.g., ListboxContext and useListbox, consumed and extended by Combobox (adding expanded to its extending type), and exported for consumption so that LocationSelector can use it.

This seems like an ideal long term fix. I'll take a look at it this afternoon. Also open to a quick huddle anytime this afternoon to come up with a plan if you are around.

andrewmcgov:
Chatted with @chloerice yesterday and we agreed that while separating the listbox and texfield contexts from Combobox is the right long term path, it is a large chunk of work that is out of scope for our project. 

We decided that this approach with allowing the parent to set the listbox ID the best path forward for now until more effort can be dedicated to decoupling the combobox contexts.

polaris-react/UNRELEASED.md
```diff
@@ -12,6 +12,7 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 - Added the ability to disable specific dates in the `DatePicker`, to go along with date ranges ([#5356](https://github.com/Shopify/polaris/pull/5356))
 
 - Added breakpoint CSS custom properties and SCSS media conditions ([#5558](https://github.com/Shopify/polaris/pull/5558))
+- Added customListId prop to Listbox ([5627](https://github.com/Shopify/polaris/pull/5627))
```
kyledurand:
```suggestion
- Added `customListId` prop to `Listbox` ([5627](https://github.com/Shopify/polaris/pull/5627))
```

kyledurand (approved):
I'm fine with this as long as @chloerice is 👍

kyledurand (approved):
🐑 🇮🇹

PR #5630, opened 2022-04-26
rm _skeleton.scss

alex-page:
**What does this PR do?**

- [x] Replace `position: absolute` from `skeleton-content` with `display: flex`
- [x] Replace `thumbnail-size` with hard coded values
- [x] Move `background-color` `border-radius` into skeleton components

**Why?**

My thought here is to remove as much abstraction and get values as close to where developers are working. We could explore grouping duplicate values at the React component layer.


polaris-react/src/components/SkeletonPage/SkeletonPage.scss
```diff
@@ -55,9 +55,16 @@ $skeleton-display-text-max-width: 120px;
 }
 
 .SkeletonTitle {
-  @include skeleton-content;
+  display: flex;
+  background-color: var(--p-surface-neutral);
+  border-radius: var(--p-border-radius-base);
+
```
alex-page:
```suggestion
```

polaris-react/UNRELEASED.md
```diff
@@ -10,8 +10,8 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 - Added `border-width-4` and `border-width-5` tokens and replaced hardcoded values ([#5528](https://github.com/Shopify/polaris/pull/5528))
 - Replaced any hardcoded `outline-width` with `border-width` ([#5528](https://github.com/Shopify/polaris/pull/5528))
 - Added the ability to disable specific dates in the `DatePicker`, to go along with date ranges ([#5356](https://github.com/Shopify/polaris/pull/5356))
-
 - Added breakpoint CSS custom properties and SCSS media conditions ([#5558](https://github.com/Shopify/polaris/pull/5558))
+- Replaced `skeleton-content` and `thumbnail-size` mixins with css values ([#5630](https://github.com/Shopify/polaris/pull/5630))
```
kyledurand: Maybe this would be better under code quality

kyledurand (approved):

polaris-react/src/components/SkeletonThumbnail/SkeletonThumbnail.scss
```diff
@@ -1,22 +1,26 @@
...
+  border-radius: var(--p-border-radius-base);
+
+  @media screen and (-ms-high-contrast: active) {
+    background-color: grayText;
+  }
 }
 
 .sizeSmall {
-  @include skeleton-thumbnail-size(small);
+  height: 40px;
```
aveline, line 14: The thing that concerns me about hard coding these is we lose the link between thumbnail component and that the values need to match. Maybe a shared custom property?

aveline (approved):

polaris-react/src/components/SkeletonThumbnail/SkeletonThumbnail.scss
alex-page: Yes. You are totally right. I was thinking of creating a `ThumnailContainer` component but it just didn't seem necessary. I know the coverage tool will pick up on these values. My only concern is losing the connection between `SkeletonThumbnail` and `Thumbnail`.

PR #5631, opened 2022-04-26
Fix build and TypeScript errors after porting token scripts

aaronccasanova:


romellogoodman:
merged in main because https://github.com/Shopify/polaris/pull/5628 took care of some of the type errors we were seeing with the site

PR #5649, opened 2022-04-27
Use next/link for components filter

### LINKED ISSUE #5636: [Polaris website refresh] Use next.js Link component for Component filters
Right now, they're using `router.push()` which means the pages aren't preloaded.

(This is a bit of a head scratcher: On the one hand, preloading full pages is a bit silly, since we're just doing a `components.filter()` to narrow the components index down. However, we also want proper URLs for each component category, so maybe full pages is the best option after all.)

martenbjork:
Fixes #5636

### DIFF
--- a/polaris.shopify.com/src/components/Button/Button.tsx
+++ b/polaris.shopify.com/src/components/Button/Button.tsx
@@ -10,7 +10,7 @@ interface Props {
 interface ButtonProps extends Props, HTMLProps<HTMLButtonElement> {}
 interface LinkButtonProps extends Props, PropsWithChildren<LinkProps> {}
 
-function Button({ small, pill, children, ...rest }: NativeButtonProps) {
+function Button({ small, pill, children, ...rest }: ButtonProps) {
   return (
     <button
       className={[
--- a/polaris.shopify.com/src/components/Button/Button.tsx
+++ b/polaris.shopify.com/src/components/Button/Button.tsx
@@ -1,5 +1,6 @@
 import Link, { LinkProps } from "next/link";
 import { HTMLProps, PropsWithChildren } from "react";
+import { className } from "../../utils/various";
 import styles from "./Button.module.scss";
 
 interface Props {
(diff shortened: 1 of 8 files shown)


polaris.shopify.com/src/components/Button/Button.tsx
```diff
@@ -1,13 +1,16 @@
...
 
-interface Props extends HTMLProps<HTMLButtonElement> {
+interface Props {
   small?: boolean;
   pill?: boolean;
-  children: React.ReactNode;
 }
 
-function Button({ small, pill, children, ...rest }: Props) {
+interface NativeButtonProps extends Props, HTMLProps<HTMLButtonElement> {}
```
aaronccasanova:
```suggestion
interface ButtonProps extends Props, HTMLProps<HTMLButtonElement> {}
```

Light suggestion to keep prop interfaces consistent with the associated component.

martenbjork: You mean the name of the interface? (Definitely agree!)

polaris.shopify.com/src/components/Button/Button.tsx
```diff
@@ -23,4 +26,27 @@ function Button({ small, pill, children, ...rest }: Props) {
...
+  ...rest
+}: LinkButtonProps) {
+  return (
+    <Link href={href} passHref>
+      <a
+        className={[
+          styles.Button,
+          small ? styles.small : null,
+          pill ? styles.pill : null,
+        ].join(" ")}
```
aaronccasanova:
```suggestion
        className={[
          styles.Button,
          small ? styles.small : null,
          pill ? styles.pill : null,
        ].filter(Boolean).join(" ")}
```

**suggestion (non-blocking):** Not a show stopper but we should probably filter out the `null` values
image

aaronccasanova (approved):

polaris.shopify.com/src/components/Button/Button.tsx
```diff
@@ -1,13 +1,16 @@
...
   small?: boolean;
   pill?: boolean;
-  children: React.ReactNode;
 }
 
-function Button({ small, pill, children, ...rest }: Props) {
+interface ButtonProps extends Props, HTMLProps<HTMLButtonElement> {}
+interface LinkButtonProps extends Props, PropsWithChildren<LinkProps> {}
+
+function Button({ small, pill, children, ...rest }: NativeButtonProps) {
```
aaronccasanova:
```suggestion
function Button({ small, pill, children, ...rest }: ButtonProps) {
```

polaris.shopify.com/src/components/Button/Button.tsx
martenbjork: Good call! I introduced a `className` helper function to do this automatically.

PR #5661, opened 2022-04-28
Update yarn.lock and newly caught linting, type check, a11y, and component errors

sam-b-rose:

This is a full-sweep effort to update our package dependencies using `yarn upgrade` to bump to the latest minors. This is to address any security vulnerabilities noted in the [Dependabot alerts](https://github.com/Shopify/polaris/security/dependabot) 🤖 ⚠️ 

**yarn.lock**

- Update yarn.lock to include latest minor version of our npm dependencies
- Updated `svgo` to address a particular sub-dependency security vulnerability

**eslint**

- Turn off the [`name-convention`](https://typescript-eslint.io/rules/naming-convention) rule due to conflicts with `aria-*` attributes in polaris-react and naming conflicts in polaris-tokens (ex: `z-1`)
- Remove unnecessary rule disables
- 

**stylelint**

- Turn on `camelCaseSvgKeywords` option for the [`value-keyword-case` rule](https://stylelint.io/user-guide/rules/list/value-keyword-case/)

**Type Check**

- Update JSDoc typings for `ruleMessage`
- Strict cast variables to `string`

**Accessibility**

- Replace incorrect usage of `aria-expanded` with `aria-hidden` on `Collapsible` component (was encountering [this error](https://dequeuniversity.com/rules/axe/4.4/aria-allowed-attr?application=axe-puppeteer) and [this usage solution](https://www.w3.org/WAI/GL/wiki/Using_the_WAI-ARIA_aria-expanded_state_to_mark_expandable_and_collapsible_regions) seemed to be out-of-scope for the Collapsible component alone).


.github/workflows/ci-polaris-react.yml
```diff
@@ -87,6 +88,7 @@ jobs:
             ${{ runner.os }}-node${{ matrix.node-version }}-visual-regression-test-v1-
 
       - run: yarn --frozen-lockfile
+      - run: yarn build
```
sam-b-rose: I don't like having to build the packages before Storybook. This is where smart build pipeline would come in handy.

Rather than building ALL packages, we should defined only the packages that need to be built prior to building other assets like storybook

.github/workflows/ci-polaris-react.yml
```diff
@@ -60,6 +60,7 @@ jobs:
             ${{ runner.os }}-node${{ matrix.node-version }}-accessibility-test-v1-
 
       - run: yarn --frozen-lockfile
+      - run: yarn build
```
sam-b-rose: These build steps were needed in order for these checks to pass. Storybook was failing to build due to now referencing the local `polaris-icons` package locally using Yarn workspaces

polaris-react/src/components/CustomProperties/CustomProperties.tsx
```diff
@@ -37,7 +38,7 @@ export function CustomProperties(props: CustomPropertiesProps) {
       <Component
         p-color-scheme={colorScheme}
         className={className}
-        style={{color: 'var(--p-text)'}}
+        style={{color: 'var(--p-text)', ...style}}
```
sam-b-rose, line 41: 👀  @aaronccasanova does this passing of the `style` object seem appropriate?

romellogoodman (approved):

polaris-react/src/components/CustomProperties/CustomProperties.tsx
aaronccasanova: Yes thank you for catching that

polaris-react/UNRELEASED.md
```diff
@@ -14,6 +14,10 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 
 ### Bug fixes
 
+- Used prop-provided `selectable` value in `IndexTable` ([#5661](https://github.com/Shopify/polaris/pull/5661))
+- Fixed passing inline styles to the root element of the `CustomProperties` component ([#5661](https://github.com/Shopify/polaris/pull/5661))
+- Replace incorrect usage of `aria-expanded` on `Collapsible` with `aria-hidden` ([#5661](https://github.com/Shopify/polaris/pull/5661))
```
aaronccasanova:
```suggestion
- Replaced incorrect usage of `aria-expanded` on `Collapsible` with `aria-hidden` ([#5661](https://github.com/Shopify/polaris/pull/5661))
```

polaris-react/src/components/CustomProperties/CustomProperties.tsx
```diff
@@ -25,6 +25,7 @@ export function CustomProperties(props: CustomPropertiesProps) {
     children,
     className,
     colorScheme = DEFAULT_COLOR_SCHEME,
+    style = {},
```
aaronccasanova:
```suggestion
    style,
```
**suggestion (non-blocking):** We don't need a fallback:
image

polaris-react/src/components/IndexTable/IndexTable.tsx
```diff
@@ -67,14 +67,15 @@ function IndexTableBase({
   emptyState,
   sort,
   lastColumnSticky = false,
+  ...props
```
aaronccasanova:
```suggestion
  ...restProps
```
**nit (non-blocking):** Prefer this indicates it's a subset of the props

polaris-react/src/components/IndexTable/IndexTable.tsx
```diff
@@ -67,14 +67,15 @@ function IndexTableBase({
...
 }: IndexTableBaseProps) {
   const {
     loading,
     bulkSelectState,
     resourceName,
     bulkActionsAccessibilityLabel,
-    selectable,
     selectMode,
+    selectable = props.selectable,
```
aaronccasanova:
```suggestion
    selectable = restProps.selectable,
```
(non-blocking) If moving forward with the above suggestion

.eslintrc
```diff
@@ -41,7 +41,8 @@
     "@shopify/jsx-prefer-fragment-wrappers": "off",
     "@shopify/no-ancestor-directory-import": "error",
     "@shopify/react-prefer-private-members": "off",
-    "@typescript-eslint/array-type": ["error", {"defaultOption": "array"}],
+    "@typescript-eslint/array-type": ["error", {"default": "array"}],
+    "@typescript-eslint/naming-convention": "off",
```
aaronccasanova, line 45: **question (non-blocking):** Should we configure this rule? Seems helpful to enforce naming conventions.

aaronccasanova (approved):

.eslintrc
sam-b-rose: I did try to keep this rule by modifying it so that `kebab-case` was valid. However, once you add your own options for it, the rule requires additional `projectOptions.project` details that was causing a lot of added complexity (requires pre-typescript-build and evaluation of the project). When adding in those details to the eslint config, it then came back with a lot more issues :(

Think it is worth adding back at some point? Maybe a follow up issue?

aaronccasanova: Oof! Yeah, I think low priority follow up issue 👍

polaris-react/src/components/IndexTable/IndexTable.tsx
sam-b-rose: Do we have this convention documented anywhere? I've seen both conventions used in `polaris-react` but mostly `...props`. I'm indifferent, just crave consistency 😄

.eslintrc
sam-b-rose: Created a follow up issue here https://github.com/Shopify/polaris/issues/5677 🎟️

PR #5671, opened 2022-04-29
[Polaris website refresh] Read icons from polaris-icons

martenbjork:
- Reads metadata from `polaris-icons`
- Copies latest icons from `polaris-icons` into next.js's `public` folder

The duplication between `polaris-icons` and next.js is a bit stupid, but not a big issue IMO. The script will ensure that the directories are 1:1. The big benefit of this method is that it allows us to use next/image with lazy loading which is needed on the icons page (so that the 400 icons aren't embedded as SVGs).


alex-page:
I think there is a better way...

alex-page (approved):

alex-page:
Happy to unblock this and ship.

End goal for me is no shenanigan's when adding icons. Adding an icon to `polaris-icons` should update the site without anything extra needing to happen. I also would love to not have icons across multiple repos as the download size of the `shopify/polaris` will grow pretty big.

sam-b-rose (approved):
Also not super thrilled about this icon strategy. Happy to approve to unblock as well, but I think we should consider other patterns for icons. 

Even though we get the imae lazy-loading benefits with icons using this method, I'd rather not duplicate content and am uncertain lazy-loading performance will be a major boost over just inlining SVG for something as small as icons (even though there could be 400+). I'd rather render a grid and load & render them after page load. 

More scripts, more duplicated files 😬

polaris.shopify.com/src/components/IconGallery/IconGallery.tsx
```diff
@@ -7,6 +7,11 @@ import styles from "./IconGallery.module.scss";
 import Longform from "../Longform";
 import { Tab } from "@headlessui/react";
 import { className } from "../../utils/various";
+const importedSvgs = require.context(
+  "../../../../polaris-icons/icons",
+  true,
+  /\.svg$/
+);
```
martenbjork, line 14: https://stackoverflow.com/a/46210355

martenbjork:
All right, good points all around! I've updated the logic so that it uses inline svgs 🙏 You're right, this is much better.

polaris.shopify.com/src/data/icons.json
```diff
@@ -3,8 +3,14 @@
     "name": "abandoned cart",
     "set": "major",
```
alex-page, line 4: Random thought. There is a script in icons that generates a JSON file. I ended up deleting it but we could move it to this site.
https://github.com/Shopify/polaris-icons/blob/main/packages/polaris-icons/scripts/metadata-json.js

alex-page (approved):
TY @martenbjork. This works for me.

sam-b-rose:
Likewise, much cleaner solution ✨  ty @martenbjork 🚀

polaris.shopify.com/src/data/icons.json
martenbjork: My script does exactly the same thing, so we should remove mine, so all the icons logic lives in polaris-icons.

If the script runs in `polaris.shopify.com`, the script needs to be manually run to update the icons page which is what you wanted to avoid. It would be better if polaris-icons handled all the logic and exposed the meta json. (Riiight?)

PR #5676, opened 2022-04-29
Create Releasing.md

romellogoodman:
Adds documentation on the current release process for each application


sam-b-rose (approved):
It's a start! Thanks for documenting a high-level overview of how each package is released. 

Good info to have when forming a strategy around what improved approach we want to take next 👍

PR #5685, opened 2022-05-02
Added min-height to Popover.Pane

dleroux:

To accommodate the new `TagAutocomplete` in web, a `height` prop was added to `Combobox`. This height prop gets passed down to the `maxHeight` and `height` of the `Popover.Pane`. However because the popover adds inline styles, a `minHeight` is also needed to prevent the popover from resizing based on viewport space. Without it, the `TagAutocomplete` overlay moved between the top and bottom of the Textfield based on its content which wasn't a great UX.

Adds `minHeight` to the `Popover.Pane` based on the height prop. The same approach used for the `maxHeight`.

### How to 🎩

I added a height tot the Single select autocomplete Combobox example here:
spin: http://240.16.151.141:6006/?path=/story/all-components-combobox--single-select-autocomplete

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist


documentation/Releasing.md
```diff
@@ -4,35 +4,35 @@ The process for releasing the monorepo's contents is in active development. This
 
 ### [polaris-for-vscode](/polaris-for-vscode)
```
dleroux: @romellogoodman, I'm not sure why your change here: https://github.com/Shopify/polaris/pull/5676 didn't fail the linter.

romellogoodman: Just pushed an update to main with these changes. Thanks for the catch

kyledurand (commented):
Should minHeight be a new prop?

dleroux:
> Should minHeight be a new prop?

I don't believe so, the same way maxHeight isn't. It's just to make sure the popover keep the height it was given.

dleroux:
@romellogoodman @kyledurand can I get a ✅

kyledurand (approved):

PR #5686, opened 2022-05-02
fix: Polaris tokens ESM interoperability

aaronccasanova:
This PR fixes and issue with ESM interoperability in  `@shopify/polaris-tokens`. While we were already exporting dual ESM/CJS builds, Node.js was unable to infer what module system to use from the output paths alone. Therefore, all `dist/esm/*` files have been updated to use `.mjs` extensions and `dist/cjs/*` files will continue to use `.js`

image

### DIFF
--- a/polaris-tokens/package.json
+++ b/polaris-tokens/package.json
@@ -3,12 +3,12 @@
   "version": "5.0.0",
   "description": "",
   "main": "dist/cjs/index.js",
-  "module": "dist/esm/index.js",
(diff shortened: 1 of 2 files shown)


romellogoodman (approved):
thanks for doing the detective work to hunt down the problems and solution!

sam-b-rose (approved):
Simple and clean ✨  nice solution!

BPScott (approved):

PR #5687, opened 2022-05-02
Fix color of `plain-button-backdrop` on colored backgrounds

### LINKED ISSUE #5674: [Button] Color of `plain-button-backdrop` on colored backgrounds
The `plain-button-backdrop` background color on `:active` and `:focus` on colored backgrounds looks incorrect:

- https://github.com/Shopify/web/pull/61917/files#r853645305
- #4968

We should remove this background color for now (✅ approved by design @sarahill) and will revisit these buttons as a whole in the future.

lgriffee:

Fixes #5674 

Removing the `plain-button-backdrop` mixin background color for now and will revisit these buttons as a whole in the future. 

<hr>

#### Note 🚨 
This change does essentially render the `plain-button-backdrop` mixin useless but it's safer than entirely removing the mixin for now until we revisit plain buttons. I added a comment to try and prevent any confusion.


dominikwilkowski (approved):
LGTM

PR #5692, opened 2022-05-02
Move Babel config to root level for better monorepo support

### LINKED ISSUE #5441: Move babel config to package.json "babel" key

sam-b-rose:

Fixes #5441 

Test and verify using `yarn build` to build all packages (including `polaris-react`, `polaris-tokens`, and `polaris-icons` which uses Babel).

* Create a shared root-level`./babel.config.js`
* Remove all other `babel.config.js` or `.babelrc` files in `polaris-react`, `polaris-icons`, and `polaris-tokens`
* Update Storybook babel config options (`polaris-react` only)
* Update Rollup configurations options
* Update Jest transform options (`polaris-tokens` only)
* Move Babel-related `devDependencies` to the root `package.json`

Co-authored: @BPScott, @aaronccasanova


polaris-react/.storybook/main.js
```diff
@@ -32,6 +32,7 @@ module.exports = {
           {
             loader: 'babel-loader',
             options: {
+              configFile: path.resolve(__dirname, '../../babel.config.js'),
```
aaronccasanova:
```suggestion
              configFile: path.join(__dirname, '../../babel.config.js'),
```

**suggestion (non-blocking):** `__dirname` is already an absolute path so no need to `.resolve()` one

polaris-react/tsconfig.json
```diff
@@ -21,5 +21,6 @@
     "./locales/**/*.json",
     "./tests",
     "./playground"
-  ]
+  ],
+  "references": [{"path": "../polaris-tokens"}]
```
aaronccasanova: **question:** Mind providing some insight on this addition?

polaris-react/.storybook/babel.config.js
```diff
@@ -1 +1 @@
-module.exports = require('../babel.config.js');
+module.exports = require('../../babel.config.js');
```
aaronccasanova: https://babeljs.io/docs/en/config-files#configuration-file-types

**question:** Should this file be renamed to `.babelrc.js`? Looks like the project wide config should be `babel.config.js` and sub projects should be `.baberc.js`.

polaris-tokens/rollup.config.js
```diff
@@ -34,6 +34,7 @@ const rollupOptions = {
     commonjs(),
     // Compile TypeScript/JavaScript files
     babel({
+      configFile: '../babel.config.js',
```
aaronccasanova: **question:** Do we need to `path.join(__dirname, '../babel.config.js')` here?

aaronccasanova (approved):
LGTM! For monorepos, I do slightly prefer using a [root level `babel.config.js` and individual `.babelrc.js` per package](https://github.com/parcel-bundler/parcel/tree/v2/packages/core/integration-tests/test/integration/babel-config-monorepo) with:
```js
// packages/<package>/.baberc.js
module.exports = {
  rootMode: 'upward',
  extends: '../babel.config.js',
  presets: [...],
  plugins: [...],
}
```

While it is a bit more verbose, I find it easier to understand the subpackage's infrastructure and overrides and have seen the alternative root `babel.config.js` `overrides: [{...}, {...}, {...}]` mechanism grow excessively large overtime.

Note: This is simply an observation and by no means blocking (:

polaris-react/.storybook/main.js
sam-b-rose: Good catch, ty!

BPScott:
https://babeljs.io/docs/en/config-files#root-babelconfigjson-file suggests that usage of a root babel.config.js and per-package babelrc files are complementary approaches rather than needing to do both. It also notes about using `rootMode:"upward"` as an alternative to specifying configPath.


https://babeljs.io/docs/en/config-files#subpackage-babelrcjson-files suggets we might also need to add a `babelrcRoots` option to the top level babel config.


I think I originally leaned slightly towards having a root babel.config, and using overrides if we ever need to customize on a per-package basis, rather than having per package configs as that seems [recommended](https://babeljs.io/docs/en/configuration#whats-your-use-case), but I'm happy to go either way. @aaronccasanova's rational of having configs next within each package where per-package overrides can happen makes sense, though I'm not aware of an example where that would need to occur in this repo currently.

aaronccasanova:
> though I'm not aware of an example where that would need to occur in this repo currently

This is key. We don't currently have that problem and it would likely be years before actually becoming an issue. I'm fine with the original approach using a single `babel.configs.js` at the root and applying `overrides` when needed.

As @BPScott mentioned, we should still set `babelrcRoots` and `rootMode: 'upward'` respectively.

PR #5696, opened 2022-05-03
Configure babelrcRoots and update sub-packages

aaronccasanova:


polaris-icons/.babelrc.js
```diff
@@ -1,3 +1,6 @@
+/**
+ * @type {import('@babel/core').TransformOptions}
+ */
 module.exports = {
-  extends: '../babel.config.js',
+  presets: [['@shopify/babel-preset', {react: true}]],
```
BPScott, line 5: I'd say leave this as it is. Enabling the typescript transforms in this package has no effect - as there are no ts/tsx files. I don't think this is worth the divergence from the base config.

polaris-react/.babelrc.js
```diff
@@ -1,3 +1,6 @@
+/**
+ * @type {import('@babel/core').TransformOptions}
+ */
 module.exports = {
-  extends: '../babel.config.js',
+  presets: [['@shopify/babel-preset', {typescript: true, react: true}]],
```
BPScott, line 5: I'd say leave this as it is. Enabling the react transforms in this package has no effect - as there are no jsx usages. I don't think this is worth the divergence from the base config.

BPScott (commented):
I don't think it's worth disabling typescript/react transforms at the root and then opting into them on a per-package basis. Just because a package doesn't use react/typescript I don't think there's any harm in having those transforms enabled as it means we have one config for the whole monorepo rather than lots of per-package configs.

I think we should keep `[['@shopify/babel-preset', {typescript: true, react: true}]]` in the root config, and have the per-package configs all use rootMode to leverage that.

aaronccasanova:
> I think we should keep [['@shopify/babel-preset', {typescript: true, react: true}]] in the root config, and have the per-package configs all use rootMode to leverage that.

No objections from me 👍

sam-b-rose (commented):
I really like this strategy! However, `rootMode: "upward"` is only specified in the `rollup.config.js` files. My only concern/thought is if we change the base `babel.config.js`, those changes won't be applied to the `.babelrc.js` for non-Rollup processes (Jest and Storybook). We could add those babel options to those tools as well, but again seems like it is just spreading config to other areas 🤷 

This might not be an issue since we may change how our tests and storybook are run in the future, but in its current state it feels disconnected. Things still work though because we really don't change anything in our per-package `.babelrc.js` files other than either flipping on typescript or react.

Anyway, I'm fine with this configuration, but I'm also not sure what advantages it provides over the single root file. And it splits the babel config to the Rollup-specific plugin

sam-b-rose (approved):

PR #5698, opened 2022-05-03
Add single CI for build, test, and linting

### LINKED ISSUE #5684: Clean up CI tasks
- [x] Use Turborepo to handle smart caching of building, linting, and testing packages
- [x] Remove redundant per-package CI workflows

sam-b-rose:

Fixes #5684 

Our CI is currently split up into multiple workflows by package. This provides a lot of control over each workflow per-package, but also creates a lot of workflow maintenance overhead.

Now that we have Turborepo in place, we can better cache and orchestraight our scripts. Build, lint, and type output is cached for each package, speeding up the time to it takes to check redundant packages that aren't affected by the PR.

- Create a single CI workflow that "Builds, Lints, and Tests" all packages (with caching) for each Node version [14, 16].
- Remove all redundant single-package workflows used for build, lint, and test validation


.github/workflows/ci.yml
```diff
@@ -0,0 +1,48 @@
+name: 'Validate'
```
BPScott: Silly nit If we're calling the filename `ci`, can we have this as `CI`?

sam-b-rose: Yes, this was a just a test filename and workflow name. Will update to a more official and descriptive name once I can "validate" the cache is working 😅 just testing it out for now

.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+on:
+  push:
+    branches:
+      - main
+      - polaris-release
+  pull_request:
+
+jobs:
+  build:
+    name: 'Build, Lint, & Test: node-${{ matrix.node_version }}'
```
aaronccasanova:
```suggestion
    name: 'Build, Lint, and Test: node-${{ matrix.node_version }}'
```

In the spirit of calling out silly nits 🤓

.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+        uses: actions/setup-node@v3
+        with:
+          node_version: ${{ matrix.node_version }}
+          cache: yarn
+
+      - name: Restore cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            *.tsbuildinfo
```
aaronccasanova:
```suggestion
            **/*.tsbuildinfo
```

I believe we need to recursively cache these `tsbuildinfo` files:
image

.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+            node_modules/.cache/turbo
```
aaronccasanova, line 31: **suggestion:** Add `**/.turbo`

Might as well cache the sub-package `.turbo` directories too:
image

.github/workflows/ci.yml
sam-b-rose: Good catch, yes. Ty!

.github/workflows/ci.yml
sam-b-rose: Good `find` 👀  👍

.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
+name: CI
+
+on:
+  push:
+    branches: [main, polaris-release]
+  pull_request:
+    types: [opened, synchronize]
```
aaronccasanova:
```suggestion
```

**suggestion (non-blocking):** The default activity types for the `pull_request` trigger are `open`, `synchronize`, and `reopened`:
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+          node_version: ${{ matrix.node_version }}
+          cache: yarn
+
+      - name: Restore cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            **/*.tsbuildinfo
+            **/.turbo
+            node_modules/.cache/turbo
```
aaronccasanova: I'm noticing some of the removed workflows cached `.eslintcache`. Should we add that here?

aaronccasanova: Additionally, we likely need to port the `polaris-react/.loom` cache (removed from the `ci-polaris-react` workflow)

aaronccasanova (commented):
All the updates are looking great! Dropping some more feedback! Just waiting to approve once the `shipit.*.yml` files are updated.

.github/workflows/ci.yml
sam-b-rose: Cool cool, didn't know this! Fixed 🔧

.github/workflows/ci.yml
```diff
@@ -0,0 +1,49 @@
+name: CI
+
+on:
+  push:
+    branches: [main, polaris-release]
+  pull_request:
+
+jobs:
+  build:
+    name: 'Build, Lint, and Validate with Node v${{ matrix.node_version }}'
```
BPScott:
```suggestion
    name: 'Validate with Node v${{ matrix.node_version }}'
```

To match the names in the shipit files

.github/workflows/ci.yml
sam-b-rose: I don't think we'll need either since we are performing caching using Turborepo. It will cache the scrip output for both `lint` and `test` commands 👍

.github/workflows/ci.yml
sam-b-rose: Good catch!

.github/workflows/ci.yml
aaronccasanova: From the [Turbo docs](https://turborepo.org/docs/core-concepts/caching#configuring-cache-outputs):

> If your task does not emit any files (e.g. unit tests with Jest), set outputs to an empty array (i.e. []). Turborepo will automagically cache the logs for you.

From what I understand, the [empty array](https://github.com/Shopify/polaris/blob/3a146996b62425393e9350394c4bc7ad0bc3f281/turbo.json#L14) only caches the logs sent to stdout so Turbo can replay the logs if there is a cache hit. I believe we'd still want to cache the [.eslintcache](https://eslint.org/docs/user-guide/command-line-interface#--cache) which allows Eslint to perform an optimized re-run of the process (either here in the workflow or in `turbo.json`).

sam-b-rose: Added restoration of test and lint cache 🔧

.github/workflows/ci-polaris-react.yml
```diff
@@ -10,36 +10,6 @@ on:
       - 'polaris-react/**'
```
alex-page:
```suggestion
      - 'polaris-react/**'
```
Should this run without the filter? The file could also be renamed?!

sam-b-rose: Probably so, since a change in other packages might cause a change in the Viz and A11y tests. Good callout

alex-page (approved):

.github/workflows/ci-polaris-react.yml
BPScott: At this point, where there's no filtering here and the `on` logic is the same as ci.yml, could these be jobs in the ci.yml file?

shipit.polaris-react.yml
```diff
@@ -16,4 +16,5 @@ dependencies:
 
 deploy:
   override:
+    - yarn turbo run build --filter=@shopify/polaris
```
BPScott, line 19: I'm assuming turbo.json has been taught that tokens needs to be built before polaris

BPScott (approved):
Looking pretty good to me.

Dropped a note inline pondering that if `ci-polaris-react.yml` and `ci.yml`'s `on` triggers shall be the same then could we make the contents of `ci-polaris-react.yml` be steps in `ci.yml` instead of a separate file.

I'd be tempted to update the eslint commands to run over the current folders - `TIMING=1 eslint --cache .` rather than specific subfolders that'll give a bit more consistency across projects.

shipit.polaris-react.yml
sam-b-rose: Yes, that is correct! Which makes this much better than just using the `yarn workspace [package] [script]` command since that doesn't execute builds for topological dependencies

.github/workflows/ci-polaris-react.yml
sam-b-rose: Decided to keep the filter and CI separate since jobs can't be filtered based on path. Only full workflows can be triggered conditionally by path.

aaronccasanova (approved):

.github/workflows/polaris-react-ci.yml
```diff
@@ -0,0 +1,94 @@
+name: '@shopify/polaris CI'
+
+on:
+  push:
+    branches: [main, polaris-release]
+    paths:
+      - 'polaris-react/**'
+  pull_request:
+    paths:
+      - 'polaris-react/**'
```
alex-page:
```suggestion
      - 'polaris-react/src/**'
      - 'polaris-tokens/src/**'
  pull_request:
    paths:
      - 'polaris-react/src/**'
      - 'polaris-tokens/src/**'
```

@bpscott made a good point that tokens would also benefit from running these actions.

alex-page: @samrose3 If we want this maybe it should be called `ci-a11y-vrt.yml`

PR #5700, opened 2022-05-04
[DataTable] BUG: Whitespace when scrolling in sticky header in DataTable

blboyle:

The issue was that we were combining both scrolling and translating for the sticky header and this caused the content to go off screen with whitespace appearing. It also caused the table to do the same.

Before: https://screenshot.click/02-11-666cm-n5493.mp4

Changed so that the sticky header scrolls when the table is scrolled. But, the scrolling the sticky header does not scroll the table.

After: https://screenshot.click/04-16-qt55s-z6b1q.mp4

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist


polaris-react/src/components/DataTable/DataTable.tsx
```diff
@@ -477,13 +501,6 @@ class DataTableInner extends PureComponent<CombinedProps, DataTableState> {
             ? currentColumn.rightEdge
             : previousColumn.leftEdge;
 
-        if (stickyTableHeadingsRow) {
```
blboyle, line 480: Removed this as we don't need to specific the moving of the sticky heading, the table will do it for us.

**Before**: 
Navigation button is clicked 
-> JS scrolls `stickyTableHeadingsRow`

**After**: 
Navigation button is clicked 
-> JS scrolls `scrollContainer` 
-> `scrollContainer` scrolls `stickyTableHeadingsRow` (because of `private stickyHeaderScrolling`)

polaris-react/src/components/DataTable/DataTable.tsx
```diff
@@ -425,46 +433,62 @@ class DataTableInner extends PureComponent<CombinedProps, DataTableState> {
     };
   };
 
-  private stickyHeaderScrolling = (event: Event) => {
+  private focusListener = (event: Event) => {
```
blboyle, line 436: This is used in order to allow for a keyboard use, who is tabbing through the heading buttons of the sticky header, to have the headings come into view if they are outside of view.

polaris-react/src/components/DataTable/DataTable.tsx
```diff
@@ -425,46 +433,62 @@ class DataTableInner extends PureComponent<CombinedProps, DataTableState> {
...
+    this.stickyHeaderFocusStateChangeOnScroll(event);
+  }
+
+  private stickyHeaderFocusStateChangeOnScroll = (event: Event) => {
```
blboyle: This is used in order to have any of the buttons inside of any of the headings of the sticky header to come into view if a user is tabbing through them.

polaris-react/src/components/DataTable/DataTable.tsx
```diff
@@ -340,6 +344,10 @@ class DataTableInner extends PureComponent<CombinedProps, DataTableState> {
 
     if (inStickyHeader) {
       this.stickyHeadings[index] = ref;
+
+      const button = ref.querySelector('button');
```
blboyle, line 337: Adding focus event listener to only the headings inside of sticky header. When tabbing through the regular table headings, normal keyboard tabbing applies.

polaris-react/src/components/DataTable/DataTable.tsx
```diff
@@ -208,7 +209,10 @@ class DataTableInner extends PureComponent<CombinedProps, DataTableState> {
         <div className={styles.StickyTable} role="presentation">
           <Sticky
             boundingElement={this.dataTable.current}
-            onStickyChange={this.changeHeadingFocus}
+            onStickyChange={(isSticky) => {
+              this.changeHeadingFocus();
+              this.stickyHeaderActive = isSticky;
```
blboyle, line 214: So that `DataTable` knows if the sticky header is active, or not. 

This is used to disable the scrolling feature when not sticky.

polaris-react/src/components/DataTable/DataTable.tsx
```diff
@@ -425,46 +433,62 @@ class DataTableInner extends PureComponent<CombinedProps, DataTableState> {
...
-      scrollContainer == null ||
-      table == null
-    ) {
+    if (stickyTableHeadingsRow == null || scrollContainer == null) {
       return;
     }
 
-    const {scrollLeft: stickyScrollLeft} = stickyTableHeadingsRow;
-    const {scrollLeft: scrollContainerScrollLeft} = scrollContainer;
+    stickyTableHeadingsRow.scrollLeft = scrollContainer.scrollLeft;
```
blboyle, line 462: Chose to only have the table scrolling affect the sticky header.
Removed the ability to scroll on sticky header and also having it affect the table.

polaris-react/src/components/DataTable/DataTable.tsx
```diff
@@ -425,46 +433,62 @@ class DataTableInner extends PureComponent<CombinedProps, DataTableState> {
...
-  private scrollListener = (event: Event) => {
+  private scrollListener = () => {
     debounce(() => {
       this.setState((prevState) => ({
         ...this.calculateColumnVisibilityData(prevState.condensed),
       }));
     }, 500);
 
-    this.stickyHeaderScrolling(event);
+    if (this.props.stickyHeader && this.stickyHeaderActive) {
```
blboyle, line 472: Sticky header scrolling will only fire when the table _uses_ sticky header and also sticky header is active.

blboyle:
@JeremyLudwigDev 

You'll notice that I also have a focus tabbing handler in here. I think it'll be best to figure out how we consolidate ours together so that we don't do any duplication in this component.

polaris-react/src/components/DataTable/DataTable.tsx
```diff
@@ -425,46 +434,62 @@ class DataTableInner extends PureComponent<CombinedProps, DataTableState> {
     };
   };
 
-  private stickyHeaderScrolling = (event: Event) => {
+  private focusListener = (event: Event) => {
```
philschoefer: Unecessary indirection? WDYT about calling `stickyHeaderFocusStateChangeOnScroll` directly?

polaris-react/src/components/DataTable/DataTable.tsx
philschoefer: The name makes me think this is called by a `scroll` handler. I think updating the name to something like `stickyHeaderFocusListener` would make clearer when this function is called. What it does I think is pretty clear based on it's content.

blboyle: Ahh yes, makes sense.

polaris-react/src/components/DataTable/DataTable.tsx
```diff
@@ -425,46 +434,62 @@ class DataTableInner extends PureComponent<CombinedProps, DataTableState> {
...
-      table.style.transform = `translateX(-${stickyScrollLeft}px)`;
-    } else {
-      this.stickyHeadings.map((heading) => {
-        heading.style.transform = `translateX(-${scrollContainerScrollLeft}px)`;
-      });
-    }
   };
 
-  private scrollListener = (event: Event) => {
+  private scrollListener = () => {
```
philschoefer, line 465: I noticed we have a `handleResize` function. I would expect this to be `handleScroll` based on that naming pattern. Totally optional nit, happy to go with whatever you prefer.

philschoefer (approved):
LGTM! One thing I noticed: I don't think we have a storybook story with the new fixed header. I think it might be worth adding for visual regression testing and to make the feature more visible within the Polaris docs.

polaris-react/src/components/DataTable/DataTable.tsx
blboyle: True, I might be early-optimizing

alex-page (approved):

PR #5702, opened 2022-05-04
Add disable option to page header action groups

kemaltatar:

Enabling disable property to the action groups in the page header, see [this PR](https://github.com/Shopify/web/pull/63718/files#diff-c09cd3ac6d132bf92463d869e74307b8a241f21005cbd907f7330f3bb692bcb5R43) for context

Screenshot 2022-05-04 at 19 10 56

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/polaris-react/src/components/ActionMenu/ActionMenu.tsx
+++ b/polaris-react/src/components/ActionMenu/ActionMenu.tsx
@@ -62,6 +62,10 @@ export function hasGroupsWithActions(groups: ActionMenuProps['groups'] = []) {
 function convertGroupToSection({
   title,
   actions,
+  disabled,
 }: MenuGroupDescriptor): ActionListSection {
-  return {title, items: actions};
+  return {
+    title,
+    items: disabled ? [] : actions,
+  };
 }
--- a/polaris-react/src/components/Page/README.md
+++ b/polaris-react/src/components/Page/README.md
@@ -413,11 +413,16 @@ Use action groups for sets of actions that relate to one another, particularly w
   actionGroups={[
     {
       title: 'Promote',
+      disabled: true,
+      actions: [{content: 'Share on Facebook'}],
+    },
+    {
+      title: 'More actions',
       actions: [
-        {
-          content: 'Share on Facebook',
-          onAction: () => {},
-        },
+        {content: 'Duplicate'},
+        {content: 'Print'},
+        {content: 'Unarchive'},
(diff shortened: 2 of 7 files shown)


polaris-react/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -62,6 +62,10 @@ export function hasGroupsWithActions(groups: ActionMenuProps['groups'] = []) {
 function convertGroupToSection({
   title,
   actions,
+  disabled,
```
dingua, line 65: may be make it get a default value set to `false` to be backward compatible ? (assuming this is the public interface of the component)

polaris-react/src/components/Page/README.md
```diff
@@ -413,11 +413,16 @@ Use action groups for sets of actions that relate to one another, particularly w
   actionGroups={[
     {
       title: 'Promote',
+      disabled: true,
```
dingua, line 416: it makes more sense to be `false` :)

polaris-react/src/components/ActionMenu/ActionMenu.tsx
kemaltatar: default state for it is `undefined` which is a falsy value and is backward compatible. I don't think we need to explicitly assign it to `false` since this is an optional field

polaris-react/src/components/Page/README.md
kemaltatar: The idea here is to have an example of a disabled action group in the documentation

polaris-react/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -62,6 +62,10 @@ export function hasGroupsWithActions(groups: ActionMenuProps['groups'] = []) {
...
 }: MenuGroupDescriptor): ActionListSection {
-  return {title, items: actions};
+  return {
+    title,
+    items: disabled ? [] : actions,
```
kemaltatar, line 69: as another option we can do something like 

```
actions.map(action => ({...action, disabled: disabled || action.disabled}))
```

jorgenunezsiri (approved):
This LGTM! 😄

alex-page (approved):

PR #5720, opened 2022-05-05
Clean up tsconfigs to use shared @shopify/typescript-configs

### LINKED ISSUE #5682: Consolidate shared TypeScript configurations based on package type (base, nextjs, node14, react-library, etc)
- [x] Create base configs
- [x] Add `paths` for packages to base config
- [x] Add package dependency references to package tsconfigs

Examples:

- [TypeScript recipe using wireit](https://github.com/google/wireit#typescript)
- [Turborepo tsconfig examples](https://github.com/vercel/turborepo/tree/main/examples/design-system/packages/acme-tsconfig)
- [Breaking down TS Project Refs](https://jakeginnivan.medium.com/breaking-down-typescript-project-references-260f77b95913)

sam-b-rose:

Fixes #5682 

- Clean up TS configurations to use [`@shopify/typescript-configs`](https://github.com/Shopify/web-configs/tree/main/packages/typescript-configs)
- Removes redundant config options
- Fixes project references to package dependencies


BPScott:
These base configs look very similar to those provided by [`@shopify/typescript-configs`](https://github.com/Shopify/web-configs/tree/main/packages/typescript-configs). There's also the 
 [`@tsconfigs` packages](https://github.com/tsconfig/bases) provided by the typescript team. Could you expand on the rational for starting for rolling our own rather over leveraging these existing approaches.

sam-b-rose:
@BPScott great point. I've revised to leverage the @shopify/typescript-configs instead! Lmwyt! 👍

package.json
```diff
@@ -20,7 +20,7 @@
     "test": "turbo run test",
     "clean": "turbo run clean && rm -rf node_modules",
     "format": "prettier . --write",
-    "type-check": "tsc --build",
+    "type-check": "tsc --build --pretty",
```
aaronccasanova: Thoughts on adding `"pretty": true,` to the root `polaris/tsconfig.json`?

polaris-for-vscode/tsconfig.json
```diff
@@ -1,18 +1,8 @@
...
-    "declarationMap": true
-  },
-  "include": ["src"],
-  "exclude": ["node_modules", "dist"],
-  "references": [{"path": "./client"}, {"path": "./server"}]
+  "files": [],
+  "references": [
+    {"path": "./client"},
+    {"path": "./server"},
+    {"path": "../polaris-tokens"}
```
aaronccasanova: **question:** Can you help me understand why we include the `../polaris-tokens` reference?

polaris.shopify.com/tsconfig.json
```diff
@@ -1,30 +1,34 @@
...
+  "$schema": "https://json.schemastore.org/tsconfig",
+  "display": "Next.js",
   "compilerOptions": {
-    "rootDir": "../",
     "target": "es5",
     "lib": ["dom", "dom.iterable", "esnext"],
     "allowJs": true,
     "skipLibCheck": true,
-    "strict": true,
+    "strict": false,
```
aaronccasanova: I'm sure you had a reason, but mind adding some insight on this change?

polaris-for-vscode/tsconfig.json
sam-b-rose: Since `@shopify/polaris-tokens` is a dependency of `polaris-for-vscode`, we should let TypeScript know that it should build the types for tokens as well when running `tsc -b`. That's what I'm gathering based on my understanding of the [project references docs](https://www.typescriptlang.org/docs/handbook/project-references.html#what-is-a-project-reference), but I could be interpreting it incorrectly 😓

polaris.shopify.com/tsconfig.json
sam-b-rose: This is what happens when you copy the [tsconfig/bases/next.json](https://github.com/tsconfig/bases/blob/main/bases/next.json) without checking the values for all properties carefully enough 😅  will change back to `strict: true`! Thank you for catching!

package.json
sam-b-rose: Looks like `"pretty": true` is the default. Going to leave it out of the config and the command flag 👍

polaris-for-vscode/package.json
```diff
@@ -33,8 +33,7 @@
     "build": "tsc -b",
     "dev": "tsc -b -w",
     "lint": "TIMING=1 eslint --cache .",
-    "clean": "rm -rf .turbo node_modules client/dist server/dist",
-    "vscode:prepublish": "yarn build"
+    "clean": "rm -rf .turbo node_modules **/dist"
```
BPScott: `**/BLAH` globs always feel very scary and overreaching to me, I prefer what we've got currently. If we want to get terser we could do `'{client,server}/dist`


Though I'd probably update the build config so that it generates content in `dist/client` and `dist/server` and then we can do a boring old `rm dist` like everywhere else rather than polaris-for-vscode differ from the trend of where everything else puts compiled output

polaris-for-vscode/tsconfig.json
BPScott: yep explanation that sounds conceptually right to me.

These tsconfigs should state all of their dependencies. It should be possible to start with a fresh checkout, and then do `yarn run tsc -p polaris-for-vscode/tsconfig.json` and it complete successfully. That would be thanks to this references list saying "oh hey build these projects before starting on polaris-for-vscode".


That said, this seems incorrect - None of the files/includes referenced in `polaris-for-vscode/tsconfig.json` mention polaris-tokens (as this project doesn't load any files itself, it just references two other projects). Meanwhile code in `polaris-for-vscode/server` references tokens, but there is no references to the polaris-tokens project in `polaris-for-vscode/server/tsconfig.json`.

I think the reference to polaris-tokens needs to be removed here and added to `polaris-for-vscode/server/tsconfig.json` - because that's the project that actually depends upon polaris-tokens.

.github/workflows/publish-polaris-for-vscode.yml
```diff
@@ -16,14 +16,17 @@ jobs:
...
-          node-version: '16'
-          cache: 'yarn'
+          node-version: 16
+          cache: yarn
 
       - name: Install dependencies
         run: yarn
 
+      - name: Build extension
+        run: yarn turbo run build --filter=polaris-for-vscode
```
BPScott, line 26: Does turborepo know that a build of polaris-for-vscode should trigger a build of polaris-tokens beforehand?

BPScott (commented):
Haven't checked that the changes are exact, but they generally seem right to me.

I've dropped some notes inline

.github/workflows/publish-polaris-for-vscode.yml
sam-b-rose: Yep! Tested with a clean setup (no build output) 👍 🎉

Screen Shot 2022-05-12 at 12 35 35 PM

polaris-for-vscode/package.json
sam-b-rose: Moved the server and client files into a single `src` folder. Then removed the extra layer of tsconfigs down to a single config for the package

polaris-for-vscode/tsconfig.json
sam-b-rose: Removed the reference entirely. Seems we could only need it if we wanted to build (as you mentioned) directly using `tsc -p polaris-for-vscode/tsconfig.json` but I think we should keep it simple to using turbo for building from the root (ex: `yarn turbo run build --filter=polaris-for-vscode`) which will build with the proper dependency order (including types) 🏗️ 👍

BPScott (approved):
Loving the standardisation of polaris-for-vscode along the way

PR #5725, opened 2022-05-05
Replace Yarn with pnpm as package manager

### LINKED ISSUE #5679: Use pnpm for optimized dependency management and monorepo support
[pnpm](https://github.com/pnpm/pnpm) is an alternative package manager for Node.js. It is a drop-in replacement for npm, but faster and more efficient (3 times faster – [see benchmarks](https://github.com/pnpm/node-package-manager-benchmark).

📺 [Yarn vs pnpm comparison](https://www.youtube.com/watch?v=d1E31WPR70g)

pnpm has all the additional features that Yarn has over npm:

1. **security**. Like Yarn, pnpm has a special file with all the installed packages’ checksums to verify the integrity of every installed package before its code is executed.
1. **offline mode**. pnpm saves all the downloaded package tarballs in a local registry mirror. It never makes requests when a package is available locally. With the --offline parameter, HTTP requests can be prohibited at all.
1. **speed**. pnpm is faster than npm and Yarn (both with cold and hot cache). Yarn copies files from cache whereas pnpm just links them from the global store.

sam-b-rose:

Fixes #5679 

Replaces all usage of Yarn with `pnpm`. This includes:

- [x] `.lock` files
- [x] GitHub Action workflows
- [x] Shipit commands
- [x] Usage of `yarn` commands in documentation


sam-b-rose:
This depends on the ability to use Rollup for the VS Code extension. This can be unblocked when this dependency fix is merged

https://github.com/microsoft/vscode-languageserver-node/pull/957

PR #5727, opened 2022-05-05
Adding polaris-for-figma plugin into the monorepo

### LINKED ISSUE #5234: [Polaris for Figma] Migrate the plugin to the Polaris monorepo
Currently the plugin lives in https://github.com/shopify/figma-plugins with all of our other plugins.

Don't feel like it needs to move to polaris-react right now before the monorepo structure is set up and before the tokens package is available because I don't feel like it needs to be public during development, open to opinions!

romellogoodman:

Fixes #5234 

Ports the code for the figma plugin into the monorepo

* Adds polaris-for-figma folder with code from https://github.com/shopify/figma-plugins
* Documents the release steps for the plugin


polaris-for-figma/webpack.config.js
```diff
@@ -0,0 +1,56 @@
...
+        test: /\.css$/,
+        use: [ 'style-loader', 'css-loader' ]
+      },
+      {test: /\.(png|jpg|gif|webp|svg)$/, loader: [{loader: 'url-loader'}]},
+    ],
+  },
+  resolve: {extensions: ['.tsx', '.ts', '.jsx', '.js']},
+  output: {
+    filename: '[name].js',
+    path: path.resolve(__dirname, 'build'),
```
sam-b-rose, line 42: Can we use `dist` here or is `build` a required output directory for Figma plugins?

polaris-for-figma/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+  "api": "1.0.0",
+  "name": "Polaris for Figma",
+  "main": "build/code.js",
+  "ui": "build/ui.html",
```
sam-b-rose, line 5: Same here, and for the `main` key. Can we use `dist/us.html` and `dist/code.js` respectively?

Also for consistency, is it weird to use `index.js` and `index.html`? Not as descriptive but also follows convention

.gitignore
```diff
@@ -9,6 +9,7 @@
 .turbo
 dist
 node_modules
+/polaris-for-figma/build
```
sam-b-rose, line 12: We can remove this if using the `dist` output directory convention

polaris-for-figma/README.md
```diff
@@ -0,0 +1,21 @@
+# Polaris for Figma
+
+Official Figma plugin for building with the Shopify [Polaris Design System](https://polaris.shopify.com/).
+
+## Development
+
+Install dependencies
+
+```bash
```
sam-b-rose, line 9: Can we use `sh` instead of `bash` here and below?

sam-b-rose (approved):
Looks good overall! We should also add it to:

- the yarn `workspaces` array in the root `package.json`
- the root `tsconfig.json` as a project reference for type checking

romellogoodman:
@samrose3 will do these changes in a follow up so that this PR maintains the status of the initial files

BPScott:
polaris-for-figma/yarn.lock shouldn't exist - dependencies should be in the root level yarn.lock

romellogoodman:
@BPScott gonna follow up in another PR later this week to integrate the plugin properly into the monorepo. We were trying to get it out of its current repo so we could archive it

PR #5731, opened 2022-05-05
Stop errors when running a repository after yarn install without yarn build

alex-page:
Every now and then someone runs `yarn install` or `dev up` then starts a project for it to have multiple errors as a build has not been ran. This change would remove the need to remember to run `yarn build` after install.

This is very strict and forces a build after every install. This might not be the right long term solution. Happy for any other ideas.

### DIFF
--- a/.github/workflows/ci-a11y-vrt.yml
+++ b/.github/workflows/ci-a11y-vrt.yml
@@ -36,11 +36,8 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-accessibility-test-v1-
 
-      - name: Install dependencies
+      - name: Install dependencies and build packages
         run: yarn --frozen-lockfile
-
-      - name: Build packages
-        run: yarn turbo run build --filter=@shopify/polaris
 
       - name: Build Storybook
         run: yarn workspace @shopify/polaris run storybook:build --quiet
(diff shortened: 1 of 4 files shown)


aaronccasanova:
Personally, I think we should avoid `postinstall` scripts. It's a bit magic and only resolves the described scenario on initialization (as in folks will still have to `yarn` or `yarn build` if interrelated deps change).

Perhaps we can address this confusion by improving the documentation such as reframing the README from `Commands` to `Getting started`. Also, `Turbo` addresses this scenario with their [`--filter` syntax](https://turborepo.org/docs/core-concepts/filtering#include-dependents-of-matched-packages). For example, the following will run Polaris and all it's dependencies in watch mode:

```sh
yarn turbo run dev --filter=@shopify/polaris...
```

sam-b-rose:
> Personally, I think we should avoid postinstall scripts. It's a bit magic and only resolves the described scenario on initialization...

Agreed with @aaronccasanova here. This solves a one-time learning curve and creates script coupling with our install step and build step. 

I'd prefer to run any build steps in a `dev` script rather than postinstall–less magic and more directly related to the desired task of developing.

Example: `"dev": "turbo run build --filter=@shopify/polaris...` will run the `dev` script of `@shopify/polaris` and all its dependencies (which could include a build but in `watch` mode `rollup -c -w`). This should respect the topological order of build dependencies as well so `polaris-tokens` is ready and running before `polaris-react`

PR #5732, opened 2022-05-05
Add icon docs to NPM

alex-page:
**What this PR does**
- [x] Generate a JSON file for icon documentation
- [x] Add `metadata.json` to the NPM package (increasing bundle size by 100kb)

**Next steps**
- [ ] Remove `polaris-icons` npm package from private repo and update private site to use docs from NPM package
- [x] Import documentation for polaris.shopify.com from NPM package dependency


BPScott:
Might be worth double checking what bits of metadata are used on https://polaris-icons.shopify.com. I'm faiiiirly sure date_added, date_modified, authors, version and exclusive_use are never surfaced there so we could drop those keys and save a few bytes.

martenbjork:
> Might be worth double checking what bits of metadata are used...

Thumbs up! These won't be used for the new website either.

martenbjork (approved):
Aces!

sam-b-rose (approved):
Looks good! Only suggestion is that we provide a way of accessing this metadata that doesn't rely on reaching into the `dist` folder. Should we add an exports key to the `package.json`

```json
"exports": {
    ".": {
      "types": "./dist/index.d.ts",
      "import": "./dist/index.mjs",
      "require": "./dist/index.js"
    },
    "./metadata/*": "./dist/metadata.json"
  },
```

Question though, why don't we just store the Yaml metadata as JSON? Or why not just include the Yaml data in the npm package. Curious why or if we need to support both.

Or an alternative would be to glob and form this data structure on the site build of `polais-icons.shopify.com`.

polaris.shopify.com/src/components/IconGallery/IconGallery.tsx
```diff
@@ -1,21 +1,24 @@
...
 import { className } from "../../utils/various";
+
+
+
+const icons: Icons = require('@shopify/polaris-icons/metadata');
 const importedSvgs = require.context(
   "../../../../polaris-icons/icons",
   true,
   /\.svg$/
 );
```
alex-page, line 19: **I left this as a follow up task.** The `@shopify/polaris-icons` package has a `dist/svg` directory where we should be able to access these files. I tried a couple of things and failed.

BPScott: > I tried a couple of things and failed.

You've added an `exports` key to the package.json. That limits what is visible to places that try to import things. I think you might need to add `dist/svg/*.svg`: `dist/svg/*.svg` or something to that to keep being able to access those files

PR #5737, opened 2022-05-07
Add icon metadata to @shopify/polaris-icons

alex-page:
| Before | After |
| --- | --- |
| [unpkg](https://unpkg.com/browse/@shopify/polaris-icons@4.19.0/) | [dist.zip](https://github.com/Shopify/polaris-icons/files/8655522/dist.zip) | 

**What this PR does**
- [x] Simplify `customTypes` plugin to run for metadata/entrypoints
- [x] Simplify rollup config to remove all deprecated and fill replacement code
- [x] Generate metadata with a script and add it to the `dist/metadata.js` file
- [x] Replaces `parseIcons` and `icons.json` with import from `@shopify/polaris-icons`
- [x] Use same `glob` version in `@shopify/polaris` and `@shopify/polaris-icons`
- [x] Bump version to `v4.20.0` for next release

**Next steps in https://github.com/Shopify/polaris-icons/pull/1518**
- [ ] Remove `packages/polaris-icons` from previous private repo
- [ ] Update https://polaris-icons.shopify.com site to use metadata from NPM package


polaris-icons/package.json
```diff
@@ -19,6 +19,14 @@
   "files": [
     "dist/*"
   ],
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.mjs",
+      "require": "./dist/index.js"
+    },
+    "./metadata": "./dist/metadata.json"
```
BPScott: Exposing a json file doesn't feel that great. You can load it with `require()` but you can't load it with `import` statement (unless you muck about with running node with CLI options but that's not general purpose yet). I think it might be better to expose the metadata as a js file if possible. That feels like it plays a lot nicer with exports notations. Alternatively if we want to stick with json, could we have the key be `./metadata.json`? To me it feels like extension-less keys should be reserved for js/mjs code.


Tangentally related - it looks like this file is generated with a script. It might be nicer to use rollup's virtual file stuff to create this file, in [the same way that it generates the index file](https://github.com/Shopify/polaris/blob/main/polaris-icons/rollup.config.js#L258-L260) - that way we get to keep the build as " run rollup" without needing extra steps

polaris.shopify.com/src/types.ts
```diff
@@ -24,3 +24,14 @@ export type LineConfig = {
   fromOrigin?: "center" | "edge" | "left" | "right";
   toOrigin?: "center" | "edge" | "left" | "right";
 };
+
+interface Icon {
```
sam-b-rose: Thoughts on calling this `IconMetadata` since it relates to the icon info rather than the icon resource itself? If we want to support metadata from the npm package, the type information should probably come from there.

To @BPScott's point, perhaps we should export the metadata as JS (but with TS declarations). That would improve import/require compatibility, include type info, and consumers could easily convert to JSON if desired with `JSON.parse(meta)`. Then we could also probably remove the export key in the `package.json` and instead just export metadata from the package index file.

polaris.shopify.com/src/utils/search.ts
```diff
@@ -1,11 +1,14 @@
...
-import icons from "../data/icons.json";
-import guidelines from "../data/guidelines.json";
 import Fuse from "fuse.js";
 import { slugify, stripMarkdownLinks } from "./various";
 
+import components from "../data/components.json";
+import guidelines from "../data/guidelines.json";
+import {Icons} from '../types';
+
+const icons: Icons = require('@shopify/polaris-icons/metadata');
```
sam-b-rose: The module mixing 🙈 this pattern of `require` & `import` together scares me a bit 😬

polaris.shopify.com/src/types.ts
BPScott: > Then we could also probably remove the export key in the package.json and instead just export metadata from the package index file.

I'd probably lean towards keeping a separate import for the metadata as it isn't inntended to be used outside of our documentation. Rather than make bundler dead-code-eliminate that big data object, I'd lean towards not including it in the main entrypoint in the first place

alex-page: I cleaned this up. It's now called metadata and I removed all export keys from package.json ✨

polaris.shopify.com/src/utils/search.ts
alex-page: Yes! I moved this to imports!

polaris-icons/package.json
alex-page: I ended up removing all the virtual file stuff as it felt magic with memory and having to console log out what is happening. I now just generate the files in the repository and run rollup on them.

.gitignore
```diff
@@ -13,3 +13,6 @@ node_modules
 /polaris-react/build
 /polaris-react/build-internal
 /polaris-react/src/styles/_media-queries.scss
+
+# Generated file for icons build
+/polaris-icons/src/index.ts
```
sam-b-rose: If the generated file is going to be in `src`, I'd rather it be checked into version control. Otherwise, we should put it in `dist`

polaris-icons/package.json
```diff
@@ -1,13 +1,13 @@
...
-  "version": "4.19.0",
+  "version": "4.20.0",
   "license": "SEE LICENSE IN LICENSE.md",
   "author": "Shopify Inc.",
   "main": "dist/index.js",
   "module": "dist/index.mjs",
   "types": "dist/index.d.ts",
   "scripts": {
-    "build": "rollup -c",
+    "build": "node ./src/generate-index.js && rollup -c",
```
sam-b-rose: An improvement thought... this is where a check like [lint-staged](https://github.com/okonet/lint-staged) could be useful.

If we check the `index.ts` file into version control, then a hook could run when paths are changed within the `icons` directory rather than needing to regenerate it on every build.

polaris.shopify.com/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "polaris-shopify-com",
+  "name": "polaris.shopify.com",
```
sam-b-rose, line 2: Sneaky sneaky... trying to break the rules of `package.json` I see. 

https://docs.npmjs.com/creating-a-package-json-file#required-name-and-version-fields

> The "name" field contains your package's name, and must be lowercase and one word, and may contain hyphens and underscores.

sam-b-rose (commented):
Looks good! Only feedback:

* Keep the name `polaris-shopify-com`
* Check the `src/index.ts` into version control
* Remove `/polaris-icons/src/index.ts` from all ignore files

Optional additional security:

* Add `lint-staged` for checking and running generation script when `/polaris-icons/icons` changes
* https://github.com/okonet/lint-staged

polaris.shopify.com/package.json
alex-page: It's not a package. The rules don't exist 🧙‍♂️

sam-b-rose (approved):

.gitignore
alex-page: I agree. I am thinking about this, we wouldn't want to include this file in the NPM package so it would need to be cleaned up. At that point it sounds like rollup virtual plugin 😅

polaris-icons/package.json
alex-page: I would also need to run a fix script to fix linting errors as the generated files have issues. Thinking about this one.

BPScott: Fun fact: Originally polaris-icons had a generated index.ts checked into version control. And then whenever people added an icon they had to regenerate the file. What we found what that everybody always forgot to regenerate it as part of their PRs.

At which point I did the rollup-plugin-virtual work to make it so people didn't need to think about it.

alex-page: It all makes sense. I still prefer being able to see the file and run rollup on it. We are sure that build gets ran before release so consumers shouldn't need to do it. Any major concerns leaving as is?

```
deploy:
  override:
    - yarn turbo run build --filter=@shopify/polaris-icons
    - npm publish polaris-icons --tag latest
```

polaris.shopify.com/src/components/IconGallery/IconGallery.tsx
```diff
@@ -1,21 +1,22 @@
...
+import { Tab } from "@headlessui/react";
+import {metadata} from '@shopify/polaris-icons';
+
+import TextField from "../TextField";
 import styles from "./IconGallery.module.scss";
 import Longform from "../Longform";
-import { Tab } from "@headlessui/react";
 import { className } from "../../utils/various";
+
+
```
aaronccasanova:
```suggestion
```

aaronccasanova (approved):

alex-page:
@aaronccasanova @samrose3 after jamming @bpscott I went back on the virtual stuff...

Cleaning it up the rollup config, removing the deprecated complexity and removing the fill replacement code (that wasn't used any more) meant that we could keep it quite clean and simple. This also allowed removing the ignores and src directory. Happy with this, will probably shipit, let me know if you have any follow up feedback.

We also split up the metadata exports. Feeling good.

PR #5746, opened 2022-05-09
[HD32][InContextLearning] Add dismiss button and onDismiss prop

annmariedoyle-shopify:
Adds an onDismiss prop and a dismiss button that calls onDismiss. 

More information on these changes can be found [here](https://docs.google.com/document/d/1Jf-6lPq5lygtFsZf9wt0vD5i7fKt6bHhrCHFL8-fjVc/edit).

Add a new function prop `onDismiss` to the component and renders a button a close button

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/polaris-react/locales/en.json
+++ b/polaris-react/locales/en.json
@@ -155,6 +155,9 @@
       "clear": "Clear",
       "clearLabel": "Clear {filterName}"
     },
+    "InContextLearning": {
+      "accessibilityLabel": "Close tutorial"
+    },
     "IndexProvider": {
       "defaultItemSingular": "Item",
       "defaultItemPlural": "Items",
--- a/polaris-react/src/components/InContextLearning/InContextLearning.tsx
+++ b/polaris-react/src/components/InContextLearning/InContextLearning.tsx
@@ -1,37 +1,57 @@
 import React, {useState} from 'react';
 
-import {Step} from './components';
+import {CancelSmallMinor} from '@shopify/polaris-icons';
+
+import {useI18n} from '../../utilities/i18n';
 import {Button} from '../Button';
+import {Step} from './components';
 
 interface Props {
-    children: React.ReactElement[];
+  children: React.ReactElement[];
+  onDismiss(): void;
 }
 
-export function InContextLearning({children}: Props) {
-    const hasMultipleSteps = children.length > 0;
-    const [currentStep, setCurrentStep] = useState(0);
+export function InContextLearning({children, onDismiss}: Props) {
+  const i18n = useI18n();
 
-    const showPrev = hasMultipleSteps && currentStep > 0;
-    const showNext = hasMultipleSteps && currentStep < (children.length - 1);
+  const hasMultipleSteps = children.length > 0;
+  const [currentStep, setCurrentStep] = useState(0);
 
-    const handleNext = () => {
-        setCurrentStep((currentStep) => currentStep + 1)
-        console.warn('Next clicked');
-    }
+  const showPrev = hasMultipleSteps && currentStep > 0;
+  const showNext = hasMultipleSteps && currentStep < children.length - 1;
 
-    const handlePrev = () => {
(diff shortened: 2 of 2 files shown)


polaris-react/locales/en.json
```diff
@@ -155,6 +155,9 @@
       "clear": "Clear",
       "clearLabel": "Clear {filterName}"
     },
+    "InContextLearning": {
+      "accessibilityLabel": "Close tutorial"
```
annmariedoyle-shopify, line 159: I got this content from this doc: https://docs.google.com/document/d/10x1ynHfpZVMDGKEhqIMhXhIT3suC8rdex4Dh1gbAJ3g/edit

I wasn't sure if we wanted to have a prop of the tutorial name passed into the component - at this point it seems unnecessary for just one label 🤷‍♀️  But I can update if we feel that:
1) having the name in this label is important
2) we have a use for the name beyond this label

polaris-react/src/components/InContextLearning/InContextLearning.tsx
```diff
@@ -1,37 +1,56 @@
...
+    console.warn('Next clicked');
+  };
+
+  const handlePrev = () => {
+    setCurrentStep((currentStep) => currentStep - 1);
+    console.warn('Prev clicked');
+  };
+
+  const dismissButton = onDismiss ? (
+    <div>
```
annmariedoyle-shopify, line 34: I assume styles will be added under a different PR

polaris-react/locales/en.json
kimhawthorne-shopify: I think this is fine for hack days — we can totally revisit if we want to make this more specific

rockinrobin714 (approved):
lgtm!

kimhawthorne-shopify (approved):
Looks good to me!

polaris-react/src/components/InContextLearning/InContextLearning.tsx
```diff
@@ -1,37 +1,58 @@
...
+
+import {useI18n} from '../../utilities/i18n';
+
 import {Step} from './components';
 import {Button} from '../Button';
 
 interface Props {
-    children: React.ReactElement[];
+  children: React.ReactElement[];
+  onDismiss?(): void;
```
arapehl: I read the doc outlining the pattern we're following from elsewhere, but since the design calls for a dismiss "X" button and `esc` behaviour, are sure we want to make this optional?

annmariedoyle-shopify: That's a good question. I was thinking that we would want things to be flexible and not force it to be dismissible but rather allow that as an option.

annmariedoyle-shopify: Looking at other Polaris components, this seems to be a common pattern - to allow some configuration/flexibility within the component.

arapehl: True, but in the case of banners, there are dismissible ones and non-dismissible ones. In our case, I think all we have are dismissible elements. We could check with UX to confirm.

annmariedoyle-shopify: Sure, I will ask in the channel.

annmariedoyle-shopify: Confirmed in the change the the `onDismiss` functionality in the [channel](https://shopify-hack-days.slack.com/archives/C03BFM4D54Y/p1652125473077789?thread_ts=1652125260.072959&cid=C03BFM4D54Y). I think the PR to close using `esc` should be a separate PR to ensure we are keeping out PRs small.

arapehl (approved):
Code is good. Design decision behind onDismiss can be hammered out even after this PR ships.

PR #5754, opened 2022-05-10
WIP steps in context learning

kimhawthorne-shopify:
- Resolve conflicts on Alex's step prototype work
- Navigate the learning popover based on class name
- Add `Learner` story for HD32 work


rockinrobin714 (approved):

polaris-react/playground/Learner.tsx
```diff
@@ -0,0 +1,64 @@
+import React, {useEffect, useState, useLayoutEffect} from 'react';
+import {Layout, Page, Card, List, InContextLearning} from '../src';
+import styles from './Learner.scss';
+
+const LEARNING_STEPS = [
+  {
+    className: '.learning-step-one',
```
alololox, line 7: small nit that came up yesterday in our pair, more flexible to use a generic `selector` here so anything can be passed in (e.g. seeing `data-onboarding-project-step-one` within a render is more descriptive than seeing a `className` you might not know is tied to styles, etc)

PR #5757, opened 2022-05-10
add close button; fix multi step logic; render step counter

kimhawthorne-shopify:
- Fix logic for determining multiple steps
- Render "Got it" button when there is one step OR on the last step
- Render step counter

### DIFF
--- a/polaris-react/src/components/InContextLearning/InContextLearning.tsx
+++ b/polaris-react/src/components/InContextLearning/InContextLearning.tsx
@@ -8,6 +8,11 @@ import {Portal} from '../Portal';
 import {PositionedOverlay} from '../PositionedOverlay';
 import {KeypressListener} from '../KeypressListener';
 import {Key} from '../../types';
+import {TextStyle} from '../TextStyle';
+import {Stack} from '../Stack';
+import {Step} from './components';
+
+import {styles} from '../CustomProperties/styles';
 
 interface InContextLearningStep {
   selector: string;
@@ -32,7 +37,8 @@ interface Props {
 
 export function InContextLearning({children, steps, onDismiss}: Props) {
   const i18n = useI18n();
-  const hasMultipleSteps = steps.length > 0;
+  const totalSteps = steps.length;
+  const hasMultipleSteps = totalSteps > 1;
   const [currentStep, setCurrentStep] = useState(0);
   const [currentActivator, setCurrentActivator] = useState<HTMLElement | null>(
     null,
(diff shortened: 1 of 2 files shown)


polaris-react/src/components/InContextLearning/InContextLearning.tsx
```diff
@@ -71,19 +87,43 @@ export function InContextLearning({children, steps, onDismiss}: Props) {
...
       {dismissButton}
       {steps[currentStep].content}
-      {showPrev && <Button onClick={handlePrev}>Prev</Button>}
-      {showNext && <Button onClick={handleNext}>Next</Button>}
+
+      <div
+        style={{
+          display: 'flex',
+          alignItems: 'center',
+          justifyContent: 'space-between',
```
kimhawthorne-shopify: Is there a "Polaris" way of laying out three columns that don't stack? I couldn't get `Layout` to work for me.

Otherwise I can move these styles to a proper stylesheet 😆

rockinrobin714: Would `stack` work?

polaris-react/src/components/InContextLearning/InContextLearning.tsx
```diff
@@ -25,7 +27,8 @@ interface Props {
 
 export function InContextLearning({children, steps, onDismiss}: Props) {
   const i18n = useI18n();
-  const hasMultipleSteps = steps.length > 0;
+  const totalSteps = steps.length;
+  const hasMultipleSteps = totalSteps > 1;
```
annmariedoyle-shopify, line 41: 👍

polaris-react/src/components/InContextLearning/InContextLearning.tsx
alololox: +1 for stack! `<Stack alignment="center" wrap={false} distribution="equalSpacing">`

kimhawthorne-shopify: YES this is exactly what I wanted — thank you!

polaris-react/src/components/InContextLearning/InContextLearning.tsx
```diff
@@ -60,6 +70,13 @@ export function InContextLearning({children, steps, onDismiss}: Props) {
     </div>
   );
 
+  const counterMarkup = hasMultipleSteps ? (
+    <p style={{textAlign: 'center'}}>
```
kimhawthorne-shopify, line 87: I have two `textAlign` inline styles to align the counter and "Back" button correctly. There is maybe a Polaris way to do this as well? Or eventually we should have a PR to move all these inline styles to a stylesheet.

alololox (approved):
🔥

annmariedoyle-shopify (approved):

PR #5765, opened 2022-05-10
switch to refs and context

chris-w-mclean-shopify:
- Replaces direct dom querying with React Context and refs
  - Allows for stateful updates of workflows
  - Prevents dom modifications from breaking the system
- Introduced a more declarative approach to bootstrapping the workflow

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/polaris-react/playground/Learner.tsx
+++ b/polaris-react/playground/Learner.tsx
@@ -41,9 +41,8 @@ function LearnerApp() {
               <Card.Section title="Items">
                 <List>
                   <List.Item>
-                    <InContextLearning.Step stepIndex={0}>
+                    <InContextLearning.Step direction="top-left" stepIndex={0}>
                       <span>1 × Oasis Glass, 4-Pack</span>
-                      <p>FOOOOOO</p>
                     </InContextLearning.Step>
                   </List.Item>
                   <List.Item>1 × Anubis Cup, 2-Pack</List.Item>
@@ -52,7 +51,7 @@ function LearnerApp() {
             </Card>
             <Card>
               <Card.Section title="Collections">
-                <InContextLearning.Step stepIndex={1}>
+                <InContextLearning.Step direction="top-right" stepIndex={1}>
                   <span>another piece of content</span>
                 </InContextLearning.Step>
               </Card.Section>
@@ -60,7 +59,7 @@ function LearnerApp() {
             </Card>
 
             <Stack distribution="trailing">
-              <InContextLearning.Step stepIndex={2}>
+              <InContextLearning.Step direction="none" stepIndex={2}>
                 <span data-learning-step-three>
                   Yet another piece of content!
                 </span>
--- a/polaris-react/playground/Learner.tsx
+++ b/polaris-react/playground/Learner.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useState} from 'react';
 
 import {
   Layout,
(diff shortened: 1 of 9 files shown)


polaris-react/playground/Learner.tsx
```diff
@@ -1,31 +1,36 @@
...
-    selector: '[data-learning-step-two]',
-    content: <span>Step 2: Testing React</span>,
-  },
-  {
-    selector: '[data-learning-step-three]',
-    content: <span>Step 3: Testing React</span>,
-  },
-];
+function StepOne() {
+  return <span>Step 1: Testing React</span>;
```
rockinrobin714: just curious, why are the steps a `React.FunctionComponent` vs a plain ole ReactNode?

rockinrobin714 (approved):

polaris-react/playground/Learner.tsx
chris-w-mclean-shopify: I'm still somewhat newish to Typescript, so I'm happy for suggestions on what is appropriate here. What we're passing along isn't a node, but the function component itself. 

Do you have any thoughts on this approach? I'm not too attached to how this is being done now vs passing along a node.

polaris-react/src/components/InContextLearning/InContextLearning.tsx
```diff
@@ -112,7 +94,7 @@ export function InContextLearning({children, steps, onDismiss}: Props) {
               >
                 <KeypressListener keyCode={Key.Escape} handler={onDismiss} />
                 {dismissButton}
-                {steps[currentStep].content}
+                {steps[currentStep].component()}
```
alololox: I think [this is going to be problematic](https://kentcdodds.com/blog/dont-call-a-react-function-component). 

e.g. try adding a `useState` method within one of the steps within `Learner.tsx` and you should see: `Invalid hook call. Hooks can only be called inside of the body of a function component.`

polaris-react/playground/Learner.tsx
rockinrobin714: Makes sense! Probably not worth examining the pros in cons for a hack day. Just curious what was the reasoning behind making them functions versus nodes. e.g. `function StepOne() {
 return <span>Step 1: Testing React</span>;
}` would be `const StepOne = <span>Step 1: Testing React</span>;`
and then in the component itself you would just use `{steps[currentStep].component}` instead of `{steps[currentStep].component()}`

rockinrobin714: also - great work!!!

rockinrobin714: oops I see @alololox already commented below

polaris-react/src/components/InContextLearning/InContextLearning.tsx
chris-w-mclean-shopify: Good catch. I'll look into this.

chris-w-mclean-shopify: I switched things to pass along a react node instead of calling the component directly.

polaris-react/playground/Learner.tsx
chris-w-mclean-shopify: Yup. This was a good callout. I adjusted the code to use React nodes instead.

alololox (approved):

PR #5775, opened 2022-05-11
polaris-for-figma, hooking up to monorepo

### LINKED ISSUE #5234: [Polaris for Figma] Migrate the plugin to the Polaris monorepo

romellogoodman:

As a follow up to #5727 this integrates the figma plugin into the monorepo infrastructure

Fixes #5234 

* Integrates the plugin into the monorepo's infrastructure
* Simplifies tsconfig
* Handles ts and linting errors


alex-page (approved):

polaris-for-figma/src/utils.ts
```diff
@@ -48,12 +47,12 @@ const backgroundTokens: Token[] = tokensArray.filter(({figmaName}) =>
 const surfaceTokens: Token[] = tokensArray.filter(({figmaName}) =>
   figmaName.startsWith('Surface'),
 );
-const shadowTokens: Token[] = tokensArray.filter(({figmaName}) =>
-  figmaName.startsWith('Shadow'),
-);
-const hintTokens: Token[] = tokensArray.filter(({figmaName}) =>
-  figmaName.startsWith('Hint'),
-);
+// const shadowTokens: Token[] = tokensArray.filter(({figmaName}) =>
```
romellogoodman, line 50: We're about to add this functionality to the plugin in https://github.com/Shopify/polaris/issues/5398 so commenting out instead of deleting

PR #5779, opened 2022-05-11
[EmptyState] Fix vertical scroll on small screens

### LINKED ISSUE #5662: [EmptyState] has vertical scroll on small screens.
## Issue summary

There is a minor vertical scroll added to `.Polaris-EmptyState__DetailsContainer` when on smaller screens. This seems to be an unintentional side effect of `overflow-x: hidden` set on that div.

## Expected behavior

When on smaller screens, the EmptyState component renders content without a vertical scroll.

## Actual behavior

Notice the small scroll bar added beside the content:

Screen Shot 2022-04-28 at 1 23 58 PM
Screen Shot 2022-04-28 at 1 24 43 PM

Add content as child of the EmptyState component.

## Reduced test case

## Specifications

- Polaris version number: v9.5.2 (latest)
- Browser: Chrome
- Device: Mac m1
- Operating System: Ios

rivaridley:

Fixes https://github.com/Shopify/polaris/issues/5662

The `overflow-x: hidden` property on the Details and Images containers doesn't hide any horizontal overflow since these elements are flex items set with `flex: 1 1 auto`. These elements will grow to the width of the content. If you add a large image in without specifying the `largeImage` prop, the overflow-x does nothing and the image causes a horizontal scroll. 

`overflow-x: hidden` also has unintended side effect of adding a vertical scroll because of [how browsers interpret `overflow-y` as default (which is scroll)](https://stackoverflow.com/questions/6179831/setting-overflow-x-hidden-adds-a-vertical-scrollbar) when an overflow-x is set.

The property was added in this [commit](https://github.com/Shopify/polaris/commit/bf4d3d01b58a4d4be46f233dab5cb277e20adce2), but @lgriffee and I both couldn't find the PR/Issue detailing why. In my opinion, with the content of EmptyState being flexed and properties to take care of large images, this style isn't needed and should be removed since it causes a vertical scroll bar on mobile screens.

Removes `overflow-x: hidden` from `.Polaris-EmptyState__DetailsContainer` and `.Polaris-EmptyState__ImageContainer` 

Before:
Screen Shot 2022-05-11 at 10 37 53 AM

After:
Screen Shot 2022-05-11 at 10 39 10 AM

### 🎩 checklist

### DIFF
--- a/polaris-react/UNRELEASED.md
+++ b/polaris-react/UNRELEASED.md
@@ -10,6 +10,8 @@ Use [the changelog guidelines](/documentation/Versioning%20and%20changelog.md) t
 
 ### Bug fixes
 
+- Fixed vertical scroll on small screens in `EmptyState` ([#5779](https://github.com/Shopify/polaris/pull/5779))
+
 ### Documentation
 
 ### Development workflow
--- a/polaris-react/src/components/EmptyState/EmptyState.scss
+++ b/polaris-react/src/components/EmptyState/EmptyState.scss
@@ -75,10 +75,6 @@ $centered-illustration-width: 226px;
   @include page-content-when-not-partially-condensed {
     flex-basis: 50%;
   }
-
-  @include page-content-when-fully-condensed {
-    overflow-x: hidden;
-  }
 }
 
 .withinContentContainer {


philmongeon:
do we need a bug fix entry in the unreleased change log ?

philmongeon:
do we need a bug fix entry in the unreleased change log ?

kyledurand (approved):
This should be ok to ship. I think I remember why this was included and it was for the old empty states that had huge illustrations.

Just be sure to add an entry for this under bug fixes in the UNRELEASED.md doc 👍

rivaridley:
> This should be ok to ship. I think I remember why this was included and it was for the old empty states that had huge illustrations.
> 
> Just be sure to add an entry for this under bug fixes in the UNRELEASED.md doc 👍

Great, I've added an entry there, thanks for the review ✌️

PR #5784, opened 2022-05-11
Root level jest config

BPScott:
This pulls the jest dependencies and config up to the root level.
You can still run `jest` from a given sub project, but now you can also
run `jest` from the root of the repository and it shall run all tests.

Improving how tests can be ran. In the case of polaris-react we means we leaning on using jest directly rather than using loom to trigger running jest. This makes the jest config more obvious and amenable to change.

This allows us to remove per-package jest dependencies, and allows us to run jest from the root of the repository - this means `jest --watch` from the root of the repo shall work without issue.

This also moves us a step further towards removing loom within polaris-react - after this loom is only used for building, and we can replace that with a plain old rollup config in a future PR.

- Remove per-package jest dependencies, moving them up to the root package.json
- Update jest.config.js in polaris-tokens and polaris-icons.
- Update a pair of tests in polaris-react so they use absolute file paths and thus no longer rely on running jest with polaris-react as the current working directory.
- Add jest.config.js to polaris-react that is based upon the jest config that loom generates.
- Run polaris-react tests with jest directly instead of using `loom test`.

The big reduction in yarn.lock is due to removing babel-jest v24.x - we now use only use v27.x.

### How to 🎩

Tests pass.

Play with some of these commands:

- `yarn jest` in the root of the repository work and runs test from all folders. Note that `yarn jest polaris-tokens` works to filter, and `yarn jest --watch` runs a file watcher that looks just for changed files
- `yarn test` still runs turborepo that triggers the test runs on a per package basis
- `yarn test` when in the polaris-react folder works


alex-page (approved):

aaronccasanova (approved):

sam-b-rose (approved):
Very nice, looks 💯

PR #5785, opened 2022-05-12
[IndexTable] Updates to IndexTable from polaris-next version

mrcthms:

The Admin Quality team are currently working on an initiative to update the index filtering across our commerce objects. A foundational piece of work for that is to update the IndexTable in Polaris with the relevant updates to the polaris-next IndexTable in shopify/web.

There is a lot of code that is in the polaris-next IndexTable that isn't in the polaris-react version, but we do not want to include everything different. This is because there are many examples of single-use changes being introduced to the polaris-next version which have no place in the polaris-react component. The only changes we are adding are:
- Adding a `className` prop to a IndexTable.Cell so that folks can set widths on cells. A `width` prop was thought about, but there is a use case where folks could need to set a min/max width as well, at which point a single prop would not suffice.
- Adding a `customClickHandler` prop to a IndexTable.Row, so folks can override the default behaviour.
- Adding a `paginatedSelectAllActionText` to the IndexTable, so folks can render a custom string in their table pagination.
- Exports a `clearSelection` method from the `use-index-resource-state` hook as this is a very common need in `shopify/web`
- Updates the styling of the `ScrollBarContainer` to use the correct border radius, and remove the top border.
- Update styling for `TableCell` to fix unpadded cell bug.
- Adds a 20px width to the `TableHeading-first` to fix the width of the checkbox column.

### 🎩 checklist

### DIFF
--- a/polaris-react/src/components/IndexTable/IndexTable.tsx
+++ b/polaris-react/src/components/IndexTable/IndexTable.tsx
@@ -46,6 +46,7 @@ export interface IndexTableBaseProps {
   children?: React.ReactNode;
   emptyState?: React.ReactNode;
   sort?: React.ReactNode;
+  paginatedSelectAllActionText?: string;
   lastColumnSticky?: boolean;
   selectable?: boolean;
 }
@@ -67,6 +68,7 @@ function IndexTableBase({
   children,
   emptyState,
   sort,
+  paginatedSelectAllActionText,
   lastColumnSticky = false,
   ...restProps
 }: IndexTableBaseProps) {
@@ -746,13 +748,17 @@ function IndexTableBase({
       return;
     }
 
+    const customActionText =
+      paginatedSelectAllActionText ??
+      i18n.translate('Polaris.IndexTable.selectAllItems', {
+        itemsLength: itemCount,
+        resourceNamePlural: resourceName.plural.toLocaleLowerCase(),
+      });
+
     const actionText =
       selectedItemsCount === SELECT_ALL_ITEMS
         ? i18n.translate('Polaris.IndexTable.undo')
-        : i18n.translate('Polaris.IndexTable.selectAllItems', {
-            itemsLength: itemCount,
-            resourceNamePlural: resourceName.plural.toLocaleLowerCase(),
-          });
+        : customActionText;
 
     return {
       content: actionText,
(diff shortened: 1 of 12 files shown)


alex-page (approved):

polaris-react/src/components/IndexTable/components/Row/Row.tsx
```diff
@@ -21,6 +21,7 @@ export interface RowProps {
   subdued?: boolean;
   status?: RowStatus;
   onNavigation?(id: string): void;
+  customClickHandler?(): void;
```
AndrewMusgrave: Nit: Should this just be `onClick`?

polaris-react/src/components/IndexTable/components/Row/tests/Row.test.tsx
```diff
@@ -220,6 +220,23 @@ describe('<Row />', () => {
     expect(onNavigationSpy).toHaveBeenCalledTimes(1);
   });
```
AndrewMusgrave, line 222: Should we test when `customClickHandler` is used the code below isn't executed?

polaris-react/src/components/IndexTable/tests/IndexTable.test.tsx
```diff
@@ -382,6 +382,26 @@ describe('<IndexTable>', () => {
       );
     });
 
+    it('renders a custom pagination string if present', () => {
```
AndrewMusgrave:
```suggestion
    it('renders a custom select all string if present', () => {
```

The terminology is a little odd. The button doesn't actually handle any pagination. It just selects "all" resources

polaris-react/src/components/IndexTable/IndexTable.tsx
```diff
@@ -806,3 +812,4 @@ export function IndexTable({
 
 IndexTable.Cell = Cell;
 IndexTable.Row = Row;
+IndexTable.ScrollContainer = ScrollContainer;
```
AndrewMusgrave: For context, why did we need to export an internal component?

AndrewMusgrave (commented):
LGTM! Left a few minor comments. Thank you for making this change 👐

polaris-react/src/components/IndexTable/IndexTable.tsx
mrcthms: It's used https://github.com/Shopify/web/blob/main/app/sections/ShippingLabels/PurchaseShippingLabels/components/TableView/TableView.tsx#L155 but it looks like that just needs to be wholesale replaced with the IndexTable rather than the own-rolled table that's currently there. Can remove 👍

PR #5793, opened 2022-05-12
Clean up recommended extensions

alex-page:
This PR removes a bunch of plugins that I see as optional or things that VSCode already supports. This leaves recommendations for extensions we strongly recommend using when developing in `shopify/polaris`.

Let me know your thoughts.


sam-b-rose (approved):

kyledurand (approved):

laurkim (approved):
💯

PR #5795, opened 2022-05-12
Remove TS project references and type check

### LINKED ISSUE #5792: Remove TS project references

sam-b-rose:

Fixes #5792 

Remove TS project references for the following reasons:

- the .tsbuildinfo files can't be cached in CI
- running into occasional build issues due to `tsbuildinfo` cache, having to `rm` them manually to get turbo builds to work
- the type check step in CI isn't providing much additional security
- the package builds should fail if TS isn't valid anyway

### DIFF
--- a/polaris-for-figma/tsconfig.json
+++ b/polaris-for-figma/tsconfig.json
@@ -1,3 +1,11 @@
 {
-  "extends": "@shopify/typescript-configs/dom"
+  "extends": "@shopify/typescript-configs/dom",
+  "compilerOptions": {
+    "composite": true,
+    "emitDeclarationOnly": true,
+    "resolveJsonModule": true,
+    "outDir": "dist",
+    "rootDir": "."
+  },
+  "include": ["src", "src/**/*.json"]
 }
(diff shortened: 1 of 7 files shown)


alex-page (commented):
The UX kind of sucks when getting type errors when running `yarn build`. If this could be cleaned up it could make sense.

screen_shot_2022-05-12_at_2 02 31_pm
screen_shot_2022-05-12_at_2 02 45_pm

BPScott:
> the .tsbuildinfo files can't be cached in CI

Huh, that's news to me? Caching these files work fine in the loom and [quilt](https://github.com/Shopify/quilt/blob/main/.github/workflows/node-ci.yml#L38-L39) repositories. You just need to make sure that the .d.ts files that are generated as part of the type check are also cached and restored at the same time.


> the package builds should fail if TS isn't valid anyway

It is very possible for code valid code to build with rollup sucessfully even if it fails type-checking. For instance `const a: string = 1` would fail type checking but would build sucessfully. Obviously this is a contrived example but things like "you're passing the wrong type into this function" would also trigger type errors but build with rollup sucessfully.

aaronccasanova:
> the package builds should fail if TS isn't valid anyway

@BPScott In regards to the above, I believe @samrose3 is referring to the fact that our [build script also builds types](https://github.com/Shopify/polaris/blob/5a01bbdba2c7178a3ff3363e57436f5802921168/polaris-tokens/package.json#L19-L22) and thus will fail the build on type errors.

sam-b-rose:
Closing in favor of #5799

PR #5796, opened 2022-05-12
stylelint use monorepo package

alex-page:
I don't know if this worked correctly. Testing the plugin everything worked as intended.

The intention of this PR is to use the `package.json` format for built packages to include `stylelint-polaris`. When running `yarn` after this change I had to specify a version and the `yarn.lock` has it there. Unsure if version bumps and edits would flow through as I intended.

```
Couldn’t find any versions for “@shopify/stylelint-polaris” that matches “*”.
Please choose a version of “@shopify/stylelint-polaris” from this list:
- 0.0.0-alpha.6
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "@shopify/eslint-plugin": "^41.0.1",
     "@shopify/prettier-config": "^1.1.2",
     "@shopify/stylelint-plugin": "^11.0.0",
+    "@shopify/stylelint-polaris": "*",
     "@shopify/typescript-configs": "^5.1.0",
     "eslint": "^8.3.0",
     "jest": "^27.5.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2759,6 +2759,13 @@
     stylelint-prettier "^2.0.0"
     stylelint-scss "^4.0.0"
 
+"@shopify/stylelint-polaris@*":
+  version "0.0.0-alpha.6"
+  resolved "https://registry.yarnpkg.com/@shopify/stylelint-polaris/-/stylelint-polaris-0.0.0-alpha.6.tgz#177bf4a6ed45fd0e3cb46f619a37a5696ab2ca9e"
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -33,6 +33,7 @@
     "@shopify/eslint-plugin": "^41.0.1",
     "@shopify/prettier-config": "^1.1.2",
     "@shopify/stylelint-plugin": "^11.0.0",
+    "@shopify/stylelint-polaris": "*",
```
sam-b-rose, line 36: Curious if the `*` doesn't work the same as it does for workspace packages since this is in the root package.json 🤔

yarn.lock
```diff
@@ -2759,6 +2759,13 @@
     stylelint-prettier "^2.0.0"
     stylelint-scss "^4.0.0"
 
+"@shopify/stylelint-polaris@*":
```
sam-b-rose, line 2762: If the local linking of the package was working, I don't think it would be added to the yarn.lock 😞

sam-b-rose (commented):
Looks like this strategy won't work for our use case. Yarn workspaces doesn't support locally linked packages at the root 😢 

> Workspaces must be descendants of the workspace root in terms of folder hierarchy. You cannot and must not reference a workspace that is located outside of this filesystem hierarchy.

https://classic.yarnpkg.com/lang/en/docs/workspaces/#toc-limitations-caveats

package.json
aaronccasanova: We may be fine here. I remember `Turbo` had a similar example where they referenced a custom eslint config workspace in the root package.json:
- [Root package.json referencing the custom eslint config workspace](https://github.com/vercel/turborepo/blob/5ad04a308d0d761e2bc2e1be76f16d6f7c4745f4/examples/basic/package.json#L15)
- [Root .eslintrc using the custom eslint config](https://github.com/vercel/turborepo/blob/5ad04a308d0d761e2bc2e1be76f16d6f7c4745f4/examples/basic/.eslintrc.js#L4)

sam-b-rose: Interesting... how'd they do it 🤯 hmm

PR #5797, opened 2022-05-12
Vsc build updates

yurm04:

* The entire monorepo was getting packaged up with the extension on `vsce package`
* Get ready for v0.2.0
* Bring up to speed with main

* updating the `.vscodeignore` to exclude everything outside of `polaris-for-vscode` and `.turbo`
* bump the version to `v0.2.0`
* merge `main`

### DIFF
--- a/polaris-for-vscode/.vscodeignore
+++ b/polaris-for-vscode/.vscodeignore
@@ -5,8 +5,6 @@
 **/*.ts
 *.vsix
 CONTRIBUTING.md
-docs
-!docs/polaris-icon.png
 scripts
 src
 node_modules
--- a/polaris-for-vscode/.vscodeignore
+++ b/polaris-for-vscode/.vscodeignore
@@ -1,9 +1,12 @@
+.turbo
+../**
 **/tsconfig.json
 **/*.map
 **/*.ts
 *.vsix
 CONTRIBUTING.md
 docs
+!docs/polaris-icon.png
 scripts
 src
 node_modules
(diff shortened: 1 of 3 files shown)


polaris-for-vscode/.vscodeignore
```diff
@@ -1,9 +1,12 @@
+.turbo
+../**
 **/tsconfig.json
 **/*.map
 **/*.ts
 *.vsix
 CONTRIBUTING.md
 docs
+!docs/polaris-icon.png
```
yurm04, line 9: this makes sure the icon image can get packaged correctly in the publish process

polaris-for-vscode/.vscodeignore
```diff
@@ -1,9 +1,10 @@
+.turbo
+../**
```
yurm04, line 2: for some reason the  `vsce` cli will package up the entire repo with the extension without this. I'm hoping this won't be necessary after @romellogoodman's PR to add rollup as a build tool #5802

aaronccasanova: Might be helpful to include a comment here since it's resolving a specific edge case. I don't know the ignore pattern off the top, but we could also try to make this self documenting with a negated pattern e.g. `ignore everything that's not polaris-for-vscode`
```
../**
!../polaris-for-vscode
```

alex-page (approved):

polaris-for-vscode/.vscodeignore
sam-b-rose: I did some digging here and seems like there still ins't official support for `vsce package` for yarn workspaces

This should be a problem if we are bundling `node_modules` with the package using webpack or rollup (I think) according to [this comment](https://github.com/microsoft/vscode-vsce/issues/300#issuecomment-697344682)

PR #5799, opened 2022-05-13
Clean `tsbuildinfo` and add common scripts to `polaris-for-figma`

sam-b-rose:

- The `tsbuildinfo` output should be cleared when cleaning the repo of build output (`yarn clean`)
- There are missing common tasks for the `polaris-for-figma` package (clean and lint)
 

- Add the `*.tsbuildinfo` files to the `clean` task where relevant
- Add lint and clean scripts to `polaris-for-figma`

Post-build | Post-clean 
--- | ---
before | after

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -70,6 +70,12 @@
   },
   "overrides": [
     {
+      "files": ["polaris-for-figma/src/**/*.{ts,tsx}"],
+      "rules": {
+        "@shopify/jsx-no-hardcoded-content": "off"
+      }
+    },
+    {
       "files": ["polaris-react/src/**/*.{ts,tsx}"],
       "extends": ["plugin:@shopify/typescript-type-checking"],
       "parserOptions": {
--- a/polaris-for-figma/package.json
+++ b/polaris-for-figma/package.json
@@ -3,7 +3,9 @@
   "version": "0.0.1",
   "scripts": {
     "build": "webpack --mode=production",
-    "dev": "webpack --mode=development --watch"
+    "dev": "webpack --mode=development --watch",
+    "lint": "TIMING=1 eslint --cache .",
+    "clean": "rm -rf .turbo node_modules dist"
   },
   "dependencies": {
     "@ant-design/icons": "^4.7.0",
--- a/polaris-for-figma/src/tokens/tokenScript.js
+++ b/polaris-for-figma/src/tokens/tokenScript.js
@@ -21,6 +21,7 @@ function addPaintStyleIDtoTokens() {
     JSON.stringify(apiColorTokens),
     (err) => {
       if (err) throw err;
+      // eslint-disable-next-line no-console
       console.log('Data written to file');
     },
   );
--- a/polaris-for-vscode/package.json
+++ b/polaris-for-vscode/package.json
@@ -33,7 +33,7 @@
     "build": "tsc -b",
     "dev": "tsc -b -w",
     "lint": "TIMING=1 eslint --cache .",
-    "clean": "rm -rf .turbo node_modules dist"
+    "clean": "rm -rf .turbo node_modules dist *.tsbuildinfo"
   },
   "dependencies": {
     "@shopify/polaris-tokens": "*",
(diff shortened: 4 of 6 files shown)


polaris-for-figma/tsconfig.json
```diff
@@ -1,3 +1,11 @@
 {
-  "extends": "@shopify/typescript-configs/dom"
+  "extends": "@shopify/typescript-configs/dom",
+  "compilerOptions": {
+    "composite": true,
+    "emitDeclarationOnly": true,
+    "resolveJsonModule": true,
+    "outDir": "dist",
+    "rootDir": "."
```
aaronccasanova: This appears to be the default with `composite: true`:
https://www.typescriptlang.org/tsconfig#rootDir

aaronccasanova: Now noticing `rootDir: "."` is set in every package. I'm thinking we should remove it since its normal behavior is overwritten. Thoughts?

polaris-tokens/tsconfig.json
```diff
@@ -4,8 +4,8 @@
     "composite": true,
     "declarationDir": "dist/types",
     "emitDeclarationOnly": true,
-    "outDir": "dist",
-    "rootDir": "src"
+    "outDir": "dist/types",
```
aaronccasanova: For my own understanding, does this then put the `*.tsbuildinfo` in `dist/types`?

polaris-for-figma/tsconfig.json
sam-b-rose: Yes, counter to my usual opinion of being implicit, I opted to be explicit for consistency with the other tsconfigs. Think we should just drop all `rootDir`s in favor of just using the default (requires people knowing that setting `composite: true` changes this behavior though)?

polaris-tokens/tsconfig.json
sam-b-rose: Yes, that's correct 👍  then we `yarn clean`, the `tsbuildinfo` is cleared as well

polaris-for-figma/tsconfig.json
aaronccasanova: > requires people knowing that setting composite: true changes this behavior though

It is also misleading. Setting `rootDir: "."` makes it appear we control this behavior where in fact it is doing nothing (If I'm understanding the docs correctly).

polaris-tokens/tsconfig.json
aaronccasanova: Awesome! So why does `tsbuildinfo` need to be emitted in `dist/types`? Is the behavior the same when `outDir` is `dist`?

polaris-tokens/tsconfig.json
```diff
@@ -4,8 +4,8 @@
...
+    "rootDir": "."
```
BPScott: I believe changing rootDir around messes with the folders structure generated within the outDir (I think it'll now have a src directory in it). If that's the case you'll need to update the `types` keys in package.json files

BPScott (commented):
Happy to see these make their way back into somewhere easily cleanable.

I believe changing rootDir around messes with the folders structure generated within the outDir (I think it'll now have a src directory in it). If that's the case you'll need to update the types keys in package.json files.

Talking of which, you'll also need to make sure that the tsbuildinfo files are excluded from package.json's files array ([like this](https://github.com/Shopify/polaris/blob/main/polaris-react/package.json#L19) so that they don't get included in published packages as they're pretty chonky.

sam-b-rose:
Looks like the output of the `*.tsbuildinfo` files is a bit trickier than I initially realized. Just going keep the tsconfigs the way they are and add the removal of `*.tsbuildinfo` to the `clean` task.

sam-b-rose:
Sorry for changing the scope here!

My original intent was to clean up project references, then that pivioted to cleaning up the `*.tsbuildinfo` files by placing them in the build output.

There were unforseen complications with both of those initiatives. Now, this PR is a just a simple addition of removing the TS build info output with the `clean` task. Also added the `clean` and `lint` tasks to the Figma package.

> Note: I tried adding `composite: true` to the tsconfig for Figma but ran into issue with ts-loader not being able to load `.json` files appropriately. We can revisit adding composite to this project later and then add it to the root project references tsconfig.

aaronccasanova (approved):
Updates look great 🚀

Issue #5678, opened 2022-05-02
Use a single bundler for all packages that require a build step


sam-b-rose:
Convert all packages to use the same build/bundler (Rollup w/ Babel):

- [ ] Use Rollup for `polaris-for-vscode`
- [x] Use Rollup for `polaris-react`
- [ ] Clean up `rollup.config.js` for `polaris-icons`
- [ ] Clean up/check `rollup.config.js` for `polaris-tokens`

PR #5802, opened 2022-05-13
Polaris for VS Code, bundle using rollup

romellogoodman:

As part of https://github.com/Shopify/polaris/issues/5678 this configures polaris-for-vscode to be built using rollup.

* Creates a rollup.config.js
* Updates vscode task problemMatcher to understand that rollup is a watch task

### DIFF
--- a/polaris-for-vscode/package.json
+++ b/polaris-for-vscode/package.json
@@ -36,12 +36,12 @@
     "clean": "rm -rf .turbo node_modules dist *.tsbuildinfo"
   },
   "dependencies": {
-    "@shopify/polaris-tokens": "*",
     "vscode-languageclient": "^8.0.1",
     "vscode-languageserver": "^8.0.1",
     "vscode-languageserver-textdocument": "^1.0.4"
   },
   "devDependencies": {
+    "@shopify/polaris-tokens": "*",
     "@types/glob": "^7.2.0",
     "@types/node": "14.x",
     "@types/vscode": "^1.64.0",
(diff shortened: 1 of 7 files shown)


polaris-for-vscode/package.json
```diff
@@ -30,23 +30,27 @@
...
-    "vscode-languageserver": "^7.0.0",
+    "vscode-languageclient": "^8.0.1",
+    "vscode-languageserver": "^8.0.1",
     "vscode-languageserver-textdocument": "^1.0.4"
   },
   "devDependencies": {
+    "@rollup/plugin-babel": "^5.3.1",
+    "@rollup/plugin-commonjs": "^21.1.0",
+    "@rollup/plugin-node-resolve": "^13.2.1",
+    "@shopify/polaris-tokens": "*",
```
aaronccasanova, line 47: @samrose3 I'm not familiar with the VS Code `client/server` architecture, but is `@shopify/polaris-tokens` not a regular dependency since it's used in `server.ts` (and has access to the `node_modules`)?

sam-b-rose: Normally yes, this would be in the `dependencies` list rather than `devDependencies`.  But, when using `vsce package`, there is an error not being able to find `@shopify/polaris-tokens` likely due to not being aware of it as part of the Yarn workspace. When using `vsce package --yarn`, the `node_modules` are omitted from being included. When trying to bundle `node_modules` using Rollup, we run into circular dependencies that causes failures in the `semver` package.

All that to say, this is the best route around all these edge-case issues. Maybe a lot of this can be resolved after migrating to pnpm 🤷 we'll see

aaronccasanova (approved):
Not tested but all the updates look sensible to me! 🚀

PR #5805, opened 2022-05-13
fix: Update config paths in stylelint-polaris tests

aaronccasanova:


lgriffee (approved):

sam-b-rose (approved):
good catch, ty for fixing!

PR #5810, opened 2022-05-13
Add useEventListener hook

kyledurand:

This is a nice simple event listener hook that takes all of the properties we cared about in `<EventListener />` and moves them into the hook. The tests are the same as well. 

This means we can eventually remove `<EventListener />`

From @aaronccasanova: 
* Components => encapsulate UI
* Hooks => encapsulate Behavior

I liked the idea so I made a quick hook. Then we can start using it in the grid component and start setting good examples of ☝️ 

Here's how to test that the listener gets removed:

https://user-images.githubusercontent.com/6844391/168386422-b87f0f04-ebd6-4898-8094-e62356bacbe2.mp4

<details>
  <summary>Playground to test unmount</summary>

```jsx
import React from 'react';

import {Page} from '../src';
import {useEventListener} from '../src/utilities/event-listener/hooks';

  const [show, setShow] = React.useState(false);
  return (
    <Page title="Playground">
      {show ? <WithEventListener /> : <div>not listening</div>}
      <button onClick={() => setShow(!show)}>toggle</button>
    </Page>
  );
}

const WithEventListener = () => {
  useEventListener('resize', () => console.log('resizing'));
  return <div>listening</div>;
};


```
</details>


alex-page (approved):
Should we deprecate `<EventListener>` in this PR? Any next steps coming out of this?

alex-page (commented):
Should we make a hooks directory? Or is utils a fine place for them?

kyledurand:
Deprecate the previous, yes I'll do that Monday

> Should we make a hooks directory? Or is utils a fine place for them?

That's what I was going to do at first but we have a mix of standard utils and hooks in this directory. So I ended up scrapping that folder and doing it all in utilities. If we did have a hooks folder we'd have to move the hooks out of utilities and that would separate the standard utils from the hooks which might be confusing

alex-page:
Happy to keep hooks in utils for now but might be worth exploring in the future.

polaris-react/src/components/EventListener/EventListener.tsx
```diff
@@ -10,7 +10,7 @@ export interface EventListenerProps extends BaseEventProps {
   passive?: boolean;
 }
 
-// see https://github.com/oliviertassinari/react-event-listener/
+/** @deprecated Use the useEventHandler hook instead. */
```
aaronccasanova:
```suggestion
/** @deprecated Use the useEventListener hook instead. */
```

polaris-react/src/utilities/event-listener/hooks.ts
```diff
@@ -0,0 +1,21 @@
+import {useEffect, useRef} from 'react';
+
+export type ElementType = Window | Document | HTMLElement;
+
+export function useEventListener(
+  event: string,
+  handler: () => void,
```
aaronccasanova:
```suggestion
  handler: (event: Event) => void,
```

While this isn't the most informative type it at least indicates an event parameter is present.

polaris-react/src/utilities/event-listener/hooks.ts
```diff
@@ -0,0 +1,21 @@
+import {useEffect, useRef} from 'react';
+
+export type ElementType = Window | Document | HTMLElement;
```
aaronccasanova:
```suggestion
type ElementType = Window | Document | HTMLElement;
```

Thoughts on holding off exporting this for now. I have a local POC enhancing the TypeScript behavior referring to these as `UseEventTarget`s (Matching the [EventTarget.addEventListener docs](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener))

polaris-react/src/utilities/event-listener/hooks.ts
```diff
@@ -0,0 +1,21 @@
...
+  options?: AddEventListenerOptions,
+) {
+  const handlerRef = useRef(handler);
+
+  useEffect(() => {
+    const handler = handlerRef.current;
+    element.addEventListener(event, handlerRef.current, options);
+
+    return () => {
+      element.removeEventListener(event, handler, options?.capture);
```
aaronccasanova:
```suggestion
      element.removeEventListener(event, handler, options);
```

Looks sufficient to pass the `options` to `removeEventListener`:
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#syntax

polaris-react/src/utilities/event-listener/hooks.ts
```diff
@@ -0,0 +1,21 @@
...
+    };
+  }, [event, element, options]);
```
aaronccasanova:
```suggestion
  }, [event, element]);
```

I believe we should also set this in a `ref` so an unmemorized object does re-run this effect each render.

polaris-react/src/utilities/event-listener/hooks.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export type ElementType = Window | Document | HTMLElement;
+
+export function useEventListener(
+  event: string,
+  handler: () => void,
+  element: ElementType = window,
+  options?: AddEventListenerOptions,
+) {
+  const handlerRef = useRef(handler);
```
aaronccasanova: We may need to add an effect to keep this ref up to date:
```tsx
const handlerRef = useRef(handler);

useEffect(() => {
  handlerRef.current = handler
}, [handler])

...
```

Same with `options`. You can learn more about this pattern [here](https://overreacted.io/making-setinterval-declarative-with-react-hooks/#:~:text=This%20useInterval%20isn%E2%80%99t%20a%20built%2Din%20React%20Hook%3B%20it%E2%80%99s%20a%20custom%20Hook%20that%20I%20wrote%3A)

polaris-react/src/utilities/event-listener/hooks.ts
kyledurand: I thought I tested with options and it didn't remove. I'll check again

polaris-react/src/utilities/event-listener/hooks.ts
kyledurand: Nice catch. I missed this from the previous implementation 👍

kyledurand:
Thanks for the suggestions @aaronccasanova I'm fairly happy with this now. Any regressions left that would conflict with your POC?

Edit: I also added support for passing in a ref object and move the files around to match other hooks and make this more easily searchable

polaris-react/src/utilities/use-event-listener.ts
```diff
@@ -0,0 +1,31 @@
...
+  element: ElementType = window,
+  options?: AddEventListenerOptions,
+) {
+  const handlerRef = useRef(handler);
+  const optionsRef = useRef(options);
+
+  useEffect(() => {
+    handlerRef.current = handler;
+    optionsRef.current = options;
+  }, [handler, options]);
```
aaronccasanova, line 17:
```suggestion
  const handlerRef = useRef(handler);
  const optionsRef = useRef(options);

  useEffect(() => {
    handlerRef.current = handler;
  }, [handler]);

  useEffect(() => {
    optionsRef.current = options;
  }, [options]);
```

Thoughts on separating `handlerRef` and `optionsRef` assignments. If `handler` was memoized and `options` wasn't we would end up reassigning `handler` every render.

polaris-react/src/utilities/use-event-listener.ts
```diff
@@ -0,0 +1,31 @@
+import {RefObject, useEffect, useRef} from 'react';
+
+type ElementType = Window | Document | HTMLElement | RefObject<HTMLElement>;
```
aaronccasanova:
```suggestion
type ElementType = Window | Document | HTMLElement;
```

Suggest removing as we can emulate the same behavior by passing in `ref.current` and also I was having type inferences issues using refs in my POC e.g.
```tsx
useEventListener('click', handler, myRef.current)
```

polaris-react/src/utilities/use-event-listener.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  useEffect(() => {
+    const target = 'current' in element ? element.current : element;
+
+    if (!target) return;
+
+    const handler = handlerRef.current;
+    const options = optionsRef.current;
+
+    target.addEventListener(event, handlerRef.current, options);
```
aaronccasanova:
```suggestion
    target.addEventListener(event, handler, options);
```

polaris-react/src/utilities/use-event-listener.ts
```diff
@@ -0,0 +1,31 @@
...
+    optionsRef.current = options;
+  }, [handler, options]);
+
+  useEffect(() => {
+    const target = 'current' in element ? element.current : element;
+
+    if (!target) return;
+
+    const handler = handlerRef.current;
+    const options = optionsRef.current;
```
aaronccasanova, line 21: Surprised your not getting a `no shadow vars` eslint error here (conflicting with the destructured props)

aaronccasanova (approved):
This looks great!! Left a few non-blocking comments after a final last pass though 🚀

polaris-react/src/utilities/use-event-listener.ts
kyledurand: These should be named `handlerRef` anyway to match our current patterns 👍

polaris-react/src/utilities/use-event-listener.ts
kyledurand: I think that's overkill personally. I'll leave it up to you in the follow up

PR #5303, opened 2022-03-17
[TextField] Add support for inline autocompletion

laurkim:
## To do:

While doing a final tophat in mobile, I found two issues that need to be resolved before shipping:
- [x] Selection styling and jank in mobile web (access and play with the playground on a mobile device to replicate)
- [x] Infinite loop (in Listbox hook?) when `loading` is true and list has been filtered (see Autocomplete or Combobox basic loading examples in the Chromatic)

### What does this pull request do?

- [x] Fixes #5151 by adding support for `aria-autocomplete="both"
- [x] Fixes #5149 by adding a `suggestion` prop that replaces the `value` when set to support inline autocompletion. The range of `suggestion` characters that don't match the `value` are given a selected with subdued styling.  `TextField` `selectTextOnFocus` and `suggestion` props are mutually exclusive, so only one or the other can be set.

This PR also unblocks [the final PR for changes needed to support the tag multi-select pattern](https://github.com/Shopify/polaris/pull/5392).

[Play in the playground](https://5d559397bae39100201eedc1-aiuwbusgwo.chromatic.com/?path=/story/playground-playground--playground) (playground will be reset before merging)

[Play with the focus/scroll fix when list is lazy loaded or long](https://5d559397bae39100201eedc1-aiuwbusgwo.chromatic.com/?path=/story/all-components-autocomplete--autocomplete-with-lazy-loading)

<details>
  <summary>Click to view demo GIF</summary>
  Typeahead demo
</details>

This PR also includes bug fixes and polish for automatic selection required for the `suggestion` prop to work as expected:
- [x] Refactored `Listbox` to reset the active option when the list values change
- [x] Fixed list navigation and styles when disabled options are present
- [x] Fixed `Listbox` `activeOption` state not updating when only an action option is active and updated
- [x] Fixed automatic selection when options are lazy loaded (see lazy load Autocomplete example)
- [x] Refactored `Listbox` scrolling
  - [x] Fixed option flashing when scrolled into view
  - [x] Fixed option jumping to top when scrolled into view
  - [x] Listbox now scrolls only enough to make option visible, accounting for the spacing above the first option as well as the loading spinner shown below the first option when lazy loading

---

##

### How to 🎩

🖥 [Local development instructions](https://github.com/Shopify/polaris-react/blob/master/README.md#development)
🗒 [General tophatting guidelines](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting.md)
[...]

PR #5812, opened 2022-05-14
[Listbox] Fix: preserve scroll position when lazy-loading additional items

### LINKED ISSUE #5809: ListBox scrolls to the top of the list when more options are loaded
## Issue summary

I'm trying to use the Polaris `Autocomplete` component to display a list of items. When the list is scrolled to the bottom, more items should be fetched from GraphQL and appended to the list. When the new items are added, however, the list scrolls to the current selection, which by default is the first element in the list.

You can see this behavior even on the code sandbox [example](https://codesandbox.io/s/k7kg1d?module=App.js&file=/App.js) for the `Autocomplete` component. When the list is scrolled to the bottom, more items get loaded, but the list resets to the top, losing the user's positioning.

You can view the discussion where we discovered exactly what was happening in Slack [here](https://shopify.slack.com/archives/C019H1T86D8/p1652294775113199).

## Expected behavior

I would expect, that even though the first element in the list is focused by default, that scrolling to the end and loading more item's shouldn't lose the scroll positioning. Instead, the new items should be appended to the bottom, allowing the user to continue scrolling from where they're at.

## Actual behavior

After scrolling to the end of the list and more items are loaded, the list resets the scroll position to the top of the list.

1. Use [this](https://codesandbox.io/s/k7kg1d?module=App.js&file=/App.js) code sandbox.
2. Open the autocomplete list; scroll to the bottom.

## Specifications

- Polaris version number: I believe its 9.6.0. Whatever is the version on the `web` repo.
- Browser: Chrome
- Device: MacBook
- Operating System: Mac

developit:
This fixes a regression in `@shopify/polaris` at version `9.4.0-rc.1`. Previously, appending additional items to a Listbox would preserve the current scroll position. Additional focus management was added (I believe in #5303) that may not have accounted for this case.

Fixes #5809 (see issue for details from @gcrevell)

It Listbox from being scrolled to the top via `scrollIntoViewIfNeeded` when items are appended to the current list of items. The focus behaviour for changing or removing items is unchanged.

#### Before:

https://user-images.githubusercontent.com/105127/168404342-befd7766-ee7c-4a36-81a1-76cc013e215f.mov

#### After:

https://user-images.githubusercontent.com/105127/168404393-65a1c3db-85dc-477b-abfc-8ff96c861709.mov

### DIFF
--- a/.changeset/curvy-rockets-provide.md
+++ b/.changeset/curvy-rockets-provide.md
@@ -1,0 +1,5 @@
+---
+'@shopify/polaris': minor
+---
+
+Fixed Listbox to preserve scroll position when lazy-loading additional items
--- a/.changeset/curvy-rockets-provide.md
+++ b/.changeset/curvy-rockets-provide.md
@@ -1,5 +1,5 @@
 ---
-'@shopify/polaris': minor
+'@shopify/polaris': patch
 ---
 
-Fixed Listbox to preserve scroll position when lazy-loading additional items
+Fixed `Listbox` to preserve scroll position when lazy-loading additional items
--- a/polaris-react/src/components/Listbox/Listbox.tsx
+++ b/polaris-react/src/components/Listbox/Listbox.tsx
@@ -238,19 +238,30 @@ export function Listbox({
       option.getAttribute(OPTION_VALUE_ATTRIBUTE),
     );
 
-    const listIsUnchangedOrAppended =
+    const listIsUnchanged =
+      nextValues.length === currentValues.length &&
+      nextValues.every((value, index) => {
+        return currentValues[index] === value;
+      });
+
+    const listIsAppended =
       currentValues.length !== 0 &&
-      nextValues.length >= currentValues.length &&
+      nextValues.length > currentValues.length &&
       currentValues.every((value, index) => {
         return nextValues[index] === value;
       });
 
-    if (listIsUnchangedOrAppended) {
+    if (listIsUnchanged) {
       if (optionIsAlreadyActive && actionContentHasUpdated) {
         setCurrentOptions(nextOptions);
         handleChangeActiveOption(nextOption);
       }
 
+      return;
+    }
+
+    if (listIsAppended) {
+      setCurrentOptions(nextOptions);
       return;
     }
 
--- a/polaris-react/src/components/Listbox/Listbox.tsx
+++ b/polaris-react/src/components/Listbox/Listbox.tsx
@@ -238,13 +238,14 @@ export function Listbox({
       option.getAttribute(OPTION_VALUE_ATTRIBUTE),
     );
 
-    const listIsUnchanged =
-      nextValues.length === currentValues.length &&
-      nextValues.every((value, index) => {
-        return currentValues[index] === value;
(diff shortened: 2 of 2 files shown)


ghost:

gcrevell:
I like the changes you made, and they make sense to me at least. Some tests would be nice though! Need to prevent regressions 😁

alex-page:
@developit thanks a ton for this contribution. Chloe and Lo are out this week and I would love their eyes on this.

Is there any urgency on getting this fixed? I am unsure of the complexity of adding a test but it would be greatly appreciated.

developit:
@alex-page hi - not urgent, was just a fix for an issue raised in Slack.

laurkim:
@developit fyi, since this PR was created, the Polaris team has moved away from entries in an `UNRELEASED.MD` file in favor of changesets. I created and pushed up the changeset in this [commit](https://github.com/Shopify/polaris/pull/5812/commits/a7a3d100634561102548bd4f8435a32ac4b23c8c)

laurkim (changes requested):
The code does fix the regression of the scroll position jumping to the top when lazy loading additional items, but breaks keyboard navigation where you cannot arrow up/down through the additional items. Video example below where after the next batch of items loads, I navigate down from first item to get to `Rustic 26` using arrow keys.

Happy to pair on this together if you'd like 👍 

https://user-images.githubusercontent.com/26749317/170118051-fa811f1e-0cfc-490f-bdac-32de6ed57c6b.mp4

developit:
I'm not an accessibility expert, but is the right thing to do here really to change the focused item when scrolling with the mouse wheel?

chloerice:
> I'm not an accessibility expert, but is the right thing to do here really to change the focused item when scrolling with the mouse wheel?

Hey @developit this is a known issue that is not priority at this time because autocomplete implementations should prioritize search over endless scroll. I don't have capacity right now, but I'll have a free cycle to dig into your PR in a couple of weeks (I think you're onto the right fix, it just currently breaks keyboard navigation so would be trading one bug for another and keyboard is higher priority in autocomplete UX).

developit:
Thanks @chloerice!

laurkim:
@developit I dug around a little more and it turns out the reason keyboard navigation wasn't working as expected was because the `currentOptions` weren't being updated to include the next set of options. When trying to fix it, it was causing an infinite loop of refreshing options so I had to separate out the logic for when the list is unchanged and when it's appended. I made a PR off your branch [here](https://github.com/Shopify/polaris/pull/6025).

cc @chloerice

.changeset/curvy-rockets-provide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/polaris': minor
+---
+
+Fixed Listbox to preserve scroll position when lazy-loading additional items
```
laurkim:
```suggestion
'@shopify/polaris': patch
---

Fixed `Listbox` to preserve scroll position when lazy-loading additional items
```

laurkim (approved):

chloerice (approved):
🚀

Issue #5467, opened 2022-04-07
[Polaris website refresh] Explore best way to author jsx/md content


martenbjork:
Our needs:

- Writing content
- Adding images 
- Adding videos
- Embedding Polaris examples
- Embedding arbitrary jsx
- ⚠️ Ability to extract content in a modular and context-aware way (to the API)

### mdx

👍 Theoretically mixes markdown and jsx seamlessly.
👎 Feels pretty hacky and fragile. I found it difficult to get eslint and prettier to play nice.

```tsx
# Title
Some markdown

<SomeComponent />

Some more markdown
```

### Plain tsx

👍 Native. Straight forward. Easy to mix HTML and React.
👎 Requires HTML knowledge. Less readable than markdown. Lots of room for divergence. 

```tsx
function Page() {
  return (
    <>
      <h1>Title</h1>
      <p>Just some text</p>

      <CustomComponent />
    </>
  )
}
```

### react-markdown
Using [react-markdown](https://github.com/remarkjs/react-markdown) to mix React and markdown content.

👍 Pretty straight forward. Easy to puzzle pieces together.
👎 If written inside the return function, Prettier will try to indent the markdown which breaks it. Therefore it needs to be pulled out into separate `const`s. A little messy.

```tsx
const intro = `
Welcome to the blog post!
`;

const end = `
That's all folks!
`;

function Page() {
  return (
    <>
      <ReactMarkdown>{intro}</ReactMarkdown>
      <CustomComponent />
      <ReactMarkdown>{end}</ReactMarkdown>
    </>
  );
}
```

We could iterate on this to abstract the internals away:

```tsx
const intro = `
Welcome to the blog post!
`;

const end = `
That's all folks!
`;

const Page = createPage(
  intro, 
  <CustomComponent />,
  end
);

export default Page;

```

alex-page:
```
---
specialComponent: special.md
---

# Title

Some markdown

Some more markdown
```

BPScott:
[...]

PR #5821, opened 2022-05-17
Add @markdoc/next.js to test rendering Markdown in Next.js site

sam-b-rose:
Huston, we have [Markdoc](https://markdoc.io/) 🧑‍🚀 🚀 

> Note: We may want to consider wrapping the `Page` component to provide layout for the markdown pages as well. See the [Next.js Layout section](https://markdoc.io/docs/nextjs#layouts).

| Getting Started page example |
| --- |
| Screen Shot 2022-05-16 at 7 22 20 PM |

### DIFF
--- a/polaris.shopify.com/next.config.js
+++ b/polaris.shopify.com/next.config.js
@@ -1,7 +1,9 @@
 /* eslint-disable require-await */
 /** @type {import('next').NextConfig} */
 
-const nextConfig = {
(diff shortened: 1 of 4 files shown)


martenbjork:
Interesting! 👀 Will do a deep dive into this in a few days when I have more time.

In the mean time, I'll add it to this issue so that we have covered all possible ways we can author content: https://github.com/Shopify/polaris/issues/5467

PR #5798, opened 2022-05-13
[Badge] 🐛 matches `new` and `info` Pip colors with the ones in `Badge`

bernbecht:
## What is the problem
The background color hue for `<Badge  />` with status  `info` and `new`  is different of when it is rendered as `<Badge.Pip />` 

#### status `info` 

|  `<Badge status='info' />`|  `<Badge.Pip status='info' />`    |
|---	|---	|
|  image |   image |

#### status `new` 

|  `<Badge status='new' />`|  `<Badge.Pip status='new' />`    |
|---	|---	|
|  image | image  |

#### When status `info` 

|  `<Badge status='info' />`|  `<Badge.Pip status='info' />`    |
|---	|---	|
|  image |  image |

#### When status `new` 

|  `<Badge status='new' />`|  `<Badge.Pip status='new' />`    |
|---	|---	|
|  image | image |

## Why it happens
Checking [Badge.scss](https://github.com/Shopify/polaris/blob/5a01bbdba2c7178a3ff3363e57436f5802921168/polaris-react/src/components/Badge/Badge.scss#L29) we see that the colors for badge and pip are inverted. 

The solution is to use the same colors Badge.scss uses for `info` and `new` status in Pip.scss

## How to tophat
<details>
<summary>Copy-paste this code in <code>playground/Playground.tsx</code>:</summary>

```jsx
import React from 'react';

import {Page, Stack, Badge} from '../src';

export function Playground() {
  return (
    <Page title="Pip">
      <Stack>
        <div>
          <Badge.Pip progress="incomplete" status="new" /> Incomplete & new
          <Badge.Pip progress="incomplete" status="info" /> Incomplete & info
        </div>
      </Stack>

      <Page title="Badges">
        <Stack>
          <Badge progress="incomplete" status="new">
            Incomplete & new
          </Badge>
          <Badge progress="incomplete" status="info">
            Incomplete & info
          </Badge>
        </Stack>
      </Page>
    </Page>
  );
}

```
</details>

It should render the following
image

PR #5840, opened 2022-05-19
[Badge] 🐛 makes `new` and `info` Pip colors the same as in `Badge`

bernbecht:
## !Important
This PR was previously reverted https://github.com/Shopify/polaris/pull/5798#issuecomment-1130313111 as it introduced breaking changes 😔

## What is the problem
The background color hue for `<Badge status ='new  />` and `<Badge status ='info'  />`  are different than the ones rendered as `<Badge.Pip status ='...' />`.  The colors should match independent of the component.

#### Current
|  `<Badge status='info' ... />`|  `<Badge.Pip status='info' ... />`    |
|---	|---	|
|  image |   image |

#### Expected
|  `<Badge status='info' ... />`|  `<Badge.Pip status='info'  ... />`    |
|---	|---	|
|  image |  image |

#### Current
|  `<Badge status='new' ... />`|  `<Badge.Pip status='new' ... `/>`    |
|---	|---	|
|  image | image  |

#### Expected
|  `<Badge status='new' ... />`|  `<Badge.Pip status='new' ... />`    |
|---	|---	|
|  image | image |

## Why the error happens
Checking [Badge.scss](https://github.com/Shopify/polaris/blob/5a01bbdba2c7178a3ff3363e57436f5802921168/polaris-react/src/components/Badge/Badge.scss#L29) we see that the colors for badge and pip are inverted. 

So part of the solution is to use the same colors Badge.scss uses for `info` and `new` status in Pip.scss

## How to tophat
<details>

```jsx
import React from 'react';

  return (
    <>
      <Page title="Pip">
        <Stack>
          <div>
            <Badge.Pip progress="incomplete" status="new" /> Incomplete & new
            <br />
            <Badge.Pip progress="incomplete" status="info" /> Incomplete & info
            <br />
            <Badge.Pip /> Default
          </div>
        </Stack>
      </Page>
      <Page title="Badges">
        <Stack>
          <Badge progress="incomplete" status="new">
            Incomplete & new
          </Badge>
          <Badge progress="incomplete" status="info">
            Incomplete & info
          </Badge>
          <Badge progress="complete">Default</Badge>
        </Stack>
      </Page>
    </>
  );
}

```
</details>

It should render the following
image

### DIFF
--- a/.changeset/unlucky-baboons-breathe.md
+++ b/.changeset/unlucky-baboons-breathe.md
@@ -1,0 +1,5 @@
+---
+'@shopify/polaris': patch
+---
+
+Fixed `Badge` and `Pip` having different background colors for `new` and `info` status ([#5840](https://github.com/Shopify/polaris/pull/5840))
--- a/polaris-react/src/components/Badge/components/Pip/Pip.scss
+++ b/polaris-react/src/components/Badge/components/Pip/Pip.scss
@@ -11,7 +11,7 @@
 }
 
 .statusInfo {
-  --pc-pip-color: var(--p-icon);
+  --pc-pip-color: var(--p-icon-highlight);
 }
 
 .statusSuccess {
@@ -19,7 +19,7 @@
 }
 
 .statusNew {
-  --pc-pip-color: var(--p-icon-highlight);
+  --pc-pip-color: var(--p-icon);
 }
 
 .statusAttention {


polaris-react/src/components/Badge/README.md
```diff
@@ -260,7 +260,9 @@ Use to indicate when a given task has been partially completed. For example, whe
 Use to indicate when a given task has been completed. For example, when merchants have fulfilled an order.
 
 ```jsx
-<Badge progress="complete">Fulfilled</Badge>
+<Badge progress="complete" status="new">
+  Fulfilled
+</Badge>
```
bernbecht: To avoid the situation above where there's a mismatch between pip hue and badge's background 

**Code**
`<Badge progress='complete'>Fulfilled</Badge>`

**Result**
Screen Shot 2022-05-19 at 10 22 43

bernbecht: Btw, shouldn't bagdes which with defined `progress` status but no status have the same color by default?

emma-boardman: > To avoid the situation above where there's a mismatch between pip hue and badge's background

Does this mean consumer's previously passing in `progress='complete'` will need to pass in `status=new` to achieve the same effect?

emma-boardman (commented):
Colour swap looks good 👍 

Is it worth adding the undocumented variants to the documentation? That will make it easier to catch any mismatches in the storybook tophat. 

For my own curiosity, why does the Pip need a default of status=info, but not the Badge itself?

polaris-react/src/components/Badge/README.md
emma-boardman: > Btw, shouldn't bagdes which with defined `progress` status but no status have the same color by default?

I think so, yes. If status isn't defined, the background colour should be surface neutral: https://github.com/Shopify/polaris/blob/866ae314cba596af78cef2e32bec929c20b65048/polaris-react/src/components/Badge/Badge.scss#L7

bernbecht: To get the result bellow `status=new` is necessary. Other wise, Pip will be `status= info` which is blue
image

There are 3 alternatives to render the default  `<Badge `progress='complete' />` with bkg and pip as gray:
1)  Make `status='new'` the default in Badge's props
2) Make `status='new'` the default  in Pip's props
3) Remove default `status` in Pip's props

bernbecht: Gotcha. However it doesn't happen right now because pip will be `info` (blue hue)

bernbecht:
> For my own curiosity, why does the Pip need a default of status=info, but not the Badge itself?

My memory can't record very well but I think Pip needed it follow the Badges's default gray background. But checking its .scss the default status in Pip  isn't needed anymore. That's why I was pushing before to get rid of the default status there

sarahill (approved):
✅

polaris-react/src/components/Badge/README.md
emma-boardman: > To get the result bellow status=new is necessary. 

Is that not a breaking change? 🤔

bernbecht: It could be 🤦 even though the component shouldn't have 2 different colors

bernbecht:
Had to move this PR as `breaking change`. So, please read the `Why this PR is a breaking change` section in PR's description.

@aaronccasanova could you please guide me on the process of getting it merged?

emma-boardman (approved):

sarahill (approved):
✅

chloerice:
Hey @Berhell 👋🏽 The `new` badge/pip should be the only one that's blue, `info` is the default and should have no color (grayscale).  I've been out of office for a few weeks, is this part of an admin quality redesign I may have missed? The blue pip will add a lot of noise to completed orders on the list view.

bernbecht:
> Hey @Berhell 👋🏽 The `new` badge/pip should be the only one that's blue, `info` is the default and should have no color (grayscale). I've been out of office for a few weeks, is this part of an admin quality redesign I may have missed? The blue pip will add a lot of noise to completed orders on the list view.

Great that you're back @chloerice. You can explain what's the expected behavior then ☺️

There's no re-design just trying to make `Badge.Pip` and `Badge` match their colors when used separately.

When you say that only  the`new` should be blue, do you mean like this?
image

or like this?
image

chloerice:
> > Hey @Berhell 👋🏽 The `new` badge/pip should be the only one that's blue, `info` is the default and should have no color (grayscale). I've been out of office for a few weeks, is this part of an admin quality redesign I may have missed? The blue pip will add a lot of noise to completed orders on the list view.
> 
> Great that you're back @chloerice. You can explain what's the expected behavior then ☺️
> 
> There's no re-design just trying to make `Badge.Pip` and `Badge` match their colors when used separately.
> 
> When you say that only the`new` should be blue, do you mean like this? image
> 
> or like this? image

`new` should have blue background + blue pip, and `info` should gray background + gray pip.

alex-page:
@Berhell can you ship this without the breaking change and do the breaking change to default in a follow up?

bernbecht:
> `new` should have blue background + blue pip, and `info` should gray background + gray pip.

Thanks for confirming the behavior, @chloerice.

Reading the code for `<Badge>`, it seems that `info` is the blue one and `new` is the gray. And it looks like this happens since v6 https://github.com/Shopify/polaris/pull/3743/commits/9f370cdf057a120f4c7cd4b99dd446db216d8a48 🤔 

If we confirm that there's an error with `Badge` since v6, then `<Badge.Pip>` code was already correct and the fix should happen in `<Badge>`

bernbecht:
> @Berhell can you ship this without the breaking change and do the breaking change to default in a follow up?

Sure thing @alex-page

polaris-react/src/components/Badge/components/Pip/Pip.tsx
```diff
@@ -15,7 +15,7 @@ export interface PipProps {
 }
 
 export function Pip({
-  status = 'info',
```
alex-page: If we leave this as `status = 'info'` this can be shipped as a non breaking change.

.changeset/shy-sheep-camp.md
```diff
@@ -0,0 +1,7 @@
+---
+'@shopify/polaris': major
+---
+
+Fixed `Badge` and `Pip` having different background colors for `new` and `info` status ([#5798](https://github.com/Shopify/polaris/pull/5840))
+
+Removed default `status` prop in `Pip`. That caused a color mismatch between `Badge` background and progress indicator in situations where `Badge` has `progress` but no informed `status`. For `Badge.Pip`, if users don't inform the `status` prop in, its color is gray ([#5798](https://github.com/Shopify/polaris/pull/5840))
```
alex-page: Feel free to delete this changeset and make a `minor` one.

chloerice:
> > `new` should have blue background + blue pip, and `info` should gray background + gray pip.
> 
> Thanks for confirming the behavior, @chloerice.
> 
> Reading the code for `<Badge>`, it seems that `info` is the blue one and `new` is the gray. And it looks like this happens since v6 [9f370cd](https://github.com/Shopify/polaris/commit/9f370cdf057a120f4c7cd4b99dd446db216d8a48) 🤔
> 
> If we confirm that there's an error with `Badge` since v6, then `<Badge.Pip>` code was already correct and the fix should happen in `<Badge>`

Hey @Berhell, thanks for digging deeper into this and so sorry for the confusion!! It's been a while since the design language and color system changed and I totally spaced on the weirdness/relics in `Badge` 🤦🏽‍♀️

Back in the "purpler" days, our color system wasn't semantic and we had a wider range and different use of colors. There used to be a "new" variant that was teal, and info was blue. But when we moved to the new design language, we paired back our use of color to be narrower and more intentional, so the concept of "new" and "info" merged into the semantic color role  "highlight" (teal) and blue was given a role of interactive. So while "info" badge moved from light blue to teal, the "new" badge got paired back to gray. This was a little contentious because that change made "new" badges not stand out at all, so teal got brought back into them with the pip that in that special case is used to call attention subtly instead of denoting progress.

All that to say, ignore me and carry on as you were 😅

bernbecht:
@alex-page all set 🙌

alex-page (approved):

PR #5844, opened 2022-05-19
Make Polaris for VS Code and Figma private to prevent publishing

sam-b-rose:


.github/workflows/unpublish.yml
```diff
@@ -0,0 +1,22 @@
...
+      - name: Setup Node with v16.13.0
+        uses: actions/setup-node@v2
+        with:
+          node-version: 16.13.0
+
+      - name: Unpublish Polaris for VS Code
+        run: npm unpublish polaris-for-vscode
+
+      - name: Unpublish Polaris for Figma
+        run: npm unpublish polaris-for-figma
```
BPScott, line 22:
```suggestion
        run: npm unpublish polaris-for-figma
        env:
          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```

polaris-for-figma/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "polaris-for-figma",
```
BPScott, line 2:
```suggestion
  "name": "@shopify/polaris-for-figma",
```

Let's stick this in the shopify namespace so in the event it happens again, at least it is part of the org and org admins can unpublish it rather than having to resort to gh actions shennanigans.

.github/workflows/unpublish.yml
```diff
@@ -0,0 +1,26 @@
+name: Unpublish
+
+on: [pull_request]
```
BPScott: This should be triggered manually, not automatically triggered as part in response to any event

```
on: [workflow_dispatch]
```


https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow#configuring-a-workflow-to-run-manually

polaris-for-vscode/package.json
```diff
@@ -14,6 +14,7 @@
     "url": "https://github.com/Shopify/polaris/issues"
   },
   "version": "0.2.0",
+  "private": true,
```
BPScott, line 17: Let's update the name of this package to `@shopify/polaris-for-vscode`

Let's stick this in the shopify namespace so in the event it happens again, at least it is part of the org and org admins can unpublish it rather than having to resort to gh actions shennanigans.

.github/workflows/unpublish.yml
```diff
@@ -0,0 +1,26 @@
...
+
+      - name: Setup Node with v16.13.0
+        uses: actions/setup-node@v2
+        with:
+          node-version: 16.13.0
+
+      - name: Unpublish Polaris for VS Code
+        run: npm unpublish polaris-for-vscode --force
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
BPScott:
```suggestion
          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```

To match what is in release.yml

.github/workflows/unpublish.yml
```diff
@@ -0,0 +1,26 @@
...
+
+      - name: Unpublish Polaris for Figma
+        run: npm unpublish polaris-for-figma --force
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
BPScott:
```suggestion
          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```

To match what is in release.yml

.github/workflows/unpublish.yml
sam-b-rose: This actually should be as-is, using setup-node action here. Not the changesets action as used in `release.yml`

https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#publish-to-npmjs-and-gpr-with-npm

.github/workflows/unpublish.yml
sam-b-rose: Good call, ty!

alex-page (approved):

BPScott (approved):

PR #5865, opened 2022-05-19
Add the underline to removeUnderline on :hover

### LINKED ISSUE #4377: [Link] removeUnderline prop should still show on :hover
> **tl;dr:** The `Link` component should actually still render a "underline" on `:hover` and `:active` in order to align with [how Polaris describes these interactions](https://polaris.shopify.com/design/interaction-states).

[This conversation took place over Slack](https://shopify.slack.com/archives/C4Y8N30KD/p1628617714099300), but here is a recap:

---

@beefchimi 

Can anyone help shed some light on when a “interactive element pointing to a url” _(in other words a link or link-like element)_ should render with a text-underline? I am slightly confused because I see the recommendation here:
https://polaris.shopify.com/design/interaction-states

But then opposite results between Button and Link components:
https://codesandbox.io/s/polaris-button-or-link-1me4y?file=/App.js

@alex-page 

Underlines help users know that links are interactive. They can easily glance at a page and figure out what to press. However when there is a list of links the underlines can create noise and friction with borders/underlines appearing in multiple places. Repeated patterns where users can determine their interactivity in other ways do not require an underline.

The general guidance is links in paragraphs definitely should have underlines. Links in UI depending on the surrounding elements should or should not have underlines.

@beefchimi 

hmm, thats not the most straight forward rule to follow, but I certainly understand the rationale. Im not sure it explains why `Button > plain` behaves the way it does - is there any additional rule for underlines + `hover/active` states?

In other words - if we like how `Button > plain` works, should `Link > removeUnderline` behave the same way _(shows an underline once `:hover`)_

@alex-page 

Yeah they should work the same. That is a bug.

alex-page:
Fixes https://github.com/Shopify/polaris/issues/4377

This gives `<Link removeUnderline />` an underline when hovered so that users do not have to perceive color to use the user interface.

### DIFF
--- a/.changeset/good-knives-taste.md
+++ b/.changeset/good-knives-taste.md
@@ -2,4 +2,4 @@
 '@shopify/polaris': minor
 ---
 
-Add underline to Link on hover when removeUnderline property is used
+Added underline to Link on hover when removeUnderline property is used
--- a/.changeset/good-knives-taste.md
+++ b/.changeset/good-knives-taste.md
@@ -1,0 +1,5 @@
+---
+'@shopify/polaris': minor
+---
+
+Add underline to Link on hover when removeUnderline property is used
--- a/polaris-react/src/components/Link/Link.scss
+++ b/polaris-react/src/components/Link/Link.scss
@@ -76,4 +76,8 @@
 
 .removeUnderline {
   text-decoration: none;
+
+  &:hover {
+    text-decoration: underline;
+  }
 }


.changeset/good-knives-taste.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/polaris': minor
+---
+
+Add underline to Link on hover when removeUnderline property is used
```
aaronccasanova:
```suggestion
Added underline to Link on hover when removeUnderline property is used
```

Are still we using past tense changelog entries?

aaronccasanova (approved):

lgriffee (approved):

PR #5874, opened 2022-05-20
[Modal] Allow close on `Backdrop` click or tap

### LINKED ISSUE #5834: Modal should close by clicking on background
## Issue summary

The Polaris modal doesn't close when clicking on the background and our guidance recommends against clicking the background to close. 

## Expected behavior

Merchants should be able to close the modal by either selecting the close button, pressing the escape key, and clicking the background. 
slack-imgs-1

## Done when

The modal behavior has been updated
The style guide guidance has been updated

kyledurand:

Fixes https://github.com/Shopify/polaris/issues/5834

- Passes `onClose` to the `onClick` handler of `Backdrop`
- Allows pointer events through the transparent section of the `Dialog` container
- Recaptures pointer events on `Dialog` content

Alternative approach:
- Refactor so Dialog renders the backdrop
- Pass a new prop to dialog. Something like `onBackdropClick`

Second approach might be more future proof but I'm not sure it's possible or the implications of that change. I'll explore when I have some time

### How to 🎩

Checkout any of the Modal stories in storybook
I also used ngrok to check the tap event on my phone

### 🎩 checklist

### DIFF
--- a/polaris-react/src/components/Modal/components/Dialog/Dialog.scss
+++ b/polaris-react/src/components/Modal/components/Dialog/Dialog.scss
@@ -19,10 +19,17 @@ $dangerous-magic-space-16: 64px;
   display: flex;
   flex-direction: column;
   justify-content: flex-end;
+  // Allow passthrough of click events to dismiss modal on backdrop click
+  pointer-events: none;
 
   @include breakpoint-after($layout-width-page-with-nav-base) {
     justify-content: center;
   }
+}
+
+.Dialog {
+  // Resume pointer events
+  pointer-events: initial;
 }
 
 .Dialog:focus {
--- a/polaris-react/src/components/Modal/components/Dialog/Dialog.scss
+++ b/polaris-react/src/components/Modal/components/Dialog/Dialog.scss
@@ -20,6 +20,7 @@ $dangerous-magic-space-16: 64px;
   flex-direction: column;
   justify-content: flex-end;
   // Allow passthrough of click events to dismiss modal on backdrop click
+  // Resume pointer-events on .Modal below
   pointer-events: none;
 
   @include breakpoint-after($layout-width-page-with-nav-base) {
@@ -32,8 +33,6 @@ $dangerous-magic-space-16: 64px;
 }
 
 .Modal {
-  --pc-dialog-horizontal-spacing: var(--p-space-16);
-  // Resume pointer events
   pointer-events: initial;
   position: fixed;
   right: 0;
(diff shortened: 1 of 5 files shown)


polaris-react/src/components/Modal/components/Dialog/Dialog.scss
```diff
@@ -19,12 +19,19 @@ $dangerous-magic-space-16: 64px;
...
 
   @include breakpoint-after($layout-width-page-with-nav-base) {
     justify-content: center;
   }
 }
 
+.Dialog {
+  // Resume pointer events
+  pointer-events: initial;
+}
```
gui-santos: I was testing these changes in Storybook and they work 🙌 
but for some reason, in the `Modal with scroll listener`, the click is on the Backdrop is not working
Well... it works, but only if you click on the margin space at the top and at the bottom, I made a video:

https://user-images.githubusercontent.com/6597467/170225114-9b2e7878-5895-42ed-a004-63dbffb05f64.mp4

---

I checked the CSS and it seems that this code here is the culprit, but I didn't understand why this is only a problem in this specific example. I bet you have more context to find out why

Here is the `pointer-events: none` you added to the `Container` class: 

scrollable-modal-pointer-events-1


... And here is the child with `pointer-events: initial` that is making the click not work:

scrollable-modal-pointer-events-2

But again this is only a problem in this particular example 🤷 
Do you have an idea of why this is happening?

gui-santos (commented):
Hey! This looks great 👌 

I only found an issue in the `Modal with scroll listener` example. I commented the code with a video showing the issue

yesenia-perezcruz:
Hey @kyledurand - I'm looking at this Storybook link and I don't see the changes. Is there another link I should use? https://storybook.polaris.shopify.com/?path=/story/all-components-modal--basic-modal

alex-page:
@yesenia-perezcruz this link should work https://5d559397bae39100201eedc1-loervswvkg.chromatic.com/?path=/story/all-components-modal--basic-modal

That storybook URL is the production link of what is shipped.

yesenia-perezcruz (approved):
This looks good!

polaris-react/src/components/Modal/components/Dialog/Dialog.scss
kyledurand: Thank Gui! I think you caught an important bug. Looks like `.Dialog-Dialog` can span the full width of the page in some cases. I'm going to try resuming pointer-events on `.Dialog-Modal`. Which seems to just wrap the content

kyledurand:
There's an important note from @jjgali in the issue [here](https://github.com/Shopify/polaris/issues/5834#issuecomment-1136166872) that I think we should address before shipping this cc @alex-page @yesenia-perezcruz @gui-santos

alex-page:
@kyledurand @jjgali it's a valid concern. From my understanding and the current implementation all Polaris modals have a close icon, escape key close and a cancel button. Due to this the risk is quite low as all modals in the admin built with Polaris assume that the user can close them. I think we should ship this to make all Polaris Modals closable with the background press.

Blocking the close functionality feels like a new feature where we might even want to explore other options like the escape key and close icon. I would love to do a follow up exploration of a couple of modal patterns. I think that is out of scope for this PR.

alex-page (approved):

gui-santos (approved):

PR #5876, opened 2022-05-20
Update TextField docs

aveline:

Update `TextField` docs to encourage properly setting `inputMode` prop so that the best virtual keyboard is presented to users on mobile devices.

Screen Shot 2022-05-27 at 9 26 54 AM

Image added in https://github.com/Shopify/polaris-styleguide/pull/5840


alex-page (approved):
Thanks @aveline

PR #4355, opened 2021-08-03
[ActionList] Make spacing consistent

AndrewMusgrave:
### WHY are these changes introduced?

Fixes linked issue (see mentions below)

### WHAT is this pull request doing?

Creating consistent spacing between the list, items, sections & titles.

### Screenies

Image 2021-08-03 at 3 22 07 PM
Image 2021-08-03 at 3 30 05 PM

### How to 🎩

 Code below displaying a bunch of different combinations between sections/items/section titles

<details>
<summary>Copy-paste this code in <code>playground/Playground.tsx</code>:</summary>

```jsx
import React, {useState, useCallback} from 'react';
import {ImportMinor, ExportMinor} from '@shopify/polaris-icons';

import {Page, Popover, Button, ActionList} from '../src';

const actionListProps = [
  {
    sections: [
      {
        title: 'File options',
        items: [
          {content: 'Import file', icon: ImportMinor},
          {content: 'Export file', icon: ExportMinor},
        ],
      },
      {
        title: 'File options',
        items: [
          {content: 'Import file', icon: ImportMinor},
          {content: 'Export file', icon: ExportMinor},
        ],
      },
      {
        items: [
          {content: 'Import file', icon: ImportMinor},
          {content: 'Export file', icon: ExportMinor},
        ],
      },
      {
        title: 'File options',
        items: [
          {content: 'Import file', icon: ImportMinor},
          {content: 'Export file', icon: ExportMinor},
        ],
      },
      {
        items: [
          {content: 'Import file', icon: ImportMinor},
          {content: 'Export file', icon: ExportMinor},
        ],
      },
    ],
  },
  {
    sections: [
      {
        items: [
          {content: 'Import file', icon: ImportMinor},
          {content: 'Export file', icon: ExportMinor},
        ],
      },
      {
        title: 'File options',
        items: [
          {content: 'Import file', icon: ImportMinor},
          {content: 'Export file', icon: ExportMinor},
        ],
      },
      {
        items: [
          {content: 'Import file', icon: ImportMinor},
          {content: 'Export file', icon: ExportMinor},
        ],
      },
      {
        title: '',
        items: [
          {content: 'Import file', icon: ImportMinor},
          {content: 'Export file', icon: ExportMinor},
        ],
      },
      {
        title: 'Test',
        items: [
          {content: 'Import file', icon: ImportMinor},
          {content: 'Export file', icon: ExportMinor},
        ],
      },
    ],
  },
  {
[...]

PR #5879, opened 2022-05-20
[ActionList] Fix spacing above action list title when first in mobile action menu popover

### LINKED ISSUE #5877: [Page] Responsive: Action group title is too close to the top when there are no secondary actions
## Issue summary

When a `Page` has no `secondaryActions`, but it does have have an `actionGroup` with a `title`, the title of that action group appears too close to the top of the popover menu on mobile.

## Expected behavior

There should be more space above the action group title, as when there are secondary actions like in the screenshot below.

Screen Shot 2022-05-20 at 4 48 11 PM

## Actual behavior

There is very little space above the action group title when there are no secondary actions.

Screen Shot 2022-05-20 at 4 49 42 PM

Create a `Page` that has at least one action group with a title, but does not have any secondary actions. View on mobile and check the spacing above the first title.

## Reduced test case

https://codesandbox.io/s/wandering-cherry-9sl4qe?file=/App.js

Click the "toggle responsive preview" icon to view the mobile size and click the ellipsis to open the menu to see the issue.

## Specifications

Output of `npx envinfo --system --binaries --browsers --npmPackages react,react-dom,@shopify/polaris`:

Paste the results here:

```bash
  System:
    OS: Linux 5.10 Ubuntu 21.10 21.10 (Impish Indri)
    CPU: (32) x64 Intel(R) Xeon(R) CPU @ 2.80GHz
    Memory: 65.43 GB / 125.90 GB
    Container: Yes
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.15.0 - ~/.nvm/versions/node/v16.15.0/bin/node
    Yarn: 1.22.15 - /usr/local/bin/yarn
    npm: 8.5.5 - ~/.nvm/versions/node/v16.15.0/bin/npm
    Watchman: 4.9.0 - /usr/bin/watchman
  npmPackages:
    @shopify/polaris: 9.7.0 => 9.7.0
    react: ~17.0.2 => 17.0.2
    react-dom: ~17.0.2 => 17.0.2
```

rmleg:

Fixes https://github.com/Shopify/polaris/issues/5877

When a page has action groups but no secondary actions, there is not enough spacing above the title of the first action group in the action list popover on mobile.

I can't reproduce the problem in an `ActionList` in isolation; it only seems to be present in a `Page`.

There was a [previous PR](https://github.com/Shopify/polaris/pull/4355) that fixed this last August by adding a `firstSection` class to the first section of the action list using the `index` when mapping across `finalSections`. However this solution is not reliable as-is because the item at the index of `0` is not always rendered.

Instead of fixing the logic, I decided to go for a CSS solution.

First, set the `padding-top` for every action list title to the increased amount.

Then, select all titles that are in sections that are not the first section and set their `padding-top` to the smaller amount.

I had originally done this by just selecting `.Section:first-child .Title` and setting its `padding-top` to the increased amount, but that broke the case where an action list is not sectioned (has only one item in its `actionGroups` array).

#### Before this PR

The first section has a title which does not have enough spacing above it.

The first section does not have title and all spacing is correct.

#### After this PR

The first section has a title and it has enough spacing above it.

The first section does not have a title and all spacing is correct. There is no change in this scenario after this PR.

### How to 🎩

This can be tophatted by viewing the `Page/Page with action groups` story in Storybook at a mobile screen size and expanding the action list menu.

You can also verify this does not break the `ActionList` component in isolation by viewing the `Action list/Action list with destructive item` and `Action list/Sectioned action list` stories.

</details>

### 🎩 checklist

### DIFF
--- a/polaris-react/src/components/ActionList/components/Section/Section.tsx
+++ b/polaris-react/src/components/ActionList/components/Section/Section.tsx
@@ -1,6 +1,5 @@
 import React from 'react';
 
-import {classNames} from '../../../../utilities/css';
 import {Item} from '../Item';
 import type {
   ActionListItemDescriptor,
--- a/polaris-react/src/components/ActionList/components/Section/Section.tsx
+++ b/polaris-react/src/components/ActionList/components/Section/Section.tsx
@@ -15,8 +15,6 @@ export interface SectionProps {
   hasMultipleSections: boolean;
   /** Defines a specific role attribute for each action in the list */
   actionRole?: 'option' | 'menuitem' | string;
-  /** Whether or not the section is the first to appear */
-  firstSection?: boolean;
   /** Callback when any item is clicked or keypressed */
   onActionAnyItem?: ActionListItemDescriptor['onAction'];
 }
@@ -25,7 +23,6 @@ export function Section({
   section,
   hasMultipleSections,
   actionRole,
-  firstSection,
   onActionAnyItem,
 }: SectionProps) {
   const handleAction = (itemOnAction: ActionListItemDescriptor['onAction']) => {
@@ -54,10 +51,7 @@ export function Section({
   );
 
   const className = section.title ? undefined : styles['Section-withoutTitle'];
-  const titleClassName = classNames(
-    styles.Title,
-    firstSection && styles.firstSectionWithTitle,
-  );
+  const titleClassName = classNames(styles.Title);
 
   const titleMarkup = section.title ? (
     <p className={titleClassName}>{section.title}</p>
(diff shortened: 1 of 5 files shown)


polaris-react/src/components/ActionList/components/Section/Section.tsx
```diff
@@ -54,10 +51,7 @@ export function Section({
   );
 
   const className = section.title ? undefined : styles['Section-withoutTitle'];
-  const titleClassName = classNames(
-    styles.Title,
-    firstSection && styles.firstSectionWithTitle,
-  );
+  const titleClassName = classNames(styles.Title);
```
laurkim: nit: you can get rid of this and reference it directly on line 57
`<p className={styles.Title}>...`

laurkim (approved):
💯

polaris-react/src/components/ActionList/components/Section/Section.tsx
rmleg: Good catch! updated

PR #5893, opened 2022-05-24
Polaris for Figma, fix bundler issues

romellogoodman:

Fixes #0000 

Since polaris-for-figma was integrated into the monorepo infrastructure this PR updates parts of its config to work in the Figma environment again. One thing to note, we can't switch over to rollup at the moment because we are compiling two files and rollup won't allow us to use the `iife` or `umd` output formats as a result.

* Renames main figma files from `ui.html` and `code.js` to `index.html` and `plugin.js`
* Adds es5 compilerOption target for the figma environment
* Cleans up webpack.config.js

### How to 🎩

* Clone branch locally
* Run `yarn workspace polaris-for-figma dev`
* Open up Figma
* Plugins > Development >  Import plugin from manifest
* Select `polaris-for-figma/manifest.json`
* Confirm the plugin loads


BPScott:
> One thing to note, we can't switch over to rollup at the moment because we are compiling two files and rollup won't allow us to use the iife or umd output formats as a result.

The fix for this is to have a rollup configuration that contains two configs - with each one producing one file. This is done by returning an array of objects from the rollup config's default export instead of a single object. See polaris-icons for a demo of this https://github.com/Shopify/polaris/blob/main/polaris-icons/rollup.config.js#L206

---

See https://rollupjs.org/guide/en/#configuration-files

> You can export an array from your config file to build bundles from several unrelated inputs at once, even in watch mode. To build different bundles with the same input, you supply an array of output options for each input:

romellogoodman:
@BPScott, nice! Started to reopen the exploration here and will handle the full conversion in another PR https://github.com/Shopify/polaris/compare/figma-rollup-convert?expand=1

sam-b-rose (approved):
Built and loaded locally 💯

PR #5912, opened 2022-05-24
Add snapshot release workflow triggered by /snapshot-release

sam-b-rose:
Trigger a [snapshot release](https://github.com/changesets/changesets/blob/main/docs/snapshot-releases.md) using a slash command in your feature PR (`/snapshot-release`)

### DIFF
--- a/.github/workflows/slash-command.yml
+++ b/.github/workflows/slash-command.yml
@@ -1,0 +1,19 @@
+name: Slash Command
+
+on:
+  issue_comment:
+    types:
+      - created
+
+jobs:
+  slashCommand:
+    name: Dispatch
+    runs-on: ubuntu-latest
+    steps:
+      - name: Slash Command Dispatch
+        uses: peter-evans/slash-command-dispatch@v3
+        with:
+          token: ${{ secrets.PAT }}
+          permission: admin
+          issue-type: pull-request
+          commands: snapshot-release
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -31,11 +31,14 @@ jobs:
           git reset --hard ${{ github.event.client_payload.pull_request.head.ref }}
 
       - name: Create release Pull Request or publish to NPM
-        id: changesets
-        uses: changesets/action@v1
-        with:
-          version: yarn version-packages --snapshot snapshot
-          publish: yarn release --no-git-tags --snapshot
-        env:
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: echo "Creating release via Changesets..."
+
+      # - name: Create release Pull Request or publish to NPM
+      #   id: changesets
+      #   uses: changesets/action@v1
+      #   with:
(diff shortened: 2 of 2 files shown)


.github/workflows/slash-command.yml
```diff
@@ -0,0 +1,19 @@
...
+    name: Dispatch
+    runs-on: ubuntu-latest
+    steps:
+      uses: peter-evans/slash-command-dispatch@v3
+      with:
+        token: ${{ secrets.PAT }}
+        permission: admin
+        issue-type: pull-request
+        commands: |
+          snapshot
```
alex-page:
```suggestion
        commands: snapshot
```

.github/workflows/slash-command.yml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+  slashCommand:
+    name: Dispatch
+    runs-on: ubuntu-latest
+    steps:
+      uses: peter-evans/slash-command-dispatch@v3
+      with:
+        token: ${{ secrets.PAT }}
+        permission: admin
```
alex-page, line 16:
```suggestion
        permission: admin
```

Are we only allowing `admin` to do snapshot releases? I think any Shopifolk should be able to use it.

.github/workflows/snapshot.yml
```diff
@@ -23,21 +23,22 @@ jobs:
...
+          git checkout -b snapshot
+          git reset --hard ${{ github.ref }}
 
       - name: Create release Pull Request or publish to NPM
         id: changesets
         uses: changesets/action@v1
         with:
-          publish: yarn release
+          version: yarn version-packages --snapshot snapshot
+          publish: yarn release --no-git-tags --snapshot
```
aaronccasanova:
```suggestion
          version: yarn version-packages -- --snapshot snapshot
          publish: yarn release -- --no-git-tags --snapshot
```

Do we need to add option terminators to ensure the additional flags are passed to the `changeset` command?

.github/workflows/snapshot.yml
```diff
@@ -23,21 +23,22 @@ jobs:
...
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Remove snapshot branch
+        run: |
+          git checkout ${{ github.ref }}
+          git branch -D snapshot
```
aaronccasanova: I know I originally proposed this step, but am now wondering if it's needed. A fresh runner is used each time this workflow is triggered and we never push the temp `snapshot-release` branch. Should we delete?

alex-page: 🚮

alex-page (commented):
- [ ] Update release documentation
- [ ] Only allow shopifolk to run `/snapshot-release`

.github/workflows/snapshot.yml
sam-b-rose: Great point! Nice catch

.github/workflows/snapshot.yml
sam-b-rose: I did test and no extra `--` needed here 👍

PR #5924, opened 2022-05-25
feat: Format snapshot release success comment

aaronccasanova:
This PR updates the `snapshot.yml` workflow to execute the Changeset `version` and `publish` commands in an [inline GitHub-Script](https://github.com/actions/github-script#use-npm-packages) and filters the results to send a formatted PR comment e.g.
image

### DIFF
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -38,7 +38,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v2
         with:
-          node-version: 16
+          node-version: 16.x
 
       - name: Install dependencies
         run: yarn --frozen-lockfile
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -75,9 +75,12 @@ jobs:
             )
 
             if (newTags.length) {
+              const multiple = newTags.length > 1
               const body = (
-                `**Thanks @${context.actor}! Your snapshots have been shipped. 🚢**\n` +
-                `Test your changes with:\n` +
+                `**Thanks @${context.actor}! ` +
+                `Your snapshot${multiple ? 's have' : ' has'} been shipped. 🚢**\n` +
+                `Test the snapshot${multiple ? 's' : ''} by updating your \`package.json\` ` +
+                `with the newly published version${multiple ? 's' : ''}:\n` +
                 newTags.map(tag => (
                   '```sh\n' +
                   `yarn add ${tag}\n` +
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -59,11 +59,11 @@ jobs:
           script: |
             const execa = require('execa')
 
-            const versionProcess = execa.command('npm run version-packages -- --snapshot snapshot-release')
+            const versionProcess = execa.command('yarn version-packages -- --snapshot snapshot-release')
             versionProcess.stdout.pipe(process.stdout)
             await versionProcess
 
-            const releaseProcess = execa.command('npm run release -- --no-git-tags --snapshot')
+            const releaseProcess = execa.command('yarn release -- --no-git-tags --snapshot')
             releaseProcess.stdout.pipe(process.stdout)
 
             const {stdout} = await releaseProcess
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -76,7 +76,7 @@ jobs:
 
             if (newTags.length) {
               const body = (
-                `**Thanks @${context.actor} your snapshot has been shipped! 🚢**\n` +
+                `**Thanks @${context.actor}! Your snapshots have been shipped. 🚢**\n` +
                 `Test your changes with:\n` +
                 newTags.map(tag => (
                   '```sh\n' +
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -80,7 +80,7 @@ jobs:
                 `Test your changes with:\n` +
                 newTags.map(tag => (
                   '```sh\n' +
-                  `npm i ${tag}\n` +
+                  `yarn add ${tag}\n` +
                   '```'
                 )).join('\n')
               )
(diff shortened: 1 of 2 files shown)


.github/workflows/snapshot.yml
```diff
@@ -45,27 +51,46 @@ jobs:
...
+            versionProcess.stdout.pipe(process.stdout)
+            await versionProcess
+
+            const releaseProcess = execa.command('npm run release -- --no-git-tags --snapshot')
+            releaseProcess.stdout.pipe(process.stdout)
+
+            const {stdout} = await releaseProcess
+
+            const newTags = Array.from(
+              stdout.matchAll(/New tag:\s+([^\s\n]+)/g),
```
aaronccasanova, line 72: Note: This pattern of extracting release tags was derived from the official Changesets GitHub action:
https://github.com/changesets/action/blob/fa2875d61732593540054835e577e28a69e7f4cb/src/run.ts#L97

.github/workflows/snapshot.yml
```diff
@@ -26,9 +26,15 @@ jobs:
...
-      - name: Checkout Repo
+      - name: Checkout default branch
         uses: actions/checkout@v2
 
+      - name: Checkout pull request branch
+        # https://github.com/actions/checkout/issues/331#issuecomment-1120113003
+        run: hub pr checkout ${{ github.event.issue.number }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
```
aaronccasanova, line 37: Without this addition we are creating snapshot releases from the Changesets on `main`. After some light research I found that there is no way to get the PR branch solely from the `issue_comment` trigger contexts. However, `hub` is built directly into GitHub actions and includes a nifty command to checkout the branch from a pull request number 🙌

.github/workflows/snapshot.yml
```diff
@@ -45,27 +51,46 @@ jobs:
...
+            const execa = require('execa')
+
+            const versionProcess = execa.command('npm run version-packages -- --snapshot snapshot-release')
+            versionProcess.stdout.pipe(process.stdout)
+            await versionProcess
+
+            const releaseProcess = execa.command('npm run release -- --no-git-tags --snapshot')
+            releaseProcess.stdout.pipe(process.stdout)
+
+            const {stdout} = await releaseProcess
```
aaronccasanova: In case folks are wondering, this `execa.command(...)/subprocess.stdout.pipe(process.stdout)` combo allows us to follow along with the logs in the GitHub action while also storing them in a `stdout` variable to filter out the snapshot release tags.

More examples can be found in the [execa docs](https://www.npmjs.com/package/execa#:~:text=Save%20and%20pipe%20output%20from%20a%20child%20process)

.github/workflows/snapshot.yml
sam-b-rose: Just want a fact-check here. Reading over the `actions/checkout` [usage docs](https://github.com/actions/checkout#usage), the default ref that is checked out is the one that triggered the workflow:

> When checking out the repository that triggered a workflow, this defaults to the reference or SHA for that event

What info are you finding that suggests this is using the default branch (`main`)?

.github/workflows/snapshot.yml
sam-b-rose: Nice source code finding!! 💯

.github/workflows/snapshot.yml
```diff
@@ -45,27 +51,46 @@ jobs:
...
-
-      - name: Create comment
-        uses: peter-evans/create-or-update-comment@v2
         with:
-          issue-number: ${{ github.event.issue.number }}
-          body: ${{ steps.publish_snapshots.outputs.body }}
+          script: |
+            const execa = require('execa')
+
+            const versionProcess = execa.command('npm run version-packages -- --snapshot snapshot-release')
```
sam-b-rose: Should we use `yarn` just for consistency? I know they are effectively the same 🤷

.github/workflows/snapshot.yml
```diff
@@ -45,27 +51,46 @@ jobs:
...
+
+            const newTags = Array.from(
+              stdout.matchAll(/New tag:\s+([^\s\n]+)/g),
+            ).map(
+              ([_, tag]) => tag,
+            )
+
+            if (newTags.length) {
+              const body = (
+                `**Thanks @${context.actor} your snapshot has been shipped! 🚢**\n` +
```
sam-b-rose:
```suggestion
                `**Thanks @${context.actor}! Your snapshots have been shipped. 🚢**\n` +
```

.github/workflows/snapshot.yml
```diff
@@ -45,27 +51,46 @@ jobs:
...
+                `Test your changes with:\n` +
```
sam-b-rose:
```suggestion
                `Test your snapshots by updating your package.json with the newly published versions:\n` +
```

Thoughts on more detail? Just a suggestion, but either works for me!

sam-b-rose (approved):
Excellent improvement!! Only semi-blocking question is around the added step for checking out the PR branch. This was an issue when using the `repository_dispatch` trigger before, but I think it isn't a problem for a normal `issue_comment` trigger.

.github/workflows/snapshot.yml
aaronccasanova: Whoops, nice catch, definitely!

.github/workflows/snapshot.yml
aaronccasanova: > What info are you finding that suggests this is using the default branch (main)?

I discovered this behavior when running `yarn version-packages -- --snapshot snapshot-release`, in my test repo, and seeing the logs state there were no changesets found. In my case, I had one `test branch` and `main` and only the `test branch` had changesets. Thus I logged the `current branch`, after checkout, and sure enough the action checked out `main`.
image

.github/workflows/snapshot.yml
aaronccasanova: I like it! Added a slightly modified version and included support for singular/plural grammar: https://github.com/Shopify/polaris/pull/5924/commits/a6f73829d7e2dbb1294010be8064ef3b9c2cfa32

image

PR #5928, opened 2022-05-25
[Page] Fix page without header section causes scroll

FCalabria:

When a `Page` component does not have any attribute that generates a `Header`, the content adds some margin to create a proper separation to the top of the page. That separation is created with a margin (contrary to the header, which uses padding to create that space) which affects the `body` position and creates an undesired scroll.

<details>
      <summary>Bug reproduction on the playground</summary>
      canvas section off the playground, empty except for a text 'Page with no header section'. The page scrolls slightly up and down
    </details>

Changed the margin by padding so it doesn't affect the body position.

Before:
playground, the body element of the canvas is selected with the element inspector. A red arrow points to a space left between the body and the top of the page. At the bottom, the body overflows the canvas

After:
playground, the body element of the canvas is selected with the element inspector. It fits the canvas area perfectly

### 🎩 checklist

### DIFF
--- a/.changeset/shy-suns-peel.md
+++ b/.changeset/shy-suns-peel.md
@@ -2,4 +2,4 @@
 '@shopify/polaris': patch
 ---
 
-Fix scroll caused by Page without header section
+Fixed  `Page` without header section causing unnecessary scrollbar
--- a/.changeset/shy-suns-peel.md
+++ b/.changeset/shy-suns-peel.md
@@ -1,0 +1,5 @@
+---
+'@shopify/polaris': patch
+---
+
+Fix scroll caused by Page without header section
--- a/polaris-react/src/styles/shared/_page.scss
+++ b/polaris-react/src/styles/shared/_page.scss
@@ -9,10 +9,10 @@
 }
 
 @mixin page-content-layout {
-  margin: var(--p-space-2) 0;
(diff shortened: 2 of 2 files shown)


kyledurand (changes requested):
I'm not sure this is the best fix for the issue. Looks like we might have been relying on margin collapses and this change will introduce new spacing. I could be wrong but I'd love to see a spin instance with the snap before approving

FCalabria:
@kyledurand yeah sure, here you have a spin with the snapshot installed on web: https://web-66hm.paqui-calabria.eu.spin.dev/
I have checked side by side pages with title, without title and with `title=""` and all looks good to me, but I might have overlooked some corner case.

kyledurand (approved):
🎩  in spin looks good to me 👍

.changeset/shy-suns-peel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/polaris': patch
+---
+
+Fix scroll caused by Page without header section
```
chloerice:
```suggestion
Fixed  `Page` without header section causing unnecessary scrollbar
```

chloerice (approved):
Tophat looks good to me as well! 👀

Issue #5900, opened 2022-05-24
[Polaris website refresh] Improving component readme parsing


yurm04:
- It’s pretty horrendous
- Changing a readme doesn’t automatically update the website
- Some duplication between the website and polaris-react

Issue #5899, opened 2022-05-24
[Polaris website refresh] Hosting the examples


yurm04:
- Extracting example code out of readmes (WIP: https://github.com/Shopify/polaris/pull/5626) 
- Using storybook? Pros and cons?
- Johan has ideas about how to make this section better.
- [Seek playroom examples?](https://github.com/seek-oss/playroom)

PR #5934, opened 2022-05-25
Website - POC for extracting, parsing, and rendering Component examples

romellogoodman:
This PR is a proposal for parsing the component examples from the README.md file so that they can live as standalone files and be linted and type-checked. The proposal builds off the work done in #5626 and uses the `AccountConnection` component to showcase how it would work. The work can be broken down into 3 parts:

* Extraction
* Parsing
* Rendering

Links

* Account Connect - http://localhost:3000/components/actions/account-connection
* Account Connect Example - http://localhost:3000/examples/default-account-connection
* #5899
* #5900

### Extraction

Extraction is straight forward as we already have a script we can use from #5626. This can either be a one-time process that removes the content or we can keep the script around in order to maintain backwards compatibility with the existing styleguide site. The end result would be the creation of a `polaris-react/src/components/$component/example` folder for each component with example files and an `index.json` file that contains the title and description of each. You can see an approximation of this work in the files of #5626.

### Parsing

Parsing the examples so that they can be used in the next.js website would require the introduction of a new script `polaris.shopify.com/src/scripts/parseExamples.mjs` and`yarn parse-examples` command. The script itself would

* Look through polaris-react/src/components
* For every example found copy it into polaris.shopify.com/src/pages/generated-examples
* Add templated content to render the example (see below)
* Look for an polaris-react/src/components/$component/example/index.json
* Add each entry into polaris.shopify.com/src/data/components.examples.json

 **Templated Content:**
 
```
// Added the bottom of each example file
  import PolarisExample from '../../components/PolarisExample';

  export function ComponentExample() {
    return (
      <PolarisExample>
        <Example />
      </PolarisExample>
    )
  }

  export default ComponentExample
```

**Commits in this PR that demonstrate this:**
* Adding the script - https://github.com/Shopify/polaris/commit/f00c3e2caf779eda756c7b922b7981992fcc2311
* Simulating the scripts input/output - https://github.com/Shopify/polaris/commit/3b00dc77757d838796ff5dceb1304bc0366937df

### Rendering

To render the examples we look through polaris.shopify.com/src/data/components.examples.json to determine if the component has examples and iframe them onto the page. By creating the `PolarisExample` component we can separate the logic needed for rendering a component on the site from the example of it being used. This gives the added benefit of maintaining the copy-paste ability of the sourcecode that exists on the current site and means that each example doesn't need boilerplate code added to it. A thing to note about the implementation of PolarisExample is that we have to dynamically add the stylesheet onto the page here instead of using next.js' _document feature. The reason for this is that the global site styles clash with the global stylesheet needed for `polaris-react` and we can't conditionally render the sheet in _document as the document does not reload on page transitions. We are left with a situation where we can either append the stylesheet into the head using a hook or render it directly into the body alongside the component.

**Commits in this PR that demonstrate this:**
* Rendering the examples - https://github.com/Shopify/polaris/commit/e845b800baaf094c02353d7f532d0e2e12794a87

## Next Steps

Gather feedback on the approach above and then break each step into it's own PR

* Extraction - Use the script from #5626 to extract all of the examples
* Parsing - Create and run the parseExamples script
* Rendering - Determine a format/design for rendering the examples


romellogoodman:
Closing in favor of building on top of the #5626 script as codemod

romellogoodman:
Closed in favor of #5944 and subsequent PRs

PR #5659, opened 2022-04-27
Tokens, port polaris-react

romellogoodman:
### WHY are these changes introduced?

Fixes https://github.com/Shopify/polaris/issues/4088

This PR updates polaris-react to use `@shopify/polaris-tokens`. It removes the existing tokens from the polaris-react package, updates code to use the new token schema and removes tests and utilities that have already been ported over.

PR #5952, opened 2022-05-25
feat: Add @shopify/polaris-tokens to @shopify/polaris

aaronccasanova:

This PR [re-introduces the feature to add `@shopify/polaris-tokens` to `@shopify/polaris`](https://github.com/Shopify/polaris/pull/5659) and updates the `Grid` component to use `@shopify/polaris-tokens`.


polaris-react/src/components/CustomProperties/CustomProperties.scss
```diff
@@ -0,0 +1,7 @@
+/* stylelint-disable scss/at-import-partial-extension-blacklist */
+
+/**
+ * Import the scss file from polaris-tokens
+ * so that the content is appended to the main polaris-react css file
+ */
+@import '../../../../node_modules/@shopify/polaris-tokens/dist/scss/styles.scss';
```
alex-page, line 7: This should get cleaned up in the future and import from the built package. Could this break for a consumer that is using the scss and building it?

polaris-react/src/tokens/token-groups/typography.json
```diff
@@ -1,31 +0,0 @@
-{
-  "font-family-sans": "-apple-system, BlinkMacSystemFont, 'San Francisco', 'Segoe UI', Roboto, 'Helvetica Neue', sans-serif",
-  "font-family-mono": "ui-monospace, SFMono-Regular, 'SF Mono', Consolas, 'Liberation Mono', Menlo, monospace",
```
alex-page, line 3: Is this a breaking change as the built files no longer include the tokens? https://unpkg.com/browse/@shopify/polaris@9.9.0/build/esm/tokens/tokens.js

I don't think this was ever documented to consumers but it might be a concern. Is there any code that we should remove that adds them to the build?

aaronccasanova: Shouldn't be. We intentionally didn't add `tokens` to the public API: https://unpkg.com/browse/@shopify/polaris@9.9.0/build/esm/index.js

alex-page: Awesome. Thanks @aaronccasanova

polaris-react/src/components/CustomProperties/CustomProperties.scss
aaronccasanova: > Could this break for a consumer that is using the scss and building it?

Did some digging and with the recent removal of the public sass API we no longer expose any raw SCSS files for consumers to build.

> This should get cleaned up in the future and import from the built package.

We tried this originally and were having issues configuring `node-sass` to resolve `styles.scss` from `@shopify/polaris-tokens`. Agree this is something we should circle back to 👍 Would love this too look more like:
```scss
@import '@shopify/polaris-tokens/scss/styles.scss'
```

alex-page: Yes. I would just adjust it slightly... so that it's clear this is not a thing we do for other tokens.
```scss
@import '@shopify/polaris-tokens/scss/breakpoints.scss'
```

BPScott: As Aaron says, scss files in polaris-react are compiled at build time and we ship css files. There are no scss file in the published `@shopify/polaris` package anymore. Thus having this funky relative import won't break anything - it just looks ugly AF.

https://github.com/Shopify/polaris/blob/main/polaris-react/config/rollup/plugin-styles.js#L155 is where we teach sass about load paths - we should be able to teach it to look in node_modules , so you can do `@import '@shopify/polaris-tokens'`. See https://sass-lang.com/documentation/at-rules/import#load-paths


---
As a side note - you'll be able to cmd+click on this relative path in vscode to go to the tokens file. I believe you'll loose that functionality if you do a `@import '@shopify/polaris-tokens'`. Is that DX tradeoff important?

alex-page: Yes! My misunderstanding I was thinking of the public API that no longer exists 🤙

aaronccasanova: @BPScott Thanks for pointing me in the right direction. I'll give that a shot!

aaronccasanova: @alex-page 

> Yes. I would just adjust it slightly... so that it's clear this is not a thing we do for other tokens.
> ```@import '@shopify/polaris-tokens/scss/breakpoints.scss'```

Do you mean `@import '@shopify/polaris-tokens/scss/custom-properties.scss`? The `styles.scss` file contains all our Polaris custom properties (no breakpoints).

alex-page: I would rename that file from `style.scss` to `breakpoints.scss` as this is an edge case for breakpoints today. This isn't a pattern we want to scale for multiple tokens.

aaronccasanova: Sorry, I think there's some miscommunication. There are no breakpoints in `styles.scss` only Polaris custom properties.

alex-page: Oh my bad. I was thinking we would import them with css not scss. That makes sense though.

alex-page (approved):

PR #5962, opened 2022-05-27
Move rollup to root package.json

alex-page:
Co-Authored-By: Ben Scott <227292+BPScott@users.noreply.github.com>

Added a skip changelog as I don't think that consumers would be affected by this change.


alex-page:
Added a skip changelog as I don't think that consumers would be affected by this change.

polaris-react/package.json
```diff
@@ -72,10 +72,6 @@
   "devDependencies": {
     "@babel/core": "^7.15.0",
     "@babel/node": "^7.14.9",
-    "@rollup/plugin-image": "^2.0.5",
-    "@rollup/plugin-json": "^4.1.0",
-    "@rollup/plugin-replace": "^2.3.3",
-    "@rollup/pluginutils": "^3.1.0",
```
BPScott, line 78: Major version bump here, We should confirm that this doesn't have an adverse effect on build output (and it might result in a yarn.lock change)

BPScott (commented):
Looking good. Happy to see this stuff be pulled up.

I see a few other rollup build related things that could be pulled up:

We should also pull up the dependencies on `rollup` (present in polaris-for-vscode, polaris-icons, polaris-tokens), `@svgr/core` (from polaris-icons), `@shopify/browserslist-config` (from polaris-tokens). Basically anything that gets mentioned from a rollup.config.js

We should also pull up and deduplicate the `typescript` dependency

polaris-for-vscode/package.json
```diff
@@ -42,9 +42,6 @@
     "vscode-languageserver-textdocument": "^1.0.4"
   },
   "devDependencies": {
-    "@rollup/plugin-babel": "^5.3.1",
-    "@rollup/plugin-commonjs": "^21.1.0",
-    "@rollup/plugin-node-resolve": "^13.2.1",
```
aaronccasanova, line 47: It's strange that these are erroring here and not in `polaris-tokens`:
image

Perhaps, the build is just failing first in the `polaris-for-vscode` `rollup.config.js`. To quickly unblock this PR we could inline disable the eslint rule for these imports. Otherwise, we likely need to add some override rules here: https://github.com/Shopify/polaris/blob/main/.eslintrc.js

alex-page: I feels weird adding override rules for moving dependencies to the root workspace. I wonder if there is a way typescript can understand the monorepo -_-

aaronccasanova:
```js
// .eslintrc.js
{
  overrides: [
    {
      files: ['**/rollup.config.js'],
      rules: {
        'import/no-extraneous-dependencies': 'off',
      }
    }
  ]
}
```

BPScott: https://github.com/import-js/eslint-plugin-import/blob/db0970a6bbdfaa84f1e19f6cbb7d60647051546f/README.md#importexternal-module-folders looks like it has some promise

BPScott: Or actually even better reading the import/no-extraneous-dependencies doc: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md#options suggests there's a `packageDir` option where you can teach it about multiple package.json files (ie. the one at the root and the ones on a per-package basis)

alex-page: OOoo thanks Ben and Aaron. Let me try the `packageDir` approach!

.eslintrc.js
```diff
@@ -53,6 +53,7 @@ module.exports = {
     '@babel/no-unused-expressions': 'off',
     'import/named': 'off',
     'import/no-default-export': ['error'],
+    'import/no-extraneous-dependencies': ['error', {packageDir: ['../', './']}],
```
alex-page: @BPScott @aaronccasanova what do you think about this? It is working locally.

polaris-react/package.json
alex-page: Ran build with `"@rollup/pluginutils": "^3.1.0",` and `"@rollup/pluginutils": "^4.1.0",` both produced the same output running `diff -r build build-og`

.eslintrc.js
BPScott: I think this is ok but is is pretty brittle. It relies upon eslint being ran in the specific package folders.

If you don't do that, e.g. running `yarn eslint ./polaris-react` in the root of the repo, causes thousands of errors for me. It also means I see errors in vscode, as that doesn't run with the cwd you're assuming. Sample:

```
/Users/ben/src/github.com/Shopify/polaris/polaris-react/tests/utilities/react-testing.tsx
0:1  error  The package.json file could not be found                                                                                  import/no-extraneous-dependencies
  1:1  error  'react' should be listed in the project's dependencies. Run 'npm i -S react' to add it                                    import/no-extraneous-dependencies
  2:1  error  '@shopify/react-testing' should be listed in the project's dependencies. Run 'npm i -S @shopify/react-testing' to add it  import/no-extraneous-dependencies
```

I suspect:

- These paths should be absolute so they don't depend upon the cwd
- We'd need to add overrides for each package to reference the correct package.json for each package.

BPScott (approved):
🎉

PR #5963, opened 2022-05-27
@shopify/polaris Use rollup instead of loom for build process

alex-page:


polaris-react/config/rollup/plugin-styles.js
```diff
@@ -119,14 +119,15 @@ module.exports.styles = function styles({
     }
 
     // Regular css file
+
```
aaronccasanova:
```suggestion
```

polaris-react/rollup.config.js
```diff
@@ -0,0 +1,96 @@
...
+      replace({
+        '{{POLARIS_VERSION}}': packageJSON.version,
+        delimiters: ['', ''],
+        preventAssignment: true,
+      }),
+      image(),
+      json({
+        compact: true,
+      }),
+
```
aaronccasanova:
```suggestion
```

polaris-react/rollup.config.js
```diff
@@ -0,0 +1,96 @@
...
+      {
+        format: 'cjs',
+        dir: 'build/cjs',
+        preserveModules: true,
+        entryFileNames: '[name][assetExtname].js',
+        exports: 'named',
+      },
+      {
+        format: 'esm',
+        dir: 'build/esm',
```
aaronccasanova: **suggestion (non-blocking):** Should we derive output `dir`s from the `package.json` like [we did in polaris-tokens](https://github.com/Shopify/polaris/blob/66b43a6892cdac31d6884c4dfd431c4588705d75/polaris-tokens/rollup.config.js#L20-L26)?

polaris-react/rollup.config.js
```diff
@@ -0,0 +1,96 @@
...
+      mode: 'esnext',
+      modules: {
+        generateScopedName: generateScopedName({includeHash: true}),
+      },
+      plugins: postcssPlugins,
+    },
+    output: [
+      {
+        format: 'esm',
+        dir: 'build/esnext',
```
aaronccasanova: Same as above

polaris-react/rollup.config.js
```diff
@@ -0,0 +1,96 @@
+const {babel} = require('@rollup/plugin-babel');
```
aaronccasanova: Would be nice to [add `JSDoc`s](https://github.com/Shopify/polaris/blob/66b43a6892cdac31d6884c4dfd431c4588705d75/polaris-tokens/rollup.config.js#L11-L13) in this file to get some type hints throughout the config. Can be done in a follow up task.

package.json
```diff
@@ -56,6 +56,14 @@
...
+    "@rollup/plugin-commonjs": "^21.1.0",
+    "@rollup/plugin-image": "^2.0.5",
+    "@rollup/plugin-json": "^4.1.0",
+    "@rollup/plugin-node-resolve": "^13.2.1",
+    "@rollup/plugin-replace": "^2.3.3",
+    "@rollup/plugin-virtual": "^2.0.3",
+    "@rollup/pluginutils": "^4.1.0",
     "stylelint": "^14.1.0",
     "turbo": "^1.2.8",
     "typescript": ">=3.0.0"
```
aaronccasanova, line 75: Do we need to adjust this `typescript` version range? Polaris was previously on `~4.3.5`. I realize that falls within `>=3`, but I'm not sure what version will be used now that `polaris-react`'s `typescript` dependency was removed.

aaronccasanova (commented):
Nice work!! I'll pull down this branch in a bit to validate the upgrade 👍

polaris-react/rollup.config.js
BPScott: That's a nice idea! I think this being an absolute path would also mean that we can revert the `generateOptions` change in plugin-styles.

polaris-react/rollup.config.js
BPScott: 💯 Love this. In an "improvements I'd like to see in all rollup config files" vein - Currently rollup.config.js files are transpiled through babel, and this leads to weirdness when importing commonjs files because how babel handles that vs how node handles that is not consistent (tis a hecking minefield). We can avoid all that by having `rollup.config.mjs` files that are pure esm, and won't get transpiled.

See https://rollupjs.org/guide/en/#configuration-files // https://rollupjs.org/guide/en/#using-untranspiled-config-files

package.json
BPScott: Yep we do. `yarn why typescript` shows multiple versions of typescript installed. We should remove mentions of typescript on a package level and just have one definition at the root. This could be done outside of this PR, possibly in #5962, that pulls other packages up to the root

BPScott:
This is still a little bit WIP - we want to merge #5962, and pull typescript up to a workspace level dependency first. We also need to delete all the loom packages in polaris-react's package.json and ensure any rollup plugins used are explictly mentioned in the workspace level package.json

package.json
alex-page: Yes. I agree a good follow up.

polaris-react/scripts/build.js
```diff
@@ -7,8 +7,14 @@ const run = (cmd) => execSync(cmd, {stdio: 'inherit', cwd: root});
 // Generate SCSS media queries in `src/styles` for internal use
 run(`yarn gen-media-queries`);
 
+run('yarn run tsc -b');
+
 // Run Sk build to generate everything
-run(`yarn run loom build`);
+run(`yarn run rollup -c`);
+
+run('yarn run downlevel-dts "./build/ts/latest" "./build/ts/3.4"');
```
BPScott: Some merge goof,  downlevel-dts / copyfiles commands are already present

polaris-icons/rollup.config.js
```diff
@@ -203,6 +203,9 @@ function replaceFillAttributeSvgoPlugin() {
   };
 }
 
+/**
+ * @type {import('rollup').RollupOptions}
+ */
```
BPScott: Nitpick: you can specify these docblocks on one line

```suggestion
/** @type {import('rollup').RollupOptions} */
```

polaris-react/rollup.config.js
```diff
@@ -0,0 +1,100 @@
...
+      }),
+      styles(stylesConfig),
+    ],
+    output,
+  };
+}
+
+/**
+ * @type {import('rollup').RollupOptions}
+ */
```
BPScott: Nitpick: you can specify these docblocks on one line

```suggestion
/** @type {import('rollup').RollupOptions} */
```

package.json
```diff
@@ -57,14 +65,7 @@
...
-    "@rollup/plugin-image": "^2.0.5",
-    "@rollup/plugin-json": "^4.1.0",
-    "@rollup/plugin-node-resolve": "^13.2.1",
-    "@rollup/plugin-replace": "^2.3.3",
-    "@rollup/plugin-virtual": "^2.0.3",
-    "@rollup/pluginutils": "^4.1.0",
+    "rollup-plugin-node-externals": "^4.0.0",
     "stylelint": "^14.1.0",
     "turbo": "^1.2.8",
     "typescript": ">=3.0.0"
```
BPScott, line 75: I reckon this should be a bit nearer to reality rather than "any TS version released in the last 4 years".

```suggestion
    "typescript": "^4.3.5"
```

Run a `yarn` after this

BPScott (commented):
Looking great. A few small bits inline.

We should also delete the `polaris-react/loom.config.ts` file.

Looks like the build is failing due to a messed up import in the rollup config file.

BPScott:
I think that's back to working again.

Next things:

- Test build output again to confirm its identical
- Update the typescript version in the root package.json - it can still resolve to the same version but `>=3.0.0` is a silly dependency range
- Follow up PR to make all the rollup.config files .mjs - so that the config files don't need to be transpiled as part of running rollup, they can be interpreted by node immediately.

BPScott (approved):
🎉

PR #5966, opened 2022-05-27
[Grid] Refactor grid-areas

kyledurand:
This really cleans up the Grid component and improves performance by removing resize event listener


polaris-react/scripts/gen-media-queries.js
```diff
@@ -59,5 +61,5 @@ function getDownMediaCondition(breakpoint) {
 }
 
 function toEm(value) {
-  return `${parseInt(value, 10) / 16}em`;
+  return `${parseInt(value, 10)}em`;
```
kyledurand: I need to revert this change once https://github.com/Shopify/polaris/pull/6012 ships

polaris-react/src/components/Grid/Grid.scss
```diff
@@ -25,6 +35,16 @@
...
+    grid-template-areas: var(
+      --pc-grid-areas-xl,
+      var(
+        --pc-grid-areas-lg,
+        var(
+          --pc-grid-areas-md,
+          var(--pc-grid-areas-sm, var(--pc-grid-areas-xs))
+        )
+      )
+    );
```
kyledurand: cc @aaronccasanova we have all the fallbacks here now (which we couldn't have had with the js solution without a closest match check) by chaining vars with fallbacks. Code looks grosssss but super nice to use

aaronccasanova: I love the idea!! Had a quick play with a similar/alternative approach: https://codepen.io/aaronccasanova/pen/QWQxpmE?editors=1100

What do you think? In the context of the `Grid`, I'm imagining we would set the base case to:
`--pc-grid-areas-xs: initial`

alex-page (approved):

aaronccasanova (approved):

polaris-react/src/components/Grid/Grid.scss
kyledurand: I missed this til just now but this is a fantastic idea. Will implement

PR #5984, opened 2022-05-31
[Page] Fix divider so that it has a padding of 32px

zakwarsame:

- Changes the page padding underneath the divider from 16px to 32px 

Fixes https://github.com/Shopify/shopify/issues/353025

Please see here for fix results: https://github.com/Shopify/web/pull/65757

Storybook example:
Screenshot of page with divider

### 🎩 checklist

### DIFF
--- a/.changeset/brave-kids-heal.md
+++ b/.changeset/brave-kids-heal.md
@@ -1,0 +1,5 @@
+---
+'@shopify/polaris': patch
+---
+
+changes divider padding on page component
--- a/polaris-react/src/components/Page/Page.scss
+++ b/polaris-react/src/components/Page/Page.scss
@@ -37,5 +37,5 @@ body {
 
 .divider {
   border-top: var(--p-border-divider);
-  padding-top: var(--p-space-4);
+  padding-top: var(--p-space-8);
 }


matallo (approved):
Code and 🎩 LGTM (in web PR), make sure to check all the cases when the divider is used.

sarahill:
I'm wondering if this change will cause unintended spacing due to varying ways a page is implemented. I quickly adjusted the divider spacing from `16px` to `32px` in the browser on some settings pages and it looks like there may be additional layout elements that are causing additional space. I suspect to achieve what you're doing here.

Screen Shot 2022-06-01 at 10 43 15 AM

---

Screen Shot 2022-06-01 at 10 48 11 AM

image

image

zakwarsame:
Thank you for the feedback!
I did consider that, and then made the necessary changes on the web repo because for each additional page being created, those UI components also had to add the margin on top as you've mentioned. This could be easily achieved by polaris hence this change.
Here are the spin URLs showing how it would look like after adding this change with a snapshot and removing the additional UI components spaces: 
- store details: https://shop7.shopify.page-divider.zakaria-warsame.us.spin.dev/admin/settings/general
- users and permissions: https://shop7.shopify.page-divider.zakaria-warsame.us.spin.dev/admin/settings/account

See this PR for overall changes: https://github.com/Shopify/web/pull/65478
I haven't yet shipped that one because they both have to be done at the same time.

sarahill (approved):
Nice update and thanks for looking at this across the admin 👏

Issue #5903, opened 2022-05-24
[Mobile-friendly] Implement navigation of the website in small screens


yurm04:
## Top navigation

- [x] Show hamburger button and side menu in small screens
- [x] Create side menu with links to different sections of the website
- [ ] Move left navigation from each section to the side menu (In Review: [#6141](https://github.com/Shopify/polaris/pull/6141); [#6226](https://github.com/Shopify/polaris/pull/6226))

### Nice to have

- [ ] Extend keyboard experience by implementing UP, DOWN, and ESC keys
- [ ] Animate closing transition of the side menu

PR #5998, opened 2022-06-01
[polaris.shopify.com] Add side menu to the website in small screens

gui-santos:

Implements part of [#5903](https://github.com/Shopify/polaris/issues/5903)

This PR creates a SideMenu component that can only be open in small screens
It is opened by clicking in a Hamburger Button that shows up in the top bar of the websites in screens smaller than 768px

<details>
      <summary>SideMenu in action</summary>
      Opening and closing side menu
</details>

This is what I came up so far, we can always improve this 🙌 
tell me what you think

### Mobile-first or Desktop-first?

Our css is currently all set with the big screens in mind, so I implemented the media queries as if we are "desktop-first", I was not sure what the team prefers so I kept it like that
BUT if the team wants I can just refactor those CSS to be mobile-first 👌 

### Help

~~There is one weird thing that is happening and I can't find out why: the website is not scaling down in Chrome when you use the device simulator in dev tools, the one that you get when clicking on this button:~~

Screenshot 2022-06-01 at 16 03 08

~~And I added the viewport meta tag to `_app.tsx` and it is being loaded in all the pages of the website~~

~~If you know what is happening, please tell me~~

ERRATA: I think it was a bug in chrome because this is working now 🤷 

### Nice TODO's for another PR (this one is already too big)

- Extend keyboard experience by implementing UP, DOWN, and ESC keys
- Animate closing transition of the side menu

### How to 🎩

### 🎩 checklist


ghost:

gui-santos:
I pushed some code to address the keyboard navigation and the experience for screen reader users

<details>
      <summary>Keyboard navigation ignores the hidden menu</summary>
      Navigating through the page with keyboard
</details>

<details>
      <summary>Keyboard navigation stays inside the menu while it's opened</summary>
      Navigating inside the menu with keyboard
</details>

polaris.shopify.com/src/components/Header/Header.tsx
```diff
@@ -0,0 +1,100 @@
...
+  currentSection?: string;
+}
+
+function Header({ currentSection }: Props) {
+  const [showMenu, setShowMenu] = useState(false);
+
+  useEffect(() => {
+    // it ensures that the menu is closed after the screen is resized
+    function handleResize() {
+      if (window.innerWidth > 768 && showMenu) {
```
martenbjork: Can we extract `768` into a variable in a config file?

polaris.shopify.com/src/components/Header/Header.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+interface Props {
+  currentSection?: string;
+}
+
+function Header({ currentSection }: Props) {
+  const [showMenu, setShowMenu] = useState(false);
+
+  useEffect(() => {
+    // it ensures that the menu is closed after the screen is resized
```
martenbjork: Shopify has taught me this pattern which I like: Always try to avoid comments and let the code do the talking instead. Can you remove this comment and instead clarify the code? Perhaps renaming `handleResize` to `setNavVisiblityOnResize` would be just as clear?

polaris.shopify.com/src/components/Header/Header.tsx
```diff
@@ -0,0 +1,100 @@
...
+    window.addEventListener("resize", handleResize);
+
+    return () => window.removeEventListener("resize", handleResize);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  const handleCloseMenu = () => {
+    setShowMenu(false);
+    // it focus back to the menu button after the user closes the menu
+    document.getElementById("menu-button")?.focus();
```
martenbjork: Could we use refs here instead? Seems like a more Reacty way of doing things.

If we can't, I would extract the `id` into a const at the top of the file so that you don't have it hard coded in 2 different places.

polaris.shopify.com/src/components/Header/Header.tsx
```diff
@@ -0,0 +1,100 @@
...
+            aria-controls="side-menu"
+            aria-expanded={showMenu}
+            onClick={() => setShowMenu(true)}
+          >
+            <Image
+              src={hamburguerIcon}
+              layout="fixed"
+              width={24}
+              height={24}
+              alt="Hamburger icon"
```
martenbjork: Nit: "Hamburger icon" is a phrase that's only used in the industry. It might more accessible to call it "navigation icon" or set the alt attribute to empty (the aria-label on the parent gives enough context).

polaris.shopify.com/src/components/Header/Header.tsx
```diff
@@ -0,0 +1,100 @@
...
+            />
+          </Button>
+
+          <SideMenu
```
martenbjork: Nit: Maybe we should call this SideNav? We're using "nav" in the rest of the project, rather than "menu".

polaris.shopify.com/src/components/SideMenu/SideMenu.module.scss
```diff
@@ -0,0 +1,99 @@
...
+    transform: translateX(0);
+    transition: transform var(--duration-200) var(--ease-in-out);
+
+    .CloseButton {
+      position: absolute;
+      display: flex;
+      top: calc(
+        var(--header-height) / 2 - 12px
+      ); // 12px is half the height of the Close Icon
+      right: calc(-24px - 1rem); // 24px is the width of the Close Icon
```
martenbjork: Rather than comments, could this be communicated through code? By extracting the icon size into a CSS variable at the top? `--mobile-nav-icon-size` and then using `calc` to factor it in.

polaris.shopify.com/src/components/SideMenu/SideMenu.tsx
```diff
@@ -0,0 +1,98 @@
...
+      className={className(styles.SideMenu, showMenu && styles.show)}
+      ref={menuRef}
+    >
+      <li>
+        <Link href="/" passHref>
+          <a
+            className={styles.Logo}
+            onClick={handleCloseMenu}
+            onKeyDown={(e) => {
+              // this is the last child of the menu, if users press SHIFT + TAB it will return focus to the last element of the menu
```
martenbjork: I don't think the comment is needed, the code below is clear enough!

polaris.shopify.com/src/components/SideMenu/SideMenu.tsx
```diff
@@ -0,0 +1,98 @@
...
+        handleCloseMenu={handleCloseMenu}
+      />
+
+      <li>
+        <button
+          aria-label="Close menu"
+          className={styles.CloseButton}
+          onClick={handleCloseMenu}
+          onKeyDown={(e) => {
+            // this is the last child of the menu, if users press TAB it will return focus to the first element of the menu
```
martenbjork: Comment not needed, code gives enough context!

martenbjork (approved):

martenbjork:
Getting good! 🚀 Left some comments above.

One layout issue: On desktop, the search should be pushed to the right a bit, to the margin of the page.

Screen Shot 2022-06-03 at 12 35 24 PM

martenbjork:
I'm not 100% sure, but I think it's recommended that the `esc` key closes modal experiences like this one. I know it will be rare that someone uses a small screen + keyboard, but for the sake of adhering to a11y guidelines...

polaris.shopify.com/src/components/Header/Header.tsx
gui-santos: Uuhmmm I thought about creating an enum and keep it at `types.ts`
At least until we have come utility functions that deal with resolution and breakpoints

what do you think?

polaris.shopify.com/src/components/Header/Header.tsx
gui-santos: Oh I think it's a good idea!

It's not nit at all, naming things is the hardest and most important thing to be fair
I will rename it to SideNav 👌

gui-santos:
@martenbjork 

> One layout issue: On desktop, the search should be pushed to the right a bit, to the margin of the page.

What do you mean? Maybe I lost some style in one of my rebases
I only changed the placement of the search in screens smaller than 768px where I put it aligned to the right of the screen

> I'm not 100% sure, but I think it's recommended that the `esc` key closes modal experiences like this one. I know it will be rare that someone uses a small screen + keyboard, but for the sake of adhering to a11y guidelines...

Yep! Also UP and DOWN arrows should work, but I thought this PR was too big and I was going to add it in a next PR

I guess since you mentioned it in this review, it means I'm being over cautious with the size of the PR hahahaha
I will add those keys now 👌

gui-santos:
I changed my mind, I will add the functionality for the arrow keys in another PR because it will a bit more complex than the Esc key code

PR #6005, opened 2022-06-01
[ActionList] Decouples ActionButton from ActionList

felipeleusin:

In Shopify Email we leverage the ActionList with a single item in a slim pattern for our sidebar. We used to resort to a hack to override polaris styling to achieve our design without having to re-implement a custom component that would be identical to ActionListItem.

This PR exposes a new ActionButton component was initial used internally by ActionList as a standalone component.

If this PR is deemed good to merge I can write the documentation for this component.

### 🎩 checklist

### DIFF
--- a/.changeset/wet-mice-share.md
+++ b/.changeset/wet-mice-share.md
@@ -1,0 +1,5 @@
+---
+'@shopify/polaris': patch
+---
+
+[ActionList] Includes variable for ActionList padding
--- a/.changeset/wet-mice-share.md
+++ b/.changeset/wet-mice-share.md
@@ -1,5 +1,5 @@
 ---
-'@shopify/polaris': patch
+'@shopify/polaris': minor
 ---
 
-[ActionList] Includes variable for ActionList padding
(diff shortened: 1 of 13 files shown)


alex-page (changes requested):
@felipeleusin we strongly recommend not overriding these custom properties. We are tracking folks using them in shopify/web. Do you have to use an ActionList component, could you use a stack instead?

felipeleusin:
> @felipeleusin we strongly recommend not overriding these custom properties. We are tracking folks using them in shopify/web. Do you have to use an ActionList component, could you use a stack instead?

Stack doesn't work, we actually care less about ActionList and more about the ActionListItem, if we could use just the ActionListItem but they are wrapped in an `li`. Sadly it seems like the solution would be to create a custom component for this, which I was trying to avoid

alex-page:
@felipeleusin can you provide some screenshots and context of what you are trying to create? Or a codesandbox?

felipeleusin:
Replaced by https://github.com/Shopify/polaris/pull/6229

PR #6009, opened 2022-06-02
[Polaris website refresh] Refactor Left Nav and TOC

yurm04:
Using the Next.js [per page layouts](https://nextjs.org/docs/basic-features/layouts#per-page-layouts) approach. This refactor:
* breaks up the left nav into it's own layout
* breaks up the left nav with TOC into a separate layout
* Refactors the `Components` index page to use the left nav layout
* Refactors the individual components page `[component].tsx` to use the left nav with TOC layout
* makes everything more modular to easily assemble pages with layout of choice

**What still ain't right?**
* pretty much all of the styling is duplicated and I made no effort to clean it up. This is just implementing the new layout approach and bringing the consuming pages to parity with the visual styles today. But boy is the code ugly still 😬 
* This creates two separate layouts `LeftNavLayout` and `LeftNavWithTOCLayout`. These two layouts are probably at least 50% the same code. Ideally we could get to a place where we can split out the TOC into a separate component and compose pages using nested layouts. I couldn't figure out a quick way to do with in CSS modules world because modules pretty much eliminates the C in CSS so I decided to put that down for now in order to unblock enablement work. Ideally a proper solution will end up like the following example. I'll make a ticket to capture this issue.
```
<LeftNavLayout>
  {children} // this is the page contents
  <TableOfContents />
</LeftNavLayout>
```

### DIFF
--- a/polaris.shopify.com/src/pages/_app.tsx
+++ b/polaris.shopify.com/src/pages/_app.tsx
@@ -1,18 +1,30 @@
+import type { ReactElement, ReactNode } from 'react'
+import type { NextPage } from 'next'
 import type { AppProps } from "next/app";
 // import "@shopify/polaris/build/esm/styles.css";
 import "../styles/globals.scss";
 import Page from "../components/Page";
 import { useRouter } from "next/router";
 
-function MyApp({ Component, pageProps }: AppProps) {
+type NextPageWithLayout = NextPage & {
+  getLayout?: (page: ReactElement) => ReactNode
+}
+
+type AppPropsWithLayout = AppProps & {
+  Component: NextPageWithLayout
+}
+
+function MyApp({ Component, pageProps }: AppPropsWithLayout) {
   const router = useRouter();
   const isPolaris = router.asPath.startsWith("/generated-examples");
+  const getLayout = Component.getLayout ?? ((page) => page)
+
   return (
-    <Page skipHeaderAndFooter={isPolaris}>
-      <div style={{ background: isPolaris ? "#fafafa" : "unset" }}>
-        <Component {...pageProps} />
-      </div>
-    </Page>
+      <Page skipHeaderAndFooter={isPolaris}>
+        <div style={{ background: isPolaris ? "#fafafa" : "unset" }}>
+          {getLayout(<Component {...pageProps} />)}
+        </div>
+      </Page>
   );
 }
 
--- a/polaris.shopify.com/src/pages/components/[category]/[component].tsx
+++ b/polaris.shopify.com/src/pages/components/[category]/[component].tsx
@@ -12,13 +12,8 @@ import {
 } from "../../../utils/various";
 import fs from "fs";
 import path from "path";
-import MaxPageWidthDiv from "../../../components/MaxPageWidthDiv";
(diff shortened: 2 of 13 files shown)


polaris.shopify.com/src/pages/_app.tsx
```diff
@@ -1,18 +1,30 @@
...
   const isPolaris = router.asPath.startsWith("/generated-examples");
+  const getLayout = Component.getLayout ?? ((page) => page)
+
   return (
-    <Page skipHeaderAndFooter={isPolaris}>
-      <div style={{ background: isPolaris ? "#fafafa" : "unset" }}>
-        <Component {...pageProps} />
-      </div>
-    </Page>
+      <Page skipHeaderAndFooter={isPolaris}>
```
yurm04, line 23: assuming we want the top bar nav on every page, if not we can break this out into a separate layout as well and compose different layouts together

polaris.shopify.com/src/pages/components/index.tsx
```diff
@@ -1,8 +1,30 @@
...
 interface Props {}
 
-const Components: NextPage<Props> = () => <ComponentsPage />;
+const navItems: NavItem[] = getComponentNav();
+
+type NextPageWithLayout = NextPage & {
+  getLayout?: (page: ReactElement) => ReactNode
+}
+
+const Components: NextPageWithLayout = () => <ComponentsPage />;
```
yurm04, line 19: Someone that know typescript better than me pls double check that this makes any sense at all 🙏

polaris.shopify.com/src/pages/components/[category]/[component].tsx
```diff
@@ -11,18 +12,55 @@ import {
...
 }
 
+const navItems: NavItem[] = getComponentNav();
+
+
+type NextPageWithLayout = NextPage & Props & {
+  getLayout?: (page: ReactElement) => ReactNode
+}
+
+const ComponentsTwo: NextPageWithLayout = ({ name, readme }) => {
```
Comment (yurm04) [inline polaris.shopify.com/src/pages/components/[category]/[component].tsx:31]: this is still giving me a typescript error

Comment (romellogoodman) [reply polaris.shopify.com/src/pages/components/[category]/[component].tsx:31]: Doing `{ name, readme }: Props` fixed one of the errors but unsure how to fix the other one with NextPageWithLayout and getLayout

**Update:**

The following seems to fix the errors

```
type NextPageWithLayout = NextPage &
  any & {
    getLayout?: (page: ReactElement) => ReactNode;
  };

const ComponentsTwo: NextPageWithLayout = ({ name, readme }: Props) => {
```

romellogoodman (approved):
looks good to me! love the refactor of the components as the functionality aligns

martenbjork:
This is two steps forward, two steps back to me.  (But I love that you shipped it so quickly! Bias towards merging FTW.)

- I don't get the `getLayout` pattern. I don't see what it does that regular composability couldn't do. It easily become "a thing" you have to understand to work on the website.
- Creating a `<TableOfContents />` component that can be inserted anywhere creates too much freedom for consumers IMO. Since the TOC needs to be injected in exactly the right place in the DOM, I'd rather have the parent layout component insert this fella so that it ends up in the right place.

Let's keep on jamming on this!

PR #6012, opened 2022-06-02
Fix breakpoint tokens

kyledurand:
We were converting breakpoint px values to `rem`, then to `em` in `gen-media-queries`. So the px value would be dived by 16, then the rem value would be divided by 16 so we had super small em values as breakpoints.

This change removes pxToRem at the token level so we go straight from pxToEm in `gen-media-queries`

### DIFF
--- a/.changeset/nasty-carpets-design.md
+++ b/.changeset/nasty-carpets-design.md
@@ -1,0 +1,6 @@
+---
+'@shopify/polaris': patch
+'@shopify/polaris-tokens': patch
+---
+
+Fixed breakpoint tokens
--- a/polaris-tokens/src/tokens.ts
+++ b/polaris-tokens/src/tokens.ts
@@ -65,10 +65,10 @@ export interface Tokens {
 }
 
 export const tokens: Tokens = {
-  legacyTokens: tokensToRems(legacyTokens),
-  breakpoints,
+  breakpoints: tokensToRems(breakpoints),
   colorSchemes,
   depth,
+  legacyTokens: tokensToRems(legacyTokens),
   motion,
   shape: tokensToRems(shape),
   spacing: tokensToRems(spacing),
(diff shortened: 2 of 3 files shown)


.changeset/nasty-carpets-design.md
```diff
@@ -0,0 +1,6 @@
+---
+'@shopify/polaris': patch
```
kyledurand, line 2: I'm not sure but changeset considered this a change in /polaris as well as tokens? Is this necessary or should I remove?

aaronccasanova: That is odd. I would think Polaris would use the updated `@shopify/polaris-tokens` bump since we use the `*` version range.

polaris-tokens/src/tokens.ts
```diff
@@ -65,10 +65,10 @@ export interface Tokens {
 }
 
 export const tokens: Tokens = {
-  breakpoints: tokensToRems(breakpoints),
+  legacyTokens: tokensToRems(legacyTokens),
```
aaronccasanova: Why re-alpha?

polaris-tokens/src/tokens.ts
```diff
@@ -65,10 +65,10 @@ export interface Tokens {
...
+  breakpoints,
```
aaronccasanova: We converted these tokens to `rem` for use in components. e.g. Layout components that want to set `max-width`s at a given breakpoint.

Would it be better to update the logic in `gen-media-queries`?

kyledurand: Ah ok, yep I can make the change there for sure. Just wanted to avoid converting them twice if possible

polaris-tokens/src/tokens.ts
kyledurand: Oh we they sorted by alphabet? Didn't notice. I wanted these up top so new ones below might override if that would ever happen, which I doubt it would. I can change back though

aaronccasanova (approved):

PR #6018, opened 2022-06-03
Simplify /snapit workflow

aaronccasanova:
Simplify the `/snapit` workflow by updating the `yarn version-packages` command to inherit the parent's `stdio` instead of manually piping `versionProcess.stdout`


sam-b-rose (approved):
TDIL!! Nice clean up!

alex-page (approved):

PR #3420, opened 2020-10-08
Actions rollup menu groups

kyledurand:
### WHY are these changes introduced?

Fixes a bunch of header issues but I'll leave the issues open until we test in web

## 

### How to 🎩

🖥 [Local development instructions](https://github.com/Shopify/polaris-react/blob/master/README.md#development)
🗒 [General tophatting guidelines](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting.md)
📄 [Changelog guidelines](https://github.com/Shopify/polaris-react/blob/master/documentation/Versioning%20and%20changelog.md)

<details>
<summary>Copy-paste this code in <code>playground/Playground.tsx</code>:</summary>

```jsx
import React from 'react';
import {Page} from '../src';

export function Playground() {
  return (
    <Page title="Playground">
      {/* Add the code you want to test in here */}
    </Page>
  );
}

```

</details>

### 🎩 checklist

* [x] Tested on [mobile](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting.md#cross-browser-testing)
* [x] Tested on [multiple browsers](https://help.shopify.com/en/manual/intro-to-shopify/shopify-admin/supported-browsers)
* [ ] Tested for [accessibility](https://github.com/Shopify/polaris-react/blob/master/documentation/Accessibility%20testing.md)
* [ ] Updated the component's `README.md` with documentation changes
* [ ] [Tophatted documentation](https://github.com/Shopify/polaris-react/blob/master/documentation/Tophatting%20documentation.md) changes in the style guide
* [ ] For visual design changes, pinged one of @ HYPD, @ mirualves, @ sarahill, or @ ry5n to update the Polaris UI kit

PR #6022, opened 2022-06-03
[Page] Add 'onActionRollup' prop to Page component.

jorgenunezsiri:

Right now, the Page component by default introduces the `More Actions` action group to perform roll ups. However, when there is an action group already present, then all secondary actions are going to roll up to this one. This might not be desirable, given that the existing action group could have nothing to do with the secondary actions, or the existing action group could be disabled as well and in this case we would roll up items into a disabled action group.

To solve the above problem, this PR adds `isRollingUp` prop to the Page component. The idea is that we are able to pass a callback `isRollingUp`, and listen on the consumer side whether a roll up is happening or not. For the above issue, one fix here would be to manually introduce `More Actions` action group when a roll up happens.

### How to 🎩

You can test by passing the prop to any `Page` component in the playground:
```JS
onActionRollup={(hasRolledUp) => { console.log('HAS ROLLED UP: ', hasRolledUp); }}
```

The results should be similar to the following:

https://user-images.githubusercontent.com/16692690/171854652-8f80ae5d-978e-4b84-b462-c7d6687f44e8.mov

### 🎩 checklist

### DIFF
--- a/polaris-react/src/components/ActionMenu/ActionMenu.tsx
+++ b/polaris-react/src/components/ActionMenu/ActionMenu.tsx
@@ -19,6 +19,8 @@ export interface ActionMenuProps {
   rollup?: boolean;
   /** Label for rolled up actions activator */
   rollupActionsLabel?: string;
+  /** Callback that returns true when secondary actions are rolled up into action groups, and false when not */
+  onActionRollup?(hasRolledUp: boolean): void;
 }
 
 export function ActionMenu({
@@ -26,6 +28,7 @@ export function ActionMenu({
   groups = [],
   rollup,
   rollupActionsLabel,
+  onActionRollup,
 }: ActionMenuProps) {
   if (actions.length === 0 && groups.length === 0) {
     return null;
@@ -47,7 +50,11 @@ export function ActionMenu({
           sections={rollupSections}
         />
       ) : (
-        <Actions actions={actions} groups={groups} />
+        <Actions
+          actions={actions}
+          groups={groups}
+          onActionRollup={onActionRollup}
+        />
       )}
     </div>
   );
--- a/polaris-react/src/components/ActionMenu/components/Actions/Actions.tsx
+++ b/polaris-react/src/components/ActionMenu/components/Actions/Actions.tsx
@@ -20,6 +20,8 @@ interface Props {
   actions?: MenuActionDescriptor[];
   /** Collection of page-level action groups */
   groups?: MenuGroupDescriptor[];
+  /** Callback that returns true when secondary actions are rolled up into action groups, and false when not */
+  onActionRollup?(hasRolledUp: boolean): void;
 }
 
 interface MeasuredActions {
@@ -29,7 +31,7 @@ interface MeasuredActions {
 
 const ACTION_SPACING = 8;
 
-export function Actions({actions = [], groups = []}: Props) {
+export function Actions({actions = [], groups = [], onActionRollup}: Props) {
   const i18n = useI18n();
   const actionsLayoutRef = useRef<HTMLDivElement>(null);
   const menuGroupWidthRef = useRef<number>(0);
@@ -37,6 +39,7 @@ export function Actions({actions = [], groups = []}: Props) {
   const actionsAndGroupsLengthRef = useRef<number>(0);
   const timesMeasured = useRef(0);
   const actionWidthsRef = useRef<number[]>([]);
+  const rollupActiveRef = useRef<boolean | null>(null);
   const [activeMenuGroup, setActiveMenuGroup] = useState<string | undefined>(
     undefined,
   );
@@ -124,6 +127,17 @@ export function Actions({actions = [], groups = []}: Props) {
       }
     });
 
+    if (onActionRollup) {
+      // Note: Do not include last group actions since we are skipping `lastMenuGroup` above
+      // as it is always rendered with its own actions
+      const isRollupActive =
+        newShowableActions.length < actionsAndGroups.length - 1;
+      if (rollupActiveRef.current !== isRollupActive) {
+        onActionRollup(isRollupActive);
+        rollupActiveRef.current = isRollupActive;
+      }
+    }
+
     setMeasuredActions({
       showable: newShowableActions,
       rolledUp: newRolledUpActions,
@@ -131,7 +145,7 @@ export function Actions({actions = [], groups = []}: Props) {
 
     timesMeasured.current += 1;
     actionsAndGroupsLengthRef.current = actionsAndGroups.length;
-  }, [actions, groups, lastMenuGroup, lastMenuGroupWidth]);
+  }, [actions, groups, lastMenuGroup, lastMenuGroupWidth, onActionRollup]);
 
   const handleResize = useMemo(
     () =>
--- a/polaris-react/src/components/Page/components/Header/Header.tsx
+++ b/polaris-react/src/components/Page/components/Header/Header.tsx
@@ -54,6 +54,8 @@ export interface HeaderProps extends TitleProps {
   additionalNavigation?: React.ReactNode;
   // Additional meta data
   additionalMetadata?: React.ReactNode | string;
+  /** Callback that returns true when secondary actions are rolled up into action groups, and false when not */
+  onActionRollup?(hasRolledUp: boolean): void;
 }
 
 const SHORT_TITLE = 20;
@@ -73,6 +75,7 @@ export function Header({
   secondaryActions = [],
   actionGroups = [],
   compactTitle = false,
+  onActionRollup,
 }: HeaderProps) {
   const i18n = useI18n();
   const {isNavigationCollapsed} = useMediaQuery();
@@ -150,6 +153,7 @@ export function Header({
             ? i18n.translate('Polaris.Page.Header.rollupActionsLabel', {title})
             : undefined
         }
+        onActionRollup={onActionRollup}
       />
     );
   } else if (isReactElement(secondaryActions)) {
--- a/.changeset/clean-rockets-beam.md
+++ b/.changeset/clean-rockets-beam.md
@@ -1,0 +1,5 @@
+---
+'@shopify/polaris': minor
+---
+
+Added 'onActionRollup' prop to Page component.
--- a/polaris-react/src/components/ActionMenu/tests/ActionMenu.test.tsx
+++ b/polaris-react/src/components/ActionMenu/tests/ActionMenu.test.tsx
@@ -5,7 +5,7 @@ import type {
   MenuGroupDescriptor,
   ActionListItemDescriptor,
 } from '../../../types';
-import {MenuGroup, RollupActions} from '../components';
+import {MenuGroup, RollupActions, Actions} from '../components';
 import {ActionMenu, ActionMenuProps} from '../ActionMenu';
 import {Button} from '../../Button';
 import {ButtonGroup} from '../../ButtonGroup';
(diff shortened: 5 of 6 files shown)


.changeset/mean-experts-scream.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/polaris': patch
```
kyledurand: Adding a prop requires a minor version bump

.changeset/mean-experts-scream.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+[Page] Add 'isRollingUp' prop to Page component.
```
kyledurand:
```suggestion
Added 'isRollingUp' prop to Page component.
```

polaris-react/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -19,13 +19,16 @@ export interface ActionMenuProps {
   rollup?: boolean;
   /** Label for rolled up actions activator */
   rollupActionsLabel?: string;
+  /** Callback that fires when secondary actions are rolled up into action groups */
```
kyledurand:
```suggestion
  /** Callback that returns true when secondary actions are rolled up into action groups, and false when not */
```

polaris-react/src/components/ActionMenu/components/Actions/Actions.tsx
```diff
@@ -20,6 +20,8 @@ interface Props {
   actions?: MenuActionDescriptor[];
   /** Collection of page-level action groups */
   groups?: MenuGroupDescriptor[];
+  /** Callback that fires when secondary actions are rolled up into action groups */
```
kyledurand:
```suggestion
  /** Callback that returns true when secondary actions are rolled up into action groups, and false when not */
```

polaris-react/src/components/Page/components/Header/Header.tsx
```diff
@@ -54,6 +54,8 @@ export interface HeaderProps extends TitleProps {
   additionalNavigation?: React.ReactNode;
   // Additional meta data
   additionalMetadata?: React.ReactNode | string;
+  /** Callback that fires when secondary actions are rolled up into action groups */
```
kyledurand:
```suggestion
  /** Callback that returns true when secondary actions are rolled up into action groups, and false when not */
```

kyledurand (commented):
This is looking good. I've 🎩 'd but not super thoroughly. I'll dig in a little more after my upcoming meetings.

I'd really like us to test the return value at the `Page.test.tsx` level but haven't found a good way to mock screen width. Not a blocker but would definitely be a required follow up IMO

chloerice:
Hey @jorgenunezsiri 👋🏽 I'm confused by what I'm seeing in the screen capture with the actions sometimes rolling up into a named group. They should all roll up into an ellipsis on small screens. Can you elaborate or show us an example of the issue you're facing with the rollup actions? It sounds like this is a complicated fix to a problem that shouldn't be happening (a regression that was missed after a past change to the Page Header, there have been several this year) 🧐

jorgenunezsiri:
> Hey @jorgenunezsiri 👋🏽 I'm confused by what I'm seeing in the screen capture with the actions sometimes rolling up into a named group. They should all roll up into an ellipsis on small screens. Can you elaborate or show us an example of the issue you're facing with the rollup actions? It sounds like this is a complicated fix to a problem that shouldn't be happening (a regression that was missed after a past change to the Page Header, there have been several this year) 🧐

@chloerice Thank you for your comment!

About action items rolling up into an ellipsis button, that is what is happening nowadays, but this only happens on small screens (as you said). So, we currently have the [RollupActions](https://github.com/Shopify/polaris/blob/main/polaris-react/src/components/ActionMenu/ActionMenu.tsx#L44) component that it is being rendered when the [navigation is collapsed](https://github.com/Shopify/polaris/blob/main/polaris-react/src/components/Page/components/Header/Header.tsx#L147) (on smaller screens). For any other case, we are using the [Actions](https://github.com/Shopify/polaris/blob/main/polaris-react/src/components/ActionMenu/ActionMenu.tsx#L50) component. This `Actions` component has a [More Actions](https://github.com/Shopify/polaris/blob/main/polaris-react/src/components/ActionMenu/components/Actions/Actions.tsx#L48) action group, where action items roll up by default. However, the issue is if there is another action group being passed to the `Page` component, then this other action group will be used to roll up action items (instead of using the default `More Actions`). Again, all of this is happening inside the `Actions` component and on smaller screens there is no issue at all.

I am not sure if this is a regression or not, but what this PR is trying to do is to provide a way to listen for when a rollup happens. This will allow us to manually introduce `More Actions` as an action group from the other side only when a rollup is happening. There might be other use cases for this callback in the future but this is one we are trying to solve now: if there is one action group present then that is the one being used to roll up action items into, and this is not always desirable to have all action items rolling up into an arbitrary action group, because this existing action group can have nothing to do with the action items, or even worse it could be `disabled` making the action items completely hidden inside a `disabled` action group.

If you have any further questions, feel free to reach out or let me know! 😄

kyledurand (approved):

kemaltatar (approved):
Looks good to me as it is, but do you think we need a micro-optimization like this?

- Trigger the callback **only** when the value of `newShowableActions.length < actionsAndGroups.length - 1` change

kemaltatar:
```js
    if (isRollingUp) {
      const isRollupActive =
        newShowableActions.length < actionsAndGroups.length - 1;
      // Note: Do not include last group actions since we are skipping lastMenuGroup above
      // as it is always rendered with its own actions
      if (rollupActiveRef.current !== isRollupActive) {
        isRollingUp(isRollupActive);
        rollupActiveRef.current = isRollupActive;
      }
    }
```

kemaltatar:
And this on the top:
```js  
const rollupActiveRef = useRef<boolean | null>(null);
```

polaris-react/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -19,13 +19,16 @@ export interface ActionMenuProps {
   rollup?: boolean;
   /** Label for rolled up actions activator */
   rollupActionsLabel?: string;
+  /** Callback that returns true when secondary actions are rolled up into action groups, and false when not */
+  isRollingUp?(hasRolledUp: boolean): void;
```
chloerice:
```suggestion
  onActionRollup?(hasRolledUp: boolean): void;
```

chloerice:
> > Hey @jorgenunezsiri 👋🏽 I'm confused by what I'm seeing in the screen capture with the actions sometimes rolling up into a named group. They should all roll up into an ellipsis on small screens. Can you elaborate or show us an example of the issue you're facing with the rollup actions? It sounds like this is a complicated fix to a problem that shouldn't be happening (a regression that was missed after a past change to the Page Header, there have been several this year) 🧐
> 
> @chloerice Thank you for your comment!
> 
> About action items rolling up into an ellipsis button, that is what is happening nowadays, but this only happens on small screens (as you said). So, we currently have the [RollupActions](https://github.com/Shopify/polaris/blob/main/polaris-react/src/components/ActionMenu/ActionMenu.tsx#L44) component that it is being rendered when the [navigation is collapsed](https://github.com/Shopify/polaris/blob/main/polaris-react/src/components/Page/components/Header/Header.tsx#L147) (on smaller screens). For any other case, we are using the [Actions](https://github.com/Shopify/polaris/blob/main/polaris-react/src/components/ActionMenu/ActionMenu.tsx#L50) component. This `Actions` component has a [More Actions](https://github.com/Shopify/polaris/blob/main/polaris-react/src/components/ActionMenu/components/Actions/Actions.tsx#L48) action group, where action items roll up by default. However, the issue is if there is another action group being passed to the `Page` component, then this other action group will be used to roll up action items (instead of using the default `More Actions`). Again, all of this is happening inside the `Actions` component and on smaller screens there is no issue at all.
> 
> I am not sure if this is a regression or not, but what this PR is trying to do is to provide a way to listen for when a rollup happens. This will allow us to manually introduce `More Actions` as an action group from the other side only when a rollup is happening. There might be other use cases for this callback in the future but this is one we are trying to solve now: if there is one action group present then that is the one being used to roll up action items into, and this is not always desirable to have all action items rolling up into an arbitrary action group, because this existing action group can have nothing to do with the action items, or even worse it could be `disabled` making the action items completely hidden inside a `disabled` action group.
> 
> If you have any further questions, feel free to reach out or let me know! 😄

Thanks for the extra context, @jorgenunezsiri! The partial rollup is not a regression. I dug around in the history of the changes in this file and found [the PR that introduced the partial rollup](https://github.com/Shopify/polaris/pull/3420). The partial rollup should not be augmenting consumer provided action groups, it should only be rolling up any that don't fit into a "More actions" group that's defined internally as `defaultRollupGroup`. So there must be a bug with the `defaultRollupGroup` related logic causing the last provided rollup to act as the default rollup group instead of the internally defined `defaultRollupGroup`. I'm digging further to see if I can find the problem 👀

polaris-react/src/components/ActionMenu/ActionMenu.tsx
jorgenunezsiri: @chloerice That is definitely a better name for the callback. I have now updated all the callback occurrences with this name. 😄

jorgenunezsiri:
> do you think we need a micro-optimization like this?
Trigger the callback only when the value of newShowableActions.length < actionsAndGroups.length - 1 change

@kemaltatar Thank you for your feedback! That is definitely better, to only have the callback firing when the value actually changes. I have now made the change! 😄

jorgenunezsiri:
> I'd really like us to test the return value at the Page.test.tsx level but haven't found a good way to mock screen width. Not a blocker but would definitely be a required follow up IMO

@kyledurand I have been trying to write such a test but I have not been successful at all. I am afraid this will be very hard to achieve within Jest. So, I was able to mock the screen width by doing:
```JS
jest.spyOn(window.screen, "width", "get").mockReturnValue(1024);
```
However, that mock did not cause any changes because we are relying on the [offsetWidth](https://github.com/Shopify/polaris/blob/be84b8203130eef06341dc174cc9fe18d1bfe8c8/polaris-react/src/components/ActionMenu/components/Actions/Actions.tsx#L171) of [this div element](https://github.com/Shopify/polaris/blob/be84b8203130eef06341dc174cc9fe18d1bfe8c8/polaris-react/src/components/ActionMenu/components/Actions/Actions.tsx#L301) to be able to execute the [measureActions callback](https://github.com/Shopify/polaris/blob/be84b8203130eef06341dc174cc9fe18d1bfe8c8/polaris-react/src/components/ActionMenu/components/Actions/Actions.tsx#L94). In the end, I believe this is a problem outside of this change, meaning that it is not that we are not testing when this callback gets called, it is more about not having tests that are testing the rollup functionality at all, which would include testing the behavior of both [measuredActions.showable](https://github.com/Shopify/polaris/blob/be84b8203130eef06341dc174cc9fe18d1bfe8c8/polaris-react/src/components/ActionMenu/components/Actions/Actions.tsx#L47) and [measuredActions.rolledUp](https://github.com/Shopify/polaris/blob/be84b8203130eef06341dc174cc9fe18d1bfe8c8/polaris-react/src/components/ActionMenu/components/Actions/Actions.tsx#L48).

kyledurand:
> In the end, I believe this is a problem outside of this change

I agree, I recently tried probably the same things you did and found the same outcome. Before this I was pairing with a Sr. Staff dev and we couldn't figure it out then either. Let's leave it for now

PR #6023, opened 2022-06-03
[Icons]  Adds major and minor Insecure icon

itsluke:

Fixes #1515

We have lovely icon we use to talk about security (see pic attached) however due to the fact we're normally telling people their X is secure it's only available with a :heavy_tick: inside. This fixes this by adding an icon to be used to demonstrate risk.

Icons


polaris-icons/icons/SecureMajor.svg
```diff
@@ -1 +1,3 @@
-<svg viewBox="0 0 20 20" xmlns="http://www.w3.org/2000/svg"><path fill-rule="evenodd" d="M9.128.233c-2.37 1.383-5.37 2.33-7.635 2.646-.821.115-1.495.79-1.493 1.62l.001.497c-.03 6.043.477 11.332 9.462 14.903a1.45 1.45 0 001.062 0c8.993-3.571 9.503-8.86 9.473-14.903v-.501c-.001-.828-.674-1.51-1.492-1.638-2.148-.337-5.281-1.274-7.65-2.628a1.733 1.733 0 00-1.728.004zm4.577 8.478a1 1 0 00-1.414-1.415L8.998 10.59 7.705 9.297A1 1 0 106.29 10.71l2 2.001a1 1 0 001.414 0l4-4.001z" fill="#5C5F62"/></svg>
+<svg width="20" height="20" viewBox="0 0 20 20" xmlns="http://www.w3.org/2000/svg">
```
alex-page:
```suggestion
<svg viewBox="0 0 20 20" xmlns="http://www.w3.org/2000/svg">
```

@itsluke we should probably not change this SVG. Also the other files should follow the same format. No width/height and no color provided. The `<Icon>` component or css can change the icon color.

itsluke: Whoops thanks for catching that, I didn't mean to push the changes to `SecureMajor.svg`

PR #5803, opened 2022-05-13
feat: Improve emitted TypeScript declarations for Polaris tokens

aaronccasanova:
### WHAT is this pull request doing?

This PR introduces a `createTokens` identity function to validate and infer Polaris tokens. The result is a major improvement to the developer experience when using the `tokens` object as TypeScript will now emit type declarations matching the exact shape of our `tokens`. Thus enabling the most accurate type checking and intellisense.

[...]

PR #6045, opened 2022-06-03
Move size-limit to the root of Polaris

aaronccasanova:

This PR moves `size-limit` to the root of Polaris and simplifies the `size-limit.yml` workflow.

Co-authored-by: @BPScott

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist


.github/workflows/size-limit.yml
```diff
@@ -17,9 +17,7 @@ jobs:
         with:
           node-version: 16
           cache: yarn
-      - run: yarn --frozen-lockfile
-      - run: yarn build
```
aaronccasanova, line 21: This is already handled by the `size-limit-action` and in fact allows a custom `build_script` which we leverage to only build `@shopify/polaris` and its dependencies.

.github/workflows/size-limit.yml
```diff
@@ -17,9 +17,7 @@ jobs:
         with:
           node-version: 16
           cache: yarn
-      - run: yarn --frozen-lockfile
```
aaronccasanova, line 20: This is also already handled by the `size-limit-action`.

.github/workflows/size-limit.yml
```diff
@@ -17,9 +17,7 @@ jobs:
...
-      - run: yarn build
-      - uses: andresz1/size-limit-action@v1.5.1
+      - uses: andresz1/size-limit-action@v1.7.0
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          directory: polaris-react/
```
aaronccasanova, line 25: Removing `directory: polaris-react` should fix the `size-limit` failures in [PR #5803](https://github.com/Shopify/polaris/pull/5803) as `yarn build` was only being executed in `polaris-react` where what we actually need is to run a build for `polaris-react` and all its dependencies.

BPScott (approved):
Looks good to me. Size-limit is currently failing as it can't see a root level config when it checks out the main branch - which makes sense as there is no config in the root level package.json on main currently.

With this move to the main branch i wonder if it'd be interesting to introduce  sizelimit stats for polaris-tokens / polaris-icons too.

kyledurand (approved):

aaronccasanova:
> introduce sizelimit stats for polaris-tokens / polaris-icons too.

Great idea! I added an issue (https://github.com/Shopify/polaris/issues/6049) to handle that in a follow up tasks 🙌

sam-b-rose (approved):

alex-page (approved):

PR #6052, opened 2022-06-06
[polaris.shopify.com reboot] Refactor layout

martenbjork:
A second take at a refactored Layout component.

Enables the following layouts:

- Nav + Content
- Nav + Content + TOC

Changes:

- Familiar API: Just wrap your page in a `<Layout />` component
- The right level of flexibility — title and TOC are optional, but the rest is enforced
- Reduces amount of code and some duplication


martenbjork:
I'll rudely merge this to unblock the responsification of the site, but I look forward to discussion this more with y'all!

PR #6062, opened 2022-06-07
[Polaris website refresh] Add keyboard shortcut to trigger search

### LINKED ISSUE #5908: [Polaris website refresh] Global: Add keyboard shortcut to trigger search
Use a similar convention to https://shopify.dev/ where you press `/` and triggers the search model

chazdean:

Fixes [#5908](https://github.com/Shopify/polaris/issues/5908)

Adds global shortcut (command + K) that opens search and sets focus

https://user-images.githubusercontent.com/59836805/172496371-e3b6a817-b0df-460c-a444-72d2526255f3.mov

### 🎩 checklist

### DIFF
--- a/polaris.shopify.com/src/components/GlobalSearch/GlobalSearch.tsx
+++ b/polaris.shopify.com/src/components/GlobalSearch/GlobalSearch.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useState, useEffect } from "react";
 import { search } from "../../utils/search";
 import {
   GroupedSearchResults,
@@ -54,6 +54,7 @@ function GlobalSearch({}: Props) {
     getComboboxProps,
     highlightedIndex,
     getItemProps,
+    openMenu,
   } = useCombobox({
     id: "global-search",
     items: resultsInRenderedOrder,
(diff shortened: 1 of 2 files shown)


polaris.shopify.com/src/components/GlobalSearch/GlobalSearch.tsx
```diff
@@ -72,6 +73,21 @@ function GlobalSearch({}: Props) {
 
   let resultIndex = -1;
 
+  useEffect(() => {
+    document.addEventListener("keydown", (event) => {
+      const searchbar = document.getElementById("searchbar");
```
martenbjork: - Maybe this ID should be called global-search to align with the component name?
- Maybe it should be extracted into a const at the top to reduce the risk of someone changing 1 but not both?

martenbjork (approved):
Looks good!

Maybe we should switch the keyboard shortcut to be `/` to align with shopify.dev. What do you think?

polaris.shopify.com/src/components/GlobalSearch/GlobalSearch.tsx
```diff
@@ -72,6 +73,21 @@ function GlobalSearch({}: Props) {
...
+      let isKKey = event.key === "k" || event.key === "K";
```
romellogoodman: You could also write this as 
```
let isKKey = event.key.toLowerCase() === "k";
```

romellogoodman (approved):

chazdean:
> Looks good!
> 
> Maybe we should switch the keyboard shortcut to be `/` to align with shopify.dev. What do you think?

I agree, I went ahead and added that change

I think in the future we can look to add a `/` icon into the searchbar as well similar to shopify.dev so users know right away

polaris.shopify.com/src/components/GlobalSearch/GlobalSearch.tsx
chazdean: Good idea, took care of that 👍🏾

polaris.shopify.com/src/components/GlobalSearch/GlobalSearch.tsx
chazdean: Good to know, this is cleaner!

PR #6068, opened 2022-06-08
Remove dark color scheme support

### LINKED ISSUE #5994: Remove dark color scheme from tokens package
Remove dark:

- [x] Remove `colors.dark` from tokens
- [x] Remove scripts and functions for creating multiple schemes

Improve usage:

- [ ] Export tokens key/value pairs as default (no metadata)
- [x] Provide separate export for metadata for improved tree-shaking

sam-b-rose:

Depends on #6043 

Closes #5994

Remove dark:

- [x] Remove `colors.dark` from tokens
- [x] Remove scripts and functions for creating multiple schemes

Improve usage:

- [x] Provide separate export for metadata for improved tree-shaking
- [x] Export tokens key/value pairs as default (no metadata)

```tsx
// All tokens
import { tokens } from '@shopify/polaris-tokens';

const primary = tokens.colors.background; // 'rgba(246, 246, 247, 1)'

// Tree shaking support
import { colors } from '@shopify/polaris-tokens';

const primary = colors.background; // 'rgba(246, 246, 247, 1)'

// Metadata
import { metaTokens } from '@shopify/polaris-tokens';

const primaryValue = metaTokens.colors.background.value; // 'rgba(246, 246, 247, 1)'
const primaryDescription = metaTokens.colors.background.description; // 'For use as a background color, in components such as Page and Frame backgrounds.'
```


polaris-tokens/scripts/toStyleSheet.ts
```diff
@@ -97,15 +50,9 @@ export async function toStyleSheet(
   }
 
   const staticCustomProperties = getStaticCustomProperties(tokens);
-  const colorSchemeDeclarations = getColorSchemeDeclarations(
-    'light',
-    tokens,
-    osColorSchemes,
-  );
-  const defaultDeclarations = `${colorSchemeDeclarations}${staticCustomProperties}`;
+  const defaultDeclarations = `${staticCustomProperties}`;
```
aaronccasanova:
```suggestion
  const defaultDeclarations = getStaticCustomProperties(tokens);
```

polaris-tokens/src/utilities.ts
```diff
@@ -40,10 +40,8 @@ export function getKeyframeNames(motionTokenGroup: TokenGroup) {
  * Result: ['--p-background', '--p-text', etc...]
  */
 export function getCustomPropertyNames(tokens: Tokens) {
-  const {colorSchemes, ...restTokenGroups} = tokens;
   const customPropertyNames = [
-    ...Object.keys(colorSchemes.light).map((token) => createVar(token)),
-    ...Object.entries(restTokenGroups)
+    ...Object.entries(tokens)
```
aaronccasanova: **suggestion:** We're no longer concatenating token groups
```js
export function getCustomPropertyNames(tokens: Tokens) {
  return Object
    .entries(tokens)
    .map(([_, tokenGroup]: [string, TokenGroup]) =>
      Object.keys(tokenGroup).map((token) => createVar(token))
    )
    .flat()
  }
```

polaris-tokens/tests/toStyleSheet.test.js
```diff
@@ -26,13 +23,8 @@ const mockMotionTokenGroup = {
...
 };
 
-const mockColorSchemes = {
-  light: mockTokenGroup,
-  dark: mockTokenGroup,
-};
-
 const mockTokens = {
-  colorSchemes: mockColorSchemes,
+  colors: {},
```
aaronccasanova, line 27: Suggest moving this under the comment

polaris-tokens/scripts/toStyleSheet.ts
```diff
@@ -1,67 +1,23 @@
...
-/**
- * Creates CSS declarations for a given color-scheme.
- */
-export function getColorSchemeDeclarations(
-  colorScheme: ColorScheme,
-  tokens: Tokens,
-  osColorSchemes: OSColorSchemes,
-) {
-  return [
-    `color-scheme:${osColorSchemes[colorScheme]};`,
```
aaronccasanova, line 60: Perhaps, we should add this to the `:root` selector to give the OS a hint how to render built-ins.

polaris-tokens/src/tokens.ts
```diff
@@ -1,8 +1,7 @@
 import {breakpoints} from './token-groups/breakpoints';
 import {depth} from './token-groups/depth';
 import {legacy as legacyTokens} from './token-groups/legacy';
-import {colors as darkColorScheme} from './token-groups/color.dark';
-import {colors as lightColorScheme} from './token-groups/color.light';
+import {colors} from './token-groups/color';
```
aaronccasanova:
```suggestion
import {colors} from './token-groups/colors';
```
**light suggestion (non-blocking)**

polaris-tokens/src/tokens.ts
```diff
@@ -66,7 +35,7 @@ export interface Tokens {
 
 export const tokens: Tokens = {
   breakpoints: tokensToRems(breakpoints),
-  colorSchemes,
```
aaronccasanova, line 70: Note: There may be a few more lingering instances in Polaris (e.g. `polaris-for-vscode/src/server.ts`)

aaronccasanova (commented):
This is coming together nicely 🙌 Transforms are going to be much cleaner now that `tokens` contain only a single level of `token groups`! Left a few small suggestions

polaris-tokens/src/tokens.ts
sam-b-rose: Should we make them all plural? Seems like we currently have a mix:

**Singular**: color, depth, motion, shape, zIndex

**Plural**: breakpoints, legacyTokens, spacing (?)

polaris-tokens/scripts/toStyleSheet.ts
sam-b-rose: Looks like the `:root` selector is [added here](https://github.com/Shopify/polaris/pull/6068/files#diff-69b734df0611ae7ed41fe371499590065e1ec9ee094231731f9317c001f30ae9R53). Does that accomplish this?

aaronccasanova: > Does that accomplish this?

Yup! Assuming you mean moving `color-scheme: light` there.

polaris-tokens/src/tokens.ts
aaronccasanova: I personally have a preference for plural token groups. While our structure is slightly different, I still like to think of token groups as properties on a theme object e.g. `tokens.colors.primary`. It won't work for all token groups, like the ones mentioned above, but I think we should use plural names where possible.

polaris-tokens/rollup.config.mjs
```diff
@@ -9,39 +9,62 @@ const pkg = JSON.parse(
...
-    }),
-  ],
-  external: [
-    ...Object.keys(pkg.dependencies ?? {}),
-    ...Object.keys(pkg.peerDependencies ?? {}),
-  ],
-};
+export default [
+  {
+    input: 'build-internal/index.ts',
```
BPScott: You can have an array of inputs. It looks like the output and other options are the same so it seems like you could do the following instead of needing multiple configs:

```
export default {
  input: ['src/index.ts', 'build-internal/metadata.ts'],
  outputs,
  plugins,
  external
}
```

polaris-tokens/rollup.config.mjs
```diff
@@ -9,39 +9,62 @@ const pkg = JSON.parse(
...
+        entryFileNames: '[name][assetExtname].mjs',
+        dir: path.dirname(pkg.module),
+        preserveModules: true,
+      },
+    ],
+    plugins,
+    external,
+  },
+  {
+    input: 'build-internal/metadata.ts',
```
BPScott: Check out what `polaris-icons` does with `rollup-plugin-virtual` to create a virtual entrypoint - this would allow you to dynamically generate the contents of this metadata file  as part of the rollup build instead of needing a two step process where you to generate metadata.ts and then call rollup (and thus no need to deal with the creation andtidying of build-internal folder)

aaronccasanova: IIRC the reasoning for generating `index.ts` and `metadata.ts` in `build-internal` is to still allow TypeScript to validate and emit declarations. Do you have suggestions or examples of how to achieve that with `rollup-plugin-virtual` (without manually generating the TS declarations)?

sam-b-rose: I did explore the virtual file route but ran into issue with TS consistency. I've cleaned up this PR to remove those exploratory commits, but after discussion with @aaronccasanova and @alex-page this approach made the most sense and resulted in the least complexity

polaris-tokens/rollup.config.mjs
sam-b-rose: Great suggestion, updated! Thanks Ben 👍

documentation/guides/migrating-from-v9-to-v10.md
```diff
@@ -81,3 +81,25 @@ const App = (props) => (
 +  </div>
 )
 ```
+
+## Tokens
+
+The dark color scheme palette has been removed. Dark color tokens are now consolidated into a single colors palette.
```
aaronccasanova:
```suggestion
The dark color scheme palette has been removed. Dark color tokens are now consolidated into a single color palette.
```

polaris-tokens/scripts/index.ts
```diff
@@ -1,8 +1,9 @@
-import {osColorSchemes, tokens} from '../src';
+import {tokens} from '../src';
 
+import {toTokenValues} from './toTokenValues';
 import {toJSON} from './toJSON';
 import {toStyleSheet} from './toStyleSheet';
 
 (async () => {
-  await Promise.all([toJSON(tokens), toStyleSheet(tokens, osColorSchemes)]);
+  await Promise.all([toTokenValues(), toJSON(tokens), toStyleSheet(tokens)]);
```
aaronccasanova:
```suggestion
  await Promise.all([toTokenValues(tokens), toJSON(tokens), toStyleSheet(tokens)]);
```

Should `toTokenValues` also accept a reference to the `tokens` object?

polaris-tokens/scripts/toJSON.ts
```diff
@@ -1,37 +1,26 @@
 import fs from 'fs';
 import path from 'path';
 
-import {Tokens, TokenProperties} from '../src';
+import {MetaTokens, TokenGroup} from '../src';
```
aaronccasanova, line 4:
```suggestion
import type {MetaTokens, TokenGroup} from '../src';
```

polaris-tokens/scripts/toJSON.ts
```diff
@@ -1,37 +1,26 @@
...
     await fs.promises.mkdir(outputDir, {recursive: true});
   }
 
   for (const entry of Object.entries(tokens)) {
-    const [tokenGroup, tokenProps] = entry as [string, TokenProperties];
+    const [tokenGroupName, tokenGroup] = entry as [
+      keyof MetaTokens,
+      TokenGroup,
+    ];
+    const fileName = getFileName(`${tokenGroupName}.json`);
```
aaronccasanova:
```suggestion
    const filePath = path.join(outputDir, `${tokenGroupName}.json`);
```

**thought:** `getFileName` is kind of an unnecessary abstraction now, what do you think?

polaris-tokens/scripts/toStyleSheet.ts
```diff
@@ -1,71 +1,27 @@
 import fs from 'fs';
 import path from 'path';
 
-import {Tokens, ColorScheme, TokenGroup, OSColorSchemes} from '../src';
+import {MetaTokens, MetaTokenGroup} from '../src';
```
aaronccasanova, line 4:
```suggestion
import type {MetaTokens, MetaTokenGroup} from '../src';
```

polaris-tokens/scripts/toStyleSheet.ts
```diff
@@ -78,34 +34,23 @@ export function getCustomProperties(tokenGroup: TokenGroup) {
...
-  const colorSchemeDeclarations = getColorSchemeDeclarations(
-    'light',
-    tokens,
-    osColorSchemes,
-  );
-  const defaultDeclarations = `${colorSchemeDeclarations}${staticCustomProperties}`;
   const styles = `
-  :root{${defaultDeclarations}}
-  ${getColorSchemeRules(tokens, osColorSchemes)}
+  :root{${getStaticCustomProperties(tokens)}}
```
aaronccasanova:
```suggestion
  :root{color-scheme: light;${getStaticCustomProperties(tokens)}}
```

I think we should still provide a hint to the browser: https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme

polaris-tokens/src/types.ts
```diff
@@ -37,3 +37,32 @@ export type Exact<
     } & {
       [Key in Exclude<keyof InputType, KeysOfUnion<ParameterType>>]: never;
     };
+
+export interface MetaTokenProperties {
```
aaronccasanova: Perhaps we should move these above line 1 since it states the following types are copied directly from `type-fest`

alex-page (approved):
I thought VSCode and polaris.shopify.com used the token metadata? I was hoping to see those imports change. I might be missing something though.

polaris-tokens/tests/toStyleSheet.test.js
aaronccasanova: Nudge (:

polaris-tokens/scripts/toTokenValues.ts
```diff
@@ -0,0 +1,95 @@
...
+    const result = valueTransforms[tokenGroupName].reduce(
+      (str, fn) => str.replaceAll(valueRegex, (_, p1) => `'${fn(p1)}',`),
+      keyValuePairs,
+    );
+
+    await fs.promises.writeFile(tokenFile, result);
+  }
+}
+
+export async function toTokenValues() {
```
aaronccasanova: I have a rough idea of how we can simplify this transform. I'll work out a POC and report back.

aaronccasanova: Had a quick pair with @samrose3 and the [results look promising](https://github.com/Shopify/polaris/pull/6068/commits/78f0bdcafa7560dd63b695a2287d1cae6a6a44f9). I basically reworked [toTokenValues.ts](https://github.com/Shopify/polaris/blob/78f0bdcafa7560dd63b695a2287d1cae6a6a44f9/polaris-tokens/scripts/toTokenValues.ts) to generate a single `build-internal/tokens.ts` file which is then passed to Rollup (to build `cjs` and `esm` outputs) and TypeScript (to build type declarations).

Here is an example of the output file:

image

Note: `toTokenValues.ts` ensures the generated values are valid `TokenGroup`s and thus we apply `as const` assertions to each export to get accurate TypeScript declarations

@samrose3 Is going to take it from here and make sure the package.json and types exports are working as expected. cc @alex-page

aaronccasanova (approved):

polaris.shopify.com/src/pages/api/v0/tokens/[tokens].tsx
```diff
@@ -1,24 +1,12 @@
...
-  tokens,
-} from "@shopify/polaris-tokens";
+import { Tokens, TokenGroup, createVar, tokens } from "@shopify/polaris-tokens";
 import type { NextApiRequest, NextApiResponse } from "next";
 
 /**
  * Color scheme independent token groups
  */
-type StaticTokenGroupKey = Exclude<keyof Tokens, "colorSchemes">;
+type TokenGroupKey = Exclude<keyof Tokens, "colorSchemes">;
```
Comment (aaronccasanova) [inline polaris.shopify.com/src/pages/api/v0/tokens/[tokens].tsx:7]:
```suggestion
type TokenGroupKey = keyof Tokens;
```

PR #6082, opened 2022-06-08
[polaris.shopify.com reboot] Fix search issue

martenbjork:
The current implementation of the "press `/` to trigger search" contained a bug that would make the page unresponsive. This PR fixes the issue :)

### DIFF
--- a/polaris.shopify.com/src/components/GlobalSearch/GlobalSearch.tsx
+++ b/polaris.shopify.com/src/components/GlobalSearch/GlobalSearch.tsx
@@ -85,7 +85,6 @@ function GlobalSearch({}: Props) {
           searchbar.focus();
         }
       }
-      console.log("a");
     });
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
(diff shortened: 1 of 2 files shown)


polaris.shopify.com/src/components/GlobalSearch/GlobalSearch.tsx
```diff
@@ -77,16 +77,17 @@ function GlobalSearch({}: Props) {
   useEffect(() => {
     document.addEventListener("keydown", (event) => {
       const searchbar = document.getElementById(globalSearchID);
-      let isKKey = "/";
```
martenbjork, line 80: This condition is always true.

polaris.shopify.com/src/components/GlobalSearch/GlobalSearch.tsx
```diff
@@ -77,16 +77,17 @@ function GlobalSearch({}: Props) {
...
         event.preventDefault();
         openMenu();
         if (searchbar !== null) {
           searchbar.focus();
         }
       }
     });
-  });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
martenbjork, line 90: If you don't include a dependency array, the hook will be run every time the component rerenders. In this case, it means that the `addEventListener` will be bound over and over again.

PR #6083, opened 2022-06-09
[Badge][Pip] Remove default `status` value

chloerice:

Fixes the `progress` `Badge` with no `status` set having a blue pip instead of gray.

- Removes the default `status` of `info` from `Pip`
- Re-simplifies the types for `Badge` `status` and `progress` props so they're parsable by `typedoc`
- Fixes the a11y label utility so that instead of always returning the label for status and progress, it outputs the correct label based on whether `status` and/or `progress` are set

|Before|After|
|---|---|
|Screen Shot 2022-06-09 at 1 04 30 PM|Screen Shot 2022-06-09 at 1 04 02 PM|
|Screen Shot 2022-06-09 at 2 07 36 PM|Screen Shot 2022-06-09 at 2 07 11 PM|
|Screen Shot 2022-06-09 at 2 07 43 PM|Screen Shot 2022-06-09 at 2 07 18 PM|

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/polaris-react/src/components/Badge/types.ts
+++ b/polaris-react/src/components/Badge/types.ts
@@ -1,3 +1,11 @@
+export type Status =
+  | 'info'
+  | 'success'
+  | 'warning'
+  | 'critical'
+  | 'attention'
+  | 'new';
+
 export enum StatusValue {
   Info = 'info',
   Success = 'success',
@@ -6,12 +14,13 @@ export enum StatusValue {
   Attention = 'attention',
   New = 'new',
 }
-// uses template literal types to get the string representation of the values of the FieldsMap enums
-export type Status = `${StatusValue}`;
+
+export type Progress = 'incomplete' | 'partiallyComplete' | 'complete';
+
 export enum ProgressValue {
   Incomplete = 'incomplete',
   PartiallyComplete = 'partiallyComplete',
   Complete = 'complete',
 }
-export type Progress = `${ProgressValue}`;
+
 export type Size = 'small' | 'medium';
--- a/.changeset/clever-scissors-lay.md
+++ b/.changeset/clever-scissors-lay.md
@@ -1,0 +1,5 @@
+---
+'@shopify/polaris': patch
+---
+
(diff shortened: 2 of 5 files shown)


polaris-react/src/components/Badge/types.ts
```diff
@@ -1,17 +1,10 @@
...
-  New = 'new',
-}
-// uses template literal types to get the string representation of the values of the FieldsMap enums
-export type Status = `${StatusValue}`;
-export enum ProgressValue {
-  Incomplete = 'incomplete',
-  PartiallyComplete = 'partiallyComplete',
-  Complete = 'complete',
-}
-export type Progress = `${ProgressValue}`;
```
chloerice, line 16: This was unnecessary and causing typedoc not to parse the types.

bernbecht:
Should this PR be a `patch`? When I was trying to fix the same thing in this [PR](https://github.com/Shopify/polaris/pull/5840) I was oriented that it was a breaking change and needed to be a major

bernbecht (approved):

chloerice:
> Should this PR be a `patch`? When I was trying to fix the same thing in this [PR](https://github.com/Shopify/polaris/pull/5840) I was oriented that it was a breaking change and needed to be a major

The discussion around breaking change was part of the misunderstanding of what Badge is supposed to look like when it does and doesn't have a status, which got slightly messed up (and you partially fixed with that PR) when the Pip was extracted from Badge.

This PR puts the Badge back at where it was pre-Pip extraction:
Default (no `status`) => Gray background, with gray pip signaling `progress` if set
Info (`status="info"`) => Blue background, with blue pip signaling `progress` if set
New (`status="new"`) => Gray background, with gray pip signaling `progress` if set (it's identical to the default badge except for having a higher font-weight, but historically was it's own color variant of teal prior to the new design language)

alex-page (approved):

PR #6096, opened 2022-06-09
[Custom properties] Add data- to p-color-scheme body attribute

dleroux:

p-color-scheme isn't a valid attribute for the body tag. 

Simply added the `data-` prefix.

### 🎩 checklist

### DIFF
--- a/polaris-react/src/components/CustomProperties/CustomProperties.tsx
+++ b/polaris-react/src/components/CustomProperties/CustomProperties.tsx
@@ -43,7 +43,7 @@ export function CustomProperties(props: CustomPropertiesProps) {
 
   return (
     <Component
-      p-color-scheme={colorScheme}
+      data-p-color-scheme={colorScheme}
       className={className}
       style={{color: 'var(--p-text)', ...style}}
     >
--- a/.changeset/real-pants-look.md
+++ b/.changeset/real-pants-look.md
@@ -1,0 +1,5 @@
+---
+'@shopify/polaris': patch
+---
+
+Prefixed body attribute with data-
--- a/polaris-react/src/components/AppProvider/AppProvider.tsx
+++ b/polaris-react/src/components/AppProvider/AppProvider.tsx
@@ -98,7 +98,7 @@ export class AppProvider extends Component<AppProviderProps, State> {
     // Inlining the following custom properties to maintain backward
     // compatibility with the legacy ThemeProvider implementation.
     document.body.setAttribute(
-      'p-color-scheme',
+      'data-p-color-scheme',
       this.props.colorScheme || DEFAULT_COLOR_SCHEME,
     );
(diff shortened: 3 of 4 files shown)


kyledurand (commented):
Looks like we'll also need to make this change in:
`polaris-tokens/scripts/toStyleSheet.ts`
`polaris-tokens/tests/toStyleSheet.test.js`

That should fix the changed UI Tests

polaris-react/src/components/CustomProperties/CustomProperties.tsx
```diff
@@ -43,7 +43,7 @@ export function CustomProperties(props: CustomPropertiesProps) {
 
   return (
     <Component
-      p-color-scheme={colorScheme}
+      data-p-color-scheme={colorScheme}
```
aaronccasanova, line 46: This would be a breaking change as overriding custom properties with an attribute selector is the recommended approach for upgrading from the `ThemeProvider` to `CustomProperties` component in the [v8 to v9 migration guide](https://github.com/Shopify/polaris/blob/main/documentation/guides/migrating-from-v8-to-v9.md#replacing-custom-properties-from-theme-overrides).

aaronccasanova:
> p-color-scheme isn't a valid attribute for the body tag.

Mind linking a resource? I thought attribute names simply had to be XML compatible.

https://html.spec.whatwg.org/#attributes-2

> **Attribute names** must consist of one or more characters other than [controls](https://infra.spec.whatwg.org/#control), U+0020 SPACE, U+0022 ("), U+0027 ('), U+003E (>), U+002F (/), U+003D (=), and [noncharacters](https://infra.spec.whatwg.org/#noncharacter).

> Attribute names are said to be XML-compatible if they match the [Name](https://www.w3.org/TR/xml/#NT-Name) production defined in XML and they contain no U+003A COLON characters

Are you encountering a specific issue with `p-color-scheme` on the body? @samrose3 is currently working on removing color schemes from the system (including this `p-color-scheme` attribute)

dleroux:
If you run it through a [validator](https://validator.w3.org/) it will fail. But if it's on the way out and it's might as well leave it for now? https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

BPScott:
> Mind linking a resource? I thought attribute names simply had to be XML compatible.

Correct but to ensure userland code doesn't conflict with some future html-spec provided attributes it says that user-provided attributes must start with `data-`, while anything else is reserved for attributes provided by future versions of the html spec.

https://html.spec.whatwg.org/#embedding-custom-non-visible-data-with-the-data-*-attributes


See also the idea that custom elements must have a hypen in them - the `something-somethingelse` is reserved for useland provided custom components, while tag names without a hypen in them are reserved for tags provided by future versions of the html spec.

PR #6113, opened 2022-06-13
Convert breakpoint before and after mixins to scss variables

lgriffee:

Prerequisite of upgrade strategy for updating media queries to use the new Polaris breakpoints aliases.

- Converts existing `breakpoint-before` and `breakpoint-after` mixins to Sass variables
- Creates Sass variables for any hardcoded media conditions to help with upgrade strategy


.prettierignore
```diff
@@ -3,3 +3,4 @@ dist
 node_modules
 /polaris-react/build
 /polaris-react/build-internal
+/polaris-react/src/styles/shared/_breakpoints.scss
```
aaronccasanova, line 6: Temporarily ignoring the `_breakpoints.scss` file as the formatting was [making the Scss variables harder to read](https://github.com/Shopify/polaris/blob/437c28ef294e7948d73252a8eff2fedc1e2d5958/polaris-react/src/styles/shared/_breakpoints.scss#L28-L55).

polaris-react/.storybook/GridOverlay/GridOverlay.css
```diff
@@ -15,8 +15,8 @@
   opacity: 0.25;
   pointer-events: none;
 }
-
-@media screen and (min-width: 768px) {
+/* #{$p-breakpoints-sm-up} */
+@media (min-width: 768px) {
```
alex-page, line 19: For these ones should we just use the new variable?

alex-page (approved):

polaris-react/.storybook/GridOverlay/GridOverlay.css
aaronccasanova, line 19: Added a comment of the associated breakpoint alias to allow the media condition to show up in search results.

polaris-react/src/styles/shared/_breakpoints.scss
```diff
@@ -1,41 +1,55 @@
...
 
-$page-max-width: $layout-width-primary-max + $layout-width-secondary-max +
-  $layout-width-inner-spacing-base;
+@function breakpoints-up($value) {
+  @return '(min-width: #{breakpoint($value)})';
+}
+
+@function breakpoints-down($value, $inclusive: false) {
+  @return '(max-width: #{breakpoint($value, $inclusive)})';
+}
```
aaronccasanova, line 26: While reviewing these `mixins to variables` updates we caught multiple typos and find/replace issues (human error) that justified temporarily introducing `breakpoints-up` and `breakpoints-down` Scss functions. The breakpoint functions follow the same mobile first strategy used in the `gen-media-queries` script where `down` conditions apply a `-0.05px` offset to avoid conflicts with `up` conditions.

One benefit of `breakpoints-down` setting `$inclusive: false` by default is we can now clearly see `down` conditions that are not following our recommended approach as they will be explicitly defined with `$inclusive: true`

e.g. `$some-legacy-breakpoint: breakpoints-down(555px, $inclusive: true)`

polaris-react/.storybook/GridOverlay/GridOverlay.css
aaronccasanova: We could but would have to convert this file to SCSS. @kyledurand I forget, was there a reason we wrote this in vanilla CSS?

alex-page: I guess storybook doesn't build scss without plugins... That makes sense.

aaronccasanova (approved):

Issue #6132, opened 2022-06-15
[polaris.shopfiy.com reboot] "Jump to main content" should be hidden until focussed


martenbjork:
Right now, it sneaks into view (see dark mode).

<img width="263" alt="image" src="https://user-images.githubusercontent.com/875708/173836400-1dcb0d5b-1842-4428-9c53-a47c579170be.png">

PR #6135, opened 2022-06-15
[polaris.shopify.com reboot] Edit skip-to-content button

chazdean:

[#6132](https://github.com/Shopify/polaris/issues/6132)

Fixes bug with current skip-content-button showing in dark-mode

button-bug

Moves content button further up Y axis and replaces background colour with `var(--surface)` to make sure that tokenized colour is used


romellogoodman (approved):
🎩 'd and looks good to me!

PR #6137, opened 2022-06-15
Bump size limit for polaris-react-*

sam-b-rose:

Size limit check was failing due to bundle size increases.

Increases the size limit for each `polaris-react-*` bundle limit.

Before | After
--- | ---
before | after

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,17 +76,17 @@
     {
       "name": "polaris-react-cjs",
       "path": "polaris-react/build/cjs/index.js",
-      "limit": "180 kB"
+      "limit": "205 kB"
     },
     {
       "name": "polaris-react-esm",
       "path": "polaris-react/build/esm/index.js",
-      "limit": "105 kB"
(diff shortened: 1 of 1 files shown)


alex-page:
@samrose3 those sizes are quite a large leap. Is there a smaller jump we could take right now?

sam-b-rose:
> Is there a smaller jump we could take right now?

@alex-page Yes, I'll reduce to just a small bit over what is currently being emitted

alex-page (approved):

chloerice (approved):
Thanks Sam!! 🚀

aaronccasanova (approved):
TY!

BPScott:
Question: Being 20-30kb over the size limits feels like a lot - the size-limit stuff started failing relatively recently so it's not really a build up of small things, it seems like one bigger change shot up the size. Are we doing anything to investigate what that was and workout alternative solutions?

PR #6141, opened 2022-06-15
[polaris.shopify.com] refactor: create one source of truth for navItems

gui-santos:

The side navigation in mobile needs to have the same links that sections like "Foundations" and "Components" have, otherwise users in smaller screens won't be able to navigate to the different pages in those sections

The final goal is to create a mobile navigation with accordions like this:

alt_B_1
_(this is just a DRAFT, it won't look like this in the end)_

BUT for that to happen, first I'm unifying the "NavItems" components that are used in `Nav` and in `SideNav` so we can have only one place to update navigation

In this PR, you won't see any visual change... At least, if my changes worked hahahahaha
but you will see that both `Nav` and in `SideNav` are using the same types and data
 
- ~~moved NavItem type to NavItems component~~
- Renamed `NavItem` component to `NavListItem` so TS doesn't mistake it for the type that has the same name
- made NavItems use the data from data/navItems
- removed some unused HTML and CSS in Page.tsx
- removed the `useRouter` from `_app.tsx` because it was only passing props to `Page` and `Page` was already internally calling it for other reasons
- moved `<div style={{ background: isPolaris ? "#fafafa" : "unset" }}>` to Page.tsx

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/.changeset/sour-penguins-roll.md
+++ b/.changeset/sour-penguins-roll.md
@@ -1,0 +1,5 @@
+---
+'polaris.shopify.com': patch
+---
+
+Create one source of truth for the side navigation in every screen size
--- a/.changeset/sour-penguins-roll.md
+++ b/.changeset/sour-penguins-roll.md
@@ -2,4 +2,5 @@
 'polaris.shopify.com': patch
 ---
 
-Create one source of truth for the side navigation in every screen size
+- Remove unused HTML and CSS from website components
+- Use data from /data/navItems in the Header component of the website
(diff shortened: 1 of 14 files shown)


alex-page (approved):

martenbjork (approved):
💯

PR #6186, opened 2022-06-16
Refactor react perf profiler for local dev, local data collection

kyledurand:

https://polaris-coverage.shopifycloud.com/profiler isn't collecting data. This is the first step to refactor the way we collect data

### DIFF
--- a/polaris-react/.storybook/manager.js
+++ b/polaris-react/.storybook/manager.js
@@ -71,13 +71,13 @@ function GridPanel(props) {
             name: 'Show grid overlay',
             description:
               'Show or hide a 4 / 12 column grid, overlaying components',
-            defaultValue: false,
+            defaultValue: 'false',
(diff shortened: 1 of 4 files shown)


polaris-react/.storybook/manager.js
```diff
@@ -71,13 +71,13 @@ function GridPanel(props) {
             name: 'Show grid overlay',
             description:
               'Show or hide a 4 / 12 column grid, overlaying components',
-            defaultValue: false,
+            defaultValue: 'false',
```
alex-page, line 74: It doesn't work with booleans?! 😰

alex-page (approved):

polaris-react/.storybook/manager.js
kyledurand: I tested with both, just making this change for consistency

PR #6226, opened 2022-06-17
[polaris.shopify.com] feat: mobile navigation uses accordions to group links

### LINKED ISSUE #5903: [Mobile-friendly] Implement navigation of the website in small screens

gui-santos:

Fixes #5903 

Screenshot 2022-06-17 at 14 29 58

- Replaces `SideNav` with `MobileNav` to make it more clear what that component does
- Add accordions to the `MobileNav` to handle the links that users find in Desktop when they are pages like "Foundations", "Components", "Contributing", etc.
- Delete `NavItems` component and move its logic to `Header` to lower code complexity in the navigation

<details>
<summary>Keyboard navigation in the accordions</summary>

</details>

### What is this pull request _NOT_ doing?

Things that are coming in a following PR:

- UP/DOWN arrow navigation
- Automagically 🌈  generate links from files in `/content`

### ⚠️ Open questions (a.k.a I don't know what to do)

You will see that I added an accordion for each group of links that exists in Desktop side navs. For example: we have Foundations, Design, Content, and Patterns groups in Foundations section of the website

Screenshot 2022-06-17 at 15 27 12

Maybe you've noticed already the problem this creates: there is no link for the "index" page of foundations the one accessed in `/foundations`. What should we do?

### How to 🎩

### 🎩 checklist


polaris.shopify.com/src/components/Header/Header.module.scss
```diff
@@ -173,12 +173,12 @@
...
 }
 
-.SideNavContainer {
+.MobileNavContainer {
   visibility: hidden;
   display: none;
-  fill: var(--text-strong);
 
   svg {
+    fill: var(--text-strong);
```
martenbjork, line 161: This could probably just be `var(--text)`. Right now it's very stark compared to the text.

polaris.shopify.com/src/components/MobileNav/MobileNav.module.scss
```diff
@@ -26,23 +20,36 @@ $navIconSize: 24px;
...
   }
+}
+
+.LinksList {
+  display: flex;
+  align-items: flex-start;
+  flex-direction: column;
+  height: 100%;
+  width: calc(100vw - 4rem);
+  max-width: 22.5rem;
```
martenbjork: This `22.5rem` value makes a few appearances. Extract into a global CSS variable? `--mobile-nav-max-width`

polaris.shopify.com/src/components/MobileNav/MobileNav.module.scss
```diff
@@ -26,23 +20,36 @@ $navIconSize: 24px;
...
+  min-width: 15rem;
+  background: var(--surface);
+  font-size: var(--font-size-100);
+  font-weight: 500;
```
martenbjork: Use a font weight token 😊

polaris.shopify.com/src/components/MobileNav/MobileNav.tsx
```diff
@@ -0,0 +1,247 @@
...
+    function hideSideNavOnResize() {
+      if (window.innerWidth > Breakpoints.Desktop && showMenu) {
+        setShowMenu(false);
+      }
+    }
+
+    window.addEventListener("resize", hideSideNavOnResize);
+
+    return () => window.removeEventListener("resize", hideSideNavOnResize);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
martenbjork, line 58: I'm always suspicious of disabled linting 👀 Why do we not want to include `showMenu` among the dependencies here?

polaris.shopify.com/src/components/MobileNav/MobileNav.tsx
```diff
@@ -0,0 +1,247 @@
...
+              <a
+                ref={logoRef}
+                className={styles.Logo}
+                onClick={handleCloseMenu}
+                onKeyDown={(e) => {
+                  if (e.key === "Tab" && e.shiftKey) {
+                    e.preventDefault();
+                    const closeButton = closeButtonRef.current;
+                    closeButton instanceof HTMLElement && closeButton.focus();
+                  }
```
martenbjork: Room for debate, but I can see this being extracted into a function above the return. That way, the JSX part of the component becomes easy to read which is inviting to designer/developer types.

polaris.shopify.com/src/components/MobileNav/MobileNav.tsx
```diff
@@ -0,0 +1,247 @@
...
+          aria-label="Close menu"
+          className={styles.CloseButton}
+          onClick={handleCloseMenu}
+          onKeyDown={(e) => {
+            if (e.key === "Tab" && !e.shiftKey) {
+              e.preventDefault();
+              const logoLink = logoRef.current;
+              logoLink instanceof HTMLElement && logoLink.focus();
+            }
+          }}
```
martenbjork: Same as above, maybe move above the return?

polaris.shopify.com/src/types.ts
```diff
@@ -81,5 +81,8 @@ export interface HighlightableSearchResult {
 }
 
 export enum Breakpoints {
-  SMALL = 768,
+  Mobile = 500,
+  Tablet = 768,
+  Desktop = 1400,
+  DesktopLarge = 1600,
```
martenbjork, line 94: For another day, but it would be nice to find a way to keep these in sync with the corresponding Sass variables.

martenbjork (changes requested):
SO GOOD! 😻 Played with this on my phone and it simply feels great.

I've added some nit-picky comments to the code. Other than that, here are some small things I noticed:

### Overscrolling
When you reach the end of the nav and keep scrolling ("overscrolling") the whole page behind the backdrop starts to scroll. You can prevent this using this CSS property:

https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior

### Focus visible
On mobile, i see a dark focus outline in the nav when it's opened. I think this might be a mistake that I did — we should use `focus-visible` for those links, not `focus-within`.

Image from iOS

### Touch callouts
There is a lot of visual feedback when you touch an item. On iOS, webkit adds a "touch callout" to links which is not needed if you provide your own `:active` styles. Maybe remove them?

https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-touch-callout

### Text selectability
Not sure if there are a11y implications, but it might be good to do `user-select: none;` on the items in the nav so that you can long-press (when thinking) without selecting text.

### Item height
Items in the nav are currently 48 ish pixels tall. I think we can reduce that towards 44px and still have decent sized touch targets.

martenbjork:
> Maybe you've noticed already the problem this creates: there is no link for the "index" page of foundations the one accessed in /foundations. What should we do?

Ignore this problem for now 🤪 We can add it to the back-log.

polaris.shopify.com/src/components/MobileNav/MobileNav.tsx
gui-santos: Because if we pass `showMenu` in the dependencies we will add/remove the event listener every time that prop changes values

When we pass an empty dependency array to `useEffect`, react will only call that effect when "mounting" the component

This listener only needs to be added once in the first load of a page and never more. It also only covers a very edge case of a user who opens the website in a small browser and resizes it while the mobile nav is opened

polaris.shopify.com/src/types.ts
gui-santos: yeah that would be cool
But I'm not sure how to use TS values in Sass or vice-versa 🤔

polaris.shopify.com/src/components/MobileNav/MobileNav.tsx
gui-santos: Do you mean something like creating a `handleShiftTabPress` function?

polaris.shopify.com/src/components/MobileNav/MobileNav.tsx
gui-santos: Do you mean something like creating a `handleTabPress` function?

gui-santos:
Thanks a lot for the review @martenbjork 🙌 
I will address all of your comments and the feedback regarding Overscrolling, focus-visible, Text selectability, and the Item height :)

> > Maybe you've noticed already the problem this creates: there is no link for the "index" page of foundations the one accessed in /foundations. What should we do?
> 
> Ignore this problem for now 🤪 We can add it to the back-log.

Cool! I think it's fine, those pages serve as navigation to their content and while they are not there users can use the mobile nav to do the same thing they would in those pages
So we are not blocking users :D

polaris.shopify.com/src/components/MobileNav/MobileNav.tsx
martenbjork: Great explaination! Thank you!

polaris.shopify.com/src/components/MobileNav/MobileNav.tsx
martenbjork: Yeah!

polaris.shopify.com/src/types.ts
martenbjork: Same! Maybe we can write a node script to create these values. Another day!

polaris.shopify.com/src/components/MobileNav/MobileNav.tsx
martenbjork: Yeah!

gui-santos:
Hey @martenbjork 
I've addressed all the suggestions, but there are two of them that I may need more clarification:

> ### Focus visible
> On mobile, i see a dark focus outline in the nav when it's opened. I think this might be a mistake that I did — we should use `focus-visible` for those links, not `focus-within`.
> 
> Image from iOS
> 

I changed to `focus-visible`, but the black border still shows to me. That border is actually the focus for keyboard
If you open the menu using your keyboard (Tabbing to the burger menu and then pressing Enter), the menu opens and I move the focus to the first item of the menu (Polaris logo) and that makes the black border show up. I do this for a11y reasons, otherwise people using the keyboard would have to press Tab again to start navigating the menu

The black border is the style of the keyboard focus that the website already had before my changes, I didn't change anything. But we can re-style how keyboard focus looks like of course 🙌 

Right now, It is being defined by [this lines](https://github.com/Shopify/polaris/blob/main/polaris.shopify.com/src/styles/globals.scss#L122-L124) 

> ### Item height
> Items in the nav are currently 48 ish pixels tall. I think we can reduce that towards 44px and still have decent sized touch targets.

The height of the items is actually defined by their padding, the padding before was `padding: 0.85rem 1rem` and it was making height to be ~52px. I changed it to `padding: 0.75rem 1rem` and now the height is ~48px, but if a decrease the padding to `padding: 0.5rem 1rem` that makes the height too short 40px

Is it ok to keep it `padding: 0.75rem 1rem` and `48px`?

Cheers

martenbjork (approved):
Worth saying again: this is great work! Gonna merge.

Issue #6155, opened 2022-06-15
[Polaris website reboot] Light/Dark mode icon gets covered by search bar


sarahill:
The light/dark mode icon is getting covered by the search bar when the browser width is `1400px` or less

<img width="305" alt="image" src="https://user-images.githubusercontent.com/8629173/173934547-a40502d1-ff92-4327-b2e9-0d65ffbb066c.png">

PR #6227, opened 2022-06-17
[Polaris website reboot] Fix/Update Top Nav Breakpoints & Styling

### LINKED ISSUE #6154: [Polaris website reboot] Make Top nav collapse at a smaller breakpoint
The top nav is collapsing at around `1400px` right now

I would expect the nav to stay open until at least `768px` and smaller.

Image

chazdean:

Fixes #6154 and #6155 

Screen Shot 2022-06-17 at 8 58 29 AM

- The top nav was collapsing at 1400px, now it stays open until 768px and smaller.
- Dark Mode toggle also no longer hidden when searchbar expands below 768px and smaller.

#### Concerns

- Navbar may still be a bit too crowded as the searchbar and dark toggle cant be shown with all of the buttons at smaller desktop sizes. #6214 May solve this issue

### DIFF
--- a/.changeset/fresh-ducks-return.md
+++ b/.changeset/fresh-ducks-return.md
@@ -1,0 +1,5 @@
+---
+'polaris.shopify.com': patch
+---
+
+Update top nav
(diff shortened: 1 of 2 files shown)


yurm04:
in the gif example above the top nav will cut off the search bar and light/dark mode button. I think the dark mode button can be removed for launch, but is there a way so that the search bar does not get cut off at the in-between screen size?

yurm04 (approved):

PR #6230, opened 2022-06-17
[Polaris website reboot] Make `Getting Started` the new homepage

### LINKED ISSUE #6140: [Polaris website refresh] `Getting Started` and `Homepage` do not add value as separate pages
In their current state, Getting Started and the homepage do essentially the same thing. Let's...

- Get rid of Homepage and keep Getting Started OR
- Find a way to make them both distinct and high value pages

yurm04:
resolves [6140](https://github.com/Shopify/polaris/issues/6140)

removed from top nav and `pages/resource`. Figure we can keep the `ResourcePage` component so that we can come back to it after launch

### DIFF
--- a/.changeset/two-crabs-rule.md
+++ b/.changeset/two-crabs-rule.md
@@ -2,4 +2,4 @@
 'polaris.shopify.com': patch
 ---
 
-Remove homepage and make Getting Started the new site index page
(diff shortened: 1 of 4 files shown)


gui-santos (approved):
🙌

PR #6297, opened 2022-06-22
[polaris.shopify.com] Generate open graph images

martenbjork:
Took a stab at easily creating open graph images for the site.

- Reverse engineers the slug to create the title
- Grabs the preview images for components if they are present
- Uses `next-sitemap` to create a map of all the pages on the site and then uses those URLs to generate images

### Dependency alert!

Run this locally to get all the things that node-canvas need: `brew install pkg-config cairo pango libpng jpeg giflib librsvg`

components--color-picker
components--modal
components--account-connection
components--combobox
components--banner
foundations--content--merchant-to-customer
tokens--typography
foundations
foundations--design--design
tokens
home
icons
components--grid


polaris.shopify.com/src/scripts/og-images/index.js
```diff
@@ -0,0 +1,165 @@
...
+
+  if (url.startsWith("/tokens/")) {
+    title = `${title} tokens`;
+  }
+
+  if (title === "") {
+    title = "Polaris";
+  }
+
+  return title;
```
gui-santos: Uhmmm in line 138 here we are reassigned the value ("Polaris") that we have already assigned in line 127

I think we can use the `slug` instead of `title` in this "if" statement and then we do not need to do line 131 or reassign "Polaris" to title, something like this:

```suggestion
  let title = "Polaris";
  const urlSegments = url.split("/");
  const slug = capitalizeFirstLetter(urlSegments[urlSegments.length - 1]).replace(/-/g, " ");

  if (slug) {
    const isToken = url.startsWith("/tokens/")
    title = `${slug}${isToken && ' tokens'}`;
  }

  return title;
```

What do you think?

gui-santos (approved):
Wow! This looks pretty cool 🙌 

I made just one suggestion, but I don't think it’s a blocker

polaris.shopify.com/src/scripts/og-images/index.js
martenbjork: Thanks! I took some hints from what you suggested and rewrote it a bit. 

The only reason I want to keep the...

```
 if (url.startsWith("/tokens/")) {
    title = `${title} tokens`;
  }
```

...around is that it's easy to just add more `if`s down the road if we want more title hacks. 🤷‍♀️

Let's clean this up more down the road! 👀

PR #6308, opened 2022-06-22
[polaris.shopify.com reboot] Update tokens content for accuracy, plus a small edit

selenehinkley:
Paired with @jjgali on the tokens content based on feedback from @alex-page. We think this new version summarizes the technical complexity in plainer language.


martenbjork (approved):
Looks good! Thanks! 🚀

PR #6309, opened 2022-06-22
Delete vocabulary.md

selenehinkley:
The vocabulary page was originally created when Polaris.Shopify.com launched and our terminology set at Shopify was small. Shopify has grown, and this format doesn't scale. At one point, we thought the terms might be useful on our site for 3p developers, but we now have .dev as a resource for these information needs. There is also a project within Shopify around terminology management to solve this problem, plus we have an existing internal spreadsheet ([product glossary](https://docs.google.com/spreadsheets/d/1YZaIAnRmiKr9r9bmUrFstZho4jNKr8Rp2IOT65l4JuE/edit#gid=0)) that is the most up to date terminology home that crafters use today.

FYI @jjgali


alex-page:
This also needs to ship on the old site `polaris-styleguide`.

selenehinkley:
[PR for polaris-styleguide](https://github.com/Shopify/polaris-styleguide/pull/5934)

@alex-page Is there a way to efficiently scan for any potential cross links to this page from the rest of the sites?

alex-page:
@selenehinkley it's usually just a search through code. I will have some time to help after launch, otherwise reach out to folks they can help!

martenbjork:
I'll just bump this one, Alex, since you're thinking about content management this week.

polaris.shopify.com/content/foundations/content/naming.md
```diff
@@ -693,8 +693,6 @@ confusing.
 
 Before naming a new product or feature:
 
-- Conduct an audit of our [vocabulary list](/content/vocabulary#navigation) to
```
selenehinkley, line 696: Could we replace what's being deleted here with:
- Conduct an audit of existing names to narrow down your naming choices

alex-page (approved):

martenbjork:
I've never seen that spreadsheet before, and I bet that's true for many crafters at Shopify. I think it makes sense for Polaris to own it.

We shouldn't just put it into a markdown file though. We can easily build a dedicated page type that feels more like a dictionary. That would also enable us to show dictionary results in the global search in a nice way. Anyone interested in me spending 20 minutes coming up with a mock-up?

selenehinkley:
Hey @martenbjork, @deannatron is away at a Burst, but when she's back I'll connect you two. She's a Content Design Lead who is looking into terminology management and the glossary at Shopify. If you're both interested I think it'd be great to collab on this together. She can at least provide the most up to date context on this area.

deannatron:
Hey @martenbjork! Looking forward to connecting—I'll set up some time to share more context!

The latest update is that Shopify recently hired a Localization Program Manager, Terminology (Laura Di Tullio) and Shopify also recently got a terminology management system (TMS) approved. This has been driven by folks in international languages since glossaries are critical for achieving high translation quality. I think this will be an amazing partnership.

On the content design side there are so many team-specific glossaries floating around. The way I see it, at this point the Shopify Product glossary is just one of many spreadsheets we need to reconcile, since a lot of unique terms (probably hundreds) are living in team or product area glossary spreadsheets. To give you an idea, for Hack Days I did a quick analysis of the Core glossaries I could find, and on average only 15% of their terms are also added to the Shopify Product Glossary.

martenbjork:
Hi @deannatron! Wow, that's super interesting! And a great opportunity for systematizing this and making it more consistent and accessible for everyone. Looking forward to chatting more!

PR #6323, opened 2022-06-22
[polaris.shopify.com] Set root font size as percentage

aveline:
Small fix to set root font size as a percentage so that user browser font size preferences are supported


martenbjork (approved):
Nice! Thanks! 🚀

PR #6344, opened 2022-06-23
Add profiler scripts

kyledurand:
https://user-images.githubusercontent.com/6844391/176679744-4b307bf4-7156-4774-928c-35a27f722587.mp4

Goes hand in hand with https://github.com/Shopify/polaris-coverage-shopifycloud-com/pull/78

### DIFF
--- a/polaris-react/.storybook/preview.js
+++ b/polaris-react/.storybook/preview.js
@@ -56,24 +56,24 @@ function ReactRenderProfiler(Story, context) {
 export const globalTypes = {
   strictMode: {
     name: 'React.StrictMode',
-    defaultValue: 'true',
+    defaultValue: true,
     toolbar: {
       items: [
-        {title: 'Disabled', value: 'false'},
-        {title: 'Enabled', value: 'true'},
+        {title: 'Disabled', value: false},
+        {title: 'Enabled', value: true},
       ],
-      showName: 'true',
+      showName: true,
     },
   },
   profiler: {
     name: 'React.Profiler',
-    defaultValue: 'false',
+    defaultValue: false,
     toolbar: {
       items: [
-        {title: 'Disabled', value: 'false'},
-        {title: 'Enabled', value: 'true'},
+        {title: 'Disabled', value: false},
+        {title: 'Enabled', value: true},
       ],
-      showName: 'true',
+      showName: true,
     },
   },
 };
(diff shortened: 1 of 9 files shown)


polaris-react/.storybook/preview.js
```diff
@@ -56,24 +56,24 @@ function ReactRenderProfiler(Story, context) {
 export const globalTypes = {
   strictMode: {
     name: 'React.StrictMode',
-    defaultValue: 'true',
```
kyledurand, line 59: These globals always seem to act weird via params especially. This doesn't seem to matter but they're not strings in the docs

polaris-react/playground/KitchenSink.tsx
```diff
@@ -11,17 +13,32 @@ const readmeReq = require.context(
...
-      .map(([storyName, Story]) => (
-        <StoryWithWrapper key={`${readmeModule.default.title}:${storyName}`}>
-          <Story />
-        </StoryWithWrapper>
-      ));
+      .map(([storyName, Story]) => {
+        const id = `${readmeModule.default.title}:${storyName}`;
+        return (
+          <div key={id}>
+            <RenderPerformanceProfiler
```
kyledurand, line 26: Need to explicitly render the perf profiler in the kitchen sink for a couple of reasons

1. Using the query param is flakey
2. The data brought back from the generic profiler is too generic. Every id is just playground/playground and there's no story id / name available

alex-page: Lets try and take a look at fixing the query param. We want that to be stable for local development and testing. 🤞

kyledurand: > The data brought back from the generic profiler is too generic. Every id is just playground/playground and there's no story id / name available

The query param isn't that flakey, but the grid one was because it wasn't getting included in the global scope for context. I've fixed that but we still need the render performance wrapper around the kitchen sink. I've somewhat fixed the issue of duplicating profiles being logged by printing out the data to a div via a prop that only gets added on the kitchen sink. 

So we have the option to print data to the DOM or log it out

polaris-react/.storybook/preview.js
kyledurand: Digging deeper it seems that these options sometimes resolve as typeof boolean, sometimes as string. Super strange

alex-page (approved):

polaris-react/scripts/profiler.js
```diff
@@ -0,0 +1,18 @@
...
+  'http://localhost:6006/iframe.html?id=playground-playground--kitchen-sink&args=&viewMode=story&globals=profiler:false';
+
+(async () => {
+  const browser = await puppeteer.launch();
+  const page = await browser.newPage();
+  await page.goto(iframePath, {waitUntil: 'networkidle2', timeout: 0});
+
+  const data = await page.evaluate(
+    () => document.getElementById('render-performance-profiler').innerHTML,
+  );
```
alex-page, line 13: This is super clean. Nice!

Issue #5714, opened 2022-05-04
Update Polaris breakpoint utilities and components within the MD breakpoint range


lgriffee:
Update breakpoint utilities between `768px-1039px`

PR #6365, opened 2022-06-23
feat: Update Dialog to Polaris MD media conditions

lgriffee:
Part of #5714    

Updating `Dialog` to use Polaris MD media conditions.

#### Checklist
- What Polaris media condition was used?
  - From: `@media (min-height: $height-limit + $vertical-spacing)`
  - To: `@media #{$p-breakpoints-md-up}`
- Did the breakpoint value change? `Yes`
  - From: `660px`
  - To: `768px`
- Was the breakpoint variable, function, or mixin used elsewhere in Polaris? `No`
- Was the breapoint variable, function, or mixin used elsewhere in Web? `Yes`
  - Search pattern: `height-limit.+vertical-spacing`
- Is the layout using a mobile first strategy? `Yes`
 
#### Before/After Examples
**Before**
_Note: The first few seconds of this video is showing that the layout is not changing at 660px._

https://user-images.githubusercontent.com/21976492/175395299-06ed3ab2-6372-4740-ae92-1d55de5f3ed0.mp4

**After**

https://user-images.githubusercontent.com/21976492/175395319-029e2594-9c8b-465e-a248-cef8d6e98d18.mp4

### DIFF
--- a/polaris-react/src/components/Modal/components/Dialog/Dialog.scss
+++ b/polaris-react/src/components/Modal/components/Dialog/Dialog.scss
@@ -12,8 +12,6 @@ $breakpoints-xsmall-width-up: breakpoints-up(
   $xsmall-width + $dangerous-magic-space-16
 );
 
-$breakpoints-height-limit-up: breakpoints-up($height-limit + $vertical-spacing);
-
 .Container {
   position: fixed;
   z-index: var(--p-z-11);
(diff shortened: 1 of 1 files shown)


polaris-react/src/components/Modal/components/Dialog/Dialog.scss
```diff
@@ -68,9 +66,7 @@ $breakpoints-height-limit-up: breakpoints-up($height-limit + $vertical-spacing);
 
   &.limitHeight {
     @media #{$p-breakpoints-md-up} {
-      @media #{$breakpoints-height-limit-up} {
```
lgriffee, line 71: I would be replacing this media condition with `@media #{$p-breakpoints-md-up} {` which would duplicate the media condition above it so I just removed it.

lgriffee:
Closing this one since it's height related and we don't want to replace it.

PR #6376, opened 2022-06-24
[Polaris website refresh] Update Search field to clear when closed

chazdean:

The search field will now clear after search window has been closed or after user clicks the search results 

<details>
      <summary>After Changes</summary>
      search-bar clear
    </details>

### DIFF
--- a/polaris.shopify.com/src/components/GlobalSearch/GlobalSearch.tsx
+++ b/polaris.shopify.com/src/components/GlobalSearch/GlobalSearch.tsx
@@ -103,6 +103,12 @@ function GlobalSearch({}: Props) {
     };
   }, [setIsOpen, router.events]);
 
+  useEffect(() => {
+    if (!isOpen) {
+      setSearchTerm("");
+    }
+  }, [isOpen]);
+
   const handleKeyboardNavigation: KeyboardEventHandler<HTMLDivElement> = (
     evt
   ) => {
--- a/.changeset/polite-shrimps-bake.md
+++ b/.changeset/polite-shrimps-bake.md
@@ -1,0 +1,5 @@
+---
+'polaris.shopify.com': patch
+---
+
+Add clear searchbar field on close


polaris.shopify.com/src/components/GlobalSearch/GlobalSearch.tsx
```diff
@@ -103,6 +103,12 @@ function GlobalSearch({}: Props) {
     };
   }, [setIsOpen, router.events]);
 
+  useEffect(() => {
+    if (!isOpen) {
+      setSearchTerm("");
+    }
+  }, [isOpen]);
```
alex-page, line 110: We probably don't want to add a useEffect for this functionality. It would be best if when the user clicked the background it automatically closed the modal and reset the search state.

What about... creating a new function called `handleClose` to replace what happens on line #L206 in the `onClick`. This function can close the modal `setIsOpen(false)` and clear the search box `setSearchTerm("")` all at once.


Let me know if you need any help! I really recommend reading https://beta.reactjs.org/learn/you-might-not-need-an-effect

martenbjork: I think useEffect is appropriate here since HeadlessUI handles a lot of the opening/closing. If we create a custom handler we have to make sure that we use it for:

- Any close buttons
- Clicking the backdrop
- Pressing escape

Any any other scenario that might trigger a close. IMO, using a useEffect here seems like a good way to simply extend "whatever decided to close the modal".

martenbjork (approved):

polaris.shopify.com/src/components/GlobalSearch/GlobalSearch.tsx
martenbjork: Headless UI provides an `onClose` prop, let's use that: https://headlessui.dev/react/dialog

That way, our handler will be run regardless of why the modal was closed (escape button, backdrop click etc).

chazdean: The reason why I opted to using useEffect as opposed to creating a handleClose function like you recommended is because the `onClick` and even the `onClose` prop for Headless UI only trigger when the model is closed by using the close button shown in the mobile view or clicking the background. But when clicking a search item it seems that the handler is not used.

Using useEffect was also an easier way to ensure that I was taking care of all the close events at once.

PR #6404, opened 2022-06-27
[polaris.shopify.com] Fix component example (page, narrow width)

### LINKED ISSUE #6384: [polaris.shopify.com] Component example is broken (Page, narrow width)
image

chazdean:

Fixes #6384  

The component example for Page (narrrow width) is no longer broken and can now be shown as a preview


martenbjork (approved):
Nice!

These examples should probably be full width, but we can do that in a separate PR since there are some different ways to achieve that. (See #6403)

PR #6398, opened 2022-06-25
[polaris.shopify.com][/contributing] Add quick start tables

chloerice:
This PR adds scannable contribution guidance to help folks take the first step.

Screen Shot 2022-06-29 at 5 14 00 PM

## To do

- Explore more concise ways of displaying this info (can also effectively serve as a way to capture data that helps us identify contribution patterns that point to gaps in the system)
  - [ ] Quick-win idea: a drop down for the contribution types (similar to the component examples) so you're only seeing the info you're looking for, eliminating the redundancy between rows
    - [x] As a two column table (https://github.com/Shopify/polaris/pull/6430)
    - [x] As a responsive description list
   - [ ] Design-thinking idea: short user story form acts as a way to dial in on exactly the info needed to take first contribution steps 
     - As a `CONTRIBUTOR TYPE` working on `COMMERCE CONCEPT`, I want to contribute `CONTRIBUTION TYPE` to make `STEP OF JOB TO BE DONE` a better experience for merchants across the Shopify Admin.
     - For example: As a `Shopifolk` working on `pricing and discounts`, I want to contribute `a new component` to make `inputing currency values` a better experience for merchants across the Shopify Admin.

PR #6430, opened 2022-06-28
[polaris.shopify.com] Polish contribution guidance

chloerice:

This PR: 
- Fixes `Layout` adding extra left and right content padding on small screens
- Adds a background to inline code snippets
- Polishes contribution guidance content
- Adds a hook for setting query params, to be used in a follow up PR to make component examples linkable

### How to 🎩

- `git checkout quickstart-table_concept-1`
- `yarn build && yarn turbo run dev --filter=polaris.shopify.com`
- Navigate to `localhost:3000/contributing`
- Things to look for in the /contributing section:
  - Broken or missing links
  - Typos
  - etc 

### 🎩 checklist


polaris.shopify.com/src/components/Layout/Layout.module.scss
```diff
@@ -17,6 +17,10 @@
   @media screen and (max-width: $breakpointTablet) {
     display: block;
   }
+
+  @media screen and (max-width: $breakpointMobile) {
+    padding: 0;
```
chloerice, line 22: This fixes small screen left and right padding being doubled up, as `Container` also adds `1rem` of left and right padding.

polaris.shopify.com/src/components/Longform/Longform.module.scss
```diff
@@ -161,6 +167,14 @@
     font-size: var(--font-size-200);
     border-collapse: collapse;
 
+    p {
```
chloerice, line 173: (Ensures paragraphs rendered in tables don't have longform content level spacing.)

polaris.shopify.com/src/components/Longform/Longform.module.scss
```diff
@@ -151,6 +151,15 @@
     }
   }
 
+  code {
```
chloerice, line 154: (This adds styles for inline code snippets)

polaris.shopify.com/src/components/QuickStartGuide/QuickStartGuide.tsx
```diff
@@ -0,0 +1,111 @@
...
+    routerIsReady && currentParams?.quickGuide
+      ? queryParamValues.indexOf(currentParams.quickGuide as string)
+      : "0";
+
+  const [guideIndex, setGuideIndex] = useState(defaultGuideIndex);
+
+  const handleChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    const nextIndex: string = event.target.value;
+    setGuideIndex(nextIndex);
+    setQueryParams({ quickGuide: `${queryParamValues[Number(nextIndex)]}` });
```
chloerice: @martenbjork When I'm back from holiday I'll put up a PR to add query params to examples so folks can link to a specific example too 😁

alex-page:
The content and direction this work is headed in is awesome.

I was a bit unsure the value we from some of the design decisions it felt like we were shipping a data structure. I also think the user experience could be pushed and evolved easier with a simpler approach.

1. The drop down switcher hides quick start guides and requires users to press them to see them. Users should be able to see all quick start guides without clicking a button.
2. Quick start guides don't get a lot of value from the table layout. They could be simplified to be markdown pages with heading structure.
3. The complexity of sharing these with them being a part of a select instead of a table isn't high value. Having to update query params isn't great.
4. Future quick start guides have to follow this structure and flexibility might be useful to make guides the best they can be for our users.

chloerice:
> The content and direction this work is headed in is awesome.
> 
> I was a bit unsure the value we from some of the design decisions it felt like we were shipping a data structure. I also think the user experience could be pushed and evolved easier with a simpler approach.
> 
> 1. The drop down switcher hides quick start guides and requires users to press them to see them. Users should be able to see all quick start guides without clicking a button.
> 2. Quick start guides don't get a lot of value from the table layout. They could be simplified to be markdown pages with heading structure.
> 3. The complexity of sharing these with them being a part of a select instead of a table isn't high value. Having to update query params isn't great.
> 4. Future quick start guides have to follow this structure and flexibility might be useful to make guides the best they can be for our users.

Thanks for this detailed feedback @alex-age!! 🙌🏽 

The goal with this design was to simplify the [initial table concept](https://github.com/Shopify/polaris/pull/6398) we drafted in the Google Doc by reusing the pattern used for component examples because it's:
- A pattern that consumers and contributors are used to that doesn't take up massive real estate on the page
- Allows the guidance to scale regardless of how many ways there are to contribute (illustrations: 1, documentation: 4)
- Cuts the redundancy of the content, as there's a lot of the same answers across contribution types
- Makes the framing of Who/What/Where/When/How info more human 
- Presents the info in a scannable format (inspired by breakout tables in textbooks 📚) in order to: 
  - Minimize overwhelm for new contributors
  - Speed up existing contributors 
  
The query params aren't needed for the select, they're prep for exploring the idea mentioned in the last bullet of the PR description. They're also there to make linking to relevant guidance more direct when supporting folks in Slack and on GitHub. It's also something I've been wanting to contribute to component examples forever for the same reason.

The data coming from frontmatter vs from within the Mardkown wasn't the initial plan, but was what I landed on to move quickly because we can't parse the individual parts of the table from ReactMarkdown (at least not from what I can tell 🧐). It's def not ideal, but I also don't think it should block shipping the MVP of the guidelines. We plan on making fast follows in the coming weeks to elevate the design of these pages to the level of the rest of the site. This feedback will be super useful as we iterate, especially now that we're gonna have a designer helping with this work (yaaay @raquelbreternitz 🎉).

alex-page:
That all makes sense @chloerice. My thought with markdown frontmatter was to remove the table all together. I think the table is limiting and increasing complexity.

polaris.shopify.com/content/contributing/illustrations.md
```diff
@@ -6,12 +6,48 @@ keywords:
...
+Illustration contributions are welcome from anyone that works at Shopify.
+
+Each illustration must be submitted for feedback from an illustrator in #illustration-guild on Slack. Polaris only supports illustrations within the Shopify Admin product and not marketing assets.
 
 ## Quick start guide
 
-table
+<!--
+
+This is a placeholder table, quick start guides render here. You can find and edit quick guide content in the metadata (front matter) at the top of this file. The placeholder table is here so the Markdown component that parses these files places the QuickStartGuide component here (see components/Markdown.tsx).
```
chloerice: > That all makes sense @chloerice. My thought with markdown was to remove the table all together. I think the table is limiting and increasing complexity.

@alex-page I think what you're referring to are [the Markdown tables from the PR this replaces](https://github.com/Shopify/polaris/pull/6398/files#r918024958)? We don't hold the content in that format anymore. The content lives in the front matter of each `/contributing` file, which gets passed to the `QuickStartGuide` component.

selenehinkley (approved):

polaris.shopify.com/content/contributing/documention.md
```diff
@@ -5,64 +5,161 @@ keywords:
...
+        answer: Copy edits can be contributed in the <a href="https://github.com/Shopify/polaris">Shopify/polaris</a> GitHub repo. Copy edits should also be contributed in Figma if relevant.
+      - relatedIssues:
+        question: Where should I search for related issues or topics?
+        answer: Search for related issues and pull requests in the <a href="https://github.com/Shopify/polaris">Shopify/polaris</a> GitHub repo.
+      - what:
+        question: What resources will my contribution impact?
+        answer: Along with related content sections on this documentation site, your contribution may impact the Figma UI Kit and component examples.
+      - how:
+        question: How can I get started?
+        answer: <a href="/contributing/shipping-your-contribution">Open a pull request</a> in the <a href="https://github.com/Shopify/polaris">Shopify/polaris</a> GitHub repo
```
alex-page: This is not a pattern we should scale on the website.

alex-page (changes requested):
The approach to quick guides content is complex and should be avoided across the website. We should create a much simpler initial solution that is just a HTML page using markdown and title hierarchy. I don't think the effort to maintain the frontmatter, parsing, QuickStartGuide.tsx etc. is worth the outcome for our users.

polaris.shopify.com/content/contributing/illustrations.md
alex-page: I added a comment to add some clarity below.

yurm04:
I agree with all the things you've brought up @alex-page , but we also want to ship this quickly since we’re still iterating on the design of all the guidelines, and the quick start table was the most difficult to come up with a great design on a tight deadline with no UXers. This does add more tech debt to the site right now but @raquelbreternitz  and @chloerice are going to be revisiting the design asap. Since all of these changes are contained they can be easily ripped out at that point. ATM I would consider this page a `Blocker for comms` and should be resolved before the website project end date. Until then, we need this page quickly to unblock internal comms so that we can start getting feedback from Shopifolk and collaborating with teams to contribute pattern guidance for a related project

alex-page (approved):
Approving. I don't feel great about the added technical debt for adding what should be simple pages. I will create a follow up PR to remove a bunch of unnecessary code.

yurm04 (approved):

PR #6445, opened 2022-06-29
Re-enable stylelint-polaris check and address misc errors

aaronccasanova:

This PR re-enables the `stylelint-polaris` check in Polaris and addresses misc errors introduced while the plugin was not active.

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist


sam-b-rose (approved):

package.json
```diff
@@ -27,7 +27,8 @@
   "scripts": {
     "build": "turbo run build",
     "dev": "turbo run dev --filter=@shopify/polaris... --parallel",
-    "lint": "turbo run lint",
+    "lint": "turbo run lint && yarn lint:styles",
+    "lint:styles": "stylelint **/*.{css,scss}",
```
sam-b-rose: Should we lint at the individual package level? That way we can leverage turbo's caching and only run stylelint on the modified packages.

alex-page (approved):

package.json
aaronccasanova: That sounds like a good follow up task 👍 We use a (minimal) [root stylelint config](https://github.com/Shopify/polaris/blob/main/.stylelintrc.js) that applies `styling-polaris/internal` to `polaris-react` and the public config to the rest of the repository. Thus, I'm not sure there's much value in breaking out to use package level configs. In regards to caching, I updated to use [the `--cache` stylelint flag](https://stylelint.io/user-guide/usage/cli#--cache) and the `ci.yml` workflow cache.

BPScott: I don't really mind if the repo goes for top-level linting or per-package linting, but we should pick one and stick to it. 

We already do per-package level linting for [eslint](https://github.com/Shopify/polaris/blob/main/polaris-react/package.json#L55), with having a root-level `.eslintrc`. As that already exists we should stick to that pattern, and then in a follow up we can plot "do we want top-level or per-package linter runs" instead of introducing a second way of handling linting here.


> stylelint caching

Stylelint's cache is flawed - [It does not re-report errors in files that were previously cached](https://github.com/stylelint/stylelint/issues/4715). This means you can run a lint, see some errors, then run lint again without changing anything and it passes. I'd suggest staying away from stylelint's --cache flag

aaronccasanova: Sounds good 👍 Removed root `lint:styles` in favor of [per package run-scripts](https://github.com/Shopify/polaris/pull/6445/commits/751a9c015c8431c36d18cee7950b1e383333b790).

.stylelintrc.js
```diff
@@ -1,5 +1,11 @@
+/** @type {import('stylelint').Config} */
 module.exports = {
   extends: ['@shopify/stylelint-plugin/prettier', './stylelint-polaris'],
+  ignoreFiles: [
```
BPScott: Can we put these ignores into https://github.com/Shopify/polaris/blob/main/.stylelintignore, so we've got one way of dealing with ignored files?

aaronccasanova: Discussed offline, stylelint is not leveraging the root level `.stylelintignore` with the recent update to use package level run-scripts, but does pick up the root level config. Thus, we decided to remove the `.stylelintignore` file in favor of the `ignoreFiles` config: https://github.com/Shopify/polaris/pull/6445/commits/50aa75a47902c98cbdf2b309f8a39260b5aa769f

.stylelintrc.js
```diff
@@ -9,11 +10,24 @@ module.exports = {
...
     },
   ],
+  ignoreFiles: [
+    '.next/**/*.{css,scss}',
+    'node_modules/**/*.{css,scss}',
+    'dist/**/*.{css,scss}',
+    'documentation/guides/legacy-polaris-v8-public-api.scss',
+    'polaris-react/build/**/*.{css,scss}',
+    'polaris-react/build-internal/**/*.{css,scss}',
+    'polaris-tokens/dist/**/*.{css,scss}',
```
BPScott: Is this already covered by `'dist/**/*.{css,scss}'` above?

BPScott (approved):
Looks good to me.

alex-page (approved):

.stylelintrc.js
aaronccasanova: It works with a [slight tweak](https://github.com/Shopify/polaris/pull/6445/commits/d610383abb0d07093df37fa122a8724a6bdf392e), thanks for the idea!

PR #6451, opened 2022-06-30
[Tooltip] Pin tooltip to cursor during hover

zakwarsame:

- Having an interactive tooltip that follows the cursor on hover is an idea proposed by Tobi and the quality team

Closes https://github.com/Shopify/shopify/issues/321500

- This PR wraps the content in a component (`ChildWrapperCommponent`) which tracks mouse movements on hover
- Introduce a string using the useState hook and use that to _transform_ the tooltip element
- Calculate the the difference in the mouse coordinates and center of the element to place the tooltip in the middle.
- Then push the left side by half the _overlay's_ width to where the cursor is

### How to 🎩

🖥 [Click here for a short demo video](https://screenshot.click/11-46-2038-45714.webm)

🖥 [Spin URL to test feature on the admin](https://shop1.shopify.tooltip-movement.zakaria-warsame.us.spin.dev/admin/products/1)
 > _Hover over the text formatting area to see feature in practice_

### 🎩 checklist

### DIFF
--- a/.changeset/chilled-impalas-sleep.md
+++ b/.changeset/chilled-impalas-sleep.md
@@ -1,0 +1,5 @@
+---
+'@shopify/polaris': minor
+---
+
+minor
--- a/polaris-react/src/components/Tooltip/Tooltip.tsx
+++ b/polaris-react/src/components/Tooltip/Tooltip.tsx
@@ -141,6 +141,7 @@ export function Tooltip({
     mouseEntered.current = false;
     handleBlur();
   }
+
   // https://github.com/facebook/react/issues/10109
   // Mouseenter event not triggered when cursor moves from disabled button
   function handleMouseEnterFix() {
--- a/polaris-react/src/components/Tooltip/Tooltip.tsx
+++ b/polaris-react/src/components/Tooltip/Tooltip.tsx
@@ -5,6 +5,7 @@ import {findFirstFocusableNode} from '../../utilities/focus';
 import {useUniqueId} from '../../utilities/unique-id';
 import {useToggle} from '../../utilities/use-toggle';
 import {Key} from '../../types';
+import {getRectForNode} from '../../utilities/geometry';
 
 import {TooltipOverlay, TooltipOverlayProps} from './components';
 
@@ -51,7 +52,7 @@ export function Tooltip({
     setFalse: handleBlur,
   } = useToggle(Boolean(originalActive));
   const [activatorNode, setActivatorNode] = useState<HTMLElement | null>(null);
-  const [positionTooltip, setPositionTooltip] = useState<TooltipPosition>({
+  const [coordinates, setCoordinates] = useState<TooltipPosition>({
     x: 0,
     y: 0,
     cursorX: 0,
@@ -92,7 +93,7 @@ export function Tooltip({
         active={active}
         accessibilityLabel={accessibilityLabel}
         onClose={noop}
-        coordinates={positionTooltip}
+        coordinates={coordinates}
       >
         {content}
       </TooltipOverlay>
@@ -151,8 +152,12 @@ export function Tooltip({
   function handleMouseMove(event: React.MouseEvent) {
     if (!active) return;
     if (childWrapperContainer.current == null) return;
-    const {x, y, width, height} =
-      childWrapperContainer.current.getBoundingClientRect();
+    const {
+      left: x,
+      top: y,
+      width,
+      height,
+    } = getRectForNode(childWrapperContainer.current) || {};
     const centerX = width / 2;
     const centerY = height / 2;
     const {clientX, clientY} = event;
@@ -163,7 +168,7 @@ export function Tooltip({
     const transformX = tooltipLeft - centerX;
     const transformY = tooltipTop - centerY;
 
-    setPositionTooltip({
+    setCoordinates({
       x: transformX,
       y: transformY,
       cursorX: clientX,
--- a/polaris-react/src/components/Tooltip/Tooltip.tsx
+++ b/polaris-react/src/components/Tooltip/Tooltip.tsx
@@ -92,7 +92,7 @@ export function Tooltip({
         active={active}
         accessibilityLabel={accessibilityLabel}
         onClose={noop}
-        transform={positionTooltip}
+        coordinates={positionTooltip}
       >
         {content}
       </TooltipOverlay>
@@ -153,14 +153,14 @@ export function Tooltip({
     const {x, y, width, height} =
       childWrapperContainer.current.getBoundingClientRect();
     const centerX = width / 2;
-    const ceterY = height / 2;
+    const centerY = height / 2;
     const {clientX, clientY} = event;
 
     const tooltipLeft = clientX - x;
     const tooltipTop = clientY - y;
 
     const transformX = tooltipLeft - centerX;
-    const transformY = tooltipTop - ceterY;
+    const transformY = tooltipTop - centerY;
 
     setPositionTooltip({
       x: transformX,
--- a/polaris-react/src/components/Tooltip/Tooltip.tsx
+++ b/polaris-react/src/components/Tooltip/Tooltip.tsx
@@ -8,6 +8,12 @@ import {Key} from '../../types';
 
 import {TooltipOverlay, TooltipOverlayProps} from './components';
 
+export interface TooltipPosition {
+  x: number;
+  y: number;
+  cursorX: number;
+}
+
 export interface TooltipProps {
   /** The element that will activate to tooltip */
   children?: React.ReactNode;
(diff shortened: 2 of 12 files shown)


.changeset/chilled-impalas-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/polaris': minor
```
mrcthms: This should be a `minor` release, so these will need deleting and running through the `yarn changeset` command again. Also, we only need one changeset file per PR. Also, these go out in the official release notes, so if you can be a descriptive as possible and use sentence case, that would be fab, thanks!

polaris-react/src/components/Tooltip/components/TooltipOverlay/TooltipOverlay.tsx
```diff
@@ -7,37 +7,59 @@ import {
...
-  preventInteraction,
   id,
   children,
   accessibilityLabel,
+  transform,
 }: TooltipOverlayProps) {
   const i18n = useI18n();
+  const [tooltipTransform, setTooltipTransform] = useState<string>('');
+  const tooltipRef = useRef<HTMLDivElement>(null);
+  const windowSize = useRef({
```
mrcthms: I think this is going to cause an error on any server-rendered pages, as the `window` object will not exist. Plus we need to account for when a user resizes their browser. It'd be safer to do something like:

```tsx
interface WindowSize {
  width: number;
  height: number; 
}
const windowSize = useRef<WindowSize>(null);
useEffect(() => {
  function handleWindowResize() {
    windowSize.current = {
      width: window.innerWidth,
      height: window.innerHeight
    }
  }
  handleWindowResize();
  window.addEventListener('resize', handleWindowResize);
  () => window.removeEventListener('resize', handleWindowResize);
}, [])
```

polaris-react/src/components/Tooltip/components/TooltipOverlay/TooltipOverlay.tsx
```diff
@@ -7,37 +7,59 @@ import {
...
+  });
+
+  useEffect(() => {
+    const tooltipRect = getRectForNode(tooltipRef.current);
+    const pushSize: number = tooltipRect.width / 2 + 10;
+
+    const {x, y, cursorX} = transform;
+    const hasHitRightBorder =
+      cursorX + tooltipRect.width + 20 > windowSize.current.width;
+    const transformPositionX = hasHitRightBorder ? x - pushSize : x + pushSize;
```
mrcthms, line 64: Very nice! Remember we'll need to do this for the Y axis, too.

polaris-react/src/components/Tooltip/components/TooltipOverlay/TooltipOverlay.tsx
```diff
@@ -7,37 +7,59 @@ import {
...
 export interface TooltipOverlayProps {
   id: string;
   active: boolean;
-  preventInteraction?: PositionedOverlayProps['preventInteraction'];
   preferredPosition?: PositionedOverlayProps['preferredPosition'];
   children?: React.ReactNode;
   activator: HTMLElement;
   accessibilityLabel?: string;
   onClose(): void;
+  transform: TooltipPosition;
```
mrcthms: We may want to rename this prop here. Within the `PositionedOverlay` we have a `transform` prop which is a `string`, whereas here it is an object of values. This could be confusing for folks. Maybe something like `coordinates` would work better here given the type of values we're passing.

polaris-react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -126,12 +141,33 @@ export function Tooltip({
...
     !mouseEntered.current && handleMouseEnter();
   }
+
+  function handleMouseMove(event: React.MouseEvent) {
+    if (!active) return;
+    if (childWrapperContainer.current == null) return;
+    const {x, y, width, height} =
+      childWrapperContainer.current.getBoundingClientRect();
+    const centerX = width / 2;
+    const ceterY = height / 2;
```
mrcthms: Nit: spelling

mrcthms (changes requested):
Really good stuff going on here! Just a few final little tweaks and polishes and we'll be good to go ☄️

polaris-react/src/components/Tooltip/components/TooltipOverlay/TooltipOverlay.tsx
zakwarsame: > replying what we discussed earlier, on github for any readers

the y-axis scroll is already handled by the `calculateVerticalPosition` function in `PositionedOverlay.tsx`

zakwarsame:
Thanks for the feedback! Should be good to go!

polaris-react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -34,22 +38,28 @@ export interface TooltipProps {
...
 }: TooltipProps) {
   const WrapperComponent: any = activatorWrapper;
+  const ChildWrapperComponent: any = activatorWrapper;
   const {
     value: active,
     setTrue: handleFocus,
     setFalse: handleBlur,
   } = useToggle(Boolean(originalActive));
   const [activatorNode, setActivatorNode] = useState<HTMLElement | null>(null);
+  const [positionTooltip, setPositionTooltip] = useState<TooltipPosition>({
```
mrcthms: Nit: rename this to `const [coordinates, setCoordinates]` to mirror the prop you're passing it down to.

polaris-react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -132,6 +147,28 @@ export function Tooltip({
   function handleMouseEnterFix() {
     !mouseEntered.current && handleMouseEnter();
   }
+
+  function handleMouseMove(event: React.MouseEvent) {
+    if (!active) return;
+    if (childWrapperContainer.current == null) return;
+    const {x, y, width, height} =
+      childWrapperContainer.current.getBoundingClientRect();
```
mrcthms: Might want to use the `getRectForNode` utility you're importing in TooltipOverlay here too, for consistency's sake.

mrcthms (commented):
The actual code is looking pretty good to me now. Just a couple of little nitpicks left.

What would be good to include would be some unit tests for some of this new functionality. Specifically:

- Testing that the `transform` prop gets passed down correctly to the` PositionedOverlay`
- Testing the `handleMouseMove` function within `Tooltip` and ensuring it gives you the expected results. (if you need help mocking things let me know)
- Testing the `useEffect` hooks introduced in `TooltipOverlay`, ensuring that the setTooltipTransform is being called with the correct value

johanstromqvist:
Demo looks great!

A few questions how this impacts UI and devs using the component:
- Can you share a little why are these changes being made from a merchant or developer POV? We need this to grasp its purpose and integrate it purposefully.
- How should it be used? Is this out-of-the box default behaviour that apply to all tooltip use cases? Or do designers and developers need to make any sort of decisions on when to use it? Does this decision making need guidance?
- Are there any performance considerations?

dGoligorsky:
👋 Reviving the thread with rationale—
🎥 [Video explainer, 4min](https://videobin.shopify.io/v/BDl1WW).

## Today

- Tooltips on charts follow the mouse, which definitely adds value to the data points underneath.
- Tooltips on text appear on hover. Clicking the text will make the tooltip persist, until clicking elsewhere to dismiss the tooltip.
- There’s a tooltip behavior on truncated text in an index row, which only work on hover.
- gif examples of [analytics card in product](https://screenshot.click/02-54-363zp-z7q1c.gif), and [marketing page in production](https://screenshot.click/02-18-tc4ay-ovge8.gif) 

## Use cases that could arise
- If it’d be useful to have a tooltip for cells in index rows.
-- e.g. for Product Status, one could hover and see active channels, instead of clicking today. [gif where it could help](https://screenshot.click/02-25-zbhk1-d0wnf.gif)
-- e.g. if the clickthrough rate were a sparkline graph and one could hover it for data. [gif where it could help](https://screenshot.click/02-18-tc4ay-ovge8.gif)

## Proposed solution for text tooltips
- Use the ["traveling tooltip" on hover](https://screenshot.click/02-57-595ug-s3n4w.gif). Fix the tooltip in place on click.

yesenia-perezcruz:
Hi all, 

To give some more context on the purpose of these traveling tool tips:

The goal is to give merchants optional but helpful context on hover in Index views. Today in Indexes, merchants need to click to see a preview of customer or products.
image (5)

The traveling tool tip proposal would allow merchants to see that information without needing to click, similar to Github's tool tips where you can see the details for a pull request or user at a glance.
Screen Shot 2022-09-02 at 12 38 12 PM

The opportunities here are that Shopify's core objects (customer, product, collection, etc) can be more easily recognizable from Indexes and the admin will feel more like a pro-tool by providing helpful context without requiring extra clicks. 

I'm on board with this change. 

@dGoligorsky is your proposal that this is default functionality to tooltips or that it's for specific use cases?

dGoligorsky:
@yesenia-perezcruz 
Based on the rationale that hovering text, help icons, and index cells is an exploration mode, I'd propose that the traveling tooltip should be the default prop for hover states. For focus and tap/click, the tooltip should become fixed.

We may want to consider introducing a nominal delay (125ms?) to trigger a tooltip on hover. Just so there's a little intentionality from the user before we have a flurry of tooltips while hovering an index table, for example.

PR #6461, opened 2022-06-30
[Button] Make disabled buttons more accessible

### LINKED ISSUE #5967: Make our disabled buttons more accessible
Currently, our buttons in Polaris have a `disabled` prop. When `true`, this sets the `disabled` property on the underlying button, which prevents a user from interacting with the button. For merchants that use a mouse to interact with the admin, this is fine, but breaks the experience for our merchants that use other means of interacting with the admin, especially keyboard users.

A better approach is to utilise the `aria-disabled` attribute, in combination with JavaScript to disable the functionality. Using [this article as a reference](https://css-tricks.com/making-disabled-buttons-more-inclusive/), we should update the Button component in Polaris to utilise this best practice.

zakwarsame:

- Buttons are currently insufficiently accessible when the `disabled` prop is set to true
- The native disabled HTML attribute causes this by ignoring the entire element when navigating with the `Tab` or `Tab + Shift` keys and also makes it difficult for **screen readers** to locate it.
- Using the `aria-disabled` instead makes it easy for screen readers and keyboard users to interpret the button.

Fixes https://github.com/Shopify/polaris/issues/5967

- Replaces `disabled` with `aria-disabled` 

### How to 🎩

- Using storybook, go to Button > Disabled state, and navigate using the `Tab` key

🖥  [Check out this short demo video](https://screenshot.click/30-28-32121-86914.webm)

### 🎩 checklist


.changeset/hot-crabs-act.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/polaris': major
```
mrcthms: This should be a minor release, so this is going to have to be deleted and re-generated using `yarn changeset`. If you just press `Enter` on the major screen, it'll move over to the minor release option and then you can use `space` to select @shopify/polaris.

polaris-react/src/components/Button/Button.tsx
```diff
@@ -200,6 +201,16 @@ export function Button({
     setDisclosureActive((disclosureActive) => !disclosureActive);
   }, []);
 
+  const preventedInteraction = eventsToPrevent.reduce(
```
mrcthms: Firstly, I want to say this is a very clever bit of code here, `reduce` can be a fairly tricky method to understand but you've done so here. However, it takes a little bit of time to understand fully what is going on here, we should try and make our code that we write as clear as possible for other folks that may need to edit it.

Furthermore, if `disabled` is true, and we spread the `preventedInteraction` variable on the `button` like in line 247 below, we are actually attaching two `onClick` handlers to the `button`. It's a better practice to include this kind of logic within the existing event handler, rather than create an alternative. So in this example, it'd look something like:

```tsx
const toggleDisclosureActive = useCallback((event: React.MouseEvent<HTMLButtonElement>) => {
  if (disabled) {
    event.preventDefault();
    return;
  }
    setDisclosureActive((disclosureActive) => !disclosureActive);
}, [disabled]);

const handleKeyPress = useCallback((event: React.KeyboardEvent<HTMLButtonElement>) => {
  if (disabled) {
    event.preventDefault();
  }
}, [disabled])
```

and where we have the button below it'd look like:

```tsx
<button
  type="button"
  className={connectedDisclosureClassName}
  aria-disabled={disabled}
  aria-label={disclosureLabel}
  aria-describedby={ariaDescribedBy}
  aria-checked={ariaChecked}
  onClick={toggleDisclosureActive}
  onKeyPress={handleKeyPress}
>

File: polaris-react/src/components/Button/Button.tsx [hunk 3]
```diff
@@ -288,7 +300,12 @@ export function Button({
   };
 
   const buttonMarkup = (
-    <UnstyledButton {...commonProps} {...linkProps} {...actionProps}>
+    <UnstyledButton
+      {...commonProps}
+      {...linkProps}
+      {...actionProps}
+      preventedInteraction={preventedInteraction}
```
mrcthms: Similar to above, we should be handling the nullifying of the event within the existing handlers within the UnstyledButton component, so we don't need to pass this prop through.

polaris-react/src/components/UnstyledButton/UnstyledButton.tsx
```diff
@@ -89,6 +90,7 @@ export function UnstyledButton({
         onKeyUp={onKeyUp}
         onKeyPress={onKeyPress}
         {...rest}
+        {...(disabled ? preventedInteraction : null)}
```
mrcthms: Similar to the Button component, we should be doing the early return within a wrapping function here, something like:

```tsx
function handleClick(event: React.MouseEvent<HTMLButtonElement>) {
  if (disabled) {
    event.preventDefault();
    return;
  }
  onClick && onClick(event)
}

function handleKeyPress(event: React.KeyboardEvent<HTMLButtonElement>) {
  if (disabled) {
    event.preventDefault();
    return;
  }
  onKeyPress && onKeyPress(event)
}
```

and then update the methods passed to the <button> element to be `onKeyPress={handleKeyPress} onClick={handleClick}`. you'll need to explicitly destructure the `onClick` variable from the `props` object, too.

polaris-react/src/components/UnstyledButton/tests/UnstyledButton.test.tsx
```diff
@@ -212,15 +212,15 @@ describe('<Button />', () => {
     it('passes to `button`', () => {
       const button = mountWithApp(<UnstyledButton disabled />);
       expect(button).toContainReactComponent('button', {
-        disabled: true,
+        'aria-disabled': true,
```
mrcthms, line 215: Would be good to add a couple of tests here to ensure that the `onClick` and `onKeyPress` props are not getting called if `disabled`. Should be fairly similar to the test you wrote for the `Button.test.tsx` file.

mrcthms (changes requested):
Overall, your logic is pretty sound here. Just a couple of semantic things to take a look at.

Also, you mentioned the disabled buttons still having a focus state when a user presses a key. Ideally we would want to avoid having the button change visually at all when it is disabled, event when focused. It's an industry best practice to do so, and also could cause potential confusion for users if it is the only button on screen and they feel they are interacting with it, but don't have an active button to compare it to note this is a disabled button, rather than us just having super low contrast regular buttons, if that makes sense.

zakwarsame:
thanks for the premium feedback marc, should all be resolved :)

polaris-react/src/components/Button/Button.scss
```diff
@@ -224,7 +224,7 @@
   }
 
   &:hover,
-  &:focus {
+  &:focus:not(.disabled) {
```
mrcthms, line 227: Nice :+1:

mrcthms (approved):
This looks good to me! We'll probably need to get somebody from @Shopify/polaris-team to take a 👀  too though before we can merge it.

polaris-react/src/components/Button/Button.tsx
```diff
@@ -200,6 +200,25 @@ export function Button({
...
+    [disabled, toggleDisclosureActive],
+  );
+  const handleKeyDown = useCallback(
+    (event: React.KeyboardEvent<HTMLButtonElement>) => {
+      if (['Enter', ' '].includes(event.key) && disabled) {
+        event.preventDefault();
+      }
+    },
+    [disabled],
+  );
```
alex-page: I don't know off the top of my head, but I assumed using `aria-disabled` would have given us this functionality for free. This logic could probably be repeated for all Polaris elements that are disabled.

sarahill:
> I don't know off the top of my head, but I assumed using `aria-disabled` would have given us this functionality for free. This logic could probably be repeated for all Polaris elements that are disabled.

+1 `aria-disabled` should give us what we want here and communicates with folks using screen readers https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled

zakwarsame:
> I don't know off the top of my head, but I assumed using aria-disabled would have given us this functionality for free. This logic could probably be repeated for all Polaris elements that are disabled.

> +1 aria-disabled should give us what we want here and communicates with folks using screen readers https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled

Yeah we could benefit from the fact that it communicates with assistive technologies. Apart from the semantics, though, we don't get features like click prevention, focus prevention:

> While adding disabled to an HTML form control causes :disabled user-agent styles to be applied, adding aria-disabled="true" doesn't, by default, alter an element's appearance. The element can be styled with the [attribute selector](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors) `[aria-disabled="true"] ...
> You should include JavaScript to disable the functionality of the element while also changing the appearance of the element so sighted users know it is disabled.  - MDN

A nice table from [this article](https://css-tricks.com/making-disabled-buttons-more-inclusive/) reference:


I'm currently exploring exporting the logic to a custom hook in order to be repeated in other Polaris elements.

alex-page:
Hey @zakwarsame I don't think we should ship changes based off the css tricks article. Have we found issues using assistive technologies? Can we replicate the issue you are fixing?

zakwarsame:
> Hey @zakwarsame I don't think we should ship changes based off the css tricks article. Have we found issues using assistive technologies? Can we replicate the issue you are fixing?

The source I quoted there is actually from MDN- I only used the table from css tricks to represent what's written in MDN visually. I can understand that we need to confirm it, so I've replicated this myself. Here is how:

**Issue**
- What we're trying to fix can be easily replicated on our current storybook by clicking through the disabled buttons with both mouse and keyboard, and observing that nothing happens.  Below I'm demonstrating the problem. Neither the keyboard, nor the assistive voice-over picks up that there is a disabled button:

   <details>
    <summary>Current disabled button walk-through video (unmute to hear 🔊) : </summary>

   https://user-images.githubusercontent.com/42528878/178357465-e1b75385-e75b-462b-bc68-d6052af48a08.mov
   </details>

**Solution**

- `aria-disabled` tackles this by enabling keyboard navigation and indicates to assistive technologies that the element upon which it is set and all focusable descendants are disabled. 


   <details>
     <summary>   Walk-through showing that with aria-disabled, we can both navigate and get feedback from an assistive technology for indication (unmute to hear 🔊) </summary>

   https://user-images.githubusercontent.com/42528878/178355307-26fbec67-e2d6-45a5-bea9-4a0343947f23.mov

   </details>

**Why we need to additionally disable clicks with javascript on `aria-disable`d elements**

- Even though we have `aria-disabled` and `pointer-events: none;` in our styling for these elements, keyboards are still able to click the buttons.

   Below, notice how nothing happens when I click with a mouse (due to our css code `pointer-events: none;`) but also nothing prevents the keyboard from clicking:

   <details>
     <summary>Here I'm showing what happens when we click the button with `aria-disabled` and our current 
   styling</summary>

   https://user-images.githubusercontent.com/42528878/178361123-3c383764-bf85-4412-a582-f9f179fd8c75.mov

   </details>

   _I've temporarily included the code `console.log('This should not be clickable')` in the handleClick function for emphasis on problem. It's not in the PR._

**Final proposal**

- With this PR, we get the best of `aria-disabled` and Javascript by stopping the button from being clicked with both mouse and keyboard while maintaining the accessibility functionality.

   <details>
     <summary>Assistive technology picks it up (unmute to hear 🔊): </summary>

    https://user-images.githubusercontent.com/42528878/178362691-22004c80-e4aa-4a92-96aa-44f2efebc58f.mov

   </details>

   <details>
     <summary>Keyboard cannot click button but can navigate</summary>

   https://user-images.githubusercontent.com/42528878/178363125-1f3a44b2-d8a5-4dc2-9138-c1ed7a31d5e7.mov

   </details>

alex-page:
@zakwarsame this is an excellent write up. Thank you. Honestly this is super.

It's interesting to me that disabled elements can receive focus. What do you think about changing the `tabindex`? This could stop disabled elements from receiving focus and then if they aren't focused they cannot be clicked on. Is there a reason they need to be focusable?

zakwarsame:
Thank you! I think having them be focusable but styling to indicate that they're not clickable is the perfect compromise.

> Keyboard accessibility is one of the most important aspects of web accessibility. Many users with motor disabilities rely on a keyboard. Some people have tremors which don't allow for fine muscle control. Others have little or no use of their hands, or no hands at all. In addition to traditional keyboards, some users may use modified keyboards or other hardware that mimics the functionality of a keyboard. Blind users also typically use a keyboard for navigation. Users without disabilities may use a keyboard for navigation because of preference or efficiency
-[WebAIM](https://webaim.org/techniques/keyboard/)

Considering that most people with visual impairments will prefer to navigate using the keyboard, wouldn't `tabIndex` do the opposite of what we're trying to achieve here?
> A negative value (usually tabindex="-1") means that the element is not reachable via sequential keyboard navigation...
The user won't be able to focus any element with a negative tabindex using the keyboard - [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex)

I'm going to explore this further before merging, for best results.

zakwarsame:
`tabIndex` is picked up by screen readers. This is the best solution for avoiding keyboard focus and allowing assistive technologies to navigate.

<details>
<summary>No focus when only navigating with keyboards: </summary>

https://user-images.githubusercontent.com/42528878/178508332-50ff99f9-608b-4195-8f0b-d6dfa0e792aa.mov
</details>

<details>
<summary>With a screen reader (🔊): </summary>


https://user-images.githubusercontent.com/42528878/178509126-5a2cf9a3-efee-47bb-946a-dc5528e77e54.mov
</details>

polaris-react/src/components/Button/Button.tsx
```diff
@@ -200,6 +204,9 @@ export function Button({
     setDisclosureActive((disclosureActive) => !disclosureActive);
   }, []);
 
+  const handleClick = useDisableClick(disabled, toggleDisclosureActive);
+  const handleKeyDown = useDisableKeyboard(disabled);
```
alex-page: I think we can remove the `handleKeyDown` now that the elements cannot receive focus.

zakwarsame: Yes, I agree.
You mentioned earlier that this logic could potentially be repeated for other Polaris elements that are disabled - do you think `aria-disabled` +  `tabIndex` could also be optimal for these? In that case, we could do without this new `useDisableInteraction` hook entirely.

polaris-react/src/utilities/use-disable-interaction.ts
```diff
@@ -0,0 +1,59 @@
...
+ *  console.log('disable me');
+ * };
+ * const handleClickEvent = useDisableInteraction(true, handleClick);
+ * return <button onClick={handleClickEvent}>Im Disabled</button>;
+ * }
+ */
+
+export function useDisableClick(
+  disabled = false as boolean,
+  handleClick: () => void = () => {},
```
mrcthms: Little nit here, and with the `useDisableKeyboard` expression too. I don't think we need to set the default parameters here given we are making them required arguments

polaris-react/src/utilities/use-disable-interaction.ts
```diff
@@ -0,0 +1,59 @@
...
+    [disabled, handleClick],
+  );
+
+  return handleClickWrapper;
+}
+
+export function useDisableKeyboard(
+  disabled = false as boolean,
+  handleKeyDown: (
+    event?: React.KeyboardEvent<HTMLButtonElement>,
```
mrcthms: Any reason why this is optional here? One would imagine the event parameter will always exist if it's being invoked from an event handler.

polaris-react/src/components/UnstyledButton/UnstyledButton.tsx
```diff
@@ -37,6 +41,7 @@ export function UnstyledButton({
   onKeyUp,
   onMouseEnter,
   onTouchStart,
+  preventedInteraction,
```
mrcthms: Think this can be removed.

mrcthms (approved):
Couple of little nits but nothing from my side that would block a release, nice one!

polaris-react/src/utilities/use-disable-interaction.ts
zakwarsame: The prop type passed in the component has both `onKeyDown` and `onClick`  as optional so initially I went with the default params but I think passing them as optional again makes sense. Otherwise, they don't pass the type check.

polaris-react/src/utilities/use-disable-interaction.ts
zakwarsame: https://github.com/Shopify/polaris/pull/6461#discussion_r925825501

arthurgouveia:
@zakwarsame I'm currently having to run Polaris Uplift on Email, and we need to overwrite some properties to make sure it matches our UI.

Normally, I target `:disabled` for that and got surprised we don't add it do the button. While I can see `aria-disabled` addition being a good thing, why would we remove the native `disabled` as well? Would putting it back cause any harm to the current implementation?

From an a11y POV, I don't see how `disabled` and `aria-disabled` wouldn't announce the same on a screenreader given both should serve the same purpose, while the first actually also natively makes the button not work.

Surely I can target the `aria-disabled` but I'm curious as to why we would remove the native browser solution as well.

Issue #6471, opened 2022-07-01
[polaris.shopify.com] Icon button names should not be subdued


chloerice:

martenbjork:
From a design POV, we want the icons to stand out. The text shouldn't distract. The subdued text passes a11y requirements.

Do you still feel that this need to be changed, @chloerice ?

<img width="228" alt="image" src="https://user-images.githubusercontent.com/875708/180439767-2c4659b2-ca93-4f18-837f-4f29a05e7010.png">

chloerice:
Hey Marten! Sorry meant to close this when the subdued color was fixed 💯

martenbjork:
NP! 😋 Just trimming the backlog ☺️

Issue #6470, opened 2022-07-01
`--surface-subdued` is indistinguishable from `--surface` in light mode


chloerice:

chloerice:
No longer an issue.

PR #6473, opened 2022-07-01
[polaris.shopify.com] Various a11y improvements

### LINKED ISSUE #6469: The video in the Polaris for VS Code section of `HomePage` should account for users who prefer reduced motion

chloerice:

Closes #6469, #6470, and #6471

- General:
  - Gave light mode `--surface-subdued` analgous weight to dark mode
- Icons: 
  - Made the icon name button text responsive and accessible
- Home Page:
  - Render an image in place of video if reduce motion preference is set in the Polaris for VS Code section

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist


polaris.shopify.com/src/components/HomePage/HomePage.module.scss
```diff
@@ -280,8 +283,29 @@
...
+  .Poster {
+    position: absolute;
+    border-radius: 24px;
+    overflow: hidden;
+    width: 60%;
+    height: 85%;
+    left: 50%;
+    margin-right: -10rem;
+    min-width: 65.5;
+    max-width: 65.5%;
```
gui-santos: The `min-width` here is missing the `%` sign

And I have another question, I see that both min- and max-width have the same value 65.5% which is bigger than the `width` value. So I think that in the end they are not doing anything
Isn't it possible to remove them and keep only `width` in this case?

And I think the margin-right is also not doing anything. At least, it seems not to do anything when I checked in Chrome's inspector

polaris.shopify.com/src/components/HomePage/HomePage.tsx
```diff
@@ -2,17 +2,22 @@ import { BrowseIcon, FigmaIcon, GitHubIcon, InstallIcon } from "./icons";
...
 import Image from "../Image";
 import PageMeta from "../PageMeta";
 import styles from "./HomePage.module.scss";
+import { useMediaQuery } from "@shopify/polaris/build/ts/latest/src/utilities/media-query";
 
 interface Props {}
 
 function HomePage({}: Props) {
+  const useMotion = useMedia("(prefers-reduced-motion: no-preference)");
+  console.log(useMotion);
```
gui-santos: Is this a leftover?

I also forget to remove console.logs in this project and I think it's because the eslint has no rules in `polaris.shopify.com`

polaris.shopify.com/src/components/HomePage/HomePage.tsx
```diff
@@ -2,17 +2,22 @@ import { BrowseIcon, FigmaIcon, GitHubIcon, InstallIcon } from "./icons";
...
+import { useMedia } from "../../utils/hooks";
 import componentsImage from "./components.png";
 import tokensImage from "./tokens.png";
 import iconsImage from "./icons.png";
 import foundationsImage from "./foundations.png";
+import vsCodeImage from "./vscode.png";
 import Image from "../Image";
 import PageMeta from "../PageMeta";
 import styles from "./HomePage.module.scss";
+import { useMediaQuery } from "@shopify/polaris/build/ts/latest/src/utilities/media-query";
```
gui-santos: I think this is not being used anywhere, is it?

polaris.shopify.com/src/components/HomePage/HomePage.module.scss
```diff
@@ -280,8 +283,29 @@
...
+    top: 50%;
+    transform: translateY(-50%) translateX(33px);
+    box-shadow: 0 50px 50px rgba(0, 0, 0, 0.15);
+
+    .Poster_Image {
+      // width: 100%;
+      // height: auto;
+      // object-fit: cover !important;
+      // position: relative;
```
gui-santos: Should this be removed, maybe?

gui-santos (changes requested):
This is great! 🙌 

I had only a few questions and suggestions 🙏

polaris.shopify.com/src/pages/api/v0/tokens/index.tsx
```diff
@@ -14,7 +14,7 @@ const getGithubUrl = (file: string, isRaw: boolean) => {
 
 const html = `
 <!DOCTYPE>
-<html>
+<html lang="en">
```
martenbjork: This HTML is actually for the scrappy landing page for the API.

To add the lang attriibute for all pages, you can add some shenanigans to next's config file, see #6475 .

polaris.shopify.com/src/styles/globals.scss
```diff
@@ -150,7 +150,8 @@ body,
   --text-subdued: #777;
   --text-link: #2e5dba;
   --surface: #fff;
-  --surface-subdued: #fbfbfb;
+  --surface-subdued: #f7f7f7;
+  --surface-inline-code: #e7e7e7;
```
martenbjork: Is this used somewhere? 🧐

polaris.shopify.com/src/components/IconGrid/IconGrid.module.scss
```diff
@@ -23,6 +23,11 @@
     border-radius: var(--border-radius-600);
     box-shadow: var(--card-shadow);
     user-select: none;
+    color: var(--text);
```
martenbjork: Thanks for removing that hard coded text color! A leftover from prototyping...

I wonder if we could use `text-subdued` for these, if we upped the contrast of text-subdued to `#767676`?

image

martenbjork (changes requested):
Thanks for fixing these issues! 🔥

polaris.shopify.com/src/styles/globals.scss
chloerice: Ah yes! This PR actually started off as part of [this other PR](https://github.com/Shopify/polaris/pull/6430/files#r913845615) and I broke it out because it kind of spiraled as I started doing a11y testing 😅 Moved `--surface-inline-code` to #6430

polaris.shopify.com/src/components/HomePage/HomePage.module.scss
chloerice: 🙏🏽

gui-santos (approved):
Looks good! 💎

polaris.shopify.com/src/components/HomePage/HomePage.tsx
```diff
@@ -208,18 +212,30 @@ function HomePage({}: Props) {
...
+                    <source src="/images/vscode.mp4" type="video/mp4" />
+                  </video>
+                </div>
+              ) : (
+                <div className={styles.Poster}>
+                  <Image
+                    width="2250"
+                    height="1440"
+                    src={vsCodeImage}
+                    className={styles.Poster_Image}
```
martenbjork: Nit: For classes, we use `ThisCasingWithoutUnderScores` in the rest of the project.

polaris.shopify.com/src/components/HomePage/HomePage.tsx
```diff
@@ -2,17 +2,21 @@ import { BrowseIcon, FigmaIcon, GitHubIcon, InstallIcon } from "./icons";
 import Container from "../Container";
 import Link from "next/link";
 import { className } from "../../utils/various";
+import { useMedia } from "../../utils/hooks";
 import componentsImage from "./components.png";
 import tokensImage from "./tokens.png";
 import iconsImage from "./icons.png";
 import foundationsImage from "./foundations.png";
+import vsCodeImage from "./vscode.png";
```
martenbjork: This single asset weighs in at 3.4mb. Simply exporting it as a JPEG reduces the size by 90%.

Next/Image will serve a compressed version, so users won't suffer, but converting it might be a tiny win towards keeping the repo lightweight...er.

martenbjork (approved):
Fantastic 🔥🔥

Issue #6329, opened 2022-06-22
Rework Layout to use Grid component


lgriffee:
Related to #6327

```scss
$secondary-basis: $layout-width-secondary-min;
$primary-basis: $layout-width-primary-min;
$one-half-basis: $layout-width-one-half-width-base;
$one-third-basis: $layout-width-one-third-width-base;
$relative-size: $primary-basis / $secondary-basis;
```

PR #6483, opened 2022-07-04
Deprecate Layout.AnnotatedSection

kyledurand:

Part of https://github.com/Shopify/polaris/issues/6329

### How to 🎩

### 🎩 checklist

### DIFF
--- a/.changeset/shy-needles-battle.md
+++ b/.changeset/shy-needles-battle.md
@@ -1,0 +1,5 @@
+---
+'@shopify/polaris': patch
+---
+
+Deprecated Layout.AnnotatedSection
--- a/polaris-react/src/components/Layout/Layout.tsx
+++ b/polaris-react/src/components/Layout/Layout.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 
+// eslint-disable-next-line import/no-deprecated
 import {AnnotatedSection, Section} from './components';
 import styles from './Layout.scss';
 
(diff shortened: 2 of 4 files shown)


alex-page (approved):

mbaumbach:
Out of curiosity, is this component deprecated because the intent is that Polaris will no longer suggest using this annotated pattern in the future? If so, is there a new pattern suggested? Thanks in advance for any clarification!

stephenkeable:
@mbaumbach I was wondering that too, found this in the commit:-
https://github.com/Shopify/polaris/commit/3b65101e88519cbc992ec417e25ea5c227dd4975#diff-f42a28126a71d9eb46c958954ef437461a4d1f184e66a598f9a4a5f97c367e35

https://github.com/Shopify/polaris/blob/3b65101e88519cbc992ec417e25ea5c227dd4975/polaris-react/src/components/Layout/README.md#annotated-layout-using-plain-layoutsections

However I've found that causes the "one third" section text to hit the edges of the display on mobile/narrow screens
Screenshot 2022-08-09 at 14 06 10
.

mbaumbach:
@stephenkeable Yea, I saw that commit as well. I would assume if this is a pattern that is suggested to continue doing that having a specialized component like the Layout.AnnotatedSection still worthwhile. That way if minor design tweaks happen, the implementation of the Annotated Section is just a version bump for consumers and we'll all share the same design.

@kyledurand Any thoughts on the future of annotation sections? Is it something that Polaris is effectively deprecating from a design suggestion standpoint which is why this component was removed? I just want to make sure we are building similar experiences in our apps that Shopify plans to use in the core UI. Thanks!

PR #6484, opened 2022-07-04
Change icon name casing

martenbjork:
Changes the casing of icons to start with an uppercase character. This will make the icon names render like any other word/sentence on the website.

image

The keys in the metadata file/object are untouched, so this shouldn't be a breaking change.

### DIFF
--- a/.changeset/fresh-drinks-change.md
+++ b/.changeset/fresh-drinks-change.md
@@ -1,0 +1,5 @@
+---
+'@shopify/polaris-icons': minor
+---
+
+Change icon names to start with an uppercase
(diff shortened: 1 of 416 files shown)


BPScott:
Drive-by thought: Thoughts on using Title Case to account for places where [icon names are multiple words](https://github.com/Shopify/polaris/pull/6484/files#diff-5d74a7b4b9fd949d4bb692483d11e40596b729e8e9d2d11f40ae7508bee61e36R1)?

alex-page (approved):

martenbjork:
> Drive-by thought: Thoughts on using Title Case to account for places where [icon names are multiple words](https://github.com/Shopify/polaris/pull/6484/files#diff-5d74a7b4b9fd949d4bb692483d11e40596b729e8e9d2d11f40ae7508bee61e36R1)?

If I read our guidelines correctly, they say to use sentence case.

https://polaris.shopify.com/foundations/content/naming

image

However, some of the words should perhaps be uppercased, like names of currencies 🤔

PR #6488, opened 2022-07-04
[polaris.shopify.com] Show component status (alpha, deprecated)

martenbjork:
There is room to give the design of these a bit more love down the road, but I thought we'd keep it simple for v1.

Screen Shot 2022-07-04 at 9 05 08 PM

Screen Shot 2022-07-04 at 9 04 59 PM

Screen Shot 2022-07-04 at 9 04 51 PM

Screen Shot 2022-07-04 at 9 05 48 PM


polaris-react/src/components/Grid/README.md
```diff
@@ -13,7 +13,7 @@ keywords:
   - full width containers
   - css grid
 notice:
-  status: alpha
+  status: Alpha
```
martenbjork: I changed these to uppercase. Could this be considered a breaking change? @kyledurand

polaris.shopify.com/src/components/ComponentGrid/ComponentGrid.module.scss
```diff
@@ -50,6 +53,22 @@
...
+    padding: 0 0.5rem;
+
+    &[data-status="Deprecated"] {
+      background: var(--surface-warning);
+    }
+
+    &[data-status="Alpha"] {
+      background: var(--surface-information);
+    }
+  }
```
alex-page: I would love this to be another component

alex-page: Oh. I see, maybe this can be removed?

polaris-react/src/components/Grid/README.md
kyledurand: It's just for the styleguide so I think you're ok. I guess there's a small chance someone somewhere could be pulling this meta data for some reason but we shouldn't have to worry about stuff like that I don't think

polaris.shopify.com/src/components/ComponentGrid/ComponentGrid.module.scss
kyledurand: Seems like it can be removed

kyledurand (approved):
This is looking good to me

polaris.shopify.com/src/components/ComponentGrid/ComponentGrid.module.scss
martenbjork: Just to clarify, there are 2 components that need these background styles: the `Badge` and the `Banner`.

I've componentized these now to make all of this a bit cleaner :)

martenbjork:
Updated PR since last review:

`notice` frontmatter meta is now called `status`. Website code is updated accordingly. This makes the status API pretty clean: `<StatusBadge status={status} />`

alex-page (approved):

PR #6489, opened 2022-07-04
[polaris.shopify.com] Add support for tooltips that point up

### LINKED ISSUE #6478: [polaris.shopify.com] Fix tooltip direction
Tooltips have the capability to point to any direction, but we currently hard-code the arrow to point down which sometimes causes them to point at the wrong thing.

image

Steps to reproduce:

- Go to global search. Search for an icon.
- Scroll so that the icon is at the top of the list.
- Hover the icon. The tooltip will render below, but the arrow will point down instead of up.

martenbjork:
Tooltips default to pointing down:

image

But can point up when needed:

image

Closes #6478

### DIFF
--- a/.changeset/stupid-cherries-press.md
+++ b/.changeset/stupid-cherries-press.md
@@ -1,0 +1,5 @@
+---
+'polaris.shopify.com': patch
+---
+
+Fix the direction of Tooltip's arrow. Now supports pointing upwards as well.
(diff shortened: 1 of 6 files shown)


alex-page (approved):
We probably shouldn't own component logic this complex on this website. We use a bunch of libraries already, do they have anything? I also wonder how much value the tooltips bring versus maintaining location logic.

martenbjork:
> We probably shouldn't own component logic this complex on this website. We use a bunch of libraries already, do they have anything? I also wonder how much value the tooltips bring versus maintaining location logic.

I agree, that's why we use an external library for this :) This component is just a small internal wrapper so that website devs can ignore the external library and just use `<Tooltip />`. Inside that component, all the heavy lifting is done by `@floating-ui/react-dom-interactions`.

PR #6493, opened 2022-07-05
[polaris.shopify.com] Search clean up

alex-page:
- [x] Remove modal backdrop with no functionality
- [ ] Remove `<>` that isn't needed
- [x] Simplify `search.ts` functionality

### DIFF
--- a/polaris.shopify.com/src/components/GlobalSearch/GlobalSearch.tsx
+++ b/polaris.shopify.com/src/components/GlobalSearch/GlobalSearch.tsx
@@ -173,170 +173,174 @@ function GlobalSearch({}: Props) {
 
       <Dialog open={isOpen} onClose={() => setIsOpen(false)}>
         <div className="dark-mode">
-          <Dialog.Panel className={styles.Results}>
-            {isOpen && (
-              <div className={styles.Header}>
-                <div className={styles.SearchIcon}>
-                  <SearchIcon />
-                </div>
+          <>
+            <Dialog.Panel className={styles.Results}>
+              {isOpen && (
+                <div className={styles.Header}>
+                  <div className={styles.SearchIcon}>
+                    <SearchIcon />
+                  </div>
 
-                <input
-                  type="search"
-                  value={searchTerm}
-                  onChange={(evt) => setSearchTerm(evt.target.value)}
-                  role="combobox"
-                  aria-controls="search-results"
-                  aria-expanded={searchResultsCount > 0}
-                  aria-activedescendant={
-                    searchResultsCount > 0
-                      ? getItemId(activeDescendant)
-                      : undefined
-                  }
-                  onKeyDown={handleKeyboardNavigation}
-                  autoComplete="off"
-                  autoCorrect="off"
-                  autoCapitalize="off"
-                  spellCheck={false}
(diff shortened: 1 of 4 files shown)


polaris.shopify.com/src/utils/search.ts
```diff
@@ -149,42 +141,14 @@ export function search(query: string): GroupedSearchResults {
...
-
-    Object.keys(groupedResults).forEach((category) => {
-      const typedCategory = category as SearchResultCategory;
-      groupedResults[typedCategory].maxScore =
-        groupedResults[typedCategory].results[0]?.score || 0;
+    scoredResults.forEach((result) => {
+      const categoryResults = groupedResults[result.category].results;
+      if (categoryResults.length < MAX_RESULTS[result.category]) {
+        if (!result) return;
+        // @ts-expect-error
```
alex-page, line 147: I would like to clean this file up. It's frustrating that we need to duplicate the same code multiple times because of the types needed by `as IconsSearchResult[]` `as TokensSearchResult[]`...

Any thoughts? cc @aaronccasanova

polaris.shopify.com/src/components/GlobalSearch/GlobalSearch.tsx
```diff
@@ -172,176 +172,171 @@ function GlobalSearch({}: Props) {
       </button>
 
       <Dialog open={isOpen} onClose={() => setIsOpen(false)}>
-        <div className={styles.ModalBackdrop}></div>
```
martenbjork, line 175: This one blocks pointer interactions with background elements quite efficiently. Without it, you can interact with the background elements which feels a bit icky.

image

martenbjork:
Removing the `<>` made the diff impossible to read, so I temporarily re-added it while we hammer out the harder parts of the PR. Unchecked the todo in the PR description so that we can re-remove it before merging. (Sorry for purposefully reintroducing bad code.)

polaris.shopify.com/src/utils/search.ts
martenbjork: I was able to clean this code up by simplifying the search result types.

The search results are almost identical so they can use the same root type. The main difference is the `meta` object which contains different things depending on what type of search result it is.

New code: https://github.com/Shopify/polaris/pull/6502/files#diff-3e0197d18c6157be3d623c9dd24dc76a4b1b2cc8e300f0f99cbfa25bc85641b4R172

Simplified types: https://github.com/Shopify/polaris/pull/6502/files#diff-45904998eafd7e76b04406ff0b277701c8732c59ab9bd8ab93b8518118769497R48

polaris.shopify.com/src/components/GlobalSearch/GlobalSearch.tsx
alex-page: Ahhh sweet. I was thinking it did something but I couldn't quickly parse what. Maybe we should rename the className for now.

alex-page: Should it have `pointer-events: none`? That would make it more obvious in the CSS, right now it looks like an unused transparent overlay.

PR #6502, opened 2022-07-05
[polaris.shopify.com] Refactor site search (ready for review)

### LINKED ISSUE #6463: [polaris.shopify.com] Improve design of Foundations articles in global search
Design on foundations page:

Screen Shot 2022-07-01 at 12 08 48 AM

Design in search results:

Screen Shot 2022-07-01 at 12 08 56 AM

One of the design principles of the site is that "items are represented consistently". For instance, an icon/component/token looks the same in both global search and on their respective pages. We break this rule for Foundations article since they look different in the global search. Let's fix this!

### LINKED ISSUE #6546: [polaris.shopify.com] In global search, change limits to # of results shown for each type
See conversation here: https://github.com/Shopify/polaris/issues/6529

TLDR: For some searches, like "list", you should see more than 3 components since there are 6+ components that have "list" in their names. It's technically easy to fix this, but let's take it slow — we should rethink how results are scored and how we decide what to show.

martenbjork:
Cleans up the code that powers the global search. The new code is still complex, but less so. Implementing new types of search results is also easier thanks to the simpler `useGlobalSearchResults()` hook.

## UX updates

This refactor shouldn't change the search experience much. One update however, is that foundations results now render with the correct icon / color.

image

Related to #6529, this PR also increases the number of search results that are visible for each category, which fixes #6546.

Also fixes #6463.


alex-page (approved):

polaris.shopify.com/src/data/foundations.json
```diff
@@ -32,7 +32,7 @@
       ]
     },
     "intro": "Our mission is to make commerce better for _everyone_. Building products for everyone means creating inclusive content.",
-    "section": "content"
+    "category": "content"
```
alex-page, line 35: Love the consolidation here. I was unsure why it's called `content` but this is fine for now.

polaris.shopify.com/src/components/GlobalSearch/GlobalSearch.tsx
```diff
@@ -1,34 +1,58 @@
...
-
-interface Props {}
+import FoundationsGrid from "../FoundationsGrid";
+import { foundationsNavItems } from "../../data/navItems";
+
+const CATEGORY_NAMES: { [key in SearchResultCategory]: string } = {
+  components: "Components",
+  foundations: "Foundations",
+  tokens: "Tokens",
+  icons: "Icons",
```
alex-page, line 22: Do we need this? Could just be uppercase the key function

polaris.shopify.com/src/data/foundations.json
martenbjork: The category maps to the folder that the foundations content lives in: `foundations | design | content | patterns`.

polaris.shopify.com/src/components/GlobalSearch/GlobalSearch.tsx
martenbjork: I know, it was tempting to remove this very verbose part of the code. It just feels icky to hide the translations strings and make assumptions about language. If we add a more complex search result type like, `GitHub repos`, it's nice to be able to be explicit about capitalization.

PR #6515, opened 2022-07-05
[polaris.shopify.com] Move search to API [WIP]

martenbjork:
- Bad: Slightly slower searches
- Good: Smaller bundle size
- Good: Enables us to index the content of all the foundations pages

Prototype below. Notice how the excerpts match the search query.

image


polaris.shopify.com/src/utils/search.ts
```diff
@@ -171,6 +177,32 @@ export function search(query: string): GroupedSearchResults {
...
+                  const match = matchez[0];
+                  if (match.index) {
+                    const resultExcerpt = `${content
+                      .slice(match.index - 20, match.index + query.length + 150)
+                      .trim()
+                      .replace(/\n/g, " ")}`;
+                    if (result.meta.foundations && resultExcerpt) {
+                      result.meta.foundations = {
+                        ...result.meta.foundations,
+                        excerpt: resultExcerpt,
```
martenbjork, line 197: This mess will be cleaned up

PR #6548, opened 2022-07-07
[polaris.shopify.com] Remove keyboard and stepper components

alex-page:
These components are mobile only and we do not provide any react code.


martenbjork (approved):
💯

PR #6552, opened 2022-07-07
Remove platforms from front-matter

alex-page:
All components are web and react based right now. Removing this front matter.


martenbjork (approved):
YEEES

PR #6556, opened 2022-07-07
Rm deleted component og images

alex-page:
These snuck through, my bad


martenbjork (approved):

martenbjork:
We should automate that script so you don't have to manage these manually ☺️

PR #6570, opened 2022-07-08
Fix breakpoints with updated token structure

sam-b-rose:

Current `next` branch is broken with mixed usage of `tokens.breakpoints`

Updates breakpoint utilities to use either `metaTokens.breakpoints` or `tokens.breakpoints` token structures.

### DIFF
--- a/polaris-tokens/scripts/toTokenValues.ts
+++ b/polaris-tokens/scripts/toTokenValues.ts
@@ -1,13 +1,11 @@
 import fs from 'fs';
 import path from 'path';
 
-import type {MetaTokens, Tokens} from '../src';
+import {removeMetadata} from '../src';
+import type {Entry, Entries, MetaTokens, Tokens} from '../src/types';
 
 const outputDir = path.join(__dirname, '../build');
 const outputFile = path.join(outputDir, 'index.ts');
-
-type Entry<T> = [keyof T, T[keyof T]];
-type Entries<T> = Entry<T>[];
 
 export async function toTokenValues(metaTokens: MetaTokens) {
   if (!fs.existsSync(outputDir)) {
@@ -15,17 +13,10 @@ export async function toTokenValues(metaTokens: MetaTokens) {
   }
 
   const tokensEntries: Entries<Tokens> = Object.entries(metaTokens).map(
-    (entry) => {
+    (entry): Entry<Tokens> => {
       const [tokenGroupName, tokenGroup] = entry as Entry<MetaTokens>;
 
-      const tokenGroupValues = Object.fromEntries(
-        Object.entries(tokenGroup).map(([tokenName, {value}]) => [
-          tokenName,
-          value,
-        ]),
-      );
-
-      return [tokenGroupName, tokenGroupValues];
+      return [tokenGroupName, removeMetadata(tokenGroup)];
     },
   );
 
--- a/polaris-tokens/src/utilities.ts
+++ b/polaris-tokens/src/utilities.ts
@@ -1,5 +1,12 @@
-import type {Entry, Exact, MetaTokenGroup, Tokens, TokenGroup} from './types';
-import type {breakpoints as breakpointsTokenGroup} from './token-groups/breakpoints';
+import type {
+  Entry,
+  Exact,
+  ExtractValues,
+  MetaTokenGroup,
+  Tokens,
+  TokenGroup,
+} from './types';
+import type {breakpoints as metaBreakpointsTokenGroup} from './token-groups/breakpoints';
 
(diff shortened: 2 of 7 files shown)


polaris-tokens/src/utilities.ts
```diff
@@ -118,7 +118,9 @@ export function getCustomPropertyNames(tokens: Tokens) {
     .flat();
 }
 
-export type BreakpointsTokenGroup = typeof breakpointsTokenGroup;
+export type BreakpointsTokenGroup = {
+  [key in keyof typeof breakpointsTokenGroup]: string | TokenGroup;
+};
```
sam-b-rose: @aaronccasanova thoughts on opening this structure to either token structure (metaTokens or just tokens)?

Or do you think we should have two different types? Maybe something like the following:

```ts
type ExtractValues<T extends { value: string }> = {
  [key in keyof T]: T['value'];
}

export type MetaBreakpointsTokenGroup = typeof breakpointTokenGroup;
export type BreakpointsTokenGroup = ExtractValues<MetaBreakpointTokenGroup>;
```

aaronccasanova: I think we should keep it scoped to the `breakpoints` token group so we can ensure `polaris-react` pulls in the treeshaken named export.

The last suggestion looks sufficient 👍
```ts
// src/types.ts
type ExtractValues<T extends MetaTokenGroup> = {
  [K in keyof T]: T[K]['value'];
}

// src/utilities.ts
import {Exact, ExtractValues} from './types'
import type {breakpoints as metaBreakpointsTokenGroup} from './token-groups/breakpoints';

export type MetaBreakpointsTokenGroup = typeof metaBreakpointTokenGroup;
export type BreakpointsTokenGroup = ExtractValues<MetaBreakpointTokenGroup>;

export function getMediaConditions<
  T extends Exact<BreakpointsTokenGroup, T>
>(breakpoints: T) {...}
```

Also, I don't mind the dual type exports and think that's a good distinguishing naming convention (which we are already using e.g. `Tokens`/`MetaTokens`, `TokenGroup`/`MetaTokenGroup`, etc.). However, if after the rework we don't need the `MetaBreakpointsTokenGroup` export let's nix it, thoughts?

sam-b-rose: Alright, updated the types and moved removal of metadata to a utility function 👍 lmk wha you think @aaronccasanova

polaris-tokens/src/utilities.ts
```diff
@@ -118,7 +125,19 @@ export function getCustomPropertyNames(tokens: Tokens) {
...
+export function removeMetadata<T extends Exact<MetaTokenGroup, T>>(
+  tokenGroup: T,
+): ExtractValues<T> {
+  return Object.fromEntries(
+    Object.entries(tokenGroup).map(([tokenName, {value}]) => [
+      tokenName,
+      value,
+    ]),
+  ) as ExtractValues<T>;
+}
```
aaronccasanova:
```suggestion
export function removeMetadata<T extends Exact<MetaTokenGroup, T>>(
  tokenGroup: T,
) {
  return Object.fromEntries(
    Object.entries(tokenGroup).map((entry): Entry<TokenGroup> => {
      const [tokenName, {value}] = entry as Entry<MetaTokenGroup>;

      return [tokenName, value];
    }),
  ) as ExtractValues<T>;
}
```

Light suggestion to cast the `entry` itself in addition to the final `Object.fromEntries` result. This way we can validate each `.map` return type matches `Entry<TokenGroup>`.

polaris-tokens/scripts/toTokenValues.ts
```diff
@@ -1,32 +1,20 @@
...
-        Object.entries(tokenGroup).map(([tokenName, {value}]) => [
-          tokenName,
-          value,
-        ]),
-      );
-
-      return [tokenGroupName, tokenGroupValues];
-    },
+    ([tokenGroupName, tokenGroup]) =>
+      [tokenGroupName, removeMetadata(tokenGroup)] as Entry<Tokens>,
```
aaronccasanova:
```suggestion
    (entry): Entry<Tokens> => {
      const [tokenGroupName, tokenGroup] = entry as Entry<MetaTokens>;

      return [tokenGroupName, removeMetadata(tokenGroup)];
    },
```

Similar to the comment below, casting the `entry` itself allows us to validate the `.map` return type matches `Entry<Tokens>`.

polaris-tokens/src/utilities.ts
aaronccasanova: Update looks great! Left a couple light suggestions

aaronccasanova (approved):

Issue #6510, opened 2022-07-05
[Typography foundations] Add new typography tokens


laurkim:
Breaking change

Display medium/large variants in [prototype](https://github.com/Shopify/polaris/pull/6333) had hardcoded values for font-size/line-height that need to be added as typography tokens and existing tokens that are replaced will need to be hardcoded.

laurkim:
Make sure [this file](https://github.com/Shopify/polaris/blob/main/polaris-react/src/styles/shared/_typography.scss) is up to date with tokens

PR #6574, opened 2022-07-08
[Typography foundations] Update type scale and font size tokens

aveline:

Resolves https://github.com/Shopify/polaris/issues/6510

- Update type scale and font size token names
Screen Shot 2022-07-08 at 12 41 46 PM

- Hardcode existing `font-size` values outside the new scale
- Update `line-height` token values and hardcode values outside the new scale

No visual changes


polaris-react/src/styles/shared/_controls.scss
```diff
@@ -8,7 +8,7 @@
 
 @function control-vertical-padding() {
   @return calc(
-    (var(--p-line-height-6) - var(--p-line-height-3) - var(--p-space-05)) / 2
+    (36px - var(--p-line-height-3) - var(--p-space-05)) / 2
   );
```
laurkim:
```suggestion
  @return calc((36px - var(--p-line-height-3) - var(--p-space-05)) / 2);
```

For lint fix

chazdean (approved):
🎉

polaris-tokens/src/token-groups/typography.ts
```diff
@@ -7,44 +7,29 @@ export const typography = {
     value:
       "ui-monospace, SFMono-Regular, 'SF Mono', Consolas, 'Liberation Mono', Menlo, monospace",
   },
-  'font-size-1': {
+  'font-fontsize-75': {
```
aaronccasanova: **suggestion:** If we are renaming tokens I think it would be a good idea to prefix with the token group name (e.g. `typography-*`). Similar to what we established for the [`breakpoints` token group](https://github.com/Shopify/polaris/blob/main/polaris-tokens/src/token-groups/breakpoints.ts).

@samrose3 and I should probably document this somewhere, but we briefly discussed using plural token group names where possible (not needed for `typography`) and prefix tokens with token group names.

The idea being to still think about token relationships as nested object (even though we use a flat structure) e.g.
```js
const tokens = {
  typography: {
    'font-size-1': '...',
    'font-size-2': '...',  
  },
  colors: {
    primary: '...',
    secondary: '...',
  },
}

tokens.typography['font-size-1']
tokens.typography['font-size-2']
tokens.colors.primary
tokens.colors.secondary
```
Represented as CSS vars
```css
--p-typography-font-size-1: ...;
--p-typography-font-size-2: ...;
--p-colors-primary: ...;
--p-colors-secondary: ...;
```

cc @alex-page

polaris-tokens/src/token-groups/typography.ts
```diff
@@ -7,44 +7,29 @@ export const typography = {
...
     value: '12px',
   },
-  'font-size-2': {
-    value: '13px',
-  },
-  'font-size-3': {
+  'font-fontsize-100': {
```
aaronccasanova: **question:** Why did we remove the `-` in `font-size`? This makes it a bit harder for us to preform transforms on the token name in the future.

polaris-tokens/src/token-groups/typography.ts
sarahill: I was proposing prefixing `font-` so that it mapped closer to css and groups all the type related tokens together. `typography` does the same job of grouping them together though so I'm fine with that. If we think it's overkill though we can remove it and go with what was there previously though.

aveline:
## Update
There's an [ongoing discussion for token naming convention](https://github.com/Shopify/polaris/discussions/6613), but to unblock this PR we landed on:
- renaming this token group from `typography` to `font`
- using the token group name as the prefix
- not duplicating the token group name when the property already contains the same word. ex. `font-font-size` is just `font-size`

sarahill (approved):
✅

PR #6583, opened 2022-07-11
[IndexTable] Fix bulk actions on small screens

### LINKED ISSUE #4156: IndexTable iPhone view
## Issue summary

When using Bulk Actions with IndexTable, when the screen is shrunk to iPhone dimensions, the BulkAction buttons do not appear when selecting items.

## Expected behavior

Bulk Action button should appear when selecting items from IndexTable, even when screen size is iPhone dimensions

## Actual behavior

When screen is desktop dimensions, when I select an item from IndexTable, the BulkAction buttons appear.
However, when I shrink the screen to iPhone dimension, the BulkAction buttons do not appear when selecting from IndexTable.

1. Add an IndexTable with Bulk Actions to a page
1. Shrink browser window to iPhone dimensions
1. Select an item from IndexTable, the BulkAction buttons should appear

## Reduced test case

## Specifications

- Polaris version number: 6.2.1
- Browser: Chrome
- Device: PC
- Operating System: Windows 10

Paste the results here:

  System:
    OS: Windows 10 10.0.19041
    CPU: (8) x64 Intel(R) Core(TM) i7-3610QM CPU @ 2.30GHz
    Memory: 6.05 GB / 15.92 GB
  Binaries:
    Node: 14.5.0 - C:\Program Files\nodejs\node.EXE
    npm: 6.14.5 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 90.0.4430.93
    Edge: Spartan (44.19041.906.0), Chromium (90.0.818.51)
    Internet Explorer: 11.0.19041.1

```bash

```

### LINKED ISSUE #5598: [BulkActions] Not appearing on IndexTable small screen examples in Storybook
## Issue summary
`BulkActions` isn't showing in `IndexTable` small screen examples in Storybook (however it seems to be functioning correctly in production so seems like just an issue with our Storybook example).

The following code in `BulkActions.scss` is responsible for hiding it at a small screen size but I'm still not sure why there's a difference between Storybook and production...

Screen Shot 2022-04-18 at 2 39 52 PM

## Expected behavior
Clicking the "Select" button inside the `IndexTable` `HeaderWrapper` triggers the "Select" button to disappear and be replaced by `BulkActions`.

https://user-images.githubusercontent.com/21976492/163875207-79746a47-dd09-40f8-9f9d-41cdcd5f160f.mp4

## Actual behavior
Clicking the "Select" button inside the `IndexTable` `HeaderWrapper` triggers the "Select" button to disappear and is not replaced with anything. 

https://user-images.githubusercontent.com/21976492/163875228-c9d6c224-2c1c-48a7-af4b-8fe97d190381.mp4

1. Go to [Simple small screen index table](https://storybook.polaris.shopify.com/?path=/story/all-components-index-table--simple-small-screen-index-table) or [Small screen IndexTable with all of its elements](https://storybook.polaris.shopify.com/?path=/story/all-components-index-table--small-screen-index-table-with-all-of-its-elements)
2. Click the "Select" button.

### LINKED ISSUE #6405: [IndexTable] Bulk action buttons disappear on small screens
## Issue summary

For the `IndexTable` when the bulk actions feature is used, on narrow screens (less than 458px) the action buttons are not visible.

## Expected behavior

Bulk action buttons are always visible

Edit - Aditional info: The `ResourceList` component uses the same bulk actions but it's configuration is coordinated with the screen size so the actions are always visible

## Actual behavior
<details>
      <summary>Reproduction</summary>
      Index table with two rows. The first is selected and the bulk actions bar is visible on top with the selection info and two buttons. The screen is then resized (narrowing) and the bulk action bar disappear. Reverting it back to the previous size shows the actions again
    </details>

`.Polaris-BulkActions__Group--largeScreen` has a `display: none` that causes the issue. On bigger screens a media query (`(min-width: 28.625em)`) sets it to `display: flex`

1. Create a index table with bulk actions
2. Select at least one element of the table
3. Resize the screen until it is less than 458px wide
4. The bulk actions are no longer visible

## Reduced test case
https://codesandbox.io/s/8plzpj

## Specifications

- Polaris version number: 9.19.0
- Browser: Tested in chrome
- Device: MacBook
- Operating System: MacOs 12.4

FCalabria:

Fixes #6405 
Fixes #4156
Fixes #5598

(it's the same bug, it is duplicated)

It copies the mechanism on `ResourceList` to keep the bulk actions visible on all screen sizes. 

Previous

Fixed

Does not affect bigger screens

### DIFF
--- a/.changeset/nasty-pianos-shout.md
+++ b/.changeset/nasty-pianos-shout.md
@@ -2,4 +2,4 @@
 '@shopify/polaris': patch
 ---
 
-Fix bulk actions not visible on IndexTable on small screens
+Fixed `IndexTable` bulk action visibility on small screens
--- a/.changeset/nasty-pianos-shout.md
+++ b/.changeset/nasty-pianos-shout.md
@@ -1,0 +1,5 @@
+---
+'@shopify/polaris': patch
+---
+
+Fix bulk actions not visible on IndexTable on small screens
--- a/polaris-react/src/components/IndexTable/IndexTable.tsx
+++ b/polaris-react/src/components/IndexTable/IndexTable.tsx
@@ -104,6 +104,7 @@ function IndexTableBase({
   const [stickyWrapper, setStickyWrapper] = useState<HTMLElement | null>(null);
   const [hideScrollContainer, setHideScrollContainer] =
     useState<boolean>(false);
+  const [smallScreen, setSmallScreen] = useState(isSmallScreen());
 
   const tableHeadings = useRef<HTMLElement[]>([]);
   const stickyTableHeadings = useRef<HTMLElement[]>([]);
@@ -250,6 +251,13 @@ function IndexTableBase({
     handleCanScrollRight();
   }, [handleCanScrollRight]);
 
+  const handleIsSmallScreen = useCallback(() => {
+    const newSmallScreen = isSmallScreen();
+    if (smallScreen !== newSmallScreen) {
+      setSmallScreen(newSmallScreen);
+    }
+  }, [smallScreen]);
+
   const handleResize = useCallback(() => {
     // hide the scrollbar when resizing
     scrollBarElement.current?.style.setProperty(
@@ -260,7 +268,13 @@ function IndexTableBase({
     resizeTableHeadings();
     debounceResizeTableScrollbar();
     handleCanScrollRight();
-  }, [debounceResizeTableScrollbar, resizeTableHeadings, handleCanScrollRight]);
+    handleIsSmallScreen();
+  }, [
+    resizeTableHeadings,
+    debounceResizeTableScrollbar,
+    handleCanScrollRight,
+    handleIsSmallScreen,
+  ]);
 
   const handleScrollContainerScroll = useCallback(
     (canScrollLeft, canScrollRight) => {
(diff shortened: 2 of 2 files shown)


.changeset/nasty-pianos-shout.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/polaris': patch
+---
+
+Fix bulk actions not visible on IndexTable on small screens
```
chloerice:
```suggestion
Fixed `IndexTable` bulk action visibility on small screens
```

polaris-react/src/components/IndexTable/IndexTable.tsx
```diff
@@ -250,6 +251,13 @@ function IndexTableBase({
     handleCanScrollRight();
   }, [handleCanScrollRight]);
 
+  const handleIsSmallScreen = useCallback(() => {
+    const newSmallScreen = isSmallScreen();
+    if (smallScreen !== newSmallScreen) {
+      setSmallScreen(newSmallScreen);
+    }
```
chloerice: I think we can remove the conditional here because states only get updated if the value passed to the setter method is not the same as the current value.

```suggestion
    setSmallScreen(isSmallScreen());
```

chloerice (approved):
Tophatted on iPhone in Safari and Chrome and changes work great, thanks for jumping in to fix this @FCalabria!!

PR #6584, opened 2022-07-11
[polaris.shopify.com] Add list of props to each component

martenbjork:
Per popular request, this PR adds a basic props table to each component. 😋

The good:

- It covers 80% of what folks have been asking for — a simple way of viewing the props.
- The code that extracts the props uses the native Typescript AST and is ~80 lines long.
- It puts required props first in the list
- It highlights deprecated components
- There is a link that asks for feedback so that we can keep learning more about users' needs.

The bad:

- The script assumes that props are defined using an interface called `ComponentNameProps`. Some components, like Avatar, uses other constructs (like union types) to define the props. The script can't yet parse these.

image


polaris.shopify.com/package.json
```diff
@@ -15,7 +15,8 @@
     "parse-guidelines": "node src/scripts/parseGuidelines.mjs",
     "copy-icons": "node src/scripts/copy-icons.mjs",
     "generate-og-images": "next build && yarn next-sitemap && node src/scripts/generate-og-images.js",
-    "generate-og-images-skip-sitemap": "node src/scripts/generate-og-images.js"
+    "generate-og-images-skip-sitemap": "node src/scripts/generate-og-images.js",
+    "get-props": "tsc ./src/scripts/get-props.ts --target esnext --module node16 --types \"react\" --outDir ./src/scripts/get-props && node ./src/scripts/get-props/scripts/get-props.js && rm -rf ./src/scripts/get-props"
```
alex-page, line 19: I am happy to start with this. I think like our other packages `polaris-react` should export metadata and this should be a part of that export. Then we would just import the metadata from that package.

polaris.shopify.com/src/components/PropsTable/PropsTable.tsx
```diff
@@ -0,0 +1,71 @@
...
+    feedbackTitle
+  )}&amp;labels=polaris.shopify.com`;
+
+  return (
+    <div className={styles.PropsTable}>
+      <h2 id="properties">Props</h2>
+      <p>
+        This feature is a work in progress. Please{" "}
+        <a href={feedbackUrl}>share your feedback</a>.
+      </p>
```
alex-page: I liked this framing for new features from GitHub:

> Want to help make code search better? Provide [feedback here!](https://github.com/github/feedback/discussions/categories/code-search-and-navigation-feedback)

polaris.shopify.com/package.json
martenbjork: Yeah maybe! I definitely agree if the data stays this simple and generic.

Down the road, this object could get a lot more complex and tailored to the website's unique needs. I don't know yet!

polaris.shopify.com/src/components/PropsTable/PropsTable.tsx
alex-page: A `new` or `alpha` badge or something might also be nice to show that things are still being worked on.

polaris.shopify.com/src/data/props.json
```diff
@@ -0,0 +1,11536 @@
+[
```
alex-page, line 1: This file probably shouldn't exist but happy to start here.

alex-page (approved):
This PR is a great improvement and much needed by users. Lets ship and iterate.

polaris.shopify.com/src/components/PropsTable/PropsTable.tsx
martenbjork: I like that! Changed it to:

> Want to help make this feature better? Please [share your feedback](https://github.com/shopify/polaris/issues/new?title=%5Bpolaris.shopify.com%5D%20Props%20table%20feedback&amp;labels=polaris.shopify.com).

polaris.shopify.com/src/data/props.json
alex-page: Another way we could do this is generate this file in memory on the server then pass the necessary props to components. This would get ugly as we are reaching into `polaris-react` but with the right output I think we could clean this up.

martenbjork: Yeah, that sounds good to me. We should probably do this with most of these data files tbh — they are just intermediate layers that could be removed.

alex-page: Yep. On my radar. Not necessary immediately.

PR #6590, opened 2022-07-11
Update token tests

aveline:
Co-Authored-By: Aaron Casanova <32409546+aaronccasanova@users.noreply.github.com>

Update the tests so they are not so brittle and break when ever an update is made to tokens

### DIFF
--- a/.changeset/warm-grapes-joke.md
+++ b/.changeset/warm-grapes-joke.md
@@ -1,5 +1,0 @@
----
-'@shopify/polaris-tokens': patch
----
-
-Update tests to use mock tokens objects
--- a/.changeset/warm-grapes-joke.md
+++ b/.changeset/warm-grapes-joke.md
@@ -1,0 +1,5 @@
+---
+'@shopify/polaris-tokens': patch
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/warm-grapes-joke.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/polaris-tokens': patch
+---
+
+Update tests to use mock tokens objects
```
aaronccasanova:
```suggestion
Updated tests to use mock tokens objects
```

aaronccasanova (approved):

.changeset/warm-grapes-joke.md
sam-b-rose: Do we need a patch changeset if this is only updating tests? I think we could drop the changelog entry for this internal test update 👍

May need to either add an empty changeset or just apply the `Skip changelog` label to the PR to pass the CI check

sam-b-rose (approved):
Looks good!

Just a note that this will need to change on the next major since we are removing the color scheme layer. I can create a PR in the next branch to update it based on those changes 👍

PR #6592, opened 2022-07-11
[polaris.shopify.com] Improve layout

### LINKED ISSUE #5647: [Polaris website refresh] Add "on this page" TOC on docs pages
Parse headings and generate it on the fly.

Example:

Screen Shot 2022-04-27 at 4 23 16 PM

### LINKED ISSUE #6138: [Polaris website reboot] Sticky `Component` and `Foundations` page nav
Make the Nav inside the Layout component stick to the top as you scroll down the page.

If it's sticky, it also needs to handle overflow...
- How do you make the scroll bars look good?
- What happens when you've scrolled down inside the nav and go back to the top of the page?
- What part of the page is actually scrollable? How do you communicate that?

### LINKED ISSUE #6205: Current selection in Table of Content sticky nav should be highlighted

### LINKED ISSUE #6392: Challenging to navigate submenus
## Issue summary

It's challenging to navigate the submenus (example: `Grammar and Mechanics`, under `Content`). 

## Expected behavior

- It's easy to view the full menu on the right hand side
- Submenu items (for example, `Hyphens` under `Punctuation`) are indexed and come up in search results

## Actual behavior

- There is a long scroll to view the full menu on the right hand side (see video)
- The submenu items don't come up in search results (for example, `Hyphens` under `Punctuation`)

https://user-images.githubusercontent.com/39194032/175578543-4a145f8a-1e9e-4958-88c0-e5b5c2b28a04.mov

1. Navigate to an item in the left hand menu, like `Grammar and Mechanics`
2. Scroll to try to view the full list
3. Search for something in the right hand menu, like `Hyphens`

## Reduced test case

## Specifications

- Polaris version number:
- Browser:
- Device:
- Operating System:

Paste the results here:

```bash

```

martenbjork:
Many improvements to the layout and reading experience.

- Makes navigation and TOC sticky and scrollable. 
- Highlights current item in TOC as you scroll
- Immediately highlights the correct TOC item when you click it (by temporarily ignoring scroll events as the browser scrolls to that item)
- Uses native dark scrollbars on the body when using dark mode

https://user-images.githubusercontent.com/875708/179526814-87e30842-a9dd-4237-9f38-919745e2a46c.mp4

Fixes: #6138
Fixes: #6205
Fixes: #6392
Fixes: #5647


polaris.shopify.com/src/components/GlobalSearch/GlobalSearch.module.scss
```diff
@@ -12,6 +12,7 @@
   border-radius: var(--border-radius-round);
   gap: 0.5rem;
   font-weight: var(--font-weight-500);
+  background: var(--surface);
```
martenbjork, line 15: The header is now slightly translucent, but the button that triggers search should be non-transparent.

polaris.shopify.com/src/components/TOC/TOC.tsx
```diff
@@ -0,0 +1,145 @@
...
+  // Scan the headings from the bottom. The heading that comes first
+  // after the "content top margin" is the current heading.
+  for (let i = headings.length - 1; i >= 0; i--) {
+    const heading = headings[i];
+    const { top } = heading.getBoundingClientRect();
+    currentHeading = heading;
+    if (top < contentTopMargin) {
+      break;
+    }
+  }
```
alex-page, line 32: This should happen in markdown parser IMO. What we can do is, every time there is a heading we create a link like this:
```
<h1>
  <a id="content-polaris" class="anchor" href="#polaris"</a>Polaris
</h1>
```

Then we can on click of the TOC item scroll to the location of the anchor. I think this would remove a lot of complexity. We can also make the titles shareable by passing the anchor.

alex-page: I think we are kind of doing this, but there is a lot of complexity that I think could be removed.

polaris.shopify.com/src/components/Layout/Layout.module.scss
```diff
@@ -1,119 +1,89 @@
...
+  &::-webkit-scrollbar-thumb {
+    background: transparent;
+  }
+
+  &:hover {
+    &::-webkit-scrollbar-thumb {
+      background-color: var(--text-subdued);
+    }
+  }
+}
```
alex-page, line 21: Should this go in `mixins.scss`?

polaris.shopify.com/src/components/TOC/TOC.tsx
martenbjork: Yeah, this is how we create the ids: https://github.com/Shopify/polaris/blob/main/polaris.shopify.com/src/components/Markdown/Markdown.tsx#L25

polaris.shopify.com/src/components/Layout/Layout.module.scss
martenbjork: Maybe, be I don't think any other component than `Layout` should use these scrollbars.

alex-page (approved):
I think this can ship but we should probably remove a bunch of complexity by updating the markdown parser to create anchors. We can also add the titles to a larger navigation tree and that can create the list.

polaris.shopify.com/src/components/TOC/TOC.tsx
martenbjork: > I think we are kind of doing this, but there is a lot of complexity that I think could be removed.

Elaborate!

alex-page: All of the scroll to, calculate on page position logic could be removed to just use the default browser anchor link behaviour. There is a lot of complexity here.

martenbjork:
> I think this can ship but we should probably remove a bunch of complexity by updating the markdown parser to create anchors. We can also add the titles to a larger navigation tree and that can create the list.

I agree. 2 things that consistenly introduce technical debt:

- Markdown parsing
- Yaml parsing

polaris.shopify.com/src/components/TOC/TOC.tsx
martenbjork: The thing is, the browser doesn't take the fixed heading into account, so it awkwardly scrolls so that the target heading is hidden below the header.

alex-page: Ok. I don't know why the heading is fixed?

martenbjork: 1. Highly subjective, but I quite like the UX of it. Makes the site feel a bit more app-like.

2. To make the side navs scrollable, they need to have an overflow and a fixed height. Having a fixed header makes the height of the sidebars consistent and predictable (`calc(100vh - var(--header-height))`). Without a fixed header, their heights will range from `calc(100vh - var(--header-height))` (header is visible) and `100vh` (header is not visible). It adds another type of headache.

You can see how the not-fixed Vercel bar here messes with the sidebars and makes them cut off in the middle of the air. I want to avoid that.

https://user-images.githubusercontent.com/875708/179785550-e3bb9f06-9ffc-4a45-a1f3-add96a7dd06b.mov

PR #6596, opened 2022-07-11
Add <SearchableListbox /> component

carolopolo:

- This component's goal is to provide a consistent `searchable list in a popover` experience to multiple consumers spanning from internal admin commerce objects to non-admin 1P apps. This consistent UI composition will help us accomplish a unified UX across our organisation, but also, provide a useful and (imo) common component pattern to the public.

- Adds a new component, `<SearchableListbox />` to `src/components`

### How to 🎩

- Instead of a playground, please check out this branch, run storybook (`yarn && yarn dev`), and find the `SearchableListbox` folder in the left nav, and play around with the different examples.

### 🎩 checklist

### DIFF
--- a/polaris-react/src/components/SearchableListbox/README.md
+++ b/polaris-react/src/components/SearchableListbox/README.md
@@ -23,7 +23,7 @@ A SearchableListbox is a vertical, searchable list of interactive, customisable 
 A SearchableListbox is composed of:
 
 1. **Activator node**: Customisable activator to toggle popover containing list items
-2. **Search field**: Textfield for performing serach that can be toggled via the `showSearch` prop
+2. **Search field**: Textfield for performing search that can be toggled via the `showSearch` prop
 3. **List items:** Customisable options inside the list that users can select or deselect
 4. **Footer action:** Optional footer action placed below the list that can be used for actions such as list expansion
 
@@ -35,6 +35,7 @@ SearchableListboxes should:
 
 - Have clear activator node that toggles popover, and content that indicates selected list item
 - Limit the number of list items displayed at once; use the `footerAction` prop for list expansion
+- Be passed all search related props if `showSearch` prop is true
 - Indicate search empty state to user when search value doesn't match any results
 - Show loading state to user when list data is being generated
 
@@ -68,10 +69,6 @@ function BaseExample() {
         </Button>
       }
       open={open}
-      showSearch={false}
-      searchValue=""
-      searchEmptyStateMessage="No results found"
-      searchLabel="Search"
       listItems={[
         {value: '1', children: 'option 1', selected: true},
         {value: '2', children: 'option 2'},
@@ -79,7 +76,6 @@ function BaseExample() {
       onClose={() => {
         setOpen(false);
       }}
-      onSearch={() => {}}
       onOptionSelect={(value) => {
         console.log(value);
         setOpen(false);
@@ -172,15 +168,10 @@ function LoadingExample() {
       }
       open={open}
       loading
-      showSearch={false}
-      searchValue=""
-      searchEmptyStateMessage="No results found"
-      searchLabel="Search"
       listItems={[]}
       onClose={() => {
         setOpen(false);
       }}
-      onSearch={() => {}}
       onOptionSelect={(value) => {
         console.log(value);
         setOpen(false);
@@ -219,15 +210,10 @@ function FooterActionExample() {
         </Button>
       }
       open={open}
-      showSearch={false}
-      searchValue=""
-      searchEmptyStateMessage="No results found"
-      searchLabel="Search"
       listItems={listItems}
       onClose={() => {
         setOpen(false);
       }}
-      onSearch={() => {}}
       onOptionSelect={(value) => {
         console.log(value);
         setOpen(false);
@@ -280,10 +266,6 @@ function CustomOptionExample() {
         </Button>
       }
       open={open}
-      showSearch={false}
-      searchValue=""
-      searchEmptyStateMessage="No results found"
-      searchLabel="Search"
       listItems={[
         {
           value: '1',
@@ -315,7 +297,6 @@ function CustomOptionExample() {
       onClose={() => {
         setOpen(false);
       }}
-      onSearch={() => {}}
       onOptionSelect={(value) => {
         console.log(value);
         setOpen(false);
--- a/polaris-react/src/components/SearchableListbox/README.md
+++ b/polaris-react/src/components/SearchableListbox/README.md
@@ -1,0 +1,326 @@
+---
+name: SearchableListbox
+category: Lists and tables
+keywords:
+  - searchable list
+  - list with search
+  - listbox
+  - list box
+  - interactive list
+  - list
+  - search
+  - textfield
+---
+
+# SearchableListbox
+
+A SearchableListbox is a vertical, searchable list of interactive, customisable options
+
+---
+
+## Anatomy
+
+A SearchableListbox is composed of:
+
+1. **Activator node**: Customisable activator to toggle popover containing list items
+2. **Search field**: Textfield for performing serach that can be toggled via the `showSearch` prop
+3. **List items:** Customisable options inside the list that users can select or deselect
+4. **Footer action:** Optional footer action placed below the list that can be used for actions such as list expansion
+
+---
+
+## Best practices
+
+SearchableListboxes should:
+
+- Have clear activator node that toggles popover, and content that indicates selected list item
+- Limit the number of list items displayed at once; use the `footerAction` prop for list expansion
+- Indicate search empty state to user when search value doesn't match any results
+- Show loading state to user when list data is being generated
+
+---
+
+## Examples
+
+### Basic SearchableListbox
+
+Basic implementation
+
+```jsx
+function BaseExample() {
+  const [open, setOpen] = useState(false);
+
+  return (
+    <SearchableListbox
+      activatorNode={
+        <Button
+          plain
+          monochrome
+          removeUnderline
+          size="large"
+          textAlign="left"
+          disclosure="down"
+          onClick={() => {
+            setOpen(!open);
+          }}
+        >
+          Option 1
+        </Button>
+      }
+      open={open}
+      showSearch={false}
+      searchValue=""
+      searchEmptyStateMessage="No results found"
+      searchLabel="Search"
+      listItems={[
+        {value: '1', children: 'option 1', selected: true},
+        {value: '2', children: 'option 2'},
+      ]}
+      onClose={() => {
+        setOpen(false);
+      }}
+      onSearch={() => {}}
+      onOptionSelect={(value) => {
+        console.log(value);
+        setOpen(false);
+      }}
+    />
+  );
+}
+```
+
+### SearchableListbox with search enabled
+
+```jsx
+function ExampleWithSearch() {
+  const [open, setOpen] = useState(false);
+  const [searchValue, setSearchValue] = useState('');
+  const initialListItems = [
+    {value: 'option 1', children: <span>option 1</span>, selected: true},
+    {value: 'choice 2', children: <span>choice 2</span>},
+  ];
+  const [listItems, setListItems] = useState(initialListItems);
+
+  const handleSearch = (searchTerm) => {
+    setSearchValue(searchTerm);
+
+    const nextListItems = searchTerm
+      ? listItems.filter(({value}) => value.includes(searchTerm))
+      : initialListItems;
+    setListItems(nextListItems);
+  };
+
+  return (
+    <SearchableListbox
+      activatorNode={
+        <Button
+          plain
+          monochrome
+          removeUnderline
+          size="large"
+          textAlign="left"
+          disclosure="down"
+          onClick={() => {
+            setOpen(!open);
+          }}
+        >
+          Option 1
+        </Button>
+      }
+      open={open}
+      showSearch
+      searchValue={searchValue}
+      searchEmptyStateMessage="No results found"
+      searchLabel="Search"
+      listItems={listItems}
+      onClose={() => {
+        setOpen(false);
+        setSearchValue('');
+      }}
+      onSearch={handleSearch}
+      onOptionSelect={(value) => {
+        console.log(value);
+        setOpen(false);
+      }}
+    />
+  );
+}
+```
+
+### SearchableListbox in loading state
+
+```jsx
+function LoadingExample() {
+  const [open, setOpen] = useState(false);
+
+  return (
+    <SearchableListbox
+      activatorNode={
+        <Button
+          plain
+          monochrome
+          removeUnderline
+          size="large"
+          textAlign="left"
+          disclosure="down"
+          onClick={() => {
+            setOpen(!open);
+          }}
+        >
+          Option 1
+        </Button>
+      }
+      open={open}
+      loading
+      showSearch={false}
+      searchValue=""
+      searchEmptyStateMessage="No results found"
+      searchLabel="Search"
+      listItems={[]}
+      onClose={() => {
+        setOpen(false);
+      }}
+      onSearch={() => {}}
+      onOptionSelect={(value) => {
+        console.log(value);
+        setOpen(false);
+      }}
+    />
+  );
+}
+```
+
+### SearchableListbox with Footer Action
+
+```jsx
+function FooterActionExample() {
+  const [open, setOpen] = useState(false);
+  const [listItems, setListItems] = useState([
+    {value: 'option 1', children: 'option 1', selected: true},
+    {value: 'option 2', children: 'option 2'},
+  ]);
+  const [infiniteScrollEnabled, setInfiniteScrollEnabled] = useState(false);
+
+  return (
+    <SearchableListbox
+      activatorNode={
+        <Button
+          plain
+          monochrome
+          removeUnderline
+          size="large"
+          textAlign="left"
+          disclosure="down"
+          onClick={() => {
+            setOpen(!open);
+          }}
+        >
+          Option 1
+        </Button>
+      }
+      open={open}
+      showSearch={false}
+      searchValue=""
+      searchEmptyStateMessage="No results found"
+      searchLabel="Search"
+      listItems={listItems}
+      onClose={() => {
+        setOpen(false);
+      }}
+      onSearch={() => {}}
+      onOptionSelect={(value) => {
+        console.log(value);
+        setOpen(false);
+      }}
+      footerAction={
+        infiniteScrollEnabled ? undefined : (
+          <Button
+            plain
+            removeUnderline
+            textAlign="left"
+            onClick={() => {
+              setInfiniteScrollEnabled(true);
+              setListItems([
+                ...listItems,
+                {value: 'option 3', children: 'option 3'},
+                {value: 'option 4', children: 'option 4'},
+              ]);
+            }}
+          >
+            Show all
+          </Button>
+        )
+      }
+    />
+  );
+}
+```
+
+### SearchableListbox with custom list options
+
+```jsx
+function CustomOptionExample() {
+  const [open, setOpen] = useState(false);
+
+  return (
+    <SearchableListbox
+      activatorNode={
+        <Button
+          plain
+          monochrome
+          removeUnderline
+          size="large"
+          textAlign="left"
+          disclosure="down"
+          onClick={() => {
+            setOpen(!open);
+          }}
+        >
+          Option 1
+        </Button>
+      }
+      open={open}
+      showSearch={false}
+      searchValue=""
+      searchEmptyStateMessage="No results found"
+      searchLabel="Search"
+      listItems={[
+        {
+          value: '1',
+          children: (
+            <Stack wrap={false} alignment="leading">
+              <Stack.Item fill>
+                <Stack vertical spacing="extraTight">
+                  <span>option 1</span>
+                </Stack>
+              </Stack.Item>
+            </Stack>
+          ),
+          selected: true,
+        },
+        {
+          value: '2',
+          children: (
+            <Stack wrap={false} alignment="leading">
+              <Stack.Item fill>
+                <Stack vertical spacing="extraTight">
+                  <span>option 2</span>
+                </Stack>
+              </Stack.Item>
+              <Badge status="info">Draft</Badge>
+            </Stack>
+          ),
+        },
+      ]}
+      onClose={() => {
+        setOpen(false);
+      }}
+      onSearch={() => {}}
+      onOptionSelect={(value) => {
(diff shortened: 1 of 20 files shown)


polaris-react/src/components/SearchableListbox/SearchableListbox.tsx
```diff
@@ -0,0 +1,120 @@
...
+      {showEmptyState ? (
+        <SearchEmptyState message={searchEmptyStateMessage} />
+      ) : (
+        <Scrollable shadow onScrolledToBottom={onScrolledToBottom}>
+          <div className={styles.ListBoxWrapper}>
+            <Listbox
+              autoSelection={AutoSelection.None}
+              onSelect={onOptionSelect}
+              customListId={listId}
+              onActiveOptionChange={handleActiveOptionChange}
```
andrewmcgov, line 95: I think we probably want `enableKeyboardControl` here. 

```suggestion
              onActiveOptionChange={handleActiveOptionChange}
              enableKeyboardControl
```

polaris-react/src/components/SearchableListbox/SearchableListbox.tsx
```diff
@@ -0,0 +1,120 @@
...
+            >
+              {listItems?.map(({children, value, selected}) => {
+                return (
+                  <Listbox.Option value={value} selected={selected} key={value}>
+                    <Listbox.TextOption selected={selected}>
+                      {children}
+                    </Listbox.TextOption>
+                  </Listbox.Option>
```
andrewmcgov, line 103: What do you think about letting the parents fully control rendering of the items? I think it might allow the component to be more composable for different use cases. 

For example in the LocationPicker we use a `Listbox.Action` for the `Show all locations` action. This means its really easy to navigate with the keyboard to the footer action, and the appearance of the focus state is the same as any other list option: 

https://user-images.githubusercontent.com/14366097/179245751-f44a77e6-6cdd-47c5-9fbe-29b913c756ea.mov

Currently the `footerAction` in this component is rendered outside of the `Listbox`, so we can't use `Listbox.Action` there. This means you can navigate the the options with arrow keys, but need to tab down to the footer action. Then you can also end up with a weird double focus state: 

Screen Shot 2022-07-15 at 10 37 49 AM

I also tried just passing a `Listbox.Action` as the children of a `listItem`, but since we wrap all children in a `Listbox.Option`, and `Listbox.TextOption`, we also end up with a double focus state: 

Screen Shot 2022-07-15 at 10 37 27 AM

All this to say, what do you think about just rendering `listItems` directly here, and having the parent component set up the `Option`s, `TextOption`s, or `Action`s as needed? It would be a bit more work for parent components, but would allow a lot more flexibility. 

We could even have the items get passed into `SearchableListbox` as children directly if we wanted to. This more closely match the component API of `Listbox` too: 

```tsx
<Listbox accessibilityLabel="Basic Listbox example">
  <Listbox.Option value="UniqueValue-1">Item 1</Listbox.Option>
  <Listbox.Option value="UniqueValue-2">Item 2</Listbox.Option>
  <Listbox.Option value="UniqueValue-3">Item 3</Listbox.Option>
</Listbox>
```

Definitely open to hearing other opinions on this!

polaris-react/src/components/SearchableListbox/SearchableListbox.scss
```diff
@@ -0,0 +1,11 @@
+.SearchFieldWrapper {
+  padding: var(--p-space-4) var(--p-space-4) var(--p-space-2) var(--p-space-4);
```
andrewmcgov: I added a bunch of items to one of the examples and noticed that there is not enough padding on the bottom of the search field here: 

Screen Shot 2022-07-15 at 10 53 14 AM

polaris-react/src/components/SearchableListbox/SearchableListbox.scss
```diff
@@ -0,0 +1,11 @@
...
+}
+
+.ListBoxWrapper {
+  padding: var(--p-space-1) 0 var(--p-space-2);
```
andrewmcgov: When you search, it looks like the popover jumps to a different width. Im not sure the best way to solve this problem, since different implementations of this might want to have different widths. but maybe this is something we should think about: 

https://user-images.githubusercontent.com/14366097/179249809-9be1aec9-2bea-47bb-bb75-b1a0d182ad6a.mov

andrewmcgov (commented):
This is looking really good so far @carolopolo! Thanks for taking this on. Lots of teams are using a similar pattern here, so having something like this in Polaris will have a huge impact 👏 

I'm going to do another tophat and more thorough review early next week, but left a couple questions and suggestions from what I saw today. Really excited with where this is going!

polaris-react/src/components/SearchableListbox/SearchableListbox.tsx
jonathanhamel4: > For example in the LocationPicker we use a Listbox.Action for the Show all locations action.

We were initially using the Listbox.Action, but it didn't entirely match the styles from our original figmas iirc (overriding polaris spacing in web or something like that), but happy to use it now. It would indeed fix this double focus state. 

> what do you think about just rendering listItems directly here, and having the parent component set up the Options, TextOptions, or Actions as needed? It would be a bit more work for parent components, but would allow a lot more flexibility.

I feel like having a shared style for the selected states Option > TextOption could be useful, but I agree that it is limiting. It depends on how extensible we want this component to be or if it should follow the base UI for the [pattern](https://polaris.shopify.com/foundations/patterns/locations) and evolve with it 🤔  The `SearchableListbox` and the pattern aren't necessarily coupled though and can totally just be a tool to get to the pattern and allow for more flexibility.

polaris-react/src/components/SearchableListbox/SearchableListbox.scss
carolopolo: yea, our implementation in `web` actually enforces a fixed width of `rem(310px)`, but i felt it was wrong to enforce this in a polaris component, so i removed it. here are a couple options i thought of:

1. consumers could enforce a fixed width 
2. we could expose a new prop with fixed widths tied to them (`size?: 'small' | 'regular' | 'large'`, defaults to 'regular', AKA `rem(310px)`)

what else? wdyt? cc @jonathanhamel4

polaris-react/src/components/SearchableListbox/SearchableListbox.tsx
carolopolo: > For example in the LocationPicker we use a Listbox.Action for the Show all locations action.

totally aligned, thanks for this!

> what do you think about just rendering listItems directly here, and having the parent component set up the Options, TextOptions, or Actions as needed? It would be a bit more work for parent components, but would allow a lot more flexibility.

hmm... our mandate to app consumers is to provide a plug (data layer) & play solution that looks & feels the same as the segment picker, but i do agree that allowing more flexible compositions would also be awesome.

polaris-react/src/components/SearchableListbox/SearchableListbox.tsx
```diff
@@ -0,0 +1,120 @@
...
+  searchValue: string;
+  searchEmptyStateMessage: string;
+  searchLabel: string;
+  searchPlaceholder?: string;
+  listItems?: SearchableListItem[];
+  loading?: boolean;
+  loadingAccessibilityLabel?: string;
+  footerAction?: ReactElement;
+  onClose(): void;
+  onSearch(search: string): void;
```
christiefelker: While trying out the component, I noticed that it felt a bit odd to have `onSearch`, `searchLabel`, `searchValue` and `showSearch` all be required props on the component. Since I wanted to try an example without the search box, I needed to explicitly set `showSearch={false}`, and create a few unused defaults for the other related props. As long as it doesn't cause other issues, it might be nice for this component to have default values for them instead :)

polaris-react/src/index.ts
```diff
@@ -290,6 +290,12 @@ export type {ResourceItemProps} from './components/ResourceItem';
 export {ResourceList} from './components/ResourceList';
 export type {ResourceListProps} from './components/ResourceList';
 
+export {
+  SearchableListbox,
+  type SearchableListboxProps,
+  type SearchableListItem,
```
christiefelker: I noticed these lines seemed to be causing a TypeScript error when I was setting up my environment with the snapshots from this branch. 
Screen Shot 2022-07-18 at 11 30 58 AM

I resolved this locally by separating the types into a different line instead. 
```
export { SearchableListbox,  } from './components/SearchableListbox';
export type { SearchableListboxProps, SearchableListItem } from './components/SearchableListbox';
```

carolopolo: interesting i'm not getting those errors! but thanks for flagging, will fix :)

polaris-react/src/components/SearchableListbox/README.md
```diff
@@ -0,0 +1,326 @@
...
+A SearchableListbox is a vertical, searchable list of interactive, customisable options
+
+---
+
+## Anatomy
+
+A SearchableListbox is composed of:
+
+1. **Activator node**: Customisable activator to toggle popover containing list items
+2. **Search field**: Textfield for performing serach that can be toggled via the `showSearch` prop
```
yurm04:
```suggestion
2. **Search field**: Textfield for performing search that can be toggled via the `showSearch` prop
```

polaris-react/src/components/SearchableListbox/SearchableListbox.tsx
carolopolo: @felipeleusin @dmcvittie as consumers, how do you feel about the point above regarding having consumers fully control what type of `Listbox.[SubComponent]` are used to render each list item? initially, on the segmentation side, we wanted to provide a plug (data) & play, consistent UI to y'all so that's our rationale for using the `Listbox.Option` directly in this component. 

@andrewmcgov do you foresee a lot of use cases where you would not use `Listbox.Option` > `Listbox.TextOption`? it looked from the location picker mockups that this would be alright. i will definitely change the footer action to be a `Listbox.Action` of course

andrewmcgov: I lean towards having consumers control the rendering of the items because it allows for full flexibility of appearance and order of items. 

That being said, if the footer action is rendered as a `Listbox.Action` I am fine either way as it supports our use case!

polaris-react/src/components/SearchableListbox/SearchableListbox.scss
carolopolo: @chloerice maybe you could help us here ^. wdyt?

carolopolo: @felipeleusin @dmcvittie do you want to stick to the same width as locations/segmentation, or do you want to be able to have different widths?

felipeleusin: I think leaving the width to the consumer makes sense. In email we will arguably want an even bigger width to match our existing design so let's not hard code it. How is this width being enforced in core?

carolopolo: https://github.com/Shopify/web/blob/main/app/sections/Customers/CustomerListNext/components/SegmentList/components/ElementPicker/ElementPicker.scss

https://github.com/Shopify/web/blob/main/app/components/LocationPicker/LocationPicker.scss

carolopolo: so if that's the case, then we will go with option 2 from above:

2. we could expose a new prop with fixed widths tied to them (size?: 'small' | 'regular' | 'large', defaults to 'regular'

for location picker & segment picker in the admin currently, we both have a width of 300/310px, so we can maybe keep that as the 'regular' one. if you just want a larger width, you can tell us what width you'd like that to be. the proposed size prop is following existing polaris conventions.

@felipeleusin and @dmcvittie can you please share your use cases so we can decide on the width props?

felipeleusin: Could we replicate the `fluidContent` prop from popover? https://polaris.shopify.com/components/popover

carolopolo: all that does is:

```
.FluidContent {
  max-height: none;
  max-width: none;
}
```
that will still result in your implemented picker jumping to different widths when the search field is activated, for eg. i think the initial comment on this thread was coming from concern for the jumpiness demonstrated in the gif. the jumping UX is strange, don't you think?

polaris-react/src/components/SearchableListbox/SearchableListbox.tsx
felipeleusin: I wouldn't be too opposed to build the components using `Listbox.Option` directly if it makes us with a bit more flexible composition

polaris-react/src/components/SearchableListbox/SearchableListbox.scss
dmcvittie: Sorry for the delayed reply.

We have would like to have it take the full width in all cases for our implementation. The segment selector will always be inside a container that has the entire parent container width to consume.

Here is a screenshot of the existing component we're replacing:
Screen Shot 2022-07-20 at 11 15 41 AM

carolopolo: so, here is a summary of our width needs:

- we need to apply a fixed width to ensure the popover doesn't jump when search is activated
- flow wants the popover to take the full width of the activator
- email, segmentation, and locations can't have the popover take the full width of the activator
- locations and segmentation have the same fixed width need: `300/310px`
- email wants a bigger fixed width: TBD what the size is

based on this, i'm thinking we could go with the size prop:

`size?: 'small' | 'regular' | 'large'` => defaults to `regular`

- `small` would be `locations/segmentation` width at `310px`
- `regular` would be the default, width at `100%` => for `flow`
- `large` would be a fixed width that `email` defines as their need

how does this sound?👂🏼 @andrewmcgov @felipeleusin @dmcvittie

polaris-react/src/components/SearchableListbox/SearchableListbox.tsx
carolopolo: ok great! then i will just do:

`<Listbox>{listItems}</Listbox>` and let y'all decide what `Listbox.[Subcomponent]` to pass in then :)

carolopolo: just so we are all crystal clear on this, @felipeleusin @dmcvittie, this would mean that you would be doing this yourselves:

```
 {listItems?.map(({children, value, selected}) => {
                return (
                  <Listbox.Option value={value} selected={selected} key={value}>
                    <Listbox.TextOption selected={selected}>
                      {children}
                    </Listbox.TextOption>
                  </Listbox.Option>
                );
              })}

**Review by chloerice — CHANGES_REQUESTED** [review]
Hey y'all 👋🏽 

I hate to throw a wrench in this group effort, but the Polaris team has very recently added undertaking a complete overhaul of the Polaris React library to the roadmap for the rest of the year (more on that in the coming weeks). We are planning to deprecate domain specific implementations of system level patterns by end of year. While we don't have a strategy for where those components will move, if at all, we are aware and will be factoring in the gap to be filled of meeting the need for the shared domain specific patterns when we start solutioning.

In the meantime, the primitive components (Button, Popover, TextField, Listbox, EmptySearchResult, Scrollable etc) needed to compose the segment picker in any context (modal, popover, side nav, sheet etc) are all available and ready to be used. Shipping a composite component isn't needed to unblock shipping the feature in different contexts and would end up adding unnecessary tech debt for Customers in Q4 when the component is deprecated. The work our team will be doing in the second half of the year aims to make our primitive components as flexible as possible and move RnD toward composition over configuration, the latter of which is too often a blocker for teams currently as their sections evolve and opinionated Polaris components do not meet their changing needs.

Here is WIP Playground code serving as a draft for adding segment picker example to Listbox documentation and/or shipping as part of the picker pattern documentation work Winter and Raschin are working with Selene and Tracy to ship this week. 
<details>
  <summary>Segment picker example composition</summary>
      

```jsx

import {
  Icon,
  Page,
  Card,
  Layout,
  Button,
  Popover,
  TextField,
  Listbox,
  AutoSelection,
  Scrollable,
  EmptySearchResult,
} from '../src';
import {useUniqueId} from '../src/utilities/unique-id';

interface CustomerSegment {
  id: string;
  label: string;
}

const segments: CustomerSegment[] = [
  {
    label: 'All customers',
    id: 'gid://shopify/CustomerSegment/1',
  },
  {
    label: 'VIP customers',
    id: 'gid://shopify/CustomerSegment/2',
  },
  {
    label: 'New customers',
    id: 'gid://shopify/CustomerSegment/3',
  },
  {
    label: 'Abandoned carts - last 30 days',
    id: 'gid://shopify/CustomerSegment/4',
  },
  {
    label: 'Wholesale customers',
    id: 'gid://shopify/CustomerSegment/5',
  },
  {
    label: 'Email subscribers',
    id: 'gid://shopify/CustomerSegment/6',
  },
  {
    label: 'From New York',
    id: 'gid://shopify/CustomerSegment/7',
  },
  {
    label: 'Repeat buyers',
    id: 'gid://shopify/CustomerSegment/8',
  },
  {
    label: 'First time buyers',
    id: 'gid://shopify/CustomerSegment/9',
  },
  {
    label: 'From Canada',
    id: 'gid://shopify/CustomerSegment/10',
  },
  {
    label: 'Bought in last 60 days',
    id: 'gid://shopify/CustomerSegment/11',
  },
  {
    label: 'Bought last BFCM',
    id: 'gid://shopify/CustomerSegment/12',
  },
];

const lazyLoadSegments: CustomerSegment[] = Array.from(Array(100)).map(
  (_, index) => ({
    label: `Other customers ${index + 13}`,
    id: `gid://shopify/CustomerSegment/${index + 13}`,
  }),
);

segments.push(...lazyLoadSegments);

const interval = 25;

  const [pickerOpen, setPickerOpen] = useState(false);
  const [query, setQuery] = useState('');
  const [lazyLoading, setLazyLoading] = useState(false);
  const [loading, setLoading] = useState(false);
  const [willLoadMoreResults, setWillLoadMoreResults] = useState(true);
  const [visibleOptionIndex, setVisibleOptionIndex] = useState(interval);
  const [activeOptionId, setActiveOptionId] = useState(segments[0].id);
  const [selectedSegmentIndex, setSelectedSegmentIndex] = useState(0);
  const [filteredSegments, setFilteredSegments] = useState<CustomerSegment[]>(
    [],
  );

  const handleFilterSegments = (query: string) => {
    const nextFilteredSegments = segments.filter((segment: CustomerSegment) => {
      return segment.label
        .toLocaleLowerCase()
        .includes(query.toLocaleLowerCase().trim());
    });

    setFilteredSegments(nextFilteredSegments);
  };

  const handleQueryChange = (query: string) => {
    setQuery(query);

    if (query.length > 2) handleFilterSegments(query);
  };

  const handleQueryClear = () => {
    handleQueryChange('');
  };

  const handleOpenPicker = () => {
    setPickerOpen(true);
  };

  const handleClosePicker = () => {
    setPickerOpen(false);
  };

  const handleSegmentSelect = (segmentIndex: string) => {
    setSelectedSegmentIndex(Number(segmentIndex));
    handleClosePicker();
  };

  const handleActiveOptionChange = (_: string, domId: string) => {
    setActiveOptionId(domId);
  };

  /* This is just to illustrate lazy loading state vs loading state. Of course this is an example, so we aren't fetching from GraphQL. You'd use `pageInfo.hasNextPage` from your GraphQL query data instead of this fake "willLoadMoreResults" state along with setting `first` your GraphQL query's variables to your app's default max edges limit (e.g., 250 in Shopify/web). */

  const handleLazyLoadSegments = () => {
    if (willLoadMoreResults) {
      setLazyLoading(true);

      const options = query ? filteredSegments : segments;

      setTimeout(() => {
        const remainingOptionCount = options.length - visibleOptionIndex;
        const nextVisibleOptionIndex =
          remainingOptionCount >= interval
            ? visibleOptionIndex + interval
            : visibleOptionIndex + remainingOptionCount;

        setLazyLoading(false);
        setVisibleOptionIndex(nextVisibleOptionIndex);

        if (remainingOptionCount <= interval) {
          setWillLoadMoreResults(false);
        }
      }, 1000);
    }
  };

  const listboxId = useUniqueId('SearchableListbox');

  /* Your app's feature/context specific activator here */
  const activator = (
    <Button plain onClick={handleOpenPicker}>
      {segments[selectedSegmentIndex].label}
    </Button>
  );

  const textFieldMarkup = (
    <div style={{padding: '12px'}}>
      <TextField
        focused
        clearButton
        labelHidden
        label="Customer segments"
        placeholder="Type to search or filter segments"
        autoComplete="off"
        value={query}
        prefix={<Icon source={SearchMinor} />}
        ariaActiveDescendant={activeOptionId}
        ariaControls={listboxId}
        onChange={handleQueryChange}
        onClearButtonClick={handleQueryClear}
      />
    </div>
  );

  const segmentOptions = query ? filteredSegments : segments;

  const segmentList =
    segmentOptions.length > 0
      ? segmentOptions
          .slice(0, visibleOptionIndex)
          .map(({label, id}, index) => {
            const selected = segmentOptions[selectedSegmentIndex].id === id;

            return (
              <Listbox.Option key={id} value={`${index}`} selected={selected}>
                <Listbox.TextOption selected={selected}>
                  {label}
                </Listbox.TextOption>
              </Listbox.Option>
            );
          })
      : null;

  const lazyLoadingMarkup = lazyLoading ? (
    <Listbox.Loading
      accessibilityLabel={`${
        query ? 'Filtering' : 'Loading'
      } customer segments`}
    />
  ) : null;

  const noResultsMarkup =
    segmentOptions.length === 0 ? (
      <EmptySearchResult
        title=""
        description={`No segments found matching "${query}"`}
      />
    ) : null;

  const listboxMarkup = (
    <Listbox
      enableKeyboardControl
      autoSelection={AutoSelection.FirstSelected}
      onSelect={handleSegmentSelect}
      onActiveOptionChange={handleActiveOptionChange}
    >
      {segmentList}
      {noResultsMarkup}
      {lazyLoadingMarkup}
    </Listbox>
  );

  return (
    <Page title="Playground">
      <Layout sectioned>
        <Card sectioned title="Customer segment picker">
          <div style={{height: '225px'}}>
            <Popover
              active={pickerOpen}
              activator={activator}
              preferredAlignment="left"
              autofocusTarget="first-node"
              onClose={handleClosePicker}
            >
              <Popover.Pane fixed>{textFieldMarkup}</Popover.Pane>
              <Scrollable
                shadow
                style={{
                  height: '265px',
                  width: '310px',
                  padding: '4px 0 12px',
                }}
                onScrolledToBottom={handleLazyLoadSegments}
              >
                {listboxMarkup}
              </Scrollable>
            </Popover>
          </div>
        </Card>
      </Layout>
    </Page>
  );
}

```
</details>

PR #6597, opened 2022-07-11
Update legacy text style mixins to use tokens

aveline:

Relates to https://github.com/Shopify/web/issues/68542

Replaces hardcoded values and scss functions with css custom properties

### How to 🎩

- [ ] Verify there are no visual changes resulting from this update


alex-page (approved):
Code looks great. We could try to test the outputted CSS to compare if they are the same?

laurkim (approved):

PR #6536, opened 2022-07-07
[polaris-react] Remove ios and android content and examples

alex-page:
Removing unused mobile examples from markdown

- [x] Remove `content-for`
- [x] Remove `example-for`
- [x] Remove storybook build script magic
- [x] Remove Stepper component
- [x] Remove Key Accessories component
- [ ] Images will be removed in a follow up

PR #6620, opened 2022-07-13
Remove non existent examples frontmatter

alex-page:
These were removed in https://github.com/Shopify/polaris/pull/6536 and https://github.com/Shopify/polaris/pull/6549. This is some cleanup I missed.

- [x] Remove example frontmatter that does not have an example file


martenbjork (approved):
🔥

PR #6623, opened 2022-07-13
[Typography foundations] Add support for code blocks with `InlineCode` component

### LINKED ISSUE #6538: [Typography foundations] Add `code` variant to Text
`TextStyle` has a `code` [variation](https://github.com/Shopify/polaris/blob/852313946f538f25b6dbd49045a67bac437526bf/polaris-react/src/components/TextStyle/TextStyle.tsx#L13).

~When the new Text component [PR](https://github.com/Shopify/polaris/pull/6521) is merged, add a `monospaced` variant to add the text style and create a new `InlineCode` component that has the code variation styling (i.e., `background-color`). Expected implementation is that you would wrap `Text` with new `InlineCode` component to recreate what was originally `<TextStyle variation="code">`.~

Add support for `code` variant and styling to `Text`.

laurkim:

Resolves #6538.

~Adds a new `code` variant to the `Text` component and support for `code` element types. Also removes the `text-style-body` mixin from `AppProvider.scss` because it was causing issues with `font-size` in Safari.~
Adds a new `InlineCode` component to support inline code implementation inside a `Text` component.

[Storybook](https://5d559397bae39100201eedc1-xoqskaitpl.chromatic.com/?path=/story/playground-playground--playground) with TextStyle and Text comparisons.
example2

### How to 🎩

<details>

```jsx
import React from 'react';

import {Page, Stack, Text, TextStyle, InlineCode} from '../src';

  return (
    <Page title="Playground">
      <Stack vertical>
        <TextStyle variation="code">TextStyle code variation</TextStyle>
        <InlineCode>Inline code component</InlineCode>
        <Text as="h2" variant="headingLg">
          Text variant headingLg with <InlineCode>Inline code</InlineCode>
        </Text>
        <Text as="p" variant="bodyMd">
          Text variant bodyMd with <InlineCode>Inline code</InlineCode>
        </Text>
      </Stack>
    </Page>
  );
}
```

</details>

### 🎩 checklist


polaris-react/src/components/AppProvider/AppProvider.scss
```diff
@@ -10,8 +10,12 @@
 
 html,
 body {
-  @include text-style-body;
   @include text-emphasis-normal;
+  font-size: 16px;
```
aveline: Could this be updated to use tokens?

```
  font-size: var(--p-font-size-4);
  line-height: var(--p-line-height-2);
```

aveline (approved):
👍

polaris-react/src/components/AppProvider/AppProvider.scss
laurkim: Later on in the [file](https://github.com/Shopify/polaris/blob/ed29689180023ae8f9a3eb0316ba9853899a0926/polaris-react/src/components/AppProvider/AppProvider.scss#L40), there's other styling on the body that needed to be hard coded as well because AppProvider doesn't have access to the tokens. So I had to hardcode these values unfortunately.

aveline: Ok got it. I couldn't remember if we have access to tokens at that point.

alex-page: This is a bit confusing why the app provider doesn't have access to the tokens. Previously this made sense but we recently shipped a change to add the tokens globally without the AppProvider injecting them. cc @aaronccasanova @samrose3

aaronccasanova: Yea, I thought we'd have access to the tokens here as well. The intension is that custom properties are [globally set here in CustomProperties.scss](https://github.com/Shopify/polaris/blob/main/polaris-react/src/components/CustomProperties/CustomProperties.scss#L7) (and IIRC soon to move to AppProvider.scss).

I just made a quick codepen, to rule out a cascade/ordering issue, and had no issues applying `:root` styles above and below an `html, body` selector. I recommend looking into this a bit more before hard coding the values here.

laurkim: My PR Storybook worked fine in Chrome and Firefox but not Safari, which is why I had to hardcode. It was overriding the `font-size` for my `Text` component from the token value (`14px`) to `16.xxx px`.
Was your codepen working in Safari @aaronccasanova?

polaris-react/src/components/Text/Text.scss
```diff
@@ -117,3 +117,13 @@
...
 }
+
+.code {
+  display: inline-block;
+  background-color: var(--p-surface-subdued);
+  border-radius: var(--p-border-radius-base);
+  box-shadow: inset 0 0 0 1px var(--p-border-subdued);
+  font-family: var(--p-font-family-mono);
+  font-size: var(--p-font-size-100);
+  padding: 0 var(--p-space-1);
```
aaronccasanova: **suggestion:** Can we extract out non-typographic properties (e.g. `display`, `background-color`, `border-radius`, `box-shadow`, and `padding`)? Text variants are commonly a composite of type styles and this appears to be conflating the collection with a particular business application.

If this is a common pattern in our codebase we can leverage composition to encapsulate the business requirement e.g.
```scss
// Code.scss
.root {
  display: inline-block;
  background-color: var(--p-surface-subdued);
  box-shadow: inset 0 0 0 1px var(--p-border-subdued);
  padding: 0 var(--p-space-1);
}
```
```tsx
// Code.tsx
function Code(props) {
  return <Text {...props} className={styles.root} variant="code" as="code" />
}
```

laurkim: I had originally had an `InlineCode` wrapper with all the non-typographic styles but removed it based on Slack thread feedback! Will add it back in 👍

polaris-react/src/components/AppProvider/AppProvider.scss
alex-page: @dominikwilkowski was telling me to run it through a `calc` which does nothing in Safari for `code` and `pre` elements. A bit odd.

laurkim: Refactoring the logic to render:

```
<p className="Text-bodyMd">lorem ipsum <code className="InlineCode-Code">hello world</code></p>
```

It resolved the issue with the `font-size` token being overwritten so undoing changes to `AppProvider.scss`.

laurkim:
Updated [Storybook](https://5d559397bae39100201eedc1-kuneahgnwq.chromatic.com/?path=/story/playground-playground--playground) with example implementation that I discussed offline with @sarahill 
inline-code-example

With this approach, `InlineCode` inherits `font-size` from the parent wrapping it, i.e., when wrapped in:

- Text variant headingLg, `<code>` font-size is `var(--p-font-size-200)` or `16px`
- Text variant bodyMd, `<code>` font-size is `var(--p-font-size-100)` or `14px`

sarahill (approved):
✅

polaris-react/src/components/InlineCode/InlineCode.scss
```diff
@@ -0,0 +1,8 @@
+.Code {
+  display: inline-block;
```
alex-page:
```suggestion
  display: inline;
```

Just wanted to double check `inline-block was needed. It is sometimes with padding/border.

polaris-react/src/components/InlineCode/InlineCode.scss
```diff
@@ -0,0 +1,8 @@
...
+  background-color: var(--p-surface-subdued);
+  border-radius: var(--p-border-radius-base);
+  box-shadow: inset 0 0 0 1px var(--p-border-subdued);
```
alex-page: Could be a border?

polaris-react/src/components/InlineCode/InlineCode.scss
```diff
@@ -0,0 +1,8 @@
+.Code {
+  display: inline;
```
aveline: We can remove this declaration since `<code>` is inline by default

aveline (approved):
Looks good in paragraphs 👍
Screen Shot 2022-07-19 at 11 19 26 AM

laurkim:
> Looks good in paragraphs 👍 Screen Shot 2022-07-19 at 11 19 26 AM

Updated fontsize to be `0.9em` in `InlineCode` based on feedback:

text-example

PR #6630, opened 2022-07-14
[polaris.shopify.com] Fix background color of code examples

martenbjork:
#6430 introduced a new background color for inline code examples `like this one`. However, this broke multi-line code examples. This PR fixes the issue so that both designs are supported.

Thanks @jbalsas for filing a very thorough bug report that helped us debug this quickly! 🔥 

Before:

Screen Shot 2022-07-14 at 12 52 08 PM

After:

Screen Shot 2022-07-14 at 12 52 11 PM

### DIFF
--- a/.changeset/silly-melons-visit.md
+++ b/.changeset/silly-melons-visit.md
@@ -1,0 +1,5 @@
+---
+'polaris.shopify.com': patch
+---
+
+Fix background colors of code examples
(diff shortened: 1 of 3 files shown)


jbalsas (approved):
Thanks for the quick fix @martenbjork!!

Makes sense to me code-wise! I haven't built this locally to 🎩 it, though... wondering if we wouldn't be generating previews for PRs to avoid having to do that!

martenbjork:
Thanks for the ✅

> wondering if we wouldn't be generating previews for PRs to avoid having to do that

Yeah, that would be great. Would love to have visual diffing and/or deploy previews a la Netlify! I think it's too hard to solve right now, but it would definitely be great down the road.

PR #6636, opened 2022-07-14
[polaris.shopify.com] Fix Do and Don't icons

martenbjork:
Before:

image

After:

image


alex-page (approved):

alex-page:
@martenbjork does this still work for components? I think the icons showed there.

martenbjork:
> does this still work for components?

Yes, I think so! It's not tied to a specific page, but rather the markdown parsing magic that adds the `.usage-block` hack.

PR #6642, opened 2022-07-15
[polaris.shopify.com] Fix up mobile padding

martenbjork:
Fixes a regression in `Layout` that was introduced in #6430.

Before:

image

After:

image

### DIFF
--- a/polaris.shopify.com/src/pages/contributing/[doc].tsx
+++ b/polaris.shopify.com/src/pages/contributing/[doc].tsx
@@ -3,7 +3,6 @@ import fs from "fs";
 import path from "path";
 import glob from "glob";
 
-import Container from "../../components/Container";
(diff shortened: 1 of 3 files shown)


polaris.shopify.com/src/pages/contributing/[doc].tsx
```diff
@@ -21,15 +20,13 @@ const contributingDirectory = path.join(process.cwd(), "content/contributing");
 
 const Contributing: NextPage<Props> = ({ readme, title }: Props) => {
   return (
-    <Container>
```
Comment (martenbjork) [inline polaris.shopify.com/src/pages/contributing/[doc].tsx:24]: @chloerice Sorry that this isn't better documented yet, but `Layout` doesn't need a `Container`, so I removed this and re-added the padding around `Layout`.

gui-santos (approved):

PR #6645, opened 2022-07-15
[polaris.shopify.com] Minor improvements to contributing page

martenbjork:
### Make the page layout consistent:
✅ The page wrapper now has the same width across all pages.

✅ The TOC is now visible on all pages. This is consistent with all the other pages on the site. This also reduces the width of the article itself which makes the line length appropriate for readability ([a11y guidelines](https://www.w3.org/WAI/tutorials/page-structure/styling/#line-length)). 

### Link to overview
✅ There was no way to get back to find the "overview" (`/contributing`) page in the side nav. I found this confusing. This adds an item to the nav called "Introduction". What do you think? "Overview"?

Before:

image

After:

image

Related to #6430

### DIFF
--- a/.changeset/tricky-paws-hope.md
+++ b/.changeset/tricky-paws-hope.md
@@ -1,0 +1,5 @@
+---
+'polaris.shopify.com': patch
+---
+
+Improve layout of contribution guidelines so that its width is consistent between pages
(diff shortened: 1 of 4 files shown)


jjgali (approved):
Thanks @martenbjork! These are great little fixes. 

I think I'm okay with `Introduction`. I was trying to see if we had an existing pattern for intro/overview titles, but I'm struggling to find one. 

Is this something we could change pretty easily later on?

martenbjork:
> Is this something we could change pretty easily later on?

Definitely! Changing that text is a 1 minute fix.

We can also rethink the navs / IA down the road if the current pattern is too inflexible.

PR #6658, opened 2022-07-18
[polaris.shopify.com] Table style tweaks

sarahill:
Slight table style adjustments

Changes:
- Updated border to use light border color
- increased left/right padding for screens 1440px +

| Before | After |
| --- | --- |
| image | image | 
| image | image |


sarahill:
@zecarlostorre here's the updates we made together on Friday. Some things to note that you can see in the before / after:
 - the increased padding causes some table headers to wrap
 - the contrast is low between surface and alternating row colors. I think this might be fine just calling it out as a discussion point.

I'm kind of leaning toward not increasing the padding or we can try removing the left/right padding altogether and making the text flush to the edge as it was before if we are really concerned about the text alignment. Open to playing around with this more though

@martenbjork as we were adjusting colors we were wanting things to be slightly lighter so I added specific variables for table colors. I don't know that we want this long term as they're just more grays. I think the better solution is to have a spectrum of grays that map to the variables.

One thing we were also wondering was why we don't just use the variables from polaris? I know they're not great but I think that just highlights the gap and with Joe's color work that should fix the issues.

polaris.shopify.com/src/components/Longform/Longform.module.scss
```diff
@@ -174,14 +174,22 @@
     font-size: var(--font-size-200);
     border-collapse: separate;
     border-spacing: 0;
-    border: var(--border);
+    border: 1px solid var(--border-color-light);
     border-radius: 0.5rem;
     width: 100%;
 
+    tr:nth-child(even) {
```
martenbjork: TIL! I've always done this `nth-child(2n)`. This is much more excellent.

polaris.shopify.com/src/components/Longform/Longform.module.scss
```diff
@@ -174,14 +174,22 @@
...
+      background: var(--table-row);
+    }
+
+    tr:nth-child(odd) {
+      background: var(--surface);
```
martenbjork: Is this needed? The background behind the table should already be `--surface` 👀

martenbjork:
> I think the better solution is to have a spectrum of grays that map to the variables.

I agree!

> One thing we were also wondering was why we don't just use the variables from polaris? 

I think this might be the outcome in the end! For now though, I think it helps the website move fast to be a little bit unsystematic and allow new tokens to be created ad hoc. I think we're still in the "explore needs" phase rather than "optimize system" phase. (Could be wrong!)

martenbjork:
- The rows look nice! Less border-y
- The padding seems a bit inefficient
- Do we need this many grays? I don't mind, just worth re-evaluating before merging.

zecarlostorre:
@sarahill 

**Padding:** Indeed, the extra padding is a bit problematic here. I agree with keeping it the way it was.
**Stripes:** Indeed the contrast is a bit low. I thought we had given up on them, but maybe they could work if we bump up the contrast a touch?

One question (CC @martenbjork): How are we dividing up the space? I mocked it up with 48px padding and they could fit. It seems like we're wasting a ton of space in between columns.

Another thought/experiment: We could also bump down the text size. Example on the right had 12pt for headings, 14pt for cells.

image

zecarlostorre (approved):
Sorry for the delay @sarahill, must have accidentally close the tab and forgot to get back to you. Generally, this looks good to me, so I'd ship it!

The only thing that still bugs me is the massive space between columns, but looks like that's in there for both before and after. Only now is a bit more noticeable because we added more space on the margins.

Issue #6650, opened 2022-07-15
Polaris style is not applied on components


denniss:
## Issue summary

I am setting up a new React app (on Ruby on Rails) and I want to use the Polaris components. After following the set up, I am able to see some styles being applied but it doesn't seem that the button UI looks correct. Here's the code that I use:

Content of `./javascript/packs/ShopifyAppIndex.jsx`

```
import React from 'react'
import enTranslations from '@shopify/polaris/locales/en.json';
import {AppProvider, Page, Card, Button} from '@shopify/polaris';

class ShopifyApp extends React.Component {
    render() {
        return (
            <AppProvider i18n={enTranslations}>
                <Page title="Example app">
                    <Card sectioned>
                        <Button>Example button</Button>
                    </Card>
                    <button className="Polaris-Button">Example button</button>
                </Page>
            </AppProvider>
        );
    };
}

export default ShopifyApp
```

Content of `./javascript/packs/ShopifyAppIndex.jsx`

```
import React from 'react'
import ReactDOM from 'react-dom'
import ShopifyApp from '../components/ShopifyApp'

ReactDOM.render(
    <ShopifyApp/>,
    document.getElementById('app')
)
```

The `react_home.html`. Note that I am using the style tag because I couldn't get the 
`import '@shopify/polaris/build/esm/styles.css';` line to work.

```
<!DOCTYPE html>
<html>
<head>
  <title>Shopify</title>
  <%= csrf_meta_tags %>
  <%= csp_meta_tag %>

  <link
    rel="stylesheet"
    href="https://unpkg.com/@shopify/polaris@9.8.0/build/esm/styles.css"
    />

  <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
</head>
<body>
<div id="app"></div>
<%= javascript_pack_tag 'ShopifyAppIndex' %>
</body>
</html>
```

And this is what I see:

<img width="871" alt="Screen Shot 2022-07-15 at 7 58 03 AM" src="https://user-images.githubusercontent.com/385638/179250057-68e02682-fc9d-4dbe-816d-7649be445cdd.png">

I notice the following warning under the console tab. I wonder if this could be the issue:
<img width="1640" alt="Screen Shot 2022-07-15 at 8 06 56 AM" src="https://user-images.githubusercontent.com/385638/179251716-c002cbbc-846d-4fd6-ab54-c153abc15435.png">

## Actual behavior

I expect the button UI to show up.

[...]

PR #6663, opened 2022-07-18
Update CSS version in readme

martenbjork:
The `polaris-react` CSS that was published to the CDN was broken for a few versions. It didn't contain the tokens / CSS Custom Properties which caused the layout to break. See #6650.

- 9.8.0: Missing tokens in `:root` https://unpkg.com/@shopify/polaris@9.8.0/build/esm/styles.css
- 9.20.0: Tokens are back in `:root` https://unpkg.com/@shopify/polaris@9.20.0/build/esm/styles.css

This PR updates the readme so that it points to a newer, fixed version of the CSS file.

### DIFF
--- a/.changeset/short-planets-cross.md
+++ b/.changeset/short-planets-cross.md
@@ -1,0 +1,5 @@
+---
+'@shopify/polaris': patch
+---
+
+Update stylesheet url in Readme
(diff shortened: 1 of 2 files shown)


alex-page (approved):

alex-page:
@samrose3 this used to be automated, now we do deploys in CI we might want to check if that still happens.

sam-b-rose:
This step is getting skipped now since it was part of the `version` npm script in `/polaris-react`. I think we can add a step to our root-level `version` command that will call any custom `preversion` scripts (like updating the README) then call the official `changesets version` command. I'll get a draft PR to see how this could work

alex-page:
Nice @samrose3 thank you!

PR #6664, opened 2022-07-18
Call package preversion scripts before running changeset version

sam-b-rose:

Fixes #6663 

This PR runs the `preversion` scripts of each package before versioning the packages (using `changesets`) for release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "format": "prettier . --write",
     "type-check": "tsc --build",
     "changeset": "changeset",
+    "preversion-packages": "turbo run preversion",
     "version-packages": "changeset version",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -34,6 +34,7 @@
     "format": "prettier . --write",
     "type-check": "tsc --build",
     "changeset": "changeset",
+    "perversion-packages": "turbo run perversion",
```
alex-page:
```suggestion
    "perversion-packages": "turbo run perversion",
```

This is a very perverted change.

sam-b-rose: Oooof good catch 😅 fixed!

alex-page (approved):

PR #6667, opened 2022-07-19
Use set versions for workspace package dependencies

### LINKED ISSUE #6666: Dependency versions should be explicit
## Issue summary

We're getting this error:

```
Attempted import error: 'getMediaConditions' is not exported from '@shopify/polaris-tokens' (imported as 'getMediaConditions').
warning ../../node_modules/@shopify/polaris/build/esm/utilities/breakpoints.js
```

`@shopify/polaris` v9.20.0 is trying to import `getMediaConditions` from `@shopify/polaris-tokens@npm:5.4.0`, but this export has only existed since v5.5.0.

## Expected behavior

`@shopify/polaris` v9.20.0 should have shipped with a dependency on `@shopify/polaris-tokens` v5.5.0 or newer. This is the only signal to a package manager that the transitive dependency needs to be upgraded.

## Actual behavior

https://github.com/Shopify/polaris/blob/b3e027c9b45fa44d1137fd409b2f8a10687c90ef/polaris-react/package.json#L65-L67

I can provide a test case if needed, but given this is a pretty obvious gap, I'm hoping you'll just accept the report as-is.

In the meantime we can work around this by directly installing and upgrading `@shopify/polaris-tokens` (and `@shopify/polaris-icons`) ourselves.

sam-b-rose:

Fixes #6666

Sets the workspace package dependencies to specific versions rather than the `*` version wildcard.

These specific versions should be automatically updated and bumped when using `changeset version` on release 🤖 🚀

### DIFF
--- a/.changeset/curly-paws-ring.md
+++ b/.changeset/curly-paws-ring.md
@@ -1,0 +1,8 @@
+---
+'polaris-for-figma': patch
+'polaris-for-vscode': patch
+'@shopify/polaris': patch
+'polaris.shopify.com': patch
+---
+
+Use exact versions for Polaris workspace dependencies.
--- a/polaris-for-figma/package.json
+++ b/polaris-for-figma/package.json
@@ -12,7 +12,7 @@
   },
   "dependencies": {
     "@ant-design/icons": "^4.7.0",
-    "@shopify/polaris": "*",
+    "@shopify/polaris": "^9.20.0",
     "antd": "^4.20.4",
     "lodash": "^4.17.15",
     "react": "^16.0.0",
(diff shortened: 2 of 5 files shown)


aaronccasanova:
Thanks @samrose3! Looks like we're just missing one `@shopify/polaris-tokens: "*"` update in `polaris-for-vscode`. Curious if you ran `yarn` after to check if there were any `yarn.lock` updates?

aaronccasanova (approved):

sam-b-rose:
@aaronccasanova good catch, will update VS Code package as well.

> Curious if you ran yarn after to check if there were any yarn.lock updates?

Yes, no lock file changes here. Will run again just to double check.

alex-page (approved):

PR #6668, opened 2022-07-19
[polaris.shopify.com] Improve examples browsing experience

### LINKED ISSUE #6424: [polaris.shopify.com] Code and design should be visible simultaneously

### LINKED ISSUE #6425: [polaris.shopify.com] Redesign examples nav to not use a <select>

martenbjork:
A few easy wins for the examples browsing experience 🚀 The examples experience will keep on evolving (towards a full playground) but in the mean time, this PR contains some nice improvements:

- No more select. Just list the examples on the page. Much easier to scan. (Using HeadlessUI's tab component for accessibility.)
- Show preview and code simultaneously. Removes the awkward tabs and reduces context switching.

Fixes #6424
Fixes #6425

image

### DIFF
--- a/polaris.shopify.com/src/components/Examples/Examples.module.scss
+++ b/polaris.shopify.com/src/components/Examples/Examples.module.scss
@@ -15,6 +15,7 @@
     border-radius: var(--border-radius-400);
     font-size: var(--font-size-100);
     line-height: 1.33;
+    color: var(--text-strong);
 
     &:hover {
       box-shadow: var(--card-shadow-hover);
--- a/polaris.shopify.com/src/components/Examples/Examples.module.scss
+++ b/polaris.shopify.com/src/components/Examples/Examples.module.scss
@@ -23,7 +23,10 @@
 
     &[aria-selected="true"] {
       background: var(--search-highlight-color);
-    }
+      &:focus {
+        box-shadow: var(--focus-outline);
+      }
+    }  
   }
 }
 
--- a/polaris.shopify.com/src/components/Examples/Examples.module.scss
+++ b/polaris.shopify.com/src/components/Examples/Examples.module.scss
@@ -1,64 +1,33 @@
-.SelectContainer {
-  position: relative;
-  display: inline-block;
-  width: 100%;
+.Examples {
+  display: flex;
+  flex-wrap: wrap;
+  grid-template-columns: 1fr 1fr 1fr 1fr;
+  gap: 0.5rem;
+  margin-bottom: 1rem;
 
-  select {
-    width: 100%;
-    height: 2.5rem;
-    line-height: 2.5rem;
-    padding-left: 0.75rem;
-    padding-right: 0.75rem;
-    background-color: var(--surface-subdued);
-    color: var(--text);
-    border-radius: var(--border-radius-500);
-    font-size: var(--font-size-400);
-    box-shadow: 0 0 0 1px var(--border-color);
+  button {
+    padding: 0.6rem 0.85rem;
+    text-align: left;
+    display: flex;
+    align-items: flex-start;
+    background: transparent;
+    box-shadow: var(--card-shadow);
+    border-radius: var(--border-radius-400);
+    font-size: var(--font-size-100);
+    line-height: 1.33;
 
-    &:focus-visible {
-      box-shadow: var(--focus-outline);
+    &:hover {
+      box-shadow: var(--card-shadow-hover);
+    }
+
+    &[aria-selected="true"] {
+      background: var(--search-highlight-color);
     }
   }
 }
 
-.SelectIcon {
-  pointer-events: none;
-  position: absolute;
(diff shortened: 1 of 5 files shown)


polaris.shopify.com/src/components/Examples/Examples.module.scss
```diff
@@ -1,68 +1,49 @@
...
-  right: 1rem;
-  transform: translateY(-50%);
-  z-index: 1;
-
-  svg {
-    height: 1.2rem;
-    width: 1.2rem;
+    &[aria-selected="true"] {
+      background: var(--search-highlight-color);
+    }
```
chloerice: When keyboard focus moves to the active tab, we should indicate visually that the tabs are in the page's tab order by adding a focus ring to the active tab:

```suggestion
      &:focus {
        box-shadow: var(--focus-outline);
      }
    }  
```

alex-page (approved):
Looks great. I was thinking the tabs might not be useful as it is just changing a URL for the iFrame and the codesandbox... HOWEVER with the code being rendered and changing for the examples it makes sense.

chloerice (approved):
This is a great iteration on the scannable list, folks will now be able to tab in and out of the list of examples without needing to go through all of them to get to the example itself 🎉 🙌🏽  Let's use this approach for the icon grid as well (tab to the list, arrows to navigate) so folks can quickly and easily tab back and forth between the icon list and the details in the sidebar 🚀

alex-page:
> Let's use this approach for the icon grid as well (tab to the list, arrows to navigate) so folks can quickly and easily tab back and forth between the icon list and the details in the sidebar 🚀

We have to be careful where we use arrow key and tab navigation. This is a website and the majority of the ui should be navigable by tabbing. My thought on the icons page is that it should be tabs as we don't want users to skip over the list of icons. On this page arrow keys are fine as we wouldn't mind if someone wanted to skip over examples.

Tabbing is also native and requires no code to be added to change that functionality.

chloerice:
> > Let's use this approach for the icon grid as well (tab to the list, arrows to navigate) so folks can quickly and easily tab back and forth between the icon list and the details in the sidebar 🚀
> 
> We have to be careful where we use arrow key and tab navigation. This is a website and the majority of the ui should be navigable by tabbing. My thought on the icons page is that it should be tabs as we don't want users to skip over the list of icons. On this page arrow keys are fine as we wouldn't mind if someone wanted to skip over examples.
> 
> Tabbing is also native and requires no code to be added to change that functionality.

I agree the site needs to be navigable by tab, but I have to push back on the way that's done. Implementing the icon grid using the layout grid pattern is needed to make the experience of exploring icons for folks using keyboard or assistive technology analogous to the visual user experience. You and I are able to quickly visually relate the icon grid to the aside with the icon details. But there is no semantic relationship between the two in the markup to enable that same experience or back and forth exploration for keyboard and screen reader users.

This website and the Shopify admin are responsible for following core web usability standards. Markup needs to be implemented so that the information architecture is not solely visual. Otherwise, we're only serving the needs of one kind of user and making pages and features extremely confusing, time consuming to parse, and often unusable for folks using assistive technologies or simply using the keyboard instead of a mouse.

This can be felt very quickly when attempting to navigate the site via keyboard. For example, navigation menu items are expected to function similar to tabs in that you tab to the nav and use arrows to navigate. Implementing a nav without menuitem roles and expected keyboard navigation makes the semantic grouping of the buttons unclear and doesn't scale. Tabbing through the top nav of the site is not arduous, but continuing on into the side nav in sections like Foundations and Components is not good experience. The "Skip to content" button isn't meaningful if the section and in page navigation are not part of the tab order as widgets for users to choose to enter into and use instead of having to tab through the entire page manually in order to understand where they are and how to use the page/site.

For the site to be usable for everyone and give every user the same experience of the IA, the section side nav should follow the [treeview navigation pattern](https://www.w3.org/WAI/ARIA/apg/example-index/treeview/treeview-navigation.html), the site nav should follow the [menubar pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menu/), tokens list and component example lists should use [the tabpanel pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabpanel/), and grids like the icon grid should use [the layout grid pattern](https://www.w3.org/WAI/ARIA/apg/example-index/grid/LayoutGrids).

alex-page:
I get what you mean @chloerice but I think that is over-engineering semantics. We can have very simple accessible solutions.

martenbjork:
I agree with Chloe, especially with regards to the icon grid. I think using tabbing to move between the icons is frustrating and unintuitive. 

In the full icons grid, I can get to the right icon so much quicker if I can just move there using the arrow keys. It's an awesome feature that is in line with the goal of the site: helping you be producitve.

In search, arrow keys are used to navigate between items. I would expect both directions to be supported. Right now, it's quite strange when you arrow down and the focus suddenly move sideways.

PR #6678, opened 2022-07-20
[polaris.shopify.com] Clean up example titles

alex-page:
- [x] Removed component names from example titles as they are on the page title
- [x] Removed `simple` or `basic` from example titles
- [x] Updated README, content and examples dir
- [x] Fixed markdown issues when investigating migration
- [x] Tophatted all examples and logged issues in https://github.com/Shopify/polaris/issues/6706


.changeset/odd-wolves-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+'polaris.shopify.com': minor
+---
+
+Example descriptions support for markdown content
```
martenbjork: You have 2 changesets in this PR

polaris-react/src/components/Card/README.md
```diff
@@ -450,7 +450,7 @@ Use to render custom content such as icons, links, or buttons in a card section
 </Card>
 ```
 
-### Card with all of its elements
+### With all of its elements
```
martenbjork: Maybe even:

> With all elements

martenbjork (approved):

polaris-react/src/components/VideoThumbnail/README.md
```diff
@@ -33,7 +33,7 @@ Video thumbnails should:
 
 ## Examples
 
-### Basic video thumbnail
+### Default
```
martenbjork, line 36: Help a tired person out: Why do we not delete these readmes? Aren't they 1:1 with what's in the website directory? Is that part of your migration strategy?

martenbjork (approved):

polaris-react/src/components/VideoThumbnail/README.md
alex-page: I did this so I could make sure we didn't lose any content as it hadn't been migrated across in a long time. We definitely should delete them but first I need to make sure the content is not lost.

martenbjork: Gotcha!

.changeset/odd-wolves-attack.md
martenbjork: Now 3 🤔

PR #6682, opened 2022-07-20
Remove fullSizeExample, omitAppProvider and hidePlayground frontmatter

alex-page:


martenbjork (approved):
Ship it! 

(I want to acknowledge that these are probably very useful and that we're likely to reinvent them soon — but the process of gaining understanding and learning is useful enough to motivate it.)

PR #6688, opened 2022-07-20
[polaris.shopify.com] Include HTML source for examples

### LINKED ISSUE #6341: [polaris.shopify.com] Feedback (on page /components/index-table)
Please include HTML markup along side of the React component markup on new site.

Thx!!

### LINKED ISSUE #6528: [polaris.shopify.com] HTML examples of components
Hey,

The new Polaris documentation looks amazing 🔥 Thanks for updating it!

However, the new version doesn't have HTML examples of components. It was important since not everyone is using react implementation of components. We have alternative implementations without React. For example, Polaris ViewComponents for Rails: https://polarisviewcomponents.org/. It's a pretty big library to create apps using server-side Rails instead of React. It would be great if you could add HTML examples to the new version of documentation 🙏 It will make the maintenance of libraries like Polaris ViewComponents easier going forward.

Kirill

### LINKED ISSUE #6531: [polaris.shopify.com] Make "intro paragraph detection" less naive/fragile
On the component pages, the code that detects the "intro paragraph" assumes that the entire paragraph lives on a single line. This is sometimes not true, as authors sometimes prefer to add some line breaks for readability.

The code should be updated to take multi-line paragraphs into account.

Problematic code:

https://github.com/Shopify/polaris/blob/main/polaris.shopify.com/src/pages/components/%5Bcomponent%5D.tsx#L68

image

### LINKED ISSUE #6559: broken content in AppProvider docs
## Issue summary

There are some content [extraction?] bugs in the [AppProvider](https://polaris.shopify.com/components/app-provider) docs, and maybe other components?

image

### LINKED ISSUE #6603: [polaris.shopify.com] View HTML and CSS classes for components
In the earlier version of the Polaris website, it was possible to view the plain HTML version of the components. 
Although we're using react for frontend, there were multiple occasions that made us use the HTML version to extend the features of existing components.

**This is really useful for multiple use cases:** 
- Our being to use a component/icon where in the react component only string is accepted
- Use the Polaris design without react. Use plain html, vue, etc instead
- Extend the functionality. For instance, the button onClick currently won't give the event object to the callback. We had a use case where this was needed

martenbjork:
🚀 Adds React/HTML switcher to component examples
🚀 Renames `Examples` component to `ComponentExamples`.
🚀 Renames `CodeExample` component to `Code`. Adds ability to contain multiple examples, rendered in tabs.
🚀 Uses this component to render code everywhere: examples, markdown, icons page.
🚀 Code now respect settings for light and dark mode. Light code in light mode, dark code in dark mode.
🚀 Fixes the broken `[object object]` code in AppProvider docs. We were parsing the markdown both in `getInitialProps` and in `ReactMarkdown` causing HTML, not markdown, to be passed to the `Markdown` component. Stringify HTML and you get `[object object]`...

Closes #6603
Closes #6528
Closes #6341
Closes #6531
Closes #6559

https://user-images.githubusercontent.com/875708/180210955-dc0089a7-96bf-480e-bd26-c9e2c90cfe9f.mp4

Tested with VoiceOver:

https://user-images.githubusercontent.com/875708/180212422-661c9795-a312-49bb-956d-4979157feb81.mov


polaris.shopify.com/src/components/Examples/Examples.tsx
```diff
@@ -74,7 +112,7 @@ const Examples = (props: Props) => {
                     height={iframeHeight}
                     width="100%"
                     onLoad={handleExampleLoad}
-                    ref={iframeRef}
```
martenbjork, line 78: I found the refs approach to be flaky. Maybe because the ref changes due to the Tabs. Using the `id` to find the iframe consistency works.

polaris.shopify.com/src/components/Examples/Examples.tsx
```diff
@@ -15,26 +15,63 @@ interface Props {
...
-      );
-    const padding = 192;
-    let newHeight = padding;
-
-    if (exampleElement) {
-      newHeight += exampleElement?.offsetHeight;
-    }
+    let attempts = 0;
+
+    const waitForExampleContent = setInterval(() => {
```
martenbjork: Just because the iframe's `onLoad` handler fires doesn't mean that React example has had the time to render. This timer polls the iframe to check if the example has _actually_ been rendered.

.changeset/nine-goats-leave.md
```diff
@@ -0,0 +1,5 @@
+---
+'polaris.shopify.com': minor
+---
+
+Simplify rendering of code blocks on site. Use the same component for every code example. This makes the experience more consistent and ensures that every code block has features like copy paste.
```
chloerice:
```suggestion
Simplified rendering of code blocks on site by using the same component for every code example. This makes the experience more consistent and ensures that every code block has features like copy/paste.
```

polaris.shopify.com/src/components/Code/Code.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { useState } from "react";
+
+interface Props {
+  tabs: {
+    title: string;
+    code: string;
+  }[];
+}
+
+function Code({ tabs }: Props) {
```
chloerice: When I first read the PR description I assumed that the `Code` component was a generic wrapper for any code sample on the site, but it looks like it's specific to `ComponentExamples`. Should we move the sub-component into a `components` directory inside of the `ComponentExamples` directory? Can the component be more generic, maybe the `tabs` prop could be optional? For example, rendering the `Code` component for the `/contributing/shipping-your-contribution` code page doesn't completely translate for things like yarn commands.

polaris.shopify.com/src/components/ComponentExamples/ComponentExamples.tsx
```diff
@@ -0,0 +1,141 @@
...
+    if (element.match(/^<?\w[^>]*[^\/]$/) && !element.startsWith("input")) {
+      indent += tab;
+    }
+  });
+
+  return result.substring(1, result.length - 3);
+}
+
+const ComponentExamples = (props: Props) => {
+  const { examples } = props;
```
chloerice:
```suggestion
const ComponentExamples = ({examples}: Props) => {
```

polaris.shopify.com/src/components/ComponentExamples/ComponentExamples.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+  useEffect(() => {
+    setSelectedIndex(0);
+  }, [examples]);
+
+  return (
+    <>
+      <h2 id="examples">Examples</h2>
+
+      <Tab.Group
```
chloerice, line 84: Is there an opportunity to make the button-like `tabs` their own sub-component? They might be useful to replace the `Select` in the `QuickStartGuides` of the `/contributing` section as well 👀

chloerice (approved):
This looks great Marten!! 🤌🏽 Had a couple questions and suggestions but nothing blocking 😁

polaris.shopify.com/src/components/ComponentExamples/ComponentExamples.tsx
martenbjork: Maaaaybe! Although I'm not sure we would want this styling for other types of tabs — it's very specific for the `Code` component. All the logic and a11y features come from `HeadlessUI` so that part is already reusable for other parts of the site.

polaris.shopify.com/src/components/Code/Code.tsx
martenbjork: It is generic! Is there anything about it that made you think otherwise?

The tabs are optional — if you pass in a single example, `Code` will render it without tabs.

polaris.shopify.com/src/components/Code/Code.tsx
```diff
@@ -0,0 +1,98 @@
+import Tooltip from "../Tooltip";
+import Prism from "prismjs";
+import { useCopyToClipboard } from "../../utils/hooks";
+import styles from "./Code.module.scss";
+import { Tab } from "@headlessui/react";
+import Image from "../Image";
+import { useState } from "react";
+
+interface Props {
+  tabs: {
```
chloerice:
```suggestion
  tabs?: {
```

polaris.shopify.com/src/components/Code/Code.tsx
```diff
@@ -0,0 +1,98 @@
...
+}
+
+function HighlightedCode({ code }: { code: string }) {
+  return (
+    <div
+      className={styles.ActualCode}
+      dangerouslySetInnerHTML={{
+        __html: Prism.highlight(code, Prism.languages.javascript, "javasript"),
+      }}
+    ></div>
```
alex-page: This should probably be `pre` instead of div

polaris.shopify.com/src/components/Code/Code.tsx
alex-page: Should we just have two components? `CodeTabs` and `CodeBlock` or something? We kind of already do with `HighlightedCode`.

polaris.shopify.com/src/components/ComponentExamples/ComponentExamples.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  const [iframeHeight, setIframeHeight] = useState("400px");
+
+  const handleExampleLoad = () => {
+    let attempts = 0;
+
+    const waitForExampleContent = setInterval(() => {
+      const exampleIframe = document.getElementById(
+        "examples-iframe"
+      ) as HTMLIFrameElement;
```
alex-page: Urgh. iFrames suck. When can we use the nextjs beta and try out layout lol

alex-page: Could we maybe add a comment about what this does or why it's needed? I could imagine coming back here in a week and being confused as heck.

polaris.shopify.com/src/data/components.json
```diff
@@ -91,9 +91,7 @@
         "customer avatar",
         "business avatar",
         "customer thumbnail",
-        "business thumbnail",
-        "ios",
-        "android"
```
alex-page: Going to make a seperate PR for this. It shouldn't be in this PR.

alex-page: Latest from main will clean this up.

alex-page (approved):
Nice clean up, thanks @martenbjork

polaris.shopify.com/src/components/Code/Code.tsx
alex-page: Maybe? It seems like prism does some stuff...

polaris.shopify.com/src/components/Examples/Examples.tsx
alex-page: yayayaya yep yep. iframes are just urgh

polaris.shopify.com/src/components/Code/Code.tsx
martenbjork: Good catch! Fixed!

polaris.shopify.com/src/components/Code/Code.tsx
martenbjork: I clarified the component props a bit which maybe (?) makes the component clearer:

To render some code:

```
<Code code={{ title: "My example", code }} />
```

To render multiple pieces of code:

```
<Code code={[
  { title: "Styles", code: cssCode },
  { title: "HTML", code: htmlCode },
]} />
```

In short: It's not about the tabs, it's about what you want to render. The tabs are just a happy side effect.

Happy to refactor this down the road if it causes confusion!

polaris.shopify.com/src/components/ComponentExamples/ComponentExamples.tsx
martenbjork: I updated the code to be even more explicit about what's going on, hopefully that makes it self documenting.

I'm also keen not to use iframes at all. Idea: Import the Polaris CSS and prefix every selector with a class?

polaris.shopify.com/src/components/Code/Code.tsx
```diff
@@ -0,0 +1,109 @@
...
+function HighlightedCode({ code }: { code: string }) {
+  return (
+    <pre>
+      <code
+        className={styles.ActualCode}
+        dangerouslySetInnerHTML={{
+          __html: Prism.highlight(
+            code,
+            Prism.languages.javascript,
+            "javasript"
```
JaKXz, line 73: Found a typo!

also with Prism, it’s best practice to set Prism.manual = true somewhere at the top of this file. It shouldn’t make a difference here because you haven’t used the classes on the containing element that prism searches for to replace the content, but, that can get funky. Prism.manual is also just a flag for the internals of prism not to run unless called so it’s a minor optimization but worth it :)

alex-page: Fixed the typo here b3e14b733119519952cc1fbf20541c5ad43022d6

Are you able to write up an issue around what `Prism.manual = true` would change or enhance with the code highlighting? I am taking a look and am not too sure I understand if it's needed.

PR #6695, opened 2022-07-20
Update Releasing.md

aveline:

Add guidance to the Releasing docs around tagging contributors for reviews


philschoefer:
thank you this will be very helpful!

philschoefer (approved):

philschoefer:
Not sure if we should check in #polaris if this is the wording that we want.

PR #6697, opened 2022-07-20
[polaris.shopify.com] Create typography-updates.md

sarahill:
Adding a page to the style guide for the typography updates post. After chatting with @alex-page we are going to add a news/updates section for content like this to live.

Fixes https://github.com/orgs/Shopify/projects/2250/views/63

Screen Shot 2022-07-25 at 8 29 06 AM

Screen Shot 2022-07-25 at 8 29 17 AM

Screen Shot 2022-07-25 at 8 29 29 AM


martenbjork:
Beautiful! 

Quick thought: You, José, Erin and I (and others) should align on how to work with graphics:

- We should probably have a Figma file where all the website assets live
- We should export as much as possible as SVGs
- We should let the website do the rounding of corners so that it stays consistent across all images. It also allows us to change this later or on different breakpoints.
- We should let the website add borders to graphics (if we want that). That way, borders remain 1px, even when scaled down.

Not a blocker for now, but something for us to discuss before we start shipping tons of assets.

sarahill:
Thanks @martenbjork that makes sense!

I updated the images to:
- remove border radius
- remove any borders
- adjusted the aspect ratio

For now I just made a new page for "Assets" with a subpage for "News" to house these images. https://www.figma.com/file/GV7HHVTBCEPDXnemkQ1rv3/Foundations-guidance?node-id=965%3A67328

We'll have to decide what we want the radius and border to be but I'm fine with removing them for now and adding it via CSS when that decision is made. Feels better for scaling and updating later.

cc: @adrianefong @zecarlostorre @erncat

martenbjork:
> I updated the images to:

Aces! Thanks, Sara! 🙌  These graphics are beautiful BTW.

sarahill:
@martenbjork does this require a changelog entry? I know this PR is also dependent on the addition of the new "News / Updates" section which I assume will be a separate PR but wasn't sure.

martenbjork:
Yeah, right now anything going into `main` needs a changelog entry. I don't mind this being merged for now, and then we'll add the news functionality very soon in a separate PR.

martenbjork (approved):

PR #6701, opened 2022-07-20
Create `@shopify/polaris-migrator` package

sam-b-rose:
Adds the beginnings of the `@shopify/polaris-migrator` package. This package applies code transformations to help update 
Polaris apps and the Polaris codebase.

## Usage

```sh
npx @shopify/polaris-migrator <migration> <path>
```

Run tests for `@shopify/polaris-migrator`

```sh
yarn test --filter=@shopify/polaris-migrator 
```

### DIFF
--- a/polaris-migrator/package.json
+++ b/polaris-migrator/package.json
@@ -25,8 +25,10 @@
   },
   "dependencies": {
     "@babel/core": "^7.18.9",
+    "@babel/plugin-transform-runtime": "^7.18.9",
     "@babel/runtime": "^7.18.9",
     "@babel/helper-plugin-utils": "^7.18.9",
+    "@shopify/babel-preset": "^24.1.5",
     "chalk": "^4.1.0",
     "globby": "11.0.1",
     "is-git-clean": "^1.1.0",
--- a/polaris-migrator/package.json
+++ b/polaris-migrator/package.json
@@ -1,0 +1,44 @@
+{
+  "name": "@shopify/polaris-migrator",
+  "version": "0.0.0",
+  "description": "Codemod transformations to help upgrade your Polaris codebase",
+  "license": "SEE LICENSE IN LICENSE.md",
+  "author": "Shopify <dev@shopify.com>",
+  "homepage": "https://polaris.shopify.com",
+  "repository": "https://github.com/Shopify/polaris",
+  "bugs": {
+    "url": "https://github.com/Shopify/polaris/issues"
+  },
+  "publishConfig": {
+    "access": "public",
+    "@shopify:registry": "https://registry.npmjs.org"
+  },
+  "bin": "dist/index.js",
+  "main": "dist/index.js",
+  "scripts": {
+    "build": "rollup -c",
+    "dev": "rollup -c -w",
+    "start": "node ./dist/index.js",
+    "lint": "TIMING=1 eslint --cache .",
+    "test": "jest",
+    "clean": "rm -rf .turbo node_modules dist *.tsbuildinfo"
+  },
+  "dependencies": {
+    "@babel/core": "^7.18.9",
+    "@babel/runtime": "^7.18.9",
+    "@babel/helper-plugin-utils": "^7.18.9",
+    "chalk": "^4.1.0",
+    "globby": "11.0.1",
+    "is-git-clean": "^1.1.0",
+    "jscodeshift": "^0.13.1",
+    "meow": "^9.0.0",
+    "p-map": "^4.0.0",
+    "postcss-scss": "^4.0.4",
+    "postcss": "^8.4.14"
+  },
+  "devDependencies": {
+    "@types/babel__helper-plugin-utils": "^7.10.0",
+    "@types/is-git-clean": "^1.1.0",
+    "rollup-plugin-preserve-shebang": "^1.0.1"
+  }
+}
--- a/polaris-migrator/rollup.config.mjs
+++ b/polaris-migrator/rollup.config.mjs
@@ -1,0 +1,52 @@
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import * as url from 'node:url';
+
+import {babel} from '@rollup/plugin-babel';
+import {nodeResolve} from '@rollup/plugin-node-resolve';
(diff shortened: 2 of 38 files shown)


polaris-migrator/src/utilities/testUtils.ts
```diff
@@ -0,0 +1,55 @@
...
+  mode?: 'only' | 'skip' | 'standard';
+  before?: () => void;
+  after?: () => void;
+}
+
+export function createCheckFn(testDir: string) {
+  return check(testDir);
+}
+
+function check(testDir: string) {
```
BPScott, line 21: jscodeshift exposes a `runTest` utility https://github.com/Shopify/web/blob/main/tooling/codemods/tests/remove-autobind.test.js#L4 (though it lacks types, so you'll need to [add those yourself](https://github.com/Shopify/web/blob/main/tooling/typescript/modules.d.ts#L73-L81))

It seems to do the same as everything you're doing here. Is it worth leveraging that already existing utility instead of writing our own?

polaris-migrator/rollup.config.mjs
```diff
@@ -0,0 +1,52 @@
...
+    // Allows node_modules resolution
+    nodeResolve({extensions, preferBuiltins: true}),
+    // Allow bundling cjs modules. Rollup doesn't understand cjs
+    commonjs(),
+    // Compile TypeScript/JavaScript files
+    babel({
+      extensions,
+      rootMode: 'upward',
+      include: ['src/**/*'],
+      babelHelpers: 'runtime',
```
BPScott, line 42: All our other packages use `babelHelpers:'bundled'`. If've found that it is much simpler than needing to deal with having a dependency on the runtime plugin.

You're not specifying a `targets` field here so babel is going to transpile EVERYTHING. I bet if you specify `targets: 'node 14.15'`. Then you'll see you'll require basically no babel helpers at all.

polaris-migrator/rollup.config.mjs
```diff
@@ -0,0 +1,52 @@
...
+    }),
+  ],
+  external: [
+    ...Object.keys(pkg.dependencies ?? {}),
+    ...Object.keys(pkg.peerDependencies ?? {}),
+    // https://www.npmjs.com/package/@rollup/plugin-babel#user-content-babelhelpers
+    /@babel\/runtime/,
+    /node_modules/,
```
BPScott, line 50: This seems suprising. Why does this need to be here? Here you're telling this "don't bundle things that are imported from node_modules" - but anything you import should be covered by the above dependencies and peerDependencies.

If you're importing some code from node_modules but it's not in your dependencies/peerDependencies then I'd expect that to potentially fail at runtime

polaris-migrator/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "lint": "TIMING=1 eslint --cache .",
+    "test": "jest",
+    "clean": "rm -rf .turbo node_modules dist *.tsbuildinfo"
+  },
+  "dependencies": {
+    "@babel/core": "^7.18.9",
+    "@babel/plugin-transform-runtime": "^7.18.9",
+    "@babel/runtime": "^7.18.9",
+    "@babel/helper-plugin-utils": "^7.18.9",
+    "@shopify/babel-preset": "^24.1.5",
```
BPScott, line 31: This shouldn't be needed at runtime. It looks like you're missing a [`files` key](https://github.com/Shopify/polaris/blob/main/polaris-react/package.json#L17) to prevent shipping your `.babelrc` as part of the package.

polaris-migrator/src/cli.ts
```diff
@@ -0,0 +1,111 @@
+import * as fs from 'node:fs';
```
BPScott, line 1: I'm guessing this is riffing on [https://github.com/reactjs/react-codemod/blob/master/bin/cli.js](react-codemod) and eventually it'll have have a nice UI to let people select the codemods they want

polaris-migrator/src/migrations/replace-sass-spacing/index.ts
```diff
@@ -0,0 +1,30 @@
...
+  'spacing(base)': 'var(--p-space-4)',
+  'spacing(loose)': 'var(--p-space-5)',
+  'spacing(extra-loose)': 'var(--p-space-8)',
+};
+
+const plugin = (): Plugin => ({
+  postcssPlugin: 'ReplaceSassSpacing',
+  Declaration(decl) {
+    decl.value = decl.value.replace(
+      createRegexFromMap(spacingMap),
```
BPScott, line 20: I'm not a fan of the usage of regex here. It looks like it can match false positives like `my-spacing()`, and miss variations in whitespace/quoting like `spacing( none )`, `spacing('none')` and `spacing( "none" )` (admittedly as we use prettier our whitespace should be pretty conventional so this shouldn't be a huuuge issue for our codebases).

I'd expect us to have a better AST here. There should be AST nodes for stuff like "this is a function with a name and these specific arguments" - you should be able to be able to say "only act when you find a function name, whose name is "spacing" and has these specific values as the first argument". We should leverage a tokeniser, rather than trying to fuzzy match ourselves. It looks like this behaviour of transforming values into "this is a function call" etc is provided by [`postcss-value-parser`](https://www.npmjs.com/package/postcss-value-parser)


stylelint has a demo of using `postcss-value-parser` https://github.com/stylelint/stylelint/blob/main/lib/rules/function-allowed-list/index.js#L36

polaris-migrator/src/migrations/replace-text-component/tests/replace-text-component.input.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import {DisplayText, Heading} from '@shopify/polaris';
+
+export default function App() {
```
BPScott, line 4: Nitpick: Remove the `import/no-default-export` rule exception in eslintrc and make this be `export function`

BPScott (commented):
Heck yeah I love me some codemods!

Dropped some comments around the place. Noteably I think it'd be very valuable to use `postcss-value-parser` to get a better AST for handling transforms of prop values.

sam-b-rose:
Closing in favor of new iteration in #6852

PR #6709, opened 2022-07-21
Support tooltip for Page primary & secondary actions

aminpaks:

Fixes https://github.com/Shopify/core-access/issues/409 

These changes are being introduced because we want to be able to provide context for a user if there is a reason why we disable a Page Action. The above ticket shows an example of this where we want to show a Tooltip so we can tell the user why this button is disabled for them. 

This change will allow us to provide this context to consumers of Polaris that utilize Page actions, it's optional and will not effect existing Page actions.

Adds an option tooltip prop to Page actions.

### How to 🎩

Top-hat instructions available in [this PR](https://github.com/Shopify/web/pull/69430) which is using a snapshot of this Polaris change. https://github.com/Shopify/web/pull/69430

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

Not tophatted on mobile since this change does not effect `Page` actions inside the mobile ... `ActionList`. Not tested for accessibility since it is not adding any new visual elements. 
- 🚫 Tested on [mobile](https://github.com/Shopify/polaris/blob/main/documentation/Tophatting.md#cross-browser-testing)
- 🚫 Tested for [accessibility](https://github.com/Shopify/polaris/blob/main/documentation/Accessibility%20testing.md)


polaris-react/src/components/Page/tests/Page.test.tsx
```diff
@@ -290,6 +292,39 @@ describe('<Page />', () => {
...
+          primaryAction={{content: 'Save', tooltip: saveTooltip}}
+        />,
+      );
+      expect(page).not.toContainReactComponent(
+        Tooltip,
+        expect.objectContaining(saveTooltip),
+      );
+    });
+
+    it('is rendered when available for secondary actions', () => {
```
peterlazzarino: I know this is a WIP so note for later, we should probably include a negative test case as well to ensure its not rendered when there isn't tooltip content.

polaris-react/src/components/Tooltip/utils.tsx
```diff
@@ -0,0 +1,22 @@
...
+import {Tooltip, TooltipProps} from './Tooltip';
+
+export function tooltipFrom(
+  tooltip: NonNullable<ReactNode> | ActionTooltip,
+  button: ReactNode,
+) {
+  const tooltipType = typeof tooltip;
+  const tooltipProps: TooltipProps =
+    tooltipType !== 'string' &&
+    tooltipType !== 'function' &&
```
peterlazzarino: to me this reads as the toolTipType may be a function, is that the case? I can't find anything in the ToolTipProps or typing for tooltips where that's the case.

aminpaks: Yeah, I wish we had `Object.isObject(...)`, I'll adjust it.
For reference the idea came from [here](https://stackoverflow.com/a/13045455)

sarahill:
This looks good on larger screens. Works on hover and focus as expected. 

Couple things on small screens:
- Disabled option is interactive
- Since I can't hover to get a tooltip here, we'll need to think about how that message shows up when it's in a dropdown. You could potentially show the message below Refund when it's in the list... would need to see how that would fit.
 
refund

peterlazzarino:
Thanks for the feedback @sarahill , for those two items, we are planning on following up on a UX pattern for mobile but IIRC the team had concerns around tooltips on mobile (stacking actions) obscuring other items, so the tooltip would have to be custom styled. They wanted to get this shipped while they come up with a solution.

As for the interact-able button, I looked at what this is like now on the Page component and it seems this is currently how that button behaves for small screen disabled actions, so not a regression from this PR. I put up a code sandbox using Polaris 9.2.1 to confirm if you want to take a look. https://codesandbox.io/s/tmg9qm?module=App.js&file=/App.js . Even though in this case the button is properly disabled, there are some Polaris styles on hover and active that make it appear interact-able.

chloerice (approved):

peterlazzarino:
Snapped and 🎩  on `web`, looking good!

peterlazzarino (approved):

Issue #6288, opened 2022-06-22
[Polaris website reboot] Update homepage copy


selenehinkley:
The homepage could use a more in-depth content design lens. For the launch, here's some quick-ish improvements:

**More than a UI kit.**
Polaris unifies 1,000+ people designing and building merchant experiences, so that our product feels like it’s made by one person.

 **Explore the foundations**
Our design foundations provide fundamental design elements and guidance for creating good merchant experiences.

**Components**
Components are reusable building blocks made of interface elements and styles. Piece them together, improve them, and create new ones when creating solutions for merchant problems.

**Tokens**
Tokens are repeatable design elements that can be combined creatively. Use them to unite our design decisions across merchant experiences.

**Icons**
The Polaris icon library has 400+ carefully designed icons for commerce and entrepreneurship. Use them as visual aids to help merchants complete their tasks.

FYI @yesenia-perezcruz 👀

alex-page:
I love the components, foundations and icon adjustments you suggested @selenehinkley. I made a [PR](https://github.com/Shopify/polaris/pull/6292/files) with all of your suggestions but want to chat about some of the content.

> **More than a UI kit.**
> Polaris unifies 1,000+ people designing and building merchant experiences, so that our product feels like it’s designed by one person.

This is the first thing users will read coming to our site. Referencing a "UI Kit" then saying we are more than that feels a bit odd to me. I would expect this title and sentence to explain our unique value that we add to Shopify and it's builders. Saying "More than a UI Kit" to me doesn't get that across. I am also not a fan of putting a number on the total people building merchant experiences. That should be growing rapidly and honestly 1000 sounds quite low for a design system at Shopify scale.

> Tokens
> Tokens are repeatable design elements that can be combined creatively. Use them to unite our design decisions across merchant experiences.

I like the "unite our design decisions across merchant experiences" but the "Use them" kind of throws me off. Who uses the tokens? Are we telling people to use them?

selenehinkley:
Great feedback! 

For the intro, let's leave it as is and we can always revisit down the road. I see that Marten updated the subcopy to the shorter version already so that's good.

For the Tokens piece, let's update it to either of these:

Option 1:
> Tokens
> Tokens are repeatable design elements that can be combined creatively. They unite our design decisions across merchant experiences.

Option 2:
> Tokens
> Tokens are repeatable design elements that can be combined creatively. They’re built-in to unite our design decisions across merchant experiences.

[...]

PR #6733, opened 2022-07-22
[polaris.shopify.com] Update home page copy

martenbjork:
See #6288.

> The HomePage remains a confused mess that we will redesign, but for now, we should ship this if it helps our messaging

image


dominikwilkowski (approved):
LGTM

martenbjork:
Shipping! Let's keep on iterating! 🚀

PR #6736, opened 2022-07-22
[polari.shopify.com] Fix issue that caused site search to re-open

### LINKED ISSUE #6638: [polaris.shopify.com] Search box reappears after typing `Enter` to select a result
Reproduction Steps:
1. Open the search pop modal
2. Type a search term that returns at least one result.
3. Select a result with the keyboard and type `Enter` to select it.

Expected Results:
The modal disappears and the user is navigated to the result page.

Actual Results:
The modal disappears **briefly** and the user is navigated to the result page, but the Search modal immediately **reappears**.

Notes:
- Occurs 100% of the time.
- Tested on Chrome and Firefox.

martenbjork:
Changes the event handler for the global search. It fixes a pesky focus bug, but makes the search feel a tiny bit less snappy which we can address later.

> OK, I've narrow this one down to this:
> 
> - When you press enter, the keydown event closes the modal and returns focus to the previously focused element (a11y best practice)
> - The onkeyup event immediately fires on that element causing the modal to re-open
> 
> You can prove this by manually setting the focus on the dark mode switcher, then opening the global search using the / key. When you press enter on a result, the search will disappear (onkeydown), focus will return to the dark mode switcher, and the site will switch to light/dark mode (onkeyup).

Fixes #6638

### DIFF
--- a/.changeset/chatty-seahorses-raise.md
+++ b/.changeset/chatty-seahorses-raise.md
@@ -1,0 +1,5 @@
+---
+'polaris.shopify.com': patch
+---
+
+Fixes an issue which caused the site search to re-open once dismissed.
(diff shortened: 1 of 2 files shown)


SeanyB (approved):
🎩 looks good! I see the problem no longer occurs when focus is on the search box.

PR #6743, opened 2022-07-22
Support React 18

### LINKED ISSUE #5477: In polaris-react, bump react version in peerDependencies to 18.0.0
## Feature request summary

In polaris-react, bump react version in peerDependencies to 18.0.0.

## Rationale

When generating Next.js project, react version by default is 18.0.0, while in peerDependencies it's "^16.14.0 || ^17.0.0" now.  
So right now, new Next.js projects can't install & use polaris-react from the start.

ryanwilsonperkin:

Fixes #5477 

Adds explicit support for React 18 in the component library, website, and figma plugin. Fixes bugs and type errors that result.

There are a few notable changes that we make in order to work with new behaviours in React 18:
1. The `AfterInitialMount` helper component gets a new prop `onMount` callback that will be invoked after the initial mount. This lets us avoid a timing issues in IndexTable that happens based on new mounting & effect processing behaviour that would otherwise cause a callback to invoked before a ref was actually available in the DOM.
2. An updated version of `@shopify/react-testing` as well as proper use of the `act` method throughout our tests. `act` was previously more lenient but gets stricter in React 18 about exactly how and when it should be invoked.
3. Removed unnecessary use of `resetModules` in some of our tests which was interfering with a global behaviour of the `@shopify/react-testing` library
4. Use a ref instead of state in the Konami code easter egg component. In React 18, state changes are now batched in more places, meaning that this handler was being invoked with stale references and was unable to track state changes appropriately. A ref works better in this scenario because we're trying to use a mutable store of value to ensure we're not relying on a stale state reference.
5. Fix type checking issues, most notably that `useCallback` functions now require parameter signatures and functional components no longer implicitly receive a `children` prop, and need to declare it explicitly now.

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist


.changeset/kind-bears-pretend.md
```diff
@@ -0,0 +1,7 @@
+---
+'polaris-for-figma': patch
+'@shopify/polaris': patch
```
kyledurand:
```suggestion
'@shopify/polaris': minor
```

> MINOR version when you add functionality in a backwards compatible manner

I think minor might make more sense here since it is an enhancement

polaris-react/src/components/KonamiCode/KonamiCode.tsx
```diff
@@ -21,21 +21,21 @@ export const KONAMI_CODE = [
...
         handler(event);
-        setPosition(0);
+        position.current = 0;
       } else {
-        setPosition(position + 1);
+        position.current++;
       }
     } else {
-      setPosition(0);
+      position.current = 0;
```
kyledurand, line 38: These changes look fine to me. @tmlayton can you double check nothing breaks here.

Should we remove this component at some point? I'm not sure it's used anywhere internally anymore

polaris-react/src/components/Page/components/Header/Header.tsx
```diff
@@ -249,6 +249,8 @@ function PrimaryActionMarkup({
     ) : (
       content
     );
+  } else {
+    actionMarkup = primaryAction;
```
kyledurand, line 253: Why was this needed?

polaris-react/src/components/Tabs/components/TabMeasurer/tests/TabMeasurer.test.tsx
```diff
@@ -47,9 +47,7 @@ describe('<TabMeasurer />', () => {
   it('calls setTimeout in development to delay measurements', () => {
     process.env.NODE_ENV = 'development';
     mountWithApp(<TabMeasurer {...mockProps} tabToFocus={0} />);
-    // The second invocation is from the react scheduler
-    // https://github.com/facebook/react/blob/0cf22a56a18790ef34c71bef14f64695c0498619/packages/scheduler/src/forks/SchedulerHostConfig.default.js#L53
-    expect(setTimeoutSpy).toHaveBeenCalledTimes(2);
+    expect(setTimeoutSpy).toHaveBeenCalled();
```
kyledurand, line 50: Did this result change somehow? Seems like we should still be using `toHaveBeenCalledTimes(2)`

kyledurand (approved):
This looks great @ryanwilsonperkin thank you for taking this on. I think next you'll want to `/snapit` for testing in web. Let me know if you want a hand with that

polaris-react/src/components/Page/components/Header/Header.tsx
ryanwilsonperkin: `actionMarkup` will be used below and must be a `ReactNode`, but the value that gets passed in from `primaryAction` can be either a `ReactNode` or an "interface" object with some properties on it. So what's happening here is that we're initially assigning it to `T | ReactNode` and then in the `isInterface` condition we're refining that to just `ReactNode`, however for some reason the type assertion of the `isInterface` utility isn't working correctly so lower down when we try to use this as a `ReactNode` TypeScript still believes that it's either `T | ReactNode`.

React itself used to be more permissive about this, but is stricter in v18 and flags this as a potential error. By telling the compiler that it can only ever be a `ReactNode` and then assigning it as such in the two different conditions, we make the compiler happy, and don't change the functionality of the code.

polaris-react/src/components/Tabs/components/TabMeasurer/tests/TabMeasurer.test.tsx
ryanwilsonperkin: No, the scheduler no longer has the second call in this situation, plus it felt more appropriate for the spirit of the test to simply assert that the set timeout spy had been called, not to specify how _many_ times it had been called

polaris-react/src/components/KonamiCode/KonamiCode.tsx
tmlayton: Yes this component should be removed. I think this change is fine for now.

BPScott:
Rebased atop latest main. Only conflicts was some small ordering adjusts in package.json files

polaris-react/src/components/KonamiCode/KonamiCode.tsx
tmlayton: FYI this has been removed targeting v11 https://github.com/Shopify/polaris/pull/7349

chloerice (approved):
Thanks for tackling this upgrade @ryanwilsonperkin!! This fixes some bugs affecting the admin, including the [IndexTable scrollbar bug](https://github.com/Shopify/polaris/issues/7307) 🎉   ([Spinstance](https://admin.web.polaris-react18-support.chloe-rice.us.spin.dev/store/shop1/orders?inContextTimeframe=none))

BPScott:
@ryanwilsonperkin is vacationing (since Sept 27th, calendars suggest he'll be back on Tuesday). If the polaris crew is happy then I'm confident that this is good to merge without waiting for him to return and do the what-have-i-missed catchup.

PR #6757, opened 2022-07-25
[polaris.shopify.com] Improve props table

### LINKED ISSUE #6371: [polaris.shopify.com] Props table is missing for components
Just browsing the new component documentation - the props table is not there (or is hidden). Where'd it go? That was helpful to see the full capabilities or constraints of the component.

### LINKED ISSUE #6529: [polaris.shopify.com] Feedback (general)
Searching for components is significantly more difficult, especially for people new to the library. The sidebar is ordered alphabetically only, whereas the previous site had categories. This would be fine except the search results are limited to three components. For example, if I search for 'list', I don't get all the list components, just the first three. I also can't find all list components quickly in the sidebar as they don't appear together because they start with a different letter. Categories help me quickly identify which components I have at my disposal for different situations. It's a shame this has been removed. Maybe a visible tag system for reference on the component page to be used as a search term and more visible results (horizontal scroll) might fix this?

Where have the props disappeared to? They're either removed or hidden away. They're essential to getting an 'at a glance' idea of how components are used without importing them and running through the type definition.

Missing information regarding the Avatar component on how the colours change (this is an issue on the legacy site too), I believe this may also be a problem with a number of other components, such as the IndexTable.Cell and the `data-primary-link` prop for handling links.

Hope this helps. Great work!

### LINKED ISSUE #6599: [polaris.shopify.com] Nested types missing from props table
Could you tell me what have in complexAction

### LINKED ISSUE #6779: [polaris.shopify.com] Props table feedback - Missing props
Props are missing on the list. Currently only `children``fullWidth`, `narrowWidth` and `divider` are listed, but the `Page` component accepts more than that like `primaryAction`, `title`, `secondaryActions` and others:

Screenshot 2022-07-27 at 16 45 05

martenbjork:
This PR adds a powerful new props table for every component.

🚀 Closes #6371 🚀 Closes #6529 🚀 Closes #6599 🚀 Closes #6779 🚀

ezgif-3-18af949102

### What's in the PR?
- A script that parses `polaris-react` and outputs `props.json`
- Tests for that script
- New front end logic for displaying the props

### Jest mess
The AST parser is complex and needs pretty thorough tests, so I added Jest. This introduces a bit of a...situation:

- Jest supports Typescript, if you add a `babel.config.json` file.
- Next.js version 12 (latest) introduces a new Rust-based compiler (SWC). If users want to opt out, they do this by...adding a `babel.config.json` file.

So, these two things collide. I want to customize Jest, but Next.js thinks I want to disable the Rust compiler. Sigh. To get around this, I placed the Jest tests and config in the `/scripts` dir. They live in their own little island that the website (and Next.js) doesn't know about.

Later on, when the Rust compiler can't be opted out of, we should move this to the website root.

## Checklist

- [x] Output default values
- [x] Highlight deprecated types ([DataTableProps.hasFixedFirstColumn](https://polaris.shopify.com/components/data-table))
- [x] Highlight optional types
- [x] Dump type definition "as is" when it can't be understood
- [x] Handle types
- [x] Handle interfaces
- [x] Handle enums
- [x] Handle properties
- [x] Handle methods
- [x] Handle index signatures like`[key: string]: boolean;` 
- [x] Inline simple references like `Alignment` (= `'top' | 'center' | 'bottom'`)
- [x] Expandable definitions for referenced types
  - [x] One reference
  - [x] Unions of interfaces (`InterfaceA | Interface B`)
  - [x] Intersections of interfaces (`InterfaceA & Interface B`)
  - [x] Methods with interfaces (`(a: string, b: InterfaceA) => void`)
  - [x] Arrays of references (`Interface[]`)
- [x] Use semantic HTML elements
- [x] Add syntax highlighting for types
- [x] Add accessible syntax highlighting theme and use for prism code blocks as well
- [x] Align syntax highlighting colors with decorative colors in tokens
- [x] Gracefully handle components that define props in a way that we can't parse ([Badge](https://polaris.shopify.com/components/badge))
- [x] Is responsive

## Potential follow-ups:

- [ ] Link to MDN for native types (`Range`)
- [ ] Handle components that accept no props
- [ ] Show links to Github
- [ ] Add copyable typescript code for types
- [ ] Support types that were imported from other polaris packages (`ColorScheme` from `polaris-tokens`)


polaris.shopify.com/tsconfig.json
```diff
@@ -20,6 +20,17 @@
...
+    "pages",
+    ".cache/*.json",
+    "**/*.json",
+    "next-env.d.ts",
+    "scripts/get-props/src/get-props.ts"
+  ],
+  "exclude": [
+    "node_modules",
+    "pages/examples/**/*",
+    "scripts/get-props/testData"
```
martenbjork, line 34: This data intentionally contains some invalid Type references, so we don't want to type check it

kyledurand (approved):
Huge W 🙌

PR #6766, opened 2022-07-25
Breakpoints migration guide

lgriffee:
Workspace for working on breakpoints migration guide without running into conflicts with other v10 migration guide work.


documentation/guides/breakpoints-migration.md
```diff
@@ -0,0 +1,792 @@
...
+$p-breakpoints-md-only: '(min-width: 48em) and (max-width: 64.996875em)';
+
+$p-breakpoints-lg-up: '(min-width: 65em)';
+$p-breakpoints-lg-down: '(max-width: 64.996875em)';
+$p-breakpoints-lg-only: '(min-width: 65em) and (max-width: 89.996875em)';
+
+$p-breakpoints-xl-up: '(min-width: 90em)';
+$p-breakpoints-xl-down: '(max-width: 89.996875em)';
+$p-breakpoints-xl-only: '(min-width: 90em)';
+```
```
gui-santos: I added this to be referenced in the examples of the mixins migration, wdyt?

documentation/guides/breakpoints-migration.md
```diff
@@ -0,0 +1,792 @@
...
+  // See https://github.com/sass-mq/sass-mq/issues/6
+  @if $adjustment == -1px {
+    $adjusted-value: -0.01em;
+  } @else if $adjustment == 1px {
+    $adjusted-value: 0.01em;
+  }
+
+  @return em($value) + $adjusted-value;
+}
+```
```
gui-santos: I'm not sure what to do with functions
They don't have a migration since they've been removed and if I just put the definitions here I'm not sure how helpful it is
Also, usually, they depend on other functions that depend on other functions and so on and I don't know if we should dump so much code in the guide. It looks a bit confusing

Maybe we link to the legacy file here in github instead of putting their definitions in the guide. Wdyt?

documentation/guides/breakpoints-migration.md
```diff
@@ -0,0 +1,792 @@
...
+
+```scss
+// Check the Variables section for the value of $not-condensed-content
+
+@mixin page-content-when-partially-condensed() {
+  @include page-content-breakpoint-before($not-condensed-content) {
+    @content;
+  }
+}
+```
```
gui-santos: So I started reviewing the examples and I noticed that they have variables and other functions so I tried to tell people to check the value of the variable at the bottom of the guide

Another thing that I thought was to write the computed values of the variables here, but some variables depend on several functions and it started getting more complicated

Wdyt? Is this helpful?

documentation/guides/breakpoints-migration.md
```diff
@@ -0,0 +1,792 @@
...
+| `$not-condensed-min-page`            | `$not-condensed-content + $not-condensed-outer-spacing`                                       |
+| `$not-condensed-outer-spacing`       | `breakpoint(2 * $layout-width-outer-spacing-max)`                                             |
+| `$partially-condensed-content`       | `em(layout-width(page-content, partially-condensed))`                                         | `28.125em`    |
+| `$partially-condensed-min-page`      | `$partially-condensed-content + $partially-condensed-outer-spacing`                           |
+| `$partially-condensed-outer-spacing` | `breakpoint(2 *$layout-width-outer-spacing-min)`                                              |
+| `$page-max-width`                    | `layout-width(primary, max) + layout-width(secondary, max) +layout-width(inner-spacing)`      | `62.375rem`   |
+| `$stacked-content`                   | `em(layout-width(primary, min) + layout-width(secondary, min) + layout-width(inner-spacing))` | `46em`        |
+| `$typography-condensed`              | `em(640px)`                                                                                   | `40em`        |
```
gui-santos: Added this column with the computed default value of this variables so people can have an idea of how the mixins are using these. Wdyt?

documentation/guides/breakpoints-migration.md
```diff
@@ -0,0 +1,792 @@
+## New breakpoints
+
+TODO: write about the 5 new breakpoints
+
+### Media queries variables
+
+We created some Sass variables to handle the new breakpoints and most of the migration you need to do is replacing old media-queries or mixins with them in your code.
```
aaronccasanova:
```suggestion
We've created a collection of Sass variables for applying responsive styles at each breakpoint alias. A majority of the migration will involve replacing legacy media-queries or mixins with the following values:
```

documentation/guides/breakpoints-migration.md
```diff
@@ -0,0 +1,792 @@
...
+$p-breakpoints-lg-only: '(min-width: 65em) and (max-width: 89.996875em)';
+
+$p-breakpoints-xl-up: '(min-width: 90em)';
+$p-breakpoints-xl-down: '(max-width: 89.996875em)';
+$p-breakpoints-xl-only: '(min-width: 90em)';
+```
+
+## Sass functions and mixins
+
+The following Sass functions and mixins have been removed. If you wish to continue using them you will need to either add them directly to your repo or replace any instances with a value equivalent.
```
aaronccasanova:
```suggestion
The following Sass functions and mixins have been removed. You will need to replace any instances with the closest media query variable (listed above) or hard code the one off media query value you need.
```

Copy/pasting legacy breakpoint functions and mixins will not be a reliable option as the values have been adjusted and would result in unsynchronized responsive behavior.

documentation/guides/breakpoints-migration.md
lgriffee: Aaron and I workshopped this slightly to follow some of the language conventions we've been using throughout the project and in previous migration guides.

documentation/guides/breakpoints-migration.md
lgriffee: Yes! Love that you included these!

documentation/guides/breakpoints-migration.md
lgriffee: Aaron and I updated this text from the v8 to v9 migration guide to really apply to our work. I had just dumped this in from my original PR for the migration guide but never got around to updating it.

documentation/guides/breakpoints-migration.md
```diff
@@ -0,0 +1,792 @@
+## New breakpoints
+
+TODO: write about the 5 new breakpoints
```
lgriffee, line 3: Wanted to drop in some content we've used for comms during the project that could be helpful inspiration for this section! Definitely not something that we could just use "as is" for the section but it has some good info we could rip out. 

> As part of our project to reduce the total number of unique breakpoints in polaris and Shopify’s admin we've created the follow set of breakpoints: 
> 
> * `xs`: 0px
> * `sm`: 490px
> * `md`: 768px
> * `lg`: 1040px
> * `xl`: 1440px
> 
> A script takes these values and generates Sass variables (which can be used in media conditions) for the breakpoint and up, for the breakpoint and down, and for only the breakpoint.
> 
> Example of generated output for `breakpoints-md`:
> ```
> @media #{$p-breakpoints-md-up} {/*...*/}
> @media #{$p-breakpoints-md-down} {/*...*/}
> @media #{$p-breakpoints-md-only} {/*...*/}
> ```

Perhaps we want to also add something about having to use/import Polaris tokens to use these values?

documentation/guides/breakpoints-migration.md
```diff
@@ -0,0 +1,792 @@
...
+$p-breakpoints-xl-only: '(min-width: 90em)';
+```
+
+## Sass functions and mixins
+
+The following Sass functions and mixins have been removed. If you wish to continue using them you will need to either add them directly to your repo or replace any instances with a value equivalent.
+
+### Functions
+
+These functions were removed because the mixins using them were also removed ([see the Mixins section](#mixins)). In case you need to use those mixins you will find the fucntions necessary to make them work them here.
```
lgriffee:
```suggestion

```
This section can be deleted since we are keeping the `breakpoint()` function around.

documentation/guides/breakpoints-migration.md
```diff
@@ -0,0 +1,792 @@
...
+}
+```
+
+</details>
+
+### Mixins
+
+Use `console.log()` to get the function output and hard code the value you need.
+If the end value of is close to one of [Polaris’ tokens](https://polaris.shopify.com/tokens/breakpoints), we recommend replacing the mixin with the proper token (see the following examples).
+Otherwise, if you really need to keep the old mixin, you can copy its definition and use it locally.
```
lgriffee:
```suggestion

```
Since we don't need the function section we can get rid of this one as well.

documentation/guides/breakpoints-migration.md
lgriffee: Also just added a line about using `console.log()` to get the function/mixin output.

documentation/guides/breakpoints-migration.md
lgriffee: 💯 Totally, love the idea of linking to the legacy file instead! Here's something from the v8 to v9 migration guide that I think captures what you're suggesting. I think it's a really slick/clean looking format: 

> #### `available-names()`
> | Deprecated Function | Source                                                                                                                                                          |
> | ------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------- |
> | `available-names()` | [definition](https://github.com/Shopify/polaris/blob/b443d114d447df15d9e72914c8ca5058439a175e/documentation/guides/legacy-polaris-v8-public-api.scss#L354-L386) |

So in the case of `breakpoint()` perhaps it would be: 
### `breakpoint()`
| Deprecated Function | Source                                                                                                                                                          |
| ------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `breakpoint()` | [definition](https://github.com/Shopify/polaris/blob/b443d114d447df15d9e72914c8ca5058439a175e/documentation/guides/legacy-polaris-v8-public-api.scss#L1109-L1122) |

We'd probably want a more up to date commit link for legacy-file but just threw that in there for now. And then if we do that any mixins too we could change the first title to "Deprecated Mixin".

documentation/guides/breakpoints-migration.md
lgriffee:
```suggestion
As part of our project to reduce the total number of unique breakpoints in Polaris and Shopify’s admin we've created the follow set of mobile first breakpoint tokens: 

* `xs`: 0px
* `sm`: 490px
* `md`: 768px
* `lg`: 1040px
* `xl`: 1440px

A transform takes these values and generates Sass variables (which can be used in media conditions) for eahc breakpoint in `up`, `down`, and `only` directions. 

Example of generated output for `breakpoints-md`:

@media #{$p-breakpoints-md-up} {/*...*/}
@media #{$p-breakpoints-md-down} {/*...*/}
@media #{$p-breakpoints-md-only} {/*...*/}
```

We should also add a note on how to access these Sass variables from the [Polaris tokens package](https://github.com/Shopify/polaris/blob/main/polaris-tokens/scripts/toMediaConditions.ts).

documentation/guides/breakpoints-migration.md
lgriffee: Chatting with @aaronccasanova  about this we're thinking that for:

1. The collection of mixins that are static (consuming `page-content-breakpoint-before|after`) we can provided a one to one mapping of the outputted and replacement values. 
2. The dynamic usages of the `page-content-*` mixins we can [port the implementations to JS as a convenience utility](https://github.com/Shopify/polaris/blob/de09047cbe58012b1d095e5b53bd10cf959152ca/breakpoints.mjs) for computing the SCSS output. Similar to [Sam's contribution to the v9 migration guide](https://github.com/Shopify/polaris/blob/main/documentation/guides/migrating-from-v8-to-v9.md#replacing-custom-properties-from-theme-overrides), we could take the JS implementation and [include it in a code sandbox](https://stackblitz.com/edit/legacy-v8-custom-props?file=README.md) with instructions on how to analyze the output and determine the appropriate Polaris breakpoint or recommended media query.

documentation/guides/breakpoints-migration.md
lgriffee: https://github.com/Shopify/polaris/pull/6766/files#r931642265

documentation/guides/breakpoints-migration.md
lgriffee: We may not need this table, Aaron and I will follow up with you about ideas.

documentation/guides/breakpoints-migration.md
gui-santos: Ohhhhhh that's a cool idea!

documentation/guides/breakpoints-migration.md
gui-santos: Yeah sure! Feel free to make suggestions of any kind!

documentation/guides/breakpoints-migration.md
```diff
@@ -0,0 +1,751 @@
...
+}
+```
+
+</details>
+
+### Adding the functions and mixins to your repo
+
+Any functions or mixins that were being consumed from `???` have been removed. If you wish to continue using them you can add them directly to your repo. All of the removed functions and mixins can found in the following file:
+
+[File Name](#)
```
gui-santos: Another thing @lgriffee 

This part was here before, but I didn't know exactly the intention behind it and at the same time it seems important so I didn't remove it hahahaha
But what should I do here? Is the file in question `legacy-polaris-v8-public-api.scss`?

lgriffee: Ah yes this is another instance of something I dumped into my original PR from the v8 to v9 guide but that we don't need for this one (and I never got around to updating). You can totally remove this section! As Aaron said above: 

> Copy/pasting legacy breakpoint functions and mixins will not be a reliable option as the values have been adjusted and would result in unsynchronized responsive behavior.

documentation/guides/breakpoints-migration.md
```diff
@@ -0,0 +1,103 @@
...
+$p-breakpoints-lg-only: '(min-width: 65em) and (max-width: 89.996875em)';
+
+$p-breakpoints-xl-up: '(min-width: 90em)';
+$p-breakpoints-xl-down: '(max-width: 89.996875em)';
+$p-breakpoints-xl-only: '(min-width: 90em)';
+```
+
+## Sass mixins
+
+The following Sass mixins have been removed. You will need to replace any instances with the closest media query variable (listed above) or hard code the one off media query value you need.
```
lgriffee, line 53:
```suggestion
The following Sass mixins have been removed. You will need to replace any instances with the closest media query variable (listed below) or hard code the one off media query value you need.
```

aaronccasanova (approved):

PR #6770, opened 2022-07-26
Bump storybook versions and clean up main.js

alex-page:


polaris-react/.storybook/main.js
```diff
@@ -2,24 +2,13 @@ const path = require('path');
...
 module.exports = {
   core: {
     builder: 'webpack5',
   },
-  // Added to work around https://github.com/storybookjs/storybook/issues/15336
-  // The line below can be removed in @storybook/react v6.5.0
-  typescript: {reactDocgen: false},
   stories: ['../playground/stories.tsx', '../src/components/**/*/README.md'],
-  addons: [
-    {name: '@storybook/addon-essentials', options: {docs: enableDocs}},
```
BPScott, line 20: We'll want to keep `@storybook/addon-essentials`

BPScott (changes requested):
We'll want to keep addon-essentials as it is [used for many useful things] like the toolbars and actions stuff. Add that back.

polaris-react/.storybook/main.js
alex-page: Oh good point. I added back the specific addon we use so we don't get a bunch of extra stuff. I couldn't see us using actions or controls. If we do in the future we should add them back. Let me know what you think.

| Before | After |
| --- | --- |
| Screen Shot 2022-07-27 at 6 32 45 AM | Screen Shot 2022-07-27 at 6 46 02 AM |

alex-page: Oh crap `React.strictMode` is gone. I remember now. Adding this back.

BPScott: I think controls and viewport and measure would be useful too, and at that point I'd be happier including the kitchen sink of addon-essentials rather than opting in individually.

PR #6771, opened 2022-07-26
Padded layout for components in storybook by default

alex-page:


polaris-react/.storybook/preview-body.html
```diff
@@ -1,5 +0,0 @@
-<style>
-  .sb-show-main.sb-main-padded {
-    padding: 0;
-  }
-</style>
```
alex-page, line 5: @kyledurand this felt like something that could be cleaned up. Let me know what you think about this PR. The other option is to make all components use `layout: fullscreen` by default and we could remove this and keep everything as is.

polaris-react/.storybook/GridOverlay/GridOverlay.tsx
```diff
@@ -28,6 +35,7 @@ export function GridOverlay({inFrame, maxWidth, layer, children}: Props) {
   const className = classNames('GridOverlay', inFrame && 'inFrame');
   const style = {
     maxWidth,
+    margin: componentLayout === 'padded' ? '0 1rem' : '',
```
BPScott: What if rather than a having to a prop on the component to control the margin, the styles set a margin detect based on the presence of the `sb-main-padded` / `sb-main-fullscreen` / `sb-main-centered` classes that get set on the body element depending on the layout in use.

```css
.sb-main-padded .GridOverlay { margin: 0 1rem; }
```

 That way you'd only need to configure the layout of the component, rather than the layout and and then this componentLayout prop to make the grid match.

BPScott (commented):
I've not tested this to prove the layout works but I think conceptually it makes sense to keep the default of `padded` layout, and explicitly opt particular stories to be `fullscreen`. Having that extra indent makes it see things like drop shadows on the top edge of an element.

I put a note inline about how we could leverage the classes than the layout prop sets on the body to determine if any extra padding is needed instead of needing the Grid component to have an additional property.

See https://storybook.js.org/docs/react/configure/story-layout for context about the layout parameter

polaris-react/.storybook/polaris-readme-loader.js
```diff
@@ -25,9 +25,14 @@ module.exports = function loader(source) {
...
-  const hasFullscreenLayout = ['App provider', 'Frame', 'Navigation'].includes(
-    readme.name,
-  );
+  const hasFullscreenLayout = [
+    'App provider',
+    'Frame',
+    'Fullscreen bar',
+    'Navigation',
+    'Sheet',
+    'Grid',
```
kyledurand: Does grid need to be full screen now? Might look a little cleaner with padding now. I think ContextualSaveBar should be full screen as well

polaris-react/.storybook/GridOverlay/GridOverlay.tsx
```diff
@@ -15,9 +15,16 @@ interface Props {
   maxWidth?: string;
   layer?: Layer;
   children?: React.ReactNode;
+  componentLayout: 'padded' | 'centered' | 'fullscreen';
```
kyledurand: Do we use centered layouts?

kyledurand (approved):
This looks good but I do like Ben's suggestion of a purely css approach. Might be slightly tough to manage that for centered layouts. Those wouldn't likely work with the grid overlay anyway though

polaris-react/.storybook/GridOverlay/GridOverlay.tsx
alex-page: I don't think so, but I wanted to support them. Looking at @BPScott's above suggestion I could remove this prop all together.

polaris-react/.storybook/polaris-readme-loader.js
alex-page: Definitely not. I will add `ContextualSavebar` and remove the `Grid`.

BPScott:
> Might be slightly tough to manage that for centered layouts. Those wouldn't likely work with the grid overlay anyway though

Yeah I'm fine with punting on centered layouts, and having the grid on them act the same as the fullscreen style grid. We don't have anything that uses the centered layout anyway.

PR #6797, opened 2022-07-28
[Typography foundations] Feat/Update Text component to support Labels

### LINKED ISSUE #6564: [Typography foundations] Add `Label` support to `Text` component
https://github.com/Shopify/polaris/blob/main/polaris-react/src/components/Label/Label.tsx

chazdean:

Fixes #6564  

Changes the default behaviour of the `Text` component allowing span elements to have an `inline` layout and not default to `block`. This change prevents issues when `Text` is a child of `Label`

### How to 🎩

<details>

```jsx
import React from 'react';

import {Label, Text, Page} from '../src';

  return (
    <Page title="Playground">
      <Label id="test" requiredIndicator>
        <Text as="span" variant="headingSm">
          This is a Label
        </Text>
      </Label>
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/.changeset/short-experts-flow.md
+++ b/.changeset/short-experts-flow.md
@@ -1,5 +1,0 @@
----
-'@shopify/polaris': patch
----
-
-Update Text componant to work with Labels
--- a/.changeset/short-experts-flow.md
+++ b/.changeset/short-experts-flow.md
@@ -1,0 +1,5 @@
+---
+'@shopify/polaris': patch
+---
+
+Update Text componant to work with Labels
--- a/polaris-react/src/components/Text/Text.tsx
+++ b/polaris-react/src/components/Text/Text.tsx
@@ -72,7 +72,6 @@ export const Text = ({
     styles.root,
     styles[variant],
     fontWeight ? styles[fontWeight] : styles[VariantFontWeightMapping[variant]],
-    (alignment || truncate) && as !== 'span' && styles.block,
     alignment && styles[alignment],
     color && styles[color],
     truncate && styles.truncate,
--- a/polaris-react/src/components/Text/Text.tsx
+++ b/polaris-react/src/components/Text/Text.tsx
@@ -72,7 +72,7 @@ export const Text = ({
     styles.root,
     styles[variant],
     fontWeight ? styles[fontWeight] : styles[VariantFontWeightMapping[variant]],
-    (alignment || truncate) && styles.block,
+    (alignment || truncate) && as !== 'span' && styles.block,
     alignment && styles[alignment],
     color && styles[color],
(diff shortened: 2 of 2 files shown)


polaris-react/src/components/Text/Text.tsx
```diff
@@ -72,7 +72,7 @@ export const Text = ({
     styles.root,
     styles[variant],
     fontWeight ? styles[fontWeight] : styles[VariantFontWeightMapping[variant]],
-    (alignment || truncate) && styles.block,
+    (alignment || truncate) && as !== 'span' && styles.block,
```
alex-page: My gut feeling is we shouldn't set an alignment ever. `h1` is `block` and `span` is `inline` by default unless there is a reset doing some magic.

aveline: Fair point. So if someone set alignment on a `span` then the expected behaviour would be nothing.

Making alignment switch the element to block was based on a previous iteration when all the `Display` elements were `span`, but that's no longer the case so we should be good to remove that.

chazdean: Thanks @aveline your explanation on why we needed this functionality previously is helpful. In that case I would also agree that we can remove it 👍🏾

alex-page (approved):
Nice @chazdean. Just wanted to confirm this approach works for `<Label>`?

.changeset/short-experts-flow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/polaris': patch
+---
+
+Update Text componant to work with Labels
```
laurkim: Since this is going into the `typography-foundations-text-component` which already has a changeset file, you can delete this file or just update the existing changeset directly

chazdean: For sure!

chazdean:
> Nice @chazdean. Just wanted to confirm this approach works for `<Label>`?

Yup, after using it in Playground it seems to work nicely

.changeset/short-experts-flow.md
alex-page: Yep add the `skip-changelog` label to pass the check

aveline (approved):
My only thought would be that we should add some documentation somewhere that `align` and `truncate` will only take effect on block level elements. I can see it causing some confusion when the prop is set but nothing happens on inline elements.

chazdean:
> 

Right now I'm thinking the best place for this would be on the Text component example page under Best Practices? I noticed that the Data table component has a section for Alignment under Best practices

laurkim (approved):
💯

PR #6804, opened 2022-07-28
Update `Text` prop descriptions

aveline:
Update prop descriptions for `truncate` and `alignment` to indicate they work on block level elements only as per https://github.com/Shopify/polaris/pull/6797


polaris-react/src/components/Text/Text.tsx
```diff
@@ -38,7 +38,7 @@ const VariantFontWeightMapping: {[V in Variant]: FontWeight} = {
 };
 
 export interface TextProps {
-  /** Adjust horizontal alignment of text */
+  /** Adjust horizontal alignment of text when `as` prop is set to a block element */
```
laurkim, line 41:
```suggestion
  /** Adjust horizontal alignment of text for block elements */
```

My suggestion might be too brief? What do you think @aveline

laurkim (approved):

polaris-react/src/components/Text/Text.tsx
```diff
@@ -38,7 +38,7 @@ const VariantFontWeightMapping: {[V in Variant]: FontWeight} = {
...
   alignment?: Alignment;
```
alex-page: This is interesting as alignment doesn't work when an element is inline. e.g. `display: inline; text-align: center;` would do nothing. Maybe we don't document it and see if there is any confusion?

polaris-react/src/components/Text/Text.tsx
```diff
@@ -48,7 +48,7 @@ export interface TextProps {
   color?: Color;
   /** Adjust weight of text */
   fontWeight?: FontWeight;
-  /** Truncate text overflow with ellipsis */
+  /** Truncate text overflow with ellipsis when `as` prop is set to a block element */
```
alex-page: Taking a look at this implementation it makes sense to me. I wonder if truncate should just apply `display: block`
https://github.com/seek-oss/braid-design-system/blob/master/lib/components/private/Truncate/Truncate.tsx

aveline: Yes the intent behind https://github.com/Shopify/polaris/pull/6797/files#diff-8ee4c2ce9fa2a15a8575923c63afff900480c13e0a001dc1007929d1fedf67e7L75 was that truncate would automatically also apply block. Maybe removing it was conflating that with alignment? We don't want alignment to apply block, but truncate is ok?

polaris-react/src/components/Text/Text.tsx
aveline: I like brevity! But I think a clear link to the `as` prop is worth calling out?

polaris-react/src/components/Text/Text.tsx
chazdean: My only question here before we change it back would be what are the cons of allowing Text to use truncate with an inline element as opposed to block?

alex-page: I feel like `truncate` and `align` should apply `block` as they don't work without it.

aveline: > what are the cons of allowing Text to use truncate with an inline element

It could create confusion when truncate is set, but nothing happens

> I feel like truncate and align should apply block as they don't work without it.

That was our original intent before the discussion to decouple them in #6797 😂

Shall I close this PR then?

alex-page: Oh I see. I am so confusing sometimes. I think the intent in 6797 is right. If a developer wants to use `align` or `truncate` it should apply block. HOWEVER it shouldn't apply block by default. Only when those properties are used. The issue before was that `alignment = 'inherit'` was setting a default value and causing everything to be `block` by default. I don't think that should be the case. We can close this PR but should...

1. Set no default value to alignment, or only set display block if alignment !== `inherit`
2. Keep the block style for when alignment or truncate is set
3. This should allow `span` without alignment or truncate to be inline by default

Hope this makes sense @aveline

PR #6811, opened 2022-07-29
Stop page width update from scrollable pages

alex-page:
Video shows this PR first on localhost then the current site with tabs resizing

https://user-images.githubusercontent.com/19199063/181815496-df2eb778-1350-4486-9bf2-85e21d5fe44c.mov

### DIFF
--- a/polaris.shopify.com/src/styles/globals.scss
+++ b/polaris.shopify.com/src/styles/globals.scss
@@ -135,6 +135,7 @@ html {
 
 html {
   font-size: 106.25%;
+  overflow-y: overlay;
 
   @media screen and (max-width: $breakpointMobile) {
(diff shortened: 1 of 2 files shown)


lgriffee (approved):

polaris.shopify.com/src/styles/globals.scss
```diff
@@ -135,6 +135,7 @@ html {
 
 html {
   font-size: 106.25%;
+  overflow-y: overlay;
```
BPScott, line 138: https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-y suggests `overlay` isn't a valid value. Did you mean `scroll`?

PR #6815, opened 2022-08-01
[AccountConnection] Change action type to LoadableAction

### LINKED ISSUE #6814: Connection action should be Loadable
## Issue summary

`AccountConnectionProps` has `action?: Action`, which does not include a `loading` prop.

## Expected behavior

I think this should be `LoadableAction`. We can pass `action.loading: true` to display a spinner while the user is being connected; I think the types should indicate this.

## Actual behavior

```
Object literal may only specify known properties, and 'loading' does not exist in type 'Action'. ts(2322)
```

## Reduced test case

https://codesandbox.io/s/shopify-polaris-react-typescript-playground-forked-k157r1?file=/src/index.tsx

aaronadamsCA:

Fixes #6814 by adding type-safe support for the `loadable` prop on `AccountConnectionProps["action"]`.

This is important because clicking "connect" should trigger an external API request, and if this takes time to complete, the user should be informed. Setting `loading: true` disables the button and gives visual feedback.

(The best action type would also support `disabled`, so that the button could be disabled for 1 second before switching to a loading state; this PR keeps it simple for now, to avoid introducing any new types.)

It's just adding type safety for an existing capability:

image

### DIFF
--- a/polaris-react/src/components/AccountConnection/AccountConnection.tsx
+++ b/polaris-react/src/components/AccountConnection/AccountConnection.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import type {Action} from '../../types';
+import type {LoadableAction} from '../../types';
 import {Avatar} from '../Avatar';
 import {buttonFrom} from '../Button';
 import {Card} from '../Card';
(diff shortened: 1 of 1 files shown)


aaronadamsCA:
Closing as there doesn't seem to be any interest in fixing this type.

PR #6824, opened 2022-08-02
[Typography foundations] Update text alignment

chazdean:

- Default alignment of Text component is set to `inherit`
- `block` style is only applied when alignment or truncate is set (ignores default)
- Without setting alignment or truncate, users can use span elements to apply `inline` styling

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page, Text} from '../src';

  return (
    <Page title="Playground">
      <Text as="span" variant="headingSm">
        Hello World inline
      </Text>
      <Text as="h1" variant="headingSm" alignment="start">
        Hello World block
      </Text>
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/polaris-react/src/components/Text/Text.scss
+++ b/polaris-react/src/components/Text/Text.scss
@@ -19,6 +19,10 @@
   @include visually-hidden;
 }
 
+.inherit {
+  text-align: inherit;
+}
+
 .start {
   text-align: start;
 }
--- a/polaris-react/src/components/Text/Text.scss
+++ b/polaris-react/src/components/Text/Text.scss
@@ -19,10 +19,6 @@
   @include visually-hidden;
 }
 
-.inherit {
-  text-align: inherit;
-}
-
 .start {
   text-align: start;
 }
--- a/polaris-react/src/components/Text/Text.tsx
+++ b/polaris-react/src/components/Text/Text.tsx
@@ -18,7 +18,7 @@ type Variant =
   | 'bodyMd'
   | 'bodyLg';
 
-type Alignment = 'inherit' | 'start' | 'center' | 'end' | 'justify';
+type Alignment = 'start' | 'center' | 'end' | 'justify';
 
 type FontWeight = 'regular' | 'medium' | 'semibold' | 'bold';
 
@@ -57,7 +57,7 @@ export interface TextProps {
 }
 
 export const Text = ({
-  alignment = 'inherit',
+  alignment,
   as,
   children,
   color,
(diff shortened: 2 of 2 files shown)


aveline (approved):

polaris-react/src/components/Text/Text.scss
```diff
@@ -19,6 +19,10 @@
   @include visually-hidden;
 }
 
+.inherit {
+  text-align: inherit;
+}
+
```
alex-page: I can't see `styles.inherit` in the `Text.tsx` file 🤔

alex-page (commented):
I would remove the default `alignment = 'inherit'`. Then set the `text-align: inherit` on the `styles.root`. Then we can move the `display: block` into the css for alignment.

```
.center,
.end,
.start {
  display: block;
}
```

aveline:
@alex-page I prefer `display: block;` decoupled from alignment because we also apply `.block` for truncate.

laurkim:
> I would remove the default alignment = 'inherit'. Then set the text-align: inherit on the styles.root.

Looks like `.root` already has `text-align: inherit`
https://github.com/Shopify/polaris/blob/f0a33df20dde7656d32911e8cda5faec5fb6e193/polaris-react/src/components/Text/Text.scss#L5

polaris-react/src/components/Text/Text.scss
chazdean: I added it to the .scss in the event that a user wanted to set alignment as `inherit` manually. `inherit` is listed as an Alignment option currently. Should it be removed and only be accessible as a default when no other alignment is set?

`type Alignment = 'inherit' | 'start' | 'center' | 'end' | 'justify';`

`alignment && styles[alignment]`

alex-page:
@aveline makes sense to me. @laurkim if we already do that maybe we can remove it as an option for alignment? Then the `block` style only gets added when `alignment !== undefined`

polaris-react/src/components/Text/Text.tsx
```diff
@@ -72,6 +72,7 @@ export const Text = ({
     styles.root,
     styles[variant],
     fontWeight ? styles[fontWeight] : styles[VariantFontWeightMapping[variant]],
+    (alignment !== 'inherit' || truncate) && styles.block,
```
laurkim:
```suggestion
    (alignment || truncate) && styles.block,
```

And then you can remove `inherit` as a type from `line 21` and as a default argument on `line 60` since it's in `.root` style.

alex-page (approved):

PR #6836, opened 2022-08-03
Use polaris-icons for /foundations landing page

alex-page:
This PR:

- [x] Replaces the majority of icons with a 1:1 match of `polaris-icons`

Changes had to be made to these icons to use the system
- [x] IA icon changed to a `polaris-icon` of documents
- [x] Internationalisation icon changed to `polaris-icon` globe as the current is minor size only
- [x] Spacing changed to `polaris-icon` that has columns instead of three bars
- [x] Voice and tone adjusted to `polaris-icon` that includes `...`
- [x] Loading changed to `polaris-icon` for refresh
- [x] HomeCard changed to `polaris-icon` for `TextBlock`
- [x] Error messages changed to correct diamond icon for errors

If we feel strongly about icons that were adjusted we should contribute to `polaris-icons` and then use them from the library.


laurkim (approved):
I think `src/components/GlobalSearch/GlobalSearch.tsx` needs to have the type for `foundationsIcons` updated as well on line 25.

`const foundationsIcons: { [title: string]: JSX.Element } = {};` -> `const foundationsIcons: { [title: string]: string } = {};`

chloerice (approved):

alex-page:
@laurkim totally missed that. Going to make a bit of a mess to fix it quickly but the clean up/simplification is coming.

polaris.shopify.com/src/data/foundations.json
```diff
@@ -2,6 +2,7 @@
   {
     "frontMatter": {
       "title": "Accessible and inclusive language",
+      "icon": "AccessibilityMajor",
```
alex-page, line 5: Going to remove this duplication with the nav in a follow up

PR #6852, opened 2022-08-04
Create `@shopify/polaris-migrator` package

### LINKED ISSUE #6943: Create prototype `@shopify/polaris-migrator` package

sam-b-rose:
Closes #6943

Adds the beginnings of the `@shopify/polaris-migrator` package. This package applies code transformations to help update 
Polaris apps and the Polaris codebase.

## Usage

```sh
npx @shopify/polaris-migrator <migration> <path>
```

Run tests for `@shopify/polaris-migrator`

```sh
yarn test --filter=@shopify/polaris-migrator 
```

Original PR branch: #6701

### DIFF
--- a/polaris-migrator/bin/migrator-cli.js
+++ b/polaris-migrator/bin/migrator-cli.js
@@ -1,0 +1,5 @@
+#!/usr/bin/env node
+
+const {run} = require('./../dist/cjs');
+
+run();
--- a/polaris-migrator/src/cli.ts
+++ b/polaris-migrator/src/cli.ts
@@ -1,0 +1,119 @@
+import fs from 'node:fs';
+import path from 'node:path';
+
+// @ts-expect-error TS can't resolve the type of this import
+import * as jscodeshift from 'jscodeshift/src/Runner';
+import chalk from 'chalk';
+import isGitClean from 'is-git-clean';
+import globby from 'globby';
+import meow from 'meow';
+
+const cli = meow({
+  description: 'Code migrations for updating Polaris apps.',
+  help: `
+    Usage
+      $ npx @shopify/polaris-migrator <migration> <path> <...options>
+        migration    One of the choices from https://polaris.shopify.com/docs/advanced-features/migrations
+        path         Files or directory to transform. Can be a glob like src/**.scss
+    Options
+      --force            Bypass Git safety checks and forcibly run migrations
+      --dry              Dry run (no changes are made to files)
+      --print            Print transformed files to your terminal
+    `,
+  flags: {
+    force: {
+      type: 'boolean',
+    },
+    dry: {
+      type: 'boolean',
+    },
+    print: {
+      type: 'boolean',
+    },
+  },
+});
+
+export function checkGitStatus(force?: boolean) {
+  let clean = false;
+  let errorMessage = 'Unable to determine if git directory is clean';
+  try {
+    clean = isGitClean.sync(process.cwd());
+    errorMessage = 'Git directory is not clean';
+  } catch (err: any) {
+    if (err && err.stderr && err.stderr.indexOf('Not a git repository') >= 0) {
+      clean = true;
+    }
+  }
+
+  if (!clean) {
+    /* eslint-disable no-console */
+    if (force) {
+      console.log(`WARNING: ${errorMessage}. Forcibly continuing.`);
+    } else {
+      console.log('Thank you for using @shopify/polaris-migrator!');
+      console.log(
+        chalk.yellow(
+          '\nBut before we continue, please stash or commit your git changes.',
+        ),
+      );
+      console.log(
+        '\nYou may use the --force flag to override this safety check.',
+      );
+      process.exit(1);
+    }
+    /* eslint-enable no-console */
+  }
+}
+
+type Args = [
+  migration: string,
+  path: string,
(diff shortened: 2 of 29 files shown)


polaris-migrator/src/cli.ts
```diff
@@ -0,0 +1,112 @@
+import fs from 'fs';
+import path from 'path';
+
+import chalk from 'chalk';
+import isGitClean from 'is-git-clean';
+import globby from 'globby';
+import meow from 'meow';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
```
BPScott: Using `@ts-expect-error` instead of ts-ignore should keep eslint happy

```suggestion
// @ts-expect-error Some Description of why this throws an error here
```

polaris-migrator/src/migrations/replace-sass-spacing/replace-sass-spacing.ts
```diff
@@ -0,0 +1,31 @@
...
+  'spacing(base)': 'var(--p-space-4)',
+  'spacing(loose)': 'var(--p-space-5)',
+  'spacing(extra-loose)': 'var(--p-space-8)',
+};
+
+const plugin = (): Plugin => ({
+  postcssPlugin: 'ReplaceSassSpacing',
+  Declaration(decl) {
+    decl.value = decl.value.replace(
+      createRegexFromMap(spacingMap),
```
BPScott: Dragging a comment over from the old PR: https://github.com/Shopify/polaris/pull/6701#discussion_r937015021

Use [`postcss-value-parser`](https://www.npmjs.com/package/postcss-value-parser) here to have more AST focused approach to parsing the value instead of string replacements. We should be able to say "only act when you find a function name, whose name is "spacing".

This will avoid false positives like `my-spacing()`, and catch variations in whitespace/quoting like `spacing( none )`, `spacing('none')` and `spacing( "none" )` (admittedly as we use prettier our whitespace should be pretty conventional so this shouldn't be a huuuge issue for our codebases).


stylelint has a demo of using `postcss-value-parser` https://github.com/stylelint/stylelint/blob/main/lib/rules/function-allowed-list/index.js#L36

polaris-migrator/src/utilities/testUtils.ts
```diff
@@ -0,0 +1,32 @@
+/* eslint-disable jest/no-export, jest/valid-title, @typescript-eslint/no-var-requires */
```
BPScott, line 1: Dragging a comment over from the old PR: https://github.com/Shopify/polaris/pull/6701#discussion_r936982008

jscodeshift exposes a `runTest` utility https://github.com/Shopify/web/blob/main/tooling/codemods/tests/remove-autobind.test.js#L4 (though it lacks types, so you'll need to [add those yourself](https://github.com/Shopify/web/blob/main/tooling/typescript/modules.d.ts#L73-L81))

It seems to do the same as everything you're doing here. Is it worth leveraging that already existing utility instead of writing our own?

polaris-migrator/src/cli.ts
sam-b-rose: Ah nice! Great suggestion, fixed 👍

polaris-migrator/src/utilities/testUtils.ts
sam-b-rose: I like the idea of not maintaining our own version of such a similar function. The `runTest` function from `jscodeshift` does make some assumptions about the directory structure that I didn't prefer (ex: needing to use the `__tests__` and `__testfixtures__` conventions).

I think having our own check function allows us to simplify the function arguments and create our own conventions. I like having the single `tests` directory where all tests and fixtures can live for a specific migration.

polaris-migrator/src/migrations/replace-sass-spacing/replace-sass-spacing.ts
sam-b-rose: Great idea, replaced the regex solution with the `postcss-value-parser`. Feels much better 👍

polaris-migrator/src/utilities/testUtils.ts
BPScott: Sounds good to me 👍

polaris-migrator/bin/migrator-cli.js
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env node
+
+const {run} = require('./../dist');
```
alex-page: Can we make this `"type": "module"` and remove require?

BPScott:
I realised an edge case: We would'nt want to trivially transform the cases where it's doing basic sass math: `margin: spacing() + spacing()`. In Sass this would compile and work at runtime; however `var(--spacing-base) + var(--spacing-base)` would compile, but would produce meaningless code that would break at runtime.

polaris-migrator/bin/migrator-cli.js
sam-b-rose: I did give this a try and it lead to multiple other issues so punted for now due to time-to-importance ratio. We can certainly revisit later.

I did add an ESM build for consuming packages like the [Shopify CLI](https://github.com/Shopify/cli)

aaronccasanova (approved):

PR #6858, opened 2022-08-05
OG image api

alex-page:
- [x] Remove the need to run the script and keep images in `/public`
- [x] Create images on request
- [x] Update metadata
- [x] Foundations icons
- [ ] Only generate images for routes that exist on the site


polaris.shopify.com/package.json
```diff
@@ -23,8 +23,10 @@
     "@shopify/polaris-tokens": "^5.5.1",
     "codesandbox": "^2.2.3",
     "fuse.js": "^6.5.3",
+    "gray-matter": "^4.0.3",
```
alex-page, line 26: This library will be used during consolidation work as well.

PR #6843, opened 2022-08-03
Add Working with Polaris page

selenehinkley:
This adds a page to Polaris.Shopify.com in the Contributing section to cover points from a Service Level Agreement but with a more approachable and less legalese angle.

### 🎩
Screen Shot 2022-08-05 at 7 59 53 AM

PR #6860, opened 2022-08-05
Revert "Add Working with Polaris page"

selenehinkley:
Reverts Shopify/polaris#6843


yurm04:
We fixed the issue in a follow up PR so no need to revert

PR #6881, opened 2022-08-09
Replaced glob with globby

alex-page:
- [x] Uses `v11.1.0` so it supports commonjs and esm
- [x] One library for globbing across the monorepo instead of multiple


polaris-for-vscode/package.json
```diff
@@ -47,7 +47,6 @@
     "@types/node": "14.x",
     "@types/vscode": "^1.64.0",
     "@vscode/test-electron": "^2.1.2",
-    "glob": "^7.2.0",
```
laurkim, line 50: I think we can also get rid of `@types/glob` on line 46 since globby has type definitions.

laurkim (approved):

sam-b-rose (changes requested):
Really like this `glob/**/*.swap` 👍 

Only suggestion would be to add `globby` to the individual `project/package.json` instead of the root. This may seem redundant, but it will help us with migration to pnpm and is more accurate given globby is used for the package's source and scripts rather than for the root monorepo's tasks and tooling

sam-b-rose (approved):
`⭐ ⭐ /⭐ .approved`

PR #6883, opened 2022-08-09
[polaris.shopify.com] Remove table specific variables

sarahill:
There are a few CSS variables specific to table styling that we don't need.

This PR removes them and uses existing variables instead.


alex-page (approved):
Thank you!

PR #6890, opened 2022-08-10
Add legacy-scss-api to stylelint test

alex-page:
When looking at `polaris-coverage` there are pages where there should be failures showing up. This didn't work as many values were missing from the deprecated scss API. `color` `spacing` etc.

- [x] Add missing scss functions
- [x] Add missing scss mixins
- [x] Add missing global variables
- [x]  **this is due to string test being an exact match**
  - Investigate why `color(` and `rem(` in screenshot below is not a failure
- [x] Convert tests to regex as it exact matches strings

Screen Shot 2022-08-09 at 6 38 27 PM

[polaris-coverage website bug](https://polaris-coverage.shopifycloud.com/shopify/web/tree/main/app/components/DevUI/components/DevUITabs/components/Performance/Performance.scss)

---

The outcome of this PR is a dramatic change in coverage.

183802267-5d442b46-157e-4a78-bf5f-bbbd4d1ee877


sam-b-rose (approved):
Love this! So much legacy Sass to remove

stylelint-polaris/utils/index.js
```diff
@@ -124,28 +124,6 @@ function testAgainstStringOrRegExp(value, comparison) {
...
-
-  // If so, create a new RegExp from it
-  if (comparisonIsRegex) {
-    const valueMatches = hasCaseInsensitiveFlag
-      ? new RegExp(comparison.slice(1, -2), 'i').test(value)
-      : new RegExp(comparison.slice(1, -1)).test(value);
-
-    return valueMatches ? {match: value, pattern: comparison} : false;
-  }
-
```
alex-page: This seems like duplicate code of the lines above `comparison instanceof RegExp`. Tests fail when I remove it. I am confused. We should probably support one type of regex?! 🤷

stylelint-polaris/utils/index.js
```diff
@@ -124,28 +124,6 @@ function testAgainstStringOrRegExp(value, comparison) {
...
   // Otherwise, it's a string. Do a strict comparison
   return value === comparison ? {match: value, pattern: comparison} : false;
```
alex-page: If it's a `string` it only exact matches and this is why so many tests didn't work in the coverage dashboard. The `plugins/global-disallowed-list/index.test.js` has regexes and the `config/coverage.js` has strings 🤷

alex-page: Should we test with the actual config? I understand why we might not... but it's a risk.

stylelint-polaris/utils/index.js
aaronccasanova: All of these `utils` were copy/pasted from Stylelint: https://github.com/stylelint/stylelint/blob/main/lib/utils/matchesStringOrRegExp.js#L16

This `matchesStringOrRegExp` util in particular is used as the foundation for many built-in Stylelint rules. Thus we copied it verbatim to expose a familiar interface. Happy for it to be reworked to only allow strings, RegExp literals, or instances of the RegExp constructor 👍

alex-page: Awesome context. Thank you @aaronccasanova. Happy to leave as is.

stylelint-polaris/plugins/global-disallowed-list/index.js
```diff
@@ -28,6 +28,35 @@ const {rule} = stylelint.createPlugin(
         );
       }
 
+      // Mixin detection
+      postcssRoot.walkAtRules((atRule) => {
+        const value = atRule.source.input.css.slice(
+          atRule.source.start.offset,
+          atRule.source.end.offset,
+        );
```
aaronccasanova: Thoughts on abstracting this out to a dedicated plugin e.g. `mixin-disallowed-list`?

We can then refine the experience by allowing users to validate simpler string/regex matches (as right now we're working within the constraints of the `global-disallowed-list` configuration).

Take [this example](https://astexplorer.net/#/gist/407137c9c881ddf2750943bae62af165/a76a42203bbdee64805f4d6793adefa42f8c0ecf):
image

image

To match the above with the current implementation:
```js
'stylelint-polaris/global-disallowed-list': [/@include\s+my-mixin/]
```
Whereas a dedicated plugin could just validate the mixin name:
```js
postcssRoot.walkAtRules('include', (atRule) => {
  const mixinName = atRule.params
  const found = primary.filter((test) => {
    return matchesStringOrRegExp(value, test);
  });
  ...
```

And is this case would be a simple string match:
```js
'stylelint-polaris/mixin-disallowed-list': ['my-mixin']
```

PR #6929, opened 2022-08-10
[polaris.shopify.com] add icons api route and delete icon images

yurm04:

Removing the local copies of the icon svg images with a dynamic api route to serve the svg code.

To test run locally and visit the api end point http://localhost:3000/api/icons/v0/HintMajor.svg

Screen Shot 2022-08-10 at 3 51 56 PM


polaris.shopify.com/src/pages/api/icons/v0/[icon].tsx
```diff
@@ -0,0 +1,26 @@
...
+      'Cache-Control',
+      'public, immutable, no-transform, s-maxage=31536000, max-age=31536000',
+    );
+    res.end(iconsvg);
+  } catch (err) {
+    res.statusCode = 500;
+    res.setHeader('Content-Type', 'text/html');
+    res.end('<h1>Internal Error</h1><p>Sorry, there was a problem</p>');
+    console.error(err);
+  }
```
Comment (alex-page) [inline polaris.shopify.com/src/pages/api/icons/v0/[icon].tsx:23]: I think there is a way to redirect to a 404. We should also double check if the `readFileSync` fails it hits this error route (it looks like it should).

polaris.shopify.com/src/pages/api/icons/v0/[icon].tsx
```diff
@@ -0,0 +1,26 @@
+import type {NextApiResponse, NextApiRequest} from 'next';
+import fs from 'fs';
+
+const handler = async (req: NextApiRequest, res: NextApiResponse) => {
+  try {
+    const {icon} = req.query;
+    const iconsvg = fs.readFileSync(
+      `../polaris-icons/dist/svg/${icon}`,
+      'utf-8',
+    );
```
Comment (aaronccasanova) [inline polaris.shopify.com/src/pages/api/icons/v0/[icon].tsx:10]: Is the entire monorepo deployed with the site? Wonder, if we should resolve the svgs from the `node_modules`? e.g. Something like:
```js
    const iconSvg = await fs.promises.readFile(
      path.join(
        require.resolve('@shopify/polaris-icons'),
        `../dist/svg/${icon}`
      ),
      'utf-8',
    );
```

Comment (alex-page) [reply polaris.shopify.com/src/pages/api/icons/v0/[icon].tsx:10]: It is deployed!

Comment (aaronccasanova) [reply polaris.shopify.com/src/pages/api/icons/v0/[icon].tsx:10]: Oh Interesting! I still think it might be safer to pull from the `node_modules` so we're accessing assets from published versions and not whatever is on `polaris/main`. That said, I'm not very familiar with the Polaris site infra or if my suggestion would even by pass the yarn workspace symlinks so I'll trust you two!

Comment (alex-page) [reply polaris.shopify.com/src/pages/api/icons/v0/[icon].tsx:10]: I don't think the current approach is the best approach but we couldn't access the files with imports. We should try the require resolve method ^

alex-page:
Closed with https://github.com/Shopify/polaris/pull/6934

PR #6934, opened 2022-08-11
Use @shopify/polaris-icons instead of /icons and copy-icons

alex-page:
We tried these options.

😢 **Copy icons to `public`**
Have to run a script, have to ignore or commit the files up.

🤷  **Create an API for icons**
Works but has to be versioned. Unsure what happens with new icon releases and making sure the API refreshes it's cache. Doesn't feel like a problem we should maintain as an API.

🤔 **Use `@shopify/polaris-icons` as intended**
Works (this PR) however we don't have a great way to get the buffer of the file. We generate it from a fetch for now.


alex-page:
Going to merge as I am cleaning up asset generation. Happy to address feedback in a follow up PR.

PR #6949, opened 2022-08-12
[IndexTable] Update headings style and add sortable functionality.

mrcthms:

Fixes https://github.com/Shopify/web/issues/70689
Fixes https://github.com/Shopify/web/issues/70690

Our IndexTables are sortable across the admin, but there is no way of showing the current sorted state within an IndexTable currently, which could cause confusion for merchants. We also want to tighten up the appearance of the IndexTable to coincide with our efforts to unify the index filtering and searching experience.

A mockup of the new IndexTable heading UI, showing a more condensed view with a subdued background, and the ability to show the sorting state in a heading cell.

This PR does two things that are related to the IndexTable headings. Firstly, it tightens up the density of the headings by shrinking the vertical padding of a heading, shrinking the font size down to 12px, and also adds in a `--p-surface-subdued` background color with `--p-text-subdued` font color to the heading.

Secondly, it adds the ability to show the sorting state of an IndexTable and controlling the sorting state by interacting with a sortable IndexTable heading. The logic introduced here mirrors the logic recently introduced for the DataTable for how that allows for sorting.

### How to 🎩

- Clone this branch locally.
- Visit the new "With Sortable Headings" story in the IndexTable 
- Confirm that the Name and Location columns are now sortable through the heading
- Confirm that the other columns function as before with no additional sortable functionality

### 🎩 checklist

### DIFF
--- a/polaris-react/src/components/IndexTable/IndexTable.scss
+++ b/polaris-react/src/components/IndexTable/IndexTable.scss
@@ -201,6 +201,36 @@ $loading-panel-height: 53px;
   width: var(--p-space-5);
 }
 
+.TableHeadingSortButton {
+  background: none;
+  padding: 0;
+  border: 0;
+  cursor: pointer;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  font-weight: var(--p-font-weight-medium);
+  color: var(--p-text-subdued);
+  font-size: var(--p-font-size-1);
+
+  &:hover,
+  &:focus {
+    .TableHeadingSortIcon {
+      opacity: 1;
+    }
+  }
+}
+
+.TableHeadingSortIcon {
+  order: 1;
+  opacity: 0;
+  transition: var(--p-ease) var(--p-duration-150) opacity;
+}
+
+.TableHeadingSortIcon-visible {
+  opacity: 1;
+}
+
 .ColumnHeaderCheckboxWrapper {
   display: flex;
 }
--- a/polaris-react/src/components/IndexTable/IndexTable.scss
+++ b/polaris-react/src/components/IndexTable/IndexTable.scss
@@ -12,6 +12,7 @@ $breakpoint-small: 458px;
   --pc-index-table-loading-panel: 37;
   position: relative;
   overflow: hidden;
+  border-radius: inherit;
 }
 
 .LoadingContainer-enter {
@@ -177,7 +178,6 @@ $loading-panel-height: 53px;
 
 .TableHeading {
   background: var(--p-surface-subdued);
-  box-shadow: 0 1px var(--p-divider) inset;
   padding: var(--p-space-2) var(--p-space-4);
   text-align: left;
   font-weight: var(--p-font-weight-medium);
--- a/polaris-react/src/components/IndexTable/IndexTable.scss
+++ b/polaris-react/src/components/IndexTable/IndexTable.scss
@@ -12,7 +12,6 @@ $breakpoint-small: 458px;
   --pc-index-table-loading-panel: 37;
   position: relative;
   overflow: hidden;
-  border-radius: inherit;
 }
 
 .LoadingContainer-enter {
@@ -177,9 +176,13 @@ $loading-panel-height: 53px;
 }
 
 .TableHeading {
-  padding: var(--p-space-4);
+  background: var(--p-surface-subdued);
+  box-shadow: 0 1px var(--p-divider) inset;
+  padding: var(--p-space-2) var(--p-space-4);
   text-align: left;
   font-weight: var(--p-font-weight-medium);
+  color: var(--p-text-subdued);
+  font-size: var(--p-font-size-1);
   white-space: nowrap;
   border: 0;
 }
--- a/polaris-react/src/components/IndexTable/IndexTable.scss
+++ b/polaris-react/src/components/IndexTable/IndexTable.scss
@@ -177,7 +177,6 @@ $loading-panel-height: 53px;
 }
 
 .TableHeading {
-  background: var(--p-surface-subdued);
   padding: var(--p-space-2) var(--p-space-4);
   text-align: left;
   font-weight: var(--p-font-weight-medium);
@@ -185,6 +184,10 @@ $loading-panel-height: 53px;
   font-size: var(--p-font-size-1);
   white-space: nowrap;
   border: 0;
+}
+
+.TableHeading-sortable {
+  background: var(--p-surface-subdued);
 }
 
 .TableHeading-flush {
--- a/polaris-react/src/components/IndexTable/IndexTable.stories.tsx
+++ b/polaris-react/src/components/IndexTable/IndexTable.stories.tsx
@@ -1403,12 +1403,10 @@ export function WithSortableHeadings() {
     setSortIndex(index);
     setSortDirection(direction);
     const newSortedRows = sortRows(rows, index, direction);
-    console.log({newSortedRows});
     setSortedRows(newSortedRows);
   }
 
   function sortRows(localRows, index, direction) {
-    console.log(index, direction);
     return [...localRows].sort((a, b) => {
       const key = index === 0 ? 'name' : 'location';
       if (a[key] < b[key]) {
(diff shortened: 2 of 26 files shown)


polaris-react/src/components/IndexTable/IndexTable.tsx
```diff
@@ -71,6 +95,11 @@ function IndexTableBase({
   sort,
   paginatedSelectAllActionText,
   lastColumnSticky = false,
+  sortable,
+  sortDirection,
+  defaultSortDirection = 'descending',
```
sam-b-rose, line 100: I like the ability to set the initial sorting direction here! Did you mean `defaultSortDirection = 'ascending'` here? That is the default value used in the `IndexTableBaseProps` comment.

polaris-react/src/components/IndexTable/IndexTable.scss
```diff
@@ -12,7 +12,6 @@ $breakpoint-small: 458px;
   --pc-index-table-loading-panel: 37;
   position: relative;
   overflow: hidden;
-  border-radius: inherit;
```
sam-b-rose: I'm a fan of removing this due to it's dependence on the parent style, but looks like it is needed for rounding the corner when the IndexTable is wrapped in a Card.

12-05-s5r0v-1w62u

polaris-react/src/components/IndexTable/IndexTable.scss
```diff
@@ -177,9 +176,13 @@ $loading-panel-height: 53px;
 }
 
 .TableHeading {
-  padding: var(--p-space-4);
+  background: var(--p-surface-subdued);
+  box-shadow: 0 1px var(--p-divider) inset;
```
sam-b-rose: More of a design question, but is the box-shadow top border intended for the heading? The border acts as a divider for the other table items, but here it is adding a one-sided border to the table that seems inconsistent.

12-05-s5r0v-1w62u

polaris-react/src/components/IndexTable/IndexTable.tsx
```diff
@@ -50,6 +59,21 @@ export interface IndexTableBaseProps {
...
+   * The direction to sort the table rows on first click or keypress of a sortable column heading. Defaults to ascending.
+   * @default 'ascending'
+   */
+  defaultSortDirection?: IndexTableSortDirection;
+  /** The current sorting direction. */
+  sortDirection?: IndexTableSortDirection;
+  /**
+   * The index of the heading that the table rows are sorted by.
+   */
+  sortColumnIndex?: number;
```
sam-b-rose, line 74: Curious if this should be handled internally rather than set by the props? Looks like the `DataTable` does an `initialSortColumnIndex` then handles subsequent column index changes internally. Thoughts on this pattern?

polaris-react/src/components/IndexTable/IndexTable.tsx
mrcthms: Oops, this should be `descending` rather than `ascending` everywhere - will change in the prop description.

polaris-react/src/components/IndexTable/IndexTable.scss
mrcthms: @samrose3 yeah, after a bit of consideration, especially with the `border-radius: inherit` being needed, I think the top dividing line as seen on the [Figma](https://www.figma.com/file/bZb9bEHNCEMiYjzO0fhK4J/Index-filters-and-search?node-id=682%3A183096) spec should exist outside of the IndexTable, and be the responsibility of the Filters component above to introduce a bottom border instead. Will remove this and reinstate the `border-radius: inherit`

polaris-react/src/components/IndexTable/IndexTable.tsx
mrcthms: @samrose3 We need to expose this in this way because the logic that does all the sorting of the data has to happen outside of the IndexTable itself, as you can also sort an IndexTable from an external component. You can see here on the [Figma](https://www.figma.com/file/bZb9bEHNCEMiYjzO0fhK4J/Index-filters-and-search?node-id=670%3A176130) that we have a Sort button that lives within the new filtering UI that also controls the sorting state. Internalizing the sorting state within the IndexTable would make it impossible, or certainly impractical, to manage this state.

sam-b-rose (approved):

polaris-react/src/components/IndexTable/IndexTable.stories.tsx
```diff
@@ -1347,6 +1347,122 @@ export function WithAllOfItsElements() {
...
+    plural: 'customers',
+  };
+
+  const rows = sortedRows ?? initialRows;
+
+  const {selectedResources, allResourcesSelected, handleSelectionChange} =
+    useIndexResourceState(rows);
+
+  function handleClickSortHeading(index, direction) {
+    setSortIndex(index);
```
sam-b-rose, line 1403: If we are setting the `sortIndex` only here then passing the `index` from the `onSort` callback directly to our sort function, do we still need to keep `sortIndex` in state?

Similarly, I think the same could be done for the `setDirection`. Matching the `DataTabel` prop interface seems more consistent and less state management on the user:

```tsx
interface IndexTableBaseProps {
 ...

  /**
   * The direction to sort the table rows on first click or keypress of a sortable column heading. Defaults to ascending.
   * @default 'ascending'
   */
  defaultSortDirection?: SortDirection;
  /**
   * The index of the heading that the table rows are initially sorted by. Defaults to the first column.
   * @default 0
   */
  initialSortColumnIndex?: number;
  ...
}

File: polaris-react/src/components/IndexTable/IndexTable.tsx (hunk 1 above)
Comment (sam-b-rose) [reply polaris-react/src/components/IndexTable/IndexTable.tsx:100]: Curious the rational of using `descending` as the default. Could we be consistent with the `DataTable` and use `ascending` here?

**Review by sam-b-rose — COMMENTED** [review]
Sorry for extra questions, @mrcthms! Just want a bit more clarity around the component API needs and rational for using a different pattern than the `DataTable`

File: polaris-react/src/components/IndexTable/IndexTable.tsx (hunk 1 above)
Comment (mrcthms) [reply polaris-react/src/components/IndexTable/IndexTable.tsx:100]: Yeah, given we're allowing the user the ability to change it anyway, no qualms with setting the default here to match the `DataTable`.

File: polaris-react/src/components/IndexTable/IndexTable.stories.tsx (hunk 5 above)
Comment (mrcthms) [reply polaris-react/src/components/IndexTable/IndexTable.stories.tsx:1403]: When using IndexTables in the admin, they are usually paired with some form of filter/sort controls, which means that the sorting state still needs to live outside of the IndexTable itself. Extremely simplified version below:

```tsx
function MyComponent() {
  const [sortIndex, setSortIndex] = useState(0);
  const [sortDirection, setSortDirection] = useState('ascending');

  function handleSort(index, direction) {
    setSortIndex(index)
    setSortDirection(direction)
  }

  return (
    <>
      <MyCustomSortingComponent onSort={handleSort} sortIndex={sortIndex} sortDirection={sortDirection} />

      <IndexTable sortColumnIndex={sortIndex} sortDirection={sortDirection} onSort={handleSort} {...otherProps} />
    </>
  )
}
```

We have access to the new index and direction due to the `onSort` handler providing us the new values, but we need to store them at the comsumer level as the `MyCustomSortingComponent` also needs to know these values.

If we have something like this which would bring the sorting state inside of the IndexTable:

```tsx
function MyInternalComponent() {

  function handleSort(index, direction) {
    console.log(index);
    console.log(direction);
  }

  return (
    <>
      <MyCustomSortingComponent onSort={handleSort} sortIndex="??" sortDirection="??" />

      <IndexTable defaultSortColumnIndex={0} sortDirection="ascending" onSort={handleSort} {...otherProps} />
    </>
  )
}
```

Then we do have access to the new values thanks to the `onSort` function, but given we also need to pass these values to the `MyCustomSortingComponent`, we can't access them without saving them at the consumer level.

Looking through the `web` codebase, in the examples I can find where a DataTable is sortable, there is no additional external components which can sort the data inside the table, so it makes sense to internalise the state in that case. But the IndexTable sorting implementations within `web` differ. Whether down the line we want to homogenise this is probably a broader discussion, and one that I'm sure the Quality Crew would love to be a part of if and when that time comes 😄

sam-b-rose: I see, thank you for the added example and clarity! Makes total sense 👍 this API looks good to me

PR #6961, opened 2022-08-15
[TextField] Fix: multiline TextField focus problem

### LINKED ISSUE #5683: [TextField] Using `multiline` causes `focused` to not focus on the `TextField`
## Issue summary

When attempting to set `focused` on a `TextField` that uses `multiline`, you won't be able to focus on the text field.

## Expected behavior

When setting `focused` on a `TextField` that uses `multiline`, it should function exactly the same as a `TextField` which does not use `multiline`.

## Actual behavior

The `TextField` which has `focused` set to `true` and `multiline` set to some value, will end up with a `TextField` that isn't focused.

### Video of the cases where it works and doesn't work.
https://user-images.githubusercontent.com/73828120/166250129-85b38bc6-10e8-496c-86b6-fb451fbd948c.mov

1. Create a `TextField` with `multiline` set to some value (example: `<TextField multiline={4} .../>`
2. Add a button component that should in turn cause `focused` on the `TextField` to be set to `True`
3. Have a TextField that isn't focused, but has a `focused` state of `true`.

## Reduced test case
The issue can be tested [here](https://codesandbox.io/s/shopify-polaris-multiline-focusing-bug-on-textfield-19odxc) (CodeSandbox).

## Specifications

- Polaris version number: `9.4.0`
- Browser: Chrome 100.0.4896.127
- Device: Macbook Pro (13-inch, 2020)
- Operating System: macOS 12.0.1

Paste the results here:

```bash
  System:
    OS: macOS 12.0.1
    CPU: (8) x64 Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
    Memory: 262.95 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 17.2.0 - /usr/local/bin/node
    Yarn: 1.22.18 - /usr/local/bin/yarn
    npm: 8.1.4 - /usr/local/bin/npm
  Browsers:
    Chrome: 100.0.4896.127
    Edge: 101.0.1210.32
    Firefox: 98.0.2
    Safari: 15.1
  npmPackages:
    @shopify/polaris: ^9.4.0 => 9.4.0
    react: ^17.0.2 => 17.0.2
    react-dom: ^17.0.2 => 17.0.2
```

berkaygure:

Fixes #5683 

The focus property does not work correctly when the TextField component has a multiline property. Because it still gets inputRef even though it is a textarea. In this PR, I added logic to fix this situation.

### 🎩 checklist

### DIFF
--- a/.changeset/silver-papayas-relax.md
+++ b/.changeset/silver-papayas-relax.md
@@ -2,4 +2,4 @@
 '@shopify/polaris': patch
 ---
 
-Fix multiline TextField focus problem
+Fixed `TextField` `focused` prop not working when `multiline` is `true`
--- a/.changeset/silver-papayas-relax.md
+++ b/.changeset/silver-papayas-relax.md
@@ -1,0 +1,5 @@
+---
+'@shopify/polaris': patch
+---
+
+Fix multiline TextField focus problem
--- a/polaris-react/src/components/TextField/TextField.tsx
+++ b/polaris-react/src/components/TextField/TextField.tsx
@@ -242,10 +242,10 @@ export function TextField({
   const spinnerRef = useRef<HTMLDivElement>(null);
 
   useEffect(() => {
-    const input = inputRef.current;
+    const input = multiline ? textAreaRef.current : inputRef.current;
     if (!input || focused === undefined) return;
     focused ? input.focus() : input.blur();
-  }, [focused, verticalContent]);
+  }, [focused, verticalContent, multiline]);
 
   useEffect(() => {
     const input = inputRef.current;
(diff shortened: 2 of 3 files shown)


.changeset/silver-papayas-relax.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/polaris': patch
+---
+
+Fix multiline TextField focus problem
```
chloerice:
```suggestion
Fixed `TextField` `focused` prop not working when `multiline` is `true`
```

berkaygure:
It didn't pass the "Visual Regression Test" and gave a "Missing project token error." Is there anything I can do about that?

chloerice:
> It didn't pass the "Visual Regression Test" and gave a "Missing project token error." Is there anything I can do about that?

Hey @berkaygure, sorry for the confusion! Your changes are fine, this check breaks on open source PRs currently.

chloerice (approved):
Great catch, thanks @berkaygure 🙌🏽

PR #6965, opened 2022-08-15
[FormLayout.Item] do not render if children is undefined

### LINKED ISSUE #6956: Empty FormLayout.Item renders unnecessary padding
## Issue summary
When a null or empty FormLayout.Item is passed to the FormLayout component, the FormLayout.Item is rendered with unnecessary padding.

image

## Expected behavior

The FormLayout component should not be rendering additional padding when an empty or null item is passed.

## Actual behavior

An empty or null FormLayout.Item causes unnecessary padding. However, this only seems to occur when the null FormLayout.Item is followed by a FormLayout.Group, as seen below.

image

1. Within a FormLayout component, include an empty/null FormLayout.Item (ie. empty div)
1. Underneath the empty item, add a FormLayout.Group
2. Inside the FormLayout.Group, add a TextField component

## Proposed Solution
Adding a style which removes the padding from an empty/null FormLayout.Item might fix this issue.

image

## Reduced test case

[CodeSandbox test case](https://codesandbox.io/s/shopify-polaris-react-typescript-playground-forked-qzi7lw?file=/src/index.tsx)

This can also be observed by going to my [spin instance](https://shop9.shopify.domains-vbjg.lazir-pascual.us.spin.dev/admin/settings/domains/confirm?host=test-domain-453.com) and clicking the `Edit address` link. The additional padding can be seen between the City/Province/Postal Code and Phone/Fax TextFields.

kyledurand:
Fixes: https://github.com/Shopify/polaris/issues/6956

Alternatively we could try the css approach i.e. `&:empty {padding: 0}`. I think this is safe enough though

### DIFF
--- a/.changeset/red-radios-drive.md
+++ b/.changeset/red-radios-drive.md
@@ -1,0 +1,5 @@
+---
+'@shopify/polaris': patch
+---
+
+Don't render if children is null / undefined
--- a/polaris-react/src/components/FormLayout/components/Item/Item.tsx
+++ b/polaris-react/src/components/FormLayout/components/Item/Item.tsx
@@ -6,6 +6,6 @@ export interface ItemProps {
   children?: React.ReactNode;
 }
 
-export function Item(props: ItemProps) {
-  return <div className={styles.Item}>{props.children}</div>;
+export function Item({children}: ItemProps) {
+  return children ? <div className={styles.Item}>{children}</div> : null;
 }
--- a/polaris-react/src/components/FormLayout/components/Item/tests/Item.test.tsx
+++ b/polaris-react/src/components/FormLayout/components/Item/tests/Item.test.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import {mountWithApp} from 'tests/utilities';
 
+import styles from '../../../FormLayout.scss';
 import {TextField} from '../../../../TextField';
 import {Item} from '../Item';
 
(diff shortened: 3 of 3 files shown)


sam-b-rose (approved):

aaronccasanova:
> Alternatively we could try the css approach i.e. `&:empty {padding: 0}`

Light suggestion to use the CSS approach. I try to avoid rendering `null` from components as it can lead to confusing layout behavior. Take this contrived example: 
```scss
.container {
  display: flex;
  justify-content: space-between;
}
```

```jsx
<Container>
  <Item />
  <Item />
</Container>
```

The end result would look significantly different if the first `Item` renders `null`

kyledurand:
Good point @aaronccasanova but I wouldn't worry too much about that. Consumers can still wrap this in a `div` if they know the item might be null. Main reason for this approach vs css is that we'd have to potentially cancel out other properties as they get added. Which is slightly harder to maintain

lazirpascual (approved):
Looks good! 💯

aaronccasanova:
@kyledurand Sounds good, I trust you!

I did have another quick thought. Isn't this technically a breaking change?

rgoessl (approved):

kyledurand:
> Isn't this technically a breaking change?

If you get really into the weeds about it, maybe. That shouldn't prevent us from fixing minor bugs like this though. I think if this breaks a layout then it probably wasn't being used correctly

PR #6966, opened 2022-08-15
feat(polaris-react): add support for ID prop on Choice in ChoiceList

### LINKED ISSUE #6959: Allow ID prop for "Choice" in ChoiceList
## Feature request summary

Allow a `Choice` component (radio or checkbox used within `ChoiceList`) to set its own `id` attribute. Currently this is not possible and the ID value is auto-generated. 

## Rationale

`id` prop is valid for individual components like `RadioButton` or `Checkbox` but not when using them within a `ChoiceList`. This would be valuable for testing purposes and/or element tracking, such as with analytics. 

---

🌟 Feature requests that are not yet planned will be closed. We then use the issue’s [:+1: upvotes](https://help.github.com/articles/about-conversations-on-github/) to track and set priorities. See the [contribution guidelines](https://github.com/Shopify/polaris/blob/main/.github/CONTRIBUTING.md) for more information.

LA1CH3:

Fixes #6959  

Allows an author using `ChoiceList` component to supply custom values for each `id` prop on `Choice`. Currently these values are auto-generated and `ChoiceList` does not respect supplying your own values. 

- Adds `id` as an available prop for `Choice` in `ChoiceList`
- Updates unit test reflecting new behavior

I didn't bother updating stories, examples, etc. because this a very minor change and should be expected behavior consistent with related components like `RadioButton` and `Checkbox`. However I am happy to add them if reviewers feel necessary. 

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page} from '../src';

  return (
    <Page title="Playground">
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/.changeset/shiny-dragons-check.md
+++ b/.changeset/shiny-dragons-check.md
@@ -2,4 +2,4 @@
 '@shopify/polaris': minor
 ---
 
-Support ID prop for Choice in ChoiceList
+Added `id` prop for Choice in ChoiceList
--- a/.changeset/shiny-dragons-check.md
+++ b/.changeset/shiny-dragons-check.md
@@ -1,0 +1,5 @@
+---
+'@shopify/polaris': minor
+---
+
+Support ID prop for Choice in ChoiceList
--- a/polaris-react/src/components/ChoiceList/ChoiceList.tsx
+++ b/polaris-react/src/components/ChoiceList/ChoiceList.tsx
@@ -14,6 +14,8 @@ interface Choice {
   value: string;
   /** Label for the choice */
   label: React.ReactNode;
+  /** A unique identifier for the choice */
+  id?: string;
   /** Disable choice */
   disabled?: boolean;
   /** Additional text to aide in use */
@@ -75,6 +77,7 @@ export function ChoiceList({
   const choicesMarkup = choices.map((choice) => {
     const {
       value,
+      id,
       label,
       helpText,
       disabled: choiceDisabled,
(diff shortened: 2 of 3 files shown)


.changeset/shiny-dragons-check.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/polaris': minor
+---
+
+Support ID prop for Choice in ChoiceList
```
kyledurand:
```suggestion
Added `id` prop for Choice in ChoiceList
```

kyledurand (commented):
This looks good but I'm wondering if there's another way to do this without adding the prop. Off the top of my head I'm thinking we could probably handle this from a callback. Have you explored anything like that at all? 

Just want to make sure this is 100% necessary before adding a new prop

LA1CH3:
@kyledurand I'm not sure I follow when you mention using a callback. Could you elaborate a bit?

kyledurand:
I meant it as in if there's an `onChange` or `onClick` you can hook into to trigger your analytics event. I'm not sure this is possible in the way the component is set up just wanted to ask first

LA1CH3:
Ah I understand what you mean. So, in this case, we actually want the `ID` for element tracking and not for a specific event. Specifically, our UX researchers would like to know how many times a specific radio button within a `ChoiceList` is being used. Our analytics service automatically tracks user events, but you need to provide a selector as part of the report to see aggregate data. With the `ID` being autogenerated, we do not have a reliable selector to use for this. There are other selectors we could potentially go with but they are much more brittle and could potentially become invalid/aren't specific enough.

kyledurand (approved):
👍

PR #6975, opened 2022-08-17
Tokens: Add missing alpha values in RGBA colors

jonathaneckmier:
This PR adds in some missing `alpha` values to a couple of our polaris token colors. All tokens are currently using the `rbga` color format, and there are a couple that didn't define an `alpha` value. In a browser, a value of `1` will be assumed, so it's not entirely necessary. However, in other implementations like React Native, the alpha value is required or the color will be ignored:

image

### How to 🎩

This is a very simple update that should have no noticeable impact where these color values are used.

### DIFF
--- a/.changeset/lovely-needles-mix.md
+++ b/.changeset/lovely-needles-mix.md
@@ -1,0 +1,5 @@
+---
+'@shopify/polaris-tokens': patch
+---
+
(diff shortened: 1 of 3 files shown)


ghost:

alex-page (approved):

PR #6976, opened 2022-08-17
[DataTable] Add `fixedFirstColumns` prop

tylernoseworthy:

We are currently using the `DataTable`, but there was a requirement in our current project for the first 2 columns within the table to be fixed.  

This PR adds a `fixFirstColumns` prop in favour of the `hasFixedFirstColumn` prop, which now has a deprecation warning if still being used in development. In the event that `hasFixedFirstColumn` is still being used, we will default to showing 1 fixed column. Additionally, if a number defined in `fixFirstColumns` exceeds or is equal to the amount of columns within the table, no fixed columns will be rendered. 

#### Before
data table before changes

#### After

**Without sticky header**

data table without sticky header

**With sticky header**

data table with sticky header

### How to 🎩

<details>

```jsx
import React from 'react';
import {Page, Card, DataTable} from '../src';

  return (
    const rows = [
    [
      'Emerald Silk Gown',
      'Formalwear',
      "Jill's formal",
      10,
      '$875.00',
      124689,
      140,
      '$426.00',
      '$122,500.00',
    ],
    [
      'Mauve Cashmere Scarf',
      'Accessories',
      "Jack's Accessories",
      253,
      '$230.00',
      124533,
      83,
      '$620.00',
      '$19,090.00',
    ],
    [
      'Ensembles',
      'Avocado Fashions',
      23,
      '$445.00',
      124518,
      32,
      '$353.00',
      '$14,240.00',
    ],
    [
      'Socks',
      'Essentials',
      'Avocado Fashions',
      465,
      '$4.00',
      124518,
      32,
      '$3.00',
      '$140.00',
    ],
  ];

  return (
      <Card>
        <DataTable
          columnContentTypes={[
            'text',
            'text',
            'text',
            'numeric',
            'numeric',
            'numeric',
            'numeric',
            'numeric',
            'numeric',
          ]}
          headings={[
            'Product',
            'Category',
            'Vendor',
            'Orders',
            'Price',
            'SKU Number',
            'Net quantity',
            'Shipping',
            'Net sales',
          ]}
          rows={rows}
          totals={['', '', '', '', '', '', 255, '$1399', '$155,830.00']}
          initialSortColumnIndex={4}
          stickyHeader
          fixedFirstColumns={3}
          truncate
          showTotalsInFooter
          footerContent={`Showing ${rows.length} of ${rows.length} results`}
        />
      </Card>
    </Page>
  );
}
```

</details>

### 🎩 checklist

### DIFF
--- a/.changeset/lazy-nails-relax.md
+++ b/.changeset/lazy-nails-relax.md
@@ -2,4 +2,5 @@
 '@shopify/polaris': minor
 ---
 
-Added `fixedFirstColumns` prop to `DataTable` so that multiple columns can be fixed
+- Added a `fixedFirstColumns` prop to `DataTable` so that multiple columns can be fixed
+- Deprecated the `DataTable` `fixedFirstColumn` prop
--- a/.changeset/lazy-nails-relax.md
+++ b/.changeset/lazy-nails-relax.md
@@ -1,0 +1,5 @@
+---
+'@shopify/polaris': minor
+---
+
+Added `fixedFirstColumns` prop to `DataTable` so that multiple columns can be fixed
--- a/polaris-react/src/components/DataTable/DataTable.stories.tsx
+++ b/polaris-react/src/components/DataTable/DataTable.stories.tsx
@@ -13,6 +13,48 @@ export default {
   },
 } as ComponentMeta<typeof DataTable>;
 
+function sortRows(rows, index, direction) {
+  return [...rows].sort((rowA, rowB) => {
+    let type;
+    if (rowA[index].length > 1) {
+      if (!isNaN(parseFloat(rowA[index]))) {
+        type = 'number';
+      } else if (!isNaN(parseFloat(rowA[index].substring(1)))) {
+        type = 'currency';
+      } else type = 'string';
+    } else {
+      type = !isNaN(parseFloat(rowA[index])) ? 'number' : 'string';
+    }
+
+    let itemA;
+    let itemB;
+    if (type === 'number') {
+      itemA = parseInt(rowA[index], 10);
+    } else if (type === 'currency') {
+      itemA = parseFloat(rowA[index].substring(1));
+    } else {
+      itemA = rowA[index];
+    }
+    if (type === 'number') {
+      itemB = parseInt(rowB[index], 10);
+    } else if (type === 'currency') {
+      itemB = parseFloat(rowA[index].substring(1));
+    } else {
+      itemB = rowA[index];
+    }
+    if (type === 'string') {
+      let result;
+
+      if (itemA > itemB) result = 1;
+      else if (itemB > itemA) result = -1;
+      else result = 0;
+      return direction === 'descending' ? result * -1 : result;
+    }
+
+    return direction === 'descending' ? itemB - itemA : itemA - itemB;
+  });
+}
+
 export function Default() {
   const rows = [
     ['Emerald Silk Gown', '$875.00', 124689, 140, '$122,500.00'],
@@ -53,7 +95,7 @@ export function Default() {
 }
 
 export function Sortable() {
-  const [sortedRows, setSortedRows] = useState(null);
+  const [sortedRows, setSortedRows] = useState<any[]>();
 
   const initiallySortedRows = [
     ['Emerald Silk Gown', '$875.00', 124689, 140, '$122,500.00'],
@@ -69,7 +111,7 @@ export function Sortable() {
   const rows = sortedRows ? sortedRows : initiallySortedRows;
 
   const handleSort = useCallback(
-    (index, direction) => setSortedRows(sortCurrency(rows, index, direction)),
+    (index, direction) => setSortedRows(sortRows(rows, index, direction)),
     [rows],
   );
 
@@ -101,15 +143,6 @@ export function Sortable() {
       </Card>
     </Page>
   );
-
-  function sortCurrency(rows, index, direction) {
-    return [...rows].sort((rowA, rowB) => {
-      const amountA = parseFloat(rowA[index].substring(1));
-      const amountB = parseFloat(rowB[index].substring(1));
-
-      return direction === 'descending' ? amountB - amountA : amountA - amountB;
-    });
-  }
 }
 
 export function WithFooter() {
@@ -300,7 +333,7 @@ export function WithRowHeadingLinks() {
 }
 
 export function WithAllOfItsElements() {
-  const [sortedRows, setSortedRows] = useState(null);
+  const [sortedRows, setSortedRows] = useState<any[]>();
 
   const initiallySortedRows = [
     [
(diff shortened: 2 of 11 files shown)


polaris-react/src/components/DataTable/DataTable.stories.tsx
```diff
@@ -13,6 +13,48 @@ export default {
   },
 } as ComponentMeta<typeof DataTable>;
 
+function sortRows(rows, index, direction) {
```
nickpresta: We also wanted to fix these examples since the sorting doesn't really work.

nickpresta:
I left comments on the stories that changed in Chromatic -- it would be great for someone else to take a look and verify my comments and accept the changes.

polaris-react/src/components/DataTable/components/Cell/Cell.tsx
```diff
@@ -29,24 +29,25 @@ export interface CellProps {
...
   fixedCellVisible?: boolean;
   firstColumnMinWidth?: string;
+  style?: React.CSSProperties;
 }
 
 export function Cell({
   content,
   contentType,
-  firstColumn,
+  nthColumn,
```
tylernoseworthy, line 42: Wondering if this should just be called `fixed` or `fixedColumn`

ryanpdaley (approved):
Looks good to me.

tylernoseworthy:
Just leaving a comment about a UI issue that potentially could happen if you pass a number of `fixedFirstColumns` that exceeds the table width. This causes there to be an overflow of the table which hides the scrollable table. I am thinking this is an edge case that should be handled by who ever uses the component, seeing as having 2+ fixed columns on certain screen sizes isn't the greatest experience.

Screen Shot 2022-08-23 at 9 42 30 AM

.changeset/lazy-nails-relax.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/polaris': minor
+---
+
+Added `fixedFirstColumns` prop to `DataTable` so that multiple columns can be fixed
```
chloerice:
```suggestion
- Added a `fixedFirstColumns` prop to `DataTable` so that multiple columns can be fixed
- Deprecated the `DataTable` `fixedFirstColumn` prop 
```

chloerice (changes requested):
Hey @tylernoseworthy 👋🏽 This works great on large screen but needs to be limited to 1 fixed column (or none at all) on small screens. Currently [the table breaks](https://5d559397bae39100201eedc1-sfqrphkjnx.chromatic.com/?path=/story/all-components-datatable--with-fixed-columns) on small screens:

Screen Shot 2022-08-23 at 11 53 10 AM

nickpresta:
@chloerice Thanks for the feedback! I'd like to understand a bit better -- is there a reason this needs to be baked into Polaris? The `fixedFirstColumns` prop is configurable, and the default state is to have _no_ fixed columns, which means by default things are fine on small screens, and any consumer opting into `fixedFirstColumns` behavior could handle the responsiveness themselves.

We don't seem to take a stance of overriding props when certain combinations wouldn't look great on small screens - for example [grid](https://polaris.shopify.com/components/grid), [button group](https://polaris.shopify.com/components/button-group), [index table](https://polaris.shopify.com/components/index-table) require the consumer to make an explicit choice about how to handle small screens.

For the data table, it would mean changing the `fixedFirstColumns` prop to something like `0` or `1` on small screens.

Thanks for any context!

chloerice:
> @chloerice Thanks for the feedback! I'd like to understand a bit better -- is there a reason this needs to be baked into Polaris? The `fixedFirstColumns` prop is configurable, and the default state is to have _no_ fixed columns, which means by default things are fine on small screens, and any consumer opting into `fixedFirstColumns` behavior could handle the responsiveness themselves.
> 
> We don't seem to take a stance of overriding props when certain combinations wouldn't look great on small screens - for example [grid](https://polaris.shopify.com/components/grid), [button group](https://polaris.shopify.com/components/button-group), [index table](https://polaris.shopify.com/components/index-table) require the consumer to make an explicit choice about how to handle small screens.
> 
> For the data table, it would mean changing the `fixedFirstColumns` prop to something like `0` or `1` on small screens.
> 
> Thanks for any context!

Hey @nickpresta! I think that approach of putting the onus on consumers is fine, as long the documentation for that prop calls it out and its examples show how to implement it.

polaris.shopify.com/pages/examples/data-table-with-fixed-first-columns.tsx
```diff
@@ -276,6 +277,8 @@ function DataTableWithFixedFirstColumnsExample() {
     ],
   ];
   const [sortedRows, setSortedRows] = useState(rows);
+  const showFixedColumns = useMedia('screen and (max-width: 850px)');
```
tylernoseworthy, line 280: I noticed this will not work when editing in codesandbox because of the `useMedia` hook. Is this something we should be worried about?

chloerice: Hey @tylernoseworthy you'll need to copy and paste the hook into the bottom of the example instead of importing it.

chloerice (approved):

polaris.shopify.com/pages/examples/data-table-with-fixed-first-columns.tsx
tylernoseworthy: Forgot to do it before merging 😅. I'll follow up with a another PR to fix the example for codesandbox.

.changeset/lazy-nails-relax.md
```diff
@@ -0,0 +1,6 @@
+---
+'@shopify/polaris': minor
```
sam-b-rose, line 2: I'm a bit late on this comment, but looking at these changes seems like this should be a `major` change since the current implementation is replacing the old component prop API.

If we want to keep it a minor, we should support both props, marking the old one as `@deprecated` and then do a followup PR to remove it in the next major.

@chloerice Should we revert for now so we aren't blocking the next Version Packages release?

tylernoseworthy: @samrose3 We added a [deprecation warning](https://github.com/Shopify/polaris/blob/53c0a981db5f6c75a4030be9972c13f294bba48a/polaris-react/src/components/DataTable/DataTable.tsx#L97) and still supported the use of the prop within our new implementation. Is there anywhere else within the documentation we need to mark this prop as deprecated?

.changeset/lazy-nails-relax.md
sam-b-rose: Looks like this caused a bit of a kerfuffle in the [generated changelog](https://github.com/Shopify/polaris/pull/6999). Can we reduce this to a single line without bullets? Not as fancy, but might improve how changesets renders it

Screen Shot 2022-08-26 at 6 29 08 AM

tylernoseworthy: @samrose3 Am I able to make that change directly on that branch? or is there a special command I have to run?

sam-b-rose: You can just modify the changeset file directly (`.changeset/lazy-nails-relax.md`) no command needed 👍

.changeset/lazy-nails-relax.md
sam-b-rose: I'm sorry, I missed this due to GitHub collapsing large file diffs 😅  disregard! All looks good to me 👍

PR #6992, opened 2022-08-22
Remove i18n test

kyledurand:
Do not merge 😅 

This is a test to see what our component library could look like if we removed the translations library.

**Things I'm curious about:**
* Bundle size
* What translations can be replaced by props and composition


alex-page:
Woah. There is a lot of translations that I didn't even realise that existed 🤯

PR #7007, opened 2022-08-24
Add initial color palette structure and next color tokens

sam-b-rose:


polaris-tokens/src/palettes.ts
```diff
@@ -0,0 +1,86 @@
+export const palettes = {
```
aaronccasanova, line 1: **thought:** Should these be individual exports? e.g. `export const gray = {...}`

sam-b-rose: We are wanting to only keep them "private" for now and only reference them through the semantic tokens. We may decide to open these up in the future, but safer to keep them protected for now.

aaronccasanova: Makes sense!

aaronccasanova:
@samrose3 Thanks for kicking this off! This PR was an awesome reference for the final implementation: https://github.com/Shopify/polaris/pull/8246

PR #7018, opened 2022-08-25
Generate all assets

alex-page:
- [x] `gen-assets` is decoupled from build
- [x] `gen-assets` creates `og-images` and `sitemap.xml`
- [x] `gen-assets` is ran in `polaris-site-container-builder` and optionally locally https://github.com/Shopify/polaris-site-container-builder/pull/18
- [x] No more `sitemap.xml` changing in PR's problems
- [x] Adds turborepo script so it can be ran at root by `polaris-site-container-builder`


sam-b-rose (approved):
Woohoo!! Thank you for fixing this! 🎉  I like this solution

But I was just now getting use to always removing `sitemap.xml` from my git changes post build 🙃

PR #7029, opened 2022-08-25
[Layout foundations] Inline component prototype

aveline:

Fixes https://github.com/Shopify/polaris/issues/6895

Introduces the `Inline` component which can be used to manage spacing between inline elements

Screen Shot 2022-08-25 at 2 45 31 PM

<details>
  <summary>Playground sample code</summary>

```
      <Text variant="bodySm" as="h3">
        default
      </Text>
      <Inline>
        <Text variant="heading4xl" as="h2">
          Inline
        </Text>
        <Badge>One</Badge>
        <Badge>Two</Badge>
        <Badge>Three</Badge>
        <Icon source={CapitalMajor} color="primary" />
      </Inline>
      <hr />
      <Text variant="bodySm" as="h3">
        alignY center
      </Text>
      <Inline alignY="center">
        <Text variant="heading4xl" as="h2">
          Inline
        </Text>
        <Badge>One</Badge>
        <Badge>Two</Badge>
        <Badge>Three</Badge>
        <Icon source={CapitalMajor} color="primary" />
      </Inline>
      <hr />
      <Text variant="bodySm" as="h3">
        alignY top
      </Text>
      <Inline alignY="top">
        <Text variant="heading4xl" as="h2">
          Inline
        </Text>
        <Badge>One</Badge>
        <Badge>Two</Badge>
        <Badge>Three</Badge>
        <Icon source={CapitalMajor} color="primary" />
      </Inline>
      <hr />
      <Text variant="bodySm" as="h3">
        alignY bottom
      </Text>
      <Inline alignY="bottom">
        <Text variant="heading4xl" as="h2">
          Inline
        </Text>
        <Badge>One</Badge>
        <Badge>Two</Badge>
        <Badge>Three</Badge>
        <Icon source={CapitalMajor} color="primary" />
      </Inline>
      <hr />
      <Text variant="bodySm" as="h3">
        alignY baseline
      </Text>
      <Inline alignY="baseline">
        <Text variant="heading4xl" as="h2">
          Inline
        </Text>
        <Badge>One</Badge>
        <Badge>Two</Badge>
        <Badge>Three</Badge>
        <Icon source={CapitalMajor} color="primary" />
      </Inline>
      <hr />
      <Text variant="bodySm" as="h3">
        align start
      </Text>
      <Inline align="start">
        <Text variant="heading4xl" as="h2">
          Inline
        </Text>
        <Badge>One</Badge>
        <Badge>Two</Badge>
        <Badge>Three</Badge>
        <Icon source={CapitalMajor} color="primary" />
      </Inline>
      <hr />
      <Text variant="bodySm" as="h3">
        align center
      </Text>
      <Inline align="center">
        <Text variant="heading4xl" as="h2">
          Inline
        </Text>
        <Badge>One</Badge>
        <Badge>Two</Badge>
        <Badge>Three</Badge>
        <Icon source={CapitalMajor} color="primary" />
      </Inline>
      <hr />
      <Text variant="bodySm" as="h3">
        align end
      </Text>
      <Inline align="end">
        <Text variant="heading4xl" as="h2">
          Inline
        </Text>
        <Badge>One</Badge>
        <Badge>Two</Badge>
        <Badge>Three</Badge>
        <Icon source={CapitalMajor} color="primary" />
      </Inline>
      <hr />
      <Text variant="bodySm" as="h3">
        align center alignY center
      </Text>
      <Inline align="center" alignY="center">
        <Text variant="heading4xl" as="h2">
          Inline
        </Text>
        <Badge>One</Badge>
        <Badge>Two</Badge>
        <Badge>Three</Badge>
        <Icon source={CapitalMajor} color="primary" />
      </Inline>
      <hr />
      <Text variant="bodySm" as="h3">
        spacing 0
      </Text>
      <Inline spacing="0">
        <Text variant="heading4xl" as="h2">
          Inline
        </Text>
        <Badge>One</Badge>
        <Badge>Two</Badge>
        <Badge>Three</Badge>
        <Icon source={CapitalMajor} color="primary" />
      </Inline>
      <hr />
      <Text variant="bodySm" as="h3">
        spacing 05
      </Text>
      <Inline spacing="05">
        <Text variant="heading4xl" as="h2">
          Inline
        </Text>
        <Badge>One</Badge>
        <Badge>Two</Badge>
        <Badge>Three</Badge>
        <Icon source={CapitalMajor} color="primary" />
      </Inline>
      <hr />
      <Text variant="bodySm" as="h3">
        spacing 4
      </Text>
      <Inline spacing="4">
        <Text variant="heading4xl" as="h2">
          Inline
        </Text>
        <Badge>One</Badge>
        <Badge>Two</Badge>
        <Badge>Three</Badge>
        <Icon source={CapitalMajor} color="primary" />
      </Inline>
      <hr />
      <Text variant="bodySm" as="h3">
        spacing 10
      </Text>
      <Inline spacing="10">
        <Text variant="heading4xl" as="h2">
          Inline
        </Text>
        <Badge>One</Badge>
        <Badge>Two</Badge>
        <Badge>Three</Badge>
        <Icon source={CapitalMajor} color="primary" />
      </Inline>
      <hr />
      <Text variant="bodySm" as="h3">
        spacing 32
      </Text>
      <Inline spacing="32">
        <Text variant="heading4xl" as="h2">
          Inline
        </Text>
        <Badge>One</Badge>
        <Badge>Two</Badge>
        <Badge>Three</Badge>
        <Icon source={CapitalMajor} color="primary" />
      </Inline>
```
  
</details>


polaris-react/src/components/Inline/Inline.scss
```diff
@@ -0,0 +1,117 @@
...
+.spacing025 {
+  --pc-stack-spacing: var(--p-space-025);
+}
+
+.spacing05 {
+  --pc-stack-spacing: var(--p-space-05);
+}
+
+.spacing1 {
+  --pc-stack-spacing: var(--p-space1);
```
chazdean: You'll want to change this
```suggestion
  --pc-stack-spacing: var(--p-space-1);
```

polaris-react/src/components/Inline/Inline.tsx
```diff
@@ -0,0 +1,56 @@
...
+  align,
+  alignY,
+  wrap,
+}: InlineProps) {
+  const className = classNames(
+    styles.Inline,
+    spacing && styles[variationName('spacing', spacing)],
+    align && styles[variationName('align', align)],
+    alignY && styles[variationName('alignY', alignY)],
+    wrap === false && styles.noWrap,
```
kyledurand: I think most of the css here can be replaced by setting the custom property to the wrapper i.e. 

`return <div className={styles.Inline} style={style}>{itemMarkup}</div>;`

where style would be similar to https://github.com/Shopify/polaris/blob/layout-foundations-prototype/polaris-react/src/components/Columns/Columns.tsx#L27-L38

Back when we had atoms it was fine but now we risk blowing up our bundle size with one off classes

.changeset/fresh-dingos-press.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/polaris': minor
+---
+
+Add alpha `Inline` component
```
laurkim:
```suggestion
Added alpha `Inline` component
```

polaris-react/src/components/Inline/Inline.scss
```diff
@@ -0,0 +1,113 @@
...
+}
+
+.alignYBaseline {
+  align-items: baseline;
+}
+
+.Item {
+  flex: 0 0 auto;
+  min-width: 0;
+  max-width: 100%;
```
kyledurand: I'm not sure we need the min-width / max-width flex box fixes these days. Can we omit them until we notice weirdness? Also `flex: 0 0 auto` feels like it might be a bit too prescriptive but I'm open to it if you feel it's better DX

kyledurand (commented):
This is looking good Aveline! My main concern is getting the total lines of css down as much as possible but overall I think this is looking great

laurkim (approved):
LGTM! 🚀 💯 

My only suggestion is to reduce the class names in the `Inline.scss` file by using `style={style}` as @kyledurand suggested 👍

chazdean (approved):
🚀!

polaris-react/src/components/Inline/Inline.tsx
aveline: Yeah I think it would be great to explore some kind of shared utility or atomic classes for this. I would almost rather leave this in as a reminder to do so. Setting the custom property inline seems less legible to me.

polaris-react/src/components/Inline/Inline.scss
aveline: Good call on all accounts. I think the `min-width` fix doesn't come into play unless there's some nested flex going on, so might not be necessary for this component. I'm good to remove unless we notice issues.

polaris-react/src/components/Inline/Inline.tsx
kyledurand: I'm curious how it's less legible. I've always found things like our classname variations hard to understand what's happening: `spacing && styles[variationName('spacing', spacing)]`

I've just done a [quick refactor](https://github.com/Shopify/polaris/pull/7174/files#diff-fa99a8ea2c4165b404905aedebdd75053975846cf7860bf30fda7a2150b7fd70R43-R48), it's very messy 😪 but I think it's nice and readable and brings the css down to ~7 lines.

Also GH copilot almost automatically brought the playground over into stories 🤯. All I did was keep pressing enter

kyledurand: In terms of getting into atomic classes. I'm very hesitant to do so, at least right now. We'd have to generate them from tokens and consume them in Polaris react. This means recreating a subset of vanilla extract functionality which is probably not easy. We'd also need some integration tests between the two  packages. 

What we're left with there is still more css and increased bundle, and documenting how to use these utility classes.

Something like `import spacingStyles from '@shopify/polaris-tokens'` 

But we can't directly match our spacing system to css because numbers so we'd have to prefix them, I think.

`const className = spacingStyles.space05`

I'm rambling but in my mind it seems like more hassle than it's worth than to use tokens directly

polaris-react/src/components/Inline/Inline.scss
```diff
@@ -0,0 +1,107 @@
+.Inline {
+  --pc-stack-spacing: var(--p-space-4);
```
kyledurand:
```suggestion
  --pc-inline-spacing: var(--p-space-4);
```

polaris-react/src/components/Inline/Inline.scss
```diff
@@ -0,0 +1,107 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  align-items: stretch;
+  gap: var(--pc-stack-spacing);
```
kyledurand:
```suggestion
  gap: var(--pc-inline-spacing);
```

kyledurand (approved):
Feel free to ship as is and iterate as you see fit 👍

polaris-react/src/components/Inline/Inline.tsx
aveline: Thanks for the refactor, I see the value in reducing the css file size. In terms of legibility, I meant adding all that inline in the dom. It isn't too bad with only a few custom properties, but using this pattern where there could be several more, it will start to look messy. Maybe it's just a personal preference. Regardless, I see your point that this is a larger issue that we shouldn't be trying to solve in the layout work. So I'm on board with going with this pattern for now.

PR #7038, opened 2022-08-26
[Avatar] Apply default background color for images

kyledurand:
Image avatars had the `styleOne` class applied by default giving them a jarring yellow background color. This change defaults to surface neutral to match our skeleton content.

Before:
image

After:
image


sarahill (approved):
Nice update! 👏 ✅

QuintonC (approved):
Looks good! Thanks for taking care of this 🙌

alex-page (approved):

